Issue #2277, opened 2023-12-06
[Tech Debt]: remove Stack and Stack Items from ToolbarGroup instances


jenny-s51:
### Type

General Tech Debt (eg. Improve the way something exists today)

### Description

cc @andrewballantyne @alexcreasy as discussed in our meeting today.

To follow PF guidelines, update toolbar label instances to sit inline with toolbar items, i.e. remove Stack where it is used within <ToolbarGroup> components

### Why?

#2268

dgutride:
Migrated to:  https://issues.redhat.com/browse/RHOAIENG-2403

PR #2609, opened 2024-03-18
Create `~/concepts/metrics` and `MetricsCommonContext` to begin reusing code between Model Metrics and Distributed Workloads

mturley:
Part of [RHOAIENG-2881](https://issues.redhat.com/browse/RHOAIENG-2881)

## Description

There is a decent amount of code written for Model Serving Metrics that is useful generically on any page that renders metrics charts. This PR creates the `~/concepts/metrics` directory to start collecting these abstractions and abstracts the `MetricsPageToolbar` component out of `~/pages/modelServing` along with its relevant types and utilities for use on the Distributed Workloads page.

* The original `MetricsPageToolbar` consumed the `ModelServingMetricsContext`, but it was only using state related to its dropdown menus which are useful outside of model serving (time range and refresh interval). This change creates a new `MetricsCommonContext` for this shared state, lifting it out of the `ModelServingMetricsProvider`. A `<MetricsCommonContextProvider>` is now wrapped around all instances of both `<ModelServingMetricsProvider>` and `<DistributedWorkloadsContextProvider>`, which both consume it. Any components used within either state can access the common state by consuming `MetricsCommonContext` directly. The new `MetricsPageToolbar` does this, which decouples it from the `ModelServingMetricsContext`.
* The state of the new Refresh Interval and Time Range dropdowns rendered by `MetricsPageToolbar` is now wired up to the parameters for data fetch hooks in `DistributedWorkloadsContextProvider` via consuming `MetricsCommonContext`.
* The `MetricsCommonContextProvider` exposes an `initialRefreshInterval` prop which defaults to 5 minutes, because this is the current default on the model metrics page. The Distributed Workloads page designs have the default at 30 minutes, so that value is passed only for Distributed Workloads.
* The `DistributedWorkloadsToolbar` component was redundant with `MetricsPageToolbar` and has been deleted. `MetricsPageToolbar` is now rendered below the Distributed Workloads page tabs.
* In addition to decoupling `MetricsPageToolbar` from model metrics, this change also factors out its two dropdowns as `MetricsTimeRangeSelect` and `MetricsRefreshIntervalSelect`, which are individual context consumers. This way it will be easy to move the Time Range selector out of the toolbar and next to a chart in the grid below [as designed in Distributed Workloads mockups](https://www.figma.com/proto/BKBS45DIWxGOhIYQbQqzYB/Distributed-Workloads?page-id=1%3A529&type=design&node-id=737-17667&viewport=326%2C658%2C0.13&t=LoGqmT1aihp9sM8q-1&scaling=contain&starting-point-node-id=405%3A9770), which will be a separate change.
* The Select component from `@patternfly/react-core/deprecated` is still in use here because we're moving existing code that uses it, and upgrading that to the new PF Select is out of scope. That will likely be a simple upgrade, but one goal here is not to change any current behavior of the model metrics pages.
* References to https://github.com/opendatahub-io/odh-dashboard/issues/2277 have been replaced with references to the equivalent Jira issue [RHOAIENG-2403](https://issues.redhat.com/browse/RHOAIENG-2403). Fixing this is currently out of scope of this PR because there are still ongoing design discussions about it ([see Slack thread here](https://redhat-internal.slack.com/archives/C05SMJ09DD2/p1710449134176569)).
* Types, constants and utilities in `~/concepts/metrics/[types|const|utils]` have been moved unchanged from `~/pages/modelServing/screens/[types|const]` and `~/pages/modelServing/screens/metrics/utils`.
* In a future PR, I plan to move the `~/pages/modelServing/screens/metrics/MetricsChart.tsx` component into `~/concepts/metrics` to use it in Distributed Workloads. This change does already update it to consume state from `MetricsCommonContext` instead of `ModelServingMetricsContext`, however.

Some files were moved and changed significantly enough that the diff considers it a delete/create. It may be useful to check the diffs of deleted files when reviewing these changes.

Note that the labels for Time Range and Refresh Interval are stacked above the dropdowns because that's how they are already implemented in model metrics. This may change depending on the ongoing discussion around [RHOAIENG-2403](https://issues.redhat.com/browse/RHOAIENG-2403) ([see Slack thread here](https://redhat-internal.slack.com/archives/C05SMJ09DD2/p1710449134176569)). Note also that the Project dropdown on the left side may be moving outside the tab container ([see another Slack thread](https://redhat-internal.slack.com/archives/C06J77CPB7S/p1710775221987839)). Addressing the inconsistency of label positions shown in this screenshot is out of scope for this PR.

Screenshot 2024-03-18 at 7 18 01 PM

FYI @xianli123

## How Has This Been Tested?

Tested by hand to see that the Time Range and Refresh Interval dropdowns behave as expected on both the Workload Metrics page and the model serving metrics page. Made sure there are no layout or functionality changes in model metrics.

## Test Impact

None of the code being moved had unit tests to begin with, and there are no significant new functions to test. Unit tests have been added for the utilities moved into `~/concepts/metrics/utils`. They are pretty trivial, but as we move more utilities over in the future we can add tests for them here. Otherwise, the changes here are covered by existing tests.

## Request review criteria:

If you have UI changes: 


gitdallas (approved):

lucferbux (approved):
/approve

PR #2612, opened 2024-03-19
Add ca-bundle.crt to add cluster bundles into the dashboard

lucferbux:
Closes https://issues.redhat.com/browse/RHOAIENG-4572

## Description

Fixes an issue that the dashboard is not getting the clobal cluster certificates when the ODH Operator Certificates functionality is not enabled.

## How Has This Been Tested?

1. Deploy this branch to a cluster
2. Test that you can create a pipeline server connection

## Test Impact

No testing converage added.

## Request review criteria:

- [ ] The developer has added tests or explained why testing cannot be added (unit or cypress tests for related changes

If you have UI changes: 

### DIFF
--- a/backend/src/server.ts
+++ b/backend/src/server.ts
@@ -39,10 +39,12 @@ app.listen({ port: PORT, host: IP }, (err) => {
     process.exit(1); // eslint-disable-line
   }
   // Load CA bundle used in our API calls
+  // tls-ca-bundle.pem is the default CA bundle used by the system in CentOS/RHEL
   // ca.crt is the default CA bundle provided by the service account for kubernetes
   // service-ca.crt is the CA bundle provided by the service account for kubernetes used by prometheus
   // odh-ca-bundle.crt and odh-trusted-ca-bundle.crt are the CA bundles provided by the ODH platform
   const caPaths = [
+    '/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem',
     '/var/run/secrets/kubernetes.io/serviceaccount/ca.crt',
     '/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt',
     '/etc/pki/tls/certs/odh-ca-bundle.crt',


lucferbux:
Checking the solution, we found that we have this other ca bundle called `ca-bundle.trust.crt` [see this link for more context](https://serverfault.com/questions/620003/difference-between-ca-bundle-crt-and-ca-bundle-trust-crt).

We have not included it yet since we don't know the implications, not sure if it's worth including it cc @LaVLaS @andrewballantyne @HumairAK what do you guys think?

lucferbux:
I need to test the image once it's built, I'll get back and do a lgtm myself once we are able to verify it

backend/src/server.ts
```diff
@@ -42,11 +42,13 @@ app.listen({ port: PORT, host: IP }, (err) => {
   // ca.crt is the default CA bundle provided by the service account for kubernetes
   // service-ca.crt is the CA bundle provided by the service account for kubernetes used by prometheus
   // odh-ca-bundle.crt and odh-trusted-ca-bundle.crt are the CA bundles provided by the ODH platform
+  // ca-bundle.crt is the default CA bundle used by the system
   const caPaths = [
     '/var/run/secrets/kubernetes.io/serviceaccount/ca.crt',
     '/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt',
     '/etc/pki/tls/certs/odh-ca-bundle.crt',
     '/etc/pki/tls/certs/odh-trusted-ca-bundle.crt',
+    '/etc/pki/tls/certs/ca-bundle.crt',
```
shalberd, line 51: are you using CentOS or RHEL? If yes, then the default system CA bundle location is
`/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem`

See
https://go.dev/src/crypto/x509/root_linux.go

@harshad16 at least in case of our standard notebook images, we always base default system trust on

https://github.com/opendatahub-io/kubeflow/pull/43/files#diff-1df80e242be90c94fce3ffb051a0bfe706fe924dab62438c7aa0a186b8067153R284

What do you think of this proposed change here?

shalberd: content of the configmap has two files still, right?
ca-bundle.crt:   being from central cluster config
odh-ca-bundle.crt:  being i.e. self-signed provided cert via DataScienceCluster

backend/src/server.ts
```diff
@@ -42,11 +42,13 @@ app.listen({ port: PORT, host: IP }, (err) => {
   // ca.crt is the default CA bundle provided by the service account for kubernetes
   // service-ca.crt is the CA bundle provided by the service account for kubernetes used by prometheus
   // odh-ca-bundle.crt and odh-trusted-ca-bundle.crt are the CA bundles provided by the ODH platform
+  // ca-bundle.crt is the default CA bundle used by the system
   const caPaths = [
     '/var/run/secrets/kubernetes.io/serviceaccount/ca.crt',
     '/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt',
     '/etc/pki/tls/certs/odh-ca-bundle.crt',
     '/etc/pki/tls/certs/odh-trusted-ca-bundle.crt',
```
shalberd, line 50: where does odh-trusted-ca-bundle.crt stem from?

backend/src/server.ts
lucferbux: Yes that's right @zdtsw we use ubi 8 as base image, symlink will point to the real path, which is the one you mentioned @shalberd, this is the standard we are going with alognside pipelines, cc @HumairAK

shalberd: is that symlink from `/etc/pki/tls/certs/ca-bundle.crt` to `/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem` there in all base images ... i.e. CentOS 9 Stream, UBI8?

When mounting in files via volume and volumeMount, the actual location has to be used, or can one mount to a symlink location? This more out of interest.

LaVLaS: This is a good discussion about which is the proper file reference as the default CA bundle.  I did a quick deep dive into CA files as the default location(s) and it appears that these locations are owned by the `ca-certificates` package on RPM based systems.

As @shalberd pointed out `/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem` is the current Red Hat default for generated certificates but `/etc/pki/tls/certs/ca-bundle.crt` is the historical default for Fedora/RHEL systems which is why the there is a symlink from `*/ca-bundle.crt --> */tls-ca-bundle.pem` in all of the generated images.

Either file reference should work as noted by the link to the `crypto/x509/root_linux.go` but if we are UBI8+ only we can safely use `/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem` as the default.  For any component that may need to support other distributions for workflows with custom user imagebuilds then it would be safer to rely on the `/etc/pki/tls/certs` as the universal default directory for openssl certificates

References:
* [Making CA certificates available to Linux command-line tools](https://www.redhat.com/sysadmin/ca-certificates-cli).
* cert-manager - [Managing public trust in kubernetes with trust-manager](https://cert-manager.io/docs/tutorials/getting-started-with-trust-manager/#configure-real-applications)

backend/src/server.ts
LaVLaS: The goal of providing ODH support for the cluster-wide and "self-provided" was to make them available as additional certificates for the application trust bundle.  I would be extremely reluctant to overwrite the default certificate bundle unless ODH was going to 100% own the management & lifecycle of the certificates trust bundle for each component.

Right now the initial compromise is we make both available in hopes that each component can provide full coverage for no self-signed certs (Default), application self-signed certs & proxy w/ self-signed certs.

backend/src/server.ts
shalberd: > For any component that may need to support other distributions for workflows with custom user imagebuilds then it would be safer to rely on the /etc/pki/tls/certs as the universal default directory for openssl certificates

Yes, at least for Guillome's notebook images in odh-contrib, those are based on CentOS Stream 9.
I have been successful relying on the location `/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem` in those.
Meaning it is working with the current odh-notebook-controller and proxy implementation. Granted, of course, we have out Openshift cluster admins added own private CAs to the openshift-config.

backend/src/server.ts
```diff
@@ -42,11 +42,13 @@ app.listen({ port: PORT, host: IP }, (err) => {
   // ca.crt is the default CA bundle provided by the service account for kubernetes
   // service-ca.crt is the CA bundle provided by the service account for kubernetes used by prometheus
   // odh-ca-bundle.crt and odh-trusted-ca-bundle.crt are the CA bundles provided by the ODH platform
+  // ca-bundle.crt is the default CA bundle used by the system
   const caPaths = [
     '/var/run/secrets/kubernetes.io/serviceaccount/ca.crt',
     '/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt',
     '/etc/pki/tls/certs/odh-ca-bundle.crt',
```
shalberd, line 49: @harshad16 @LaVLaS @VaishnaviHire in odh notebook controller in future, for setting the env variables value of SSL_CERT_FILE, REQUESTS_CA_BUNDLE and so on, we could do a similar approach as suggested by @LaVLaS.

i.e. check mounted-in new files from new configmap in the following order

`/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem`  (the system default of CentOS Stream 9, UBI8+, RHEL9 ..., will aways work and not be mucked up)
`/etc/pki/tls/certs/odh-trusted-ca-bundle.crt ` (here the one from the injection annotation configmap part. ie. ca-bundle.crt in the [original configmap](https://docs.openshift.com/container-platform/4.10/networking/configuring-a-custom-pki.html#certificate-injection-using-operators_configuring-a-custom-pki) that was configured by openshift cluster admins in line with the proxy config, what Landon calls proxy w/ self-signed certs or I call proxy with corporate CA non-public CAs. Will no be mucked up provided openshift admins do their job, which they do in our firm)
`/etc/pki/tls/certs/odh-ca-bundle.crt` the one from new configmap odh-ca-bundle.crt data section, i.e. self-signed certs given via [DataScienceClusterInitialization CR 
spec.trustedCABundle.customCAbundle](https://github.com/opendatahub-io/opendatahub-operator/pull/848#issue-2120714285) for dev purposes mostly

backend/src/server.ts
shalberd: yes, agreed, this is safer, it the sense that the linux default location is not overwritten. It just means a bit more work when i.e. setting default env vars related to CA trust in components, like I allude to for odh-notebook-controller.
I like the approach you took a lot, with having three different bundles used (linux system default location, application DSCI self-signed certs, central cluster config proxy w self-signed certs and private corporate CAs).

backend/src/server.ts
lucferbux: Ok, I've updated the PR, adding `'/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem'` as the default file and putting that at top of the toolchain. That's what I got from the conversation. Is that right? @LaVLaS @shalberd 
And thanks a lot for the great feedback here.

lucferbux:
Ok, I'm adding the lgtm and approve cause it's ready to go, tested and it's working

cc @dgutride @christianvogt @shalberd @LaVLaS

backend/src/server.ts
lucferbux: I'm gonna resolve this conversation to allow the PR to be merged, we can open a discussion on slack/a github issue if we wanna continue this topic.

Thanks again @shalberd for this input, it's really appreciated.

shalberd: very good, yes, what Landon wanted to make sure that the Linux default CA bundle location is on top of the list .... which it is now. Nice

PR #2613, opened 2024-03-19
proxy for access to mlmd service

christianvogt:
https://issues.redhat.com/browse/RHOAIENG-2975

## Description

Proxy enabled only when `DEV_MODE` is enabled.

Creates a proxy service at `/api/mlmd` which proxies to the cluster internal `ds-pipeline-metadata-envoy-dspa` `Service`.

Proxy URL: `/api/mlmd/<namespace>/<name>`
Since the service is per namespace, the frontend provides the namespace in the URL. Additionally, the dspa name is also provided at the same time as this is used first to confirm access to the DSPA. The name is then used with the prefix `ds-pipeline-metadata-envoy-` to come up with the `Service` name. Assumes port service `9090`.

`src/third_party/mlmd`: Sourced from https://github.com/kubeflow/pipelines/blob/master/frontend/src/third_party/mlmd/index.ts

## How Has This Been Tested?

Frontend:
Used custom test pages to test frontend APIs.

Backend:
Create a pipeline server in a project.
Postman:
- method: POST
- URL: `http://localhost:4000/api/mlmd/<namespace>/dspa/ml_metadata.MetadataStoreService/GetExecutionTypes`
- Add header `Content-Type: application/grpc-web+proto`
- Body (5 null characters): `�����` 

## Request review criteria:

If you have UI changes: 


backend/src/routes/api/k8s/pass-through.ts
```diff
@@ -9,7 +9,7 @@ import { proxyCall, ProxyError, ProxyErrorType } from '../../../utils/httpUtils'
 
 export type PassThroughData = {
   method: string;
-  requestData: string;
+  requestData?: string;
```
andrewballantyne, line 12: 🤔 intentional change? I guess for pure query params?

backend/src/types.ts
```diff
@@ -265,7 +265,7 @@ export type KubeFastifyInstance = FastifyInstance & {
 
 // TODO: constant-ize the x-forwarded header
 export type OauthFastifyRequest<Data extends RouteGenericInterface = RouteGenericInterface> =
-  FastifyRequest<{ Headers: { 'x-forwarded-access-token': string } & Data['Headers'] } & Data>;
+  FastifyRequest<{ Header?: { 'x-forwarded-access-token'?: string } & Data['Headers'] } & Data>;
```
andrewballantyne, line 268: Uh -- isn't it `Headers`? 😕 

That's fair that it's optional -- local development...

backend/src/routes/api/k8s/pass-through.ts
christianvogt: This aligns with the types for `proxyCall` it's optional. I have no data for a `GET` request.

backend/src/routes/api/mlmd/index.ts
```diff
@@ -0,0 +1,78 @@
...
+      const kc = fastify.kube.config;
+      const cluster = kc.getCurrentCluster();
+
+      // see `prefix` for named params
+      const namespace = getParam(request, 'namespace');
+      const name = getParam(request, 'name');
+
+      // retreive the metadata envoy service
+      passThroughResource<V1Service>(fastify, request, {
+        url: `${cluster.server}/api/v1/namespaces/${namespace}/services/ds-pipeline-metadata-envoy-${name}`,
```
alexcreasy, line 41: You're potentially introducing an SSRF vulnerability here as you're directly building the URL from data provided by the client, without first validating it.

christianvogt: This call goes to the k8s API server only based on `cluster.server` which is not user input. The API server the handles whether the user has access to the resource.

The `params` are parsed by `http-proxy` matching a specific segment of the incoming URL. I am assuming that the `http-proxy` parser here does not allow inclusion of `/` within the param value but I didn't explicitly test that. It shouldn't though, otherwise that would defeat the purpose of the prefix path parser.

@alexcreasy Is there something else specific that you're referring to?

Gkrumbach07:
From a "it works as expected" side of things, it looks good. I was a bale to use the frontend util to fetch executions and so on. I imagine we might want to include a wrapper around this when we use it for things, like how we are accessing the KF api, but probably on a simpler scale.

I did not dig into the security side of things but it looks like you addressed this in the pr comments. So from my side this is good to go. I will let you give the final a-ok to merge.

/lgtm

christianvogt:
After further discussions, 2 follow up changes have been made:
- Assume the port without having to first fetch the `Service` to perform a lookup.
- Replace fetching `Service` with fetching the `DSPA`. Use the incoming name to `GET` the `DSPA` to confirm the name is valid before continuing.

jwforres:
We need to make sure any pod to pod communication is happening over TLS as well.  I believe the current state of this PR has it communicating over non TLS connection

christianvogt:
Correct. There doesn't seem to be any service certificate annotation: https://github.com/red-hat-data-services/data-science-pipelines-operator/blob/main/config/internal/ml-metadata/metadata-envoy.service.yaml.tmpl

christianvogt:
Last change enables the proxy only when `DEV_MODE` is enabled.

Gkrumbach07:
The frontend is throwing an error when trying to initialize the client for mlmd. weirdly it is saying
`_third_party_mlmd__WEBPACK_IMPORTED_MODULE_10__.MetadataStoreServicePromiseClient is not a constructor`
Which is odd becuase the types for this indicate it has a a constructor.

Gkrumbach07:
Ran some tests on cluster and can confirm the route is not accessible. We will need to make sure we try catch all the calls because the service doesnt seem to have a am available functionality

christianvogt:
I could squash both commits. However I created the 2 commits intentionally because of the logical separation between the third party lib (first commit) and then the integration (2nd commit).

andrewballantyne:
I think we can go ahead with this if everyone is happy with it. It's behind the dev flag right? So we should be able to merge and start iterating in the dev cycles on it.

andrewballantyne:
/cc @Gkrumbach07 

I think the approval is on you now that you're an approver (might need to add it manually given the history of the PR)

Gkrumbach07:
oh my lgtm never was added

/lgtm

PR #2616, opened 2024-03-20
Add wrap text option for pipeline run logs

DaoDaoNoCode:
JIRA: [RHOAIENG-3830](https://issues.redhat.com/browse/RHOAIENG-3830)

## Description

There is some problem with the PF log viewer, which is causing the incorrect rendering for the extremely long log lines. In the OpenShift console, the solution is to truncate the line. However, it's limited currently in the dashboard for now because we haven't applied websocket on the logs, so it's hard to truncate a single line of the logs (it's a full string now). After a discussion with the UX, we decided to add an option to allow the users to check whether they want to wrap the text in the log viewer or not. In this way, they can view the incorrectly rendered logs by unchecking the `Wrap text` checkbox in the toolbar.

Screenshot 2024-03-20 at 11 17 46 AM

Also, update the text in the alert because we don't truncate the exceptionally long lines :)

Screenshot 2024-03-20 at 11 18 51 AM

## How Has This Been Tested?

1. Create a pipeline run
2. Check the logs in the steps, try to wrap and unwrap it
3. Make sure nothing is broken
4. Check the text in the alert

## Test Impact

N/A, it's a simple update with the built-in log viewer interface.

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/concepts/dashboard/DashboardLogViewer.tsx
+++ b/frontend/src/concepts/dashboard/DashboardLogViewer.tsx
@@ -8,7 +8,8 @@ const DashboardLogViewer: React.FC<{
   footer: JSX.Element | false;
   onScroll: React.ComponentProps<typeof LogViewer>['onScroll'];
   height?: number | string;
-}> = ({ data, logViewerRef, toolbar, footer, onScroll, height = '100%' }) => (
+  isTextWrapped?: boolean;
+}> = ({ data, logViewerRef, toolbar, footer, onScroll, height = '100%', isTextWrapped }) => (
   <LogViewer
     data-testid="logs"
     hasLineNumbers={false}
@@ -18,6 +19,7 @@ const DashboardLogViewer: React.FC<{
     toolbar={toolbar}
     footer={footer}
     onScroll={onScroll}
+    isTextWrapped={isTextWrapped}
   />
 );
 
--- a/frontend/src/concepts/pipelines/content/pipelinesDetails/pipelineRun/runLogs/LogsTab.tsx
+++ b/frontend/src/concepts/pipelines/content/pipelinesDetails/pipelineRun/runLogs/LogsTab.tsx
@@ -1,6 +1,7 @@
 import React, { useEffect } from 'react';
 import {
   Button,
+  Checkbox,
   DropdownList,
   Icon,
   Spinner,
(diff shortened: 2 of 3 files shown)


DaoDaoNoCode:
@yih-wang Could you check the screenshot to see if it's accurate?

yih-wang:
@DaoDaoNoCode Can you give me a screenshot of how it looks like in the minSize of the drawer? I think it's fine to have the 'wrap text' and 'download' both outside the kebab. But I want to confirm that it doesn't cause any wrapping issues when the drawer is minSize.

DaoDaoNoCode:
@yih-wang Sure, this is the min size, I cannot shrink it anymore.

Screenshot 2024-03-20 at 10 36 07 PM

yih-wang:
@DaoDaoNoCode Any mess if you start searching and input a single character in the search box?

DaoDaoNoCode:
@yih-wang Ah yes, it gets a little bit weird.
Screenshot 2024-03-21 at 8 02 56 AM

yih-wang:
so that's what I meant. We need to extend the minSize of the drawer, or hide one of the 'wrap text' and 'download' into the kebab to make it neat.

yih-wang:
That said, this kind of wrap seems acceptable to me... I'm not sure how many users would see this wrapping layout. If only 10% of them, I think we can keep it as is since keep both the wrap text and download in the toolbar would bring most convinience.

DaoDaoNoCode:
@yih-wang I can extend the min size a little bit. The following screenshot extends it for about 50px, do you think this is acceptable? If so I will go with this solution.
Screenshot 2024-03-21 at 8 25 27 AM

yih-wang:
can you try to type a character in the search box? Because it will add the length of the search box.

DaoDaoNoCode:
@yih-wang I see, it breaks the layout again, so I added 20px more to it, then it's OK now as the screenshot shows.

Screenshot 2024-03-21 at 8 56 37 AM

yih-wang:
I would say add at least 50px more... Since we will also have a status icon when the log is still running. And if the number (1/256) in the search box is longer, it would even take more space.

yih-wang:
And I'm not sure if it's too large for a minSize after adding another 50px... If so, let's go back to what you originally have (with wrapped 'wrap text') since it looks not so bad.
image

DaoDaoNoCode:
@yih-wang How does this look like? It's 100px more compared to the very original one. If not enough I can make it to 120px more.

Screenshot 2024-03-21 at 9 01 26 AM

DaoDaoNoCode:
Yeah, maybe I won't change anything about the min size if you think it's not a big deal.

yih-wang:
Sounds good to me. Let's stick to your original version.

jpuzz0 (approved):

Gkrumbach07 (approved):
/lgtm
/approve

PR #2617, opened 2024-03-20
Fix indentation for labels in serving manifests

Xaenalt:
Small nit noticed in https://issues.redhat.com/browse/RHOAIENG-4547

### DIFF
--- a/manifests/modelserving/ovms-kserve-ootd.yaml
+++ b/manifests/modelserving/ovms-kserve-ootd.yaml
@@ -13,13 +13,13 @@ metadata:
 objects:
   - apiVersion: serving.kserve.io/v1alpha1
     kind: ServingRuntime
-    labels:
-      opendatahub.io/dashboard: 'true'
     metadata:
       annotations:
         openshift.io/display-name: OpenVINO Model Server
         opendatahub.io/recommended-accelerators: '["nvidia.com/gpu"]'
       name: kserve-ovms
+      labels:
+        opendatahub.io/dashboard: 'true'
     spec:
       multiModel: false
       annotations:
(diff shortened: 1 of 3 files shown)


jiridanek:
> Small nit noticed in https://issues.redhat.com/browse/RHOAIENG-4547

If forces me to do

```java
            // WORKAROUND(RHOAIENG-4547) ServingRuntime should not have top level `labels` key
            genericKubernetesResource.getAdditionalProperties().remove("labels");
            return castResource(it, ServingRuntime.class);
```

https://github.com/skodjob/odh-e2e/blob/46c2482fc88daf5ad1d72e5956cbf7b34f5415df/src/test/java/io/odh/test/e2e/standard/ModelServingST.java#L237-L239

and I guess it either means the label is missing where it should be after the template is instantiated, or how do you call it, or that something has to be adding that label. In case of which maybe the label does not have to be specified in the template in the first place.

manifests/modelserving/ovms-kserve-ootd.yaml
```diff
@@ -20,6 +20,8 @@ objects:
         openshift.io/display-name: OpenVINO Model Server
         opendatahub.io/recommended-accelerators: '["nvidia.com/gpu"]'
       name: kserve-ovms
+      labels:
+        opendatahub.io/dashboard: 'true'
```
jiridanek, line 24: please remove the `labels:` on lines 16,17

jiridanek (changes requested):

andrewballantyne (approved):
Looks good, didn't test this on a cluster but looks good on paper.

PR #2620, opened 2024-03-20
[WIP] Home page demo

jeff-phillips-18:
Demo purposes only


andrewballantyne:
@jeff-phillips-18 I think we can close this -- yes? Or are you using this to build images for the cluster?

jeff-phillips-18:
Yeah, I use this to deploy an image with everything enabled so designers have a playground.

jeff-phillips-18:
No more need for this.

PR #2610, opened 2024-03-18
feat(DW): add workloads table

gitdallas:
Closes: [RHOAIENG-3583](https://issues.redhat.com//browse/RHOAIENG-3583)

## Description
Implement first iteration of workloads table. More information is necessary before including links from the table. After selecting a project with workloads, go to the workload metrics -> workload status tab.
image

## Test Impact
Cypress tests to see that the table is there with a workload name and with the placeholder text if no workload. Unit test on a new utility function

Self checklist (all need to be checked):
- [x] The developer has manually tested the changes and verified that the changes work
- [x] Commits have been squashed into descriptive, self-contained units of work (e.g. 'WIP' and 'Implements feedback' style messages have been removed)
- [x] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [x] The developer has added tests or explained why testing cannot be added (unit or cypress tests for related changes)

If you have UI changes: 
- [x] Included any necessary screenshots or gifs if it was a UI change.
- [ ] Included tags to the UX team if it was a UI/UX change (find relevant UX in the [SMEs](https://github.com/opendatahub-io/odh-dashboard/tree/main/docs/smes.md) section).

After the PR is posted & before it merges:
- [ ] The developer has tested their solution on a cluster by using the image produced by the PR to `main`

PR #2623, opened 2024-03-21
Implement Status overview donut chart on DW Workload Status page, temporarily remove Status trends chart

mturley:
Closes [RHOAIENG-2881](https://issues.redhat.com/browse/RHOAIENG-2881)

## Description

* Replaces the old Prometheus-based `workloadCurrentMetrics` with aggregate status counts based on the conditions of Workload CRs. A new `getStatusCounts` helper uses the `getStatusInfo` helper introduced by https://github.com/opendatahub-io/odh-dashboard/pull/2610 for this.
* Temporarily removes the "Status trends" card from the page since it has been descoped from GA due to RBAC issues.
   * Leaves the unused `workloadTrendMetrics` fetching code in place for now because it will be reused when the RBAC issues are resolved. A comment to this effect has been added.

cc @xianli123 

Screenshot 2024-03-21 at 12 10 30 PM

## How Has This Been Tested?

Tested that the donut chart and card layout look correct and accurately represent the data.

## Test Impact

Unit test added for the new `getStatusCounts` helper.

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/pages/distributedWorkloads/global/workloadStatus/DWStatusOverviewDonutChart.tsx
+++ b/frontend/src/pages/distributedWorkloads/global/workloadStatus/DWStatusOverviewDonutChart.tsx
@@ -1,23 +1,31 @@
 import * as React from 'react';
 import { Card, CardTitle, CardBody, Bullseye, Spinner } from '@patternfly/react-core';
+import { ChartDonut, ChartThemeColor } from '@patternfly/react-charts';
 import EmptyStateErrorMessage from '~/components/EmptyStateErrorMessage';
 import { DistributedWorkloadsContext } from '~/concepts/distributedWorkloads/DistributedWorkloadsContext';
+import {
+  WorkloadStatusColorAndIcon,
+  WorkloadStatusType,
+  getStatusCounts,
+} from '~/concepts/distributedWorkloads/utils';
 
 export const DWStatusOverviewDonutChart: React.FC = () => {
-  const { workloadCurrentMetrics } = React.useContext(DistributedWorkloadsContext);
+  const { workloads } = React.useContext(DistributedWorkloadsContext);
 
-  if (workloadCurrentMetrics.error) {
+  const statusCounts = React.useMemo(() => getStatusCounts(workloads.data), [workloads.data]);
+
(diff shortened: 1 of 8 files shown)


gitdallas (approved):

lucferbux:
@mturley can you clean up the commits please?

frontend/src/pages/distributedWorkloads/global/workloadStatus/DWStatusOverviewDonutChart.tsx
```diff
@@ -1,23 +1,31 @@
...
+  WorkloadStatusType,
+  getStatusCounts,
+} from '~/concepts/distributedWorkloads/utils';
 
 export const DWStatusOverviewDonutChart: React.FC = () => {
-  const { workloadCurrentMetrics } = React.useContext(DistributedWorkloadsContext);
+  const { workloads } = React.useContext(DistributedWorkloadsContext);
 
-  if (workloadCurrentMetrics.error) {
+  const statusCounts = React.useMemo(() => getStatusCounts(workloads.data), [workloads.data]);
```
lucferbux, line 15: FYI, just adding the thing we discussed on the peer review. We are memoizing here a k8s object, bear in mind that state changes on complex objects like this wont bring anything relevant into the table, since pretty much all the refecth will get a different reference, maybe we'll wanna revisit this for a performance standpoint in the future

mturley: (Just repeating my thoughts from the meeting as well so I remember in the future 😆)
Yeah, that's a fair point. It's probably an overoptimization to begin with because `getStatusCounts` isn't that heavy a calculation, I just wanted to prevent it from getting re-run even when the data hasn't been refetched (if the tree re-renders for a different reason). The way to memoize it correctly though would be to only recalculate it if the `status.conditions` array on any workload has changed, and checking that via `useDeepCompareMemoize` would be as much (or more) computation than just re-running `getStatusCounts` I think

lucferbux (approved):
/lgtm

PR #2632, opened 2024-03-25
DW Workload Status: Add Admitted and Evicted statuses, improve logic for resolving status, add better tests

mturley:
Closes [RHOAIENG-4717](https://issues.redhat.com/browse/RHOAIENG-4717)

## Description

Based on discussion with @astefanutti ([see slack thread here](https://redhat-internal.slack.com/archives/C06J77CPB7S/p1710925990577769?thread_ts=1710359684.868669&cid=C06J77CPB7S)), this PR corrects our logic for deriving the discrete status of a workload from its conditions, and our rendering of those statuses in the UI.

* There are 2 new statuses:
  * Admitted - a state after Pending where the workload has its resources (QuotaReserved) but is not yet running (PodsReady).
  * Evicted - a workload is removed from the queue after it starts but before it completes.
* The Unknown status will now only be shown in the status overview chart and legend if there is at least one workload with unknown status. This is because Unknown is a fail-safe that we should never hit - If the user sees Unknown, we have a bug in `getStatusInfo` and we show that rather than have the chart not add up to the total workloads.
* To accommodate the additional statuses, the legend of the status overview chart now appears in two columns.

Screenshot 2024-03-25 at 7 15 16 PM

## How Has This Been Tested?

Tested by hand that I didn't break the existing behavior (viewing existing workloads and running new ones). Added unit tests and cypress tests.

## Test Impact

* Added new Cypress tests to check that the status overview chart is rendering properly both with and without the Unknown segment.
* Added new unit tests to make sure the `getStatusInfo` and `getStatusCounts` util functions are correctly handling all possible workload statuses.
* Enhanced `mockWorkloadK8sResource` with a `mockStatus` argument so it can simulate workloads of different statuses. This mock file contains example conditions for each status which are the basis of the assertions in the new tests.

## Request review criteria:

If you have UI changes: 


mturley:
Marking WIP for now just to make sure I confirm this new logic with @astefanutti before we merge this. I don't anticipate any further changes if my interpretation of his explanation on Slack is correct.

frontend/src/pages/distributedWorkloads/global/workloadStatus/DWStatusOverviewDonutChart.tsx
```diff
@@ -13,6 +13,7 @@ export const DWStatusOverviewDonutChart: React.FC = () => {
   const { workloads } = React.useContext(DistributedWorkloadsContext);
 
   const statusCounts = React.useMemo(() => getStatusCounts(workloads.data), [workloads.data]);
+  const statusTypes = Object.keys(statusCounts) as WorkloadStatusType[];
```
andrewballantyne, line 16: Why not an enum for WorkloadStatus? You seem to be wanting to do that a lot...

mturley: Previously I had an `Unknown` status that was only conditionally included in `statusCounts` and wanted to hide it from the chart if it wasn't there, but that has been replaced by falling back to `Pending` so maybe it makes sense to have an enum now. Hmm.

mturley: ahh -- part of the reason for using Object.keys (which returns `string[]` and needs this assertion) is that in `utils.tsx`, the order of keys in `knownStatusConditions` matters because we want the first matching condition found there, and if we just did `Object.values(Enum)` we'd be making the order of the enum values affect behavior elsewhere which feels fragile. But maybe we need to explicitly just make a `statusConditionsInOrder` array to go through instead of relying on the key order there.

mturley: Rebased and updated to use an enum here, `Object.values()` around it with no type assertion where we were using `Object.keys` before with a type assertion, and use a `statusesInEvalOrder` to be explicit about that behavior and not rely on the order of keys in the `knownStatusConditions` object.

Srihari1192:
@mturley   Came across below observations while testing these changes , not sure its already a known issue.

- Hovering on donut chart Workload status for more than 2 seconds is showing multiple hover info text as `Status overview donut chart` 


Screenshot 2024-03-27 at 7 04 50 PM



-  Hovering on empty states in donut charts is showing as non existing status of the workload with zero count.

Screenshot 2024-03-27 at 7 05 55 PM


Screenshot 2024-03-27 at 7 09 27 PM

mturley:
@Srihari1192 the second point should be fixed now.

Srihari1192:
@mturley Verified with the new changes. looks good to me. Thanks

gitdallas (approved):

andrewballantyne (approved):

PR #2636, opened 2024-03-26
Add compare runs functionality

Gkrumbach07:
closes: https://issues.redhat.com/browse/RHOAIENG-4657

## Description

- adds a context for compare runs page that handles selections and runs in the array
- added routing to extend the context into both the index and /add pages
- added data fetching from url and handling error states
Screenshot 2024-03-26 at 10 31 11 AM
Screenshot 2024-03-27 at 4 27 49 PM
Screenshot 2024-03-27 at 4 27 17 PM
Screenshot 2024-03-27 at 4 26 45 PM

## How Has This Been Tested?

1. navigate to `/experiments/{namespace}/{experimentId}/compareRuns`
2. should show an error state
3. navigate to `/experiments/{namespace}/{experimentId}/compareRuns?runs={run_id}`
4. should not show and error state
5. navigate to `/experiments/{namespace}/{experimentId}/compareRuns?runs=not a run id`
6. should remove the run from the url
7. should show error state

## Test Impact

tests added for error states

## Request review criteria:

If you have UI changes: 

@yannnz @xianli123 

### DIFF
--- a/frontend/src/concepts/pipelines/apiHooks/usePipelineRunById.ts
+++ b/frontend/src/concepts/pipelines/apiHooks/usePipelineRunById.ts
@@ -9,7 +9,7 @@ import { PipelineRunKFv2, RuntimeStateKF, runtimeStateLabels } from '~/concepts/
 import { FAST_POLL_INTERVAL } from '~/utilities/const';
 import { computeRunStatus } from '~/concepts/pipelines/content/utils';
 
-const usePipelineById = (
+const usePipelineRunById = (
   pipelineRunId?: string,
   refreshForDetails?: boolean,
 ): FetchState<PipelineRunKFv2 | null> => {
@@ -48,4 +48,4 @@ const usePipelineById = (
   return runData;
 };
 
-export default usePipelineById;
+export default usePipelineRunById;
--- a/frontend/src/concepts/pipelines/content/compareRuns/CompareRunsContext.tsx
+++ b/frontend/src/concepts/pipelines/content/compareRuns/CompareRunsContext.tsx
@@ -1,0 +1,97 @@
+import * as React from 'react';
+import { useSearchParams } from 'react-router-dom';
+import { conditionalArea, SupportedArea } from '~/concepts/areas';
+import { PipelineRunKFv2 } from '~/concepts/pipelines/kfTypes';
+import { CompareRunsSearchParam } from '~/concepts/pipelines/content/types';
+import useNotification from '~/utilities/useNotification';
+import { allSettledPromises } from '~/utilities/allSettledPromises';
+import useFetchState, { FetchStateCallbackPromise } from '~/utilities/useFetchState';
+import { usePipelinesAPI } from '~/concepts/pipelines/context';
+
+type CompareRunsContextType = {
+  runs: PipelineRunKFv2[];
+  selectedRuns: PipelineRunKFv2[];
+  setRuns: (runs: PipelineRunKFv2[]) => void;
+  setSelectedRuns: React.Dispatch<React.SetStateAction<PipelineRunKFv2[]>>;
+  loaded: boolean;
+};
+
+const CompareRunsContext = React.createContext<CompareRunsContextType>({
+  runs: [],
+  loaded: false,
+  setRuns: () => undefined,
+  selectedRuns: [],
+  setSelectedRuns: () => undefined,
+});
+
+type CompareRunsContextProviderProps = {
+  children: React.ReactNode;
+};
+
+export const CompareRunsContextProvider = conditionalArea<CompareRunsContextProviderProps>(
+  SupportedArea.DS_PIPELINES,
+  true,
+)(({ children }) => {
+  // get run ids from url
+  const [searchParams, setSearchParams] = useSearchParams();
+  const runIdsArray = React.useMemo(() => {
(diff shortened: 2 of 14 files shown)


frontend/src/concepts/pipelines/content/tables/pipelineRun/PipelineRunTable.tsx
```diff
@@ -101,6 +102,54 @@ const PipelineRunTable: React.FC<PipelineRunTableProps> = ({
...
+              },
+            ]),
+      ]}
+    />
+  );
+
+  const actions = [primaryToolbarAction, toolbarDropdownAction];
+  if (compareRunsAction) {
+    actions.push(compareRunsAction);
+  }
```
jpuzz0, line 151: Can this not be conditionally spread into the array of actions instead?

frontend/src/routes/pipelines/experiments.ts
```diff
@@ -64,3 +64,19 @@ export const experimentScheduleDetailsRoute = (
   !experimentId || !recurringRunId
     ? experimentsBaseRoute(namespace)
     : `${experimentSchedulesRoute(namespace, experimentId)}/${recurringRunId}`;
+
+const compareRunsSelect = (runIds: string[]) => `?compare=${runIds.join(',')}`;
```
jpuzz0, line 68: nit: I'm not sure if the name `compareRunsSelect` for this constant describes the value very well.

jpuzz0:
The placement of the compare button is meant to be to the immediate right of the primary action according to the design.

Actual
image

Design
image

Also, shouldn't the button be disabled unless you've selected at least 2 runs? That seems to be how it works in KF UI.

jpuzz0:
Not sure if we want to handle this in this story, but there is a scenario where just playing with URL params, anything following `compareRuns` that isn't the `runs` array of IDs leads to a blank page: 
image

frontend/src/concepts/pipelines/content/compareRuns/CompareRunsContext.tsx
```diff
@@ -0,0 +1,94 @@
...
+  React.useEffect(() => {
+    if (runs.length > 0) {
+      // 5 = Not found https://github.com/googleapis/googleapis/blob/master/google/rpc/code.proto#L78
+      const validRuns = runs.filter(({ error: runError }) => runError?.code !== 5);
+
+      // if there exists invalid runs, remove them from the url and notify the user
+      if (validRuns.length !== runIdsArray.length) {
+        searchParams.set(
+          CompareRunsSearchParam.RUNS,
+          validRuns.map(({ run_id: runId }) => runId).join(','),
```
jpuzz0, line 57: Based on the comment above, should this not be a `filter` instead of a `map`? To filter out run IDs with no associated run?

frontend/src/concepts/pipelines/apiHooks/usePipelineRunById.ts
```diff
@@ -8,8 +8,36 @@ import { usePipelinesAPI } from '~/concepts/pipelines/context';
...
+  const call = React.useCallback<FetchStateCallbackPromise<PipelineRunKFv2[]>>(
+    (opts) =>
+      allSettledPromises(pipelineRunIds.map((id) => api.getPipelineRun(opts, id))).then(
+        ([successful]) => successful.map(({ value }) => value),
+      ),
+    [api, pipelineRunIds],
+  );
+
+  const [data, loaded, , refresh] = useFetchState(call, []);
+  return [data, loaded, refresh];
```
jpuzz0, line 37: Is there a reason we wouldn't return an error if one exists?

Gkrumbach07: allSettledPromises wont throw an error, it splits calls into successful responses and error responses. And in our use case we still want the data to be returned despite there being an error on some of the calls (becuase we trim out the ids that returned an error from the url). I went many different directions with this. and becuase of how useFetchState works, having an error means that there cant also be data loaded in.

jpuzz0:
I think we agreed yesterday post-standup that the list pages would require at least 2 selections for `Compare runs` to be enabled for use. I'm still seeing only 1 needing to be selected.

frontend/src/concepts/pipelines/apiHooks/usePipelineRunById.ts
jpuzz0: If an error exists, then I would expect no data. So I'm not sure what the issue is there as you're describing how useFetchState works. 

I'm not seeing why we can't also use error responses from allSettledPromises.

jpuzz0: The function seems tailored then to this 1 use-case but yet has a name that implies it would behave just like any of our other API call hooks.

Gkrumbach07: In some fashion the hook needs to give back data despite there being an error, if not then there would be no way to prune the bad runs. I can move this hook into the compare runs directory and rename it to match this usecase.

Gkrumbach07: i ended up just pruning it down and putting it in the context provider component becuase no one else would use it

jpuzz0 (approved):
/approve

PR #2644, opened 2024-03-28
feat(dw): add empty state and placeholders

gitdallas:
Closes: #https://issues.redhat.com/browse/RHOAIENG-4746

## Description
Add empty state, error state, loading state, and placeholders where we don't have the data/charts implemented yet

Showing multiple states (manually manipulated code to do so):
image

With a project that has no workloads:
image

A project with workloads:
image

No quota:
image

no project (already existed, will show selector when merged with the "move project selector" PR)
image

## How Has This Been Tested?
tested manually loading data with and without workload data

## Test Impact
added tests to make sure the various empty states are showing up

## Request review criteria:

Self checklist (all need to be checked):
- [ x] The developer has manually tested the changes and verified that the changes work
- [ x] Commits have been squashed into descriptive, self-contained units of work (e.g. 'WIP' and 'Implements feedback' style messages have been removed)
- [ x] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [ x] The developer has added tests or explained why testing cannot be added (unit or cypress tests for related changes)

If you have UI changes: 

- [ x] Included any necessary screenshots or gifs if it was a UI change.
- [ ] Included tags to the UX team if it was a UI/UX change (find relevant UX in the [SMEs](https://github.com/opendatahub-io/odh-dashboard/tree/main/docs/smes.md) section).

After the PR is posted & before it merges:
- [ ] The developer has tested their solution on a cluster by using the image produced by the PR to `main`

PR #2650, opened 2024-03-29
DW: Fetch LocalQueues, pass them down in context, use them for "quota is not set" empty state

mturley:
Closes [RHOAIENG-4746](https://issues.redhat.com/browse/RHOAIENG-4746) (followup to #2644, cc @gitdallas)
Needed for [RHOAIENG-2844](https://issues.redhat.com/browse/RHOAIENG-2844)

## Description

* Adds types, utilities, hooks and tests for fetching Kueue [LocalQueues](https://kueue.sigs.k8s.io/docs/reference/kueue.v1beta1/#kueue-x-k8s-io-v1beta1-LocalQueue). LocalQueues are the last piece of data the UI needs to fetch to implement features for our initial Distributed Workloads GA. The `LocalQueueKind` type is based on referencing [the CRD here](https://github.com/kubernetes-sigs/kueue/blob/main/config/components/crd/bases/kueue.x-k8s.io_localqueues.yaml) and [the API docs here](https://kueue.sigs.k8s.io/docs/reference/kueue.v1beta1/#kueue-x-k8s-io-v1beta1-LocalQueue).
* Also cleans up some repetition in the types for ClusterQueueKind and adds API doc links to all 3 Kueue CR types in that file.
* Completes the empty-state work that started in #2644 by providing the logic for the "No quota set" empty state on the Project Metrics tab. Per @astefanutti, that message should be shown if there are is no ClusterQueue, or the ClusterQueue's `spec.resourceGroups` is missing or empty, or there are no LocalQueues in the selected namespace.
* Simplifies the logic for working with a singleton ClusterQueue: Previously, we were passing down the whole array of `clusterQueues` via context, but this UI will only support a single ClusterQueue being created in the cluster. The context provider now only passes down a single `clusterQueue` which will either be the first ClusterQueue found that has `resourceGroups`, or it will be `undefined`.

Screenshot 2024-03-29 at 6 19 45 PM

cc @xianli123 @cfullam1 

## How Has This Been Tested?

Sanity checked everything locally.

## Test Impact

* Unit tests have been added for `listLocalQueues` and `useLocalQueues`, similar to the existing tests for `listWorkloads` and `useWorkloads` etc.
* Cypress tests have been added to make sure the "Quota is not set" message appears under the correct circumstances. I also reorganized the Cypress tests to group them into separate `describe` blocks per tab.

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/pages/distributedWorkloads/global/projectMetrics/GlobalDistributedWorkloadsProjectMetricsTab.tsx
+++ b/frontend/src/pages/distributedWorkloads/global/projectMetrics/GlobalDistributedWorkloadsProjectMetricsTab.tsx
@@ -1,5 +1,15 @@
 import * as React from 'react';
-import { Bullseye, Spinner, Stack, StackItem } from '@patternfly/react-core';
+import {
+  Bullseye,
+  EmptyState,
+  EmptyStateBody,
+  EmptyStateHeader,
+  EmptyStateIcon,
+  Spinner,
+  Stack,
+  StackItem,
+} from '@patternfly/react-core';
+import { WrenchIcon } from '@patternfly/react-icons';
 import { DistributedWorkloadsContext } from '~/concepts/distributedWorkloads/DistributedWorkloadsContext';
 import EmptyStateErrorMessage from '~/components/EmptyStateErrorMessage';
 import { ResourceUsage } from './sections/ResourceUsage';
@@ -8,18 +18,21 @@ import { WorkloadResourceMetrics } from './sections/WorkloadResourceMetrics';
 import { DWSectionCard } from './sections/DWSectionCard';
 
 const GlobalDistributedWorkloadsProjectMetricsTab: React.FC = () => {
-  const { projectCurrentMetrics } = React.useContext(DistributedWorkloadsContext);
+  const { clusterQueue, localQueues } = React.useContext(DistributedWorkloadsContext);
+  const requiredFetches = [clusterQueue, localQueues];
+  const error = requiredFetches.find((f) => !!f.error)?.error;
+  const loaded = requiredFetches.every((f) => f.loaded);
 
-  if (projectCurrentMetrics.error) {
+  if (error) {
     return (
       <EmptyStateErrorMessage
         title="Error loading distributed workload metrics"
-        bodyText={projectCurrentMetrics.error.message}
+        bodyText={error.message}
       />
     );
   }
 
-  if (!projectCurrentMetrics.loaded) {
+  if (!loaded) {
     return (
       <Bullseye style={{ minHeight: 150 }}>
         <Spinner />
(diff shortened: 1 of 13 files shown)


xianli123:
Thanks @mturley. It looks great.

frontend/src/pages/distributedWorkloads/global/projectMetrics/GlobalDistributedWorkloadsProjectMetricsTab.tsx
```diff
@@ -8,43 +18,41 @@ import { WorkloadResourceMetrics } from './sections/WorkloadResourceMetrics';
...
-  //         titleText="Quota is not set"
-  //         headingLevel="h4"
-  //         icon={<EmptyStateIcon icon={WrenchIcon} />}
-  //       />
-  //       <EmptyStateBody>Select another project or set the quota for this project.</EmptyStateBody>
-  //     </EmptyState>
-  //   );
-  // }
+  // clusterQueue.data will only be defined here if it has spec.resourceGroups (see DistributedWorkloadsContext)
+  if (!clusterQueue.data || localQueues.data.length === 0) {
```
andrewballantyne, line 44: For one -- Don't describe your data outside of its definition -- this will quickly get out of date.

For two -- Why is clusterQueue important here? If you don't have a local queue, doesn't that invalidate the project?

mturley: One -- maybe this comment just isn't useful anyway but what it describes is obvious from the code where it is defined. The point of this comment was "the design calls for this empty state if the cluster queue has no resource groups, that case is covered here even though it looks like we're only checking for it to exist". Confusing and probably unnecessary now that I return to it.

Two -- it's possible for a user to create a ClusterQueue and LocalQueue but not have any `resourceGroups` defined on the ClusterQueue, which is what sets the resource quotas.

Maybe we should be checking spec.resourceGroups here directly instead of in the context to resolve both of these?

andrewballantyne: Perhaps this is a DW thing I don't understand... how is
```
          titleText="Quota is not set"
```
valid when you don't have resources? 🤔 Are Quotas set inside resources?

LocalQueue is about having none... where ClusterQueue is about having no resources? 🤔 This feels weird.

mturley: You need both Cluster and Local. ClusterQueue defines the available resources for the whole cluster, LocalQueue defines some subset of those resources to claim (typically one per namespace, but you can have more than one). I think the idea is to cover this in product documentation initially, but one of the future epics is to provide a UI settings page for configuring these quotas and add a button here to take you there.

mturley: although hang on, checking my assumption here... a localQueue doesn't have resourceRequests in its spec. I think you just need to have one existing to be ready to start workloads and that got lumped into the "set a quota" logic. I was just implementing this based on what @astefanutti asked for

mturley: > Are Quotas set inside resources?

My understanding is that the "Quota" is the available resources in `clusterQueue.spec.resourceRequests`

gitdallas (approved):

andrewballantyne:
/override ci/prow/odh-dashboard-pr-image-mirror ci/prow/images

PR #2655, opened 2024-04-01
[RHOAIENG-4660] Compare runs - Manage runs page

jpuzz0:
Closes: [RHOAIENG-4660](https://issues.redhat.com/browse/RHOAIENG-4660)

## Description
Manage runs page uses selected runs from the Compare runs page initially as pre-selected table rows (runs), and the user has the ability to select up to 10 total (and change what was initially selected if they want). Selections are maintained between pages, cancelling uses what was initially selected by the Compare runs page, and "Update" will update the selections and bring the user back to the Compare runs page.

## Tests
Cypress tests for main interactions

## How to test
1. Navigate to `http://localhost:4010/experiments/jps-fun-world/85c9d729-bf63-441f-af87-d6898e5db5f0/compareRuns/add`
2. Verify the initial Experiment filter is set to the experiment associated with the ID in your URL path
3. Select up to 10 table rows on a single or multiple pages
4. "Update" and verify you're navigated to the Compare runs page with URL search string updated with all run valid IDs
5. Navigate to the same path from step 1
6. "Cancel" and verify you're navigated to the Compare runs page with no "runs" in the search param
7. Navigate to the same path from step 1, except with `?runs=` following with a comma delimited list of run IDs to verify initial rows are checked (assuming those IDs are valid)
8. Verify filter, row links, pagination, sorting works for the Manage runs table
9. Verify selections between pages persist
10. Verify breadcrumbs bring you to the correct locations

image
image

(cc @yannnz)

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/components/table/TableBase.tsx
+++ b/frontend/src/components/table/TableBase.tsx
@@ -35,6 +35,7 @@ type Props<DataType> = {
   enablePagination?: boolean | 'compact';
   truncateRenderingAt?: number;
   toolbarContent?: React.ReactElement<typeof ToolbarItem | typeof ToolbarGroup>;
+  bottomToolbarContent?: React.ReactElement<typeof ToolbarItem | typeof ToolbarGroup>;
   emptyTableView?: React.ReactNode;
   caption?: string;
   footerRow?: (pageNumber: number) => React.ReactElement<typeof Tr> | null;
@@ -77,6 +78,7 @@ const TableBase = <T,>({
   rowRenderer,
   enablePagination,
   toolbarContent,
+  bottomToolbarContent,
   emptyTableView,
   caption,
   disableRowRenderSupport,
@@ -222,7 +224,11 @@ const TableBase = <T,>({
   return (
     <>
       {(toolbarContent || showPagination) && (
-        <Toolbar inset={{ default: 'insetNone' }} customChipGroupContent={<></>}>
+        <Toolbar
+          inset={{ default: 'insetNone' }}
+          className="pf-v5-u-w-100"
+          customChipGroupContent={<></>}
+        >
           <ToolbarContent>
             {toolbarContent}
             {showPagination && (
@@ -249,15 +255,17 @@ const TableBase = <T,>({
           {emptyTableView}
         </div>
       )}
-      {showPagination && (
-        <Toolbar>
-          <ToolbarContent>
+
+      <Toolbar inset={{ default: 'insetNone' }} className="pf-v5-u-w-100">
+        <ToolbarContent alignItems="center">
+          {bottomToolbarContent}
+          {showPagination && (
             <ToolbarItem variant="pagination" align={{ default: 'alignRight' }}>
               {pagination('bottom')}
             </ToolbarItem>
-          </ToolbarContent>
-        </Toolbar>
-      )}
+          )}
+        </ToolbarContent>
+      </Toolbar>
     </>
   );
 };
--- a/frontend/src/components/table/useCheckboxTableBase.ts
+++ b/frontend/src/components/table/useCheckboxTableBase.ts
@@ -16,7 +16,7 @@ const useCheckboxTableBase = <T>(
   selectedData: T[],
   setSelectedData: React.Dispatch<React.SetStateAction<T[]>>,
   dataMappingHelper: (selectData: T) => string,
-  selectAll?: { selected?: boolean; disabled?: boolean },
+  options?: { selectAll?: { selected?: boolean; disabled?: boolean }; persistSelections?: boolean },
 ): UseCheckboxTableBaseProps<T> => {
   const dataIds = React.useMemo(() => data.map(dataMappingHelper), [data, dataMappingHelper]);
 
@@ -29,6 +29,10 @@ const useCheckboxTableBase = <T>(
 
   // remove selected ids that are no longer present in the provided dataIds
   React.useEffect(() => {
+    if (options?.persistSelections) {
+      return;
+    }
+
     const newSelectedIds = intersection(selectedDataIds, dataIds);
     const newSelectedData = newSelectedIds
       .map((id) => data.find((d) => dataMappingHelper(d) === id))
@@ -36,7 +40,15 @@ const useCheckboxTableBase = <T>(
     if (selectedData.length !== newSelectedData.length) {
       setSelectedData(newSelectedData);
     }
-  }, [data, dataIds, dataMappingHelper, selectedData, selectedDataIds, setSelectedData]);
+  }, [
+    data,
+    dataIds,
+    dataMappingHelper,
+    options?.persistSelections,
+    selectedData,
+    selectedDataIds,
+    setSelectedData,
+  ]);
 
   const disableCheck = React.useCallback<UseCheckboxTableBaseProps<T>['disableCheck']>(
     (item, disabled) =>
@@ -74,7 +86,7 @@ const useCheckboxTableBase = <T>(
             setSelectedData(value ? checkable : []);
           },
           selected: headerSelected,
-          ...selectAll,
+          ...options?.selectAll,
         },
       },
       disableCheck,
@@ -95,10 +107,10 @@ const useCheckboxTableBase = <T>(
     selectedDataIds,
     dataMappingHelper,
     selectedData,
-    selectAll,
-    disableCheck,
     disabledData,
     setSelectedData,
+    options?.selectAll,
+    disableCheck,
   ]);
 };
 
--- a/frontend/src/__tests__/cypress/cypress/e2e/pipelines/ManageRuns.cy.ts
+++ b/frontend/src/__tests__/cypress/cypress/e2e/pipelines/ManageRuns.cy.ts
@@ -1,0 +1,171 @@
+/* eslint-disable camelcase */
+import {
+  buildMockExperimentKF,
+  buildMockPipelineV2,
+  buildMockPipelineVersionV2,
+  buildMockPipelineVersionsV2,
+  buildMockPipelines,
+  buildMockRunKF,
+} from '~/__mocks__';
+import { PipelineRunKFv2 } from '~/concepts/pipelines/kfTypes';
+import {
+  manageRunsPage,
+  manageRunsTable,
+} from '~/__tests__/cypress/cypress/pages/pipelines/manageRuns';
+import { configIntercept, dspaIntercepts, projectsIntercept, statusIntercept } from './intercepts';
+
+const projectName = 'test-project-name';
+const pipelineVersionId = 'test-version-id';
+const pipelineId = 'test-pipeline-id';
+const experimentId = 'test-experiment-id';
+const initialRunIds = ['test-run-1', 'test-run-2'];
+
+const mockRuns = Array(11)
+  .fill(buildMockRunKF())
+  .map((mockRun: Partial<PipelineRunKFv2>, index) => ({
+    ...mockRun,
+    display_name: `Test run ${index + 1}`,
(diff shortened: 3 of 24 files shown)


frontend/src/components/table/TableBase.tsx
```diff
@@ -222,7 +224,11 @@ const TableBase = <T,>({
   return (
     <>
       {(toolbarContent || showPagination) && (
-        <Toolbar inset={{ default: 'insetNone' }} customChipGroupContent={<></>}>
+        <Toolbar
+          inset={{ default: 'insetNone' }}
+          className="pf-v5-u-w-100"
```
Gkrumbach07, line 229: why was this needed? `className="pf-v5-u-w-100"`

frontend/src/concepts/pipelines/content/compareRuns/CompareRunsContext.tsx
```diff
@@ -38,32 +38,38 @@ export const CompareRunsContextProvider = conditionalArea<CompareRunsContextProv
     const runIds = searchParams.get(CompareRunsSearchParam.RUNS);
     return runIds ? runIds.split(',') : [];
   }, [searchParams]);
+  const loadedRef = React.useRef(false);
```
Gkrumbach07, line 41: why switch to using a ref here and not the loaded state from useFetchState? It looks like now on initial navigation to the page, the error/empty state is flashed instead of the loading indicator

Gkrumbach07: i am guessing this is becuase loaded is set before runs is set, causing a few frames to be incorrect

frontend/src/pages/pipelines/global/runs/ManageRunsPage.tsx
```diff
@@ -0,0 +1,171 @@
...
+
+  if (!loaded && !initialLoaded) {
+    return (
+      <Bullseye>
+        <Spinner />
+      </Bullseye>
+    );
+  }
+
+  if (loaded && totalSize === 0 && !tableProps.filter) {
```
Gkrumbach07, line 81: can you add a screenshot for this

frontend/src/pages/pipelines/global/runs/ManageRunsTable.tsx
```diff
@@ -0,0 +1,173 @@
...
+  filterProps,
+  ...tableProps
+}) => {
+  const navigate = useNavigate();
+  const [isSelectAllChecked, setIsSelectAllChecked] = React.useState(false);
+  const { namespace } = usePipelinesAPI();
+  const pageRunIds = runs.map(({ run_id: runId }) => runId);
+
+  const { tableProps: checkboxTableProps, toggleSelection } = useCheckboxTable(pageRunIds);
+  const selectedIdsRef = React.useRef<string[]>(selectedRunIds);
```
Gkrumbach07, line 45: useCheckboxTable manages selections and select all, why are you managing them outside the hook?

frontend/src/pages/pipelines/global/runs/ManageRunsTable.tsx
```diff
@@ -0,0 +1,173 @@
...
+
+            if (isSelectAllChecked) {
+              setIsSelectAllChecked(false);
+            }
+          }
+        }}
+        selectAll={{
+          onSelect: onSelectAll,
+          selected: isSelectAllChecked,
+        }}
```
Gkrumbach07, line 115: this overrides checkboxTableProps.selectAll. is this on purpose?

frontend/src/pages/pipelines/global/runs/ManageRunsTable.tsx
```diff
@@ -0,0 +1,173 @@
...
+                  onClick={() =>
+                    navigate(
+                      experimentsCompareRunsRoute(namespace, experimentId, selectedIdsRef.current),
+                    )
+                  }
+                  isDisabled={selectedIdsRef.current.length < 1}
+                >
+                  Update
+                </Button>
+              </Tooltip>
```
Gkrumbach07, line 153: missing the tooltip when the button is disabled. You can use isAriaDisabled so the tooltip still works in a disabled state.

frontend/src/components/table/TableBase.tsx
jpuzz0: Without the width of 100% added here, when there are no actions next to the filter input, the pagination for the table gets pushed to the immediate left of the filter input instead of floating to the far right as intended.

frontend/src/concepts/pipelines/content/compareRuns/CompareRunsContext.tsx
jpuzz0: Once the compare runs page loads, that useFetchState's "loaded" state never changes from "true" even when the array of IDs changes. If you want to solve this some other way, let me know.

frontend/src/pages/pipelines/global/runs/ManageRunsTable.tsx
jpuzz0: Because the checkbox selections otherwise get reset between page changes.

frontend/src/pages/pipelines/global/runs/ManageRunsTable.tsx
jpuzz0: Yes, for the same reason to your comment here: https://github.com/opendatahub-io/odh-dashboard/pull/2655#discussion_r1548285555

frontend/src/pages/pipelines/global/runs/ManageRunsPage.tsx
jpuzz0: image

Gkrumbach07: added it to the description

Gkrumbach07:
organizational suggestion: can you put the ManageRunsPage under the compare runs folder instead of the runs folder

frontend/src/concepts/pipelines/content/compareRuns/CompareRunsContext.tsx
Gkrumbach07: thats correct, it wont. thats just how useFetchState works. loaded will be true when the data is not an error and has loaded before. You should be able to react on runs changing instead of having the loaded ref flip.

jpuzz0: I don't understand what you're suggesting, but your if-statement in the effect added here; https://github.com/opendatahub-io/odh-dashboard/blob/b7576b6a87b63daf8479d5a498595c7b77938092/frontend/src/concepts/pipelines/content/compareRuns/CompareRunsContext.tsx#L64 is checking for a difference in runs already. This effect triggers before the request for new runs occurs when switching from Manage runs to Compare runs. 

This means that we get invalid runs and those run IDs in the URL get stripped before the request even gets made. How would you like to solve this problem?

jpuzz0: Alternatively, if we don't have Manage runs be a sub-route of Compare runs, the whole context would re-render when visiting Compare runs (coming from Manage runs) and this wouldn't be an issue.

frontend/src/pages/pipelines/global/runs/ManageRunsTable.tsx
Gkrumbach07: i tinkered around. you can just pass a default set of strings to the usecvheckboxtable hook and then set that to selections right away. this means you dont need the custom select all, on toggle, and use ref selections

frontend/src/pages/pipelines/global/runs/ManageRunsTable.tsx
Gkrumbach07: moving convo there then

frontend/src/concepts/pipelines/content/compareRuns/CompareRunsContext.tsx
Gkrumbach07: lets do that. move the context under each page respectively. really the context will be mostly useful in the first compare runs page as it will have many components relying on the selections

frontend/src/pages/pipelines/global/runs/ManageRunsTable.tsx
jpuzz0: What default set of strings would one pass to useCheckboxTable? Set what to selections right away?

Gkrumbach07: that table has a prop for the selectedRunIds. and the state inside the checkbox table hook can eb set with an additional optional input param for defaults

jpuzz0: I think your last sentence is suggesting updating the useCheckboxTable hook with a new param for default IDs set in state, but I'm not seeing how that would result in the same table behavior as the current implementation. Have you tested this and everything behaves the same way?

Gkrumbach07: i have. you will also need to then use the selections from the checkbox table hook

jpuzz0: Those "selections" from the checkbox hook are reset between page changes. Maybe you could elaborate.

jpuzz0: I asked this question to Yan last week: https://www.figma.com/proto/etbN0ccCcZR1qKCpJXQNAg?node-id=1750-107510#751898481. So there's a different expectation here it seems.

jpuzz0: pagination

Gkrumbach07: when i tested it on my end the selection seems to work as expected across pagination

frontend/src/components/table/useCheckboxTableBase.ts
```diff
@@ -17,6 +17,7 @@ const useCheckboxTableBase = <T>(
   setSelectedData: React.Dispatch<React.SetStateAction<T[]>>,
   dataMappingHelper: (selectData: T) => string,
   selectAll?: { selected?: boolean; disabled?: boolean },
+  persistSelections?: boolean,
```
Gkrumbach07, line 20: code quality wise we should put these two into a `opts` param so we can avoid putting undefined in when we dont need to set selectAll

Gkrumbach07 (commented):
looks good. just one comment then we can look to merge

frontend/src/components/table/useCheckboxTableBase.ts
jpuzz0: I disagree that this suggestion improves code quality, and this seems more like a personal preference, but have updated anyways.

Gkrumbach07: My reasoning is that as function params start to have multiple optional params you can start to have things like this in an extreme case:

```tsx
const someFunc = (p1: string, p2: string, op1?: string, op2?: string, op3?: string, op4?: string) => {...}

someFunc("one", "two", undefined, undefined, undefined, "optional")
```

With one optional param its not a big deal, two sure, but three definitely want to do something about it. The reason I  flagged it is becuase it was talked about at some point prior as something we wanted to shy away from.

Gkrumbach07: There was a jira to look to add a linter rule for it, but i dont think it was added as one doesn't exist

https://issues.redhat.com/browse/RHOAIENG-513

PR #2662, opened 2024-04-03
[RHOAIENG-4980] Fix RBAC on every tab for the action button

jeff-phillips-18:
Closes: [RHOAIENG-4980](https://issues.redhat.com/browse/RHOAIENG-4980)

## Description
Remove erroneous RBAC checks

## How Has This Been Tested?
Logged in as non-admin, verified I could create necessary items.

## Test Impact
None

## Request review criteria:

### DIFF
--- a/frontend/src/components/EmptyDetailsView.tsx
+++ b/frontend/src/components/EmptyDetailsView.tsx
@@ -13,7 +13,7 @@ type EmptyDetailsViewProps = {
   description?: string;
   iconImage?: string;
   imageAlt?: string;
-  allowCreate: boolean;
+  allowCreate?: boolean;
   createButton?: React.ReactNode;
   imageSize?: string;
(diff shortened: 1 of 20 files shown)


andrewballantyne (changes requested):
Have you tested this with an edit user? The Permissions tab should not be visible to those users.

jeff-phillips-18:
@andrewballantyne Rebase and added check before showing the permissions card.

christianvogt:
@jeff-phillips-18 Logged in as a self-provisioner user who was granted `edit` access to a project and I still see the permissions tab as well as permissions card:
image

image

christianvogt:
There's also this issue when trusty isn't disabled even though it should be. Maybe related to how tabs are being enabled and the permissions tab issue.
image

PR #2690, opened 2024-04-08
[RHOAIENG-4727] Compare runs - Parameters expandable section

jpuzz0:
Closes: [RHOAIENG-4727](https://issues.redhat.com/browse/RHOAIENG-4727)

## Description
Compare runs - Parameters expandable section renders runtime config data associated with all selected runs from the `Run list`, where if a run does not share a parameter with another, empty cells are shown. The feature has a switch that allows for showing and hiding of parameter rows based on whether rows possess all of the same value. When row data extends beyond a certain point, horizontal scroll will kick in and the first column representing parameter names will stick while scrolling left and right.

**NOTE:** The design diverged slightly from what the mocks show at the moment. The implementation here is what was recently agreed upon with @yannnz.

image

**Only showing parameters with values that differ between runs**
image

**Empty state**
image

## How Has This Been Tested?
Added cypress tests demonstrating interactions between the `Run list` and `Parameters` tables, along with hiding of parameters with values that are all equal.

## How to test
Navigate to the Compare runs page `/experiments/jps-fun-world/85c9d729-bf63-441f-af87-d6898e5db5f0/compareRuns?runs=[your list of runs here]`, verify the table design and interactions with the Run list above it.

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/pages/pipelines/global/experiments/compareRuns/CompareRunParamsSection.tsx
+++ b/frontend/src/pages/pipelines/global/experiments/compareRuns/CompareRunParamsSection.tsx
@@ -1,0 +1,168 @@
+import React from 'react';
+
+import {
+  EmptyState,
+  EmptyStateBody,
+  EmptyStateHeader,
+  EmptyStateVariant,
+  ExpandableSection,
+  Flex,
+  Switch,
+} from '@patternfly/react-core';
+import { InnerScrollContainer, TableVariant, Td, Tr } from '@patternfly/react-table';
+
+import { SortableData, Table } from '~/components/table';
+import { useCompareRuns } from '~/concepts/pipelines/content/compareRuns/CompareRunsContext';
+import { RuntimeConfigParamValue } from '~/concepts/pipelines/kfTypes';
+import { normalizeInputParamValue } from '~/concepts/pipelines/content/pipelinesDetails/pipelineRun/utils';
+
+export const CompareRunParamsSection: React.FunctionComponent = () => {
+  const { loaded, selectedRuns } = useCompareRuns();
+  const [isSectionOpen, setIsSectionOpen] = React.useState(true);
+  const [isHideSameRowsChecked, setIsHideSameRowsChecked] = React.useState<boolean>(false);
+
+  const runNameColumns: SortableData<[string, Record<string, RuntimeConfigParamValue>]>[] =
+    React.useMemo(
+      () => [
+        {
+          label: 'Run name',
+          field: 'run-name',
+          isStickyColumn: true,
+          hasRightBorder: true,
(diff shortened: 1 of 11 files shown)


frontend/src/pages/pipelines/global/experiments/compareRuns/CompareRunParamsSection.tsx
```diff
@@ -0,0 +1,170 @@
...
+    React.useMemo(
+      () =>
+        hasSelectedRuns
+          ? [
+              {
+                label: 'Run name',
+                field: 'run-name',
+                isStickyColumn: true,
+                hasRightBorder: true,
+                backgroundColor: '#ededed',
```
Gkrumbach07, line 35: I am pretty sure there is not a way to do this through patternfly, so first i want to double check with @yannnz @xianli123 -- PF does not allow us to set individual columns a certain darker color (only alternating rows). Is this something that is needed? If so we can continue using the backgroundColor custom styling. If not we should remove it in favor of staying more consistent with PF.

Gkrumbach07 (commented):
code looks good besides the question on background color. will test it locally tomorrow

frontend/src/pages/pipelines/global/experiments/compareRuns/CompareRunParamsSection.tsx
jpuzz0: I meant to switch this to a PF className, so I've updated with that instead. This was demoed to both Yan & Kun a couple days ago and the idea is that we contribute this guideline back to them at some point for use-cases similar to ours.

Gkrumbach07: Ok. Do you mind opening up a upstream patternfly feature request ticket with the use case. Then comment that ticket url in the scss file next to where the class is defined for the custom styling as a TODO. This is how we have been tracking upstream PF tickets so we can go back and remove them when they close

jpuzz0: It's not a custom class name, it's a PF class name that we're using here. Yan started a conversation here: https://patternfly.slack.com/archives/CGR204RM2/p1712631660224649.

I can open a discussion in github and link that in a comment where the patternfly className is used, but beyond that, this has yet to be deemed something PF would work on.

Gkrumbach07: ok. yeah then that is fine. If an issue does get made we should make sure to document it somewhere. in the meantime, this doesnt need to block the pr

Gkrumbach07 (approved):
/lgtm
/approve

PR #2692, opened 2024-04-09
Fix pipeline version reference in pipeline run details

Gkrumbach07:
closes: https://issues.redhat.com/browse/RHOAIENG-5245

## Description

- handles sdk flow where user creates a run using `kfp.Client.create_run_from_pipeline_package` which produces a run without a reference to a pipeline/pipeline version. This PR adjusts the types to make it the reference optional so we dont run into this type error in the future
- also now will fallback to run.pipeline_spec (is only available when creating a run using `create_run_from_pipeline_package`) instead of fetching the version and getting the spec from there.

## How Has This Been Tested?

1. create a pipeline run using the v2 sdk
   - I used this tutorial: https://www.kubeflow.org/docs/components/pipelines/v2/hello-world/
   - I had ssl issues, you may not, but to get around it see: https://v1-5-branch.kubeflow.org/docs/components/pipelines/sdk/connect-api/#connect-to-kubeflow-pipelines-from-outside-your-cluster
   - i ended up portforwarding like this: `oc port-forward svc/ds-pipeline-dspa 4000:8888 --namespace <namespace>`
   - and then set the host i your python script to `http://localhost:4000`
2. this should kick of a pipeline with no reference to a pipeline version.
3. you should be able to interact fully with the run by going to the details and seeing all the run details and task details
4. you cna clone the run, but it wont pull in the pipeline spec so the pipeline dropdown will be empty - not sure if thats possible or not from the api so i left it out of this PR

## Test Impact

no impact to tests other than removing an unused param that could cause future confusion

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/concepts/pipelines/content/__tests__/utils.spec.tsx
+++ b/frontend/src/concepts/pipelines/content/__tests__/utils.spec.tsx
@@ -19,7 +19,6 @@ import { computeRunStatus } from '~/concepts/pipelines/content/utils';
 const run: PipelineRunKFv2 = {
   created_at: '2023-09-05T16:23:25Z',
   storage_state: StorageStateKF.AVAILABLE,
-  pipeline_version_id: 'version-id',
   finished_at: '2023-09-05T16:24:34Z',
   run_id: 'dc66a214-4df2-4d4d-a302-0d02d8bba0e7',
   display_name: 'flip-coin',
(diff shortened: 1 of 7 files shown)


jpuzz0 (approved):
/lgtm

Gkrumbach07 (commented):
/approve

PR #2695, opened 2024-04-09
Read pipeline run node details from mlmd context

DaoDaoNoCode:
JIRA: https://issues.redhat.com/browse/RHOAIENG-4544

## Description

Read pipeline executions information from the mlmd service instead of parsing it by ourselves so it's more accurate. If the mlmd service is not available, it will fall back to our current logic.

Screenshot 2024-04-09 at 5 39 21 PM

## How Has This Been Tested?

1. Create a pipeline run
2. Click on the nodes
3. Compare it with the KubeFlow Pipeline runs UI, make sure the information is the same with the ones in that UI

## Test Impact

N/A, not sure how we should test the mlmd service, I can add it if needed after the PR gets reviewed.

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/concepts/pipelines/apiHooks/mlmd/useMlmdContext.ts
+++ b/frontend/src/concepts/pipelines/apiHooks/mlmd/useMlmdContext.ts
@@ -1,0 +1,45 @@
+import React from 'react';
+import { MlmdContext, MlmdContextTypes } from '~/concepts/pipelines/apiHooks/mlmd/types';
+import { usePipelinesAPI } from '~/concepts/pipelines/context';
+import { GetContextByTypeAndNameRequest } from '~/third_party/mlmd';
+import useFetchState, {
+  FetchState,
+  FetchStateCallbackPromise,
+  NotReadyError,
+} from '~/utilities/useFetchState';
+
+/**
+ * A hook used to use the MLMD service and fetch the MLMD context
+ * If being used without name/type, this hook will throw an error
+ * @param name The identifier to query a specific type of MLMD context. e.g. The runID for a pipeline run
+ */
+export const useMlmdContext = (
+  name?: string,
+  type?: MlmdContextTypes,
+  refreshRate?: number,
+): FetchState<MlmdContext | null> => {
+  const { metadataStoreServiceClient } = usePipelinesAPI();
+
+  const call = React.useCallback<FetchStateCallbackPromise<MlmdContext | null>>(async () => {
+    if (!type) {
+      return Promise.reject(new NotReadyError('No context type'));
+    }
+    if (!name) {
+      return Promise.reject(new NotReadyError('No context name'));
+    }
+
+    const request = new GetContextByTypeAndNameRequest();
+    request.setTypeName(type);
+    request.setContextName(name);
+    const res = await metadataStoreServiceClient.getContextByTypeAndName(request);
+    const context = res.getContext();
+    if (!context) {
+      return Promise.reject(new Error('Cannot find specified context'));
+    }
+    return context;
+  }, [metadataStoreServiceClient, type, name]);
+
+  return useFetchState(call, null, {
+    refreshRate,
+  });
+};
--- a/frontend/src/concepts/pipelines/apiHooks/mlmd/usePipelineRunMlmdContext.ts
+++ b/frontend/src/concepts/pipelines/apiHooks/mlmd/usePipelineRunMlmdContext.ts
@@ -1,0 +1,8 @@
+import { FetchState } from '~/utilities/useFetchState';
+import { MlmdContext, MlmdContextTypes } from '~/concepts/pipelines/apiHooks/mlmd/types';
+import { useMlmdContext } from '~/concepts/pipelines/apiHooks/mlmd/useMlmdContext';
+
+export const usePipelineRunMlmdContext = (
+  runID?: string,
+  refreshRate?: number,
+): FetchState<MlmdContext | null> => useMlmdContext(runID, MlmdContextTypes.RUN, refreshRate);
--- a/frontend/src/concepts/pipelines/content/pipelinesDetails/pipelineRun/runLogs/LogsTab.tsx
+++ b/frontend/src/concepts/pipelines/content/pipelinesDetails/pipelineRun/runLogs/LogsTab.tsx
@@ -40,12 +40,12 @@ import DownloadDropdown from '~/concepts/pipelines/content/pipelinesDetails/pipe
 import { PodStepStateType } from '~/types';
 import useDebounceCallback from '~/utilities/useDebounceCallback';
 import { PipelineTask } from '~/concepts/pipelines/topology';
-import { RuntimeStateKF } from '~/concepts/pipelines/kfTypes';
+import { ExecutionStateKF } from '~/concepts/pipelines/kfTypes';
 
 // TODO: If this gets large enough we should look to make this its own component file
 const LogsTab: React.FC<{ task: PipelineTask }> = ({ task }) => {
   const podName = task.status?.podName;
-  const isFailedPod = task.status?.state === RuntimeStateKF.FAILED;
+  const isFailedPod = task.status?.state === ExecutionStateKF.FAILED;
 
   if (!podName) {
     return <>No content</>;
--- a/frontend/src/concepts/pipelines/content/pipelinesDetails/pipelineRun/useExecutionsForPipelineRun.ts
+++ b/frontend/src/concepts/pipelines/content/pipelinesDetails/pipelineRun/useExecutionsForPipelineRun.ts
@@ -1,0 +1,21 @@
+import { useExecutionsFromMlmdContext } from '~/concepts/pipelines/apiHooks/mlmd/useExecutionsFromMlmdContext';
+import { usePipelineRunMlmdContext } from '~/concepts/pipelines/apiHooks/mlmd/usePipelineRunMlmdContext';
+import { isPipelineRunFinished } from '~/concepts/pipelines/apiHooks/usePipelineRunById';
+import { PipelineRunKFv2 } from '~/concepts/pipelines/kfTypes';
+import { Execution } from '~/third_party/mlmd';
+import { FAST_POLL_INTERVAL } from '~/utilities/const';
+
+const useExecutionsForPipelineRun = (
+  run: PipelineRunKFv2 | null,
+): [executions: Execution[] | null, loaded: boolean, error?: Error] => {
+  const isFinished = isPipelineRunFinished(run);
+  const refreshRate = isFinished ? 0 : FAST_POLL_INTERVAL;
+  // contextError means mlmd service is not available, no need to check executions
+  const [context, , contextError] = usePipelineRunMlmdContext(run?.run_id, refreshRate);
+  // executionsLoaded is the flag to show the spinner or not
+  const [executions, executionsLoaded] = useExecutionsFromMlmdContext(context, refreshRate);
+
+  return [executions, executionsLoaded, contextError];
+};
+
+export default useExecutionsForPipelineRun;
--- a/frontend/src/concepts/pipelines/topology/parseUtils.ts
+++ b/frontend/src/concepts/pipelines/topology/parseUtils.ts
@@ -1,6 +1,8 @@
 import { RunStatus } from '@patternfly/react-topology';
 import {
+  ArtifactStateKF,
   DAG,
+  ExecutionStateKF,
   InputOutputArtifactType,
   InputOutputDefinition,
   PipelineComponentsKF,
@@ -10,6 +12,7 @@ import {
   TaskDetailKF,
 } from '~/concepts/pipelines/kfTypes';
 import { VolumeMount } from '~/types';
+import { Artifact, Execution } from '~/third_party/mlmd';
 import { PipelineTaskInputOutput, PipelineTaskRunStatus } from './pipelineTaskTypes';
 
 export const composeArtifactType = (data: InputOutputArtifactType): string =>
@@ -135,14 +138,13 @@ export const lowestProgress = (details: TaskDetailKF[]): PipelineTaskRunStatus['
   )[0].state;
 };
 
-export const parseRuntimeInfo = (
+export const parseRuntimeInfoFromRunDetails = (
   taskId: string,
   runDetails?: RunDetailsKF,
 ): PipelineTaskRunStatus | undefined => {
   if (!runDetails) {
     return undefined;
   }
-
   const { task_details: taskDetails } = runDetails;
 
   // taskId should always be first, as it's the most direct item, but it may not drive the entire details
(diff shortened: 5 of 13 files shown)


Gkrumbach07:
/hold

will go in next sprint

frontend/src/concepts/pipelines/apiHooks/mlmd/useExecutionsFromContext.ts
```diff
@@ -0,0 +1,31 @@
...
+import { Context, Execution, GetExecutionsByContextRequest } from '~/third_party/mlmd';
+import { FAST_POLL_INTERVAL } from '~/utilities/const';
+import useFetchState, {
+  FetchState,
+  FetchStateCallbackPromise,
+  NotReadyError,
+} from '~/utilities/useFetchState';
+
+export const useExecutionsFromContext = (
+  context: Context | null,
```
jpuzz0, line 12: Is it not confusing to have the concept of React's Context and also Context from MLMD? Maybe we should alias Context as `MlContext` or `MlmdContext` and name these functions `useExecutionsFromMlContext` or `useExecutionsFromMlmdContext`?

jpuzz0: And looking at your other file, it appears this was done to an extent already (e.g. `useMlmdContext`).

frontend/src/concepts/pipelines/apiHooks/mlmd/usePipelineRunMlmdContext.ts
```diff
@@ -0,0 +1,48 @@
...
+} from '~/utilities/useFetchState';
+import { GetContextByTypeAndNameRequest } from '~/third_party/mlmd/generated/ml_metadata/proto/metadata_store_service_pb';
+import { FAST_POLL_INTERVAL } from '~/utilities/const';
+
+const KFP_V2_RUN_CONTEXT_TYPE = 'system.PipelineRun';
+
+const useMlmdContext = (
+  name?: string,
+  type?: string,
+  pipelineFinished?: boolean,
```
jpuzz0, line 17: What does "pipelineFinished" mean exactly? Finished loading or something else?

frontend/src/concepts/pipelines/apiHooks/mlmd/usePipelineRunMlmdContext.ts
```diff
@@ -0,0 +1,48 @@
...
+  NotReadyError,
+} from '~/utilities/useFetchState';
+import { GetContextByTypeAndNameRequest } from '~/third_party/mlmd/generated/ml_metadata/proto/metadata_store_service_pb';
+import { FAST_POLL_INTERVAL } from '~/utilities/const';
+
+const KFP_V2_RUN_CONTEXT_TYPE = 'system.PipelineRun';
+
+const useMlmdContext = (
+  name?: string,
+  type?: string,
```
jpuzz0, line 16: Is there a known set of MLMD context types we can make a typescript type for here?

frontend/src/concepts/pipelines/apiHooks/mlmd/usePipelineRunMlmdContext.ts
jpuzz0: Is it not the pipeline but the pipeline run that is indicated as finished?

frontend/src/concepts/pipelines/content/pipelinesDetails/pipelineRun/runLogs/LogsTab.tsx
```diff
@@ -40,12 +40,11 @@ import DownloadDropdown from '~/concepts/pipelines/content/pipelinesDetails/pipe
 import { PodStepStateType } from '~/types';
 import useDebounceCallback from '~/utilities/useDebounceCallback';
 import { PipelineTask } from '~/concepts/pipelines/topology';
-import { RuntimeStateKF } from '~/concepts/pipelines/kfTypes';
 
 // TODO: If this gets large enough we should look to make this its own component file
 const LogsTab: React.FC<{ task: PipelineTask }> = ({ task }) => {
   const podName = task.status?.podName;
-  const isFailedPod = task.status?.state === RuntimeStateKF.FAILED;
+  const isFailedPod = task.status?.state === 'Failed';
```
jpuzz0, line 47: Can you explain why we are not using an enum value here any longer?

frontend/src/concepts/pipelines/topology/usePipelineTaskTopology.ts
```diff
@@ -2,23 +2,37 @@ import { PipelineRunKFv2, PipelineSpecVariable } from '~/concepts/pipelines/kfTy
...
 import { KubeFlowTaskTopology } from './pipelineTaskTypes';
 
-const EMPTY_STATE: KubeFlowTaskTopology = { taskMap: {}, nodes: [] };
+const EMPTY_STATE: KubeFlowTaskTopology = { taskMap: {}, nodes: [], loaded: true };
 
 export const usePipelineTaskTopology = (
   spec?: PipelineSpecVariable,
   run?: PipelineRunKFv2,
 ): KubeFlowTaskTopology => {
+  const isComplete = isPipelineRunComplete(run);
```
jpuzz0, line 26: nit: The concepts of "complete" and "finished" are similar, but maybe just stick with 1 or the other across the board.

jpuzz0: Complete can sometimes be associated with a success, so maybe "finished" would be more appropriate.

frontend/src/concepts/pipelines/topology/usePipelineTaskTopology.ts
```diff
@@ -2,23 +2,37 @@ import { PipelineRunKFv2, PipelineSpecVariable } from '~/concepts/pipelines/kfTy
...
+  parseRuntimeInfoFromExecutions,
+  parseRuntimeInfoFromRunDetails,
   parseTasksForArtifactRelationship,
   parseVolumeMounts,
   translateStatusForNode,
 } from './parseUtils';
 import { KubeFlowTaskTopology } from './pipelineTaskTypes';
 
-const EMPTY_STATE: KubeFlowTaskTopology = { taskMap: {}, nodes: [] };
+const EMPTY_STATE: KubeFlowTaskTopology = { taskMap: {}, nodes: [], loaded: true };
```
jpuzz0, line 20: Are we sure we want an empty state to be considered "loaded"? Would there be a request made and finished at the point of this object being returned?

frontend/src/concepts/pipelines/topology/pipelineTaskTypes.ts
```diff
@@ -31,7 +31,7 @@ export type PipelineTaskRunStatus = {
   startTime: string;
   completeTime?: string;
   podName?: string;
-  state?: RuntimeStateKF;
+  state?: string;
```
jpuzz0, line 34: I think we should try to keep the state an enum here unless there's a good reason not to.

frontend/src/concepts/pipelines/topology/parseUtils.ts
```diff
@@ -159,24 +159,131 @@ export const parseRuntimeInfo = (
...
+
+  if (!execution) {
+    return undefined;
+  }
+
+  const lastUpdatedTime = execution.getLastUpdateTimeSinceEpoch();
+  let completeTime;
+  if (
+    lastUpdatedTime &&
+    (execution.getLastKnownState() === Execution.State.COMPLETE ||
```
jpuzz0, line 187: nit: can we add a constant for `execution.getLastKnownState()`?

jpuzz0: Also, what sets these states apart from the rest? Maybe they should be stored in their own array, or at least we can add a comment as to why these states are chosen?

frontend/src/concepts/pipelines/topology/parseUtils.ts
```diff
@@ -159,24 +159,131 @@ export const parseRuntimeInfo = (
...
     case RuntimeStateKF.CANCELING:
       return RunStatus.Cancelled;
-    case RuntimeStateKF.PAUSED:
-      return RunStatus.Pending;
+    case 'Running':
+      return RunStatus.Running;
+    case 'Failed':
     case RuntimeStateKF.FAILED:
       return RunStatus.Failed;
+    case 'Pending':
```
jpuzz0, line 280: This is confusing, since the value being returned for the cases you're adding here match. (e.g. `RunStatus.Pending` = "Pending")

I'm not sure it makes sense to use this already unit tested function for this other usage.

frontend/src/concepts/pipelines/topology/parseUtils.ts
```diff
@@ -159,24 +159,131 @@ export const parseRuntimeInfo = (
...
+    const state = props.resource.getState();
+    switch (state) {
+      case Artifact.State.UNKNOWN:
+        return undefined; // when state is not set, it defaults to UNKNOWN
+      case Artifact.State.PENDING:
+        return 'Pending';
+      case Artifact.State.LIVE:
+        return 'Live';
+      case Artifact.State.MARKED_FOR_DELETION:
+        return 'Marked for deletion';
```
jpuzz0, line 239: Can we store these constant values as an enum elsewhere?

frontend/src/concepts/pipelines/topology/parseUtils.ts
```diff
@@ -159,24 +159,131 @@ export const parseRuntimeInfo = (
...
+
+export type ResourceInfoProps = ArtifactProps | ExecutionProps;
+
+// Get text representation of resource state.
+// Works for both artifact and execution.
+export const getResourceStateText = (props: ResourceInfoProps): string | undefined => {
+  if (props.resourceType === ResourceType.ARTIFACT) {
+    const state = props.resource.getState();
+    switch (state) {
+      case Artifact.State.UNKNOWN:
```
jpuzz0, line 232: Seems like this case can be stacked with the default.

frontend/src/concepts/pipelines/topology/parseUtils.ts
```diff
@@ -159,24 +159,131 @@ export const parseRuntimeInfo = (
...
+      case Artifact.State.DELETED:
+        return 'Deleted';
+      default:
+        return undefined;
+    }
+  } else {
+    // type == EXECUTION
+    const state = props.resource.getLastKnownState();
+    switch (state) {
+      case Execution.State.UNKNOWN:
```
jpuzz0, line 249: This case can be stacked with the default one since they both return undefined.

frontend/src/concepts/pipelines/topology/parseUtils.ts
```diff
@@ -159,24 +159,131 @@ export const parseRuntimeInfo = (
...
+  resourceType: ResourceType.EXECUTION;
+  resource: Execution;
+  typeName: string;
+}
+
+export type ResourceInfoProps = ArtifactProps | ExecutionProps;
+
+// Get text representation of resource state.
+// Works for both artifact and execution.
+export const getResourceStateText = (props: ResourceInfoProps): string | undefined => {
```
jpuzz0, line 228: At one point recently all parseUtils were unit tested and the file had 100% line coverage. I think it best we continue to unit test these functions since there was a ticket dedicated to adding tests for this file I assume for good reason.

frontend/src/concepts/pipelines/apiHooks/mlmd/usePipelineRunMlmdContext.ts
```diff
@@ -0,0 +1,48 @@
...
+  FetchState,
+  FetchStateCallbackPromise,
+  NotReadyError,
+} from '~/utilities/useFetchState';
+import { GetContextByTypeAndNameRequest } from '~/third_party/mlmd/generated/ml_metadata/proto/metadata_store_service_pb';
+import { FAST_POLL_INTERVAL } from '~/utilities/const';
+
+const KFP_V2_RUN_CONTEXT_TYPE = 'system.PipelineRun';
+
+const useMlmdContext = (
```
jpuzz0, line 14: Ideally, this is unit tested.

frontend/src/concepts/pipelines/apiHooks/mlmd/usePipelineRunMlmdContext.ts
DaoDaoNoCode: Not as I know of after checking the kubeflow code https://github.com/opendatahub-io/data-science-pipelines/blob/680e4452ceaa209f80c6fe8d689a02aaded786f7/frontend/src/mlmd/MlmdUtils.ts#L55, we can type them more when we get to pass more and more types into it later, but it should be good for now.

frontend/src/concepts/pipelines/content/pipelinesDetails/pipelineRun/runLogs/LogsTab.tsx
DaoDaoNoCode: Hmm, I think I was testing some new states that were brought by the V2 runs at that time and forgot to type it. I will add the types.

frontend/src/concepts/pipelines/topology/parseUtils.ts
DaoDaoNoCode: `RunStatus` is used only for showing the correct icon for topology nodes. It's an enum imported from the PF topology library and be used later. `Pending` here is the status we get from the pipeline run and mlmd context, sometimes the state could be different from the icon we want to show (i.e `RuntimeStateKF.RUNNING` is `RunStatus.InProgress` icon), I think it's good to deal with them separately.

frontend/src/concepts/pipelines/topology/parseUtils.ts
DaoDaoNoCode: The states are from here https://github.com/opendatahub-io/data-science-pipelines/blob/master/frontend/src/components/tabs/RuntimeNodeDetailsV2.tsx#L245-L253, I will add a comment to it.

frontend/src/concepts/pipelines/apiHooks/mlmd/usePipelineRunMlmdContext.ts
DaoDaoNoCode: We need to mock up some MLMD objects to unit test this, I prefer to push it later to add with other MLMD hooks.

frontend/src/concepts/pipelines/apiHooks/mlmd/usePipelineRunMlmdContext.ts
jpuzz0: I don't see why we would offer a type parameter in a file scoped to pipeline runs, when it seems the only type one would use here is the one you just referenced from KF's codebase.

frontend/src/concepts/pipelines/apiHooks/mlmd/useMlmdContext.ts
```diff
@@ -0,0 +1,41 @@
...
+import useFetchState, {
+  FetchState,
+  FetchStateCallbackPromise,
+  NotReadyError,
+} from '~/utilities/useFetchState';
+
+export const useMlmdContext = (
+  name?: string,
+  type?: string,
+  activelyRefresh?: boolean,
```
jpuzz0, line 15: Why not just have this be the refreshRate itself?

frontend/src/concepts/pipelines/apiHooks/mlmd/usePipelineRunMlmdContext.ts
jpuzz0: Looks like the file might have moved, but either way, it makes more sense to me to add the 1 type we're using now to an enum that can be expanded upon later.

frontend/src/concepts/pipelines/content/pipelinesDetails/pipelineRun/useExecutionsForPipelineRun.ts
```diff
@@ -0,0 +1,19 @@
...
+import { Execution } from '~/third_party/mlmd';
+
+const useExecutionsForPipelineRun = (
+  run: PipelineRunKFv2 | null,
+): [executions: Execution[] | null, loaded: boolean, error?: Error] => {
+  const isFinished = isPipelineRunFinished(run);
+  // contextError means mlmd service is not available, no need to check executions
+  const [context, , contextError] = usePipelineRunMlmdContext(run?.run_id, !isFinished);
+  // executionsLoaded is the flag to show the spinner or not
+  const [executions, executionsLoaded] = useExecutionsFromContext(context, !isFinished);
```
jpuzz0, line 14: Looks like this is still called "FromContext" instead of "FromMlmdContext". Is this intentional? If so, why?

DaoDaoNoCode: Ah, I forgot to rename it, the hooks are mostly from the KubeFlow codebase, where they call everything context. Will rename it, thanks!

frontend/src/concepts/pipelines/topology/parseUtils.ts
```diff
@@ -159,26 +161,133 @@ export const parseRuntimeInfo = (
...
+      case Execution.State.CACHED:
+        return ExecutionStateKF.CACHED;
+      default:
+        return undefined;
+    }
+  }
+};
+
+export const translateStatusForNode = (
+  state?: RuntimeStateKF | ExecutionStateKF | ArtifactStateKF,
```
jpuzz0, line 269: nit: I think a more comprehensive approach would be to have separate switch statements for the different types of state being used here.

frontend/src/concepts/pipelines/apiHooks/mlmd/usePipelineRunMlmdContext.ts
jpuzz0: I'm not sure why it would be done with other MLMD hooks. Why do you prefer that?

DaoDaoNoCode: I am just unsure how to write unit test for this... any ideas?

frontend/src/concepts/pipelines/topology/parseUtils.ts
```diff
@@ -159,26 +161,133 @@ export const parseRuntimeInfo = (
...
+
+export type ResourceInfoProps = ArtifactProps | ExecutionProps;
+
+// Get text representation of resource state.
+// Works for both artifact and execution.
+export const getResourceStateText = (
+  props: ResourceInfoProps,
+): ArtifactStateKF | ExecutionStateKF | undefined => {
+  if (props.resourceType === ResourceType.ARTIFACT) {
+    const state = props.resource.getState();
```
jpuzz0, line 233: This `Artifact` switch doesn't appear to be unit tested.

frontend/src/concepts/pipelines/topology/parseUtils.ts
jpuzz0, line 1: nit: parseVolumeMounts hasn't been unit tested, which was a part of a separate PR of yours I believe, but does prevent full coverage of this file.

frontend/src/concepts/pipelines/topology/parseUtils.ts
DaoDaoNoCode: That would be complex because we need to change the type of the state to something like `{type: xxx, state: xxx}` and I checked the state is used across a few files, which need some refactors to the use of these files, and I don't think it worth the effort just to try to separate the types. Basically, we are still doing the switch statement based on the string value.

jpuzz0: Why wouldn't there just be 3 functions, 1 for artifacts, 1 for executions and 1 for runtime?

DaoDaoNoCode: For example, if you separate this into 3 functions, which function should be called here? https://github.com/opendatahub-io/odh-dashboard/blob/main/frontend/src/concepts/pipelines/content/pipelinesDetails/pipelineRun/SelectedNodeDetailsTab.tsx#L39

jpuzz0: Theoretically there is something there that you could use to determine which function to use, and you'd conditionally use them based on whether you're dealing with artifact, experiment, or runtime state. If that's not possible then my idea would not work.

DaoDaoNoCode: Yes I wanted, but it's not something that could be easily implemented for now, because the `RuntimeState` is different. We always try to use MLMD first, and when it's not available, we parse the runtime state as a backup plan instead of showing users an error state. So it's hard to separate them.

frontend/src/concepts/pipelines/apiHooks/mlmd/usePipelineRunMlmdContext.ts
DaoDaoNoCode: OK, I tried to create some unit tests but it doesn't work good and I don't know why. I decided to create a JIRA ticket for it and just add some comments on this hook for better future use so that I can wrap this PR up.

DaoDaoNoCode:
@jpuzz0 Updated the PR and let me know, if it looks good I can squash the commits.

frontend/src/concepts/pipelines/apiHooks/mlmd/usePipelineRunMlmdContext.ts
DaoDaoNoCode: https://issues.redhat.com/browse/RHOAIENG-5510

frontend/src/concepts/pipelines/apiHooks/mlmd/useMlmdContext.ts
```diff
@@ -0,0 +1,45 @@
...
+import { GetContextByTypeAndNameRequest } from '~/third_party/mlmd';
+import useFetchState, {
+  FetchState,
+  FetchStateCallbackPromise,
+  NotReadyError,
+} from '~/utilities/useFetchState';
+
+/**
+ * A hook used to use the MLMD service and fetch the MLMD context
+ * If being used without name/type, this hook will throw an error
```
jpuzz0, line 13: I'm not sure why name and type are optional if the function doesn't work if they're not provided.

DaoDaoNoCode: I didn't test what would happen if they are not provided. This is also the logic from the KubeFlow codebase. I can test it now.

frontend/src/concepts/pipelines/topology/parseUtils.ts
jpuzz0: nit: The default case is not tested, but the bulk was done here so not a blocker.

frontend/src/concepts/pipelines/apiHooks/mlmd/useMlmdContext.ts
jpuzz0: Ok, I changed it to a nitpick, so maybe a no-op. Otherwise your changes lgtm. Feel free to squash your commits.

DaoDaoNoCode: Ok, it's just
```
request.setTypeName(type);
request.setContextName(name);
```
from the mlmd library doesn't allow the `undefined` to be set, not sure what will happen if I pass an empty string to it if it doesn't exist. But definitely, we always need to deal with the null value.

jpuzz0 (approved):
/approve

DaoDaoNoCode:
/unhold
New sprint.

Gkrumbach07:
this looks good.

/approve

PR #2696, opened 2024-04-10
Fixes Image selection dropdown scroll issue

manaswinidas:
Closes: https://issues.redhat.com/browse/RHOAIENG-3412

## Description

Fixes the image selection dropdown scroll issue. Upgraded `SimpleDropdownSelect` component to PF5.

Screenshot 2024-04-10 at 7 23 06 PM

https://github.com/opendatahub-io/odh-dashboard/assets/22885912/3d8af954-c3cf-4d6c-8702-b14b5096b359

## How Has This Been Tested?

Try creating a workbench after creating many custom images using the `Notebook images` section in the `Settings` section in the Dashboard. You should see a scroll bar while trying to select images in the `Image selection`, which doesn't break the layout. 

## Test Impact

Fixed failing tests due to the PF5 migration. Added new tests for scrollable dropdown selection

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/__tests__/cypress/cypress/support/commands/application.ts
+++ b/frontend/src/__tests__/cypress/cypress/support/commands/application.ts
@@ -1,6 +1,5 @@
 import { MatcherOptions } from '@testing-library/cypress';
 import { Matcher, MatcherOptions as DTLMatcherOptions } from '@testing-library/dom';
-
 /* eslint-disable @typescript-eslint/no-namespace */
 declare global {
   namespace Cypress {
@@ -27,6 +26,12 @@ declare global {
        */
       findDropdownItem(name: string | RegExp): Cypress.Chainable<JQuery>;
 
+      /**
+       * Finds a patternfly dropdown item by data-testid, first opening the dropdown if not already opened.
+       *
+       * @param testId the name of the item
+       */
+      findDropdownItemByTestId(testId: string): Cypress.Chainable<JQuery>;
       /**
        * Finds a patternfly select option by first opening the select menu if not already opened.
        *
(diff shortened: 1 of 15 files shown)


lucferbux (changes requested):
Ok a couple of things here:

Have you tested wether this fix the issue? I would think it's still happening if we still have the css.

We should have coverage with mockup data that might be recreating the issue, we can talk about that async if you want.

Is this refactor gonna help other instances of this? Maybe we should just fix the issue of the height and do the refactor in another issue OR just do the refactor in all the instances, but it's not a great idea having both components.

dpanshug (approved):
/lgtm

manaswinidas:
@lucferbux Fixed all test failures arising due to the migration. Added a new test to test scrollable dropdown selection as requested. Even manually tested the views on the pages, seems to work fine. The CSS file you mentioned seems to fix some width issues instead of height and I think it's required coz the extra padding persists if we remove that CSS. This is ready for review.

frontend/src/__tests__/cypress/cypress/support/commands/application.ts
```diff
@@ -27,6 +26,12 @@ declare global {
        */
       findDropdownItem(name: string | RegExp): Cypress.Chainable<JQuery>;
 
+      /**
+       * Finds a patternfly dropdown item by data-testid, first opening the dropdown if not already opened.
+       *
+       * @param testId the name of the item
+       */
+      findByDropdownItemId(testId: string): Cypress.Chainable<JQuery>;
```
christianvogt, line 34:
```suggestion
      findDropdownItemByTestId(testId: string): Cypress.Chainable<JQuery>;
```

lucferbux (approved):
/lgtm

Had to check the part of `testId` prop, but I think that's the only solution we can go for now, since that's how patternfly recommends using the testing id.

PR #2703, opened 2024-04-11
Fix issue when selecting a pipeline without version on the clone run page

DaoDaoNoCode:
JIRA: https://issues.redhat.com/browse/RHOAIENG-4258

## Description

When switching to a pipeline that contains no versions in it, we always trigger the hook `useUpdatePipelineFormData` and set the version to the incorrect one. The solution is to add a check before setting it to make sure we are only doing this when setting the same pipeline but not switching to another.
Also, fix a small issue that could crash the page.

## How Has This Been Tested?

1. Create 2 pipelines (pipeline 1 and pipeline 2)
2. Delete the version in pipeline 2, make sure there is no version in it
3. Create a run using pipeline 1
4. Duplicate this run, and switch the pipeline selection to pipeline 2
5. Verify the version selector is disabled and showing "no versions available" in it

## Test Impact

Add some conditions in the cypress test to verify the change works.

## Request review criteria:

If you have UI changes: 


Gkrumbach07 (approved):
tested live and works. low impact -- very nice. good stuff.

/lgtm
/approve

PR #2557, opened 2024-03-04
RHOAIENG-3984 Use the proxy at Red Hat.

pilhuhn:
https://issues.redhat.com/browse/RHOAIENG-3984

## Description

Make segment connections go through the proxy at console.redhat.com

## How Has This Been Tested?

Look at Browsers network tab

## Test Impact

Should be none, as the tracking itself has not changed.

## Request review criteria:

Self checklist (all need to be checked):
- [X] The developer has manually tested the changes and verified that the changes work
- [X] Commits have been squashed into descriptive, self-contained units of work (e.g. 'WIP' and 'Implements feedback' style messages have been removed)
- [ ] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [X] The developer has added tests or explained why testing cannot be added (unit tests & storybook for related changes)

After the PR is posted & before it merges:
- [ ] The developer has tested their solution on a cluster by using the image produced by the PR to `main`

PR #2716, opened 2024-04-16
[RHOAIENG-3984] Make sure analytics.js is loaded via proxy

pilhuhn:
## Description

The previous setting (from #2557 ) was still loading the analytics.js directly from Segment.
The cdn setting in the options did not override this.
This is a followup on [RHOAIENG-3984](https://issues.redhat.com//browse/RHOAIENG-3984).

## How Has This Been Tested?

Manual testing in browser console without dev-mode.

Network calls should look like this:
Bildschirmfoto 2024-04-16 um 16 27 18

## Test Impact

Should be none, as the tracking itself has not changed.

## Request review criteria:

- [X]  Commits have been squashed into descriptive, self-contained units of work (e.g. 'WIP' and 'Implements feedback' style messages have been removed)

### DIFF
--- a/frontend/src/utilities/segmentIOUtils.tsx
+++ b/frontend/src/utilities/segmentIOUtils.tsx
@@ -96,7 +96,9 @@ export const initSegment = async (props: {
       const t = document.createElement('script');
       t.type = 'text/javascript';
       t.async = true;
-      t.src = `https://cdn.segment.com/analytics.js/v1/${encodeURIComponent(key)}/analytics.min.js`;
+      t.src = `https://console.redhat.com/connections/cdn/analytics.js/v1/${encodeURIComponent(
+        key,
+      )}/analytics.min.js`;
       const n = document.getElementsByTagName('script')[0];
       if (n.parentNode) {
         n.parentNode.insertBefore(t, n);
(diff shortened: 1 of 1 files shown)


pilhuhn:
Screenshot on situation before:
Bildschirmfoto 2024-04-16 um 16 47 47

and after:

Bildschirmfoto 2024-04-16 um 16 45 09

lucferbux (approved):
/lgtm

Had a private conversation with @pilhuhn to ensure the validity of the proxy in a working cluster

frontend/src/utilities/segmentIOUtils.tsx
```diff
@@ -96,7 +96,9 @@ export const initSegment = async (props: {
       const t = document.createElement('script');
       t.type = 'text/javascript';
       t.async = true;
-      t.src = `https://cdn.segment.com/analytics.js/v1/${encodeURIComponent(key)}/analytics.min.js`;
+      t.src = `https://console.redhat.com/connections/cdn/analytics.js/v1/${encodeURIComponent(
```
shalberd, line 99: this will not work if access to console.redhat.com is not possible directly. As is the case in any corporate setup ... i.e. HTTP_PROXY and HTTPS_PROXY env vars need to be set and honored by your connections here. 

https://docs.openshift.com/container-platform/4.14/networking/enable-cluster-wide-proxy.html#nw-proxy-configure-object_config-cluster-wide-proxy

https://docs.openshift.com/container-platform/4.9/rest_api/config_apis/proxy-config-openshift-io-v1.html#apisconfig-openshift-iov1proxies



So that, from the proxy, if allowed, access continues to console.redhat.com.

Can this functionality be disabled? If not, this would in its current form be a breaking change. I am not talking public cloud and commercial Red Hat, I am talking ODH.

PR #2727, opened 2024-04-18
add support to run tests with different user permissions

christianvogt:
https://issues.redhat.com/browse/RHOAIENG-5259

## Description

Added utility to run tests as users with various permissions:
- disallowed user: no access to load dashboard
- cluster admin: full access to cluster
- project admin: full access specific projects
- edit user: full access to specific projects without ability to create role bindings

Optionally a user may also be a self provisioner (true by default). Non self provisioners cannot create new projects.

Default user for all tests is now a project admin.

Added some small tests that use the various users as well as updated breaking tests.

## How Has This Been Tested?

Run cypress tests.

## Test Impact

N/A

## Request review criteria:

If you have UI changes: 


frontend/src/__tests__/cypress/cypress/e2e/Application.cy.ts
```diff
@@ -0,0 +1,38 @@
...
+  });
+
+  it('should not show Settings nav section for non-cluster admins', () => {
+    asProjectAdminUser();
+    appChrome.visit();
+    appChrome.findNavSection('Settings').should('not.exist');
+  });
+
+  it('should show Settings nav section for cluster admins', () => {
+    asClusterAdminUser();
```
andrewballantyne, line 23: Can I suggest we either have a test for "cluster admin" or "product admin" -- Cluster Admin sounds like `cluster-admin` and that's technically different than just the admins of our product.

frontend/src/__tests__/cypress/cypress/utils/users.ts
```diff
@@ -0,0 +1,75 @@
...
+ */
+export const asDisallowedUser = (): void => {
+  cy.interceptOdh('GET /api/status', mockStatus({ isAllowed: false }));
+  interceptSelfSubjectAccessReview();
+};
+
+/**
+ * Cluster admins have access to perform any action on resource in every project.
+ */
+export const asClusterAdminUser = (): void => {
```
andrewballantyne, line 48: Ah I see -- you're aiming explicitly at `cluster-admin` -- recommend adding `asProductAdmin` as well. Which should not grant access to any project but will the settings stuff. Not sure if that's possible with this current infra atm.

frontend/src/__tests__/cypress/cypress/e2e/Application.cy.ts
christianvogt: ah yes. 
Having a distinct difference between product admin vs cluster admin is key. In fact almost all usage of cluster admin I did should be product admin.

frontend/src/__tests__/cypress/cypress/utils/users.ts
christianvogt: I will add `asProductAdmin` as suggested.

christianvogt:
Rebased.
Added user `asProductAdminUser`.
Moved core intercept initialization for k8s objects from `e2e.ts` to `users.ts`. Expectation now is that tests must establish a user before applying their own test specific intercepts as default intercepts are user specific. `e2e.ts` will still establish the default user as a project admin.

andrewballantyne (approved):

PR #2728, opened 2024-04-19
added user management cypress test

pnaik1:
Closes: https://issues.redhat.com/browse/RHOAIENG-3813
Closes: https://issues.redhat.com/browse/RHOAIENG-5310

## Description
added cypress test for user management

## How Has This Been Tested?
npm run test

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/__tests__/cypress/cypress/e2e/userManagement/userManagement.cy.ts
+++ b/frontend/src/__tests__/cypress/cypress/e2e/userManagement/userManagement.cy.ts
@@ -1,0 +1,67 @@
+import { mockGroupSettings } from '~/__mocks__/mockGroupConfig';
+import { userManagement } from '~/__tests__/cypress/cypress/pages/userManagement';
+import { asProductAdminUser, asProjectAdminUser } from '~/__tests__/cypress/cypress/utils/users';
+import { pageNotfound } from '~/__tests__/cypress/cypress/pages/pageNotFound';
+
+it('Cluster settings should not be available for non product admins', () => {
+  asProjectAdminUser();
+  userManagement.visit(false);
+  pageNotfound.findPage().should('exist');
+  userManagement.findNavItem().should('not.exist');
+});
+
+describe('User Management', () => {
+  beforeEach(() => {
+    asProductAdminUser();
+    cy.interceptOdh('GET /api/groups-config', mockGroupSettings());
+    userManagement.visit();
+  });
+
+  it('Administrator group setting', () => {
+    const administratorGroupSection = userManagement.getAdministratorGroupSection();
+    userManagement.findSubmitButton().should('be.disabled');
+    administratorGroupSection.findChipItem(/^odh-admins$/).should('exist');
+    administratorGroupSection.shouldHaveAdministratorGroupInfo();
+    administratorGroupSection.clearMultiChipItem();
+    administratorGroupSection.selectMultiGroup('odh-admins');
+    administratorGroupSection.findMultiGroupInput().type('odh-admin');
+    administratorGroupSection.findMultiGroupOptions('odh-admins-1').click();
+    administratorGroupSection.removeChipItem('odh-admins');
+    administratorGroupSection.findChipItem(/^odh-admins$/).should('not.exist');
+    administratorGroupSection.removeChipItem('odh-admins-1');
+    administratorGroupSection.findErrorText().should('exist');
+    administratorGroupSection.findMultiGroupOptions('odh-admins').click();
+    administratorGroupSection.findErrorText().should('not.exist');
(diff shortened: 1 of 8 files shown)


frontend/src/__tests__/cypress/cypress/support/commands/odh.ts
```diff
@@ -57,6 +58,16 @@ declare global {
         response?: OdhResponse,
       ): Cypress.Chainable<null>;
 
+      interceptOdh(
+        type: 'GET /api/groups-config',
+        response?: OdhResponse<GroupsConfig>,
```
christianvogt, line 63: I think I made a mistake with the accelerator endpoints using `response?`. I believe responses should be required. Can you also update other 3.
```suggestion
        response: OdhResponse<GroupsConfig>,
```

frontend/src/__tests__/cypress/cypress/pages/userManagement.ts
```diff
@@ -0,0 +1,68 @@
...
+    return this.find().findByTestId('multi-group-selection').findByRole('option', { name });
+  }
+
+  private findChipGroup() {
+    return this.find().findByRole('list', { name: 'Chip group category' });
+  }
+
+  findChipItem() {
+    return this.findChipGroup().find('li');
+  }
```
christianvogt, line 27: This would be more useful as a parameterized function that attempts to locate the specific item by name. That way `exists` and `not.exists` checks can be chained.

frontend/src/__tests__/cypress/cypress/e2e/userManagement/userManagement.cy.ts
```diff
@@ -0,0 +1,51 @@
...
+    const userGroupSection = userManagement.getUserGroupSection();
+    userManagement.findSubmitButton().should('be.disabled');
+    userGroupSection.findErrorText().should('exist');
+    userGroupSection.selectMultiGroup('odh-admins');
+    userGroupSection.findMultiGroupOptions('system:authenticated').click();
+    userGroupSection.findChipItem('odh-admins').should('exist');
+    userGroupSection.removeChipItem('odh-admins');
+    userManagement.findSubmitButton().should('be.enabled');
+    userManagement.findSubmitButton().click();
+    cy.wait('@saveGroupSetting').then((interception) => {
```
manaswinidas, line 38: Is it possible to check whether the alert is visible after this `@saveGroupSetting` is successfully executed? I believe there should be something to check whether this operation is completed. Correct me if I'm wrong.

manaswinidas (changes requested):

frontend/src/__tests__/cypress/cypress/e2e/userManagement/userManagement.cy.ts
```diff
@@ -0,0 +1,62 @@
...
+  userManagement.visit(false);
+  pageNotfound.findPage().should('exist');
+  userManagement.findNavItem().should('not.exist');
+});
+
+describe('User Management', () => {
+  beforeEach(() => {
+    asProductAdminUser();
+    cy.interceptOdh('GET /api/groups-config', mockGroupSettings());
+    cy.interceptOdh('PUT /api/groups-config', mockGroupSettings()).as('saveGroupSetting');
```
andrewballantyne, line 17: you only use this in one test, is there a reason it's run before each?

frontend/src/__tests__/cypress/cypress/e2e/userManagement/userManagement.cy.ts
```diff
@@ -0,0 +1,62 @@
...
+    administratorGroupSection.findErrorText().should('exist');
+    administratorGroupSection.findMultiGroupOptions('odh-admins').click();
+    administratorGroupSection.findErrorText().should('not.exist');
+    userManagement.findSubmitButton().should('be.disabled');
+  });
+
+  it('User group setting', () => {
+    const userGroupSection = userManagement.getUserGroupSection();
+    userManagement.findSubmitButton().should('be.disabled');
+    userGroupSection.findErrorText().should('exist');
```
andrewballantyne, line 41: This is an invalid state -- which is fine to test, but you should test success states that the App goes through.

andrewballantyne (changes requested):
You stumbled upon an invalid state -- which IS possible for the page to load in that state. We should have a test for an invalid admin/allowed group to make sure the page is not submittable when loaded in that state.

However, your mock is starting in an invalid state, and I think you need to make it start in a valid state and then manipulate it into an invalid state (using props to the mock or modifying the mock after it is returned from the function).

andrewballantyne (approved):
Not quite what I asked for, but covers the needed tests.

PR #2731, opened 2024-04-19
Use project display names in DW, move related utils from `~/pages/projects` to `~/concepts/projects` and `~/concepts/k8s`

mturley:
Closes [RHOAIENG-5291](https://issues.redhat.com/browse/RHOAIENG-5291)

## Description

* Moves a group of utilities out of `~/pages/projects` into sensible places in `~/concepts` so we're not introducing more import pollution
* Adds missing unit tests for the moved utils
* Uses `getProjectDisplayName` from its new home to expose a `projectDisplayName` value in DistributedWorkloadsContext
* Uses `projectDisplayName` instead of `namespace` in the "Requested resources" charts, which is the only remaining place we were rendering a namespace in Distributed Workload Metrics after Summit design iterations

Screenshot 2024-04-19 at 12 43 45 PM

## How Has This Been Tested?

Made sure updated imports don't cause any type errors, sanity-checked that other pages are still working, and verified that a project display name is now appearing in the chart

## Test Impact

Added new unit tests for each of the 8 utility functions being moved

## Request review criteria:

If you have UI changes: 


gitdallas (approved):

frontend/src/concepts/distributedWorkloads/DistributedWorkloadsContext.tsx
```diff
@@ -25,6 +26,7 @@ type DistributedWorkloadsContextType = {
   projectCurrentMetrics: DWProjectCurrentMetrics;
   refreshAllData: () => void;
   namespace?: string;
+  projectDisplayName?: string;
```
andrewballantyne, line 29: Not optional -- you can always give an answer

frontend/src/concepts/distributedWorkloads/DistributedWorkloadsContext.tsx
```diff
@@ -109,6 +111,7 @@ export const DistributedWorkloadsContextProvider =
           projectCurrentMetrics,
           refreshAllData,
           namespace,
+          projectDisplayName: project ? getProjectDisplayName(project) : namespace,
```
andrewballantyne, line 114: Namespace and Project are synonyms as far as the UI is concerned. This is very weird logic.

If you cannot find a project... how do you have a namespace? You shouldn't render your context if you don't have the bare min.  

So an error if you're without the min required for anything to work.

I get this is not a change you're doing with this PR -- but this exposed a gap in your code you need to close with this PR or do another PR and close it there and put this on top. This blocks this from merging.

frontend/src/concepts/projects/utils.ts
```diff
@@ -7,3 +10,12 @@ export const isAvailableProject = (projectName: string, dashboardNamespace: stri
     projectName === 'openshift' ||
     projectName === dashboardNamespace
   );
+
+export const getProjectDisplayName = (project: ProjectKind): string =>
+  getDisplayNameFromK8sResource(project);
+export const getProjectDescription = (project: ProjectKind): string =>
+  getDescriptionFromK8sResource(project);
```
andrewballantyne, line 17: Not your fault this is just refactoring -- but these may be short lived. We started with not being able to get direct access to the base level `getDisplayNameFromK8sResource` -- and now it seems to have mixed usage. 

https://issues.redhat.com/browse/RHOAIENG-6359 will address this going forward.

frontend/src/pages/distributedWorkloads/global/projectMetrics/sections/RequestedResources.tsx
```diff
@@ -58,7 +58,7 @@ const RequestedResourcesBulletChart: React.FC<RequestedResourcesBulletChartProps
   });
 
   const requestedByThisProjectData = getDataItem({
-    name: `Requested by ${namespace}`,
+    name: `Requested by ${projectDisplayName}`,
```
andrewballantyne, line 61: In your current state this is optional -- this is something you should have checked for. We need to make sure we trust our types 100% and there is no ambiguity. I wonder if we can get a linter for this 🤔 no `undefined` in string construction.

Logged this https://issues.redhat.com/browse/RHOAIENG-6361 -- you're going to make yours required, you don't need to worry about this per se.

andrewballantyne (changes requested):
Great refactor, thanks for taking this on. However, your implementation has exposed an issue in your context. This will block this PR. See comments. Let me know if you have any questions.

mturley:
Good feedback, thanks @andrewballantyne. Will fix

frontend/src/concepts/distributedWorkloads/DistributedWorkloadsContext.tsx
mturley: I think the reason for the fallback here was strictly a TypeScript one: `project` can be null because it comes from a `projects.find()` and TS doesn't know that we will definitely always find a project. I could use a `!` here to assert it's always defined, but then what if there is some error loading projects from the API and we really do have a null value? suddenly we have a null value at run time passed down for a statically non-null type. .... although thinking about it, if we have no projects this whole page won't render anyway, so it's moot. Maybe a `!` assertion up where `project` is defined on line 53 is the way to go?
```ts
const project = projects.find(byName(namespace))!;
```

frontend/src/concepts/distributedWorkloads/DistributedWorkloadsContext.tsx
mturley: Ah --- I remember now why these were optional to begin with. In the initial design for Distributed Workload Metrics we had our Cluster Metrics page which was not namespaced, so the context actually legitimately could have no namespace selected. Hopefully this is the last vestige of that we missed.

andrewballantyne (approved):
Addressed my only concern about the awkward logic in the context.

One parting question (for another PR) @mturley what happens if you put in an invalid namespace into your context (I assume this comes from the URL at some point)

mturley:
@andrewballantyne if the URL has an invalid namespace on this page, you'll hit this case before the context is rendered and it'll render an "invalid project" message: https://github.com/opendatahub-io/odh-dashboard/blob/main/frontend%2Fsrc%2Fpages%2FdistributedWorkloads%2Fglobal%2FGlobalDistributedWorkloads.tsx#L43-L55

If in the future someone renders the context provider elsewhere and gives it an invalid namespace, I believe the fetch hooks would hit 404 errors and you'd have a defined `fetchError` which would render that first encountered 404 error message (ClusterQueues probably) without trying to render children. Maybe it's worth having an "Invalid project" case there too, but we won't hit that with the code as it is now.

PR #2744, opened 2024-04-24
[RHOAIENG-2977] Artifacts table

jpuzz0:
Closes: [RHOAIENG-2977](https://issues.redhat.com/browse/RHOAIENG-2977)

## Description
Artifacts table, filter toolbar, pagination. 

image
(cc @yannnz)

## Demos
**Filtering**
https://github.com/opendatahub-io/odh-dashboard/assets/96431149/53a04653-b833-49ec-94ec-70d7bcc666a1

**Pagination**
https://github.com/opendatahub-io/odh-dashboard/assets/96431149/a532700a-2edd-4288-a013-9731fc5eef89

## Testing instructions
Navigate to the Artifacts page from the left nav. Using a project with multiple artifacts, verify pagination, and toolbar filtering works as expected.

## How Has This Been Tested?
Cypress does not support intercepting `application/grpc-web+proto` content type requests.

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/concepts/pipelines/context/MlmdListContext.tsx
+++ b/frontend/src/concepts/pipelines/context/MlmdListContext.tsx
@@ -1,0 +1,61 @@
+import React from 'react';
+
+interface MlmdOrderBy {
+  field: string;
+  direction: 'asc' | 'desc';
+}
+
+interface MlmdListContextProps {
+  filterQuery: string | undefined;
+  pageToken: string | undefined;
+  maxResultSize: number;
+  orderBy: MlmdOrderBy | undefined;
+  setFilterQuery: (filterQuery: string | undefined) => void;
+  setPageToken: (pageToken: string | undefined) => void;
+  setMaxResultSize: (maxResultSize: number) => void;
+  setOrderBy: (orderBy: MlmdOrderBy | undefined) => void;
+}
+
+const MlmdListContext = React.createContext({} as MlmdListContextProps);
+
+export const MlmdListContextProvider: React.FC<React.PropsWithChildren> = ({ children }) => {
+  const [filterQuery, setFilterQuery] = React.useState<string>();
+  const [pageToken, setPageToken] = React.useState<string>();
+  const [maxResultSize, setMaxResultSize] = React.useState(10);
+  const [orderBy, setOrderBy] = React.useState<MlmdOrderBy>();
+  const value = React.useMemo(
+    () => ({
+      filterQuery,
+      pageToken,
+      maxResultSize,
+      orderBy,
+      setFilterQuery,
+      setPageToken,
+      setMaxResultSize,
+      setOrderBy,
+    }),
+    [filterQuery, maxResultSize, orderBy, pageToken],
+  );
+
+  return <MlmdListContext.Provider value={value}>{children}</MlmdListContext.Provider>;
+};
+
+export const useMlmdListContext = (nextPageToken?: string): MlmdListContextProps => {
+  // https://github.com/patternfly/patternfly-react/issues/10312
+  // Force disabled state to pagination when there is no nextPageToken
+  React.useEffect(() => {
+    const paginationNextButtons = document.querySelectorAll('button[aria-label="Go to next page"]');
+
+    if (paginationNextButtons.length > 0) {
+      paginationNextButtons.forEach((button) => {
+        if (!nextPageToken) {
+          button.setAttribute('disabled', '');
+        } else {
+          button.removeAttribute('disabled');
+        }
+      });
+    }
+  }, [nextPageToken]);
+
+  return React.useContext(MlmdListContext);
+};
--- a/frontend/src/pages/pipelines/global/experiments/artifacts/ArtifactsTable.tsx
+++ b/frontend/src/pages/pipelines/global/experiments/artifacts/ArtifactsTable.tsx
@@ -1,0 +1,200 @@
+import React from 'react';
+import { Link } from 'react-router-dom';
+
+import { Flex, FlexItem, TextInput, Truncate } from '@patternfly/react-core';
+import { ExternalLinkAltIcon } from '@patternfly/react-icons';
+import { TableVariant, Td, Tr } from '@patternfly/react-table';
+
+import { Artifact } from '~/third_party/mlmd';
+import { TableBase } from '~/components/table';
+import DashboardEmptyTableView from '~/concepts/dashboard/DashboardEmptyTableView';
+import PipelinesTableRowTime from '~/concepts/pipelines/content/tables/PipelinesTableRowTime';
+import { FilterToolbar } from '~/concepts/pipelines/content/tables/PipelineFilterBar';
+import SimpleDropdownSelect from '~/components/SimpleDropdownSelect';
+import { ArtifactType } from '~/concepts/pipelines/kfTypes';
+import { useMlmdListContext } from '~/concepts/pipelines/context';
+import { FilterOptions, columns, initialFilterData, options } from './constants';
+
+interface ArtifactsTableProps {
+  artifacts: Artifact[] | null | undefined;
+  nextPageToken: string | undefined;
+  isLoaded: boolean;
+}
+
+export const ArtifactsTable: React.FC<ArtifactsTableProps> = ({
+  artifacts,
+  isLoaded,
+  nextPageToken,
+}) => {
+  const {
+    maxResultSize,
+    setFilterQuery,
+    setPageToken: setRequestToken,
+    setMaxResultSize,
+  } = useMlmdListContext(nextPageToken);
+  const [page, setPage] = React.useState(1);
+  const [filterData, setFilterData] = React.useState(initialFilterData);
+  const onClearFilters = React.useCallback(() => setFilterData(initialFilterData), []);
+  const [pageTokens, setPageTokens] = React.useState<Record<number, string>>({});
+
+  const onFilterUpdate = React.useCallback(
+    (key: string, value: string | { label: string; value: string } | undefined) =>
+      setFilterData((prevValues) => ({ ...prevValues, [key]: value })),
+    [],
+  );
+
+  const onNextPageClick = React.useCallback(
+    (_: React.SyntheticEvent<HTMLButtonElement>, nextPage: number) => {
+      if (nextPageToken) {
+        setPageTokens((prevTokens) => ({ ...prevTokens, [nextPage]: nextPageToken }));
+        setRequestToken(nextPageToken);
+        setPage(nextPage);
+      }
+    },
+    [nextPageToken, setRequestToken],
+  );
+
+  const onPrevPageClick = React.useCallback(
+    (_: React.SyntheticEvent<HTMLButtonElement>, prevPage: number) => {
+      if (pageTokens[prevPage]) {
+        setRequestToken(pageTokens[prevPage]);
+        setPage(prevPage);
+      } else {
+        setRequestToken(undefined);
+      }
+    },
+    [pageTokens, setRequestToken],
+  );
+
+  React.useEffect(() => {
+    if (Object.values(filterData).some((filterOption) => !!filterOption)) {
+      let filterQuery = '';
+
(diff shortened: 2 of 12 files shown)


frontend/src/pages/pipelines/global/experiments/artifacts/useGetArtifactsList.ts
```diff
@@ -2,17 +2,47 @@ import React from 'react';
...
+    }
+
+    listOperationOptions.setMaxResultSize(maxResultSize ?? defaultMaxResultSize);
+    request.setOptions(listOperationOptions);
+
+    const response = await metadataStoreServiceClient.getArtifacts(request);
+    const nextPageToken = response.getNextPageToken();
+    listOperationOptions.setNextPageToken(nextPageToken);
+
+    return { artifacts: response.toObject().artifactsList, nextPageToken };
```
DaoDaoNoCode, line 44: You can use `response.getArtifactsList()` instead of the conversion by yourself.

frontend/src/pages/pipelines/global/experiments/artifacts/useGetArtifactsList.ts
```diff
@@ -2,17 +2,47 @@ import React from 'react';
...
+    if (pageToken) {
+      listOperationOptions.setNextPageToken(pageToken);
+    }
+
+    listOperationOptions.setMaxResultSize(maxResultSize ?? defaultMaxResultSize);
+    request.setOptions(listOperationOptions);
+
+    const response = await metadataStoreServiceClient.getArtifacts(request);
+    const nextPageToken = response.getNextPageToken();
+    listOperationOptions.setNextPageToken(nextPageToken);
```
DaoDaoNoCode, line 42: Is the setting here needed?

frontend/src/pages/pipelines/global/experiments/artifacts/useGetArtifactsList.ts
jpuzz0: They're not the same. That's a different set of data.

frontend/src/pages/pipelines/global/experiments/artifacts/useGetArtifactsList.ts
jpuzz0: It is not needed, I will remove it.

frontend/src/pages/pipelines/global/experiments/artifacts/useGetArtifactsList.ts
DaoDaoNoCode: Wow, what's the difference because I am using `response.getExecutionList()` to get the executions

DaoDaoNoCode: Yes, because I think the intention of the KubeFlow MLMD class is to let you use the method in the class, though it's not documented anywhere. So if you are using `response.getArtifactsList()`, then `artifact.getCreateTimeSinceEpoch()` should give you the same result compared to this line https://github.com/opendatahub-io/odh-dashboard/pull/2744/files#diff-a85ef6d786c1a8509e52597a67c0273d4302a82018ad211b237f1d45bb1f82baR180

jpuzz0: Are you claiming the theoretical intention of the MLMD class and the accessing of values via method call is better in some way than just accessing from the object keys?

DaoDaoNoCode: Yes, because this is how they do in the codebase. Also, it matches the type `Artifact` instead of type `Artifact.AsObject`

jpuzz0: How is it better? They expose AsObject for a reason. Just because they're not using it in their code is not justification for it being better. Unless there's a good reason, I'll be leaving as-is. The consumption of data in the table is more straightforward this way.

frontend/src/pages/pipelines/global/experiments/MlmdListContext.tsx
DaoDaoNoCode, line 1: It's better to put this file under `/src/concepts/pipelines/context/`, what do you think? It's not an experiment-specific stuff.

jpuzz0: Sure, I can move it.

frontend/src/pages/pipelines/global/experiments/artifacts/ArtifactsTable.tsx
```diff
@@ -0,0 +1,213 @@
...
+    const paginationNextButtons = document.querySelectorAll('button[aria-label="Go to next page"]');
+
+    paginationNextButtons.forEach((button) => {
+      if (!nextPageToken) {
+        button.setAttribute('disabled', '');
+      } else {
+        button.removeAttribute('disabled');
+      }
+    });
+  }, [nextPageToken]);
```
DaoDaoNoCode, line 51: Can you make this a hook file like `useDisablePaginationNextButtons(nextPageToken)`? It would be easier to reuse it in the execution table as well.

jpuzz0: I don't think that's necessary and that's a confusing name, where its also only applicable to MLMD list tables.

jpuzz0: I moved the logic inside of `useMlmdListContext` so you get it for free with executions.

DaoDaoNoCode (approved):
Tested, works well.
/lgtm

frontend/src/pages/pipelines/global/experiments/artifacts/useGetArtifactsList.ts
Gkrumbach07: MLMD generated functions exposes it, not KF. We should use the full functionality of the mlmd resource types which include more than just the data, there are other functions we would not have if we convert the artifact into an object right away. We may not use them know, but maybe in the future we will. I would suggest we keep the main hooks returning the base type, and then further down into implementation, we we can choose to convert that artifact into an object.

frontend/src/concepts/pipelines/context/MlmdListContext.tsx
```diff
@@ -0,0 +1,60 @@
...
+  orderBy: MlmdOrderBy | undefined;
+  setFilterQuery: (filterQuery: string | undefined) => void;
+  setPageToken: (pageToken: string | undefined) => void;
+  setMaxResultSize: (maxResultSize: number) => void;
+  setOrderBy: (orderBy: MlmdOrderBy | undefined) => void;
+}
+
+const MlmdListContext = React.createContext({} as MlmdListContextProps);
+
+export const MlmdListContextProvider: React.FC<React.PropsWithChildren> = ({ children }) => {
```
Gkrumbach07, line 21: Why do we need a context provider for just the table?

frontend/src/pages/pipelines/global/experiments/artifacts/ArtifactsTable.tsx
```diff
@@ -0,0 +1,200 @@
...
+
+      if (filterData[FilterOptions.Type]) {
+        const artifactTypeQuery = `type LIKE '%${filterData[FilterOptions.Type]}%'`;
+        filterQuery += filterQuery.length ? ` AND ${artifactTypeQuery}` : artifactTypeQuery;
+      }
+
+      setFilterQuery(filterQuery);
+    } else {
+      setFilterQuery('');
+    }
```
Gkrumbach07, line 93: can this filter logic be abstracted so the executions can reuse this?

DaoDaoNoCode: I can do this in my PR after this is merged.

frontend/src/concepts/pipelines/context/MlmdListContext.tsx
```diff
@@ -0,0 +1,60 @@
...
+    [filterQuery, maxResultSize, orderBy, pageToken],
+  );
+
+  return <MlmdListContext.Provider value={value}>{children}</MlmdListContext.Provider>;
+};
+
+export const useMlmdListContext = (nextPageToken?: string): MlmdListContextProps => {
+  // Force disabled state to pagination when there is no nextPageToken
+  React.useEffect(() => {
+    const paginationNextButtons = document.querySelectorAll('button[aria-label="Go to next page"]');
```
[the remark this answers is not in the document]
jpuzz0: The pagination arrow has no exposed disabled handler, and due to the way data is only fetched per page for MLMD resources, this was necessary to disable the next arrow on the last page.

frontend/src/concepts/pipelines/context/MlmdListContext.tsx
jpuzz0: I added the context so that Juntao can use it for Executions as well. It would be for any list table using MLMD resources.

frontend/src/pages/pipelines/global/experiments/artifacts/useGetArtifactsList.ts
jpuzz0: Why would we want anything other than the data here? Is that not what every other API fetching hook we have does?

frontend/src/concepts/pipelines/context/MlmdListContext.tsx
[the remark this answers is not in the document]
Gkrumbach07: can you do something where once you have a next page token, you set the item count to some defiend value. then it will disable on its own 

for ref this is how the page size is calculated
```tsx
 // when itemCount is not known let's set lastPage as page+1 as we don't know the total count
itemCount || itemCount === 0 ? Math.ceil(itemCount / perPage) || 0 : page + 1;
```

frontend/src/pages/pipelines/global/experiments/artifacts/useGetArtifactsList.ts
Gkrumbach07: I am trying to keep consistency across the feature. If we are using one type everywhere else, but then another version of the type in one other spot, it can get confusing. And you can still use that object further down in implementation.

jpuzz0: Where would we use Artifact as a non-object form and why?

jpuzz0: I would like to know what the answers to these questions are, but since this seems to be blocking the PR, I went ahead and updated my hook to return the list in a non-object form, and am just having to iterate over all artifacts now for the table, which is more code than what was there prior.

frontend/src/concepts/pipelines/context/MlmdListContext.tsx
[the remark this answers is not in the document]
Gkrumbach07: OK in that case, can you log a PF issue and add that too the code as a comment please. This really should not be a way we do it, but if there is no other option then its fine but should be tracked

frontend/src/concepts/pipelines/context/MlmdListContext.tsx
Gkrumbach07: closing as convo is in slack

frontend/src/pages/pipelines/global/experiments/artifacts/useGetArtifactsList.ts
Gkrumbach07: We are not using it now, but for consistency across all of our other MLMD hooks

frontend/src/concepts/pipelines/context/MlmdListContext.tsx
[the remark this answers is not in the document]
jpuzz0: I created the issue and added a comment with a link to that PF issue.

Gkrumbach07:
all good to me.

/approve

DaoDaoNoCode (commented):
/lgtm

PR #2748, opened 2024-04-25
[RHOAIENG-5492] Landing page: admin section

jeff-phillips-18:
Closes: [RHOAIENG-5492](https://issues.redhat.com/browse/RHOAIENG-5492)

## Description
Adds an admin section to the landing page for those users who have admin privileges.
Fixes issue where the nav bar group items are not automatically expanded when navigating to a child section.

## How Has This Been Tested?
Tested locally by using admin role and non-admin role. Turned off different sections in the admin section to verify only the enabled sections are shown and if all are hidden then the entire section is not shown.

## Test Impact
Added e2e tests.

### Full access
image

### Full access collapsed
image

### 3 sections visible
image

### 2 sections visible
image

### 1 section visible
image

## Request review criteria:

/cc @jgiardino


frontend/src/concepts/design/CollapsibleSection.tsx
```diff
@@ -46,28 +44,21 @@ const CollapsibleSection: React.FC<CollapsibleSectionProps> = ({
             aria-expanded={open}
             isInline
             variant="link"
```
jgiardino, line 46: Is this an existing component within the product that we're reusing for this page?

When I review the latest implementation on the poc cluster, there are a few observations about the styles for this button:

### Button variant
Currently this button is use 
```
isInline
variant="link"
```
But for an icon button, the ideal variant is `variant="plain"` without the `isInline` modifier, similar to the icon button in this [PF react example](https://www.patternfly.org/components/button#variant-examples). 

### Spacing
I get why the `isInline` modifier is used to remove padding, and it makes sense to remove padding for this. If you remove that, the plain variant would add padding to the button. Another option for removing the left & right padding is to use a utility class `pf-v5-u-px-0`.

### Sizing
Ideally any icons that accompany text would have the same size as the text. You can do this by adding a utility class to match the size of the header. In this example, you would want it to be `pf-v5-u-font-size-2xl` but if the header size ever changes, then the size of this icon would need to be updated, too.

jeff-phillips-18: We do not import the pf utility classes

jeff-phillips-18: I will update as you suggest though but I'll use styles.

frontend/src/__tests__/cypress/cypress/e2e/home/homeAdmin.cy.ts
```diff
@@ -0,0 +1,118 @@
...
+
+describe('Home page Admin section', () => {
+  beforeEach(() => {
+    cy.interceptOdh('GET /api/docs', mockDocs());
+    cy.interceptOdh('GET /api/components', null, mockComponents());
+    cy.interceptOdh('GET /api/quickstarts', mockQuickStarts());
+  });
+  it('should show the admin section for admins', () => {
+    initHomeIntercepts({ disableHome: false, disableProjects: true });
+    setAdminUser(true);
```
andrewballantyne, line 16: Christian recently merged this change, I think that's what you're after here.
```suggestion
    asProductAdmin();
```

If I am right, can you change all `setAdminUser` function calls and remove the utility? (see [the user calls here](https://github.com/opendatahub-io/odh-dashboard/blob/main/frontend/src/__tests__/cypress/cypress/utils/users.ts))

frontend/src/__tests__/cypress/cypress/e2e/home/homeUtils.ts
```diff
@@ -13,3 +17,7 @@ export const initHomeIntercepts = (config: HandlersProps = {}): void => {
   };
   cy.interceptOdh('GET /api/config', mockDashboardConfig(dashboardConfig));
```
andrewballantyne, line 18: FWIW, you can just type your utility with the dashboardConfig settings for the mock... saving you the dupe effort for your type.

Since you're trying to mimic the first argument to this function, you can just set your function to take the same input.

```
export const initHomeIntercepts = (
  config: Parameters<typeof mockDashboardConfig>[0] = {},
): void => {
```

Or if you want, you can go one step further:
```
export const initHomeIntercepts = (...args: Parameters<typeof mockDashboardConfig>): void => {
  cy.interceptOdh('GET /api/config', mockDashboardConfig(...args));
};
```

andrewballantyne (changes requested):
Testing now, looking pretty good on the code front, just two points.

andrewballantyne (approved):

PR #2754, opened 2024-04-26
[RHOAIENG-6519] DW: Support workloads owned by either a RayCluster or a Job when querying CPU and Memory usage

mturley:
Fixes the **blocker** https://issues.redhat.com/browse/RHOAIENG-6519

## Description

We made the incorrect assumption that all Kueue Workload CRs will have an ownerReference to a Job CR which we can reference in a Prometheus query to get CPU and Memory usage for Jobs. In reality, when Kueue workloads are created from RayCluster, no Job CR gets created and the RayCluster itself is the owner. This meant our existing Prometheus query was not sufficient for getting CPU and memory usage for all workloads.

@Fiona-Waters and @KPostOffice, among others, worked out that we can use a different Prometheus query based on `kube_pod_owner` to include metrics from pods owned by both RayClusters and Jobs. This meant updating the UI logic so we index these usage values not just based on the job name, but based on the owner kind and owner name (in case there is a RayCluster and a Job with the same name).

This change adjusts logic that was based on job name to handle this extra dimension of first indexing by owner kind, and considers ownerReferences of a workload that are of either RayCluster or Job kind instead of just Jobs.

Screenshot 2024-04-26 at 4 49 02 PM

## How Has This Been Tested?

Tested in local dev env while connected to a QE cluster with running RayClusters and Jobs. Will test an image on a cluster when this PR builds one.

## Test Impact

Unit and cypress tests have been updated to cover the changes and make sure RayClusters and Jobs are both covered in the mock data.

## Request review criteria:

If you have UI changes: 


gitdallas (approved):

mturley:
Self-approving with @andrewballantyne's permission. This has been reviewed by @gitdallas and I've verified the fix on the cluster where the issue was initially identified by @Srihari1192

PR #2758, opened 2024-04-29
Add dscinitializations resource to cluster-role.yaml and handle missing dsci status in ServingRuntimeList.cy.ts

lucferbux:
Closes https://issues.redhat.com/browse/RHOAIENG-6578 

## Description

Add permissions to fetch dsci by the dashboard serviceaccount and fixes a flaw in the area enablement util method that returned true if the dsci was missing.

## How Has This Been Tested?

1. Run the changes in a cluster with dsci but no permissions
2. Use `npm run start:dev:ext` to get the deployed dashboard serviceaccount capabilities
3. Check that the kserve auth feature should be disabled
4. Now add the permissions to the cluster (see bellow)
5. Check that the feature should be enabled

## Request review criteria:

If you have UI changes: 

**Tested with npm run start:dev:ext which sholud mimic this same behavior**

### DIFF
--- a/frontend/src/__tests__/cypress/cypress/e2e/modelServing/ServingRuntimeList.cy.ts
+++ b/frontend/src/__tests__/cypress/cypress/e2e/modelServing/ServingRuntimeList.cy.ts
@@ -598,6 +598,7 @@ describe('Serving Runtime List', () => {
       initIntercepts({
         disableModelMeshConfig: false,
         disableKServeConfig: false,
+        disableKServeAuthConfig: true,
         servingRuntimes: [],
       });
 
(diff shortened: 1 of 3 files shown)


frontend/src/concepts/areas/utils.ts
```diff
@@ -72,18 +72,18 @@ export const isAreaAvailable = (
...
-            ...acc,
-            [capability]: dsciStatus.conditions.some(
-              (c) => c.type === capability && c.status === 'True',
-            ),
-          }),
-          {},
-        )
-      : null;
+  // If DSCI is not available, we shouldn't assume it has met the required capabilities
+  const requiredCapabilitiesState = requiredCapabilities
```
andrewballantyne, line 76: Why the change 🤔 We have this for DSC

lucferbux: This is part of the issue that i discovered, if DSCI returns a 404 (it's happening in a cluster cause it lacks the permission), it will later return true as if the condition is fulfilled.

lucferbux: We shouldn't get that state as dsci should be present every time, but we met this specific edge case @andrewballantyne

andrewballantyne: This is true of everything -- including the DSC -- this is not a good reason to change the logic on the frontend.

andrewballantyne: There is a reason we have the template "test on cluster" checkbox. Because of weird permission things like this.

lucferbux: > There is a reason we have the template "test on cluster" checkbox. Because of weird permission things like this.
Yes, and that was on me, I missed that cause when we shipped this the operator side wasn't even completed, so I didn't have the ability to test it on cluster, just mocking my own responses, and that's on me.

On the other hand, I've tested it right now and it seems that we can get the same result with `npm run start:dev:ext` which will give us much more flexibility to test permissions on clusters

lucferbux: I'm gonna remove the logic change, keep the tests and keep the permission.

andrewballantyne (approved):
Lucas tested with connecting to the Dashboard Pod and overriding permissions on cluster -- I'm approving code.

PR #2759, opened 2024-04-29
[RHOAIENG-6606] Correctly nav to application from Jupyter card

jeff-phillips-18:
Fixes: [RHOAIENG-6606](https://issues.redhat.com/browse/RHOAIENG-6606)

## Description
Update the link URL for the Jupyter card Launch application link to not be a relative path.

## How Has This Been Tested?
Tested with the home page disabled and enabled.

## Test Impact
Added an e2e test to check the launch application link with and without the home page feature enabled.

## Request review criteria:

### DIFF
--- a/frontend/src/components/OdhAppCard.tsx
+++ b/frontend/src/components/OdhAppCard.tsx
@@ -105,7 +105,11 @@ const OdhAppCard: React.FC<OdhAppCardProps> = ({ odhApp }) => {
     <CardFooter className="odh-card__footer">
       {odhApp.metadata.name === 'jupyter' ? (
         odhApp.spec.internalRoute ? (
-          <Link to={odhApp.spec.internalRoute} className="odh-card__footer__link">
+          <Link
+            data-testid="jupyter-app-link"
+            to="/notebookController"
+            className="odh-card__footer__link"
+          >
             Launch application
           </Link>
         ) : null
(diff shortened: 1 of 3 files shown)


frontend/src/components/OdhAppCard.tsx
```diff
@@ -105,7 +105,11 @@ const OdhAppCard: React.FC<OdhAppCardProps> = ({ odhApp }) => {
...
         odhApp.spec.internalRoute ? (
-          <Link to={odhApp.spec.internalRoute} className="odh-card__footer__link">
+          <Link
+            data-testid="jupyter-app-link"
+            to={`/${odhApp.spec.internalRoute}`}
+            className="odh-card__footer__link"
+          >
             Launch application
           </Link>
         ) : null
```
andrewballantyne, line 115: Weird legacy code... Lets just change this to `/notebookController` please (instead of using the spec property).

We have 5 other instances that hardcode the path. JH was a wild alternative back in the day and it needed more dynamic nature. This is an internal route, we should just sync it up imo.

andrewballantyne (changes requested):
One tiny change for organization, everything else is good to go. Works well, just I feel for maintenance we can clean this up a tad while we are here.

andrewballantyne (approved):

PR #2767, opened 2024-05-01
[RHOAIENG-5497] Enable the landing page

jeff-phillips-18:
Closes: [RHOAIENG-5497](https://issues.redhat.com/browse/RHOAIENG-5497)

## Description
Changes the default for `disableHome` to true in the dashboard config thus enabling the new Landing Page by default.

## How Has This Been Tested?
Tested locally

## Test Impact
Updated tests to verify that the default state shows the landing page.

## Request review criteria:


andrewballantyne:
/hold

Needs a review from PM before we enable it for the product.

frontend/src/__tests__/cypress/cypress/pages/enabled.ts
```diff
@@ -1,6 +1,6 @@
 class EnabledPage {
-  visit(homeEnabled = false) {
-    cy.visit(homeEnabled ? '/enabled' : '/');
+  visit() {
+    cy.visit('/enabled');
```
andrewballantyne, line 3: I don't understand why you changed this... the home page can still be disabled

jeff-phillips-18: I could add it back but it's not set in any of the callers.

andrewballantyne: You had a test to make sure the URL redirect was properly tested with and without the feature enabled. I'm feeling you had to have dropped something or we are missing something if it's no longer used.

This cannot be right I would think 🤔 Not sure where the gap is...

jeff-phillips-18: I changed the test to verify that `/` takes you to the enabled page if the flag is disabled:
https://github.com/opendatahub-io/odh-dashboard/pull/2767/files#diff-34ec9e23c634748dc1ed7b69552b385bf9a11a5a746b7e4779c48e3302f9699aR19

andrewballantyne (approved):

PR #2760, opened 2024-04-29
Model Registry: Initial content of model details tab with stub editable fields and tab navigation handling

mturley:
Part of [RHOAIENG-2235](https://issues.redhat.com/browse/RHOAIENG-2235)

## Description

Implements most of the "Details" tab of the registered model page. Note that the "Properties" section is not functional and will be implemented in a followup PR. Also note that while the Description and Labels sections have editable functionality implemented, the actual last step of persisting the edits to the API is a stub promise that simply waits 2 seconds in order to show the "saving" state (edits are not allowed while saving). Implementing the real API patch requests will be in a separate PR.

Model details tab:
Screenshot 2024-04-29 at 7 31 32 PM

Description field in edit mode:
Screenshot 2024-04-29 at 7 15 02 PM

Description field disabled while saving (currently a stub 2-second-delay promise):
Screenshot 2024-04-29 at 7 24 38 PM

Labels in edit mode:
Screenshot 2024-04-29 at 7 15 13 PM

Editing an existing label:
Screenshot 2024-04-29 at 7 15 21 PM

Add label modal:
Screenshot 2024-04-29 at 7 24 49 PM
Screenshot 2024-04-29 at 7 24 56 PM

Label editing disabled while saving (currently a stub 2-second-delay promise):
Screenshot 2024-04-29 at 7 25 53 PM

## How Has This Been Tested?

1. Go to `Model Registry` tab
2. Click on any registered model in the `Registered Models` table and then click the `Details` tab, or click the kebab menu at the right of a registered model and click `View details`
3. You will be taken to the model details view shown above
4. Click between the tabs and note that the URL route is changing and using the new `/details` and `/versions` subpaths

## Test Impact

Will be added shortly.

## Request review criteria:

Self checklist (all need to be checked):
- [x] The developer has manually tested the changes and verified that the changes work
- [x] Commits have been squashed into descriptive, self-contained units of work (e.g. 'WIP' and 'Implements feedback' style messages have been removed)
- [x] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [ ] The developer has added tests or explained why testing cannot be added (unit or cypress tests for related changes)

If you have UI changes: 

- [x] Included any necessary screenshots or gifs if it was a UI change.
- [ ] Included tags to the UX team if it was a UI/UX change (find relevant UX in the [SMEs](https://github.com/opendatahub-io/odh-dashboard/tree/main/docs/smes.md) section).

After the PR is posted & before it merges:
[...]

PR #2770, opened 2024-05-02
Model Details and Model Version Details: add editable properties section, set up API patches for description, labels and properties

mturley:
Closes [RHOAIENG-2235](https://issues.redhat.com/browse/RHOAIENG-2235) and [RHOAIENG-2238](https://issues.redhat.com/browse/RHOAIENG-2238)
Followup to https://github.com/opendatahub-io/odh-dashboard/pull/2760

## Description

* Implements the contents of the Properties section using a new `ModelPropertiesDescriptionListGroup` component which provides view/add/edit/delete controls
* Implements API patch requests when saving edits to Description, Labels and Properties (replacing the stub Promise for the former two)
* Adds input validation to ensure labels and property keys are unique. See **"Input validation screencaps"** below.
* None of these components/utils are specific to `RegisteredModel`, they work with `ModelRegistryCustomProperties` objects and generic `T extends ModelRegistryBase` objects, so it should be easy to reuse these for the ModelVersion detail page.
* Note that the patch requests must use the new `getPatchBody` utility in order to make sure the request isn't clearing other fields due to the way patches are performed in the API. We must supply all editable fields when patching an object even if we aren't modifying them all, but we can't supply some fields like the create/edit timestamps or we'll get 400 errors.

### Demo screencaps

* Editing the Description of a RegisteredModel

  https://github.com/opendatahub-io/odh-dashboard/assets/811963/08721aae-5897-4fa3-a00e-a7e7bee78ecb

* Editing the Labels of a RegisteredModel

  https://github.com/opendatahub-io/odh-dashboard/assets/811963/6bed2da9-99cc-4880-942b-e5e981cc4609

* Editing the Properties of a RegisteredModel

  https://github.com/opendatahub-io/odh-dashboard/assets/811963/4aba24ad-f97e-42b5-a49b-4be3432fc6cf

### Input validation screencaps
* When adding a new label, in addition to the existing validation that the label text must be no longer than 63 characters, we also now validate that a newly added label is unique (does not match any existing key in `customProperties`, whether that is a label or the key of a property in the table below). If the label is not unique, the modal submission is blocked and an error message appears:
  
  Screenshot 2024-05-02 at 3 44 56 PM
    
* When editing an existing label, the same validation is applied. However, with the [current beta version of PF editable labels](https://www.patternfly.org/components/label#editable-labels), there is no way to express input validation errors while editing a label's text. As a workaround for now, if the user attempts to use an invalid label (either longer than 63 characters or matching an existing label/property), when they click away the change is not applied and the label remains as it was before they clicked it:
  
  https://github.com/opendatahub-io/odh-dashboard/assets/811963/d88261bf-0743-4f36-9c41-2fe9a43964cc

* When adding or editing a property, the same validation is applied to the property key. Errors in this case appear inline below the key input field:
  
  Screenshot 2024-05-02 at 3 53 40 PM
  Screenshot 2024-05-02 at 3 52 26 PM

cc @yih-wang @vconzola 

## How Has This Been Tested?

Go to `Model Registry` in the nav, click a registered model and click the Details tab. Play around with the edit controls on the Description, Labels and Properties sections. (see the screen recordings above)

## Test Impact

Unit tests have been added for the new utility functions `mergeUpdatedLabels`, `getProperties`, `mergeUpdatedProperty` and `getPatchBody`. Cypress tests for all these views are to come soon.

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/components/EditableLabelsDescriptionListGroup.tsx
+++ b/frontend/src/components/EditableLabelsDescriptionListGroup.tsx
@@ -20,7 +20,8 @@ type EditableTextDescriptionListGroupProps = Partial<
   Pick<DashboardDescriptionListGroupProps, 'title' | 'contentWhenEmpty'>
 > & {
   labels: string[];
-  saveEditedLabels: (labels: string[]) => Promise<void>;
+  saveEditedLabels: (labels: string[]) => Promise<unknown>;
+  allExistingKeys?: string[];
 };
 
 const EditableLabelsDescriptionListGroup: React.FC<EditableTextDescriptionListGroupProps> = ({
@@ -28,6 +29,7 @@ const EditableLabelsDescriptionListGroup: React.FC<EditableTextDescriptionListGr
   contentWhenEmpty = 'No labels',
   labels,
   saveEditedLabels,
+  allExistingKeys = labels,
 }) => {
   const [isEditing, setIsEditing] = React.useState(false);
   const [unsavedLabels, setUnsavedLabels] = React.useState(labels);
@@ -54,10 +56,22 @@ const EditableLabelsDescriptionListGroup: React.FC<EditableTextDescriptionListGr
     setUnsavedLabels([...unsavedLabels, text]);
   };
 
+  // Don't allow a label that matches a non-label property key or another label (as they stand before saving)
+  // Note that this means if you remove a label and add it back before saving, that is valid
+  const reservedKeys = [
+    ...allExistingKeys.filter((key) => !labels.includes(key)),
+    ...unsavedLabels,
+  ];
+
   const [isAddLabelModalOpen, setIsAddLabelModalOpen] = React.useState(false);
   const [addLabelInputValue, setAddLabelInputValue] = React.useState('');
   const addLabelInputRef = React.useRef<HTMLInputElement>(null);
-  const addLabelInputTooLong = addLabelInputValue.length > 63;
+  let addLabelValidationError: string | null = null;
+  if (reservedKeys.includes(addLabelInputValue)) {
+    addLabelValidationError = 'Label must not match an existing label or property key';
+  } else if (addLabelInputValue.length > 63) {
+    addLabelValidationError = "Label text can't exceed 63 characters";
+  }
 
   const toggleAddLabelModal = () => {
     setAddLabelInputValue('');
@@ -68,7 +82,8 @@ const EditableLabelsDescriptionListGroup: React.FC<EditableTextDescriptionListGr
       addLabelInputRef.current.focus();
     }
   }, [isAddLabelModalOpen]);
-  const addLabelModalSubmitDisabled = !addLabelInputValue || addLabelInputTooLong;
+
+  const addLabelModalSubmitDisabled = !addLabelInputValue || !!addLabelValidationError;
   const submitAddLabelModal = (event?: React.FormEvent) => {
     event?.preventDefault();
     if (!addLabelModalSubmitDisabled) {
@@ -107,7 +122,11 @@ const EditableLabelsDescriptionListGroup: React.FC<EditableTextDescriptionListGr
                 editableProps={{ 'aria-label': `Editable label with text ${label}` }}
                 onClose={() => removeUnsavedLabel(label)}
                 closeBtnProps={{ isDisabled: isSavingEdits }}
-                onEditComplete={(_event, newText) => editUnsavedLabel(newText, index)}
+                onEditComplete={(_event, newText) => {
+                  if (!reservedKeys.includes(newText) && newText.length <= 63) {
+                    editUnsavedLabel(newText, index);
+                  }
+                }}
               >
                 {label}
               </Label>
@@ -172,13 +191,13 @@ const EditableLabelsDescriptionListGroup: React.FC<EditableTextDescriptionListGr
               }
               ref={addLabelInputRef}
               isRequired
-              validated={addLabelInputTooLong ? 'error' : 'default'}
+              validated={addLabelValidationError ? 'error' : 'default'}
             />
-            {addLabelInputTooLong && (
+            {addLabelValidationError && (
               <FormHelperText>
                 <HelperText>
                   <HelperTextItem icon={<ExclamationCircleIcon />} variant="error">
-                    Label text can&apos;t exceed 63 characters
+                    {addLabelValidationError}
                   </HelperTextItem>
                 </HelperText>
               </FormHelperText>
--- a/frontend/src/concepts/modelRegistry/types.ts
+++ b/frontend/src/concepts/modelRegistry/types.ts
@@ -35,6 +35,57 @@ export enum ModelArtifactState {
   REFERENCE = 'REFERENCE',
 }
 
+export enum ModelRegistryMetadataType {
+  INT = 'MetadataIntValue',
+  DOUBLE = 'MetadataDoubleValue',
+  STRING = 'MetadataStringValue',
+  STRUCT = 'MetadataStructValue',
+  PROTO = 'MetadataProtoValue',
+  BOOL = 'MetadataBoolValue',
+}
+
+export type ModelRegistryCustomPropertyInt = {
+  metadataType: ModelRegistryMetadataType.INT;
+  int_value: string; // int64-formatted string
+};
+
+export type ModelRegistryCustomPropertyDouble = {
+  metadataType: ModelRegistryMetadataType.DOUBLE;
+  double_value: number;
+};
+
+export type ModelRegistryCustomPropertyString = {
+  metadataType: ModelRegistryMetadataType.STRING;
+  string_value: string;
+};
+
+export type ModelRegistryCustomPropertyStruct = {
+  metadataType: ModelRegistryMetadataType.STRUCT;
+  struct_value: string; // Base64 encoded bytes for struct value
+};
+
+export type ModelRegistryCustomPropertyProto = {
+  metadataType: ModelRegistryMetadataType.PROTO;
+  type: string; // url describing proto value
+  proto_value: string; // Base64 encoded bytes for proto value
+};
+
+export type ModelRegistryCustomPropertyBool = {
+  metadataType: ModelRegistryMetadataType.BOOL;
+  bool_value: boolean;
+};
+
+export type ModelRegistryCustomProperty =
+  | ModelRegistryCustomPropertyInt
+  | ModelRegistryCustomPropertyDouble
+  | ModelRegistryCustomPropertyString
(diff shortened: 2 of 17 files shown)


frontend/src/pages/modelRegistry/screens/ModelPropertiesTableRow.tsx
```diff
@@ -0,0 +1,176 @@
...
+  saveEditedProperty,
+}) => {
+  const { key, value } = keyValuePair;
+  const [unsavedKey, setUnsavedKey] = React.useState(key);
+  const [unsavedValue, setUnsavedValue] = React.useState(value);
+
+  const keyValidationError =
+    unsavedKey !== key && allExistingKeys.includes(unsavedKey)
+      ? 'Key must not match an existing property key or label'
+      : unsavedKey.length > 63
```
gitdallas, line 49: nit.. wonder if it would be better to just use normal if/else blocks here just for readability

gitdallas (approved):

frontend/src/pages/modelRegistry/screens/ModelPropertiesDescriptionListGroup.tsx
```diff
@@ -0,0 +1,108 @@
...
+type ModelPropertiesDescriptionListGroupProps = {
+  customProperties: ModelRegistryCustomProperties;
+  saveEditedCustomProperties: (properties: ModelRegistryCustomProperties) => Promise<unknown>;
+};
+
+const ModelPropertiesDescriptionListGroup: React.FC<ModelPropertiesDescriptionListGroupProps> = ({
+  customProperties = {},
+  saveEditedCustomProperties,
+}) => {
+  const [editingPropertyKeys, setEditingPropertyKeys] = React.useState<string[]>([]);
```
manaswinidas, line 20: I'm not seeing the "Show more" functionality in Properties section. It's breaking the layout.

Screenshot 2024-05-03 at 6 32 58 PM

manaswinidas (changes requested):

manaswinidas (dismissed):
Also, I cannot see the dragging functionality for description. As per the mocks
Screenshot 2024-05-03 at 6 39 45 PM

vconzola:
LGTM. @mturley I think the way you're handling error cases is fine for now. I also like the enhancement you suggested in Slack. I assume that'll be in a future PR once UXD has a chance to review your proposal.

mturley:
@manaswinidas I'm working on the overflow in the Properties section, but for the Description textarea I do get a draggable control in the bottom-right on Chrome. Although it should probably default to that 24-line height, I can change that.

mturley:
@manaswinidas I addressed your feedback: the description textarea will now default to 24 lines and be resizable only vertically (I was missing the `resizeOrientation` prop), and the Properties section now properly wraps and truncates text:

Screenshot 2024-05-03 at 11 41 44 AM

Screenshot 2024-05-03 at 11 41 50 AM

Screenshot 2024-05-03 at 11 42 01 AM

Really good catch there, thank you!

frontend/src/pages/modelRegistry/screens/ModelPropertiesDescriptionListGroup.tsx
mturley: Resolved, see screenshots in comment below

mturley:
/hold
we'll merge https://github.com/opendatahub-io/odh-dashboard/pull/2774 first

frontend/src/components/EditableLabelsDescriptionListGroup.tsx
```diff
@@ -107,7 +122,11 @@ const EditableLabelsDescriptionListGroup: React.FC<EditableTextDescriptionListGr
                 editableProps={{ 'aria-label': `Editable label with text ${label}` }}
                 onClose={() => removeUnsavedLabel(label)}
                 closeBtnProps={{ isDisabled: isSavingEdits }}
-                onEditComplete={(_event, newText) => editUnsavedLabel(newText, index)}
+                onEditComplete={(_event, newText) => {
+                  if (!reservedKeys.includes(newText) && newText.length <= 63) {
```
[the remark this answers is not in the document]
mturley: Yeah, I'll keep the discussion open about this and open an issue if we decide to change it. ([slack thread here](https://redhat-internal.slack.com/archives/C06EJG57Z0X/p1714681683554989))

frontend/src/concepts/modelRegistry/types.ts
```diff
@@ -35,14 +35,65 @@ export enum ModelArtifactState {
...
+  DOUBLE = 'MetadataDoubleValue',
+  STRING = 'MetadataStringValue',
+  STRUCT = 'MetadataStructValue',
+  PROTO = 'MetadataProtoValue',
+  BOOL = 'MetadataBoolValue',
+}
+
+export type ModelRegistryCustomPropertyInt = {
+  metadataType: ModelRegistryMetadataType.INT;
+  int_value: string; // int64-formatted string
```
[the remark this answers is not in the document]
mturley: This is the way the data is structured in the API, I was going from the [swagger docs here](https://editor.swagger.io/?url=https://raw.githubusercontent.com/opendatahub-io/model-registry/main/api/openapi/model-registry.yaml). We could bring that feedback up with the model-registry backend team maybe, although I'm not sure we'll ever use the non-string types here. I just added them for completeness

frontend/src/pages/modelRegistry/screens/ModelPropertiesTableRow.tsx
mturley: Good call

gitdallas (approved):

yih-wang:
Hi @mturley, thanks for all the work, they look really great! I got two comments regarding the current screens, but feel free to close this pr and address the comments in future enhancements.

* It seems that the validation for duplicated labels/properties occurs while the user is typing. This could be annoying if the user hasn't finished inputting and accidentally duplicates an existing label. I believe it would be better to validate this after the user submits the modal/row.
* I didn't see from the screen cap how it works if the properties are more than 5 and collapsed, and the user clicks 'add property' at this moment. Just to clarify, the ideal behavior would be for the collapsed list to expand, and the cursor to automatically locate to the newly added key field.

mturley:
Hi @yih-wang,

Thanks for the feedback! Let's make sure to follow up with a UX issue / discussion to improve this after this PR.
* I'm not sure how input validation works elsewhere in the dashboard, I think it would be important to be consistent with that. I implemented the real-time validation because it is simpler to implement that way in React and I've seen it done that way elsewhere, but we can certainly change that behavior. This is one area where I may need to keep my non-designer opinion to myself, but personally as a user I always hate when I don't see validation until after I submit something because it takes me longer to try multiple times than to get the proactive feedback as I type and make sure I'm doing it right the first time. But I think I may be the weird one here!
* Currently if you click "Add property" while the properties are collapsed, it puts the new row at the bottom of the currently shown rows without expanding the rest. That's something we could easily change.

frontend/src/pages/modelRegistry/screens/ModelDetailsView.tsx
```diff
@@ -1,97 +1,103 @@
...
+            value={rm.description || ''}
+            saveEditedValue={(value) =>
+              apiState.api
+                .patchRegisteredModel(
+                  {},
+                  // TODO remove the getPatchBody* functions when https://issues.redhat.com/browse/RHOAIENG-6652 is resolved
+                  getPatchBodyForRegisteredModel(rm, { description: value }),
+                  rm.id,
+                )
+                .then(refreshAPIState)
```
lucferbux, line 39: Why do you want to call refresh api here? This just reload the fetching of the CR, which is not needed when PATCH right?

We should refresh ONLY the version details here, we are reloading pretty much everything.

frontend/src/pages/modelRegistry/screens/ModelDetailsView.tsx
```diff
@@ -1,97 +1,103 @@
...
+              apiState.api
+                .patchRegisteredModel(
+                  {},
+                  // TODO remove the getPatchBody* functions when https://issues.redhat.com/browse/RHOAIENG-6652 is resolved
+                  getPatchBodyForRegisteredModel(rm, {
+                    customProperties: mergeUpdatedLabels(rm.customProperties, editedLabels),
+                  }),
+                  rm.id,
+                )
+                .then(refreshAPIState)
```
lucferbux, line 55: Same here

frontend/src/pages/modelRegistry/screens/ModelDetailsView.tsx
lucferbux: For this you can get the `refresh` value form `useModelVersionById` that we get in `ModelVersionDetails` and pass it to the component.

lucferbux: I've just created a commit (cause i wanted to test it) which is more self-explanatory, I'll leave it here so you can check the suggestion: https://github.com/lucferbux/odh-dashboard/commit/6b921040d92a6d6cb0b366e8ac81a1fcb65051d1

frontend/src/pages/modelRegistry/screens/ModelVersionDetails/ModelVersionDetailsView.tsx
```diff
@@ -1,20 +1,26 @@
...
 import useModelArtifactsByVersionId from '~/concepts/modelRegistry/apiHooks/useModelArtifactsByVersionId';
+import { ModelRegistryContext } from '~/concepts/modelRegistry/context/ModelRegistryContext';
 
 type ModelVersionDetailsViewProps = {
   modelVersion: ModelVersion;
 };
 
 const ModelVersionDetailsView: React.FC<ModelVersionDetailsViewProps> = ({ modelVersion: mv }) => {
   const [modelArtifact] = useModelArtifactsByVersionId(mv.id);
+  const { apiState, refreshAPIState } = React.useContext(ModelRegistryContext);
```
lucferbux, line 23: Same here

lucferbux (changes requested):
Ok, so far everything seems great, I think there was still a discussion open right? But so far it looks great, just a small nit about refreshing data, we should use the right refreshing state for each component

frontend/src/pages/modelRegistry/screens/ModelDetailsView.tsx
mturley: Ah, good call thank you! I was looking at `apiState` for a way to do that, of course it makes sense that it would be on `useModelVersionById`

lucferbux (approved):
/lgtm

PR #2777, opened 2024-05-06
Fix unit test date - Artifacts table

jpuzz0:
Relates to: [RHOAIENG-2977](https://issues.redhat.com/browse/RHOAIENG-2977)
Fixing Artifacts table unit test to always the use current date to compare against.

### DIFF
--- a/frontend/src/pages/pipelines/global/experiments/artifacts/__tests__/ArtifactsTable.spec.tsx
+++ b/frontend/src/pages/pipelines/global/experiments/artifacts/__tests__/ArtifactsTable.spec.tsx
@@ -79,7 +79,7 @@ describe('ArtifactsTable', () => {
                 ],
               ],
               state: 2,
-              createTimeSinceEpoch: 1711113121829,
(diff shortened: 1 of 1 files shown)


frontend/src/pages/pipelines/global/experiments/artifacts/__tests__/ArtifactsTable.spec.tsx
```diff
@@ -103,7 +103,7 @@ describe('ArtifactsTable', () => {
                 ],
               ],
               state: 2,
-              createTimeSinceEpoch: 1611399342384,
+              createTimeSinceEpoch: new Date(),
```
jeff-phillips-18, line 106: This one could be left alone to test the date string format

jpuzz0: Good point. Reverted this to the original.

jeff-phillips-18 (approved):

PR #1046, opened 2023-03-27
add dynamic namespace to delegate-url oauth arg

strangiato:
Closes #1045 

## Description
Resolve issue with Service Accounts being unable to authenticate to the dashboard without cluster scope permissions

## How Has This Been Tested?
Deployed kfdef manually and followed reproducability instructions in #1045

1. Install ODH with a kfdef in a namespace called `opendatahub` with the following manifests url:

```
  repos:
    - name: manifests
      uri: >-
        https://github.com/strangiato/odh-manifests/tarball/dashboard-oauth-delegate-url-fix
```

2. Create a test service account:
```
oc create sa test-connection
```
3. Grant permissions to the service account:
```
oc policy add-role-to-user view system:serviceaccount:opendatahub:test-connection
```
4. Verify SA has permissions to list service:
```
TOKEN=$(oc create token test-connection -n opendatahub)
oc login --token=${TOKEN}
oc get services odh-dashboard -n opendatahub
```
5.  Attempt to authenticate to OAuth
```
URL="https://$(oc get routes odh-dashboard -n opendatahub -o jsonpath='{.spec.host}' )"
curl --location --request GET $URL \
--header "Authorization: Bearer ${TOKEN}" -I
```

Response will return a 200 instead of a 403.

## Request review criteria:

- [x] The commits have meaningful messages (squashes happen on merge by the bot).
- [N/A] Included any necessary screenshots or gifs if it was a UI change.
- [N/A] Included tags to the UX team if it was a UI/UX change.
- [x] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [x] The developer has manually tested the changes and verified that the changes work
- [ ] The developer has added tests or explained why testing cannot be added (unit tests & storybook for related changes)

PR #2789, opened 2024-05-08
update oauth-proxy openshift-delegate-urls

christianvogt:
https://issues.redhat.com/browse/RHOAIENG-6479

## Description

The oauth-proxy `openshift-delegate-urls` option grants access to the underlying service via `Authorization: Bearer` header. The current access check is restricted to users with access to the namespace in which the dashboard is deployed. This limits access to privileged users and restricts other users, such as self-provisioners.

When running a local frontend dev server, developers want to target external clusters and test with users of varying access. Currently this method of development only works for cluster admin users. The change in this PR will loosen the access check so that any user with the ability to list projects can get through.

## How Has This Been Tested?

Apply the deployment changes for the oauth-proxy configuration.

Create a self-provisioner user on the cluster
In a terminal, `oc login` ... to your cluster with the unprivileged user
Now start the dev server: `npm run start:dev:ext`

Visit http://localhost:4010 to check the UI is running as expected.

## Test Impact

None.

## Request review criteria:

If you have UI changes: 

cc @andrewballantyne

### DIFF
--- a/manifests/base/deployment.yaml
+++ b/manifests/base/deployment.yaml
@@ -70,12 +70,6 @@ spec:
             name: odh-ca-cert
             subPath: odh-ca-bundle.crt
       - name: oauth-proxy
-        env:
-          - name: NAMESPACE
-            valueFrom:
-              fieldRef:
-                apiVersion: v1
-                fieldPath: metadata.namespace
         args:
           - --https-address=:8443
           - --provider=openshift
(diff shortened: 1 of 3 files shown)


christianvogt:
fyi @andrewballantyne @lucferbux

manifests/base/deployment.yaml
```diff
@@ -88,7 +88,7 @@ spec:
           - --cookie-secret-file=/etc/oauth/config/cookie_secret
           - --cookie-expire=23h0m0s
           - --pass-access-token
-          - '--openshift-delegate-urls={"/": {"resource": "services", "verb": "get", "name": "odh-dashboard", "namespace": "$(NAMESPACE)"}}'
+          - '--openshift-delegate-urls={"/": {"resource": "projects", "verb": "list"}}'
```
andrewballantyne, line 91: The env on line 74 could probably be dropped.

christianvogt:
@andrewballantyne referred me to https://github.com/opendatahub-io/odh-dashboard/pull/1046 for testing because it made changes to the oauth-proxy `openshift-delegate-urls` and added the `NAMESPACE` arg so that service accounts without cluster access could be used to authenticate with a bearer token. I tested its use case and the same service account with view access can list projects and authenticate with this change.

The change in this PR should allow any authenticated user access.

PR #2798, opened 2024-05-09
Add Compare Runs metric section

Gkrumbach07:
closes: https://issues.redhat.com/browse/RHOAIENG-4772
closes: https://issues.redhat.com/browse/RHOAIENG-4771

## Description

Adds compare runs section for the following metrics:
- scalar metric table
- confusion matrix
- roc curve
- I added a markdown tab but kept it disabled until fetching from s3 is done

**Scalar metric table**
- should be able to see all scalar metrics for each execution.
- will react to selected runs
- horizontal scroll
- ability to hide same metrics

**Confusion matrix**
- horizontal scroll
- has expand button for graph which opens graph by itself.
   - i moved away from the mocks on how the dropdown works in this state becuase the dropdown is a checkbox. @yannnz  you can clarify what is intended on this 

**roc curve**
- curve has legend that matches table under it
- table can be sorted and filtered
Screenshot 2024-05-09 at 5 27 14 PM
Screenshot 2024-05-09 at 5 27 04 PM
Screenshot 2024-05-09 at 5 26 52 PM
Screenshot 2024-05-09 at 5 26 47 PM
Screenshot 2024-05-09 at 5 26 42 PM
Screenshot 2024-05-09 at 5 26 28 PM

## How Has This Been Tested?

- compare runs that have all metrics. use pipeline linked
[Pipeline metrics-visualization-pipeline.yaml.zip](https://github.com/opendatahub-io/odh-dashboard/files/15268325/Pipeline.metrics-visualization-pipeline.yaml.zip)
- go to metrics section and verify description

## Test Impact

Tests are complicated becuase they involve graphs and mlmd. I am working on a solution now but i want to get this PR up so it unblocks

## Request review criteria:

If you have UI changes: 


frontend/src/concepts/pipelines/apiHooks/mlmd/useGetArtifactById.ts
jpuzz0, line 1: Doesn't this already exist in `src/pages/pipelines/global/experiments/artifacts`?

frontend/src/concepts/pipelines/apiHooks/mlmd/useGetArtifactsList.ts
jpuzz0, line 1: Doesn't this already exist in `src/pages/pipelines/global/experiments/artifacts`?

frontend/src/concepts/pipelines/content/artifacts/charts/ConfusionMatrix.tsx
```diff
@@ -0,0 +1,185 @@
...
+              }}
+            >
+              <span style={{ left: 15, position: 'absolute', top: -7 }}>
+                {Math.floor((i / 5) * max)}
+              </span>
+            </div>
+          ))}
+        </div>
+        <div
+          style={{
```
jpuzz0, line 173: A lot of the inline styles in this file can be achieved by using PF util classnames.

jpuzz0: The styles that aren't available via PF classnames, seems like there's enough to be pushed inside of separate stylesheets for the ConfusionMatrix in the form of our own classnames.

frontend/src/concepts/pipelines/content/artifacts/charts/ConfusionMatrix.tsx
```diff
@@ -0,0 +1,185 @@
...
+  // Determine the size for all cells, including labels
+  const cellSize = `${size}px`;
+
+  // Generate the gradient for the legend
+  const gradientLegend = `linear-gradient(to bottom, rgba(41, 121, 255, 1) 0%, rgba(41, 121, 255, 0) 100%)`;
+
+  return (
+    <div
+      style={{
+        display: 'flex',
```
jpuzz0, line 51: Can use `Flex` PF component instead of this.

frontend/src/concepts/pipelines/content/compareRuns/metricsSection/confusionMatrix/ConfusionMatrixCompare.tsx
```diff
@@ -0,0 +1,143 @@
...
+    fullArtifactPaths
+      .map((fullPath) => {
+        const customProperties = fullPath.linkedArtifact.artifact.getCustomPropertiesMap();
+        return {
+          run: fullPath.run,
+          title: getFullArtifactPathLabel(fullPath),
+          data: customProperties.get('confusionMatrix')?.getStructValue()?.toJavaScript(),
+        };
+      })
+      .filter((confidenceMetrics) => !!confidenceMetrics.data)
```
jpuzz0, line 64: At least, this map & filter can be simplified with 1 reduce, where probably the forEach could be combined with that reduce as well.

frontend/src/concepts/pipelines/content/compareRuns/metricsSection/markdown/MarkdownCompare.tsx
```diff
@@ -0,0 +1,133 @@
...
+
+    fullArtifactPaths
+      .map((fullPath) => ({
+        run: fullPath.run,
+        title: getFullArtifactPathLabel(fullPath),
+        uri: fullPath.linkedArtifact.artifact.getUri(),
+      }))
+      .filter((markdown) => !!markdown.uri)
+      .forEach(async ({ uri, title, run }) => {
+        const data = uri; // TODO: fetch data from uri
```
jpuzz0, line 58: nit: Is there a JIRA associated with this TODO?

frontend/src/concepts/pipelines/content/compareRuns/metricsSection/markdown/MarkdownSelect.tsx
jpuzz0, line 1: This and the `ConfusionMatrixSelect` are close to identical. Seems like we should create a shared component for this.

frontend/src/concepts/pipelines/content/compareRuns/metricsSection/roc/RocCurveTableRow.tsx
```diff
@@ -0,0 +1,51 @@
...
+    <CheckboxTd
+      id={getLinkedArtifactId(fullArtifactPath.linkedArtifact)}
+      isChecked={isChecked}
+      onToggle={onToggleCheck}
+    />
+    <Td dataLabel="Execution name > Artifact name">{getFullArtifactPathLabel(fullArtifactPath)}</Td>
+    <Td dataLabel="Run name">{fullArtifactPath.run.display_name}</Td>
+    <Td dataLabel="Curve legend">
+      <ChartLegend
+        colorScale={'red' as unknown as string[]}
```
jpuzz0, line 33: This is a confusing cast. Does this not work `colorScale={['red']}`?

frontend/src/concepts/pipelines/content/compareRuns/metricsSection/roc/utils.ts
```diff
@@ -0,0 +1,24 @@
...
+  );
+};
+
+export const buildRocCurveConfig = (
+  confidenceMetricsArray: ConfidenceMetric[],
+  index: number,
+): ROCCurveConfig => ({
+  index,
+  data: confidenceMetricsArray.map((metric) => ({
+    name: metric.confidenceThreshold as unknown as string,
```
jpuzz0, line 19: Why cast a number type as a string? Shouldn't the number be converted to a string instead?

frontend/src/concepts/pipelines/content/compareRuns/metricsSection/scalar/ScalarMetricTable.tsx
```diff
@@ -0,0 +1,112 @@
...
+      <Switch
+        label="Hide parameters with no differences"
+        isChecked={isHideSameRowsChecked}
+        onChange={(_, checked) => setIsHideSameRowsChecked(checked)}
+        id="hide-same-scalar-metrics-switch"
+        data-testid="hide-same-scalar-metrics-switch"
+      />
+
+      <InnerScrollContainer>
+        <Table
```
jpuzz0, line 97: The style of these comparison tables will need to change due to this new PF guideline: https://patternfly-org-pr-4006-site.surge.sh/components/table/design-guidelines#comparison-table

frontend/src/concepts/pipelines/content/compareRuns/metricsSection/scalar/utils.ts
```diff
@@ -0,0 +1,116 @@
...
+import { SortableData } from '~/components/table';
+import {
+  getExecutionDisplayName,
+  getArtifactName,
+} from '~/concepts/pipelines/content/compareRuns/metricsSection/utils';
+import { RunArtifact } from '~/concepts/pipelines/apiHooks/mlmd/types';
+import { Value } from '~/third_party/mlmd';
+import { ScalarTableProps, XParentLabel, ScalarRowData, ScalarTableData } from './types';
+
+export function getMetadataValue(value?: Value): string | number | Struct | undefined {
```
jpuzz0, line 11: This helper already exists and is called `getMlmdMetadataValue`

frontend/src/concepts/pipelines/content/compareRuns/metricsSection/types.ts
```diff
@@ -0,0 +1,19 @@
+// An artifact which has associated event.
+
+import { LinkedArtifact } from '~/concepts/pipelines/apiHooks/mlmd/types';
+import { PipelineRunKFv2 } from '~/concepts/pipelines/kfTypes';
+import { Artifact, Event, Execution } from '~/third_party/mlmd';
+
+// each run can have multiple executions, artifacts, and events
+export type MlmdPackage = {
```
jpuzz0, line 8: nit: is "Package" the best way to describe what this object represents? If all of the data relates to the run, should the name not indicate this in some way?

frontend/src/pages/pipelines/global/experiments/compareRuns/CompareRunsMetricsSection.tsx
```diff
@@ -0,0 +1,106 @@
...
+          title={<TabTitleText>{MetricSectionTabLabels.ROC_CURVE}</TabTitleText>}
+        >
+          <TabContentBody hasPadding>
+            <RocCurveCompare runArtifacts={rocCurveArtifactData} isLoaded={isLoaded} />
+          </TabContentBody>
+        </Tab>
+        <Tab
+          eventKey={MetricSectionTabLabels.MARKDOWN}
+          title={<TabTitleText>{MetricSectionTabLabels.MARKDOWN}</TabTitleText>}
+          isDisabled // TODO enable when markdown can be fetched from storage (s3)
```
jpuzz0, line 97: nit: Ideally we have a JIRA link for this comment.

jpuzz0 (commented):
Using some existing runs that probably don't have matrix data to display, I'm seeing an infinite loading spinner: 
image

frontend/src/concepts/pipelines/content/artifacts/charts/ConfusionMatrix.tsx
Gkrumbach07: Yeah probably, this component is a mess bc it has to override many styles, i will revisit this and see what i ca do to reuse patternfly. But some PF layout components add extra styles that i cant have in this weird usage

frontend/src/concepts/pipelines/content/compareRuns/metricsSection/markdown/MarkdownCompare.tsx
Gkrumbach07: not yet, im working on the spike or the data fetching

frontend/src/concepts/pipelines/content/compareRuns/metricsSection/markdown/MarkdownSelect.tsx
Gkrumbach07: i was so close to doing so, i will go ahead and do that

frontend/src/concepts/pipelines/content/compareRuns/metricsSection/roc/utils.ts
Gkrumbach07: this was part of a CP from KF. will fix

frontend/src/concepts/pipelines/content/compareRuns/metricsSection/scalar/ScalarMetricTable.tsx
Gkrumbach07: that link leads nowhere, can you send more context

Gkrumbach07: I am going to have UX follow up on this becuase to align with the guidelines, we would have to drop some labels. I will log an issue and link it here.

frontend/src/concepts/pipelines/content/compareRuns/metricsSection/types.ts
Gkrumbach07: its what KF used, i can come up with a better one

frontend/src/concepts/pipelines/content/artifacts/charts/ConfusionMatrix.tsx
Gkrumbach07: the Flex broke it so i just put everything into a scss file to clean it up

frontend/src/concepts/pipelines/content/compareRuns/metricsSection/markdown/MarkdownCompare.tsx
Gkrumbach07: added

frontend/src/concepts/pipelines/content/artifacts/charts/ConfusionMatrix.tsx
```diff
@@ -0,0 +1,136 @@
...
+                  lineHeight: cellSize,
+                  minWidth: cellSize,
+                }}
+              >
+                <Text>{labels[rowIndex]}</Text>
+              </td>
+              {row.map((cell, cellIndex) => (
+                <td
+                  key={labels[cellIndex] + labels[rowIndex]}
+                  className="ConfusionMatrix-cell"
```
jpuzz0, line 67: classnames are not typically lowercase across the board.

frontend/src/concepts/pipelines/content/artifacts/charts/ConfusionMatrix.tsx
jpuzz0, line 1: Why would this not be placed inside of the folder called `confusionMatrix`?

jpuzz0:
Are you sure the intention of `Expand` link was to just span the whole width of the drawer? Based on [this Figma conversation](https://www.figma.com/design/etbN0ccCcZR1qKCpJXQNAg?node-id=1682-101180#801138424), it seems like the intention for all `Expand`s might be to open them in a modal, although I'm questioning that in this conversation and waiting for Yan's response.

jpuzz0:
Why would anything be hidden if only 1 run is shown here? 
image

image

Also, why is only 1 run shown, when 3 are selected in this comparison example?
image

jpuzz0:
When selecting the already checked/selected item for the Confusion matrix, this is what I see:
image

Seems like maybe the click handler should do nothing when attempting to click an item that is already selected to prevent this.

jpuzz0:
Should we not show empty state when there are no parameters with differences? 
image

jpuzz0:
Wondering why we would use "Series" instead of the run name for the ROC curve legend: 
image

jpuzz0:
Should we still show ROC curve data if the search finds none?
image

Gkrumbach07:
> Are you sure the intention of Expand link was to just span the whole width of the drawer? Based on [this Figma conversation](https://www.figma.com/design/etbN0ccCcZR1qKCpJXQNAg?node-id=1682-101180#801138424), it seems like the intention for all Expands might be to open them in a modal, although I'm questioning that in this conversation and waiting for Yan's response.

The UX is different on the compare runs page vs the run details for expand

Gkrumbach07:
> Should we still show ROC curve data if the search finds none?

This is what is done in KF, and i did not have a ref for what yan wanted so i fell back on what upstream did

Gkrumbach07:
> Wondering why we would use "Series" instead of the run name for the ROC curve legend:

there could be multiple per run so it would have to be run name > execution name > artifact name.... this can get large so i assume thats why its just using series name. this can be iterated on late if it is confusing

Gkrumbach07:
> Should we not show empty state when there are no parameters with differences?

We could, this is what upstream had and there was no mock for what to do in our case

Gkrumbach07:
> Why would anything be hidden if only 1 run is shown here?

I can remove the switch if there is only one run

Gkrumbach07:
> Also, why is only 1 run shown, when 3 are selected in this comparison example?

it only shows the runs with scalar metrics

Gkrumbach07:
> When selecting the already checked/selected item for the Confusion matrix, this is what I see:

I think that would be confusing to the user if they could not deselect it. 

The blank skeleton was the closest PF like thing i could use to simulate what KF used.

I think a lot of the concerns need UX to decide what is best, maybe we can wait until demo and @yannnz  can give her opinion

jpuzz0:
> > When selecting the already checked/selected item for the Confusion matrix, this is what I see:
> 
> I think that would be confusing to the user if they could not deselect it.
> 
> The blank skeleton was the closest PF like thing i could use to simulate what KF used.
> 
> I think a lot of the concerns need UX to decide what is best, maybe we can wait until demo and @yannnz can give her opinion

The skeleton is used for loading states, but this doesn't load anything, so I think if we are meant to deselect the item, the content should instead show empty state?

yannnz:
- For the ROC Curve, there is an interaction been missed. When the user hover the legend of the series, the corresponding curve highlighted with a thicker curve. 
Screenshot 2024-05-15 at 21 32 51

- Experiment column of the run table list should be clickable if we keep parity with KFP UI.
Screenshot 2024-05-15 at 21 36 20

- Not quite sure why the Markdown tab is disable. If it is because there is no markdown, I think we need an empty states.
Screenshot 2024-05-15 at 21 38 22

- On my screen, the space between the Metrics section and the Parameter section can be larger to keep the space consistency between sections. The space between the Run table list under the ROC Curve and the ROC Curve graph can be smaller because they related with each other.
Screenshot 2024-05-15 at 21 35 53
Screenshot 2024-05-15 at 21 35 39

yannnz:
> Should we still show ROC curve data if the search finds none? image

I wonder how KFP does it?  I still cannot figure out how their filters work.   I can see how to use this filter is to filter the listed artifact names, if there are several results matching the results, the UI will only display those filtered ones, right? If this is the logic for filtering, filtering ended with no result means nothing to show on the graph.

yannnz:
> Wondering why we would use "Series" instead of the run name for the ROC curve legend: image
> 
> If we do keep Series, it seems like there should be a tooltip for each series color in the table showing which series corresponds to the color: image

I still don't get what is Series TBH. We just followed KFP UI.

yannnz:
> Should we not show empty state when there are no parameters with differences? image

Yes I don't think we need to show empty state for this scenario, because there are still parameters. It might confuse the user. WDYT?

yannnz:
> When selecting the already checked/selected item for the Confusion matrix, this is what I see: image
> 
> Seems like maybe the click handler should do nothing when attempting to click an item that is already selected to prevent this.

Thanks for raising this. Yes this is not a good experience when you already selected something, but it started reloading just because you clicked it again.

yannnz:
> Why would anything be hidden if only 1 run is shown here? image
> 
> image
> Also, why is only 1 run shown, when 3 are selected in this comparison example? image

It is weird. Is it because of the KFP API behavior?

jpuzz0:
> > Wondering why we would use "Series" instead of the run name for the ROC curve legend: image
> > If we do keep Series, it seems like there should be a tooltip for each series color in the table showing which series corresponds to the color: image
> 
> I still don't get what are Series TBH. Are those runs? We just followed KFP UI....

@yannnz I'm not sure either. I understood it as them being the runs since that's what is unique for each table row in this case, but I could be wrong.

jpuzz0:
> > Why would anything be hidden if only 1 run is shown here? image
> > image
> > Also, why is only 1 run shown, when 3 are selected in this comparison example? image
> 
> It is weird. Is it because of the KFP API behavior?

@yannnz I don't think so. At least how the Parameters compare table works, if 1 run has data, the other 2 show to compare against empty values. This should be achievable here as well, but not sure what was expected.

jpuzz0:
> > Should we not show empty state when there are no parameters with differences? image
> 
> Yes I don't think we need to show empty state for this scenario, because there are still parameters. It might confuse the user. WDYT?

@yannnz Does it still have parameters? I think the 2nd row here doesn't represent a parameter, just the execution name.

This has a parameter called `accuracy`:
image

This has no parameters as far as my understanding goes:
image

jpuzz0:
> > Should we still show ROC curve data if the search finds none? image
> 
> I wonder how KFP does it? I still cannot figure out how their filters work. I can see how to use this filter is to filter the listed artifact names, if there are several results matching the results, the UI will only display those filtered ones, right? If this is the logic for filtering, filtering ended with no result means nothing to show on the graph.

In KF UI, When adding filter text to the search box and no results are found, they don't update the table data, and instead still show the rows. In our case, we allow users to filter, which hides the rows, but yet they're still selected:

KF UI:
image

frontend/src/concepts/pipelines/content/compareRuns/metricsSection/roc/RocCurveCompare.tsx
```diff
@@ -0,0 +1,125 @@
...
+        const customProperties = fullArtifactPath.linkedArtifact.artifact.getCustomPropertiesMap();
+        return {
+          data: customProperties.get('confidenceMetrics')?.getStructValue()?.toJavaScript(),
+          fullArtifactPath,
+        };
+      })
+      .filter((confidenceMetrics) => !!confidenceMetrics.data)
+      .map(({ data, fullArtifactPath }, i) => {
+        // validate the custom properties
+        const confidenceMetricsArray = data?.list as unknown[];
```
jpuzz0, line 54: Why is this case to unknown?

jpuzz0:
When the width is small enough and there's enough data, the x-axis values overlap for the ROC curve: 
image

frontend/src/concepts/pipelines/content/compareRuns/metricsSection/roc/RocCurveCompare.tsx
Gkrumbach07: becuase it is coming from custom properties

Gkrumbach07:
> For the ROC Curve, there is an interaction been missed. When the user hover the legend of the series, the corresponding curve highlighted with a thicker curve.
 This is not possible in victory charts

Gkrumbach07:
> @yannnz I don't think so. At least how the Parameters compare table works, if 1 run has data, the other 2 show to compare against empty values. This should be achievable here as well, but not sure what was expected.

KF does not show empty ones, just ones that gave the param. otherwise it may give the impression that the execution has that scalar metric. And if we do go down this route, then we would be showing a bunch of empty cells

Gkrumbach07:
> @yannnz In KF UI, When adding filter text to the search box and no results are found, they don't update the table data, and instead still show the rows. In our case, we allow users to filter, which hides the rows, but yet they're still selected:

filtering should not deselect things from the checkboxes. that is not our behavior in other areas so it should not be here. i guess the idea is to help find your artifact to show if you have a lot?

frontend/src/concepts/pipelines/content/compareRuns/metricsSection/roc/RocCurveCompare.tsx
Gkrumbach07: i changed the logic to be based on the actual value now

PR #2745, opened 2024-04-25
Cypress test for pagination

ppadti:
Closes: [RHOAIENG-5117](https://issues.redhat.com/browse/RHOAIENG-5117)

## Description
This PR aims to create a utility to test pagination. Follow up PR will have pagination for pipelines.

## How Has This Been Tested?
`npm run test`

## Request review criteria:

Self checklist (all need to be checked):
- [x] The developer has manually tested the changes and verified that the changes work
- [x] Commits have been squashed into descriptive, self-contained units of work (e.g. 'WIP' and 'Implements feedback' style messages have been removed)
- [x] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [ ] The developer has added tests or explained why testing cannot be added (unit or cypress tests for related changes)

If you have UI changes: 

- [ ] Included any necessary screenshots or gifs if it was a UI change.
- [ ] Included tags to the UX team if it was a UI/UX change (find relevant UX in the [SMEs](https://github.com/opendatahub-io/odh-dashboard/tree/main/docs/smes.md) section).

After the PR is posted & before it merges:
- [ ] The developer has tested their solution on a cluster by using the image produced by the PR to `main`

PR #2800, opened 2024-05-10
Cypress test for pipeline pagination

ppadti:
Closes: [RHOAIENG-5117](https://issues.redhat.com/browse/RHOAIENG-5117)

## Description
follow up to https://github.com/opendatahub-io/odh-dashboard/pull/2745
This PR to add pagination to pipeline and pipeline run tables.

## How Has This Been Tested?
`npm run test`

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/__tests__/cypress/cypress/e2e/pipelines/PipelineRuns.cy.ts
+++ b/frontend/src/__tests__/cypress/cypress/e2e/pipelines/PipelineRuns.cy.ts
@@ -32,6 +32,7 @@ import {
   ProjectModel,
   RouteModel,
 } from '~/__tests__/cypress/cypress/utils/models';
+import { tablePagination } from '~/__tests__/cypress/cypress/pages/components/Pagination';
 
 const projectName = 'test-project-filters';
 const pipelineId = 'test-pipeline';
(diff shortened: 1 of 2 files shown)


ppadti:
For Archived run, table is not properly getting updated with mock data. So I was not able to add pagination tests for that table.

frontend/src/__tests__/cypress/cypress/e2e/pipelines/PipelineRuns.cy.ts
```diff
@@ -170,6 +171,92 @@ describe('Pipeline runs', () => {
...
+        pagination.findPreviousButton().should('be.disabled');
+        cy.intercept(
+          {
+            method: 'GET',
+            pathname: `/api/service/pipelines/${projectName}/dspa/apis/v2beta1/runs`,
+          },
+          {
+            runs: mockRuns.slice(10, 15),
+            total_size: 15,
+            next_page_token: 'page-2-token',
```
christianvogt, line 217: There shouldn't be another page token, especially of the same value as the previous page token, when there are no more items to fetch.

frontend/src/__tests__/cypress/cypress/e2e/pipelines/PipelineRuns.cy.ts
```diff
@@ -170,6 +171,92 @@ describe('Pipeline runs', () => {
...
+          {
+            method: 'GET',
+            pathname: `/api/service/pipelines/${projectName}/dspa/apis/v2beta1/runs`,
+          },
+          {
+            runs: mockRuns.slice(0, 10),
+            total_size: 15,
+            next_page_token: 'page-2-token',
+          },
+        );
```
christianvogt, line 200: Pagination relies on the UI sending the correct query params in the network request as the user navigates the pages. The test should assert the query params are correct, especially the page token as it controls the ability to retrieve the next page. Page tokens are cached in the UI and therefore going back a page should use the cached value. Any new page token received from the backend should be used in subsequent next page requests and not the old cached value.

Interceptors should be configured not only with a path, but also query params.

christianvogt (commented):
There's also opportunity to avoid duplication of tests here and create utilities that can test the pagination functionality of the various tables on different pages since they all function the same way.

PR #2810, opened 2024-05-10
feat(mr): mr settings scaffolding

gitdallas:
Closes: https://issues.redhat.com/browse/RHOAIENG-7057

## Description
Scaffolding for model registry admin settings, component with just empty state for now
image
image

## How Has This Been Tested?
manually tested that the nav item shows up based on useArea, and that the page renders the empty state properly

## Test Impact
none, just scaffolding for now

## Request review criteria:
make sure the nav item appears properly and that the page renders empty state with an action button that navs to model registries.

i had to run `/backend` vs `:ext` from `/frontend` as the `:ext` doesn't have the setting for it yet.

If you have UI changes: 

### DIFF
--- a/backend/src/types.ts
+++ b/backend/src/types.ts
@@ -18,6 +18,7 @@ export type DashboardConfig = K8sResourceCommon & {
       disableInfo: boolean;
       disableSupport: boolean;
       disableClusterManager: boolean;
+      disableModelRegistryManager: boolean;
       disableTracking: boolean;
       disableBYONImageStream: boolean;
       disableISVBadges: boolean;
--- a/backend/src/types.ts
+++ b/backend/src/types.ts
@@ -18,7 +18,6 @@ export type DashboardConfig = K8sResourceCommon & {
       disableInfo: boolean;
       disableSupport: boolean;
       disableClusterManager: boolean;
-      disableModelRegistryManager: boolean;
       disableTracking: boolean;
       disableBYONImageStream: boolean;
       disableISVBadges: boolean;
--- a/frontend/src/concepts/areas/const.ts
+++ b/frontend/src/concepts/areas/const.ts
@@ -10,11 +10,6 @@ export const SupportedAreasStateMap: SupportedAreasState = {
   [SupportedArea.CLUSTER_SETTINGS]: {
     featureFlags: ['disableClusterManager'],
   },
-  [SupportedArea.MODEL_REGISTRY_SETTINGS]: {
-    featureFlags: ['disableModelRegistry'],
-    requiredComponents: [StackComponent.MODEL_REGISTRY],
-    requiredCapabilities: [StackCapability.SERVICE_MESH, StackCapability.SERVICE_MESH_AUTHZ],
-  },
   [SupportedArea.CUSTOM_RUNTIMES]: {
     featureFlags: ['disableCustomServingRuntimes'],
     reliantAreas: [SupportedArea.MODEL_SERVING],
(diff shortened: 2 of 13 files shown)


backend/src/types.ts
```diff
@@ -18,6 +18,7 @@ export type DashboardConfig = K8sResourceCommon & {
       disableInfo: boolean;
       disableSupport: boolean;
       disableClusterManager: boolean;
+      disableModelRegistryManager: boolean;
```
lucferbux, line 21: @andrewballantyne I'll defer this decision to you, there's an admin section for model registry which is an integral part of the feature, do we use the same feature flag for the enablement/disablement of the feature or should we use a different one?

frontend/src/concepts/areas/const.ts
```diff
@@ -10,6 +10,9 @@ export const SupportedAreasStateMap: SupportedAreasState = {
   [SupportedArea.CLUSTER_SETTINGS]: {
     featureFlags: ['disableClusterManager'],
   },
+  [SupportedArea.MODEL_REGISTRY_SETTINGS]: {
+    featureFlags: ['disableModelRegistryManager'],
```
lucferbux, line 14: This should have ate least:

```typescript
    requiredComponents: [StackComponent.MODEL_REGISTRY],
```

and 

```typescript
    requiredCapabilities: [StackCapability.SERVICE_MESH, StackCapability.SERVICE_MESH_AUTHZ],
```

(this is just cause we need to use it with authorino)

lucferbux (changes requested):
I wanna have a conversation about whether we should have an extra feature flag or not, I see the value on that but I'm not sure 100% if it's needed.

And we should add the extra checks in the supported area.

backend/src/types.ts
gitdallas: @andrewballantyne prefers not adding the extra flag. "sounds like you should have 1 flag. Don’t make it complicated for the user just because"

removed that

andrewballantyne (changes requested):
As I noted on Slack... not sure you should be creating a new area... a new flag is a bad idea, but the new area seems confusing. Why not use the existing area & feature flag?

backend/src/types.ts
lucferbux: Great, thanks for adding the reply here!

lucferbux (commented):
/lgtm

frontend/src/concepts/areas/const.ts
```diff
@@ -77,5 +77,6 @@ export const SupportedAreasStateMap: SupportedAreasState = {
   [SupportedArea.MODEL_REGISTRY]: {
     featureFlags: ['disableModelRegistry'],
     requiredComponents: [StackComponent.MODEL_REGISTRY],
+    requiredCapabilities: [StackCapability.SERVICE_MESH, StackCapability.SERVICE_MESH_AUTHZ],
```
mturley, line 80: @lucferbux this is further restricting the existing Model Registry page to only work if you have service mesh, is that intended?

mturley (approved):
LGTM other than a question for @lucferbux above

frontend/src/concepts/areas/const.ts
mturley: Assuming the answer is yes since you /lgtm'd, I missed that

lucferbux: Yes, that's right, we need both service mesh and authorino to be working, we didn't added it before cause when we started development weren't part of the operator, now it can be safely added.

PR #2812, opened 2024-05-13
Add backend routes and frontend service for admin CRUD of ModelRegistries

mturley:
Closes https://issues.redhat.com/browse/RHOAIENG-7058

## Description

Adds the following admin-only backend API routes:
* `GET /api/modelRegistries` - lists ModelRegistries
* `POST /api/modelRegistries` - creates a ModelRegistry
* `GET /api/modelRegistries/:modelRegistryName` - gets a ModelRegistry
* `PATCH /api/modelRegistries/:modelRegistryName` - updates a ModelRegistry
* `DELETE /api/modelRegistries/:modelRegistryName` - deletes a ModelRegistry 

And adds the new `frontend/src/services/modelRegistryService.ts` with functions to call these routes using axios.

Also adds a new constant `MODEL_REGISTRY_NAMESPACE` to `backend/src/utils/constants.ts` since there are some places where we were hard-coding the use of the `odh-model-registries` namespace.

## How Has This Been Tested?

The admin page where this will be used does not exist yet (it is being created in https://github.com/opendatahub-io/odh-dashboard/pull/2810). I tested the two GET routes by inserting this temporary code into another existing page:

```ts
  React.useEffect(() => {
    const fetchData = async () => {
      const registries = await listModelRegistriesBackend();
      console.log('Registries!', registries);
      const registry = await getModelRegistryBackend('internal-registry');
      console.log('Registry!', registry);
    };
    fetchData();
  }, []);
```

This worked as expected. I did not yet test the POST, PATCH or DELETE routes because I don't want to mess with what's currently set up on the shared cluster. Those will be tested as part of upcoming PRs to consume these routes, and any issues we find at that time can be addressed in those PRs.

## Test Impact

No tests here currently. There are no existing tests for any of these backend routes or corresponding services. They are also a temporary workaround to be removed when we access these services via a passthrough API (per @lucferbux) so tests here would be removed soon anyway.

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/backend/src/routes/api/modelRegistries/index.ts
+++ b/backend/src/routes/api/modelRegistries/index.ts
@@ -1,0 +1,180 @@
+import { FastifyReply, FastifyRequest } from 'fastify';
+import { PatchUtils } from '@kubernetes/client-node';
+import { secureAdminRoute } from '../../../utils/route-security';
+import { KubeFastifyInstance, ModelRegistryKind, RecursivePartial } from '../../../types';
+import { MODEL_REGISTRY_NAMESPACE } from '../../../utils/constants';
+
+export default async (fastify: KubeFastifyInstance): Promise<void> => {
+  fastify.get(
+    '/',
+    secureAdminRoute(fastify)(
+      async (
+        request: FastifyRequest<{ Querystring: { labelSelector: string } }>,
+        reply: FastifyReply,
+      ) => {
+        const { labelSelector } = request.query;
+        try {
+          const response = await fastify.kube.customObjectsApi.listNamespacedCustomObject(
+            'modelregistry.opendatahub.io',
+            'v1alpha1',
+            MODEL_REGISTRY_NAMESPACE,
+            'modelregistries',
+            undefined,
+            undefined,
+            undefined,
+            labelSelector,
+          );
+          return response.body;
+        } catch (e) {
+          fastify.log.error(
+            `ModelRegistries could not be listed, ${e.response?.body?.message || e.message}`,
+          );
+          reply.send(e);
+        }
+      },
+    ),
+  );
+
+  fastify.post(
+    '/',
+    secureAdminRoute(fastify)(
+      async (
+        request: FastifyRequest<{
+          Querystring: { dryRun?: string };
+          Body: ModelRegistryKind;
+        }>,
+        reply: FastifyReply,
+      ) => {
+        const { dryRun } = request.query;
+        const modelRegistry = request.body;
+        try {
+          const response = await fastify.kube.customObjectsApi.createNamespacedCustomObject(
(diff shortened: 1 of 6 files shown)


backend/src/routes/api/modelRegistries/index.ts
```diff
@@ -0,0 +1,141 @@
+import { FastifyReply, FastifyRequest } from 'fastify';
+import { PatchUtils } from '@kubernetes/client-node';
+import { secureAdminRoute } from '../../../utils/route-security';
+import { KubeFastifyInstance, ModelRegistryKind, RecursivePartial } from '../../../types';
+import { MODEL_REGISTRY_NAMESPACE } from '../../../utils/constants';
+
+export default async (fastify: KubeFastifyInstance): Promise<void> => {
```
lucferbux, line 7: Ok, first thing, checking this, we need to add permissions to the Dashboard Serviceaccount to make sure it can fetch Model Registry CR. I would advise to add that to the [cluster role file](https://github.com/opendatahub-io/odh-dashboard/blob/main/manifests/base/cluster-role.yaml)

backend/src/routes/api/service/modelregistry/index.ts
```diff
@@ -20,7 +20,7 @@ export default async (fastify: KubeFastifyInstance): Promise<void> => {
           // kubectl port-forward -n <namespace> svc/<service-name> <local.port>:<service.port>
           `http://${process.env.MODEL_REGISTRY_SERVICE_HOST}:${process.env.MODEL_REGISTRY_SERVICE_PORT}`
         : // Construct service URL
-          `http://${name}.odh-model-registries.svc.cluster.local:8080`;
+          `http://${name}.${MODEL_REGISTRY_NAMESPACE}.svc.cluster.local:8080`;
```
lucferbux, line 23: Yes, really great catch, we didn't parametrize this. 👌

lucferbux (changes requested):
Just add the permissions and I think it would be ready to go

backend/src/routes/api/modelRegistries/index.ts
mturley: Done, pushed -- I gave it all the verbs (same as we do for notebooks, for example), is that right?

gitdallas (approved):

mturley:
@lucferbux @gitdallas I updated this to add the DELETE route, if you don't mind re-reviewing

backend/src/routes/api/modelRegistries/index.ts
lucferbux: Yes, that's fine cause we need to do all the operations 👌

lucferbux (approved):
/lgtm

PR #2815, opened 2024-05-15
Model version and details archive

dpanshug:
[RHOAIENG-6639](https://issues.redhat.com/browse/RHOAIENG-6639)
[RHOAIENG-6640](https://issues.redhat.com/browse/RHOAIENG-6640)

## Description
Archiving model versions.

See below video for archiving and restoring flow, along with archive model version table and details view.
[Screencast from 2024-05-20 16-07-17.webm](https://github.com/opendatahub-io/odh-dashboard/assets/97534722/cc36b216-b2eb-41ab-ac24-30d82c129f97)

## How Has This Been Tested?
1. Move to model registry section and open a register model.
2. Try archiving and restoring model version.

## Test Impact
Added unit tests and cypress tests

## Request review criteria:

If you have UI changes: 

@yih-wang

### DIFF
--- a/frontend/src/__tests__/cypress/cypress/e2e/modelRegistry/ModelVersionArchive.cy.ts
+++ b/frontend/src/__tests__/cypress/cypress/e2e/modelRegistry/ModelVersionArchive.cy.ts
@@ -1,0 +1,292 @@
+/* eslint-disable camelcase */
+import { mockK8sResourceList } from '~/__mocks__';
+import { mockDashboardConfig } from '~/__mocks__/mockDashboardConfig';
+import { MODEL_REGISTRY_API_VERSION } from '~/concepts/modelRegistry/const';
+import { mockModelRegistry } from '~/__mocks__/mockModelRegistry';
+import { mockRegisteredModelList } from '~/__mocks__/mockRegisteredModelsList';
+import { ModelRegistryModel } from '~/__tests__/cypress/cypress/utils/models';
+import { mockModelVersionList } from '~/__mocks__/mockModelVersionList';
+import { mockModelVersion } from '~/__mocks__/mockModelVersion';
+import { ModelVersion, ModelVersionState } from '~/concepts/modelRegistry/types';
+import { mockRegisteredModel } from '~/__mocks__/mockRegisteredModel';
+import { verifyRelativeURL } from '~/__tests__/cypress/cypress/utils/url';
+import {
+  archiveVersionModal,
+  modelVersionArchive,
+  restoreVersionModal,
+} from '~/__tests__/cypress/cypress/pages/modelRegistry/modelVersionArchive';
+import { labelModal, modelRegistry } from '~/__tests__/cypress/cypress/pages/modelRegistry';
+
+type HandlersProps = {
+  registeredModelsSize?: number;
+  modelVersions?: ModelVersion[];
+};
+
+const initIntercepts = ({
+  registeredModelsSize = 4,
+  modelVersions = [
+    mockModelVersion({
+      name: 'model version 1',
+      author: 'Author 1',
+      id: '1',
+      labels: [
+        'Financial data',
+        'Fraud detection',
+        'Test label',
+        'Machine learning',
+        'Next data to be overflow',
+        'Test label x',
+        'Test label y',
+        'Test label z',
+      ],
+      state: ModelVersionState.ARCHIVED,
+    }),
+    mockModelVersion({ id: '2', name: 'model version 2', state: ModelVersionState.ARCHIVED }),
+    mockModelVersion({ id: '3', name: 'model version 3' }),
+  ],
+}: HandlersProps) => {
+  cy.interceptOdh(
+    'GET /api/config',
+    mockDashboardConfig({
+      disableModelRegistry: false,
+    }),
+  );
+
+  cy.interceptK8sList(
+    ModelRegistryModel,
+    mockK8sResourceList([mockModelRegistry({}), mockModelRegistry({ name: 'test-registry' })]),
+  );
+
+  cy.interceptK8s(ModelRegistryModel, mockModelRegistry({}));
+
+  cy.interceptOdh(
+    'GET /api/service/modelregistry/:serviceName/api/model_registry/:apiVersion/registered_models',
+    { path: { serviceName: 'modelregistry-sample', apiVersion: MODEL_REGISTRY_API_VERSION } },
+    mockRegisteredModelList({ size: registeredModelsSize }),
+  );
+
+  cy.interceptOdh(
+    'GET /api/service/modelregistry/:serviceName/api/model_registry/:apiVersion/registered_models/:registeredModelId/versions',
+    {
+      path: {
+        serviceName: 'modelregistry-sample',
+        apiVersion: MODEL_REGISTRY_API_VERSION,
+        registeredModelId: 1,
+      },
+    },
+    mockModelVersionList({
+      items: modelVersions,
+    }),
+  );
+
+  cy.interceptOdh(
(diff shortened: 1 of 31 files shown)


yih-wang:
Thanks @dpanshug for all the work! That looks great!

Several comments:
* After a version is restored, the user should be navigated to the version details page rather than the versions list. That will provide convenience for the following actions to that version.
* Not sure whether that should be fixed here, but in the empty state of the versions list, the link button should be put under the primary button, rather than next to each other. That follows the PF guidelines.
* I've noticed numerous tooltips are appearing when interacting with the interface. Are these tooltips part of the default PF template? It seems like they are too much and not necessary in all cases. Is that possible to display only the tooltips specified in the designs?
image
image

dpanshug:
@yih-wang Pls check the latest screencast video. 

> After a version is restored, the user should be navigated to the version details page rather than the versions list. That will provide convenience for the following actions to that version.

I've updated this so restoring from the details page navigates to the version details page. Should the same happen when restoring from the Archive version table?

> Not sure whether that should be fixed here, but in the empty state of the versions list, the link button should be put under the primary button, rather than next to each other. That follows the PF guidelines.

Corrected this.

> I've noticed numerous tooltips are appearing when interacting with the interface. Are these tooltips part of the default PF template? It seems like they are too much and not necessary in all cases. Is that possible to display only the tooltips specified in the designs?

That's because we are using [Truncate](https://www.patternfly.org/components/truncate) component for the name. I am not sure if we can remove the unwanted tooltip. cc @lucferbux

frontend/src/pages/modelRegistry/screens/routeUtils.ts
```diff
@@ -1,8 +1,21 @@
...
+
+export const modelVersionArchiveUrl = (rmId?: string, preferredModelRegistry?: string): string =>
+  `/modelRegistry/${preferredModelRegistry}/registeredModels/${rmId}/versions/archive`;
+
+export const modelVersionArchiveDetailsUrl = (
+  mvId: string,
+  rmId?: string,
+  preferredModelRegistry?: string,
+): string =>
+  `/modelRegistry/${preferredModelRegistry}/registeredModels/${rmId}/versions/archive/${mvId}`;
```
manaswinidas, line 21: I think this route can be changed to `archived` instead of `archive`, since it's already archived?

manaswinidas (changes requested):

frontend/src/__tests__/cypress/cypress/e2e/modelRegistry/ModelVersionArchive.cy.ts
```diff
@@ -0,0 +1,320 @@
...
+};
+
+const initIntercepts = ({
+  registeredModelsSize = 4,
+  modelVersions = [
+    mockModelVersion({
+      name: 'model version 1',
+      author: 'Author 1',
+      id: '1',
+      customProperties: {
```
manaswinidas, line 36: Do we need the `customProperties` here ? If yes, can we move this to another shared file since a lot of mocks use this?

manaswinidas (changes requested):

frontend/src/pages/modelRegistry/screens/routeUtils.ts
dpanshug: `archive` sounds better to me, as it denotes a list of archived items. @lucferbux, any thoughts on this?

frontend/src/pages/modelRegistry/ModelRegistryRoutes.tsx
```diff
@@ -47,6 +49,29 @@ const ModelRegistryRoutes: React.FC = () => (
             />
             <Route path="*" element={<Navigate to="." />} />
           </Route>
+          <Route path="versions/archive">
```
lucferbux, line 52: I'm leaving that here, not just for your PR, maybe I'll create a task for this, but we should start thinking about splitting routes to decouple a little bit this section.

frontend/src/pages/modelRegistry/screens/components/RestoreModelVersionModal.tsx
```diff
@@ -0,0 +1,65 @@
...
+  const onClose = React.useCallback(() => {
+    onCancel();
+  }, [onCancel]);
+
+  const onConfirm = React.useCallback(async () => {
+    setIsSubmitting(true);
+
+    try {
+      await onSubmit();
+      setIsSubmitting(false);
```
lucferbux, line 32: If you end up always setting submit to false use `finally`

frontend/src/pages/modelRegistry/screens/routeUtils.ts
lucferbux: That's something that we'll need to run down with the UX team for microcopy, IMO if it were `archivedModels` i would go with the past but given that it's a state `archive` sounds better.

frontend/src/pages/modelRegistry/screens/ModelVersionsArchive/ModelVersionArchiveDetails.tsx
```diff
@@ -0,0 +1,127 @@
...
+  const [rm] = useRegisteredModelById(rmId);
+  const [mv, mvLoaded, mvLoadError, refresh] = useModelVersionById(mvId);
+  const [isRestoreModalOpen, setIsRestoreModalOpen] = React.useState(false);
+
+  return (
+    <>
+      <ApplicationsPage
+        {...pageProps}
+        breadcrumb={
+          <Breadcrumb>
```
lucferbux, line 55: This seems long enought to convert in it's own component don't you think?

lucferbux (changes requested):
There are a couples of nits in the code, haven't had the time to properly test the functionality, will do tomorrow, tests seemed great!

yih-wang:
@dpanshug Thanks for all the updates! The new video looks good.

> I've updated this so restoring from the details page navigates to the version details page. Should the same happen when restoring from the Archive version table?

Yes, lets keep both flows ending up with the details page.

> That's because we are using Truncate component for the name. I am not sure if we can remove the unwanted tooltip. cc @lucferbux

What makes sense to me is only the truncated names have hovering tooltips. Not sure whether it's an easy way to estimate whether it's a truncated name or not. Can we remove the tooltip also for the field below? Seems it's not due to the truncate component.
image

lucferbux:
> What makes sense to me is only the truncated names have hovering tooltips. Not sure whether it's an easy way to estimate whether it's a truncated name or not. Can we remove the tooltip also for the field below? Seems it's not due to the truncate component. image

@yih-wang @dpanshug Oh i assumed that wasn't a tooltip but a native autofill in the browser, let me review the code but I couldn't recall a tooltip in the confirmation modal

yih-wang:
@lucferbux Ah yes, the one in the confirmation modal might be a browser-native component. It looks different from the tooltip above.

lucferbux:
> That's because we are using [Truncate](https://www.patternfly.org/components/truncate) component for the name. I am not sure if we can remove the unwanted tooltip. cc @lucferbux
I was assuming we are mostly reusing what we have in other views, if we have tooltips and truncate text in say model serving and pipelines we should stick to that.

yih-wang:
> I was assuming we are mostly reusing what we have in other views, if we have tooltips and truncate text in say model serving and pipelines we should stick to that.

Agree. I'm fine if all the other places do the same thing. We could just follow the pattern and improve this as a whole in the future.

manaswinidas (changes requested):

manaswinidas (approved):
/lgtm Thanks for making all the changes.

lucferbux:
/retest

It seems a problem with cypress and the cdn

PR #2819, opened 2024-05-16
add taskName parameter to parseRuntimeInfoFromExecutions function

Gkrumbach07:
closes: https://issues.redhat.com/browse/RHOAIENG-7079

## Description

- run details now uses the task name first to match an execution to a task in the dag
- if no name exists, we fall back to using the key to match the execution
- and if nothing exists at all in terms of matching an execution, we fall back to the pipeline spec

## How Has This Been Tested?

- check that logs are equal from our UI to KF

## Test Impact

none, small mlmd fix in for backend

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/concepts/pipelines/topology/parseUtils.ts
+++ b/frontend/src/concepts/pipelines/topology/parseUtils.ts
@@ -182,6 +182,7 @@ export const parseRuntimeInfoFromRunDetails = (
 
 export const parseRuntimeInfoFromExecutions = (
   taskId: string,
+  taskName: string,
   executions?: Execution[] | null,
 ): PipelineTaskRunStatus | undefined => {
   if (!executions) {
@@ -189,7 +190,7 @@ export const parseRuntimeInfoFromExecutions = (
   }
 
   const execution = executions.find(
-    (e) => e.getCustomPropertiesMap().get('task_name')?.getStringValue() === taskId,
+    (e) => e.getCustomPropertiesMap().get('task_name')?.getStringValue() === (taskName || taskId),
   );
 
   if (!execution) {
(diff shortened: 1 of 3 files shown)


frontend/src/concepts/pipelines/topology/parseUtils.ts
```diff
@@ -182,14 +182,15 @@ export const parseRuntimeInfoFromRunDetails = (
...
+  taskName: string,
   executions?: Execution[] | null,
 ): PipelineTaskRunStatus | undefined => {
   if (!executions) {
     return undefined;
   }
 
   const execution = executions.find(
-    (e) => e.getCustomPropertiesMap().get('task_name')?.getStringValue() === taskId,
+    (e) => e.getCustomPropertiesMap().get('task_name')?.getStringValue() === (taskName || taskId),
```
jpuzz0, line 193: If both of `taskName` & `taskId` props are required and not possibly `undefined` (since the type is simply `string`, this will always use the taskName. Maybe taskName is meant to be optional and just used to override the usage of taskId?

jpuzz0 (commented):
Just a small comment, otherwise looks good.

frontend/src/concepts/pipelines/topology/parseUtils.ts
Gkrumbach07: not necessarily. I used `||` which will used the taskId if taskName is `""`, which is what i noticed in cases where the label is empty for some reason.

in JS:
```js
a = "";
b = "defined"
a || b == "defined"
```

jpuzz0 (approved):
/lgtm

PR #2827, opened 2024-05-20
Add s3 storage artifact route and ui integration of it

Gkrumbach07:
closes: https://issues.redhat.com/browse/RHOAIENG-7081

## Description

- created a new /api/storage GET route that takes bucket and namespace along with the path as a query param and gives back what is in that s3 bucket.
- using that in compare runs markdown compare and in artifacts list uri link
- added preview and peek file size warnings

Screenshot 2024-06-05 at 4 32 42 PM
Screenshot 2024-06-05 at 4 12 18 PM
Screenshot 2024-06-05 at 4 11 55 PM
Screenshot 2024-06-05 at 4 05 31 PM

## How Has This Been Tested?

- go to artifacts list page and navigate to pipelines with output artifacts
   (reach out to @Gkrumbach07 i you need a test cluster for this PR)

## Test Impact

no tests add bc this is a backend route, will add util tests

## Request review criteria:

If you have UI changes: 

cc @yannnz

### DIFF
--- a/backend/src/routes/api/storage/index.ts
+++ b/backend/src/routes/api/storage/index.ts
@@ -1,0 +1,105 @@
+import { FastifyInstance, FastifyReply } from 'fastify';
+import { getObjectSize, getObjectStream, setupMinioClient } from './storageUtils';
+import { getDashboardConfig } from '../../../utils/resourceUtils';
+import { getDirectCallOptions } from '../../../utils/directCallUtils';
+import { getAccessToken } from '../../../utils/directCallUtils';
+import { OauthFastifyRequest } from '../../../types';
+
+export default async (fastify: FastifyInstance): Promise<void> => {
+  fastify.get(
+    '/:namespace/size',
+    async (
+      request: OauthFastifyRequest<{
+        Querystring: { key: string };
+        Params: { namespace: string };
+      }>,
+      reply: FastifyReply,
+    ) => {
+      try {
+        const dashConfig = getDashboardConfig();
+        if (dashConfig?.spec.dashboardConfig.disableS3Endpoint !== false) {
+          reply.code(404).send('Not found');
+          return reply;
+        }
+
+        const { namespace } = request.params;
+        const { key } = request.query;
+
+        const requestOptions = await getDirectCallOptions(fastify, request, request.url);
+        const token = getAccessToken(requestOptions);
+
+        const { client, bucket } = await setupMinioClient(fastify, token, namespace);
+
+        const size = await getObjectSize({
+          client,
+          key,
+          bucket,
+        });
+
+        reply.send(size);
+      } catch (err) {
+        reply.code(500).send(err.message);
+        return reply;
+      }
+    },
+  );
+
+  fastify.get(
+    '/:namespace',
+    async (
+      request: OauthFastifyRequest<{
+        Querystring: { key: string; peek?: number };
+        Params: { namespace: string };
+      }>,
+      reply: FastifyReply,
+    ) => {
+      try {
+        const dashConfig = getDashboardConfig();
+        if (dashConfig?.spec.dashboardConfig.disableS3Endpoint !== false) {
+          reply.code(404).send('Not found');
+          return reply;
+        }
+
+        const { namespace } = request.params;
+        const { key, peek } = request.query;
+
+        const requestOptions = await getDirectCallOptions(fastify, request, request.url);
+        const token = getAccessToken(requestOptions);
+
+        const { client, bucket } = await setupMinioClient(fastify, token, namespace);
+
+        const stream = await getObjectStream({
+          client,
+          key,
+          bucket,
+          peek,
+        });
+
+        reply.type('text/plain');
+
+        await new Promise<void>((resolve, reject) => {
+          stream.on('data', (chunk) => {
+            reply.raw.write(chunk);
+          });
+
+          stream.on('end', () => {
+            reply.raw.end();
+            resolve();
+          });
+
+          stream.on('error', (err) => {
+            fastify.log.error('Stream error:', err);
+            reply.raw.statusCode = 500;
+            reply.raw.end(err.message);
+            reject(err);
+          });
+        });
+
+        return;
+      } catch (err) {
+        reply.code(500).send(err.message);
+        return reply;
+      }
+    },
+  );
+};
--- a/backend/src/routes/api/storage/storageUtils.ts
+++ b/backend/src/routes/api/storage/storageUtils.ts
@@ -1,0 +1,185 @@
+import { Client as MinioClient } from 'minio';
+import { DSPipelineKind, KubeFastifyInstance } from '../../../types';
+import { Transform, TransformOptions } from 'stream';
+
+export interface PreviewStreamOptions extends TransformOptions {
+  peek: number;
+}
+
+/**
+ * Transform stream that only stream the first X number of bytes.
+ */
+export class PreviewStream extends Transform {
+  constructor({ peek, ...opts }: PreviewStreamOptions) {
+    // acts like passthrough
+    let transform: TransformOptions['transform'] = (chunk, _encoding, callback) =>
+      callback(undefined, chunk);
+    // implements preview - peek must be positive number
+    if (peek && peek > 0) {
+      let size = 0;
+      transform = (chunk, _encoding, callback) => {
+        const delta = peek - size;
+        size += chunk.length;
+        if (size >= peek) {
+          callback(undefined, chunk.slice(0, delta));
+          this.resume(); // do not handle any subsequent data
+          return;
+        }
+        callback(undefined, chunk);
+      };
+    }
+    super({ ...opts, transform });
+  }
+}
+
+export async function getDspa(
+  fastify: KubeFastifyInstance,
+  token: string,
+  namespace: string,
+): Promise<DSPipelineKind> {
+  const dspaResponse = await fastify.kube.customObjectsApi
+    .listNamespacedCustomObject(
+      'datasciencepipelinesapplications.opendatahub.io',
+      'v1alpha1',
+      namespace,
+      'datasciencepipelinesapplications',
+      undefined,
+      undefined,
+      undefined,
+      undefined,
+      undefined,
+      undefined,
+      undefined,
+      undefined,
+      {
+        headers: {
+          authorization: `Bearer ${token}`,
+        },
+      },
+    )
+    .catch((e) => {
+      throw `A ${e.statusCode} error occurred when trying to fetch dspa aws storage credentials: ${
+        e.response?.body?.message || e?.response?.statusMessage
+      }`;
+    });
+
+  const dspas = (
+    dspaResponse?.body as {
+      items: DSPipelineKind[];
+    }
+  )?.items;
+
+  if (!dspas || !dspas.length) {
+    throw 'No Data Science Pipeline Application found';
+  }
+
+  return dspas[0];
+}
+
+async function getDspaSecretKeys(
+  fastify: KubeFastifyInstance,
+  token: string,
+  namespace: string,
+  dspa: DSPipelineKind,
+): Promise<{ accessKey: string; secretKey: string }> {
+  try {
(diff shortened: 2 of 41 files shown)


Gkrumbach07:
keeping this as WIP until security review is done. I will also add tests in the meantime

backend/src/routes/api/storage/index.ts
```diff
@@ -0,0 +1,47 @@
+import { FastifyInstance, FastifyReply, FastifyRequest } from 'fastify';
+import { createMinioClient, getObjectStream } from './storageUtils';
+
+export default async (fastify: FastifyInstance): Promise<void> => {
+  fastify.get('/:namespace/:bucket', async (request: FastifyRequest, reply: FastifyReply) => {
+    try {
+      const { namespace, bucket } = request.params as {
+        namespace: string;
+        bucket: string;
+        key: string;
```
andrewballantyne, line 10: Key shouldn't be here, right? key is in the query...

backend/src/routes/api/storage/index.ts
```diff
@@ -0,0 +1,47 @@
+import { FastifyInstance, FastifyReply, FastifyRequest } from 'fastify';
+import { createMinioClient, getObjectStream } from './storageUtils';
+
+export default async (fastify: FastifyInstance): Promise<void> => {
+  fastify.get('/:namespace/:bucket', async (request: FastifyRequest, reply: FastifyReply) => {
```
andrewballantyne, line 5: Technically the Request & Reply are generics

backend/src/routes/api/storage/index.ts
Gkrumbach07: fixed

backend/src/routes/api/storage/index.ts
Gkrumbach07: fixed, i think

Gkrumbach07:
@andrewballantyne I took this out of WIP, the safety check for the storage status is used, and the feature flag blocks the endpoint

frontend/src/concepts/pipelines/content/artifacts/ArtifactUriLink.tsx
```diff
@@ -0,0 +1,47 @@
...
+    // Otherwise check if the uri is s3
+    // If it is not s3, return undefined as we only support fetching from s3 buckets
+    const uriComponents = extractS3UriComponents(uri);
+    if (!uriComponents) {
+      return;
+    }
+
+    const { bucket, path } = uriComponents;
+
+    // /api/storage/${namespace}/${bucket}?key=${path}
```
andrewballantyne, line 34: Not sure what this comment is telling that the code immediately under it doesn't show almost directly.

backend/src/routes/api/storage/storageUtils.ts
```diff
@@ -0,0 +1,96 @@
...
+/**
+ * Create minio client with aws instance profile credentials if needed.
+ * @param config minio client options where `accessKey` and `secretKey` are optional.
+ */
+export async function createMinioClient(
+  fastify: KubeFastifyInstance,
+  namespace: string,
+): Promise<MinioClient> {
+  try {
+    const dspaResponse = await fastify.kube.customObjectsApi
```
andrewballantyne, line 14: This is problematic -- this is done with the service account. We need to do this with the user's token.

Otherwise you could provide a namespace you do not have access to and get back a response.

See the `utils/proxy.ts` for an example.

backend/src/routes/api/storage/storageUtils.ts
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    // always get the first one
+    const externalStorage = dspas[0].spec.objectStorage.externalStorage;
+
+    if (externalStorage) {
+      const { region, host: endPoint, s3CredentialsSecret } = externalStorage;
+
+      // get secret
+      const secret = await fastify.kube.coreV1Api.readNamespacedSecret(
```
andrewballantyne, line 54: This should also be done with the user's token

backend/src/routes/api/storage/index.ts
```diff
@@ -0,0 +1,50 @@
+import { FastifyInstance } from 'fastify';
+import { createMinioClient, getObjectStream } from './storageUtils';
+import { getDashboardConfig } from '../../../utils/resourceUtils';
+
+export default async (fastify: FastifyInstance): Promise<void> => {
+  const dashConfig = getDashboardConfig();
+  if (!dashConfig.spec.dashboardConfig.disableS3Endpoint) {
```
andrewballantyne, line 7: It's a stupid config -- sadly, this won't work.

disableS3Endpoint equals:
* `true` (flag off) => !true => false ✅ 
* `false` (flag on) => !false => true ✅ 
* `undefined` (flag not set) => !undefined => true ❌  

Explicitly check to see if it's `=== false`

andrewballantyne (changes requested):
If you would like to talk about permission flows more, let me know

frontend/src/concepts/pipelines/content/artifacts/ArtifactUriLink.tsx
```diff
@@ -0,0 +1,46 @@
...
+    // Otherwise check if the uri is s3
+    // If it is not s3, return undefined as we only support fetching from s3 buckets
+    const uriComponents = extractS3UriComponents(uri);
+    if (!uriComponents) {
+      return;
+    }
+
+    const { bucket, path } = uriComponents;
+
+    return `/api/storage/${namespace}/${bucket}?key=${encodeURIComponent(path)}`;
```
andrewballantyne, line 34: Codecov says you don't have a test for this -- can you make sure you have a mocked cypress test for this?

backend/src/routes/api/storage/index.ts
```diff
@@ -0,0 +1,63 @@
...
+              resolve();
+            });
+
+            stream.on('error', (err) => {
+              fastify.log.error('Stream error:', err);
+              reply.raw.statusCode = 500;
+              reply.raw.end(err.message);
+              reject(err);
+            });
+          });
```
andrewballantyne, line 53: My leading concern -- which I'm not sure we need to address it here -- but should before we release it. 

If the resource you're pulling is so large the data crashes our pod, it's not going to be awesome. You're also holding up the connection until you're done -- which may time out.

We should probably find a way to peek only the first "bunch" of characters. Are we responsible for the "full download / view in a new tab"?

andrewballantyne: It would appear we are on the hook for the new tab too... if my scanning of the code was accurate

frontend/src/services/storageService.ts
andrewballantyne, line 1: fwiw, you don't need to use the services folder -- I went back and forth on this, but I think I need to deprecate this folder entirely.

backend/src/routes/api/storage/index.ts
Gkrumbach07: As far as i could tell, the KF ui will attempt to pull all in the new tab unless told to preview

backend/src/routes/api/storage/index.ts
```diff
@@ -0,0 +1,65 @@
...
+    '/:namespace/:bucket',
+    async (
+      request: OauthFastifyRequest<{
+        Querystring: { key: string; peek?: number };
+        Params: { namespace: string; bucket: string };
+      }>,
+      reply: FastifyReply,
+    ) => {
+      const dashConfig = getDashboardConfig();
+      if (dashConfig?.spec.dashboardConfig.disableS3Endpoint === false) {
```
Gkrumbach07, line 19: switch condition

backend/src/routes/api/storage/index.ts
```diff
@@ -0,0 +1,65 @@
...
+import { getAccessToken } from '../../../utils/directCallUtils';
+import { OauthFastifyRequest } from '../../../types';
+
+export default async (fastify: FastifyInstance): Promise<void> => {
+  fastify.get(
+    '/:namespace/:bucket',
+    async (
+      request: OauthFastifyRequest<{
+        Querystring: { key: string; peek?: number };
+        Params: { namespace: string; bucket: string };
```
Gkrumbach07, line 14: take bucket out

backend/src/routes/api/storage/index.ts
```diff
@@ -0,0 +1,65 @@
+import { FastifyInstance, FastifyReply } from 'fastify';
+import { createMinioClient, getObjectStream } from './storageUtils';
+import { getDashboardConfig } from '../../../utils/resourceUtils';
+import { getDirectCallOptions } from '../../../utils/directCallUtils';
+import { getAccessToken } from '../../../utils/directCallUtils';
+import { OauthFastifyRequest } from '../../../types';
+
+export default async (fastify: FastifyInstance): Promise<void> => {
+  fastify.get(
+    '/:namespace/:bucket',
```
andrewballantyne, line 10: Remove bucket, get from DSPA -- maybe make `createMinioClient` into a `setupMinioClient` function that returns the bucket and the client... then use them in the `getObjectStream`.

Gkrumbach07, line 10: create new endpoint for getting size

backend/src/routes/api/storage/index.ts
```diff
@@ -0,0 +1,65 @@
...
+      const dashConfig = getDashboardConfig();
+      if (dashConfig?.spec.dashboardConfig.disableS3Endpoint === false) {
+        try {
+          const { namespace, bucket } = request.params;
+          const { key, peek } = request.query;
+
+          const requestOptions = await getDirectCallOptions(fastify, request, request.url);
+          const token = getAccessToken(requestOptions);
+
+          const stream = await getObjectStream({
```
andrewballantyne, line 27: Can you get the stat information on the file before downloading it so we know how large it is... maybe make that another endpoint and allow this one as a text-stream endpoint.

backend/src/routes/api/storage/index.ts
```diff
@@ -0,0 +1,65 @@
...
+            client: await createMinioClient(fastify, token, namespace),
+            key,
+            peek,
+          });
+
+          reply.type('text/plain');
+
+          await new Promise<void>((resolve, reject) => {
+            stream.on('data', (chunk) => {
+              reply.raw.write(chunk);
```
andrewballantyne, line 38: Use Peek to avoid getting more than 100 megabtypes...

frontend/src/concepts/pipelines/content/artifacts/ArtifactUriLink.tsx
Gkrumbach07: there is a jest test for it. it is hard to create a cypress test becuase this data is from mlmd

backend/src/utils/constants.ts
```diff
@@ -62,6 +62,7 @@ export const blankDashboardCR: DashboardConfig = {
       disableModelMesh: false,
       disableAcceleratorProfiles: false,
       disablePipelineExperiments: true,
+      disableS3Endpoint: false,
```
andrewballantyne, line 65: default needs to be `true`

backend/src/routes/api/storage/index.ts
```diff
@@ -0,0 +1,105 @@
...
+    ) => {
+      try {
+        const dashConfig = getDashboardConfig();
+        if (dashConfig?.spec.dashboardConfig.disableS3Endpoint !== false) {
+          reply.code(404).send('Not found');
+          return reply;
+        }
+
+        const { namespace } = request.params;
+        const { key, peek } = request.query;
```
andrewballantyne, line 64: We need to make sure peek doesn't exceed our limit -- it doesn't matter what the UI says

andrewballantyne (approved):
Okay, let's go with this implementation. We'll be looking at replacing this logic with more structured backend soon, so this is temporary. 

We need loading spinners for the artifact peek loading & we definitely need to consider why the Artifact page does not have the peek -- it's awkward that the direct source (the artifact page itself) of the data doesn't have it but the embedded sources (on the runs page) do... 

/hold 

Not sure if you're addressing the loading spinners now or later, unhold it if you'll follow up in another PR.

andrewballantyne (approved):
lgtm

/unhold

PR #2828, opened 2024-05-20
feat(mr): create mr modal

gitdallas:
Closes: [RHOAIENG-2230](https://issues.redhat.com/browse/RHOAIENG-2230)

## Description
Create MR form/modal
image

## How Has This Been Tested?
Created an MR, checked form cannot be submitted if form validations exist.

## Test Impact

- added cypress tests for access to MR Settings
- added cypress tests for create MR form validation

## Request review criteria:
Modal opens, form looks right, validation shows up. form can be submitted with good content.

If you have UI changes: 

### DIFF
--- a/frontend/src/__tests__/cypress/cypress/e2e/modelRegistrySettings/ModelRegistrySettings.cy.ts
+++ b/frontend/src/__tests__/cypress/cypress/e2e/modelRegistrySettings/ModelRegistrySettings.cy.ts
@@ -1,0 +1,101 @@
+import { mockDashboardConfig, mockDscStatus } from '~/__mocks__';
+import { mockDsciStatus } from '~/__mocks__/mockDsciStatus';
+import { StackCapability, StackComponent } from '~/concepts/areas/types';
+import { modelRegistrySettings } from '~/__tests__/cypress/cypress/pages/modelRegistrySettings';
+import { pageNotfound } from '~/__tests__/cypress/cypress/pages/pageNotFound';
+import { asProductAdminUser, asProjectAdminUser } from '~/__tests__/cypress/cypress/utils/users';
+
+const setupMocksForMRSettingAccess = () => {
+  asProductAdminUser();
+  cy.interceptOdh(
+    'GET /api/config',
+    mockDashboardConfig({
+      disableModelRegistry: false,
+    }),
+  );
+  cy.interceptOdh(
+    'GET /api/dsc/status',
+    mockDscStatus({
+      installedComponents: {
+        [StackComponent.MODEL_REGISTRY]: true,
+        [StackComponent.MODEL_MESH]: true,
+      },
+    }),
+  );
+  cy.interceptOdh(
+    'GET /api/dsci/status',
+    mockDsciStatus({
+      requiredCapabilities: [StackCapability.SERVICE_MESH, StackCapability.SERVICE_MESH_AUTHZ],
+    }),
+  );
+};
+
+it('Model registry settings should not be available for non product admins', () => {
+  asProjectAdminUser();
+  modelRegistrySettings.visit(false);
+  pageNotfound.findPage().should('exist');
+  modelRegistrySettings.findNavItem().should('not.exist');
+});
+
+it('Model registry settings should be available for product admins with capabilities', () => {
+  setupMocksForMRSettingAccess();
+  // check page is accessible
+  modelRegistrySettings.visit(true);
+  // check nav item exists
+  modelRegistrySettings.findNavItem().should('exist');
+});
+
+describe('CreateModal', () => {
+  beforeEach(() => {
+    setupMocksForMRSettingAccess();
+  });
+
+  it('should display error messages when form fields are invalid and not allow submit', () => {
+    modelRegistrySettings.visit(true);
+    cy.findByText('Create model registry').click();
+    // Fill in the form with invalid data
+    cy.get('#mr-name').clear();
+    cy.get('#mr-host').clear();
+    cy.get('#mr-port').clear();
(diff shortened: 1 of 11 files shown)


frontend/src/pages/modelRegistrySettings/CreateModal.scss
```diff
@@ -0,0 +1,6 @@
+// targeting specific form section title elements to override --pf-v5-c-form__section-title--FontSize, --pf-v5-c-form__section-title--FontWeight
+.form-subtitle-text {
+    color: var(--pf-v5-global--Color--200);
```
mturley, line 3: I wonder if we can use utility classes for these instead?

```ts
import text from '@patternfly/react-styles/css/utilities/Text/text';
```

then you can use `text.color_200` in the className. Although there's no utility class for `font-size: initial` or `font-weight: initial`, you'd have to explicitly use the specific size and weight classes.

mturley (approved):
LGTM, one comment that we could leave for another PR if you want.

frontend/src/pages/modelRegistrySettings/CreateModal.scss
gitdallas: maybe best to leave for another PR, i actually grabbed this CSS from another spot that was doing the same thing. we'd likely want to either update it there as well. i'd say we could make a custom component for this "section title/description", but there is a PF issue to add a description property to the form section.

mturley: For posterity: the existing example of this being copied here is from https://github.com/opendatahub-io/odh-dashboard/blob/main/frontend/src/concepts/pipelines/content/configurePipelinesServer/ConfigurePipelinesServerModal.scss#L10

mturley:
Approving, I'll follow up on some minor things we discussed as part of #2829

frontend/src/__tests__/cypress/cypress/e2e/modelRegistrySettings/ModelRegistrySettings.cy.ts
```diff
@@ -0,0 +1,101 @@
...
+      .should('be.visible')
+      .contains(
+        "Must consist of lower case alphanumeric characters or '-', and must start and end with an alphanumeric character",
+      );
+    cy.findByTestId('mr-host-error').should('be.visible').contains('Host cannot be empty');
+    cy.findByTestId('mr-port-error').should('be.visible').contains('Port cannot be empty');
+    cy.findByTestId('mr-username-error').should('be.visible').contains('Username cannot be empty');
+    cy.findByTestId('mr-password-error').should('be.visible').contains('Password cannot be empty');
+    cy.findByTestId('mr-database-error').scrollIntoView();
+    cy.findByTestId('mr-database-error').should('be.visible').contains('Database cannot be empty');
```
christianvogt, line 77: @dpanshug Selectors should be part of page objects so they are reusable when we get to e2e tests.

frontend/src/__tests__/cypress/cypress/e2e/modelRegistrySettings/ModelRegistrySettings.cy.ts
```diff
@@ -0,0 +1,101 @@
...
+    cy.get('#mr-database').clear();
+    cy.get('#mr-database').blur();
+    // Assert the submit button is disabled
+    cy.findByTestId('modal-submit-button').should('be.disabled');
+    // Assert that error messages are displayed
+    cy.findByTestId('mr-name-error')
+      .should('be.visible')
+      .contains(
+        "Must consist of lower case alphanumeric characters or '-', and must start and end with an alphanumeric character",
+      );
```
christianvogt, line 71: Asserting static text as defined in in our react components is often a bad idea because it's a 1:1 relationship and will only fail if we change the react component.

mturley: Ack. I plan to give these tests a second pass as part of adding more in #2829, I'll address this feedback there and request a review from you for that one when it's ready if you don't mind.

mturley: @christianvogt looking back at this, I think it has value beyond what you're saying -- we shouldn't just assert that components render certain text for the sake of it, but this is trying to assert that the error is shown vs not shown. We could do it with a test-id instead if you prefer but imo it doesn't matter much as long as the test is asserting logic and not just content.

christianvogt: My concern was the `contains` check and not so much the check that the error node was visible since this is the only available error that's shown in this node.

frontend/src/__tests__/cypress/cypress/e2e/modelRegistrySettings/ModelRegistrySettings.cy.ts
mturley: @christianvogt FYI I addressed this in updates to https://github.com/opendatahub-io/odh-dashboard/pull/2829

frontend/src/__tests__/cypress/cypress/e2e/modelRegistrySettings/ModelRegistrySettings.cy.ts
mturley: @christianvogt FYI I addressed this in updates to https://github.com/opendatahub-io/odh-dashboard/pull/2829

PR #2833, opened 2024-05-21
[RHOAIENG-4999] Details page for archived runs

jenny-s51:
Closes https://issues.redhat.com/browse/RHOAIENG-4999

## Description

- Adds a label indicating the run is archived in the run title (see [mocks](https://www.figma.com/design/etbN0ccCcZR1qKCpJXQNAg/Pipeline-phase-2?node-id=7456-75320&t=pnUdepwkzdUli9Ot-0#752615976))
- Adds dropdown actions for an archived run

## How Has This Been Tested?

1. Experiments and Runs
2. Select a run
3. Toggle dropdown in upper right
4. Test that dropdown actions are in parity with the mocks and the methods work as expected. 

## Test Impact

Updated PipelinesTopology.cy.ts. 

## Request review criteria:

If you have UI changes: 

Screenshot 2024-05-21 at 3 27 41 PM

### DIFF
--- a/frontend/src/concepts/pipelines/content/pipelinesDetails/pipelineRun/PipelineRunDetailsActions.tsx
+++ b/frontend/src/concepts/pipelines/content/pipelinesDetails/pipelineRun/PipelineRunDetailsActions.tsx
@@ -1,30 +1,54 @@
 import * as React from 'react';
+import { Tooltip } from '@patternfly/react-core';
 import {
   Dropdown,
   DropdownItem,
   DropdownSeparator,
   DropdownToggle,
 } from '@patternfly/react-core/deprecated';
-import { useNavigate, useParams } from 'react-router-dom';
+import { useNavigate } from 'react-router-dom';
 import { usePipelinesAPI } from '~/concepts/pipelines/context';
 import useNotification from '~/utilities/useNotification';
 import { PipelineRunKFv2, RuntimeStateKF, StorageStateKF } from '~/concepts/pipelines/kfTypes';
 import { cloneRunRoute, experimentsCompareRunsRoute } from '~/routes';
 import { SupportedArea, useIsAreaAvailable } from '~/concepts/areas';
+import useExperimentById from '~/concepts/pipelines/apiHooks/useExperimentById';
 
 type PipelineRunDetailsActionsProps = {
   run?: PipelineRunKFv2 | null;
+  onArchive: () => void;
   onDelete: () => void;
 };
 
-const PipelineRunDetailsActions: React.FC<PipelineRunDetailsActionsProps> = ({ onDelete, run }) => {
+const PipelineRunDetailsActions: React.FC<PipelineRunDetailsActionsProps> = ({
+  onDelete,
+  onArchive,
+  run,
+}) => {
   const navigate = useNavigate();
-  const { experimentId } = useParams();
-  const { namespace, api } = usePipelinesAPI();
+  const { namespace, api, refreshAllAPI } = usePipelinesAPI();
   const notification = useNotification();
   const [open, setOpen] = React.useState(false);
   const isExperimentsAvailable = useIsAreaAvailable(SupportedArea.PIPELINE_EXPERIMENTS).status;
   const isRunActive = run?.storage_state === StorageStateKF.AVAILABLE;
+  const [experiment] = useExperimentById(run?.experiment_id);
+  const isExperimentActive = experiment?.storage_state === StorageStateKF.AVAILABLE;
+
+  const RestoreDropdownItem = (
+    <DropdownItem
+      isDisabled={!isExperimentActive}
+      key="restore-run"
+      onClick={() =>
+        run &&
+        api
+          .unarchivePipelineRun({}, run.run_id)
+          .catch((e) => notification.error('Unable to restore pipeline run', e.message))
+          .then(() => refreshAllAPI())
+      }
+    >
+      Restore
+    </DropdownItem>
+  );
 
   return (
     <Dropdown
@@ -42,6 +66,31 @@ const PipelineRunDetailsActions: React.FC<PipelineRunDetailsActionsProps> = ({ o
           ? []
           : [
               <DropdownItem
+                key="retry-run"
+                isDisabled={run.state !== RuntimeStateKF.FAILED || !!run.error}
+                onClick={() =>
+                  api
+                    .retryPipelineRun({}, run.run_id)
+                    .catch((e) => notification.error('Unable to retry pipeline run', e.message))
+                }
+              >
+                Retry
+              </DropdownItem>,
+              <DropdownItem
+                key="clone-run"
+                onClick={() =>
+                  navigate(
+                    cloneRunRoute(
+                      namespace,
+                      run.run_id,
+                      isExperimentsAvailable ? run.experiment_id : undefined,
+                    ),
+                  )
+                }
+              >
+                Duplicate
+              </DropdownItem>,
+              <DropdownItem
                 key="stop-run"
                 isDisabled={run.state !== RuntimeStateKF.RUNNING}
                 onClick={() =>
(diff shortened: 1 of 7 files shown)


jpuzz0:
"Clone" vs "Duplicate" was discussed [here](https://github.com/opendatahub-io/odh-dashboard/pull/2518#discussion_r1496326786) a few months ago. Unless something has changed, I believe the intention was to keep the "Duplicate" wording.

jenny-s51 (commented):
Thank you for providing clarity on this @jpuzz0 - based on that discussion it looks like [the mocks are out of date](https://www.figma.com/design/etbN0ccCcZR1qKCpJXQNAg/Pipeline-phase-2?node-id=7456-75320&t=KsJA9kNFjs0pM9WJ-0#752615976), will revert back to `Duplicate`. cc @yannnz @xianli123

jenny-s51 (commented):
Following the conversation in the demo of this work during the Pipelines Review meeting this morning, based on feedback from @yannnz we want to disable the Restore button on Archived runs, and add a tooltip as shown below. 
Screenshot 2024-05-23 at 11 13 57 AM

It was also noted by Fede that, when navigating to an archived experiment from "Experiments and Runs", we want to set the Archived tab as the active tab:
Screenshot 2024-05-23 at 11 17 31 AM

frontend/src/concepts/pipelines/content/tables/experiment/ExperimentTableRow.tsx
```diff
@@ -24,11 +24,20 @@ const ExperimentTableRow: React.FC<ExperimentTableRowProps> = ({
...
+
   return (
     <Tr>
       <CheckboxTd id={experiment.experiment_id} isChecked={isChecked} onToggle={onToggleCheck} />
       <Td dataLabel="Experiment">
-        <Link to={experimentRunsRoute(namespace, experiment.experiment_id)} state={{ experiment }}>
+        <Link
+          to={
+            isArchived
+              ? experimentArchivedRunsRoute(namespace, experiment.experiment_id)
```
jpuzz0, line 36: `experimentArchivedRunsRoute` is the same as `experimentRunsRoute` except with a querystring of runType, so why not just dynamically add the querystring to either an additional parameter or just to a template string here instead of making a new function?

frontend/src/concepts/pipelines/content/utils.tsx
```diff
@@ -30,6 +31,9 @@ export type RunStatusDetails = {
 const UNKNOWN_ICON = <QuestionCircleIcon />;
 const UNKNOWN_STATUS = 'warning';
 
+export const getStorageState = (run?: PipelineRunKFv2 | null): StorageStateKF | undefined =>
+  run?.storage_state;
```
jpuzz0, line 35: nit: I don't think accessing storage_state off of run constitutes a helper function, where this is seemingly used in only 1 place as well.

frontend/src/concepts/pipelines/content/pipelinesDetails/pipelineRun/PipelineRunDetailsActions.tsx
```diff
@@ -78,6 +90,32 @@ const PipelineRunDetailsActions: React.FC<PipelineRunDetailsActionsProps> = ({ o
               ) : (
                 <React.Fragment key="compare-runs" />
               ),
+              !isRunActive ? (
+                <Tooltip
+                  position="left"
+                  content={
+                    <div>
+                      Archived runs cannot be restored until its associated experiment is restored.
```
jpuzz0, line 98: I might be missing something here, but if this messaging is not behind a feature flag, will it not be confusing to see this without the context of being within an experiment?

frontend/src/concepts/pipelines/content/pipelinesDetails/pipelineRun/PipelineRunDetailsActions.tsx
```diff
@@ -78,6 +90,32 @@ const PipelineRunDetailsActions: React.FC<PipelineRunDetailsActionsProps> = ({ o
...
+                    </div>
+                  }
+                >
+                  <DropdownItem
+                    isDisabled={!isRunActive}
```
Gkrumbach07, line 103: this should be disabled if the experiment is archived. so you would need to do this
```ts
  const [experiment] = useExperimentById(experimentId);
  const isExperimentActive = experiment?.storage_state === StorageStateKF.AVAILABLE;
```

Then the tooltip should only be displayed if the dropdown item is disabled

frontend/src/concepts/pipelines/content/utils.tsx
Gkrumbach07: yeah does not need a helper

Gkrumbach07 (changes requested):

frontend/src/concepts/pipelines/content/tables/experiment/ExperimentTableRow.tsx
```diff
@@ -24,11 +24,20 @@ const ExperimentTableRow: React.FC<ExperimentTableRowProps> = ({
 }) => {
   const { namespace } = usePipelinesAPI();
 
+  const isArchived = window.location.href.includes('/archived');
```
jpuzz0, line 27: I think we typically identify where we are location using react router helpers, and this could be identified using useLocation or useParams. I'm also thinking this row could identify this via a prop passed down from the table as well.

frontend/src/concepts/pipelines/content/tables/experiment/ExperimentTableRow.tsx
jenny-s51: Great point, thank you - updated

frontend/src/concepts/pipelines/content/pipelinesDetails/pipelineRun/PipelineRunDetailsActions.tsx
Gkrumbach07: I think its ok. Becuase the user will not be able to archive their experiment unless they have the feature flag flipped. So they would never get into the case. And if for some reason they do flip the flag, then archive, then flip back. We should still say why it is disabled

frontend/src/concepts/pipelines/content/pipelinesDetails/pipelineRun/PipelineRunDetailsActions.tsx
```diff
@@ -78,6 +93,32 @@ const PipelineRunDetailsActions: React.FC<PipelineRunDetailsActionsProps> = ({ o
               ) : (
                 <React.Fragment key="compare-runs" />
               ),
+              !isExperimentActive ? (
```
Gkrumbach07, line 96: the logic is still wrong here. This check here should be if the run is not active. The the check to see if the item is disabled would be what you have now, isExperimentActive. And you still need to not show the tooltip when it is not disabled

frontend/src/concepts/pipelines/content/tables/experiment/ExperimentTableRow.tsx
```diff
@@ -23,12 +23,20 @@ const ExperimentTableRow: React.FC<ExperimentTableRowProps> = ({
   actionColumnItems,
 }) => {
   const { namespace } = usePipelinesAPI();
+  const location = useLocation();
+
+  const isArchived = location.pathname.includes('/archived');
```
Gkrumbach07, line 28: cant you just check if the experiment is archived or not instead of the route itself?

like 

experiment.storage_state === StorageStateKF.ARCHIVED

Gkrumbach07:
Need to verify this

> An active run cannot be deleted. Should remove the Delete action in the Action dropdown.
Miss Retry action.

Gkrumbach07 (approved):
/lgtm
/approve

frontend/src/concepts/pipelines/content/pipelinesDetails/pipelineRun/PipelineRunDetailsActions.tsx
```diff
@@ -78,10 +108,33 @@ const PipelineRunDetailsActions: React.FC<PipelineRunDetailsActionsProps> = ({ o
...
+              ) : (
+                <React.Fragment key="restore-run" />
+              ),
+              !isRunActive ? (
+                <React.Fragment key="delete-run">
+                  <DropdownSeparator key="separator" />
+                  <DropdownItem onClick={() => onDelete()}>Delete</DropdownItem>
+                </React.Fragment>
+              ) : (
+                <React.Fragment key="delete-run" />
```
Gkrumbach07, line 136: this looks like it should be archive not none

https://www.figma.com/proto/etbN0ccCcZR1qKCpJXQNAg/Pipeline-phase-2?page-id=615%3A34927&node-id=1585-95721&viewport=3206%2C1251%2C0.65&t=oARGr6aU5d4T3KjF-1&scaling=min-zoom&starting-point-node-id=1585%3A95719

frontend/src/concepts/pipelines/content/pipelinesDetails/pipelineRun/PipelineRunDetailsActions.tsx
```diff
@@ -66,6 +85,17 @@ const PipelineRunDetailsActions: React.FC<PipelineRunDetailsActionsProps> = ({ o
               >
                 Duplicate
```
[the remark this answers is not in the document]
jenny-s51: I left "Duplicate" per @jpuzz0 's comment [here](https://github.com/opendatahub-io/odh-dashboard/pull/2833#issuecomment-2127205019) - would we want to update this one to Clone?

frontend/src/concepts/pipelines/content/pipelinesDetails/pipelineRun/PipelineRunDetailsActions.tsx
jenny-s51: Updated

PR #2834, opened 2024-05-21
show special error state for ChunkLoadError

christianvogt:
https://issues.redhat.com/browse/RHOAIENG-2690

## Description

_@ UX please suggest icon and text_

The times in which this screen may be shown is very very small. A possible way to encounter this issue in production is to have the web app opened while the dashboard updates behind the scene. Then the user attempts to navigate the already opened application.

A `ChunkLoadError` is a special case that indicates a failure to load a build artifact. This generally occurs because the web page was opened on a previous version of the application and the build artifacts have changed on the server. It _may_ also occur as a result of cache not getting the latest version although this case hasn't been proven in our environment where it's expected that we do not cache such files.

Whenever a `ChunkLoadError` is detected, the best course of action for the user is to simply refresh the page. As such this change will provide the user with a clear action button for how to proceed. Furthermore in case the error persists for some reason, we give the option to display the error as is done with all other errors. Since it's a possibility that other errors may also be resolved by reloading the page, a general message and reload action is provided in the header of the standard error page. eg. if after an update, the endpoint of an action on longer exists and throws an error.

image

Clicking the `Show error` button will display the standard error screen for data collection purposes:
image

## How Has This Been Tested?

Added unit tests.
This has been tested locally.

1. Build the frontend `npm run build`
2. Start the backend `npm run start:dev`
3. Open your web browser to `localhost:4000`
4. Open web inspector.
5. Navigate to the `Explore` page.
6. From the network tab web inspector look, for `###.js` javascript file request. This is the chunk related to the explore page.
7. Rename `frontend/public/###.js` to anything else. eg `###.js.bak`
8. Reload the page or first go to `localhost:4000` and then click on the `Explore` page link.
9. Observe the new error state.

## Request review criteria:

If you have UI changes: 

cc @andrewballantyne @kywalker-rh @kaedward @vconzola

### DIFF
--- a/frontend/src/components/error/ErrorBoundary.tsx
+++ b/frontend/src/components/error/ErrorBoundary.tsx
@@ -2,6 +2,7 @@ import * as React from 'react';
 import { Button, Split, SplitItem, Title } from '@patternfly/react-core';
 import { TimesIcon } from '@patternfly/react-icons';
 import ErrorDetails from './ErrorDetails';
+import UpdateState from './UpdateState';
 
 type ErrorBoundaryProps = {
   children?: React.ReactNode;
@@ -13,6 +14,7 @@ type ErrorBoundaryState =
       hasError: true;
       error: Error;
       errorInfo: React.ErrorInfo;
+      isUpdateState: boolean;
     };
 
 class ErrorBoundary extends React.Component<ErrorBoundaryProps, ErrorBoundaryState> {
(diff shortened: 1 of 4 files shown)


andrewballantyne (commented):
/approve

Might want to consider tests more so because this is a near impossible reproduce state in the app.

christianvogt:
Added unit tests.

andrewballantyne (approved):

PR #2855, opened 2024-05-29
Convert templates to use websocket

ppadti:
Closes: [RHOAIENG-5368](https://issues.redhat.com/browse/RHOAIENG-5368)

## Description
Convert templates to use websockets. Replaces templates#listTemplates with useTemplates.
Created hook `useCustomK8sWatchResource` to perform type check on errors.

## How Has This Been Tested?
In settings, open serving runtimes:
1. Add/Edit/Duplicate/Delete the serving runtimes and check network inspector for websocket connection.

## Test Impact
Added unit test to the `useCustomK8sWatchResource` hook.
Updated unit and cypress test to use websocket.

## Request review criteria:

If you have UI changes: 


ppadti:
@christianvogt  I tried sending resource to websocket while editing serving templates in cypress test it doesn't seems to work. But I am unsure what should be the `modelOrOption`  while sending Patch request.  
This is what I tried for editing serving runtime: 
 ```
cy.wsK8s(
      'MODIFIED',
      TemplateModel,
incrementResourceVersion(
      mockServingRuntimeTemplateK8sResource({
        name: 'template-1',
        displayName: 'Updated Multi platform',
        platforms: [ServingRuntimePlatform.SINGLE],
      }),)
    );
```

File: frontend/src/__tests__/cypress/cypress/e2e/customServingRuntimes/CustomServingRuntimes.cy.ts [hunk 1]
```diff
@@ -81,37 +78,20 @@ describe('Custom serving runtimes', () => {
...
-    cy.wait('@createSingleModelServingRuntime').then((interception) => {
-      expect(interception.request.url).to.include('?dryRun=All');
-      expect(interception.request.body).to.containSubset({
-        metadata: {
-          name: 'template-new',
-          annotations: { 'openshift.io/display-name': 'New OVMS Server' },
-          namespace: 'opendatahub',
-        },
-      });
-    });
```
christianvogt, line 94: It's still a requirement to assert the request content. Do not remove these.

frontend/src/pages/modelServing/screens/projects/ModelServingPlatform.tsx
```diff
@@ -167,7 +167,7 @@ const ModelServingPlatform: React.FC = () => {
         }
         isLoading={!servingRuntimesLoaded && !templatesLoaded}
         isEmpty={shouldShowPlatformSelection}
-        loadError={platformError || servingRuntimeError || templateError}
+        loadError={platformError || servingRuntimeError || (templateError as Error | undefined)}
```
christianvogt, line 170: The cast is not ideal. I forget why they type error as `unknown` instead of some super of `Error` union with `undefined`.
@andrewballantyne any recollection here?

frontend/src/api/k8s/templates.ts
```diff
@@ -39,18 +47,35 @@ export const assembleServingRuntimeTemplate = (
...
+  const modelServingEnabled = useModelServingEnabled();
+  const customServingRuntimesEnabled = useCustomServingRuntimesEnabled();
+
+  if (!namespace) {
+    throw new NotReadyError('No namespace provided');
+  }
+
+  if (!modelServingEnabled) {
+    throw new NotReadyError('Model serving is not enabled');
+  }
```
christianvogt, line 60: This is different behavior than previously. This will throw in the middle of a render instead of before it would return an error object as part of the result.

frontend/src/api/k8s/templates.ts
```diff
@@ -39,18 +47,35 @@ export const assembleServingRuntimeTemplate = (
...
+
+  const [templates, loaded, error] = useK8sWatchResource<TemplateKind[]>(
+    {
+      isList: true,
+      groupVersionKind: groupVersionKind(TemplateModel),
+      namespace,
+    },
```
christianvogt, line 67: You will need to handle `namespace` being undefined.
```suggestion
    namespace ? {
      isList: true,
      groupVersionKind: groupVersionKind(TemplateModel),
      namespace,
    } : null,
```

frontend/src/api/k8s/templates.ts
```diff
@@ -39,18 +47,35 @@ export const assembleServingRuntimeTemplate = (
...
+  );
+
+  if (!customServingRuntimesEnabled) {
+    return [
+      templates.filter((template) => template.metadata.labels?.['opendatahub.io/ootb'] === 'true'),
+      loaded,
+      error,
+    ];
+  }
+  return [templates, loaded, error];
```
christianvogt, line 78: Store the filtered templates with `useMemo` and return a stable templates array.

Something like this:
```suggestion
  const t = React.useMemo(() => customServingRuntimesEnabled ? templates : templates.filter((template) => template.metadata.labels?.['opendatahub.io/ootb'] === 'true'), [templates, customServingRuntimesEnabled]);
  return [t, loaded, error];
```

frontend/src/pages/modelServing/screens/projects/ModelServingPlatform.tsx
andrewballantyne: No recollection -- anyway we can wrap and override that instead of casting everywhere?

There was some gaps in the K8s CRUD usages too -- I think it was just short sightedness that had not been vetted.

christianvogt: Agreed. We can create our own hook that can perform a typeof check on the error and if it's of type error use it, otherwise create a new Error object. Then we use this hook everywhere in place of `useK8sWatchResource`

ppadti: yes, I will work on this.

frontend/src/utilities/useCustomK8sWatchResource.ts
```diff
@@ -0,0 +1,33 @@
...
+  const [data, loaded, error] = useK8sWatchResource<T>(initResource, initModel, options);
+
+  const loadError = React.useMemo(() => {
+    if (error instanceof Error || typeof error === 'undefined') {
+      return error;
+    }
+
+    if (typeof error === 'string') {
+      return undefined;
+    }
```
christianvogt, line 25: TBH i don't think the SDK has any case where it's not `Error` or `undefined`. But if it does, I would expect if `error` is defined, we should be returning an instance of `Error`.

frontend/src/types.ts
```diff
@@ -689,6 +689,8 @@ export type ImageStreamAndVersion = {
   imageVersion?: ImageStreamSpecTagType;
 };
 
+export type CustomWatchK8sResult<T> = [data: T, loaded: boolean, error?: Error | undefined];
```
christianvogt, line 692: You should reference the type from `WatchK8sResult`. Note that `K8sResourceCommon | K8sResourceCommon` must come from the SDK and not our own definitions.
A comment as to why this exists would be useful until the SDK gets updated at which point we can remove this.
```suggestion
export type CustomWatchK8sResult<R extends K8sResourceCommon | K8sResourceCommon[]> = [
  data: WatchK8sResult<R>[0],
  loaded: WatchK8sResult<R>[1],
  loadError: Error | undefined,
];
```

frontend/src/pages/modelServing/screens/projects/ModelServingPlatform.tsx
christianvogt, line 170:
```suggestion
        loadError={platformError || servingRuntimeError || templateError}
```

frontend/src/api/k8s/templates.ts
christianvogt: Do not throw errors here, return the error as part of the watch result.

frontend/src/utilities/useCustomK8sWatchResource.ts
ppadti: `useK8sWatchResource` is returning an empty string on success.  How will we handle that.  `[data, true,'']`

frontend/src/types.ts
ppadti: We already have a definition for `K8sResourceCommon` in `types.ts`. So you want me to move this type declaration to `k8sTypes.ts` file?

frontend/src/utilities/useCustomK8sWatchResource.ts
christianvogt: Ok in that case, do something like this:
```
    if (error instanceof Error) {
      return error;
    }

    if (!error) {
      return undefined;
    }
```

frontend/src/types.ts
christianvogt: Our definition is `deprecated`. Since this new type is to work with utils from the SDK, it should use the types from the SDK as well.

frontend/src/api/k8s/templates.ts
```diff
@@ -39,18 +46,42 @@ export const assembleServingRuntimeTemplate = (
...
+  const modelServingEnabled = useModelServingEnabled();
+  const customServingRuntimesEnabled = useCustomServingRuntimesEnabled();
+
+  const initResource: WatchK8sResource | null = namespace
+    ? {
+        isList: true,
+        groupVersionKind: groupVersionKind(TemplateModel),
+        namespace,
+      }
+    : null;
```
pnaik1, line 59: initResource should be null when `modelServingEnabled` is false?
```suggestion
  const initResource: WatchK8sResource | null = namespace && modelServingEnabled
    ? {
        isList: true,
        groupVersionKind: groupVersionKind(TemplateModel),
        namespace,
      }
    : null;
```

frontend/src/pages/modelServing/customServingRuntimes/CustomServingRuntimeContext.tsx
```diff
@@ -10,37 +10,34 @@ import {
...
   const { dashboardNamespace } = useDashboardNamespace();
 
-  // TODO: Disable backend workaround when we migrate admin panel to Passthrough API
-  const servingRuntimeTemplates = useContextResourceData<TemplateKind>(
-    useTemplates(dashboardNamespace, true),
-  );
+  const servingRuntimeTemplates = useTemplates(dashboardNamespace);
 
-  // TODO: Disable backend workaround when we migrate admin panel to Passthrough API
+  // TODO: Disable backend workaround when we migrate admin panelk to Passthrough API
```
pnaik1, line 40: nit
```suggestion
  // TODO: Disable backend workaround when we migrate admin panel to Passthrough API
```

pnaik1 (changes requested):

frontend/src/api/k8s/templates.ts
```diff
@@ -39,18 +46,43 @@ export const assembleServingRuntimeTemplate = (
...
+  const [templatesData, loaded, error] = useCustomK8sWatchResource<TemplateKind[]>(
+    initResource,
+    TemplateModel,
+  );
+
+  const templates = React.useMemo(
+    () =>
+      customServingRuntimesEnabled
+        ? templatesData
+        : templatesData.filter(
```
christianvogt, line 71: Isn't it possible that `templatesData` is `undefined` here if `initResource == null` ?

frontend/src/api/k8s/templates.ts
```diff
@@ -39,18 +46,43 @@ export const assembleServingRuntimeTemplate = (
...
+    [templatesData, customServingRuntimesEnabled],
+  );
+
+  if (!namespace) {
+    return [[], true, new NotReadyError('No namespace provided')];
+  }
+
+  if (!modelServingEnabled) {
+    return [[], true, new NotReadyError('Model serving is not enabled')];
+  }
```
christianvogt, line 83: I forgot that `NotReadyError` is a special error handled by `useFetchState` as a way to delay any response. When `useFetchState` encounters this error, it does not set `loaded` and instead returns: `[<default value>, loaded = false, error = undefined]`

I believe this would be the equivalent behavior to the old `useTemplates`. 
Note that we need to make sure that `templatesData` is an array because the old implementation defaults to empty array. We also cannot simply return `[]` here because that would be an unstable reference. Either `templatesData` needs to ensure it is always a value so that `useMemo` will make it stable or you need to define an empty array constant.

It may be possible to also return `[undefined, false, undefined]` since data shouldn't be looked at unless `loaded === true` but it could pose other problems.

```suggestion
  if (!namespace || !modelServingEnabled) {
    return [templatesData, false, undefined];
  }
```

frontend/src/api/k8s/templates.ts
ppadti: yes. I will handle that case.

frontend/src/utilities/useCustomK8sWatchResource.ts
```diff
@@ -0,0 +1,33 @@
...
+
+    if (!error) {
+      return undefined;
+    }
+
+    return new Error('Unknown error occured');
+  }, [error]);
+
+  return [data, loaded, loadError];
+};
```
christianvogt, line 31: Can we focus this hook on listing resources so that we can also work around the type issue in the SDK where `data` can be `undefined`.

```
const useK8sWatchResourceList = <T extends K8sResourceCommon | K8sResourceCommon[]>(
  initResource: WatchK8sResource | null,
  initModel?: K8sModelCommon,
  options?: Partial<WebSocketOptions & RequestInit & { wsPrefix?: string; pathPrefix?: string }>,
): CustomWatchK8sResult<T> => {
  const initListResource = React.useMemo(
    () => (initResource != null ? { ...initResource, isList: true } : null),
    [initResource],
  );
  const [data, loaded, error] = useK8sWatchResource<T>(initListResource, initModel, options);

  const loadError = React.useMemo(() => {
    if (error instanceof Error) {
      return error;
    }

    if (!error) {
      return undefined;
    }

    return new Error('Unknown error occured');
  }, [error]);

  // disable as data can be `undefined` by the type in the SDK is incorrect
  // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
  return [React.useMemo(() => data ?? [], [data]), loaded, loadError];
};
```

If you do this, then you can also fix the issue in `templates.ts` where you have to disable the eslint rule `eslint-disable-next-line @typescript-eslint/no-unnecessary-condition`.

PR #2866, opened 2024-05-30
fixed end date issue and issues scheduling pipelines in Firefox

rsun19:
Closes: [RHOAIENG-1197](https://issues.redhat.com/browse/RHOAIENG-1197)

## Description

The issue at hand was that in Firefox, when trying to schedule a pipeline run, returned incoherent values for the end date. This was because of the usage of the Date() constructor, which is inconsistent at runtime if the parameter isn't an `ISO 8601 formatted` string. Here is the [documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/Date#several_ways_to_create_a_date_object) if you are interested.

Unfortunately, it seemed that I could not schedule any pipelines in Firefox, due to this runtime issue, which is a larger issue. Although this is now fixed, I need guidance on whether or not the testing I have done is adequate, or if it needs more.

Reproducible for Firefox on Mac

## How Has This Been Tested?

Run `npm run test`.

## Test Impact

I added a test for my newly created function in `frontend/src/utilities/__tests__/time.spec.ts`

## Request review criteria:

### DIFF
--- a/frontend/src/concepts/pipelines/content/createRun/contentSections/RunTypeSectionScheduled.tsx
+++ b/frontend/src/concepts/pipelines/content/createRun/contentSections/RunTypeSectionScheduled.tsx
@@ -6,6 +6,7 @@ import EndDateBeforeStartDateError from '~/concepts/pipelines/content/createRun/
 import CatchUp from '~/concepts/pipelines/content/createRun/contentSections/CatchUp';
 import MaxConcurrencyField from '~/concepts/pipelines/content/createRun/contentSections/MaxConcurrencyField';
 import TriggerTypeField from '~/concepts/pipelines/content/createRun/contentSections/TriggerTypeField';
+import { convertToDate } from '~/utilities/time';
 
 type RunTypeSectionScheduledProps = {
   data: RunTypeScheduledData;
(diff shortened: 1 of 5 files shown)


rsun19:
Need guidance on whether more testing is necessary or what I have is adequate.

christianvogt:
After setting a value for the end date, if I then toggle the switch off and then on again I get the following error:
image

frontend/src/concepts/pipelines/content/createRun/contentSections/RunTypeSectionScheduled.tsx
```diff
@@ -39,7 +40,8 @@ const RunTypeSectionScheduled: React.FC<RunTypeSectionScheduledProps> = ({ data,
         onChange={(end) => onChange({ ...data, end })}
         adjustNow={(now) => {
           if (data.start) {
-            const start = new Date(`${data.start.date} ${data.start.time}`);
+            const convertedTime = convertToTwentyFourHourTime(data.start.time);
+            const start = new Date(`${data.start.date}T${convertedTime}:00.000`);
```
christianvogt, line 44: I'd like to see this conversion moved to a function and reused elsewhere.
Something like this:
```
export const toDate = ({ date, time }: RunDateTime) =>
  new Date(`${date}T${convertToTwentyFourHourTime(time)}:00.000`);
```

rsun19:
^ Code is ready for review

PR #2879, opened 2024-06-05
first round of updates to disable type assertions

rsun19:
Closes: [RHOAIENG-4336](https://issues.redhat.com/browse/RHOAIENG-4336)

## Description

We would like to limit the use of type assertions as much as possible. However, a large amount of these assertions are made in a type-cast like way making it very tricky to type check them. We may need to resort to using `as` in the type guard in those cases.

## How Has This Been Tested?

`npm run test`

## Request review criteria:

### DIFF
--- a/frontend/src/components/DocCardBadges.tsx
+++ b/frontend/src/components/DocCardBadges.tsx
@@ -5,7 +5,7 @@ import { QuickStartContext, QuickStartContextValues } from '@patternfly/quicksta
 import { OdhDocument, OdhDocumentType } from '~/types';
 import { getQuickStartCompletionStatus, CompletionStatusEnum } from '~/utilities/quickStartUtils';
 import { DOC_TYPE_TOOLTIPS } from '~/utilities/const';
-import { getLabelColorForDocType, getDuration } from '~/utilities/utils';
+import { getLabelColorForDocType, getDuration, asEnumMember } from '~/utilities/utils';
 import { DOC_TYPE_LABEL } from '~/pages/learningCenter/const';
 
 import './OdhCard.scss';
@@ -19,7 +19,7 @@ const DocCardBadges: React.FC<DocCardBadgesProps> = ({ odhDoc }) => {
   const [completionStatus, setCompletionStatus] = React.useState<
     CompletionStatusEnum | undefined
   >();
-  const docType = odhDoc.spec.type as OdhDocumentType;
+  const docType = asEnumMember(odhDoc.spec.type, OdhDocumentType) ?? OdhDocumentType.Documentation;
   const docName = odhDoc.metadata.name;
   const duration = odhDoc.spec.durationMinutes;
 
--- a/frontend/src/concepts/metrics/utils.ts
+++ b/frontend/src/concepts/metrics/utils.ts
@@ -1,12 +1,12 @@
 import { SelectOptionObject } from '@patternfly/react-core/deprecated';
 import { TimeframeTitle, RefreshIntervalTitle } from '~/concepts/metrics/types';
+import { isEnumMember } from '~/utilities/utils';
 
 export const isTimeframeTitle = (
   timeframe: string | SelectOptionObject,
-): timeframe is TimeframeTitle =>
-  Object.values(TimeframeTitle).includes(timeframe as TimeframeTitle);
+): timeframe is TimeframeTitle => isEnumMember(timeframe.toString(), TimeframeTitle);
 
 export const isRefreshIntervalTitle = (
   refreshInterval: string | SelectOptionObject,
 ): refreshInterval is RefreshIntervalTitle =>
-  Object.values(RefreshIntervalTitle).includes(refreshInterval as RefreshIntervalTitle);
+  isEnumMember(refreshInterval.toString(), RefreshIntervalTitle);
--- a/frontend/src/concepts/modelRegistry/context/ModelRegistryContext.tsx
+++ b/frontend/src/concepts/modelRegistry/context/ModelRegistryContext.tsx
@@ -28,6 +28,7 @@ export const ModelRegistryContext = React.createContext<ModelRegistryContextType
   hasCR: false,
   crInitializing: false,
   serverTimedOut: false,
+  // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
   apiState: { apiAvailable: false, api: null as unknown as ModelRegistryAPIState['api'] },
   ignoreTimedOut: () => undefined,
   refreshState: async () => undefined,
--- a/frontend/src/concepts/pipelines/content/compareRuns/metricsSection/roc/utils.ts
+++ b/frontend/src/concepts/pipelines/content/compareRuns/metricsSection/roc/utils.ts
@@ -2,14 +2,15 @@ import { JavaScriptValue } from 'google-protobuf/google/protobuf/struct_pb';
 import { ROCCurveConfig } from '~/concepts/pipelines/content/artifacts/charts/ROCCurve';
 import { ConfidenceMetric } from './types';
 
-export const isConfidenceMetric = (obj: JavaScriptValue): obj is ConfidenceMetric => {
-  const metric = obj as ConfidenceMetric;
-  return (
-    typeof metric.confidenceThreshold === 'number' &&
-    typeof metric.falsePositiveRate === 'number' &&
-    typeof metric.recall === 'number'
-  );
-};
+export const isConfidenceMetric = (obj: JavaScriptValue): obj is ConfidenceMetric =>
+  typeof obj === 'object' &&
+  obj !== null &&
+  'confidenceThreshold' in obj &&
+  'falsePositiveRate' in obj &&
+  'recall' in obj &&
+  typeof obj.confidenceThreshold === 'number' &&
+  typeof obj.falsePositiveRate === 'number' &&
+  typeof obj.recall === 'number';
 
 export const buildRocCurveConfig = (
   confidenceMetricsArray: ConfidenceMetric[],
--- a/frontend/src/concepts/pipelines/content/tables/PipelineFilterBar.tsx
+++ b/frontend/src/concepts/pipelines/content/tables/PipelineFilterBar.tsx
@@ -41,11 +41,13 @@ export function FilterToolbar<T extends string>({
   testId = 'filter-toolbar',
   ...toolbarGroupProps
 }: ToolbarFilterProps<T>): React.JSX.Element {
+  // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
   const keys = Object.keys(filterOptions) as Array<T>;
   const [open, setOpen] = React.useState(false);
   const [currentFilterType, setCurrentFilterType] = React.useState<T>(keys[0]);
   const isToolbarChip = (v: unknown): v is ToolbarChip & { key: T } =>
-    !!v && Object.keys(v as ToolbarChip).every((k) => ['key', 'node'].includes(k));
+    !!v && Object.keys(v).every((k) => ['key', 'node'].includes(k));
+  const filterItem = filterData[currentFilterType];
 
   return (
     <>
@@ -111,9 +113,7 @@ export function FilterToolbar<T extends string>({
           {filterOptionRenders[currentFilterType]({
             onChange: (value, label) =>
               onFilterUpdate(currentFilterType, label && value ? { label, value } : value),
-            ...(typeof filterData[currentFilterType] === 'string'
-              ? { value: filterData[currentFilterType] as string }
-              : (filterData[currentFilterType] as { label: string; value: string })),
+            ...(typeof filterItem === 'string' ? { value: filterItem } : filterItem),
           })}
         </ToolbarFilter>
       </ToolbarGroup>
--- a/frontend/src/concepts/pipelines/content/tables/usePipelineFilter.ts
+++ b/frontend/src/concepts/pipelines/content/tables/usePipelineFilter.ts
@@ -14,13 +14,11 @@ export enum FilterOptions {
   PIPELINE_VERSION = 'pipeline_version',
 }
 
-export const getDataValue = <T extends FilterProps['filterData'], R = T[keyof T]>(
-  data: R,
-): string | undefined => {
-  if (typeof data === 'string') {
+export const getDataValue = (data: string | { value: string } | undefined): string | undefined => {
+  if (typeof data === 'string' || typeof data === 'undefined') {
     return data;
   }
-  return (data as { label: string; value: string } | undefined)?.value;
+  return data.value;
 };
 
 const defaultFilterData: FilterProps['filterData'] = {
--- a/frontend/src/concepts/pipelines/context/PipelinesContext.tsx
+++ b/frontend/src/concepts/pipelines/context/PipelinesContext.tsx
@@ -51,6 +51,7 @@ const PipelinesContext = React.createContext<PipelineContext>({
   serverTimedOut: false,
   ignoreTimedOut: () => undefined,
   namespace: '',
+  // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
   project: null as unknown as ProjectKind,
   refreshState: async () => undefined,
   refreshAPIState: () => undefined,
@@ -58,7 +59,9 @@ const PipelinesContext = React.createContext<PipelineContext>({
     loading: false,
     data: null,
   }),
+  // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
   apiState: { apiAvailable: false, api: null as unknown as PipelineAPIState['api'] },
+  // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
   metadataStoreServiceClient: null as unknown as MetadataStoreServicePromiseClient,
 });
 
@@ -106,7 +109,7 @@ export const PipelineContextProvider = conditionalArea<PipelineContextProviderPr
     );
     if (DEV_MODE) {
       // Enables the use of this browser extension: https://github.com/SafetyCulture/grpc-web-devtools
-      // eslint-disable-next-line @typescript-eslint/no-explicit-any, @typescript-eslint/no-empty-function
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any, @typescript-eslint/no-empty-function, @typescript-eslint/consistent-type-assertions
       const enableDevTools = (window as any).__GRPCWEB_DEVTOOLS__ || (() => {});
       enableDevTools([client]);
     }
--- a/frontend/src/concepts/topology/PipelineDefaultTaskGroup.tsx
+++ b/frontend/src/concepts/topology/PipelineDefaultTaskGroup.tsx
@@ -33,7 +33,7 @@ type PipelinesDefaultGroupInnerProps = Omit<PipelinesDefaultGroupProps, 'element
 
 const DefaultTaskGroupInner: React.FunctionComponent<PipelinesDefaultGroupInnerProps> = observer(
   ({ element, selected, onSelect }) => {
-    const [hover, hoverRef] = useHover();
+    const [hover, hoverRef] = useHover<SVGGElement>();
     const popoverRef = React.useRef<SVGGElement>(null);
     const detailsLevel = element.getGraph().getDetailsLevel();
 
@@ -80,7 +80,7 @@ const DefaultTaskGroupInner: React.FunctionComponent<PipelinesDefaultGroupInnerP
 
     return (
       <Layer id={detailsLevel !== ScaleDetailsLevel.high && hover ? TOP_LAYER : DEFAULT_LAYER}>
-        <g ref={hoverRef as React.LegacyRef<SVGGElement>}>
+        <g ref={hoverRef}>
           {element.isCollapsed() ? (
             <Popover
               triggerRef={popoverRef}
--- a/frontend/src/concepts/topology/PipelineTaskEdge.tsx
+++ b/frontend/src/concepts/topology/PipelineTaskEdge.tsx
@@ -2,10 +2,10 @@ import * as React from 'react';
 import {
   DEFAULT_SPACER_NODE_TYPE,
   GraphElement,
-  Edge,
   EdgeTerminalType,
   observer,
   TaskEdge,
+  isEdge,
 } from '@patternfly/react-topology';
 
 interface PipelineTaskEdgeProps {
@@ -13,7 +13,10 @@ interface PipelineTaskEdgeProps {
 }
 
 const PipelineTaskEdge: React.FC<PipelineTaskEdgeProps> = ({ element, ...props }) => {
-  const edge = element as Edge;
+  if (!isEdge(element)) {
+    throw new Error('Element is not Edge');
+  }
+  const edge = element;
   return (
     <TaskEdge
       element={edge}
--- a/frontend/src/pages/learningCenter/DocTypeFilters.tsx
+++ b/frontend/src/pages/learningCenter/DocTypeFilters.tsx
@@ -4,6 +4,7 @@ import { FilterSidePanelCategory } from '@patternfly/react-catalog-view-extensio
 import FilterSidePanelCategoryItem from '~/components/FilterSidePanelCategoryItem';
 import { OdhDocument, OdhDocumentType } from '~/types';
 import { removeQueryArgument, setQueryArgument } from '~/utilities/router';
+import { asEnumMember, enumIterator } from '~/utilities/utils';
 import { DOC_TYPE_FILTER_KEY } from './const';
 import { useQueryFilters } from './useQueryFilters';
 
@@ -16,10 +17,11 @@ const DocTypeFilters: React.FC<DocTypeFiltersProps> = ({ categoryApps }) => {
   const docTypeFilters = useQueryFilters(DOC_TYPE_FILTER_KEY);
   const docCounts = React.useMemo(
     () =>
-      categoryApps.reduce(
+      categoryApps.reduce<{ [key in OdhDocumentType]: number }>(
         (acc, docApp) => {
-          if (docApp.spec.type in acc) {
-            acc[docApp.spec.type as keyof typeof acc]++;
+          const enumMember = asEnumMember(docApp.spec.type, OdhDocumentType);
+          if (enumMember) {
+            acc[enumMember]++;
           }
           return acc;
         },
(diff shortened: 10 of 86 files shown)


frontend/src/utilities/useAcceleratorProfileState.ts
```diff
@@ -65,7 +65,7 @@ const useAcceleratorProfileState = (
       } else {
         // check if there is accelerator usage in the container
         // this is to handle the case where the accelerator is disabled, deleted, or empty
-        const containerResourceAttributes = Object.values(ContainerResourceAttributes) as string[];
+        const containerResourceAttributes: string[] = Object.values(ContainerResourceAttributes);
         const possibleAcceleratorRequests = Object.entries(resources.requests ?? {})
           .filter(([key]) => !containerResourceAttributes.includes(key))
```
christianvogt, line 70: Can we use `isEnumMember` here?

frontend/src/pages/projects/screens/detail/overview/serverModels/deployedModels/DeployedModelsSection.tsx
```diff
@@ -57,12 +58,13 @@ const DeployedModelsSection: React.FC<DeployedModelsSectionProps> = ({ isMultiPl
...
     }
-    if (isMultiPlatform) {
+    const inferenceServiceKindArr: unknown = [];
+    if (isMultiPlatform && isInferenceServiceKindArr(inferenceServiceKindArr)) {
       const modelInferenceServices = modelServers.reduce((acc, modelServer) => {
         const services = getInferenceServiceFromServingRuntime(inferenceServices, modelServer);
         acc.push(...services);
         return acc;
-      }, [] as InferenceServiceKind[]);
+      }, inferenceServiceKindArr);
```
christianvogt, line 67: Parameterize `reduce` instead:

```
const modelInferenceServices = modelServers.reduce<InferenceServiceKind[]>(
```

frontend/src/pages/modelServing/screens/metrics/utils.tsx
```diff
@@ -220,7 +221,7 @@ export const checkConfigurationFieldsValid = (
 export const isMetricType = (
   metricType: string | SelectOptionObject,
 ): metricType is BiasMetricType =>
-  Object.values(BiasMetricType).includes(metricType as BiasMetricType);
+  Object.values(BiasMetricType).includes(asEnumMember(metricType.toString(), BiasMetricType)!);
```
christianvogt, line 224: Isn't this just:
```suggestion
  isEnumMember(metricType.toString(), BiasMetricType);
```

frontend/src/pages/modelServing/screens/global/utils.ts
```diff
@@ -11,10 +12,15 @@ export const getTokenDisplayName = (secret: SecretKind): string =>
...
-  InferenceServiceModelState.UNKNOWN;
+): InferenceServiceModelState => {
+  if (typeof is.status?.modelStatus.states.activeModelState !== 'undefined') {
+    return asEnumMember(is.status.modelStatus.states.activeModelState, InferenceServiceModelState)!;
+  }
+  if (typeof is.status?.modelStatus.states.targetModelState !== 'undefined') {
+    return asEnumMember(is.status.modelStatus.states.targetModelState, InferenceServiceModelState)!;
+  }
+  return InferenceServiceModelState.UNKNOWN;
+};
```
christianvogt, line 23: Update the enum functions to support `undefined` as an input.
```suggestion
): InferenceServiceModelState =>
  asEnumMember(is.status?.modelStatus.states.activeModelState, InferenceServiceModelState) ||
  asEnumMember(is.status?.modelStatus.states.targetModelState, InferenceServiceModelState) ||
  InferenceServiceModelState.UNKNOWN;
```

frontend/src/pages/modelRegistry/screens/RegisteredModelsArchive/RegisteredModelsArchiveListView.tsx
```diff
@@ -68,7 +69,7 @@ const RegisteredModelsArchiveListView: React.FC<RegisteredModelsArchiveListViewP
                   }))}
                   value={searchType}
                   onChange={(newSearchType) => {
-                    setSearchType(newSearchType as SearchType);
+                    setSearchType(asEnumMember(newSearchType, SearchType)!);
```
christianvogt, line 72: You need to guard against `null` before assigning. There's other code areas with the same issue.

frontend/src/concepts/pipelines/content/tables/usePipelineFilter.ts
```diff
@@ -20,7 +20,17 @@ export const getDataValue = <T extends FilterProps['filterData'], R = T[keyof T]
...
+    typeof data === 'object' &&
+    data !== null &&
+    'label' in data &&
+    typeof data.label === 'string' &&
+    'value' in data &&
+    typeof data.value === 'string'
+  ) {
+    return data.value;
+  }
+  return undefined;
```
christianvogt, line 33: Simplify entire function to:
```
export const getDataValue = (data: string | { value: string } | undefined): string | undefined => {
  if (typeof data === 'string' || typeof data === 'undefined') {
    return data;
  }
  return data.value;
};
```

frontend/src/concepts/pipelines/content/compareRuns/metricsSection/roc/utils.ts
```diff
@@ -2,14 +2,15 @@ import { JavaScriptValue } from 'google-protobuf/google/protobuf/struct_pb';
...
-};
+export const isConfidenceMetric = (obj: JavaScriptValue): obj is ConfidenceMetric =>
+  typeof obj === 'object' &&
+  obj !== null &&
+  'confidenceThreshold' in obj &&
+  'falsePositiveRate' in obj &&
+  'recall' in obj &&
+  obj.confidenceThreshold === 'number' &&
+  obj.falsePositiveRate === 'number' &&
+  obj.recall === 'number';
```
christianvogt, line 13:
```suggestion
  typeof obj.confidenceThreshold === 'number' &&
  typeof obj.falsePositiveRate === 'number' &&
  typeof obj.recall === 'number';
```

frontend/src/concepts/metrics/utils.ts
```diff
@@ -1,12 +1,15 @@
...
-  Object.values(TimeframeTitle).includes(timeframe as TimeframeTitle);
+  Object.values(TimeframeTitle).includes(asEnumMember(timeframe.toString(), TimeframeTitle)!);
 
 export const isRefreshIntervalTitle = (
   refreshInterval: string | SelectOptionObject,
 ): refreshInterval is RefreshIntervalTitle =>
-  Object.values(RefreshIntervalTitle).includes(refreshInterval as RefreshIntervalTitle);
+  Object.values(RefreshIntervalTitle).includes(
+    asEnumMember(refreshInterval.toString(), RefreshIntervalTitle)!,
+  );
```
christianvogt, line 15: Use `isEnumMember`

frontend/src/concepts/metrics/utils.ts
```diff
@@ -1,12 +1,15 @@
 import { SelectOptionObject } from '@patternfly/react-core/deprecated';
 import { TimeframeTitle, RefreshIntervalTitle } from '~/concepts/metrics/types';
+import { asEnumMember } from '~/utilities/utils';
 
 export const isTimeframeTitle = (
   timeframe: string | SelectOptionObject,
 ): timeframe is TimeframeTitle =>
-  Object.values(TimeframeTitle).includes(timeframe as TimeframeTitle);
+  Object.values(TimeframeTitle).includes(asEnumMember(timeframe.toString(), TimeframeTitle)!);
```
christianvogt, line 8: Use `isEnumMember`

frontend/src/components/DocCardBadges.tsx
```diff
@@ -32,35 +32,38 @@ const DocCardBadges: React.FC<DocCardBadgesProps> = ({ odhDoc }) => {
...
-          Failed
-        </Label>
-      ) : null}
-    </LabelGroup>
-  );
+    return (
+      <LabelGroup defaultIsOpen numLabels={3}>
+        <Tooltip content={DOC_TYPE_TOOLTIPS[docType]}>
+          <Label color={getLabelColorForDocType(docType)}>{label}</Label>
+        </Tooltip>
```
christianvogt, line 42: This is the only node that should be conditionally rendered against the `docType` and not the entire block.

frontend/src/utilities/useAcceleratorProfileState.ts
rsun19: Used `const possibleAcceleratorRequests = Object.entries(resources.requests ?? {})
          .filter(([key]) => isEnumMember(key, containerResourceAttributes))
          .map(([key, value]) => ({ identifier: key, count: value }));`

frontend/src/pages/modelRegistry/screens/RegisteredModelsArchive/RegisteredModelsArchiveListView.tsx
```diff
@@ -68,7 +69,10 @@ const RegisteredModelsArchiveListView: React.FC<RegisteredModelsArchiveListViewP
                   }))}
                   value={searchType}
                   onChange={(newSearchType) => {
-                    setSearchType(newSearchType as SearchType);
+                    const newSearchTypeInput = asEnumMember(newSearchType, SearchType);
+                    if (newSearchTypeInput !== null) {
+                      setSearchType(newSearchTypeInput);
+                    }
```
christianvogt, line 75: Can you explain to me why this is an error that required fixing while in `RegisteredModelListView` it wasn't?

frontend/src/pages/modelServing/screens/metrics/bias/BiasChart.tsx
```diff
@@ -19,28 +19,33 @@ const BiasChart: React.FC<BiasChartProps> = ({ biasMetricConfig }) => {
...
-    const metricData = data[modelMetricKey].data as PrometheusQueryRangeResponseDataResult[];
-
-    const values = metricData.find((x) => x.metric.request === id)?.values || [];
-
-    return {
-      ...data[modelMetricKey],
-      data: values,
-    };
+    const metricData = data[modelMetricKey].data;
+    if (isPrometheusQueryRangeResponseDataResultArr(metricData)) {
```
christianvogt, line 23: Look into fixing the type from the context so that the keys always gives back the matching typed value.

frontend/src/pages/modelServing/screens/metrics/bias/BiasConfigurationPage/BiasConfigurationModal/MetricTypeField.tsx
```diff
@@ -37,9 +39,15 @@ const MetricTypeField: React.FC<MetricTypeFieldProps> = ({ fieldId, value, onCha
           <SelectOption
             key={type}
             value={type}
-            description={METRIC_TYPE_DESCRIPTION[type as keyof typeof BiasMetricType]}
+            description={
+              asEnumMember(type, BiasMetricType) !== null
+                ? METRIC_TYPE_DESCRIPTION[asEnumMember(type, BiasMetricType)!]
+                : null
```
christianvogt, line 45: We need a function `mapEnum(<enum>)` which should return all the enum literals as an array.
I think this would solve several other areas in in this PR related to iterating over enums.

frontend/src/pages/modelServing/screens/metrics/performance/ModelMeshMetrics.tsx
```diff
@@ -5,11 +5,13 @@ import {
...
   ModelServingMetricsContext,
 } from '~/pages/modelServing/screens/metrics/ModelServingMetricsContext';
-import { ContextResourceData, PrometheusQueryRangeResultValue } from '~/types';
 import EnsureMetricsAvailable from '~/pages/modelServing/screens/metrics/EnsureMetricsAvailable';
+import { isPrometheusQueryRangeResultValueNotResponseDataResultWithContext } from '~/utilities/utils';
 
 const ModelMeshMetrics: React.FC = () => {
   const { data } = React.useContext(ModelServingMetricsContext);
+  const requestSuccess = data[ModelMetricType.REQUEST_COUNT_SUCCESS];
+  const requestCountFailed = data[ModelMetricType.REQUEST_COUNT_FAILED];
```
christianvogt, line 14: See previous comment related to improving the context type.

frontend/src/pages/modelServing/screens/metrics/performance/ServerGraphs.tsx
```diff
@@ -9,80 +9,82 @@ import {
...
+  isPrometheusQueryRangeResultValueNotResponseDataResultWithContext,
+} from '~/utilities/utils';
 
 const ServerGraphs: React.FC = () => {
   const { data } = React.useContext(ModelServingMetricsContext);
 
+  const requestCount = data[ServerMetricType.REQUEST_COUNT];
+  const avgResponseTime = data[ServerMetricType.AVG_RESPONSE_TIME];
+  const cpuUtils = data[ServerMetricType.CPU_UTILIZATION];
+  const memoryUtils = data[ServerMetricType.MEMORY_UTILIZATION];
```
christianvogt, line 24: See previous comment related to improving the context type.

frontend/src/pages/projects/screens/spawner/environmentVariables/EnvTypeSelectField.tsx
```diff
@@ -32,10 +33,13 @@ const EnvTypeSelectField: React.FC<EnvTypeSelectFieldProps> = ({
...
               onSelect={(e, value) => {
                 if (typeof value === 'string') {
-                  onUpdate({
-                    type: value as EnvironmentVariableType,
-                  });
-                  setOpen(false);
+                  const enumValue = asEnumMember(value, EnvironmentVariableType);
+                  if (enumValue !== null) {
+                    onUpdate({
+                      type: asEnumMember(value, EnvironmentVariableType),
```
christianvogt, line 39:
```suggestion
                      type: enumValue,
```

frontend/src/utilities/NavData.tsx
```diff
@@ -29,8 +29,10 @@ export type NavDataGroup = NavDataCommon & {
 export type NavDataItem = NavDataHref | NavDataGroup;
 
 export const isNavDataHref = (navData: NavDataItem): navData is NavDataHref =>
+  // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
   !!(navData as NavDataHref).href;
 export const isNavDataGroup = (navData: NavDataItem): navData is NavDataGroup =>
+  // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
   !!(navData as NavDataGroup).children;
```
christianvogt, line 36:
```suggestion
export const isNavDataHref = (navData: NavDataItem): navData is NavDataHref => 'href' in navData;
export const isNavDataGroup = (navData: NavDataItem): navData is NavDataGroup => 'children' in navData;
```

frontend/src/utilities/useDraggableTable.ts
```diff
@@ -159,6 +159,7 @@ const useDraggableTable = (
   );
 
   const onDragEnd = React.useCallback<UseDraggableTable['rowProps']['onDrop']>((evt) => {
+    // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
     const target = evt.currentTarget as HTMLTableRowElement;
```
christianvogt, line 163: Can you remove the cast altogether?

frontend/src/concepts/pipelines/context/PipelinesContext.tsx
```diff
@@ -50,14 +50,17 @@ const PipelinesContext = React.createContext<PipelineContext>({
...
   refreshState: async () => undefined,
   refreshAPIState: () => undefined,
   getJobInformation: () => ({
     loading: false,
     data: null,
   }),
+  // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
   apiState: { apiAvailable: false, api: null as unknown as PipelineAPIState['api'] },
+  // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
   metadataStoreServiceClient: null as unknown as MetadataStoreServicePromiseClient,
```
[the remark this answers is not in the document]
andrewballantyne: Agreed, I don't think we care about outside of Context -- that's a dev problem, and it would be pretty evident and fixed before code ever hit runtime.

🤔 So... we could change this to `{} as Type` but my concern is we have code looking at some value being a truthy value -- like in a loading state internal to the context component.

I don't see that as a high chance, but it is a non-zero chance. Perhaps ignoring this `as` usage and coming back with a Tech Debt is the better way to go. In the end this PR is not going to get super great at test coverage due to the wide reaching changes. Hopefully Cypress & Jest play ball and catch all the corners, but we are not 100% test covered. 

@christianvogt does that sound reasonable?

frontend/src/concepts/topology/PipelineDefaultTaskGroup.tsx
```diff
@@ -80,6 +80,7 @@ const DefaultTaskGroupInner: React.FunctionComponent<PipelinesDefaultGroupInnerP
 
     return (
       <Layer id={detailsLevel !== ScaleDetailsLevel.high && hover ? TOP_LAYER : DEFAULT_LAYER}>
+        {/* eslint-disable-next-line @typescript-eslint/consistent-type-assertions*/}
         <g ref={hoverRef as React.LegacyRef<SVGGElement>}>
```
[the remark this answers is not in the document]
rsun19: Do you mean `<g ref={useHover<SVGGElement>()}>`? Because that doesn't work for me.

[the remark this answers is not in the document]
christianvogt: `const [hover, hoverRef] = useHover<SVGGElement>();`

frontend/src/concepts/topology/PipelineTaskEdge.tsx
```diff
@@ -13,6 +13,7 @@ interface PipelineTaskEdgeProps {
 }
 
 const PipelineTaskEdge: React.FC<PipelineTaskEdgeProps> = ({ element, ...props }) => {
+  // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
   const edge = element as Edge;
```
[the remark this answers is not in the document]
christianvogt: @rsun19 hmmm this may not be fixable in topology. My suggestion is causing other issues.

Instead we can use `isEdge` from the topology package to first ensure it is the correct type

```
if (!isEdge(element)) {
  throw new Error(...);
}
```

frontend/src/concepts/pipelines/context/PipelinesContext.tsx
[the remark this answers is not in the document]
christianvogt: Some code reaching for the context today without the provider is very likely to break.

We are always render the providers where we expect child components to depend on them. The use case you bring up would be very edge and likely overlooked by the dev who would have implemented such a case. ie to render a loading state that depends on the context in which the loading state is rendered within thus using the default value. <_snake eats itself_>

@andrewballantyne I think it's fair that we tackle these contexts in a separate PR instead of along with the rest.

frontend/src/pages/modelRegistry/screens/RegisteredModelsArchive/RegisteredModelsArchiveListView.tsx
rsun19: I must have missed the null check

frontend/src/pages/modelServing/screens/metrics/bias/BiasConfigurationPage/BiasConfigurationModal/MetricTypeField.tsx
rsun19: Can I have more clarification on what that would look like? i.e. what other functionality it will bring that `asEnumMember` doesn't?

frontend/src/concepts/pipelines/content/tables/PipelineFilterBar.tsx
```diff
@@ -41,11 +42,12 @@ export function FilterToolbar<T extends string>({
   testId = 'filter-toolbar',
   ...toolbarGroupProps
 }: ToolbarFilterProps<T>): React.JSX.Element {
+  // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
   const keys = Object.keys(filterOptions) as Array<T>;
```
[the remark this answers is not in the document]
rsun19: Agreed to tackle this in another PR

frontend/src/utilities/useAcceleratorProfileState.ts
```diff
@@ -65,9 +65,9 @@ const useAcceleratorProfileState = (
       } else {
         // check if there is accelerator usage in the container
         // this is to handle the case where the accelerator is disabled, deleted, or empty
-        const containerResourceAttributes = Object.values(ContainerResourceAttributes) as string[];
+        const containerResourceAttributes = Object.values(ContainerResourceAttributes);
         const possibleAcceleratorRequests = Object.entries(resources.requests ?? {})
-          .filter(([key]) => !containerResourceAttributes.includes(key))
+          .filter(([key]) => isEnumMember(key, containerResourceAttributes))
```
rsun19, line 70: Any idea why the original code opted to include keys that aren't in `containerResourceAttributes`? Am I reading this correctly?

christianvogt: Keep the same logic as the old code. Any changes to logic should log a new issue if necessary.

Including @Gkrumbach07 since he authored this code.

frontend/src/pages/modelServing/screens/global/utils.ts
```diff
@@ -12,18 +13,31 @@ export const getTokenDisplayName = (secret: SecretKind): string =>
 export const getInferenceServiceActiveModelState = (
   is: InferenceServiceKind,
 ): InferenceServiceModelState =>
-  <InferenceServiceModelState | undefined>is.status?.modelStatus?.states?.activeModelState ||
-  <InferenceServiceModelState | undefined>is.status?.modelStatus?.states?.targetModelState ||
+  (is.status &&
```
rsun19, line 16: Made more explicit type guarding due to these type-check tests:

```
src/pages/modelServing/screens/global/utils.ts:16:16 - error TS18048: 'is.status.modelStatus' is possibly 'undefined'.

16   asEnumMember(is.status?.modelStatus.states.activeModelState, InferenceServiceModelState) ||
```

frontend/src/utilities/useAcceleratorProfileState.ts
rsun19: Ok, I raised this comment because I got unexpected behavior with the `isEnumMember` function. But that's because I put the array as the arg instead of the enum. Oops... Resolving.

frontend/src/pages/modelServing/screens/global/utils.ts
christianvogt:
```
export const getInferenceServiceActiveModelState = (
  is: InferenceServiceKind,
): InferenceServiceModelState =>
  asEnumMember(is.status?.modelStatus?.states?.activeModelState, InferenceServiceModelState) ||
  asEnumMember(is.status?.modelStatus?.states?.targetModelState, InferenceServiceModelState) ||
  InferenceServiceModelState.UNKNOWN;
```

frontend/src/pages/modelServing/screens/global/utils.ts
```diff
@@ -12,18 +13,31 @@ export const getTokenDisplayName = (secret: SecretKind): string =>
...
+    typeof is.status.modelStatus.states !== 'undefined'
+  ) {
+    activeModelState = is.status.modelStatus.states.activeModelState;
+    targetModelState = is.status.modelStatus.states.targetModelState;
+    lastFailureMessage = is.status.modelStatus.lastFailureInfo?.message;
+  }
+
   const isFailedToLoad = activeModelState === failedToLoad || targetModelState === failedToLoad;
 
-  const lastFailureMessage = is.status?.modelStatus?.lastFailureInfo?.message;
```
christianvogt, line 26: Why change this function?

frontend/src/typeHelpers.ts
```diff
@@ -162,6 +162,7 @@ export type ExactlyOne<T> = AtMostOne<T> & AtLeastOne<T>;
 export const isInEnum =
   <T extends { [s: string]: unknown }>(e: T) =>
   (token: unknown): token is T[keyof T] =>
+    // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
     Object.values(e).includes(token as T[keyof T]);
```
christianvogt, line 166: Can we remove this function and use `isEnumMember`.

frontend/src/pages/learningCenter/DocTypeFilters.tsx
```diff
@@ -52,26 +53,18 @@ const DocTypeFilters: React.FC<DocTypeFiltersProps> = ({ categoryApps }) => {
...
-                (e.target as React.AllHTMLAttributes<HTMLInputElement>).checked || false,
-              )
-            }
-            title={docType}
-          >
-            {`${docType} (${docCounts[value]})`}
-          </FilterSidePanelCategoryItem>
-        );
-      })}
+      {Object.entries(OdhDocumentType).map(([docType, value]) => (
```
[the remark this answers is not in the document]
rsun19: the subsequent code references docType in `[docType, value]` though; how should I handle that? Since I think `enumIterator` only returns the value part of it

frontend/src/pages/modelServing/screens/global/utils.ts
rsun19: > Made more explicit type guarding due to these type-check tests:
> 
> ```
> src/pages/modelServing/screens/global/utils.ts:16:16 - error TS18048: 'is.status.modelStatus' is possibly 'undefined'.
> 
> 16   asEnumMember(is.status?.modelStatus.states.activeModelState, InferenceServiceModelState) ||
> ```

Was a side effect of this; I just reverted the change.

frontend/src/concepts/modelRegistry/context/ModelRegistryContext.tsx
```diff
@@ -28,6 +28,7 @@ export const ModelRegistryContext = React.createContext<ModelRegistryContextType
   hasCR: false,
   crInitializing: false,
   serverTimedOut: false,
+  // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
   apiState: { apiAvailable: false, api: null as unknown as ModelRegistryAPIState['api'] },
```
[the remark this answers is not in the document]
rsun19: Yes

frontend/src/pages/modelServing/screens/metrics/bias/BiasConfigurationPage/BiasConfigurationModal/MetricTypeField.tsx
```diff
@@ -33,13 +35,9 @@ const MetricTypeField: React.FC<MetricTypeFieldProps> = ({ fieldId, value, onCha
...
-        {Object.keys(BiasMetricType).map((type) => (
-          <SelectOption
-            key={type}
-            value={type}
-            description={METRIC_TYPE_DESCRIPTION[type as keyof typeof BiasMetricType]}
-          >
-            {METRIC_TYPE_DISPLAY_NAME[type as keyof typeof BiasMetricType]}
+        {enumIterator(BiasMetricType).map(([, type]) => (
+          <SelectOption key={type} value={type} description={METRIC_TYPE_DESCRIPTION[type]}>
+            {METRIC_TYPE_DISPLAY_NAME[type]}
```
christianvogt, line 40: Shouldn't this use the first entry in the iterator since the original code used `Object.keys`
```suggestion
        {enumIterator(BiasMetricType).map(([type]) => (
          <SelectOption key={type} value={type} description={METRIC_TYPE_DESCRIPTION[type]}>
            {METRIC_TYPE_DISPLAY_NAME[type]}
```

PR #2894, opened 2024-06-10
Added migration for disabling trusty bias for RHOAI

andrewballantyne:
https://issues.redhat.com/browse/RHOAIENG-7811

## Description

Disable Trusty Bias when the app starts.

## How Has This Been Tested?

Manually, connecting to a RHOAI cluster (and an ODH as sanity nothing happens).

## Test Impact

None, we don't have a framework for testing on the backend.

## Request review criteria:

If you have UI changes: 

/cc @alexcreasy

### DIFF
--- a/backend/src/utils/resourceUtils.ts
+++ b/backend/src/utils/resourceUtils.ts
@@ -33,6 +33,7 @@ import { blankDashboardCR } from './constants';
 import { getIsAppEnabled, getRouteForApplication, getRouteForClusterId } from './componentUtils';
 import { createCustomError } from './requestUtils';
 import { getDetectedAccelerators } from '../routes/api/accelerators/acceleratorUtils';
+import { RecursivePartial } from '../typeHelpers';
 
 const dashboardConfigMapName = 'odh-dashboard-config';
 const consoleLinksGroup = 'console.openshift.io';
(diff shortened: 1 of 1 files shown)


backend/src/utils/resourceUtils.ts
```diff
@@ -905,6 +910,64 @@ export const migrateTemplateDisablement = async (
     });
 };
 
+/**
+ * As we release Dashboard v2.24.0 we will be re-enabling Trusty for RHOAI, but not for Bias Metrics.
+ * This means we will need to make sure Bias is in a state of disabled while potentially the client
+ * reads Bias as
```
mturley, line 916: Looks like there was more to this sentence?

backend/src/utils/resourceUtils.ts
```diff
@@ -905,6 +910,64 @@ export const migrateTemplateDisablement = async (
...
+ */
+export const migrateBiasTrustyForRHOAI = async (
+  fastify: KubeFastifyInstance,
+  dashboardConfig: DashboardConfig,
+): Promise<DashboardConfig> => {
+  const { namespace, customObjectsApi } = fastify.kube;
+
+  // TODO: There should be a better way to go about this... but the namespace is unlikely to ever change
```
mturley, line 924: Should we just create a constant for `'redhat-ods-applications'` in `backend/src/utils/constants.ts`, or maybe a `isRHOAI` util?

mturley (commented):
Mostly LGTM, couple of nits

backend/src/utils/resourceUtils.ts
andrewballantyne: hahah, I'll fix.

backend/src/utils/resourceUtils.ts
andrewballantyne: It's a valid question -- I feel this is not something I want to encourage... but maybe that's a good use for a utility. I'll go ahead and document the function. Thanks!

lucferbux: We used to have something like that for prometheus, I think we remove it in a refactoring.

Also, DSC has a `release` section, detailing wether it's open data hub or rhoai, maybe that's something that we could use to determine wether we are using ODH or rhoai right?

mturley (approved):
LGTM 👍

backend/src/utils/resourceUtils.ts
```diff
@@ -56,24 +57,28 @@ let buildsWatcher: ResourceWatcher<BuildStatus>;
...
+  dashboardName: process.env['DASHBOARD_CONFIG'] || dashboardConfigMapName,
+};
+
 const fetchDashboardCR = async (fastify: KubeFastifyInstance): Promise<DashboardConfig[]> => {
-  const dashboardName = process.env['DASHBOARD_CONFIG'] || dashboardConfigMapName;
-  return fetchOrCreateDashboardCR(fastify, dashboardName).then((dashboardCR) => {
-    return migrateTemplateDisablement(fastify, dashboardCR).then((dashboardCR) => [dashboardCR]);
-  });
+  return fetchOrCreateDashboardCR(fastify)
+    .then((dashboardCR) => migrateTemplateDisablement(fastify, dashboardCR))
```
lucferbux, line 69: `migrateTemplateDisablement` was created almost a year ago for a upgrade path from 2.11, it was intended to be removed once we dropped support from that version, isn't that the case right now?

backend/src/utils/resourceUtils.ts
```diff
@@ -905,6 +910,66 @@ export const migrateTemplateDisablement = async (
     });
 };
 
+/**
+ * TODO: There should be a better way to go about this... but the namespace is unlikely to ever change
+ */
+export const isRHOAI = (fastify: KubeFastifyInstance): boolean =>
+  fastify.kube.namespace === 'redhat-ods-applications';
```
lucferbux, line 917: DSC has a `release` attribute under status, and they detail in there wether we use ODH or RHOAI, it would be yet another request and I don't think the namespace would change in the future, but that's another possibility to determine the installation

lucferbux (commented):
Just left two comments, just suggestions, I can approve it in case we don't wanna go with either those two things.

backend/src/utils/resourceUtils.ts
andrewballantyne: 🤷 wanna log something @lucferbux ? This is well outside of scope.

backend/src/utils/resourceUtils.ts
andrewballantyne: It does -- but it's custom text and requires a fetch to find out -- I'd rather not build in the logic to fetch it to do this... I did think of that. This is reliable if not awkward.

backend/src/utils/resourceUtils.ts
lucferbux: Yes, I agree, I've created https://issues.redhat.com/browse/RHOAIENG-8499 for this.

backend/src/utils/resourceUtils.ts
lucferbux: Yeah, we haven't changed the namespace and I would assume if done, we should have a long list of requirements to change and that can fall into that bucket, I think this a reliable way to check the operator.

lucferbux (approved):
/lgtm

PR #2906, opened 2024-06-12
Restructure the manifest files

andrewballantyne:
https://issues.redhat.com/browse/RHOAIENG-5506

## Description

Restructure the manifest files for the 3 types of deployments.

The goal is to make pure entry points and self link all needed parts, that way the operator doesn't need to know the internals of our manifests folder layout. 

The three entry points are:
* `/manifests/odh/kustomization.yaml` - ODH Operator
* `/manifests/rhoai/onprem/kustomization.yaml` - RHOAI Operator for Self Managed deployments (onprem is an old name)
* `/manifests/rhoai/addon/kustomization.yaml` - RHOAI OPerator for Managed deployments

## How Has This Been Tested?

Added testing steps in the root README.md -- these steps will only apply going forward and don't help with the change itself as the operator owned a lot of the functionality.

Need to get the answers to the following questions:
- [x] Where would `params.env` need to live for RHOAI to make use of it
    - Might be okay in its current position -- will need to check with Operator Team
- [x] Operator needs a mirror effort here to restructure their code to point at root kustomize.yamls
    - https://github.com/opendatahub-io/opendatahub-operator/pull/1065
- [x] How to best test this ... custom operator?
    - Using `kustomize build` commands -- see the root README for more
    - Using the custom operator image: http://quay.io/wenzhou/rhods-operator-catalog:v2.13.851-1 (replace the operator image)

[Topic with Operator Team (slack)](https://redhat-internal.slack.com/archives/C064Z5S84F3/p1718223737561859)

**OUT OF SCOPE** of this PR is handling the out of date kustomize format. I'll follow up with proper updates to that when I'm not moving 100 some odd files.

## Test Impact

None, manifest files.

## Request review criteria:

If you have UI changes: 

I'll need help working through this -- cc @zdtsw @VaishnaviHire @lucferbux


manifests/README.md
```diff
@@ -1,32 +1,55 @@
...
 
+## Installation (ODH)
+
+You can use the `kustomize` tool to process the manifest for the `oc apply` command.
+
+```markdown
+# Set the namespace in the manifest where you want to deploy the dashboard
+kustomize edit set namespace <DESTINATION NAMESPACE>
+kustomize build common | oc apply -f -
+kustomize build overrides/base | oc apply -f -
```
lucferbux, line 27: Now that we've changed the folders, we had some scripts under `install` (`deploy.sh`, `undeploy.sh`), to automate the development section, we can either remove them or adapt them to this new reestructure.

manifests/README.md
```diff
@@ -1,32 +1,55 @@
...
+## Testing Changes
+
+One way to test changes locally is to generate the full structure before your changes, and then again after your changes.
+
+Before:
+```markdown
+# Generate the files before your changes for a baseline
+kustomize build rhoai/addon > before-rhoai-addon-test.yaml
+kustomize build rhoai/onprem > before-rhoai-onprem-test.yaml
+kustomize build odh > before-odh-test.yaml
```
lucferbux, line 39: We don't use `makefile` a lot, it's only scoped right now for builds and deployments, but it might be nice to have the manifest testing under a phony target there too.

manifests/README.md
```diff
@@ -1,32 +1,55 @@
...
-For more information, visit the project [GitHub repo](https://github.com/opendatahub-io/odh-dashboard).
+## Adding/Modifying Manifests
 
-### Folders
-1. base: contains all the necessary yaml files to install the dashboard
+Rules for keeping the manifest files in a sane order:
 
-##### Installation
-Use the `kustomize` tool to process the manifest for the `oc apply` command.
+1. When adding a new type of thing, always make it have its own folder; be sure to add the reference to the parent folder's `kustomziation.yaml` (if applicable)
```
lucferbux, line 14: nit: remove the empty space at the end of the line

manifests/overrides/README.md
```diff
@@ -0,0 +1,5 @@
+# Overridable Manifests
+
+These are manifest files that are intended to be used as-is or to be overridden.
+
+> **Note:** There should be note root `kustomization.yaml` as each folder should be treated as a specific target for inclusion or overrides.
```
lucferbux, line 5: I get the new structure, and I'm not against it given our particular set of deployments, but I find the override folder a little bit misleading, kustomize is set on the premise of [bases and overlays](https://kubernetes.io/docs/tasks/manage-kubernetes-objects/kustomization/#bases-and-overlays). You tend to find that kind of structure so this might confuse newcomers that are used to the common kustomize flow.

lucferbux (changes requested):
I've just tested the odh deployment and it seems to be working flawlesly, I've left a couple of thoughts and nits, nothing serious.

A part from that, other suggestion would be having documented in the operator the installation instructions (steps that the operator follows) and link that into our docs, I tend to go to the operator manifest when I have questions about manifest additions. cc @zdtsw

lucferbux:
And my other comment is that I find that both `common` and `overrides` folder seem to be the `base` files from manifest and `odh`, `rhoai/addon`, `rhoai/onprem` as overlays. I get the difference, but why don't label the first two under a `base` folder and the other three under a `overlay`?

manifests/README.md
andrewballantyne: Interesting, I'll look into it.

manifests/overrides/README.md
andrewballantyne: I'll respond here since these threads seem related and GitHub sucks at root-level conversations 🙂 

> And my other comment is that I find that both `common` and `overrides` folder seem to be the `base` files from manifest and `odh`, `rhoai/addon`, `rhoai/onprem` as overlays. I get the difference, but why don't label the first two under a `base` folder and the other three under a `overlay`?
> 
> It's just semantics but might stick to what kustomize suggests organizing the files.

([from here](https://github.com/opendatahub-io/odh-dashboard/pull/2906#issuecomment-2186648170))

@lucferbux The way we used overlays were highly confusing to me, so I did away with it in the first pass. If we find having overlays to be specifically beneficial, we can craft those as needed and maybe create a new overlays folder; I didn't see anything useful in the last one and deleted them in the restructure effort.

We don't want structure out manifests to be "layers on top of each other" -- since there are 3 semi-unique entry points. It's not a matter of "a is the base" "b is additions on that"... the way we had it was "overlays were a mixed bag of RHOAI manifests and optional overlays"; I'm not sure this was the intended design of kustomize's "bases and overlays".

I'll think on if there is a better way to shape the use of overlays -- but honestly, it feels like we have a use-case that doesn't work for that design and for greater management and simplicity, we do away with it. I appreciate the input, but our old usage of `overlays` felt confusing and left a lot of questions about "what is what" and "is this for x deployment". Overlays always struck me as "additions to an existing setup" and not "ways to configure multiple different core deployment types". I feel overlays need to speak to "optional opt-ins" and not core fundamental stacking points.

Happy to discuss this further if you feel this is a confusing setup for `kustomize` experts. Maybe there is a middle ground here, but I don't currently see a need for overlays.

andrewballantyne: Also, I forgot to address the other point -- `overrides` is perhaps not the best name -- I could call them `overlays` but I feel that's still confusing use of `overlays`. I'm open to discussing alternatives here.

andrewballantyne: The text on the kustomize website definitely feels like maybe we call `overrides` something like "custom-bases" or "core-bases" -- and that sorta fits the design overlap... but it'll be "different" than kustomizes suggestion.

manifests/odh/params.env
```diff
@@ -0,0 +1,5 @@
+# Injected variables from the Operator
+odh-dashboard-image=quay.io/opendatahub/odh-dashboard:main
+admin_groups=odh-admins
+dashboard-url=https://www.redhat.com
+section-title=Open Data Hub
```
zdtsw, line 5:
```suggestion
section-title=OpenShift Open Data Hub
```
should we just have this set to the correct value by default? same as for the other 2 params.env files. 
Operator can still inject this value

manifests/rhoai/addon/params.env
```diff
@@ -0,0 +1,5 @@
+# Injected variables from the Operator
+odh-dashboard-image=quay.io/opendatahub/odh-dashboard:main
+admin_groups=rhods-admins
+dashboard-url=https://www.redhat.com
+section-title=OpenShift AI
```
zdtsw, line 5:
```suggestion
admin_groups=dedicated-admins
section-title=OpenShift Managed Services
```

manifests/rhoai/onprem/params.env
```diff
@@ -0,0 +1,5 @@
+# Injected variables from the Operator
+odh-dashboard-image=quay.io/opendatahub/odh-dashboard:main
+admin_groups=rhods-admins
+dashboard-url=https://www.redhat.com
+section-title=OpenShift AI
```
zdtsw, line 5:
```suggestion
admin_groups=rhods-admins
section-title=OpenShift Self Managed Services
```

manifests/odh/params.env
andrewballantyne: Not sure it matters -- I was just trying to put _some value_ there -- I could make it more of a description than an actual value. I do not think it's important to be aligned on this. It was more of an example than anything else.

manifests/rhoai/addon/params.env
andrewballantyne: [See other comment](https://github.com/opendatahub-io/odh-dashboard/pull/2906#discussion_r1652958821)

manifests/rhoai/onprem/params.env
andrewballantyne: [See other comment](https://github.com/opendatahub-io/odh-dashboard/pull/2906#discussion_r1652958821)

manifests/odh/params.env
zdtsw: ok, i thought you want to put the "real" value here. but if it is just placeholder, i am fine with any value

manifests/overrides/README.md
andrewballantyne: Okay, I have updated it somewhat based on your feedback... @lucferbux I still don't believe we need to be so basic with our definitions... `bases` & `overlays` is a naive structure that makes it difficult to manage 3 deployments in a way that is both readable and functional. If that causes friction with kustomize experts, then it is a price to pay for maintainability imo.

lucferbux: Ok, that's fair tbh, we can keep that structure, it's fine to me having this layout.

manifests/README.md
andrewballantyne: I'm thinking in order to do this via the makefile -- it would need to sorta do:

* be on your branch with changes
* take a snapshot of the output (after files)
* checkout "main"(?) (hopefully your local main is up to date?)
* take a snapshot of the output (before files)
* report the difference (output files)

And then I guess try to get you back on your branch again? Not sure how that works out. It's brittle and requires everyone to have the same structure of git. I'm not inclined to do this as the steps are copy-able and paste-able into the console pretty easily. 

It was an interesting thought though. We should log something @lucferbux if there is a path forward I'm not seeing here.

andrewballantyne:
Tested with Wen on her custom catalog source images -- going to be merging this on Tuesday. Monday I'm OOO and today is Code Freeze for 2.11 -- just to avoid the noise, we will wait until post that.

lucferbux (changes requested):
I've left a nit and a question, so far the rest looks great.
I've managed to get a deployment in ODH and I've managed to get a cluster to install the rhoai image, it's giving me some issues so I'll take a look tomorrow with @zdtsw

manifests/rhoai/shared/consolelink/consolelink.yaml
```diff
@@ -9,7 +9,7 @@
   value: Red Hat OpenShift AI
 - op: replace
   path: /spec/applicationMenu/section
-  value: <section-title>
+  value: $(section-title)
```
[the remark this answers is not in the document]
zdtsw: we (Operator) will not do the replacement in the manifests for these variable, 
instead the new solution is to dump these variable into params.env, then you (Dashboard) and use these variable in your manifests.

manifests/rhoai/onprem/params.env
```diff
@@ -0,0 +1,5 @@
+# Injected variables from the Operator
+odh-dashboard-image=quay.io/opendatahub/odh-dashboard:main
+admin_groups=rhods-admins
+dashboard-url=https://www.redhat.com
```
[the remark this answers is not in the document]
zdtsw: for the odh build, the one will be overriden is in the manifests/odh/params.env
this file is only used for self-management case

lucferbux:
/lgtm

Everything seems great un rhoai too

PR #2913, opened 2024-06-13
Support model metrics for KServe (RHOAIENG-6560) (RHOAIENG-6561)

alexcreasy:
Closes: [RHOAIENG-6560](https://issues.redhat.com//browse/RHOAIENG-6560) [RHOAIENG-6561](https://issues.redhat.com//browse/RHOAIENG-6561)

## Description

Implements metrics graphs for supported Kserve runtimes. Implements empty state for when a specific serving runtime is not supported.

Graphs and queries are now defined by a config map created by the KServe operator.

odh-dashboard-opendatahub apps rosa vedant-metrics chbf p3 openshiftapps com_projects_models_metrics_model_mnist_performance (1)

odh-dashboard-opendatahub apps rosa vedant-metrics chbf p3 openshiftapps com_projects_models_metrics_model_mnist_performance (2)

odh-dashboard-opendatahub apps rosa vedant-metrics chbf p3 openshiftapps com_projects_models_metrics_model_mnist_performance (3)

odh-dashboard-opendatahub apps rosa vedant-metrics chbf p3 openshiftapps com_projects_models_metrics_model_mnist_performance

NOTE: There are some outstanding bugs in the backend:
- The definition provided for the requests chart is invalid, right now we can only verify the dashboard handles this gracefully
- Graph titles are defined on the backend, these need updating to fix case / match modelmesh charts.
- There is an error in the Memory Usage query that returns an absolute value, instead of a percentage. As such you won't see any data on that chart.

## How Has This Been Tested?

Environment setup:
- Install latest ODH/RHOAI
- Use the following [DSC](https://gist.github.com/VedantMahabaleshwarkar/634b4e1b4e434280775c477135d81454#file-gistfile1-txt-L2-L18)
- Add the following [SR ](https://gist.github.com/VedantMahabaleshwarkar/634b4e1b4e434280775c477135d81454#file-gistfile1-txt-L20-L69)for single model serving
- Use the serving runtime to create a mnist model. Can use the [rhods-public s3 bucket](https://s3.console.aws.amazon.com/s3/buckets/rhods-public/?region=us-east-1&tab=objects) with your access+secret key
model path is : kserve-samples/mnist/
(can keep the require service account box unchecked for easier curls..)
To perform requests against the model
curl --insecure {isvc_url}/v2/models/mnist/infer -d @./input-onnx.json 
where,
isvc_url=$(oc get isvc -n <your_project> mnist -o jsonpath='{.status.url}') 
and
you can find input-onnx.json [here](https://github.com/opendatahub-io/modelmesh-serving/blob/main/opendatahub/quickstart/common_manifests/input-onnx.json)

Instructions:
- Navigate to kserve model `mnist`
- Hit the model with some data (spam a number of requests as directed above)
- Validate charts function correctly (use the timeframe / refresh interval controls.
- Navigate to model `sklearn-v2-iris metrics`
- Validate the unsupported runtime state is visible.

## Test Impact

- Cypress tests added
- Unit tests of utility functions
- Tests of new hooks to follow after.

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/api/prometheus/kservePerformanceMetrics.ts
+++ b/frontend/src/api/prometheus/kservePerformanceMetrics.ts
@@ -1,0 +1,209 @@
+import React from 'react';
+import { KserveMetricGraphDefinition } from '~/concepts/metrics/kserve/types';
+import { defaultResponsePredicate } from '~/api/prometheus/usePrometheusQueryRange';
+import { SupportedArea, useIsAreaAvailable } from '~/concepts/areas';
+import { TimeframeTitle } from '~/concepts/metrics/types';
+import useQueryRangeResourceData from '~/api/prometheus/useQueryRangeResourceData';
+import { PendingContextResourceData, PrometheusQueryRangeResultValue } from '~/types';
+import { DEFAULT_PENDING_CONTEXT_RESOURCE } from '~/api/prometheus/const';
+
+type RequestCountData = {
+  data: {
+    successCount: PendingContextResourceData<PrometheusQueryRangeResultValue>;
+    failedCount: PendingContextResourceData<PrometheusQueryRangeResultValue>;
+  };
+  refreshAll: () => void;
+};
+
+export const useFetchKserveRequestCountData = (
+  metricsDef: KserveMetricGraphDefinition,
+  timeframe: TimeframeTitle,
+  endInMs: number,
+  namespace: string,
+): RequestCountData => {
+  const active = useIsAreaAvailable(SupportedArea.K_SERVE_METRICS).status;
+
+  const successQuery = metricsDef.queries[0].query;
+  const failedQuery = metricsDef.queries[1].query;
+
+  const successCount = useQueryRangeResourceData(
+    active,
+    successQuery,
+    endInMs,
+    timeframe,
+    defaultResponsePredicate,
+    namespace,
+  );
+
+  const failedCount = useQueryRangeResourceData(
+    active,
+    failedQuery,
+    endInMs,
+    timeframe,
+    defaultResponsePredicate,
+    namespace,
+  );
+
+  const data = React.useMemo(
+    () => ({
+      successCount,
+      failedCount,
+    }),
+    [failedCount, successCount],
+  );
+
+  return useAllSettledContextResourceData(data, {
+    successCount: DEFAULT_PENDING_CONTEXT_RESOURCE,
+    failedCount: DEFAULT_PENDING_CONTEXT_RESOURCE,
+  });
+};
+
+type MeanLatencyData = {
+  data: {
+    inferenceLatency: PendingContextResourceData<PrometheusQueryRangeResultValue>;
+    requestLatency: PendingContextResourceData<PrometheusQueryRangeResultValue>;
+  };
+  refreshAll: () => void;
+};
+
+export const useFetchKserveMeanLatencyData = (
+  metricsDef: KserveMetricGraphDefinition,
+  timeframe: TimeframeTitle,
+  endInMs: number,
+  namespace: string,
+): MeanLatencyData => {
+  const active = useIsAreaAvailable(SupportedArea.K_SERVE_METRICS).status;
+
+  const inferenceLatency = useQueryRangeResourceData(
+    active,
+    metricsDef.queries[0].query,
+    endInMs,
+    timeframe,
+    defaultResponsePredicate,
+    namespace,
+  );
+
+  const requestLatency = useQueryRangeResourceData(
+    active,
+    metricsDef.queries[1].query,
+    endInMs,
+    timeframe,
+    defaultResponsePredicate,
+    namespace,
+  );
+
+  const data = React.useMemo(
+    () => ({
+      inferenceLatency,
+      requestLatency,
+    }),
+    [inferenceLatency, requestLatency],
+  );
+
+  return useAllSettledContextResourceData(data, {
+    inferenceLatency: DEFAULT_PENDING_CONTEXT_RESOURCE,
+    requestLatency: DEFAULT_PENDING_CONTEXT_RESOURCE,
+  });
+};
+
+type CpuUsageData = {
+  data: {
+    cpuUsage: PendingContextResourceData<PrometheusQueryRangeResultValue>;
+  };
+  refreshAll: () => void;
+};
+
+export const useFetchKserveCpuUsageData = (
+  metricsDef: KserveMetricGraphDefinition,
+  timeframe: TimeframeTitle,
+  endInMs: number,
+  namespace: string,
+): CpuUsageData => {
+  const active = useIsAreaAvailable(SupportedArea.K_SERVE_METRICS).status;
+
+  const cpuUsage = useQueryRangeResourceData(
+    active,
+    metricsDef.queries[0].query,
+    endInMs,
+    timeframe,
+    defaultResponsePredicate,
+    namespace,
+  );
+
+  const data = React.useMemo(
+    () => ({
+      cpuUsage,
+    }),
+    [cpuUsage],
+  );
+
+  return useAllSettledContextResourceData(data, {
+    cpuUsage: DEFAULT_PENDING_CONTEXT_RESOURCE,
+  });
+};
+
+type MemoryUsageData = {
+  data: {
+    memoryUsage: PendingContextResourceData<PrometheusQueryRangeResultValue>;
+  };
+  refreshAll: () => void;
+};
+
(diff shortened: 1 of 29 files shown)


frontend/src/concepts/metrics/kserve/utils.ts
```diff
@@ -0,0 +1,27 @@
...
+import {
+  KserveMetricsConfigMapKind,
+  KserveMetricsDataObject,
+} from '~/concepts/metrics/kserve/types';
+
+export const isKserveMetricsConfigMapKind = (
+  configMapKind: ConfigMapKind,
+): configMapKind is KserveMetricsConfigMapKind =>
+  (configMapKind as KserveMetricsConfigMapKind).data.supported === 'true' ||
+  (configMapKind as KserveMetricsConfigMapKind).data.supported === 'false';
```
christianvogt, line 11: Not sure exactly what kind of values will be passed into this, but `data` is possibly `undefined` for a config map and there's no defensive checks here.

frontend/src/api/prometheus/kservePerformanceMetrics.ts
```diff
@@ -0,0 +1,199 @@
...
+  metricsDef: KserveMetricGraphDefinition,
+  timeframe: TimeframeTitle,
+  endInMs: number,
+  namespace: string,
+): RequestCountData => {
+  const active = useIsAreaAvailable(SupportedArea.K_SERVE_METRICS).status;
+
+  //TODO: Necessary due to bug on backend - must be removed before release.
+  const successQuery = metricsDef.queries[0].query; //.replace('[${rate_interval}]', '[5m]');
+  const failedQuery = metricsDef.queries[1].query; //.replace('[${rate_interval}]', '[5m]');
```
christianvogt, line 27: Is this expected to remain as `TODO`?

frontend/src/api/prometheus/kservePerformanceMetrics.ts
```diff
@@ -0,0 +1,199 @@
...
+  // store the result in a reference and only update the reference so long as there are no pending queries
+  //TODO: this creates a bug where the first render may contain a mix of pending / non pending data, causing a flash.
+  //      need to take a default empty value to return first time.
+  const resultRef = React.useRef(result);
+
+  if (!Object.values(result.data).some((x) => x.pending)) {
+    resultRef.current = result;
+  }
+
+  return resultRef.current;
```
christianvogt, line 198: Do you want all data to be `pending` if any one piece of data is pending? If so, return a new data structure with `pending` set to the overall pending state:

```suggestion
  return React.useMemo(() => {
    const pending = Object.values(data).some((x) => x.pending);
    return {
      data: Object.keys(data).reduce((acc, d) => ({ ...acc, [d]: { ...data[d], pending } }), {
        ...data,
      }),
      refreshAll,
    };
  }, [data, refreshAll]);
```

frontend/src/api/prometheus/kservePerformanceMetrics.ts
```diff
@@ -0,0 +1,199 @@
...
+export const useFetchKserveRequestCountData = (
+  metricsDef: KserveMetricGraphDefinition,
+  timeframe: TimeframeTitle,
+  endInMs: number,
+  namespace: string,
+): RequestCountData => {
+  const active = useIsAreaAvailable(SupportedArea.K_SERVE_METRICS).status;
+
+  //TODO: Necessary due to bug on backend - must be removed before release.
+  const successQuery = metricsDef.queries[0].query; //.replace('[${rate_interval}]', '[5m]');
```
mturley, line 26: Is the commented-out code part of the TODO?

frontend/src/api/prometheus/kservePerformanceMetrics.ts
alexcreasy: Missed in final cleanup, removed now.

frontend/src/api/prometheus/kservePerformanceMetrics.ts
alexcreasy: The commented out code was effectively me fulfilling the TODO... just after testing it, rather than going back and removing the commented code and the todo I moved onto the next thing d'oh!

frontend/src/concepts/metrics/kserve/useKserveMetricsConfigMap.ts
```diff
@@ -0,0 +1,36 @@
...
+        .then((c) => {
+          if (!isKserveMetricsConfigMapKind(c)) {
+            return Promise.reject(
+              'Received invalid ConfigMap format for kserve metrics definition',
+            );
+          }
+          return c;
+        })
+        .catch((e) => {
+          // if (e.statusObject?.code === 404) {
```
mturley, line 24: Some commented out code here as well

frontend/src/concepts/metrics/kserve/utils.ts
christianvogt: If you can try to do this without using the keyword `as` that would be good as we're attempting to remove the usage of casting. You can do checks like `'data' in configMapKind`

alexcreasy: Ahh didn't see the followup, I've removed the `as` - is this up to standard now?

frontend/src/api/prometheus/kservePerformanceMetrics.ts
alexcreasy: This doesn't fix the issue at hand - chiefly that you can get data that's "pending" returned in the ref - I've updated this to add a default empty state to initially populate the ref with, this fixes the bug.

My solution doesn't win any points for style I imagine, so happy to take any suggestions there :)

frontend/src/concepts/metrics/kserve/useKserveMetricsConfigMap.ts
alexcreasy: Good spot, thanks!

mturley (approved):
Pulled and tested locally. LGTM.

frontend/src/concepts/metrics/kserve/types.ts
```diff
@@ -0,0 +1,31 @@
+import { ConfigMapKind } from '~/k8sTypes';
+import { KserveMetricsGraphTypes } from '~/concepts/metrics/kserve/const';
+
+export type KserveMetricsConfigMapKind = ConfigMapKind & {
+  data: {
+    supported: 'true' | 'false';
+    metrics: string;
+  };
```
mturley, line 8: (I already approved, but then realized what you were saying above @christianvogt)
Is this type incorrect then, if `data` could be undefined in a configmap? maybe this should be `data?: {` and then optional chaining should be used where it is consumed

frontend/src/api/prometheus/kservePerformanceMetrics.ts
alexcreasy: Yeah, for posterity - I think the better way of implementing this, is to recreate `usePrometheusQueryRange` to enable it to take a batch of queries and use `Promise.all` to make the update atomic. Ran out of time though, I'll try to improve this or log tech debt.

frontend/src/concepts/metrics/kserve/types.ts
alexcreasy: Hmmm it shouldn't be optional for the contract we have with the backend - that's why I'm using a validation check when receiving the configmap - if those properties aren't there the UI will show an error state

alexcreasy: This check ensures that we don't ever run into the issue of those props being undefined.

https://github.com/alexcreasy/odh-dashboard/blob/b0dde63bf4040f9dc75037d1f44c35c612096742/frontend/src/concepts/metrics/kserve/useKserveMetricsConfigMap.ts

christianvogt: @mturley I'm assuming this type is correct for kserve metrics as it expects both of these to be defined if the configmap is valid for this usecase.

In the case of a generic configmap this is invalid because data is optional. But once we go through a type guard check and confirm the configmap to be well formed, then it's correct that these data types aren't optional.

mturley: (my last comment was made before your last two comments loaded, disregard that 😄 )
sounds good.

frontend/src/api/prometheus/kservePerformanceMetrics.ts
christianvogt: I guess my assumption of what you were trying to do was wrong. I'll need more clarity before being able to help further.

PR #2923, opened 2024-06-18
add always() condition to upload cypress artifacts

christianvogt:
## Description

Cypress artifact upload was only done on success but not when test failed.
According to the [github documentation](https://docs.github.com/en/actions/learn-github-actions/expressions#status-check-functions), the default conditional for actions is `success()`. Added the condition of `always()` to the artifact upload step.


christianvogt:
CI passed. artifacts link is still present for successful test run.

DaoDaoNoCode (approved):
/lgtm

PR #2934, opened 2024-06-20
configured default environment for cypress

rsun19:
Closes:  [RHOAIENG-5298](https://issues.redhat.com/browse/RHOAIENG-5298)

## Description

Need consistency within test files on how `ODH_PRODUCT_NAME` is used. Added an eslint rule to enforce that it is used in Cypress testing to avoid test failures.

## How Has This Been Tested?

`npm run test`

Alternatively, you can run the test files that were changed. Run them, then change the `.env` var for `ODH_PRODUCT_NAME`, then run again, and see the name change.

## Test Impact

I implemented the tests.

## Request review criteria:

### DIFF
--- a/frontend/.eslintrc
+++ b/frontend/.eslintrc
@@ -327,6 +327,22 @@
           }
         ]
       }
+    },
+    {
+      "files": ["src/__tests__/cypress/**"],
+      "rules": {
+        "no-restricted-syntax": [
+          "error",
+          {
+            "selector": "Literal[value=/\\bRed Hat OpenShift AI\\b/i],JSXText[value=/\\bRed Hat OpenShift AI\\b/i]",
+            "message": "Do not hard code product name `Red Hat OpenShift AI`. Use `Cypress.env('ODH_PRODUCT_NAME')` instead."
+          },
+          {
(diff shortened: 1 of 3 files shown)


pnaik1:
Verified locally
/lgtm

rsun19:
Rebased to resolve merge conflicts

frontend/.eslintrc
```diff
@@ -327,6 +327,22 @@
           }
         ]
       }
+    },
+    {
+      "files": ["src/__tests__/cypress/**"],
+      "rules": {
```
christianvogt, line 333: The existing rule has this exclusion: `"excludedFiles": ["src/__mocks__/**", "src/__tests__/**"],`

Should we just remove `"src/__tests__/**"` instead of cloning the rule? I suspect we still need to exclude mocks.

rsun19: So the intention for doing this was to change the text in the error message to reflect how we could call on the environment variable in Cypress. I just couldn't find a way to do it when using it in only one usage.

christianvogt: I didn't realize you had changed the message. This is fine then.

Issue #1513, opened 2023-06-28
Container Size becomes unknown when an administrator updates the notebook sizes. 


erwangranger:
## Solution

Instead of saying `Unknown` with an error, display `Custom` instead when we cannot render a matching name.

---

So, today, after consulting with my co-administrator, we decided that the default Notebook Sizes were too generous and needed a haircut. 

The default values in the CR were: 

```yaml
  notebookSizes:
    - name: Small
      resources:
        limits:
          cpu: '2'
          memory: 8Gi
        requests:
          cpu: '1'
          memory: 8Gi
    - name: Medium
      resources:
        limits:
          cpu: '6'
          memory: 24Gi
        requests:
          cpu: '3'
          memory: 24Gi
    - name: Large
      resources:
        limits:
          cpu: '14'
          memory: 56Gi
        requests:
          cpu: '7'
          memory: 56Gi
    - name: X Large
      resources:
        limits:
          cpu: '30'
          memory: 120Gi
        requests:
          cpu: '15'
          memory: 120Gi

```

We changed that to the following: 

```yaml 
  notebookSizes:
    - name: Small
      resources:
        limits:
          cpu: '2'
          memory: 4Gi
        requests:
          cpu: '1'
          memory: 2Gi
    - name: Medium-Small
      resources:
        limits:
          cpu: '4'
          memory: 8Gi
        requests:
          cpu: '2'
          memory: 4Gi
    - name: Medium
      resources:
        limits:
          cpu: '6'
          memory: 16Gi
        requests:
          cpu: '3'
          memory: 8Gi
    - name: Large
      resources:
        limits:
          cpu: '12'
          memory: 24Gi
        requests:
          cpu: '6'
          memory: 16Gi
    - name: X Large
      resources:
        limits:
          cpu: '32'
          memory: 64Gi
        requests:
          cpu: '16'
          memory: 32Gi
    - name: XX Large
      resources:
        limits:
          cpu: '64'
          memory: 512Gi
        requests:
          cpu: '32'
          memory: 256Gi
```

You will note how careful we were to keep the old names, only change their values, and add a couple new names. 
We figured that replacing the name "small" with "tiny" might cause issues, for example. 
Because previously-created workbenches with a size name that would no longer exist would result in ... bad things probably. 

I really wanted to confirm everything worked right away, but this settings takes a couple minutes to get applied, during which I got pinged by someone and spent 30 minutes talking to. 

[...]

PR #2944, opened 2024-06-21
Add custom container size to workbenches table

manaswinidas:
Closes: https://issues.redhat.com/browse/RHOAIENG-578

## Description

When an administrator updates the notebook sizes, the container size now changes to `Custom` instead of `Unknown`.

<details>
  <summary>Workbenches table(not expanded):</summary>
  
Screenshot 2024-07-09 at 11 02 41 PM

</details>

<details>
  <summary>Workbenches table(expanded):</summary>
Screenshot 2024-07-09 at 11 02 54 PM

</details>

</details>

<details>
  <summary>Edit workbenches for deleted notebook sizes(expanded):</summary>
Screenshot 2024-07-16 at 6 10 24 PM

</details>

## How Has This Been Tested?

1. Create a workbench with any container size
2. Go to console and update the `notebookSizes` in `OdhDashboardConfig`([for example](https://github.com/opendatahub-io/odh-dashboard/issues/1513#issue-1803778644))
3. Go to your Workbenches table on your dashboard and see `Custom` in the `Container size`. Expand the table row and see `limits` and `requests`.
4. Edit this workbench > now you should see a selected "Keep custom size" option in the expanded Container size dropdown, along with its `limits` and `requests`.
5. Try editing the workbench name while keeping others intact. The container size should still be "Custom" in the workbenches table after updating the workbench name.

## Test Impact

Added e2e test

## Request review criteria:

If you have UI changes: 

- [X] Included tags to the UX team if it was a UI/UX change (find relevant UX in the [SMEs](https://github.com/opendatahub-io/odh-dashboard/tree/main/docs/smes.md) section).

- [ ] The developer has tested their solution on a cluster by using the image produced by the PR to `main`.

### DIFF
--- a/frontend/src/__tests__/cypress/cypress/pages/workbench.ts
+++ b/frontend/src/__tests__/cypress/cypress/pages/workbench.ts
@@ -299,6 +299,10 @@ class CreateSpawnerPage {
   findBucketInput() {
     return cy.findByTestId('field AWS_S3_BUCKET');
   }
+
+  findContainerSizeInput(name: string) {
+    return cy.findByTestId('container-size-group').contains(name);
+  }
 }
 
 class EditSpawnerPage extends CreateSpawnerPage {
@@ -325,6 +329,10 @@ class EditSpawnerPage extends CreateSpawnerPage {
   shouldHaveContainerSizeInput(name: string) {
     cy.findByTestId('container-size-group').contains(name).should('exist');
     return this;
+  }
+
+  findCancelButton() {
+    return cy.findByTestId('workbench-cancel-button');
   }
 }
 
--- a/frontend/src/__tests__/cypress/cypress/tests/mocked/projects/workbench.cy.ts
+++ b/frontend/src/__tests__/cypress/cypress/tests/mocked/projects/workbench.cy.ts
@@ -1,4 +1,5 @@
 import {
+  mockDashboardConfig,
   mockDscStatus,
   mockK8sResourceList,
   mockNotebookK8sResource,
@@ -32,20 +33,45 @@ import {
   SecretModel,
 } from '~/__tests__/cypress/cypress/utils/models';
 import { mock200Status } from '~/__mocks__/mockK8sStatus';
+import type { NotebookSize } from '~/types';
 
 const configYamlPath = '../../__mocks__/mock-upload-configmap.yaml';
 
 type HandlersProps = {
   isEmpty?: boolean;
+  notebookSizes?: NotebookSize[];
 };
 
-const initIntercepts = ({ isEmpty = false }: HandlersProps) => {
+const initIntercepts = ({
+  isEmpty = false,
+  notebookSizes = [
+    {
+      name: 'Medium',
+      resources: {
+        limits: {
+          cpu: '6',
+          memory: '24Gi',
+        },
+        requests: {
+          cpu: '3',
+          memory: '24Gi',
+        },
+      },
+    },
+  ],
+}: HandlersProps) => {
   cy.interceptOdh(
     'GET /api/dsc/status',
     mockDscStatus({
       installedComponents: {
         workbenches: true,
       },
+    }),
+  );
+  cy.interceptOdh(
+    'GET /api/config',
+    mockDashboardConfig({
+      notebookSizes,
     }),
   );
   cy.interceptK8sList(ProjectModel, mockK8sResourceList([mockProjectK8sResource({})]));
@@ -149,7 +175,37 @@ describe('Workbench page', () => {
   });
 
   it('Create workbench', () => {
-    initIntercepts({ isEmpty: true });
+    initIntercepts({
+      isEmpty: true,
+      notebookSizes: [
+        {
+          name: 'XSmall',
+          resources: {
+            limits: {
+              cpu: '0.5',
+              memory: '500Mi',
+            },
+            requests: {
+              cpu: '0.1',
+              memory: '100Mi',
+            },
+          },
+        },
+        {
+          name: 'Small',
+          resources: {
+            limits: {
+              cpu: '2',
+              memory: '8Gi',
+            },
+            requests: {
+              cpu: '1',
+              memory: '8Gi',
+            },
+          },
+        },
+      ],
+    });
     workbenchPage.visit('test-project');
     workbenchPage.findCreateButton().click();
     createSpawnerPage.findSubmitButton().should('be.disabled');
@@ -262,7 +318,37 @@ describe('Workbench page', () => {
   });
 
   it('Create workbench with numbers', () => {
-    initIntercepts({ isEmpty: true });
+    initIntercepts({
+      isEmpty: true,
+      notebookSizes: [
+        {
+          name: 'XSmall',
+          resources: {
+            limits: {
+              cpu: '0.5',
+              memory: '500Mi',
+            },
+            requests: {
+              cpu: '0.1',
+              memory: '100Mi',
+            },
+          },
+        },
+        {
+          name: 'Small',
+          resources: {
+            limits: {
+              cpu: '2',
+              memory: '8Gi',
+            },
+            requests: {
+              cpu: '1',
+              memory: '8Gi',
+            },
+          },
+        },
+      ],
+    });
     workbenchPage.visit('test-project');
     workbenchPage.findCreateButton().click();
     createSpawnerPage.findSubmitButton().should('be.disabled');
@@ -300,7 +386,36 @@ describe('Workbench page', () => {
   });
 
   it('list workbench and table sorting', () => {
-    initIntercepts({});
+    initIntercepts({
+      notebookSizes: [
+        {
+          name: 'XSmall',
+          resources: {
+            limits: {
+              cpu: '0.5',
+              memory: '500Mi',
+            },
+            requests: {
+              cpu: '0.1',
+              memory: '100Mi',
+            },
+          },
+        },
+        {
+          name: 'Small',
+          resources: {
+            limits: {
+              cpu: '2',
+              memory: '8Gi',
+            },
+            requests: {
+              cpu: '1',
+              memory: '8Gi',
+            },
+          },
+        },
+      ],
+    });
     workbenchPage.visit('test-project');
     const notebookRow = workbenchPage.getNotebookRow('Test Notebook');
     notebookRow.shouldHaveNotebookImageName('Test Image');
@@ -397,7 +512,36 @@ describe('Workbench page', () => {
   });
 
   it('Edit workbench', () => {
-    initIntercepts({});
+    initIntercepts({
+      notebookSizes: [
+        {
+          name: 'XSmall',
+          resources: {
+            limits: {
+              cpu: '0.5',
+              memory: '500Mi',
+            },
+            requests: {
+              cpu: '0.1',
+              memory: '100Mi',
+            },
+          },
+        },
+        {
+          name: 'Small',
+          resources: {
+            limits: {
+              cpu: '2',
+              memory: '8Gi',
+            },
+            requests: {
+              cpu: '1',
+              memory: '8Gi',
+            },
+          },
+        },
+      ],
+    });
     editSpawnerPage.visit('test-notebook');
     editSpawnerPage.findNameInput().should('have.value', 'Test Notebook');
     editSpawnerPage.shouldHaveNotebookImageSelectInput('Test Image');
@@ -437,8 +581,39 @@ describe('Workbench page', () => {
     });
 
     cy.get('@editWorkbench.all').then((interceptions) => {
-      expect(interceptions).to.have.length(2); // 1 dry run request and 1 actaul request
+      expect(interceptions).to.have.length(2); // 1 dry run request and 1 actual request
     });
+  });
+
+  it('Handle deleted notebook sizes in workbenches table', () => {
+    initIntercepts({
+      notebookSizes: [
+        {
+          name: 'XSmall',
+          resources: {
+            limits: {
+              cpu: '0.5',
+              memory: '500Mi',
+            },
+            requests: {
+              cpu: '0.1',
+              memory: '100Mi',
+            },
+          },
+        },
+      ],
+    });
+    workbenchPage.visit('test-project');
+    const notebookRow = workbenchPage.getNotebookRow('Test Notebook');
+    notebookRow.shouldHaveNotebookImageName('Test Image');
+    notebookRow.shouldHaveContainerSize('Custom');
+    notebookRow.findKebabAction('Edit workbench').click();
+    editSpawnerPage.shouldHaveContainerSizeInput('Keep custom size');
+    cy.go('back');
+    workbenchPage.findCreateButton().click();
+    verifyRelativeURL('/projects/test-project/spawner');
+    // Custom container size dropdown option should not be present for create workbench
+    createSpawnerPage.findContainerSizeInput('Keep custom size').should('not.exist');
   });
 
   it('Validate that updating invalid workbench will navigate to the new page with an error message', () => {
--- a/frontend/src/pages/projects/screens/detail/notebooks/NotebookTableRow.tsx
+++ b/frontend/src/pages/projects/screens/detail/notebooks/NotebookTableRow.tsx
@@ -1,17 +1,8 @@
 import * as React from 'react';
 import { ActionsColumn, ExpandableRowContent, Tbody, Td, Tr } from '@patternfly/react-table';
-import {
-  Button,
-  Flex,
-  FlexItem,
-  Icon,
-  Popover,
-  Split,
-  SplitItem,
-  Tooltip,
-} from '@patternfly/react-core';
+import { Button, Flex, FlexItem, Icon, Popover, Split, SplitItem } from '@patternfly/react-core';
 import { useNavigate } from 'react-router-dom';
-import { ExclamationCircleIcon, InfoCircleIcon } from '@patternfly/react-icons';
+import { InfoCircleIcon } from '@patternfly/react-icons';
 import { NotebookState } from '~/pages/projects/notebook/types';
 import NotebookRouteLink from '~/pages/projects/notebook/NotebookRouteLink';
 import NotebookStatusToggle from '~/pages/projects/notebook/NotebookStatusToggle';
@@ -22,6 +13,7 @@ import { TableRowTitleDescription } from '~/components/table';
 import { ProjectObjectType, typedObjectImage } from '~/concepts/design/utils';
 import DashboardPopupIconButton from '~/concepts/dashboard/DashboardPopupIconButton';
 import { getDescriptionFromK8sResource, getDisplayNameFromK8sResource } from '~/concepts/k8s/utils';
+import { NotebookSize } from '~/types';
 import useNotebookDeploymentSize from './useNotebookDeploymentSize';
 import useNotebookImage from './useNotebookImage';
 import NotebookSizeDetails from './NotebookSizeDetails';
@@ -51,7 +43,14 @@ const NotebookTableRow: React.FC<NotebookTableRowProps> = ({
   const { currentProject } = React.useContext(ProjectDetailsContext);
   const navigate = useNavigate();
   const [isExpanded, setExpanded] = React.useState(false);
-  const { size: notebookSize, error: sizeError } = useNotebookDeploymentSize(obj.notebook);
+  const { size: notebookSize } = useNotebookDeploymentSize(obj.notebook);
+  const lastDeployedSize: NotebookSize = {
+    name: 'Custom',
+    resources: obj.notebook.spec.template.spec.containers[0].resources ?? {
+      limits: {},
+      requests: {},
+    },
+  };
   const [notebookImage, loaded, loadError] = useNotebookImage(obj.notebook);
 
   return (
@@ -124,7 +123,7 @@ const NotebookTableRow: React.FC<NotebookTableRowProps> = ({
                   }
                 >
                   <DashboardPopupIconButton
-                    aria-label="Notebook image has out of date Elrya version"
+                    aria-label="Notebook image has out of date Elyra version"
                     data-testid="outdated-elyra-info"
                     icon={
                       <Icon status="info">
@@ -143,14 +142,7 @@ const NotebookTableRow: React.FC<NotebookTableRowProps> = ({
               spaceItems={{ default: 'spaceItemsXs' }}
               alignItems={{ default: 'alignItemsCenter' }}
             >
-              <FlexItem>{notebookSize?.name ?? 'Unknown'}</FlexItem>
-              {sizeError && (
-                <Tooltip content={sizeError}>
-                  <Icon aria-label="error icon" role="button" status="danger" tabIndex={0}>
-                    <ExclamationCircleIcon />
-                  </Icon>
-                </Tooltip>
-              )}
+              <FlexItem>{notebookSize?.name ?? <i>{lastDeployedSize.name}</i>}</FlexItem>
             </Flex>
           </Td>
         ) : null}
@@ -212,7 +204,7 @@ const NotebookTableRow: React.FC<NotebookTableRowProps> = ({
           </Td>
           <Td dataLabel="Limits">
             <ExpandableRowContent>
-              {notebookSize && <NotebookSizeDetails notebookSize={notebookSize} />}
+              <NotebookSizeDetails notebookSize={notebookSize || lastDeployedSize} />
             </ExpandableRowContent>
           </Td>
           <Td />
--- a/frontend/src/pages/projects/screens/detail/notebooks/useNotebookDeploymentSize.ts
+++ b/frontend/src/pages/projects/screens/detail/notebooks/useNotebookDeploymentSize.ts
@@ -5,9 +5,7 @@ import { getNotebookSizes } from '~/pages/notebookController/screens/server/useP
 import { NotebookKind } from '~/k8sTypes';
 import { isCpuResourceEqual, isMemoryResourceEqual } from '~/utilities/valueUnits';
 
-const useNotebookDeploymentSize = (
-  notebook?: NotebookKind,
-): { size: NotebookSize | null; error: string } => {
+const useNotebookDeploymentSize = (notebook?: NotebookKind): { size: NotebookSize | null } => {
   const { dashboardConfig } = React.useContext(AppContext);
 
   const container: PodContainer | undefined = notebook?.spec.template.spec.containers.find(
@@ -15,7 +13,7 @@ const useNotebookDeploymentSize = (
   );
 
   if (!container) {
-    return { size: null, error: 'Failed to get workbench information.' };
+    return { size: null };
   }
 
   const sizes = getNotebookSizes(dashboardConfig);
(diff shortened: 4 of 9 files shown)


manaswinidas:
@kywalker-rh Is there a need to state the resource limits and requests below `Custom` 

For example, 
Screenshot 2024-06-22 at 1 23 03 AM

kywalker-rh:
@manaswinidas It would be helpful for the user to see what the limits and requests are, but as it states in the other comment, this can be shown when the row is expanded.

frontend/src/pages/projects/screens/detail/notebooks/NotebookTableRow.tsx
```diff
@@ -51,7 +42,7 @@ const NotebookTableRow: React.FC<NotebookTableRowProps> = ({
   const { currentProject } = React.useContext(ProjectDetailsContext);
   const navigate = useNavigate();
   const [isExpanded, setExpanded] = React.useState(false);
-  const { size: notebookSize, error: sizeError } = useNotebookDeploymentSize(obj.notebook);
+  const { selectedSize: notebookSize } = usePreferredNotebookSize();
```
mturley, line 45: I don't know if `usePreferredNotebookSize` is giving you the right value here -- it doesn't know anything about the current row's notebook, it just gives the `currentUserState.lastSelectedSize`. I created a workbench with container size "Small", then I created one with size "Large", and they both display as "Small" with this change but they are correct on main.

mturley (changes requested):

frontend/src/pages/projects/screens/detail/notebooks/NotebookTableRow.tsx
manaswinidas: Yes, I saw the same too an hour or two ago .. there are a few things I am testing .. will update this shortly

manaswinidas:
> @manaswinidas It would be helpful for the user to see what the limits and requests are, but as it states in the other comment, this can be shown when the row is expanded.

@kywalker-rh Yes, this is there but the limits and requests are not in the expanded row now that this container size is deleted. You can check this in the screen recording attached in the PR description. 

I have added screenshots to show the container size dropdown for this "Custom" state. Can you check them when you have time?

As shown in the screenshot, I'm adding "Custom" as the first option in the dropdown and it doesn't show limits and requests in the option description coz it's all deleted. If I show them, it will look something like this: 

Screenshot 2024-07-02 at 11 26 01 PM

Let me know if the screenshots and screen recording look fine.

andrewballantyne:
@manaswinidas can we not show whatever is in the notebook? It's only on edit we show custom, right?

manaswinidas:
@andrewballantyne Oh yes, missed it. Updated to show "Custom" only on Edit.

@kywalker-rh is there a need to change the content in this alert?
Screenshot 2024-07-03 at 11 01 43 AM

dpanshug:
When creating a new workbench the alert pops up. 

https://github.com/opendatahub-io/odh-dashboard/assets/97534722/74111aa4-fed6-4037-b25d-eb402316cfe9

manaswinidas:
@dpanshug fixed!

kywalker-rh:
> @kywalker-rh Yes, this is there but the limits and requests are not in the expanded row now that this container size is deleted. You can check this in the screen recording attached in the PR description.

@manaswinidas I'm sorry I misunderstood what you were asking before. With more context I'm okay moving forward as is.

manaswinidas:
@kywalker-rh We now have the limits and requests in the expanded row. The screenshots and screen recording have been updated.

dpanshug (changes requested):
Just a small nit, otherwise lgtm

dpanshug (approved):
/lgtm

frontend/src/__tests__/cypress/cypress/tests/mocked/projects/workbench.cy.ts
```diff
@@ -399,10 +400,72 @@ describe('Workbench page', () => {
...
+      expect(interceptions).to.have.length(2); // 1 dry run request and 1 actual request
     });
   });
 
+  it('Handle deleted notebook sizes in workbenches table', () => {
+    initIntercepts({});
+    cy.interceptOdh(
+      'GET /api/config',
+      mockDashboardConfig({
+        notebookSizes: [
```
andrewballantyne, line 412: Could this be added to initIntercepts and just configured with these values? 

I imagine the test is not finding any mock today for this and soft erroring out... the mock dashboard config seems to handle no data -- so it shouldn't impact things 🤔 but it will allow for wider usage by other tests down the road.

Also, you don't need to make the mock so much -- you're just looking to set it to something else for the sake of this test... verbose test data doesn't necessarily help test it more.

frontend/src/__tests__/cypress/cypress/tests/mocked/projects/workbench.cy.ts
```diff
@@ -399,10 +400,72 @@ describe('Workbench page', () => {
...
+    workbenchPage.visit('test-project');
+    const notebookRow = workbenchPage.getNotebookRow('Test Notebook');
+    notebookRow.shouldHaveNotebookImageName('Test Image');
+    notebookRow.shouldHaveContainerSize('Custom');
+    notebookRow.findKebabAction('Edit workbench').click();
+    editSpawnerPage.shouldHaveContainerSizeInput('Custom');
+    workbenchPage.visit('test-project');
+    workbenchPage.findCreateButton().click();
+    createSpawnerPage.findSubmitButton().should('be.disabled');
+    verifyRelativeURL('/projects/test-project/spawner');
```
andrewballantyne, line 464: What is checking the submit button to be disabled testing here? The form doesn't start in a submitting format before your change 🤔

frontend/src/__tests__/cypress/cypress/tests/mocked/projects/workbench.cy.ts
```diff
@@ -399,10 +400,72 @@ describe('Workbench page', () => {
...
+    // Custom container size dropdown option should not be present for create workbench
+    createSpawnerPage.shouldNotHaveCustomContainerSizeInput();
```
andrewballantyne, line 466: Ah -- your usage here seems weird. Why not support the inverse of the usage a few lines above
```suggestion
    createSpawnerPage.shouldNotHaveContainerSizeInput('Custom');
```

andrewballantyne, line 466: Are these not two different tests? -- this one (create) and the other one above this (edit)?

frontend/src/pages/projects/screens/detail/notebooks/NotebookTableRow.tsx
```diff
@@ -51,7 +43,14 @@ const NotebookTableRow: React.FC<NotebookTableRowProps> = ({
   const { currentProject } = React.useContext(ProjectDetailsContext);
   const navigate = useNavigate();
   const [isExpanded, setExpanded] = React.useState(false);
-  const { size: notebookSize, error: sizeError } = useNotebookDeploymentSize(obj.notebook);
```
andrewballantyne, line 54: So you didn't address the `error` source here... you replaced the only code that used it -- and technically there are still avenues to address an error here. Please look at the `useNotebookDeploymentSize` and determine if you need to make changes to satisfy this.

frontend/src/pages/projects/screens/spawner/SpawnerPage.tsx
```diff
@@ -97,10 +97,16 @@ const SpawnerPage: React.FC<SpawnerPageProps> = ({ existingNotebook }) => {
...
   const { size: notebookSize } = useNotebookDeploymentSize(existingNotebook);
   React.useEffect(() => {
+    const customSize: NotebookSize = { name: 'Custom', resources: {} };
     if (notebookSize) {
       setSelectedSize(notebookSize.name);
+    } else {
+      if (existingNotebook && !sizes.some((size) => size.name === 'Custom')) {
+        sizes.unshift(customSize);
+      }
+      setSelectedSize(customSize.name);
```
andrewballantyne, line 107: This logic prevents the ability for the customer to use `Custom` as a value name. Why did you do that?

Can we not make our "Custom" significantly more ambitious with an internal id/name that we have a constant for -- and then use that to render special options?

frontend/src/pages/projects/screens/spawner/deploymentSize/ContainerSizeSelector.tsx
```diff
@@ -35,7 +35,11 @@ const ContainerSizeSelector: React.FC<ContainerSizeSelectorProps> = ({
         {sizes.map((size) => {
           const { name } = size;
           const desc = getSizeDescription(size);
-          return <SelectOption key={name} value={name} description={desc} />;
+          return name === 'Custom' ? (
```
andrewballantyne, line 38: You should avoid disjointed magic strings; always use a constant when you have to share a value across files -- it makes it difficult to maintain, especially with common words like `Custom`.

frontend/src/pages/projects/screens/spawner/SpawnerPage.tsx
andrewballantyne: This is very weird logic. Please have a test for this.

Example rendering if I set a 'Custom' NotebookSize in the OdhDashboardConfig:
image

frontend/src/pages/projects/screens/spawner/useNotebookSize.ts
```diff
@@ -28,15 +30,15 @@ export const useNotebookSize = (): {
...
         notification.warning(
           'The size you select is no longer available, we have set the size to the default one.',
         );
       }
-      setSelectedSize(foundSize);
+      setSelectedSize(foundSize ?? sizes[0]);
     },
-    [notification, sizes],
+    [notification, sizes, isExistingNotebook],
   );
```
andrewballantyne, line 42: What's this logic doing? If it's a new notebook you're skipping the block anyways as there is no size selected 🤔 

Wouldn't we want to just do away with this callback?

andrewballantyne (changes requested):
You seem to be reacting to comments and not addressing our codebase.

Maintenance is 80% of the life of any code you ever write. Please do not leave our code in a worse state. If you're modifying something, ask yourself "is this the best way to continue doing this" -- just slapping some logic to make your PR comment "go away" is not an acceptable way to solve feedback. 

Please read my comments and let me know if you need help figuring out some of the answers to the questions I asked.

frontend/src/__tests__/cypress/cypress/pages/workbench.ts
```diff
@@ -299,6 +299,11 @@ class CreateSpawnerPage {
   findBucketInput() {
     return cy.findByTestId('field AWS_S3_BUCKET');
   }
+
+  shouldNotHaveCustomContainerSizeInput() {
+    cy.findByTestId('container-size-group').contains('Custom').should('not.exist');
+    return this;
+  }
```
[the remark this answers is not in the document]
andrewballantyne: FWIW, I got a response -- It is a pattern that seems to be documented: https://github.com/opendatahub-io/odh-dashboard/blob/main/docs/testing.md#page-objects but it does not explicitly state to do this like this -- so up to interpretation. However I would encourage you to take this feedback and combine it with [this feedback](https://github.com/opendatahub-io/odh-dashboard/pull/2944#discussion_r1666010528) and address them together.

frontend/src/__tests__/cypress/cypress/tests/mocked/projects/workbench.cy.ts
andrewballantyne: Conversation [is going on here](https://redhat-internal.slack.com/archives/C069KSM8T9N/p1720121573050779) -- this may not be necessarily. Waiting on the end of the thread.

andrewballantyne: Seems I am contradicting Christian -- I'll resolve this comment as not needed.

frontend/src/__tests__/cypress/cypress/tests/mocked/projects/workbench.cy.ts
```diff
@@ -399,10 +400,72 @@ describe('Workbench page', () => {
...
+        ],
+      }),
+    );
+    workbenchPage.visit('test-project');
+    const notebookRow = workbenchPage.getNotebookRow('Test Notebook');
+    notebookRow.shouldHaveNotebookImageName('Test Image');
+    notebookRow.shouldHaveContainerSize('Custom');
+    notebookRow.findKebabAction('Edit workbench').click();
+    editSpawnerPage.shouldHaveContainerSizeInput('Custom');
+    workbenchPage.visit('test-project');
```
andrewballantyne, line 461: Does this cause a page refresh in the Cypress UI?

frontend/src/pages/projects/screens/spawner/useNotebookSize.ts
manaswinidas: This logic is to prevent notifications from popping up in case of creating workbench. https://github.com/opendatahub-io/odh-dashboard/pull/2944#issuecomment-2205447763

frontend/src/__tests__/cypress/cypress/tests/mocked/projects/workbench.cy.ts
manaswinidas: yes

frontend/src/pages/projects/screens/spawner/useNotebookSize.ts
andrewballantyne: I understand what you've tried to do -- what I'm asking is why are you doing it and why are you duct taping the logic onto the side of the function and not just readdressing the function's purpose? You've effectively changed the logic to avoid a side-effect because your trying to avoid the side effect -- I'm asking you to spend time on understanding the code before just slapping a change to the logic. What's the use-cases here for this function now that use made this change?

frontend/src/__tests__/cypress/cypress/tests/mocked/projects/workbench.cy.ts
andrewballantyne: @christianvogt How do you redirect without a refresh?

christianvogt: Looks to me like you just want to go back to the previous page. In which case you can manipulate the history using `cy.go('back')`. There's no way to go directly to a URL without a page refresh AFAIK. Therefore another alternative is to utilize the app navigation and manually navigate.

frontend/src/__tests__/cypress/cypress/tests/mocked/projects/workbench.cy.ts
manaswinidas: I'm doing this: ` createSpawnerPage.findContainerSizeInput('Keep custom size').should('not.exist');`

manaswinidas:
@kywalker-rh We are getting rid of the error icon in the notebook icon for this issue coz it's not an error. So how do we distinguish between deleted container size name and `Custom` predefined notebook size? @andrewballantyne suggested that we keep deleted ones in italics as of now - let me know if there are other ways we can show this. 
Screenshot 2024-07-09 at 10 47 17 PM

andrewballantyne (changes requested):

frontend/src/pages/projects/screens/spawner/deploymentSize/ContainerSizeSelector.tsx
```diff
@@ -3,6 +3,7 @@ import { FormGroup, FormHelperText, HelperText, HelperTextItem } from '@patternf
 import { Select, SelectOption } from '@patternfly/react-core/deprecated';
 import { NotebookSize } from '~/types';
 import { getSizeDescription } from '~/pages/projects/screens/spawner/spawnerUtils';
+import { customSize } from '~/pages/projects/screens/spawner/SpawnerPage';
```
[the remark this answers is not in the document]
manaswinidas: Oh yes, I should have done this earlier.

frontend/src/__tests__/cypress/cypress/pages/workbench.ts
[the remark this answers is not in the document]
manaswinidas: Resolved here: https://github.com/opendatahub-io/odh-dashboard/pull/2944#discussion_r1670800662

frontend/src/pages/projects/screens/spawner/SpawnerPage.tsx
```diff
@@ -99,8 +101,13 @@ const SpawnerPage: React.FC<SpawnerPageProps> = ({ existingNotebook }) => {
   React.useEffect(() => {
     if (notebookSize) {
       setSelectedSize(notebookSize.name);
+    } else {
+      if (existingNotebook && !sizes.some((size) => size === customSize)) {
+        sizes.unshift(customSize);
+      }
```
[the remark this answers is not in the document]
manaswinidas: For this in the Edit screen:
Screenshot 2024-07-09 at 11 03 21 PM

We can't select a size that's not in sizes.

frontend/src/__tests__/cypress/cypress/tests/mocked/projects/workbench.cy.ts
manaswinidas: This test is failing without a page refresh for me: `createSpawnerPage.findContainerSizeInput('Keep custom size').should('not.exist');`

frontend/src/pages/projects/screens/spawner/SpawnerPage.tsx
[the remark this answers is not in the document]
andrewballantyne: You shouldn't mutate a variable you didn't create like this. `unshift` mutates the array, which could have negative impacts on React's ability to diff between renders -- in this case, you're just mutating another runtime variable, but it could have side-effects on other functions run before or after your code. In React, you should be having a single source of truth for what are the available sizes and getting that updated with the `Keep custom size` and not trying to adhoc change the value later.

So maybe we need to rework this effort so it's more aligned... maybe something like this?

Can you take all of:
```
  const { size: notebookSize } = useNotebookDeploymentSize(existingNotebook);
  React.useEffect(() => {
    if (notebookSize) {
      setSelectedSize(notebookSize.name);
    } else {
      if (existingNotebook && !sizes.some((size) => size === CUSTOM_NOTEBOOK_SIZE)) {
        sizes.unshift(CUSTOM_NOTEBOOK_SIZE);
      }
      setSelectedSize(CUSTOM_NOTEBOOK_SIZE.name);
    }
  }, [notebookSize, setSelectedSize, sizes, existingNotebook]);
```

And create a new custom hook that uses this:
```
const { selectedSize, setSelectedSize, sizes } = useNotebookSize();
```

And produces a set of sizes & selectedSize to be used by the form, which manages both the Edit with custom, the create, and the edit with a known size.

frontend/src/__tests__/cypress/cypress/tests/mocked/projects/workbench.cy.ts
andrewballantyne: Wait, why is it failing? Did you find a bug that you're trying to just ignore or is this a cypress-ism and some state is weirdly set because of it? Did you debug the issue @manaswinidas ? Have you ran the test manually to see if the issue is there too?

frontend/src/pages/projects/screens/spawner/SpawnerPage.tsx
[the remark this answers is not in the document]
manaswinidas: Added a custom hook and this custom hook works better.

frontend/src/pages/projects/screens/spawner/useNotebookSizeState.ts
```diff
@@ -0,0 +1,32 @@
...
+  existingNotebook: NotebookKind | undefined,
+): {
+  selectedSize: NotebookSize;
+  setSelectedSize: (size: string) => void;
+  sizes: NotebookSize[];
+} => {
+  const { size: notebookSize } = useNotebookDeploymentSize(existingNotebook);
+  const { selectedSize, setSelectedSize, sizes } = useNotebookSize();
+
+  const CUSTOM_NOTEBOOK_SIZE = getCustomNotebookSize(existingNotebook);
```
christianvogt, line 18: This is going to trigger the `useEffect` on every render.

frontend/src/pages/projects/screens/spawner/useNotebookSizeState.ts
```diff
@@ -0,0 +1,32 @@
...
+  React.useEffect(() => {
+    if (notebookSize) {
+      setSelectedSize(notebookSize.name);
+    } else {
+      if (existingNotebook && !sizes.some((size) => size.name === CUSTOM_NOTEBOOK_SIZE.name)) {
+        sizes.unshift(CUSTOM_NOTEBOOK_SIZE);
+      }
+      setSelectedSize(CUSTOM_NOTEBOOK_SIZE.name);
+    }
+  }, [notebookSize, setSelectedSize, sizes, existingNotebook, CUSTOM_NOTEBOOK_SIZE]);
```
christianvogt, line 29: Every time i see a `useEffect` without any async behavior it's likely possibly to do the same using `useMemo` and properly assigning the default state when `useState` is first defined. In this particular case the default selected size comes from `useNotebookSize` and assigns `sizes[0]` which isn't correct. It will not always be the first in the list and a good solution would be to accept the default state as a parameter to the hook.

manaswinidas: This fixes the [testing issue](https://github.com/opendatahub-io/odh-dashboard/pull/2944#discussion_r1671997549) too.

frontend/src/pages/projects/screens/spawner/useNotebookSizeState.ts
manaswinidas: resolving this as I'm not using `useEffect` anymore.

frontend/src/__tests__/cypress/cypress/tests/mocked/projects/workbench.cy.ts
manaswinidas: Fixed after changes following [this comment](https://github.com/opendatahub-io/odh-dashboard/pull/2944#discussion_r1679436931)

frontend/src/pages/projects/screens/spawner/useNotebookSizeState.ts
christianvogt: It is still used as a dependency of `useMemo` in this hook. As an unstable reference, it forces all dependent hooks to be re-evaluated.

You'll want to utilize `useDeepCompareMemoize` on this.

frontend/src/pages/projects/screens/spawner/useNotebookSizeState.ts
```diff
@@ -0,0 +1,43 @@
...
+  const { sizes } = useNotebookSize();
+
+  const CUSTOM_NOTEBOOK_SIZE = getCustomNotebookSize(existingNotebook);
+
+  const allSizes = React.useMemo(() => {
+    if (existingNotebook && !sizes.some((size) => size.name === CUSTOM_NOTEBOOK_SIZE.name)) {
+      return [CUSTOM_NOTEBOOK_SIZE, ...sizes];
+    }
+    return sizes;
+  }, [sizes, existingNotebook, CUSTOM_NOTEBOOK_SIZE]);
```
christianvogt, line 25: Won't this add `CUSTOM_NOTEBOOK_SIZE` into the array always even for notebooks without a custom size?

frontend/src/pages/projects/screens/spawner/useNotebookSizeState.ts
manaswinidas: I'm computing when it's needed.

frontend/src/pages/projects/screens/spawner/useNotebookSizeState.ts
manaswinidas: yes, not sure why I removed the condition when I changed `useEffect` to `useMemo` 🤕

emilys314:
I verified the changes in this branch on my cluster, everything checks out functionality wise:
<details>
  <summary>Screenshots</summary>
  Screenshot from 2024-07-23 11-13-47
  Screenshot from 2024-07-23 11-14-41
</details>

frontend/src/__tests__/cypress/cypress/tests/mocked/projects/workbench.cy.ts
```diff
@@ -149,7 +175,37 @@ describe('Workbench page', () => {
   });
 
   it('Create workbench', () => {
-    initIntercepts({ isEmpty: true });
+    initIntercepts({
+      isEmpty: true,
+      notebookSizes: [
```
ppadti, line 180: Is this necessary to pass `notebookSizes` in every tests? Can we just pass this only in required test?

frontend/src/pages/projects/screens/detail/notebooks/useNotebookDeploymentSize.ts
```diff
@@ -31,12 +29,10 @@ const useNotebookDeploymentSize = (
   if (!size) {
     return {
       size: null,
-      error:
-        'Workbench size is currently unavailable. Check your dashboard configuration to view size information.',
     };
   }
 
-  return { size, error: '' };
+  return { size };
```
christianvogt, line 35: If `requests` differ but `limits` match a predefined size, that size is shown and the UI displays the incorrect `requests` in the UI.

eg:
```
          resources:
            limits:
              cpu: '2'
              memory: 8Gi
            requests:
              cpu: '2'
              memory: 9Gi
```
image
image


Expected to see `custom` and the correct values in the UI.


As a result of this, when editing this notebook, the size is selected as `small` and even if I do not change the size, I lose my custom values and it changes to the small container size values.

frontend/src/pages/projects/screens/spawner/useNotebookSizeState.ts
```diff
@@ -0,0 +1,45 @@
...
+} => {
+  const { size: notebookSize } = useNotebookDeploymentSize(existingNotebook);
+  const { sizes } = useNotebookSize();
+
+  const allSizes = React.useMemo(() => {
+    if (notebookSize) {
+      return sizes;
+    }
+    const CUSTOM_NOTEBOOK_SIZE = getCustomNotebookSize(existingNotebook);
+    if (existingNotebook && !sizes.some((size) => size.name === CUSTOM_NOTEBOOK_SIZE.name)) {
```
christianvogt, line 23: `sizes` will never contain the `CUSTOM_NOTEBOOK_SIZE` entry. This logic is left over from a previous iteration.

frontend/src/pages/projects/screens/spawner/useNotebookSizeState.ts
```diff
@@ -0,0 +1,45 @@
...
+    }
+    return sizes;
+  }, [notebookSize, sizes, existingNotebook]);
+
+  const defaultSize = React.useMemo(() => {
+    const foundSize = allSizes.find((size) => size.name === notebookSize?.name);
+    return foundSize || allSizes[0];
+  }, [allSizes, notebookSize]);
+
+  const [selectedSize, setSelectedSize] = React.useState<NotebookSize>(defaultSize);
```
christianvogt, line 34: Use the `useState` function initializer param since the value of `defaultSize` is only used once to initialize the state.

```suggestion
  const [selectedSize, setSelectedSize] = React.useState<NotebookSize>(
    () => allSizes.find((size) => size.name === notebookSize?.name) || allSizes[0],
  );
```

andrewballantyne:
/uncc
/unassign

Christian is on the review side of things now. I'll step out. Tag me if you need me.

frontend/src/pages/projects/screens/spawner/useNotebookSizeState.ts
mturley: TIL you can do that 👀

manaswinidas: done, thanks for pointing this out.

manaswinidas:
@christianvogt made the requested changes and the build passed.

gitdallas (approved):

christianvogt:
Retested the previous failing scenarios I noted in comments and they are now working correctly.

/approve

PR #2958, opened 2024-06-26
[RHOAIENG-7495] Lift ModelRegistrySelectorContextProvider to top level, fall back gracefully if MR feature is unavailable

mturley:
Resolves [RHOAIENG-7495](https://issues.redhat.com/browse/RHOAIENG-7495)

## Description

From the Jira issue:

> Right now, the ModelRegistryContextCreator provider is wrapping the [Model Registry route](https://github.com/opendatahub-io/odh-dashboard/blob/main/frontend/src/pages/modelRegistry/ModelRegistryRoutes.tsx#L16), which is fine but is leading to two issues:
> * We don't get this same data in the home page, which might be needed in the near future.
> * We loose the ability to store a registry selection if we move away from the model registry mounted route.
> 
> For that we should implement the following:
> * Add the Model Registry Selector Context in top of the dashboard tree context provider ([right below the project selector](https://github.com/opendatahub-io/odh-dashboard/blob/main/frontend/src/app/App.tsx#L118)).
> * We need to take into account that Model Registry might not exist if the feature is not enabled, so we'll need to add some logic to check if dsc is not mounted and skip the fetch.

This PR:
* Lifts the `ModelRegistrySelectorContextProvider` from where it was wrapping the MR page route to the top-level `App.tsx`
* Renames the existing provider to `EnabledModelRegistrySelectorContextProvider` and replaces it with a wrapper that conditionally renders the inner provider only when the MR feature is available.
  * This prevents us from having to worry about checking the feature availability in all the hooks used in the provider - we just don't mount them at all unless/until the feature is available. Any place that consumes the context will fall back to its default empty values.
  * This means we don't need `conditionalArea` there anymore, so I unwrapped that and now the inner `Enabled*` provider is a regular `React.FC`. However, this triggered an additional eslint rule that wasn't taking effect before (`react/jsx-no-constructed-context-values`), so to make eslint happy I had to also wrap the context value in a `useMemo`.
  * However, now the `ModelRegistryCoreLoader` is not protected by the `conditionalArea` we removed, so I added `conditionalArea` to that component so it still won't render if the feature is unavailable.
* Moves the `useModelRegistryEnabled` hook from `~/pages/modelRegistry` to `~/concepts/modelRegistry` since we now import it within the latter and we don't want imports to things in `~/pages/*` outside each page.

## How Has This Been Tested?

* Tested that the model registry page and registry selector still work
* Tested that the selected registry is retained when switching away from the page and back
* Tested that the MR page is still unavailable when the MR feature is disabled
* Tested that the other pages still work correctly with no MR-related errors when the MR feature is disabled

## Test Impact

N/A, covered by existing tests. If the model registry page and other pages still work both with and without MR enabled, that's sufficient to verify this.

## Request review criteria:

If you have UI changes: 


frontend/src/concepts/modelRegistry/context/ModelRegistrySelectorContext.tsx
```diff
@@ -20,52 +20,66 @@ export const ModelRegistrySelectorContext = React.createContext<ModelRegistrySel
...
-            {error.message}
-          </Alert>
-        </Bullseye>
-      );
-    }
+  const updatePreferredModelRegistry = React.useCallback<
+    ModelRegistrySelectorContextType['updatePreferredModelRegistry']
+  >((modelRegistry) => {
+    setPreferredModelRegistry(modelRegistry);
+  }, []);
```
christianvogt, line 55: Is this necessary, doesn't `setPreferredModelRegistry` satisfy the type for the context?

frontend/src/concepts/modelRegistry/context/ModelRegistrySelectorContext.tsx
```diff
@@ -20,52 +20,66 @@ export const ModelRegistrySelectorContext = React.createContext<ModelRegistrySel
...
+      modelRegistries,
+      preferredModelRegistry,
+      updatePreferredModelRegistry,
+    }),
+    [modelRegistries, preferredModelRegistry, updatePreferredModelRegistry],
+  );
+
+  if (!isLoaded) {
+    return <Bullseye>Loading model registries...</Bullseye>;
+  }
```
christianvogt, line 68: With the context pulled up to to the top of the app, this is going to hold up the entire application waiting to load model registries where it isn't required.

Is there a more graceful way for dependant components to handle this?

frontend/src/pages/modelRegistry/ModelRegistryCoreLoader.tsx
```diff
@@ -19,67 +20,72 @@ type ModelRegistryCoreLoaderProps = {
...
-    return <Navigate to={getInvalidRedirectPath(redirectModelRegistry.metadata.name)} replace />;
-  }
+      // They ended up on a non-valid project path
+      renderStateProps = {
+        empty: true,
+        emptyStatePage: <InvalidModelRegistry modelRegistry={modelRegistry} />,
+      };
+    } else {
+      // Redirect the namespace suffix into the URL
+      const redirectModelRegistry = preferredModelRegistry ?? modelRegistries[0];
```
christianvogt, line 67: Interesting to have a fallback here even though the same is done in the context.

frontend/src/concepts/modelRegistry/context/ModelRegistrySelectorContext.tsx
mturley: Everything inside `EnabledModelRegistrySelectorContextProvider` is existing code, just the whitespace changed because I removed the wrapping `conditionalArea`. I could improve this while I'm in here but I wonder if that's outside the scope of this PR?

frontend/src/concepts/modelRegistry/context/ModelRegistrySelectorContext.tsx
```diff
@@ -20,52 +20,66 @@ export const ModelRegistrySelectorContext = React.createContext<ModelRegistrySel
...
-    }, []);
+  const firstModelRegistry = modelRegistries.length > 0 ? modelRegistries[0] : null;
 
-    if (!isLoaded) {
-      return <Bullseye>Loading model registries...</Bullseye>;
+  React.useEffect(() => {
+    if (firstModelRegistry && !preferredModelRegistry) {
+      setPreferredModelRegistry(firstModelRegistry);
     }
+  }, [firstModelRegistry, preferredModelRegistry]);
```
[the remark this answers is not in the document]
mturley: Ditto, do you think I should improve this existing code here or maybe in a separate followup PR?

frontend/src/concepts/modelRegistry/context/ModelRegistrySelectorContext.tsx
mturley: Ah, fair point. Will look into that.

frontend/src/pages/modelRegistry/ModelRegistryCoreLoader.tsx
mturley: I wonder if that was done simply because the type can be undefined, which is necessary because there's no default value for the context until registries are loaded. Maybe the solution is to instead just not return the `<Navigate>` here if it's undefined (which as you say won't really happen)

frontend/src/concepts/modelRegistry/context/ModelRegistrySelectorContext.tsx
mturley: I'll just keep improvements to the preexisting code here in a separate commit in case we need to revert them I think.

frontend/src/concepts/modelRegistry/context/ModelRegistrySelectorContext.tsx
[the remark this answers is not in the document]
mturley: I'll just keep improvements to the preexisting code here in a separate commit in case we need to revert them I think.

frontend/src/pages/modelRegistry/ModelRegistryCoreLoader.tsx
christianvogt: Oh i thought i deleted this comment. This is fine to keep because types require you to do something.

frontend/src/concepts/modelRegistry/context/ModelRegistrySelectorContext.tsx
christianvogt: I think any time we touch code it's fair to also improve it. 
But I'll leave it to you if you choose to do something in this PR or capture it for later.

christianvogt: The `set` function returned from `React.useState` is considered stable.

lucferbux:
Oh I came late to the party, but I've reviewed it nonetheless and it's exactly what we needed. lgtm

PR #2960, opened 2024-06-27
Create NPM Cache directories for CI and run `npm ci` in Dockerfile build

manosnoam:
## Description
This reduces the cypress job duration in the CI (Jenkins) container runtime, by using preexisting npm packages that were installed during `scripts/ci/Dockerfile` build.
In addition updated to the latest minor release of NPM 20 (v20.15.0).

## How Has This Been Tested?
Running the CI Job that uses an image built from this Dockerfile, takes ~10 minutes instead of ~16 previously:
image

In the job output we can see that the `npm ci` command did not pull new packages, since all required Node packages were already available:
```
14:03:50  + git clone -b main --single-branch https://github.com/opendatahub-io/odh-dashboard odh-dashboard
14:03:50  Cloning into 'odh-dashboard'...
[Pipeline] dir
14:03:56  Running in /home/jenkins/workspace/cypress/cypress-tests/odh-dashboard
[Pipeline] {
[Pipeline] echo
14:03:56  Installing NodeJS dependencies (package-lock.json) for Cypress 'frontend' tests
[Pipeline] sh
14:03:58  + whoami
14:03:58  cypress
[Pipeline] sh
14:04:01  + npm --verbose --prefix frontend ci
14:04:01  npm verbose cli /home/cypress/nvm/versions/node/v20.15.0/bin/node /home/cypress/nvm/versions/node/v20.15.0/bin/npm
14:04:01  npm info using npm@10.7.0
14:04:01  npm info using node@v20.15.0
14:04:01  npm verbose title npm ci
14:04:01  npm verbose argv "--loglevel" "verbose" "--prefix" "frontend" "ci"
14:04:01  npm verbose logfile logs-max:10 dir:/opt/app-root/src/.npm-global/_logs/2024-06-27T11_04_01_288Z-
14:04:01  npm verbose logfile /opt/app-root/src/.npm-global/_logs/2024-06-27T11_04_01_288Z-debug-0.log
...
14:05:56  npm info run cypress@13.10.0 postinstall { code: 0, signal: null }
14:05:56  added 1954 packages, and audited 1955 packages in 2m
14:05:56  found 0 vulnerabilities
14:05:56  npm verbose exit 0
14:05:56  npm info ok
```

### DIFF
--- a/scripts/ci/Dockerfile
+++ b/scripts/ci/Dockerfile
@@ -3,11 +3,11 @@ FROM quay.io/fedora/fedora:40
 ARG USER=cypress
 ARG USER_HOME=/home/$USER
 ARG NPM_CACHE=/opt/app-root/src/.npm-global
-ARG CYPRESS_CACHE=/home/jenkins/.cypress-cache
+ARG CYPRESS_CACHE=$USER_HOME/.cypress-cache
 ARG CHROME=https://dl.google.com/linux/direct/google-chrome-stable_current_x86_64.rpm
 ARG OCP_CLI=https://mirror.openshift.com/pub/openshift-v4/clients/ocp/stable/openshift-client-linux.tar.gz
 ARG NVM_INSTALLER=https://raw.githubusercontent.com/creationix/nvm/v0.39.7/install.sh
-ARG NODE_VERSION=v20.10.0
+ARG NODE_VERSION=v20.15.0
 
 USER root
 
@@ -16,9 +16,16 @@ RUN mkdir -p "$USER_HOME"
 RUN useradd -m -g root "$USER" -d "$USER_HOME" --uid 1000
 RUN chgrp -R 0 "$USER_HOME"
 RUN chmod -R g=u "$USER_HOME"
-RUN mkdir -p $NPM_CACHE
+
+# Create global NPM cache directory
+RUN mkdir -p "$NPM_CACHE"
 RUN chgrp -R 0 "$NPM_CACHE"
-RUN chmod -R 777 $NPM_CACHE
+RUN chmod -R 777 "$NPM_CACHE"
+
+# Create local Cypress cache directory
+RUN mkdir -p "$CYPRESS_CACHE"
+RUN chgrp -R 0 "$CYPRESS_CACHE"
+RUN chmod -R 777 "$CYPRESS_CACHE"
 
 # Install tools including Xvfb and Chrome
 RUN dnf update -y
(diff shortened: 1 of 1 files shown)


scripts/ci/Dockerfile
```diff
@@ -55,16 +62,16 @@ ENV NODE_PATH "$NVM_DIR/versions/node/$NODE_VERSION/lib/node_modules"
...
 RUN /bin/bash -c "source $NVM_DIR/nvm.sh && \
     nvm install $NODE_VERSION && \
     nvm use --delete-prefix $NODE_VERSION && \
     npm config set prefix '$NPM_CACHE' && \
     echo 'export PATH=$PATH' >> '$USER_HOME/.profile' && \
     source $USER_HOME/.profile && \
-    npm install -g npm@latest && \
-    npm cache -g clean --force"
+    npm cache -g clean --force && \
+    npm ci --cache=$NPM_CACHE"
```
christianvogt, line 74: From what I read, it seems that install will still fetch from the internet and fallback to cache if not available. There were talks about using `--prefer-offline` to reverse this lookup. Can you also give this a try and provide some comparison times between with and without this PR.

lucferbux: @christianvogt Are you sure? I just took a look at [npm cache](https://docs.npmjs.com/cli/v10/commands/npm-cache) and how the [cache flag works](https://docs.npmjs.com/cli/v10/using-npm/config#cache) and I haven't been able to find any reference about that. I only see that they do a internet check first to ensure cache integrity (which i think is needed) so I would assume this is the right ting to do.
You can even do that setting the cache with `npm set cache ...`, but I think this is more straightforward.
But I don't get the part of cleaning the global cache though, if we are creating the container and we only use a custom cache folder I would assumen the global folder will be empty right?
I've built this image and it seems to be working just fine but it will be interesting checking the running difference as @christianvogt is mentioning.

lucferbux (changes requested):

scripts/ci/Dockerfile
christianvogt: I didn't get a chance yet to test this as I was just doing some googling on the topic. So take what I've said as with a grain of salt :)

After reading your response and looking a bit more I think this is correct.

At the end of the day I don't expect the npm install to be a big contributor to the overall time.

christianvogt: On a side note. When we get to actually running e2e tests, we don't need to install all the packages of the frontend; rather only those needed to run cypress. We may look at creating a sub package strictly for cypress.

PR #2969, opened 2024-06-30
feat(mr settings): permissions table

gitdallas:
Closes: https://issues.redhat.com/browse/RHOAIENG-6636

## Description
Add model registry settings > permissions page. Factored out components from ProjectSharing (project permissions).
image

## How Has This Been Tested?
Ran through creating/editing/deleting rolebindings for users/groups in model registries, also tested project permissions since that was factored out.

## Test Impact
added cypress testing for the new page

## Request review criteria:
make sure that you can create/edit/delete user/group rolebindings for model registries and that project permissions isn't broken

If you have UI changes: 


gitdallas:
i just noticed the ? icon for the default group permission in the new mock and if i aggregate the previous mocks i need it to have a tooltip that says "This group is created by default. You can add users to this group via the API."

so i'll add that

frontend/src/concepts/roleBinding/RoleBindingPermissions.tsx
```diff
@@ -0,0 +1,129 @@
...
+      variant="light"
+      id={ProjectSectionID.PERMISSIONS}
+    >
+      <Stack hasGutter>
+        <StackItem>{description}</StackItem>
+        <StackItem>
+          {projectName === MODEL_REGISTRY_DEFAULT_NAMESPACE ? groupTable : userTable}
+        </StackItem>
+        <StackItem>
+          {projectName === MODEL_REGISTRY_DEFAULT_NAMESPACE ? userTable : groupTable}
```
mturley, line 122: I think because this component is under its own `~/concepts/roleBinding` it shouldn't have any logic specific to Model Registry. Can you add a prop like `groupsBeforeUsers: boolean` or something to pass instead?

frontend/src/concepts/roleBinding/RoleBindingPermissionsTableRow.tsx
```diff
@@ -41,18 +43,20 @@ const ProjectSharingTableRow: React.FC<ProjectSharingTableRowProps> = ({
...
-  const [roleBindingRoleRef, setRoleBindingRoleRef] = React.useState<ProjectSharingRoleType>(
-    defaultValueRole(obj),
-  );
+  const [roleBindingRoleRef, setRoleBindingRoleRef] =
+    React.useState<RoleBindingPermissionsRoleType>(defaultValueRole(obj));
   const [isLoading, setIsLoading] = React.useState(false);
   const createdDate = new Date(obj.metadata.creationTimestamp || '');
+  const isModelRegistry = obj.metadata.namespace === MODEL_REGISTRY_DEFAULT_NAMESPACE;
+  const disableActions =
+    isModelRegistry && obj.metadata.name === `${obj.metadata.labels?.app}-users`;
```
mturley, line 52: Same here, can we remove `isModelRegistry` and references to `MODEL_REGISTRY_DEFAULT_NAMESPACE` and instead lift out props that describe what the actual effects of the logic are? Like for example `RoleBindingPermissionsTableRow` could have a `hasPermissionSelection: boolean` prop and maybe a `disableActionsForRbApp: string`... although...

Are we sure this is the right `disableActions` logic? We want to disable actions for the rolebinding associated with the default group. Maybe we instead want to create a prop like `disableActionsForSubject?: RoleBindingSubject`, and then the logic here can be something like:

```ts
  const disableActions =
    disableActionsForSubject &&
    !!obj.subjects.find(
      ({ kind, name }) =>
        kind === disableActionsForSubject.kind && name === disableActionsForSubject.name,
    );
```

and then you'd pass down the prop:

```ts
  disableActionsForSubject={{
    kind: 'Group',
    apiGroup: 'rbac.authorization.k8s.io',
    name: `${modelRegistry.metadata.name}-users`,
  }}
```

For both of these examples, you'd need to lift out these new props `hasPermissionSelection` and `disableActionsForSubject` (or whatever you call them) a few layers, through `RoleBindingPermissionsTable`, `RoleBindingPermissionsTableSection`, `RoleBindingPermissions`, until you get to `ModelRegistriesPermissions` where you can pass their values specific to MR. The goal here is to not have any imports from `~/concepts/modelRegistry` inside `~/concepts/roleBinding` if possible. It'll be more extensible/maintainable that way.

frontend/src/concepts/roleBinding/RoleBindingPermissionsTableRow.tsx
```diff
@@ -67,8 +71,8 @@ const ProjectSharingTableRow: React.FC<ProjectSharingTableRowProps> = ({
           )}
         </Td>
         <Td dataLabel="Permission">
-          {isEditing ? (
-            <ProjectSharingPermissionSelection
+          {isEditing && !isModelRegistry ? (
```
mturley, line 74: On second thought, do you even really need the `hasPermissionSelection` logic? Maybe you simply only show the selection dropdown if there's more than one option, since MR will only have the single "Default" option?

```ts
isEditing && permissionOptions.length > 1
```

That would keep you from having to lift out `hasPermissionSelection` through all those component layers like I described above.

Also, it looks like you're not passing down `permissionOptions` into `RoleBindingPermissionsPermissionSelection` here.

frontend/src/concepts/roleBinding/RoleBindingPermissionsTableRowAdd.tsx
```diff
@@ -2,35 +2,41 @@ import * as React from 'react';
...
+    type: RoleBindingPermissionsRoleType;
+    description: string;
+  }[];
+  onChange: (name: string, roleType: RoleBindingPermissionsRoleType) => void;
   onCancel: () => void;
 };
 
-/** @deprecated - this should use ProjectSharingTableRow */
-const ProjectSharingTableRowAdd: React.FC<ProjectSharingTableRowPropsAdd> = ({
+/** @deprecated - this should use RoleBindingPermissionsTableRow */
```
mturley, line 20: Would it be a lot of effort to actually deduplicate these and address this deprecation?

frontend/src/concepts/roleBinding/RoleBindingPermissionsTableRowAdd.tsx
```diff
@@ -42,7 +48,8 @@ const ProjectSharingTableRowAdd: React.FC<ProjectSharingTableRowPropsAdd> = ({
           />
         </Td>
         <Td dataLabel="Permission">
-          <ProjectSharingPermissionSelection
+          <RoleBindingPermissionsPermissionSelection
+            permissionOptions={permissionOptions}
```
mturley, line 52: You probably don't want to show this selection dropdown if there's only one option, just like when editing the regular `RoleBindingPermissionsTableRow`.

frontend/src/concepts/roleBinding/RoleBindingPermissionsTableSection.tsx
```diff
@@ -13,23 +13,34 @@ import {
 import { RoleBindingKind } from '~/k8sTypes';
 import HeaderIcon from '~/concepts/design/HeaderIcon';
 import { ProjectObjectType } from '~/concepts/design/utils';
-import ProjectSharingTable from './ProjectSharingTable';
-import { ProjectSharingRBType } from './types';
+import RoleBindingPermissionsTable from './RoleBindingPermissionsTable';
+import { RoleBindingPermissionsRBType, RoleBindingPermissionsRoleType } from './types';
 
-export type ProjectSharingTableSectionAltProps = {
+export type RoleBindingPermissionsTableSectionAltProps = {
```
mturley, line 19: Why `Alt` here?

frontend/src/concepts/roleBinding/utils.ts
```diff
@@ -0,0 +1,30 @@
...
+  }
+  return RoleBindingPermissionsRoleType.DEFAULT;
+};
+
+export const firstSubject = (roleBinding: RoleBindingKind): string =>
+  roleBinding.subjects[0]?.name || '';
+
+export const roleLabel = (value: RoleBindingPermissionsRoleType): string => {
+  if (value === RoleBindingPermissionsRoleType.EDIT) {
+    return 'Contributer';
```
mturley, line 27: Typo, `Contributer` -> `Contributor` with an O.

frontend/src/k8sTypes.ts
```diff
@@ -22,6 +22,7 @@ export enum KnownLabels {
   PROJECT_SHARING = 'opendatahub.io/project-sharing',
   MODEL_SERVING_PROJECT = 'modelmesh-enabled',
   DATA_CONNECTION_AWS = 'opendatahub.io/managed',
+  MODEL_REGISTRY_SHARING = 'opendatahub.io/model-registry-sharing',
```
mturley, line 25: You're not actually applying this label to your new roleBindings I think. We probably don't need it now that we're filtering by `app.kubernetes.io/name` in `ModelRegistriesManagePermissions` and using the `component=model-registry` labelSelector in `useRoleBindings`.

frontend/src/pages/modelRegistrySettings/ModelRegistriesPermissions.tsx
```diff
@@ -0,0 +1,106 @@
...
+import { MODEL_REGISTRY_DEFAULT_NAMESPACE } from '~/concepts/modelRegistry/const';
+import { SupportedArea } from '~/concepts/areas';
+import { RoleBindingPermissionsRoleType } from '~/concepts/roleBinding/types';
+import useRoleBindings from './useRoleBindings';
+
+const ModelRegistriesManagePermissions: React.FC = () => {
+  const [activeTabKey, setActiveTabKey] = React.useState('users');
+  const [groups] = useGroups();
+  const roleBindings = useContextResourceData<RoleBindingKind>(useRoleBindings());
+  const { mrName } = useParams();
```
mturley, line 27: This page should probably redirect back to `/modelRegistrySettings` if the `mrName` doesn't match an existing modelRegistry (as part of 404 handling). You could do this by fetching modelRegistries here with `useModelRegistriesBackend` and then if that fetch is done and there isn't a MR matching `mrName`, return a `<Navigate to="/modelRegistrySettings" replace />`. or maybe you could just do that if `filteredRoleBindings.length === 0` since in theory there's always the RB for the default group.

frontend/src/pages/modelRegistrySettings/useRoleBindings.ts
```diff
@@ -0,0 +1,21 @@
+import * as React from 'react';
+import { listRoleBindings } from '~/api';
+import { RoleBindingKind } from '~/k8sTypes';
+import useFetchState, { FetchState } from '~/utilities/useFetchState';
+
+const useRoleBindings = (): FetchState<RoleBindingKind[]> => {
```
mturley, line 6: Can we name this something that reflects its filter? `useModelRegistryRoleBindings` maybe?

frontend/src/pages/modelRegistrySettings/useRoleBindings.ts
```diff
@@ -0,0 +1,21 @@
...
+  const getRoleBindings = React.useCallback(
+    () =>
+      listRoleBindings('odh-model-registries', 'component=model-registry').catch((e) => {
```
mturley, line 9: Instead of hard-coding the `odh-model-registries` namespace we should import and use `MODEL_REGISTRY_DEFAULT_NAMESPACE`.

Also, instead of hard-coding `component=model-registry` can we create a new `LABEL_SELECTOR_MODEL_REGISTRY` in `frontend/src/const.ts`? And as part of that can you create `KnownLabels.COMPONENT` in `frontend/src/k8sTypes.ts`?

mturley (changes requested):
Really good stuff @gitdallas . I have a couple of things, hopefully it won't be too bad to address.

frontend/src/__mocks__/mockRoleBindingK8sResource.ts
```diff
@@ -22,22 +23,38 @@ export const mockRoleBindingK8sResource = ({
...
+      name,
+      namespace,
+      uid,
+      creationTimestamp: '2023-02-14T21:43:59Z',
+      labels,
+    },
+    subjects,
+    roleRef: {
+      apiGroup: 'rbac.authorization.k8s.io',
+      kind: 'ClusterRole',
```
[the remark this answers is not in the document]
gitdallas: @lucferbux - do you know if Model Registries should have `Role` vs `ClusterRole`? Would it be different for User vs Group?

frontend/src/concepts/roleBinding/RoleBindingPermissionsTableRow.tsx
gitdallas: hm.. with latest change i'm putting "This group is created by default. You can add users to this group via the API", so maybe I need to pass an `isDefaultGroup` prop of some sort

frontend/src/concepts/roleBinding/RoleBindingPermissionsTableRowAdd.tsx
gitdallas: i haven't looked into it too much, but would prefer that be another story/concern

frontend/src/concepts/roleBinding/RoleBindingPermissionsTableRow.tsx
gitdallas: with the tooltip now also on the defaultGroup... might need an `isDefaultGroup` prop of some sort

frontend/src/concepts/roleBinding/RoleBindingPermissionsTableSection.tsx
gitdallas: idk, that's just the previous name with my ProjectSharing->RoleBindingPermissions swap

frontend/src/k8sTypes.ts
gitdallas: does the "probably" mean we should get confirmation from @lucferbux ?

mturley:
@gitdallas while you're in here, can you actually also please look at the `DeleteModelRegistryModal` and replace the placeholder text "TODO: group here" with `{mr.metadata.name}-users`? You can just inline it and get rid of the `defaultPermissionsGroup` variable. I thought we'd have to fetch something to look that up but since we're making the naming convention assumption here, we can use it there too.

frontend/src/__mocks__/mockRoleBindingK8sResource.ts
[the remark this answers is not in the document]
mturley: I'll see if Lucas says anything different, but it just needs to match whatever the kind is of the actual role that gets generated by the backend that we're assuming exists with the name `registry-user-${mrName}`. Looking at the one generated for the MR you created yesterday, it's a `Role`:

Screenshot 2024-07-03 at 12 34 22 PM

frontend/src/api/k8s/roleBindings.ts
```diff
@@ -46,22 +46,23 @@ export const generateRoleBindingServingRuntime = (
...
       name: `dashboard-permissions-${genRandomChars()}`,
       namespace,
-      labels: {
-        [KnownLabels.DASHBOARD_RESOURCE]: 'true',
-        [KnownLabels.PROJECT_SHARING]: 'true',
-      },
+      labels: rbLabels,
     },
     roleRef: {
       apiGroup: 'rbac.authorization.k8s.io',
```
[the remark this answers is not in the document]
mturley: The roleRef will be the same for user and group subjects, the RB is binding either a user or a group to the single role that gets generated automatically when the MR is created, `registry-user-${mrName}`. Don't let the word "user" in the role name throw you off -- a group bound to it just means "all users in this group are users of this MR".

For Role vs ClusterRole: A Role is a namespaced resource, so it exists inside a single namespace and only grants access to things in that namespace (in our case, inside odh-model-registries). A ClusterRole is a cluster-scoped resource (it has no namespace, exists globally) and it grants access to other cluster-scoped things or to things across multiple namespaces.

For example, to make a user a cluster admin, you create a cluster-scoped RoleBinding with that user as a subject and the ClusterRole called `cluster-admin` in its roleRef.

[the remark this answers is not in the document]
mturley: More info about Roles/ClusterRoles/RoleBindings/ClusterRoleBindings: https://kubernetes.io/docs/reference/access-authn-authz/rbac/

frontend/src/concepts/roleBinding/RoleBindingPermissionsTableRow.tsx
mturley: Yeah maybe instead of `disableActionsForSubject` it's just `defaultGroupName` as an optional prop, and if you pass a default group it disables actions on that group and also adds that help text to it.

frontend/src/concepts/roleBinding/RoleBindingPermissionsTableRowAdd.tsx
mturley: Sounds good. I'll make a note to open a tech debt followup task.

frontend/src/concepts/roleBinding/RoleBindingPermissionsTableSection.tsx
mturley: Fair enough

frontend/src/k8sTypes.ts
mturley: I think we can just ditch it, it was my idea and turns out I was wrong about it being necessary

frontend/src/pages/modelRegistrySettings/ModelRegistriesPermissions.tsx
mturley: If you prefer we can also open a followup bug for this, but we just need to make sure the page doesn't crash if the MR with `mrName` doesn't exist.

frontend/src/concepts/roleBinding/RoleBindingPermissions.tsx
mturley: This could be followup tech debt, I'll open an issue.

frontend/src/concepts/roleBinding/RoleBindingPermissionsTableRow.tsx
mturley: We could open a tech debt issue for this too if you can't get to it today, just because I guess technically the auto-generated RoleBinding has the name `${obj.metadata.labels?.app}-users` in addition to the group itself so this logic will work.

frontend/src/pages/modelRegistrySettings/useRoleBindings.ts
mturley: This could be a follow-up tech debt task if you can't get to it today.

frontend/src/concepts/roleBinding/RoleBindingPermissionsTableRowAdd.tsx
gitdallas: kept forgetting that edit/add were two separate things :p

mturley:
Looks like the test failure is just Prettier

frontend/src/api/k8s/roleBindings.ts
```diff
@@ -46,26 +46,28 @@ export const generateRoleBindingServingRuntime = (
...
 
-export const generateRoleBindingProjectSharing = (
+export const generateRoleBindingPermissions = (
   namespace: string,
-  rbSubjectType: ProjectSharingRBType,
+  rbSubjectType: RoleBindingPermissionsRBType,
   rbSubjectName: string,
-  rbRoleRefType: ProjectSharingRoleType,
+  rbRoleRefType: RoleBindingPermissionsRoleType | string, //string because with MR this can include MR name
+  rbRoleRefKind: string,
```
mturley, line 54: This should be `rbRoleRefKind: 'Role' | 'ClusterRole'`

frontend/src/concepts/roleBinding/RoleBindingPermissionsTable.tsx
```diff
@@ -1,56 +1,68 @@
...
             onChange={(name, roleType) => {
-              const newRBObject = generateRoleBindingProjectSharing(
-                currentProject.metadata.name,
+              const newRBObject = generateRoleBindingPermissions(
+                projectName,
                 type,
                 name,
-                roleType,
+                roleRef || roleType,
+                'Role',
```
mturley, line 64: Instead of passing in `'Role'` here you need to lift the `rbRoleRefKind` prop out with the others so you can pass in Role for MR and ClusterRole for projects. I think that's what's causing your test failure

frontend/src/concepts/roleBinding/RoleBindingPermissionsTable.tsx
```diff
@@ -66,18 +78,22 @@ const ProjectSharingTable: React.FC<ProjectSharingTableProps> = ({
...
           onChange={(name, roleType) => {
-            const newRBObject = generateRoleBindingProjectSharing(
-              currentProject.metadata.name,
+            const newRBObject = generateRoleBindingPermissions(
+              projectName,
               type,
               name,
-              roleType,
+              roleRef || roleType,
+              'Role',
```
mturley, line 95: Same here

frontend/src/concepts/roleBinding/RoleBindingPermissionsTableRow.tsx
```diff
@@ -2,57 +2,67 @@ import * as React from 'react';
...
 }) => {
   const [roleBindingName, setRoleBindingName] = React.useState(defaultValueName(obj));
-  const [roleBindingRoleRef, setRoleBindingRoleRef] = React.useState<ProjectSharingRoleType>(
-    defaultValueRole(obj),
-  );
+  const [roleBindingRoleRef, setRoleBindingRoleRef] =
+    React.useState<RoleBindingPermissionsRoleType>(defaultValueRole(obj));
   const [isLoading, setIsLoading] = React.useState(false);
   const createdDate = new Date(obj.metadata.creationTimestamp || '');
+  const isDefaultGroup = obj.metadata.name === defaultGroupName;
```
mturley, line 58: Ok so don't kill me, but technically `obj.metadata.name` is the rolebinding name, not the group name (the rolebinding's subject). The logic I was imagining is that `isDefaultGroup` is true if the `obj.subjects` contains a subject of kind `'Group'` and name equal to `defaultGroupName`. Buuut you could also just rename this prop to `defaultRoleBindingName`.

frontend/src/concepts/roleBinding/RoleBindingPermissionsTableRowAdd.tsx
mturley: Opened a task to follow up on this: https://issues.redhat.com/browse/RHOAIENG-9457

mturley (commented):
Tested and it's working well for me on the MR side, and the failures on the project side are caused by the lines I comment on below. Also one last nit. Otherwise all the changes LGTM.

frontend/src/api/k8s/roleBindings.ts
gitdallas: i had that originally, but then noticed that the `RoleBindingSubject.kind` type was `string`, so thought I should match that

frontend/src/concepts/roleBinding/RoleBindingPermissionsTable.tsx
gitdallas: oh, gah.. lol i think i just did a project wide copy/paste regex on generateRoleBindingPermissions and didn't check where it was 🤦

frontend/src/concepts/roleBinding/RoleBindingPermissionsTableRow.tsx
gitdallas: 🔪

frontend/src/api/k8s/roleBindings.ts
lucferbux: That's correct, and that's a great catch, rolebindings could only be `Role` or `ClusterRole`, so we should limit

frontend/src/api/k8s/roleBindings.ts
```diff
@@ -46,26 +46,28 @@ export const generateRoleBindingServingRuntime = (
...
 };
 
-export const generateRoleBindingProjectSharing = (
+export const generateRoleBindingPermissions = (
   namespace: string,
-  rbSubjectType: ProjectSharingRBType,
+  rbSubjectType: RoleBindingPermissionsRBType,
   rbSubjectName: string,
-  rbRoleRefType: ProjectSharingRoleType,
+  rbRoleRefType: RoleBindingPermissionsRoleType | string, //string because with MR this can include MR name
```
lucferbux, line 53: btw, one enhancement we should add is `ownerReferences` for the rolebinding, we should either add it into the service or the role, if we delete the modelregistry we would want all the rolebindings to be gone

lucferbux (changes requested):
Ok I agree on @mturley changes and I added one more thing to do. Rest lgtm

frontend/src/api/k8s/roleBindings.ts
gitdallas: @lucferbux - should i change the `RoleBindngSubject.kind` type in `k8sTypes`?

frontend/src/api/k8s/roleBindings.ts
gitdallas: @lucferbux sorry, i'm not sure what you mean exactly. looks like `ownerReference` is part of the metadata for `K8sResourceCommon` so I could add it to the metadata of the roleBindingObject?

what values would i use though?
```
    apiVersion: string;
    kind: string;
    name: string;
    uid: string;
    controller?: boolean;
    blockOwnerDeletion?: boolean;
```

lucferbux: I mean [garbage collection](https://kubernetes.io/docs/concepts/architecture/garbage-collection/#owners-dependents). TBH I thought rolebindings were dependant of roles so when you deleted a role the rolebinding was deleted (not sure why i thought that was the behavior). But testing in the cluster, when I delete a model registry, all resoruces are deleted execpt rolebindings. Steps to replicate are:

1. Create a model registry in the admin panel
2. Manage access and add some users/groups
3. Delete the model registry
4. Check rolebindings in the namespace, rolebindings are still there

Just add a [owner reference](https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/#owner-references-in-object-specifications) to the role.

lucferbux: We have a util method to add owner references [here](https://github.com/opendatahub-io/odh-dashboard/blob/main/frontend/src/api/k8sUtils.ts#L7) let me know if you need anything.

frontend/src/api/k8s/roleBindings.ts
lucferbux: @gitdallas yup, that should be the way to go, we should be mindful of our types, and try to be as precise as possible.

gitdallas: @lucferbux - actually, do we need `RoleBindingSubject` in both both `frontend/src/types.ts` and `frontend/src/k8sTypes.ts`?

lucferbux: Oh no, not at all, good catch, we should delete all duplicates, it could happen when the codebase grows that large.

Keep the one in `k8sTypes`

frontend/src/api/k8s/roleBindings.ts
gitdallas: @lucferbux - i tried in [this commit](https://github.com/opendatahub-io/odh-dashboard/pull/2969/commits/dad2e6c1d3d4235da0c3f3bab1242674215caa25), but it doesn't seem to clear the rolebindings with the mr

gitdallas: @lucferbux - i ended up using the MR CR, which required more changes to retrieve it and pass it through... but it seems to be working now

lucferbux: Great, MR CR can be the owner reference since that's the higher order parent, so that's perfect.

gitdallas:
failing test seems unrelated (compareRuns graph)

frontend/src/pages/modelRegistrySettings/ModelRegistriesPermissions.tsx
```diff
@@ -0,0 +1,130 @@
...
+                'app.kubernetes.io/component': SupportedArea.MODEL_REGISTRY,
+                'app.kubernetes.io/part-of': SupportedArea.MODEL_REGISTRY,
+                'app.kubernetes.io/name': mrName || '',
+                component: SupportedArea.MODEL_REGISTRY,
+              }}
+              projectName={MODEL_REGISTRY_DEFAULT_NAMESPACE}
+              description={
+                <>
+                  To enable access for all cluster users, add{' '}
+                  <ClipboardCopy variant="inline-compact">system.authenticated</ClipboardCopy> to
```
lucferbux, line 109: nit: This is `system:authenticated`

lucferbux (changes requested):
ok, just a nit, rest of code looks great, once you have that change we can merge it!

lucferbux (approved):
/lgtm

lucferbux:
mmmm weird, some pipeline test has failed. let me try and retest it
/retest

gitdallas:
@lucferbux - yea, mentioned that https://github.com/opendatahub-io/odh-dashboard/pull/2969#issuecomment-2213787442

seems consistent

PR #2981, opened 2024-07-04
Handle only numbers workbench names

manaswinidas:
Closes: https://issues.redhat.com/browse/RHOAIENG-824

## Description

Workbenches start fine with names containing only numbers. This PR adds a prefix wb- for names starting with only numbers.

Screenshot 2024-07-04 at 3 47 51 PM

## How Has This Been Tested?

1. Try creating a workbench that contains only numbers and check whether it starts fine

## Test Impact

Added both unit and e2e tests

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/concepts/k8s/utils.ts
+++ b/frontend/src/concepts/k8s/utils.ts
@@ -5,12 +5,24 @@ export const getDisplayNameFromK8sResource = (resource: K8sDSGResource): string 
 export const getDescriptionFromK8sResource = (resource: K8sDSGResource): string =>
   resource.metadata.annotations?.['openshift.io/description'] || '';
 
-export const translateDisplayNameForK8s = (name: string): string =>
-  name
+/**
+ * Converts a display name to a k8s safe variant, if needed a `safeK8sPrefix` can be provided to customize the prefix used to align with k8s standards if it is needed.
+ */
+export const translateDisplayNameForK8s = (name: string, safeK8sPrefix?: string): string => {
+  const translatedName: string = name
     .trim()
     .toLowerCase()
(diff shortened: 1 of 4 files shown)


dpanshug (changes requested):

frontend/src/concepts/k8s/utils.ts
```diff
@@ -12,5 +12,14 @@ export const translateDisplayNameForK8s = (name: string): string =>
     .replace(/\s/g, '-')
     .replace(/[^A-Za-z0-9-]/g, '');
 
+export const translateNameForWb = (name: string): string => {
```
andrewballantyne, line 15: I like the way you did this -- but I think we're still going to cause an issue with `translateDisplayNameForK8s` on other resources -- can we not just take an optional prefix or something that makes the signature:

```
/** 
 * Converts a display name to a k8s safe variant, if needed a `safeK8sPrefix` can be provided to customize the prefix used to align with k8s standards if it is needed.
 */
export const translateDisplayNameForK8s = (name: string, safeK8sPrefix?: string = 'ai-'): string
```
And we can write the logic for checking if it is valid to add the prefix as needed?

I don't imagine we want to inject a common pattern always, so like your if-statement below, we should have that condition still. 

And you can provide `wb-` for the workbench resource.

andrewballantyne (changes requested):

frontend/src/concepts/k8s/utils.ts
manaswinidas: Good idea! Updated accordingly.

andrewballantyne (approved):

PR #2992, opened 2024-07-10
refactor model registry proxy onto proxy util

christianvogt:
https://issues.redhat.com/browse/RHOAIENG-8891

## Description

Model registry had a slight variation in requires for their backend proxy than the other proxy services and therefore the common proxy util was not usable without being refactored.

## How Has This Been Tested?

Find a cluster with model-registry setup or perform the setup on your cluster.
Login to the cluster on the terminal.
- run a local backend: `npm run start:dev`
- run a local frontend: `npm run start:dev`
- from the root dir, `NAMESPACE=default make port-forward`
  - make sure the env var `MODEL_REGISTRY_NAME` in the `.env.development` is correct for your cluster
- visit `http://localhost:4010/modelRegistry`

Make sure the page loads and functions as expected

## Test Impact

None, unfortunately backend doesn't have tests at this time.

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/backend/src/routes/api/service/modelregistry/index.ts
+++ b/backend/src/routes/api/service/modelregistry/index.ts
@@ -1,32 +1,18 @@
-import httpProxy from '@fastify/http-proxy';
-import { KubeFastifyInstance } from '../../../../types';
-import { DEV_MODE, MODEL_REGISTRY_NAMESPACE } from '../../../../utils/constants';
-import { getParam, setParam } from '../../../../utils/proxy';
+import { MODEL_REGISTRY_NAMESPACE } from '../../../../utils/constants';
+import { proxyService } from '../../../../utils/proxy';
 
-export default async (fastify: KubeFastifyInstance): Promise<void> => {
-  fastify.register(httpProxy, {
-    upstream: '',
-    prefix: '/:name',
-    rewritePrefix: '',
-    replyOptions: {
-      // preHandler must set the `upstream` param
-      getUpstream: (request) => getParam(request, 'upstream'),
-    },
-    preHandler: (request, _, done) => {
-      const name = getParam(request, 'name');
-
-      const upstream = DEV_MODE
-        ? // Use port forwarding for local development:
-          // kubectl port-forward -n <namespace> svc/<service-name> <local.port>:<service.port>
-          `http://${process.env.MODEL_REGISTRY_SERVICE_HOST}:${process.env.MODEL_REGISTRY_SERVICE_PORT}`
-        : // Construct service URL
-          `http://${name}.${MODEL_REGISTRY_NAMESPACE}.svc.cluster.local:8080`;
-
-      // assign the `upstream` param so we can dynamically set the upstream URL for http-proxy
-      setParam(request, 'upstream', upstream);
-
-      fastify.log.info(`Proxy ${request.method} request ${request.url} to ${upstream}`);
-      done();
-    },
-  });
-};
+export default proxyService(
+  null,
+  {
+    port: 8080,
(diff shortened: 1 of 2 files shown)


christianvogt:
cc @lucferbux

backend/src/routes/api/service/modelregistry/index.ts
```diff
@@ -1,32 +1,18 @@
...
+    port: 8080,
+    namespace: MODEL_REGISTRY_NAMESPACE,
+  },
+  {
+    // Use port forwarding for local development:
+    // kubectl port-forward -n odh-model-registries svc/<service-name> 8085:8080
+    host: process.env.MODEL_REGISTRY_SERVICE_HOST,
+    port: process.env.MODEL_REGISTRY_SERVICE_PORT,
+  },
+  null,
```
lucferbux, line 16: We didn't implement at first status check for model registry, but it's based on the `ModelRegistry` `CR` availability, in this case we have

```yaml
status:
  conditions:
    - lastTransitionTime: '2024-07-11T19:47:12Z'
      message: Deployment is available
      reason: DeploymentAvailable
      status: 'True'
      type: Available
```
So something like:

```typescript
  (resource) =>
    !!resource.status?.conditions?.find((c) => c.type === 'Available' && c.status === 'True'),
```
would do the trick.

backend/src/routes/api/service/modelregistry/index.ts
```diff
@@ -1,32 +1,18 @@
...
+  false,
```
lucferbux, line 17: @mturley If we set this to true we will be covering part of https://issues.redhat.com/browse/RHOAIENG-9037, would you want to split that into other task and have it done right away or should we wait until we get to that story?

lucferbux (changes requested):
Ok, I've tested the logic of `proxy.ts` and it's working just fine, and even if we apply the `statusCheck`, I've left just a couple of questions that I would wanna have settled before approving the PR.

backend/src/routes/api/service/modelregistry/index.ts
mturley: @lucferbux setting this to true means Istio will become required to use the MR API, right? We can ask the backend devs to help us enable it on our shared cluster, but it means until they have that configuration automatic in the operator (https://issues.redhat.com/browse/RHOAIENG-9907) any fresh installations of model registry will be broken. Not sure if we should do that.

backend/src/routes/api/service/modelregistry/index.ts
mturley: +1, this will be the same logic we'll soon use for the Available status in the UI as part of https://issues.redhat.com/browse/RHOAIENG-7485

backend/src/utils/proxy.ts
```diff
@@ -27,12 +27,13 @@ const notFoundError = (kind: string, name: string, e?: any, overrideMessage?: st
 };
 
 export const proxyService =
-  <K extends K8sResourceCommon>(
-    model: { apiGroup: string; apiVersion: string; plural: string; kind: string },
+  <K extends K8sResourceCommon = never>(
```
mturley, line 30: ha, I never considered this as a way to make a generic optional. clever.

lucferbux: Yes, indeed, I thought the same hahahha

backend/src/routes/api/service/modelregistry/index.ts
lucferbux: Ok, we'll enable it later

lucferbux (approved):
/lgtm

After some internal discussion we can use it like this and tweak it when ready for enabling tls

PR #2989, opened 2024-07-10
RHOAIENG-9230 Restructure InitSegment

pilhuhn:
Restructure InitSegment
https://issues.redhat.com/browse/RHOAIENG-9230

## Description
Split off Segment initialisation from usage.
Remove username from init code and only use when needed to fire an identify event.
No user-visible changes.

## How Has This Been Tested?
Manual test by looking at browser console 

## Request review criteria:

Self checklist (all need to be checked):
- [X] The developer has manually tested the changes and verified that the changes work
- [X] Commits have been squashed into descriptive, self-contained units of work (e.g. 'WIP' and 'Implements feedback' style messages have been removed)
- [ ] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [X] The developer has added tests or explained why testing cannot be added (unit or cypress tests for related changes)

After the PR is posted & before it merges:
- [ ] The developer has tested their solution on a cluster by using the image produced by the PR to `main`

cc @andrewballantyne

PR #2995, opened 2024-07-11
[RHOAIENG-9476] Create base firing event

pilhuhn:
Closes https://issues.redhat.com/browse/RHOAIENG-9476

## Description
Break out the various types of segment interaction from the one function into specialised ones.
Have typed methods for custom-event tracking (i.e. events other than page-view).
Get rid of the free-form fireTrackingEventRaw function.

## How Has This Been Tested?
Manual testing and watching the developer console

## Request review criteria:

@andrewballantyne does this go into the direction you thought of?
Note this builds on top of #2989

### DIFF
--- a/frontend/src/components/ExternalLink.tsx
+++ b/frontend/src/components/ExternalLink.tsx
@@ -1,7 +1,7 @@
 import * as React from 'react';
 import { Button } from '@patternfly/react-core';
 import { ExternalLinkAltIcon } from '@patternfly/react-icons';
-import { fireTrackingEventRaw } from '~/concepts/analyticsTracking/segmentIOUtils';
+import { fireLinkTrackingEvent } from '~/concepts/analyticsTracking/segmentIOUtils';
 
 type ExternalLinkProps = {
   text: string;
@@ -14,7 +14,7 @@ const ExternalLink: React.FC<ExternalLinkProps> = ({ text, to }) => (
     isInline
     onClick={() => {
       window.open(to);
-      fireTrackingEventRaw('ExternalLink Clicked', { href: to, from: window.location.pathname });
+      fireLinkTrackingEvent('ExternalLink Clicked', { href: to, from: window.location.pathname });
     }}
     icon={<ExternalLinkAltIcon />}
     iconPosition="right"
--- a/frontend/src/concepts/analyticsTracking/segmentIOUtils.tsx
+++ b/frontend/src/concepts/analyticsTracking/segmentIOUtils.tsx
@@ -1,22 +1,59 @@
 import { DEV_MODE, INTERNAL_DASHBOARD_VERSION } from '~/utilities/const';
-import { TrackingEventProperties } from '~/concepts/analyticsTracking/trackingProperties';
+import {
+  BaseTrackingEventProperties,
+  FormTrackingEventProperties,
+  IdentifyEventProperties,
+  LinkTrackingEventProperties,
+  MiscTrackingEventProperties,
+} from '~/concepts/analyticsTracking/trackingProperties';
 
-// The following is like the original method below, but allows for more 'free form' properties.
-// eslint-disable-next-line @typescript-eslint/no-explicit-any,@typescript-eslint/explicit-module-boundary-types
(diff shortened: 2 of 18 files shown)


frontend/src/concepts/analyticsTracking/segmentIOUtils.tsx
```diff
@@ -1,22 +1,59 @@
...
+export const fireMiscTrackingEvent = (
+  eventType: string,
+  properties: MiscTrackingEventProperties,
+): void => {
+  if (DEV_MODE) {
+    /* eslint-disable-next-line no-console */
+    console.warn('This tracking event type is a last resort for legacy purposes');
+  }
+  fireTrackingEvent(eventType, properties);
+};
```
andrewballantyne, line 33: Should these not be `eventName: string` as well?

frontend/src/components/ExternalLink.tsx
```diff
@@ -14,7 +14,7 @@ const ExternalLink: React.FC<ExternalLinkProps> = ({ text, to }) => (
     isInline
     onClick={() => {
       window.open(to);
-      fireTrackingEventRaw('ExternalLink Clicked', { href: to, from: window.location.pathname });
+      fireLinkTrackingEvent('ExternalLink Clicked', { href: to, from: window.location.pathname });
```
andrewballantyne, line 17: Question, not request... did we want to create an Enum for all of these `eventName` or `eventType` like system? That way we can look at it in one spot and know what is already in place.

frontend/src/pages/home/projects/ProjectCard.tsx
```diff
@@ -38,10 +39,11 @@ const ProjectCard: React.FC<ProjectCardProps> = ({ project }) => {
           isInline
           onClick={() => {
             navigate(`/projects/${project.metadata.name}`);
-            fireTrackingEventRaw('HomeCardClicked', {
+            const props: LinkTrackingEventProperties = {
               to: `/projects/${project.metadata.name}`,
               type: 'project',
-            });
+            };
+            fireLinkTrackingEvent('HomeCardClicked', props);
```
andrewballantyne, line 46: You can do this all in 1 call:

```
fireLinkTrackingEvent('HomeCardClicked', {
    to: `/projects/${project.metadata.name}`,
    type: 'project',
} satisfies LinkTrackingEventProperties);
```

frontend/src/pages/projects/screens/spawner/SpawnerFooter.tsx
```diff
@@ -87,7 +87,7 @@ const SpawnerFooter: React.FC<SpawnerFooterProps> = ({
 
   const afterStart = (name: string, type: 'created' | 'updated') => {
     const { acceleratorProfile, notebookSize, image } = startNotebookData;
-    fireTrackingEventRaw(`Workbench ${type === 'created' ? 'Created' : 'Updated'}`, {
+    const tep: FormTrackingEventProperties = {
```
andrewballantyne, line 90: There are several created items here and I think we need to figure out a way to just pass it to the event call. If we need to build this complexity, I wonder if we should look at making custom hooks and utilities to craft this to avoid it polluting the forms with this functionality. 

We may need to look at all the events after we have done the effort to make sure we aren't losing some good refactoring and cleanup items. The goal is to structure all these in such a fashion to support segment more cleanly throughout our app -- so maybe we need to have higher layer functions for each form or something. Not sure. Like I said, we need to work through this once we have an understanding of each usage better.

pilhuhn: I am not sure if this is now an ask to just inline `tep` or a forward looking comment for future work (potentially outside this PR).

andrewballantyne (approved):

frontend/src/pages/projects/screens/spawner/SpawnerFooter.tsx
andrewballantyne: The variable was a bit odd, but basically this kind of data probably needs to be filtered through a function and not so "on the form" so segment functionality is a little less "in the middle of stuff" and more documented for the use-cases it wants to provide. Definitely future work.

Issue #1932, opened 2023-10-09
[Feature Request]: Disable "Edit project" action from the menu for users without Admin permissions


bdattoma:
### Feature description

When users with Edit access to a DS project tries to edit the project details, they get correctly a permission error:
![image](https://github.com/opendatahub-io/odh-dashboard/assets/88311595/28974e5c-7b5e-46cd-a9a2-abe3ce992088)

However, I think it would be better not showing the Edit project menu when the user cannot actually edit the project. Similar to what is done for "Edit permissions" action which is available for project admin only

dgutride:
Migrated to:  https://issues.redhat.com/browse/RHOAIENG-1163

PR #3003, opened 2024-07-15
Defer project kebab check and disable

emilys314:
Closes [RHOAIENG-1163](https://issues.redhat.com/browse/RHOAIENG-1163) and https://github.com/opendatahub-io/odh-dashboard/issues/1932

## Description

When a user doesn't have permission to Edit Project, Delete Project, or Edit Permissions, the actions will be grayed out and disabled. This prevents a user from trying to do the action and later getting an insufficient permissions error. (We decided to not hide them, because then we would have to hide the entire kebab menu if there were no permissions, which requires a check on display of the table.)

This also defers the project permission checks until the menu will be opened. This saves us from having to make 1 (or more) api calls per project being displayed on the table on page load. 

## How Has This Been Tested?

Tested locally with varying user permissions. Checked that projects with sufficient permissions allow actions, and without required permissions, the options will be disabled and the tooltip will show.

Added a test to check the insufficient permissions behaviour.

## Test Impact

Mocked access review calls should be done if actions in the kebab menu are going to be used, since they are gated now. 

## Request review criteria:

If you have UI changes: 

`aria-disabled` being used to add tooltip for a disabled item:

image

Throttled testing to show permissions disabled while network call is still loading:

https://github.com/user-attachments/assets/98a36897-830a-47bb-86d2-95acf0a07eda

### DIFF
--- a/frontend/src/pages/projects/screens/projects/useProjectTableRowItems.ts
+++ b/frontend/src/pages/projects/screens/projects/useProjectTableRowItems.ts
@@ -1,12 +1,15 @@
+import * as React from 'react';
 import { useNavigate } from 'react-router-dom';
+import { TooltipProps } from '@patternfly/react-core';
 import { useAccessReview } from '~/api';
 import { AccessReviewResourceAttributes, ProjectKind } from '~/k8sTypes';
 
 type KebabItem = {
   title?: string;
-  isDisabled?: boolean;
+  isAriaDisabled?: boolean;
   isSeparator?: boolean;
   onClick?: () => void;
+  tooltipProps?: TooltipProps;
 };
 const accessReviewResource: AccessReviewResourceAttributes = {
   group: 'rbac.authorization.k8s.io',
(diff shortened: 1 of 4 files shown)


frontend/src/pages/projects/screens/projects/useProjectTableRowItems.ts
```diff
@@ -18,40 +21,76 @@ const useProjectTableRowItems = (
...
+  const [allowDelete, allowDeleteLoaded] = useAccessReview(
+    {
+      ...accessReviewResource,
+      namespace: project.metadata.name,
+      verb: 'delete',
+    },
+    shouldRunCheck,
+  );
+
+  const runAccesCheck = () => {
```
andrewballantyne, line 52: When dealing with custom hooks, we need to make sure that any functions that are passed out of the hook are referentially stable as it's difficult for it to be done later. Data is less important because you can just compare data to past data and make it referentially stable between renders (we have our utility, `useDeepCompareMemoize`).

So this is to say, turn it into a useCallback.
```suggestion
  const runAccesCheck = React.useCallback(() => {
```

frontend/src/pages/projects/screens/projects/useProjectTableRowItems.ts
```diff
@@ -18,40 +21,76 @@ const useProjectTableRowItems = (
...
     {
       title: 'Edit project',
-      isDisabled: isRefreshing,
+      isAriaDisabled: isRefreshing || !allowUpdate || !allowUpdateLoaded,
       onClick: () => {
         setEditData(project);
       },
+      ...(!allowUpdate &&
+        allowUpdateLoaded && {
+          tooltipProps: { content: 'You do not have permissions to perform this action' },
```
andrewballantyne, line 65: Might want to make this a static block as a variable, since it's just duplicated between each option.

andrewballantyne (changes requested):
Great, thanks for working on this. Not the easiest and clearest first ticket, but we got there 🙂 

Couple comments about implementation, let me know if you have any questions.

frontend/src/pages/projects/screens/projects/useProjectTableRowItems.ts
```diff
@@ -18,40 +21,72 @@ const useProjectTableRowItems = (
...
+      verb: 'delete',
+    },
+    shouldRunCheck,
+  );
+
+  const runAccesCheck = React.useCallback(() => {
+    setShouldRunCheck(true);
+  }, []);
+
+  const noPermissionToolTip = (allow: boolean, loaded: boolean) =>
```
andrewballantyne, line 56: Technically a nit because our linter should be doing this if we wanted it...

But typically you always want to type functions for both input and return -- in this case, it's a little less important, but I think it's a good practice to get into.

If you create this function in the root, I believe our linter yells at you to add a return type.

```suggestion
  const noPermissionToolTip = (allow: boolean, loaded: boolean): Partial<KebabItem> =>
```

emilys314: Sure thing, I just added the `Partial<KebabItem> | undefined` return type onto it. Oddly the linter doesn't complain if the function is at the root either.

andrewballantyne: Interesting... I swore we had that linter. cc @christianvogt ?

emilys314: Looks like the eslint rule is disabled for that 
in `frontend/.eslintrc` on line 156 
```
    "@typescript-eslint/explicit-function-return-type": "off",
```

christianvogt: We do have this rule enabled: `@typescript-eslint/explicit-module-boundary-types`
This forces a type for all exported properties. Anything else internal to the module can use inference.

PR #3008, opened 2024-07-16
[RHOAIENG-4702] Improve unique naming for runs and versions in pipelines

jpuzz0:
Closes: [RHOAIENG-4702](https://issues.redhat.com/browse/RHOAIENG-4702)

## Description
- Added duplicate name info helper text when uploading a pipeline, version or creating a run or schedule.
- Fixed issue where if > 20 versions existed within a particular pipeline, only the latest 20 were being used in the list of versions used on every runs table - this would result in pipeline versions not rendered in the table - instead just a hyphen, and consequently effected the list of pipeline versions available for the table filter toolbar dropdown.
- Fixed duplicate running of ~30 tests that was leading to an unnecessary additional ~2.5 minutes of CI cypress test runtime.

image

## How Has This Been Tested?
Manual testing, and updated existing tests to verify the existence of the duplicate name helper text.

## Test Impact
Creation of runs, schedules, and importing of pipelines and versions.

## How to test
1. When creating a new run, specify a "Name" of a run that already exists within the project you're in, verify Info helper text exists, with the ability to still create the run, since this is not an API limitation for runs or recurring runs.
2. Perform step 1, but instead use the "Create schedule" page/form.
3. Open the import pipeline modal from the pipelines list page, specify a "Name" of a pipeline that already exists within the project you're in, verify helper text exists.
4. Perform step 3 except from the "Create run" page (accessed from below the Pipeline form dropdown).
5. Open the import pipeline version modal from the pipelines list page, specify a "Version name" of a pipeline version that already exists within the project you're in and the pipeline specified above it, verify helper text exists.
6. Perform step 5 except from the "Create run" page (accessed from below the Pipeline version form dropdown).

If you have UI changes: 

### DIFF
--- a/frontend/src/concepts/k8s/NameDescriptionField.tsx
+++ b/frontend/src/concepts/k8s/NameDescriptionField.tsx
@@ -23,6 +23,8 @@ type NameDescriptionFieldProps = {
   showK8sName?: boolean;
   disableK8sName?: boolean;
   maxLength?: number;
+  nameHelperText?: React.ReactNode;
+  onNameChange?: (value: string) => void;
 };
 
 const NameDescriptionField: React.FC<NameDescriptionFieldProps> = ({
@@ -34,6 +36,8 @@ const NameDescriptionField: React.FC<NameDescriptionFieldProps> = ({
   showK8sName,
   disableK8sName,
   maxLength,
+  nameHelperText,
+  onNameChange,
 }) => {
   const autoSelectNameRef = React.useRef<HTMLInputElement | null>(null);
 
@@ -62,14 +66,20 @@ const NameDescriptionField: React.FC<NameDescriptionFieldProps> = ({
             data-testid={nameFieldId}
             name={nameFieldId}
             value={data.name}
-            onChange={(e, name) => setData({ ...data, name })}
+            onChange={(_e, value) => {
+              setData({ ...data, name: value });
+              onNameChange?.(value);
+            }}
             maxLength={maxLength}
           />
+
           {maxLength && (
             <HelperText>
               <HelperTextItem>{`Cannot exceed ${maxLength} characters`}</HelperTextItem>
             </HelperText>
           )}
+
+          {nameHelperText}
         </FormGroup>
       </StackItem>
       {showK8sName && (
--- a/frontend/src/concepts/pipelines/content/createRun/RunForm.tsx
+++ b/frontend/src/concepts/pipelines/content/createRun/RunForm.tsx
@@ -5,10 +5,19 @@ import { RunFormData, RunTypeOption } from '~/concepts/pipelines/content/createR
 import { ValueOf } from '~/typeHelpers';
 import { ParamsSection } from '~/concepts/pipelines/content/createRun/contentSections/ParamsSection';
 import RunTypeSectionScheduled from '~/concepts/pipelines/content/createRun/contentSections/RunTypeSectionScheduled';
-import { PipelineVersionKFv2, RuntimeConfigParameters } from '~/concepts/pipelines/kfTypes';
+import {
+  PipelineRecurringRunKFv2,
+  PipelineRunKFv2,
+  PipelineVersionKFv2,
+  RuntimeConfigParameters,
+} from '~/concepts/pipelines/kfTypes';
 import ProjectAndExperimentSection from '~/concepts/pipelines/content/createRun/contentSections/ProjectAndExperimentSection';
 import { getDisplayNameFromK8sResource } from '~/concepts/k8s/utils';
 import { useLatestPipelineVersion } from '~/concepts/pipelines/apiHooks/useLatestPipelineVersion';
+import { getNameEqualsFilter } from '~/concepts/pipelines/utils';
+import { DuplicateNameHelperText } from '~/concepts/pipelines/content/DuplicateNameHelperText';
+import { usePipelinesAPI } from '~/concepts/pipelines/context';
+import useDebounceCallback from '~/utilities/useDebounceCallback';
 import PipelineSection from './contentSections/PipelineSection';
 import { RunTypeSection } from './contentSections/RunTypeSection';
 import { CreateRunPageSections, RUN_NAME_CHARACTER_LIMIT, runPageSectionTitles } from './const';
@@ -21,12 +30,42 @@ type RunFormProps = {
 };
 
 const RunForm: React.FC<RunFormProps> = ({ data, onValueChange, isCloned }) => {
+  const { api } = usePipelinesAPI();
   const [latestVersion] = useLatestPipelineVersion(data.pipeline?.pipeline_id);
   // Use this state to avoid the pipeline version being set as the latest version at the initial load
   const [initialLoadedState, setInitialLoadedState] = React.useState(true);
   const selectedVersion = data.version || latestVersion;
   const paramsRef = React.useRef(data.params);
   const isSchedule = data.runType.type === RunTypeOption.SCHEDULED;
+  const { name } = data.nameDesc;
+  const [hasDuplicateName, setHasDuplicateName] = React.useState(false);
+
+  const checkForDuplicateName = useDebounceCallback(
+    React.useCallback(
+      async (value: string) => {
+        if (value) {
+          let duplicateRuns: PipelineRunKFv2[] | PipelineRecurringRunKFv2[] | undefined = [];
+
+          if (isSchedule) {
+            const { recurringRuns } = await api.listPipelineRecurringRuns(
+              {},
+              getNameEqualsFilter(value),
+            );
+            duplicateRuns = recurringRuns;
+          } else {
+            const { runs } = await api.listPipelineActiveRuns({}, getNameEqualsFilter(value));
+            duplicateRuns = runs;
+          }
+
+          if (duplicateRuns?.length) {
+            setHasDuplicateName(true);
+          }
+        }
+      },
+      [api, isSchedule],
+    ),
+    500,
+  );
 
   const updateInputParams = React.useCallback(
     (version: PipelineVersionKFv2 | undefined) =>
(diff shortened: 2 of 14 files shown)


frontend/src/concepts/pipelines/content/createRun/utils.ts
```diff
@@ -45,6 +46,7 @@ export const isFilledRunFormData = (formData: RunFormData): formData is SafeRunF
 
   return (
     !!formData.nameDesc.name &&
+    formData.nameDesc.name.length <= RUN_NAME_CHARACTER_LIMIT &&
```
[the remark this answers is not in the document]
jpuzz0: Yes, we still allow users to create runs with duplicate names. That's what the text the JIRA screenshot had when the icon is of type "info", so that's a question for Yan and/or Katie.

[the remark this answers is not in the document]
DaoDaoNoCode: @yannnz @kaedward ^ What do you think? We can also merge this PR and create a ticket for this later.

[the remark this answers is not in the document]
yannnz: I think it makes sense to info the user about the duplicated name. IMO, if I am the user and uses the same names for different runs, it will confuse me later to find the correct run I am looking for. I will try to avoid using duplicated name as a user for better filtering and management. I don't understand why technically the Platform allows so, but if it is how it works nowadays, I would think it is good to use info icon to warn the user.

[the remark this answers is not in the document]
jpuzz0: Thanks @yannnz. Unless @kaedward has a suggestion for different wording here, I think we roll with this for now at least.

frontend/src/concepts/pipelines/content/import/PipelineImportModal.tsx
```diff
@@ -121,8 +126,11 @@ const PipelineImportModal: React.FC<PipelineImportModalProps> = ({ isOpen, onClo
                 data-testid="pipeline-name"
                 name="pipeline-name"
                 value={name}
-                onChange={(e, value) => setData('name', value)}
+                onChange={(_e, value) => setData('name', value)}
+                validated={hasDuplicateNameError ? 'error' : 'default'}
               />
+
+              {hasDuplicateNameError && <DuplicateNameHelperText name={name} isError />}
```
Gkrumbach07, line 133: why are we showing an error state here? i thought it was just info. we are not blocking the user from doing this just informing them

frontend/src/concepts/pipelines/content/import/PipelineImportModal.tsx
```diff
@@ -28,11 +30,14 @@ const PipelineImportModal: React.FC<PipelineImportModalProps> = ({ isOpen, onClo
   const [error, setError] = React.useState<Error | undefined>();
   const [{ name, description, fileContents, pipelineUrl, uploadOption }, setData, resetData] =
     usePipelineImportModalData();
+  const [{ items: pipelines }] = useAllPipelines();
```
Gkrumbach07, line 33: do we need to fetch all the pipelines for this. can we get away with just doing a regular fetch pipelines call with a filter?

frontend/src/concepts/pipelines/content/import/PipelineVersionImportModal.tsx
```diff
@@ -35,6 +37,8 @@ const PipelineVersionImportModal: React.FC<PipelineVersionImportModalProps> = ({
     setData,
     resetData,
   ] = usePipelineVersionImportModalData(existingPipeline);
+  const [{ items: versions }] = useAllVersionsByPipelineId(pipeline?.pipeline_id ?? '');
```
Gkrumbach07, line 40: same thing here, can we just add a filter for the name instead of fetching all versions

frontend/src/concepts/pipelines/content/createRun/RunForm.tsx
```diff
@@ -27,6 +29,8 @@ const RunForm: React.FC<RunFormProps> = ({ data, onValueChange, isCloned }) => {
   const selectedVersion = data.version || latestVersion;
   const paramsRef = React.useRef(data.params);
   const isSchedule = data.runType.type === RunTypeOption.SCHEDULED;
+  const [{ items: activeRuns }] = useAllPipelineActiveRuns();
+  const [{ items: recurringRuns }] = useAllPipelineRecurringRuns();
```
Gkrumbach07, line 34: we should avoid doing this recursive fetching.

Gkrumbach07:
Added comments. Biggest thing is that we dont want to be doing a recursive fetching resources. This is the same reason why we dont have the feature for "delete a pipeline and it deletes all the versions too"

frontend/src/concepts/pipelines/content/import/PipelineImportModal.tsx
jpuzz0: The user is unable to create duplicate pipeline names within the same project. That results in an API error, so no, for this modal, and the version modal, it is considered an error, and we're catching it before they experience that API error.

frontend/src/concepts/pipelines/content/import/PipelineImportModal.tsx
jpuzz0: I was thinking we weren't able to filter by name and only ID, but you are correct and I have updated this. The recursive fetching is required for the versions displayed in the table though. I've scrapped the others.

frontend/src/concepts/pipelines/content/import/PipelineImportModal.tsx
Gkrumbach07: do we block the submit?

jpuzz0: With this PR, if there is a duplicate name identified, we are preventing the ability to submit the modal. If the user were to click submit without a duplicate name identified, an API error would show in the modal and the user would be forced to enter a new name regardless.

Gkrumbach07: to keep with consistency across our forms, we should just use the info version and not block them from submitting and let the api throw an error. I believe will eventually start to show errors on fields but not until we adopt a form lib to do it.

frontend/src/concepts/k8s/NameDescriptionField.tsx
```diff
@@ -61,15 +69,24 @@ const NameDescriptionField: React.FC<NameDescriptionFieldProps> = ({
...
+            }}
+            onBlur={onNameBlur}
+            validated={hasMaxLengthError ? 'error' : 'default'}
           />
-          {maxLength && (
+
+          {hasMaxLengthError && (
             <HelperText>
-              <HelperTextItem>{`Cannot exceed ${maxLength} characters`}</HelperTextItem>
+              <HelperTextItem variant="error" hasIcon>
```
Gkrumbach07, line 83: keep this info for the same reason

jpuzz0: Design wanted to avoid having stacked helper text, which is why this was converted to an error. cc @yannnz

frontend/src/concepts/pipelines/content/createRun/RunForm.tsx
```diff
@@ -75,6 +107,9 @@ const RunForm: React.FC<RunFormProps> = ({ data, onValueChange, isCloned }) => {
           data={data.nameDesc}
           setData={(nameDesc) => onValueChange('nameDesc', nameDesc)}
           maxLength={RUN_NAME_CHARACTER_LIMIT}
+          onNameBlur={onNameBlur}
+          onNameChange={() => setHasDuplicateName(false)}
+          {...(hasDuplicateName && { nameHelperText: <DuplicateNameHelperText name={name} /> })}
```
Gkrumbach07, line 112: do you need toi use the spread operator here? can you just

```tsx
nameHelperText: hasDuplicateName  ? <DuplicateNameHelperText name={name} /> : undefined

Comment (jpuzz0) [reply frontend/src/concepts/pipelines/content/createRun/RunForm.tsx:112]: What you suggested also works, but why would you prefer it?

File: frontend/src/concepts/pipelines/content/createRun/RunForm.tsx [hunk 8]
```diff
@@ -75,6 +107,9 @@ const RunForm: React.FC<RunFormProps> = ({ data, onValueChange, isCloned }) => {
           data={data.nameDesc}
           setData={(nameDesc) => onValueChange('nameDesc', nameDesc)}
           maxLength={RUN_NAME_CHARACTER_LIMIT}
+          onNameBlur={onNameBlur}
```
Gkrumbach07, line 110: can you use lodash debounce instead. a user wouldn't see the helper text unless if they never click out

Gkrumbach07: or scroll down and click out

jpuzz0: A debounce of how many seconds would you like?

frontend/src/concepts/pipelines/content/createRun/RunForm.tsx
Gkrumbach07: just code clarity. i would see it used if we were spreading more than one prop conditionally

frontend/src/concepts/pipelines/content/createRun/RunForm.tsx
Gkrumbach07: the `useDebounceCallback` hook uses 250ms

frontend/src/concepts/pipelines/content/import/PipelineImportModal.tsx
Gkrumbach07: thats true, but i believe consistency efforts are covered by someone else in UXD. So im not sure if everything is always in sync

PR #3010, opened 2024-07-16
disable access to notebook from other users

rsun19:
Closes: [RHOAIENG-536](https://issues.redhat.com/browse/RHOAIENG-536)

## Description

Previously, a user without 'get' permissions for notebooks could access the API url for notebook events. Now, they are unable to, and they would be greeted with a `403` status code.

## How Has This Been Tested?

Steps to test:
1. Create a new notebook server (`Applications` -> `Enabled` ->  `Jupyter`)
2. That itself should be an event. If time has elapsed since then, just open up the notebook server, make a file, and type in `print("hello world")` or anything else that would constitute an event.
3. Create a user with no permissions
4. Change the URL in `dev_impersonate` to:
```
 const url = `https://oauth.${namedHost}/oauth/authorize?response_type=token&client_id=openshift-challenging-client`; 
```
We are changing the URL because currently it doesn't work in ROSA clusters. This issue has already been [raised](https://issues.redhat.com/browse/RHOAIENG-9996)
5. Follow the steps [ here ](https://github.com/opendatahub-io/odh-dashboard/blob/279d95daa575379ca4fb27e86877ffd6a9a9edb7/docs/SDK.md#pass-through-impersonate-user-dev-mode) to impersonate said user with no permissions.
6. Create a role for that user which gives it access to view notebooks. Then, make a rolebinding to that user for that role.
```
rules:
  - verbs:
      - get
    apiGroups:
      - kubeflow.org
    resources:
      - notebooks
    resourceName:
      - <notebook name>
```
7. Access through this link  `/api/nb-events/:namespace/:notebookName/:podUID`
8. Get the notebookName and podUID from the terminal with this command:

```
oc get pods -n opendatahub -o custom-columns=notebookName:.metadata.labels.notebook-name,podUID:.metadata.uid
```
9. This should work. However, when you go back to the role you just created previously and remove the 'get' verb OR set the <notebook name> to literally anything else, you should be greeted with a 403 status code.

## Test Impact

SSAR is hard to test in test suites.

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/backend/src/routes/api/namespaces/namespaceUtils.ts
+++ b/backend/src/routes/api/namespaces/namespaceUtils.ts
@@ -8,7 +8,7 @@ import { K8sStatus, KubeFastifyInstance, OauthFastifyRequest } from '../../../ty
 import { createCustomError } from '../../../utils/requestUtils';
 import { isK8sStatus, passThroughResource } from '../k8s/pass-through';
 
-const createSelfSubjectAccessReview = (
+export const createSelfSubjectAccessReview = (
   fastify: KubeFastifyInstance,
   request: OauthFastifyRequest,
   resourceAttributes: V1ResourceAttributes,
--- a/backend/src/routes/api/nb-events/eventUtils.ts
+++ b/backend/src/routes/api/nb-events/eventUtils.ts
@@ -1,5 +1,7 @@
-import { V1Event, V1EventList } from '@kubernetes/client-node';
-import { KubeFastifyInstance } from '../../../types';
+import { V1Event, V1SelfSubjectAccessReview } from '@kubernetes/client-node';
+import { K8sStatus, KubeFastifyInstance, OauthFastifyRequest } from '../../../types';
+import { createSelfSubjectAccessReview } from '../namespaces/namespaceUtils';
+import { createCustomError } from '../../../utils/requestUtils';
 
 export const getNotebookEvents = async (
   fastify: KubeFastifyInstance,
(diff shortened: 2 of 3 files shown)


rsun19:
CC @christianvogt

christianvogt (commented):
Since `#getNotebookEvents` doesn't use `notebookName` if `podUID` is supplied, I can supply any notebook name I have access to but still provide the pod UID for a notebook I do not have access to.

There needs to be some link between the events and the notebook name.

In `#getNotebookEvents`. Can we make use of the podUID and the notebook name at the same time?
```
      podUID
        ? `involvedObject.kind=Pod,involvedObject.uid=${podUID}`
        : `involvedObject.kind=StatefulSet,involvedObject.name=${notebookName}`,
```
cc @andrewballantyne

rsun19:
> Since `#getNotebookEvents` doesn't use `notebookName` if `podUID` is supplied, I can supply any notebook name I have access to but still provide the pod UID for a notebook I do not have access to.
> 
> There needs to be some link between the events and the notebook name.
> 
> In `#getNotebookEvents`. Can we make use of the podUID and the notebook name at the same time?
> 
> ```
>       podUID
>         ? `involvedObject.kind=Pod,involvedObject.uid=${podUID}`
>         : `involvedObject.kind=StatefulSet,involvedObject.name=${notebookName}`,
> ```
> 
> cc @andrewballantyne

I have a suggested solution in the commit; it uses `listNamespacedPod`

rsun19:
Also, note that in the Self subject access review, the notebooks rule with no resource name such as:
```
rules:
  - verbs:
      - get
    apiGroups:
      - kubeflow.org
    resources:
      - notebooks
```
allows a user to get all notebook events

is this the intended behavior?

christianvogt (commented):
Tested with a cluster admin and a regular user. 
This correctly restricted access for the regular user trying to get events for the admin notebook pod. While the cluster admin could get the events for the regular user's notebook pod.

rsun19:
Made requested changes

christianvogt:
/lgtm
/approve

Used `curl` for testing locally. Note that I had to disable the `DEV_MODE` condition in `directCallUtils.ts` for local testing in this way:
```
curl -H "x-forwarded-access-token: ${TOKEN}" "https://localhost:4000/api/nb-events/opendatahub/jupyter-nb-clusteradmin/<podUID>"
curl -H "x-forwarded-access-token: ${TOKEN}" "https://localhost:4000/api/nb-events/opendatahub/jupyter-nb-clusteradmin"
curl -H "x-forwarded-access-token: ${TOKEN}" "https://localhost:4000/api/nb-events/opendatahub/jupyter-nb-regularuser/<podUID>"
curl -H "x-forwarded-access-token: ${TOKEN}" "https://localhost:4000/api/nb-events/opendatahub/jupyter-nb-regularuser"
```

The admin should have access to all while the regular user should only have access to their own.

Also do some ad hoc testing by changing the namespace, notebook name and podUID to varying values of real and inaccessible.

PR #3012, opened 2024-07-17
[RHOAIENG-9754] Use npm-Version of the Segment client

pilhuhn:
Closes https://issues.redhat.com/browse/RHOAIENG-9754

## Description
Use the npm-version of the Segment client (Analytics(-next).js) and don't load it on the fly 
on the first load of the UI. 
This change has no impact of the visible appearance and functionality of the UI.

## How Has This Been Tested?
Manual testing by looking at network tab of the browser tools in non-dev mode

## Test Impact
See previous item

## Request review criteria:

Self checklist (all need to be checked):
- [X] The developer has manually tested the changes and verified that the changes work
- [ ] Commits have been squashed into descriptive, self-contained units of work (e.g. 'WIP' and 'Implements feedback' style messages have been removed)
- [ ] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [X] The developer has added tests or explained why testing cannot be added (unit or cypress tests for related changes)

After the PR is posted & before it merges:
- [ ] The developer has tested their solution on a cluster by using the image produced by the PR to `main`

This currently builds on the code base of #2995  , but I can also rebase that against main 
cc @christianvogt @andrewballantyne

PR #3014, opened 2024-07-17
[RHOAIENG-9754] Use the npm-Version of the Segment libraries.

pilhuhn:
Closes https://issues.redhat.com/browse/RHOAIENG-9754

## Description
Use the npm-version of the Segment client (Analytics(-next).js) and don't load it on the fly 
on the first load of the UI. 
This change has no impact of the visible appearance and functionality of the UI.

## How Has This Been Tested?
Manual testing by looking at network tab of the browser tools in non-dev mode

## Test Impact
See previous item

## Request review criteria:

cc @christianvogt @andrewballantyne 
Superseeds #3012

### DIFF
--- a/frontend/src/concepts/analyticsTracking/initSegment.tsx
+++ b/frontend/src/concepts/analyticsTracking/initSegment.tsx
@@ -1,74 +1,26 @@
+import { AnalyticsBrowser } from '@segment/analytics-next';
+
 export const initSegment = async (props: {
   segmentKey: string;
   enabled: boolean;
 }): Promise<void> => {
   const { segmentKey, enabled } = props;
-  const analytics = (window.analytics = window.analytics || []);
-  if (analytics.initialize) {
+  if (!enabled || !segmentKey) {
     return;
   }
-  if (analytics.invoked) {
-    /* eslint-disable-next-line no-console */
-    console.error('Segment snippet included twice.');
-  } else {
-    analytics.invoked = true;
-    analytics.methods = [
-      'trackSubmit',
-      'trackClick',
-      'trackLink',
-      'trackForm',
-      'pageview',
-      'identify',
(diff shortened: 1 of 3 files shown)


andrewballantyne (commented):
Much cleaner. I wonder how we find out if this is supported by the CPaaS build system. Started a [slack thread here](https://redhat-internal.slack.com/archives/C05NXTEHLGY/p1721232422365439).

christianvogt:
@andrewballantyne are we gtg with this yet?

andrewballantyne:
> @andrewballantyne are we gtg with this yet?

I was having issues getting a key to test this -- seems we have that as an option now. I have not yet tested this and I'm not sure I'll have today to do so. Maybe tomorrow.

@christianvogt feel free to test this and remove the hold whenever you see fit.

frontend/src/concepts/analyticsTracking/initSegment.tsx
```diff
@@ -1,74 +1,27 @@
...
+
 export const initSegment = async (props: {
   segmentKey: string;
   enabled: boolean;
 }): Promise<void> => {
   const { segmentKey, enabled } = props;
-  const analytics = (window.analytics = window.analytics || []);
-  if (analytics.initialize) {
+  if (!enabled) {
+    window.analytics = window.analytics || [];
```
christianvogt, line 9: This causes an exception to be thrown when a segment key is provided but `disableTracking=true`. `window.analytics = []` causes the exception in `segmentIOUtils`.
```
TypeError: window.analytics.identify is not a function
    at fireTrackingEvent (segmentIOUtils.tsx:48:26)
```

pilhuhn: Good catch

PR #3019, opened 2024-07-19
feat(mr): add display name and description for mr selector

gitdallas:
closes: https://issues.redhat.com/browse/RHOAIENG-8453
closes: https://issues.redhat.com/browse/RHOAIENG-10025

## Description
add display name and description (truncated with tooltip) in the selector, add a "view description" button link off to the side with a popver containing the name/description.
image

image

## How Has This Been Tested?
tested with MRs that have display name different from resource name. tested with MR that contains a long description, short description, no description.

## Test Impact
just adding some info, tests all still pass

## Request review criteria:
test selector looks correct with various display name, resource name, and description.

If you have UI changes: 

### DIFF
--- a/frontend/src/pages/modelRegistry/screens/ModelRegistrySelector.tsx
+++ b/frontend/src/pages/modelRegistry/screens/ModelRegistrySelector.tsx
@@ -5,11 +5,29 @@ import {
   SelectGroup,
   SelectOption,
 } from '@patternfly/react-core/deprecated';
-import { Bullseye, Flex, FlexItem } from '@patternfly/react-core';
+import {
+  Bullseye,
+  Button,
+  DescriptionList,
+  DescriptionListDescription,
+  DescriptionListGroup,
+  DescriptionListTerm,
+  Flex,
+  FlexItem,
+  Popover,
+  Tooltip,
+} from '@patternfly/react-core';
+import truncateStyles from '@patternfly/react-styles/css/components/Truncate/truncate';
+import { InfoCircleIcon } from '@patternfly/react-icons';
 import { useBrowserStorage } from '~/components/browserStorage';
 import { ModelRegistrySelectorContext } from '~/concepts/modelRegistry/context/ModelRegistrySelectorContext';
 import { ProjectObjectType, typedObjectImage } from '~/concepts/design/utils';
-import { getDisplayNameFromK8sResource } from '~/concepts/k8s/utils';
+import {
+  getDescriptionFromK8sResource,
+  getDisplayNameFromK8sResource,
+  getResourceNameFromK8sResource,
+} from '~/concepts/k8s/utils';
+import { ModelRegistryKind } from '~/k8sTypes';
 
 const MODEL_REGISTRY_FAVORITE_STORAGE_KEY = 'odh.dashboard.model.registry.favorite';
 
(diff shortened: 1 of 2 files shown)


frontend/src/pages/modelRegistry/screens/ModelRegistrySelector.tsx
```diff
@@ -44,9 +49,12 @@ const ModelRegistrySelector: React.FC<ModelRegistrySelectorProps> = ({
         <SelectOption
           id={mr.metadata.name}
           key={mr.metadata.name}
-          value={getDisplayNameFromK8sResource(mr)}
+          value={getResourceNameFromK8sResource(mr)}
+          description={<Truncate content={getDescriptionFromK8sResource(mr)} />}
```
mturley, line 53: Ok, this is interesting. So there are two things about the design that this doesn't match:
1. The tooltip has a bold header "<registry name> description" above its full content
2. The tooltip appears to be positioned relative to the entire SelectOption, not just the description, and positioned to the right.

In theory we could fix this by not using the `<Truncate>` component and instead using just the CSS styles for truncating the displayed text without adding a tooltip ([like in this HTML example](https://www.patternfly.org/components/truncate/html#default), except with the classNames imported from `@patternfly/react-styles`), and then add our own custom tooltip around the SelectOption. However I messed with that a little bit, and if you wrap SelectOption in Tooltip it causes an error because there is special logic in SelectGroup that expects SelectOption to be a direct child. I tinkered with other ways to position the tooltip where we want it and gave up.

The main problem here is that we're still using the deprecated Select component. If we were using the new PF5 Select, SelectOption would be inheriting props from MenuItem and we'd be able to pass `tooltipProps` to it to get our custom tooltip with the right positioning and the header text.

So I see two options here:
1. Upgrade this whole component to the new Select as part of this PR and use the above solution. @gitdallas I'll let you decide if that's practical to do quickly.
2. Do a half-measure for now (maybe just pass `tooltipPosition="right"` to Truncate so at least that is positioned a bit better when you hover the description text) and either open a followup tech debt issue to improve this later or let it go.

@yih-wang are you ok with option (2) above for MVP? We'd have a tooltip that only appears when you hover the small description text, and it won't have the bold "<registry name> description" header text in it. If so, should we open an issue to make it match the mockup exactly after MVP or is that fine to keep?

mturley (changes requested):
There's an inconsistency with the mockup here that's tricky to address. Some options outlined below, @yih-wang we may need your input.

yih-wang:
@mturley Thanks for summarizing the problems and the solutions. 

I'm fine with positioning the tooltip relative to the description rather than the entire selectOption. It's acceptable if we don't change this after the MVP. However, I think it's crucial to have a header for the tooltip to ensure the context is clear. What we have now (as shown in the screenshot) is good for the MVP. Please open a ticket to track the update for adding the bold header.

yih-wang:
BTW regarding the tooltip header, can we simply steal the codes from the workbench creation? This is what its tooltip looks like: it includes a bold header and all the text is left-aligned, which seems easier to read when the text is long.
image

gitdallas:
@yih-wang - It's not so simple for two reasons. Workbench creation is using the new Select and they are not using Truncate.

gitdallas:
@yih-wang - updated by not actually using the `Truncate` component, then wrapping with a `Tooltip` that allows the customization we need:
image

frontend/src/pages/modelRegistry/screens/ModelRegistrySelector.tsx
```diff
@@ -38,15 +55,42 @@ const ModelRegistrySelector: React.FC<ModelRegistrySelectorProps> = ({
...
+          <DescriptionListTerm>{`${getDisplayNameFromK8sResource(
+            mr,
+          )} description`}</DescriptionListTerm>
+          <DescriptionListDescription>{desc}</DescriptionListDescription>
+        </DescriptionListGroup>
+      </DescriptionList>
+    );
+    return (
+      <Tooltip content={tooltipContent} isContentLeftAligned>
+        <span className="pf-v5-c-truncate">
```
mturley, line 75: So that TypeScript catches any broken CSS class usage when we upgrade PF, we should try to always import classNames from `@patternfly/react-styles` if we are using them directly like this. Can you add this import:

```ts
import truncateStyles from '@patternfly/react-styles/css/components/Truncate/truncate';
```

and then use properties of it here:

```tsx
  return (
      <Tooltip content={tooltipContent} isContentLeftAligned>
        <span className={truncateStyles.truncate}>
          <span className={truncateStyles.truncateStart}>{desc}</span>
        </span>
      </Tooltip>
    );
```

mturley (changes requested):
One more tiny thing... otherwise LGTM

PR #3022, opened 2024-07-19
Remove unused docs & cleanup .git folder

andrewballantyne:
Cleanup a lot of misc files.

* Removed our unused docs around process; since we moved our issues to Jira & our process to Confluence -- these are outdated now
* SMEs for PR templates is now gone, these were outdated
* Removed unused .git items
    * Issue templates are reduced back to `bug` & `feature request` in case someone using ODH wishes to provide feedback
    * Removed some of the workflows we had because those are documented internal now


manifests/core-bases/base/deployment.yaml
andrewballantyne, line 1: Fixed indentation of this file with this for ease of use for anyone with VS Code -- it prettier'ed the file when editing.

PR #3024, opened 2024-07-22
RHOAIENG-9232 Create useTrackUser

pilhuhn:
Closes https://issues.redhat.com/browse/RHOAIENG-9232

## Description
Put obtaining user properties into its own hook. This is meant to be extended in the future.

## How Has This Been Tested?
Manual testing by developer by inspecting browser console and (for non-dev mode) network tab.

## Request review criteria:

cc @andrewballantyne

### DIFF
--- a/frontend/src/concepts/analyticsTracking/trackingProperties.ts
+++ b/frontend/src/concepts/analyticsTracking/trackingProperties.ts
@@ -3,7 +3,10 @@ export type ODHSegmentKey = {
 };
 
 export type IdentifyEventProperties = {
+  isAdmin: boolean;
   anonymousID?: string;
+  userId?: string;
+  canCreateProjects: boolean;
 };
 
 export const enum TrackingOutcome {
--- a/frontend/src/concepts/analyticsTracking/useSegmentTracking.ts
+++ b/frontend/src/concepts/analyticsTracking/useSegmentTracking.ts
@@ -2,6 +2,7 @@ import React from 'react';
 import { useAppContext } from '~/app/AppContext';
 import { useAppSelector } from '~/redux/hooks';
 import { fireIdentifyEvent, firePageEvent } from '~/concepts/analyticsTracking/segmentIOUtils';
+import { useTrackUser } from '~/concepts/analyticsTracking/useTrackUser';
 import { useWatchSegmentKey } from './useWatchSegmentKey';
 import { initSegment } from './initSegment';
 
@@ -10,28 +11,27 @@ export const useSegmentTracking = (): void => {
   const { dashboardConfig } = useAppContext();
   const username = useAppSelector((state) => state.user);
   const clusterID = useAppSelector((state) => state.clusterID);
+  const [userProps, uPropsLoaded] = useTrackUser(username);
 
   React.useEffect(() => {
-    if (segmentKey && loaded && !loadError && username && clusterID) {
-      const computeUserId = async () => {
-        const anonymousIDBuffer = await crypto.subtle.digest(
-          'SHA-1',
-          new TextEncoder().encode(username),
-        );
-        const anonymousIDArray = Array.from(new Uint8Array(anonymousIDBuffer));
-        return anonymousIDArray.map((b) => b.toString(16).padStart(2, '0')).join('');
-      };
-
+    if (segmentKey && loaded && !loadError && username && clusterID && uPropsLoaded) {
       window.clusterID = clusterID;
       initSegment({
         segmentKey,
         enabled: !dashboardConfig.spec.dashboardConfig.disableTracking,
       }).then(() => {
-        computeUserId().then((userId) => {
-          fireIdentifyEvent({ anonymousID: userId });
-          firePageEvent();
-        });
+        fireIdentifyEvent(userProps);
+        firePageEvent();
       });
     }
-  }, [clusterID, loadError, loaded, segmentKey, username, dashboardConfig]);
+  }, [
+    clusterID,
+    loadError,
+    loaded,
+    segmentKey,
+    username,
+    dashboardConfig,
+    userProps,
+    uPropsLoaded,
+  ]);
 };
--- a/frontend/src/concepts/analyticsTracking/useTrackUser.ts
+++ b/frontend/src/concepts/analyticsTracking/useTrackUser.ts
@@ -1,0 +1,50 @@
+import React from 'react';
+import { useUser } from '~/redux/selectors';
+import { useAccessReview } from '~/api';
+import { AccessReviewResourceAttributes } from '~/k8sTypes';
+import { IdentifyEventProperties } from '~/concepts/analyticsTracking/trackingProperties';
+
+export const useTrackUser = (username?: string): [IdentifyEventProperties, boolean] => {
+  const { isAdmin } = useUser();
+  const [anonymousId, setAnonymousId] = React.useState<string | undefined>(undefined);
+
+  const [loaded, setLoaded] = React.useState(false);
+  const createReviewResource: AccessReviewResourceAttributes = {
+    group: 'project.openshift.io',
+    resource: 'projectrequests',
+    verb: 'create',
+  };
+  const [allowCreate, acLoaded] = useAccessReview(createReviewResource);
+
+  React.useEffect(() => {
+    const computeAnonymousUserId = async () => {
+      const anonymousIDBuffer = await crypto.subtle.digest(
+        'SHA-1',
+        new TextEncoder().encode(username),
+      );
+      const anonymousIDArray = Array.from(new Uint8Array(anonymousIDBuffer));
+      const aId = anonymousIDArray.map((b) => b.toString(16).padStart(2, '0')).join('');
+      return aId;
+    };
+
+    if (!anonymousId) {
+      computeAnonymousUserId().then((val) => {
+        setAnonymousId(val);
+      });
+    }
+    if (acLoaded && anonymousId) {
+      setLoaded(true);
+    }
+  }, [username, anonymousId, acLoaded]);
+
+  const props: IdentifyEventProperties = React.useMemo(
+    () => ({
+      isAdmin,
+      canCreateProjects: allowCreate,
+      anonymousID: anonymousId,
+    }),
+    [isAdmin, allowCreate, anonymousId],
+  );
+
+  return [props, loaded];
+};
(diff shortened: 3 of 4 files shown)


frontend/src/concepts/analyticsTracking/useTrackUser.ts
```diff
@@ -0,0 +1,39 @@
...
+      new TextEncoder().encode(username),
+    );
+    const anonymousIDArray = Array.from(new Uint8Array(anonymousIDBuffer));
+    const aId = anonymousIDArray.map((b) => b.toString(16).padStart(2, '0')).join('');
+    setAnonymousId(aId);
+  };
+
+  if (!anonymousId) {
+    computeAnonymousUserId();
+  }
```
christianvogt, line 30: This needs to be wrapped in a `useEffect`. Do not execute directly in the render function.
```suggestion
 React.useEffect(() => {
  const computeAnonymousUserId = async () => {
    const anonymousIDBuffer = await crypto.subtle.digest(
      'SHA-1',
      new TextEncoder().encode(username),
    );
    const anonymousIDArray = Array.from(new Uint8Array(anonymousIDBuffer));
    const aId = anonymousIDArray.map((b) => b.toString(16).padStart(2, '0')).join('');
    setAnonymousId(aId);
  };

  if (!anonymousId) {
    computeAnonymousUserId();
  }
  }, [username]);
```

frontend/src/concepts/analyticsTracking/useTrackUser.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  const props: IdentifyEventProperties = {
+    isAdmin,
+    canCreateProjects: allowCreate,
+    anonymousID: anonymousId,
+  };
```
christianvogt, line 36: If you use `props` as a dependency, then it must be a stable reference as much as possible.
```suggestion
  const props: IdentifyEventProperties = React.useMemo(() =>({
    isAdmin,
    canCreateProjects: allowCreate,
    anonymousID: anonymousId,
  }), [isAdmin, allowCreate, anonymousId];
```

frontend/src/concepts/analyticsTracking/useSegmentTracking.ts
```diff
@@ -10,28 +11,18 @@ export const useSegmentTracking = (): void => {
...
-        computeUserId().then((userId) => {
-          fireIdentifyEvent({ anonymousID: userId });
-          firePageEvent();
-        });
+        fireIdentifyEvent(userProps);
+        firePageEvent();
       });
     }
-  }, [clusterID, loadError, loaded, segmentKey, username, dashboardConfig]);
+  }, [clusterID, loadError, loaded, segmentKey, username, dashboardConfig, userProps]);
```
christianvogt, line 27: `userProps` is not a stable reference and will cause this function to be called over and over again on each render.

frontend/src/concepts/analyticsTracking/useSegmentTracking.ts
```diff
@@ -10,28 +11,18 @@ export const useSegmentTracking = (): void => {
...
       window.clusterID = clusterID;
       initSegment({
         segmentKey,
         enabled: !dashboardConfig.spec.dashboardConfig.disableTracking,
       }).then(() => {
-        computeUserId().then((userId) => {
-          fireIdentifyEvent({ anonymousID: userId });
-          firePageEvent();
-        });
+        fireIdentifyEvent(userProps);
```
christianvogt, line 23: It seems possible that `userProps.anonymousId` may not yet be defined because it is computed asynchronously.

pilhuhn:
Thank you @christianvogt your review was super helpful.

frontend/src/concepts/analyticsTracking/useSegmentTracking.ts
christianvogt: This comment still holds true. 
Perhaps you should capture a value to know when segment is initialized, then use a `useEffect` on that value along with `userProps` so that you fire an identify event whenever 'userProps.anonymousID' changes?

The old implementation didn't suffer from this issue.

frontend/src/concepts/analyticsTracking/useTrackUser.ts
```diff
@@ -0,0 +1,44 @@
...
+export const useTrackUser = (username?: string): IdentifyEventProperties => {
+  const { isAdmin } = useUser();
+  const [anonymousId, setAnonymousId] = React.useState<string | undefined>(undefined);
+
+  const createReviewResource: AccessReviewResourceAttributes = {
+    group: 'project.openshift.io',
+    resource: 'projectrequests',
+    verb: 'create',
+  };
+  const [allowCreate] = useAccessReview(createReviewResource);
```
christianvogt, line 16: This makes a network request but the value isn't used in the current implementation.

pilhuhn:
I am afk tomorrow. Will have a look on Thursday

frontend/src/concepts/analyticsTracking/useTrackUser.ts
pilhuhn: It is used in the returned props.

christianvogt: It may be returned but I do not see it being consumed by the current implementation.

pilhuhn: It is used inside the `fireIdentifyEvent() `and passed to segment as user properties.

christianvogt: `fireIdentifyEvent` is only using `properties.anonymousID` but no one uses `canCreateProjects` (which comes from `allowCreate` in this function)

pilhuhn:
I think https://github.com/opendatahub-io/odh-dashboard/pull/3024#discussion_r1688419274 is now solved.

jeff-phillips-18 (approved):

emilys314:
Okay, I enabled the segment stuff on my test cluster, and then I added this PR image to it.

I can confirm the identify event still run on load correctly (with the same parameters)



<details>
<summary> Before vs After screenshots </summary>
the timestamp and session id's are different

before:

Screenshot from 2024-08-09 14-01-25

after:

image


</details>

frontend/src/concepts/analyticsTracking/useTrackUser.ts
```diff
@@ -0,0 +1,50 @@
...
+import { useUser } from '~/redux/selectors';
+import { useAccessReview } from '~/api';
+import { AccessReviewResourceAttributes } from '~/k8sTypes';
+import { IdentifyEventProperties } from '~/concepts/analyticsTracking/trackingProperties';
+
+export const useTrackUser = (username?: string): [IdentifyEventProperties, boolean] => {
+  const { isAdmin } = useUser();
+  const [anonymousId, setAnonymousId] = React.useState<string | undefined>(undefined);
+
+  const [loaded, setLoaded] = React.useState(false);
```
christianvogt, line 11: Unnecessary to track this `loaded` as a separate state. 
Compute it as `!!anonymousId && acLoaded`

frontend/src/concepts/analyticsTracking/useTrackUser.ts
```diff
@@ -0,0 +1,50 @@
...
+  const props: IdentifyEventProperties = React.useMemo(
+    () => ({
+      isAdmin,
+      canCreateProjects: allowCreate,
+      anonymousID: anonymousId,
+    }),
+    [isAdmin, allowCreate, anonymousId],
+  );
+
+  return [props, loaded];
```
christianvogt, line 49:
```suggestion
  return [props, acLoaded && !!anonymousId];
```

frontend/src/concepts/analyticsTracking/useTrackUser.ts
```diff
@@ -0,0 +1,50 @@
...
+
+    if (!anonymousId) {
+      computeAnonymousUserId().then((val) => {
+        setAnonymousId(val);
+      });
+    }
+    if (acLoaded && anonymousId) {
+      setLoaded(true);
+    }
+  }, [username, anonymousId, acLoaded]);
```
christianvogt, line 38: If `username` changes, you aren't recomputing the anonymous id. Although I don't think `username` should change and you are only going to compute the anonymous id once. But should you recompute?

If you don't care to recompute based on `username` changes, then you can remove all hook dependencies.
```suggestion
    computeAnonymousUserId().then((val) => {
      setAnonymousId(val);
    });
  // compute anonymousId only once
  // eslint-disable-next-line react-hooks/exhaustive-deps
  }, []);
```

frontend/src/concepts/analyticsTracking/useSegmentTracking.ts
```diff
@@ -10,28 +11,27 @@ export const useSegmentTracking = (): void => {
...
       });
     }
-  }, [clusterID, loadError, loaded, segmentKey, username, dashboardConfig]);
+  }, [
+    clusterID,
+    loadError,
+    loaded,
+    segmentKey,
+    username,
+    dashboardConfig,
```
christianvogt, line 33: Remove `dashboardConfig` from the list of dependencies because it's too broad.
Assign `dashboardConfig.spec.dashboardConfig.disableTracking` to a local variable and use it as a dependency instead.

frontend/src/concepts/analyticsTracking/trackingProperties.ts
```diff
@@ -3,7 +3,10 @@ export type ODHSegmentKey = {
 };
 
 export type IdentifyEventProperties = {
+  isAdmin: boolean;
   anonymousID?: string;
+  userId?: string;
```
christianvogt, line 8: `userId` isn't used anywhere. Should this be removed? I also thought that we shouldn't include identifiable information in events.

pilhuhn: It could indeed be removed from _this_ PR.  UserId would be the web-user-id for the sandbox case.

frontend/src/concepts/analyticsTracking/useTrackUser.ts
pilhuhn: The usename is set at login. So I don't think it will ever change during one session.

pilhuhn:
@christianvogt This was merged in the meanwhile. Should I open a subsequent PR with your suggestions?

christianvogt:
@pilhuhn  Yes please open a new PR for additional changes. This shouldn't have been merged.

PR #3028, opened 2024-07-23
Test: Import and run a Pipeline

FedeAlonso:
https://issues.redhat.com/browse/RHOAIENG-9869

## Description
E2E test: Import a Pipeline and run it using Cypress

## How Has This Been Tested?
Locally, against a Live Cluster. Screenshots of the results will be posted here

## Test Impact
This is already a test.

## Request review criteria:

### DIFF
--- a/frontend/src/__tests__/cypress/cypress/fixtures/resources/yaml/data_connection.yml
+++ b/frontend/src/__tests__/cypress/cypress/fixtures/resources/yaml/data_connection.yml
@@ -1,0 +1,41 @@
+kind: Secret
+apiVersion: v1
+metadata:
+  name: aws-connection-ods-ci-ds-pipelines
+  namespace: {{NAMESPACE}}
+  labels:
+    opendatahub.io/dashboard: 'true'
+    opendatahub.io/managed: 'true'
+  annotations:
+    opendatahub.io/connection-type: s3
+    openshift.io/display-name: ods-ci-ds-pipelines
+  managedFields:
+    - manager: unknown
+      operation: Update
+      apiVersion: v1
+      fieldsType: FieldsV1
+      fieldsV1:
+        'f:data':
+          .: {}
+          'f:AWS_ACCESS_KEY_ID': {}
+          'f:AWS_DEFAULT_REGION': {}
+          'f:AWS_S3_BUCKET': {}
+          'f:AWS_S3_ENDPOINT': {}
+          'f:AWS_SECRET_ACCESS_KEY': {}
+        'f:metadata':
+          'f:annotations':
+            .: {}
+            'f:opendatahub.io/connection-type': {}
+            'f:openshift.io/display-name': {}
+          'f:labels':
+            .: {}
+            'f:opendatahub.io/dashboard': {}
+            'f:opendatahub.io/managed': {}
+        'f:type': {}
+data:
+  AWS_ACCESS_KEY_ID: {{AWS_ACCESS_KEY_ID}}
+  AWS_DEFAULT_REGION: {{AWS_DEFAULT_REGION}}
+  AWS_S3_BUCKET: {{AWS_S3_BUCKET}}
+  AWS_S3_ENDPOINT: {{AWS_S3_ENDPOINT}}
+  AWS_SECRET_ACCESS_KEY: {{AWS_SECRET_ACCESS_KEY}}
+type: Opaque
--- a/frontend/src/__tests__/cypress/cypress/tests/e2e/pipelines/pipelines.cy.ts
+++ b/frontend/src/__tests__/cypress/cypress/tests/e2e/pipelines/pipelines.cy.ts
@@ -1,0 +1,124 @@
+import { createOpenShiftProject, deleteOpenShiftProject, applyOpenShiftYaml } from '~/__tests__/cypress/cypress/utils/ocCommands'; 
+import { replacePlaceholdersInYaml } from '~/__tests__/cypress/cypress/utils/yaml_files';
+import { ADMIN_USER } from '~/__tests__/cypress/cypress/utils/e2eUsers';
+import { AWS_PIPELINES_BUCKET } from '~/__tests__/cypress/cypress/utils/s3Buckets';
+
+const projectName = 'test-pipelines-prj';
+
+describe('An admin user can import and run a pipeline', { testIsolation: false }, () => {
+  before(() => {
+    // Provision a Project
+    createOpenShiftProject(projectName).then((result) => {
+      expect(result.code).to.eq(0, `ERROR provisioning ${projectName} Project
+                                    stdout: ${result.stdout}
+                                    stderr: ${result.stderr}`);
+    
+    
+    })
+    // Create a pipeline compatible Data Connection
+    const replacements = {
+      NAMESPACE: projectName,
+      AWS_ACCESS_KEY_ID: Buffer.from(AWS_PIPELINES_BUCKET.AWS_ACCESS_KEY_ID).toString('base64'),
+      AWS_DEFAULT_REGION: Buffer.from(AWS_PIPELINES_BUCKET.AWS_REGION).toString('base64'),
+      AWS_S3_BUCKET: Buffer.from(AWS_PIPELINES_BUCKET.BUCKET_NAME).toString('base64'),
+      AWS_S3_ENDPOINT: Buffer.from(AWS_PIPELINES_BUCKET.AWS_ENDPOINT).toString('base64'),
+      AWS_SECRET_ACCESS_KEY: Buffer.from(AWS_PIPELINES_BUCKET.AWS_SECRET_ACCESS_KEY).toString('base64')
+    };
+    cy.fixture('resources/yaml/data_connection.yml').then((yamlContent) => {
+      const modifiedYamlContent = replacePlaceholdersInYaml(yamlContent, replacements);
+      const tempFilePath = 'cypress/temp_data_connection.yaml';
+      applyOpenShiftYaml(modifiedYamlContent).then((result) => {
+        expect(result.code).to.eq(0, `ERROR applying YAML content\nstdout: ${result.stdout}\nstderr: ${result.stderr}`);
+      });
+    });
+  });
+  
+  after(() => {
+    // Delete provisioned Project
+    deleteOpenShiftProject(projectName).then((result) => {
+      expect(result.code).to.eq(0, `ERROR deleting ${projectName} Project
+                                    stdout: ${result.stdout}
+                                    stderr: ${result.stderr}`);
+    });
+  });
+
+  it('should login and load page', () => {
+    cy.log(Cypress.env());
+    cy.visitWithLogin('/', ADMIN_USER);
+    cy.findByRole('banner', { name: 'page masthead' }).contains(ADMIN_USER.USERNAME);
+  });
+});
+
+
+
+
+// import { ADMIN_USER } from '~/__tests__/cypress/cypress/utils/e2eUsers';
+// import { createOpenShiftProject, deleteOpenShiftProject } from '~/__tests__/cypress/cypress/utils/ocCommands';
+
+// const projectName = 'test-pipelines-prj';
+
+// describe.only('An admin user can import and run a pipeline', { testIsolation: false }, () => {
+//   before(() => {
+//     // Provision a Project
+//     createOpenShiftProject(projectName).then((result) => {
+//       expect(result.code).to.eq(0, 'Failed to provision a Project');
+//     });
+//     // cy.request('https://api.spacexdata.com/v3/missions').its('body').should('have.length', 10)
+//   })
+
+//   after(() => {
+//     // Delete provisioned Project
+//     deleteOpenShiftProject(projectName).then((result) => {
+//       expect(result.code).to.eq(0, 'Project deletion should succeed');
+//     });
+//   })
+  
+//   it('should login and load page', () => {
+//     cy.visitWithLogin('/');
+//     cy.findByRole('banner', { name: 'page masthead' }).contains(ADMIN_USER.USERNAME);
+//   });
+// });
+
+
+/**
+ * Steps for base test
+ * 
+ * Before (Provision):
+ *  1. Create "test-e2e-pipelines" DS Project if doesn't exist
+ *      Name (Mandatory)
+ *      Resource Name (Mandatory but can be extracted from the name)
+ *      Description
+ * oc new-project test-pipelines-prj --display-name 'test-pipelines-prj'
+ * 
+ * 
+ *  2. Create Data Connection if doesn't already exist
+ *      Name (Mandatory)
+ *      Access Key (Mandatory)
+ *      Secret Key (Mandatory)
+ *      Endpoint (Mandatory)
+ *      Region (Mandatory)
+ *      Bucket (Mandatory)
+ *  
+ *  3. Configure Pipeline Server if it's not already configured
+ *      (Using Data Connection)
+ * 
+ *  TEST:
+ *  0. Login as an admin
+ *  1. Navigate to "Overview" tab of "test-e2e-pipelines" DS Project
+ *  2. Click in "Import pipeline" link
+ *  3. Fill "Pipeline name" from "Import pipeline" modal
+ *  4. (unrequired) Fill "Pipeline description" from "Import pipeline" modal 
+ *  5. Chose "Import by url" in "Import pipeline" modal 
+ *  6. Fill with URL (https://raw.githubusercontent.com/red-hat-data-services/ods-ci/master/ods_ci/tests/Resources/Files/pipeline-samples/v2/flip_coin_compiled.yaml) in "Import pipeline" modal 
+ *  7. Click "Import pipeline" button  from "Import pipeline" modal 
+ *  8. Navigate to "Data Science Pipelines"
+ *  9. Click on the kebab button from our created pipeline, and click on Create run
+ *  10. Select the experiment "Default"
+ *  11. Fill the name
+ *  12. (unrequired) Fill the description
+ *  13. Click on "Create run" button
+ *  14. Wait for the status not to be "Waiting" or "Running" but "Succeeded"
+ * 
+ * After (Clean-up):
+ *  1. Delete"test-e2e-pipelines" DS Project if does exist (Waterfall deletion??)
+ */
--- a/frontend/src/__tests__/cypress/cypress/tests/e2e/pipelines/pipelines.cy.ts
+++ b/frontend/src/__tests__/cypress/cypress/tests/e2e/pipelines/pipelines.cy.ts
@@ -27,7 +27,7 @@ describe('An admin user can import and run a pipeline', { testIsolation: false }
     cy.fixture('resources/yaml/data_connection.yml').then((yamlContent) => {
       const modifiedYamlContent = replacePlaceholdersInYaml(yamlContent, replacements);
       const tempFilePath = 'cypress/temp_data_connection.yaml';
-      applyOpenShiftYaml(modifiedYamlContent).then((result) => {
+      applyOpenShiftYaml(modifiedYamlContent, tempFilePath).then((result) => {
         expect(result.code).to.eq(0, `ERROR applying YAML content\nstdout: ${result.stdout}\nstderr: ${result.stderr}`);
       });
     });
--- a/frontend/src/__tests__/cypress/cypress/tests/e2e/pipelines/pipelines.cy.ts
+++ b/frontend/src/__tests__/cypress/cypress/tests/e2e/pipelines/pipelines.cy.ts
@@ -13,10 +13,10 @@ describe('An admin user can import and run a pipeline', { testIsolation: false }
                                     stdout: ${result.stdout}
                                     stderr: ${result.stderr}`);
     
-    
     })
+    
     // Create a pipeline compatible Data Connection
-    const replacements = {
+    const dataConnectionReplacements = {
       NAMESPACE: projectName,
       AWS_ACCESS_KEY_ID: Buffer.from(AWS_PIPELINES_BUCKET.AWS_ACCESS_KEY_ID).toString('base64'),
       AWS_DEFAULT_REGION: Buffer.from(AWS_PIPELINES_BUCKET.AWS_REGION).toString('base64'),
@@ -25,12 +25,29 @@ describe('An admin user can import and run a pipeline', { testIsolation: false }
       AWS_SECRET_ACCESS_KEY: Buffer.from(AWS_PIPELINES_BUCKET.AWS_SECRET_ACCESS_KEY).toString('base64')
     };
     cy.fixture('resources/yaml/data_connection.yml').then((yamlContent) => {
-      const modifiedYamlContent = replacePlaceholdersInYaml(yamlContent, replacements);
+      const modifiedYamlContent = replacePlaceholdersInYaml(yamlContent, dataConnectionReplacements);
       const tempFilePath = 'cypress/temp_data_connection.yaml';
       applyOpenShiftYaml(modifiedYamlContent, tempFilePath).then((result) => {
         expect(result.code).to.eq(0, `ERROR applying YAML content\nstdout: ${result.stdout}\nstderr: ${result.stderr}`);
       });
     });
+
+    // Configure Pipeline server
+    // Create DSPA Secret
+    const dspaSecretReplacements = {
+      NAMESPACE: projectName,
+      AWS_ACCESS_KEY_ID: Buffer.from(AWS_PIPELINES_BUCKET.AWS_ACCESS_KEY_ID).toString('base64'),
+      AWS_SECRET_ACCESS_KEY: Buffer.from(AWS_PIPELINES_BUCKET.AWS_SECRET_ACCESS_KEY).toString('base64')
+    };
+    cy.fixture('resources/yaml/dspa_secret.yml').then((yamlContent) => {
+      const modifiedYamlContent = replacePlaceholdersInYaml(yamlContent, dataConnectionReplacements);
+      const tempFilePath = 'cypress/dspa_secret.yaml';
+      applyOpenShiftYaml(modifiedYamlContent, tempFilePath).then((result) => {
+        expect(result.code).to.eq(0, `ERROR applying YAML content\nstdout: ${result.stdout}\nstderr: ${result.stderr}`);
+      });
+    });
+
+    // Create DSPA
   });
   
   after(() => {
--- a/frontend/src/__tests__/cypress/cypress/utils/ocCommands.ts
+++ b/frontend/src/__tests__/cypress/cypress/utils/ocCommands.ts
@@ -1,0 +1,117 @@
+/**
+ * Applies the given YAML content using the `oc apply` command.
+ * 
+ * @param yamlContent YAML content to be applied
+ * @returns Cypress Chainable
+ */
+export const applyOpenShiftYaml = (yamlContent: string) => {
+    const tempFilePath = 'cypress/temp.yaml';
+    cy.writeFile(tempFilePath, yamlContent);
+  
+    const ocCommand = `oc apply -f ${tempFilePath}`;
+  
+    return cy.exec(ocCommand, { failOnNonZeroExit: false }).then((result) => {
+    //   cy.log('Command Output:', result.stdout);
+    //   cy.log('Command Error:', result.stderr);
+  
+    //   // Clean up the temporary file
+    //   cy.task('deleteFile', tempFilePath);
+  
+    //   if (result.code === 0) {
+    //     cy.log(`YAML content applied successfully.`);
+    //   } else {
+    //     cy.log(`Failed to apply YAML content.`);
+    //   }
+  
+      return result;
+    });
+  };
+
+/**
+ * Create an Openshift Project
+ * 
+ * @param projectName Project Name
+ * @param displayName Project Display Name
+ * @returns Result Object of the operation
+ */
+export const createOpenShiftProject = (projectName: string, displayName?: string) => {
+  
+  const finalDisplayName = displayName || projectName;  
+  const ocCommand = `oc new-project ${projectName} --display-name='${finalDisplayName}'`;
+
+  cy.log(Cypress.env())
+
+  return cy.exec(ocCommand, { failOnNonZeroExit: false }).then((result) => {
+    return result;
+  });
+};
+
+
+/**
+ * Delete an Openshift Project given its name
+ * 
+ * @param projectName OpenShift Project name
+ * @returns Result Object of the operation
+ */
+export const deleteOpenShiftProject = (projectName: string) => {
+  const ocCommand = `oc delete project ${projectName}`;
+  return cy.exec(ocCommand, { failOnNonZeroExit: false }).then((result) => {
+    return result;
+  });
+};
+  
+
+
+
+
+
+// /**
+//  * Create an Openshift Project
+//  * 
+//  * @param projectName Project Name
+//  * @param displayName Project Display Name
+//  * @returns Result Object of the operation
+//  */
+// export const createOpenShiftProject = (projectName: string, displayName?: string) => {
+  
+//     const finalDisplayName = displayName || projectName;  
+//     const ocCommand = `oc new-project ${projectName} --display-name='${finalDisplayName}'`;
+  
+//     return cy.exec(ocCommand, { failOnNonZeroExit: false }).then((result) => {
+  
+//       cy.log('Command Output:', result.stdout);
+//       cy.log('Command Error:', result.stderr);
+          
+//       if (result.code === 0) {
+//         cy.log(`Project ${projectName} created successfully.`);
+//       } else {
+//         cy.log(`Failed to create project ${projectName}.`);
+//       }
+//       return result;
+//     });
+//   };
+  
+  
+//   /**
+//    * Delete an Openshift Project given its name
+//    * 
+//    * @param projectName OpenShift Project name
+//    * @returns Result Object of the operation
+//    */
+//   export const deleteOpenShiftProject = (projectName: string) => {
+//     const ocCommand = `oc delete project ${projectName}`;
+    
+//     return cy.exec(ocCommand, { failOnNonZeroExit: false }).then((result) => {
+//       cy.log('Command Output:', result.stdout);
+//       cy.log('Command Error:', result.stderr);
+    
+//       if (result.code === 0) {
+//         cy.log(`Project ${projectName} deleted successfully.`);
+//       } else {
+//         cy.log(`Failed to delete project ${projectName}.`);
+//       }
+    
+//       return result;
+//     });
+//   };
+    
--- a/frontend/src/__tests__/cypress/cypress/utils/s3Buckets.ts
+++ b/frontend/src/__tests__/cypress/cypress/utils/s3Buckets.ts
@@ -1,0 +1,3 @@
+import type { AWSS3Bucket } from '~/__tests__/cypress/cypress/types';
+
+export const AWS_PIPELINES_BUCKET: AWSS3Bucket = Cypress.env('AWS_PIPELINES_BUCKET');
--- a/frontend/src/__tests__/cypress/cypress/utils/testConfig.ts
+++ b/frontend/src/__tests__/cypress/cypress/utils/testConfig.ts
@@ -3,7 +3,7 @@ import path from 'path';
 import { env } from 'process';
 import dotenv from 'dotenv';
 import YAML from 'yaml';
-import type { UserAuthConfig, TestConfig } from '~/__tests__/cypress/cypress/types';
+import type { UserAuthConfig, TestConfig , AWSS3Bucket } from '~/__tests__/cypress/cypress/types';
 
 [
   `.env.cypress${env.CY_MOCK ? '.mock' : ''}.local`,
@@ -34,10 +34,19 @@ const ADMIN_USER: UserAuthConfig = testConfig?.OCP_ADMIN_USER ?? {
   PASSWORD: env.ADMIN_USER_PASSWORD || '',
 };
 
+const AWS_PIPELINES_BUCKET: AWSS3Bucket = testConfig?.AWS_PIPELINES_BUCKET ?? {
+  BUCKET_NAME: env.AWS_PIPELINES_BUCKET_NAME || '',
+  AWS_ACCESS_KEY_ID: env.AWS_PIPELINES_BUCKET_ACCESS_KEY_ID || '',
+  AWS_SECRET_ACCESS_KEY: env.AWS_PIPELINES_BUCKET_SECRET_ACCESS_KEY || '',
+  AWS_ENDPOINT: env.AWS_PIPELINES_BUCKET_ENDPOINT || '',
+  AWS_REGION: env.AWS_PIPELINES_BUCKET_REGION || '',
+};
+
 // spread the cypressEnv variables into the cypress config
 export const cypressEnv = {
   TEST_USER,
   ADMIN_USER,
+  AWS_PIPELINES_BUCKET
 };
 
 // re-export the updated process env
--- a/frontend/src/__tests__/cypress/.gitignore
+++ b/frontend/src/__tests__/cypress/.gitignore
@@ -2,3 +2,5 @@ cypress/downloads
 coverage
 results
 .nyc_output
+test-variables.yml
+cypress/temp*.yaml
--- a/frontend/src/__tests__/cypress/cypress/fixtures/resources/yaml/dspa_secret.yml
+++ b/frontend/src/__tests__/cypress/cypress/fixtures/resources/yaml/dspa_secret.yml
@@ -1,0 +1,26 @@
+kind: Secret
+apiVersion: v1
+metadata:
+  name: dashboard-dspa-secret
+  namespace: {{NAMESPACE}}
+  labels:
+    opendatahub.io/dashboard: 'true'
+  managedFields:
+    - manager: unknown
+      operation: Update
+      apiVersion: v1
+      fieldsType: FieldsV1
+      fieldsV1:
+        'f:data':
+          .: {}
+          'f:AWS_ACCESS_KEY_ID': {}
+          'f:AWS_SECRET_ACCESS_KEY': {}
+        'f:metadata':
+          'f:labels':
+            .: {}
+            'f:opendatahub.io/dashboard': {}
+        'f:type': {}
+data:
+  AWS_ACCESS_KEY_ID: {{AWS_ACCESS_KEY_ID}}
+  AWS_SECRET_ACCESS_KEY: {{AWS_SECRET_ACCESS_KEY}}
+type: Opaque
--- a/frontend/src/__tests__/cypress/cypress/types.ts
+++ b/frontend/src/__tests__/cypress/cypress/types.ts
@@ -25,8 +25,17 @@ export type UserAuthConfig = {
   PASSWORD: string;
 };
 
+export type AWSS3Bucket = {
+  BUCKET_NAME: string;
+  AWS_ACCESS_KEY_ID: string;
+  AWS_SECRET_ACCESS_KEY: string;
+  AWS_ENDPOINT: string;
+  AWS_REGION: string;
+};
+
 export type TestConfig = {
   ODH_DASHBOARD_URL: string;
   TEST_USER: UserAuthConfig;
   OCP_ADMIN_USER: UserAuthConfig;
+  AWS_PIPELINES_BUCKET: AWSS3Bucket;
 };
--- a/frontend/src/__tests__/cypress/cypress/utils/yaml_files.ts
+++ b/frontend/src/__tests__/cypress/cypress/utils/yaml_files.ts
@@ -1,0 +1,15 @@
+/**
+ * Replaces placeholders in a YAML content with environment variable values.
+ * 
+ * @param yamlContent YAML content as a string
+ * @param replacements Object containing placeholder keys and their corresponding replacement values
+ * @returns Modified YAML content
+ */
+export const replacePlaceholdersInYaml = (yamlContent: string, replacements: { [key: string]: string }): string => {
+    let modifiedYaml = yamlContent;
+    for (const [key, value] of Object.entries(replacements)) {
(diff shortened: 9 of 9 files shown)


frontend/src/__tests__/cypress/cypress/tests/e2e/pipelines/pipelines.cy.ts
```diff
@@ -0,0 +1,194 @@
...
+describe('An admin user can import and run a pipeline', { testIsolation: false }, () => {
+  before(() => {
+    // Provision a Project
+    createOpenShiftProject(projectName).then((result) => {
+      expect(result.code).to.eq(
+        0,
+        `ERROR provisioning ${projectName} Project
+                                    stdout: ${result.stdout}
+                                    stderr: ${result.stderr}`,
+      );
```
christianvogt, line 32: All this error handling can be part of the underlying function instead of repeating each time.

frontend/src/__tests__/cypress/cypress/tests/e2e/pipelines/pipelines.cy.ts
```diff
@@ -0,0 +1,194 @@
...
+    const dataConnectionReplacements = {
+      NAMESPACE: projectName,
+      AWS_ACCESS_KEY_ID: Buffer.from(AWS_PIPELINES_BUCKET.AWS_ACCESS_KEY_ID).toString('base64'),
+      AWS_DEFAULT_REGION: Buffer.from(AWS_PIPELINES_BUCKET.AWS_REGION).toString('base64'),
+      AWS_S3_BUCKET: Buffer.from(AWS_PIPELINES_BUCKET.BUCKET_NAME).toString('base64'),
+      AWS_S3_ENDPOINT: Buffer.from(AWS_PIPELINES_BUCKET.AWS_ENDPOINT).toString('base64'),
+      AWS_SECRET_ACCESS_KEY: Buffer.from(AWS_PIPELINES_BUCKET.AWS_SECRET_ACCESS_KEY).toString(
+        'base64',
+      ),
+    };
```
christianvogt, line 45: We need types and utility functions for this instead of tests duplicating it.

frontend/src/__tests__/cypress/cypress/tests/e2e/pipelines/pipelines.cy.ts
```diff
@@ -0,0 +1,194 @@
...
+      const tempFilePath = 'cypress/temp_data_connection.yaml';
+      applyOpenShiftYaml(modifiedYamlContent, tempFilePath).then((result) => {
+        expect(result.code).to.eq(
+          0,
+          `ERROR applying YAML content
+                                      stdout: ${result.stdout}
+                                      stderr: ${result.stderr}`,
+        );
+      });
+    });
```
christianvogt, line 60: Fixtures / resource creation should be placed in reusable utility functions.

frontend/src/__tests__/cypress/cypress/tests/e2e/pipelines/pipelines.cy.ts
```diff
@@ -0,0 +1,194 @@
...
+        `ERROR deleting ${projectName} Project
+                                    stdout: ${result.stdout}
+                                    stderr: ${result.stderr}`,
+      );
+    });
+  });
+
+  it('An admin User can Import and Run a Pipeline', () => {
+    // Login as an admin
+    cy.visitWithLogin('/', ADMIN_USER);
```
christianvogt, line 118: Is it normal for e2e to always start from the root or can we save time and jump directly to the area we want to test?

frontend/src/__tests__/cypress/cypress/tests/e2e/pipelines/pipelines.cy.ts
```diff
@@ -0,0 +1,194 @@
...
+    cy.findByRole('banner', { name: 'page masthead' }).contains(ADMIN_USER.USERNAME);
```
christianvogt, line 119: This assertion is irrelevant to the test case.

frontend/src/__tests__/cypress/cypress/tests/e2e/pipelines/pipelines.cy.ts
```diff
@@ -0,0 +1,194 @@
...
+    // Check if "All available projects" is selected in the DS Projects view
+    projectListPage
+      .findProjectsTypeDropdown()
+      .invoke('text')
+      .then((text) => {
+        if (!text.includes(allAvailableProjectsText)) {
+          projectListPage.findProjectsTypeDropdown().click();
+          projectListPage.findProjectsTypeDropdownByText(allAvailableProjectsText).click();
+        }
+      });
```
christianvogt, line 135: Should probably add the annotation on project creation to avoid this step.

frontend/src/__tests__/cypress/cypress/tests/e2e/pipelines/pipelines.cy.ts
```diff
@@ -0,0 +1,194 @@
...
+
+    // Open the project
+    projectListPage.findProjectLink(projectName).click();
+
+    // Increasing the timeout to ~2mins so the DSPA can be loaded
+    projectDetails.findImportPipelineButton(120000).click();
```
christianvogt, line 141: We need to find a better way to avoid simply increasing wait times. 
If there's nothing in the UI to provide a proper status then perhaps we need a mechanism to also watch resources and react to error states.

frontend/src/__tests__/cypress/cypress/tests/e2e/pipelines/pipelines.cy.ts
```diff
@@ -0,0 +1,194 @@
...
+
+    // Fill tue Import Pipeline modal
+    pipelineImportModal.findPipelineNameInput().type(testPipelineName);
+    pipelineImportModal.findPipelineDescriptionInput().type('Pipeline Description');
+    pipelineImportModal.findImportPipelineRadio().click();
+    pipelineImportModal
+      .findPipelineUrlInput()
+      .type(
+        'https://raw.githubusercontent.com/red-hat-data-services/ods-ci/master/ods_ci/tests/Resources/Files/pipeline-samples/v2/flip_coin_compiled.yaml',
```
christianvogt, line 150: Is this acceptable in ODS-CI to rely on content from this location or should it be from a red hat source?

sthundat:
@FedeAlonso Can you  add the command to run the e2e tests locally under 'How Has This Been Tested'  for example `CY_TEST_CONFIG=<path_to>/test-variables.yml npm run cypress:open`

manosnoam:
Thank you for this new pipeline test @FedeAlonso !
I tried to execute it in **Jenkins cypress-tests/22/** but it failed on `AssertionError: Timed out retrying after 10000ms: Unable to find a label with the text of: Options menu`  see screenshots and full output in Jenkins.

frontend/src/__tests__/cypress/cypress/utils/s3Buckets.ts
```diff
@@ -0,0 +1,3 @@
+import type { AWSS3Bucket } from '~/__tests__/cypress/cypress/types';
+
+export const AWS_PIPELINES_BUCKET: AWSS3Bucket = Cypress.env('AWS_PIPELINES_BUCKET');
```
manosnoam, line 3: In our Jenkins CI we use `S3.BUCKET_2` for the pipeline bucket (according to our test-variables.yaml). 
Would this also work if calling it in bash: 
`export AWS_PIPELINES_BUCKET=S3.BUCKET_2`

manosnoam (changes requested):

frontend/src/__tests__/cypress/cypress/fixtures/resources/yaml/data_connection.yml
```diff
@@ -0,0 +1,41 @@
...
+metadata:
+  name: aws-connection-ods-ci-ds-pipelines
+  namespace: {{NAMESPACE}}
+  labels:
+    opendatahub.io/dashboard: 'true'
+    opendatahub.io/managed: 'true'
+  annotations:
+    opendatahub.io/connection-type: s3
+    openshift.io/display-name: ods-ci-ds-pipelines
+  managedFields:
```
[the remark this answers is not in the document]
FedeAlonso: Fixed in 
505bb28

frontend/src/__tests__/cypress/cypress/fixtures/resources/yaml/dspa.yml
```diff
@@ -0,0 +1,133 @@
+apiVersion: datasciencepipelinesapplications.opendatahub.io/v1alpha1
+kind: DataSciencePipelinesApplication
+metadata:
+  finalizers:
+    - datasciencepipelinesapplications.opendatahub.io/finalizer
+  generation: 2
+  managedFields:
```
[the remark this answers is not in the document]
FedeAlonso: Fixed in 
505bb28

frontend/src/__tests__/cypress/cypress/tests/e2e/pipelines/pipelines.cy.ts
FedeAlonso: This is not a passive wait, is an active one. It's polling fot the element until it's available. And it's the same behaviour an end user will do (wait until the spinner is gone and the Import Pipeline button is available). 
If it reach the 2 minutes (now 3 minutes because sometimes the Pipeline Server takes a little bit longer to be available) it will fail. But if it's available in 2 seconds, the click will be performed after those 2 seconds

christianvogt: I agree it's not just a `cy.wait`. What I'm getting at here is how do we know what is a long enough wait time to set for a test case.
We can keep increasing the timeout until it's "good enough". This is fine for the time being and I'm not requesting a change here today.

In general this is a bad approach but maybe all we got for now. If we can be more precise with watching events on a more timely basis, we can avoid having very long timeouts if we can expect a status event to occur within 10 seconds (as an example), then we watch for such events.

I want us to find a better way for our tests to be written to not have flakes because in future the timeout needs to be increased. Perfect example is you started with 2 mins but now increased to 3 mins.

Similarly, if we can find a way to create the test as to have wait for success or error states that could also alleviate some unnecessary wait times when a test is failing.

frontend/src/__tests__/cypress/cypress/pages/pipelines/pipelinesTable.ts
```diff
@@ -194,6 +194,35 @@ class PipelinesTable {
...
+    return new PipelineVersionsTableRow(() =>
+      this.find()
+        .contains('td', pipelineName)
+        .parent('tr')
+        .next('tr')
+        .findByTestId('pipeline-versions-table')
+        .contains('td', pipelineVersionName)
+        .parent('tr'),
+    );
+  }
```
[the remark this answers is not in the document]
FedeAlonso: I've deleted that code [10e7da1](https://github.com/opendatahub-io/odh-dashboard/pull/3028/commits/10e7da176f88945358e91c33ae3b817b15aa102b)

frontend/src/__tests__/cypress/cypress/tests/e2e/pipelines/pipelines.cy.ts
FedeAlonso: This does not apply anymore, as in 2.12 there is no selector (all projects can be DS Projects)

frontend/src/__tests__/cypress/cypress/tests/e2e/pipelines/pipelines.cy.ts
```diff
@@ -0,0 +1,194 @@
...
+      });
+
+    // Expand the pipeline row
+    pipelinesTable.expandRowByPipelineName(testPipelineName);
+
+    // Open the create run View using the 'Create run' option from the version kebab
+    pipelinesTable
+      .findPipelineVersionRowByVersionName(testPipelineName, testPipelineName)
+      .findKebab()
+      .click();
```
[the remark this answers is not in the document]
FedeAlonso: It's already deleted

frontend/src/__tests__/cypress/cypress/pages/pipelines/pipelinesTable.ts
```diff
@@ -194,6 +194,35 @@ class PipelinesTable {
...
+            // Check if "aria-expanded" is not "true" from the expand button
+            if (ariaExpanded !== 'true') {
+              // Click the expand button
+              cy.get('td:first-child button').click();
+            }
+          });
+      });
+  }
+
+  findPipelineVersionRowByVersionName(pipelineName: string, pipelineVersionName: string) {
```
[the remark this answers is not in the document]
FedeAlonso: It's already deleted

frontend/src/__tests__/cypress/cypress/utils/ocCommands.ts
```diff
@@ -0,0 +1,46 @@
...
+ * Applies the given YAML content using the `oc apply` command.
+ *
+ * @param yamlContent YAML content to be applied
+ * @param tempFilePath Path to the temporary file
+ * @returns Cypress Chainable
+ */
+export const applyOpenShiftYaml = (yamlContent: string, tempFilePath: string) => {
+  cy.writeFile(tempFilePath, yamlContent);
+
+  const ocCommand = `oc apply -f ${tempFilePath}`;
```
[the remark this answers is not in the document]
FedeAlonso: Fixed in [4d9a76b](https://github.com/opendatahub-io/odh-dashboard/pull/3028/commits/4d9a76b0f90b4a6b104a784dd800b2cd2f40bf91)

frontend/src/__tests__/cypress/cypress/tests/e2e/pipelines/pipelines.cy.ts
FedeAlonso: Fixed in [4d9a76b](https://github.com/opendatahub-io/odh-dashboard/pull/3028/commits/4d9a76b0f90b4a6b104a784dd800b2cd2f40bf91)

frontend/src/__tests__/cypress/cypress/tests/e2e/pipelines/pipelines.cy.ts
FedeAlonso: We could do it, but I think that we should emulate the behaviour of an end user, that's also why I'm using navigate() instead of visit().
This one it's not very important, as it's only going to catch things like "this page shouldn't be accessible through navigation".
I can change it if you think it's better.

frontend/src/__tests__/cypress/cypress/tests/e2e/pipelines/pipelines.cy.ts
FedeAlonso: Fixed in [ea6d831](https://github.com/opendatahub-io/odh-dashboard/pull/3028/commits/ea6d8314940ab4737edffbf374ae294ec55400c1)

frontend/src/__tests__/cypress/cypress/tests/e2e/pipelines/pipelines.cy.ts
FedeAlonso: I agree with you that we should not let the tests go on too long. In this case, the entire test takes ~3 minutes. Much of this time is spent on resource provisioning (deploying a DSP, and getting everything ready to deploy a pipeline). 
I also believe that tests should be atomic and deterministic, so that each test can be run independently (without having dependencies with another test). But that does not mean that each test should have its own specific resources. There are resources that will be provisioned once for all tests, other resources that will be provisioned for a complete suite, and others that will be provisioned for a single test. So if a single test is run, all granularity levels will need to be provisioned (which is not very efficient), but if we have a large test suite many of the resources will be provisioned only once, which increases efficiency.
For this case, I can imagine that in this suite other tests related to pipelines will be added to make use of these resources.

frontend/src/__tests__/cypress/cypress/utils/ocCommands.ts
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * @param projectName OpenShift Project name
+ * @returns Result Object of the operation
+ */
+export const deleteOpenShiftProject = (projectName: string) => {
+  const ocCommand = `oc delete project ${projectName}`;
+  // The default timeout is 60 seconds, and the deletion can take longer
+  return cy.exec(ocCommand, { failOnNonZeroExit: false, timeout: 180000 }).then((result) => {
+    return result;
+  });
```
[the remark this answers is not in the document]
FedeAlonso: Fixed 
4a286f3

frontend/src/__tests__/cypress/cypress/utils/s3Buckets.ts
manosnoam: @FedeAlonso if that is not working well, then you can add the required bucket details to our test-variables downstream (But for the long term, we should try to use the same variable names if they already exist in test-variables. Thanks!

FedeAlonso: Fixed in 
28d8d14

frontend/src/__tests__/cypress/cypress/tests/e2e/pipelines/pipelines.cy.ts
FedeAlonso: Fixed in 
28d8d14

manosnoam:
@FedeAlonso  A new run in CI cypress-tests/23 passed, however another one failed /cypress-tests/24 on:
`AssertionError: Timed out retrying after 10000ms: Unable to find an accessible element with the role "link" and name "test-pipelines-prj"`. Please check the build, thanks!

frontend/src/__tests__/cypress/cypress/tests/e2e/pipelines/pipelines.cy.ts
FedeAlonso: Fixed in [6a949ea](https://github.com/opendatahub-io/odh-dashboard/pull/3028/commits/6a949ea4f0fbc70b412b2e9462af4a8873ad8d03)

frontend/src/__tests__/cypress/cypress/tests/e2e/pipelines/pipelines.cy.ts
FedeAlonso: That's a good point. Probably it does not make sense to use an external repo.
I've created this new PR: https://github.com/opendatahub-io/odh-dashboard/pull/3050 
which contains a dummy pipeline so we can make use of that one.

FedeAlonso:
@manosnoam I've made some fixes and did a new run (cypress-tests/30):
image
image

frontend/src/__tests__/cypress/cypress/tests/e2e/pipelines/pipelines.cy.ts
```diff
@@ -0,0 +1,109 @@
...
+describe('An admin user can import and run a pipeline', { testIsolation: false }, () => {
+  before(() => {
+    // Provision a Project
+    createOpenShiftProject(projectName);
+
+    // Create a pipeline compatible Data Connection
+    const dataConnectionReplacements = {
+      NAMESPACE: projectName,
+      AWS_ACCESS_KEY_ID: Buffer.from(AWS_BUCKETS.AWS_ACCESS_KEY_ID).toString('base64'),
+      AWS_DEFAULT_REGION: Buffer.from(AWS_BUCKETS.BUCKET_2.REGION).toString('base64'),
```
manosnoam, line 33: @FedeAlonso Can you make "BUCKET_2" not hardcoded, but an env variable, like "CY_AWS_BUCKET" ?
This way the user (or CI) can set it according to their bucket value (e.g. `CY_AWS_BUCKET=BUCKET_2`

manosnoam (changes requested):

frontend/src/__tests__/cypress/cypress/fixtures/resources/yaml/data_connection.yml
```diff
@@ -0,0 +1,18 @@
+kind: Secret
+apiVersion: v1
+metadata:
+  name: aws-connection-ods-ci-ds-pipelines
```
manosnoam, line 4: @FedeAlonso Can you rename the file and inner fields without "ods-ci" ?
For example:

- Rename this file to: `aws-connection-ds-pipelines.yaml` (btw, yaml instead of yml)
- Remove "ods-ci" from any yaml field or code

frontend/src/__tests__/cypress/cypress/tests/e2e/pipelines/pipelines.cy.ts
```diff
@@ -0,0 +1,111 @@
+import {
+  createOpenShiftProject,
+  deleteOpenShiftProject,
+} from '~/__tests__/cypress/cypress/utils/oc_commands/project';
```
manosnoam, line 4: Is it common to use `~` when importing packages?

frontend/src/__tests__/cypress/cypress/tests/e2e/pipelines/pipelines.cy.ts
```diff
@@ -0,0 +1,111 @@
...
+  before(() => {
+    // Provision a Project
+    createOpenShiftProject(projectName);
+
+    // Create a pipeline compatible Data Connection
+    const dataConnectionReplacements = {
+      NAMESPACE: projectName,
+      AWS_ACCESS_KEY_ID: Buffer.from(AWS_BUCKETS.AWS_ACCESS_KEY_ID).toString('base64'),
+      AWS_DEFAULT_REGION: Buffer.from(AWS_BUCKETS.BUCKET_2.REGION).toString('base64'),
+      AWS_S3_BUCKET: Buffer.from(AWS_BUCKETS.BUCKET_2.NAME).toString('base64'),
```
manosnoam, line 35: See my previous comment about hardcoding "BUCKET_2" in the code.

manosnoam (changes requested):

frontend/src/__tests__/cypress/cypress/pages/pipelines/topology.ts
```diff
@@ -94,6 +94,25 @@ class RunDetails extends PipelinesTopology {
     return new DetailsItem(() => cy.findByTestId(`detail-item-${key}`));
   }
 
+  private getStatusLabel() {
+    return cy.get(
+      'h1[data-testid="app-page-title"] .pf-v5-l-split__item:last-child .pf-v5-c-label__text',
```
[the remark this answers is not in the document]
FedeAlonso: Fixed in 
cebadf7

frontend/src/__tests__/cypress/cypress/tests/e2e/pipelines/pipelines.cy.ts
FedeAlonso: yes, it works as a base directory

frontend/src/__tests__/cypress/cypress/fixtures/resources/yaml/data_connection.yml
FedeAlonso: I have modifed the extensions here [209f98c](https://github.com/opendatahub-io/odh-dashboard/pull/3028/commits/209f98c9b856407ae07ef9eb2ba953abcd0de054) 
But we need to keep some ods-ci string, as it is literally part of the name of the bucket we use for pipelines

frontend/src/__tests__/cypress/cypress/tests/e2e/pipelines/pipelines.cy.ts
FedeAlonso: same

frontend/src/__tests__/cypress/cypress/pages/projects.ts
```diff
@@ -95,6 +95,26 @@ class ProjectListPage {
...
     return cy.findByRole('button', { name: 'Create a workbench' });
   }
+
+  findProjectsTypeDropdown() {
+    return cy.findByLabelText('Options menu');
+  }
+
+  findProjectsTypeDropdownByText(projectType: string) {
+    return cy.contains('button', projectType);
+  }
```
christianvogt, line 105: Remove unused code.

frontend/src/__tests__/cypress/cypress/pages/projects.ts
```diff
@@ -95,6 +95,26 @@ class ProjectListPage {
...
+
+  findProjectsDropdownFilter(){
+    return cy.findByTestId('filter-dropdown-select');
```
christianvogt, line 108: I believe you can use `.findFilterMenuOption('filter-dropdown-select', 'Name')` as we have custom commands to work with dropdowns.

frontend/src/__tests__/cypress/cypress/tests/e2e/pipelines/pipelines.cy.ts
```diff
@@ -0,0 +1,111 @@
...
+      AWS_SECRET_ACCESS_KEY: Buffer.from(AWS_BUCKETS.AWS_SECRET_ACCESS_KEY).toString('base64'),
+    };
+    createDSPASecret(dspaSecretReplacements);
+
+    // Configure Pipeline server: Create DSPA
+    const dspaReplacements = {
+      DSPA_SECRET_NAME: dspaSecretName,
+      NAMESPACE: projectName,
+      AWS_S3_BUCKET: AWS_BUCKETS.BUCKET_2.NAME,
+    };
```
christianvogt, line 55: These need proper types and you need to type `createDSPA` and other functions like it.

frontend/src/__tests__/cypress/cypress/tests/e2e/pipelines/pipelines.cy.ts
```diff
@@ -0,0 +1,111 @@
...
+    // Fill tue Import Pipeline modal
+    pipelineImportModal.findPipelineNameInput().type(testPipelineName);
+    pipelineImportModal.findPipelineDescriptionInput().type('Pipeline Description');
+    pipelineImportModal.findImportPipelineRadio().click();
+    pipelineImportModal
+      .findPipelineUrlInput()
+      //TODO: modify this URL once the PR is merged
+      .type(
+        'https://raw.githubusercontent.com/opendatahub-io/odh-dashboard/caab82536b4dd5d39fb7a06a6c3248f10c183417/frontend/src/__tests__/resources/pipelines_samples/dummy_pipeline_compiled.yaml',
+      );
```
christianvogt, line 89: Why a separate PR? Why not include the file in this PR.

frontend/src/__tests__/cypress/cypress/tests/e2e/pipelines/pipelines.cy.ts
```diff
@@ -0,0 +1,111 @@
...
+
+    //Fill the Create run fields
+    createRunPage.findExperimentSelect().click();
+    createRunPage.selectExperimentByName('Default');
+    createRunPage.fillName(testRunName);
+    createRunPage.fillDescription('Run Description');
+    createRunPage.findSubmitButton().click();
+
+    //Redirected to the Graph view of the created run
+    pipelineRunDetails.expectStatusLabelToBe('Succeeded');
```
christianvogt, line 109: Default timeout is 10 seconds. Is that going to be sufficient while waiting for the run to complete?
Here's a scenario where it would be great to have some kind of conditional checks where if the status turns to error it fails.

I don't know how exactly to do this. But perhaps we note it for future improvements.

frontend/src/__tests__/cypress/cypress/utils/oc_commands/dataConnection.ts
```diff
@@ -0,0 +1,28 @@
...
+ *                  AWS_ACCESS_KEY_ID: <AWS ACCESS KEY ID>,
+ *                  AWS_DEFAULT_REGION: <AWS REGION>,
+ *                  AWS_S3_BUCKET: <AWS BUCKET NAME>,
+ *                  AWS_S3_ENDPOINT: <AWS ENDPOINT>,
+ *                  AWS_SECRET_ACCESS_KEY: <AWS SECRET>,
+ *               }
+ * @param yamlFilePath
+ */
+export const createDataConnection = (
+  dataConnectionReplacements: { [key: string]: string },
```
christianvogt, line 19: Add proper types.

frontend/src/__tests__/cypress/cypress/utils/oc_commands/dspa.ts
```diff
@@ -0,0 +1,49 @@
...
+ *              dspaSecretReplacements = {
+ *                  DSPA_SECRET_NAME: <DSPA SECRET NAME>,
+ *                  NAMESPACE: <PROJECT NAME>,
+ *                  AWS_ACCESS_KEY_ID: <AWS ACCESS KEY ID>,
+ *                  AWS_SECRET_ACCESS_KEY: <AWS SECRET>,
+ *               }
+ * @param yamlFilePath
+ */
+export const createDSPASecret = (
+  dspaSecretReplacements: { [key: string]: string },
```
christianvogt, line 17: Add proper types.

frontend/src/__tests__/cypress/cypress/utils/oc_commands/dspa.ts
```diff
@@ -0,0 +1,49 @@
...
+ *      Dict Structure:
+ *              dspaSecretReplacements = {
+ *                  DSPA_SECRET_NAME: <DSPA SECRET NAME>,
+ *                  NAMESPACE: <PROJECT NAME>,
+ *                  AWS_S3_BUCKET: <AWS BUCKET NAME>
+ *               }
+ * @param yamlFilePath
+ */
+export const createDSPA = (
+  dspaReplacements: { [key: string]: string },
```
christianvogt, line 40: Add proper types.

frontend/src/__tests__/cypress/cypress/utils/oc_commands/project.ts
```diff
@@ -0,0 +1,31 @@
+/**
+ * Create an Openshift Project
+ *
+ * @param projectName Project Name
+ * @param displayName Project Display Name
+ * @returns Result Object of the operation
+ */
+export const createOpenShiftProject = (projectName: string, displayName?: string) => {
+  const finalDisplayName = displayName || projectName;
+  const ocCommand = `oc new-project ${projectName} --display-name='${finalDisplayName}'`;
```
christianvogt, line 10: Perhaps better to just omit the display name if one isn't supplied instead of setting to the resource name?

frontend/src/__tests__/cypress/cypress/utils/projects.ts
```diff
@@ -0,0 +1,11 @@
...
+
+/**
+ * Filter Project by name using the Project filter from the Data Science Projects view
+ * @param projectName Project Name
+ */
+export const filterProjectByName = (projectName: string) => {
+    projectListPage.findProjectsDropdownFilter().click();
+    projectListPage.findProjectsDropdownFilterButtonByText("Name").click();
+    projectListPage.findProjectsFilterInput().type(projectName);
+};
```
christianvogt, line 11: This should be part of the project list page object as a helper function.

frontend/src/__tests__/cypress/cypress/pages/projects.ts
FedeAlonso: Fixed in [3bafb74](https://github.com/opendatahub-io/odh-dashboard/pull/3028/commits/3bafb7494eb97e74a5c537657c3242a1c73a196e)

frontend/src/__tests__/cypress/cypress/pages/projects.ts
FedeAlonso: It worked [fd7be8c](https://github.com/opendatahub-io/odh-dashboard/pull/3028/commits/fd7be8c15bf8ddea1712bc40d7b9caba6cd2f53e) 
Thanks!

frontend/src/__tests__/cypress/cypress/tests/e2e/pipelines/pipelines.cy.ts
FedeAlonso: It was waiting up to 3 minutes under the hood. But I have made it to be an argument here: [8c5da97](https://github.com/opendatahub-io/odh-dashboard/pull/3028/commits/8c5da97a11378ba311868612725265ba0bd9c2e1)

frontend/src/__tests__/cypress/cypress/utils/oc_commands/project.ts
FedeAlonso: Done [5c9f0f5](https://github.com/opendatahub-io/odh-dashboard/pull/3028/commits/5c9f0f5f97f074bed7f0e7b4888779667ffaca34)

frontend/src/__tests__/cypress/cypress/utils/projects.ts
FedeAlonso: @christianvogt I've moved it to the same file as the project lists. Is that OK or should I do something different?

3f4d1ed

FedeAlonso:
image

Test results after latest changes

frontend/src/__tests__/cypress/cypress/utils/oc_commands/dspa.ts
FedeAlonso: Fixed in [caef85f](https://github.com/opendatahub-io/odh-dashboard/pull/3028/commits/caef85f876c45971b62538a3c03ca5f91e3a3627)

frontend/src/__tests__/cypress/cypress/utils/oc_commands/dataConnection.ts
FedeAlonso: Fixed in [caef85f](https://github.com/opendatahub-io/odh-dashboard/pull/3028/commits/caef85f876c45971b62538a3c03ca5f91e3a3627)

frontend/src/__tests__/cypress/cypress/tests/e2e/pipelines/pipelines.cy.ts
FedeAlonso: Fixed in [caef85f](https://github.com/opendatahub-io/odh-dashboard/pull/3028/commits/caef85f876c45971b62538a3c03ca5f91e3a3627)

frontend/src/__tests__/cypress/cypress/utils/oc_commands/project.ts
```diff
@@ -0,0 +1,43 @@
...
+ * @param projectName OpenShift Project name
+ * @returns Result Object of the operation
+ */
+export const deleteOpenShiftProject = (
+  projectName: string,
+): Cypress.Chainable<CommandLineResult> => {
+  const ocCommand = `oc delete project ${projectName}`;
+  return cy.exec(ocCommand, { failOnNonZeroExit: false }).then((result) => {
+    return result;
+  });
```
christianvogt, line 42:
```suggestion
  return cy.exec(ocCommand, { failOnNonZeroExit: false });
```

frontend/src/__tests__/cypress/cypress/utils/oc_commands/project.ts
```diff
@@ -0,0 +1,43 @@
...
+ * @returns Result Object of the operation
+ */
+export const createOpenShiftProject = (
+  projectName: string,
+  displayName?: string,
+): Cypress.Chainable<CommandLineResult> => {
+  const finalDisplayName = displayName || projectName;
+  const ocCommand = finalDisplayName
+    ? `oc new-project ${projectName} --display-name='${finalDisplayName}'`
+    : `oc new-project ${projectName}`;
```
christianvogt, line 17: `finalDisplayName` will always be defined therefore the condition isn't necessary for the `ocCommand`.
Any particular reason to even bother setting the display name to be the same as the project name?

frontend/src/__tests__/cypress/cypress/utils/oc_commands/dspa.ts
```diff
@@ -0,0 +1,52 @@
...
+ */
+export const createDSPA = (
+  dspaReplacements: DspaReplacements,
+  yamlFilePath = 'resources/yaml/dspa.yaml',
+): Cypress.Chainable<CommandLineResult> => {
+  return cy.fixture(yamlFilePath).then((yamlContent) => {
+    const modifiedYamlContent = replacePlaceholdersInYaml(yamlContent, dspaReplacements);
+    applyOpenShiftYaml(modifiedYamlContent).then((result) => {
+      return result;
+    });
```
christianvogt, line 50: Did you intend to return the result of the apply?
```suggestion
    return applyOpenShiftYaml(modifiedYamlContent);
```

frontend/src/__tests__/cypress/cypress/pages/pipelines/topology.ts
```diff
@@ -122,6 +122,14 @@ class RunDetails extends PipelinesTopology {
     return new DetailsItem(() => cy.findByTestId(`detail-item-${key}`));
   }
 
+  private findStatusLabel(timeout = 10) {
+    return cy.get('[data-testid="status-icon"]', { timeout });
+  }
+
+  expectStatusLabelToBe(statusValue: string, timeout = 180000) {
+    this.findStatusLabel(timeout).should('have.text', statusValue);
+  }
```
christianvogt, line 131: I don't think we should be having a large timeout by default. Let the individual test cases supply larger timeouts on a per need basis. Omit the timeout to use the default.

```suggestion
  private findStatusLabel(timeout?: number) {
    return cy.get('[data-testid="status-icon"]', { timeout });
  }

  expectStatusLabelToBe(statusValue: string, timeout?: number) {
    this.findStatusLabel(timeout).should('have.text', statusValue);
  }
```

frontend/src/__tests__/cypress/cypress/pages/pipelines/topology.ts
```diff
@@ -169,8 +177,8 @@ class PipelineDetails extends PipelinesTopology {
     return new DashboardCodeEditor(() => cy.findByTestId('pipeline-dashboard-code-editor'));
   }
 
-  findPageTitle() {
-    return cy.findByTestId('app-page-title');
+  findPageTitle(timeout = 10000) {
+    return cy.findByTestId('app-page-title', { timeout });
   }
```
christianvogt, line 182: Timeouts should use the default but allow for overriding.
```suggestion
  findPageTitle(timeout?: number) {
    return cy.findByTestId('app-page-title', { timeout });
  }
```

frontend/src/__tests__/cypress/cypress/pages/projects.ts
```diff
@@ -159,6 +169,10 @@ class ProjectDetails {
     );
   }
 
+  findImportPipelineButton(timeout = 4000) {
+    return cy.findByTestId('import-pipeline-button', { timeout });
+  }
```
christianvogt, line 174: Same comment here.
```suggestion
  findImportPipelineButton(timeout?: undefined) {
    return cy.findByTestId('import-pipeline-button', { timeout });
  }
```

frontend/src/__tests__/cypress/cypress/support/e2e.ts
```diff
@@ -28,6 +28,14 @@ Cypress.Keyboard.defaults({
   keystrokeDelay: 0,
 });
 
+//TODO: Delete this once https://issues.redhat.com/browse/RHOAIENG-10719 is fixed
+Cypress.on('uncaught:exception', () => false);
```
christianvogt, line 32: Can we move this to the individual test case instead of globally?
Use `cy.on('uncaught:exception', () => false);` when in a single test case.

frontend/src/__tests__/cypress/cypress/utils/oc_commands/project.ts
FedeAlonso: Yes, that's a redundancy :S
Fixed in [2f556c2](https://github.com/opendatahub-io/odh-dashboard/pull/3028/commits/2f556c2e54428d515ef4ef2a1dc21a318f8b06e9)

frontend/src/__tests__/cypress/cypress/support/e2e.ts
FedeAlonso: I don't really see the perks of moving it from globally, as the next test could also fail due to this uncaught exception and then, the line should be added to the new test too.
Done here: [2b126d1](https://github.com/opendatahub-io/odh-dashboard/pull/3028/commits/2b126d1c2b1db16307a66a2e75cfafcd47ddcb68)

FedeAlonso:
Test results after latest changes: 
image

frontend/src/__tests__/cypress/cypress/utils/testConfig.ts
```diff
@@ -3,7 +3,12 @@ import path from 'path';
...
-import type { UserAuthConfig, TestConfig } from '~/__tests__/cypress/cypress/types';
+import type {
+  UserAuthConfig,
+  TestConfig,
+  AWSS3BucketDetails,
+  AWSS3Buckets,
+} from '~/__tests__/cypress/cypress/types';
 
 [
   `.env.cypress${env.CY_MOCK ? '.mock' : ''}.local`,
```
manosnoam, line 14: Can you add Env variable and a default value for the Pipeline S3 Bucket, maybe something like:
```
`.env.cypress${env.CY_AWS_BUCKET ? '.S3.BUCKET' : 'BUCKET_2'}.local`,
```

Then replace all `BUCKET_2` to `env.CY_AWS_BUCKET`.

This way can `export CY_AWS_BUCKET=ANY_BUCKET_NAME`, while the default value would be BUCKET_2. @christianvogt what do you think ?

manosnoam:
Hi Fede, I tried to run it on RHOAI 2.16, and it failed on:
`Timed out retrying after 180000ms: Expected to find element: `[data-testid="status-icon"]`, but never found it.`
See Job cypress/job/cypress-tests/37

FedeAlonso:
> Hi Fede, I tried to run it on RHOAI 2.16, and it failed on: `Timed out retrying after 180000ms: Expected to find element: `[data-testid="status-icon"]`, but never found it.` See Job cypress/job/cypress-tests/37

@manosnoam This is because there was no data-testid in one of the elements I needed, so I added it:
https://github.com/opendatahub-io/odh-dashboard/blob/dd738769def171b3a2800ff2be50bce069395b3c/frontend/src/concepts/pipelines/content/pipelinesDetails/PipelineDetailsTitle.tsx#L36
This is something I spoke with @christianvogt , and the way of testing it is by running the UI locally with the local changes

frontend/src/__tests__/cypress/cypress/pages/pipelines/topology.ts
```diff
@@ -122,6 +122,14 @@ class RunDetails extends PipelinesTopology {
     return new DetailsItem(() => cy.findByTestId(`detail-item-${key}`));
   }
 
+  private findStatusLabel(timeout?: number) {
+    return cy.get('[data-testid="status-icon"]', { timeout });
```
christianvogt, line 126:
```suggestion
    return cy.findByTestId('status-icon', { timeout });
```

frontend/src/__tests__/cypress/cypress/utils/oc_commands/baseCommands.ts
```diff
@@ -0,0 +1,21 @@
+import type { CommandLineResult } from '~/__tests__/cypress/cypress/types';
+
+/**
+ * Applies the given YAML content using the `oc apply` command.
+ *
+ * @param yamlContent YAML content to be applied
+ * @returns Cypress Chainable
+ */
+export const applyOpenShiftYaml = (yamlContent: string): Cypress.Chainable<CommandLineResult> => {
+  const ocCommand = `oc apply -f - <<EOF\n${yamlContent}EOF`;
```
christianvogt, line 10: Does this need an extra `\n` to be safe?
```suggestion
  const ocCommand = `oc apply -f - <<EOF\n${yamlContent}\nEOF`;
```

frontend/src/__tests__/cypress/cypress/utils/testConfig.ts
christianvogt: I'm indifferent on how to manage these types of env vars.

frontend/src/__tests__/cypress/cypress/tests/e2e/pipelines/pipelines.cy.ts
manosnoam: @FedeAlonso can you try to set `CY_AWS_BUCKET=BUCKET_2` and then call:
`AWS_DEFAULT_REGION: Buffer.from(AWS_BUCKETS[CY_AWS_BUCKET].REGION).toString('base64')`

frontend/src/__tests__/cypress/cypress/tests/e2e/pipelines/pipelines.cy.ts
```diff
@@ -0,0 +1,118 @@
...
+const testPipelineName = 'test-pipelines-pipeline';
+const testRunName = 'test-pipelines-run';
+
+describe('An admin user can import and run a pipeline', { testIsolation: false }, () => {
+  before(() => {
+    // Provision a Project
+    createOpenShiftProject(projectName);
+
+    // Create a pipeline compatible Data Connection
+    const dataConnectionReplacements: DataConnectionReplacements = {
```
Gkrumbach07, line 34: can all the pipelines setup be grouped into a helper

frontend/src/__tests__/cypress/cypress/utils/oc_commands/project.ts
```diff
@@ -0,0 +1,40 @@
...
+ * Delete an Openshift Project given its name
+ *
+ * @param projectName OpenShift Project name
+ * @returns Result Object of the operation
+ */
+export const deleteOpenShiftProject = (
+  projectName: string,
+): Cypress.Chainable<CommandLineResult> => {
+  const ocCommand = `oc delete project ${projectName}`;
+  return cy.exec(ocCommand, { failOnNonZeroExit: false });
```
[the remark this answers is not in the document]
FedeAlonso: Good catch
Fixed [6499015](https://github.com/opendatahub-io/odh-dashboard/pull/3028/commits/64990154dcee90b17eff17bd1eff62c57a819d45) 
Thanks!

manosnoam (approved):

FedeAlonso:
Test results with the latest changes: 
image

FedeAlonso:
Test results with the latest changes: 
image

PR #3029, opened 2024-07-23
Show correct state text on pipeline run node side drawer

DaoDaoNoCode:
JIRA: https://issues.redhat.com/browse/RHOAIENG-8903

## Description

We should not translate the state in the side panel because it's only used to show the PF topology node status. Instead, we should show whatever state is returned by the KFP server.

## How Has This Been Tested?

Follow the instructions in the JIRA ticket, check the node status, and make sure the text is correct and the same as the KFP UI.

## Test Impact

N/A

## Request review criteria:

If you have UI changes: 


Gkrumbach07 (approved):
Screenshot 2024-07-24 at 12 01 44 PM
shows cached now
/lgtm

PR #3033, opened 2024-07-24
[WIP] merging eslint files

rsun19:
[RHOAIENG-2534](https://issues.redhat.com/browse/RHOAIENG-2534)

## Description

merging eslint config files

## How Has This Been Tested?

Break some rules in various eslint files. It should show, and `npm run test` should work in root, backend, and frontend directories.

## Test Impact

N/A

## Request review criteria:

If you have UI changes: 


package.json
```diff
@@ -51,5 +52,22 @@
     "dotenv-expand": "^5.1.0",
     "npm-run-all": "^4.1.5",
     "prettier": "^2.2.1"
+  },
+  "devDependencies": {
+    "@typescript-eslint/eslint-plugin": "^7.1.1",
```
rsun19, line 57: For some reason node version `18.18.0` is needed :(

rsun19: resolved by moving to optionalDependencies

frontend/tsconfig.json
```diff
@@ -17,6 +17,7 @@
     "esModuleInterop": true,
     "allowSyntheticDefaultImports": true,
     "strict": true,
+    "strictNullChecks": true,
```
christianvogt, line 20: This is redundant because we use `strict: true`

.eslintrc
```diff
@@ -0,0 +1,224 @@
...
+    // "jsx-a11y/no-autofocus": ["error", { "ignoreNonDOM": true }],
+    // "jsx-a11y/anchor-is-valid": [
+    //   "error",
+    //   {
+    //     "components": ["Link"],
+    //     "specialLink": ["to"],
+    //     "aspects": ["noHref", "invalidHref", "preferButton"]
+    //   }
+    // ],
+    // "arrow-body-style": "error",
```
christianvogt, line 77: Instead of commenting out all these rules, move the common rules that we want to support from the frontend to the root. Then in the backend add the same rule as disabled if it's causing issues an we can address them later.

But for a first step this is ok.

frontend/src/concepts/pipelines/topology/usePipelineTaskTopology.tsx
```diff
@@ -152,8 +152,10 @@ const getNodesForTasks = (
         (e) =>
           e.getCustomPropertiesMap().get('task_name')?.getStringValue() === (taskName || taskId),
       );
-      if (execution) {
-        linkedArtifacts = executionLinkedArtifactMap[execution.getId()] ?? [];
+      if (execution && execution.getId() in executionLinkedArtifactMap) {
+        linkedArtifacts = executionLinkedArtifactMap[execution.getId()];
+      } else {
+        linkedArtifacts = [];
```
christianvogt, line 158: Why the change?

.eslintrc
```diff
@@ -0,0 +1,224 @@
...
+    "plugin:prettier/recommended",
+    "prettier"
+  ],
+  "ignorePatterns": [
+    "package.json",
+    "./frontend/src/third_party/mlmd",
+    "./frontend/src/__tests__/cypress/.nyc_output",
+    "./frontend/src/__tests__/cypress/coverage",
+    "./frontend/package.json",
+    "./frontend/config"
```
christianvogt, line 28: Can these stay in frontend or does it end up overriding the root level ignore patterns?

rsun19:
@christianvogt made requested changes. For this to work, we will need to notify people to get rid of their working directory eslint setting, since our combined eslintrc file seems to confuse the vscode eslint extension.

PR #3034, opened 2024-07-25
Allow all users to list model registries

manaswinidas:
Closes: [RHOAIENG-9332](https://issues.redhat.com/browse/RHOAIENG-9332)

## Description

List model registry services for non-admin users as well as admin users. We are fetching services instead of modelregistries. For admin users, fetching a list of services in the odh-model-registries namespace, using a label selector for `component=model-registry` and for non-admin users: 
- Create a `SelfSubjectRulesReview` and dig out the list of service names we can see from its response
- In a Promise.all, fetch each of those services by name
- `.filter()` that resulting array by the label `component=model-registry` , and pass that filtered array down through context

Screenshot 2024-07-25 at 7 01 18 PM

## How Has This Been Tested?

1. Log in as a `regularuser` and Model Registry page still loads with data

## Test Impact

1. Fixed tests to use `ServiceModel` instead of `ModelRegistryModel`
2. Added unit and Cypress tests

## Request review criteria:

If you have UI changes: 


emilys314:
I checked out your changes in my branch, but I seem to be getting an error in the model registry page. It might be related to the failing tests though?

<details>
<summary>screenshot</summary>

image

</details>

mturley:
@emilys314 are you running the frontend with `npm run start:dev:ext`? That will use the backend on the cluster and take care of port-forwarding for the `modelregistry-sample` registry's API for you. If you run the backend locally you need to run additional port forwarding commands for model registry to work, but for PRs like this with no backend changes we can just use `start:dev:ext`

frontend/src/concepts/modelRegistry/apiHooks/useModelRegistries.ts
```diff
@@ -1,11 +1,65 @@
...
+  const [loaded, statuses] = useRulesReview(MODEL_REGISTRY_DEFAULT_NAMESPACE);
+  const serviceNames = React.useMemo(() => {
+    if (!loaded) {
+      return [];
+    }
+    return (
+      statuses?.resourceRules.find(
+        ({ resources, verbs }) =>
+          resources?.includes(ServiceModel.plural) && verbs.some((verb) => verb === 'get'),
+      )?.resourceNames || []
```
mturley, line 31: Ok @manaswinidas I apologize, I was wrong about the logic here. After granting permission to regularuser1 to see dallas-mr on our shared cluster, I found that when I log in as regularuser1 only dallas-mr is appearing in the dropdown and modelregistry-sample disappeared. I inspected the SelfSubjectRulesReview status we get back and I see the culprit here -- It's not just one item in  `resourceRules` that includes multiple service names, each appears to get its own item there:

Screenshot 2024-07-26 at 12 01 03 PM

I'm pretty sure that if you just replace the `.find(` with `.flatMap(` on line 28 here you'll get the right result.

frontend/src/concepts/modelRegistry/apiHooks/useModelRegistries.ts
```diff
@@ -1,11 +1,65 @@
...
+    adminServices,
+    nonAdminServices,
+  ]);
+
+  return useFetchState<ServiceKind[]>(getModelRegistryServices, [], {
+    initialPromisePurity: true,
+  });
+};
+
+export default useModelRegistryServices;
```
mturley, line 65: We should rename the file `useModelRegistryServices.ts` and make sure its imports are using it by the new name as well (it's still called `useModelRegistries` in `ModelRegistrySelectorContext`).

mturley (changes requested):
A couple of things we need to address here. Once the logic settles we should also add some unit tests for `useModelRegistryServices` and maybe a Cypress test that asserts we are not calling the API endpoints for ModelRegistry CRs when we're a regular user.

emilys314:
Okay so with running only the frontend and as admin I can see everything. Logged in as reguleruser1, I'm only seeing "Dallas MR" from the dropdown 
image

frontend/src/concepts/modelRegistry/context/useModelRegistryNamespaceCR.ts
```diff
@@ -31,6 +31,10 @@ export const useModelRegistryNamespaceCR = (namespace: string, name: string): Fe
           // Not finding is okay, not an error
           return null;
         }
+        if (e.statusObject?.code === 403) {
+          // Forbidden is okay, not an error
+          return null;
+        }
```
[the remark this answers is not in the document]
mturley: [I brought this up on a Slack thread here](https://redhat-internal.slack.com/archives/C06EJG57Z0X/p1722010483074919). The conclusion was that if we confirm I'm right that the `serverTimedOut` stuff is really not used anywhere in the ModelRegistryContext consumers, we should remove it and its related code there, and then since it will no longer call `useModelRegistryNamespaceCR` we can revert the change you have here @manaswinidas so that 403s are still thrown on the admin settings page.

mturley:
@emilys314 that's expected if regularuser1 hasn't been given access to "Dallas MR" by the admin under Model Registry Settings -> "manage permissions". I changed that on the cluster, can you check again?

frontend/src/concepts/modelRegistry/apiHooks/useModelRegistryServices.ts
```diff
@@ -0,0 +1,87 @@
...
+  const services = allowList
+    ? await listServices(MODEL_REGISTRY_DEFAULT_NAMESPACE)
+    : await fetchServices(serviceNames || [], MODEL_REGISTRY_DEFAULT_NAMESPACE);
+
+  return services;
+};
+
+export const useModelRegistryServices = (): FetchState<ServiceKind[]> => {
+  const [allowList, accessReviewLoaded] = useAccessReview(accessReviewResource);
+  const [rulesReviewLoaded, statuses] = useRulesReview(MODEL_REGISTRY_DEFAULT_NAMESPACE);
```
mturley, line 59: Nit, take it or leave it - I wonder if we should swap the order of the booleans returned by `useRulesReview` so it's consistent with `useAccessReview`?

mturley (commented):
New logic looks perfect!

frontend/src/concepts/modelRegistry/apiHooks/useModelRegistryServices.ts
```diff
@@ -0,0 +1,87 @@
...
+  return services;
+};
+
+const listServicesOrFetchThemByNames = async (
+  allowList: boolean,
+  accessReviewLoaded: boolean,
+  rulesReviewLoaded: boolean,
+  serviceNames?: string[],
+): Promise<ServiceKind[]> => {
+  if (!accessReviewLoaded && !rulesReviewLoaded) {
```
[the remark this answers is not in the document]
mturley: Copying from Slack: the solution here is to switch this back to `||` and then in `useModelRegistryServices.spec.ts` where you mock the return value of `useRulesReview`, give it a true response with a valid but empty status object:

```ts
    (useRulesReview as jest.Mock).mockReturnValue([
      true,
      { incomplete: false, nonResourceRules: [], resourceRules: [] },
    ] satisfies ReturnType<typeof useRulesReview>);
```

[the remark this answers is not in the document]
manaswinidas: Done. Thanks.

frontend/src/concepts/modelRegistry/apiHooks/useModelRegistryServices.ts
manaswinidas: Done.

mturley (approved):
LGTM! Great stuff, thanks for all the hard work here @manaswinidas

PR #3036, opened 2024-07-26
[RHOAIENG-10454] mock cypress test flake: pipelines.cy.ts 'imports a new pipeline'

jpuzz0:
[RHOAIENG-10454](https://issues.redhat.com/browse/RHOAIENG-10454)

## Description
The reloading of pipelines after uploading a new pipeline in the 2 tests updated here was causing the page to reload and in some cases that reload would impede on the redirect to the pipeline details page.

## Test Impact
2 cypress tests were updated to address the issue

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/__tests__/cypress/cypress/pages/pipelines/topology.ts
+++ b/frontend/src/__tests__/cypress/cypress/pages/pipelines/topology.ts
@@ -1,6 +1,10 @@
 import { Contextual } from '~/__tests__/cypress/cypress/pages/components/Contextual';
 import { DashboardCodeEditor } from '~/__tests__/cypress/cypress/pages/components/DashboardCodeEditor';
-import type { PipelineRecurringRunKFv2 } from '~/concepts/pipelines/kfTypes';
+import type {
+  PipelineKFv2,
+  PipelineRecurringRunKFv2,
+  PipelineVersionKFv2,
+} from '~/concepts/pipelines/kfTypes';
 
 class TaskDrawer extends Contextual<HTMLElement> {
   findInputArtifacts() {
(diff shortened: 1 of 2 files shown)


Gkrumbach07 (approved):
/lgtm

christianvogt:
@jpuzz0 The test still ends on the table and not the details page, although now without waiting for the refresh, we don't see the new entry.
image

When running locally, the browser navigates very briefly to the details page before redirecting back to the table. 

I suspect that this will continue to be a flakey test.

Why is the final page not the details page?

jpuzz0:
@christianvogt 

Every time I upload a pipeline via url or yaml in dev, I see the redirect happen, and I've run these 2 tests successfully locally ~10 times in a row, and I am seeing a redirect happen right after the upload for both, why the page reloads in the test I'm not sure, but the assertion is no longer interrupted by refreshing the pipelines fetched.

Since the expectation is that the page redirects and not that the new resource exists in the table, that new resource showing in the table shouldn't matter IMO as the user wouldn't see that until navigating back to the list page anyways.

christianvogt:
Agreed that we do not need to wait for the refresh to return. However the test should end on the details page if that's the next page to be asserted. It's still a race condition but now has a smaller chance to occur.

jpuzz0:
> Agreed that we do not need to wait for the refresh to return. However the test should end on the details page if that's the next page to be asserted. It's still a race condition but now has a smaller chance to occur.

@christianvogt should be updated now. Both of these tests land on the pipeline details page now and do not redirect afterwards.

christianvogt:
Ran locally to confirm tests pass and land now on the correct page.

/lgtm
/approve

PR #3045, opened 2024-07-30
define the connection type config map typescript types

christianvogt:
https://issues.redhat.com/browse/RHOAIENG-10310

## Description

Defines the typescript types used to describe connection types. 
A connection type is stored in a config map. 

A connection type is an ordered collection of field descriptors. Each descriptor typed by a unique string identifier which identifies the corresponding field type. The field type is used to associate a rendered which knows how to consume their specific properties.

Note. The current field types are based on a non-finalized version of the UX. Therefore the fields and field properties are subject to change. However the general structure will remain the same.

Added mock util to be used for early development and testing.

Added serialization/deserialization utils to convert a configmap with `string` data into a json object and back. These utils will be used when post fetch and before sending a request to the backend.

## How Has This Been Tested?

Manually.
Added test cases for serialization utils using mock data.

## Test Impact

N/A

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/__mocks__/mockConnectionType.ts
+++ b/frontend/src/__mocks__/mockConnectionType.ts
@@ -1,0 +1,393 @@
+import {
+  ConnectionTypeConfigMap,
+  ConnectionTypeConfigMapObj,
+  ConnectionTypeField,
+} from '~/concepts/connectionTypes/types';
+import { toConnectionTypeConfigMap } from '~/concepts/connectionTypes/utils';
+
+type MockConnectionTypeConfigMap = {
+  name?: string;
+  namespace?: string;
+  displayName?: string;
+  description?: string;
+  enabled?: boolean;
+  username?: string;
+  preInstalled?: boolean;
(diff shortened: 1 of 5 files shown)


christianvogt:
@andrewballantyne we use annotations for storing display name and descriptions for k8s resources. While that approach continues to work here, it could be argued that these properties should be part of the connection type definition and therefore live within the config map data itself. I'm indifferent to either approach. Do you have an opinion?

frontend/src/api/connectionTypes/types.ts
andrewballantyne, line 1: Can we put this in concepts? It's going to be the type we use and destrcture everywhere -- it's not just the API transfer structure

andrewballantyne (changes requested):
One request, we make a concepts area for this effort. Everything else looks good.

frontend/src/api/connectionTypes/types.ts
christianvogt: ha, i had it there first and 2nd guessed the location

PR #3046, opened 2024-07-30
[RHOAIENG-7485] Admin - Database status in Model Registry Table

jpuzz0:
https://issues.redhat.com/browse/RHOAIENG-7485

## Description
Show various statuses depending on model registry conditions within a new column labeled "Status" for the model registry list table page.

image

**Degrading popover**
image

**Unavailable with multiple messages**
image
(cc @yih-wang)

## How Has This Been Tested?
Manually tested with sample data, created unit tests with possible status scenarios

## Test Impact
Added unit tests

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/pages/modelRegistrySettings/ModelRegistryTableRowStatus.tsx
+++ b/frontend/src/pages/modelRegistrySettings/ModelRegistryTableRowStatus.tsx
@@ -1,0 +1,157 @@
+import React from 'react';
+
+import { Label, Popover, Stack, StackItem } from '@patternfly/react-core';
+import {
+  CheckCircleIcon,
+  DegradedIcon,
+  ExclamationCircleIcon,
+  ExclamationTriangleIcon,
+  InProgressIcon,
+} from '@patternfly/react-icons';
+
+import { K8sCondition } from '~/k8sTypes';
+
+enum ModelRegistryStatus {
+  Progressing = 'Progressing',
+  Degraded = 'Degraded',
+  Available = 'Available',
+  IstioAvailable = 'IstioAvailable',
+  GatewayAvailable = 'GatewayAvailable',
+}
+
+enum ModelRegistryStatusLabel {
+  Progressing = 'Progressing',
+  Available = 'Available',
+  Degrading = 'Degrading',
+  Unavailable = 'Unavailable',
+}
+
+enum ConditionStatus {
+  True = 'True',
+  False = 'False',
+}
+
+interface ModelRegistryTableRowStatusProps {
+  conditions: K8sCondition[] | undefined;
+}
+
+export const ModelRegistryTableRowStatus: React.FC<ModelRegistryTableRowStatusProps> = ({
+  conditions,
+}) => {
+  const conditionsMap =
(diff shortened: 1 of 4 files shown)


yih-wang:
We should use the [Alert popover](https://www.patternfly.org/components/popover#alert-popover) to display messages. Use variant Danger for `Unavailable` and variant Warning (or possibly Custom if the warning icon is hard-coded) for `Degrading`.

frontend/src/pages/modelRegistrySettings/ModelRegistryTableRowStatus.tsx
```diff
@@ -0,0 +1,157 @@
...
+      [ModelRegistryStatus.Progressing]: progressCondition,
+      [ModelRegistryStatus.Degraded]: degradedCondition,
+    } = conditionsMap;
+    const lastAvailableConditionTime = new Date(
+      availableCondition?.lastTransitionTime ?? '',
+    ).getTime();
+
+    popoverMessages =
+      availableCondition?.status === ConditionStatus.False
+        ? Object.values(conditionsMap).reduce((messages: string[], condition) => {
```
mturley, line 64: We want to show the messages from any conditions with `ConditionStatus.False` even if they aren't in the conditionsMap, since there may be additional condition types added in the future that affect availability. I think instead of reducing `Object.values(conditionsMap)` you can just reduce the whole `conditions` array here.

mturley (changes requested):
Just one small nit, everything else LGTM

frontend/src/pages/modelRegistrySettings/ModelRegistryTableRowStatus.tsx
jpuzz0: `conditionsMap` uses all of the `conditions`, so I'm confused as to why this is a desired change.

mturley: Nevermind, I misread the code. Thanks.

frontend/src/pages/modelRegistrySettings/ModelRegistryTableRowStatus.tsx
```diff
@@ -0,0 +1,157 @@
...
+    // Available
+    if (availableCondition?.status === ConditionStatus.True) {
+      statusLabel = ModelRegistryStatusLabel.Available;
+      icon = <CheckCircleIcon />;
+      color = 'green';
+    }
+    // Progressing
+    else if (
+      progressCondition?.status === ConditionStatus.True &&
+      lastAvailableConditionTime <= new Date(progressCondition.lastTransitionTime ?? '').getTime()
```
mturley, line 82: Testing this on the modelregistry-ui cluster, the registry "Dallas MR" has a Progressing condition that is the exact same time as its Available:False condition, and it's been stuck that way for days. I think we should change this `<=` to a `<`, we only want something to take precedence over Available:False if it's actually newer. Same goes for the similar comparison on line 91 for Degrading.

mturley (approved):
LGTM! Thanks for the help!

PR #3048, opened 2024-07-31
[RHOAIENG-10311] Backend service endpoints for connection types

jeff-phillips-18:
[RHOAIENG-10311](https://issues.redhat.com/browse/RHOAIENG-10311)

## Description
Adds backend endpoints for ConnectionType objects.
Includes GET, POST, PUSH, PATH, DELETE.
Adds a frontend service to access these endpoints.

## How Has This Been Tested?
This has been tested locally by adding temporary calls in the UI

## Test Impact
Tests will need to be added when the UI accesses these new functions.

## Request review criteria:

### DIFF
--- a/backend/src/routes/api/connection-types/connectionTypeUtils.ts
+++ b/backend/src/routes/api/connection-types/connectionTypeUtils.ts
@@ -1,0 +1,190 @@
+import { PatchUtils, V1ConfigMap } from '@kubernetes/client-node';
+import { KnownLabels, KubeFastifyInstance, RecursivePartial } from '../../../types';
+import { getNamespaces } from '../../../utils/notebookUtils';
+import { errorHandler } from '../../../utils';
+
+const isConnectionTypeConfigMap = (configMap: V1ConfigMap): boolean =>
+  configMap.metadata.labels &&
+  configMap.metadata.labels[KnownLabels.DASHBOARD_RESOURCE] === 'true' &&
+  configMap.metadata.labels[KnownLabels.CONNECTION_TYPE] === 'true';
+
+const isExistingConnectionType = async (
+  fastify: KubeFastifyInstance,
+  name: string,
+): Promise<boolean> => {
+  const coreV1Api = fastify.kube.coreV1Api;
+  const { dashboardNamespace } = getNamespaces(fastify);
+
+  const response = await coreV1Api.readNamespacedConfigMap(name, dashboardNamespace);
+  return isConnectionTypeConfigMap(response.body);
+};
+
+export const listConnectionTypes = async (fastify: KubeFastifyInstance): Promise<V1ConfigMap[]> => {
+  const { dashboardNamespace } = getNamespaces(fastify);
+  const coreV1Api = fastify.kube.coreV1Api;
+  const connectionTypes: V1ConfigMap[] = [];
+
+  let _continue: string = undefined;
+  let remainingItemCount = 1;
+  try {
+    while (remainingItemCount) {
+      const response = await coreV1Api.listNamespacedConfigMap(
+        dashboardNamespace,
+        undefined,
+        undefined,
+        _continue,
+        undefined,
+        `${KnownLabels.DASHBOARD_RESOURCE} = true, ${KnownLabels.CONNECTION_TYPE} = true`,
+      );
+      connectionTypes.push(...(response?.body?.items ?? []));
+      remainingItemCount = response?.body.metadata?.remainingItemCount;
+      _continue = response?.body.metadata?._continue;
+    }
+    return connectionTypes;
+  } catch (e) {
+    fastify.log.error(`Error fetching configmaps for connection types: `, e);
+    throw new Error(`Failed to list connection types: ${errorHandler(e)}.`);
+  }
+};
+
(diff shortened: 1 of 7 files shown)


frontend/src/services/connectionTypesService.ts
```diff
@@ -0,0 +1,63 @@
...
+
+export const createConnectionType = (connectionType: ConfigMapKind): Promise<ResponseStatus> => {
+  const url = `/api/connection-types`;
+  return axios
+    .post(url, connectionType)
+    .then((response) => response.data)
+    .catch((e) => {
+      throw new Error(e.response.data.message);
+    });
+};
```
christianvogt, line 33: Can we do the conversion between types? Do you think it makes sense that in app we always work with the parsed object. I don't know of a time when we would need to use the raw config map.

```suggestion
export const fetchConnectionType = (name: string): Promise<ConnectionTypeConfigMapObj> => {
  const url = `/api/connection-types/${name}`;
  return axios
    .get(url)
    .then((response) => toConnectionTypeConfigMapObj(response.data))
    .catch((e) => {
      throw new Error(e.response.data.message);
    });
};

export const createConnectionType = (connectionType: ConnectionTypeConfigMap): Promise<ResponseStatus> => {
  const url = `/api/connection-types`;
  return axios
    .post(url, toConnectionTypeConfigMap(connectionType))
    .then((response) => response.data)
    .catch((e) => {
      throw new Error(e.response.data.message);
    });
};
```

backend/src/routes/api/connection-types/connectionTypeUtils.ts
```diff
@@ -0,0 +1,191 @@
...
+    const validConnectionType = await isExistingConnectionType(fastify, name);
+    if (!validConnectionType) {
+      const error = `Unable to delete connection type, object is not a connection type`;
+      fastify.log.error(error);
+      return { success: false, error };
+    }
+    await coreV1Api.deleteNamespacedConfigMap(name, dashboardNamespace);
+    return { success: true, error: '' };
+  } catch (e) {
+    const error = `Unable to delete connection type: ${e.message}.`;
```
christianvogt, line 187: Use `errorHandler` as we will eventually have stricter ts / eslint rules.
```suggestion
    const error = `Unable to delete connection type: ${errorHandler(e)}.`;
```

backend/src/routes/api/connection-types/connectionTypeUtils.ts
```diff
@@ -0,0 +1,191 @@
...
+import { getNamespaces } from '../../../utils/notebookUtils';
+
+const ODH_DASHBOARD_KEY = 'opendatahub.io/dashboard';
+const CONNECTION_TYPE_KEY = 'opendatahub.io/connection-type';
+
+const CONNECTION_TYPE_LABEL_SELECTOR = `${ODH_DASHBOARD_KEY} = true, ${CONNECTION_TYPE_KEY} = true`;
+
+const isConnectionTypeConfigMap = (configMap: V1ConfigMap): boolean => {
+  const labels = configMap.metadata.labels;
+  return labels[ODH_DASHBOARD_KEY] === 'true' || labels[CONNECTION_TYPE_KEY] === 'true';
```
christianvogt, line 12: Can't labels be undefined? Also we need both labels to be present.
```suggestion
  return labels?.[ODH_DASHBOARD_KEY] === 'true' && labels?.[CONNECTION_TYPE_KEY] === 'true';
```

frontend/src/services/connectionTypesService.ts
```diff
@@ -0,0 +1,78 @@
...
+  toConnectionTypeConfigMap,
+  toConnectionTypeConfigMapObj,
+} from '~/concepts/connectionTypes/utils';
+
+export const fetchConnectionTypes = (): Promise<ConnectionTypeConfigMapObj[]> => {
+  const url = `/api/connection-types`;
+  return axios
+    .get(url)
+    .then((response) =>
+      response.data.map((cm: ConnectionTypeConfigMap) => toConnectionTypeConfigMapObj(cm)),
```
christianvogt, line 17: I believe this will be inside `data.items`.
```suggestion
      response.data.items.map((cm: ConnectionTypeConfigMap) => toConnectionTypeConfigMapObj(cm)),
```

frontend/src/services/connectionTypesService.ts
```diff
@@ -0,0 +1,78 @@
...
+    .put(url, toConnectionTypeConfigMap(connectionType))
+    .then((response) => response.data)
+    .catch((e) => {
+      throw new Error(e.response.data.message);
+    });
+};
+
+export const patchConnectionType = (
+  connectionType: ConnectionTypeConfigMapObj,
+): Promise<ResponseStatus> => {
```
christianvogt, line 60: Let's omit this generic patch function and add a specific one that constructs a patch just for setting the `enabled` annotation:
```
             {
                op: 'replace',
                path: '/metadata/annotations/opendatahub.io~1enabled',
                value: ...,
              }
```

frontend/src/services/connectionTypesService.ts
christianvogt: nm! the backend point returns a list

PR #3050, opened 2024-08-01
test: Add a dummy pipeline sample

FedeAlonso:
## Description
Add a sample pipeline to our repo, so we don't need to depend on other repos when testing the "import pipeline" feature.
This is a dummy pipeline, in order to run an e2e in a live cluster as fast as possible

## How Has This Been Tested?
In a 2.13 live cluster
image

## Test Impact
NA
This is going to be used in e2e testing

## Request review criteria:
Import the dummy_compiled.yaml pipeline and run it
The folder structure is good?


manosnoam (commented):
lgtm
I would rename these pipeline test files like:
"dummy_pipeline.py"
"dummy_pipeline_compiled.yaml"

PR #3056, opened 2024-08-01
connection type preview form generation

christianvogt:
https://issues.redhat.com/browse/RHOAIENG-10319

## Description

Generates a preview form from the connection type resource and field descriptors.
Uses the existing mock which encompasses all variations of fields.

Form fields are rendered in their default states but do not react to changes so they act as read only.

Provides a drawer component to render the content preview given the connection type resource.

There are still several UX discussions on going but any changes should be minimal to incorporate.
There's also a quirk with dropdowns that is causing the drawer to scroll back to the top whenever it is opened. This can be dealt with later.

image
image
image
image
image
image
image
image
image
image

## How Has This Been Tested?

Manually. Test pages.
Added unit tests.

## Test Impact

Unit tests.

## Request review criteria:

If you have UI changes: 

cc @simrandhaliw

### DIFF
--- a/frontend/src/concepts/connectionTypes/fields/DropdownFormField.tsx
+++ b/frontend/src/concepts/connectionTypes/fields/DropdownFormField.tsx
@@ -1,0 +1,88 @@
+import * as React from 'react';
+import { Badge, MenuToggle, Select, SelectList, SelectOption } from '@patternfly/react-core';
+import { DropdownField } from '~/concepts/connectionTypes/types';
+import DataFormFieldGroup from '~/concepts/connectionTypes/fields/DataFormFieldGroup';
+
+type Props = {
+  field: DropdownField;
+  isPreview?: boolean;
+  onChange?: (value: string[]) => void;
+  value?: string[];
+};
+
+const DropdownFormField: React.FC<Props> = ({ field, isPreview, onChange, value }) => {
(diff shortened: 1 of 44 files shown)


christianvogt:
@emilys314 @jeff-phillips-18 rebased this PR

emilys314:
Looks good to me I think

frontend/src/concepts/connectionTypes/fields/DropdownFormField.tsx
```diff
@@ -0,0 +1,88 @@
...
+import DataFormFieldGroup from '~/concepts/connectionTypes/fields/DataFormFieldGroup';
+
+type Props = {
+  field: DropdownField;
+  isPreview?: boolean;
+  onChange?: (value: string[]) => void;
+  value?: string[];
+};
+
+const DropdownFormField: React.FC<Props> = ({ field, isPreview, onChange, value }) => {
```
jeff-phillips-18, line 13: Nit: Calling it a Dropdown but using a Select could be confusing.

christianvogt: That's fair. It coincides with the type. And the type name will be shown to the user as an available type.

Perhaps we should discuss the type name with us?

PR #3057, opened 2024-08-02
Admin - Model Registry RBAC Management Projects

ppadti:
Closes: [RHOAIENG-8632](https://issues.redhat.com/browse/RHOAIENG-8632)

## Description
This PR aims to add projects tab under MR Permissions.

Screenshot from 2024-08-08 18-39-47

Screenshot from 2024-08-08 16-59-38

Screenshot from 2024-08-08 17-55-41

## How Has This Been Tested?
In MR permission, click on Projects tab
create/ edit/ delete rolebindings for projects.

## Test Impact
Added cypress tests.

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/__tests__/cypress/cypress/tests/mocked/modelRegistrySettings/modelRegistryPermissions.cy.ts
+++ b/frontend/src/__tests__/cypress/cypress/tests/mocked/modelRegistrySettings/modelRegistryPermissions.cy.ts
@@ -1,17 +1,18 @@
-import { mockK8sResourceList } from '~/__mocks__';
+import { mockK8sResourceList, mockProjectK8sResource } from '~/__mocks__';
 import { mock200Status } from '~/__mocks__/mockK8sStatus';
 import { mockRoleBindingK8sResource } from '~/__mocks__/mockRoleBindingK8sResource';
 import { be } from '~/__tests__/cypress/cypress/utils/should';
 import {
   GroupModel,
   ModelRegistryModel,
+  ProjectModel,
   RoleBindingModel,
 } from '~/__tests__/cypress/cypress/utils/models';
 import type { RoleBindingSubject } from '~/k8sTypes';
 import { asProductAdminUser, asProjectEditUser } from '~/__tests__/cypress/cypress/utils/mockUsers';
 import { mockModelRegistry } from '~/__mocks__/mockModelRegistry';
 import { mockGroup } from '~/__mocks__/mockGroup';
-import { usersTab } from '~/__tests__/cypress/cypress/pages/modelRegistryPermissions';
+import { modelRegistryPermissions } from '~/__tests__/cypress/cypress/pages/modelRegistryPermissions';
 
 const MODEL_REGISTRY_DEFAULT_NAMESPACE = 'odh-model-registries';
 
@@ -28,6 +29,14 @@ const groupSubjects: RoleBindingSubject[] = [
     kind: 'Group',
     apiGroup: 'rbac.authorization.k8s.io',
     name: 'example-mr-users',
+  },
+];
+
+const projectSubjects: RoleBindingSubject[] = [
+  {
+    kind: 'Group',
+    apiGroup: 'rbac.authorization.k8s.io',
+    name: 'system:serviceaccounts:test-project',
   },
 ];
 
@@ -51,6 +60,13 @@ const initIntercepts = ({ isEmpty = false, hasPermission = true }: HandlersProps
   cy.interceptK8sList(
     { model: GroupModel },
     mockK8sResourceList([mockGroup({ name: 'example-mr-group-option' })]),
+  );
+  cy.interceptK8sList(
+    ProjectModel,
+    mockK8sResourceList([
+      mockProjectK8sResource({}),
+      mockProjectK8sResource({ k8sName: 'project-name', displayName: 'Project' }),
+    ]),
   );
   cy.interceptK8sList(
     { model: RoleBindingModel, ns: MODEL_REGISTRY_DEFAULT_NAMESPACE },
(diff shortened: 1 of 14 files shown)


frontend/src/pages/modelRegistrySettings/ProjectsTab/ProjectsSettingsTab.tsx
```diff
@@ -0,0 +1,179 @@
...
+            namespace={projectName}
+            roleRefKind="Role"
+            isProject={isProject}
+            roleRefName={roleRefName}
+            labels={labels}
+            typeAhead={
+              filteredProjects.length > 0
+                ? filteredProjects.map((project) => project.metadata.name)
+                : undefined
+            }
```
mturley, line 135: So as you have this implemented, the options in the dropdown are project names (raw namespaces) rather than project display names.

It's going to be tricky... but we should show the display name of projects in the dropdown and the non-edit-mode cell of the table instead of the `project.metadata.name`.

One way to do that might be to add an optional `getSubjectDisplayName` callback prop that takes a string (the `project.metadata.name`) and returns the project's display name. You'd define it here since you already have projects in memory and you could do the lookup here (by finding the matching project object and passing it to `getDisplayNameFromK8sResource` imported from `~/concepts/k8s/utils`). Then you'd feed that `getSubjectDisplayName` prop all the way down to where your select options are rendered in `RoleBindingPermissionsNameInput` and use it to render the display name as children of `<SelectOption>`, and also use it in the non-editing mode of `RoleBindingPermissionsTableRow` instead of directly displaying `roleBindingName`. If it's not passed in it could just default to `roleBindingName`.

mturley (changes requested):
Hey @ppadti. This is looking good! Just a few things to address.

frontend/src/__tests__/cypress/cypress/tests/mocked/modelRegistrySettings/modelRegistryPermissions.cy.ts
```diff
@@ -368,17 +387,143 @@ describe('MR Permissions', () => {
...
+      projectTable.findTableHeaderButton('Date added').should(be.sortDescending);
+    });
+
+    it('Add project', () => {
+      cy.interceptK8s(
+        'POST',
+        RoleBindingModel,
+        mockRoleBindingK8sResource({
+          namespace: MODEL_REGISTRY_DEFAULT_NAMESPACE,
+          subjects: [{ ...projectSubjects[0], name: 'project-1' }],
```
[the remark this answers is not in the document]
ppadti: ah yes, I missed this...

frontend/src/concepts/roleBinding/utils.ts
```diff
@@ -5,7 +5,15 @@ import { RoleBindingPermissionsRBType, RoleBindingPermissionsRoleType } from './
...
   type: RoleBindingPermissionsRBType,
-): RoleBindingKind[] => roleBindings.filter((roles) => roles.subjects[0]?.kind === type);
+  isProject?: boolean,
+): RoleBindingKind[] =>
+  roleBindings.filter(
+    (roles) =>
+      roles.subjects[0]?.kind === type &&
+      (isProject
+        ? roles.subjects[0].name.startsWith('system:serviceaccounts:')
+        : !roles.subjects[0].name.startsWith('system:serviceaccounts:')),
```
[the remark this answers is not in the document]
ppadti: Yeah I was worried about this but I wasn't aware that we can add labels to solve this. Apologies, I forgot to add this as comment when raised a PR. Thanks!! I will update this.

frontend/src/pages/modelRegistrySettings/ProjectsTab/ProjectsSettingsTab.tsx
ppadti: oh sorry. I didn't notice that. I will update this.

frontend/src/concepts/roleBinding/utils.ts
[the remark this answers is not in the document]
mturley: No apology necessary! This is great work, thanks!

frontend/src/pages/modelRegistrySettings/ProjectsTab/ProjectsSettingsTab.tsx
mturley: No worries! Thanks for the hard work here.

frontend/src/concepts/roleBinding/RoleBindingPermissionsTable.tsx
```diff
@@ -40,12 +42,31 @@ const RoleBindingPermissionsTable: React.FC<RoleBindingPermissionsTableProps> =
...
+  );
+
+  const getSubjectResourceName = React.useCallback(
+    (name: string): string =>
+      projects.find(
+        (project) => project.metadata.annotations?.['openshift.io/display-name'] === name,
+      )?.metadata.name || name,
+    [projects],
+  );
+
```
mturley, line 69: Sorry to give you one more round of changes --

`getSubjectDisplayName` and `getSubjectResourceName` only apply to projects, and every usage of them is behind `isProjectSubject &&`. I think now that we're defining them in here instead of lifting them out, we're losing the benefit of the abstraction and we have a lot of unnecessary prop-drilling. We already have `isProjectSubject` drilled down into each of the places we're using these, and those components could just consume `ProjectsContext` directly.

I think maybe what we should do is change these to static utils - maybe add them to `frontend/src/concepts/projects/utils.ts`, make them take the projects array and name them more explicitly. You can also use `getDisplayNameFromK8sResource` from `~/concepts/k8s/utils` for the first one.

Maybe something like this:

```ts
export const namespaceToProjectDisplayName = (
  namespace: string,
  projects: ProjectKind[],
): string => {
  const project = projects.find((p) => p.metadata.name === namespace);
  return project ? getDisplayNameFromK8sResource(project) : namespace;
};

export const projectDisplayNameToNamespace = (
  displayName: string,
  projects: ProjectKind[],
): string => {
  const project = projects.find(
    (p) => p.metadata.annotations?.['openshift.io/display-name'] === displayName,
  );
  return project?.metadata.name || displayName;
};
```

And then you could get rid of all the props where you pass around `getSubjectDisplayName` and `getSubjectResourceName`, and instead consume ProjectsContext as close to where you call these as possible so you have the projects array to pass into them. I think that would be less prop drilling?

---

Now that I write all that, I'm questioning it. If you disagree with the value of that refactor, can you instead just keep what you have but rename `getSubjectDisplayName` to `getProjectDisplayName` and `getSubjectResourceName` to `getProjectResourceName`? I think that would at least solve the readability.

I'm fine with either solution, I leave it up to you. Thanks for bearing with my rambling.

mturley (changes requested):
One more thing --- everything is working great here, I just think I may have steered you wrong in terms of code readability.

yih-wang:
The dropdown is missing the placeholder text `Select or enter a project`. It's important for users to know that they can either select from the menu items or free type a new name.

frontend/src/concepts/roleBinding/RoleBindingPermissionsTable.tsx
ppadti: Yeah, this makes more sense. I have updated this. Thanks!!

ppadti:
> The dropdown is missing the placeholder text `Select or enter a project`. It's important for users to know that they can either select from the menu items or free type a new name.
@yih-wang I have updated the screenshots above. Please have a look. Thanks for pointing this out.

mturley (approved):
LGTM! works great and reads well. Thanks for the iteration on this.

gitdallas (approved):

PR #3064, opened 2024-08-05
[RHOAIENG-8816] Automatic discovery of Accelerator profile handles AMD GPUs like they were NVIDIA

jpuzz0:
https://issues.redhat.com/browse/RHOAIENG-8816

## Description
Only create accelerator when nvidia type nodes are detected from the server's list of nodes.

## How Has This Been Tested?
Manually tested. This doesn't seem to be re-creatable in dev without some hard-coded intervention. The condition to create the accelerator has been adjusted to make sure nvidia nodes exist, however I was not able to simulate the "configured" state being true, which is the other part of the condition.

### Test steps
Testing this is not straight forward and requires deleting resources in the openshift console and restarting the backend afterwards + manually editing the condition to bypass the "configured" state. I believe this change is one we can have the reporter of the issue test once released to be sure this resolves his issue.

1. Update https://github.com/opendatahub-io/odh-dashboard/blob/3f026963c167bc011c8aa3ab00f5c4a762d6b0e7/backend/src/utils/resourceUtils.ts#L684 locally and remove the `acceleratorDetected.configured` portion of the condition.
2. Stop your locally running dev "backend"
3. Delete console resources:
Pasted Graphic 2
Pasted Graphic 3
5. Start your local dev "backend" and see the accelerator profile created. If no Nvidia nodes exist, then no accelerator should be created.

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/backend/src/utils/resourceUtils.ts
+++ b/backend/src/utils/resourceUtils.ts
@@ -677,8 +677,11 @@ export const cleanupGPU = async (fastify: KubeFastifyInstance): Promise<void> =>
     ) {
       // if gpu detected on cluster, create our default migrated-gpu
       const acceleratorDetected = await getDetectedAccelerators(fastify);
+      const hasNvidiaNodes = Object.keys(acceleratorDetected.total).some(
+        (nodeKey) => nodeKey === 'nvidia.com/gpu',
+      );
 
-      if (acceleratorDetected.configured) {
+      if (acceleratorDetected.configured && hasNvidiaNodes) {
         const payload: AcceleratorProfileKind = {
           kind: 'AcceleratorProfile',
           apiVersion: 'dashboard.opendatahub.io/v1',


pnaik1:
@jpuzz0, I was trying to test your pr locally, followed your Test steps, after restarting my backend, I see 
`migration-gpu-status` config automatically created in the openshift console, That means I have nvidia node??
If so how can I test for no nvidia node??

jpuzz0:
> @jpuzz0, I was trying to test your pr locally, followed your Test steps, after restarting my backend, I see `migration-gpu-status` config automatically created in the openshift console, That means I have nvidia node?? If so how can I test for no nvidia node??

No, `migration-gpu-status` will be automatically created in the console regardless. If you want to create different node types I suggest reaching out to [#forum-openshift-ai-dashboard](https://redhat-internal.slack.com/archives/C07CQ4R4HMY).

backend/src/utils/resourceUtils.ts
```diff
@@ -677,8 +677,11 @@ export const cleanupGPU = async (fastify: KubeFastifyInstance): Promise<void> =>
     ) {
       // if gpu detected on cluster, create our default migrated-gpu
       const acceleratorDetected = await getDetectedAccelerators(fastify);
+      const hasNvidiaNodes = Object.values(acceleratorDetected.nodeInfo).some((nodeInfo) =>
+        nodeInfo.architecture.includes('nvidia'),
+      );
```
Gkrumbach07, line 682: you should be use the acceleratorDetected.total object to get the available accelerators and if one of those keys is equal to `nvidia.com/gpu` then you can say that there exists a nvidia accelerator.

The reason for this is that we are creating a resource that has `identifier: 'nvidia.com/gpu'` so we should check directly that the resource exists in the node, not that the node is of some type.

Gkrumbach07 (changes requested):

Gkrumbach07:
perfect thank you

/lgtm

PR #3065, opened 2024-08-05
API integration for Register Model form

mturley:
Closes https://issues.redhat.com/browse/RHOAIENG-2240
Unblocks and partially addresses https://issues.redhat.com/browse/RHOAIENG-7571

## Description

Implements API calls on submitting the Register Model form, and adjusts some presentation details of the page to match the latest mockups.

* Adds new API client functions to `useModelRegistryAPIState`: `createModelVersionForRegisteredModel` and `createModelArtifactForModelVersion`. The existing `createModelVersion` and `createModelArtifact` functions use the base endpoint for those resources, but (1) in the case of creating artifacts we need to use the sub-route under the model version URL to associate the artifact with that version, and (2) it made sense to be consistent and do the same for creating versions under the associated model's URL.
* Adds the `artifactType` property to the required params for creating artifacts since it is now required by the API.
* Adds new utils under `~/concepts/modelRegistry/utils`: `objectStorageFieldsToUri` and `uriToObjectStorageFields`. The former is used for constructing the URI when we create our artifact here, and the latter will be used in the future to destructure this URI for prefilling the deployment form (for [RHOAIENG-8396](https://issues.redhat.com/browse/RHOAIENG-8396), cc @DaoDaoNoCode ). The URI format we're using here is defined in [RHOAIENG-3979](https://issues.redhat.com/browse/RHOAIENG-3979).
* Changes to existing RegisterModel form:
  * Connects the submit button to the API and navigates to the new model's page after submission
  * Removes `modelRegistryName` from the state held by `useRegisterModelData` - it never changes and we already have a prop for it
  * Rearranges the PageSections so we can have a floating footer in its own PageSection with `stickyOnBreakpoint` below the form as specified in the mockups
    * Unfortunately this changes whitespace on a lot of extra lines, the diff may be hard to read.
  * Sets `isWidthLimited` on the Form
  * Updates some text in the form ([see latest mockup here](https://www.figma.com/proto/KTsYhANYQqB0NUrXW7wQH1/Model-Registry?page-id=6782%3A296930&type=design&node-id=7742-12333&viewport=985%2C-1054%2C0.1&t=Tar1fktYL7sXp9c3-1&scaling=min-zoom&starting-point-node-id=7742%3A12333&show-proto-sidebar=1))
  * Adds the "Source model format version" field per slack discussion
  * Connects the cancel button so it navigates back to the MR landing page
* Splits out some fields of `useRegisterModelData` into `useRegisterVersionData` in anticipation of reusing this state for the upcoming Register Version form

Note - See the TODO comment in `frontend/src/pages/modelRegistry/screens/RegisterModel/utils.ts`: Once we have prefilling from data connections in this form ([RHOAIENG-9914](https://issues.redhat.com/browse/RHOAIENG-9914)) we should update the artifact creation so it includes a `storageKey` property set to the data connection name. cc @gitdallas 

Top of form with sticky footer
Screenshot 2024-08-06 at 12 00 35 PM

Bottom of form in object storage mode
Screenshot 2024-08-06 at 12 00 41 PM

Bottom of form in URI mode
Screenshot 2024-08-06 at 12 00 48 PM

cc @yih-wang 

## How Has This Been Tested?

Tested on the modelregistry-ui cluster. Also tested after disconnecting the VPN to make sure it would render network errors properly.

Any fake data is sufficient to test this - just put anything in all the required fields and it should create the required resources and redirect you to the new registered model's page.

## Test Impact

* Added unit tests for the `objectStorageFieldsToUri` and `uriToObjectStorageFields` utils
* Added Cypress tests for submitting the form in both object storage and URI modes and asserting the request bodies for the created model, version and artifact. These tests sufficiently cover the utilities `registerModel` and `registerVersion` in `frontend/src/pages/modelRegistry/screens/RegisterModel/utils.ts`.

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/concepts/modelRegistry/utils.ts
+++ b/frontend/src/concepts/modelRegistry/utils.ts
@@ -1,0 +1,36 @@
+export type ObjectStorageFields = {
+  endpoint: string;
+  bucket: string;
+  region?: string;
+  path: string;
+};
+
+export const objectStorageFieldsToUri = (fields: ObjectStorageFields): string => {
+  const { endpoint, bucket, region, path } = fields;
+  const searchParams = new URLSearchParams();
+  searchParams.set('endpoint', endpoint);
+  if (region) {
+    searchParams.set('defaultRegion', region);
+  }
+  return `s3://${bucket}/${path}?${searchParams.toString()}`;
+};
+
+export const uriToObjectStorageFields = (uri: string): ObjectStorageFields | null => {
+  try {
+    const urlObj = new URL(uri);
+    // Some environments include the first token after the protocol (our bucket) in the pathname and some have it as the hostname
+    const [bucket, ...pathSplit] = `${urlObj.hostname}/${urlObj.pathname}`
+      .split('/')
+      .filter(Boolean);
+    const path = pathSplit.join('/');
+    const searchParams = new URLSearchParams(urlObj.search);
+    const endpoint = searchParams.get('endpoint');
+    const region = searchParams.get('defaultRegion');
+    if (endpoint && bucket && path) {
+      return { endpoint, bucket, region: region || undefined, path };
+    }
+    return null;
+  } catch {
+    return null;
+  }
+};
--- a/frontend/src/concepts/modelRegistry/utils.ts
+++ b/frontend/src/concepts/modelRegistry/utils.ts
@@ -5,8 +5,11 @@ export type ObjectStorageFields = {
   path: string;
 };
 
-export const objectStorageFieldsToUri = (fields: ObjectStorageFields): string => {
+export const objectStorageFieldsToUri = (fields: ObjectStorageFields): string | null => {
   const { endpoint, bucket, region, path } = fields;
+  if (!endpoint || !bucket || !path) {
+    return null;
+  }
   const searchParams = new URLSearchParams();
   searchParams.set('endpoint', endpoint);
   if (region) {
--- a/frontend/src/__tests__/cypress/cypress/pages/modelRegistry.ts
+++ b/frontend/src/__tests__/cypress/cypress/pages/modelRegistry.ts
@@ -167,6 +167,10 @@ class ModelRegistry {
   findModelVersionsTableFilter() {
     return cy.findByTestId('model-versions-table-filter');
   }
+
+  findRegisterModelButton() {
+    return cy.findByRole('button', { name: 'Register model' });
+  }
 }
 
 export const modelRegistry = new ModelRegistry();
--- a/frontend/src/__tests__/cypress/cypress/pages/modelRegistry/registerModelPage.ts
+++ b/frontend/src/__tests__/cypress/cypress/pages/modelRegistry/registerModelPage.ts
@@ -1,0 +1,37 @@
+export enum FormFieldSelector {
+  MODEL_NAME = '#model-name',
+  MODEL_DESCRIPTION = '#model-description',
+  VERSION_NAME = '#version-name',
+  VERSION_DESCRIPTION = '#version-description',
+  SOURCE_MODEL_FORMAT = '#source-model-format',
+  SOURCE_MODEL_FORMAT_VERSION = '#source-model-format-version',
+  LOCATION_TYPE_OBJECT_STORAGE = '#location-type-object-storage',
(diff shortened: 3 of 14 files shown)


frontend/src/concepts/modelRegistry/utils.ts
```diff
@@ -0,0 +1,36 @@
...
+  endpoint: string;
+  bucket: string;
+  region?: string;
+  path: string;
+};
+
+export const objectStorageFieldsToUri = (fields: ObjectStorageFields): string => {
+  const { endpoint, bucket, region, path } = fields;
+  const searchParams = new URLSearchParams();
+  searchParams.set('endpoint', endpoint);
```
lucferbux, line 11: Why don't we test first if the attributes are correct to sanetize it? I know we get the fields as mandatory but what happens if somehow endpoint can be input as an empty string? the uri is malformed or it's not set as a search param?

frontend/src/concepts/modelRegistry/utils.ts
```diff
@@ -0,0 +1,36 @@
...
+  if (region) {
+    searchParams.set('defaultRegion', region);
+  }
+  return `s3://${bucket}/${path}?${searchParams.toString()}`;
+};
+
+export const uriToObjectStorageFields = (uri: string): ObjectStorageFields | null => {
+  try {
+    const urlObj = new URL(uri);
```
lucferbux, line 20: I just told you that cause it strikes me that here we try catch to retrieve the fields but the method above we don't sanetize the inputs.

lucferbux (changes requested):
Ok, I've just done a quick first pass, I'll try to test it later running the code, so far it looks really great!

yih-wang:
Overall it looks great! Just a comment to the breadcrumb - 
I'm not sure where this registration form is triggered, but my assumption is it's been opened from the registered models list. If that's the case, then we should follow the general pattern for breadcrumb of `Model Registry - {MR Name} > Register model`
Now there's an additional middle level in the breadcrumb and it will lead the user to nowhere.

frontend/src/concepts/modelRegistry/utils.ts
mturley: Yeah the try-catch here was just specifically because the constructor for `new URL` can throw an error. It makes sense to check on both ends though.

mturley:
@lucferbux @yih-wang addressed your feedback and updated the screenshot.

gitdallas (approved):

gitdallas (approved):

gitdallas (approved):

frontend/src/concepts/modelRegistry/utils.ts
```diff
@@ -0,0 +1,39 @@
...
+  }
+  const searchParams = new URLSearchParams();
+  searchParams.set('endpoint', endpoint);
+  if (region) {
+    searchParams.set('defaultRegion', region);
+  }
+  return `s3://${bucket}/${path}?${searchParams.toString()}`;
+};
+
+export const uriToObjectStorageFields = (uri: string): ObjectStorageFields | null => {
```
DaoDaoNoCode, line 21: Can this method take `undefined` uri as a prop and return `null`? Sometimes we might want to use it like `uriToObjectStorageFields(modelArtifact.uri)` while this could be `undefined`.

DaoDaoNoCode: It's also OK to keep it as it it now, I can do the null check before using this function.

lucferbux: Is `modelArtifact.uri` marked as optional? if so you can check emptyness BEFORE executing the function. IMO we should make the function agnostic to the attribute we are gonna pass, we are converting a uri (string) if the parameter can  be optional we do the checks before the function

lucferbux (approved):
I'll wait to @DaoDaoNoCode comment before adding the lgmt, approving this in the meantime since it's working as expected

frontend/src/concepts/modelRegistry/utils.ts
mturley: Ah, sorry I missed this before merge. Feel free to make that change in your PR when you call it if you want, or I can include that in my next PR

PR #3069, opened 2024-08-06
[RHOAIENG-7572] Registered Models - Empty View Redesign

jpuzz0:
https://issues.redhat.com/browse/RHOAIENG-7572

## Description
Updated icon/text for empty state of the registered models list page.

Pasted Graphic 1

**With page height large enough to display the popover underneath the "Who's my administrator" link**
Pasted Graphic

(cc @yih-wang)

## How Has This Been Tested?
Manually tested. Can be tested with a model registry that has no registered models.

## Request review criteria:

If you have UI changes: 


yih-wang:
Copy my comment from slack for the tracking purpose:
User will see this empty state in two cases - when *no model registry has been created in the current cluster* or *the user has no access to any of the existing model registries*.
In either case, there won't be a context selector for selecting registries.

jpuzz0:
> Copy my comment from slack for the tracking purpose: User will see this empty state in two cases - when _no model registry has been created in the current cluster_ or _the user has no access to any of the existing model registries_. In either case, there won't be a context selector for selecting registries.

@yih-wang updated & added new screenshots in the description.

yih-wang:
LGTM, thanks!

manaswinidas (changes requested):
According to the JIRA, I think the empty state is for "No model registries" rather than "No registered models in a model registry". I can still see this empty state for "No model registries" I think you have to change this empty state in `ModelRegistryCoreLoader` instead of `ModelRegistry`. There is also a `TODO: Replace this with a component for empty registries once we have the designs` for this one.
Screenshot 2024-08-08 at 1 55 57 PM

jpuzz0:
> According to the JIRA, I think the empty state is for "No model registries" rather than "No registered models in a model registry". I can still see this empty state for "No model registries" I think you have to change this empty state in `ModelRegistryCoreLoader` instead of `ModelRegistry`. There is also a `TODO: Replace this with a component for empty registries once we have the designs` for this one.

@manaswinidas thanks for pointing this out. I've updated the PR with this change.

frontend/src/pages/modelRegistry/ModelRegistryCoreLoader.tsx
```diff
@@ -53,16 +57,37 @@ const ModelRegistryCoreLoader: React.FC<ModelRegistryCoreLoaderProps> =
...
         emptyStatePage: (
-          // TODO: Replace this with a component for empty registries once we have the designs
           <EmptyModelRegistryState
-            title="No model registries found"
-            description="No model registries found in the cluster. Configure a new one before registering models."
-            primaryActionText="Configure model registry"
-            primaryActionOnClick={() => {
-              // TODO: Add primary action
-            }}
+            testid="empty-registered-models"
```
manaswinidas, line 61: This is not the correct test-id. We need new Cypress tests with new test-id, for example, `empty-model-registry` or `empty-state-model-registry`(same as the image name).

frontend/src/pages/modelRegistry/ModelRegistryCoreLoader.tsx
```diff
@@ -89,18 +114,16 @@ const ModelRegistryCoreLoader: React.FC<ModelRegistryCoreLoaderProps> =
     return (
       <ApplicationsPage
         title={
-          <TitleWithIcon
-            title="Registered models"
-            objectType={ProjectObjectType.registeredModels}
-          />
+          <TitleWithIcon title="Model registry" objectType={ProjectObjectType.registeredModels} />
```
manaswinidas, line 117: I didn't find `Model registry` in the title in the mocks(the ones attached to the JIRA). Let me know if I'm missing something. If `Model registry` is the title, please update the screenshots in the PR description. I saw a new description "View and manage all of your registered models. Registering models to model registry allows you to manage their content, metadata, versions, and user access settings.", is this description for all pages with `Registered models` as title? This is perhaps a question for @yih-wang I found another page with `Registered Models` and a description - "View and manage your registered models.". If this description is for all pages, please update it accordingly.

manaswinidas (changes requested):
Works fine now. We need Cypress tests with a new test-id.

frontend/src/pages/modelRegistry/ModelRegistryCoreLoader.tsx
jpuzz0: That copy change is a part of this separate task: https://issues.redhat.com/browse/RHOAIENG-6574

jpuzz0: I ended up updating this specific text anyways for the sake of cypress tests passing.

frontend/src/pages/modelRegistry/ModelRegistryCoreLoader.tsx
jpuzz0: Updated.

yih-wang:
@jpuzz0 A following comment - The screenshot of the hovering state is removed, but I kinda remember that the button has an underline while hovering. Could you please check whether you're using the **link button variant** or the **inline button variant** for this component? The **link button variant** should be used, and when applied correctly, the link will not have an underline while hovering.

jpuzz0:
> @jpuzz0 A following comment - The screenshot of the hovering state is removed, but I kinda remember that the button has an underline while hovering. Could you please check whether you're using the **link button variant** or the **inline button variant** for this component? The **link button variant** should be used, and when applied correctly, the link will not have an underline while hovering.

@yih-wang I added the screenshot back which shows the underline on hover.

frontend/src/pages/modelRegistry/ModelRegistryCoreLoader.tsx
```diff
@@ -52,16 +56,37 @@ const ModelRegistryCoreLoader: React.FC<ModelRegistryCoreLoaderProps> =
...
+                    <ListItem>
+                      The person who gave you your username, or who helped you to log in for the
+                      first time
+                    </ListItem>
+                    <ListItem>Someone in your IT department or help desk</ListItem>
+                    <ListItem>A project manager or developer</ListItem>
+                  </List>
+                }
+              >
+                <Button variant="link" icon={<OutlinedQuestionCircleIcon />} isInline>
```
yih-wang, line 82: The `isInline` property should only be used when the button is placed inline with other texts thus isn't needed here.

jpuzz0: Without `isInline`, the underline which you explicitly asked for no longer shows up on hover of the link. So which would you prefer? No underline on hover or remove `isInline` because it has no text next to it?

mturley: If I'm understanding this earlier comment correctly:

> The link button variant should be used, and when applied correctly, the link will not have an underline while hovering.

I think @yih-wang was asking for no underline on hover, which would be consistent with removing `isInline` here

mturley:
LGTM pending resolution of the `isInline` prop discussed above

frontend/src/pages/modelRegistry/ModelRegistryCoreLoader.tsx
yih-wang: Yes, I meant the button shouldn't display an underline when hovering.

jpuzz0: Your design had the underline, but it appears it doesn't any longer. I've now removed the property.

manaswinidas (approved):
/lgtm

Works fine. The state pops up just fine for no model registry and no registered models in a model registry.

mturley (approved):
LGTM, thanks @jpuzz0 !

frontend/src/pages/modelRegistry/ModelRegistryCoreLoader.tsx
yih-wang: Right it appears to be a bug (incorrect style) in the Figma library, so I then fixed that. Thanks for the updates!

PR #3074, opened 2024-08-07
Add deployment modal for model registry

DaoDaoNoCode:
JIRA: [RHOAIENG-8396](https://issues.redhat.com/browse/RHOAIENG-8396)

## Description

When you deploy a model based on a model registry, the following modal will show:

Screenshot 2024-08-07 at 1 27 46 PM

If anything wrong with the project, the corresponding error message will pop up:

Screenshot 2024-08-07 at 1 28 32 PM

Otherwise, the current KServe modal or InferenceService modal will pop up depending on the project's serving platform. The name will be pre-filled as `{modelName} - {versionName}`:

Screenshot 2024-08-07 at 1 29 29 PM

There is a helper text if the model registry version contains a prompt model format:

Screenshot 2024-08-07 at 1 32 08 PM

The data connections part will be decided by the model artifact URI.
If there are more than 1 matches, we don't pre-select the data connection:

Screenshot 2024-08-07 at 1 32 32 PM

If there is only one match, we pre-select the recommended data connection:

Screenshot 2024-08-07 at 1 34 19 PM

If there is no match, we pre-select the new data connection field:

Screenshot 2024-08-07 at 1 35 14 PM

The path will always be pre-filled based on the URI parsing result:

Screenshot 2024-08-07 at 1 36 58 PM

## How Has This Been Tested?

1. Create project A, B and C
2. For project A, make it an unsupported serving platform (or don't select a serving platform if both platforms are supported)
3. Go to Model registry page, choose one model registry and click on `Deploy` on the version action
4. Choose project A, check the error message
5. For project B and C, make the serving platform selected (or make your cluster only support one platform)
6. Choose project B, check if the name field is auto-filled
7. Check the data connection field is pre-selected as `New data connection` and `path` field is pre-filled
8. Create one matching secret in project B and 2 matching secrets in project C ("matching" means the secret has the same `bucket`, `endpoint` and `region` as the model registry version artifact URI you are going to deploy with)
9. Choose project B again, check the data connection field is pre-selected as `Existing data connection` and the only matching data connection is labeled and pre-selected
10. Choose project C, check the data connection field is pre-selected as `Existing data connection` and both data connections in the dropdown menu are labeled as recommended, but neither is pre-selected

## Test Impact

Add a unit test to test the util function that shows the error message when choosing the project.

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/concepts/modelRegistry/context/ModelRegistryContext.tsx
+++ b/frontend/src/concepts/modelRegistry/context/ModelRegistryContext.tsx
@@ -1,10 +1,21 @@
 import * as React from 'react';
 import { SupportedArea, conditionalArea } from '~/concepts/areas';
+import { useTemplates } from '~/api';
+import { useDashboardNamespace } from '~/redux/selectors';
+import { useContextResourceData } from '~/utilities/useContextResourceData';
+import useTemplateOrder from '~/pages/modelServing/customServingRuntimes/useTemplateOrder';
+import { ContextResourceData, CustomWatchK8sResult } from '~/types';
+import { TemplateKind } from '~/k8sTypes';
+import { DEFAULT_CONTEXT_DATA, DEFAULT_LIST_WATCH_RESULT } from '~/utilities/const';
+import useTemplateDisablement from '~/pages/modelServing/customServingRuntimes/useTemplateDisablement';
 import useModelRegistryAPIState, { ModelRegistryAPIState } from './useModelRegistryAPIState';
 
 export type ModelRegistryContextType = {
   apiState: ModelRegistryAPIState;
   refreshAPIState: () => void;
+  servingRuntimeTemplates: CustomWatchK8sResult<TemplateKind[]>;
+  servingRuntimeTemplateOrder: ContextResourceData<string>;
+  servingRuntimeTemplateDisablement: ContextResourceData<string>;
 };
 
 type ModelRegistryContextProviderProps = {
@@ -16,12 +27,25 @@ export const ModelRegistryContext = React.createContext<ModelRegistryContextType
   // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
   apiState: { apiAvailable: false, api: null as unknown as ModelRegistryAPIState['api'] },
   refreshAPIState: () => undefined,
+  servingRuntimeTemplates: DEFAULT_LIST_WATCH_RESULT,
+  servingRuntimeTemplateOrder: DEFAULT_CONTEXT_DATA,
+  servingRuntimeTemplateDisablement: DEFAULT_CONTEXT_DATA,
 });
 
 export const ModelRegistryContextProvider = conditionalArea<ModelRegistryContextProviderProps>(
   SupportedArea.MODEL_REGISTRY,
   true,
 )(({ children, modelRegistryName }) => {
+  const { dashboardNamespace } = useDashboardNamespace();
+
+  const servingRuntimeTemplates = useTemplates(dashboardNamespace);
+  const servingRuntimeTemplateOrder = useContextResourceData<string>(
+    useTemplateOrder(dashboardNamespace),
+  );
+  const servingRuntimeTemplateDisablement = useContextResourceData<string>(
+    useTemplateDisablement(dashboardNamespace),
+  );
+
   const hostPath = modelRegistryName ? `/api/service/modelregistry/${modelRegistryName}` : null;
 
   const [apiState, refreshAPIState] = useModelRegistryAPIState(hostPath);
@@ -31,6 +55,9 @@ export const ModelRegistryContextProvider = conditionalArea<ModelRegistryContext
       value={{
         apiState,
         refreshAPIState,
+        servingRuntimeTemplates,
+        servingRuntimeTemplateOrder,
+        servingRuntimeTemplateDisablement,
       }}
     >
       {children}
--- a/frontend/src/pages/modelRegistry/screens/RegisteredModels/usePrefillDeployModalFromModelRegistry.ts
+++ b/frontend/src/pages/modelRegistry/screens/RegisteredModels/usePrefillDeployModalFromModelRegistry.ts
@@ -1,0 +1,87 @@
+import { AlertVariant } from '@patternfly/react-core';
+import React from 'react';
+import { ProjectKind } from '~/k8sTypes';
+import useLabeledDataConnections from '~/pages/modelRegistry/screens/RegisteredModels/useLabeledDataConnections';
+import { RegisteredModelDeployInfo } from '~/pages/modelRegistry/screens/RegisteredModels/useRegisteredModelDeployInfo';
+import {
+  CreatingInferenceServiceObject,
+  InferenceServiceStorageType,
+  LabeledDataConnection,
+} from '~/pages/modelServing/screens/types';
+import { AwsKeys, EMPTY_AWS_SECRET_DATA } from '~/pages/projects/dataConnections/const';
+import useDataConnections from '~/pages/projects/screens/detail/data-connections/useDataConnections';
+import { DataConnection, UpdateObjectAtPropAndValue } from '~/pages/projects/types';
+
+const usePrefillDeployModalFromModelRegistry = (
+  projectContext: { currentProject: ProjectKind; dataConnections: DataConnection[] } | undefined,
+  createData: CreatingInferenceServiceObject,
+  setCreateData: UpdateObjectAtPropAndValue<CreatingInferenceServiceObject>,
+  registeredModelDeployInfo?: RegisteredModelDeployInfo,
+): [LabeledDataConnection[], boolean, Error | undefined] => {
+  const [fetchedDataConnections, dataConnectionsLoaded, dataConnectionsLoadError] =
+    useDataConnections(projectContext ? undefined : createData.project);
+  const allDataConnections = projectContext?.dataConnections || fetchedDataConnections;
+  const { dataConnections, storageFields } = useLabeledDataConnections(
+    registeredModelDeployInfo?.modelArtifactUri,
+    allDataConnections,
+  );
+
+  React.useEffect(() => {
+    if (registeredModelDeployInfo) {
+      setCreateData('name', registeredModelDeployInfo.modelName);
+      const recommendedDataConnections = dataConnections.filter(
+        (dataConnection) => dataConnection.isRecommended,
+      );
+
+      if (!storageFields) {
+        setCreateData('storage', {
+          awsData: EMPTY_AWS_SECRET_DATA,
+          dataConnection: '',
+          path: '',
+          type: InferenceServiceStorageType.EXISTING_STORAGE,
+        });
+      } else {
+        const prefilledAWSData = [
+          { key: AwsKeys.NAME, value: registeredModelDeployInfo.modelArtifactStorageKey || '' },
+          { key: AwsKeys.AWS_S3_BUCKET, value: storageFields.bucket },
+          { key: AwsKeys.S3_ENDPOINT, value: storageFields.endpoint },
+          { key: AwsKeys.DEFAULT_REGION, value: storageFields.region || '' },
+          ...EMPTY_AWS_SECRET_DATA,
+        ];
+        if (recommendedDataConnections.length === 0) {
+          setCreateData('storage', {
+            awsData: prefilledAWSData,
+            dataConnection: '',
+            path: storageFields.path,
+            type: InferenceServiceStorageType.NEW_STORAGE,
+            alert: {
+              type: AlertVariant.info,
+              title:
+                "We've auto-switched to create a new data connection and pre-filled the details for you.",
+              message:
+                'Model location info is available in the registry but no matching data connection in the project. So we automatically switched the option to create a new data connection and prefilled the information.',
+            },
+          });
+        } else if (recommendedDataConnections.length === 1) {
+          setCreateData('storage', {
+            awsData: prefilledAWSData,
+            dataConnection: recommendedDataConnections[0].dataConnection.data.metadata.name,
+            path: storageFields.path,
+            type: InferenceServiceStorageType.EXISTING_STORAGE,
+          });
+        } else {
+          setCreateData('storage', {
+            awsData: prefilledAWSData,
+            dataConnection: '',
+            path: storageFields.path,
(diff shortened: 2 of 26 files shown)


yih-wang:
Thanks for the work @DaoDaoNoCode! Just a few comments - 
1. Can you stretch the modal width in the first image to the same width as the other modals?
2. Regarding the pre-filled model name, I can't see what happens if a version has been deployed multiple times. The ideal behavior is if the name has already been used within the selected project, add a postfix such as ‘-2’, ‘-3’ to differentiate it.
3. Regarding the helper text to the prompt model format - the helper text will only show up after a model server is selected, and the `model framework` dropdown is available.
4. For data connection, when there is no match and the new data connection is selected by default - a warning is missing, check details in the [mockup](https://www.figma.com/proto/KTsYhANYQqB0NUrXW7wQH1/Model-Registry?page-id=6782%3A296930&node-id=8238-695915&viewport=941%2C-3865%2C0.11&t=28gM6CEeyGJs3Cpp-1&scaling=min-zoom&content-scaling=fixed&starting-point-node-id=8238%3A699060)

frontend/src/pages/modelRegistry/screens/RegisteredModels/useRegisteredModelContext.ts
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
+import useModelArtifactsByVersionId from '~/concepts/modelRegistry/apiHooks/useModelArtifactsByVersionId';
+import useRegisteredModelById from '~/concepts/modelRegistry/apiHooks/useRegisteredModelById';
+import { ModelVersion } from '~/concepts/modelRegistry/types';
+
+export type RegisteredModelContext = {
```
mturley, line 6: I wonder if we should use a word other than `Context` here to name `RegisteredModelContext` and `useRegisteredModelContext`, since it's not a React Context. Maybe `RegisteredModelDeployInfo` or something? I guess we have the preexisting `projectContext` prop in the modals so maybe it's fine. Up to you.

Even if we keep the word Context I think it should have the word Deploy in it to indicate its purpose - `RegisteredModelDeployContext` / `useRegisteredModelDeployContext` maybe?

frontend/src/pages/modelRegistry/screens/components/DeployRegisteredModelModal.tsx
```diff
@@ -0,0 +1,112 @@
+import React from 'react';
+import { Button, Form, Modal } from '@patternfly/react-core';
+import { ModelVersion } from '~/concepts/modelRegistry/types';
+import { ProjectKind } from '~/k8sTypes';
+import useProjectErrorForRegisteredModel from '~/pages/modelRegistry/screens/RegisteredModels/useProjectErrorForRegisteredModel';
+import ProjectSelector from '~/pages/modelServing/screens/projects/InferenceServiceModal/ProjectSelector';
+import { ManageKServeModalWrapper } from '~/pages/modelServing/screens/projects/kServeModal/ManageKServeModal';
+import useServingPlatformStatuses from '~/pages/modelServing/useServingPlatformStatuses';
+import { getProjectModelServingPlatform } from '~/pages/modelServing/screens/projects/utils';
```
mturley, line 9: There's also a bunch of stuff being imported across `~/pages/modelServing` and `~/pages/modelRegistry` here - maybe we should take a pass and see if any of it is a good candidate to move to `~/concepts/modelServing`....

...ahh, we don't yet have a `~/concepts/modelServing`. Maybe we should open a followup tech tebt issue to create it and lift some of these things out there after MR MVP. I may be stirring up too much refactoring here with these comments 😄.

Yeah ok, for stuff you're importing from `~/pages/modelServing` let's leave them alone for now and I'll open a followup issue, but for new stuff you're introducing we should consider putting stuff in `~/concepts/modelRegistry` if we can do so cleanly. I'll leave it up to your judgement how much of that is within the scope of this PR.

frontend/src/pages/modelServing/screens/projects/InferenceServiceModal/ManageInferenceServiceModal.tsx
```diff
@@ -33,11 +37,16 @@ type ManageInferenceServiceModalProps = {
   }
 >;
 
-const ManageInferenceServiceModal: React.FC<ManageInferenceServiceModalProps> = ({
+export const ManageInferenceServiceModalWrapper: React.FC<
+  ManageInferenceServiceModalProps & { children: React.ReactNode }
+> = ({
```
mturley, line 42: Feel free to leave this as-is, but for each of these modals I wonder if you really need the Wrapper components and introduction of `children` - You could instead just add an optional `projectSection` prop that defaults to the default children you pass below and let the DeployRegisteredModelModal pass its own in that overrides it.

frontend/src/pages/modelServing/screens/projects/kServeModal/ManageKServeModal.tsx
```diff
@@ -73,12 +77,17 @@ type ManageKServeModalProps = {
   }
 >;
 
-const ManageKServeModal: React.FC<ManageKServeModalProps> = ({
+export const ManageKServeModalWrapper: React.FC<
```
mturley, line 80: Same here, you could avoid the wrapper if you want to.

mturley (changes requested):
This is looking really good! I have one actual change request (the comment in `useLabeledDataConnections`) and then just some questions and suggestions.

frontend/src/pages/modelRegistry/screens/RegisteredModels/useRegisteredModelContext.ts
DaoDaoNoCode: I agree with you and I will go with the `RegisteredModelDeployInfo`. Thanks!

frontend/src/pages/modelServing/screens/projects/InferenceServiceModal/ManageInferenceServiceModal.tsx
DaoDaoNoCode: Good suggestion, will update!

frontend/src/pages/modelRegistry/screens/components/DeployRegisteredModelModal.tsx
DaoDaoNoCode: Yeah I agree, I will leave this PR as it is for now because I checked the usage in those hooks, and it seems it's cross-page using many enums/const so I think they should be restructured afterward to lift all of them up to the `/concepts` folder in another ticket.

frontend/src/pages/modelRegistry/screens/RegisteredModels/useRegisteredModelContext.ts
```diff
@@ -0,0 +1,37 @@
...
+  modelName: string;
+  modelFormat?: string;
+  modelArtifactUri?: string;
+  modelArtifactStorageKey?: string;
+};
+
+const useRegisteredModelContext = (modelVersion: ModelVersion): RegisteredModelContext => {
+  const [registeredModel] = useRegisteredModelById(modelVersion.registeredModelId);
+  const [modelArtifactList] = useModelArtifactsByVersionId(modelVersion.id);
```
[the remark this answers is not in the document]
DaoDaoNoCode: Hmm, I am not sure, the only side-effect here is if there is an error, the modal will not pre-fill those values, but that doesn't block users from deploying the model, they just need to select by themselves. I am not sure the best way to handle the loading and error state here, WDYT?

frontend/src/concepts/modelRegistry/context/ModelRegistryContext.tsx
```diff
@@ -49,6 +64,14 @@ export const ModelRegistryContextProvider = conditionalArea<ModelRegistryContext
...
   }, []);
 
+  const servingRuntimeTemplates = useTemplates(dashboardNamespace);
+  const servingRuntimeTemplateOrder = useContextResourceData<string>(
+    useTemplateOrder(dashboardNamespace),
+  );
+  const servingRuntimeTemplateDisablement = useContextResourceData<string>(
+    useTemplateDisablement(dashboardNamespace),
+  );
+
```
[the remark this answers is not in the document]
DaoDaoNoCode: I think it's OK not to capture the state here, the submit function should be able to catch if there is an error when submitting, the error will only affect the order/disablement status of the templates, which is not a big deal.

frontend/src/pages/modelServing/screens/projects/kServeModal/ManageKServeModal.tsx
```diff
@@ -116,6 +131,11 @@ const ManageKServeModal: React.FC<ManageKServeModalProps> = ({
     }
   }, [currentProjectName, setCreateDataInferenceService, isOpen]);
 
+  // Refresh model format selection when changing serving runtime template selection
+  React.useEffect(() => {
+    setCreateDataInferenceService('format', { name: '' });
+  }, [createDataServingRuntime.servingRuntimeTemplateName, setCreateDataInferenceService]);
+
```
DaoDaoNoCode, line 138: Fixed a small bug here. When selecting a new template, it will reset the model frame selection now (it didn't before). Not sure if there is a ticket logged.

frontend/src/pages/modelRegistry/screens/RegisteredModels/usePrefillDeployModalFromModelRegistry.ts
```diff
@@ -0,0 +1,87 @@
...
+          type: InferenceServiceStorageType.EXISTING_STORAGE,
+        });
+      } else {
+        const prefilledAWSData = [
+          { key: AwsKeys.NAME, value: registeredModelDeployInfo.modelArtifactStorageKey || '' },
+          { key: AwsKeys.AWS_S3_BUCKET, value: storageFields.bucket },
+          { key: AwsKeys.S3_ENDPOINT, value: storageFields.endpoint },
+          { key: AwsKeys.DEFAULT_REGION, value: storageFields.region || '' },
+          ...EMPTY_AWS_SECRET_DATA,
+        ];
```
DaoDaoNoCode, line 50: Pre-filled all the fields if the URI is parsed correctly. Forgot to do that before, and now it's fixed.

DaoDaoNoCode:
> 1. Can you stretch the modal width in the first image to the same width as the other modals?
> 2. Regarding the pre-filled model name, I can't see what happens if a version has been deployed multiple times. The ideal behavior is if the name has already been used within the selected project, add a postfix such as ‘-2’, ‘-3’ to differentiate it.
> 3. Regarding the helper text to the prompt model format - the helper text will only show up after a model server is selected, and the model framework dropdown is available.
> 4. For data connection, when there is no match and the new data connection is selected by default - a warning is missing, check details in the [mockup](https://www.figma.com/proto/KTsYhANYQqB0NUrXW7wQH1/Model-Registry?page-id=6782%3A296930&node-id=8238-695915&viewport=941%2C-3865%2C0.11&t=28gM6CEeyGJs3Cpp-1&scaling=min-zoom&content-scaling=fixed&starting-point-node-id=8238%3A699060)

Based on the comments above, I updated the following:
1. For your first comment, I updated the size of the initial modal

Screenshot 2024-08-08 at 7 48 23 PM

2. For the second comment, I found it's a little bit hard from the technical side to implement what you requested. We will need to fetch all the models in that project and it's hard to decide the postfix number (e.g. There are model-1, model-3 and model-4, what should the auto-generated number be? model-2 or model-5?). I prefer we keep it as it is now, and if the name is duplicated, the server will always prompt an error when the user submits the model. (cc @mturley Might need your input here)
3. For the third comment, I updated the helper text to make it only show when there are items in the dropdown menu.

No items in the menu:

Screenshot 2024-08-08 at 7 56 50 PM

Items in the menu:

Screenshot 2024-08-08 at 7 57 01 PM

4. For the fourth comment, I updated the alert:

Screenshot 2024-08-08 at 7 58 13 PM

@yih-wang Let me know what you think, thanks!

yih-wang:
@DaoDaoNoCode All the changes look good. Thanks for the rapid work!
Regarding the auto-generated deployment name, my primary concern is to avoid duplicate names and prevent user frustration from seeing duplicated names error continuously. If the original proposal doesn't work, we can implement a workaround by adding a timestamp as postfix to each name. So we'll have a format like [modelname-versionname-yyyymmddhhmmss], which should effectively eliminate duplicates as well.

yih-wang:
LGTM, thanks @DaoDaoNoCode !

frontend/src/pages/modelRegistry/screens/RegisteredModels/useRegisteredModelContext.ts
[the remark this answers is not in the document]
mturley: I suppose the user can't get to this deployment action unless they've already gone through the Model/Version details pages which will be making these same API requests, so it's not critical to catch errors here - but technically if the user e.g. loses their internet connection and then clicks Deploy, we should ideally show an error instead of failing silently.

Can we add two more properties to `RegisteredModelDeployInfo`:
```ts
loaded: boolean;
loadError?: Error;
```
and where we call `useRegisteredModelById` and `useModelArtifactsByVersionId`, can we pull out the `loaded` and `loadError` values from those returned arrays, and then when we return `RegisteredModelDeployInfo` we can aggregate them like:
```ts
loaded: registeredModelLoaded && modelArtifactsLoaded,
loadError: registeredModelLoadError || modelArtifactsLoadError,
```
and then in the `DeployRegisteredModelModal`, in the body of the modal can we render an Alert if there is a `loadError` in that object or a Spinner if it's `!loaded`?

frontend/src/concepts/modelRegistry/context/ModelRegistryContext.tsx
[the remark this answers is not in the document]
mturley: Sounds reasonable to me.

frontend/src/pages/modelServing/screens/projects/kServeModal/ManageKServeModal.tsx
mturley: I can't find a ticket, I think it's probably ok to just fix this here. @lucferbux do you happen to know if there's a bug logged for this?

mturley (approved):
Pulled and tested on the modelregistry-ui cluster - everything looks and works great here!! Thanks so much for the hard work @DaoDaoNoCode !

PR #3086, opened 2024-08-13
Removed disableS3Endpoint and disableArtifactsAPI flags

Gkrumbach07:
closes: https://issues.redhat.com/browse/RHOAIENG-9698

## Description

- removed `disableS3Endpoint` and `disableArtifactsAPI` feature flags and their uses
- removed tests that tested this flag
- updated other tests
- removed `/api/storage` api endpoint and frontend usages
- removed geStorageObject function as it is not supported due to CORS issues
- removed size warnings as it is not relevant anymore

## How Has This Been Tested?

1. create a pipeline run with markdown, html, model, and text/csv metric outputs
2. check that you can visualize the html formatted
3. check that markdown is visible but it will not be formatted
4. do this for both compare runs and artifact  viz tab
5. check that artifact uri link opens all in a new tab, except for model which will also download

## Test Impact

removed tests that tested the flags, updated others

## Request review criteria:

If you have UI changes: 


DaoDaoNoCode:
There are some places that's not used anymore and might need removal:
1. In https://github.com/Gkrumbach07/odh-dashboard/blob/cleanup-s3-feature/frontend/src/concepts/pipelines/content/artifacts/utils.ts. Should we remove the 2 functions in this files and remove the unit tests related to them?
2. In https://github.com/Gkrumbach07/odh-dashboard/blob/cleanup-s3-feature/frontend/src/__tests__/cypress/cypress/support/commands/odh.ts#L578-L593, should we remove the 2 support for the old `/api/storage` endpoint?
3. In https://github.com/Gkrumbach07/odh-dashboard/blob/cleanup-s3-feature/frontend/src/__tests__/cypress/cypress/tests/mocked/pipelines/compareRuns.cy.ts#L218-L237, should we remove these 2 mocked API calls?

DaoDaoNoCode (approved):
Tested on the cluster. No feature flags are needed and the feature is working well.
/lgtm

PR #3087, opened 2024-08-13
add actions to create and edit section fields

christianvogt:
https://issues.redhat.com/browse/RHOAIENG-10317

## Description

cc @simrandhaliw 

Adds the ability to create, edit and delete field sections.
Tests will be added. This PR provides some plumbing for integrating the add field modal as well.

Creation modal:
image

Editing modal:
image

Popovers:
image
image

Empty field state:
image

Section row with actions:
image

Buttons beneath fields table:
image

## How Has This Been Tested?

Construct a connection type:
<details>
<summary>ConfigMap</summary>
```
kind: ConfigMap
apiVersion: v1
metadata:
  name: test-ct
  namespace: opendatahub
  labels:
    opendatahub.io/connection-type: 'true'
  annotations:
    opendatahub.io/enabled: 'true'
    opendatahub.io/username: test
    openshift.io/description: this is the description
    openshift.io/display-name: test connection type
```

</details>

Go to `/connectionTypes/duplicate/test-ct`.
Try all actions related to section headings. Delete, edit, add...

## Test Impact

Added a unit test

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/concepts/connectionTypes/fields/ConnectionTypeFormFields.tsx
+++ b/frontend/src/concepts/connectionTypes/fields/ConnectionTypeFormFields.tsx
@@ -17,9 +17,6 @@ type Props = {
 
 type FieldGroup = { section: SectionField | undefined; fields: ConnectionTypeDataField[] };
 
-const createKey = (field: ConnectionTypeField) =>
-  `${field.type}-${field.type === ConnectionTypeFieldType.Section ? field.name : field.envVar}`;
-
 const ConnectionTypeFormFields: React.FC<Props> = ({ fields, isPreview, onChange }) => {
   const fieldGroups = React.useMemo(
     () =>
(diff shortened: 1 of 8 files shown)


frontend/src/concepts/connectionTypes/fields/ConnectionTypeFormFields.tsx
```diff
@@ -48,9 +45,9 @@ const ConnectionTypeFormFields: React.FC<Props> = ({ fields, isPreview, onChange
 
   return (
     <>
-      {fieldGroups?.map((fieldGroup) =>
+      {fieldGroups?.map((fieldGroup, i) =>
         fieldGroup.section ? (
-          <SectionFormField field={fieldGroup.section} key={createKey(fieldGroup.section)}>
+          <SectionFormField field={fieldGroup.section} key={i}>
             {renderDataFields(fieldGroup.fields)}
```
emilys314, line 51: Yeah.. we don't have any unique identifiers for the items :(  But here they can also be reordered, so using the index as a key probably shouldn't be done?

Since key is just used by react for tracking the lifecycles I'm pretty sure, I think it wouldn't hurt to use a combination of index and the field properties?

christianvogt: Once index is included in the key, there's really no point in concatenating with another string.

jeff-phillips-18 (approved):

PR #3089, opened 2024-08-14
Fix error in loading Enabled applications after uninstalling an ISV

manaswinidas:
Closes: https://issues.redhat.com/browse/RHOAIENG-1135

## Description

Added a catch block to `fastify.kube.customObjectsApi.getNamespacedCustomObject`

## How Has This Been Tested?

1. Install an ISV operator(e.g. Elastic(ECK) operator/NVIDIA GPU(Certified) operator).
2. If the corresponding `OdhApplication` isn't present, create the corresponding `OdhApplication` using https://github.com/opendatahub-io/odh-dashboard/blob/main/manifests/rhoai/addon/apps/starburst/starburst-app.yaml with `enabled: true` inside `spec`
3. Go to Dashboard > Enabled and wait for the card to appear.
4. Uninstall the operator and head back to the Enabled page
5. You should not see an `HTTP request failed` page.
6. For the Elastic tile.. when you have the operator installed you shouldn't see `Disabled` in the tile and vice-versa. 

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/backend/src/utils/componentUtils.ts
+++ b/backend/src/utils/componentUtils.ts
@@ -12,6 +12,7 @@ import {
   getInstalledKfdefs,
   getSubscriptions,
 } from './resourceUtils';
+import { isHttpError } from '../utils';
 
 type RoutesResponse = {
   body: {
@@ -205,10 +206,11 @@ const getCSVForApp = (
       return undefined;
     })
     .catch((e) => {
-      if (e?.statusCode === 404) {
+      if (isHttpError(e) && e.statusCode === 404) {
         fastify.log.error(e);
+        return undefined;
       }
-      return undefined;
+      throw e;
     });
 };
 
--- a/backend/src/utils/componentUtils.ts
+++ b/backend/src/utils/componentUtils.ts
@@ -203,6 +203,12 @@ const getCSVForApp = (
         return csv;
       }
       return undefined;
+    })
+    .catch((e) => {
+      if (e?.statusCode === 404) {
+        fastify.log.error(e);
+      }
+      return undefined;
     });
 };
 


backend/src/utils/componentUtils.ts
```diff
@@ -180,7 +180,8 @@ const getCSVForApp = (
 
   const installedCSV = subStatus.installedCSV;
 
-  if (!installedCSV) {
+  // eslint-disable-next-line
+  if (!!installedCSV) {
```
gitdallas, line 184: any reason you couldn't just remove the `!` and not have to disable eslint?

manaswinidas: @gitdallas If `installedCSV` is present => the `fastify.kube.customObjectsApi.getNamespacedCustomObject` should execute successfully and not result in a `HTTP request failed`. I was trying to handle a case when this string is falsy... if it's falsy.. we should execute `Promise.resolve(undefined)` instead of `fastify.kube.customObjectsApi.getNamespacedCustomObject`

gitdallas: `!!falsy === false` though, both `falsy` and `!!falsy` would evaluate to false here.

if you want "falsy to execute `Promise.resolve(undefined)`", the previous way was doing that. `!falsy === true`

mturley: Agreed, this change would prevent `fastify.kube.customObjectsApi.getNamespacedCustomObject` from ever getting called with a defined `installedCSV` value, there must be something else going on to cause the bug

backend/src/utils/componentUtils.ts
```diff
@@ -203,6 +203,12 @@ const getCSVForApp = (
         return csv;
       }
       return undefined;
+    })
+    .catch((e) => {
+      if (e?.statusCode === 404) {
+        fastify.log.error(e);
+      }
+      return undefined;
```
mturley, line 211: This will swallow non-404 errors that we probably still want to throw. Maybe instead of `return undefined;` we just want `throw e;` here?

mturley: I'll test it myself when I get a chance, but what does the user see instead of the error page? Is it ok that the fetch silently fails? I wonder if we should get help from Berto verifying the fix (not sure of his GitHub username)

manaswinidas: The user shouldn't see an error page.. but if you want to test for the Elastic tile.. when you have the operator installed you shouldn't see `Disabled` in the title and when you don't have it installed.. there should be a test implying `Disabled` in the Elastic tile.

mturley:
I tried to reproduce/test this on the modelregistry-ui cluster (it's what I had available) by installing the Starburst Enterprise Helm Operator and creating the OdhApplication from that template. It complained that `spec.enabled` isn't a valid field, and I can't get the card to appear in the dashboard.

@manaswinidas maybe we can meet when I'm back and you can walk me through the repro, or feel free to ask someone else on the team to review/approve.

mturley (approved):

PR #3092, opened 2024-08-14
[RHOAIENG-6573] Model Detail - Registered Deployments Section

jpuzz0:
https://issues.redhat.com/browse/RHOAIENG-6573

## Description
- Added labels to creation of model deployments
- Added registered deployments to model version details and archived model version details.

image

(cc @yih-wang)

## How to test
Navigate to the version details of a given model version (archived or otherwise). Click on the registered deployments tab, verify the contents match the design in terms of the alert, table columns, and table data. If no table data exists, that means the version has no registered deployments, so you can create one using the toolbar action "Deploy".

## Test Impact
Added unit tests and a cypress test verifying data can render in the registered deployments table.

## Request review criteria:

If you have UI changes: 


mturley:
LGTM at a quick glance - will do a more complete review tomorrow morning. Thanks for the work @jpuzz0 .

yih-wang:
Thanks for the efforts @jpuzz0! One comment and two questions -
* The microcopy of the tab name and the alert should be aligned with what is provided in the handoff doc, reference [this row](https://docs.google.com/document/d/1-YvEYVaVJXWKGGTsZTsVyxzKpaHYfSvGNzz6rIrlvDE/edit#bookmark=id.1b3tcqwugh80) 
* I see a column for `API protocol` in the table on the live Model Serving page, but it doesn't show up in the screenshots, has that column been removed in the latest RHOAI version?
* I'm a little confused when I see the case that deployments are showing up for an archived version. Because the user is not allowed to archive the version/model which includes active deployments, so for an archived version, this tab should always be empty.

jpuzz0:
> * this row

@yih-wang 

1. I'll update that microcopy

2. Your mockups don't have that API protocol column, so therefore I removed it for this instance of the table:
image

3. We currently don't have that restriction setup, and that isn't a part of this task's scope, so we can make a separate JIRA task/issue for that.

mturley (approved):
download

(tested for RBAC issues with an image on modelregistry-ui cluster after fetches were implemented, tested new changes on localhost)

PR #3093, opened 2024-08-15
RHOAIENG-9480 Update useTrackUser for Developer Sandbox

pilhuhn:
https://issues.redhat.com/browse/RHOAIENG-9480

## Description
In cases where we have a sso-user-id , supply this to segment tracking instead of the anonymous one.

## How Has This Been Tested?
So far.. manual inspection only.

## Request review criteria:

If you have UI changes: 

cc @christianvogt

### DIFF
--- a/backend/src/routes/api/status/statusUtils.ts
+++ b/backend/src/routes/api/status/statusUtils.ts
@@ -1,6 +1,6 @@
 import { FastifyRequest } from 'fastify';
 import { KubeFastifyInstance, KubeStatus } from '../../../types';
-import { getUserName } from '../../../utils/userUtils';
+import { getUserInfo } from '../../../utils/userUtils';
 import { createCustomError } from '../../../utils/requestUtils';
 import { isUserAdmin, isUserAllowed } from '../../../utils/adminUtils';
 import { isImpersonating } from '../../../devFlags';
@@ -19,7 +19,7 @@ export const status = async (
 
   const { server } = currentCluster;
 
-  const userName = await getUserName(fastify, request);
+  const { userName, userID } = await getUserInfo(fastify, request);
   const isAdmin = await isUserAdmin(fastify, userName, namespace);
   const isAllowed = isAdmin ? true : await isUserAllowed(fastify, userName);
 
@@ -37,6 +37,7 @@ export const status = async (
       currentUser,
       namespace,
       userName,
+      userID,
       clusterID,
       clusterBranding,
       isAdmin,
--- a/backend/src/utils/userUtils.ts
+++ b/backend/src/utils/userUtils.ts
@@ -29,6 +29,9 @@ export type OpenShiftUser = {
     name: string;
     uid: string;
     resourceVersion: string;
+    annotations?: {
+      'toolchain.dev.openshift.com/sso-user-id': string;
+    };
   };
   fullName: string;
   identities: string[];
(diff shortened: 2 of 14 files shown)


backend/src/routes/api/status/statusUtils.ts
```diff
@@ -20,6 +20,7 @@ export const status = async (
   const { server } = currentCluster;
 
   const userName = await getUserName(fastify, request);
+  const userID: string = await getUserId(fastify, request);
```
christianvogt, line 23: both `getUserName` and `getUserId` perform a network request to obtain user data. These should be combined into a single `getUserInfo` and then deconstruct the specifics you want here:
```suggestion
  const { userName, userID } = await getUserInfo(fastify, request);
```

backend/src/utils/userUtils.ts
```diff
@@ -29,6 +29,9 @@ export type OpenShiftUser = {
     name: string;
     uid: string;
     resourceVersion: string;
+    annotations: {
```
christianvogt, line 32: Annotations are optional.
```suggestion
    annotations?: {
```

frontend/src/concepts/analyticsTracking/segmentIOUtils.tsx
```diff
@@ -108,6 +110,10 @@ export const fireIdentifyEvent = (properties: IdentifyEventProperties): void =>
       isAdmin: properties.isAdmin,
       canCreateProjects: properties.canCreateProjects,
     };
-    window.analytics.identify(properties.anonymousID, traits);
+    if (properties.userID) {
+      window.analytics.identify(properties.userID, traits);
+    } else {
+      window.analytics.identify(properties.anonymousID, traits);
+    }
```
christianvogt, line 117:
```suggestion
    window.analytics.identify(properties.userID || properties.anonymousID, traits);
```

Although this works, shouldn't the `anonymousID` be set to `userID` in the first place instead? That way we don't bother generating the `anonymousID` and there's always only a single way to deal with the ID.

Doesn't sandbox also want the `toolchain.dev.openshift.com/sso-account-id` as part of the identify payload?

pilhuhn: Wrt the latter, we only need the web-user-id, which is the sso-id. Actually for my account the 2 seem to even be the same

frontend/src/concepts/analyticsTracking/useTrackUser.ts
```diff
@@ -26,9 +29,11 @@ export const useTrackUser = (username?: string): [IdentifyEventProperties, boole
...
     };
 
-    computeAnonymousUserId().then((val) => {
-      setAnonymousId(val);
-    });
+    if (!userID) {
+      computeAnonymousUserId().then((val) => {
+        setAnonymousId(val);
+      });
+    }
```
christianvogt, line 36: If you never compute the anonymousID, then the user props will never be considered loaded according to the last line:
```
return [props, acLoaded && !!anonymousId];
```

Shouldn't the anonymousId be set to the incoming userID instead?

pilhuhn: Done

frontend/src/concepts/analyticsTracking/useTrackUser.ts
```diff
@@ -26,9 +29,13 @@ export const useTrackUser = (username?: string): [IdentifyEventProperties, boole
...
-    computeAnonymousUserId().then((val) => {
-      setAnonymousId(val);
-    });
+    if (!userID) {
+      computeAnonymousUserId().then((val) => {
+        setAnonymousId(val);
+      });
+    } else {
+      setAnonymousId(userID);
+    }
```
christianvogt, line 38: If `userID` is the anonymous ID, then you can init the anonymous ID as follows instead of this if condition:
```
const [anonymousId, setAnonymousId] = React.useState<string | undefined>(userID);
```

Furthermore, if there's no reason to make use of `userID` aside from it being the anonymous ID, then you can remove the `userID` return value as well as the conditional tracking event based on user ID in `segmentIOUtils`

pilhuhn: Tbh I am not sure about the latter as the segment library seems to do some magic by introspecting property names and acting accordingly.

pilhuhn: I suspect that the segment library is under the hood evaluating the name of the first argument's variable and setting things accordingly

pilhuhn:
Closing in favour of #3107

PR #3095, opened 2024-08-15
changed the source of Owner and updated field to be searchable and so…

YuliaKrimerman:
…rtable

https://issues.redhat.com/browse/RHOAIENG-7566
## Description
Update the source of the Owner - now it's just a field that we get from the API response. I've attached the screenshot that shows where this field on the page now is coming from. 
In addition I've made the owner a sortable field as well as a Search option. See screenshots attached.

## How Has This Been Tested?
Tested on the Model Registry Cluster

## Test Impact
Updated the tests to cover now both field options scenarios( with Author and when "-" )

## Request review criteria:

If you have UI changes: 

Screenshot 2024-08-15 at 9 00 24 AM
Screenshot 2024-08-15 at 9 00 37 AM
Screenshot 2024-08-15 at 9 02 16 AM

### DIFF
--- a/frontend/src/__mocks__/mockRegisteredModel.ts
+++ b/frontend/src/__mocks__/mockRegisteredModel.ts
@@ -4,6 +4,8 @@ import { createModelRegistryLabelsObject } from './utils';
 type MockRegisteredModelType = {
   id?: string;
   name?: string;
+  author?: string;
+  owner?: string;
   state?: ModelState;
   description?: string;
   labels?: string[];
@@ -11,6 +13,8 @@ type MockRegisteredModelType = {
 
 export const mockRegisteredModel = ({
   name = 'test',
+  author = 'Author 1',
+  owner = 'Author 1',
   state = ModelState.LIVE,
   description = '',
   labels = [],
@@ -23,5 +27,7 @@ export const mockRegisteredModel = ({
   lastUpdateTimeSinceEpoch: '1710404288975',
   name,
   state,
+  author,
+  owner,
   customProperties: createModelRegistryLabelsObject(labels),
 });
--- a/frontend/src/__mocks__/mockRegisteredModel.ts
+++ b/frontend/src/__mocks__/mockRegisteredModel.ts
@@ -4,7 +4,6 @@ import { createModelRegistryLabelsObject } from './utils';
 type MockRegisteredModelType = {
   id?: string;
   name?: string;
-  author?: string;
   owner?: string;
   state?: ModelState;
   description?: string;
@@ -13,7 +12,6 @@ type MockRegisteredModelType = {
 
 export const mockRegisteredModel = ({
   name = 'test',
-  author = 'Author 1',
   owner = 'Author 1',
   state = ModelState.LIVE,
   description = '',
@@ -27,7 +25,6 @@ export const mockRegisteredModel = ({
   lastUpdateTimeSinceEpoch: '1710404288975',
   name,
   state,
-  author,
   owner,
   customProperties: createModelRegistryLabelsObject(labels),
 });
--- a/frontend/src/__tests__/cypress/cypress/tests/mocked/modelRegistry/modelRegistry.cy.ts
+++ b/frontend/src/__tests__/cypress/cypress/tests/mocked/modelRegistry/modelRegistry.cy.ts
@@ -66,7 +66,10 @@ const initIntercepts = ({
       ],
     }),
   ],
-  modelVersions = [mockModelVersion({ name: 'model version' })],
+  modelVersions = [
+    mockModelVersion({ author: 'Author 1' }),
+    mockModelVersion({ name: 'model version' }),
+  ],
   allowed = true,
 }: HandlersProps) => {
   cy.interceptOdh(
--- a/frontend/src/__tests__/cypress/cypress/tests/mocked/modelRegistry/modelRegistry.cy.ts
+++ b/frontend/src/__tests__/cypress/cypress/tests/mocked/modelRegistry/modelRegistry.cy.ts
@@ -178,7 +178,14 @@ describe('Model Registry core', () => {
         .contains(
           'A machine learning model trained to detect fraudulent transactions in financial data',
         );
+<<<<<<< HEAD
       registeredModelRow.findAuthor().contains('Author 1');
+=======
+      registeredModelRow.findOwner().should(($owner) => {
+        const text = $owner.text();
+        expect(text).to.be.oneOf(['Author 1', '-']);
+      });
+>>>>>>> 8ca179e7 (fixed lint and test)
 
       // Label popover
       registeredModelRow.findLabelPopoverText().contains('2 more');
--- a/frontend/src/__tests__/cypress/cypress/tests/mocked/modelRegistry/modelRegistry.cy.ts
+++ b/frontend/src/__tests__/cypress/cypress/tests/mocked/modelRegistry/modelRegistry.cy.ts
@@ -178,21 +178,7 @@ describe('Model Registry core', () => {
         .contains(
           'A machine learning model trained to detect fraudulent transactions in financial data',
         );
-<<<<<<< HEAD
-<<<<<<< HEAD
-      registeredModelRow.findAuthor().contains('Author 1');
-=======
-=======
       registeredModelRow.findOwner().contains('Author 1');
-<<<<<<< HEAD
->>>>>>> f9da61fb (updated tests)
-      registeredModelRow.findOwner().should(($owner) => {
-        const text = $owner.text();
-        expect(text).to.be.oneOf(['Author 1', '-']);
-      });
->>>>>>> 8ca179e7 (fixed lint and test)
-=======
->>>>>>> 05749f70 (removed assertion)
 
       // Label popover
       registeredModelRow.findLabelPopoverText().contains('2 more');
--- a/frontend/src/__tests__/cypress/cypress/tests/mocked/modelRegistry/modelRegistry.cy.ts
+++ b/frontend/src/__tests__/cypress/cypress/tests/mocked/modelRegistry/modelRegistry.cy.ts
@@ -181,12 +181,15 @@ describe('Model Registry core', () => {
 =======
 =======
       registeredModelRow.findOwner().contains('Author 1');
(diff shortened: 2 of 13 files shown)


frontend/src/__tests__/cypress/cypress/tests/mocked/modelRegistry/modelRegistry.cy.ts
```diff
@@ -178,7 +178,10 @@ describe('Model Registry core', () => {
         .contains(
           'A machine learning model trained to detect fraudulent transactions in financial data',
         );
-      registeredModelRow.findOwner().contains('Author 1');
+      registeredModelRow.findOwner().should(($owner) => {
+        const text = $owner.text();
+        expect(text).to.be.oneOf(['Author 1', '-']);
```
gitdallas, line 183: I think we should test both ways here, and that we'll want to test both ways (that the author shows up if the author prop is populated, that the '-' shows up if not).

gitdallas (changes requested):

frontend/src/__tests__/cypress/cypress/tests/mocked/modelRegistry/modelRegistry.cy.ts
YuliaKrimerman: That line of code started failing with the new change, and I wasn't quite able to figure out why ( findOwner has the same textId 'registered-model-owner' so it should be good) . It's complaining that :"Timed out retrying after 10000ms: Expected to find content: 'Author 1' within the element: <p> but never did". Do you have maybe a suggestion what can be missing?

gitdallas: It looks like you haven't updated the mock types and data. They still have the author in the modelVersion.

YuliaKrimerman: So if I understood correctly I should delete "mockModelVersion({ author: 'Author 1' })" line 70 in that test file? - I tried, that didn't help. I also tried to delete that field from all the 3 parts within MockModelVersionType but then the test suite didn't even run.

gitdallas: You'll need to also add the author prop to the ModelRegistry where your code is looking for it.

As for the tests not running in MockModelVersionType after removing all the model versions. I'm not sure, is that completely removed from everywhere or just within this case? If removed generally and tests fail after mocking that way, then there must be some test that is still falsely depending on that.

gitdallas: looks like you need both `owner` and `author` on the `mockRegisteredModel` stuff

gitdallas: ah, actually i can't do that as the `mockRegisteredModel.ts` file wasn't touched yet. but basically:

image


then you could update the calls to `mockRegisteredModel` in `modelRegistry.cy.ts` to whatever owner/author you want, and the test to match. could use `Author 1` if you want, but i think something with "Owner" would make more sense for the owner field so people don't get the two mixed up in reviewing/making tests.

frontend/src/pages/modelRegistry/screens/ModelVersions/ModelDetailsView.tsx
```diff
@@ -83,7 +82,7 @@ const ModelDetailsView: React.FC<ModelDetailsViewProps> = ({ registeredModel: rm
             </ClipboardCopy>
           </DashboardDescriptionListGroup>
           <DashboardDescriptionListGroup title="Owner">
-            <RegisteredModelOwner registeredModelId={rm.id} />
+            <Text data-testid="registered-model-owner">{rm.author || '-'}</Text>
```
gitdallas, line 85: should this actually be `rm.owner`?

frontend/src/__tests__/cypress/cypress/tests/mocked/modelRegistry/modelRegistry.cy.ts
gitdallas: or is there even an author field? you may have just had a typo below in`ModelDetailView`

gitdallas (changes requested):

frontend/src/__tests__/cypress/cypress/tests/mocked/modelRegistry/modelRegistry.cy.ts
gitdallas: ah, actually i can't do that as the `mockRegisteredModel.ts` file wasn't touched yet. but basically:

image


then you could update the calls to `mockRegisteredModel` in `modelRegistry.cy.ts` to whatever owner/author you want, and the test to match. could use `Author 1` if you want, but i think something with "Owner" would make more sense for the owner field so people don't get the two mixed up in reviewing/making tests.

or is there even an author field? you may have just had a typo below in`ModelDetailView`

gitdallas: Sorry, I just wrapped my head around the story and changes a bit more. I think you need to still treat `author` the same way and keep it in modelVersion and just add `owner` to the registeredModel. So the RM has an "owner" while the version has an "author".

Maybe @mturley has some input on this

mturley: Yes, it's confusing but the RegisteredModel has an `owner` but no author and the ModelVersion (and ModelArtifact, though that's not relevant) has an `author` but no owner. Currently in the UI we are showing "Owner" and under the hood we're using the version's `author` field.

There are two changes that prompted this story:
1. The database now has `owner` on RegisteredModel, which previously didn't have either owner or author (which is why we were fetching the author from the model's latest version).
2. The UX microcopy is being updated so that we call it in the UI what it is in the database -- for models we call it "Owner" and use the `.owner` field, and for versions we'll call it "Author" and use the `.author` field.

mturley: That latter change to use "Author" in the UI wasn't originally part of this Jira issue, it was from a followup discussion, but I think making that change in this PR would be least confusing. Is that ok @YuliaKrimerman ?

frontend/src/__mocks__/mockRegisteredModel.ts
```diff
@@ -4,13 +4,17 @@ import { createModelRegistryLabelsObject } from './utils';
 type MockRegisteredModelType = {
   id?: string;
   name?: string;
+  author?: string;
+  owner?: string;
```
mturley, line 8: we should only have `owner` here, not `author`. Same for the other two instances below in this file.

frontend/src/concepts/modelRegistry/types.ts
```diff
@@ -85,6 +85,8 @@ export type ModelRegistryBase = {
   id: string;
   name: string;
   externalID?: string;
+  author?: string;
+  owner?: string;
```
mturley, line 89: We shouldn't add either of these to `ModelRegistryBase` - I know I had originally added author here when we were pairing on this @YuliaKrimerman but I was wrong, sorry. I hadn't realized it was called `owner` for RegisteredModels.

The type updates we need are:

ModelRegistryBase - no change
RegisteredModel - add `owner`
ModelArtifact - no change (don't remove `author`)
ModelVersion - no change (don't remove `author`)

frontend/src/pages/modelRegistry/screens/RegisterModel/utils.ts
```diff
@@ -33,6 +33,7 @@ export const registerModel = async (
       name: formData.modelName,
       description: formData.modelDescription,
       customProperties: {},
+      author,
```
mturley, line 36: This line will need to be:
```
owner: author,
```
because the property we're adding on RegisteredModel is `owner`, but we can leave the function argument called `author` because we're using the same value for the `author` of the ModelVersion and ModelArtifact in the `registerVersion` call below.

Alternatively, if you want you can change this `registerModel` function so it takes an `owner` instead of `author`, then this line is just `owner,` and then when we call `registerVersion()` below you can pass `owner` instead of `author`.

I feel like I'm talking in circles, hopefully this makes sense.

frontend/src/pages/modelRegistry/screens/__tests__/utils.spec.ts
```diff
@@ -270,6 +270,8 @@ describe('getPatchBody', () => {
   it('returns a given RegisteredModel with id/name/timestamps removed, customProperties updated and other values unchanged', () => {
     const registeredModel = mockRegisteredModel({
       id: '1',
+      author: 'Author 1',
+      owner: 'Author 1',
```
mturley, line 274: Same deal here, we only need `owner` for models.

frontend/src/pages/modelRegistry/screens/__tests__/utils.spec.ts
```diff
@@ -286,9 +288,11 @@ describe('getPatchBody', () => {
     );
     expect(result).toEqual({
       description: 'Description here',
+      author: 'Author 1',
```
mturley, line 291: So we'll want to remove `author` here as well.

frontend/src/__tests__/cypress/cypress/tests/mocked/modelRegistry/modelRegistry.cy.ts
mturley: Actually --- scratch that last point, sorry. We don't have to touch anything with how Versions display the owner/author, that's out of scope here.

mturley (commented):
A couple changes here.

YuliaKrimerman:
Ready for hopefully final review @mturley @gitdallas

gitdallas (approved):

manaswinidas (approved):
/lgtm

Checked with Postman... the owner field gets picked up fine

mturley (approved):
LGTM!

PR #3099, opened 2024-08-16
Add drag and drop to connection type fields table

emilys314:
Closes https://issues.redhat.com/browse/RHOAIENG-10314 by completing the table functionality. 

## Description

Adds drag and drop. The example on patternfly makes the rows switch by changing the html in the `move()`, which is difficult to sync up with state. There is an existing hook that uses this method. But a new hook was created and used which will make the drag and drop directly update the react state of the rows. It also will update if the parent state updates (so if a new row is added, deleted, or modified, the hook can handle it)

dragdrop

## How Has This Been Tested?

Tested locally with different types of fields, and adding, deleting rows, and editing existing rows. To make sure all the ids line up when dragging them around.

Go to the create / edit / duplicate page, you'll see the new drag icon cell on the rows. You can mouse down anywhere on the row and drag and drop it. 

## Test Impact

I've added a cypress mock test for the edit command, which makes use of function for triggering drag and drop commands. This function can be used for reuse.

## Request review criteria:

If you have UI changes: 


frontend/src/pages/connectionTypes/manage/ManageConnectionTypeFieldsTableRow.tsx
```diff
@@ -65,13 +72,16 @@ const ManageConnectionTypeFieldsTableRow: React.FC<Props> = ({
...
-    <Tr id={row.name} data-testid="row">
+    <Tr id={row.id} draggable data-testid="row" {...props}>
+      <Td
+        draggableRow={{
+          id: `draggable-row-${row.id}`,
+        }}
+      />
       <Td dataLabel={columns[0]} data-testid="field-name">
         {row.name}
         <TextContent>
```
jeff-phillips-18, line 83:
```suggestion
        <div className="pf-v5-u-color-200">
```

For the section heading above, we should add the same `<Truncate>` for the description.

frontend/src/pages/connectionTypes/manage/ManageConnectionTypeFieldsTableRow.tsx
```diff
@@ -65,13 +72,16 @@ const ManageConnectionTypeFieldsTableRow: React.FC<Props> = ({
...
-          <Text className="pf-v5-u-color-200">
-            <Truncate content={row.description ?? ''} />
-          </Text>
+          <Truncate className="pf-v5-u-color-200" content={row.description ?? ''} />
         </TextContent>
```
jeff-phillips-18, line 85:
```suggestion
        </div>
```

frontend/src/utilities/useDraggableTableControlled.ts
```diff
@@ -0,0 +1,172 @@
+import React from 'react';
+import styles from '@patternfly/react-styles/css/components/Table/table';
+
+export type Row<T> = T & {
+  id: string;
+};
+
+type UseDraggableTable<T> = {
```
jeff-phillips-18, line 8:
```suggestion
type UseDraggableTableControlled<T> = {
```

frontend/src/utilities/useDraggableTableControlled.ts
```diff
@@ -0,0 +1,172 @@
...
+  };
+  rowProps: {
+    onDragStart: React.DragEventHandler<HTMLTableRowElement>;
+    onDragEnd: React.DragEventHandler<HTMLTableRowElement>;
+    onDrop: React.DragEventHandler<HTMLTableRowElement>;
+  };
+  rowsToRender: Row<T>[];
+};
+
+const useDraggableTable = <T>(
```
jeff-phillips-18, line 25:
```suggestion
const useDraggableTableControlled = <T>(
```

frontend/src/utilities/useDraggableTableControlled.ts
```diff
@@ -0,0 +1,172 @@
+import React from 'react';
+import styles from '@patternfly/react-styles/css/components/Table/table';
+
+export type Row<T> = T & {
+  id: string;
+};
+
+type useDraggableTableControlled<T> = {
```
jeff-phillips-18, line 8:
```suggestion
type useDraggableTableControlledProps<T> = {
```

emilys314: Technically not the props, but the return type of the hook.

I did drop the capitalization accidentally. 

Looking at the other hooks in the utilities, it could  be `DraggableTableControlled`?

jeff-phillips-18: Yeah, `DraggableTableControlled` seems appropriate

PR #3100, opened 2024-08-16
model to manage field creation

christianvogt:
https://issues.redhat.com/browse/RHOAIENG-10320
https://issues.redhat.com/browse/RHOAIENG-10321
https://issues.redhat.com/browse/RHOAIENG-10323
https://issues.redhat.com/browse/RHOAIENG-10324
https://issues.redhat.com/browse/RHOAIENG-10325

partial: https://issues.redhat.com/browse/RHOAIENG-10322

## Description

Refactored the form components used for the preview and instance forms to now be used as part of the default value in the field modal.

Not all forms are complete as some forms still require setup such as dropdown before the default value input is usable.

The approach to reuse the same inputs for the default value needs a conversation with UX because their current mocks don't use these for all inputs. But reusing the existing form fields gives a very simple solution to assigning a default value.

image
image

cc @simrandhaliw 

## How Has This Been Tested?

Manually by creating and editing fields.

## Test Impact

Updated existing tests.

## Request review criteria:

If you have UI changes: 


frontend/src/concepts/connectionTypes/fields/UriFormField.tsx
```diff
@@ -1,29 +1,71 @@
...
+  const isPreview = mode === 'preview';
+
+  const isValid = validateUrl(value ?? '');
+
+  React.useEffect(() => {
+    onValidate?.(isValid);
+    // do not run when callback changes
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [isValid]);
+
```
jeff-phillips-18, line 44: This is updating validity as soon as the user types. I think `isValid` should be a state variable that get set on entry and then updates only `onBlur`.

jeff-phillips-18 (approved):

PR #3109, opened 2024-08-21
pipeline scheduled toggle is disabled for schedules with an archived …

pnaik1:
…experiment

https://issues.redhat.com/browse/RHOAIENG-10393

## Description
In pipeline version schedules table, disable toggling when experiment is archived

## How Has This Been Tested?

1. Create a schedule with an active experiment
2. Archive that experiment
3. Go to the pipeline version schedules table, and check that schedule, you can see the toggle button is disabled

## Test Impact
added cypress test cases

## Request review criteria:

If you have UI changes: 


DaoDaoNoCode (changes requested):
This is not the optimum solution and will create a lot of duplicated network calls.
1. In https://github.com/opendatahub-io/odh-dashboard/blob/main/frontend/src/concepts/pipelines/content/tables/pipelineRun/PipelineRunTableRowExperiment.tsx, we will make a call to fetch the experiment for each row already, we should look into extract that logic and make it reused by both the experiment column and the status toggle so that we can reduce the API calls, otherwise every row will make 2 same API calls to fetch the experiment.
2. The changes you made to `useContextExperimentArchived` hook will also produce some duplicated API calls on the experiment schedules pages, which is unnecessary. On the experiment schedules table, everything is under the experiment context, and we should make sure we don't make extra API calls for the experiment again and fetch all the information we need from the experiment context because all the schedules on that page are from the same experiment.

DaoDaoNoCode (approved):
Tested, works great
/lgtm
Screenshot 2024-08-26 at 12 53 48 PM

Gkrumbach07 (approved):
/approve

PR #3110, opened 2024-08-21
 removed the use of Patch Body

YuliaKrimerman:
https://issues.redhat.com/browse/RHOAIENG-10997)

## Description

Removed the workaround for Patching the fields from API. As described in the issue, the description, labels, and Properties should remain when I archive the model. See attached Screenshots 
Screenshot 2024-08-21 at 10 18 06 AM
Screenshot 2024-08-21 at 10 18 23 AM

## How Has This Been Tested?

Tested on the UI see above sreenshots
## Test Impact

I've deleted some tests that were related to the Patching functions, Also updated some fields in the tests now to match the behavior as the remaining fields in those tests are not needed anymore, because they are coming from a different place now.

## Request review criteria:

If you have UI changes: 


mturley (approved):
Good stuff, LGTM and tested locally against the modelregistry-ui cluster.

PR #3117, opened 2024-08-22
[RHOAIENG-11529] Add numeric field advanced settings for connection type fields

jeff-phillips-18:
Closes [RHOAIENG-11529](https://issues.redhat.com/browse/RHOAIENG-11529)

## Description
Adds the advanced settings for defining a numeric connection type field

## How Has This Been Tested?
Tested manually. Added RTL tests.

## Screen shots
image

image

image

image

image

## Request review criteria:

If you have UI changes: 

/cc @simrandhaliw


frontend/src/components/EditableLabelsDescriptionListGroup.tsx
```diff
@@ -200,6 +200,7 @@ const EditableLabelsDescriptionListGroup: React.FC<EditableTextDescriptionListGr
               isRequired
               validated={addLabelValidationError ? 'error' : 'default'}
             />
+            `{' '}
```
christianvogt, line 203: Why this change?

frontend/src/components/NumberInputWrapper.scss
```diff
@@ -0,0 +1,15 @@
+.odh-number-input-wrapper {
+  &.m-full-width {
```
christianvogt, line 2: Why not push back to UX if the the PF component doesn't support full width?
is there a bug logged to PF and does PF approve of such designs?

frontend/src/concepts/pipelines/content/createRun/contentSections/MaxConcurrencyField.tsx
```diff
@@ -21,7 +21,7 @@ const MaxConcurrencyField: React.FC<MaxConcurrencyFieldProps> = ({ onChange, val
       data-testid="maxConcurrencyField"
       min={1}
       max={10}
-      onChange={onChange}
+      onChange={(newValue) => onChange(newValue ?? 0)}
```
christianvogt, line 24: Should we put this on everyone or give `NumberInput` a prop that can provide the fallback value and have it default to `0` as previously done?

frontend/src/pages/connectionTypes/manage/advanced/NumericAdvancedPropertiesForm.tsx
```diff
@@ -0,0 +1,141 @@
...
+  }, [properties.unit]);
+
+  return (
+    <ExpandableSection
+      toggleText="Advanced settings"
+      onToggle={(_ev, isExpanded) => setExpanded(isExpanded)}
+      isExpanded={expanded}
+      data-testid="advanced-settings-toggle"
+    >
+      <FormGroup className="pf-v5-u-ml-lg" label="Unit" fieldId="unit-select">
```
christianvogt, line 58: Use `FormFieldGroupExpandable` ?

frontend/src/components/EditableLabelsDescriptionListGroup.tsx
jeff-phillips-18: oops

frontend/src/concepts/pipelines/content/createRun/contentSections/MaxConcurrencyField.tsx
jeff-phillips-18: We need it to come back as `undefined` for inputs that want to be able to unset the value after it has been set (undefined rather that 0). Allowing the fallback to be `undefined` would require this change anyhow (though we'd know it would never really receive undefined).

frontend/src/components/NumberInputWrapper.scss
jeff-phillips-18: I can log a request for PF support. I think it looks better and is more usable at full width so I didn't want to push back. Also, I am seeing an issue with the units value wrapping for some reason, hence the:
```
    .pf-v5-c-number-input__unit {
      white-space: nowrap;
    }
```

frontend/src/pages/connectionTypes/manage/advanced/NumericAdvancedPropertiesForm.tsx
jeff-phillips-18: Created the `ExpandableFormSection` common component

emilys314:
I checked out the PR image and tested out the advanced dropdown. It all seems to work, but I did find one behavior which I'm not 100% sure about.

So if you have a min or max, the default value - / + is disable from going beyond these. But if you use the up or down arrow keys to change the value, you can go beyond the min / max. It still gives an error and is handled correctly, but it's a little interesting. 

 <details>
 <summary>gif</summary>

increment

</details>

jeff-phillips-18:
> So if you have a min or max, the default value - / + is disable from going beyond these. But if you use the up or down arrow keys to change the value, you can go beyond the min / max. It still gives an error and is handled correctly, but it's a little interesting.

@simrandhaliw What are your thoughts on this? Should we prevent them completely from entering a default value that is out of bounds or should we allow them to enter an invalid default value (they will see the error text) and then adjust the value or the thresholds in order to correct it?

jeff-phillips-18:
@simrandhaliw Updated error messages:

image


------


image

------


image

PR #3119, opened 2024-08-22
fix cypress flake in modelVersionDeployModal.cy.ts

christianvogt:
https://issues.redhat.com/browse/RHOAIENG-11791

## Description

The `DeployRegistrationModelModal` renders 3 separate modals depending on the project selected. When the cypress tests switches modals, it acts too fast to interact again with the dropdown only to have the UI unmount the current modal and mount a new one. Since our custom commands store the selector dom node in order to perform multiple actions, this node becomes stale and no longer applicable to the current DOM. Thus the command fails.

Changed the selector for clicking on them project selector menu item to be two separate commands.

Had to update the code to move the project selector open state management to the parent so that all modal instances share the same state, otherwise the dropdown could close when rendering a new modal.

## How Has This Been Tested?

Run cypress test `modelVersionDeployModal.cy.ts` for the primary fix.

## Test Impact

N/A

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/__tests__/cypress/cypress/pages/modelRegistry/modelVersionDeployModal.ts
+++ b/frontend/src/__tests__/cypress/cypress/pages/modelRegistry/modelVersionDeployModal.ts
@@ -6,11 +6,12 @@ class ModelVersionDeployModal extends Modal {
   }
 
   findProjectSelector() {
-    return cy.findByTestId('deploy-model-project-selector');
+    return this.find().findByTestId('deploy-model-project-selector');
   }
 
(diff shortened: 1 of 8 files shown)


christianvogt:
Had to update some selectors in the pipeline tests to work with changes in custom commands for dropdowns.

Also rebased.

christianvogt:
Ran into another flake in `modelVersionArchive.cy.ts` which needed a fix to suppress uncaught exceptions.

jeff-phillips-18 (approved):

PR #3124, opened 2024-08-26
surface conflicting env var names as warnings and errors

christianvogt:
https://issues.redhat.com/browse/RHOAIENG-11538

## Description

Provides user with feedback when creating fields which have conflicting env var values.
Note that microcopy is not finalized.

image

image

cc @simrandhaliw 

## How Has This Been Tested?

- Go to `Settings` -> `Connection types` as an admin with the connection type feature flag enabled.
- Create a new connection type.
- Add two fields with the same env var name.
- Observe the error in the table and fields modal and that the form save button is disabled
- Edit a conflicting field to have a unique env var
- Observe the form save button is no longer disabled (providing the rest of the form is valid)

## Test Impact

Added unit tests.

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/pages/connectionTypes/manage/ConnectionTypeDataFieldModal.tsx
+++ b/frontend/src/pages/connectionTypes/manage/ConnectionTypeDataFieldModal.tsx
@@ -14,14 +14,19 @@ import {
   SelectOption,
   TextArea,
   TextInput,
-  ValidatedOptions,
 } from '@patternfly/react-core';
-import { ExclamationCircleIcon, OutlinedQuestionCircleIcon } from '@patternfly/react-icons';
+import {
+  ExclamationCircleIcon,
+  OutlinedQuestionCircleIcon,
+  WarningTriangleIcon,
+} from '@patternfly/react-icons';
 import DashboardModalFooter from '~/concepts/dashboard/DashboardModalFooter';
 import {
   ConnectionTypeDataField,
   connectionTypeDataFields,
+  ConnectionTypeField,
   ConnectionTypeFieldType,
+  isConnectionTypeDataField,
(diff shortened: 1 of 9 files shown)


frontend/src/pages/connectionTypes/manage/ConnectionTypeDataFieldModal.tsx
```diff
@@ -75,14 +82,13 @@ export const ConnectionTypeDataFieldModal: React.FC<Props> = ({
...
-    !envVar || ENV_VAR_NAME_REGEX.test(envVar) ? ValidatedOptions.default : ValidatedOptions.error;
-  const isValid =
-    !!fieldType &&
-    isPropertiesValid &&
-    !!name &&
-    !!envVar &&
-    envVarValidation === ValidatedOptions.default;
+  const isEnvVarConflict = !!fields?.find(
+    (f) => f !== field && isConnectionTypeDataField(f) && f.envVar === envVar,
+  );
```
emilys314, line 87: Do we need to memo-ize this? I don't think it's that computationally expensive, but I think this would still rerun on every key stroke when `name` or `description` states update?

emilys314:
I tested this on my cluster and it looks like it's all working to me

frontend/src/pages/connectionTypes/manage/ConnectionTypeDataFieldModal.tsx
christianvogt: Updated with `useMemo`

christianvogt:
@emilys314 updated with `useMemo` where env vars array is computed

christianvogt:
rebased

christianvogt:
rebased

PR #3125, opened 2024-08-26
[RHOAIENG-1105] Add Storage Class section in Admin Panel and Feature Flag

jpuzz0:
https://issues.redhat.com/browse/RHOAIENG-1105

## Description
- Added feature flag for storage classes
- Showing empty state when no storage classes are found
- Showing blank page in anticipation for a table to be rendered as a part of [RHOAIENG-1106](https://issues.redhat.com/browse/RHOAIENG-1106) coming soon
- Added logic on render to add annotations to storage class metadata if those storage classes are absent of the `sc-config` annotation.

**Empty state**
Pasted Graphic 1

**With feature turned off or as a non-admin user**
image

(cc @xianli123)

## How Has This Been Tested?
Manual testing with creation of annotations, although that data can be inspected in the network tab. This will be visualized in [RHOAIENG-1106](https://issues.redhat.com//browse/RHOAIENG-1106), so the expectation is to test this logic more thoroughly there. Otherwise, cypress tests have been added to verify what can and can't be seen as a regular user and admin user, along with empty state.

This can be tested by navigating to `/storageClasses?devFeatureFlags=disableStorageClasses`. If storage classes exist, the title will show without content since the table will eventually be rendered there. If no storage classes exist, you'll see the empty state as shown in the screenshot above. If you do not enable the feature via the dev feature flag queryparam or by other means, the nav item won't show and the "page not found" page will be rendered as seen in the screenshot above.

## Test Impact
Added initial cypress tests

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/app/AppRoutes.tsx
+++ b/frontend/src/app/AppRoutes.tsx
@@ -73,7 +73,6 @@ const AppRoutes: React.FC = () => {
   const isJupyterEnabled = useCheckJupyterEnabled();
   const isHomeAvailable = useIsAreaAvailable(SupportedArea.HOME).status;
   const isConnectionTypesAvailable = useIsAreaAvailable(SupportedArea.CONNECTION_TYPES).status;
-  const isStorageClassesAvailable = useIsAreaAvailable(SupportedArea.STORAGE_CLASSES).status;
 
   if (!isAllowed) {
     return (
@@ -131,9 +130,7 @@ const AppRoutes: React.FC = () => {
             {isConnectionTypesAvailable ? (
               <Route path="/connectionTypes/*" element={<ConnectionTypeRoutes />} />
             ) : null}
-            {isStorageClassesAvailable && (
-              <Route path="/storageClasses/*" element={<StorageClassesPage />} />
-            )}
+            <Route path="/storageClasses/*" element={<StorageClassesPage />} />
             <Route path="/modelRegistrySettings/*" element={<ModelRegistrySettingsRoutes />} />
             <Route path="/groupSettings" element={<GroupSettingsPage />} />
           </>
--- a/frontend/src/app/AppRoutes.tsx
+++ b/frontend/src/app/AppRoutes.tsx
@@ -64,6 +64,8 @@ const AcceleratorProfileRoutes = React.lazy(
   () => import('../pages/acceleratorProfiles/AcceleratorProfilesRoutes'),
 );
 
+const StorageClassesPage = React.lazy(() => import('../pages/storageClasses/StorageClassesPage'));
+
 const ModelRegistryRoutes = React.lazy(() => import('../pages/modelRegistry/ModelRegistryRoutes'));
 
 const AppRoutes: React.FC = () => {
@@ -71,6 +73,7 @@ const AppRoutes: React.FC = () => {
   const isJupyterEnabled = useCheckJupyterEnabled();
   const isHomeAvailable = useIsAreaAvailable(SupportedArea.HOME).status;
   const isConnectionTypesAvailable = useIsAreaAvailable(SupportedArea.CONNECTION_TYPES).status;
+  const isStorageClassesAvailable = useIsAreaAvailable(SupportedArea.STORAGE_CLASSES).status;
 
   if (!isAllowed) {
     return (
@@ -128,6 +131,9 @@ const AppRoutes: React.FC = () => {
             {isConnectionTypesAvailable ? (
               <Route path="/connectionTypes/*" element={<ConnectionTypeRoutes />} />
             ) : null}
+            {isStorageClassesAvailable && (
+              <Route path="/storageClasses/*" element={<StorageClassesPage />} />
+            )}
             <Route path="/modelRegistrySettings/*" element={<ModelRegistrySettingsRoutes />} />
             <Route path="/groupSettings" element={<GroupSettingsPage />} />
           </>
--- a/frontend/src/k8sTypes.ts
+++ b/frontend/src/k8sTypes.ts
@@ -55,13 +55,19 @@ export type StorageClassConfig = {
   description?: string;
 };
 
+export enum Annotation {
+  StorageClassIsDefault = 'storageclass.kubernetes.io/is-default-class',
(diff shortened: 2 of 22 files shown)


xianli123:
Hi @jpuzz0 I have a question about the second screenshot you shared above that is when users will see this page? Thanks!

image

frontend/src/app/AppRoutes.tsx
```diff
@@ -128,6 +131,9 @@ const AppRoutes: React.FC = () => {
             {isConnectionTypesAvailable ? (
               <Route path="/connectionTypes/*" element={<ConnectionTypeRoutes />} />
             ) : null}
+            {isStorageClassesAvailable && (
+              <Route path="/storageClasses/*" element={<StorageClassesPage />} />
+            )}
             <Route path="/modelRegistrySettings/*" element={<ModelRegistrySettingsRoutes />} />
```
Gkrumbach07, line 137: dont prevent users from accessing this page, just hide the nav item. follow how the other admin routes are.

I do see ConnectionTypes is like this, but it shouldn't be. unless Andrew is changing the way he wants the flags to work recently

frontend/src/k8sTypes.ts
```diff
@@ -55,13 +55,19 @@ export type StorageClassConfig = {
   description?: string;
 };
 
+export enum Annotation {
```
Gkrumbach07, line 58: if you are exporting this, i would be more specific about what annotations these are. Something like CommonAnnotations might be fine. but could be anything

frontend/src/pages/storageClasses/StorageClassesPage.tsx
```diff
@@ -0,0 +1,111 @@
...
+  React.useEffect(() => {
+    storageClasses.forEach((storageClass, index) => {
+      const { metadata } = storageClass;
+      const { name: storageClassName } = metadata;
+
+      if (!metadata.annotations?.[Annotation.OdhStorageClassConfig]) {
+        const isDefault = defaultStorageClass?.metadata.uid === metadata.uid;
+        const storageClassConfig: StorageClassConfig = {
+          displayName: storageClassName,
+          ...(!defaultStorageClass
```
Gkrumbach07, line 43: imagine someone adds a new storage class after we have already configured the configs for the previous classes. 

if there was no openshift default set, then the previous should all be set to enabled. which is what you have here

but now you later add a new storage class, what happens, becuase there is still no default, it will set it as enabled, however you would have already st a ODH default class. 

so what i am coming to is that we should only enable if there is no odh default or there is not openshift default

wdyt

frontend/src/pages/storageClasses/StorageClassesPage.tsx
```diff
@@ -0,0 +1,111 @@
...
+            ? {
+                isEnabled: true,
+                isDefault: index === 0,
+              }
+            : {
+                isDefault,
+                isEnabled: isDefault,
```
Gkrumbach07, line 50: we need to constrain this more. if you have existing configured storage classes, but then later on create a new one and then set it as the new default. it would try to set it as default here when we already have a odh default set

Gkrumbach07 (commented):
looks good. just some edge cases on the setup of the storage classes

DaoDaoNoCode (approved):
Tested locally, works well, all the configs are auto added. I think just need to address the changes requested by Gage and it's good to go.
/lgtm

jpuzz0:
> Hi @jpuzz0 I have a question about the second screenshot you shared above that is when users will see this page? Thanks!
> 
> image

@xianli123 As the description and title above that screenshot states, `With feature turned off or as a non-admin user` meaning when the route doesn't exist because the feature is turned off (Gage asked me just now to remove this logic, so this will no longer be true) or if the user is a non-admin user and tries visiting the `/storageClasses` url manually from the browser.

frontend/src/app/AppRoutes.tsx
jpuzz0: Seemed to make sense to me to not allow even admins to view the page if the feature isn't ready, but I can remove the condition.

frontend/src/k8sTypes.ts
jpuzz0: I went with `MetadataAnnotation` to be more specific. Outside of that, they are k8s annotations used for metadata and they're in a file called k8sTypes, so I think that should suffice.

frontend/src/pages/projects/screens/spawner/storage/usePreferredStorageClass.ts
```diff
@@ -10,10 +10,9 @@ const usePreferredStorageClass = (): StorageClassKind | undefined => {
...
-  const defaultClusterStorageClasses = storageClasses.filter((storageclass) =>
-    storageclass.metadata.annotations?.['storageclass.kubernetes.io/is-default-class']?.includes(
-      'true',
-    ),
+  const defaultClusterStorageClasses = storageClasses.filter(
+    (storageclass) =>
+      storageclass.metadata.annotations?.[Annotation.StorageClassIsDefault] === 'true',
   );
 
   const configStorageClassName = notebookController?.storageClassName ?? '';
```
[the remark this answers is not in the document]
jpuzz0: I can add some logic to not use it when the feature flag is on.

frontend/src/pages/storageClasses/StorageClassesPage.tsx
jpuzz0: That makes sense. Will update.

frontend/src/pages/storageClasses/StorageClassesPage.tsx
jpuzz0: I think addressing you other comment above should address this.

xianli123:
Thanks @jpuzz0 . That makes sense to me.

frontend/src/app/AppRoutes.tsx
Gkrumbach07: More of just a consistency things. I tried to do it the way you had before and was told to not disable the route

Gkrumbach07 (approved):
/lgtm
/approve

code looks great

PR #3127, opened 2024-08-26
add backend route for storage class metadata

Gkrumbach07:
closes: https://issues.redhat.com/browse/RHOAIENG-11855

## Description

adds backend route `/api/storage-class/{name}/metadata` and a frontend service function to call it

## How Has This Been Tested?

backend -- no tests

## Test Impact

none

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/backend/src/routes/api/storage-class/index.ts
+++ b/backend/src/routes/api/storage-class/index.ts
@@ -1,0 +1,26 @@
+import { FastifyReply, FastifyRequest } from 'fastify';
+import { StorageClassConfig, KubeFastifyInstance } from '../../../types';
+import { secureAdminRoute } from '../../../utils/route-security';
+import { updateStorageClassMetadata } from './utils';
+
+export default async (fastify: KubeFastifyInstance): Promise<void> => {
+  fastify.put(
+    '/:storageClassName/metadata',
+    secureAdminRoute(fastify)(
+      async (
+        request: FastifyRequest<{
+          Body: StorageClassConfig;
+        }>,
+        reply: FastifyReply,
+      ) => {
+        return updateStorageClassMetadata(fastify, request)
+          .then((res) => {
+            return res;
+          })
+          .catch((res) => {
+            reply.send(res);
+          });
+      },
+    ),
+  );
+};
--- a/backend/src/routes/api/storage-class/index.ts
+++ b/backend/src/routes/api/storage-class/index.ts
@@ -1,19 +1,20 @@
 import { FastifyReply, FastifyRequest } from 'fastify';
 import { StorageClassConfig, KubeFastifyInstance } from '../../../types';
 import { secureAdminRoute } from '../../../utils/route-security';
-import { updateStorageClassMetadata } from './utils';
+import { updateStorageClassConfig } from './utils';
 
 export default async (fastify: KubeFastifyInstance): Promise<void> => {
   fastify.put(
-    '/:storageClassName/metadata',
+    '/:storageClassName/config',
     secureAdminRoute(fastify)(
       async (
         request: FastifyRequest<{
           Body: StorageClassConfig;
+          Params: { storageClassName: string };
         }>,
         reply: FastifyReply,
       ) => {
-        return updateStorageClassMetadata(fastify, request)
+        return updateStorageClassConfig(fastify, request)
           .then((res) => {
             return res;
           })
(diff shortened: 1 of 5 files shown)


DaoDaoNoCode (commented):
The function looks good to me. Just need to fix the lint to avoid duplicated import paths.
One quick question: Should we change the endpoint name and the function name here? Looks like the function is trying to update the SC annotation only, but not the whole metadata, should we still call them `/:storageClassName/metadata` and `updateStorageClassMetadata`?

frontend/src/services/StorageClassService.ts
```diff
@@ -0,0 +1,16 @@
+import axios from '~/utilities/axios';
+import { StorageClassConfig } from '~/k8sTypes';
+import { ResponseStatus } from '~/types';
+
+export const updateStorageClassMetadata = (
```
jpuzz0, line 5: Maybe this should be called `updateStorageClassConfig` since it only updates the config and at that point metadata could be implied.

backend/src/types.ts
```diff
@@ -624,6 +624,14 @@ export type ImageTagInfo = {
 
 export type ImageType = 'byon' | 'jupyter' | 'other';
 
+export type StorageClassConfig = {
```
jpuzz0, line 627: Doesn't this already exist in `src/k8sTypes`?

backend/src/routes/api/storage-class/index.ts
```diff
@@ -0,0 +1,26 @@
+import { FastifyReply, FastifyRequest } from 'fastify';
+import { StorageClassConfig, KubeFastifyInstance } from '../../../types';
+import { secureAdminRoute } from '../../../utils/route-security';
+import { updateStorageClassMetadata } from './utils';
+
+export default async (fastify: KubeFastifyInstance): Promise<void> => {
```
jpuzz0, line 6: Why not name this function? This is more of a specialized action for a file called "index" in a folder called "storage-class"

backend/src/routes/api/storage-class/utils.ts
```diff
@@ -0,0 +1,58 @@
+import { FastifyRequest } from 'fastify';
+import { K8sResourceCommon, KubeFastifyInstance, StorageClassConfig } from '../../../types';
+import { isHttpError } from '../../../utils';
+import { errorHandler } from '../../../utils';
+
+export async function updateStorageClassMetadata(
+  fastify: KubeFastifyInstance,
+  request: FastifyRequest<{ Body: StorageClassConfig }>,
+): Promise<{ success: boolean; error: string }> {
```
jpuzz0, line 9: I think `{ success: boolean; error: string }` is a type we have already called `ResponseStatus` in `src/types`

backend/src/routes/api/storage-class/utils.ts
```diff
@@ -0,0 +1,58 @@
...
+  const params = request.params as { storageClassName: string };
```
jpuzz0, line 10: Can we not adjust the incoming request type so we don't have to cast here?

backend/src/routes/api/storage-class/utils.ts
```diff
@@ -0,0 +1,58 @@
...
+  const body = request.body as StorageClassConfig;
```
jpuzz0, line 11: Is this cast necessary?

backend/src/routes/api/storage-class/utils.ts
```diff
@@ -0,0 +1,58 @@
+import { FastifyRequest } from 'fastify';
+import { K8sResourceCommon, KubeFastifyInstance, StorageClassConfig } from '../../../types';
+import { isHttpError } from '../../../utils';
+import { errorHandler } from '../../../utils';
+
+export async function updateStorageClassMetadata(
```
jpuzz0, line 6: Just to reiterate Juntao's comment, I think this is meant to be updated to be more specific: `updateStorageClassConfigAnnotation` or without the `Annotation`.

backend/src/types.ts
Gkrumbach07: this is for the backend. they dont share the same type files

backend/src/routes/api/storage-class/utils.ts
Gkrumbach07: for body not anymore, but params yes

backend/src/routes/api/storage-class/utils.ts
Gkrumbach07: sure, did that

backend/src/routes/api/storage-class/utils.ts
Gkrumbach07: that is a frontend type. i can add it to the backend too

backend/src/routes/api/storage-class/index.ts
Gkrumbach07: - the api works by finding the default export for each folder in `/api`, so there is no need to name the function then export it as default from another file
- this is also the pattern we use in all the other routes

jpuzz0 (approved):

DaoDaoNoCode (approved):
/lgtm

PR #3133, opened 2024-08-28
Update model registry status

ppadti:
Closes: [RHOAIENG-11278](https://issues.redhat.com/browse/RHOAIENG-11278)

## Description
This PR aims to update the status of the model registry by adding refresh interval.

## How Has This Been Tested?
1. create a model registry.
2. check that the status of the model registry will update when it is available.

## Test Impact
NA

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/concepts/modelRegistrySettings/useModelRegistriesBackend.ts
+++ b/frontend/src/concepts/modelRegistrySettings/useModelRegistriesBackend.ts
@@ -2,10 +2,11 @@ import React from 'react';
 import useFetchState, { FetchState } from '~/utilities/useFetchState';
 import { ModelRegistryKind } from '~/k8sTypes';
 import { listModelRegistriesBackend } from '~/services/modelRegistrySettingsService';
+import { POLL_INTERVAL } from '~/utilities/const';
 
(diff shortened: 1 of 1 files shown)


ppadti:
@mturley I am not sure should we use `FAST_POLL_INTERVAL` here until the status gets updated like we have for workbenches here: https://github.com/opendatahub-io/odh-dashboard/blob/main/frontend/src/pages/projects/screens/detail/notebooks/NotebookList.tsx#L25

gitdallas (approved):
lgtm, but i'm unsure of the answer to your question about using the faster interval. possibly something to discuss with design or pm

mturley:
@ppadti let's leave it as the 30-second `POLL_INTERVAL` for now. There's a separate problem we need to address post-MVP where the registry is showing as Unavailable right after it's created ([see thread here](https://redhat-internal.slack.com/archives/C07C0447EAV/p1724847598941919), I don't think we have an issue yet) which would complicate the logic of determining whether to use the faster poll interval here. If we decide we want to add that behavior after that is fixed we can revisit it I think.

mturley (approved):
LGTM, thanks @ppadti

PR #3139, opened 2024-08-29
Feat/5463 mr new design refactor

gitdallas:
closes: https://issues.redhat.com/browse/RHOAIENG-5463

## Description
several design changes

new emptystate images:
image
image
image

breadcrumbs/titles/icons:
image
image
image

model location parts
image
with uri
image

view details is back
image
on archived as well
image

## How Has This Been Tested?
locally, mr cluster

## Test Impact
updated tests where necessary

If you have UI changes: 

### DIFF
--- a/frontend/src/pages/modelRegistry/screens/ModelVersionDetails/ModelVersionRegisteredDeploymentsView.tsx
+++ b/frontend/src/pages/modelRegistry/screens/ModelVersionDetails/ModelVersionRegisteredDeploymentsView.tsx
@@ -6,8 +6,8 @@ import { Alert, Stack } from '@patternfly/react-core';
 import EmptyModelRegistryState from '~/pages/modelRegistry/screens/components/EmptyModelRegistryState';
 import InferenceServiceTable from '~/pages/modelServing/screens/global/InferenceServiceTable';
 import { getVersionDetailsInferenceServiceColumns } from '~/pages/modelServing/screens/global/data';
-import ModelVersionDetailsTabs from './ModelVersionDetailsTabs';
 import { typedEmptyImage, ProjectObjectType } from '~/concepts/design/utils';
+import ModelVersionDetailsTabs from './ModelVersionDetailsTabs';
 
 type ModelVersionRegisteredDeploymentsViewProps = Pick<
   React.ComponentProps<typeof ModelVersionDetailsTabs>,
@@ -24,7 +24,10 @@ const ModelVersionRegisteredDeploymentsView: React.FC<
       <EmptyModelRegistryState
         title="No deployments from model registry"
         headerIcon={() => (
-          <img src={typedEmptyImage(ProjectObjectType.registeredModels, "MissingDeployment")} />
+          <img
+            src={typedEmptyImage(ProjectObjectType.registeredModels, 'MissingDeployment')}
+            alt="missing deployment"
+          />
         )}
         description="No deployments initiated from model registry for this model version."
         testid="model-version-deployments-empty-state"
--- a/frontend/src/pages/modelRegistry/screens/ModelVersionDetails/ModelVersionRegisteredDeploymentsView.tsx
+++ b/frontend/src/pages/modelRegistry/screens/ModelVersionDetails/ModelVersionRegisteredDeploymentsView.tsx
@@ -7,6 +7,7 @@ import EmptyModelRegistryState from '~/pages/modelRegistry/screens/components/Em
 import InferenceServiceTable from '~/pages/modelServing/screens/global/InferenceServiceTable';
 import { getVersionDetailsInferenceServiceColumns } from '~/pages/modelServing/screens/global/data';
 import ModelVersionDetailsTabs from './ModelVersionDetailsTabs';
+import { typedEmptyImage, ProjectObjectType } from '~/concepts/design/utils';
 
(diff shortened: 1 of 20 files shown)


yih-wang:
Hi @gitdallas, thanks for all the updates! It looks good! Just several comments -

Regarding the empty state of the versions table, it looks like the spacing above and below the image is larger than on other screens. I just checked the source file, and it turns out the white spaces of the source image are bigger than in other images... My apologies for that. I'll follow up with the visual designer and see if we can get an image with less white space to you by Tuesday.

Regarding the model location,
- I can't see how the long text behaves from the screenshots. Just a quick reminder that the text should be truncated by one line if it exceeds the container length.
image

- For the S3 type of model location, would it be possible to group 'endpoint, region, bucket, and path' all under a single 'Model Location' section and assign the group a smaller spacing value? That would help clearly distinguish it from the following attributes and make the grouping more apparent. Check the [mockup here.](https://www.figma.com/proto/KTsYhANYQqB0NUrXW7wQH1/Model-Registry?page-id=6782%3A296930&node-id=9376-719497&node-type=FRAME&viewport=2220%2C-10581%2C0.33&t=Mfnz37mG3eAYGIfg-1&scaling=min-zoom&content-scaling=fixed&starting-point-node-id=9250%3A234038)
image

frontend/src/pages/modelRegistry/screens/ModelVersionDetails/ModelVersionRegisteredDeploymentsView.tsx
```diff
@@ -22,6 +23,12 @@ const ModelVersionRegisteredDeploymentsView: React.FC<
     return (
       <EmptyModelRegistryState
         title="No deployments from model registry"
+        headerIcon={() => (
+          <img
+            src={typedEmptyImage(ProjectObjectType.registeredModels, 'MissingDeployment')}
+            alt="missing deployment"
+          />
+        )}
```
manaswinidas, line 31: I don't find a testid here, so I think we don't have a Cypress test for empty-registered-deployments. Can you please add the test for this?

frontend/src/pages/modelRegistry/screens/RegisteredModels/RegisteredModelTableRow.tsx
```diff
@@ -32,19 +33,21 @@ const RegisteredModelTableRow: React.FC<RegisteredModelTableRowProps> = ({
   const [isRestoreModalOpen, setIsRestoreModalOpen] = React.useState(false);
   const rmUrl = registeredModelUrl(rm.id, preferredModelRegistry?.metadata.name);
 
-  const actions = isArchiveRow
-    ? [
-        {
+  const actions = [
+    {
+      title: 'View details',
```
manaswinidas, line 38: Can we add a Cypress test for `View details`(maybe tweak the existing tests to add this option)?

manaswinidas (changes requested):

mturley:
@gitdallas checking the model location fields, if the text is longer than the container width it does wrap to multiple lines instead of truncating:

Screenshot 2024-09-03 at 3 00 34 PM

I was looking at [the ClipboardCopy component docs](https://www.patternfly.org/components/clipboard-copy) though, and I don't think there's a way to make the inline variant truncate to one line. We could instead switch to the expandable variant, but it might be overkill if these values will usually fit on one line. What do you think @yih-wang ? I'm thinking maybe if there's no quick solution here we should create a post-MVP issue to follow up on this.

gitdallas:
@mturley - would something like this be okay? little hacky and you gotta override the `oncopy` https://codesandbox.io/p/sandbox/red-wind-zdgyqf?file=%2Findex.js%3A20%2C37

mturley:
Oh, interesting... yeah @gitdallas I think that solution looks good to me. I wonder if we should open an issue with PF to propose adding a prop or other variant for this, like `inline-compact-truncate` or something. Maybe for now let's factor out a. `ClipboardCopyInlineTruncate` component in `src/components` (or whatever you want to name it) and let's put a comment on that inline style line explaining that it's a workaround.

gitdallas:
@yih-wang - regarding the "smaller spacing", that feels off to me personally and i don't really notice it myself... if i do, it just seems like some sort of accident. if it's what is decided on, i can do it.

seems sort of like it might make sense to do a left margin, similar to how form sections work. although being a split screen already it's a bit low on horizontal space.

yih-wang:
@gitdallas thanks for sharing your thoughts! I had the same worry before about whether it was clear enough to group things in different spaces. Seems it definitely needs better ways to convey that. I'd like to think more about this so let's keep it as is. We have another post-MVP issue to add another new field 'source model format version' on this page and we can address this layout enhancement with that together.

gitdallas:
@yih-wang - with truncated clipboard copy:

image

frontend/src/pages/modelRegistry/screens/ModelVersionDetails/ModelVersionRegisteredDeploymentsView.tsx
gitdallas: there's a test-id a couple lines below this, and the test checking for it is in `modelVersionDetails.cy.ts:192`

mturley (approved):
LGTM, thanks @gitdallas

PR #3149, opened 2024-09-03
[RHOAIENG-11528] Update connection types admin screens with final microcopy

jeff-phillips-18:
Closes [RHOAIENG-11528](https://issues.redhat.com/browse/RHOAIENG-11528)

## Description
Updates the connection type admin screens with the latest micro copy found in [Figma mocks](https://www.figma.com/proto/97akqulXiGIGtvxkdPLjMm/Data-connections-enhancement?node-id=1168-156206&t=QAO1jyYn2P3dP1yX-0&scaling=scale-down&content-scaling=fixed&page-id=24%3A74&starting-point-node-id=1168%3A156206)

## How Has This Been Tested?
Manually tested

## Test Impact
Updated e2e tests where necessary

## Screen shots
image

image

image

image

image

image

## Request review criteria:

If you have UI changes: 

/cc @simrandhaliw

### DIFF
--- a/frontend/src/pages/connectionTypes/manage/ManageConnectionTypePage.tsx
+++ b/frontend/src/pages/connectionTypes/manage/ManageConnectionTypePage.tsx
@@ -134,7 +134,7 @@ const ManageConnectionTypePage: React.FC<Props> = ({ prefill, isEdit, onSave }) 
             <Alert
               isInline
               variant="warning"
-              title="Editing this connection will not affect existing connections of this type."
+              title="Editing this connection type does not affect existing connections of this type."
             />
           </PageSection>
         ) : undefined}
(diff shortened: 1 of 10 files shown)


frontend/src/pages/connectionTypes/manage/ManageConnectionTypePage.tsx
```diff
@@ -131,11 +131,11 @@ const ManageConnectionTypePage: React.FC<Props> = ({ prefill, isEdit, onSave })
...
-            <Alert
-              isInline
-              variant="warning"
-              title="Editing this connection will not affect existing connections of this type."
-            />
+            <Alert isInline variant="warning" title="Consider renaming before saving changes">
+              Editing this connection type does not affect existing connections of this type. To
+              keep track of previous versions of this connection type, consider renaming it before
+              saving changes.{' '}
+            </Alert>
```
christianvogt, line 138: There's a trailing `{' '}` that is unnecessary.

The copy here doesn't make sense because renaming in this form does not create a new copy. The user must duplicate the connection type prior to editing.

frontend/src/pages/connectionTypes/manage/ManageConnectionTypePage.tsx
```diff
@@ -166,7 +166,7 @@ const ManageConnectionTypePage: React.FC<Props> = ({ prefill, isEdit, onSave })
               </FormGroup>
               <FormGroup label="Enable" fieldId="connection-type-enable">
                 <Checkbox
-                  label="Enable users in your organization to use this connection type when adding connections."
+                  label="Connection is enabled and therefore available to use by users in your org"
```
christianvogt, line 169: Is this correct? The mocks have the original text.

jeff-phillips-18: The `Create` page had this text. I commented in the Jira and they updated the create page to match the original text. I will revert this.

frontend/src/pages/connectionTypes/manage/ManageConnectionTypePage.tsx
jeff-phillips-18: I think the thought is that if you rename it the existing connections of this type will still be named the old name while new names with these edits will have a different name. At least then the users could tell them apart even though the old connection type name would no longer exist.

PR #3157, opened 2024-09-05
Fix for URI form field regex validation

jeff-phillips-18:
Closes [RHOAIENG-12292](https://issues.redhat.com/browse/RHOAIENG-12292)

## Description
Fixes the regular expression test for the connection type URI field input

## How Has This Been Tested?
Added additional RTL tests to validate/invalidate different URI entries

## Request review criteria:


christianvogt:
`oci://rhuss/kserving-example-sklearn:1.0"` needs to validate but it fails

PR #3158, opened 2024-09-05
Show connection type creation error message on failure

jeff-phillips-18:
Closes [RHOAIENG-12439](https://issues.redhat.com/browse/RHOAIENG-12439)

## Description
Fix handling of the error response when creating a connection type.

## How Has This Been Tested?
Tested manually

## Test Impact
Added e2e test to verify response error text is shown on the page after an error.

## Request review criteria:

### DIFF
--- a/frontend/src/__tests__/cypress/cypress/pages/connectionTypes.ts
+++ b/frontend/src/__tests__/cypress/cypress/pages/connectionTypes.ts
@@ -84,6 +84,10 @@ class CreateConnectionTypePage {
     return new CreateConnectionTypeTableRow(() => this.findAllFieldsTableRows().eq(index));
   }
 
+  findFooterError() {
+    return cy.findByTestId('connection-type-footer-error');
+  }
+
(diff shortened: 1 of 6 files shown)


emilys314:
I think it wouldn't hurt to also update `frontend/src/pages/connectionTypes/manage/EditConnectionTypePage.tsx` 

But I tested the create connection type and it looked good 
image

jeff-phillips-18:
> I think it wouldn't hurt to also update `frontend/src/pages/connectionTypes/manage/EditConnectionTypePage.tsx`

Good point. I tried this and you can edit the name of the resource to be the same name as another connection type. I think we probably want to prevent this. Leading back to the other discussion of checking for duplicate names before attempting to save. @christianvogt Thoughts?

christianvogt:
@jeff-phillips-18 let's start a conversation about showing the k8s resource name in the UI. No need to handle duplicates as part of this work.

PR #3160, opened 2024-09-05
MR selector doesn't update without page reload

YuliaKrimerman:
https://issues.redhat.com/browse/RHOAIENG-10026
## Description

Please verify that the MR selector shows a new registered model, without the need of refreshing a page

## How Has This Been Tested?

on the UI

## Test Impact

I had to rewrite a couple of tests to match the new logic

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/concepts/modelRegistry/apiHooks/__tests__/useModelRegistryServices.spec.ts
+++ b/frontend/src/concepts/modelRegistry/apiHooks/__tests__/useModelRegistryServices.spec.ts
@@ -5,6 +5,7 @@ import { ServiceKind } from '~/k8sTypes';
 import { useModelRegistryServices } from '~/concepts/modelRegistry/apiHooks/useModelRegistryServices';
 import { standardUseFetchState, testHook } from '~/__tests__/unit/testUtils/hooks';
 import { mockModelRegistryService } from '~/__mocks__/mockModelRegistryService';
+import { ModelRegistryServicesResult } from '../useModelRegistryServices'; // Make sure this import exists
 
 jest.mock('@openshift/dynamic-plugin-sdk-utils', () => ({
   k8sListResource: jest.fn(),
@@ -22,6 +23,10 @@ jest.mock('~/concepts/modelRegistry/apiHooks/useModelRegistryServices', () => ({
   fetchServices: jest.fn(),
 }));
 
+jest.mock('~/api/useRulesReview', () => ({
+  useRulesReview: jest.fn(),
+}));
+
 const mockGetResource = jest.mocked(k8sGetResource<ServiceKind>);
 const mockListServices = jest.mocked(listServices);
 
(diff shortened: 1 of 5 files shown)


frontend/src/pages/modelRegistrySettings/ModelRegistrySettings.tsx
```diff
@@ -57,14 +65,17 @@ const ModelRegistrySettings: React.FC = () => {
       >
         <ModelRegistriesTable
           modelRegistries={modelRegistries}
-          refresh={refresh}
-          onCreateModelRegistryClick={() => setCreateModalOpen(true)}
+          refresh={refreshAll}
+          onCreateModelRegistryClick={() => {
+            setCreateModalOpen(true);
+            return Promise.resolve();
```
mturley, line 71: Why do you need this `return Promise.resolve();`?

mturley (commented):
LGTM, but one line I think we can remove @YuliaKrimerman

frontend/src/concepts/modelRegistry/apiHooks/__tests__/useModelRegistryServices.spec.ts
```diff
@@ -22,102 +25,164 @@ jest.mock('~/concepts/modelRegistry/apiHooks/useModelRegistryServices', () => ({
...
+    mockGetResource.mockResolvedValueOnce(
+      mockModelRegistryService({ name: 'service-1', namespace: 'test-namespace' }),
+    );
+    mockGetResource.mockResolvedValueOnce(
+      mockModelRegistryService({ name: 'service-2', namespace: 'test-namespace' }),
+    );
+
+    const { result } = testHook(() => useModelRegistryServices('test-namespace'))();
+
+    await act(async () => {
```
gitdallas, line 83: i wonder if the uses of `act` like this could be replaced with something like:

    await waitFor(() => {
        const [, isLoaded] = objectToStandardUseFetchState(result.current);
        expect(isLoaded).toBe(true);
    });
    const [services, isLoaded] = objectToStandardUseFetchState(result.current);

frontend/src/pages/modelRegistrySettings/ModelRegistrySettings.tsx
YuliaKrimerman: Removed

frontend/src/concepts/modelRegistry/apiHooks/__tests__/useModelRegistryServices.spec.ts
YuliaKrimerman: Those tests using waitFor, were there before and I really had some issues with them, so I just took this approach

YuliaKrimerman: But I can try your suggestion and see if it works

gitdallas: there are several such uses of `act`, it just popped out to me because of the deprecated strikethrough in my IDE 😆 

but more of a nit i suppose so if it's a lot of effort to address i'm okay approving.

mturley (approved):
LGTM assuming the new tests pass, works for me locally against the modelregistry-ui cluster

gitdallas (approved):

PR #3161, opened 2024-09-06
remove useEffect from state management in useDraggableTableControlled

christianvogt:
https://issues.redhat.com/browse/RHOAIENG-12561

## Description

`useDraggableTableControlled` was using `useEffect` to keep track of draggable items internally. While this was functional it would also lead to a double render whenever the incoming items array changed. The negative side effect of this is that the first render would supply data to the consumer that was stale until the next render. As a result, components which rely on object identity of the table items would be rendering against stale data until the new state was applied in the `useEffect`.

As noted in the linked issue, the connection type field would "blip" render an error icon indicating that the field's env var was a duplicate. The cause of this is because the check for duplicates would filter the given row data from the full set of table row data by object identity. Since the component was using stale data on the first render pass, the object identity check would fail until the 2nd pass.

The fix here is to eliminate the use of `useEffect` for setting state. The internal state is only required when dragging because it's a temporary state held by the hook until the drag completes.

## How Has This Been Tested?

- Create a new connection type
- Create a field
- Toggle the required switch and observe no error icon is rendered

To ensure the error icon still renders, create two fields with the same env var value.

- Create multiple fields and test dragging them to reorder.
- Save the connection type and edit it again
- Ensure the order is maintained as previously dragged.

## Test Impact

Tests already exist for testing drag.

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/utilities/useDraggableTableControlled.ts
+++ b/frontend/src/utilities/useDraggableTableControlled.ts
@@ -37,11 +37,14 @@ const useDraggableTableControlled = <T>(
   const [isDragging, setIsDragging] = React.useState(false);
   const bodyRef = React.useRef<HTMLTableSectionElement | null>(null);
 
-  const [tempItemOrder, setTempItemOrder] = React.useState<Row<T>[]>([]);
+  const [draggingItemOrder, setDraggingItemOrder] = React.useState<Row<T>[]>([]);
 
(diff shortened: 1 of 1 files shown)


emilys314:
The code looks good. I also tested the image and the table and the dropdown items both worked correctly

PR #3162, opened 2024-09-06
fix alignment of text for connection type form elements

christianvogt:
https://issues.redhat.com/browse/RHOAIENG-12573

## Description

Removed the enable checkbox from the type details section which allows it more room to display without text wrapping.
image

Removed the col span from the section title cell in order to constrain the text to its column.
image

Added missing icon to search inputs:
image

cc @simrandhaliw 

## How Has This Been Tested?

Manually.
Visual only changes.

## Test Impact

N/A

## Request review criteria:

If you have UI changes: 


emilys314:
The checkbox does indeed go wider. Also the col span for name / desc is only on it's own column.

However, in your screenshot i see a 2 line truncated description, i don't seem to see that? 
image

christianvogt:
@emilys314 I've updated `TruncatedText` to include `overflowWrap: 'anywhere'`. This is the style you applied to the preview to address the same issue there.

PR #3164, opened 2024-09-06
Fixes to connection type field modal

emilys314:
## Description

- Auto selects the name field
- update labels that no longer have "field"
- "Text - Short" is auto selected
- field type dropdown no longer has focused styling
- button is now "Save" and not "Edit"

Initial load render screenshot:
image
image

## How Has This Been Tested?

Locally, mainly visual changes

## Test Impact

The field labels have been changed so the test-id's could also be updated, but I haven't (they still have "field")

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/pages/connectionTypes/manage/ConnectionTypeDataFieldModal.tsx
+++ b/frontend/src/pages/connectionTypes/manage/ConnectionTypeDataFieldModal.tsx
@@ -43,7 +43,6 @@ const isConnectionTypeFieldType = (
 
 type Props = {
   field?: ConnectionTypeDataField;
-  isOpen?: boolean;
   onClose: () => void;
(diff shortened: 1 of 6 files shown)


frontend/src/pages/connectionTypes/manage/ConnectionTypeDataFieldModal.tsx
```diff
@@ -107,23 +107,24 @@ export const ConnectionTypeDataFieldModal: React.FC<Props> = ({
 
   return (
     <Modal
-      isOpen
+      isOpen={isOpen}
```
christianvogt, line 110: Oh we actually had a recent conversation about this. All modals must supply `isOpen` at all times and we use conditional rendering. 
Please remove the incoming prop `isOpen` and return this to the previous change.

PR #3165, opened 2024-09-06
Move test file -- weird backend setup

andrewballantyne:
We have an issue with where our backend test files are for development -- moving to where the others are. We will follow up with an investigation of our tsconfig to figure out how to fix this more generically.


christianvogt:
/lgtm

Tested the following commands successfully:
```
npm run build
npm run test:jest
npm run start:dev
```

PR #3170, opened 2024-09-09
Don't show tooltip on truncated text that isn't truncated

jeff-phillips-18:
Closes [RHOAIENG-12566](https://issues.redhat.com/browse/RHOAIENG-12566)

## Description
Determine if the text to be displayed will be truncated and if so wrap it with the tooltip otherwise just show the text.
Also fixes an issue causing a `validateDomNesting` issue for the Connection Types list page.

## How Has This Been Tested?
Tested Home page project cards, documentation cards, and connection type descriptions (in table and in preview pane).

## Test Impact
None, visual change only.

## Request review criteria:

### DIFF
--- a/frontend/src/components/TruncatedText.tsx
+++ b/frontend/src/components/TruncatedText.tsx
@@ -6,21 +6,47 @@ type TruncatedTextProps = {
   content: string;
 } & React.HTMLProps<HTMLSpanElement>;
 
-const TruncatedText: React.FC<TruncatedTextProps> = ({ maxLines, content, ...rest }) => (
-  <Tooltip content={content}>
+const TruncatedText: React.FC<TruncatedTextProps> = ({ maxLines, content, ...props }) => {
+  const outerElementRef = React.useRef<HTMLElement>(null);
+  const textElementRef = React.useRef<HTMLElement>(null);
+  const [isTruncated, setIsTruncated] = React.useState<boolean>(false);
+
+  const updateTruncation = React.useCallback(() => {
+    if (textElementRef.current && outerElementRef.current) {
+      setIsTruncated(textElementRef.current.offsetHeight > outerElementRef.current.offsetHeight);
+    }
+  }, []);
+
+  const truncateBody = (
     <span
+      {...props}
       style={{
         display: '-webkit-box',
-        overflow: 'hidden',
         WebkitBoxOrient: 'vertical',
-        WebkitLineClamp: maxLines,
         overflowWrap: 'anywhere',
+        overflow: 'hidden',
+        WebkitLineClamp: maxLines,
+        ...(props.style || {}),
       }}
-      {...rest}
+      ref={outerElementRef}
+      onMouseEnter={(e) => {
+        props.onMouseEnter?.(e);
+        updateTruncation();
+      }}
+      onFocus={(e) => {
+        props.onFocus?.(e);
+        updateTruncation();
+      }}
     >
-      {content}
+      <span ref={textElementRef}>{content}</span>
     </span>
-  </Tooltip>
-);
+  );
+
+  return (
+    <Tooltip hidden={!isTruncated ? true : undefined} content={content}>
+      {truncateBody}
+    </Tooltip>
+  );
+};
 
 export default TruncatedText;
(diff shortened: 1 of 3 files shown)


frontend/src/components/TruncatedText.tsx
```diff
@@ -6,21 +6,55 @@ type TruncatedTextProps = {
   content: string;
 } & React.HTMLProps<HTMLSpanElement>;
 
-const TruncatedText: React.FC<TruncatedTextProps> = ({ maxLines, content, ...rest }) => (
-  <Tooltip content={content}>
+const TruncatedText: React.FC<TruncatedTextProps> = ({ maxLines, content, ...props }) => {
+  const [outerElement, setOuterElement] = React.useState<HTMLElement | null>(null);
+  const [textElement, setTextElement] = React.useState<HTMLElement | null>(null);
+  const [isHover, setIsHover] = React.useState<boolean>(false);
```
christianvogt, line 12: `isHover` state is unnecessary as an intermediate state.

frontend/src/components/TruncatedText.tsx
```diff
@@ -6,21 +6,55 @@ type TruncatedTextProps = {
...
         WebkitBoxOrient: 'vertical',
-        WebkitLineClamp: maxLines,
         overflowWrap: 'anywhere',
+        overflow: 'hidden',
+        WebkitLineClamp: maxLines,
       }}
-      {...rest}
+      ref={outerRef}
+      onMouseEnter={() => setIsHover(true)}
+      onMouseLeave={() => setIsHover(false)}
```
christianvogt, line 45: It is only necessary to determine at the time of mouse over whether the tooltip needs to be shown or not. No need to do a reset on mouse leave.
Furthermore, if we remove the `isHover` state, update the `isTruncated` state immediately.
```
  const updateTruncation = React.useCallback(() => {
    if (textRef.current && outerRef.current) {
      setIsTruncated(textRef.current.offsetHeight > outerRef.current.offsetHeight);
    }
  }, []);
```
```suggestion
      onMouseEnter={updateTruncation}
```

frontend/src/components/TruncatedText.tsx
```diff
@@ -6,21 +6,55 @@ type TruncatedTextProps = {
...
+      <span ref={textRef} {...props}>
+        {content}
+      </span>
     </span>
-  </Tooltip>
-);
+  );
+
+  return (
+    <Tooltip hidden={!isTruncated ? true : undefined} content={content}>
```
christianvogt, line 54:
```suggestion
    <Tooltip hidden={!isTruncated} content={content}>
```

christianvogt (commented):
The current approach can toggle between hidden and shown tooltip too quickly as the user mouses over, leaves, repeats very quickly.

frontend/src/components/TruncatedText.tsx
```diff
@@ -6,21 +6,55 @@ type TruncatedTextProps = {
...
       style={{
         display: '-webkit-box',
-        overflow: 'hidden',
         WebkitBoxOrient: 'vertical',
-        WebkitLineClamp: maxLines,
         overflowWrap: 'anywhere',
+        overflow: 'hidden',
+        WebkitLineClamp: maxLines,
       }}
-      {...rest}
```
christianvogt, line 19: I think you should keep props spread to this outer element but before style:
```
    <span
      {...props}
      style={{
        ...props.style,
        display: '-webkit-box',
        WebkitBoxOrient: 'vertical',
        overflowWrap: 'anywhere',
        overflow: 'hidden',
        WebkitLineClamp: maxLines,
      }}
```

frontend/src/components/TruncatedText.tsx
```diff
@@ -6,21 +6,47 @@ type TruncatedTextProps = {
...
+        updateTruncation();
+      }}
+      onFocus={(e) => {
+        props.onFocus?.(e);
+        updateTruncation();
       }}
-      {...rest}
     >
-      {content}
+      <span ref={textElementRef} {...props}>
```
christianvogt, line 39: I think `props` should be applied to the outer span. For instance if someone wanted to apply margins or borders, it should be applied to the outer because the inner one is being truncated and not fully visible.

Spread the `props` before `style` and also spread `props.style` into the style before any of the fixed styles.

jeff-phillips-18: I spread the style props after the fixed styles in case, for some reason, someone wants to override one of them. Maybe there would be a reason.

PR #3186, opened 2024-09-10
fix(5463): fix axe issue in model version detail

gitdallas:
This section will be looked at again by UX for style post MVP according to this comment by Haley: https://github.com/opendatahub-io/odh-dashboard/pull/3139#issuecomment-2327882720

For now, fixing an issue that was introduced in https://github.com/opendatahub-io/odh-dashboard/pull/3139 which results in an a11y warning. The a11y issue was due to a non datalist element as a child of a datalist. To fix this, I split up the datalist into two and put the heading in between. Had to add some space that was lost after pulling the heading out of the datalist, but this should be temporary with some refresh post-MVP.

## Test Impact
no difference, just a small DOM change.

## Request review criteria:
make sure everything looks good and AXE doesn't complain about header being in DL


DaoDaoNoCode (approved):
/lgtm

PR #3194, opened 2024-09-11
fix(12787): fix 20 qty limit on registered models

gitdallas:
fixes: https://issues.redhat.com/browse/RHOAIENG-12787

showing with 26 total.

live
image
archived
image

## Description
changed model registry list limit (via API call) to 99999 as the default was 20 and when we reached 20 we would not see any new model registries. there is a spike to followup by implementing server side filter/sort/pagination https://issues.redhat.com/browse/RHOAIENG-12800

## How Has This Been Tested?
tested locally by adding more than 20 (combo of live and archived) model registries and making sure they are still visible and behavior properly in the table.

## Test Impact
updated a test to make sure that the pageSize is included in the call. can't really test beyond that as we mock the data that comes as a response.

## Request review criteria:
see more than 20 registered models in the table

If you have UI changes: 

### DIFF
--- a/frontend/src/api/modelRegistry/__tests__/custom.spec.ts
+++ b/frontend/src/api/modelRegistry/__tests__/custom.spec.ts
@@ -279,7 +279,7 @@ describe('getListRegisteredModels', () => {
     expect(proxyGETMock).toHaveBeenCalledTimes(1);
     expect(proxyGETMock).toHaveBeenCalledWith(
       'hostPath',
-      `/api/model_registry/${MODEL_REGISTRY_API_VERSION}/registered_models`,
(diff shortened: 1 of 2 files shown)


YuliaKrimerman:
I thought that we should add this line in all of our api endpoints?

gitdallas:
added for artifacts and versions as well

YuliaKrimerman:
I know I don't qualify yet, but lgtm

mturley (approved):
LGTM, perfect, thanks @gitdallas

mturley:
@YuliaKrimerman you qualify 😄

PR #3200, opened 2024-09-12
fix(vllm metrics): error stack trace

gitdallas:
closes: https://issues.redhat.com/browse/RHOAIENG-11522

this is a situation that would result in the error stack and ui crash without this change:
image

## Description
prevent ui from crashing. let the query be undefined if it doesn't exist, which will result in an empty data and no errors. vince said he did not want an error message at all as it might convey to the user that it might resolve with a refresh or something.

## How Has This Been Tested?
tested code on a previous deploy that would crash the ui on metrics page, it no longer crashes. used MR cluster to test. existing tests still pass.

## Test Impact
added a new test using mock data that only contains 1 query and made sure that the 4 charts show up (instead of an error stack page). i also updated the test mock for `prometheus/serving` to return empty results if the request body includes `query=undefined\b` as it would in the real endpoint. Here's a screenshot from a test with a missing query resulting in no data for one of the serving endpoints (it still shows the data):
image

## Request review criteria:
test a vllm deploy, view the metrics. also view metrics of other types.

If you have UI changes: 

### DIFF
--- a/frontend/src/__mocks__/mockKserveMetricsConfigMap.ts
+++ b/frontend/src/__mocks__/mockKserveMetricsConfigMap.ts
@@ -62,6 +62,51 @@ export const MOCK_KSERVE_METRICS_CONFIG_1 = `
   ]
 }`;
 
+export const MOCK_KSERVE_METRICS_CONFIG_MISSING_QUERY = `{
+    "config": [
+			{
+				"title": "Requests per 5 minutes",
+				"type": "REQUEST_COUNT",
+				"queries": [
+					{
+						"title": "Number of successful incoming requests",
+						"query": "round(sum(increase(vllm:request_success_total{namespace='test-project',model_name='test vllm'}[5m])))"
+					}
+				]
+			},
+			{
+				"title": "Average response time (ms)",
+				"type": "MEAN_LATENCY",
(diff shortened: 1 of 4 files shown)


vconzola:
LGTM.

frontend/src/__mocks__/mockKserveMetricsConfigMap.ts
```diff
@@ -62,6 +62,51 @@ export const MOCK_KSERVE_METRICS_CONFIG_1 = `
...
+export const MOCK_KSERVE_METRICS_CONFIG_MISSING_QUERY = `{
+    "config": [
+			{
+				"title": "Requests per 5 minutes",
+				"type": "REQUEST_COUNT",
+				"queries": [
+					{
+						"title": "Number of successful incoming requests",
+						"query": "round(sum(increase(vllm:request_success_total{namespace='juntao-test-2',model_name='test vllm'}[5m])))"
+					}
```
ppadti, line 74: can we have a generic namespace here rather than `juntao-test-2`?

ppadti:
tested locally, works fine. 
/lgtm

mturley (approved):

PR #3203, opened 2024-09-12
Add support for external routes to proxyService, use for model registry

mturley:
Closes https://issues.redhat.com/browse/RHOAIENG-10133 and https://issues.redhat.com/browse/RHOAIENG-9037

## Description

Updates `proxyService` to support configuring services with an optional `addressAnnotation`. If the `addressAnnotation` is omitted, the existing behavior will apply, so this PR should not affect the behavior of proxied services other than model registry. If the `addressAnnotation` is `EXTERNAL_REST` or `EXTERNAL_GRPC`, the proxy will fetch the k8s Service named in the request URL and look for a corresponding annotation. If present, the proxy will use that annotation's value as the upstream URL. If it's missing, the proxy will fall back to the existing behavior (constructing an internal `.svc.cluster.local` URL), which will provide compatibility with non-Istio model registry services that don't require the external URL.

## How Has This Been Tested?

On ods-qe-psi-08, we deployed a model registry [from the istio TLS samples](https://github.com/opendatahub-io/model-registry-operator/tree/main/config/samples/istio/mysql-tls) and observed that the Service has the `routing.opendatahub.io/external-address-rest` annotation on it. We then deployed the image for this PR using a devFlag and saw that the model registry page works when proxying requests through this new route.

We then also created a new registry using the settings page and reused the database details from the sample registry, and saw that the new istio spec provided on creation works correctly. That registry also became available and worked with the new proxy logic.

## Test Impact

No tests for backend code, unfortunately.

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/backend/src/types.ts
+++ b/backend/src/types.ts
@@ -1215,3 +1215,8 @@ export type ResponseStatus = {
   success: boolean;
   error: string;
 };
+
+export enum ServiceAddressAnnotation {
+  EXTERNAL_REST = 'routing.opendatahub.io/external-address-rest',
+  EXTERNAL_GRPC = 'routing.opendatahub.io/external-address-grpc',
+}
--- a/backend/src/utils/proxy.ts
+++ b/backend/src/utils/proxy.ts
@@ -143,6 +143,9 @@ export const proxyService =
           done();
         };
 
+        // If `model` is passed, we first check if the user is able to get a resource with that model and the given namespace/name.
+        // We can use this to only proxy for users that can access some resource that manages the service.
+        // If `statusCheck` is also passed, we can also make sure that resource passes some check before we proxy to the service.
         const doServiceRequestWithGatingResource = async () => {
           try {
             // Retreive the gating resource by name and namespace
--- a/backend/src/utils/proxy.ts
+++ b/backend/src/utils/proxy.ts
@@ -1,11 +1,12 @@
 import { FastifyRequest } from 'fastify';
 import httpProxy from '@fastify/http-proxy';
-import { K8sResourceCommon, KubeFastifyInstance } from '../types';
+import { K8sResourceCommon, KubeFastifyInstance, ServiceAddressAnnotation } from '../types';
 import { isK8sStatus, passThroughResource } from '../routes/api/k8s/pass-through';
 import { DEV_MODE } from './constants';
 import { createCustomError } from './requestUtils';
 import { getAccessToken, getDirectCallOptions } from './directCallUtils';
 import { EitherNotBoth } from '../typeHelpers';
+import { V1Service } from '@kubernetes/client-node';
 
 export const getParam = <F extends FastifyRequest<any, any>>(req: F, name: string): string =>
   (req.params as { [key: string]: string })[name];
@@ -26,12 +27,14 @@ const notFoundError = (kind: string, name: string, e?: any, overrideMessage?: st
     404,
   );
 };
+
 export const proxyService =
   <K extends K8sResourceCommon = never>(
     model: { apiGroup: string; apiVersion: string; plural: string; kind: string } | null,
     service: EitherNotBoth<
       {
-        port: number | string;
+        addressAnnotation?: ServiceAddressAnnotation;
+        internalPort: number | string;
         prefix?: string;
         suffix?: string;
         namespace?: string;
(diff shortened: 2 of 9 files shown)


backend/src/types.ts
```diff
@@ -1205,3 +1205,9 @@ export type ResponseStatus = {
   success: boolean;
   error: string;
 };
+
+export enum ServiceAddressType {
+  INTERNAL = 'internal',
+  EXTERNAL_REST = 'rest',
+  EXTERNAL_GRPC = 'grpc',
+}
```
christianvogt, line 1213: Couldn't this and `SERVICE_ADDRESS_ANNOTATIONS` be combined so that the values of each are annotations ?

backend/src/utils/proxy.ts
```diff
@@ -70,53 +71,71 @@ export const proxyService =
...
+            // kubectl port-forward -n <namespace> svc/<service-name> <local.port>:<service.port>
+            return `${scheme}://${local.host}:${local.port}`;
+          }
+          const annotation = SERVICE_ADDRESS_ANNOTATIONS[service.addressType];
+          if (annotation) {
+            try {
+              const k8sService = (
+                await fastify.kube.coreV1Api.readNamespacedService(serviceName, namespace)
+              ).body;
+              const address = k8sService.metadata.annotations[annotation];
```
christianvogt, line 88: annotations are optional.

How come typescript isn't complaining about this in our back end :(

backend/src/utils/proxy.ts
```diff
@@ -70,53 +71,71 @@ export const proxyService =
...
+                return address;
+              }
+              fastify.log.error(
+                `Proxy could not find address annotation on k8s service ${serviceName} in namespace ${namespace}, falling back to internal address. Annotation expected: ${annotation}`,
+              );
+            } catch (e) {
+              fastify.log.error(
+                e,
+                `Proxy failed to read k8s service ${serviceName} in namespace ${namespace}, falling back to internal address`,
+              );
```
christianvogt, line 99: Shouldn't this just fail instead of fall back; what's the use case?

backend/src/utils/proxy.ts
mturley: Oof, good question...

backend/src/utils/proxy.ts
mturley: Hm, fair point. I don't think there is one, I'll change this to fail.

backend/src/utils/proxy.ts
mturley: I think our backend tsconfig doesn't have `strict` or `strictNullChecks`. Maybe we want a separate tech debt issue for that.

backend/src/utils/proxy.ts
mturley: Done

backend/src/types.ts
mturley: Done

mturley:
@christianvogt addressed your feedback

backend/src/types.ts
```diff
@@ -1215,3 +1215,8 @@ export type ResponseStatus = {
   success: boolean;
   error: string;
 };
+
+export enum ServiceAddressAnnotation {
+  EXTERNAL_REST = 'routing.opendatahub.io/external-address-rest',
+  EXTERNAL_GRPC = 'routing.opendatahub.io/external-address-grpc',
```
lucferbux, line 1221: Are we expecting connecting the proxy server through grpc? I know between servers can be done, but since it's a proxified call browsers don't support grpc natively

backend/src/utils/proxy.ts
```diff
@@ -74,56 +77,93 @@ export const proxyService =
...
-
-                  doServiceRequest(resource);
-                } else {
-                  done(notFoundError(model.kind, name, undefined, 'service unavailable'));
-                }
-              });
-            })
-            .catch((e) => {
-              done(notFoundError(model.kind, name, e));
+        const doServiceRequestWithGatingResource = async () => {
```
lucferbux, line 146: Can you add some comments to understand the difference between with/without gating resources? it took me a little bit to understand and every other step is documented.

lucferbux (commented):
/lgtm

Not approving until there's an extra review, I've tested the cluster and it seems to be connecting to the right route.

Gkrumbach07:
the pipelines construct url stuff looks to be in working order

backend/src/types.ts
mturley: We aren't currently using that annotation, I just added it in the constant here because it was added at the same time on the service in case we may need it later. Maybe I should just remove it?

andrewballantyne: I think you can keep it because it's in the generic types folder -- but I agree with Lucas, there is not a likelihood of this working through the http-proxy -- I think Pipelines MLMD uses envoy proxy which is a http endpoint that converts it... @Gkrumbach07 would have to confirm that though.

Gkrumbach07:
Tested on cluster and looks to be working and connecting to the annotation route.
/lgtm

mturley:
Self-approving with the LGTM from Gage and no more unaddressed feedback. Thanks guys

PR #3206, opened 2024-09-13
Upadte archived models and archived versions components

ppadti:
Closes: [RHOAIENG-11509](https://issues.redhat.com/browse/RHOAIENG-11509)

## Description
This PR aims to update
- Archived models to read-only by removing the toolbar actions, toolbar kebab and kebab menu from the table.

Screenshot from 2024-09-13 13-47-03

-  Archived model details page to read-only by removing all the actions.

Screenshot from 2024-09-05 00-07-05

- Archived version details page to read-only by removing all the actions.

Screenshot from 2024-09-05 00-40-37

- implementation for versions of archived model.

Screenshot from 2024-09-13 13-47-25

## How Has This Been Tested?
1. Archived models details page:
    - Archive a model
    - Go to the `view archived models`
    - see the archived model versions tab and details tab
2. Archived version details page:
    - Archive a version
    - Go to `view archived version` in model details
    - Go to the details page of the archived version
3. Version details page of archived model:
    - Archive a model with versions
    - Go to a version inside it

## Test Impact
Added cypress test.

## Request review criteria:

If you have UI changes: 


frontend/src/pages/modelRegistry/screens/ModelVersions/ModelVersionListView.tsx
```diff
@@ -55,8 +58,18 @@ const ModelVersionListView: React.FC<ModelVersionListViewProps> = ({
...
   if (unfilteredModelVersions.length === 0) {
+    if (isArchiveModel) {
+      return (
+        <EmptyModelRegistryState
+          testid="empty-archive-model-versions"
+          title="No versions"
+          description={`${rm?.name} has no registered versions.`}
+        />
+      );
+    }
```
ppadti, line 72: @mturley @yih-wang I am not sure, should we have Empty state for versions of Archived models? for now, I have just re-used the one - empty state for versions by removing the primary action. Let me know your thoughts on this?

ppadti: Had discussion with @yih-wang and I will update the Empty state for version with archived model.

frontend/src/pages/modelRegistry/screens/ModelVersionDetails/ModelVersionDetailsView.tsx
```diff
@@ -51,22 +62,39 @@ const ModelVersionDetailsView: React.FC<ModelVersionDetailsViewProps> = ({
...
+            >
+              <LabelGroup data-testid="label-group">
+                {getLabels(mv.customProperties).map((label) => (
+                  <Label textMaxWidth="40ch" key={label} color="blue" data-testid="label">
+                    {label}
+                  </Label>
+                ))}
+              </LabelGroup>
+            </DashboardDescriptionListGroup>
+          ) : (
```
mturley, line 79: You don't need this ternary / new block. If you add an `isEditable` prop to `EditableLabelsDescriptionListGroup`, you can pass its value down into `DashboardDescriptionListGroup` which already has an `isEditable` prop that is just only set to true currently. Then pass `isEditable={!isArchiveVersion}` to `EditableLabelsDescriptionListGroup` here.

frontend/src/pages/modelRegistry/screens/ModelVersions/ModelDetailsView.tsx
```diff
@@ -40,22 +54,39 @@ const ModelDetailsView: React.FC<ModelDetailsViewProps> = ({ registeredModel: rm
...
+            >
+              <LabelGroup data-testid="label-group">
+                {getLabels(rm.customProperties).map((label) => (
+                  <Label textMaxWidth="40ch" key={label} color="blue" data-testid="label">
+                    {label}
+                  </Label>
+                ))}
+              </LabelGroup>
+            </DashboardDescriptionListGroup>
+          ) : (
```
mturley, line 71: Same as above here.

frontend/src/pages/modelRegistry/screens/ModelVersions/ModelVersionListView.tsx
```diff
@@ -81,95 +100,122 @@ const ModelVersionListView: React.FC<ModelVersionListViewProps> = ({
...
-                  isExpanded={isArchivedModelVersionKebabOpen}
-                  aria-label="View archived versions"
+    <>
+      {isArchiveModel && (
+        <Alert
+          variant="warning"
+          isInline
+          title={`All the versions have been archived along with the model by ${
+            rm?.owner ?? '--'
+          } on  ${
```
mturley, line 110: @yih-wang @vconzola I don't think we can actually display who the models/versions were archived by. This code will display that they have been archived by whoever is the model owner, which may not be true if multiple users have access to the registry and someone else archived it. The API currently has no way to record who made a change unless we change the owner when someone archives something, which doesn't seem right.

I think maybe we need to instead not name a user here and say "All the versions have been archived along with the model on [date]". What do you think?

In the future we could ask the MR team to add a `lastModifiedBy` field to the API for us to track who made changes.

mturley (changes requested):
@ppadti there's one fix missing here - it is possible to see the wrong view for a model or version by going to the URL as if it was archived when it's not or vice versa. We should have the pages check whether the model and/or version is actually archived or not archived, and redirect to the correct place if you're at the wrong URL. This is a corner case, but it can come up if you have an old tab open on a model or version before its archived state changes or if you bookmark one.

Example:

If I have a model that is not archived and go its details page, that URL is:
`/modelRegistry/modelregistry-sample/registeredModels/13/versions`
and if I edit that to:
`/modelRegistry/modelregistry-sample/registeredModels/archive/13/versions`
I incorrectly see the archived view even though the model is not archived. The page should redirect me back to that first URL.
If I then archive that model and navigate to it, then change its URL to remove `/archive/`, we get the opposite problem. Either way it should redirect me to the URL corresponding to the real model state.

The same is true of the archived / non-archived version pages. They should detect which page you should be on based on whether model or version are archived and redirect you there if you aren't already there.

A few other comments below as well, but the rest LGTM! Thanks for the hard work on this.

frontend/src/pages/modelRegistry/screens/ModelVersions/ModelVersionListView.tsx
yih-wang: That makes sense to me since we don't record any model activities for now. 

> I think maybe we need to instead not name a user here and say "All the versions have been archived along with the model on [date]". What do you think?
Sounds good!

ppadti:
> @ppadti there's one fix missing here - it is possible to see the wrong view for a model or version by going to the URL as if it was archived when it's not or vice versa. We should have the pages check whether the model and/or version is actually archived or not archived, and redirect to the correct place if you're at the wrong URL. This is a corner case, but it can come up if you have an old tab open on a model or version before its archived state changes or if you bookmark one.
> 
> Example:
> 
> If I have a model that is not archived and go its details page, that URL is: `/modelRegistry/modelregistry-sample/registeredModels/13/versions` and if I edit that to: `/modelRegistry/modelregistry-sample/registeredModels/archive/13/versions` I incorrectly see the archived view even though the model is not archived. The page should redirect me back to that first URL. If I then archive that model and navigate to it, then change its URL to remove `/archive/`, we get the opposite problem. Either way it should redirect me to the URL corresponding to the real model state.
> 
> The same is true of the archived / non-archived version pages. They should detect which page you should be on based on whether model or version are archived and redirect you there if you aren't already there.

ahh... Some how I missed this,  I thought only the archived models logic is wrong for versions under that. I will update this. Thanks!!

frontend/src/pages/modelRegistry/screens/ModelVersionDetails/ModelVersionDetails.tsx
```diff
@@ -41,6 +47,26 @@ const ModelVersionsDetails: React.FC<ModelVersionsDetailProps> = ({ tab, ...page
...
+    } else if (mv?.state === ModelState.ARCHIVED) {
+      navigate(
+        modelVersionArchiveDetailsUrl(
+          mv.id,
+          mv.registeredModelId,
+          preferredModelRegistry?.metadata.name,
+        ),
+      );
+    }
+  });
```
mturley, line 68: Should all these new useEffects have dependency arrays? I'm not sure whether it matters since the `if` conditions are the same perf impact as comparing dependencies across renders, but it may still be better to not run the effects every render in case we modify them later.

mturley (commented):
LGTM except the one question here and the test failures @ppadti

mturley (approved):
LGTM! This is great, thank you @ppadti !

PR #3207, opened 2024-09-13
Projects list view: expand workbenches column to a table

jeff-phillips-18:
Closes [RHOAIENG-10494](https://issues.redhat.com/browse/RHOAIENG-10494)

## Description
Update the projects list view from having nested subcolumns to show workbenches to using an expandable column to show an expanded view with a table listing the workbenches for the project.

## How Has This Been Tested?
On a system with multiple projects, navigate to the Data Science Projects page.
Notice that workbenches are now indicated as an icon (wrench) and a count.
Click on the icon/count for a project with existing workbenches.
See the row expand to show a table of the workbenches belonging to the project.

Manually tested

## Test Impact
Updated the e2e tests

## Screen shot
image

## Request review criteria:

If you have UI changes: 

/cc @tobiastal


frontend/src/images/icons/NotebookIcon.ts
```diff
@@ -0,0 +1,13 @@
+import { createIcon } from '@patternfly/react-icons/dist/esm/createIcon';
+
+const NotebookIcon = createIcon({
```
jenny-s51, line 3: According to the [mocks](https://www.figma.com/design/xejiiAIdT9ezaBKCNTss6b/Projects-table-view?node-id=89-113418&node-type=frame&t=YJhlCks2RUKXaDas-0), it looks like this icon should be solid. Wondering if we can update the svg path here? WDYT @jeff-phillips-18

jenny-s51 (commented):
These changes are looking good @jeff-phillips-18 . Tested in the UI against the [mocks](https://www.figma.com/design/xejiiAIdT9ezaBKCNTss6b/Projects-table-view?node-id=89-113418&node-type=frame&t=YJhlCks2RUKXaDas-0). Left a comment below.

frontend/src/images/icons/NotebookIcon.ts
tobiastal: @jenny-s51 You're looking at the old mocks, In the updated[mocks](https://www.figma.com/design/xejiiAIdT9ezaBKCNTss6b/Projects-table-view?node-id=1521-112631&t=V9wgRDW5KiryjAqJ-1) (which are ready for dev), we're using the outlined icon, just like Jeff added.

One small difference is that the icon appears slightly smaller than it should be (compared to the mocks). 
@jeff-phillips-18 Could you take a look at this?

jeff-phillips-18: The size in the table is based on the font size (14px) while the mocks show 16px. I have updated the code to set the icons size to 16px. Screen shot above has been updated.

tobiastal:
LGTM

jenny-s51 (approved):
LGTM, retested locally and compared against updated mocks. Icon size update looks good.

andrewballantyne:
Does the stopped state have a pause icon? It's not paused... there is no data persistence in memory

jeff-phillips-18:
> Does the stopped state have a pause icon? It's not paused... there is no data persistence in memory

@tobiastal Mocks show a `paused` icon. Is this what we want?

jeff-phillips-18:
Got an updated icon for the workbench in the row:

image

frontend/src/pages/projects/notebook/NotebookActionsColumn.tsx
```diff
@@ -82,55 +74,70 @@ const NotebookStatusToggle: React.FC<NotebookStatusToggleProps> = ({
...
+          },
+          {
+            isDisabled: isStarting || isStopping,
+            title: 'Edit workbench',
+            onClick: () => {
+              navigate(
+                `/projects/${project.metadata.name}/spawner/${notebookState.notebook.metadata.name}`,
+              );
+            },
+          },
```
christianvogt, line 119: If I cancel out of editing the workbench, I don't end up back at the project list view.
If I press the browser back button, I end up on the project list view however I am on the wrong page or the expanded section is no longer open.

frontend/src/pages/projects/notebook/NotebookStateStatus.tsx
```diff
@@ -0,0 +1,127 @@
...
+        label: 'Starting',
+        color: 'blue',
+        icon: <InProgressIcon style={{ animation: rotate }} />,
+      };
+    }
+    if (isStopping) {
+      return {
+        label: 'Stopping',
+        color: 'grey',
+        icon: <SyncAltIcon style={{ animation: rotate }} />,
```
christianvogt, line 53: We use these icons in several places.

PF doesn't have a means to spin icons? If not then, make then reusable components with their custom animation so we can update those other locations at a later date.

frontend/src/api/k8s/notebooks.ts
```diff
@@ -206,7 +206,7 @@ export const getStopPatch = (): Patch => ({
   value: getStopPatchDataString(),
 });
 
-export const getNotebooks = (namespace: string): Promise<NotebookKind[]> =>
+export const getNotebooks = (namespace?: string): Promise<NotebookKind[]> =>
```
christianvogt, line 209: Why this change?

frontend/src/pages/projects/screens/projects/ProjectTableRowNotebookTable.tsx
```diff
@@ -0,0 +1,55 @@
...
+          />
+          <DeleteNotebookModal
+            notebook={notebookToDelete}
+            onClose={(deleted) => {
+              if (deleted) {
+                refresh();
+              }
+              setNotebookToDelete(undefined);
+            }}
+          />
```
christianvogt, line 48: Modals should be conditionally rendered.

frontend/src/pages/projects/screens/projects/ProjectTableRowNotebookTableRow.tsx
```diff
@@ -0,0 +1,45 @@
...
+}) => {
+  const [ActionColumn, stopNotebook] = useNotebookActionsColumn(
+    project,
+    notebookState,
+    enablePipelines,
+    onNotebookDelete,
+  );
+  return (
+    <Tr style={{ border: 'none' }} data-testid="project-notebooks-table-row">
+      <Td dataLabel="workbenchName" style={{ paddingLeft: 'var(--pf-v5-global--spacer--sm)' }}>
```
christianvogt, line 30: cell value is misaligned with header:
image

frontend/src/pages/projects/notebook/NotebookActionsColumn.tsx
jeff-phillips-18: Updated the spawner paged to navigate back on a cancel

frontend/src/api/k8s/notebooks.ts
jeff-phillips-18: oops... reverted

tobiastal:
> > Does the stopped state have a pause icon? It's not paused... there is no data persistence in memory
> 
> @tobiastal Mocks show a `paused` icon. Is this what we want?

In the mockups, I used the 'paused' icon since the behavior allows you to stop a running workbench, showing it as inactive but reversible—you can press 'Start', and it will resume. However, if that's not the case, as @andrewballantyne mentioned, we can switch to the "fa-ban" icon if that’s clearer.
image

christianvogt:
While the workbench expandable table is loading, an extra border at the bottom is rendered:
image

Starting icon isn't spinning. Should it?
image

jeff-phillips-18:
> Do we care what it looks like on stacked layout?

Updated:

image

jgiardino:
In the mockups, the nested table has the column header and table cell contents left-aligned
image

But in the example you shared, the column header contents are indented. 
image

When I check how [this same use case displays in PF|https://www.patternfly.org/components/table/react-demos/compound-expansion/], the column header and table cell contents are also left-aligned. However the table itself is not indented. I'm only pointing this out in case the misalignment is due to indenting the table? 🤷‍♀️ 
image

Is it possible to address this alignment issue between the table header and table cell contents?

christianvogt:
@jgiardino 
here's what the latest changes look like:
image

christianvogt:
lgtm
Any further changes from UX required? @tobiastal @jeff-phillips-18

PR #3209, opened 2024-09-13
Model registry deploy modal has faulty submit button disable detection

YuliaKrimerman:
(https://issues.redhat.com/browse/RHOAIENG-12116)
## Description

Updated the UI to enable the Deploy button once all the required fields are filled, without the need to update the pre-filled fields/ 

## How Has This Been Tested?

On the UI in - try to deploy a version of a model with pre-existing connection fields
## Test Impact

Added a unit test that makes sure the awsData array never has more than one element per key, as this was the issues that was causing the buggy behavior. ## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/pages/modelRegistry/screens/RegisteredModels/usePrefillDeployModalFromModelRegistry.ts
+++ b/frontend/src/pages/modelRegistry/screens/RegisteredModels/usePrefillDeployModalFromModelRegistry.ts
@@ -41,23 +41,18 @@ const usePrefillDeployModalFromModelRegistry = (
           type: InferenceServiceStorageType.EXISTING_STORAGE,
         });
       } else {
-        const prefilledKeys = [
+        const prefilledKeys: (typeof EMPTY_AWS_SECRET_DATA)[number]['key'][] = [
           AwsKeys.NAME,
           AwsKeys.AWS_S3_BUCKET,
           AwsKeys.S3_ENDPOINT,
           AwsKeys.DEFAULT_REGION,
         ];
-        const filteredEmptyAwsSecretData = EMPTY_AWS_SECRET_DATA.filter(
-          (item): item is (typeof EMPTY_AWS_SECRET_DATA)[number] =>
-            !prefilledKeys.includes(item.key),
-        );
-
         const prefilledAWSData = [
           { key: AwsKeys.NAME, value: registeredModelDeployInfo.modelArtifactStorageKey || '' },
           { key: AwsKeys.AWS_S3_BUCKET, value: storageFields.bucket },
           { key: AwsKeys.S3_ENDPOINT, value: storageFields.endpoint },
           { key: AwsKeys.DEFAULT_REGION, value: storageFields.region || '' },
-          ...filteredEmptyAwsSecretData,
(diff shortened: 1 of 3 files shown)


frontend/src/pages/modelServing/screens/projects/kServeModal/ManageKServeModal.tsx
```diff
@@ -165,7 +165,10 @@ const ManageKServeModal: React.FC<ManageKServeModalProps> = ({
   // Inference Service Validation
   const storageCanCreate = (): boolean => {
     if (createDataInferenceService.storage.type === InferenceServiceStorageType.EXISTING_STORAGE) {
-      return createDataInferenceService.storage.dataConnection !== '';
+      return (
+        createDataInferenceService.storage.dataConnection !== '' ||
+        (dataConnections.length > 0 && createDataInferenceService.storage.path !== '')
```
mturley, line 170: Is this line necessary now? I'm not sure I fully understand the case here. If we're using EXISTING_STORAGE we should always have a dataConnection name right?

frontend/src/pages/modelRegistry/screens/RegisteredModels/usePrefillDeployModalFromModelRegistry.ts
```diff
@@ -41,12 +41,23 @@ const usePrefillDeployModalFromModelRegistry = (
...
         });
       } else {
+        const prefilledKeys = [
+          AwsKeys.NAME,
+          AwsKeys.AWS_S3_BUCKET,
+          AwsKeys.S3_ENDPOINT,
+          AwsKeys.DEFAULT_REGION,
+        ];
+        const filteredEmptyAwsSecretData = EMPTY_AWS_SECRET_DATA.filter(
+          (item): item is (typeof EMPTY_AWS_SECRET_DATA)[number] =>
```
mturley, line 51: I think the `is` keyword here is slightly misused, what was the reason for it here? `is` creates a type guard: the function here `(item): item is T => { ... }` tells TypeScript "if this returns true, `item` has the type `T`" and it can be used for narrowing the type of a variable or parameter when it's too broad. I don't think it does anything for us when used as a callback like this, removing it doesn't seem to cause any problems.

If you were trying to make sure all your `prefilledKeys` are keys present in `EMPTY_AWS_SECRET_DATA`, you could instead change the `const prefilledKeys` line to have a type assertion like:

```ts
        const prefilledKeys: (typeof EMPTY_AWS_SECRET_DATA)[number]['key'][] = [
          AwsKeys.NAME,
          AwsKeys.AWS_S3_BUCKET,
          AwsKeys.S3_ENDPOINT,
          AwsKeys.DEFAULT_REGION,
        ];
```

mturley (changes requested):
A couple small things

frontend/src/pages/modelServing/screens/projects/kServeModal/ManageKServeModal.tsx
YuliaKrimerman: yeah It was part of me trying to secure all parts of button enabling, but you're right it's not necessary.  removed it.

PR #3212, opened 2024-09-13
Storage class dropdown select

Gkrumbach07:
Closes: [RHOAIENG-1109](https://issues.redhat.com/browse/RHOAIENG-1109)

## Description
This is a continuation of @dpanshug PR
- added default selection
- added default label
- corrected autofill for default
- added alerts for select

> Added new field to select storage class to create pvc.

Screenshot 2024-09-13 at 2 17 21 PM
Screenshot 2024-09-13 at 2 17 07 PM
Screenshot 2024-09-13 at 2 17 00 PM

## How Has This Been Tested?
Create a new cluster storage and you check list of storage classes in the field.
Create new workbench and select storage class in cluster storage.

## Test Impact

added test for default being set

## Request review criteria:

Self checklist (all need to be checked):
- [x] The developer has manually tested the changes and verified that the changes work
- [x] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [ ] The developer has added tests or explained why testing cannot be added (unit or cypress tests for related changes)

If you have UI changes: 

- [x] Included any necessary screenshots or gifs if it was a UI change.
- [x] Included tags to the UX team if it was a UI/UX change.

After the PR is posted & before it merges:
- [ ] The developer has tested their solution on a cluster by using the image produced by the PR to `main`
@xianli123

PR #3214, opened 2024-09-17
Display storage class selected in workbenches

pnaik1:
https://issues.redhat.com/browse/RHOAIENG-1110

## Description

1. Added a new column for the storage class in the cluster storage table when the feature flag is enabled
2. shows class display name and description and sortable
3. shows the deprecated storage class alert when atleast one of the storage class is disabled and attached to cluster storage and added a label 

Cluster storage page

Screenshot 2024-09-17 at 1 04 34 PM

when storage class is deprecated/disabled
[
Screenshot 2024-09-16 at 5 31 10 PM
](url)

when storage class is deleted

Screenshot 2024-09-16 at 5 29 28 PM

## How Has This Been Tested?
When feature flag is off

1. Check the functionality remains same 

When feature flag is on

1. Go to the cluster storage tab from the Data science project Page
2. Check for the new storage column
3. Check whether its sortable
4. Check when the storage class is disabled for the attached cluster storage the deprecated label along with the alert message is shown
5. Check when the storage class is deleted for the attached cluster storage the deleted label is shown

## Test Impact
Added cypress test to it

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/pages/projects/screens/detail/storage/StorageTable.tsx
+++ b/frontend/src/pages/projects/screens/detail/storage/StorageTable.tsx
@@ -1,10 +1,15 @@
 import * as React from 'react';
+import { Alert, AlertActionCloseButton } from '@patternfly/react-core';
 import { Table } from '~/components/table';
 import { PersistentVolumeClaimKind } from '~/k8sTypes';
 import DeletePVCModal from '~/pages/projects/pvc/DeletePVCModal';
+import { SupportedArea, useIsAreaAvailable } from '~/concepts/areas';
+import { getStorageClassConfig } from '~/pages/storageClasses/utils';
+import useStorageClasses from '~/concepts/k8s/useStorageClasses';
 import StorageTableRow from './StorageTableRow';
 import { columns } from './data';
 import ManageStorageModal from './ManageStorageModal';
+import { StorageTableData } from './types';
 
 type StorageTableProps = {
   pvcs: PersistentVolumeClaimKind[];
@@ -15,20 +20,64 @@ type StorageTableProps = {
 const StorageTable: React.FC<StorageTableProps> = ({ pvcs, refresh, onAddPVC }) => {
   const [deleteStorage, setDeleteStorage] = React.useState<PersistentVolumeClaimKind | undefined>();
   const [editPVC, setEditPVC] = React.useState<PersistentVolumeClaimKind | undefined>();
+  const isStorageClassesAvailable = useIsAreaAvailable(SupportedArea.STORAGE_CLASSES).status;
+  const [storageClasses, storageClassesLoaded] = useStorageClasses();
+  const [alertDismissed, setAlertDismissed] = React.useState<boolean>(false);
+  const storageTableData: StorageTableData[] = pvcs.map((pvc) => ({
+    pvc,
+    storageClass: storageClasses.find((sc) => sc.metadata.name === pvc.spec.storageClassName),
+  }));
+  const isDeprecatedAlert = React.useMemo(
+    () =>
+      storageClassesLoaded &&
+      storageTableData.some(
+        (data) => !data.storageClass || !getStorageClassConfig(data.storageClass)?.isEnabled,
+      ),
+    [storageClassesLoaded, storageTableData],
+  );
+  const shouldShowAlert = isDeprecatedAlert && !alertDismissed && isStorageClassesAvailable;
+
+  const getStorageColumns = () => {
+    let storageColumns = columns;
+
+    if (!isStorageClassesAvailable) {
+      storageColumns = columns.filter((column) => column.field !== 'storage');
+    }
+    return storageColumns;
+  };
 
   return (
     <>
+      {shouldShowAlert && (
+        <Alert
+          data-testid="storage-class-deprecated-alert"
+          variant="warning"
+          isInline
+          title="Deprecated storage class"
+          actionClose={
+            <AlertActionCloseButton
+              data-testid="storage-class-deprecated-alert-close-button"
+              onClose={() => setAlertDismissed(true)}
+            />
+          }
+        >
+          A storage class has been deprecated by your administrator, but the cluster storage using
+          it is still active. If you want to migrate your data to cluster storage instance using a
+          different storage class, contact your administrator.
+        </Alert>
+      )}
       <Table
-        data={pvcs}
-        columns={columns}
+        data={storageTableData}
+        columns={getStorageColumns()}
         disableRowRenderSupport
         data-testid="storage-table"
         variant="compact"
-        rowRenderer={(pvc, i) => (
+        rowRenderer={(data, i) => (
           <StorageTableRow
-            key={pvc.metadata.uid}
+            key={data.pvc.metadata.uid}
             rowIndex={i}
-            obj={pvc}
+            obj={data}
+            storageClassesLoaded={storageClassesLoaded}
             onEditPVC={setEditPVC}
             onDeletePVC={setDeleteStorage}
             onAddPVC={onAddPVC}
--- a/frontend/src/pages/projects/screens/detail/storage/StorageTableRow.tsx
+++ b/frontend/src/pages/projects/screens/detail/storage/StorageTableRow.tsx
@@ -7,40 +7,56 @@ import {
   Td,
   Tr,
 } from '@patternfly/react-table';
-import { Flex, FlexItem, Text } from '@patternfly/react-core';
-import { HddIcon } from '@patternfly/react-icons';
+import {
+  Flex,
+  FlexItem,
+  Label,
+  Skeleton,
+  Text,
+  TextVariants,
+  Tooltip,
+} from '@patternfly/react-core';
+import { ExclamationTriangleIcon, HddIcon } from '@patternfly/react-icons';
 import { PersistentVolumeClaimKind } from '~/k8sTypes';
 import StorageSizeBar from '~/pages/projects/components/StorageSizeBars';
 import ConnectedNotebookNames from '~/pages/projects/notebook/ConnectedNotebookNames';
 import { ConnectedNotebookContext } from '~/pages/projects/notebook/useRelatedNotebooks';
 import { TableRowTitleDescription } from '~/components/table';
 import { getDescriptionFromK8sResource, getDisplayNameFromK8sResource } from '~/concepts/k8s/utils';
+import { SupportedArea, useIsAreaAvailable } from '~/concepts/areas';
+import { getStorageClassConfig } from '~/pages/storageClasses/utils';
 import useIsRootVolume from './useIsRootVolume';
 import StorageWarningStatus from './StorageWarningStatus';
+import { StorageTableData } from './types';
 
 type StorageTableRowProps = {
-  obj: PersistentVolumeClaimKind;
   rowIndex: number;
+  obj: StorageTableData;
+  storageClassesLoaded: boolean;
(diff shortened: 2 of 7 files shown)


pnaik1:
/hold
until  this [PR-3212](https://github.com/opendatahub-io/odh-dashboard/pull/3212) gets merged

pnaik1:
@xianli123 Can you please have a look at the screenshot attached above
When the class is deleted from OpenShift , I have added the `deleted` label, Can you verify this??
and do we still need the alert message here??

frontend/src/pages/projects/screens/detail/storage/data.ts
```diff
@@ -15,16 +15,23 @@ export const columns: SortableData<PersistentVolumeClaimKind>[] = [
     sortable: (a, b) =>
       getDisplayNameFromK8sResource(a).localeCompare(getDisplayNameFromK8sResource(b)),
   },
+  {
+    field: 'storage',
+    label: 'Storage class',
+    width: 30,
+    sortable: (a, b) =>
+      (a.spec.storageClassName ?? '').localeCompare(b.spec.storageClassName ?? ''),
```
Gkrumbach07, line 23: this will not always be accurate. sort via the config display name stored in the annotation. Fall back to the storage class k8s name

frontend/src/pages/projects/screens/detail/storage/StorageTable.tsx
```diff
@@ -15,20 +19,66 @@ type StorageTableProps = {
...
+  const [storageClasses, storageClassesLoaded] = useStorageClasses();
+  const storageClassConfigMap = React.useMemo(
+    () =>
+      storageClasses.reduce((acc: Record<string, StorageClassConfig | undefined>, sc) => {
+        acc[sc.metadata.name] = getStorageClassConfig(sc);
+
+        return acc;
+      }, {}),
+    [storageClasses],
+  );
```
Gkrumbach07, line 38: why use a map of the configs and pass it in. just fetch the storageClass metadata in the row itself

frontend/src/pages/projects/screens/detail/storage/StorageTableRow.tsx
```diff
@@ -29,13 +42,28 @@ type StorageTableRowProps = {
...
+  const storageClass = storageClassConfigMap[obj.spec.storageClassName ?? ''];
+  const isStorageClassDisabled = storageClass && !storageClass.isEnabled;
+  const isStorageClassDeleted = storageClassesLoaded && !storageClass;
+  React.useEffect(() => {
+    setAlertStatuses((prevState) => ({
+      ...prevState,
+      [obj.metadata.name]: !!isStorageClassDisabled || isStorageClassDeleted,
+    }));
+  }, [isStorageClassDeleted, isStorageClassDisabled, obj.metadata.name, setAlertStatuses]);
+
```
Gkrumbach07, line 66: We should move all this alert logic back up a level because alert statuses is not even used in this row component.

frontend/src/pages/projects/screens/detail/storage/StorageTable.tsx
```diff
@@ -15,20 +19,66 @@ type StorageTableProps = {
 const StorageTable: React.FC<StorageTableProps> = ({ pvcs, refresh, onAddPVC }) => {
   const [deleteStorage, setDeleteStorage] = React.useState<PersistentVolumeClaimKind | undefined>();
   const [editPVC, setEditPVC] = React.useState<PersistentVolumeClaimKind | undefined>();
+  const isStorageClassesAvailable = useIsAreaAvailable(SupportedArea.STORAGE_CLASSES).status;
+  const [alertStatuses, setAlertStatuses] = React.useState<Record<string, boolean>>({});
+
+  const isDeprecatedAlert = React.useMemo(
+    () => Object.values(alertStatuses).includes(true),
+    [alertStatuses],
+  );
```
Gkrumbach07, line 28: this whole alertStatuses is to just put the banner on the top. So instead of state, let us just have a useMemo that iterates over each pvc's storage class and returns true or false if to show the alert banner. Then you also dont need to have the setAlertStatuses anywhere as it reacts just to the pvc.

so something like 

```tsx
const showDeprecatedStorageClassAlert = React.useMemo(() => pvcs.some(pvc => // get storage class config and return true if deprecated ...
```

frontend/src/pages/projects/screens/detail/storage/StorageTable.tsx
```diff
@@ -15,20 +19,66 @@ type StorageTableProps = {
...
+      {isDeprecatedAlert && isStorageClassesAvailable && (
+        <Alert
+          data-testid="storage-class-deprecated-alert"
+          variant="warning"
+          isInline
+          title="Deprecated storage class"
+          actionClose={
+            <AlertActionCloseButton
+              data-testid="storage-class-deprecated-alert-close-button"
+              onClose={() => setAlertStatuses({})}
```
Gkrumbach07, line 59: you will now have to have a state for is the alert has been dismissed. dont tie it into the use memo

Gkrumbach07 (changes requested):

xianli123:
Thanks @pnaik1！ That works for me.

frontend/src/pages/projects/screens/detail/storage/StorageTable.tsx
pnaik1: The reason to map the storageClass with storageConfig , so we can directly get the storageConfig for the corresponding to `pvc.spec.storageClassName`

Do you want me to do this instead in the `StorageTableRow`??
 ```
const storageClassAvailable = storageClasses.find(
    (sc) => sc.metadata.name === obj.spec.storageClassName,
  );

  const storageClass = storageClassAvailable
    ? getStorageClassConfig(storageClassAvailable)
    : undefined;

```

We only get `storageClassName` which is a `string` from `PersistentVolumeClaimKind`
and we need to iterate over  the storageClasses to find ClassConfig

File: frontend/src/pages/projects/screens/detail/storage/data.ts (hunk 1 above)
Comment (pnaik1) [reply frontend/src/pages/projects/screens/detail/storage/data.ts:23]: I checked this ,  The `a` here is of the type `PersistentVolumeClaimKind` there is no config display name stored in the annotation 
I think you are mentioning about the type `StorageClassKind[]`???
We only have `name` which is stored in `a.spec.storageClassName`

Comment (Gkrumbach07) [reply frontend/src/pages/projects/screens/detail/storage/data.ts:23]: right but you are sorting based on the name in the config, so you may need to get it in their to be able to sort it. for example creating a type for the table like

```ts
type StorageTableData = {
   pvc: PersistentVolumeClaimKind;
   storageClass: StorageClassKind;
}

File: frontend/src/pages/projects/screens/detail/storage/StorageTable.tsx (hunk 2 above)
Comment (Gkrumbach07) [reply frontend/src/pages/projects/screens/detail/storage/StorageTable.tsx:38]: i see. how about you just pass in the storage class config into the row. so instead of passing in the whole map, you just pass in the storage class config like so

```tsx
...
storageClassConfig={storageClassConfigMap[obj.spec.storageClassName ?? '']}
...

File: frontend/src/pages/projects/screens/detail/storage/StorageTableRow.tsx [hunk 6]
```diff
@@ -79,6 +107,47 @@ const StorageTableRow: React.FC<StorageTableRowProps> = ({
           </Flex>
           <Text>{getDescriptionFromK8sResource(obj)}</Text>
         </Td>
+
+        {isStorageClassesAvailable && (
+          <Td dataLabel="Storage class">
+            <TextContent>
+              <Text>
+                {obj.spec.storageClassName}{' '}
```
Gkrumbach07, line 115: this should fall back to the metadata name if the storage class config is undefined

frontend/src/pages/projects/screens/detail/storage/StorageTableRow.tsx
```diff
@@ -79,6 +107,47 @@ const StorageTableRow: React.FC<StorageTableRowProps> = ({
           </Flex>
           <Text>{getDescriptionFromK8sResource(obj)}</Text>
         </Td>
+
+        {isStorageClassesAvailable && (
+          <Td dataLabel="Storage class">
```
Gkrumbach07, line 112: the spacing is not the same as the other column

do what the name of the pvc does

```tsx
 <Flex
          >
            <FlexItem>
              <TableRowTitleDescription title={getDisplayNameFromK8sResource(obj)} resource={obj} />
            </FlexItem>
            <FlexItem>
              <StorageWarningStatus obj={obj} onEditPVC={onEditPVC} onAddPVC={onAddPVC} />
            </FlexItem>
          </Flex>
```

frontend/src/pages/projects/screens/detail/storage/StorageTable.tsx
```diff
@@ -15,20 +20,65 @@ type StorageTableProps = {
 const StorageTable: React.FC<StorageTableProps> = ({ pvcs, refresh, onAddPVC }) => {
   const [deleteStorage, setDeleteStorage] = React.useState<PersistentVolumeClaimKind | undefined>();
   const [editPVC, setEditPVC] = React.useState<PersistentVolumeClaimKind | undefined>();
+  const isStorageClassesAvailable = useIsAreaAvailable(SupportedArea.STORAGE_CLASSES).status;
+  const [storageClasses, storageClassesLoaded] = useStorageClasses();
+  const [alertDismissed, setAlertDismissed] = React.useState<boolean>(false);
+  const storageTableData: StorageTableData[] = pvcs.map((pvc) => ({
+    pvc,
+    storageClass: storageClasses.find((sc) => sc.metadata.name === pvc.spec.storageClassName),
```
pnaik1, line 28: `storageClass` takes undefined value when storageClass is deleted from `openshift` cluster

Gkrumbach07: thats fine. so you will use the pvc name as fallback

pnaik1: yeah, I have updated the pr according to what was discussed

frontend/src/pages/projects/screens/detail/storage/StorageTable.tsx
```diff
@@ -15,20 +20,65 @@ type StorageTableProps = {
...
-        rowRenderer={(pvc, i) => (
+        rowRenderer={(data, i) => (
           <StorageTableRow
-            key={pvc.metadata.uid}
+            key={data.pvc.metadata.uid}
             rowIndex={i}
-            obj={pvc}
+            storageClass={data.storageClass}
+            storageClassesLoaded={storageClassesLoaded}
+            pvc={data.pvc}
```
Gkrumbach07, line 81: you dont need to pass in storageClass or storageClassLoaded

data already has the storage class so just pass in `obj={data}`

Gkrumbach07 (changes requested):
just one comment

frontend/src/pages/projects/screens/detail/storage/StorageTable.tsx
pnaik1: The `storageClassesLoaded` must be passed because when the `storageClass` is in a loading state, it is undefined, which can be misinterpreted as a `deleted storageClass` for fraction of second. The `storageClassesLoaded` value comes from const [storageClasses, storageClassesLoaded] = useStorageClasses().

Screenshot 2024-09-20 at 11 14 05 AM

I have made the first changes

Gkrumbach07 (approved):
/lgtm
/approve

this looks good, thanks for the changes

PR #3094, opened 2024-08-15
[RHOAIENG-2404] Replace remaining PF deprecated components

jeff-phillips-18:
Closes [RHOAIENG-2404](https://issues.redhat.com/browse/RHOAIENG-2404)

## Description
Remove use of Patternfly deprecated components

## How Has This Been Tested?
Tested manually

## Test Impact
The following components are effected:

- Data Science Pipelines -> Filter type select
- Experiments -> Experiments and runs -> click an experiment -> select a run -> Compare runs -> change filter to Experiment -> Search input
- Data Science Projects -> select a project -> workbench tab -> create a workbench -> Use a data connection -> use existing data connection -> Select data connection
- Data Science Projects -> select a project -> workbench tab -> create a workbench -> Cluster storage -> use existing persistent storage -> Select a persistent storage
- Science Projects -> select a project -> Cluster storage tab -> Add cluster storage -> connected workbench select

## Request review criteria:
Self checklist (all need to be checked):
- [x] The developer has manually tested the changes and verified that the changes work
- [x] Commits have been squashed into descriptive, self-contained units of work (e.g. 'WIP' and 'Implements feedback' style messages have been removed)
- [x] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [x] The developer has added tests or explained why testing cannot be added (unit or cypress tests for related changes)

PR #3219, opened 2024-09-18
MR Permissions - Group and Project dropdown issues 

YuliaKrimerman:
[RHOAIENG-12025](https://issues.redhat.com/browse/RHOAIENG-12025)

## Description

 1.  Verify that once you've selected an item in the dropdown, if you then open the dropdown, you can NOW focus the typeahead text field without the need to close the dropdown( it was blocking the user from doing it) - try typing something in the typeahead in addition to the selected option 
 2. The Option in the bottom of the dropdown list is now Select and not Create 
 3. After you choose an option in the drop down, try deleting it in the typeahead and verify that it's not creating duplicate entries of that field in the dropdown(see screenshot in the issue) 

## How Has This Been Tested?

Tested on the UI 

## Test Impact

One test change to update the Create to Select . Also When changing the label to select I didn't change all the variable names of that label so they are still mostly shown as "create ...." in the code and not as select (except at the original line) 

## Request review criteria:

- [x
Screenshot 2024-09-18 at 10 40 12 AM
Screenshot 2024-09-18 at 10 40 29 AM
] The developer has added tests or explained why testing cannot be added (unit or cypress tests for related changes)

If you have UI changes: 

### DIFF
--- a/frontend/src/components/TypeaheadSelect.tsx
+++ b/frontend/src/components/TypeaheadSelect.tsx
@@ -250,12 +250,7 @@ const TypeaheadSelect: React.FunctionComponent<TypeaheadSelectProps> = ({
     if (onInputChange) {
       onInputChange(value);
     }
-    if (selected && value !== selected.content) {
-      // Clear the selection when the input value changes
-      if (onSelect) {
-        onSelect(undefined, '');
-      }
-    }
+
     resetActiveAndFocusedItem();
   };
 
--- a/frontend/src/components/TypeaheadSelect.tsx
+++ b/frontend/src/components/TypeaheadSelect.tsx
@@ -69,7 +69,7 @@ export interface TypeaheadSelectProps extends Omit<SelectProps, 'toggle' | 'onSe
 }
 
 const defaultNoOptionsFoundMessage = (filter: string) => `No results found for "${filter}"`;
-const defaultCreateOptionMessage = (newValue: string) => `Create "${newValue}"`;
+const defaultSelectOptionMessage = (newValue: string) => `Select "${newValue}"`;
 
 const defaultFilterFunction = (filterValue: string, options: TypeaheadSelectOption[]) =>
   options.filter((o) => String(o.content).toLowerCase().includes(filterValue.toLowerCase()));
@@ -88,7 +88,7 @@ const TypeaheadSelect: React.FunctionComponent<TypeaheadSelectProps> = ({
   noOptionsFoundMessage = defaultNoOptionsFoundMessage,
   isCreatable = false,
   isCreateOptionOnTop = false,
-  createOptionMessage = defaultCreateOptionMessage,
+  createOptionMessage = defaultSelectOptionMessage,
   isDisabled,
   toggleWidth,
   toggleProps,
@@ -212,12 +212,10 @@ const TypeaheadSelect: React.FunctionComponent<TypeaheadSelectProps> = ({
   const onInputClick = () => {
     if (!isOpen) {
       openMenu();
-      setTimeout(() => {
-        textInputRef.current?.focus();
-      }, 100);
-    } else if (isFiltering) {
-      closeMenu();
     }
+    setTimeout(() => {
+      textInputRef.current?.focus();
+    }, 100);
   };
 
   const selectOption = (
@@ -248,12 +246,17 @@ const TypeaheadSelect: React.FunctionComponent<TypeaheadSelectProps> = ({
   };
 
   const onTextInputChange = (_event: React.FormEvent<HTMLInputElement>, value: string) => {
-    setFilterValue(value || '');
+    setFilterValue(value);
     setIsFiltering(true);
     if (onInputChange) {
       onInputChange(value);
     }
-
+    if (selected && value !== selected.content) {
+      // Clear the selection when the input value changes
+      if (onSelect) {
+        onSelect(undefined, '');
+      }
+    }
     resetActiveAndFocusedItem();
   };
 
@@ -371,7 +374,7 @@ const TypeaheadSelect: React.FunctionComponent<TypeaheadSelectProps> = ({
     >
       <TextInputGroup isPlain>
         <TextInputGroupMain
-          value={isFiltering ? filterValue : selected?.content ?? ''}
+          value={isFiltering ? filterValue : (selected?.content ?? '')}
           onClick={onInputClick}
           onChange={onTextInputChange}
           onKeyDown={onInputKeyDown}
--- a/frontend/src/components/TypeaheadSelect.tsx
+++ b/frontend/src/components/TypeaheadSelect.tsx
@@ -69,8 +69,7 @@ export interface TypeaheadSelectProps extends Omit<SelectProps, 'toggle' | 'onSe
 }
 
 const defaultNoOptionsFoundMessage = (filter: string) => `No results found for "${filter}"`;
-const defaultSelectOptionMessage = (newValue: string) => `Select "${newValue}"`;
-
+const defaultCreateOptionMessage = (newValue: string) => `Create "${newValue}"`;
 const defaultFilterFunction = (filterValue: string, options: TypeaheadSelectOption[]) =>
   options.filter((o) => String(o.content).toLowerCase().includes(filterValue.toLowerCase()));
 
@@ -88,7 +87,7 @@ const TypeaheadSelect: React.FunctionComponent<TypeaheadSelectProps> = ({
   noOptionsFoundMessage = defaultNoOptionsFoundMessage,
   isCreatable = false,
   isCreateOptionOnTop = false,
-  createOptionMessage = defaultSelectOptionMessage,
+  createOptionMessage = defaultCreateOptionMessage,
   isDisabled,
   toggleWidth,
   toggleProps,
@@ -246,7 +245,7 @@ const TypeaheadSelect: React.FunctionComponent<TypeaheadSelectProps> = ({
   };
 
   const onTextInputChange = (_event: React.FormEvent<HTMLInputElement>, value: string) => {
-    setFilterValue(value);
+    setFilterValue(value || '');
     setIsFiltering(true);
     if (onInputChange) {
       onInputChange(value);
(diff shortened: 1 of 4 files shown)


frontend/src/components/TypeaheadSelect.tsx
```diff
@@ -248,12 +246,17 @@ const TypeaheadSelect: React.FunctionComponent<TypeaheadSelectProps> = ({
   };
 
   const onTextInputChange = (_event: React.FormEvent<HTMLInputElement>, value: string) => {
-    setFilterValue(value || '');
+    setFilterValue(value);
```
mturley, line 249: I'm curious if removing this was necessary?

frontend/src/components/TypeaheadSelect.tsx
```diff
@@ -248,12 +246,17 @@ const TypeaheadSelect: React.FunctionComponent<TypeaheadSelectProps> = ({
...
     if (onInputChange) {
       onInputChange(value);
     }
-
+    if (selected && value !== selected.content) {
+      // Clear the selection when the input value changes
+      if (onSelect) {
+        onSelect(undefined, '');
+      }
+    }
```
mturley, line 259: This seems possibly undesirable, the selection shouldn't change until the user clicks another option. I assume this was to fix the value being duplicated when there's a selected option that has a space and you type in the input? I wonder if there's another way to fix that.

frontend/src/components/TypeaheadSelect.tsx
```diff
@@ -212,12 +212,10 @@ const TypeaheadSelect: React.FunctionComponent<TypeaheadSelectProps> = ({
...
-      setTimeout(() => {
-        textInputRef.current?.focus();
-      }, 100);
-    } else if (isFiltering) {
-      closeMenu();
     }
+    setTimeout(() => {
+      textInputRef.current?.focus();
+    }, 100);
   };
```
mturley, line 219: This change seems reasonable to me and does fix the focus issue, but I'm curious if @jeff-phillips-18 has an opinion on it since he wrote this code in https://github.com/opendatahub-io/odh-dashboard/pull/3094.

frontend/src/components/TypeaheadSelect.tsx
mturley: Actually @YuliaKrimerman I think we should remove these lines, I troubleshooted a little and I think maybe the problem is outside this component. If you console.log `selectOptions` in `RoleBindingPermissionsNameInput` you'll see that actually when that option gets duplicated it is really being added in the list of options we pass down here. So it's something strange we're doing out there. It's also very strange that it only seems to happen for items that have a space in them.

I think it has something to do with these lines in `RoleBindingPermissionsNameInput`:

```ts
  if (value && !typeAhead.includes(value)) {
    selectOptions.push({ value, content: value });
  }
```

Ah -- it's because of the `namespaceToProjectDisplayName` thing we do there. When you select a project that includes spaces, that's a display name like "savitha MR test", which is what ends up in our `selectOptions` and is selected as our `value` there. But the `typeAhead` array is actual project names, like "savitha-mr-test". So the above lines are triggering adding the option to the list because it thinks it's missing.

I think the fix is to change the above line to:

```ts
  if (value && !selectOptions.some((option) => option.value === value)) {
    selectOptions.push({ value, content: value });
  }
```

mturley (changes requested):
A few things here @YuliaKrimerman . There's one thing I'd love to get @jeff-phillips-18 's opinion on - and if we can't get that settled within the week we should probably exclude it from this issue and open a followup issue for just that focus bug that isn't a MR blocker.

frontend/src/components/TypeaheadSelect.tsx
jeff-phillips-18: Looking at this, I would say it is ok to remove the timeout, but in the `<Select>` component below, add parameter:
```
shouldFocusFirstItemOnOpen={!selected}
```

This will keep focus on the input when the user already has something selected.

frontend/src/components/TypeaheadSelect.tsx
YuliaKrimerman: Screenshot 2024-09-19 at 2 03 59 PM
This is how it's behaving now when I implement it with the suggested fix. It does that and then after a few seconds it wipes out those funny duplicates and behave normally

frontend/src/components/TypeaheadSelect.tsx
```diff
@@ -401,6 +403,7 @@ const TypeaheadSelect: React.FunctionComponent<TypeaheadSelectProps> = ({
       onSelect={handleSelect}
       onOpenChange={(open) => !open && closeMenu()}
       toggle={toggle}
+      shouldFocusFirstItemOnOpen={false}
```
YuliaKrimerman, line 406: Added as suggested from Jeff

frontend/src/concepts/roleBinding/RoleBindingPermissionsNameInput.tsx
```diff
@@ -52,10 +52,6 @@ const RoleBindingPermissionsNameInput: React.FC<RoleBindingPermissionsNameInputP
     return { value: displayName, content: displayName };
   });
 
-  if (value && !typeAhead.includes(value)) {
-    selectOptions.push({ value, content: value });
-  }
-
```
YuliaKrimerman, line 58: Removed the code that was adding the current value to `selectOptions` if it wasn't in `typeAhead`. This ensures that custom items don't persist in the dropdown.

frontend/src/concepts/roleBinding/RoleBindingPermissionsNameInput.tsx
```diff
@@ -68,6 +64,18 @@ const RoleBindingPermissionsNameInput: React.FC<RoleBindingPermissionsNameInputP
         }
       }}
       placeholder={placeholderText}
+      createOptionMessage={(newValue) => `Select "${newValue}"`}
+      filterFunction={(filterValue, options) => {
```
YuliaKrimerman, line 68: Added this filterFunction in order to achieve the [requirement](https://issues.redhat.com/browse/RHOAIENG-12025#:~:text=It%20should%20act%20like%20this%20example%20(but%20with%20Select%20instead%20of%20Use)%20%2D%20if%20the%20user%20selects%20a%20custom%20item%20like%20this%2C%20then%20changes%20their%20selection%2C%20the%20custom%20item%20should%20disappear.) It filters the existing options based on the input and adds a new option with the exact input value if no matches are found.

frontend/src/components/TypeaheadSelect.tsx
YuliaKrimerman: Addressed all other comments and kept the original implementation for this, let me know if it's ok to keep it, or if we want to take out this issue for the future.

frontend/src/components/TypeaheadSelect.tsx
mturley: @jeff-phillips-18 what about the removal of the `closeMenu()` call here? I don't think I'm seeing the menu close when clicking the input anyway, I wonder if that's getting counteracted.

frontend/src/concepts/roleBinding/RoleBindingPermissionsNameInput.tsx
mturley: Ah fair point, I guess removing this entirely solves that as well

frontend/src/components/TypeaheadSelect.tsx
mturley: @YuliaKrimerman with your other change in place to remove the `selectOptions.push` code, I tried removing the lines you added here (clear the selection when the input value changes) and it seems to behave like I would expect. I'm not seeing the above duplicates. Can you try removing this now and see if it works for you?

frontend/src/concepts/roleBinding/RoleBindingPermissionsNameInput.tsx
mturley: I don't see any different behavior here when I try removing this `filterFunction`. With or without it, I'm actually seeing that if I click the new option that doesn't match any select items, it doesn't remain selected in the dropdown. I think that's because we still need it to be one of the `selectOptions` passed above for it to show as selected.

I also notice that with this `filterFunction` in place, if you type text that doesn't even partially match any other options, it gets shown as the sole option without the "Select ___" text. So I think we don't want this `filterFunction` and we need to fix the issue some other way.

mturley: @YuliaKrimerman I think I figured it out - Try this. Remove this `filterFunction`, and above where you removed the other code, add this:

```ts
  // If we've selected an option that doesn't exist via isCreatable, include it in the options so it remains selected
  if (!selectOptions.some((option) => option.value === value)) {
    selectOptions.push({ value, content: value });
  }
```

That ensures that our "newly created" item is in the list once it is selected, but disappears from the list if it is deselected.

mturley (commented):
Sorry for the roundabout comments, I have a solution for the problem I identified here:

YuliaKrimerman:
So I ended up removing the stuff you suggested to remove and not adding the code you proposed because it looks like it was creating this pre selected empty field. At the current state everything looks like it's working as desired, let me know what you think 
Screenshot 2024-09-20 at 8 49 46 AM
Screenshot 2024-09-20 at 8 50 54 AM

mturley:
@YuliaKrimerman without the code I proposed, when I type "istio" and click the "Select 'istio'" option, the dropdown appears as if nothing is selected even though the `value` has updated (the confirm button in the row becomes enabled).

I do notice the issue you pointed out with my suggestion, it's because I forgot to make sure `value` is defined before pushing it as an option. This should fix it:

```ts
  // If we've selected an option that doesn't exist via isCreatable, include it in the options so it remains selected
  if (value && !selectOptions.some((option) => option.value === value)) {
    selectOptions.push({ value, content: value });
  }
```

YuliaKrimerman:
@mturley Yep, you're right, now it works good. Pushed the latests and ready for re-review

mturley (approved):
Perfect, thanks @YuliaKrimerman !!

PR #3224, opened 2024-09-18
fix(12787): high page size for model versions and artifacts

gitdallas:
additional URLs for versions/artifacts for
https://issues.redhat.com/browse/RHOAIENG-12787

If you have UI changes: 


mturley (approved):
Thanks @gitdallas

PR #3225, opened 2024-09-18
connection type compatibility column

christianvogt:
https://issues.redhat.com/browse/RHOAIENG-13111

## Description

Compatibility is assumed based on inspection of the environment variables stored in the connection type and secret by name. A model serving compatible connection requires the following env vars: 'AWS_ACCESS_KEY_ID', 'AWS_SECRET_ACCESS_KEY', 'AWS_S3_ENDPOINT', 'AWS_S3_BUCKET'

If all of these env vars are present, then we assume it is model serving compatible.

Connection instances:
image

Connection types:
image

## How Has This Been Tested?

Enable the connection types feature flag.

Connection types:
- Create a connection type in the admin section. 
- Add fields with at least the following required env vars: 'AWS_ACCESS_KEY_ID', 'AWS_SECRET_ACCESS_KEY', 'AWS_S3_ENDPOINT', 'AWS_S3_BUCKET'
- Observe the connection type displays `Model serving` in the compatibility column

Connection instances:
- Create a data connection. Make sure to include a value for the bucket.
- Navigate to the project details `Connections` tab
- Observe the data connection displays `Model serving` in the compatibility column

## Test Impact

Added unit and cypress tests.

## Request review criteria:

If you have UI changes: 

cc @simrandhaliw

### DIFF
--- a/frontend/src/pages/projects/screens/detail/connections/ConnectionsTableRow.tsx
+++ b/frontend/src/pages/projects/screens/detail/connections/ConnectionsTableRow.tsx
@@ -1,9 +1,11 @@
 import * as React from 'react';
 import { ActionsColumn, Td, Tr } from '@patternfly/react-table';
-import { Truncate } from '@patternfly/react-core';
+import { LabelGroup, Truncate } from '@patternfly/react-core';
 import { Connection, ConnectionTypeConfigMapObj } from '~/concepts/connectionTypes/types';
 import { TableRowTitleDescription } from '~/components/table';
 import { getDescriptionFromK8sResource, getDisplayNameFromK8sResource } from '~/concepts/k8s/utils';
+import { getCompatibleTypes } from '~/concepts/connectionTypes/utils';
+import CompatibilityLabel from '~/concepts/connectionTypes/CompatibilityLabel';
 
 type ConnectionsTableRowProps = {
   obj: Connection;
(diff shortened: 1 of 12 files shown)


frontend/src/pages/projects/screens/detail/connections/ConnectionsTableRow.tsx
```diff
@@ -28,6 +30,8 @@ const ConnectionsTableRow: React.FC<ConnectionsTableRowProps> = ({
     );
   }, [obj, connectionTypes]);
 
+  const compatibleTypes = getCompatibleTypes(Object.keys(obj.data));
```
emilys314, line 33: On my cluster's project, I had a connection without any fields set (so an empty `data`). This caused the page to through an exception since it was undefined. Right now in the `Connection` type, it's required, but I was thinking of making it optional since for creation you use `stringData` instead.

Anyways.. should we allow creation of connections with no fields?? In my upcoming add connection modal PR, i made data an optional field anyways

christianvogt: I'll add more defensive code here. Looks like types may be wrong for secrets.

christianvogt: @emilys314 I changed the type for `Connection` to have optional `data` and also updated this code to check not only that the key exists but also that there is a value.

christianvogt:
rebased
Needed to kick start the retest

emilys314:
/lgtm

looks good, everything is working

<details>
<summary> screenshots </summary>

image
image

</details>

PR #3229, opened 2024-09-19
Fix stack chart issue when there is no data for one area

DaoDaoNoCode:
JIRA: https://issues.redhat.com/browse/RHOAIENG-11048

## Description

This is a known issue in Victory chart, and I am taking the workaround as suggested by the PF team here https://github.com/patternfly/patternfly-react/issues/10978#issuecomment-2346917642. If there is no data, we render it as `[null]` instead of `[]`, which fixes the issue.

## How Has This Been Tested?

I met some problems with generating the real data points, but you can go to the cypress test server and check the test case `Serving Chart Shows Data`

Way to compare:
1. Run `cypress:server` and `cypress:open:mock` in 2 terminals
2. Add `it.only` to the `Serving Chart Shows Data` test so we can check the screenshot
3. Go to `modelMetrics` test to check if it renders incorrectly
4. Run `cypress:server:build` and re-run step 1-3
5. You can see the data renders correctly

## Test Impact

Update the test cases to see the mocked data better.

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/__tests__/cypress/cypress/tests/mocked/modelServing/modelMetrics.cy.ts
+++ b/frontend/src/__tests__/cypress/cypress/tests/mocked/modelServing/modelMetrics.cy.ts
@@ -105,10 +105,21 @@ const initIntercepts = ({
     ServingRuntimeModel,
     mockServingRuntimeK8sResource({ name: 'test-model', namespace: 'test-project' }),
   );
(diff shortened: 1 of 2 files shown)


Gkrumbach07 (approved):
/lgtm
/approve

this looks good to me

PR #3217, opened 2024-09-18
Add connection modal

emilys314:
https://issues.redhat.com/browse/RHOAIENG-11555

## Description

Creates a new connection in the form of a secret on the project.

<details>
<summary> SCREENSHOTS </summary>

image
image

image
result:
image

image

</details>

## How Has This Been Tested?

If you have a connection type named s3, then you can view the results in data connections. otherwise check the secrets to verify the creation

## Test Impact

cypress and jest tests added

## Request review criteria:

Self checklist (all need to be checked):
- [X] The developer has manually tested the changes and verified that the changes work
- [X] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [x] The developer has added tests or explained why testing cannot be added (unit or cypress tests for related changes)

If you have UI changes: 

- [x] Included any necessary screenshots or gifs if it was a UI change.
- [x] Included tags to the UX team if it was a UI/UX change.

@simrandhaliw 

After the PR is posted & before it merges:
- [ ] The developer has tested their solution on a cluster by using the image produced by the PR to `main`

PR #3234, opened 2024-09-20
adjust resource name form spacing and help icon

christianvogt:
## Description

It was observed while reviewing the new [create connection modal](https://github.com/opendatahub-io/odh-dashboard/pull/3217) that the k8s name / description component didn't fit exactly the UX for connection types.

Two differences observed:
- The spacing between form elements is 16px as oppose to 24px. PF `Form` and PF `FormSection` provide specific spacing between elements of the form. This component was using a stack layout with a different gap.
- The help icon and tooltip vs popover differed from those in the connection type form for form field descriptions.

The change here replaces the stack layout with a `FormSection`. Since there is no title to this section, the margin is eliminated to allow for these form elements to fit into existing forms.

The 2nd change is to use a popover and `OutlinedQuestionCircleIcon` icon. There's a lot of inconsistencies in our product today for using tooltips vs popovers and `HelpIcon` vs `OutlinedQuestionCircleIcon` in many of our forms. While both methods could be considered correct, it looks inconsistent within the connection form. The `K8sNameDescriptionField` component currently is not used in a form which has another `?` icon therefore this change does not produce further inconsistency within a single form.

Before:
image
image
image
Gap is 16px:
image

After:
image
image
image
Gap is 24px as prescribed by PF forms:
image

## How Has This Been Tested?

From the project list view page, click to create a new project to open the modal.

## Test Impact

N/A, visual change only

## Request review criteria:

If you have UI changes: 

cc @andrewballantyne @simrandhaliw

### DIFF
--- a/frontend/src/concepts/k8s/K8sNameDescriptionField/K8sNameDescriptionField.tsx
+++ b/frontend/src/concepts/k8s/K8sNameDescriptionField/K8sNameDescriptionField.tsx
@@ -3,10 +3,9 @@ import {
   Button,
   FormGroup,
   FormHelperText,
+  FormSection,
   HelperText,
   HelperTextItem,
-  Stack,
-  StackItem,
   TextArea,
   TextInput,
 } from '@patternfly/react-core';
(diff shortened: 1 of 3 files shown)


andrewballantyne (commented):
Sweet, I didn't even consider the usage without stack -- I mirrored what we had already done. Better to use the form spacing anyways 👍

simrandhaliw:
Thank you for observing and fixing these inconsistencies! LGTM!

PR #3236, opened 2024-09-20
Fix validation in the storage modal mount path

DaoDaoNoCode:
JIRA: https://issues.redhat.com/browse/RHOAIENG-7680

## Description

There is something wrong with the logic in the helper text and what error text we should show. I made some changes to that to correct the logic. Also, change the regex to allow a trailing `/` in the path.

Initial state:

Screenshot 2024-09-20 at 3 12 01 PM

Trailing slash:

Screenshot 2024-09-20 at 3 12 24 PM

Duplicate path:

Screenshot 2024-09-20 at 3 12 47 PM

Empty input:

Screenshot 2024-09-20 at 3 13 05 PM

Invalid value (contains number):

Screenshot 2024-09-20 at 3 13 30 PM

Invalid value (slash is not the trailing slash):

Screenshot 2024-09-20 at 3 14 32 PM

## How Has This Been Tested?

1. Create a cluster storage
2. Attach it to a workbench
3. Input in the mouth path field, check if it's allowing a trailing slash
4. Check all the error states as listed above

## Test Impact

Updated the cypress test to check the error cases.

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/pages/projects/pvc/MountPathField.tsx
+++ b/frontend/src/pages/projects/pvc/MountPathField.tsx
@@ -37,11 +37,11 @@ const MountPathField: React.FC<MountPathFieldProps> = ({
           onChange={(e, value) => {
             let error = '';
             if (value.length === 0) {
-              error = 'Required';
-            } else if (!/^[a-z-]+$/.test(value)) {
-              error = 'Must only consist of lower case letters and dashes';
+              error = 'Enter a path to a model or folder. This path cannot point to a root folder.';
+            } else if (!/^[a-z-]+\/?$/.test(value)) {
+              error = 'Must only consist of lower case letters and dashes.';
             } else if (inUseMountPaths.includes(`/${value}`)) {
-              error = 'Mount folder is already in use for this workbench';
+              error = 'Mount folder is already in use for this workbench.';
             }
             setMountPath({ value, error });
           }}
@@ -50,10 +50,11 @@ const MountPathField: React.FC<MountPathFieldProps> = ({
     </InputGroup>
     <FormHelperText>
       <HelperText>
-        <HelperTextItem variant={mountPath.error ? 'error' : 'default'}>
-          {mountPath.error
-            ? 'Enter a path to a model or folder. This path cannot point to a root folder.'
-            : 'Must consist of lower case letters and dashes.'}
+        <HelperTextItem
+          variant={mountPath.error ? 'error' : 'default'}
+          data-testid="mount-path-folder-helper-text"
+        >
+          {mountPath.error || 'Must consist of lower case letters and dashes.'}
         </HelperTextItem>
       </HelperText>
     </FormHelperText>
--- a/frontend/src/__mocks__/mockNotebookK8sResource.ts
+++ b/frontend/src/__mocks__/mockNotebookK8sResource.ts
@@ -1,7 +1,7 @@
 import _ from 'lodash-es';
 import { KnownLabels, NotebookKind } from '~/k8sTypes';
 import { DEFAULT_NOTEBOOK_SIZES } from '~/pages/projects/screens/spawner/const';
-import { ContainerResources, TolerationEffect, TolerationOperator } from '~/types';
+import { ContainerResources, TolerationEffect, TolerationOperator, VolumeMount } from '~/types';
 import { genUID } from '~/__mocks__/mockUtils';
 import { RecursivePartial } from '~/typeHelpers';
 
@@ -17,6 +17,7 @@ type MockResourceConfigType = {
   lastImageSelection?: string;
   opts?: RecursivePartial<NotebookKind>;
   uid?: string;
+  additionalVolumeMounts?: VolumeMount[];
 };
 
 export const mockNotebookK8sResource = ({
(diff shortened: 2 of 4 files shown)


DaoDaoNoCode:
@andrewballantyne The bug you mentioned in https://github.com/opendatahub-io/odh-dashboard/pull/3231#pullrequestreview-2318816652 is not because of the connection path starts with a preexisting connection path, it's because you have a number `2` in your path, which is not allowed. The code had wrong logic of checking the error cases so it was showing an incorrect error message. The logic is fixed now in this PR.

andrewballantyne:
That logic is likely wrong fwiw -- but I suppose this PR doesn't have to deal with that

frontend/src/pages/projects/pvc/MountPathField.tsx
```diff
@@ -50,10 +50,11 @@ const MountPathField: React.FC<MountPathFieldProps> = ({
...
       <HelperText>
-        <HelperTextItem variant={mountPath.error ? 'error' : 'default'}>
-          {mountPath.error
-            ? 'Enter a path to a model or folder. This path cannot point to a root folder.'
-            : 'Must consist of lower case letters and dashes.'}
+        <HelperTextItem
+          variant={mountPath.error ? 'error' : 'default'}
+          data-testid="mount-path-folder-helper-text"
+        >
+          {mountPath.error ? mountPath.error : 'Must consist of lower case letters and dashes.'}
```
jpuzz0, line 57: nit, can be simplified to:
```suggestion
          {mountPath.error || 'Must consist of lower case letters and dashes.'}
```

frontend/src/pages/projects/pvc/MountPathField.tsx
```diff
@@ -37,8 +37,8 @@ const MountPathField: React.FC<MountPathFieldProps> = ({
           onChange={(e, value) => {
             let error = '';
             if (value.length === 0) {
-              error = 'Required';
-            } else if (!/^[a-z-]+$/.test(value)) {
+              error = 'Enter a path to a model or folder. This path cannot point to a root folder.';
```
jpuzz0, line 40: Was this text been provided by UX? Maybe @kaedward could verify this tech content.

jpuzz0 (approved):
Verified locally. Although you did not write the error messages of `Must only consist of lower case letters and dashes` and `Mount folder is already in use for this workbench`, I think these both are meant to end with periods.

Worth having @kaedward take a look at all of these screenshots of yours to be sure anyways. I don't consider this blocking.

frontend/src/pages/projects/pvc/MountPathField.tsx
DaoDaoNoCode: Yeah I just moved the old text here https://github.com/opendatahub-io/odh-dashboard/blob/main/frontend/src/pages/projects/pvc/MountPathField.tsx#L55 to above because the helper text render logic was incorrect before.

frontend/src/pages/projects/pvc/MountPathField.tsx
DaoDaoNoCode: Updated.

Gkrumbach07:
@DaoDaoNoCode it will except the path if you put a just a trailing slash, but will reject without it or if you add another folder to the end of the trailing slash

DaoDaoNoCode:
@Gkrumbach07 Reading the comment that @andrewballantyne left in the ticket:
```
I read the issue fast and missed your final remark – I was on a time crunch in the triage meeting and this was one of our last ones we were trying to get through.

I'll update my UX ping and just leave this on the backlog. I didn't know `/something/` was valid and I would think that this is equiv to `/something` anyways.
```
I think `/something` is different from `/something/`, did I understand wrongly? 🤔

andrewballantyne:
Read the last line of the description

bq. The trailing slash is still valid for the notebook object and should not throw an error in the UI. If you manually update the notebook object with a trailing slash it works perfectly fine.

Allow them to add a slash if they want.

Gkrumbach07:
if 
> /something is different from /something/

is true then this looks good, otherwise needs to be fixed?

DaoDaoNoCode:
@Gkrumbach07 I tested with both `/something` and `/something/` set as the path and it works well, I think the notebook itself could handle the path. If that's not the case we can refine it later I guess.

Gkrumbach07:
no it can handle the path. i am saying it allows me to create two pvcs one with /something and the other with /something/... both attached to the same workbench. i tested it and it wont let the notebook start

DaoDaoNoCode:
@Gkrumbach07 My notebook start and running

Screenshot 2024-09-25 at 3 42 13 PM

Gkrumbach07:
i still cannot replicate it. Are you restarting the workbench pods after you add both PVCs to it

DaoDaoNoCode:
@Gkrumbach07 Hmm, I added it to a stopped notebook, and click toggle the start manually.

Gkrumbach07:
well i cant get a workbench to spawn at all im pretty sure so something is wrong on my end. I wont hold this PR up then

/lgtm
/approve

PR #3238, opened 2024-09-23
workaround PF bug with modals causing cypress selectors to fail

christianvogt:
https://issues.redhat.com/browse/RHOAIENG-11946

## Description

A bug in PF `Modal` is causing the modal DOM node container to have `aria-hidden="true"` when running in our dev server which uses `React.StrictMode`. As a result, the cypress selector fails to find the modal.

The workaround here is to supply `hidden: true` to the selector. As per the documentation, this will include the hidden element in the result:
```
  /**
   * If true includes elements in the query set that are usually excluded from
   * the accessibility tree. `role="none"` or `role="presentation"` are included
   * in either case.
   */
  hidden?: boolean
```

## How Has This Been Tested?

`npm run cypress:server:dev`
`npm run cypress:open:mock`

Run the `connectionTypes.cy.ts` test suite. Specifically the delete test.

## Test Impact

N/A

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/__tests__/cypress/cypress/pages/components/Modal.ts
+++ b/frontend/src/__tests__/cypress/cypress/pages/components/Modal.ts
@@ -12,7 +12,10 @@ export class Modal {
   }
 
   find(): Cypress.Chainable<JQuery<HTMLElement>> {
-    return cy.findByRole('dialog', { name: this.title });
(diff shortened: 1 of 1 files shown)


jpuzz0 (approved):

christianvogt:
rebased to kick all actions to re-run

PR #3239, opened 2024-09-23
Replace model registry hard-coded namespace with namespace from DSC

ppadti:
Closes: [RHOAIENG-12311](https://issues.redhat.com/browse/RHOAIENG-12311)

## Description
This PR aims to replace the model registry hard-coded namespace with the namespace from DSC.

## How Has This Been Tested?
make sure the UI works as expected. 

## Test Impact
Updated unit tests and cypress tests. 

## Request review criteria:

If you have UI changes: 


mturley (commented):
@ppadti do you also mind editing the `Makefile` to use a `MODEL_REGISTRY_NAMESPACE` env var, and adding that to `.env.development` with the default set to odh-model-registries? That'll make it easier to test against RHOAI clusters locally if we need to.

Reviewing the rest now.

backend/src/utils/proxy.ts
```diff
@@ -37,7 +37,7 @@ export const proxyService =
         internalPort: number | string;
         prefix?: string;
         suffix?: string;
-        namespace?: string;
+        namespace?: string | ((fastify: KubeFastifyInstance) => Promise<string>);
```
mturley, line 40: It looks like you never await the function version of this in the usage below. This means when we use `namespace` in string templates below we actually get the string `"[Function: getModelRegistryNamespace]"` instead of the namespace. We need to check the type and await the function if it is not a string - or if we want to avoid run-time type checking we could just change this to `getNamespace` and always take it as a function to be awaited, and then in the usages that currently pass a string they can pass `() => Promse.resolve("some-namespace")`.

This is a bit more complex actually... `fastify.register()` needs the namespace as part of its `prefix` param, so we need to check if `service.namespace` is a function and await it before we even call `fastify.register()`. I assume that will still work since the function is async, but I'm not actually certain. It'll need testing.

backend/src/routes/api/modelRegistries/modelRegistryUtils.ts
```diff
@@ -1,11 +1,16 @@
...
 import { KubeFastifyInstance, ModelRegistryKind, RecursivePartial } from '../../../types';
 import { PatchUtils, V1Secret, V1Status } from '@kubernetes/client-node';
+import { getClusterStatus } from '../../../utils/dsc';
 
 const MODEL_REGISTRY_API_GROUP = 'modelregistry.opendatahub.io';
 const MODEL_REGISTRY_API_VERSION = 'v1alpha1';
 const MODEL_REGISTRY_PLURAL = 'modelregistries';
 
+export const getModelRegistryNamespace = async (fastify: KubeFastifyInstance): Promise<string> => {
+  const modelRegistryNamespace = await getClusterStatus(fastify);
```
mturley, line 10: In case you look at this @andrewballantyne - I considered giving the feedback here based on our discussion that we should instead add a `resourceWatcher` to `backend/src/utils/resourceUtils.ts` and use that, but I don't think that will work because we need this resolved at proxy registration time and the initial fetch of the watchers may not be complete by then (if I understand it correctly).

For the regular user's consumption of the MR API via `proxyService`, we'll only fetch the namespace once (at proxy registration time), so that part is fine. However, as implemented, we will be fetching the namespace each time the admin lists/creates/deletes model registries. Perhaps we could also just await it before calling the functions to register routes in `backend/src/routes/api/modelRegistries/index.ts`. @ppadti that might be worth trying - await getting the namespace once when we set up those routes (before calling `fastify.get/post/patch/delete`) and pass the string namespace into the utilities below (`listModelRegistries`, etc) instead of awaiting the namespace in those utils. That'll save us from having an extra network call for the namespace on every interaction in the settings page.

mturley (changes requested):
A few important things here about how/when we await the function to get the namespace.

mturley (commented):
One more small thing - testing the rest now

mturley (approved):
LGTM!! Tested on the modelregistry-ui cluster with the quay image, both in prod and with start:dev:ext locally. Thanks so much @ppadti

PR #3247, opened 2024-09-24
Restart all connected notebooks when the pvc size changes

DaoDaoNoCode:
JIRA: [RHOAIENG-147](https://issues.redhat.com/browse/RHOAIENG-147)

## Description

Find all the connected notebooks to the PVC, and when detects the PVC size changes, add the `notebooks.opendatahub.io/notebook-restart=true` annotation to the workbench to trigger a restart.

## How Has This Been Tested?

1. Create a workbench
2. Make sure it's running
3. Find the attached PVC to that workbench
4. Update that PVC, increase the size by 1 GB
5. Make sure the workbench restarts
6. You can try to attach the PVC to more workbenches and make sure the size increase could trigger the restart of all the attached workbenches

## Test Impact

Added a unit test to verify the restart notebook function triggers the correct patch method.

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/api/k8s/__tests__/notebooks.spec.ts
+++ b/frontend/src/api/k8s/__tests__/notebooks.spec.ts
@@ -30,6 +30,7 @@ import {
   getStopPatch,
   startPatch,
   mergePatchUpdateNotebook,
+  restartNotebook,
 } from '~/api/k8s/notebooks';
 
 import {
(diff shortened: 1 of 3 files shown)


Gkrumbach07 (approved):
this looks good and tested and works
/lgtm

frontend/src/api/k8s/__tests__/notebooks.spec.ts
```diff
@@ -955,3 +956,30 @@ describe('removeNotebookSecret', () => {
...
+  it('should add restart notebook annotation', async () => {
+    const name = 'test-notebook';
+    const namespace = 'test-project';
+    const notebookMock = mockNotebookK8sResource({ uid });
+
+    k8sGetResourceMock.mockResolvedValue(notebookMock);
+    k8sPatchResourceMock.mockResolvedValue(notebookMock);
+
+    const renderResult = await restartNotebook(name, namespace);
+    expect(k8sPatchResourceMock).toHaveBeenCalledWith({
```
jiridanek, line 970: Just a point of curiosity; in this test, why are you checking the arguments of the patch call and not asserting on the resulting CR after it was patched?

Essentially, I'm asking about this
> One can write unit tests either the classicist or the mockist way as per http://martinfowler.com/articles/mocksArentStubs.html
(quote from https://stackoverflow.com/questions/8666495/is-there-any-advantage-to-writing-both-classist-and-mockist-unit-tests-for-a-sin)

---

edit: never mind, I get it now; the patch is applied on the server (k8s api server), not in your program.

frontend/src/api/k8s/notebooks.ts
```diff
@@ -490,3 +490,28 @@ export const removeNotebookSecret = (
...
+
+export const restartNotebook = (
+  notebookName: string,
+  namespace: string,
+  opts?: K8sAPIOptions,
+): Promise<NotebookKind> => {
+  const patches: Patch[] = [
+    {
+      op: 'add',
+      path: '/metadata/annotations/notebooks.opendatahub.io~1notebook-restart',
```
jiridanek, line 502: There's this notebook updating (and restarting!) logic in a different place? Could it have been used instead?

https://github.com/opendatahub-io/odh-dashboard/blob/d5938bfbe56de729fa6927691ad06d6aecdeb9b2/frontend/src/pages/projects/screens/spawner/SpawnerFooter.tsx#L164-L167

DaoDaoNoCode: Hello @jiridanek! Thanks for your reviews. These 2 places have different logic. The PR is trying to find out all the connected notebooks to a PVC and simply patch a restart annotation to all the notebooks to trigger the restart, and the code you mentioned here is trying to update one notebook in one single network call, it's not only patching the annotations, but also containing many other fields, and update them all together. So I think it's hard to merge these 2 places.

PR #3248, opened 2024-09-24
Don't crash if modelregistries namespace is not present in the DSC

mturley:
Followup to https://github.com/opendatahub-io/odh-dashboard/pull/3239
Needed for https://issues.redhat.com/browse/RHOAIENG-12311

## Description

[See slack thread here](https://redhat-internal.slack.com/archives/C07C0447EAV/p1727200812929609). The above PR creates a situation where if model registry is not enabled on a cluster, the dashboard will crash on server start. This PR is an immediate fix for the crash, and I'll follow up with a second PR to switch us to using a ResourceWatcher so we make sure the namespace value is current in case it is not available yet when the dashboard starts.

This PR also fixes the same type in the frontend to make sure we are always being defensive when we try to read this value.

## How Has This Been Tested?

Tested on https://console-openshift-console.apps.ods-qe-psi-01.osp.rh-ods.com/ which is old enough that its DSC has no modelregistry spec or status. Running the backend locally on main, verified that it crashes when using this cluster's API. Running the same way on this branch there is no crash.

## Test Impact

N/A, no tests in the backend.

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/backend/src/routes/api/modelRegistries/modelRegistryUtils.ts
+++ b/backend/src/routes/api/modelRegistries/modelRegistryUtils.ts
@@ -6,9 +6,20 @@ const MODEL_REGISTRY_API_GROUP = 'modelregistry.opendatahub.io';
 const MODEL_REGISTRY_API_VERSION = 'v1alpha1';
 const MODEL_REGISTRY_PLURAL = 'modelregistries';
 
-export const getModelRegistryNamespace = async (fastify: KubeFastifyInstance): Promise<string> => {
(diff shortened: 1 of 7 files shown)


gitdallas (approved):

andrewballantyne (approved):

mturley:
/hold
We're also not strict enough with the same type in the frontend, will fix momentarily

gitdallas (approved):

gitdallas (approved):

PR #3249, opened 2024-09-24
feat(13045): mr permissions and rolebindings

gitdallas:
use a separate endpoint for MR based rolebinding calls which will automatically use the MR namespace and labels

## How Has This Been Tested?
tested locally, things still work as expected

## Test Impact
mocks changed in tests to match new urls

## Request review criteria:
make sure rolebinding still works as expected for given role. this would include listing, adding, and deleting

### DIFF
--- a/backend/src/routes/api/modelRegistryRoleBindings/index.ts
+++ b/backend/src/routes/api/modelRegistryRoleBindings/index.ts
@@ -1,0 +1,73 @@
+import { FastifyReply, FastifyRequest } from 'fastify';
+import { secureAdminRoute } from '../../../utils/route-security';
+import { KubeFastifyInstance } from '../../../types';
+import {
+  createModelRegistryRoleBinding,
+  deleteModelRegistriesRolebinding,
+  listModelRegistryRoleBindings,
+} from './modelRegistryRolebindingsUtils';
+import { V1RoleBinding } from '@kubernetes/client-node';
+import { getModelRegistryNamespace } from '../modelRegistries/modelRegistryUtils';
+
+export default async (fastify: KubeFastifyInstance): Promise<void> => {
+  fastify.get(
+    `/`,
+    secureAdminRoute(fastify)(async (request: FastifyRequest, reply: FastifyReply) => {
+      try {
+        const mrNamespace = getModelRegistryNamespace(fastify);
+        return listModelRegistryRoleBindings(fastify, mrNamespace);
+      } catch (e) {
+        fastify.log.error(
+          `ModelRegistry RoleBindings could not be listed, ${
+            e.response?.body?.message || e.message
+          }`,
+        );
+        reply.send(e);
+      }
+    }),
+  );
+
+  fastify.post(
+    '/',
+    secureAdminRoute(fastify)(
+      async (request: FastifyRequest<{ Body: V1RoleBinding }>, reply: FastifyReply) => {
+        const rbRequest = request.body;
+        try {
+          const mrNamespace = getModelRegistryNamespace(fastify);
+          return createModelRegistryRoleBinding(fastify, rbRequest, mrNamespace);
+        } catch (e) {
+          if (e.response?.statusCode === 409) {
+            fastify.log.warn(`Rolebinding already present, skipping creation.`);
+            return {};
+          }
+
+          fastify.log.error(
+            `rolebinding could not be created: ${e.response?.body?.message || e.message}`,
+          );
+          reply.send(new Error(e.response?.body?.message));
+        }
+      },
+    ),
+  );
+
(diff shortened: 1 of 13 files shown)


backend/src/types.ts
```diff
@@ -1170,6 +1171,27 @@ export type TrustyAIKind = K8sResourceCommon & {
...
+};
+
+export type RoleBindingKind = K8sResourceCommon & {
+  metadata: {
+    name: string;
+    namespace: string;
+  };
+  subjects: RoleBindingSubject[];
+  roleRef: RoleBindingRoleRef;
+};
```
mturley, line 1193: @gitdallas I just realized you should be able to just import `V1RoleBinding`, `V1RoleRef` and `V1Subject` from `'@kubernetes/client-node'` as we do in `backend/src/routes/api/rolebindings/index.ts` instead of creating these types

backend/src/routes/api/modelRegistryRoleBindings/modelRegistryRolebindingsUtils.ts
```diff
@@ -0,0 +1,55 @@
...
+  roleBindingName: string,
+  mrNamespace: string,
+): Promise<K8sStatus> => {
+  const { body } = (await fastify.kube.customObjectsApi.deleteNamespacedCustomObject(
+    MODEL_REGISTRY_ROLE_BINDING_API_GROUP,
+    MODEL_REGISTRY_ROLE_BINDING_API_VERSION,
+    mrNamespace,
+    MODEL_REGISTRY_ROLE_BINDING_PLURAL,
+    roleBindingName,
+  )) as { response: IncomingMessage; body: K8sStatus };
```
mturley, line 52: Can we make the pattern consistent with the other two functions?
```ts
const response = await (/*...*/) as Promise<{ body: K8sStatus }>;
return response.body
```
Then you also don't have to import `IncomingMessage`

frontend/src/pages/modelRegistrySettings/ModelRegistriesPermissions.tsx
```diff
@@ -121,6 +125,8 @@ const ModelRegistriesManagePermissions: React.FC = () => {
               }
               roleBindingPermissionsRB={{ ...roleBindings, data: filteredRoleBindings }}
               groups={groups}
+              createRoleBinding={createModelRegistryRoleBinding}
+              deleteRoleBinding={deleteModelRegistryRoleBinding}
```
mturley, line 129: I'm actually kinda surprised typescript allows us to assign `(roleBindingName: string) => Promise<K8sStatus>` to type `(name: string, namespace: string) => Promise<K8sStatus>`, but considering the extra arg will just be ignored in the MR case this seems fine

mturley (commented):
Mostly LGTM, a couple minor things

backend/src/routes/api/modelRegistryRoleBindings/index.ts
```diff
@@ -0,0 +1,72 @@
...
+import {
+  createModelRegistryRoleBinding,
+  deleteModelRegistriesRolebinding,
+  listModelRegistryRoleBindings,
+} from './modelRegistryRolebindingsUtils';
+import { V1RoleBinding } from '@kubernetes/client-node';
+import { getModelRegistryNamespace } from '../modelRegistries/modelRegistryUtils';
+
+export default async (fastify: KubeFastifyInstance): Promise<void> => {
+  const mrNamespace = await getModelRegistryNamespace(fastify);
```
mturley, line 13: @gitdallas sorry I didn't catch this when you first rebased - `getModelRegistryNamespace` changed, it's no longer async and instead of calling it at the top of this exported function it needs to be called within the `try` inside each of the fastify callbacks below (right before `mrNamespace` is used in each one). See how it is used now in `backend/src/routes/api/modelRegistries/index.ts`.

mturley (changes requested):
One thing that came up in the rebase.

Otherwise, I've tested this on the modelregistry-ui cluster by adding regularuser1 to the odh-admins group, and that user is now able to interact with the permission management for a registry. If you just fix this one thing we're good here.

backend/src/routes/api/modelRegistryRoleBindings/index.ts
gitdallas: Too many moving targets 😆

mturley (approved):
Working as expected on the cluster. When testing this I hit an error related to recent changes to the route-security ([see thread here](https://redhat-internal.slack.com/archives/C07C0447EAV/p1727364581165319)), but that's out of the scope of this PR and I have a followup PR coming to address it. I'll reference this PR and make sure we check that these routes still work as part of that.

PR #3250, opened 2024-09-24
Edit for existing project connection instances

emilys314:
https://issues.redhat.com/browse/RHOAIENG-11556

## Description

Hooks up the edit kebab action to the manageConnectionModal. it will pass in the connection of that row, and pass `isEdit`. It will then prefill connection data. 

The secret stores data in base64, so `atob()` is used to decode, this is done in the modal itself (which is similar to the data connections). This makes it easier to reference a corresponding connection type, to determine how to decode the value. 

If a connection has env var names that do not match up to a connection type field, they will be appended to the bottom as short text fields. 

The multi dropdown is stored as a JSON array, and decoded as such. 

The "Save" button isn't enabled until the field values have been updated. 

image

image

Additionally the k8sNameDesc is now intractable in the preview to mimic the actual form.

image

## How Has This Been Tested?

Tested locally.
Just click the kebab menu on an existing connection row and click "Edit". You can update the values and click Save.

To test non matching values, you modify the env var for the field from the connection type, or edit it directly in openshift

## Test Impact

A cypress test added to check the edit functionally
Jest tests to test the modal in edit form

## Request review criteria:

If you have UI changes: 

@simrandhaliw 

### DIFF
--- a/frontend/src/concepts/connectionTypes/ConnectionTypeForm.tsx
+++ b/frontend/src/concepts/connectionTypes/ConnectionTypeForm.tsx
@@ -30,6 +30,7 @@ type Props = Pick<
   connectionValues?: {
     [key: string]: ConnectionTypeValueType;
   };
+  disableTypeSelection?: boolean;
 };
 
 const ConnectionTypeForm: React.FC<Props> = ({
@@ -42,6 +43,7 @@ const ConnectionTypeForm: React.FC<Props> = ({
   connectionValues,
   onChange,
   onValidate,
+  disableTypeSelection,
 }) => {
   const options: TypeaheadSelectOption[] = React.useMemo(() => {
     if (isPreview && connectionType?.metadata.annotations?.['openshift.io/display-name']) {
@@ -73,7 +75,7 @@ const ConnectionTypeForm: React.FC<Props> = ({
           onSelect={(_, selection) =>
             setConnectionType?.(connectionTypes?.find((c) => c.metadata.name === selection))
           }
-          isDisabled={isPreview || connectionTypes?.length === 1}
+          isDisabled={isPreview || disableTypeSelection}
           placeholder={
             isPreview && !connectionType?.metadata.annotations?.['openshift.io/display-name']
               ? 'Unspecified'
@@ -90,7 +92,7 @@ const ConnectionTypeForm: React.FC<Props> = ({
         )}
       </FormGroup>
       {(isPreview || connectionType?.metadata.name) && (
-        <FormSection title="Connection details" style={{ marginTop: 0 }}>
+        <FormSection title="Connection details">
           <K8sNameDescriptionField
             dataTestId="connection-name-desc"
             nameLabel="Connection name"
@@ -102,7 +104,7 @@ const ConnectionTypeForm: React.FC<Props> = ({
                 k8sName: {
                   value: '',
                   state: {
-                    immutable: true,
+                    immutable: false,
                     invalidCharacters: false,
                     invalidLength: false,
                     maxLength: 0,
@@ -111,7 +113,7 @@ const ConnectionTypeForm: React.FC<Props> = ({
                 },
               }
             }
-            onDataChange={setConnectionNameDesc}
+            onDataChange={setConnectionNameDesc ?? (() => undefined)} // onDataChange needs to be truthy to show resource name
           />
           <ConnectionTypeFormFields
             fields={connectionType?.data?.fields}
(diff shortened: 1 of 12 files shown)


frontend/src/concepts/connectionTypes/ConnectionTypeForm.tsx
```diff
@@ -111,7 +113,7 @@ const ConnectionTypeForm: React.FC<Props> = ({
                 },
               }
             }
-            onDataChange={setConnectionNameDesc}
+            onDataChange={setConnectionNameDesc ?? (() => undefined)} // onDataChange needs to be truthy to show resource name
```
christianvogt, line 116: That's a strange requirement from `K8sNameDescriptionField`. Can we update it?

frontend/src/concepts/connectionTypes/fields/ConnectionTypeFormFields.tsx
```diff
@@ -47,6 +48,22 @@ const ConnectionTypeFormFields: React.FC<Props> = ({
     [fields],
   );
 
+  const unmatchedValues: ConnectionTypeDataField[] = React.useMemo(() => {
+    const unmatched: ConnectionTypeDataField[] = [];
+    for (const key in connectionValues) {
+      const matching = fields?.find((f) => isConnectionTypeDataField(f) && f.envVar === key);
+      if (!matching) {
+        unmatched.push({
+          type: ConnectionTypeFieldType.ShortText,
```
christianvogt, line 57: we _could_ be smart. If it contains `\n` make it long text, or if it is `true|false` make it a boolean. But otherwise I think this works.

frontend/src/concepts/connectionTypes/utils.ts
```diff
@@ -158,3 +163,36 @@ export const assembleConnectionSecret = (
...
+    } else if (matchingField?.type === ConnectionTypeFieldType.Numeric) {
+      response[key] = Number(decodedString);
+    } else if (
+      matchingField?.type === ConnectionTypeFieldType.Dropdown &&
+      matchingField.properties.variant === 'multi'
+    ) {
+      try {
+        response[key] = JSON.parse(decodedString);
+      } catch {
+        response[key] = decodedString;
```
christianvogt, line 190: This is expected to be an array correct? If so, perhaps ensure it parsed correctly and callback to `[<decodedString>]`?

frontend/src/concepts/connectionTypes/ConnectionTypeForm.tsx
emilys314: Okay it's been updated. I'm not 100% certain on all the use cases for the component, but i think if you want to old preview behavior, they can still pass in the `k8sName.state.immutable = false`

frontend/src/concepts/connectionTypes/fields/ConnectionTypeFormFields.tsx
```diff
@@ -74,6 +91,7 @@ const ConnectionTypeFormFields: React.FC<Props> = ({
           <React.Fragment key={i}>{renderDataFields(fieldGroup.fields)}</React.Fragment>
         ),
       )}
+      {unmatchedValues.length > 0 && renderDataFields(unmatchedValues)}
```
christianvogt, line 94: @simrandhaliw when editing a connection instance where the env var has no matching field in the connection type, should we section these off with a header and description which informs the user of their situation?

frontend/src/concepts/connectionTypes/utils.ts
```diff
@@ -158,3 +163,36 @@ export const assembleConnectionSecret = (
...
+    if (matchingField?.type === ConnectionTypeFieldType.Boolean) {
+      response[key] = decodedString === 'true';
+    } else if (matchingField?.type === ConnectionTypeFieldType.Numeric) {
+      response[key] = Number(decodedString);
+    } else if (
+      matchingField?.type === ConnectionTypeFieldType.Dropdown &&
+      matchingField.properties.variant === 'multi'
+    ) {
+      try {
+        response[key] = JSON.parse(decodedString);
```
christianvogt, line 188: Is this overkill what I'm suggesting to ensure that the parsed value conforms to an array of strings?
```suggestion
        const parsed = JSON.parse(decodedString);
        if (Array.isArray(parsed)) {
          response[key] = parsed.map(v => String(v));
        } else {
          response[key] = [decodedString];
        }
```

emilys314:
updates to the connection type select 
image

emilys314:
rebasing to run CI again

PR #3252, opened 2024-09-24
Use a ResourceWatcher for DSC status in the backend instead of fetching it once at app startup to resolve model registries namespace

mturley:
Followup to #3248 and #3239.
Closes https://issues.redhat.com/browse/RHOAIENG-12311

## Description

A potential problem with the implementation in #3239 is that it depends on the registriesNamespace being present in the DSC status immediately when the dashboard pod starts. In theory on a fresh install it is possible for the dashboard to start before the operator has reconciled the DSC and put the registriesNamespace into its status, in which case we would never refetch it and all model registry API calls would fail until the dashboard pods were restarted.

To fix this, we can use a ResourceWatcher so that the server is regularly refetching the DSC and any usage of `getModelRegistryNamespace` will use its latest value. This means that function is no longer async, and also should now be called during each fastify request instead of at app startup time. However, it still throws errors so it should only be called within a `try` block. Any clusters that do not have model registry enabled will have an error thrown by `getModelRegistryNamespace` and must handle that appropriately.

## How Has This Been Tested?

Tested in modelregistry-ui cluster - we can still access and interact with both the admin-side MR settings page and the user-side Model Registry page.

## Test Impact

N/A, no unit tests on the backend :(

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/backend/src/routes/api/modelRegistries/modelRegistryUtils.ts
+++ b/backend/src/routes/api/modelRegistries/modelRegistryUtils.ts
@@ -1,23 +1,16 @@
 import { KubeFastifyInstance, ModelRegistryKind, RecursivePartial } from '../../../types';
 import { PatchUtils, V1Secret, V1Status } from '@kubernetes/client-node';
-import { getClusterStatus } from '../../../utils/dsc';
+import { getClusterStatus } from '../../../utils/resourceUtils';
 
 const MODEL_REGISTRY_API_GROUP = 'modelregistry.opendatahub.io';
 const MODEL_REGISTRY_API_VERSION = 'v1alpha1';
 const MODEL_REGISTRY_PLURAL = 'modelregistries';
 
-export const getModelRegistryNamespace = async (
-  fastify: KubeFastifyInstance,
-): Promise<string | undefined> => {
-  let registriesNamespace: string | undefined;
-  try {
-    const clusterStatus = await getClusterStatus(fastify);
-    registriesNamespace = clusterStatus.components?.modelregistry?.registriesNamespace;
-  } catch (e) {
-    fastify.log.error(e, 'Failed to fetch DSC status');
-  }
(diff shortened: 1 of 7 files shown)


backend/src/utils/route-security.ts
andrewballantyne, line 1: Woah, not sure you should be messing in this area. What is the purpose of any usage here?

backend/src/routes/api/modelRegistries/modelRegistryUtils.ts
```diff
@@ -1,23 +1,16 @@
...
-    registriesNamespace = clusterStatus.components?.modelregistry?.registriesNamespace;
-  } catch (e) {
-    fastify.log.error(e, 'Failed to fetch DSC status');
-  }
+export const getModelRegistryNamespace = (fastify: KubeFastifyInstance): string => {
+  const clusterStatus = getClusterStatus(fastify);
+  const registriesNamespace = clusterStatus.components?.modelregistry?.registriesNamespace;
   if (!registriesNamespace) {
-    fastify.log.warn('Model registry namespace not found in DSC status');
+    throw new Error('Model registry namespace not found in DSC status');
```
andrewballantyne, line 13: I imagine this is not a problem in the `preHandler` of a fastify route 🤔

andrewballantyne: Yeah, it'll do normal things and just fail the request
```
{"statusCode":500,"error":"Internal Server Error","message":"Model registry namespace not found in DSC status"}
```

backend/src/utils/route-security.ts
mturley: For posterity - [discussed on slack here](https://redhat-internal.slack.com/archives/C07C0447EAV/p1727281666525129) and we determined we don't need the model registry namespace in any logic in this file so this PR now removes it.

andrewballantyne (approved):

PR #3257, opened 2024-09-25
Use K8sNameDescriptionField in connection type page

emilys314:
Closes https://issues.redhat.com/browse/RHOAIENG-13400

## Description

Use the new `K8sNameDescriptionField` component. This prevents the user from entering a too long of a name which causes a resource name error when saving. 

image
image
image
image

## How Has This Been Tested?

Tested locally by running through the different connection type operations. 
For the bug referenced in the jira issue, paste a lot of text https://www.lipsum.com/feed/html and click save. It will now truncate the resource name

## Test Impact

No updates, existing tests cover the previous functionality. (and the long name is a server side error)

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/concepts/k8s/utils.ts
+++ b/frontend/src/concepts/k8s/utils.ts
@@ -7,7 +7,8 @@ export const PreInstalledName = 'Pre-installed';
 export const ownedByDSC = (resource: K8sResourceCommon): boolean =>
   !!resource.metadata?.ownerReferences?.find((owner) => owner.kind === 'DataScienceCluster');
 
-export const isK8sDSGResource = (x?: K8sResourceCommon): x is K8sDSGResource => !!x?.metadata?.name;
+export const isK8sDSGResource = (x?: K8sResourceCommon): x is K8sDSGResource =>
+  x?.metadata?.name != null;
 export const getDisplayNameFromK8sResource = (resource: K8sDSGResource): string =>
   resource.metadata.annotations?.['openshift.io/display-name'] || resource.metadata.name;
 export const getResourceNameFromK8sResource = (resource: K8sDSGResource): string =>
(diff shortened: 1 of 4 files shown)


frontend/src/concepts/k8s/utils.ts
```diff
@@ -7,7 +7,8 @@ export const PreInstalledName = 'Pre-installed';
 export const ownedByDSC = (resource: K8sResourceCommon): boolean =>
   !!resource.metadata?.ownerReferences?.find((owner) => owner.kind === 'DataScienceCluster');
 
-export const isK8sDSGResource = (x?: K8sResourceCommon): x is K8sDSGResource => !!x?.metadata?.name;
+export const isK8sDSGResource = (x?: K8sResourceCommon): x is K8sDSGResource =>
+  !!x?.metadata && 'name' in x.metadata;
```
christianvogt, line 11: Why the change?

christianvogt (commented):
I can still submit the form if there are errors with the resource name.

frontend/src/concepts/k8s/utils.ts
emilys314: For prefilling the k8sNameDesc fields, it runs the following:
```
export const setupDefaults = ({
  initialData,
  limitNameResourceType,
  safePrefix,
}: UseK8sNameDescriptionDataConfiguration): K8sNameDescriptionFieldData => {
  let initialName = '';
  let initialDescription = '';
  let initialK8sNameValue = '';
  let configuredMaxLength = MAX_RESOURCE_NAME_LENGTH;

  if (isK8sDSGResource(initialData)) {
    initialName = getDisplayNameFromK8sResource(initialData);
    initialDescription = getDescriptionFromK8sResource(initialData);
    initialK8sNameValue = initialData.metadata.name;
  }
  if (limitNameResourceType != null) {
    configuredMaxLength = ROUTE_BASED_NAME_LENGTH;
  }
  ```
  
So specifically for the duplicate scenario, there will be data like the description and display name filled out, but no resource name is defined yet, and it will fail the check since "" is false. This change makes "" still count as true

emilys314: Okay I reverted this change due to other changes anyways. For duplicate, it will now pass in the resource name, so the type check will always pass.

Passing in the resource name would originally block you from editing it, but i added `editK8sNameValue` to allow edit of the prefilled value

PR #3262, opened 2024-09-26
[RHOAIENG-11326] Creating new pipeline from Create Run Page redirects to the pipeline detail page

jpuzz0:
https://issues.redhat.com/browse/RHOAIENG-11326

## Description
Add condition to pipeline import modal to only redirect when not located in the `/create` path associated with runs/schedules

**Demo:**
https://github.com/user-attachments/assets/28f10956-9b84-465e-8db6-bf9fda2d5f6d

## How Has This Been Tested?
1. Import a pipeline from the Pipelines list page and verify you're redirected to the pipeline version details page after creation
2. Import a pipeline from the Create run page and verify you are not redirected to the pipeline version details page
3. Repeat step 2 for the Create schedule page

## Test Impact
No new tests added

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/concepts/pipelines/content/createRun/contentSections/PipelineSection.tsx
+++ b/frontend/src/concepts/pipelines/content/createRun/contentSections/PipelineSection.tsx
@@ -60,6 +60,7 @@ const PipelineSection: React.FC<PipelineSectionProps> = ({
               variant="link"
               icon={<PlusCircleIcon />}
               onCreate={onPipelineChange}
+              redirectAfterImport={false}
             >
               Create new pipeline
             </ImportPipelineButton>
--- a/frontend/src/concepts/pipelines/content/import/ImportPipelineButton.tsx
+++ b/frontend/src/concepts/pipelines/content/import/ImportPipelineButton.tsx
@@ -6,11 +6,13 @@ import { PipelineKFv2 } from '~/concepts/pipelines/kfTypes';
 
 type ImportPipelineButtonProps = {
   onCreate?: (pipeline: PipelineKFv2) => void;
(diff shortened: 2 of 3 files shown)


DaoDaoNoCode (commented):
This is a nice solution, but I do have another suggestion. I feel we could avoid inferring from the pathname, instead, we can pass `isRedirectingToDetailsPage` or something like this into the modal, and only when it's set to true, navigate to the details page... And we can default it to true and set it only to false when importing on the create run/schedules page...
Matching with the pathname is not always secure because if we change the pathname in the future, it's hard to find the issue here. This is OK for now, but just a suggestion.

jpuzz0:
> This is a nice solution, but I do have another suggestion. I feel we could avoid inferring from the pathname, instead, we can pass `isRedirectingToDetailsPage` or something like this into the modal, and only when it's set to true, navigate to the details page... And we can default it to true and set it only to false when importing on the create run/schedules page... Matching with the pathname is not always secure because if we change the pathname in the future, it's hard to find the issue here. This is OK for now, but just a suggestion.

That's fair, I don't mind either way, so I made the change to add a prop instead.

DaoDaoNoCode (approved):
Tested, works well
/lgtm

Gkrumbach07:
@jpuzz0 we also should prevent navigation when importing from the empty states

jpuzz0:
> @jpuzz0 we also should prevent navigation when importing from the empty states bc it is pretty much the same page anyway

So inside of `GlobalNoPipelines`, prevent the redirect there? I'm not sure how this would be different from the redirect that we want from the pipelines table action. Maybe I'm missing something here though.

PR #3263, opened 2024-09-26
MR fixes identified during architecture review

mturley:
Followup to https://github.com/opendatahub-io/odh-dashboard/pull/3252 and https://github.com/opendatahub-io/odh-dashboard/pull/3249 as well as some misc discussion points from arch review (https://issues.redhat.com/browse/RHOAIENG-13439).

## Description

* After #3249 was rebased on the changes from #3252, testing it revealed that because we removed the model registry namespace from the checks in `backend/src/utils/route-security.ts` and rolebindings we create in MR settings include that namespace in the request body, creation of rolebindings failed due to the namespace checking in `requestSecurityGuard`. We do not want to modify route-security because it is a fragile workaround and modifying it involves high risk. The fix instead is to simply not include a namespace in the rolebinding object when we send it to the backend and instead inject it just before calling the cluster API.
* When deleting a registry in the backend, we were not returning the promise from the `deleteModelRegistryAndSecret` util which means we weren't catching errors that came up in this deletion.
* There was a TODO comment asking if we needed some code in our error handler util. That code is also present in `frontend/src/api/pipelines/errorUtils.ts`, and it is needed because some requests in MR throw a NotReadyError which should be ignored by this handler. Removed the TODO comment.

## How Has This Been Tested?

Tested on modelregistry-ui cluster, we can now create rolebindings in MR settings again

## Test Impact

N/A, no tests in the backend

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/backend/src/routes/api/modelRegistryRoleBindings/modelRegistryRolebindingsUtils.ts
+++ b/backend/src/routes/api/modelRegistryRoleBindings/modelRegistryRolebindingsUtils.ts
@@ -27,12 +27,19 @@ export const createModelRegistryRoleBinding = async (
   rbRequest: V1RoleBinding,
   mrNamespace: string,
 ): Promise<V1RoleBinding> => {
+  // Re-inject the namespace value that was omitted by the client
+  //   (see createModelRegistryRoleBinding in frontend/src/services/modelRegistrySettingsService.ts)
+  // This will be unnecessary when we remove the backend service as part of https://issues.redhat.com/browse/RHOAIENG-12077
+  const roleBindingWithNamespace = {
+    ...rbRequest,
+    metadata: { ...rbRequest.metadata, namespace: mrNamespace },
+  };
   const response = await (fastify.kube.customObjectsApi.createNamespacedCustomObject(
     MODEL_REGISTRY_ROLE_BINDING_API_GROUP,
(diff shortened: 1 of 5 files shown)


backend/src/routes/api/modelRegistryRoleBindings/modelRegistryRolebindingsUtils.ts
```diff
@@ -27,12 +27,19 @@ export const createModelRegistryRoleBinding = async (
   rbRequest: V1RoleBinding,
   mrNamespace: string,
```
andrewballantyne, line 28: why are you passing around this value? isn't it practically free from the config?

andrewballantyne (commented):
This is more-or-less what I expected. Let me test it out

andrewballantyne (approved):
Lgtm

backend/src/routes/api/modelRegistryRoleBindings/modelRegistryRolebindingsUtils.ts
mturley: This is a holdover from when we used to rely on getting the namespace from the client. You're right, we should probably not pass it around... although for this release I'm hesitant to refactor it because the helper to get that namespace can throw errors and we'll need to double check the error handling around these utils. I'll include it in my tech debt follow up issues.

PR #3271, opened 2024-09-30
Search - 'Clear all Filters' doesn't work

YuliaKrimerman:
https://issues.redhat.com/browse/RHOAIENG-12861

## Description

1. Clear all filters button is now working on all existing screens 
2.  Changed label that was mentioned in RHOAI sprint demos.

## How Has This Been Tested?

on the UI

## Test Impact

Updated tests to pass after the label change. In order to test the change, verify that now "Clear all Filters' button is now working everywhere it exists 

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/pages/modelRegistry/screens/RegisteredModels/RegisteredModelListView.tsx
+++ b/frontend/src/pages/modelRegistry/screens/RegisteredModels/RegisteredModelListView.tsx
@@ -105,12 +105,19 @@ const RegisteredModelListView: React.FC<RegisteredModelListViewProps> = ({
     </ToolbarGroup>
   );
 
+  const clearAllFilters = resetFilters;
+
   return (
     <RegisteredModelTable
       refresh={refresh}
-      clearFilters={resetFilters}
+      clearFilters={clearAllFilters}
       registeredModels={filteredRegisteredModels}
-      toolbarContent={<RegisteredModelsTableToolbar toggleGroupItems={toggleGroupItems} />}
+      toolbarContent={
+        <RegisteredModelsTableToolbar
+          toggleGroupItems={toggleGroupItems}
+          onClearAllFilters={clearAllFilters}
+        />
+      }
     />
   );
 };
--- a/frontend/src/__tests__/cypress/cypress/tests/mocked/projects/clusterStorage.cy.ts
+++ b/frontend/src/__tests__/cypress/cypress/tests/mocked/projects/clusterStorage.cy.ts
@@ -155,14 +155,14 @@ describe('ClusterStorage', () => {
     addClusterStorageModal.findMountField().fill('test2');
     addClusterStorageModal
       .findMountFieldHelperText()
-      .should('have.text', 'Must only consist of lower case letters and dashes.');
+      .should('have.text', 'Must only consist of lowercase letters and dashes.');
 
     // Allow trailing slash
     addClusterStorageModal.findMountField().clear();
     addClusterStorageModal.findMountField().fill('test/');
     addClusterStorageModal
       .findMountFieldHelperText()
-      .should('have.text', 'Must consist of lower case letters and dashes.');
+      .should('have.text', 'Must consist of lowercase letters and dashes.');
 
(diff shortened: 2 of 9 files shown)


ppadti:
@YuliaKrimerman I think we need to fix this issue in Deployed models page as well.

YuliaKrimerman:
@ppadti I was trying to add that fix into InferenceServiceTable inside the Table, but even though it has emptyTableView that whole chip group is not showing, what am I missing?

manaswinidas (commented):
@ppadti I don't think we have a toolbar with filters for the deployments tab here as per the [mocks](https://www.figma.com/proto/KTsYhANYQqB0NUrXW7wQH1/Model-Registry?page-id=6782%3A296930&node-id=8238-700071&viewport=1846%2C28%2C0.43&t=anh0eGvvgwFUHiMb-1&scaling=min-zoom&content-scaling=fixed&starting-point-node-id=8238%3A699060&show-proto-sidebar=1). Let me know if you're referring to some other mock links.

frontend/src/pages/modelRegistry/screens/RegisteredModels/RegisteredModelListView.tsx
```diff
@@ -105,12 +105,19 @@ const RegisteredModelListView: React.FC<RegisteredModelListViewProps> = ({
     </ToolbarGroup>
   );
 
+  const clearAllFilters = resetFilters;
```
manaswinidas, line 108: I don't think there is a need to create a new variable when we can just use `resetFilters` everywhere. Also, I see `() => setSearch('')` present in the same file numerous times when `resetFilters` function is doing the same thing. Can we replace all of them with `resetFilters`? Otherwise, we can rename `resetFilters` to `clearAllFilters` but all `() => setSearch('')` should be replaced with that function.

manaswinidas (changes requested):

ppadti:
> @ppadti I don't think we have a toolbar with filters for the deployments tab here as per the [mocks](https://www.figma.com/proto/KTsYhANYQqB0NUrXW7wQH1/Model-Registry?page-id=6782%3A296930&node-id=8238-700071&viewport=1846%2C28%2C0.43&t=anh0eGvvgwFUHiMb-1&scaling=min-zoom&content-scaling=fixed&starting-point-node-id=8238%3A699060&show-proto-sidebar=1). Let me know if you're referring to some other mock links.

Yeah, I was referring to this [slack conversation](https://redhat-internal.slack.com/archives/C07C0447EAV/p1727361566973249) here.

YuliaKrimerman:
@ppadti , @manaswinidas  As I mentioned above I was trying to add that fix into InferenceServiceTable(it's the right file correct?) inside the Table, but even though it has emptyTableView that whole chip group is not showing, what am I missing? Any suggestions? And yes, I suggested to fix it as part of that PR in that conversation, but just got stuck, and preferred to, at least put the required solution for the actual bug

ppadti:
> @ppadti , @manaswinidas As I mentioned above I was trying to add that fix into InferenceServiceTable(it's the right file correct?) inside the Table, but even though it has emptyTableView that whole chip group is not showing, what am I missing? Any suggestions? And yes, I suggested to fix it as part of that PR in that conversation, but just got stuck, and preferred to, at least put the required solution for the actual bug

May be because it is using DashboardSearchField and not ToolbarFilter and we have this issue in some other places as well for ex: Accelerator profile and Notebook images page. I think we need to discuss this with the team and can have a separate issue to address this. wdyt?

YuliaKrimerman:
Sounds good to me, I can open an issue for it, and will address the related PR comments then.

YuliaKrimerman:
@manaswinidas Addressed comment, ready for re-review. Also created a story for us to discuss https://issues.redhat.com/browse/RHOAIENG-13924 . @ppadti . Thank you!

ppadti:
tested locally, clear all filters works well.
/lgtm

mturley (approved):
LGTM, thanks @YuliaKrimerman

PR #3274, opened 2024-09-30
update useGenericObjectState types

christianvogt:
https://issues.redhat.com/browse/RHOAIENG-9748

## Description

Improves the typing of `useGenericObjectState` such that the value is properly typed depending on the supplied property key. As a result, any unmatched value will result in a typescript error.

Extracted common types for model serving into a separate type definition to be used by common form components.

Additionally, updated `useGenericObjectState` to accept a function to coincide with the API of `useState`. This function argument can be used when the computation of default data is expensive.

## How Has This Been Tested?

`npm run test:type-check`
Ran existing tests.

## Test Impact

N/A

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/pages/projects/types.ts
+++ b/frontend/src/pages/projects/types.ts
@@ -9,14 +9,16 @@ import {
   Volume,
   VolumeMount,
 } from '~/types';
-import { ValueOf } from '~/typeHelpers';
 import { AWSSecretKind } from '~/k8sTypes';
 import { AcceleratorProfileState } from '~/utilities/useAcceleratorProfileState';
 import { AcceleratorProfileSelectFieldState } from '~/pages/notebookController/screens/server/AcceleratorProfileSelectField';
 import { K8sNameDescriptionFieldData } from '~/concepts/k8s/K8sNameDescriptionField/types';
 import { AwsKeys } from './dataConnections/const';
 
-export type UpdateObjectAtPropAndValue<T> = (propKey: keyof T, propValue: ValueOf<T>) => void;
+export type UpdateObjectAtPropAndValue<T> = <K extends keyof T>(
+  propKey: K,
+  propValue: T[K],
+) => void;
 
 export type NameDescType = {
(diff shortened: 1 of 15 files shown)


frontend/src/pages/projects/types.ts
```diff
@@ -9,14 +9,16 @@ import {
...
 import { AWSSecretKind } from '~/k8sTypes';
 import { AcceleratorProfileState } from '~/utilities/useAcceleratorProfileState';
 import { AcceleratorProfileSelectFieldState } from '~/pages/notebookController/screens/server/AcceleratorProfileSelectField';
 import { K8sNameDescriptionFieldData } from '~/concepts/k8s/K8sNameDescriptionField/types';
 import { AwsKeys } from './dataConnections/const';
 
-export type UpdateObjectAtPropAndValue<T> = (propKey: keyof T, propValue: ValueOf<T>) => void;
+export type UpdateObjectAtPropAndValue<T> = <K extends keyof T>(
+  propKey: K,
+  propValue: Partial<T>[K],
```
andrewballantyne, line 20: Won't this break the sanctity of our type behind the object store?

 I can send `undefined` to a value required by the type.

andrewballantyne (changes requested):

frontend/src/pages/projects/types.ts
christianvogt: The current types are already broken in that I can supply a value from _any_ prop for _any_ key.
This isn't a perfect solution but a step that's better IMO. Enforces correct type but is allowing `undefined`.

I suppose we can always keep looking for the perfect solution...

christianvogt: Went with another approach that required altering several components which utilize this type. I was not able to find a solution whereby only the `UpdateObjectAtPropAndValue` could be updated and solve all use cases.

jeff-phillips-18 (approved):

PR #3276, opened 2024-10-01
Config map nim model name fetch

olavtar:
## Description

Retrieving ConfigMap using service account and added ${uid} to PVC name to avoid waiting for resource termination after deletion

## How Has This Been Tested?

Tested locally.

## Test Impact

Deployed a NIM model, deleted it, and deployed another model while the PVC was still in the 'Terminating' state. 

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/api/k8s/servingRuntimes.ts
+++ b/frontend/src/api/k8s/servingRuntimes.ts
@@ -35,6 +35,7 @@ export const assembleServingRuntime = (
   initialAcceleratorProfile?: AcceleratorProfileState,
   selectedAcceleratorProfile?: AcceleratorProfileSelectFieldState,
   isModelMesh?: boolean,
+  nimPVCName?: string,
 ): ServingRuntimeKind => {
   const {
     name: displayName,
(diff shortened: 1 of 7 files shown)


frontend/src/api/k8s/servingRuntimes.ts
```diff
@@ -171,8 +181,33 @@ export const assembleServingRuntime = (
     volumes.push(getshmVolume('2Gi'));
   }
 
-  updatedServingRuntime.spec.volumes = volumes;
+  if (nimPVCName) {
```
andrewballantyne, line 184: This code should never have been here -- you folks need to work on lowering your points of contact when fixing a blocker fix. You snuck this PVC thing into a high trafficked area for 2 other flows you're not touching instead of just adding this to your serving runtime template before ever calling this flow.

This change has already been tested and I'm going to let it in. But please do not sneak more "nice to have" code in when you need to fix critical blocking issues.

frontend/src/api/k8s/servingRuntimes.ts
```diff
@@ -242,6 +277,7 @@ export const updateServingRuntime = (options: {
   initialAcceleratorProfile?: AcceleratorProfileState;
   selectedAcceleratorProfile?: AcceleratorProfileSelectFieldState;
   isModelMesh?: boolean;
+  nimPVCName?: string;
```
andrewballantyne, line 280: This was unnecessary -- you don't support edit, updating the serving runtime is not done currently.

andrewballantyne (approved):
None of these are critical blockers. Your dodge a typescript issue because Secrets & Configmaps are very close in structure. 

I'm going to approve this to start updating downstream. Please fire up another PR and fix this in main asap.

PR #3281, opened 2024-10-01
Fix: Cypress Pipeline e2e test

FedeAlonso:
https://issues.redhat.com/browse/RHOAIENG-13384

## Description
After some changes in the DSProjects view search toolbar, the Pipelines  e2e test was failing.
Also I've updated the URL used to import the pipeline.

## How Has This Been Tested?
cypress/job/dashboard-tests/18
image


manosnoam (approved):
Thanks Fede!

PR #3286, opened 2024-10-01
Project Details Page - Allow Edit and View Details

jeff-phillips-18:
Closes [RHOAIENG-8571](https://issues.redhat.com/browse/RHOAIENG-8571)

## Description
In the project details page, add a popover to show users the project's resource name and kind.
Also, add actions to allow a provisioned user to edit or delete the project.

## How Has This Been Tested?
As a provisioned user, navigate to a project page.
Verify there is a popover indicator next to the project name.
Clicking the indicator shows a popover that shows the project's resource name and show `Project` for the resource type.
Click on the actions menu to the far right, verify both `Edit project` and `Delete project` actions are shown.
Verify that `Edit project` bring up the `Edit project` modal.
Verify that `Delete project` brings up the `Delete project` modal.

As a non-provisioning user, navigate to a project page.
Verify there is a popover indicator next to the project name.
Clicking the indicator shows a popover that shows the project's resource name and show `Project` for the resource type.
Verify there is no actions menu shown to the far right.

## Test Impact
Added cypress e2e tests.

## Mocks
Ref: [Figma prototype](https://www.figma.com/proto/ANpg8c6Ltq2sCVNCAyXzOR/Project-Enhancements?page-id=355%3A32349&node-id=355-32350&viewport=334%2C247%2C0.34&t=cu3i51XnVnoIzTvW-1&scaling=scale-down-width&content-scaling=fixed&starting-point-node-id=355%3A32350)

image

image

## Screen shots
image

image

## Request review criteria:

If you have UI changes: 

/cc  @jgiardino

### DIFF
--- a/frontend/src/pages/projects/screens/detail/ProjectActions.tsx
+++ b/frontend/src/pages/projects/screens/detail/ProjectActions.tsx
@@ -1,0 +1,100 @@
+import * as React from 'react';
+import { Dropdown, DropdownItem, MenuToggle, DropdownList } from '@patternfly/react-core';
+import { useNavigate } from 'react-router-dom';
+import { getDashboardMainContainer } from '~/utilities/utils';
+import { AccessReviewResourceAttributes, ProjectKind } from '~/k8sTypes';
+import { useAccessReview } from '~/api';
+import DeleteProjectModal from '~/pages/projects/screens/projects/DeleteProjectModal';
+import ManageProjectModal from '~/pages/projects/screens/projects/ManageProjectModal';
+
+const projectEditAccessReview: AccessReviewResourceAttributes = {
+  group: 'project.openshift.io',
+  resource: 'projectrequests',
+  verb: 'update',
+};
+
+const projectDeleteAccessReview: AccessReviewResourceAttributes = {
+  group: 'project.openshift.io',
+  resource: 'projectrequests',
+  verb: 'delete',
+};
+
+type Props = {
+  project: ProjectKind;
+};
+
+const ProjectActions: React.FC<Props> = ({ project }) => {
+  const navigate = useNavigate();
+  const [canEdit, editRbacLoaded] = useAccessReview({
+    ...projectEditAccessReview,
+    namespace: project.metadata.name,
+  });
+  const [canDelete, deleteRbacLoaded] = useAccessReview({
+    ...projectDeleteAccessReview,
+    namespace: project.metadata.name,
+  });
+  const [open, setOpen] = React.useState(false);
+  const [deleteOpen, setDeleteOpen] = React.useState(false);
+  const [editOpen, setEditOpen] = React.useState(false);
+
+  if (!editRbacLoaded || !deleteRbacLoaded || (!canEdit && !canDelete)) {
+    return null;
+  }
+
+  const DropdownComponent = (
+    <Dropdown
+      onOpenChange={setOpen}
+      onSelect={() => setOpen(false)}
+      toggle={(toggleRef) => (
+        <MenuToggle
+          aria-label="Actions"
+          data-testid="project-actions"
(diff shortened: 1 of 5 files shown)


frontend/src/pages/projects/screens/detail/ProjectActions.tsx
```diff
@@ -0,0 +1,98 @@
...
+            setOpen(!open);
+          }}
+        >
+          Actions
+        </MenuToggle>
+      )}
+      isOpen={open}
+      popperProps={{ position: 'right', appendTo: getDashboardMainContainer() }}
+    >
+      <DropdownList>
```
jenny-s51, line 64: Based on the mocks, should there be a divider between these "Edit" and "Delete" dropdown items?

jenny-s51 (commented):
Updates look great @jeff-phillips-18 . Tested both self-provisioned and admin users, edit and view details working as expected. Left a comment below.

frontend/src/pages/projects/screens/detail/ProjectActions.tsx
jgiardino: @jenny-s51 - We don't have design guidelines for action menus yet. When I look at other areas of the product, there are pages in the dashboard that consistently display a divider above Delete. But there are places where we have just Edit and Delete in a menu without a divider (Workbenches and Permissions list items in the Project details page). 

The only guidelines I see on this topic within PF are under [Menu > Red text menu](https://www.patternfly.org/components/menus/menu/design-guidelines#red-text-menu):
> A divider should be used to separate the destructive menu items from the non-destructive items.

We aren't using this variant on the Delete menu action (and maybe we should consider that when we draft guidelines), but the point about separating destructive menu items from the non-destructive menu items is the piece that I think is relevant here. I think this recommendation should apply regardless of how many menu items there are. So I am in favor of implementing the UI as designed in this case.

jeff-phillips-18: @jenny-s51 @jgiardino Looking around the UI, I see:

- Connections (recently added) does not have the separator.
- Data connections does not have the separator
- Cluster storage does not have the separator
- Permissions (both user and group) do not have the separator
- Workbenches does not have the separator
- Pipeline servers does have the separator
- Pipeline versions does have the separator

I can add it here but it would be mostly inconsistent. I would suggest UX file an issue to make all of the menus consistent and we address all of them at once.

simrandhaliw: @jeff-phillips-18 @jenny-s51 @jgiardino I agree with Jeff. I’d prefer to create a separate issue once we have the menu guidelines ready to update the styling for all menus at once and ensure consistency. For now, I’d vote to drop the divider from the implementation of the Project Details action.

jenny-s51: Thank you for your clarity here @jgiardino. Filing a separate ticket to make all of these menus consistent sounds like a good plan @jeff-phillips-18 @simrandhaliw.

jenny-s51 (approved):

frontend/src/pages/projects/screens/detail/ProjectActions.tsx
```diff
@@ -0,0 +1,98 @@
...
+      {editOpen ? (
+        <ManageProjectModal editProjectData={project} onClose={() => setEditOpen(false)} />
+      ) : null}
+      {deleteOpen ? (
+        <DeleteProjectModal
+          deleteData={project}
+          onClose={() => {
+            setDeleteOpen(false);
+            navigate('/projects');
+          }}
```
christianvogt, line 91: Only navigate to the projects page if the project was actually deleted.

PR #3288, opened 2024-10-01
Provide the functionality to search models and model version by "labels"

YuliaKrimerman:
https://issues.redhat.com/browse/RHOAIENG-12589

## Description

 
When using the Keyword filtering option, it will now query the labels as well. This should be in both models and versions

## How Has This Been Tested?

on The UI  
Screenshot 2024-10-01 at 3 45 38 PM

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/pages/modelRegistry/screens/utils.ts
+++ b/frontend/src/pages/modelRegistry/screens/utils.ts
@@ -88,13 +88,12 @@ export const filterModelVersions = (
     }
 
     switch (searchType) {
-      case SearchType.KEYWORD: {
+      case SearchType.KEYWORD:
         return (
           mv.name.toLowerCase().includes(searchLower) ||
           (mv.description && mv.description.toLowerCase().includes(searchLower)) ||
           getLabels(mv.customProperties).some((label) => label.toLowerCase().includes(searchLower))
         );
-      }
 
       case SearchType.AUTHOR: {
         return mv.author && mv.author.toLowerCase().includes(searchLower);
--- a/frontend/src/pages/modelRegistry/screens/utils.ts
+++ b/frontend/src/pages/modelRegistry/screens/utils.ts
@@ -89,29 +89,15 @@ export const filterModelVersions = (
 
     switch (searchType) {
       case SearchType.KEYWORD: {
-        const nameMatch = mv.name.toLowerCase().includes(searchLower);
-        const descriptionMatch =
-          mv.description && mv.description.toLowerCase().includes(searchLower);
-
-        let labelMatch = false;
-        labelMatch = Object.keys(mv.customProperties).some((key) => {
-          const prop = mv.customProperties[key];
-          return (
-            key.toLowerCase().includes(searchLower) ||
-            (prop.metadataType === ModelRegistryMetadataType.STRING &&
-              prop.string_value.toLowerCase().includes(searchLower))
-          );
-        });
-
-        return nameMatch || descriptionMatch || labelMatch;
+        return (
+          mv.name.toLowerCase().includes(searchLower) ||
+          (mv.description && mv.description.toLowerCase().includes(searchLower)) ||
+          getLabels(mv.customProperties).some((label) => label.toLowerCase().includes(searchLower))
+        );
       }
 
       case SearchType.AUTHOR: {
-        return (
-          mv.author &&
-          (mv.author.toLowerCase().includes(search.toLowerCase()) ||
-            (mv.author && mv.author.toLowerCase().includes(search.toLowerCase())))
-        );
+        return mv.author && mv.author.toLowerCase().includes(searchLower);
       }
 
       default:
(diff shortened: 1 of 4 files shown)


frontend/src/pages/modelRegistry/screens/utils.ts
```diff
@@ -79,30 +79,46 @@ export const filterModelVersions = (
...
-          (mv.description && mv.description.toLowerCase().includes(search.toLowerCase()))
-        );
+      case SearchType.KEYWORD: {
+        const nameMatch = mv.name.toLowerCase().includes(searchLower);
+        const descriptionMatch =
+          mv.description && mv.description.toLowerCase().includes(searchLower);
 
-      case SearchType.AUTHOR:
+        let labelMatch = false;
+        labelMatch = Object.keys(mv.customProperties).some((key) => {
```
gitdallas, line 97: this will get all customProperties which include properties, so although you don't see properties on the list view, you'll be able to filter on them which i don't think we want.

you could filter out anything where the `value.string_value` is truthy, but you may be better off using the `getLabels` function above which will return a string array... so something like

`getLabels(mv.customProperties).some( label => label.toLowerCase().includes(searchLower)`

also, rather than creating 3 vars here, you could just check them in a way that would shortcircuit if it finds a match.. like

        return (
          mv.name.toLowerCase().includes(searchLower) ||
          (mv.description && mv.description.toLowerCase().includes(searchLower)) ||
          getLabels(mv.customProperties).some( label => label.toLowerCase().includes(searchLower)
        );

i also wonder if parts of these two (filterModelVersions and filterRegisteredModels) can be factored out a bit. i think the only difference is author vs owner... so you could at least factor out the keyword search and accept `ModelVersion | RegisteredModel` or even potentially just have a more generic filter function that would take both and include both the `SearchType.OWNER` and `SearchType.AUTHOR` cases.

gitdallas (changes requested):
one request (making sure that properties aren't included in the search) and a couple other suggestions.

gitdallas:
also could add tests. there are already search tests (for example, modelVersions.cy.ts has tests "filtering by keyword". you'll need to add the labels to the mock and search on those

YuliaKrimerman:
@gitdallas Got it, thank you. Addressed the comments, ready for re-review.

gitdallas (approved):

frontend/src/pages/modelRegistry/screens/utils.ts
```diff
@@ -79,30 +79,32 @@ export const filterModelVersions = (
...
+  const searchLower = search.toLowerCase();
+
+  return unfilteredModelVersions.filter((mv: ModelVersion) => {
     if (!search) {
       return true;
     }
 
     switch (searchType) {
-      case SearchType.KEYWORD:
+      case SearchType.KEYWORD: {
```
andrewballantyne, line 91: Small nit -- you only need `{` braces around a case when creating a variable inside it to scope it lexically. Other than that it's not needed.

frontend/src/pages/modelRegistry/screens/utils.ts
```diff
@@ -115,23 +117,29 @@ export const filterRegisteredModels = (
...
+          rm.name.toLowerCase().includes(searchLower) ||
+          (rm.description && rm.description.toLowerCase().includes(searchLower)) ||
+          getLabels(rm.customProperties).some((label) => label.toLowerCase().includes(searchLower))
         );
+      }
 
-      case SearchType.OWNER:
-        return rm.owner && rm.owner.toLowerCase().includes(search.toLowerCase());
+      case SearchType.OWNER: {
+        return rm.owner && rm.owner.toLowerCase().includes(searchLower);
```
andrewballantyne, line 138: image

Seems you're missing a test?

andrewballantyne (dismissed):

frontend/src/pages/modelRegistry/screens/utils.ts
YuliaKrimerman: I think I added _because_ it was throwing a lexical error when I ran npm run test

frontend/src/pages/modelRegistry/screens/utils.ts
YuliaKrimerman: It's interesting but it's probably missing a test even before that because I applied a small change to those lines( and not really added them from scratch). Now I'm trying to cover owner search, but we don't have owner anywhere in our types, only in author. So I wonder if this functionality is even valid in the first place? @andrewballantyne

andrewballantyne: haha -- interesting -- I know very little about MR, so if the code is never run -- we probably should trim it?

gitdallas: @YuliaKrimerman registered models have owners while model versions have authors.

gitdallas: i offered a suggestion about merging or filtering out some of the common code between the two at the end of this comment  https://github.com/opendatahub-io/odh-dashboard/pull/3288#discussion_r1785535731

frontend/src/pages/modelRegistry/screens/utils.ts
mturley: @YuliaKrimerman I think we should consider doing this refactor @gitdallas is describing where we combine `filterRegisteredModels` and `filterModelVersions` using generics, but we can think about doing that in a followup PR since this one is blocking your other branch. I'll include that on a list of small tech debt items I plan to open shortly (there are some other things that came from the arch review meeting with @andrewballantyne).

frontend/src/pages/modelRegistry/screens/utils.ts
mturley: The versions tests were added, and I replied on that other comment, I think we can follow up on that refactor in another PR

mturley (approved):
LGTM, thanks @YuliaKrimerman

PR #3289, opened 2024-10-02
test: Storage Classes e2e cypress tests

FedeAlonso:
https://issues.redhat.com/browse/RHOAIENG-11613

## Description
e2e Cypress test cases for Storage Class Feature

Settings->Storage class view:
- A non admin user can not acccess to Settings -> Storage classes view
- The Default label is shown in the grid
- An admin user can enable a disabled Storage Class
- An admin user can disable an enabled Storage Class
- An admin user can set an enabled Storage Class as the default one
- An admin user can edit an Storage Class

Cluster Storage view from the Cluster Storage DSP tab:
- Regular user can create a cluster storage using a new storage class
- A Cluster Storage with a disabled SC shows the deprecated warning
- If all SC are disabled except one, the SC dropdown should be disabled

## How Has This Been Tested?
As there were some UI changes, the tests ran against the local UI and a 2.14 cluster as the backend

image

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/.env
+++ b/frontend/.env
@@ -8,4 +8,4 @@ ODH_FAVICON=odh-favicon.svg
 ODH_NOTEBOOK_REPO=opendatahub
 
 ########## Change this version with each dashboard release ###########
-INTERNAL_DASHBOARD_VERSION=v2.26.0
+INTERNAL_DASHBOARD_VERSION=v2.27.0
--- a/frontend/src/__tests__/cypress/cypress/fixtures/resources/yaml/storage_class.yaml
+++ b/frontend/src/__tests__/cypress/cypress/fixtures/resources/yaml/storage_class.yaml
@@ -1,0 +1,11 @@
+kind: StorageClass
+apiVersion: storage.k8s.io/v1
+metadata:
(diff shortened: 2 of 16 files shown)


manosnoam (commented):
Thank you Fede!
Can you share job results from Jenkins ?

FedeAlonso:
> Thank you Fede! Can you share job results from Jenkins ?

@manosnoam As this PR has UI modifications, I'm not running this in jenkins. 
The results:

Cluster Storage:
Screenshot from 2024-10-07 13-20-30

Storage Class:
Screenshot from 2024-10-07 13-24-26
(The one that's failing is because when running the UI locally it uses the token from the oc login instead of performing a login. As the OC login is done with an admin user in order to provision, teardown, and have access to admin views, and the test which fails is testing that a non admin user has no access to the storage classes view which is one of those only admin views)

FedeAlonso:
@manosnoam here you have the result of the failing test when executed against a live cluster :
image

FedeAlonso:
I've deleted some tests as they were already done for the mockend environment.
Results:
image

manosnoam (approved):

PR #3292, opened 2024-10-02
fix(MR): 12864 disable archive for mr/mv if has deploys

gitdallas:
closes: https://issues.redhat.com/browse/RHOAIENG-12864

## Description
disable archiving models/versions with deploys

mr list page:
image

mr detail and versions list page:
image
image

mr version detail page:
image

## How Has This Been Tested?
tested locally with mr cluster

## Test Impact
tried adding tests, but cypress doesn't seem to call the k8s inferencemodel endpoint? will try again when i come back from vacation if still necesssary

cypress activity:
image
network activity for k8s inferencemodel in the actual app:
image

## Request review criteria:
make sure that in all places where you can archive a model or version, you cannot do it if it has a deploy... but otherwise you can.

If you have UI changes: 


frontend/src/pages/modelRegistry/screens/ModelVersions/ModelVersionsTable.tsx
```diff
@@ -18,25 +21,37 @@ const ModelVersionsTable: React.FC<ModelVersionsTableProps> = ({
...
+            modelVersion={mv}
+            isArchiveModel={isArchiveModel}
+            refresh={refresh}
+          />
+        );
+      }}
+    />
+  );
+};
+// >>>>>>> 980ca3be (fix(MR): disable archive for mr/mv if has deploys)
```
manaswinidas, line 55: Is this comment line needed?

manaswinidas (changes requested):
Works fine.

About the tests, @gitdallas did you try adding an intercept with mock data ?

gitdallas:
@manaswinidas yea i just threw this PR up prior to going on PTO incase it needed to get picked up urgently or something happened 😆 

i was mocking it, but after debugging with mike it turns out the intercept wasn't working because it wasn't also intercepting the call to projects, which affected the URL of the inferenceService intercept 😵  updating PR soon with tests and cleanup

ppadti:
@gitdallas I think we need to hide the deployments tab in archived versions details page and version details page of archived models as well.

gitdallas:
@ppadti - ah i didn't catch that "add-on" 😆 ... will do

frontend/src/__tests__/cypress/cypress/tests/mocked/modelRegistry/modelVersionArchive.cy.ts
```diff
@@ -272,6 +281,63 @@ describe('Archiving version', () => {
     });
   });
 
+  it('Non archived version details page does have deployment tab', () => {
```
manaswinidas, line 284: nit: can we add "has" instead of "does have"? Also "Deployments" instead of "deployment".

frontend/src/__tests__/cypress/cypress/tests/mocked/modelRegistry/modelVersionArchive.cy.ts
```diff
@@ -272,6 +281,63 @@ describe('Archiving version', () => {
...
+    initIntercepts({});
+    modelVersionArchive.visitModelVersionDetails();
+    modelVersionArchive.findVersionDetailsTab().should('exist');
+    modelVersionArchive.findVersionDeploymentTab().should('exist');
+  });
+
+  it('Archived version details page does not have deployment tab', () => {
```
manaswinidas, line 291: same nit: "Deployments" instead of "deployment".

manaswinidas (changes requested):
just small nits. Otherwise lgtm.

@gitdallas Tests look good, thanks for adding them. Also, checked whether "Deployments" tab is hidden for Archived versions(coz this changed since I last reviewed).

manaswinidas:
@yih-wang can you have a look at the screenshots?

yih-wang:
LGTM. The only thing is one microcopy issue I'm not sure if we addressed anywhere - the microcopy in the last two screens should be `archive/restore model version` instead of `archive/restore version`.

mturley:
@gitdallas yeah let's knock that out while we're in here since it's just microcopy

manaswinidas (approved):
Thanks @gitdallas 

/lgtm

mturley (approved):
LGTM, thanks @gitdallas

PR #3293, opened 2024-10-03
add pipeline input parameters

pnaik1:
https://issues.redhat.com/browse/RHOAIENG-9425

## Description
Added back the input parameter tab which was removed previously
Both for runs and scheduled runs
Screenshot 2024-10-03 at 12 01 56 PM

Screenshot 2024-10-03 at 12 03 14 PM

## How Has This Been Tested?

1. Create a run with parameters
2. Check if the input parameter tab is present
3. Click on the input parameter and check for if parameters are correctly displayed

## Test Impact
Updated cypress test

## Request review criteria:

If you have UI changes: 


frontend/src/concepts/pipelines/content/pipelinesDetails/pipelineRun/PipelineRunTabParameters.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+  const details: DetailItem[] = parameters.map(([key, initialValue]) => {
+    let value = initialValue;
+
+    if (typeof value === 'boolean') {
+      value = value ? 'True' : 'False';
+    }
+
+    if (typeof value === 'object') {
+      value = JSON.stringify(value);
```
Gkrumbach07, line 50: pipeline spec for the run has input definition types that can be used
```json
  inputDefinitions:
    parameters:
      datastore:
        parameterType: STRUCT
      hyperparameters:
        parameterType: STRUCT
```

These should match up the the run runtime_config parameters, if not you can fall back to just a string, but if there is a match you can use the same type of parsing we do here: https://github.com/opendatahub-io/odh-dashboard/blob/33ea6969abe77445c2e0d8985d083ed22cb79e91/frontend/src/concepts/pipelines/content/pipelinesDetails/pipelineRun/SelectedNodeInputOutputTab.tsx#L56-L90

make sure to create a reusable util for this then to avoid duplicate code

Gkrumbach07 (approved):
/lgtm

PR #3295, opened 2024-10-03
UX updates for pipeline logs

DaoDaoNoCode:
JIRA: https://issues.redhat.com/browse/RHOAIENG-6582

## Description

Follow the mockup to do the following updates:
1. Update the alert text, add 2 links to allow users to view raw logs or download the current step in the alert

Screenshot 2024-10-03 at 12 04 19 PM

2. Add a mousedown event to capture the highlight and click the event in the log window and pause the logs as needed
3. Updated the tooltip and icon for pause/resume logs

Screenshot 2024-10-03 at 12 20 48 PM

Screenshot 2024-10-03 at 12 21 10 PM

4. Remove the log viewer footer

## How Has This Been Tested?

1. Check the JIRA comment, use the log-forever-pipeline and create a run
2. Check the node logs, and verify all the changes based on the [mockup](https://www.figma.com/proto/etbN0ccCcZR1qKCpJXQNAg/Pipeline-phase-2?page-id=12031%3A86470&node-id=12534-92933&node-type=frame&viewport=714%2C-1503%2C0.35&t=214XjsqaSYtM6ngc-1&scaling=scale-down-width&content-scaling=fixed&starting-point-node-id=12534%3A92933)

## Test Impact

N/A, mostly UX updates, and needs manual verification.

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/concepts/dashboard/DashboardLogViewer.tsx
+++ b/frontend/src/concepts/dashboard/DashboardLogViewer.tsx
@@ -5,7 +5,7 @@ const DashboardLogViewer: React.FC<{
   data: string;
   logViewerRef: React.MutableRefObject<{ scrollToBottom: () => void } | undefined>;
   toolbar: JSX.Element | boolean;
-  footer: JSX.Element | false;
+  footer?: JSX.Element | false;
   onScroll: React.ComponentProps<typeof LogViewer>['onScroll'];
   height?: number | string;
   isTextWrapped?: boolean;
--- a/frontend/src/concepts/pipelines/content/pipelinesDetails/pipelineRun/runLogs/LogsTab.tsx
+++ b/frontend/src/concepts/pipelines/content/pipelinesDetails/pipelineRun/runLogs/LogsTab.tsx
@@ -17,6 +17,7 @@ import {
   ToolbarItem,
   Tooltip,
   Truncate,
+  Icon,
 } from '@patternfly/react-core';
 import { DownloadIcon } from '@patternfly/react-icons/dist/esm/icons/download-icon';
 import { LogViewer, LogViewerSearch } from '@patternfly/react-log-viewer';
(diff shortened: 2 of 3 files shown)


DaoDaoNoCode:
@yannnz Can you check the screenshots above?

yannnz:
> @yannnz Can you check the screenshots above?

@DaoDaoNoCode, wonder what are the sizes of the pause and resume action button? It seems a bit smaller. 
For the second point of adding mousedown events, just double check if I understand it correctly: we already have the scroll to pause interaction, so we only need to add highlight to pause and click to pause the log viewer streaming in this ticket, right?

DaoDaoNoCode:
Hi @yannnz, the icon is 14x14, which is the same size as the search icon
Screenshot 2024-10-09 at 9 40 25 AM
Screenshot 2024-10-09 at 9 40 54 AM
Screenshot 2024-10-09 at 9 41 13 AM

For the second point, what you said is correct.

dpanshug (commented):
When i expand or un-expand the LogTabStatus the log scrolls to the start. I would expect it to be at the same position. wdyt @yannnz ?

https://github.com/user-attachments/assets/d8e58cb7-a08e-47d7-bbcc-8d27b7f30aa6

dpanshug:
@DaoDaoNoCode regarding the pause/resume icon size, i agree with @yannnz that seems a bit smaller to me also. can we increase to `lg`? https://www.patternfly.org/components/icon#sizes

DaoDaoNoCode:
@yannnz @dpanshug I updated the icon size to medium, does it look better?

Screenshot 2024-10-09 at 12 29 26 PM

Screenshot 2024-10-09 at 12 27 13 PM

DaoDaoNoCode:
As for the issue you mentioned above @dpanshug, I think it's a known issue in the PF log viewer and a limitation there because everything got rendered dynamically. When the container height changes, it's hard to estimate which line it remained at.

Gkrumbach07 (approved):
/lgtm

tested and works as expected. I am able to highlight and copy without it jumping around

PR #3296, opened 2024-10-03
fix: pvc fixes

olavtar:
## Description

PVC Name update in the servingRuntime fixes

## How Has This Been Tested?

Tested locally

## Test Impact

Tested both fast and slow deletion, followed by model recreation.

## Request review criteria:

If you have UI changes: 


frontend/src/pages/modelServing/screens/projects/nimUtils.ts
```diff
@@ -1,6 +1,7 @@
 // NGC stands for NVIDIA GPU Cloud.
 
-import { ProjectKind, SecretKind, TemplateKind } from '~/k8sTypes';
+import { K8sResourceCommon } from '@openshift/dynamic-plugin-sdk-utils';
+import { ProjectKind, SecretKind, ServingRuntimeKind, TemplateKind } from '~/k8sTypes';
 import { getTemplate } from '~/api';
 
 const NIM_SECRET_NAME = 'nvidia-nim-access';
```
andrewballantyne, line 7: Can we add to the list of constants at the top of the file? While reviewing, I noticed you have a duplicate constant for the template name.

There are two instances of:
```
const TEMPLATE_NAME = 'nvidia-nim-serving-template';
```

frontend/src/pages/modelServing/screens/projects/nimUtils.ts
```diff
@@ -77,3 +80,44 @@ export const getNIMServingRuntimeTemplate = async (
     return undefined;
   }
 };
+
+export const updatePVCName = (
+  servingRuntime: ServingRuntimeKind,
+  pvcName: string,
```
andrewballantyne, line 86: Perhaps just generate the PVC in here? Do you need it anywhere else?

frontend/src/pages/modelServing/screens/projects/NIMServiceModal/DeployNIMServiceModal.tsx
```diff
@@ -156,11 +159,18 @@ const DeployNIMServiceModal: React.FC<DeployNIMServiceModalProps> = ({
...
-      setServingRuntimeSelected(getServingRuntimeFromTemplate(nimTemplate));
+      if (nimTemplate) {
+        const servingRuntime = getServingRuntimeFromTemplate(nimTemplate);
+        setServingRuntimeSelected(
+          servingRuntime ? updatePVCName(servingRuntime, nimPVCName) : undefined,
+        );
+      } else {
+        setServingRuntimeSelected(undefined);
+      }
     };
```
andrewballantyne, line 170: Hmmm -- I think it's an interesting thought to modify it directly as you get it -- you can keep this if you want. I'd have thought to do the action on submit.

If you keep this action, generate the PVC name part of the effort -- using `useState` without the setter function being used is always a bit odd to see; I imagine there are good use-cases for it -- but I'm not sure it matters unless you submit.

frontend/src/pages/modelServing/screens/projects/nimUtils.ts
```diff
@@ -77,3 +80,44 @@ export const getNIMServingRuntimeTemplate = async (
     return undefined;
   }
 };
+
+export const updatePVCName = (
```
andrewballantyne, line 84:
```suggestion
export const updateServingRuntimeTemplate = (
```

I think it's more explicit to say you're updating the template -- not the PVC Name -- as you're applying the pvc name to the template.

andrewballantyne, line 84: Can we get a unit test to make sure this performs the changes as you see them? Simple providing different sets of information and asserting the response. Also helps not keep reducing the test coverage we have with new NIM code.

andrewballantyne (changes requested):

frontend/src/pages/modelServing/screens/projects/nimUtils.ts
olavtar: Yes. https://github.com/olavtar/odh-dashboard/blob/pvc-fixes/frontend/src/pages/modelServing/screens/projects/NIMServiceModal/DeployNIMServiceModal.tsx#L240

frontend/src/pages/modelServing/screens/projects/nimUtils.ts
olavtar: Updated

frontend/src/pages/modelServing/screens/projects/NIMServiceModal/DeployNIMServiceModal.tsx
olavtar: Updated

frontend/src/pages/modelServing/screens/projects/nimUtils.ts
olavtar: Updated

olavtar: Unit test added.

andrewballantyne (approved):

PR #3301, opened 2024-10-04
Dry run all workbench create/update calls before submit

DaoDaoNoCode:
JIRA: [RHOAIENG-1142](https://issues.redhat.com/browse/RHOAIENG-1142)

## Description

Check all the API calls that are executed when creating/updating a workbench, make sure dry run them before truly submitting.

## How Has This Been Tested?

1. Go to the workbench creation page
2. Create a Config Map environment variables and give it an invalid key like `test key` (space is not permitted in the key)
3. Create a new PVC
4. Create a new data connection
5. Click submit, and you should get an error message because the key is invalid
6. Change that key to valid like `test-key`
7. Submit again, you should not get any error because the previous step dry runs everything, so it didn't create a PVC and data connection. Otherwise, you would get an error message saying `xxx already exists`

## Test Impact

Added a unit test to verify all the networks have dry run options.

## Request review criteria:

If you have UI changes: 


pnaik1:
Followed this step:
Submit again, you should not get any error because the previous step dry runs everything, so it didn't create a PVC and data connection. Otherwise, you would get an error message saying xxx already exists
I still get that secret exists, Am I missing something here??

Screenshot 2024-10-07 at 6 07 33 PM

DaoDaoNoCode:
@pnaik1 Nice catch, it's a bug and I updated it.

Gkrumbach07 (approved):
/lgtm

this looks good. everything looks to be dry run first now

PR #3306, opened 2024-10-07
Disable manage permissions when no rolebinding present for the default group

ppadti:
Closes: [RHOAIENG-13091](https://issues.redhat.com/browse/RHOAIENG-13091)

## Description
This PR aims to disable the Manage Permissions when model registry is in not ready state.
Screenshot 2024-10-04 at 8 02 54 PM

Screenshot 2024-10-10 at 8 53 55 PM

## How Has This Been Tested?
Create a model registry
when Model registry is in progressing state, Manage permissions on the model registry table will be disabled.

## Test Impact
Updated cypress test.

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/pages/modelRegistrySettings/ModelRegistriesTableRow.tsx
+++ b/frontend/src/pages/modelRegistrySettings/ModelRegistriesTableRow.tsx
@@ -1,23 +1,34 @@
 import React from 'react';
 import { ActionsColumn, Td, Tr } from '@patternfly/react-table';
-import { Link } from 'react-router-dom';
-import { ModelRegistryKind } from '~/k8sTypes';
+import { useNavigate } from 'react-router-dom';
+import { Button, Tooltip } from '@patternfly/react-core';
+import { ModelRegistryKind, RoleBindingKind } from '~/k8sTypes';
 import ResourceNameTooltip from '~/components/ResourceNameTooltip';
+import { ContextResourceData } from '~/types';
 import ViewDatabaseConfigModal from './ViewDatabaseConfigModal';
 import DeleteModelRegistryModal from './DeleteModelRegistryModal';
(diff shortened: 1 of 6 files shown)


frontend/src/pages/modelRegistrySettings/ModelRegistriesTableRow.tsx
```diff
@@ -35,12 +45,18 @@ const ModelRegistriesTableRow: React.FC<ModelRegistriesTableRowProps> = ({
...
         <Td modifier="fitContent">
-          <Link
-            aria-label={`Manage permissions for model registry ${mr.metadata.name}`}
-            to={`/modelRegistrySettings/permissions/${mr.metadata.name}`}
-          >
-            Manage permissions
-          </Link>
+          {filteredRoleBindings.length === 0 ? (
+            <Tooltip content="You can manage permissions when the model registry becomes available.">
+              <span className="pf-v5-u-disabled-color-100">Manage permissions</span>
```
mturley, line 50: Instead of the span here, we can use a PF Button with isAriaDisabled so it can receive keyboard focus.
```tsx
<Button isAriaDisabled variant="link" isInline>Manage permissions</Button>
```

For consistency, we should probably also use a `<Button variant="link" isInline>` in the enabled case below as well, just in case there are spacing differences. We can add an `onClick={() => navigate(...)}` instead of the `to` prop, and get `navigate` from `useNavigate()` imported from `react-router-dom`.

I'm actually not sure if we need `isInline` on them as well, whichever has the right spacing is fine.

mturley (commented):
Mostly LGTM, one small nitpick @ppadti

mturley (approved):
LGTM, thanks @ppadti !

PR #3310, opened 2024-10-07
NIM Model size propagated to the expanded section

olavtar:
https://issues.redhat.com/browse/RHOAIENG-14039

## Description

The fix to display the correct Model server sizes on the expanded section.

## How Has This Been Tested?

Tested locally

## Test Impact

Tested with setting the sizes of the Model to Medium, Large and Custom. 

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/pages/modelServing/screens/projects/ModelMeshSection/ServingRuntimeDetails.tsx
+++ b/frontend/src/pages/modelServing/screens/projects/ModelMeshSection/ServingRuntimeDetails.tsx
@@ -25,7 +25,7 @@ const ServingRuntimeDetails: React.FC<ServingRuntimeDetailsProps> = ({ obj, isvc
   const enabledAcceleratorProfiles = acceleratorProfile.acceleratorProfiles.filter(
     (ac) => ac.spec.enabled,
   );
-  const resources = obj.spec.containers[0].resources || isvc?.spec.predictor.model?.resources;
+  const resources = isvc?.spec.predictor.model?.resources || obj.spec.containers[0].resources;
   const sizes = getServingRuntimeSizes(dashboardConfig);
   const size = sizes.find(
     (currentSize) => getResourceSize(sizes, resources || {}).name === currentSize.name,
--- a/frontend/src/pages/modelServing/screens/projects/NIMServiceModal/DeployNIMServiceModal.tsx
+++ b/frontend/src/pages/modelServing/screens/projects/NIMServiceModal/DeployNIMServiceModal.tsx
@@ -210,7 +210,7 @@ const DeployNIMServiceModal: React.FC<DeployNIMServiceModalProps> = ({
       NamespaceApplicationCase.KSERVE_NIM_PROMOTION,
       projectContext?.currentProject,
       servingRuntimeName,
-      true,
+      false,
     );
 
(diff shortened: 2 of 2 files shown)


frontend/src/pages/modelServing/screens/projects/ModelMeshSection/ServingRuntimeDetails.tsx
```diff
@@ -25,7 +25,7 @@ const ServingRuntimeDetails: React.FC<ServingRuntimeDetailsProps> = ({ obj, isvc
   const enabledAcceleratorProfiles = acceleratorProfile.acceleratorProfiles.filter(
     (ac) => ac.spec.enabled,
   );
-  const resources = obj.spec.containers[0].resources || isvc?.spec.predictor.model?.resources;
+  const resources = isvc?.spec.predictor.model?.resources || obj.spec.containers[0].resources;
```
andrewballantyne, line 28: Did you determine why NIM model only has this issue? It seems the KServe Model does not.

mpaulgreen: We could have resource request and limits defined in template, servingruntime and inferenceservice. Currently the code in dashboard is picking it from runtime first. Since resource requests in inferenceservice takes precedence over servingruntime, it needs to display the values from inferenceservice. The code currently in KServe is working since the CR generated for runtime may not have these attr that is not expected as per KServe definitions of CR. If. it is a wider impact on testing, then we can look to implement some other solution.

andrewballantyne: I'm not against this change, and I believe you are right this is a solid change -- but I still would like to know why NIM model appears to be the only one who sees it. Can you reproduce this in KServe using a similar ServingRuntime? If it happens in KServe, it was a bug we dodged. If it happens in NIM and not KServe, then there could be a bug floating in the NIM Modal. Solving this and not understanding what caused the bug leaves us open to other bugs that this will mask.

mpaulgreen: No AFAIK this cannot be reproduced in Single Model Server deployment as engineer has not defined the resource requests and limits in ServingRuntime generated from dashboard. I would say that this is a bug that was dodged. If I were you I would fix this before something comes up. The current runtime and inferenceservice CR for NIMs are correct as per KServe CR definition. `oc explain servingruntime` will help to inspect.

andrewballantyne: Okay, this fix is fine... but the bug you're seeing is that you're submitting as `Model Mesh` not `KServe` -- please fix this with this PR @olavtar.

https://github.com/opendatahub-io/odh-dashboard/blob/main/frontend/src/pages/modelServing/screens/projects/NIMServiceModal/DeployNIMServiceModal.tsx#L213

olavtar: Fixed.

andrewballantyne (approved):

PR #3314, opened 2024-10-08
Validation for model names and disabled NIM metrics hyperlink

olavtar:
https://issues.redhat.com/browse/RHOAIENG-13881
https://issues.redhat.com/browse/RHOAIENG-14044

## Description

Added validation for model names and disabled the NIM models metrics hyperlink on the Models tab

## How Has This Been Tested?

Tested locally.

## Test Impact

Tested functionality: Verified that KServe still displays the metrics link.
NIM validation: Confirmed that NIM does not display the metrics link.

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/pages/modelServing/screens/global/InferenceServiceTableRow.tsx
+++ b/frontend/src/pages/modelServing/screens/global/InferenceServiceTableRow.tsx
@@ -35,25 +35,24 @@ const InferenceServiceTableRow: React.FC<InferenceServiceTableRowProps> = ({
   isGlobal,
   columnNames,
 }) => {
+  const { projects } = React.useContext(ProjectsContext);
+  const project = projects.find(byName(inferenceService.metadata.namespace)) ?? null;
+  const isKServeNIMEnabled = project ? isProjectNIMSupported(project) : false;
+
   const [modelMetricsEnabled] = useModelMetricsEnabled();
   const kserveMetricsEnabled = useIsAreaAvailable(SupportedArea.K_SERVE_METRICS).status;
 
   const modelMesh = isModelMesh(inferenceService);
   const modelMeshMetricsSupported = modelMetricsEnabled && modelMesh;
-  const kserveMetricsSupported = modelMetricsEnabled && kserveMetricsEnabled && !modelMesh;
+  const kserveMetricsSupported =
+    modelMetricsEnabled && kserveMetricsEnabled && !modelMesh && !isKServeNIMEnabled;
   const displayName = getDisplayNameFromK8sResource(inferenceService);
-
-  const { projects } = React.useContext(ProjectsContext);
-  const project = projects.find(byName(inferenceService.metadata.namespace)) ?? null;
-  const isKServeNIMEnabled = project ? isProjectNIMSupported(project) : false;
 
   return (
     <>
       <Td dataLabel="Name">
         <ResourceNameTooltip resource={inferenceService}>
-          {isKServeNIMEnabled ? (
-            displayName
-          ) : modelMeshMetricsSupported || kserveMetricsSupported ? (
+          {modelMeshMetricsSupported || kserveMetricsSupported ? (
             <Link
               data-testid={`metrics-link-${displayName}`}
               to={
--- a/frontend/src/pages/modelServing/screens/global/InferenceServiceTableRow.tsx
+++ b/frontend/src/pages/modelServing/screens/global/InferenceServiceTableRow.tsx
@@ -51,18 +51,9 @@ const InferenceServiceTableRow: React.FC<InferenceServiceTableRowProps> = ({
     <>
       <Td dataLabel="Name">
         <ResourceNameTooltip resource={inferenceService}>
-          {modelMeshMetricsSupported ? (
-            <Link
-              data-testid={`metrics-link-${displayName}`}
-              to={
-                isGlobal
-                  ? `/modelServing/${inferenceService.metadata.namespace}/metrics/${inferenceService.metadata.name}`
-                  : `/projects/${inferenceService.metadata.namespace}/metrics/model/${inferenceService.metadata.name}`
-              }
-            >
-              {displayName}
-            </Link>
-          ) : kserveMetricsSupported ? (
+          {isKServeNIMEnabled ? (
+            displayName
+          ) : modelMeshMetricsSupported || kserveMetricsSupported ? (
             <Link
               data-testid={`metrics-link-${displayName}`}
               to={
--- a/frontend/src/pages/modelServing/screens/projects/InferenceServiceModal/ProjectSection.tsx
+++ b/frontend/src/pages/modelServing/screens/projects/InferenceServiceModal/ProjectSection.tsx
@@ -1,14 +1,19 @@
 import * as React from 'react';
 import { FormGroup, Text } from '@patternfly/react-core';
+import { translateDisplayNameForK8sAndReport } from '~/concepts/k8s/utils';
 
 type ProjectSectionType = {
   projectName: string;
 };
 
-const ProjectSection: React.FC<ProjectSectionType> = ({ projectName }) => (
-  <FormGroup label="Project">
-    <Text>{projectName}</Text>
-  </FormGroup>
-);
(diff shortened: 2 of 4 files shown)


frontend/src/pages/modelServing/screens/global/InferenceServiceTableRow.tsx
```diff
@@ -51,18 +51,9 @@ const InferenceServiceTableRow: React.FC<InferenceServiceTableRowProps> = ({
...
-              data-testid={`metrics-link-${displayName}`}
-              to={
-                isGlobal
-                  ? `/modelServing/${inferenceService.metadata.namespace}/metrics/${inferenceService.metadata.name}`
-                  : `/projects/${inferenceService.metadata.namespace}/metrics/model/${inferenceService.metadata.name}`
-              }
-            >
-              {displayName}
-            </Link>
-          ) : kserveMetricsSupported ? (
```
andrewballantyne, line 65: Curious why you didn't just update the kserveMetricsSupported to include "not NIM"?

Although this code block we have looks very weird lol

olavtar: I did in the beginning but then I saw that modelMeshMetricsSupported and kserveMetricsSupported had duplicated logic, so I chose to add the isKServeNIMEnabled check separately for clarity and to make the logic explicit. I can update the kserveMetricsSupported to handle the "not NIM" case if you think it would make it cleaner.

andrewballantyne: Yeah, I noticed that -- the more you change the longer it takes for me to verify and approve. I'll try to get this before my PTO -- but typically stay short and to the point unless the logic is not compatible. Every cleanup you do for us has me testing more to make sure we don't have some other brittle aspect that can slip out in non NIM ways. In the end it's likely a net positive, just expressing how I look at everything.

frontend/src/pages/modelServing/screens/projects/InferenceServiceModal/ProjectSection.tsx
```diff
@@ -1,14 +1,19 @@
...
   projectName: string;
 };
 
-const ProjectSection: React.FC<ProjectSectionType> = ({ projectName }) => (
-  <FormGroup label="Project">
-    <Text>{projectName}</Text>
-  </FormGroup>
-);
+const ProjectSection: React.FC<ProjectSectionType> = ({ projectName }) => {
+  const [translatedName] = translateDisplayNameForK8sAndReport(projectName);
```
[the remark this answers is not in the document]
olavtar: Updated

frontend/src/pages/modelServing/screens/global/InferenceServiceTableRow.tsx
lucferbux: Hi @olavtar pardon me cause I don't have all the context, but shouldn't this produce the same result?

```typescript
        <ResourceNameTooltip resource={inferenceService}>
          {modelMeshMetricsSupported || kserveMetricsSupported ? (
            <Link
              data-testid={`metrics-link-${displayName}`}
              to={
                isGlobal
                  ? `/modelServing/${inferenceService.metadata.namespace}/metrics/${inferenceService.metadata.name}`
                  : `/projects/${inferenceService.metadata.namespace}/metrics/model/${inferenceService.metadata.name}`
              }
            >
              {displayName}
            </Link>
          ) : (
            displayName
          )}
        </ResourceNameTooltip>
```

There's value on refactoring duplicate code, but how NIM is going to affect this component if the fallback is the same as having the feature selected? Maybe for readability, but in case we wanna add later metrics, we can add a comment explaining that nim will be enabled when we support metrics.

olavtar: Hi @lucferbux, the isKServeNIMEnabled check is important because we don’t want to show a link when NIM Metrics isn't enabled, even if other metrics conditions are met. So, removing this condition would change the behavior, as it would allow the link to be displayed when NIM Metrics are not enabled, which is not the intended outcome.

andrewballantyne: Yeah we are going to definitely not want fix this logic and also just duplicate other logic.

Go with something like this:
```typescript
const InferenceServiceTableRow: React.FC<InferenceServiceTableRowProps> = ({
  obj: inferenceService,
  servingRuntime,
  onDeleteInferenceService,
  onEditInferenceService,
  isGlobal,
  columnNames,
}) => {
  const { projects } = React.useContext(ProjectsContext);
  const project = projects.find(byName(inferenceService.metadata.namespace)) ?? null;

  const isKServeNIMEnabled = project ? isProjectNIMSupported(project) : false;
```
You'll have to move this logic to the top of the component to use it where we need it

```typescript
  const [modelMetricsEnabled] = useModelMetricsEnabled();
  const kserveMetricsEnabled = useIsAreaAvailable(SupportedArea.K_SERVE_METRICS).status;

  const modelMesh = isModelMesh(inferenceService);
  const modelMeshMetricsSupported = modelMetricsEnabled && modelMesh;
  const kserveMetricsSupported = modelMetricsEnabled && kserveMetricsEnabled && !modelMesh && !isKServeNIMEnabled;
```
KServe Metrics are not supported when you're NIM -- flip this flag

```typescript
  const displayName = getDisplayNameFromK8sResource(inferenceService);

  return (
    <>
      <Td dataLabel="Name">
        <ResourceNameTooltip resource={inferenceService}>
          {modelMeshMetricsSupported || kserveMetricsSupported ? (
            <Link
              data-testid={`metrics-link-${displayName}`}
              to={
                isGlobal
                  ? `/modelServing/${inferenceService.metadata.namespace}/metrics/${inferenceService.metadata.name}`
                  : `/projects/${inferenceService.metadata.namespace}/metrics/model/${inferenceService.metadata.name}`
              }
            >
              {displayName}
            </Link>
          ) : (
            displayName
          )}
        </ResourceNameTooltip>
      </Td>
```

frontend/src/pages/modelServing/screens/projects/NIMServiceModal/NIMModelDeploymentNameSection.tsx
```diff
@@ -19,6 +20,10 @@ const NIMModelDeploymentNameSection: React.FC<NIMModelDeploymentNameSectionProps
       data-testid="model-deployment-name-section"
       value={data.name}
       onChange={(e, name) => setData('name', name)}
+      onBlur={() => {
+        const [translatedName] = translateDisplayNameForK8sAndReport(data.name);
+        setData('name', translatedName);
```
andrewballantyne, line 25: Would this not just convert the field for what the user typed? Can we not support this under the hood and then supply it at the end? We have a typical two named set in our world... display name & k8s name -- we don't want to lose the display name to have a clean k8s name... let me see if I can see what guidance I can provide more specifically to solve this.

olavtar: @andrewballantyne This doesn't resolve the issue of allowing illegal characters. For example, if you enter the name "אקדא אעגכדע" (which contains no digits and includes a space), the problem persists.

andrewballantyne: If you make a display name with those characters it should work fine 🤔 (or so I would think) -- k8s names cannot have it. Today we trim those out, which got you 12456 as a resulting username -- this had problems being just numbers.

What error are you getting? @olavtar

andrewballantyne: This utility is used on the create project modal... this is the result of just using those Hebrew characters: 
image

olavtar: Updated

andrewballantyne (approved):

PR #3320, opened 2024-10-10
[RHOAIENG-1101] Cannot remove additional storage from the workbench configuration

jpuzz0:
https://issues.redhat.com/browse/RHOAIENG-1101

## Description
Workbench create/edit now has a table view to allow for multiple PVC instances to be created, where currently only 1 row is creatable/editable. The buttons above the table are not actionable yet and are placeholders for now. Some logic has been updated as a part of the submission process to attach multiple volumes (from PVC creations) for the notebook when available.

**Default table data on creation**
image

**Detach modal**
image

**Empty state when all are detached**
image

**Edit modal on creation**
> **NOTE:** this edit modal is a starting point and meant to be expanded upon with a separate jira; https://issues.redhat.com/browse/RHOAIENG-7681
image

(cc @xianli123)

## How Has This Been Tested?
When creating a workbench, verify the set of fields that were previously visible are now replaced with the editable table view. The Edit modal itself is a part of a different story, but one was made here as a starting point, where mainly the mount path and some show/hide of alerts will change (along with some implementation).

## Test Impact
Updated mocked cypress tests to simply verify table data for now as a part of creating/editing a workbench. Since this feature has other parts, e2e mock testing will be dependent on their completion.

## Request review criteria:

If you have UI changes: 


xianli123:
Thanks @jpuzz0 ! Overall LGTM! 
As a tip (it may be outside the scope of https://issues.redhat.com/browse/RHOAIENG-1101), when creating a workbench or storage, the storage size should display "Max XXGi" instead of "XXGi."

jpuzz0:
> Thanks @jpuzz0 ! Overall LGTM! As a tip (it may be outside the scope of https://issues.redhat.com/browse/RHOAIENG-1101), when creating a workbench or storage, the storage size should display "Max XXGi" instead of "XXGi."

@xianli123 I've made this change. Thanks for calling it out!

xianli123:
Thanks @jpuzz0 . Once it's changed, it will look good!

frontend/src/pages/projects/screens/spawner/SpawnerPage.tsx
```diff
@@ -68,19 +72,30 @@ const SpawnerPage: React.FC<SpawnerPageProps> = ({ existingNotebook }) => {
...
+          description: defaultClusterStorage.description,
+        },
+        size: defaultClusterStorage.size || defaultNotebookSize,
+        storageClassName: defaultStorageClassName || defaultClusterStorage.storageClassName,
+        mountPath: defaultClusterStorage.mountPath,
+      },
+      existing: {
+        storage: getRootVolumeName(existingNotebook),
+      },
+    },
```
Gkrumbach07, line 97: so this exisitng data should include all attached PVCs. create some hook that is basically the inverse of `useRelatedNotebooks` called `useRelatedPVCs`

frontend/src/pages/projects/screens/spawner/spawnerUtils.ts
```diff
@@ -395,22 +369,19 @@ export const isEnvVariableDataValid = (envVariables: EnvVariable[]): boolean =>
...
     projectName &&
     isK8sNameDescriptionDataValid(notebookData) &&
     image.imageStream &&
     image.imageVersion
   );
 
-  const newStorageFieldInvalid = storageType === StorageType.NEW_PVC && !creating.nameDesc.name;
-  const existingStorageFieldInvalid = storageType === StorageType.EXISTING_PVC && !existing.storage;
-  const isStorageDataValid = !newStorageFieldInvalid && !existingStorageFieldInvalid;
+  const isStorageDataValid = storageData.length > 0;
```
Gkrumbach07, line 384: storage data is not required so this is not needed

frontend/src/pages/projects/screens/spawner/storage/ClusterStorageEditModal.tsx
```diff
@@ -0,0 +1,131 @@
+import React from 'react';
```
Gkrumbach07, line 1: this file is just the placeholder until dipanshu merges right?

frontend/src/pages/projects/screens/spawner/SpawnerPage.tsx
Gkrumbach07: this may also have to be restructured then because existing would need to be able to update the mount path

my suggestions is something along the lines of

```tsx
{
  storageType: existingNotebook ? StorageType.EXISTING_PVC : StorageType.NEW_PVC,
  name: getRootVolumeName(existingNotebook),
  nameDesc: {
     name: existingName || k8sNameDescriptionData.data.name || defaultClusterStorage.name,
     description: existingDescription || defaultClusterStorage.description,
  },
  size: existingSize || defaultClusterStorage.size || defaultNotebookSize,
  storageClassName: existingStorageClassName || defaultStorageClassName || defaultClusterStorage.storageClassName,
  mountPath: existingMountPath || defaultClusterStorage.mountPath,
}
```

Gkrumbach07 (changes requested):

frontend/src/pages/projects/screens/spawner/storage/ClusterStorageEditModal.tsx
jpuzz0: Yes, I think he can use it and expand upon it though.

frontend/src/pages/projects/screens/spawner/SpawnerPage.tsx
jpuzz0: Updated, but using a variation of what was suggested here.

xianli123:
Hi @jpuzz0 , Thank you for updating the description above and adding screenshots. 
I have some suggestions for the empty state you shared above.
image

- As Katie recently suggested in Figma, the description of this empty state will be `To save your project's data, attach cluster storage to your workbench. Your data will not persist without storage.`. We can follow this latest content.

- As for the empty state style, we can remove the table's header. PF has [a guideline](https://www.patternfly.org/components/table/react-demos/#empty-states) for when we can use a table empty state. I quoted it here: "Empty states are useful in a table when a filter returns no results, while data is loading, or when any type of error or exception condition occurs." In our case, the empty state should be a common empty state instead of a table empty state. WDYT?

DaoDaoNoCode (commented):
Hi, I don't see creation or attach existing PVC button on my spawner page, is there anything wrong with my settings?

Screenshot 2024-10-18 at 4 19 17 PM

frontend/src/pages/projects/screens/spawner/storage/ClusterStorageEditModal.tsx
```diff
@@ -0,0 +1,132 @@
...
+      isOpen
+      variant={ModalVariant.small}
+      title="Edit storage"
+      onClose={onClose}
+      footer={
+        <DashboardModalFooter
+          onCancel={onClose}
+          onSubmit={onSubmit}
+          submitLabel="Save"
+          isSubmitDisabled={!name}
```
DaoDaoNoCode, line 59: Is the submit button disabled status check also implemented in the other ticket?

xianli123:
@jpuzz0 Thanks for doing that!

frontend/src/pages/projects/screens/spawner/storage/ClusterStorageEditModal.tsx
jpuzz0: As shared [here](https://github.com/opendatahub-io/odh-dashboard/pull/3320#discussion_r1801825868), this is just a start to the modal, so yes

jpuzz0:
> Hi, I don't see creation or attach existing PVC button on my spawner page, is there anything wrong with my settings?
> 
> Screenshot 2024-10-18 at 4 19 17 PM

@DaoDaoNoCode 
I commented out the buttons with a comment for now since they're not hooked up to anything yet and could lead to confusion.

DaoDaoNoCode (commented):
Looks good, just found one issue, not sure if it should be fixed in this PR.
When there are 2 storages that have the same mount path, the spawner page won't throw any error.

Screenshot 2024-10-21 at 6 01 25 PM

While the workbench couldn't start because the mount path needs to be unique.

Screenshot 2024-10-21 at 6 03 07 PM

frontend/src/pages/projects/screens/spawner/storage/ClusterStorageTable.tsx
```diff
@@ -0,0 +1,154 @@
...
+  const [isDetachModalOpen, setIsDetachModalOpen] = React.useState(false);
+  const [selectedId, setSelectedId] = React.useState<number>();
+  const selectedStorage = storageData.find((formData) => formData.id === selectedId);
+  const hasUpdatedDefaultNameRef = React.useRef(false);
+
+  const updateStorageData = React.useCallback(
+    (newStorageData: StorageData) => {
+      setStorageData((prevData) =>
+        prevData.map((pvcData, pvcIndex) => {
+          if (pvcIndex === selectedId) {
```
dpanshug, line 34: On line 27, we correctly compare selectedId with the storageData.id. The same comparison logic should be applied here for consistency.

Since `pvcData.id` is the actual unique identifier for the storage data item, it should be matched against selectedId rather than using the array index.

This will ensure that the correct storage data item is updated, regardless of its position in the array.


```suggestion
          if (pvcData.id === selectedId) {
```

frontend/src/pages/projects/screens/spawner/storage/ClusterStorageTable.tsx
```diff
@@ -0,0 +1,154 @@
...
+            </Td>
+          </Tr>
+        )}
+      />
+
+      {isEditModalOpen && selectedStorage && (
+        <ClusterStorageEditModal
+          storageData={selectedStorage}
+          onUpdate={updateStorageData}
+          onClose={() => setIsEditModalOpen(false)}
```
dpanshug, line 138: set the `selectedId` to `undefined` when the modal is closed.

frontend/src/pages/projects/screens/spawner/storage/ClusterStorageTable.tsx
```diff
@@ -0,0 +1,154 @@
...
+      )}
+
+      {isDetachModalOpen && selectedStorage && (
+        <ClusterStorageDetachModal
+          onConfirm={() => {
+            setStorageData((prevData) => prevData.filter((_, pvcIndex) => pvcIndex !== selectedId));
+            setIsDetachModalOpen(false);
+          }}
+          storageName={selectedStorage.name}
+          onClose={() => setIsDetachModalOpen(false)}
```
dpanshug, line 149: Also here 
set the `selectedId` to `undefined` when the modal is closed.

frontend/src/pages/projects/screens/spawner/storage/ClusterStorageTable.tsx
```diff
@@ -0,0 +1,154 @@
...
+          storageData={selectedStorage}
+          onUpdate={updateStorageData}
+          onClose={() => setIsEditModalOpen(false)}
+        />
+      )}
+
+      {isDetachModalOpen && selectedStorage && (
+        <ClusterStorageDetachModal
+          onConfirm={() => {
+            setStorageData((prevData) => prevData.filter((_, pvcIndex) => pvcIndex !== selectedId));
```
dpanshug, line 145:
```suggestion
            setStorageData((prevData) => 
      prevData.filter((storage) => storage.id !== selectedId)
    );
```

dpanshug (changes requested):

frontend/src/pages/projects/screens/spawner/storage/ClusterStorageTable.tsx
jpuzz0: The position in the array is what the ID is, so it wouldn't matter either way.

jpuzz0:
> Looks good, just found one issue, not sure if it should be fixed in this PR. When there are 2 storages that have the same mount path, the spawner page won't throw any error.
> 
> Screenshot 2024-10-21 at 6 01 25 PM
> While the workbench couldn't start because the mount path needs to be unique.
> 
> Screenshot 2024-10-21 at 6 03 07 PM

@DaoDaoNoCode The edit modal is not finalized and there is no form validation yet, so this is not an expectation for this PR. There is a JIRA dedicated to just this edit modal. Again, as stated [here](https://github.com/opendatahub-io/odh-dashboard/pull/3320#discussion_r1801825868) and in the description of this PR, this is just a starting point.

frontend/src/pages/projects/screens/spawner/storage/ClusterStorageTable.tsx
dpanshug: But if we detach a storageData, the subsequent items shift positions, but their id remains unchanged. In that case wouldn't the index will be different from the id?

frontend/src/pages/projects/screens/spawner/storage/ClusterStorageTable.tsx
```diff
@@ -0,0 +1,154 @@
...
+    }
+  }, [workbenchName, storageData, setStorageData]);
+
+  // When the initial storage data has yet to be edited by the user,
+  // prepend the workbench name when one exists.
+  React.useEffect(() => {
+    if (!hasUpdatedDefaultNameRef.current && storageData.length) {
+      updateInitialStorageName();
+    }
+  }, [storageData.length, updateInitialStorageName]);
```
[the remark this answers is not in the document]
jpuzz0: Good catch. I've updated this.

frontend/src/concepts/topology/customNodes/ArtifactTaskNode.tsx
dpanshug, line 1: Is this commit required in this PR?

jpuzz0: No, it certainly is not. I've removed it.

dpanshug (approved):
/lgtm The functionality works as expected according to the Figma design and JIRA requirements.

Gkrumbach07:
Ok so this needs @dpanshug PR to make it able to merge. BC as of right now there is no way to attach an existing PVC and create a new one after one is detached

jpuzz0:
> Ok so this needs @dpanshug PR to make it able to merge. BC as of right now there is no way to attach an existing PVC and create a new one after one is detached

@Gkrumbach07 
We could temporarily comment out the detach-ability. I'm ok with leaving this open though if that's what you'd prefer.

Gkrumbach07:
- it seems like the edit modal does not save the changes
- edit modal opens to custom path every time and parses paths incorrectly sometimes
- validation seems incorrect. it allows you to submit plain spaces. this is only for attach storage
- you can submit empty for standard path but only after you switch between custom and standard once
- cannot detach attached storage or existing storage
- names are allowed to be the same
- this can happen if you attach an existing storage with the same name
- or create a new one with the same name
- storage that has already been created does not disable storage class dropdown

frontend/src/pages/projects/screens/spawner/storage/CreateNewStorageSection.tsx
```diff
@@ -31,8 +31,11 @@ const CreateNewStorageSection = <D extends CreatingStorageObject>({
         <NameDescriptionField
```
jeff-phillips-18, line 31: It would be great to replace this with `K8sNameDescriptionField` to allow the user to manually set the K8s resource name if desired and would work towards [RHOAIENG-14746](https://issues.redhat.com/browse/RHOAIENG-14746)

Gkrumbach07:
Untitled

on edit of a root mount path, it wont let me save because it thinks that the path is taken, but that path is used by the one i am editing

frontend/src/pages/projects/screens/spawner/storage/CreateNewStorageSection.tsx
jpuzz0: There's a lot in-flight with this PR as it is right now, and since the replacement will involve some rework of this PR, I would prefer it be a separate issue if not just a part of the JIRA you've cited here.

Gkrumbach07:
some bugs i found this iteration
- created storage, either first initial one or newly by pressing create storage, will result in a blank path
- you can add the same pvc multiple times
- duplicate storage path is not being validated with all paths in the table

Untitled

Gkrumbach07 (approved):
/lgtm

PR #3324, opened 2024-10-11
Fix misleading label after creating model registry

manaswinidas:
Closes: https://issues.redhat.com/browse/RHOAIENG-12278

## Description

Without this change the MR showed `Unavailable` immediately after creating model registry, when the reason was `ContainerCreating`, which was misleading to the user.

Before:
Screenshot 2024-10-11 at 8 04 06 PM

Screenshot 2024-10-11 at 8 08 40 PM

After:
Screenshot 2024-10-11 at 7 54 42 PM

## How Has This Been Tested?

1. Go to model registry setting page
2. Try creating model registry, if the reason is `ContainerCreating` the status of Model registry shows `Progressing` instead of `Unavailable`.
3. Show `Unavailable` otherwise.

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/pages/modelRegistrySettings/ModelRegistryTableRowStatus.tsx
+++ b/frontend/src/pages/modelRegistrySettings/ModelRegistryTableRowStatus.tsx
@@ -67,7 +67,10 @@ export const ModelRegistryTableRowStatus: React.FC<ModelRegistryTableRowStatusPr
         : [];
 
     // Unavailable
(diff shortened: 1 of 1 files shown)


mturley (changes requested):
@manaswinidas do you mind adding a test case to `frontend/src/pages/modelRegistrySettings/__tests__/ModelRegistryTableRowStatus.spec.tsx` to make sure we show the correct status when we have ContainerCreating?

PR #3332, opened 2024-10-14
Make Label Cursor Hoverable 

YuliaKrimerman:
https://issues.redhat.com/browse/RHOAIENG-13047

## Description

Made the labels indicate they are hoverable in Settings page

https://github.com/user-attachments/assets/31648c76-ee87-4b39-80b8-316f4e594ced

## How Has This Been Tested?

On the UI
## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/pages/modelRegistrySettings/ModelRegistryTableRowStatus.tsx
+++ b/frontend/src/pages/modelRegistrySettings/ModelRegistryTableRowStatus.tsx
@@ -30,7 +30,6 @@ enum ConditionStatus {
   True = 'True',
   False = 'False',
 }
-
 interface ModelRegistryTableRowStatusProps {
   conditions: K8sCondition[] | undefined;
 }
@@ -118,9 +117,13 @@ export const ModelRegistryTableRowStatus: React.FC<ModelRegistryTableRowStatusPr
     }
   }
 
+  const isClickable =
+    statusLabel === ModelRegistryStatusLabel.Unavailable ||
+    statusLabel === ModelRegistryStatusLabel.Degrading;
+
   const label = (
     <Label
-      onClick={() => void 0}
+      {...(isClickable ? { onClick: () => {} } : {})}
       data-testid="model-registry-label"
       icon={icon}
       color={color}
--- a/frontend/src/pages/modelRegistrySettings/ModelRegistryTableRowStatus.tsx
+++ b/frontend/src/pages/modelRegistrySettings/ModelRegistryTableRowStatus.tsx
@@ -119,7 +119,14 @@ export const ModelRegistryTableRowStatus: React.FC<ModelRegistryTableRowStatusPr
   }
 
   const label = (
-    <Label data-testid="model-registry-label" icon={icon} color={color} isCompact>
+    <Label
+      onClick={() => void 0}
+      data-testid="model-registry-label"
+      icon={icon}
+      color={color}
+      isCompact
+      className="pf-m-link"
+    >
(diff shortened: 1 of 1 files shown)


manaswinidas (changes requested):
I think the label should only be hoverable for "Unavailable" and any other labels where we are showing a popover. There are no actions for "Available" and "Progressive" - so no use making it hoverable. Also, I think it's a bad practice to use `void 0`(makes it inaccessible for screen readers- [reference link](https://coreui.io/blog/what-does-javascript-void-0-mean/#:~:text=Accessibility%3A%20Links%20that%20use%20javascript,and%20reserve%20links%20for%20navigation.)) Can we use a button instead? - I think button is hoverable by default.

frontend/src/pages/modelRegistrySettings/ModelRegistryTableRowStatus.tsx
```diff
@@ -118,8 +117,24 @@ export const ModelRegistryTableRowStatus: React.FC<ModelRegistryTableRowStatusPr
...
+  const isClickable =
+    statusLabel === ModelRegistryStatusLabel.Unavailable ||
+    statusLabel === ModelRegistryStatusLabel.Degrading;
+
   const label = (
-    <Label data-testid="model-registry-label" icon={icon} color={color} isCompact>
+    <Label
+      {...(isClickable
+        ? {
+            onClick: () => {
```
manaswinidas, line 128: not sure if we should prefer this over ` style={{ cursor: 'pointer' }}`, inspiration from https://github.com/opendatahub-io/odh-dashboard/pull/3344 - let me know if this works though.

YuliaKrimerman: I tried to add that style but it's not changing anything from what I see - it remains the pointing hand ( like in the picture but just white and slightly bent)
Screenshot 2024-10-17 at 12 29 18 PM

mturley: I think maybe the suggestion was to avoid the noop function, but I think it's probably fine either way - inline styles are also not great 🤷 

@YuliaKrimerman can you just add more detail to the comment in that noop function, something like "Click event is handled by the Popover parent, this prop enables clickable styles in the PatternFly Label"?

mturley (changes requested):

frontend/src/pages/modelRegistrySettings/ModelRegistryTableRowStatus.tsx
```diff
@@ -118,8 +117,24 @@ export const ModelRegistryTableRowStatus: React.FC<ModelRegistryTableRowStatusPr
     }
   }
 
+  const isClickable =
```
mturley, line 120: I think you can have this logic based on whether there's a popover instead of based on status.
```ts
const isClickable = popoverTitle && popoverMessages.length;
```

mturley:
@manaswinidas regarding the `void 0` issue, the way I interpret that document is to say a link shouldn't have `void 0` since links should navigate to a URL, and if it's an action a button is better - in this case we have a button and the click event does "do something" (it bubbles up to the popover parent), the empty onClick function is simply because that's what triggers the PF clickable styles. So I think we're ok on a11y. But changing it from `void 0` to be an empty function with a comment looks more readable to me.

mturley (approved):
LGTM - @manaswinidas if you disagree with the approval let's follow up

PR #3333, opened 2024-10-15
Display internal addresses even when models are external

ppadti:
Closes: [RHOAIENG-7579](https://issues.redhat.com/browse/RHOAIENG-7579)

## Description
This PR aims to display internal addresses even when models are external.

Screenshot 2024-10-15 at 5 03 33 PM

Screenshot 2024-10-15 at 5 03 51 PM

## How Has This Been Tested?
check the inference endpoint:
when external route is enabled: 
  - it should display both internal and external address
  
when external route is not enabled
  - it should only show internal address

## Test Impact
Added cypress tests.

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/__tests__/cypress/cypress/pages/modelServing.ts
+++ b/frontend/src/__tests__/cypress/cypress/pages/modelServing.ts
@@ -311,6 +311,14 @@ class InferenceServiceRow extends TableRow {
   findInternalServicePopover() {
     return cy.findByTestId('internal-service-popover');
   }
+
+  findExternalServiceButton() {
+    return this.find().findByTestId('internal-external-service-button');
+  }
+
+  findExternalServicePopover() {
+    return cy.findByTestId('external-service-popover');
+  }
 }
 class ServingPlatformCard extends Contextual<HTMLElement> {
   findDeployModelButton() {
--- a/frontend/src/__tests__/cypress/cypress/tests/mocked/modelServing/servingRuntimeList.cy.ts
+++ b/frontend/src/__tests__/cypress/cypress/tests/mocked/modelServing/servingRuntimeList.cy.ts
@@ -481,6 +481,10 @@ describe('Serving Runtime List', () => {
           mockInferenceServiceK8sResource({
             name: 'another-inference-service',
             displayName: 'Another Inference Service',
+            statusPredictor: {
+              grpcUrl: 'grpc://modelmesh-serving.app:8033',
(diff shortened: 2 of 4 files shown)


ppadti:
@vconzola Could you please take a look at the above screenshots?

manaswinidas (approved):
/lgtm

Tested on a model with both enabled and only internal address enabled. Works fine

Screenshot 2024-10-15 at 8 22 48 PM

gitdallas (approved):
lgtm other than i wonder if "Internal" should be lowercase in the button "Endpoint details (internal)" and "Endpoint details (internal and external)"

mturley:
Also LGTM aside from @gitdallas 's nitpicks on capitalization @ppadti

yih-wang:
Look into the [PatternFly Capitalization guidelines](https://www.patternfly.org/ux-writing/capitalization) and the relevant rules are:
- Capitalize proper nouns, product names, acronyms, and initialisms. For example: React, PatternFly, and HTML.
- Write all components in lowercase unless they start a sentence.
I guess we won't treat 'internal' and 'external' as proper nouns then I agree with @gitdallas 's comment that internal should be lowercase.
cc @kaedward for confirmation.

mturley (approved):

kaedward:
Do we need to display "(internal and external)" in the button text? Could we just use "Endpoint details"? This is my preference unless we absolutely need to include the types in the button text. 

Alternatively, if we do need to include the types, we could say "Internal and external endpoint details". WDYT?

vconzola:
@kaedward I think the types are important, so a user can see if a model is accessible both internally and externally without having to open the pop-up. I like your idea of saying "Internal endpoint details" or "Internal and external endpoint details" and getting rid of the parentheses.

vconzola:
@ppadti @mturley I see this is already merged. Do we need to open another issue for the text change @kaedward suggested?

mturley:
@ppadti if you want to open a followup PR I can review it tomorrow, but if you don't have time then yeah we should open a separate issue. Sorry @vconzola.

ppadti:
> @ppadti if you want to open a followup PR I can review it tomorrow, but if you don't have time then yeah we should open a separate issue. Sorry @vconzola.

yeah sure, I will open a followup PR for this.

ppadti:
Followup PR: https://github.com/opendatahub-io/odh-dashboard/pull/3351

PR #3335, opened 2024-10-15
Add P0 OOTB connection types

jeff-phillips-18:
[RHOAIENG-13105](https://issues.redhat.com/browse/RHOAIENG-13105)

## Description
Adds yaml files for OOTB connection types. Updates the connection type page to not allow editing of the OOTB connection types. Fixes an issue where the connection types table does not update when new connections are added or modified.

## How Has This Been Tested?
- Run the UI and navigate to Settings -> Connection types
- Install the connection types:
  - Set the project to opendatahub: `oc project opendatahub`
  - oc apply -f manifests/common/connection-types/s3.yaml
  - oc apply -f manifests/common/connection-types/oci-compliant-registry-v1.yaml
  - oc apply -f manifests/common/connection-types/uri-v1.yaml
- Verify the connection types are shown in the UI without needing to refresh the page within 30 seconds

- From the OpenShift Console, go to Administrator -> Home -> Search
  - Select the opendatahub project
  - Select Resource = OdhApplication
  - Find the `jupyter` OdhApplication and copy its `ownerReference` section of the Yaml file
  - Go back to the Search page and Select the ConfigMap Resource
  - Edit the s3, oci-compliant-registry-v1, and uri-v1 configmaps:
    -  Add the copied `ownerReference` section to each of them
  - Check the UI and see that each now has `Pre-installed` for the `creator` column
  - Verify the `Edit` action in the kebab is disabled for each.

## Test Impact
None, test already covered this functionality

## Screen shots

image

image

image

image

## Request review criteria:

/cc @simrandhaliw

### DIFF
--- a/frontend/src/pages/connectionTypes/ConnectionTypesTableRow.tsx
+++ b/frontend/src/pages/connectionTypes/ConnectionTypesTableRow.tsx
@@ -151,6 +151,7 @@ const ConnectionTypesTableRow: React.FC<ConnectionTypesTableRowProps> = ({
             {
               title: 'Edit',
               onClick: () => navigate(`/connectionTypes/edit/${obj.metadata.name}`),
+              isDisabled: ownedByDSC(obj),
             },
             {
               title: 'Duplicate',
@@ -160,6 +161,7 @@ const ConnectionTypesTableRow: React.FC<ConnectionTypesTableRowProps> = ({
             {
               title: 'Delete',
               onClick: () => handleDelete(obj),
+              isDisabled: ownedByDSC(obj),
             },
           ]}
         />
--- a/frontend/src/pages/connectionTypes/manage/EditConnectionTypePage.tsx
+++ b/frontend/src/pages/connectionTypes/manage/EditConnectionTypePage.tsx
@@ -1,13 +1,21 @@
 import * as React from 'react';
 import { useParams } from 'react-router';
+import { useNavigate } from 'react-router-dom';
 import ApplicationsPage from '~/pages/ApplicationsPage';
 import { useConnectionType } from '~/concepts/connectionTypes/useConnectionType';
 import { updateConnectionType } from '~/services/connectionTypesService';
+import { ownedByDSC } from '~/concepts/k8s/utils';
 import ManageConnectionTypePage from './ManageConnectionTypePage';
 
 const EditConnectionTypePage: React.FC = () => {
+  const navigate = useNavigate();
   const { name } = useParams();
   const [existingConnectionType, isLoaded, error] = useConnectionType(name);
+
+  if (existingConnectionType && ownedByDSC(existingConnectionType)) {
+    navigate('/connectionTypes');
+    return null;
+  }
 
   if (!isLoaded || error) {
     return <ApplicationsPage loaded={isLoaded} loadError={error} empty />;
--- a/manifests/common/connection-types/oci-compliant-registry-v1.yaml
+++ b/manifests/common/connection-types/oci-compliant-registry-v1.yaml
@@ -1,0 +1,13 @@
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: oci-compliant-registry-v1
+  labels:
+    opendatahub.io/connection-type: 'true'
+    opendatahub.io/dashboard: 'true'
+  annotations:
+    openshift.io/description: 'Connect to an OCI-compliant container registry, enabling integration with containerized applications and services. Use this connection type to pull and manage container images and artifacts that adhere to the Open Container Initiative (OCI) standards, ensuring compatibility with OCI-compliant tools and workflows.'
+    openshift.io/display-name: OCI compliant registry - v1
+data:
+  category: '["URI"]'
+  fields: '[{"type":"uri","name":"URI","envVar":"URI","required":true,"properties":{}}]'
--- a/manifests/common/connection-types/s3.yaml
+++ b/manifests/common/connection-types/s3.yaml
@@ -1,0 +1,13 @@
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: s3
+  labels:
+    opendatahub.io/connection-type: 'true'
+    opendatahub.io/dashboard: 'true'
+  annotations:
+    openshift.io/description: 'Connect to storage systems that are compatible with Amazon S3, enabling integration with other S3-compatible services and applications.'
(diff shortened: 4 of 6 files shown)


manifests/common/connection-types/oci-compliant-registry-v1.yaml
```diff
@@ -0,0 +1,17 @@
...
+metadata:
+  name: oci-compliant-registry-v1
+  namespace: opendatahub
+  creationTimestamp: null
+  labels:
+    opendatahub.io/connection-type: 'true'
+    opendatahub.io/dashboard: 'true'
+  annotations:
+    opendatahub.io/enabled: 'true'
+    opendatahub.io/username: cluster-admin
```
christianvogt, line 12: I don't believe we should include the namespace, creationTimestamp or username annotation.

christianvogt (commented):
While the `Edit` action is disabled, i can select the duplicate action, then edit the URL from `/duplicate/` to `/edit/` and then I can save any changes. I believe the operator should just revert the change, however I don't think the user should be able to perform this action. The backend should prevent any request to update the preinstalled resources.

The delete action should not be enabled either. Should it be hidden? @simrandhaliw 

Do we know if we should allow the admin to disable these connection types?

jeff-phillips-18:
### Screen shot if user modifies the URL to edit an OOTB connection type

image

frontend/src/pages/connectionTypes/ConnectionTypesTableRow.tsx
```diff
@@ -152,6 +152,7 @@ const ConnectionTypesTableRow: React.FC<ConnectionTypesTableRowProps> = ({
             {
               title: 'Edit',
               onClick: () => navigate(`/connectionTypes/edit/${obj.metadata.name}`),
+              isDisabled: ownedByDSC(obj),
```
christianvogt, line 155: Disable the `Delete` action as well.

christianvogt:
/hold

Since we are not yet releasing this feature, I want to be absolutely sure these resources are correct because I'm also unsure how these resources would be reconciled if we need to make additional changes.

manifests/common/connection-types/s3.yaml
```diff
@@ -0,0 +1,14 @@
...
+  labels:
+    opendatahub.io/connection-type: 'true'
+    opendatahub.io/dashboard: 'true'
+  annotations:
+    opendatahub.io/enabled: 'true'
+    openshift.io/description: 'Connect to storage systems that are compatible with Amazon S3, enabling integration with other S3-compatible services and applications.'
+    openshift.io/display-name: S3 compatible object storage - v1
+data:
+  category: '["Object storage"]'
+  fields: '[{"type":"short-text","name":"Access key","envVar":"AWS_ACCESS_KEY_ID","properties":{},"required":true},{"type":"hidden","name":"Secret key","envVar":"AWS_SECRET_ACCESS_KEY","required":true,"properties":{}},{"type":"short-text","name":"Endpoint","envVar":"AWS_S3_ENDPOINT","required":true,"properties":{}},{"type":"short-text","name":"Region","envVar":"AWS_DEFAULT_REGION","required":false,"properties":{}},{"type":"short-text","name":"Bucket","envVar":"AWS_S3_BUCKET","required":false,"properties":{}},{"type":"short-text","name":"Path","envVar":"AWS_S3_PATH","properties":{}}]'
```
christianvogt, line 14: `path` is not part of data connections today. I do not believe it should be part of the the connection type. Will start a discussion on this.

frontend/src/pages/connectionTypes/manage/EditConnectionTypePage.tsx
```diff
@@ -3,12 +3,24 @@ import { useParams } from 'react-router';
...
+  if (existingConnectionType && ownedByDSC(existingConnectionType)) {
+    return (
+      <ApplicationsPage
+        loaded={isLoaded}
+        errorMessage="Unable to edit"
+        loadError={new Error('This connection type is not editable')}
+        empty
+      />
+    );
+  }
```
christianvogt, line 22: I feel like we should just redirect the user back to `/conectionTypes` instead of displaying this page. The only way they can get here is by direct URL anyways.

manifests/common/connection-types/oci-compliant-registry-v1.yaml
andrewballantyne, line 1: Do we have support for this in the UI yet?

manifests/common/connection-types/s3.yaml
andrewballantyne: Agreed -- we should make sure that field pops up when selecting an S3-esk connection in the Model Serving modal.

andrewballantyne (changes requested):
@christianvogt 

> Since we are not yet releasing this feature, I want to be absolutely sure these resources are correct because I'm also unsure how these resources would be reconciled if we need to make additional changes.

Can we make sure we can do this no matter what -- We'll need to do updates almost guaranteed to these resources at some point on upgrade. Can we not make this a first-class feature to allow for manifests (eg. devflags) to change?

christianvogt:
@andrewballantyne these resources will be updated on upgrade like the other resources. However they aren't reconciled immediately on change in case a privileged user alters them.

Seems like some churn is still happening on the definition side of these resources.

andrewballantyne:
@christianvogt Sorry, I'm not clear what you mean by "a privileged user alters them"... either they are reconciled by the operator or they are not I thought 🤔 If you oc apply these resources, they are not quite the same as if you used devFlags or an operator upgrade (which I believe to be the same if memory serves).

christianvogt:
@andrewballantyne we found out that the resources are not reconciled on change by the operator but only on upgrade or when the operator controller is restarted. So what I was getting at is if a privileged user happens to change the resource, it doesn't get reset until one of those scenarios hit. This is something @lucferbux and I found strange and would need some further looking into if we wanted to change that behavior.

manifests/common/connection-types/oci-compliant-registry-v1.yaml
```diff
@@ -0,0 +1,14 @@
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: oci-compliant-registry-v1
+  labels:
+    opendatahub.io/connection-type: 'true'
+    opendatahub.io/dashboard: 'true'
+  annotations:
+    opendatahub.io/enabled: 'true'
```
christianvogt, line 9: Remove all `opendatahub.io/enabled` annotations.

PR #3338, opened 2024-10-15
 Rename Gi/Mi size names to GiB/MiB

jenny-s51:
https://issues.redhat.com/browse/RHOAIENG-291

## Description

Use `GiB` and `MiB` as the visual representation in the UI while keeping the current usage of `Gi` and `Mi` under the hood.

Before:
modal-before

Screenshot 2024-10-15 at 3 31 50 PM

After:
Screenshot 2024-10-15 at 3 44 39 PM

Screenshot 2024-10-15 at 3 16 58 PM

## How Has This Been Tested?

Tested on Model Serving cluster.
1. Create a DS project. 
2. Add a workbench.
3. Go to the "Cluster Storage" tab in the selected DS project. 
4. Click "Add storage".
5. Verify "Persistent Storage Size" units are GiB/MiB.
6. For an existing cluster storage, go to kebab menu on table row. 
8. Click "Edit storage".
9. Verify "Persistent Storage Size" units are GiB/MiB.
10. Go to "Workbenches" in a Data Science Project.
11. Expand a table row.
12. Verify units shown in table are GiB/MiB.

## Test Impact

Tests already exist, units have been updated.

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/pages/projects/screens/detail/notebooks/NotebookSizeDetails.tsx
+++ b/frontend/src/pages/projects/screens/detail/notebooks/NotebookSizeDetails.tsx
@@ -21,13 +21,13 @@ const NotebookSizeDetails: React.FC<NotebookSizeDetailsProps> = ({ notebookSize 
       <DescriptionListGroup>
         <DescriptionListTerm>Limits</DescriptionListTerm>
         <DescriptionListDescription>
-          {limits?.cpu ?? 'Unknown'} CPU, {limits?.memory ?? 'Unknown'} Memory
+          {limits?.cpu ?? 'Unknown'} CPU, {limits?.memory + 'B' ?? 'Unknown'} Memory
         </DescriptionListDescription>
       </DescriptionListGroup>
       <DescriptionListGroup>
         <DescriptionListTerm>Requests</DescriptionListTerm>
         <DescriptionListDescription>
-          {requests?.cpu ?? 'Unknown'} CPU, {requests?.memory ?? 'Unknown'} Memory
+          {requests?.cpu ?? 'Unknown'} CPU, {requests?.memory + 'B' ?? 'Unknown'} Memory
         </DescriptionListDescription>
       </DescriptionListGroup>
     </DescriptionList>
--- a/frontend/src/pages/projects/screens/detail/notebooks/NotebookSizeDetails.tsx
+++ b/frontend/src/pages/projects/screens/detail/notebooks/NotebookSizeDetails.tsx
@@ -16,26 +16,28 @@ const NotebookSizeDetails: React.FC<NotebookSizeDetailsProps> = ({ notebookSize 
     resources: { requests, limits },
   } = notebookSize;
 
-  function formatMemory(memory: string | undefined): string {
-    if (!memory) return 'Unknown';
+  const formatMemory = (memory: string | undefined): string => {
+    if (!memory) {
+      return 'Unknown';
+    }
     if (/(Gi|Mi)$/.test(memory)) {
-        return memory + "B";
+      return `${memory}B`;
     }
     return memory;
-}
+  };
 
   return (
     <DescriptionList>
       <DescriptionListGroup>
         <DescriptionListTerm>Limits</DescriptionListTerm>
         <DescriptionListDescription>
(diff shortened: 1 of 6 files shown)


jenny-s51 (commented):
Updated based on your suggestions @jeff-phillips-18 - would be great to get your thoughts on this approach as well @christianvogt.

frontend/src/pages/projects/screens/detail/notebooks/NotebookSizeDetails.tsx
```diff
@@ -16,18 +16,28 @@ const NotebookSizeDetails: React.FC<NotebookSizeDetailsProps> = ({ notebookSize
...
 
+  const formatMemory = (memory: string | undefined): string => {
+    if (!memory) {
+      return 'Unknown';
+    }
+    if (/(Gi|Mi)$/.test(memory)) {
+      return `${memory}B`;
+    }
+    return memory;
+  };
```
christianvogt, line 27: I suggest creating and making use of a reusable function in `valueUnits` which will convert any string in similar fashion to `splitValueUnit`:

```suggestion
const formatMemory = (value: ValueUnitString): ValueUnitString => {
  const match = value.match(/^(\d*\.?\d*)(.*)$/);
  if (!(match && match[1] && match[2])) {
    return value;
  }
  return `${match[1]}${MEMORY_UNITS_FOR_PARSING.find((o) => o.unit === match[2])?.name || match[2]}`;
}
```

jenny-s51: Thank you @christianvogt for this code suggestion - added `formatMemory` as a reusable function to `valueUnits`.

frontend/src/utilities/valueUnits.ts
```diff
@@ -155,3 +155,17 @@ export const isMemoryLimitLarger = (
 
   return isLarger(limitMemory, requestMemory, MEMORY_UNITS_FOR_PARSING, isEqualOkay);
 };
+
+export const formatMemory = (value?: string | ValueUnitString): ValueUnitString => {
+  if (!value) {
+    return 'Unknown';
+  }
```
christianvogt, line 162: As a utility function, I prefer to not impose an opinionated return value. The consumer can choose to render an alternative string in place of empty or undefined.
```suggestion
export const formatMemory = (value?: ValueUnitString): ValueUnitString => {
  if (!value) {
    return value;
  }
```

frontend/src/utilities/valueUnits.ts
```diff
@@ -155,3 +155,17 @@ export const isMemoryLimitLarger = (
 
   return isLarger(limitMemory, requestMemory, MEMORY_UNITS_FOR_PARSING, isEqualOkay);
 };
+
+export const formatMemory = (value?: ValueUnitString): ValueUnitString | undefined => {
+  if (!value) {
+    return;
+  }
```
christianvogt, line 162: If the value is undefined or an empty string, return itself.
```suggestion
  if (!value) {
    return value
  }
```

jenny-s51: Thanks @christianvogt ! Updated

frontend/src/pages/projects/utils.ts
```diff
@@ -5,8 +5,10 @@ import { NotebookState } from './notebook/types';
 export const getNotebookStatusPriority = (notebookState: NotebookState): number =>
   notebookState.isRunning ? 1 : notebookState.isStarting ? 2 : 3;
 
-export const getPvcTotalSize = (pvc: PersistentVolumeClaimKind): string =>
-  pvc.status?.capacity?.storage || pvc.spec.resources.requests.storage;
+export const getPvcTotalSize = (pvc: PersistentVolumeClaimKind): string => {
+  const storage = pvc.status?.capacity?.storage || pvc.spec.resources.requests.storage;
+  return /(Gi|Mi)$/.test(storage) ? `${storage}B` : storage;
+};
```
christianvogt, line 11: @jenny-s51 why doesn't this use the utility function?

jenny-s51: Thank you @christianvogt - applying the utility function here now

PR #3344, opened 2024-10-17
Change models tooltip to popover

ashley-o0o:
Closes: [RHOAIENG-10992](https://issues.redhat.com/browse/RHOAIENG-10992)

## Description
Updates the data science projects models page to use a popover instead of a tooltip for the status. The message will no longer disappear when trying to select the text, and removes the issue of the screen shaking if the message is too long.

## How Has This Been Tested?
Tested locally

## Test Impact
Adjusted cypress tests to check for popover instead of tooltip

## Screenshots 
Before (hover for tooltip):
Screenshot 2024-10-16 at 3 47 40 PM

After (click for popover)"
Screenshot 2024-10-16 at 3 47 13 PM

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/pages/modelServing/screens/global/InferenceServiceStatus.tsx
+++ b/frontend/src/pages/modelServing/screens/global/InferenceServiceStatus.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { Text, Icon, Tooltip } from '@patternfly/react-core';
+import { Text, Icon, Popover } from '@patternfly/react-core';
 import {
   CheckCircleIcon,
   ExclamationCircleIcon,
@@ -45,6 +45,7 @@ const InferenceServiceStatus: React.FC<InferenceServiceStatusProps> = ({
             isInline
             tabIndex={0}
             iconSize={iconSize}
+            style={{ cursor: 'pointer' }}
           >
             <CheckCircleIcon />
           </Icon>
@@ -59,6 +60,7 @@ const InferenceServiceStatus: React.FC<InferenceServiceStatusProps> = ({
             isInline
             tabIndex={0}
             iconSize={iconSize}
+            style={{ cursor: 'pointer' }}
           >
             <ExclamationCircleIcon />
           </Icon>
@@ -80,6 +82,7 @@ const InferenceServiceStatus: React.FC<InferenceServiceStatusProps> = ({
             isInline
             tabIndex={0}
             iconSize={iconSize}
+            style={{ cursor: 'pointer' }}
           >
             <OutlinedQuestionCircleIcon />
           </Icon>
@@ -94,6 +97,7 @@ const InferenceServiceStatus: React.FC<InferenceServiceStatusProps> = ({
             isInline
             tabIndex={0}
             iconSize={iconSize}
+            style={{ cursor: 'pointer' }}
           >
             <OutlinedQuestionCircleIcon />
           </Icon>
@@ -102,10 +106,9 @@ const InferenceServiceStatus: React.FC<InferenceServiceStatusProps> = ({
   };
 
   return (
-    <Tooltip
-      role="none"
-      data-testid="model-status-tooltip"
-      content={
+    <Popover
+      position="right"
+      bodyContent={
         modelStatus?.failedToSchedule ? (
           <Text>Insufficient resources</Text>
         ) : (
@@ -114,7 +117,7 @@ const InferenceServiceStatus: React.FC<InferenceServiceStatusProps> = ({
       }
     >
       {statusIcon()}
-    </Tooltip>
+    </Popover>
   );
 };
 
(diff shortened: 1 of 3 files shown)


frontend/src/pages/modelServing/screens/global/InferenceServiceStatus.tsx
```diff
@@ -73,27 +75,29 @@ const InferenceServiceStatus: React.FC<InferenceServiceStatusProps> = ({
...
           <Icon
-            data-testid="status-tooltip"
+            data-testid="status-popover"
             role="button"
             aria-label="warning icon"
             status="warning"
             isInline
             tabIndex={0}
             iconSize={iconSize}
+            style={{ cursor: 'pointer' }}
```
christianvogt, line 100: You should use a PF `<Button variant="link" isInline icon={...} />` as a child of the `Popover` to create a semantic element that can gain focus and has the correct cursor by default.

frontend/src/pages/modelServing/screens/global/InferenceServiceStatus.tsx
```diff
@@ -73,27 +75,29 @@ const InferenceServiceStatus: React.FC<InferenceServiceStatusProps> = ({
...
           >
             <OutlinedQuestionCircleIcon />
           </Icon>
         );
       default:
         return (
           <Icon
-            data-testid="status-tooltip"
+            data-testid="status-popover"
```
christianvogt, line 93: QE doesn't want us to change test ids unless necessary. It's fine to keep the old ids here.

frontend/src/pages/modelServing/screens/global/InferenceServiceStatus.tsx
```diff
@@ -102,10 +106,10 @@ const InferenceServiceStatus: React.FC<InferenceServiceStatusProps> = ({
   };
 
   return (
-    <Tooltip
-      role="none"
-      data-testid="model-status-tooltip"
-      content={
+    <Popover
+      data-testid="model-status-popover"
+      position="right"
```
christianvogt, line 111: Why default the position to the right? makes more sense to default it upwards as to not overlap the row.

frontend/src/pages/modelServing/screens/global/InferenceServiceStatus.tsx
```diff
@@ -102,10 +106,10 @@ const InferenceServiceStatus: React.FC<InferenceServiceStatusProps> = ({
   };
 
   return (
-    <Tooltip
-      role="none"
-      data-testid="model-status-tooltip"
-      content={
+    <Popover
```
christianvogt, line 109: It's possible to make PF popovers trigger on `hover`, do we know if UX wants to move to on click? or keep hover like a tooltip?

ashley-o0o: The figma makes it seem that they want a click. One of the issues of this bug was with the hover, when you went to select the text, moving the mouse away from the status caused the message to disappear. I'm unsure if that would be an issue with the popover as well though

frontend/src/pages/modelServing/screens/global/InferenceServiceStatus.tsx
```diff
@@ -102,19 +102,19 @@ const InferenceServiceStatus: React.FC<InferenceServiceStatusProps> = ({
...
-      role="none"
+    <Popover
       data-testid="model-status-tooltip"
-      content={
+      position="top"
+      bodyContent={
         modelStatus?.failedToSchedule ? (
           <Text>Insufficient resources</Text>
         ) : (
           <Text>{getInferenceServiceStatusMessage(inferenceService)}</Text>
```
jenny-s51, line 112: Not a blocker but currently this causes an empty popover to render for the pending states. Wondering if we can add some logic here to prevent rendering a popover in those cases? This was an existing bug with the popovers as well.

Before:
Screenshot 2024-10-17 at 4 27 11 PM

After:
Screenshot 2024-10-17 at 4 23 55 PM

jenny-s51 (commented):
Models popover is looking awesome for error states @ashley-o0o!  Left a comment below about the progress states .

frontend/src/pages/modelServing/screens/global/InferenceServiceStatus.tsx
christianvogt: The UX has some bigger change requests for the contents.
For now I suggest to simply pre-compute the body content and then omit the popover or control its visibility by `isVisible={bodyContent ? undefined : false}`.

frontend/src/pages/modelServing/screens/global/InferenceServiceStatus.tsx
```diff
@@ -101,20 +101,21 @@ const InferenceServiceStatus: React.FC<InferenceServiceStatusProps> = ({
     }
   };
 
+  const bodyContent = modelStatus?.failedToSchedule ? (
+    <Text>Insufficient resources</Text>
+  ) : (
+    <Text>{getInferenceServiceStatusMessage(inferenceService)}</Text>
+  );
```
christianvogt, line 108: You cannot include `Text` in both paths if you want to use this later for a truthy check because otherwise it will always be truthy. Capture just the data here instead:
```suggestion
  const bodyContent = modelStatus?.failedToSchedule ? (
    'Insufficient resources'
  ) : (
    getInferenceServiceStatusMessage(inferenceService)
  );
```

PR #3345, opened 2024-10-17
Update one-option Select UX

DaoDaoNoCode:
JIRA: [RHOAIENG-10833](https://issues.redhat.com/browse/RHOAIENG-10833)

## Description

Update one-option dropdown UX as mentioned in the JIRA ticket.

I created a [doc](https://docs.google.com/document/d/1tHpzrmqDCEuEdmkqoPscCCm-oYFtUDCV8QwcvebwhA0/edit) to track the effort in case I missed anything, please refer to it.

There will be several commits in this PR to make it easier for the reviewers:
- commit 1: Update the `SimpleSelect` component to make it default to be disabled when there is only one or fewer options in the dropdown and auto-select that option. Also, make the description to be previewed under the dropdown toggle and truncate it to 2 lines if it's too long. Also, updated some usages in the code, please refer to the doc
- commit 2: Replace some use cases of using `Select` component directly with the `SimpleSelect` generic component
- commit 3: Update all the tests, which will ensure the simple select is selected and disabled when there is only one option
- commit 4: Update the linter settings to prevent users from directly importing `Select`. Also, update all the rest use cases, either disable the linter rule to explain why or update the `Select` with `MultiSelection` or `SimpleSelect`.

## How Has This Been Tested?

Make sure nothing is broken and all the single selects are working well.

## Test Impact

See the description of commit 3.

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/components/SimpleSelect.tsx
+++ b/frontend/src/components/SimpleSelect.tsx
@@ -2,6 +2,7 @@ import * as React from 'react';
 import {
   Truncate,
   MenuToggle,
+  // eslint-disable-next-line no-restricted-imports
   Select,
   SelectList,
   SelectOption,
@@ -23,6 +24,7 @@ export type SimpleSelectOption = {
   dropdownLabel?: React.ReactNode;
   isPlaceholder?: boolean;
   isDisabled?: boolean;
+  isFavorited?: boolean;
   dataTestId?: string;
 };
 
@@ -91,6 +93,7 @@ const SimpleSelect: React.FC<SimpleSelectProps> = ({
     if (totalOptionsKey) {
       onChange(totalOptionsKey, false);
     }
+    // We don't want the callback function to be a dependency
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [totalOptionsKey]);
 
@@ -137,6 +140,7 @@ const SimpleSelect: React.FC<SimpleSelectProps> = ({
                     label,
                     dropdownLabel,
                     description,
+                    isFavorited,
                     isDisabled: optionDisabled,
                     dataTestId: optionDataTestId,
                   }) => (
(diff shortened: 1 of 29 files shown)


frontend/src/components/SimpleSelect.tsx
```diff
@@ -56,88 +62,135 @@ const SimpleSelect: React.FC<SimpleSelectProps> = ({
...
+    [options, groupedOptionsFlat],
+  );
+
+  // If there is only one option, call the onChange function
+  const totalOptionsKey = totalOptions.length === 1 ? totalOptions[0].key : null;
+  React.useEffect(() => {
+    if (totalOptionsKey) {
+      onChange(totalOptionsKey, false);
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
jpuzz0, line 94: Maybe the onChange can be memoized here so that we don't have to disable the lint rule here.

DaoDaoNoCode: Do you mean we should do `React.useCallback` on `onChange` here and make it depend on the empty array? I am not sure whether it's working here but I can try. I think disable the linter rule does no harm here, though.

christianvogt: I would say this is the type of use case where disabling the lint rule is expected because we do not want the callback to be a dependency. 

Note that while disabling rules is discouraged, there are times when it's necessary. Please always add a comment as to why the rule is disabled when doing so.

jpuzz0: That's fair. I'll consider this a nitpick then, with the expectation of a possible comment explaining why the lint rule is disabled.

jpuzz0 (approved):

DaoDaoNoCode:
/unhold
Linter rule added and PR is ready for review now.

Gkrumbach07 (approved):
/lgtm

PR #3359, opened 2024-10-22
Add feature flag for configuration parameters in Kserve modal

manaswinidas:
Part of [RHOAIENG-14688](https://issues.redhat.com/browse/RHOAIENG-14688)

## Description

Adding feature flag `disableServingRuntimeParams` to add "Configuration parameters" below "Source model location" `FormSection` to `ManageKserveModal` to unblock a part of the above ticket as well as [RHOAIENG-14689](https://issues.redhat.com/browse/RHOAIENG-14689) 

## How Has This Been Tested?

1. Scale down `odh-dashboard` deployment pod.
2. Add the feature flag to `OdhDashboardConfig` CRD and to `odh-dashboard-config` in `opendatahub` namespace. You can switch the feature flag to false for testing this PR.
3. Switch `disableServingRuntimeParams` `true` and `false` and check whether you see "Configuration parameters" in the Kserve modal. You should see it when the feature flag is `false` and vice-versa

## Test Impact

Will add tests in a follow-up PR that adds fields within this FormSection

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/concepts/areas/const.ts
+++ b/frontend/src/concepts/areas/const.ts
@@ -26,6 +26,7 @@ export const allFeatureFlags: string[] = Object.keys({
   disableAcceleratorProfiles: false,
   disableDistributedWorkloads: false,
   disableModelRegistry: false,
+  disableServingRuntimeParams: false,
   disableConnectionTypes: false,
   disableStorageClasses: false,
   disableNIMModelServing: true,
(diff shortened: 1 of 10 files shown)


gitdallas (approved):

frontend/src/concepts/areas/const.ts
```diff
@@ -115,6 +116,12 @@ export const SupportedAreasStateMap: SupportedAreasState = {
     requiredComponents: [StackComponent.MODEL_REGISTRY],
     requiredCapabilities: [StackCapability.SERVICE_MESH, StackCapability.SERVICE_MESH_AUTHZ],
   },
+  [SupportedArea.SERVING_RUNTIME_PARAMS]: {
+    featureFlags: ['disableServingRuntimeParams'],
+    requiredComponents: [StackComponent.K_SERVE],
+    reliantAreas: [SupportedArea.MODEL_SERVING],
+    requiredCapabilities: [StackCapability.SERVICE_MESH, StackCapability.SERVICE_MESH_AUTHZ],
```
mturley, line 123: I'm not sure this does actually rely on servicemesh.. is that something I missed in refinement? If we're allowing Kserve model deployment we should be able to support these params. I think this area has the same requirements as `disableKServeMetrics`, so it can probably just have `reliantAreas: [SupportedArea.K_SERVE, SupportedArea.MODEL_SERVING],` and no `requiredComponents` or `requiredCapabilities`

manaswinidas: Oh yes, I forgot to comment... I needed help with the changes in this file. Thanks. This is addressed.

Force-pushed coz there was a typo in the commit message.

mturley (approved):
LGTM, thanks @manaswinidas !

PR #3107, opened 2024-08-21
RHOAIENG-9480 Update useTrackUser for Developer Sandbox

pilhuhn:
Clean version of #3093, closes #3093
Closes https://issues.redhat.com/browse/RHOAIENG-9480

## Description
In cases where we have a sso-user-id , supply this to segment tracking instead of the anonymous one.

## How Has This Been Tested?
So far.. manual inspection only.

## Test Impact

Manual testing via inspection of browser console

## Request review criteria:
Self checklist (all need to be checked):
- [X] The developer has manually tested the changes and verified that the changes work
- [X] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [ ] The developer has added tests or explained why testing cannot be added (unit or cypress tests for related changes)

After the PR is posted & before it merges:
- [ ] The developer has tested their solution on a cluster by using the image produced by the PR to `main`

cc @christianvogt

PR #3362, opened 2024-10-22
RHOAIENG-14811 web-user-id for Sandbox is not correctly set.

pilhuhn:
Closes https://issues.redhat.com/browse/RHOAIENG-14811
We now get the userID from the status api endpoint if set. 
This reverts parts of RHOAIENG-9480/PR #3107

## Description

The previous code assumes that the (sso/web-)userId is passed in an annotation in the 
get user object. This seems not be true in the real world. The underlying Kube is exposing
that id in the response to the /status call

The code reverts the try to fetch from the annotation and instead takes it from the Kube
object if present .

## How Has This Been Tested?

Manual test in dev-mode running `npm run start:dev:ext` against an instance on Dev-Sandbox and a non-Sandbox one.

Then looking at the console output for the the Identify event

```
Identify event triggered: {... "userID":"b09067a69...99"}
```

If the userID looks like this (whith chars) it is an anonymous one. If it is numeric (may be negative) and short (probably < 8 digits), then it is a web-user-id. 

## Request review criteria:

cc @andrewballantyne

### DIFF
--- a/frontend/src/concepts/analyticsTracking/useTrackUser.ts
+++ b/frontend/src/concepts/analyticsTracking/useTrackUser.ts
@@ -4,12 +4,9 @@ import { useAccessReview } from '~/api';
 import { AccessReviewResourceAttributes } from '~/k8sTypes';
 import { IdentifyEventProperties } from '~/concepts/analyticsTracking/trackingProperties';
 
-export const useTrackUser = (
-  username?: string,
-  ssoUserID?: string,
-): [IdentifyEventProperties, boolean] => {
-  const { isAdmin } = useUser();
-  const [userID, setUserID] = React.useState<string | undefined>(ssoUserID);
+export const useTrackUser = (username?: string): [IdentifyEventProperties, boolean] => {
+  const { isAdmin, userID } = useUser();
+  const [finalUserID, setUserID] = React.useState<string | undefined>(userID);
 
   const createReviewResource: AccessReviewResourceAttributes = {
     group: 'project.openshift.io',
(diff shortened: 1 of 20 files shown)


frontend/src/concepts/analyticsTracking/useTrackUser.ts
```diff
@@ -4,12 +4,10 @@ import { useAccessReview } from '~/api';
...
 import { IdentifyEventProperties } from '~/concepts/analyticsTracking/trackingProperties';
 
-export const useTrackUser = (
-  username?: string,
-  ssoUserID?: string,
-): [IdentifyEventProperties, boolean] => {
+export const useTrackUser = (username?: string): [IdentifyEventProperties, boolean] => {
   const { isAdmin } = useUser();
-  const [userID, setUserID] = React.useState<string | undefined>(ssoUserID);
+  const { userID } = useUser();
```
christianvogt, line 9: You can deconstruct this value from the line above:
```
const { isAdmin, userID } = useUser();
```

That being said, where is `userID` coming from. It seems to me that this value is removed from the backend response.

pilhuhn: `userID` comes via the /api/status call if the backend is sso-enabled.   `useDetectUser()` -> `getUserFulfilled()`

If the cluster has the web-user-id, we get it in the status response like this:

Bildschirmfoto 2024-10-18 um 15 15 48 Kopie

christianvogt: @pilhuhn I don't see the userID in the response when testing. Make sure you're running your backend changes.

Your change to `backend/src/routes/api/status/statusUtils.ts` specifically removes the `userID` from the object.

pilhuhn:
Superseded by #3387

PR #3365, opened 2024-10-23
fix validation error markers when dnd field row

christianvogt:
https://issues.redhat.com/browse/RHOAIENG-13664

## Description

If multiple connection type fields have the same env var name, then an error will appear in the field row adjacent to the env var name. During DnD of the fields, the error marker incorrectly jumps to another row because while dragging the index of the field is changing however the index of the validation error is not. Therefore the two get out of sync.

The fix is to compute the index of the row from the original field set so that the index aligns with the indicies of the validation error.

## How Has This Been Tested?

- enable connection types feature using the `disableConnectionTypes` feature flag
- navigate to `Settings -> Connection types`
- create a connection type
- add multiple fields and sections
- add two fields with the same env var name
- observe the two fields have error markers in the field row
- drag various fields to reorder

## Test Impact

N/A

## Request review criteria:

If you have UI changes: 


jeff-phillips-18 (approved):
Works well.

PR #3366, opened 2024-10-23
enable edit form submit button only when form is dirty

christianvogt:
https://issues.redhat.com/browse/RHOAIENG-14750

## Description

Updates the connection type form, along with section and data field modals. When entering these forms through an edit scenario, the form submit button will remain disabled until the form becomes dirty. Dirty is a simple computation in that as soon as any value changes, the form is considered dirty. However returning back to the original value will still be considered dirty.

## How Has This Been Tested?

- enable connection types feature flag 'disableConnectionTypes'
- navigate to `Settings -> Connection types'

Connection type:
- create a connection type
- observe the submit button is disabled until the form is valid
- complete the connection type and save
- edit the connection type
- observe the submit button is disabled
- modify any value or add a section or field
- observe the submit button is enabled
- save the changes
- duplicate the connection type
- observe the submit button is enabled

Section:
- while editing or creating a connection type
- add a new section
- observe the submit button is disabled until the form is valid
- save the section
- edit the section
- observe the submit button is disabled
- modify any value 
- observe the submit button is enabled
- save the changes
- duplicate the section
- observe the submit button is enabled

Data field:
- while editing or creating a connection type
- add a new field
- observe the submit button is disabled until the form is valid
- save the field
- edit the field
- observe the submit button is disabled
- modify any value 
- observe the submit button is enabled
- save the changes
- duplicate the field
- observe the submit button is enabled

## Test Impact

Updated existing tests and added new tests.

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/__mocks__/mockConnectionType.ts
+++ b/frontend/src/__mocks__/mockConnectionType.ts
@@ -308,7 +308,9 @@ const mockFields: ConnectionTypeField[] = [
     description: 'Test boolean',
     envVar: 'boolean_1',
     required: false,
-    properties: {},
+    properties: {
+      label: 'Input label',
+    },
   },
   {
     type: 'boolean',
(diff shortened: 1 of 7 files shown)


jeff-phillips-18 (approved):
Works well.

PR #3368, opened 2024-10-23
Deleting workbench resource makes workbench editable

pnaik1:
https://issues.redhat.com/browse/RHOAIENG-1891

## Description

- If a cluster admin deletes secrets or config maps that a Notebook resource (workbench) uses, then editing notebook will display a alert warning stating which secret or configmap is deleted 
- if you click on the update workbench, the notebook will be updated successfully without any error and the secrets/ configmaps will be removed from envFrom
- You can even delete the notebook which has secrets/configMaps deletd by clusteradmin successfully without any error

## How Has This Been Tested?

1. Create a workbench with secret or configMap env variables and 
2. Delete one of the env variables from console
3. Stop the workbench and edit the workbench
4. check the workbench with alert warning
5. check whether the workbench updates successfully
6. delete the workbench which has has secrets/configMaps deletd by clusteradmin and check if its deleted successfully

Screenshot 2024-10-23 at 5 31 38 PM

Screenshot 2024-10-23 at 5 29 03 PM

## Test Impact
added cypress test

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/pages/projects/screens/spawner/SpawnerPage.tsx
+++ b/frontend/src/pages/projects/screens/spawner/SpawnerPage.tsx
@@ -50,6 +50,7 @@ import useDefaultStorageClass from './storage/useDefaultStorageClass';
 import usePreferredStorageClass from './storage/usePreferredStorageClass';
 import { ConnectionsFormSection } from './connections/ConnectionsFormSection';
 import { getConnectionsFromNotebook } from './connections/utils';
+import AlertWarningText from './environmentVariables/AlertWarningText';
 
 type SpawnerPageProps = {
   existingNotebook?: NotebookKind;
@@ -90,7 +91,9 @@ const SpawnerPage: React.FC<SpawnerPageProps> = ({ existingNotebook }) => {
     defaultStorageClassName,
   );
 
-  const [envVariables, setEnvVariables] = useNotebookEnvVariables(existingNotebook);
+  const [envVariables, setEnvVariables, envVariablesLoaded, deletedConfigMaps, deletedSecrets] =
+    useNotebookEnvVariables(existingNotebook);
+
   const [dataConnectionData, setDataConnectionData] = useNotebookDataConnection(
     dataConnections.data,
     existingNotebook,
@@ -216,6 +219,12 @@ const SpawnerPage: React.FC<SpawnerPageProps> = ({ existingNotebook }) => {
               id={SpawnerPageSectionID.ENVIRONMENT_VARIABLES}
               aria-label={SpawnerPageSectionTitles[SpawnerPageSectionID.ENVIRONMENT_VARIABLES]}
             >
+              {envVariablesLoaded && (
+                <AlertWarningText
+                  deletedConfigMaps={deletedConfigMaps}
+                  deletedSecrets={deletedSecrets}
+                />
+              )}
               <EnvironmentVariables envVariables={envVariables} setEnvVariables={setEnvVariables} />
             </FormSection>
             <FormSection
--- a/frontend/src/pages/projects/screens/spawner/environmentVariables/AlertWarningText.tsx
+++ b/frontend/src/pages/projects/screens/spawner/environmentVariables/AlertWarningText.tsx
@@ -1,0 +1,91 @@
+import { Alert, List, ListItem, Stack, StackItem } from '@patternfly/react-core';
+import React from 'react';
+
+type AlertWarningTextProps = {
+  deletedConfigMaps: string[];
+  deletedSecrets: string[];
+};
+
+const AlertWarningText: React.FC<AlertWarningTextProps> = ({
+  deletedConfigMaps,
+  deletedSecrets,
+}) => {
+  if (deletedConfigMaps.length === 0 && deletedSecrets.length === 0) {
+    return null;
+  }
+
+  const formattedText = (items: string[], type: 'config map' | 'secret') => {
+    if (items.length === 1) {
+      return (
+        <>
+          <strong>{items[0]}</strong> {type}
+        </>
+      );
+    }
+    const lastItem = items[items.length - 1];
+    const displayItems = items.slice(0, -1);
+    return (
+      <>
+        {displayItems.map((item, index) => (
+          <>
+            <strong key={index}>{item}</strong>
+            {index < displayItems.length - 1 ? ', ' : ' and '}
+          </>
+        ))}
+        <strong>{lastItem}</strong> {type}s
+      </>
+    );
+  };
+
+  let description;
+  const commonDescriptionText = `To remove the orphaned connection, save the workbench.`;
+  const formattedConfigMapsText = formattedText(deletedConfigMaps, 'config map');
+  const formattedSecretsText = formattedText(deletedSecrets, 'secret');
+
+  if (deletedConfigMaps.length > 0 && deletedSecrets.length > 0) {
+    description = (
+      <Stack hasGutter>
+        <StackItem>
+          The following environment variables in this workbench cannot be found:
+        </StackItem>
+        <StackItem>
+          <List>
+            {[formattedConfigMapsText, formattedSecretsText].map((text, i) => (
+              <ListItem key={i}>{text}</ListItem>
+            ))}
+          </List>
+        </StackItem>
+        <StackItem>{commonDescriptionText}</StackItem>
+      </Stack>
+    );
+  } else if (deletedConfigMaps.length > 0) {
+    description = (
+      <>
+        The {formattedConfigMapsText}
+        {deletedConfigMaps.length === 1 ? ' cannot be found. ' : ' are not found. '}
+        {commonDescriptionText}
+      </>
+    );
+  } else {
+    description = (
+      <>
+        The {formattedSecretsText}
+        {deletedSecrets.length === 1 ? ' cannot be found. ' : ' are not found. '}
+        {commonDescriptionText}
+      </>
+    );
+  }
+
+  return (
+    <Alert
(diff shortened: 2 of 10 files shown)


frontend/src/pages/projects/screens/spawner/environmentVariables/AlertWarningText.tsx
```diff
@@ -0,0 +1,98 @@
...
+        The {formattedText(deletedSecret, 'secret')}
+        {deletedSecret.length === 1 ? ' cannot be found. ' : ' are not found. '}
+        {commonDescriptionText}
+      </>
+    );
+  }
+
+  return (
+    <>
+      {loaded && (
```
DaoDaoNoCode, line 84: Suggestion: It's better to early exit at the top:
```
if (!loaded) {
  return null;
}
...
```
Then you only need to focus on the logic below and return `Alert` here

frontend/src/pages/projects/screens/spawner/environmentVariables/useNotebookEnvVariables.tsx
```diff
@@ -61,17 +75,26 @@ export const fetchNotebookEnvVariables = (notebook: NotebookKind): Promise<EnvVa
 
 export const useNotebookEnvVariables = (
   notebook?: NotebookKind,
-): [envVariables: EnvVariable[], setEnvVariables: (envVars: EnvVariable[]) => void] => {
+): [
+  envVariables: EnvVariable[],
+  setEnvVariables: (envVars: EnvVariable[]) => void,
+  EnvVariableLoaded: boolean,
```
DaoDaoNoCode, line 81: Nit:
```suggestion
  envVariableLoaded: boolean,
```

DaoDaoNoCode (commented):
Left some comments, code lgtm overall. I still need to test it on a live cluster.

frontend/src/pages/projects/screens/spawner/environmentVariables/AlertWarningText.tsx
```diff
@@ -0,0 +1,98 @@
+import { Alert, List, ListItem, Stack, StackItem } from '@patternfly/react-core';
+import React from 'react';
+
+type AlertWarningTextProps = {
+  deletedConfigMap: string[];
+  deletedSecret: string[];
```
jpuzz0, line 6: nit: Wondering if this component needs the context of whether the resources are "deleted" or not. Maybe just calling these `configMap` and `secret` would suffice.

jpuzz0: Also, should these be pluralized since they're lists of items?

frontend/src/pages/projects/screens/spawner/environmentVariables/utils.ts
```diff
@@ -18,3 +19,25 @@ export const isStringKeyValuePairObject = (object: unknown): object is Record<st
...
+export const deletedConfigMapOrSecretVariables = (
+  notebook: NotebookKind | undefined,
+  envVariables: EnvVariable[],
+): {
+  deletedSecret: string[];
+  deletedConfigMap: string[];
+} => {
+  const existingEnvVariables = new Set(envVariables.map((env) => env.existingName));
+  const deletedConfigMap: string[] = [];
+  const deletedSecret: string[] = [];
```
jpuzz0, line 32: nit: deletedConfigMaps, deletedSecrets*

frontend/src/pages/projects/screens/spawner/service.ts
```diff
@@ -268,6 +274,7 @@ export const updateConfigMapsAndSecretsForNotebook = async (
 
   const deletingNames = deleteResources.map((resource) => resource.existingName || '');
   deletingNames.push(...removeDataConnections);
+  deletingNames.push(...deletedSecret, ...deletedConfigMap);
```
jpuzz0, line 277: Can lines 276 & 277 not be combined into 1 push?

frontend/src/pages/projects/screens/spawner/environmentVariables/AlertWarningText.tsx
```diff
@@ -0,0 +1,98 @@
...
+        <StackItem>
+          The following environment variables in this workbench cannot be found:
+        </StackItem>
+        <StackItem>
+          <List>
+            {[
+              formattedText(deletedConfigMap, 'configMap'),
+              formattedText(deletedSecret, 'secret'),
+            ].map((text, i) => (
+              <ListItem key={i}>{text}</ListItem>
```
jpuzz0, line 57: Can we not use the text of these config maps and secrets as the keys instead? I think maybe these originate from a Set elsewhere so maybe those are all unique?

frontend/src/pages/projects/screens/spawner/environmentVariables/AlertWarningText.tsx
```diff
@@ -0,0 +1,98 @@
...
+            ))}
+          </List>
+        </StackItem>
+        <StackItem>{commonDescriptionText}</StackItem>
+      </Stack>
+    );
+  } else if (deletedConfigMap.length > 0) {
+    description = (
+      <>
+        The {formattedText(deletedConfigMap, 'configMap')}
```
jpuzz0, line 67: This is called a couple times so might be better to hoist this as a variable of its own, along with the usage of `formattedText(deletedSecret, 'secret')`

frontend/src/pages/projects/screens/spawner/SpawnerPage.tsx
```diff
@@ -90,7 +92,13 @@ const SpawnerPage: React.FC<SpawnerPageProps> = ({ existingNotebook }) => {
...
   );
 
-  const [envVariables, setEnvVariables] = useNotebookEnvVariables(existingNotebook);
+  const [envVariables, setEnvVariables, envVariableloaded] =
+    useNotebookEnvVariables(existingNotebook);
+  const { deletedConfigMap, deletedSecret } = deletedConfigMapOrSecretVariables(
+    existingNotebook,
+    envVariables,
+  );
+
```
Gkrumbach07, line 101: in my opinion, becuase the deletedConfigMap and deletedSecret are only used in the alrt component, why not call this function in that component and then pass in existingNotebook and  envVariables instead. keeps things cleaner on the spawner page

Gkrumbach07 (changes requested):

frontend/src/pages/projects/screens/spawner/environmentVariables/AlertWarningText.tsx
pnaik1: here text is of type `React.JSX.Element` so cannot be used as a key

frontend/src/pages/projects/screens/spawner/SpawnerPage.tsx
```diff
@@ -90,7 +91,9 @@ const SpawnerPage: React.FC<SpawnerPageProps> = ({ existingNotebook }) => {
     defaultStorageClassName,
   );
 
-  const [envVariables, setEnvVariables] = useNotebookEnvVariables(existingNotebook);
+  const [envVariables, setEnvVariables, envVariableloaded] =
```
DaoDaoNoCode, line 94: A very very small nit: should be `envVariableLoaded`

frontend/src/pages/projects/screens/spawner/environmentVariables/useNotebookEnvVariables.tsx
```diff
@@ -61,17 +75,26 @@ export const fetchNotebookEnvVariables = (notebook: NotebookKind): Promise<EnvVa
...
+  const callback = React.useCallback(() => {
+    if (!notebook) {
+      return Promise.reject(new NotReadyError('No notebook'));
     }
+
+    return fetchNotebookEnvVariables(notebook);
   }, [notebook]);
 
-  return [envVariables, setEnvVariables];
+  const [fetchedVariable, envVariableloaded] = useFetchState(callback, []);
```
DaoDaoNoCode, line 93: Same here, should be `envVariableLoaded`

frontend/src/pages/projects/screens/spawner/environmentVariables/useNotebookEnvVariables.tsx
```diff
@@ -61,17 +75,26 @@ export const fetchNotebookEnvVariables = (notebook: NotebookKind): Promise<EnvVa
...
+
+  React.useEffect(() => {
+    setEnvVariables(fetchedVariable);
+  }, [fetchedVariable]);
```
DaoDaoNoCode, line 97: There is a very small issue with this method that could lead to a flickering with the alert. Imagine the following situation: You have some values in `notebook.spec.template.spec.containers[0].envFrom` and the secrets/config maps are not deleted. Then you will call `useNotebookEnvVariables` hook. The `envVariableLoaded` will become `true` as soon as the fetch finishes, while there is a very short time before it re-renders and `setEnvVariables` here. That will cause a flickering to show the alert very quickly and disappear. Although it's very hard to notice that and it normally won't have any issue, it's still better to get fixed if there is a way to do so. Maybe use a second loaded state and `useRef` to set it? 🤔

pnaik1:
FYI, I have modified the existing `useNotebookEnvVariables` hook  to include a call to ` getDeletedConfigMapOrSecretVariables`,  thats because current `envVariables` lists even the newly added/changed envVariables leading to  unnecessary alerts when previously  present envVariables are deleted from edit workbench page, open to any different approaches

DaoDaoNoCode (approved):
/lgtm

Gkrumbach07 (approved):
this looks good

/lgtm

PR #3373, opened 2024-10-23
Remove vuln_scan as it has rate limit issues blocking PRs

andrewballantyne:
Should stop having a blocking issue with rate limiting because of this. We have other mechanisms to catch vulnerabilities (like synk).


zdtsw:
i was the bad guy, and i totally forgot this one <3

PR #3377, opened 2024-10-24
Add character length limit to both model name and version name

ppadti:
Closes: [RHOAIENG-12844](https://issues.redhat.com/browse/RHOAIENG-12844)

## Description
This PR aims to add character length limit to both model name and version name in Register model and Register version forms. Also to use the given model name for the RegisteredModel and the given version name for both the ModelVersion and the ModelArtifact.

Screenshot 2024-10-25 at 3 49 09 PM

Screenshot 2024-10-25 at 3 48 31 PM

Screenshot 2024-10-25 at 3 53 23 PM

## How Has This Been Tested?
1. In model registry, click on Register model.
2. check the helper text "Cannot exceed 128 characters" is under model name and version name field.
3. Try to type in a long string, length greater than 128, you will see an error and it will disable the Register model button.
4.  Same goes for Register version form, try to Register a version with lang name and you will see the error in helper text and Register version button will be disabled. 
5. For Register version, when we select a model we will have multiple helper text under version name. 
Ref: [Doc](https://docs.google.com/document/d/1qEKzhWEDKQEPce3Zm_4ET6ci7Kkc55vegOZ0a5ORHBU/edit#bookmark=id.bmkvrjbzuhgf)

## Test Impact
Added cypress tests.

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/pages/modelRegistry/screens/RegisterModel/RegisterModel.tsx
+++ b/frontend/src/pages/modelRegistry/screens/RegisterModel/RegisterModel.tsx
@@ -4,6 +4,9 @@ import {
   BreadcrumbItem,
   Form,
   FormGroup,
+  FormHelperText,
+  HelperText,
+  HelperTextItem,
   PageSection,
   Stack,
   StackItem,
@@ -17,11 +20,12 @@ import FormSection from '~/components/pf-overrides/FormSection';
 import ApplicationsPage from '~/pages/ApplicationsPage';
 import { modelRegistryUrl, registeredModelUrl } from '~/pages/modelRegistry/screens/routeUtils';
 import { useRegisterModelData } from './useRegisterModelData';
-import { isRegisterModelSubmitDisabled, registerModel } from './utils';
+import { isNameValid, isRegisterModelSubmitDisabled, registerModel } from './utils';
 import RegistrationCommonFormSections from './RegistrationCommonFormSections';
 import { useRegistrationCommonState } from './useRegistrationCommonState';
 import PrefilledModelRegistryField from './PrefilledModelRegistryField';
 import RegistrationFormFooter from './RegistrationFormFooter';
+import { MR_CHARACTER_LIMIT } from './const';
 
 const RegisterModel: React.FC = () => {
   const { modelRegistry: mrName } = useParams();
@@ -31,6 +35,7 @@ const RegisterModel: React.FC = () => {
     useRegistrationCommonState();
 
   const [formData, setData] = useRegisterModelData();
+  const isModelNameValid = isNameValid(formData.modelName);
   const isSubmitDisabled = isSubmitting || isRegisterModelSubmitDisabled(formData);
   const { modelName, modelDescription } = formData;
 
@@ -75,7 +80,17 @@ const RegisterModel: React.FC = () => {
                     name="model-name"
                     value={modelName}
                     onChange={(_e, value) => setData('modelName', value)}
+                    validated={isModelNameValid ? 'default' : 'error'}
                   />
+                  {!isModelNameValid && (
+                    <FormHelperText>
+                      <HelperText>
+                        <HelperTextItem variant="error">
+                          Cannot exceed {MR_CHARACTER_LIMIT} characters
+                        </HelperTextItem>
+                      </HelperText>
+                    </FormHelperText>
+                  )}
                 </FormGroup>
                 <FormGroup label="Model description" fieldId="model-description">
                   <TextArea
--- a/frontend/src/pages/modelRegistry/screens/RegisterModel/RegistrationCommonFormSections.tsx
+++ b/frontend/src/pages/modelRegistry/screens/RegisterModel/RegistrationCommonFormSections.tsx
@@ -21,6 +21,8 @@ import FormSection from '~/components/pf-overrides/FormSection';
 import { ModelVersion } from '~/concepts/modelRegistry/types';
 import { ModelLocationType, RegistrationCommonFormData } from './useRegisterModelData';
 import { ConnectionModal } from './ConnectionModal';
+import { MR_CHARACTER_LIMIT } from './const';
+import { isNameValid } from './utils';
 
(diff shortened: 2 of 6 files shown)


ppadti:
@yih-wang could you please take a look at the above screenshots?

yih-wang:
@ppadti thanks for the updates!
I know the new validation guidelines suggest having the char limit displayed as help text under the field. However, looking at how it looks under the version field, I feel it's not conveying the info effectively:
- The `current version name` is not something that can be validated, so we shouldn't use the multi-condition validation pattern here.
- However, having both help texts displayed as two lines of plain text also feels strange.
- Also, the char limit isn't too important here IMO because a user will rarely touch the 128 char limit. It's basically a mechanism we add to avoid registration failure.
So I'm wondering are we supposed to always have the char limit displayed as help text, or should it be optional? @nding @simrandhaliw 

I personally prefer not to display the char limit as help text here (for both model name and version name), but only show the error message when user exceeds the limitation.

manaswinidas:
@yih-wang I see two help texts in the `Create Project` modal.

Screenshot 2024-10-24 at 11 47 37 PM

yih-wang:
@manaswinidas the two help text in project creation are both conditions to be validated. So they both have that 'validation status' to show whether the condition is satisfied. However, in our case, `current version name is xxx` is not a validation condition. We can never tell that when it's satisfied or not. It's just a reference for users to create the next version name. So it doesn't make sense to have it in that 'condition to be validated' format.

I chat with @angesDing and @simrandhaliw offline and they both agree that the char limit is not essential to display as help text in our case. We are likely to make improvements to the guidelines to make sure it's clear that the help text for char limit is optional.

So, @ppadti for both model name and version name, let's remove the char limit as the help text, and only have it as error message when the user input exceeds the char limit. Basically just copy what we have for MR labels. For version name, there could be both help text and error message (see the image below). I'm not sure which one would be on top by default, but I'm fine with both ways. We should just follow PF's default and avoid any customization.
image

ppadti:
> @manaswinidas the two help text in project creation are both conditions to be validated. So they both have that 'dynamic validation status' at the very beginning to show whether the condition is satisfied. However, in our case, `current version name is xxx` is not a validation condition. We can never tell that when it's satisfied or not. It's just a reference for users to create the next version name. So it doesn't make sense to have it in that 'condition to be validated' format.
> 
> I chat with @angesDing and @simrandhaliw offline and they both agree that the char limit is not essential to display as help text in our case. We are likely to make improvements to the guidelines to make sure it's clear that the help text for char limit is optional.
> 
> So, @ppadti for both model name and version name, let's remove the char limit as the help text, and only have it as error message when the user input exceeds the char limit. Basically just copy what we have for MR labels. For version name, there could be both help text and error message (see the image below). I'm not sure which one would be on top by default, but I'm fine with both ways. We should just follow PF's default and avoid any customization. image

@yih-wang just wanted a clarification, here in label we show `error icon` before error text `Label text can't exceed 63 characters`
Screenshot 2024-10-25 at 2 32 02 PM
that is missing in img that you shared above. So you want me to keep the error icon before error text under model name and version name field like this:
Screenshot 2024-10-25 at 2 34 18 PM

ppadti:
oh, in Register version form with icon it looks odd,
Screenshot 2024-10-25 at 2 55 50 PM

so I will go ahead with the error text without error icon as you mentioned above @yih-wang sorry for the confusion here...Thanks!!

yih-wang:
@ppadti Seems the new guidelines don't include the error icon in the error messages. So I think the error icon we have for label error will also be removed when we apply the new rules across the console.

ppadti:
@yih-wang I have updated the screenshots, take a look them when you have time.

frontend/src/pages/modelRegistry/screens/RegisterModel/RegistrationCommonFormSections.tsx
```diff
@@ -86,14 +88,22 @@ const RegistrationCommonFormSections = <D extends RegistrationCommonFormData>({
...
+                <HelperTextItem variant="indeterminate">
+                  Current version is {latestVersion.name}
+                </HelperTextItem>
               </HelperText>
-            </FormHelperText>
-          )}
+            )}
+            {!isVersionNameValid && (
+              <HelperText>
+                <HelperTextItem variant="error">Cannot exceed 128 characters</HelperTextItem>
```
manaswinidas, line 103: small nit: can we use the `NAME_CHARACTER_LIMIT` const here instead?

frontend/src/pages/modelRegistry/screens/RegisterModel/RegisterModel.tsx
```diff
@@ -75,7 +84,17 @@ const RegisterModel: React.FC = () => {
                     name="model-name"
                     value={modelName}
                     onChange={(_e, value) => setData('modelName', value)}
+                    validated={isModelNameValid ? 'default' : 'error'}
                   />
+                  {!isModelNameValid && (
+                    <FormHelperText>
+                      <HelperText>
+                        <HelperTextItem variant="error">
+                          Cannot exceed 128 characters
```
manaswinidas, line 93: can we use the `NAME_CHARACTER_LIMIT` const here too?

manaswinidas (dismissed):
small nit, otherwise lgtm

frontend/src/pages/modelRegistry/screens/RegisterModel/RegisterModel.tsx
manaswinidas: also can we rename `NAME_CHARACTER_LIMIT` to `MR_CHARACTER_LIMIT` or anything that states this is solely for MR?

gitdallas (approved):

mturley (commented):
Just one nitpick, otherwise looks good. If you disagree with this let me know, we could leave it as-is.

yih-wang:
LGTM, thanks for the updates, @ppadti !

mturley (approved):
LGTM, thanks @ppadti !

PR #3382, opened 2024-10-24
Fix type error in DSC to prevent crashing when operator first deployed

DaoDaoNoCode:
JIRA: https://issues.redhat.com/browse/RHOAIENG-14095

## Description

This is because the `installedComponents` is possibly `undefined` in the first few minutes after the operator is deployed. It will be called `Array.reduce` later, which causes the error because you cannot call `Array.reduce` on `undefined`. Change the type to optional could help with this issue.

## How Has This Been Tested?

You can change the backend to mockup the DSC return value to see the result on the frontend. At https://github.com/opendatahub-io/odh-dashboard/blob/main/backend/src/utils/dsc.ts#L14, change it to
```
.then((res) => {
      const body = (res.body as DataScienceClusterList).items[0];
      delete body.status.installedComponents;
      return body;
    })
```
And see the change of the frontend. Before the change in this PR, you should get a crashed dashboard with the error message as described in the JIRA ticket. After applying the changes in this PR, the dashboard could show correctly the minimum availability of the components.

## Test Impact

N/A

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/concepts/areas/utils.ts
+++ b/frontend/src/concepts/areas/utils.ts
@@ -61,7 +61,7 @@ export const isAreaAvailable = (
   const requiredComponentsState =
     requiredComponents && dscStatus
       ? requiredComponents.reduce<IsAreaAvailableStatus['requiredComponents']>(
-          (acc, component) => ({ ...acc, [component]: dscStatus.installedComponents[component] }),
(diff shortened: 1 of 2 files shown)


pnaik1:
/lgtm
Tested locally doesnt crash the dashboard

PR #3387, opened 2024-10-25
RHOAIENG-14811 The userdID was not correctly transferred to the frontend

pilhuhn:
Closes https://issues.redhat.com/browse/RHOAIENG-14811
We now get the userID from the status api endpoint if set. 
This builds on RHOAIENG-9480/PR #3107.
This closes the first try in #3362

## Description

The previous code did not correctly pass the userID from the backend around in the frontend.
The backend already has it correctly on a 2.14 cluster and provides it via /api/status call.

## How Has This Been Tested?

Manual test in dev-mode running `npm run start:dev:ext` against an instance on Dev-Sandbox and a non-Sandbox one.

Then looking at the console output for the the Identify event

```
Identify event triggered: {... "userID":"b09067a69...99"}
```

If the userID looks like this (whith chars) it is an anonymous one. If it is numeric (may be negative) and short (probably < 8 digits), then it is a web-user-id. 

## Request review criteria:

cc @andrewballantyne

### DIFF
--- a/frontend/src/__mocks__/mockStatus.ts
+++ b/frontend/src/__mocks__/mockStatus.ts
@@ -12,6 +12,7 @@ export const mockStatus = (options?: {
     },
     namespace: 'opendatahub',
     userName: 'test-user',
+    userID: '1234',
     clusterID: '16855612-2bb7-4a4c-9ff0-72rasdfd5',
     clusterBranding: 'ocp',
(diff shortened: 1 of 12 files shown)


christianvogt:
/lgtm
/approve

tested against my dev sandbox instance and another odh cluster instance. Sandbox cluster included my numeric user ID.

PR #3388, opened 2024-10-25
Cypress e2e Project Creation/Deletion

antowaddle:
https://issues.redhat.com/browse/RHOAIENG-14367

## Description
Migration of the below Robot test to Cypress

ODS-1775

https://opendatascience-jenkins-csb-rhods.apps.ocp-c1.prod.psi.redhat.com/job/rhoai/job/2.13/job/selfmanaged/job/cli/job/aws/job/rhoai-smoke/1/robot/report/log.html#s1-s4-s2-s4-s1-t3 

## How Has This Been Tested?

## How to run
After exporting the test-variables.yml we have 2 different ways:

Using the UI
Go to odh-dashboard/frontend/src/__tests__/cypress and run the command npx cypress open . This will open the Cypress UI where verifyDSPCreation.cy.ts can be run.

Headless
Go to odh-dashboard/frontend/src/__tests__/cypress and run the command npx cypress run --spec "cypress/tests/e2e/dataScienceProjects/testProjectCreation.cy.ts" --browser chrome

## How Has This Been Tested?
Locally, against a live cluster.
image

## Test Impact
This is already a test

This PR also introduces a plugin - https://github.com/filiphric/cypress-plugin-steps This plugin will display text contained within cy.step('...') numbered and highlighted on the Cypress Console to allow for improved test bugging, maintainability and readability.
image


frontend/src/__tests__/cypress/cypress/utils/oc_commands/project.ts
```diff
@@ -71,3 +71,43 @@ export const addUserToProject = (
...
+
+    if (projectExists) {
+      cy.log(`Project '${projectName}' exists.`);
+    } else {
+      cy.log(`Project '${projectName}' does not exist.`);
+    }
+
+    return cy.wrap(projectExists);
+  });
+};
```
FedeAlonso, line 113:
```suggestion
/**
 * Get OpenShift Project 
 *
 * @param projectName OpenShift Project name to get (in the format like 'cypress-test-project')
 * @returns A Cypress chainable with the project information or null if not found
 */
export const getOpenShiftProject = (projectName: string): Cypress.Chainable<string | null> => {
  if (!projectName || typeof projectName !== 'string') {
    cy.log(`ERROR: Invalid project name provided: ${projectName}`);
    throw new Error(`Invalid project name: ${projectName}`);
  }

  const checkCommand = `oc get project "${projectName}" -o name`;

  cy.log(`Executing command: ${checkCommand}`);

  return cy.exec(checkCommand, { failOnNonZeroExit: false }).then((result: Cypress.Exec) => {
    cy.log(`Command result: ${JSON.stringify(result)}`);

    if (result.code !== 0 && result.code !== 1) {
      // Code 1 is expected when the project doesn't exist, so we only throw for other non-zero codes
      cy.log(`ERROR: Command execution failed
              stdout: ${result.stdout}
              stderr: ${result.stderr}`);
      throw new Error(`Command failed with code ${result.code}`);
    }

    return result.code === 0 ? result.stdout.trim() : null;
  });
};

/**
 * Verify if an OpenShift Project exists
 *
 * @param projectName OpenShift Project name to verify (in the format like 'cypress-test-project')
 * @returns A Cypress chainable boolean indicating whether the project exists (true) or not (false)
 */
export const verifyOpenShiftProjectExists = (projectName: string): Cypress.Chainable<boolean> => {
  return getOpenShiftProject(projectName).then((projectInfo) => {
    const projectExists = projectInfo === `project.project.openshift.io/${projectName}`;

    if (projectExists) {
      cy.log(`Project '${projectName}' exists.`);
    } else {
      cy.log(`Project '${projectName}' does not exist.`);
    }

    return cy.wrap(projectExists);
  });
};
```

FedeAlonso: Maybe it's better approach split this method in two:
1. Getting the DSP (which is going to be reusable)
2. Do the verification (which maybe should not go here as not contains anything oc related)

frontend/src/__tests__/cypress/cypress/types.ts
```diff
@@ -84,3 +84,10 @@ export type TestConfig = {
   OCP_ADMIN_USER: UserAuthConfig;
   S3: AWSS3Buckets;
 };
+
+export type DataScienceProjectData = {
+  dspCreationProjectName: string;
+  dspCreationProjectDescription: string;
+  dspCreationURL: string;
+  dspOCName: string;
```
FedeAlonso, line 92: The types should not have the "-Creation-". Also as the type is the DSP, shouldn't contain the "dsp-" in the fields
Why it does have the OCName and the URL?

FedeAlonso (changes requested):

frontend/src/__tests__/cypress/cypress/pages/projects.ts
```diff
@@ -314,6 +314,22 @@ class ProjectDetails {
   getKserveModelMetricLink(name: string) {
     return this.findKserveModelsTable().findByTestId(`metrics-link-${name}`);
   }
+
+  verifyProjectName(project: string) {
+    return cy.get('[data-testid="app-page-title"]').should('contain.text', project);
```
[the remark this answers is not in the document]
antowaddle: @FedeAlonso the approach currently in the mock tests seem to be that the page objects are also linked with common methods like clicking etc. For now, I've just continued that approach for now for consistency. it's probably something though we need to align on and confirm over time.

frontend/src/__tests__/cypress/cypress/utils/oc_commands/project.ts
antowaddle: @antowaddle before I split this up, would there be any scenario where we would only want to do a Verification that wouldn't require getting the project first?

frontend/src/__tests__/cypress/cypress/types.ts
antowaddle: I changed the types as requested. 

The test data in the fixtures field is pre-apprended with 'dsp..' in the 'dataScienceProject.yaml', this allows us to then build upon this test data file for other data science projects.

frontend/src/__tests__/cypress/cypress/fixtures/e2e/dataScienceProject.yaml
```diff
@@ -0,0 +1,4 @@
+# cypress/fixtures/e2e/dataScienceProject.yaml
```
manosnoam, line 1: Can you set the test and data files for the E2E with more common file names, for example:

- ds_projects_test.cy.ts
- ds_projects_data.cy.json

Also note that JSON is more common than YAML for Cypress data files.

frontend/src/__tests__/cypress/cypress/pages/projects.ts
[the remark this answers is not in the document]
manosnoam: I agree, let's keep the same approach as in Mock for now.

frontend/src/__tests__/cypress/cypress/support/commands/application.ts
```diff
@@ -116,6 +116,26 @@ declare global {
...
+       * cy.step('Verifying dashboard elements');
+       *
+       * In the Cypress test runner and console, this will appear as:
+       * STEP: Logging in to the application
+       * STEP: Verifying dashboard elements
+       *
+       * This helps in quickly identifying the current stage of the test execution
+       * and provides a clear overview of the test flow.
+       */
+      step: (description: string) => Chainable;
```
manosnoam, line 138: I think this framework option to print into test runner and console should be in a more common file (like utils.js or helpers.js), not related to `application.ts`. @christianvogt wdyt ?

frontend/src/__tests__/cypress/cypress/support/commands/application.ts
```diff
@@ -237,6 +257,10 @@ Cypress.Commands.overwriteQuery('findAllByTestId', function findAllByTestId(...a
   return enhancedFindByTestId(this, ...args);
 });
 
+Cypress.Commands.add('step', (description: string) => {
+  cy.log(`\n--- 🔷 Step: ${description} 🔷---\n`);
+});
+
```
manosnoam, line 263: What about printing Steps with a counter like:

```
let stepCounter = 0;

Cypress.Commands.add('step', (description) => {
  stepCounter += 1;
  const stepMessage = `Step ${stepCounter.toString().padStart(2, '0')}: ${description}`;
  cy.log(stepMessage);
  console.log(stepMessage);
});
```

frontend/src/__tests__/cypress/cypress/tests/e2e/ui/dataScienceProject/verifyDSPCreation.cy.ts
manosnoam, line 1: As mentioned before - Can you set the test and data files for the E2E with more common file names, for example:

    ds_projects_test.cy.ts
    ds_projects_data.cy.json

frontend/src/__tests__/cypress/cypress/tests/e2e/ui/dataScienceProject/verifyDSPCreation.cy.ts
```diff
@@ -0,0 +1,95 @@
...
+      });
+  });
+
+  it('Create a Data Science Project in RHOAI', () => {
+    // Step 1: Log into the application
+    cy.step('Log into the application');
+    cy.visitWithLogin('/', ADMIN_USER);
+    projectListPage.navigate();
+
+    // Step 2: Open the Create Data Science Project modal
```
manosnoam, line 51: If you implement the steps with counter as I suggested earlier, we can skip adding the obvious comments like:
`// Step 2: Open the Create Data Science Project modal`

frontend/src/__tests__/cypress/cypress/utils/oc_commands/project.ts
```diff
@@ -71,3 +71,43 @@ export const addUserToProject = (
...
+  const checkCommand = `oc get project "${projectName}" -o name`;
+
+  cy.log(`Executing command: ${checkCommand}`);
+
+  return cy.exec(checkCommand, { failOnNonZeroExit: false }).then((result: Cypress.Exec) => {
+    cy.log(`Command result: ${JSON.stringify(result)}`);
+
+    if (result.code !== 0 && result.code !== 1) {
+      // Code 1 is expected when the project doesn't exist, so we only throw for other non-zero codes
+      cy.log(`ERROR: Command execution failed
```
manosnoam, line 96: I think we should rather use a common function as you do with `cy.step` , for example, add this to utils or helpers:
```
// utils.js or helpers.js
function logAndThrowError({ code, stdout, stderr }) {
  const message = `Command failed with code ${code}
                   stdout: ${stdout}
                   stderr: ${stderr}`;
  console.error(message); // Print to console
  cy.log(message); // Log to Cypress
  throw new Error(message);
}
```

manosnoam (changes requested):
Thanks @antowaddle for your first PR!
See some suggestions I have.

frontend/src/__tests__/cypress/cypress/fixtures/e2e/dataScienceProject.yaml
antowaddle: @manosnoam can you review the file names for the Mock tests? They seem to use camel case and not underscores.

frontend/src/__tests__/cypress/cypress/support/commands/application.ts
antowaddle: Let me give this a try, there is a plugin we can use for this. I'm not sure how the steps would work when running multiple tests which might be a problem.

frontend/src/__tests__/cypress/cypress/utils/oc_commands/project.ts
FedeAlonso: @antowaddle I'm not sure we want to make a  verifyOpenShiftProjectExists that doesn't contain the get. In fact in the example I have proposed, in line 113 the get is called. 
What I do see probable is that, at some point, we want to make use of the get, without it being inside the verify. Hence my proposal to divide them in two parts:
1. Get
2. Verify (which uses the Get)

frontend/src/__tests__/cypress/cypress/tests/e2e/ui/dataScienceProject/verifyDSPCreation.cy.ts
antowaddle: Understood, I'll add some general code comments to the file in place of this.

frontend/src/__tests__/cypress/cypress/utils/oc_commands/project.ts
antowaddle: Fede made a suggestion here to split the methods in two, I'm going to implement that so there will be two separate commands for Get and Verify.

frontend/src/__tests__/cypress/cypress/fixtures/e2e/dataScienceProject.yaml
manosnoam: ack, in that case let's use same naming conventions.
In Mock they are called for example: `mockAllowedUsers.ts`
We can use:
`testDSProjects.ts` to indicate its e2e tests.

And according to existing fixtures for the tests, maybe we should also keep the current structure:
`frontend/src/__tests__/cypress/cypress/fixtures/resources/yaml`
In there place the relevant data for DS projects tests as `testDSProjects.yaml`

?

frontend/src/__tests__/cypress/cypress/utils/oc_commands/project.ts
manosnoam: ack, but if you see that you need to use many times these prints, I would create common method for:
```
console.error(message);
cy.log(message); 
throw new Error(message);
```

antowaddle: I've made the changes suggested by Fede and will commit them shortly. Note that the other methods in the 'projects.ts' file do not use a common function for the error handling.

Would you like me to add this to the current PR or could we refactor this later? If I add it to the current PR, should all the other methods in the 'projects.ts' file also be refactored for consistency?

frontend/src/__tests__/cypress/cypress/tests/e2e/ui/dataScienceProject/verifyDSPCreation.cy.ts
antowaddle: Latest commit has more general comments relating to the steps within the test.

frontend/src/__tests__/cypress/cypress/support/commands/application.ts
antowaddle: If we go with the above the 'before' method needs have the counter added for each test which i feel is unnecessary overhead. There is a plugin I referenced in the Framework document that can be used, would you like me to implement that?

manosnoam: Yes, and an existing plugin for printing steps is preferable, but I think it's better to add it as a new PR

frontend/src/__tests__/cypress/cypress/tests/e2e/ui/dataScienceProject/verifyDSPCreation.cy.ts
manosnoam: I would make the api in a different directory (sibling to e2e):
`frontend/src/tests/cypress/cypress/tests/api`

frontend/src/__tests__/cypress/cypress/support/commands/application.ts
antowaddle: Agreed, thanks for confirming.

frontend/src/__tests__/cypress/cypress/utils/oc_commands/project.ts
manosnoam: It would be nice to use a common method. If it's not an issue to add it to this PR it would be nice.
We probably need a new PR only for the other infra improvement using `cy.step`, since it imports an external plugin.

frontend/src/__tests__/cypress/cypress/tests/e2e/ui/dataScienceProject/verifyDSPCreation.cy.ts
antowaddle: In the Framework document I've suggested we do the following:

 │   ├── e2e
  │   ├── UI
  │   │   └── dataScienceProject/
  │   │       ├── verifyDSPCreation.cy.ts
  │   │       └── verifyDSPDeletion.cy.ts
  │   ├── API
  │   │   └── healthChecks/
  │   │       └── verifyRHOAIEndpoints.cy.ts

This will allow us to have e2e tests for both UI and API Tests. It's more common to have UI e2e tests for sure, but we can also write tests that use the underlying APIs to do things like logging in, submitting projects etc. as a chained set of actions. This approach can be extremely beneficial to get feedback quickly and validate the backend is up/operational before running the larger more complex and time heavy UI tests.

Using 'API' as a label is broader and encompasses all types of application programming interfaces, while 'REST' specifically refers to a subset of APIs, potentially limiting the context and flexibility of your tests.

frontend/src/__tests__/cypress/cypress/fixtures/e2e/dataScienceProject.yaml
antowaddle: Would 'testProjectCreationDeletion' be more relevant? 

The folder structure allows us to use '...dataScienceProject.yaml' for all the test cases related to dataScienceProjects. We should then create similar files for workbenches etc. 
frontend/src/__tests__/cypress/cypress/fixtures/e2e/dataScienceProject.yaml

As we scale the tests over time, having this granularity becomes particularly useful.

antowaddle: My recent commit has resolved this issue - test names were changed along with directories.

frontend/src/__tests__/cypress/cypress/tests/e2e/ui/dataScienceProject/verifyDSPCreation.cy.ts
antowaddle: My recent commit has resolved this issue also, file names and directories have been changed/renamed.

frontend/src/__tests__/cypress/cypress/utils/oc_commands/project.ts
antowaddle: https://github.com/opendatahub-io/odh-dashboard/pull/3388/files#diff-0800fa4ff711a8478b16382cabb7407275a1245800c19c72970c196197731573

frontend/src/__tests__/cypress/cypress/support/commands/application.ts
antowaddle: My latest commit has removed the custom step written here in favour of using the cypress-step-plugin.

FedeAlonso (approved):
/lgtm

manosnoam (approved):
Looks good, thanks!

frontend/src/__tests__/cypress/cypress/tests/e2e/dataScienceProjects/testProjectCreation.cy.ts
```diff
@@ -0,0 +1,89 @@
...
+    createProjectModal.k8sNameDescription
+      .findDescriptionInput()
+      .type(testData.dsProjectDescription);
+
+    // Submit project creation
+    cy.step('Save the project');
+    createProjectModal.findSubmitButton().click();
+
+    // Verify project creation
+    cy.step(`Verify that the project ${testData.dsProjectName} has been created`);
```
jiridanek, line 68: Is it normal practice to always have a comment and then a cy.step("") with a string that says the same thing as the comment? Or is it just because the test was written without cy.step initially and the comments are leftovers?

antowaddle: I've intentionally left comments also, in some more intricate tests we may also want to add comments but not necessarily print that information to the console in cy.step.

PR #3389, opened 2024-10-25
Apply proper styling for 'Who is my admin?' button and popover content

jeff-phillips-18:
Closes [RHOAIENG-14755](https://issues.redhat.com/browse/RHOAIENG-14755)

## Description
Updates instances of the `Who's my administrator?` link and adds it to locations where it is needed.

## How Has This Been Tested?
As a non-admin who cannot create projects navigate to the following and verify the link:
- Data Science Projects page when there are no projects
- Data Science Projects page when there are projects (`Need another project?` link)
- Project details page, Models tab when no model platform is selected
- Model Registry page when there are no models
- Registered Models page when there are no registered models
- Distributed Workload metrics where the cluster is not configured
- Settings -> Storage classes when there are no storage classes

## Test Impact
There is already a test for this in locations

## Screen shots

image

image

image

image

image

image

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/components/WhosMyAdministrator.tsx
+++ b/frontend/src/components/WhosMyAdministrator.tsx
@@ -1,0 +1,50 @@
+import * as React from 'react';
+import { Button, Popover } from '@patternfly/react-core';
+import { OutlinedQuestionCircleIcon } from '@patternfly/react-icons';
+import PopoverListContent from '~/components/PopoverListContent';
+import { FindAdministratorOptions } from '~/pages/projects/screens/projects/const';
+
+type Props = {
+  buttonLabel?: string;
+  headerContent?: string;
+  leadText?: string;
+  isInline?: boolean;
+  contentTestId?: string;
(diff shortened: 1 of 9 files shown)


frontend/src/components/WhosMyAdministrator.tsx
```diff
@@ -0,0 +1,50 @@
...
+    }
+  >
+    <Button
+      isInline={isInline}
+      variant="link"
+      icon={<OutlinedQuestionCircleIcon />}
+      data-testid={linkTestId}
+    >
+      {buttonLabel}
+    </Button>
```
christianvogt, line 46: Requires an `aria-label` if there's no button label.

Likely always need a label prop and add a separate prop to hide the label.

christianvogt: Must have made the comment before realizing there was also a default value. So the intent is to always have a label.

PR #3393, opened 2024-10-28
Show deprecated alert correctly when the storage class is new created

DaoDaoNoCode:
JIRA: https://issues.redhat.com/browse/RHOAIENG-15026

## Description

The deprecated alert should only be set when a storage class is configured (has the annotation) and `isEnabled` is set to `false`. Otherwise, we should not show the deprecated alert and label.

## How Has This Been Tested?

1. Create a storage class
2. Don't configure it, use it to create a cluster storage
3. You should not see any deprecated alert and the deprecated label on the cluster storage table row

## Test Impact

A cypress test was added to verify that the deprecated alerts are not shown when the annotation doesn't exist.

## Request review criteria:

If you have UI changes: 


Gkrumbach07 (approved):
/lgtm

PR #3395, opened 2024-10-29
Fix popover overflow

ashley-o0o:
Closes: [RHOAIENG-10993](https://issues.redhat.com/browse/RHOAIENG-10993)

## Description
Fixes popover overflow when the messages is too long. Adds a scrollable option

## How Has This Been Tested?
Tested locally using mock data for the popover. Edit the `bodyContent` to include more text, and test the scroll bar functionality 

## Test Impact
No new tests added

## Screenshots 
Before:
Screenshot 2024-10-29 at 8 20 49 AM
After: (With scroll option)
Screenshot 2024-10-29 at 8 21 49 AM

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/pages/modelServing/screens/global/InferenceServiceStatus.tsx
+++ b/frontend/src/pages/modelServing/screens/global/InferenceServiceStatus.tsx
@@ -124,9 +124,10 @@ const InferenceServiceStatus: React.FC<InferenceServiceStatusProps> = ({
   return (
     <Popover
       data-testid="model-status-tooltip"
+      className="odh-u-scrollable"
       position="top"
       headerContent={headerContent()}
-      bodyContent={<div style={{ maxHeight: '95vh', overflow: 'auto' }}>{bodyContent}</div>}
+      bodyContent={bodyContent}
       isVisible={bodyContent ? undefined : false}
     >
       {statusIcon()}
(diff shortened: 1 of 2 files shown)


frontend/src/pages/modelServing/screens/global/InferenceServiceStatus.tsx
```diff
@@ -109,7 +109,7 @@ const InferenceServiceStatus: React.FC<InferenceServiceStatusProps> = ({
     <Popover
       data-testid="model-status-tooltip"
       position="top"
-      bodyContent={bodyContent}
+      bodyContent={<div style={{ maxHeight: '65vh', overflow: 'auto' }}>{bodyContent}</div>}
```
christianvogt, line 112: @ashley-o0o is there a specific reason to use 65% of the available height? Why not allow the popover to consume the most amount of available vertical space?

ashley-o0o: @christianvogt When I used 95%, it still extended beyond the window on my laptop. Obviously this might change depending on the monitor, however, visually it looked better to me to have the popover overall smaller as opposed to using most of the screen. I can change it to 95% though if you believe that would be best.

christianvogt: @ashley-o0o i wonder if that has to do with the content being updated dynamically after the popover has already positioned itself and then the new larger content doesn't trigger a revaluation of the popover position. If that's the case then 65% can still cause the same issue. I'll test this out.

christianvogt:
Testing with mock data the resulting popover looks like this:
image

ashley-o0o:
@simrandhaliw when you have a moment..

PR #3405, opened 2024-10-30
Submit additional configuration parameters

manaswinidas:
Closes: [RHOAIENG-14701](https://issues.redhat.com/browse/RHOAIENG-14701)
Closes: [RHOAIENG-14700](https://issues.redhat.com/browse/RHOAIENG-14700)

## Description

You can now submit configuration parameters(both serving runtime arguments and environment variables).

## How Has This Been Tested?

1. Try adding arguments(add new, update existing and submit) and environment updates(add new, update existing and submit)
2. This feature is behind a feature flag `disableServingRuntimeParams` - you may need to scale down Deployment and add this feature flag to CRD and `OdhDashboardConfig` to see these fields

## Test Impact

Tests will be added in a follow up PR.

## Request review criteria:

If you have UI changes: 


frontend/src/pages/modelServing/screens/projects/kServeModal/EnvironmentVariablesSection.tsx
```diff
@@ -16,31 +16,44 @@ import {
...
+    if (additionalEnvVars) {
+      const newVars = [...additionalEnvVars, { name: '', value: '' }];
+      setAdditionalEnvVars(newVars);
+      setData('servingRuntimeEnvVars', newVars);
+    }
   };
 
   const removeEnvVar = (indexToRemove: number) => {
-    setAdditionalEnvVars((prevVars) => prevVars.filter((_, i) => i !== indexToRemove));
+    if (additionalEnvVars) {
```
manaswinidas, line 43: `spec.predictor.model.env` is optional so had to add these `if` statements

frontend/src/pages/modelServing/screens/projects/kServeModal/EnvironmentVariablesSection.tsx
```diff
@@ -16,31 +16,44 @@ import {
...
+  const [additionalEnvVars, setAdditionalEnvVars] = React.useState<ServingContainer['env']>(
+    data.servingRuntimeEnvVars,
+  );
 
   const addEnvVar = () => {
-    setAdditionalEnvVars((prevVars) => [...prevVars, { name: '', value: '' }]);
+    if (additionalEnvVars) {
+      const newVars = [...additionalEnvVars, { name: '', value: '' }];
+      setAdditionalEnvVars(newVars);
+      setData('servingRuntimeEnvVars', newVars);
```
mturley, line 38: We're duplicating state here - now that the `data.servingRuntimeEnvVars` exists, I don't think we still need our `additionalEnvVars` useState here. Every place that updates it also updates the data in the form state, so we can just rely on those `setData` calls and use `data.servingRuntimeEnvVars` I think.

frontend/src/pages/modelServing/screens/projects/kServeModal/ServingRuntimeArgsSection.tsx
```diff
@@ -13,48 +13,43 @@ import { UpdateObjectAtPropAndValue } from '~/pages/projects/types';
...
+        }
+      >
+        <Icon aria-label="Additional serving runtime arguments info" role="button">
+          <OutlinedQuestionCircleIcon />
+        </Icon>
+      </Popover>
+    }
+    fieldId="serving-runtime-arguments"
+  >
+    <TextArea
```
mturley, line 39: This TextArea needs either an id or an aria-label - I missed this in the last PR, but it's giving a console warning.

It should probably have the id `"serving-runtime-arguments"` because that's the `fieldId` on the FormGroup.

frontend/src/pages/modelServing/screens/projects/kServeModal/EnvironmentVariablesSection.tsx
```diff
@@ -16,31 +16,44 @@ import {
...
+  data,
+  setData,
+}) => {
+  const [additionalEnvVars, setAdditionalEnvVars] = React.useState<ServingContainer['env']>(
+    data.servingRuntimeEnvVars,
+  );
 
   const addEnvVar = () => {
-    setAdditionalEnvVars((prevVars) => [...prevVars, { name: '', value: '' }]);
+    if (additionalEnvVars) {
```
mturley, line 35: This condition is making it so that if `additionalEnvVars` is undefined (which ends up happening if we are editing a model that was deployed before this PR), clicking the "Add variable" button doesn't do anything. I think we should remove this `if` statement and use `(additionalEnvVars || [])` here.

Now, for the other two cases (`removeEnvVar` and `updateEnvVar`) we know there will be at least one env var in state if we can reach those functions, so it's probably fine to leave the `if` statement for those, especially since the update function expects `additionalEnvVars[indexToUpdate]` to be defined.

frontend/src/api/k8s/inferenceServices.ts
```diff
@@ -79,6 +81,8 @@ export const assembleInferenceService = (
                 key: dataConnectionKey,
                 path,
               },
+              args: servingRuntimeArgs,
+              env: servingRuntimeEnvVars,
```
mturley, line 85: Either here or before we get here, we should filter out any empty strings from `servingRuntimeArgs` and filter out any env vars with an empty string for the `name`. Then we'll need to update the tests/mocks changed by this PR so they have `args: []` instead of `args: [""]` as the empty value.

frontend/src/api/k8s/inferenceServices.ts
```diff
@@ -121,6 +125,8 @@ export const assembleInferenceService = (
...
```
mturley, line 129: Same as above

mturley (approved):
Mostly looking good and working as expected. Approving to unblock the next work - let's leave the Jira issue in progress until we address feedback here though. @manaswinidas I'll see if we can get a PR together for it before you're back, otherwise it can wait until Monday.

Also note to others - we discussed adding tests for this as a followup PR as well.

PR #3369, opened 2024-10-23
"Source model format" and "source model format version" should be editable

YuliaKrimerman:
[RHOAIENG-14812](https://issues.redhat.com/browse/RHOAIENG-14812)

## Description

In the version Details page- Verify that "Source model format" and "source model format version" should be editable. Make sure that the value is preserved when the page is refreshed 
Screenshot 2024-10-22 at 12 34 35 PM
Screenshot 2024-10-25 at 2 40 54 PM

## How Has This Been Tested?

On the UI
## Test Impact

Added a Cypress test

## Request review criteria:

Self checklist (all need to be checked):
- [x] The developer has manually tested the changes and verified that the changes work
- [x] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [x] The developer has added tests or explained why testing cannot be added (unit or cypress tests for related changes)

If you have UI changes: 

- [ ] Included any necessary screenshots or gifs if it was a UI change.
- [ ] Included tags to the UX team if it was a UI/UX change.

After the PR is posted & before it merges:
- [ ] The developer has tested their solution on a cluster by using the image produced by the PR to `main`

PR #3411, opened 2024-10-31
MR API patch error fix

YuliaKrimerman:
[HOAIENG-14812](https://issues.redhat.com/browse/RHOAIENG-14812)
## Description

A follow up fix for the saving wasn't working because of an API changed made. Added the needed value the API was needing. 

## How Has This Been Tested?

on MR UI- PLEASE use only modelregistry-sample-2 or a new model registry you create, as there is a bug on the API side that doesn't patch the old registries.

## Test Impact

Test were written in the original feature PR

## Request review criteria:

If you have UI changes: 


mturley (approved):
Approving for now so we can verify the story, but we may need to make sure the MR team fixes their issue with existing registries not updating when the operator updates, or this patch won't work for existing users who update. Will follow up, we may want to revert this and #3369 before we release.

PR #3416, opened 2024-11-01
Follow-up fixes in Model Deployment modals for Select component

DaoDaoNoCode:
JIRA: Follow-up of [RHOAIENG-10833](https://issues.redhat.com/browse/RHOAIENG-10833)

## Description

Fix issues as discussed in https://redhat-internal.slack.com/archives/C05SMJ09DD2/p1730402472208489

## How Has This Been Tested?

Make sure the serving runtime/model framework selections are working well on both KServe/ModelMesh modals when deploying/editing a model.

## Test Impact

N/A

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/pages/modelServing/screens/projects/InferenceServiceModal/InferenceServiceServingRuntimeSection.tsx
+++ b/frontend/src/pages/modelServing/screens/projects/InferenceServiceModal/InferenceServiceServingRuntimeSection.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { Alert, FormGroup, Skeleton, Text } from '@patternfly/react-core';
+import { Alert, FormGroup, Text } from '@patternfly/react-core';
 import { UpdateObjectAtPropAndValue } from '~/pages/projects/types';
 import { CreatingInferenceServiceObject } from '~/pages/modelServing/screens/types';
 import { ServingRuntimeKind } from '~/k8sTypes';
@@ -28,10 +28,6 @@ const InferenceServiceServingRuntimeSection: React.FC<
   const placeholderText =
     servingRuntimes.length === 0 ? 'No model servers available to select' : 'Select a model server';
 
-  if (!loaded && !currentServingRuntime && data.project !== '') {
-    return <Skeleton />;
-  }
-
   if (loadError) {
     return (
       <Alert title="Error loading model servers" variant="danger">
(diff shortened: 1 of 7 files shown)


frontend/src/pages/modelServing/screens/projects/InferenceServiceModal/InferenceServiceServingRuntimeSection.tsx
```diff
@@ -58,16 +58,17 @@ const InferenceServiceServingRuntimeSection: React.FC<
         }))}
         toggleProps={{ id: 'inference-service-model-selection' }}
         isFullWidth
-        value={data.servingRuntimeName}
```
mturley, line 61: I think maybe we should still pass this value in in case SimpleSelect changes in the future to make it matter. The fact that passing `toggleLabel` makes `value` no longer used is not part of the interface contract of the component, just an implementation detail that could change.

DaoDaoNoCode: Thanks, updated

mturley (approved):
LGTM! Works well locally, Good stuff

frontend/src/pages/modelServing/screens/projects/InferenceServiceModal/InferenceServiceServingRuntimeSection.tsx
```diff
@@ -56,6 +52,7 @@ const InferenceServiceServingRuntimeSection: React.FC<
           key: servingRuntime.metadata.name,
           label: getDisplayNameFromK8sResource(servingRuntime),
         }))}
+        isSkeleton={!loaded && data.project !== ''}
```
mturley, line 55: This removes the `&& !currentServingRuntime` case from the existing logic here, was that intentional?

mturley (commented):
One last thing, sorry @DaoDaoNoCode

frontend/src/pages/modelServing/screens/projects/InferenceServiceModal/InferenceServiceServingRuntimeSection.tsx
DaoDaoNoCode: Hi, yes. If `currentServingRuntime` is present, it will always fall into
```
return (
      <FormGroup label="Model server">
        <Text>{getDisplayNameFromK8sResource(currentServingRuntime)}</Text>
      </FormGroup>
    );
```
above.
So the logic of checking `!currentServingRuntime` is not needed here anymore.

mturley: Ah! Good call, ok.

mturley (approved):
LGTM, thanks @DaoDaoNoCode !

PR #3417, opened 2024-11-01
update model serving compatibility for connection types

christianvogt:
https://issues.redhat.com/browse/RHOAIENG-14813

## Description

Connections:
image

Connection types:
image

image

## How Has This Been Tested?

- Enable the connection types feature flag `disableConnectionTypes`
- As an admin, navigate to `Settings -> Connection types`
- Observe the preinstalled types show the correct model serving compatibility labels.
- Duplicate a preinstalled type or hand craft a new type with appropriate model serving fields with appropriate env var names (can use the dropdown to auto add matching types).
- Observe the alert indicates the compatible model serving types.
- Create a new project
- Go to the project details -> Data Connections tab (testing data connection compatibility
- Create a new data connection (include a value for the Bucket field)
- Go to the Connections tab
- Observe the new connection is compatible with s3 object storage
- Create a new connection of type URI or OCI or s3
- Observe the new connection is compatible with the respective type

## Test Impact

Updated tests.

## Request review criteria:

If you have UI changes: 

cc @simrandhaliw

### DIFF
--- a/frontend/src/concepts/connectionTypes/utils.ts
+++ b/frontend/src/concepts/connectionTypes/utils.ts
@@ -1,5 +1,5 @@
 import { SecretKind } from '~/k8sTypes';
-import { translateDisplayNameForK8s } from '~/concepts/k8s/utils';
+import { getDisplayNameFromK8sResource, translateDisplayNameForK8s } from '~/concepts/k8s/utils';
 import { K8sNameDescriptionFieldData } from '~/concepts/k8s/K8sNameDescriptionField/types';
 import {
   Connection,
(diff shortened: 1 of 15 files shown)


frontend/src/concepts/connectionTypes/utils.ts
```diff
@@ -121,56 +118,83 @@ export const S3ConnectionTypeKeys = [
...
   }
 > = {
-  [ModelServingCompatibleConnectionTypes.ModelServing]: {
-    name: 'S3 compatible object storage',
+  [ModelServingCompatibleTypes.S3ObjectStorage]: {
+    name: ModelServingCompatibleTypes.S3ObjectStorage,
+    resource: 's3',
     envVars: S3ConnectionTypeKeys,
+    isConnectionCompatible: (connection) =>
+      connection.metadata.annotations['opendatahub.io/connection-type'] === 's3' &&
```
emilys314, line 141: So this will not show model serving compatibility for connections with aws env vars but not the type of `s3` I believe? 
image

christianvogt: Correct because the `type` makes it into the `storage-config` and is necessary for model serving to work.
That being said, if all env vars for s3 are populated then we would add the annotation on edit / save. Unless I missed a case.

jeff-phillips-18 (approved):
Works well. LGTM

PR #3421, opened 2024-11-01
update to use connection-type-ref annotation

christianvogt:
Depends on https://github.com/opendatahub-io/odh-dashboard/pull/3417

## Description

Problem
- Can’t use `opendatahub.io/connection-type` label arbitrarily because a controller reads this value to populate `storage-config` `Secret` with `s3` properties used by `InferenceService`
- Doesn’t seem to be a way to change the type of the referenced secret in the `InferenceService` outside of our connection `Secret`.
- The `opendatahub.io/managed=true` annotation is used by a controller to identify secrets which will be used to populate the `storage-config` `InferenceService` will fail unless the type related to the reference `Secret` is of type `s3`	

Solution
- At connection creation / edit time, set `opendatahub.io/managed=true` and `opendatahub.io/connection-type=s3` if the connection is an s3 model serving compatible secret (ie it also has a bucket)
- `opendatahub.io/connection-type-ref` instead to reference back the connection type with a fallback to `opendatahub.io/connection-type`
- Ensure model serving only includes URI compatible and s3 compatible (including annotations and labels) in the list of available connection types

No visual impact to the UI.
However if a connection is created from an s3 connection type but the bucket field is omitted, it will not be marked as model serving compatible in the connections table.
cc @simrandhaliw 

## How Has This Been Tested?

- enable the connection types feature flag `disableConnectionTypes`
- create duplicate s3 compatible connection types so that the configmap resource isn't named `s3`
- create some older data connection to test compatibility
- create some new connections:
    - s3 with and without bucket
    - s3 from the non pre-installed type
    - uri
    - some other arbitrary connections
- Observe only s3 connection with a bucket are marked as model serving compatible
- Observe connections with a URI env var are marked as model serving compatible
- Ensure all connections are assignable to a workbench
- Ensure only model serving compatible connections are usable in model deployments

## Test Impact

- updated some tests to align with the new use of the labels

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/__tests__/cypress/cypress/tests/mocked/connectionTypes/connectionTypes.cy.ts
+++ b/frontend/src/__tests__/cypress/cypress/tests/mocked/connectionTypes/connectionTypes.cy.ts
@@ -100,6 +100,9 @@ describe('Connection types', () => {
       },
       { success: true },
     ).as('delete');
+
+    connectionTypesPage.shouldHaveConnectionTypes();
+
     cy.interceptOdh('GET /api/connection-types', [
       mockConnectionTypeConfigMap({}),
       mockConnectionTypeConfigMap({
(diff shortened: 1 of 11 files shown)


jeff-phillips-18 (approved):
works well, LGTM

PR #3426, opened 2024-11-04
Labels editing and validation

YuliaKrimerman:
[RHOAIENG-6994](https://issues.redhat.com/browse/RHOAIENG-6994) 
[RHOAIENG-6995](https://issues.redhat.com/browse/RHOAIENG-6995)
## Description

[

https://github.com/user-attachments/assets/3d4622f7-1874-4617-bc4c-e28f83faf65a

](url)

When going to model/version Detail, editing the labels are behaving as inline labels and also now have a validation for a case a new label already exists ( side not the background of the error message is red in paternfly )

## How Has This Been Tested?

on the UI

## Test Impact

Working on adding test while getting feedback on the implementation 

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/components/EditableLabelsDescriptionListGroup.tsx
+++ b/frontend/src/components/EditableLabelsDescriptionListGroup.tsx
@@ -22,6 +22,7 @@ export const EditableLabelsDescriptionListGroup: React.FC<EditableLabelsProps> =
 }) => {
   const [isEditing, setIsEditing] = useState(false);
   const [isSavingEdits, setIsSavingEdits] = useState(false);
+  const [hasSavedEdits, setHasSavedEdits] = useState(false);
   const [unsavedLabels, setUnsavedLabels] = useState(labels);
 
   const validateLabels = (): string[] => {
@@ -222,6 +223,7 @@ export const EditableLabelsDescriptionListGroup: React.FC<EditableLabelsProps> =
         try {
           await onLabelsChange(unsavedLabels);
         } finally {
+          setHasSavedEdits(true);
           setIsSavingEdits(false);
           setIsEditing(false);
         }
@@ -231,7 +233,11 @@ export const EditableLabelsDescriptionListGroup: React.FC<EditableLabelsProps> =
         setIsEditing(false);
       }}
     >
-      <LabelGroup data-testid="display-label-group">
+      <LabelGroup
+        data-testid="display-label-group"
+        defaultIsOpen={hasSavedEdits}
+        key={String(hasSavedEdits)} // Force this to fully remount when we change defaultIsOpen
+      >
         {labels.map((label) => (
           <Label key={label} color="blue" data-testid="label">
             {label}


YuliaKrimerman:
@yih-wang Haley please take a look as well, and let me know what you think

frontend/src/components/EditableLabelsDescriptionListGroup.tsx
```diff
@@ -1,223 +1,170 @@
...
                   color="blue"
                   variant="outline"
-                  isOverflowLabel
-                  onClick={toggleAddLabelModal}
+                  isEditable
+                  editableProps={{
+                    'aria-label': 'Add label',
+                    defaultValue: '',
+                    'data-testid': 'add-label-input',
+                  }}
```
mturley, line 88: The way I understand the mockup, I don't think we want the "Add label" button to be an editable label itself. It can just be a button like before, but instead of its `onClick` opening a modal, it should add a new label called "New label" via `setUnsavedLabels()`.

mturley (commented):
One quick comment from looking at the video - I will give a more thorough review when I get a chance

frontend/src/components/EditableLabelsDescriptionListGroup.tsx
yih-wang: +1 to @mturley.
@YuliaKrimerman here's a PF demo of how should the `Add label` work: https://www.patternfly.org/components/label/#editable-label-group-with-add-button Hope it helps.

frontend/src/components/EditableLabelsDescriptionListGroup.tsx
```diff
@@ -1,223 +1,170 @@
...
             ))}
           </LabelGroup>
+          {Object.keys(labelErrors).length > 0 && (
+            <Alert
+              data-testid="label-error-alert"
+              variant={AlertVariant.danger}
+              isInline
+              title={Object.values(labelErrors)[0]}
+              aria-live="polite"
+            />
```
ppadti, line 135: use `isPlain` prop to remove colored background.

YuliaKrimerman: Got it, thank you!

YuliaKrimerman:
@mturley @yih-wang I think I got it now. Updated the video.

yih-wang:
Thanks for the updates @YuliaKrimerman! My only comment is when handling duplicate label errors, the newly created label should be marked in red to indicate the error. From the video, it looks like the existing label is flagged instead.

yih-wang:
And one question to the loading time - from the video the newly added label takes a while to load after saving. Would this delay also occur in the production environment, or is it only in development environment? It’s a bit confusing to me before the label is loaded, so if this is expected in production, adding a loading state might help improve clarity.

yih-wang:
Thanks for the updates @YuliaKrimerman! All look good now!

One thing that wasn’t part of the original design proposal but stood out to me as potentially confusing from the video - currently, after saving an edit, the label list keeps its original expanded/collapsed state. This means if the list was collapsed before editing, and the user enters the edit mode then saves any changes, the newly added label isn’t visible because the list stays collapsed. To avoid this potential confusion, it might be clearer to have the label list automatically expand after any edit is saved, regardless of its original collapsed/expanded state.

Anyway it doesn't belong to the original scope of the issue, and I don't know how much work it needs, so not sure whether we should create a new story to improve this experience separately in a future sprint @mturley

manaswinidas (changes requested):
@YuliaKrimerman can you add testing instructions in the "How has this been tested?" section of PR description?

frontend/src/components/EditableLabelsDescriptionListGroup.tsx
```diff
@@ -1,223 +1,193 @@
...
-      <DashboardDescriptionListGroup
-        title={title}
-        isEmpty={labels.length === 0}
-        contentWhenEmpty={contentWhenEmpty}
-        isEditable={!isArchive}
-        isEditing={isEditing}
-        isSavingEdits={isSavingEdits}
-        contentWhenEditing={
+    <DashboardDescriptionListGroup
+      data-testid="editable-labels-group"
```
manaswinidas, line 94: Is this test-id necessary? I don't see this test-id used anywhere.

frontend/src/components/EditableLabelsDescriptionListGroup.tsx
```diff
@@ -1,223 +1,193 @@
...
+      contentWhenEmpty={contentWhenEmpty}
+      isEditable={!isArchive}
+      isEditing={isEditing}
+      isSavingEdits={isSavingEdits}
+      isSaveDisabled={Object.keys(labelErrors).length > 0}
+      contentWhenEditing={
+        <>
           <LabelGroup
-            data-testid="label-group"
+            data-testid="editable-label-group"
```
manaswinidas, line 108: I don't see this testid used anywhere.

frontend/src/pages/modelRegistry/screens/ModelVersions/ModelDetailsView.tsx
```diff
@@ -62,6 +64,7 @@ const ModelDetailsView: React.FC<ModelDetailsViewProps> = ({
                 )
                 .then(refresh)
             }
+            data-testid="model-labels"
```
manaswinidas, line 67: I don't see this testid anywhere

manaswinidas (changes requested):

frontend/src/components/EditableLabelsDescriptionListGroup.tsx
```diff
@@ -1,223 +1,191 @@
...
+      setUnsavedLabels((prev) => {
+        if (currentLabel) {
+          return [...prev, newText];
+        }
+        const filtered = prev.filter((label) => label !== currentLabel);
+        return [...filtered, newText];
+      });
+      setLabelErrors({});
+    }
+  };
```
mturley, line 67: I'm not sure why the `setUnsavedLabels` behavior is different if there's an error vs if there isn't. This may resolve itself if you move the validation from here to render time, but I'm pretty sure `handleEditComplete` should update the unsaved labels state the same way no matter whether there is an error to show. I think this first call in your `if (error)` case is correct:

```ts
      setUnsavedLabels((prev) => {
        const filtered = prev.filter((label) => label !== currentLabel);
        return [...filtered, newText];
      });
```

Your other case is making it so any time edit focus is lost, if the user didn't enter duplicate label text, the old label is not filtered out on update and we end up with another duplicate:

https://github.com/user-attachments/assets/dfc40c98-19b2-4f21-9661-ea4637c4cca0

mturley (changes requested):
Hey @YuliaKrimerman, progress is looking great here but we need a bit of refactoring.

frontend/src/components/EditableLabelsDescriptionListGroup.tsx
```diff
@@ -1,223 +1,191 @@
...
-    setUnsavedLabels(copy);
+    return null;
   };
+
+  const handleEditComplete = (
+    _event: MouseEvent | KeyboardEvent,
+    newText: string,
+    currentLabel?: string,
+  ) => {
+    const error = validateLabel(newText, currentLabel);
```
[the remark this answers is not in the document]
YuliaKrimerman: @yih-wang Haley, we would need your input in this case on how to handle double duplicates, should we show them both like this? Should we even allow multiple duplicates at edit time? Let me know what you think : this is how it looks with what Mike suggested ( if we want to take this route please let me know which text to use)
Screenshot 2024-11-18 at 10 59 22 AM
Screenshot 2024-11-18 at 11 05 34 AM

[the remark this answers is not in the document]
yih-wang: Hi @YuliaKrimerman, I've reviewed this pattern with Anges, who is finalizing the ux validation guidelines. Her suggestion is to allow multiple duplicates at the edit time.

@kaedward, I'd like to hear your input on the validation message here -  Currently, we show the error message below when there's only one label duplicated
`‘mylabel’ already exists. Use a unique name that does not match any existing label or property key.` 

But how should the error message look like when there're multiple duplicates? Should it be a general message like
`Labels duplicated. Use a unique name that does not match any existing label or property key.` 
or a specific message which list all the duplicated labels' names like
`‘mylabel’ and 'new' already exist. Use a unique name that does not match any existing label or property key.`

yih-wang:
@YuliaKrimerman When I reviewed the validation pattern with Anges, she pointed out another inconsistency with the upcoming consistency guidelines - the ✓ should always remain enabled even when there're duplication errors. This provides better accessibility and aligns with the new form validation pattern.

Apologies for the adjustment from the original design, I think it's better to address this comment and change the ✓ behavior to be always enabled.

frontend/src/components/EditableLabelsDescriptionListGroup.tsx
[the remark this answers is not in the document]
kaedward: @yih-wang How do these sound? Do we need to specify that the label name must not match a property? 
The label name should be bolded, not in quotation marks. 

ONE ERROR:
**label-name** appears more than once. Ensure that each label is unique.

MORE THAN 1 ERROR:
**label-name** and **label-name2** appear more than once. Ensure that each label is unique.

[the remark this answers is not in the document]
yih-wang: > Do we need to specify that the label name must not match a property?
@kaedward Yes, to be accurate, the label shouldn't match any existing labels or property keys. The pattern you suggested looks good to me. But I'm wondering whether we should say 'label-name already exists' rather than 'label-name appears more than once' because it's already used elsewhere for duplication validation.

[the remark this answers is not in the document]
kaedward: ONE ERROR:
label-name appears more than once. Ensure that each label is unique and does not match any existing property key.

MORE THAN 1 ERROR:
label-name and label-name2 appear more than once. Ensure that each label is unique and does not match any existing property key.

@yih-wang Here ya go!

YuliaKrimerman:
Hi @yih-wang @kaedward @mturley. Please take a look at the video recording. It has all the requested changes so far : Handling double duplications, proper error display based on amount of duplications and enabling the save button even when error exist. Also added some top padding above the red alert as Mike suggested (see screenshot, not video). Mike if you have a chance you can also do a re-review while I will work on fixing the last test, and brace for any incoming comments if needed. 
https://github.com/user-attachments/assets/ce4bfed8-d39b-476c-a408-da627a19d74e

Screenshot 2024-11-20 at 1 07 58 PM

mturley:
@YuliaKrimerman I'll do a full review asap, but just looking at that screenshot it looks like you lost the red color on the errored labels?

YuliaKrimerman:
@mturley Haha no, it's just I screenshot with only the original showing, and not the duplicate. In the video the red is there :)

mturley:
Ohh, I misread sorry!

yih-wang:
Thanks @YuliaKrimerman, the latest updates look good! 

The only question I have is do we decide whether or not address my [earlier comment](https://github.com/opendatahub-io/odh-dashboard/pull/3426#issuecomment-2461755196) with this pr? Just make sure it's not overlooked.

> One thing that wasn’t part of the original design proposal but stood out to me as potentially confusing from the video - currently, after saving an edit, the label list keeps its original expanded/collapsed state. This means if the list was collapsed before editing, and the user enters the edit mode then saves any changes, the newly added label isn’t visible because the list stays collapsed. To avoid this potential confusion, it might be clearer to have the label list automatically expand after any edit is saved, regardless of its original collapsed/expanded state.
> 
> Anyway it doesn't belong to the original scope of the issue, and I don't know how much work it needs, so not sure whether we should create a new story to improve this experience separately in a future sprint

YuliaKrimerman:
@yih-wang Yeah I didn't address it as part of this issue. I think opening a separate story for this might be a good idea.

gitdallas (approved):

frontend/src/components/EditableLabelsDescriptionListGroup.tsx
```diff
@@ -1,223 +1,243 @@
...
+
+    const duplicatesMap = new Map<string, number>();
+    labelsList.forEach((label) => {
+      duplicatesMap.set(label, (duplicatesMap.get(label) || 0) + 1);
+    });
+
+    const duplicateLabels: string[] = [];
+    duplicatesMap.forEach((count, label) => {
+      if (count > 1) {
+        duplicateLabels.push(label);
```
mturley, line 37: @YuliaKrimerman I hate to drag this out a bit longer but can we add a test or two that covers the duplicate label validation to make codecov happier?

frontend/src/components/EditableLabelsDescriptionListGroup.tsx
```diff
@@ -1,223 +1,243 @@
...
+        const lastLabel = duplicateLabels.pop();
+        const formattedLabels = duplicateLabels.map((label) => `**${label}**`).join(', ');
+        errors.push(
+          `${formattedLabels} and **${lastLabel}** already exist. Ensure that each label is unique and does not match any existing property key.`,
+        );
+      }
+    }
+
+    labelsList.forEach((label) => {
+      if (!labelsList.includes(label) && allExistingKeys.includes(label)) {
```
mturley, line 56: This logic is wrong. `!labelsList.includes(label)` will always be false here because we're looping over labelsList, the label will always be in labelsList. So this validation case is never hit, and I am allowed to create a label that already exists as a key in the properties table below, which replaces it and loses that property value.

I think instead, we need to either consider `allExistingKeys` when we are setting up the `duplicatesMap` or we need to have this logic be somehow `if (weDontAlreadyHaveAnErrorForThisLabel && allExistingKeys.includes(label))`. In that case, the message could be specific to the fact that the label matches an existing property key, like `**${label}** already exists as a property key. Ensure that each label is unique and does not match any existing property key.`

frontend/src/components/EditableLabelsDescriptionListGroup.tsx
```diff
@@ -1,223 +1,243 @@
...
-  const submitAddLabelModal = (event?: React.FormEvent) => {
-    event?.preventDefault();
-    if (!addLabelModalSubmitDisabled) {
-      addUnsavedLabel(addLabelInputValue);
-      toggleAddLabelModal();
+
+    setUnsavedLabels((prev) => [...prev, newLabel]);
+  };
+
+  const labelErrors = validateLabels(unsavedLabels);
```
mturley, line 116: `validateLabels` doesn't really need to take `unsavedLabels` as an argument at all, it can just directly use `unsavedLabels` in its body instead of its `labelsList`.

frontend/src/components/EditableLabelsDescriptionListGroup.tsx
```diff
@@ -1,223 +1,243 @@
...
+              data-testid="add-label-button"
+              variant="plain"
+              className="pf-v5-c-label pf-m-outline"
+              onClick={addNewLabel}
+              isDisabled={isSavingEdits}
+              style={{
+                border: '2px solid #d2d2d2',
+                color: '#0066CC',
+                backgroundColor: 'transparent',
+              }}
```
mturley, line 183: These inline styles override the border/color at all times so there is no longer a hover state. We should avoid inline styles where possible.

I don't think we need our own Button here. Can we not use the `addLabelControl` prop of LabelGroup like we had before? That automatically adds a button with the correct hover styles while editing like in this PF example. https://www.patternfly.org/components/label#editable-label-group
I locally reverted that part of this change and put back that `addLabelControl` but just had it call your new `addNewLabel` function, and it appears to work fine.

mturley (changes requested):
@YuliaKrimerman I'm so sorry for the delay, don't kill me, I have comments.....

frontend/src/components/EditableLabelsDescriptionListGroup.tsx
```diff
@@ -1,223 +1,243 @@
...
+            <Alert
+              ref={alertRef}
+              data-testid="label-error-alert"
+              variant={AlertVariant.danger}
+              isInline
+              title="Label validation errors:"
+              aria-live="polite"
+              isPlain
+              tabIndex={-1}
+              style={{ marginTop: '16px' }}
```
mturley, line 198: Instead of inline styles here, you can `import spacing from '@patternfly/react-styles/css/utilities/Spacing/spacing';` and then use `className={spacing.mtMd}` which will set a margin-top to the PF medium spacer, which is 16px.

frontend/src/components/EditableLabelsDescriptionListGroup.tsx
```diff
@@ -1,223 +1,245 @@
...
+      contentWhenEditing={
+        <>
           <LabelGroup
-            data-testid="label-group"
+            data-testid="editable-label-group"
             isEditable={!isSavingEdits}
-            numLabels={unsavedLabels.length}
+            numLabels={10}
+            expandedText="Show Less"
+            collapsedText="Show More"
```
mturley, line 153: Ooooone last thing - we should set `numLabels` back to `unsavedLabels.length` and remove the `expandedText` and `collapsedText` props. When editing labels they should always be expanded, the show more/less behavior is only when they are not being edited which is handled by the other LabelGroup rendered in non-edit mode. Otherwise, if they are collapsed and you click "Add label" the new label doesn't appear until you expand.

mturley (changes requested):

mturley:
@yih-wang for posterity since this work was discussed privately - we did address your comment above in this PR about expanded/collapsed state when saving changes:

> currently, after saving an edit, the label list keeps its original expanded/collapsed state. This means if the list was collapsed before editing, and the user enters the edit mode then saves any changes, the newly added label isn’t visible because the list stays collapsed. To avoid this potential confusion, it might be clearer to have the label list automatically expand after any edit is saved, regardless of its original collapsed/expanded state.

When the user saves their edited labels the labels will now become expanded.

PR #3427, opened 2024-11-04
Show no-default alert correctly when first loading storage classes settings page

DaoDaoNoCode:
JIRA: https://issues.redhat.com/browse/RHOAIENG-15219

## Description

A one-liner change to update the condition to show the alert correctly.

Screenshot 2024-11-04 at 12 37 41 PM

## How Has This Been Tested?

1. Go to OpenShift console and find all the storage classes
2. Find the default storage class, change the annotation `storageclass.kubernetes.io/is-default-class: 'true'` to `storageclass.kubernetes.io/is-default-class: 'false'`
3. Remove the annotation `opendatahub.io/sc-config` from all the storage classes
4. Go back to the Dashboard storage classes settings page and refresh
5. You should be able to see the alert
6. Change the storage class in step 2 back to default
7. Remove the annotation `opendatahub.io/sc-config` from all the storage classes again
8. Go back to the Dashboard storage classes settings page and refresh
9. You shouldn't see the alert even if the annotations are patch

## Test Impact

Added a cypress test to verify the change

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/__tests__/cypress/cypress/pages/storageClasses.ts
+++ b/frontend/src/__tests__/cypress/cypress/pages/storageClasses.ts
@@ -30,6 +30,10 @@ class StorageClassesPage {
     return cy.findByTestId('storage-classes-empty-state');
   }
 
+  findNoDefaultAlert() {
+    return cy.findByTestId('no-default-storage-class-alert');
+  }
+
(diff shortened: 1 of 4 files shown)


Gkrumbach07 (approved):
/lgtm

nice, thanks for adding tests

PR #3430, opened 2024-11-05
Fix conda notebook image reference

tarukumar:
During testing of the downstream build, it was observed that when we activate Anaconda from the dashboard, the activation is successful, but the notebook image is not created. This is because the Cornjob, which is responsible for creating the Conda notebook image stream resource, has an incorrect image reference."

From:
```python
"quay.io/modh/odh-anaconda-notebook@sha256:sha256:acfa4bc06bbd0b4640844c5403272d6138070f391264dba238441c5dc64de505"
``` 

This:
```shell
"quay.io/modh/odh-anaconda-notebook@sha256:acfa4bc06bbd0b4640844c5403272d6138070f391264dba238441c5dc64de505"
```


manosnoam (commented):
/lgtm

PR #3439, opened 2024-11-05
Model Registry Selector: don't navigate on selection unless necessary

mturley:
Resolves https://issues.redhat.com/browse/RHOAIENG-15471

## Description

Prevents infinite loop on mounting the ModelRegistrySelectorNavigator when there is only one registry, caused by  https://github.com/opendatahub-io/odh-dashboard/pull/3345. The `onSelection` callback will be called on first render, which was triggering a navigation that would call it again, triggering an infinite loop. Now it will only navigate if we are not already on the registry we want to navigate to.

## How Has This Been Tested?

Tested locally.

## Test Impact

I attempted to write a cypress test for this, but the infinite loop was not happening in the Cypress environment so I couldn't get the test to fail without the fix. I gave up 🤷 

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/pages/modelRegistry/screens/ModelRegistrySelectorNavigator.tsx
+++ b/frontend/src/pages/modelRegistry/screens/ModelRegistrySelectorNavigator.tsx
@@ -11,15 +11,17 @@ const ModelRegistrySelectorNavigator: React.FC<ModelRegistrySelectorNavigatorPro
   ...modelRegistrySelectorProps
 }) => {
   const navigate = useNavigate();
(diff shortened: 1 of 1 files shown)


ppadti:
Tested locally, works fine
/lgtm

manaswinidas (approved):
/lgtm

I don't see the `useEffect: maximum depth exceeded` error in the developer console when we click on `Model Registry` in the nav bar.

gitdallas (approved):

PR #3440, opened 2024-11-06
guard against corrupt connection types and support previous categories

christianvogt:
Address review comments from https://issues.redhat.com/browse/RHOAIENG-14740

## Description

When creating a new connection type, previously used categories will be shown as options in the categories dropdown along with pre-defined set:

When data in a connection type fails to parse, it was causing the app to error out. Now whenever fetch a list of connection types, the corrupt ones are omitted from the list and no error is thrown which will error out the app.
When using a direct link to edit or duplicate a corrupt connection type, the following error is shown:
image

## How Has This Been Tested?

- enable the connection type feature flag `disableConnectionTypes`

Categories:
- Navigate to Settings -> Connection types
- Create a new connection type
- add custom categories and save
- create a 2nd connection type
- Observe all previously added categories are available in the select menu

Corrupt data:
- create a connection type (name it `corrupt`
- using OpenShift console edit the connection type configmap to have invalid json 
- OR using the cmd line: `oc patch configmaps -n opendatahub corrupt -p '{"data":{"category":"[]","fields": "[[]"}}'`
- navigate directly to project list / details pages and the connections tab in the project details
- Observe no errors
- Navigate to Settings -> Connection types
- Observe no errors
- go directly to edit the corrupt connection type via URL: `/connectionTypes/edit/corrupt`
- Observe the error screen is shown

## Test Impact

Added unit and cypress tests.

## Request review criteria:

If you have UI changes: 

cc @simrandhaliw @andrewballantyne

### DIFF
--- a/frontend/src/services/connectionTypesService.ts
+++ b/frontend/src/services/connectionTypesService.ts
@@ -12,22 +12,29 @@ import {
 export const fetchConnectionTypes = (): Promise<ConnectionTypeConfigMapObj[]> => {
   const url = `/api/connection-types`;
   return axios
-    .get(url)
+    .get<ConnectionTypeConfigMap[]>(url)
     .then((response) =>
-      response.data.map((cm: ConnectionTypeConfigMap) => toConnectionTypeConfigMapObj(cm)),
+      response.data.reduce<ConnectionTypeConfigMapObj[]>((acc, cm) => {
+        try {
+          acc.push(toConnectionTypeConfigMapObj(cm));
+        } catch (e) {
+          // ignore those which fail to parse
+        }
+        return acc;
+      }, []),
     )
     .catch((e) => {
-      throw new Error(e.response.data.message);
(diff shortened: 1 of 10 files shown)


frontend/src/services/connectionTypesService.ts
```diff
@@ -14,10 +14,18 @@ export const fetchConnectionTypes = (): Promise<ConnectionTypeConfigMapObj[]> =>
   return axios
     .get(url)
     .then((response) =>
-      response.data.map((cm: ConnectionTypeConfigMap) => toConnectionTypeConfigMapObj(cm)),
+      response.data
+        .map((cm: ConnectionTypeConfigMap) => {
```
jeff-phillips-18, line 18: use `reduce` instead of `map` and `filter`?

jeff-phillips-18 (approved):
LGTM

PR #3447, opened 2024-11-06
support static prefix for k8s resource name

christianvogt:
https://issues.redhat.com/browse/RHOAIENG-15509

## Description

Adds a `staticPrefix` option to the `K8sNameDescriptionField` component.
When this is used in conjunction with `safePrefix`, the prefix will always be applied.

Updated connection types form to use `staticPrefix` along with a `safePrefix` of `ct-`.

When the user types the expected prefix, the prefix isn't duplicated:
image
image

Similarly, if the prefix is omitted by the user, the prefix is applied:
image
image

The user can change still customize the resource name.
image

When the name is left blank:
image

## How Has This Been Tested?

- enable the connection types feature flag
- as an admin create a new connection type
- use various permutations of names and custom resources
- ensure the resource is always created with a prefix of `ct-`

## Test Impact

Updated tests

## Request review criteria:

If you have UI changes: 

cc @andrewballantyne @simrandhaliw


simrandhaliw:
LGTM!

frontend/src/concepts/k8s/utils.ts
```diff
@@ -44,12 +46,13 @@ type AdditionalCriteriaApplied = Record<
...
 export const translateDisplayNameForK8sAndReport = (
   name: string,
-  { safeK8sPrefix, maxLength }: AdditionalCriteriaForTranslation = {},
+  { safeK8sPrefix, isStaticPrefix, maxLength }: AdditionalCriteriaForTranslation = {},
 ): [string, AdditionalCriteriaApplied] => {
   const appliedCriteria: AdditionalCriteriaApplied = {
     autoGenerated: false,
     safeK8sPrefix: false,
     maxLength: false,
+    isStaticPrefix: false,
```
andrewballantyne, line 55:
```suggestion
    staticPrefix: false,
```
Maybe the wording here is weird -- they are all booleans, because they are status of what happened... `appliedCriteria.staticPrefix` imo is better than `appliedCriteria.isStaticPrefix` -- it's all the applied criteria so everything is prefixed with `is` or none of them are.

andrewballantyne (commented):
Minor comment -- for consistency...

Testing this on cluster.

frontend/src/concepts/k8s/utils.ts
andrewballantyne: @christianvogt this is not a blocker to me... but are you in the middle of doing this?

andrewballantyne:
/approve

Works as expected on cluster 👍

frontend/src/concepts/k8s/utils.ts
christianvogt: I'll fix this. I agree with you.

christianvogt: Added a new commit with just the rename of this.

PR #3448, opened 2024-11-07
fix adding dropdown fields to section

christianvogt:
## Description

Fixes validation with connection type dropdown items. Labels are not required and therefore empty labels should not cause the form to become invalid.

Fixes adding fields to sections.

## How Has This Been Tested?

-  enable the connection type feature flag
- create a new connection type
- add a section
- click the add field button for the section
- select the field type as drop down
- add a name and env var
- add an item and supply value `v1`
- add another item and supply value `v2`
- Ensure the modal can be saved
- save the modal
- observe the item is correctly added to the section
- try adding another field to the same or another section

## Test Impact

added test case

## Request review criteria:

If you have UI changes: 

cc @simrandhaliw


emilys314:
looks good! I tested the empty labels on the dropdown and the adding of a field to a section

/lgtm

PR #3458, opened 2024-11-08
speed up development env with swc-loader and hmr

christianvogt:
With this change, startup of dev server is faster and provides hot module reloading without full page refresh.

- Replace `ts-loader` with `swc-loader` but only for dev mode.
- Add webpack plugin `ReactRefreshWebpackPlugin` for fast hot module reloading.
- Add webpack plugin `ForkTsCheckerWebpackPlugin` to fork type checking to a separate thread.
- Add webpack plugin https://github.com/stephencookdev/speed-measure-webpack-plugin for measuring startup times.
- Update tsconfig target to `es2016`

To measure startup times use: `MEASURE=true npm run start:dev:ext`.

Start up times for frontend `npm run start:dev:ext`
```
swc-loader:
General output time took 33.89 secs
General output time took 31.31 secs
General output time took 31.66 secs

ts-loader:
General output time took 46.79 secs
General output time took 45.35 secs
General output time took 41.036 secs
```

Prior to any changes:
```
General output time took 1 min, 27.1 secs
General output time took 1 min, 25.41 secs
General output time took 1 min, 27.89 secs
```


jeff-phillips-18:
Start up is faster
Updates after a code change is a **_lot_** faster.

I haven't seen any issues.

/lgtm

(now if only we could get it to stop faster on a cntrl-c)

DaoDaoNoCode (approved):
Tested, hot reload is amazing. Loading is a lot faster.
/lgtm

mturley (commented):
This looks amazing, can't wait to get it merged and use it.

@christianvogt I did notice one weird thing - when I pass `MEASURE=true`, hot reloading isn't working and in fact the server appears to break after a change, and then reloading the page gives me a blank white screen with this console error:

Screenshot 2024-11-12 at 4 16 27 PM


Without `MEASURE` set, everything works great. Do we care about that? If we do want to fix it maybe that can be a followup PR.

mturley:
Ah got it. No worries then, I don't see a problem with merging.

mturley (approved):
Real good stuff.

PR #3459, opened 2024-11-11
Remove explicit support for OCI model serving compatibility

emilys314:
towards [[RHOAIENG-14748] enable connection types feature](https://issues.redhat.com/browse/RHOAIENG-14748)

## Description

After review and discussion from Jeff DeMoss, we decided to remove the OCI ootb connection type, because it's not fully supported yet. Many OCI model registries may require authentication, and requires additional kserve support before being usable (https://kserve.github.io/website/latest/modelserving/storage/oci/#enabling-modelcars)

image
image

## How Has This Been Tested?

- Make sure there are no references to "OCI" in the connection types admin page
- The ootb OCI connection type will also no longer exist (but you'd have to do this manually on your cluster if you already have it)
  - And fyi it will still show up in model serving as a "URI" compatible type

## Test Impact

- tests updated for removing OCI

## Request review criteria:

If you have UI changes: 


jenny-s51 (approved):
Updates LGTM @emilys314 ! Deleted the `oci-compliant-registry-v1` configMap from the shared cluster as discussed since it was showing up in the connection types list by default. No longer showing in the connect types list as expected. 

"Model serving compatible type" dropdown looks good too and no longer references OCI registry.

mturley (approved):
LGTM. Tested locally against the ods-qe-psi-08 cluster. I temporarily scaled down the operator in order to delete the `oci-compliant-registry-v1` configmap and verified that it is not an option in model serving or the connection types page.

PR #3462, opened 2024-11-11
Fix workbench new connection env warning / editing new connection / and save failure handling

emilys314:
Closes https://issues.redhat.com/browse/RHOAIENG-15584

## Description

When creating a new connection is created, it uses the connection made from `assembleConnection` which has its data inside `stringData` not `data` so check those. 

So this fixes 3 different issues oddly:
1. When creating a new connection to attach, it would not warn of duplicate env variables in the table anymore. This was because the `assembleConnection()` put the data into `stringData` not `data`
2. When saving a (new) connection, it would always add the new connection to the table, even if there was a failure. There was no success gate for adding it.
3. If you tried editing a connection you just created, it would not load the existing data you just entered, because it was in `stringData`

All of these are fixed by using the response of the save / replace and verifying it was successful 

Before:
image
After:
image

Note the fields
before:
image
after:
image

image
image

## How Has This Been Tested?

Duplicate env:
1. Create / edit a workbench
2. Attach an existing connection 
3. Create a new connection of the same type as the existing one and click save
4. See there is now a warning

Connection saving success gate:
1. Create / edit a workbench
2. Create a new connection and save it
3. Create another connection with the same name and save it
4. You'll see an error, close the modal.
5. With the fix, now that connection was not added to the table

Editing a new connection:
1. Create / edit a workbench
2. Create a new connection and save it
3. Edit the connection 
4. Make sure env var data has been populated with what you entered in step 2

## Test Impact

No updates, it's a bit tedious to test these edge cases since it requires a few steps across different components.

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/pages/projects/screens/spawner/connections/ConnectionsFormSection.tsx
+++ b/frontend/src/pages/projects/screens/spawner/connections/ConnectionsFormSection.tsx
@@ -20,6 +20,7 @@ import { Connection, ConnectionTypeConfigMapObj } from '~/concepts/connectionTyp
 import {
   filterEnabledConnectionTypes,
   getConnectionTypeDisplayName,
+  isConnection,
 } from '~/concepts/connectionTypes/utils';
 import { useWatchConnectionTypes } from '~/utilities/useWatchConnectionTypes';
 import { useNotebooksStates } from '~/pages/projects/notebook/useNotebooksStates';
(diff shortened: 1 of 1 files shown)


christianvogt:
Editing an existing connection ends up with a duplicate entry in the table.

emilys314:
> Editing an existing connection ends up with a duplicate entry in the table.

Fixed, a 4th issue.
1. Create / edit a workbench
2. Have an existing connection on the workbench
3. Edit the existing connection and save it (update the description and fields)
4. Save the connection
5. Make sure the description is updated in the table, and the fields are updated when editing again

christianvogt:
works as expected

/lgtm
/approve

PR #3470, opened 2024-11-13
Execution reference on the artifact view what execution produced the artifacts

pnaik1:
https://issues.redhat.com/browse/RHOAIENG-14415

## Description
added the execution reference in the artifact detail page that it came from and if it was cached then show the root execution

Screenshot 2024-11-20 at 3 10 13 PM

## How Has This Been Tested?

1. Run a pipeline which contains artifacts
2. Go to Artifact Page-> click on the artifactName
3. check if the Reference section is present 
4. Check if the execution link redirects to execution page
5. Check execution link is a root execution link if its cached

## Test Impact
Added cypress test cases

## Request review criteria:

If you have UI changes: 

cc @yannnz

### DIFF
--- a/frontend/src/__mocks__/mlmd/mockGetContextType.ts
+++ b/frontend/src/__mocks__/mlmd/mockGetContextType.ts
@@ -1,0 +1,14 @@
+import { ContextType } from '~/__mocks__/third_party/mlmd';
+import { GetContextTypeResponse } from '~/__mocks__/third_party/mlmd/generated/ml_metadata/proto/metadata_store_service';
+import createGrpcResponse, { GrpcResponse } from './utils';
+
+const mockedContextType: ContextType = {
+  id: 11,
+  name: 'system.PipelineRun',
+  properties: {},
+};
+
+export const mockGetContextType = (): GrpcResponse => {
+  const binary = GetContextTypeResponse.encode({ contextType: mockedContextType }).finish();
+  return createGrpcResponse(binary);
+};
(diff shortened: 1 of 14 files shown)


Gkrumbach07:
@pnaik1 this looks good. however i think we should also add the reference to the pipeline run in the list of references if possible. PMs request was looking to get back to run directly from the artifact

> I can go from the run view to an artifact, but I can’t go from an artifact list to which run it was produced by.

Gkrumbach07 (commented):
Code and functionality looks good otherwise

pnaik1:
@Gkrumbach07  `PMs request was looking to get back to run directly from the artifact`
Does this mean to list all the pipeline runs (even runs from cached execution) connected to the artifact or just the root pipeline run which created it
Ex:
```
Execution-1 -> run-id-1-> artifact-1
Execution-2(cached)->run-id-2-> artifact-1
```

the artifact details page should look something like this???

```
Pipeline run.      run-id-1
                   run-id-2
```
or just original/root runs

`Pipeline run.        run-id-1`

yannnz:
For the cached one, would it be something like this in the screenshot? Under the Reference section, list the Run and execution cached this artifact, then list the original run and execution created this artifact.  `Root` might sounds more accurate than `Original `, but we name it as`Original ` execution under the Execution detail page. 

Screenshot 2024-11-15 at 21 37 45

yannnz:
I left comments in the Slack: https://redhat-internal.slack.com/archives/C07C0FNHVPT/p1731912641438389

pnaik1:
@yannnz @Gkrumbach07 I have updated the screenshot?? I have given the name `Original run` and `Original execution` since this are the execution/ runs which actually produced the artifact, Can you have a look at it, if thats fine

PR #3471, opened 2024-11-13
run ForkTsCheckerWebpackPlugin only in dev env

christianvogt:
## Description

`ForkTsCheckerWebpackPlugin` was type checking all files including cypress files. Since we omit optional dependencies, errors were found type checking cypress files. 

The fix here is to only run `ForkTsCheckerWebpackPlugin` in dev environment as previous done before the change that introduced it.

## How Has This Been Tested?

- delete your `node_modules` folders
- `npm ci --omit=optional` from the root
- build the frontend dir: `npm run build`
- should succeed without error

## Test Impact

N/A

## Request review criteria:

If you have UI changes: 


mturley (approved):
Works for me locally

PR #3477, opened 2024-11-14
Add global pipeline runs table and related pages

DaoDaoNoCode:
JIRA: [RHOAIENG-15515](https://issues.redhat.com/browse/RHOAIENG-15515)

## Description

Mockup: https://www.figma.com/proto/etbN0ccCcZR1qKCpJXQNAg/Pipeline-phase-2?page-id=12031%3A86470&node-id=13557-54401&node-type=frame&viewport=684%2C2924%2C0.16&t=EJ7m1wDhh89XnSGs-1&scaling=scale-down-width&content-scaling=fixed&starting-point-node-id=13557%3A54381

Follow the mockup to add the global pipeline runs page and make all the changes as needed to adapt to it and reuse the components.

Screenshot 2024-11-14 at 4 12 45 PM

This PR does the following things:
1. Add the global pipeline runs page, which will show all the pipeline runs, and you can view both the experiment and pipeline version column on the table
2. Optimize the way to fetch the experiment for each run. Now it fetches all the experiments on the initial load, which significantly reduces the number of network calls
3. Allow users to compare runs from different experiments, and scope the experiment compare runs page within the experiment context (which means if you are comparing runs from an experiment, the Experiment column will be hidden and you can not add runs from other experiments to compare)
4. Add separate routes for run/schedule details, create run/schedule, and duplicate run/schedule so it won't affect other pages.
5. Fixing a bug that the runs table was making mlmd metrics calls that are not used, which made it very slowly loaded, now it's fixed
6. Re-organized some codes to make it more readable and easier to maintain

## How Has This Been Tested?

Click around everything on the new page, make sure nothing is broken and everything works as expected.

## Test Impact

N/A for this round, may add it later. A lot of tests are covered by other pages, as long as the URL is jumping correctly, the components should be fine.

## Request review criteria:

If you have UI changes: 


frontend/src/concepts/pipelines/content/tables/pipelineRun/PipelineRunTableRowExperiment.tsx
```diff
@@ -1,31 +1,45 @@
...
   experiment,
   loaded,
+  error,
 }) => {
   const { namespace } = usePipelinesAPI();
+
+  if (error) {
+    return (
+      <Tooltip content={error.message} position="right">
+        <div className="pf-v5-u-disabled-color-100 pf-v5-c-truncate__start">{displayName}</div>
```
Gkrumbach07, line 27: whats with the class names? are they needed

Gkrumbach07 (commented):
this all looks good. just had a comment about a styling thing. Things function in code well too

Gkrumbach07 (commented):
Found that it is allowed to restore a run from the global runs page via both bulk and row restore buttons. this gives an error if its experiment is archived. in experiments we disable the buttons.

same thing happens when restoring a run when experiment is deleted

frontend/src/concepts/pipelines/content/tables/pipelineRun/PipelineRunTableRowExperiment.tsx
DaoDaoNoCode: Hmm, good question. I just copied what it has in `PipelineVersionLink`, I will double-check and see if I can remove it.

pnaik1:
The functionality looks good, working great, some comments here

- I still see the error which Gage mentioned above

Screenshot 2024-11-18 at 4 14 21 PM In experiment the restore is been disabled with the reason provided in tooltip

- Its mentioned in the  [jira-issue](https://issues.redhat.com/browse/RHOAIENG-15515#:~:text=Add%20%22/pipelineRuns/%24%7Bnamespace%7D/runs/%24%7BrunId%7D%22%20to%20make%20sure%20we%20can%20view%20the%20Run%20Details%20page%20on%20this%20URL%2C%20and%20set%20the%20Breadcrumb%20to%20%22Runs%20%2D%20%7Bnamespace%7D%20%3E%20Runs%20%3E%20%7Brun%20name%7D%22), breadcrumb should be of this form, But this is what I see in UI
Screenshot 2024-11-18 at 4 56 07 PM

I see most of the work from this [issue-15518](https://issues.redhat.com/browse/RHOAIENG-15518) is covered here
But there are some changes which I can take into account

- The route should be of this form [route](https://issues.redhat.com/browse/RHOAIENG-15518#:~:text=/pipelineRuns/%24%7Bnamespace%7D/runs,runs%20table%20toolbar) (missing /runs/ here) 
- Navigating to `manage runs page` from experiment list, I can see all the runs being displayed also from different experiments

DaoDaoNoCode:
@pnaik1 Hi! Thanks for the review. I will double-check the first issue you mentioned. As for the breadcrumb, after consideration, I think there is no need to show `Runs` in the breadcrumb item because all the pages are about runs. So I will update the JIRA ticket.
It's really hard not to cover most of the [issue-15518](https://issues.redhat.com/browse/RHOAIENG-15518) because there are related and reused a lot, now we don't have feature flags so I cannot push something up that's not working. In that case, I need to make sure they are linked correctly. I will also check the 2 points you mentioned at the bottom to see if I can fix them in this PR. Thanks!

pnaik1:
@DaoDaoNoCode Thanks for the changes, I see `manage run page` correctly filters the runs when navigated from experiment page.

I still see the error  when experiment is archived and I try to deleted the `archived run` of the `archived experiment` from `global run page`, is this expected?? 
Screenshot 2024-11-18 at 4 14 21 PM
 
and also I am not able to delete the archived run of which experiment is `deleted` from global run page

https://github.com/user-attachments/assets/bdef8702-d54e-4ef0-afe8-031a57292f8d

dpanshug (approved):
/lgtm All the functionality and error mentioned above are working fine.

DaoDaoNoCode:
hi @pnaik1, this is the expected behavior for now. I will create another JIRA ticket because we might need UX support here. We should not block the bulk selection of runs that can be either restorable or not for now. What we can do in only to disable the row action.

Gkrumbach07:
This looks good, bulk restore will be talked about with UX
/lgtm

PR #3478, opened 2024-11-14
Search params pipelines

Gkrumbach07:
https://issues.redhat.com/browse/RHOAIENG-15514

## Description

- Modified usePipelineFilter hook to support URL search parameters
- Updated pipeline run tables and job tables to use the new URL persistence feature (you should be able to flip this on for anyone using the usePipelineFilter, i kept some disabled)
- Changed compare runs parameter from runs to compareRuns in URL to prevent conflicts in search params

## How Has This Been Tested?

- Pipeline Run Table Filters
  - Navigate to pipeline runs table
  - Apply various filters:
    - Name filter
    - Created At filter
    - Status filter
    - Experiment filter
    - Pipeline Version filter
  - Verify the URL updates with the filter parameters
  - Refresh the page and confirm filters persist
- Recurring Run Table (same stuff)
- Compare Runs Feature
  - Select runs to compare
  - Verify URL uses compareRuns parameter instead of runs
   - Check that search filter params and compare selections work correctly
- Edge Cases
  - Test clearing filters and verify URL parameters are removed
  - Test invalid URL parameters (they should not be deleted)

## Test Impact

added tests for hook and updated compare runs tests

## Request review criteria:

If you have UI changes: 


DaoDaoNoCode (commented):
If you put some invalid string like `%` in the search field, it will crash the page. Maybe we should look into either silence the error (probably not a safe way) or prevent them from inputting the invalid characters?
Screenshot 2024-11-15 at 9 31 38 AM

DaoDaoNoCode:
Maybe we can follow what the `Create after` field is doing.
Screenshot 2024-11-15 at 9 40 53 AM

Gkrumbach07:
@DaoDaoNoCode I think i can escape the character. so just convert it to something url safe?

DaoDaoNoCode:
@Gkrumbach07 Yeah, that solution sounds good to me to escape all the special characters in the URL. Just need to make sure to revert them back in the filter chips on UI.

frontend/src/concepts/pipelines/content/tables/usePipelineFilter.ts
```diff
@@ -49,9 +80,28 @@ const usePipelineFilter = (
...
+            pathname: window.location.pathname,
+            search: currentSearchParams.toString(),
+          },
+          { replace: true },
+        );
+      }
+    }, [persistInUrl, navigate]),
   };
 
+  const cleanPredicateValue = (value: string) => encodeURIComponent(value.replace(/\\/g, '\\\\'));
```
DaoDaoNoCode, line 103: This is not needed any more when you do `encodeURIComponent` on https://github.com/opendatahub-io/odh-dashboard/blob/main/frontend/src/api/pipelines/custom.ts#L45 and change it to `encodeURIComponent(JSON.stringify({ predicates: params.filter.predicates }))`

DaoDaoNoCode (commented):
Besides the comment I left above, there are 2 places that still need encoding - [Artifacts table](https://github.com/opendatahub-io/odh-dashboard/blob/main/frontend/src/pages/pipelines/global/experiments/artifacts/ArtifactsTable.tsx#L78) and [Executions table](https://github.com/opendatahub-io/odh-dashboard/blob/main/frontend/src/pages/pipelines/global/experiments/executions/ExecutionsTableToolbar.tsx#L35). These 2 lines need to add `encodeURIComponent` to avoid crashing.

One more question - Can we parse the filter to the display name of an experiment/pipeline version when first loading the page? For examp
le, now you can go to the experiment runs list, and filter it by the pipeline version, and the filter chip is showing correctly with its display name.
Screenshot 2024-11-15 at 1 08 18 PM
However, when you refresh the page, the filter still applies while the chip is showing the pipeline version id as it's shown in the search params, but not parsing it to the display name. Is it possible to find and parse it? If the version does not exist, I think it's fine to show the id (or is there a better way if the user randomly inputs a non-exist version in the search params?)
Screenshot 2024-11-15 at 1 08 48 PM

DaoDaoNoCode (approved):
It lgtm, as long as the tests are fixed it should be good to go. Thanks for the work!
/lgtm

DaoDaoNoCode (approved):
/lgtm

DaoDaoNoCode (approved):
/lgtm

PR #3482, opened 2024-11-15
Fix NIM selection issue

andrewballantyne:
(I'll have to create an issue -- don't think we have it yet)

## Description

Fixes the issue where if KServe is the only thing installed, NIM doesn't appear as an option and it auto-selects KServe.

## How Has This Been Tested?

Not having MM installed then installing the NIM tile & then creating a new project

## Test Impact

Test fixes -- no new tests

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/pages/modelServing/screens/global/EmptyModelServing.tsx
+++ b/frontend/src/pages/modelServing/screens/global/EmptyModelServing.tsx
@@ -2,15 +2,12 @@ import * as React from 'react';
 import { Button } from '@patternfly/react-core';
 import { useNavigate } from 'react-router-dom';
 import { ModelServingContext } from '~/pages/modelServing/ModelServingContext';
-import { getProjectModelServingPlatform } from '~/pages/modelServing/screens/projects/utils';
-import { ServingRuntimePlatform } from '~/types';
 import useServingPlatformStatuses from '~/pages/modelServing/useServingPlatformStatuses';
 import EmptyDetailsView from '~/components/EmptyDetailsView';
 import { ProjectObjectType, typedEmptyImage } from '~/concepts/design/utils';
 import { ProjectSectionID } from '~/pages/projects/screens/detail/types';
 import ServeModelButton from '~/pages/modelServing/screens/global/ServeModelButton';
 import { getDisplayNameFromK8sResource } from '~/concepts/k8s/utils';
-import { isProjectNIMSupported } from '~/pages/modelServing/screens/projects/nimUtils';
 
 const EmptyModelServing: React.FC = () => {
   const navigate = useNavigate();
@@ -19,14 +16,9 @@ const EmptyModelServing: React.FC = () => {
     project,
   } = React.useContext(ModelServingContext);
   const servingPlatformStatuses = useServingPlatformStatuses();
-  const isKServeNIMEnabled = project ? isProjectNIMSupported(project) : false;
 
-  if (
-    (getProjectModelServingPlatform(project, servingPlatformStatuses).platform !==
-      ServingRuntimePlatform.SINGLE ||
-      isKServeNIMEnabled) &&
-    servingRuntimes.length === 0
-  ) {
+  if (servingPlatformStatuses.modelMesh.enabled && servingRuntimes.length === 0) {
+    // Server needed -- must deploy from the project
     return (
       <EmptyDetailsView
         title="No deployed models"
--- a/frontend/src/pages/modelServing/screens/projects/utils.ts
+++ b/frontend/src/pages/modelServing/screens/projects/utils.ts
@@ -295,32 +295,61 @@ export const useCreateInferenceServiceObject = (
 export const getProjectModelServingPlatform = (
   project: ProjectKind | null,
   platformStatuses: ServingPlatformStatuses,
-): { platform?: ServingRuntimePlatform; error?: Error } => {
+): {
+  platform?: ServingRuntimePlatform;
+  hasMultiplePlatformOptions?: boolean;
+  noPlatformsActive?: boolean;
+  error?: Error;
+} => {
   const {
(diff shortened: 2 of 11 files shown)


andrewballantyne:
cc @olavtar

frontend/src/pages/modelServing/screens/global/EmptyModelServing.tsx
```diff
@@ -19,14 +16,9 @@ const EmptyModelServing: React.FC = () => {
...
   const servingPlatformStatuses = useServingPlatformStatuses();
-  const isKServeNIMEnabled = project ? isProjectNIMSupported(project) : false;
 
-  if (
-    (getProjectModelServingPlatform(project, servingPlatformStatuses).platform !==
-      ServingRuntimePlatform.SINGLE ||
-      isKServeNIMEnabled) &&
-    servingRuntimes.length === 0
-  ) {
+  if (servingPlatformStatuses.modelMesh.enabled && servingRuntimes.length === 0) {
```
mturley, line 20: Is removing `|| isKServeNIMEnabled` from this logic desired? Seems like yes but I'm not sure.

frontend/src/pages/modelServing/screens/projects/utils.ts
```diff
@@ -313,13 +319,21 @@ export const getProjectModelServingPlatform = (
...
       return { platform: ServingRuntimePlatform.SINGLE };
     }
-  }
-  if (project.metadata.labels?.[KnownLabels.MODEL_SERVING_PROJECT] === 'true') {
+    if (nimEnabled) {
+      // TODO: this is weird, it relies on KServe today... so it's never "only installed"
+      return { platform: ServingRuntimePlatform.SINGLE };
+    }
+
+    // TODO: unreachable code unless adding a new platform? probably should throw an error
```
mturley, line 327: Is this really unreachable? If the modelmesh label is true (btw, `MODEL_SERVING_PROJECT` is a bad name, the value is `'modelmesh-enabled'`) we'll hit this case right?

mturley (commented):
Mostly LGTM, just a few points of confusion

frontend/src/pages/modelServing/screens/global/EmptyModelServing.tsx
andrewballantyne: Yeah, I think it was a gap -- I could put it back for sure -- but NIM is just KServe so deploying the modal should be the same process -- I'll go double check this flow carefully to make sure. Thanks for calling it out.

frontend/src/pages/modelServing/screens/projects/utils.ts
andrewballantyne: All 3 if-statements directly above this line are for the 3 "auto select states" -- the last if really is stupid, but it fills out this area better from a readability point of view.

The first if-statement jumps out if there are no serving or more than 1 serving... so that leaves us with "one serving that we don't have checks for" -- aka unreachable code unless there is a problem. I didn't want to put a throw error here in the last moments of 2.16 -- but I think it should have one to help devs know when there is a 4th serving that this is an area that will fall through.

andrewballantyne:
Rebased on main to get backend support that recently merged.

frontend/src/pages/modelServing/screens/global/EmptyModelServing.tsx
mturley: I think you're right, this was unnecessarily directing the user to the project which is necessary for model servers, but not for KServe/NIM

frontend/src/pages/modelServing/screens/projects/utils.ts
mturley: This util isn't just for auto-selecting the platform for a new project, it's also for identifying whether a project already has modelmesh or kserve selected. This `else` case is under the `if (project.metadata.labels?.[KnownLabels.MODEL_SERVING_PROJECT] === undefined)` case. The 3 if statements you're referring to are all under that if block, which runs if the project does not have the modelmesh-enabled label (it's a new project in need of auto-select). This else case is for a project that already has modelmesh selected, which correctly returns `ServingRuntimePlatform.MULTI`. The way I read this, it's not unreachable or an error. Am I missing something?

mturley: Ohh --- your comment isn't about the line below it, it's about the empty space in the if block above it - I misunderstood. Maybe one more newline would help 😅

mturley: The indentation should have been clear enough, I was just being dumb. This is fine.

mturley (commented):
Tested a case that failed: On a cluster with KServe and NIM enabled, in Project Details when you select a platform you get the Change button to un-select it on the Models tab but not on the Overview tab. Looking into why

Edit: this is just because the NIM enabled state takes some extra time to load and I wasn't being patient. Disregard.

mturley (approved):
Tested cases (on dashboard shared cluster kserve-pm):
* With modelmesh and kserve enabled, I see both platforms available and selectable/unselectable in the Overview and Models tabs of a project, and the Model Serving page correctly directs me to the project to select a platform / deploy a model.
* With only modelmesh enabled, the Overview and Models tabs of a project both properly autoselect modelmesh, and the Model Serving page correctly directs me to the project to deploy a model.
* With only kserve enabled, the Overview and Models tabs of a project both properly autoselect kserve, and the Model Serving page correctly allows me to deploy a model.

Tested case (on borrowed cluster ai-dev04.kni.syseng.devcluster.openshift.com):
* With kserve and NIM enabled, I see both platforms available and selectable/unselectable in the Overview and Models tabs of a project, and the Model Serving page correctly allows me to deploy a model.

PR #3485, opened 2024-11-15
Fix the NIM Tile issue on prod

andrewballantyne:
Grants the Dashboard Service Account permission over the new NIM resource.


Gkrumbach07 (approved):
/lgtm

changes look good

PR #3488, opened 2024-11-18
Make SimpleSelect one-option hook triggered on key instead of the options

DaoDaoNoCode:
JIRA: https://issues.redhat.com/browse/RHOAIENG-15526

## Description

Update the `SimpleSelect` component, make it not depend on the whole options array (which is not stable). Instead, make it only depend on the first option key. Added a prop to the option named `optionKey` to adapt to some special cases. Here is the reason why we need the `optionKey`:

Normally, `key` in the `SelectOption` is used to differentiate the options and will be triggered an auto-update if there is only one option in the Select component, but there is a special use case in the Model Serving area. When deploying a KServe model, user needs to choose a template, and the options for the `Model format` will be changed by the choice of the template. However, 2 different templates could have the same one option (For example, template A and template B could provide same one and only option `model-format-xyz` in the Model format field). In this case, the one-option `onChange` hook cannot be triggered because it thinks the 2 options are the same, while they are different.
In this PR, I added a new key named `optionKey`, the hook will check that key first, and if it's not provided, then it will check the `key` prop. The `onChange` update will still be on the `key` prop like before. The `optionKey` is only used to trigger the one option hook if the options are actually different.

## How Has This Been Tested?

1. Make sure you can edit the environment variable fields for the notebook controller (This is the related JIRA ticket)
2. Make sure you can deploy a KServe and Model mesh normally. (Recommend test this on KServe shared cluster, try to go to the deploy modal and switch between different templates, make sure nothing is broken)

## Test Impact

N/A

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/components/SimpleSelect.tsx
+++ b/frontend/src/components/SimpleSelect.tsx
@@ -27,6 +27,7 @@ export type SimpleSelectOption = {
   isDisabled?: boolean;
   isFavorited?: boolean;
   dataTestId?: string;
+  optionKey?: string; // Used to differentiate the only option with the same key to trigger the one-option hook in the component
 };
 
 export type SimpleGroupSelectOption = {
@@ -37,7 +38,6 @@ export type SimpleGroupSelectOption = {
 
 type SimpleSelectProps = {
   options?: SimpleSelectOption[];
-  isLoadingOptions?: boolean;
   groupedOptions?: SimpleGroupSelectOption[];
   value?: string;
   toggleLabel?: React.ReactNode;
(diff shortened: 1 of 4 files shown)


frontend/src/components/SimpleSelect.tsx
```diff
@@ -27,6 +27,7 @@ export type SimpleSelectOption = {
   isDisabled?: boolean;
   isFavorited?: boolean;
   dataTestId?: string;
+  optionKey?: string; // Used to differentiate the only option with the same key to trigger the one-option hook in the component
```
christianvogt, line 30: This new key shouldn't be necessary. When there is a single option and there is no selected value then call `onChange` to simulate selecting the option, otherwise do nothing. Currently the selected value is not taken into consideration at all. If a caller requires `onChange` to be called even when supplying the same selected value then there's something wrong with the caller.

mturley (approved):
Tested locally, works as expected in model serving

PR #3491, opened 2024-11-19
'Run every' - 'Select' dropdown greys out and cannot be selected

dpanshug:
[RHOAIENG-14431](https://issues.redhat.com/browse/RHOAIENG-14431)

## Description
Refactored the logic of Run every numeric and text part. Also removed dead code previously used. 

https://github.com/user-attachments/assets/4b7c5aea-0e80-4c1c-9c01-ac17a86612c3

## How Has This Been Tested?
1. Open create schedule pipeline page.
2. Try to enter very long input in Run every numeric field. It should work without breaking.

## Test Impact
Added unit tests for utility function.

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/concepts/pipelines/content/createRun/contentSections/TriggerTypeField.tsx
+++ b/frontend/src/concepts/pipelines/content/createRun/contentSections/TriggerTypeField.tsx
@@ -2,6 +2,7 @@ import * as React from 'react';
 import {
   ClipboardCopy,
   FormGroup,
+  NumberInput,
   Split,
   SplitItem,
   Stack,
@@ -13,12 +14,11 @@ import {
   RunTypeScheduledData,
   ScheduledType,
 } from '~/concepts/pipelines/content/createRun/types';
-import NumberInputWrapper from '~/components/NumberInputWrapper';
-import { replaceNonNumericPartWithString, replaceNumericPartWithString } from '~/utilities/string';
 import {
   DEFAULT_CRON_STRING,
   DEFAULT_PERIODIC_OPTION,
 } from '~/concepts/pipelines/content/createRun/const';
+import { extractNumberAndTimeUnit } from './utils';
 
 type TriggerTypeFieldProps = {
   data: RunTypeScheduledData;
@@ -27,6 +27,8 @@ type TriggerTypeFieldProps = {
 
 const TriggerTypeField: React.FC<TriggerTypeFieldProps> = ({ data, onChange }) => {
   let content: React.ReactNode | null;
+  const [numberPart, unitPart] = extractNumberAndTimeUnit(data.value);
+
   switch (data.triggerType) {
     case ScheduledType.CRON:
       content = (
@@ -50,15 +52,19 @@ const TriggerTypeField: React.FC<TriggerTypeFieldProps> = ({ data, onChange }) =
         <FormGroup label="Run every" data-testid="run-every-group">
           <Split hasGutter>
             <SplitItem>
-              <NumberInputWrapper
+              <NumberInput
                 min={1}
-                value={parseInt(data.value) || 1}
-                onChange={(value) =>
+                value={numberPart}
+                onChange={(newNumberPart) => {
+                  const updatedValue = `${Number(newNumberPart.currentTarget.value).toLocaleString(
+                    'fullwide',
+                    { useGrouping: false },
+                  )}${unitPart}`;
                   onChange({
                     ...data,
-                    value: replaceNumericPartWithString(data.value, value ?? 0),
-                  })
-                }
+                    value: updatedValue,
+                  });
+                }}
               />
             </SplitItem>
             <SplitItem>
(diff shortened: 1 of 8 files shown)


frontend/src/concepts/pipelines/content/createRun/contentSections/TriggerTypeField.tsx
```diff
@@ -52,13 +54,14 @@ const TriggerTypeField: React.FC<TriggerTypeFieldProps> = ({ data, onChange }) =
...
               <NumberInputWrapper
                 min={1}
-                value={parseInt(data.value) || 1}
-                onChange={(value) =>
+                value={numberPart}
+                onChange={(newNumberPart) => {
+                  const updatedValue = `${newNumberPart}${unitPart}`;
                   onChange({
                     ...data,
-                    value: replaceNumericPartWithString(data.value, value ?? 0),
```
Gkrumbach07, line 59: so a few things wrong here, once it is converted to a string, it cannot be edited unless you delete it. also when you submit when the units turn into something bigger it throws an error

Failed to create a recurring run due to conversion error: Invalid input error: Found invalid period schedule interval 0. Set at interval to least 1 second

Gkrumbach07 (changes requested):

dpanshug:
@Gkrumbach07 can you share the steps to reproduce the error, because at my end its working fine

Gkrumbach07:
@dpanshug I basically made a number big enough that it changed to display as scientific format. then try to submit

DaoDaoNoCode (commented):
When I input something big in the value and even if the unit is small, it will not be able to process it.

Screenshot 2024-11-27 at 12 44 16 PM

Gkrumbach07:
Yeah this issue is still happening @dpanshug. Maybe its best to just have hard limit

dpanshug:
@Gkrumbach07 @DaoDaoNoCode The error is because the backend can't take this long value.

I have set the hard limit to 9999999999? Let me know if you think this number should be dynamic w.r.t the unit choosen. Also do we need UX review for this change?

frontend/src/concepts/pipelines/content/createRun/contentSections/TriggerTypeField.tsx
```diff
@@ -13,20 +13,24 @@ import {
...
 } from '~/concepts/pipelines/content/createRun/const';
+import NumberInputWrapper from '~/components/NumberInputWrapper';
+import { extractNumberAndTimeUnit } from './utils';
 
 type TriggerTypeFieldProps = {
   data: RunTypeScheduledData;
   onChange: (scheduledData: RunTypeScheduledData) => void;
 };
 
+const INPUT_MAX_VALUE = 9999999999;
```
DaoDaoNoCode, line 28: Alright, after testing with the current KF UI. We should not hard code the input value here because all the values will be finally transformed into seconds. When you choose a different unit, the input max limit shouldn't be the same. Let's revert this change and let the KF server throw the error.
cc @Gkrumbach07 let me know if you have a better solution because I don't know what the limit here should be. In KF UI when you input some value big (depends on the unit because all the values will be converted into seconds) it will throw the same error as well. We can leave this to KF server to handle. User should not input that big value for recurring runs.

dpanshug:
@Gkrumbach07 @DaoDaoNoCode as per our discussion, i've reverted back the changes. Pls check

frontend/src/concepts/pipelines/content/createRun/contentSections/TriggerTypeField.tsx
```diff
@@ -50,15 +52,16 @@ const TriggerTypeField: React.FC<TriggerTypeFieldProps> = ({ data, onChange }) =
...
-                onChange={(value) =>
+                value={numberPart}
+                onChange={(newNumberPart) => {
+                  const updatedValue = `${newNumberPart.currentTarget.value}${unitPart}`;
                   onChange({
                     ...data,
-                    value: replaceNumericPartWithString(data.value, value ?? 0),
-                  })
-                }
+                    value: updatedValue,
```
Gkrumbach07, line 62: so what you can do is clean the value like this. it will convert form the scientific notation to regular number

```ts
value.toLocaleString('fullwide', {useGrouping:false})
```

Gkrumbach07: you are doing ti on the string. it needs to be on the number

```tsx
onChange({
                    ...data,
                    value:
                      numberPart.toLocaleString('fullwide', { useGrouping: false }) + newUnitPart,
                  });
                }}
```

dpanshug: the example you're mentioning is for the changing the unitPart, i am not understanding why need to clean the scientific value over here? 

If i try to replicate this for the NumberInput field, it's not working.

Can you explain why the clean up is required, and what issue will persist if we continue with the existing solution?

PR #3492, opened 2024-11-19
handle workbench not enabled

Gkrumbach07:
https://issues.redhat.com/browse/RHOAIENG-15252

## Description

hides workbench related things when the area is not enabled

## How Has This Been Tested?

disable workbenches and check that everything from jira is hidden

## Test Impact

added tests to areas project list and project details

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/__tests__/cypress/cypress/tests/mocked/projects/projectDetails.cy.ts
+++ b/frontend/src/__tests__/cypress/cypress/tests/mocked/projects/projectDetails.cy.ts
@@ -58,6 +58,7 @@ type HandlersProps = {
   pipelineServerInitializing?: boolean;
   pipelineServerErrorMessage?: string;
   rejectAddSupportServingPlatformProject?: boolean;
+  disableWorkbenches?: boolean;
 };
 
 const initIntercepts = ({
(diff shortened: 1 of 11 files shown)


DaoDaoNoCode (commented):
I got 2 questions:
1. How are we going to deal with the `Existi
ng connected workbenches` and `Connected workbench` fields in this situation?

Screenshot 2024-11-22 at 2 47 34 PM

2. Should we block fetch to all the notebooks if the workbench flag is disabled?
3. This button on Global project page should be disabled or hidden definitely

Screenshot 2024-11-22 at 2 51 35 PM

4. I can still access the notebook controller via URL (not through the UI), should we disable the route as well?

Gkrumbach07:
1. i say we hide them
2. block all fetch - no, lets just hide where we would try to display it (i know its probably still fetching in the background but i think that is ok)
3. i will disable it
4. url no, we dont do that in the dashboard

dpanshug (changes requested):
~~Can you add the feature flag here https://github.com/opendatahub-io/odh-dashboard/blob/f0b961cca9a09670cf0dcdce1103184a5286b6bd/frontend/src/concepts/areas/const.ts#L93~~

~~Also here so we can disable the feature flag from disableFeatureFlags banner.
https://github.com/opendatahub-io/odh-dashboard/blob/f0b961cca9a09670cf0dcdce1103184a5286b6bd/frontend/src/concepts/areas/const.ts#L4~~

Edit: I see now, we're not adding a feature flag, taking it back.

DaoDaoNoCode (approved):
/lgtm

PR #3495, opened 2024-11-19
add data-testid to connection form fields

christianvogt:
## Description

Adds `field <env var>` `data-testid` to field control elements.
Replaces generated field IDs with `field-<env var>` values to ensure predicable IDs can be used with bitwarden.
Add `data-testid` to sections.

image

## How Has This Been Tested?

Web inspector

## Test Impact

Updated unit tests.

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/__tests__/cypress/cypress/tests/mocked/projects/connections.cy.ts
+++ b/frontend/src/__tests__/cypress/cypress/tests/mocked/projects/connections.cy.ts
@@ -162,7 +162,7 @@ describe('Connections', () => {
     projectDetails.visitSection('test-project', 'connections');
 
     connectionsPage.getConnectionRow('test2').findKebabAction('Edit').click();
-    cy.findByTestId(['field_env']).fill('new data');
+    cy.findByTestId(['field', 'field_env']).fill('new data');
     cy.findByTestId('modal-submit-button').click();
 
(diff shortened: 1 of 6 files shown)


christianvogt:
The original change had a test failure because of a test selector matching too many elements.
Made a small change to qualify the test id with `field` and also removed the test id from the label. This change aligns better with the original test ids in the data connections feature.

christianvogt:
@emilys314 there was a conflict with test ids and one of the cypress tests so I made a new change to use a static test id for the control element which can be chained from the parent selector. Also noticed that one element did not pass through the test id.

emilys314:
yeah that makes sense. 

I don't know if this version would fix Fede's and Jorge's original issue of the input fields themselves not having a unique identifier in the dom though. I don't know if bitwarden is able to do chaining.

Also the original data connections had the `data-testid` on the input themselves and not on the parent form group, is there any reason we have them there instead? (I suppose you can narrow down different elements for more complicated fields)

Screenshot From 2024-11-20 11-02-10

emilys314:
/lgtm 

image

jgarciao:
I've tested today's odh-nightly and now I'm able again to create bitwarden custom fields to populate the values. Thanks a lot, this will save us plenty of time!

PR #3505, opened 2024-11-21
Change token secret to have middle truncation

ashley-o0o:
Closes: [RHOAIENG-11861](https://issues.redhat.com/browse/RHOAIENG-11861)

## Description
Updated token secrets to have middle truncation.
Due to PatternFly not having a true fix for this yet, I’m using our InlineTruncatedClipoardCopy.tsx work around for this fix. A few modifications added to that file to allow for middle truncation. 

## How Has This Been Tested?
Tested locally. 

- Create a model mesh server with a model that requires a token
- View the Token secret in the Models and model servers table 

## Test Impact
No new tests added 

## Screenshots 
Before:
Screenshot 2024-11-21 at 11 32 28 AM

After:
Screenshot 2024-11-21 at 11 32 45 AM

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/components/InlineTruncatedClipboardCopy.tsx
+++ b/frontend/src/components/InlineTruncatedClipboardCopy.tsx
@@ -3,6 +3,7 @@ import * as React from 'react';
 
 type Props = {
   textToCopy: string;
+  truncatePosition?: 'start' | 'middle' | 'end';
   testId?: string;
   maxWidth?: number;
 };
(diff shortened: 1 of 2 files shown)


christianvogt:
@ashley-o0o i think the tooltip is likely ok. Before the change I could scroll the input to see the full value. With the truncation and no tooltip there's no way to see the value without copying it.

I suggest you keep the tooltip and have UX see it. At the very least having the tooltip remains consistent with all other truncation scenarios.

jenny-s51 (approved):
Workaround looks good @ashley-o0o ! FWIW, PF Topology has a [`truncateMiddle` util](https://github.com/patternfly/react-topology/blob/b88e9eb7ad15aa52bb33a92ea295633f144c446a/packages/module/src/utils/truncate-middle.ts#L19) which would have been awesome to use here, but it's not available for consumers to import outside of Topology. 

Left a comment / suggested solution for this in https://github.com/patternfly/patternfly-react/issues/10892#issuecomment-2494822963 - looks like the PF team has prioritized this issue for the Q1 milestone 🎉

PR #3506, opened 2024-11-21
Fix search param issues on pipeline runs table

DaoDaoNoCode:
JIRA: [RHOAIENG-16026](https://issues.redhat.com/browse/RHOAIENG-16026)

## Description

Fix issues as described in the JIRA.
Added a hook named `usePipelineFilterSearchParams` to use the search params as the only source of truth for the pipeline filters on the pipeline runs table.
1. When loading with search params in the URL, it will automatically find the valid filters and apply it
2. When switching between the tabs, it will unmount and mount the tabs in case the search params change on one tab affects another
3. The tab will remember it's last search params when switching between tabs, so users won't lose track of the filters when they change tabs in the pipeline runs table. If you navigate to another page and click back to the page, all the filters will be reset. (However, if you use the `Back` button on the browser, the filters will be kept because the search params are in the history)

## How Has This Been Tested?

1. Go to the global pipeline runs page
2. Set some filters, see how the URL changes with the search params
3. Go to other tabs on the same page, set some different filters
4. Switch between tabs and see if all the filters are kept for different tabs
5. Refresh the page with the search params, see if the search filters are auto set according to the search params

## Test Impact

Added some tests to verify the new hook.

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/__tests__/cypress/cypress/pages/pipelines/pipelineRunTable.ts
+++ b/frontend/src/__tests__/cypress/cypress/pages/pipelines/pipelineRunTable.ts
@@ -49,13 +49,8 @@ class PipelineRunsTable {
     return cy.findByTestId(this.testId);
   }
 
-  shouldRowNotBeVisible(name: string) {
-    this.find()
-      .parents()
-      .find('tr')
-      .find(`[data-label=Name]`)
-      .contains(name)
-      .should('not.be.visible');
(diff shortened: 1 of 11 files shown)


DaoDaoNoCode:
Great thanks to @christianvogt for the solution to record the history of each tab 👏

PR #3507, opened 2024-11-21
feat: added route configurations

olavtar:
https://issues.redhat.com/browse/NVPE-28

## Description

Added route configurations to the NIM model deployments similar to KServe.
image

## How Has This Been Tested?

Tested to have access through the external route. 

## Test Impact

Tested to have access through the external route through creating a workbench.

## Request review criteria:

If you have UI changes: 


andrewballantyne (changes requested):
image

Do you see this on your table in the model tab? (it's also on the overview card) 

I see this after submitting the modal. No sense in having route logic if we can't use it on the flip side.

frontend/src/__tests__/cypress/cypress/tests/mocked/projects/modelServingNim.cy.ts
```diff
@@ -346,4 +346,47 @@ describe('NIM Model Serving', () => {
...
+      nimDeployModal.findExternalRouteError().should('not.exist');
+
+      // Re-enable authentication
+      nimDeployModal.findModelRouteCheckbox().check();
+      nimDeployModal.findAuthenticationCheckbox().check();
+      nimDeployModal.findExternalRouteError().should('not.exist');
+
+      // should display the default service account name when "Model route" is checked
+      nimDeployModal.findModelRouteCheckbox().check();
+      nimDeployModal.findServiceAccountNameInput().should('have.value', 'default-name');
```
andrewballantyne, line 389: Seems we have a gap here -- I went ahead and logged a tech debt here. https://issues.redhat.com/browse/RHOAIENG-16190

We should definitely test more of this area -- but I don't think you need to do this as we didn't do it in KServe modal either 😞

andrewballantyne:
> image
> Do you see this on your table in the model tab? (it's also on the overview card)
> 
> I see this after submitting the modal. No sense in having route logic if we can't use it on the flip side.

I actually think this is more because the model is not deploying -- I checked a few other configurations -- and I'm having similar issues but other things are working. Let me go to the channel and see if we can get a proper deployment of a model

frontend/src/__tests__/cypress/cypress/tests/mocked/projects/modelServingNim.cy.ts
olavtar: After successfully deploying I see this:
image

andrewballantyne (approved):

PR #3512, opened 2024-11-22
[RHOAIENG-1154]: Apply modal to alert user error

jenny-s51:
Towards https://issues.redhat.com/browse/RHOAIENG-1154.

## Description

Before:
Screenshot 2024-12-19 at 5 43 39 PM

After:
Screenshot 2024-12-18 at 5 13 52 PM

## How Has This Been Tested?

Run `application.cy.ts` to simulate 403 status code and verify the modal is rendered. 

## Test Impact

Added new test to `application.cy.ts`. 

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/__tests__/cypress/cypress/pages/loginDialog.ts
+++ b/frontend/src/__tests__/cypress/cypress/pages/loginDialog.ts
@@ -1,0 +1,23 @@
+import { Modal } from '~/__tests__/cypress/cypress/pages/components/Modal';
+
+class LoginDialog extends Modal {
+  constructor() {
+    super(/Session Expired/);
+  }
+
+  // FIXME Remove once PF version is upgraded to 6.1.0.
+  // https://issues.redhat.com/browse/RHOAIENG-11946
+  // https://github.com/patternfly/patternfly-react/issues/11041
+  shouldBeOpen(open = true): void {
+    if (open) {
+      this.find();
+    } else {
+      this.find().should('not.exist');
+    }
+  }
+
+  findLoginButton() {
+    return this.findFooter().findByTestId('modal-login-button');
+  }
+}
+export const loginDialog = new LoginDialog();
--- a/frontend/src/__tests__/cypress/cypress/tests/mocked/application.cy.ts
+++ b/frontend/src/__tests__/cypress/cypress/tests/mocked/application.cy.ts
@@ -7,6 +7,7 @@ import {
 import { mockDashboardConfig } from '~/__mocks__';
 import { aboutDialog } from '~/__tests__/cypress/cypress/pages/aboutDialog';
 import { mockConsoleLinks } from '~/__mocks__/mockConsoleLinks';
+import { loginDialog } from '~/__tests__/cypress/cypress/pages/loginDialog';
 
 describe('Application', () => {
   it('should disallow access to the dashboard', () => {
@@ -112,5 +113,29 @@ describe('Application', () => {
 
     aboutDialog.findText().should('contain.text', 'OpenShift');
     aboutDialog.findProductName().should('contain.text', 'OpenShift AI');
+  });
+  it('should show the login modal when receiving a 403 status code', () => {
+    // Mock the intercept to return a 403 status code
+    cy.interceptOdh('GET /api/config', {
+      statusCode: 403,
+    }).as('getData403');
+
+    // Set up the sign-out intercept before visiting the page
+    cy.interceptOdh('GET /oauth/sign_out').as('signOut');
+
+    // Visit the page where the request is triggered
+    cy.visit('/');
+
+    // Wait for the intercept to be triggered
+    cy.wait('@getData403');
+
+    // Verify that the login modal is displayed
+    loginDialog.shouldBeOpen();
+
+    // Simulate clicking the Log in button
+    loginDialog.findLoginButton().click();
+
+    // Wait for the sign out intercept to be triggered
+    cy.wait('@signOut');
   });
 });
--- a/frontend/src/app/App.tsx
+++ b/frontend/src/app/App.tsx
@@ -34,6 +34,7 @@ import TelemetrySetup from './TelemetrySetup';
 import { logout } from './appUtils';
 import QuickStarts from './QuickStarts';
 import DevFeatureFlagsBanner from './DevFeatureFlagsBanner';
+import SessionExpiredModal from './SessionExpiredModal';
 
 import './App.scss';
 
@@ -68,9 +69,16 @@ const App: React.FC = () => {
     [buildStatuses, dashboardConfig, storageClasses],
   );
 
+  const isUnauthorized = fetchConfigError?.request?.status === 403;
+
   // We lack the critical data to startup the app
   if (userError || fetchConfigError) {
-    // There was an error fetching critical data
+    // Check for unauthorized state
+    if (isUnauthorized) {
+      return <SessionExpiredModal />;
+    }
+
+    // Default error handling for other cases
     return (
       <Page>
         <PageSection hasBodyWrapper={false}>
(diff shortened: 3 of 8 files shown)


emilys314:
Would there be a way to detect specifically that the user's session timed out to show the new modal? I feel like it may be slightly misleading because it looks like the code is a catch all, but the message is specific for session expiration? Unless that was the idea, it also doesn't seem that straightfoward

christianvogt:
I agree, we should look to identify the exact error response that comes from an expired session to show this modal otherwise we show the old error or update it with some other text if it's not related to expired session.

xianli123:
Thanks @jenny-s51. The implementation looks good to me. 
As for the description in the modal, it's based on the context of [RHOAIENG-1154](https://issues.redhat.com/browse/RHOAIENG-1154). I agree with @emilys314 @christianvogt. It would be better to identify the exact errors, and then we can update the description in the modal. If you have any concrete idea, please let me know and I can provide new design if needed.

jenny-s51 (commented):
Thank you for your feedback here @emilys314 @christianvogt @xianli123 - after investigating with @christianvogt we identified that a 403 status code indicates an expired session. 

I've updated the implementation to render the modal when that status code is received, and added a corresponding Cypress test.

frontend/src/__tests__/cypress/cypress/pages/loginDialog.ts
```diff
@@ -0,0 +1,9 @@
+import Chainable = Cypress.Chainable;
+
+export class LoginDialog {
+  findText(): Chainable<JQuery<HTMLElement>> {
+    return cy.findByTestId('timeout-text');
+  }
+}
```
christianvogt, line 7: For modals we normally create a contextual page object which extends the `Modal` class:
```suggestion
class LoginDialog extends Modal {
  constructor() {
    super('Session Expired');
  }
}
```

With a new instance of this class you can then do `loginDialog().shouldBeOpen()` instead of the modal text assertion.

frontend/src/app/App.tsx
```diff
@@ -67,9 +72,37 @@ const App: React.FC = () => {
...
+          <ModalFooter>
+            <Button
+              key="confirm"
+              variant="primary"
+              onClick={() => logout().then(() => window.location.reload())}
+            >
+              Log in
+            </Button>
+          </ModalFooter>
+        </Modal>
```
christianvogt, line 101: Can you move this Modal to a separate file for better separation.

frontend/src/__tests__/cypress/cypress/tests/mocked/application.cy.ts
```diff
@@ -113,4 +114,22 @@ describe('Application', () => {
     aboutDialog.findText().should('contain.text', 'OpenShift');
     aboutDialog.findProductName().should('contain.text', 'OpenShift AI');
   });
+  it('should show the login modal when receiving a 403 status code', () => {
```
christianvogt, line 117: This test can also intercept the call to `/oauth/sign_out` when the user clicks the `Log in` button.

frontend/src/services/dashboardConfigService.ts
```diff
@@ -7,6 +8,15 @@ export const fetchDashboardConfig = (): Promise<DashboardConfigKind> => {
...
+      const message = e.response.data?.message;
+
+      // Throw the AxiosError with status code
+      throw new AxiosError(
+        message, // Error message from the server
+        message, // The error message also serves as the "code" argument for AxiosError
+        undefined, // Optional: request config that was used
+        e.response, // Optional: the full response object
+        e,
+      );
```
christianvogt, line 20: is `e` not already an axios error?

jenny-s51: Good point, thanks! Updated.

frontend/src/__tests__/cypress/cypress/pages/loginDialog.ts
jenny-s51: Applied this change with `loginDialog.shouldBeOpen()` - the test isn't passing though because the `find()` call in `shouldBeOpen` doesn't identify this modal correctly. Investigating...

frontend/src/app/App.tsx
jenny-s51: Yes, thanks, created a separate component for this.

frontend/src/__tests__/cypress/cypress/pages/loginDialog.ts
```diff
@@ -0,0 +1,20 @@
...
+  find() {
+    return cy.findByTestId(this.testId);
+  }
+
+  findLoginButton() {
+    return this.findFooter().findByTestId('modal-login-button');
+  }
+}
+
+export const loginDialog = new LoginDialog('Session Expired', 'session-expired-modal');
```
christianvogt, line 20: No need to parameterize the constructor since you're not going to instantiate different login dialogs:

```suggestion
class LoginDialog extends Modal {

  constructor() {
    super('Session Expired');
  }

  find() {
    return cy.findByTestId('session-expired-modal');
  }

  findLoginButton() {
    return this.findFooter().findByTestId('modal-login-button');
  }
}

export const loginDialog = new LoginDialog();
```

frontend/src/__tests__/cypress/cypress/tests/mocked/application.cy.ts
```diff
@@ -113,4 +114,28 @@ describe('Application', () => {
...
+    // Mock the intercept to return a 403 status code
+    cy.interceptOdh('GET /api/config', {
+      statusCode: 403,
+    }).as('getData403');
+
+    // Set up the sign-out intercept before visiting the page
+    cy.intercept('GET', '/oauth/sign_out').as('signOut');
```
christianvogt, line 124: It would be good to add a typed interface function to the `interceptOdh` method. We tend to do this for all our well known URLs.

frontend/src/app/App.tsx
```diff
@@ -68,9 +69,21 @@ const App: React.FC = () => {
     [buildStatuses, dashboardConfig, storageClasses],
   );
 
+  const isUnauthorized = React.useMemo(() => {
+    if (fetchConfigError?.request?.status === 403) {
+      return true;
+    }
+    return false;
+  }, [fetchConfigError]);
```
christianvogt, line 77: No need to memoize this because the compute is negligible and the result is a stable boolean value.
```suggestion
  const isUnauthorized = fetchConfigError?.request?.status === 403;
```

frontend/src/app/SessionExpiredModal.tsx
```diff
@@ -0,0 +1,42 @@
...
+  Button,
+} from '@patternfly/react-core';
+import { logout } from './appUtils';
+
+const SessionExpiredModal: React.FC = () => (
+  <Modal
+    data-testid="session-expired-modal"
+    variant={ModalVariant.small}
+    isOpen
+    ouiaId="BasicModal"
```
christianvogt, line 18: Remove the `ouiaId` since it's value is generic and not referenced by any test.

frontend/src/app/SessionExpiredModal.tsx
```diff
@@ -0,0 +1,42 @@
...
+    aria-labelledby="session-expired-modal-title"
+  >
+    <ModalHeader
+      title="Session Expired"
+      titleIconVariant="warning"
+      labelId="session-expired-modal-title"
+    />
+    <ModalBody data-testid="timeout-text">
```
christianvogt, line 26: `data-testid` here doesn't have much value on a static block of text.

frontend/src/services/dashboardConfigService.ts
```diff
@@ -7,6 +7,7 @@ export const fetchDashboardConfig = (): Promise<DashboardConfigKind> => {
     .get(url)
     .then((response) => response.data)
     .catch((e) => {
-      throw new Error(e.response.data.message);
+      // Throw the AxiosError with status code
+      throw e;
     });
```
christianvogt, line 12: You can omit the catch and rethrow since this is the default behavior and there's no manipulation going on any longer with the error.

PR #3513, opened 2024-11-22
Add model redeploying status

emilys314:
Closes https://issues.redhat.com/browse/RHOAIENG-3409

## Description

Makes the status a spinner after editing a model, popover message says "Redeploying".

Basically the status functions will prioritize looking at the target state, since that is more indicative of what is going on. 
I'm not 100% on what the correct message is to display. But the special return case could be removed and then it would default to the state of the target i.e. "Loading". 

image

## How Has This Been Tested?

1. Deploy a model
2. Edit the model and change something (most things will reload it)
3. The status should change to loading (the polling may be slow to update it)

## Test Impact

Jest unit tests could be added to the utils functions, but there aren't any so i didn't add any.

## Request review criteria:

If you have UI changes: 

@vconzola 

### DIFF
--- a/frontend/src/pages/modelServing/screens/global/utils.ts
+++ b/frontend/src/pages/modelServing/screens/global/utils.ts
@@ -14,7 +14,7 @@ export const getInferenceServiceStatusMessage = (is: InferenceServiceKind): stri
   const activeModelState = is.status?.modelStatus?.states?.activeModelState;
   const targetModelState = is.status?.modelStatus?.states?.targetModelState;
 
-  const stateMessage = targetModelState || activeModelState || 'Unknown';
+  const stateMessage = (targetModelState || activeModelState) ?? 'Unknown';
 
   if (
     activeModelState === InferenceServiceModelState.FAILED_TO_LOAD ||
--- a/frontend/src/pages/modelServing/screens/global/utils.ts
+++ b/frontend/src/pages/modelServing/screens/global/utils.ts
@@ -3,24 +3,36 @@ import { getDisplayNameFromK8sResource } from '~/concepts/k8s/utils';
 import { InferenceServiceModelState, ModelStatus } from '~/pages/modelServing/screens/types';
 import { asEnumMember } from '~/utilities/utils';
 
-export const getInferenceServiceActiveModelState = (
+export const getInferenceServiceModelState = (
   is: InferenceServiceKind,
 ): InferenceServiceModelState =>
-  asEnumMember(is.status?.modelStatus?.states?.activeModelState, InferenceServiceModelState) ||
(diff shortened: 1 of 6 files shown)


vconzola:
@kaedward Would you please take a quick peek at this? I think it looks OK, but you might have a different suggestion for the popover message.

frontend/src/pages/modelServing/screens/global/utils.ts
```diff
@@ -3,24 +3,36 @@ import { getDisplayNameFromK8sResource } from '~/concepts/k8s/utils';
...
+  }
 
-  const lastFailureMessage = is.status?.modelStatus?.lastFailureInfo?.message;
-  const stateMessage = activeModelState ?? targetModelState ?? 'Unknown';
+  if (
+    activeModelState === InferenceServiceModelState.LOADED &&
+    (targetModelState === InferenceServiceModelState.LOADING ||
+      targetModelState === InferenceServiceModelState.PENDING)
+  ) {
+    return 'Redeploying';
```
christianvogt, line 32: @emilys314 did you get a response yet on suggestion of the new text? If not, we can always do a followup to update the text if everything else is good to go with the PR.

Should add unit tests for this util function.

christianvogt (commented):
The jira issue mentions this:
> When activeModelState is in a pending state we will show the spinner without a tooltip.

Seems to me the fallback message right now is always going to be `Unknown` ?

frontend/src/pages/modelServing/screens/global/utils.ts
emilys314: No updates on the wording, i believe Katie is out until next week

Sure I'll add tests

emilys314:
> The jira issue mentions this:
> 
> > When activeModelState is in a pending state we will show the spinner without a tooltip.
> 
> Seems to me the fallback message right now is always going to be `Unknown` ?

Yeah so the active vs target states don't seem to be well documented anywhere, but here are some tests from kserve 
- https://github.com/search?q=repo%3Akserve%2Fkserve+ModelState&type=code
- https://github.com/kserve/kserve/blob/master/pkg/apis/serving/v1beta1/inference_service_status_test.go#L405

I don't think  `activeModelState` can ever just be pending.

During initial deployment it would have `activeModelState: '', targetModelState: Loading`

The popover text will be hidden if the function returns "", it can do that now because it looks at activeModelState which becomes that commonly. But targetModelState is more up to date so this PR would make it happen less.

However there does seem merit to return an emptry string sometimes so I can make a small update by changing 
`const stateMessage = targetModelState || activeModelState || 'Unknown';`
to
`const stateMessage = (targetModelState || activeModelState) ?? 'Unknown';`

alexcreasy (approved):
Not tested, but the code looks good - thanks Emily!

PR #3516, opened 2024-11-25
Enable Cypress E2E tests login to Dashboard URL with devFeatureFlags

manosnoam:
https://issues.redhat.com/browse/RHOAIENG-16164

## Description
- Pass query parameters (e.g devFeatureFlags) if the relative url is '/'.
- Login should override the user/password values (instead of append).

## How Has This Been Tested?
All Cypress E2E tests were run using Jenkins on a real cluster (dashboard-e2e):
image

## Test Impact
This PR enables us to run E2E on ODH Dashboard URL that includes `devFeatureFlags` in the query parameters.
To use, update your ODH_DASHBOARD_URL with the desired devFeatureFlags, in the Cypress variables file (`frontend/src/__tests__/cypress/test-variables.yml`), for example:
`ODH_DASHBOARD_URL: https://odh-dashboard-opendatahub.apps.your-server.domain.com?devFeatureFlags=true`

## Request review criteria:

- [ x] The developer has manually tested the changes and verified that the changes work
- [ x] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [ x] The developer has added tests or explained why testing cannot be added (unit or cypress tests for related changes)

- [x ] The developer has tested their solution on a cluster by using the image produced by the PR to `main`

### DIFF
--- a/frontend/src/__tests__/cypress/cypress/support/commands/application.ts
+++ b/frontend/src/__tests__/cypress/cypress/support/commands/application.ts
@@ -12,13 +12,17 @@ declare global {
   namespace Cypress {
     interface Chainable {
       /**
-       * Visits the URL and performs a login if necessary.
-       * Uses credentials supplied by environment variables if not provided.
+       * Visits relative URL (page) and performs a login if necessary.
+       * If relativeUrl is '/' or empty string, uses base URL with query parameters (i.e. DevFeatureFlags).
+       * If User credentials not provided, uses Admin credentials supplied by environment variables.
        *
-       * @param url the URL to visit
+       * @param relativeUrl the page to visit.
        * @param credentials login credentials
        */
-      visitWithLogin: (url: string, user?: UserAuthConfig) => Cypress.Chainable<void>;
+      visitWithLogin: (
+        relativeUrl: string,
+        credentials?: UserAuthConfig,
+      ) => Cypress.Chainable<void>;
 
       /**
        * Find a patternfly kebab toggle button.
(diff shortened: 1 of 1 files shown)


frontend/src/__tests__/cypress/cypress/support/commands/application.ts
```diff
@@ -148,31 +152,34 @@ declare global {
...
   } else {
-    cy.intercept('GET', url, { log: false }).as('visitWithLogin');
-
-    cy.visit(url, { failOnStatusCode: false });
-
-    cy.wait('@visitWithLogin', { log: false }).then((interception) => {
-      if (interception.response?.statusCode === 403) {
-        cy.log('Do login');
-        // do login
+    let fullUrl: string;
```
antowaddle, line 159: Can we also add some logging here of the full constructued URL?

Something like 
cy.log(`Constructed full URL: ${fullUrl}`);

frontend/src/__tests__/cypress/cypress/support/commands/application.ts
```diff
@@ -148,31 +152,34 @@ declare global {
...
+      fullUrl = new URL(relativeUrl, Cypress.config('baseUrl') || '').href;
+    } else {
+      fullUrl = new URL(Cypress.config('baseUrl') || '').href;
+    }
+    cy.step(`Go to: ${fullUrl}`);
+    cy.intercept('GET', fullUrl, { log: false }).as('page');
+    cy.visit(fullUrl, { failOnStatusCode: false });
+    cy.wait('@page', { log: false }).then((interception) => {
+      const statusCode = interception.response?.statusCode;
+      if (statusCode === 403) {
```
antowaddle, line 170: Could we also print the error code if there is a general error?
 
 cy.log(`Unexpected status code ${statusCode}. Throwing error.`);
throw new Error(`Failed to visit '${fullUrl}'. Status code: ${statusCode || 'unknown'}`);

frontend/src/__tests__/cypress/cypress/support/commands/application.ts
manosnoam: It's already there on line 165:
`cy.step(`Go to: ${fullUrl}`);`

frontend/src/__tests__/cypress/cypress/support/commands/application.ts
manosnoam: Why to print both log and error messages about the same status code ?
The error message already mentions it, so I don't think log is required.

frontend/src/__tests__/cypress/cypress/support/commands/application.ts
antowaddle: Okay, that should be fine then.

frontend/src/__tests__/cypress/cypress/support/commands/application.ts
antowaddle: Those were examples, I think either or is fine really. The key point I was highlighting was that if this fails we need to know why, what the error code was and what the constructed URL is/was, otherwise it will be difficult to debug.

manosnoam: Yes, good point. In line 182 it prints it via error message.

frontend/src/__tests__/cypress/cypress/tests/e2e/storageClasses/storageClasses.cy.ts
```diff
@@ -76,7 +76,7 @@ describe('An admin user can manage Storage Classes from Settings -> Storage clas
   });
 
   it('An admin user can set an enabled Storage Class as the default one', () => {
-    cy.visitWithLogin('/', HTPASSWD_CLUSTER_ADMIN_USER);
+    cy.visitWithLogin('', HTPASSWD_CLUSTER_ADMIN_USER);
```
christianvogt, line 79: @manosnoam I think using `/` as a relative URL or `/foo/` should be valid. Whether we need to strip trailing slashes should be something left to the underlying function.

frontend/src/__tests__/cypress/cypress/support/commands/application.ts
```diff
@@ -148,31 +152,34 @@ declare global {
...
+        cy.findAllByRole('link', credentials.AUTH_TYPE ? { name: credentials.AUTH_TYPE } : {})
           .last()
-          .click();
-        cy.get('input[name=username]').type(user.USERNAME);
-        cy.get('input[name=password]').type(user.PASSWORD);
+          .then(($link) => {
+            cy.wrap($link).click();
+          });
+        cy.get('input[name=username]').invoke('val', '').type(credentials.USERNAME);
+        cy.get('input[name=password]').invoke('val', '').type(credentials.PASSWORD);
```
christianvogt, line 179: Can we use `fill` instead?
```suggestion
        cy.get('input[name=username]').fill(credentials.USERNAME);
        cy.get('input[name=password]').fill(credentials.PASSWORD);
```

frontend/src/__tests__/cypress/cypress/tests/e2e/storageClasses/storageClasses.cy.ts
manosnoam: sure, I changed tests back to '/', and now passing the query parameters - only if defined in ODH_DASHOBAR_URL and if the relativeUrl is just '/' or empty.

FedeAlonso (approved):

PR #3523, opened 2024-11-26
Cypress e2e Test - Verify RHODS Explore Section Contains Only Expected ISVs

antowaddle:
## Request review criteria:

If you have UI changes: 


frontend/src/__tests__/cypress/cypress/tests/e2e/applications/explore/testExploreDefaultISVs.cy.ts
```diff
@@ -0,0 +1,38 @@
...
+
+const applicationNamespace = Cypress.env('TEST_NAMESPACE');
+
+describe('Verify RHODS Explore Section Contains Only Expected ISVs', () => {
+  let expectedISVs: string[];
+
+  before(() => {
+    // Setup: Retrieve the names of OdhApplication resources in the specified namespace.
+    getOcResourceNames(applicationNamespace, 'OdhApplication').then((metadataNames) => {
+      // Filter out the 'rhoai' card from the expected ISVs which displays in RHOAI
```
manosnoam, line 13: Why to exclude RHOAI ISVs, and not verify that the number of resources in the backend is found in UI as well ?

antowaddle: The 'RHOAI' ISV doesn't display on the UI but does in the backend call. I had to add this exclusion in order for the test to be successful on a RHOAI cluster. 

All of the resources are verified, see line 28. Each of the resources are verified based on the backend call, I specifically match them with the metadata.name.

antowaddle: @manosnoam It's perhaps a bug, the issue only seems to occur in RHOAI (not ODH). The current implementation just excludes any 'RHOAI' isv's as in a PSI Cluster it appears in the payload but is not present on the UI.

The only other viable workaround would be to use a predefined JSON file (like the previous Robot test), but this JSON file would need regular updates which is not the case with this test. 
image

manosnoam: Thanks @antowaddle for opening the bug that may cause it: 
https://issues.redhat.com/browse/RHOAIENG-16226

manosnoam: /lgtm

FedeAlonso:
/lgtm
I've tested it using the latest version of the Dashboard in RHOAI, and got the expected bug.

FedeAlonso (approved):

PR #3529, opened 2024-11-28
Update cluster storage modal

pnaik1:
https://issues.redhat.com/browse/RHOAIENG-1098

## Description
Refer the mock [the design](https://www.figma.com/proto/372he4ynUVkovm3lwfFVbg/Storage-classes?page-id=1755%3A49861&node-id=1755-49862&node-type=frame&viewport=353%2C373%2C0.03&t=DArVL64y9lG7wVhh-1&scaling=scale-down-width&content-scaling=fixed&starting-point-node-id=1755%3A49862)

https://github.com/user-attachments/assets/9ea3acfa-225d-43aa-9e18-e6b6fb982024

Screenshot 2024-11-28 at 5 00 19 PM

Screenshot 2024-11-28 at 5 00 10 PM
Screenshot 2024-11-28 at 4 59 56 PM

Screenshot 2024-11-28 at 5 03 31 PM

## How Has This Been Tested?
Refer mocks to cover all the checks
1. Create a cluster storage and attach the existing workbench
2. Check for default mount path
3. Verify the validation path ( We are now allowing numbers)
4. create the cluster storage and check if the mountpath is correctly updated to workbench 
5. Edit  the existing cluster and check and change the workbench
6. 

## Test Impact
Added cypress test

## Request review criteria:

If you have UI changes: 


pnaik1:
cc @xianli123 , I have updated the description with required screenshot

when custom path is empty, Currently I have displayed this error, let me know your thoughts around here

Screenshot 2024-11-28 at 5 13 08 PM

frontend/src/pages/projects/screens/detail/storage/ClusterStorageTableRow.tsx
```diff
@@ -0,0 +1,268 @@
...
+              if (typeof selectedValue === 'string') {
+                addNotebookData(selectedValue);
+              }
+            }}
+          />
+        </Td>
+        <Td dataLabel="Path format">
+          <SimpleSelect
+            isFullWidth
+            popperProps={{ direction: 'down' }}
```
DaoDaoNoCode, line 167: You can also add `maxWidth: undefined` here to override the default `maxWidth: 'trigger'` in the SimpleSelect so you can see the options menu show the correct width. Otherwise, it will be limited the width same as the toggle width.

pnaik1: @DaoDaoNoCode thanks, I have updated it to `maxWidth: undefined`  but the width is limited to some extend i have a added `width:30vw`

frontend/src/pages/projects/screens/detail/storage/utils.ts
```diff
@@ -24,3 +27,87 @@ export const isPvcUpdateRequired = (
...
+
+  // Regex to allow empty string for Standard format
+  const regex =
+    format === MountPathFormat.STANDARD
+      ? /^(\/?[a-z0-9-]+(\/[a-z0-9-]+)*\/?|)$/
+      : /^(\/?[a-z0-9-]+(\/[a-z0-9-]+)*\/?)?$/;
+
+  if (!regex.test(value)) {
+    return 'Must consist of lowercase letters, numbers and hyphens (-). Use slashes (/) to indicate the subdirectories';
+  }
```
dpanshug, line 101: Numbers are not allowed in the mount path, check the mount path field in Spawner page.
image

frontend/src/pages/projects/screens/detail/storage/clusterTableColumns.ts
```diff
@@ -0,0 +1,33 @@
...
+    sortable: false,
+    width: 30,
+  },
+  {
+    label: 'Mount path',
+    field: 'mountPath',
+    sortable: false,
+    info: {
+      popover:
+        'The directory within a container where a volume is mounted and accessible. Must consist of lowercase letters, numbers and hyphens (-). Use slashes (/) to indicate subdirectories',
```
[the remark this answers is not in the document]
pnaik1: @dpanshug , there was a conversation around this in the slack channel we are now allowing numbers as well

frontend/src/pages/projects/screens/detail/storage/utils.ts
pnaik1: @dpanshug , there was a conversation around this in the slack channel we are now allowing numbers as well

DaoDaoNoCode:
There is an edge case where the newly added mount path will be reset to default. Here is the step to reproduce:
1. Edit storage A which already has a workbench connection
2. Add workbench connection and edit the mount path
3. Click minus icon to remove the existing connection
4. See the newly added connection mount path is set to default, user will lose the edited value in this case

frontend/src/pages/projects/screens/detail/storage/utils.ts
xianli123: @pnaik1 @dpanshug Here is [the thread link](https://redhat-internal.slack.com/archives/C07CQ4R4HMY/p1732630861994889) around this point. Users will be able to use the number in the path.

xianli123:
Thanks @pnaik1. 
I want to confirm how to trigger the popover as you shared (I attached it below). Does the 'View connection paths' trigger the popover? 
A small enhancement: The bottom border of the row is missing. It will LGTM, once the border is added.

Screenshot 2024-12-02 at 19 28 01

pnaik1:
@xianli123 Thanks for the review, the popover will trigger when clicked on `View connected paths` , I have added the border in the latest changes

https://github.com/user-attachments/assets/1d905418-68c3-4a9e-9f66-28d08cf871b5

xianli123:
@pnaik1 Thanks for doing it! LGTM!

Gkrumbach07:
wide screens have dropdown show over the modal
image

Gkrumbach07:
there should be a slash in front of the custom ones

image

Gkrumbach07:
root standard storage path doesn't get a checkmark
image

frontend/src/pages/projects/screens/detail/storage/ClusterStorageModal.tsx
```diff
@@ -1,58 +1,117 @@
...
+              : '',
+            error: '',
+          },
+          existingPvc: true,
+          isUpdatedValue: false,
+        })),
+      );
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [existingPvc, hasExistingNotebookConnections]);
```
dpanshug, line 70: This can be refactored to

```
React.useEffect(() => {
  if (hasExistingNotebookConnections && existingPvc) {
    const mappedData = connectedNotebooks.map((connectedNotebook) => ({
      name: connectedNotebook.metadata.name,
      mountPath: {
        value: getNotebookPVCMountPathMap(connectedNotebook)[existingPvc.metadata.name] || '',
        error: '',
      },
      existingPvc: true,
      isUpdatedValue: false,
    }));
    setNotebookData(mappedData);
  }
}, [existingPvc, hasExistingNotebookConnections, connectedNotebooks]);
```

frontend/src/pages/projects/screens/detail/storage/ClusterStorageModal.tsx
```diff
@@ -130,28 +200,82 @@ const ClusterStorageModal: React.FC<ClusterStorageModalProps> = ({ existingPvc,
...
+
+                          setNotebookData(copiedArray);
+                        }}
+                        onDelete={() => {
+                          const copiedArray = [...notebookData];
+                          copiedArray.splice(rowIndex, 1);
+                          setNotebookData(copiedArray);
+                        }}
+                      />
+                    )}
```
dpanshug, line 259: Can we extract these to functions `addNotebookData` and `deleteNotebookRow` ? we can add useCallback hook to it.

frontend/src/pages/projects/screens/detail/storage/ClusterStorageTableRow.tsx
```diff
@@ -0,0 +1,292 @@
...
+
+  let placeholderText: string;
+
+  if (!availableNotebooks.loaded) {
+    placeholderText = 'Loading workbenches';
+  } else if (selectOptions.length === 0) {
+    placeholderText = 'No existing workbench available';
+  } else {
+    placeholderText = 'Select a workbench';
+  }
```
dpanshug, line 148: This can be refactored to useMemo

```
const placeholderText = React.useMemo(() => {
 -----
}, [availableNotebooks.loaded, selectOptions.length]);
```

dpanshug (changes requested):

frontend/src/pages/projects/screens/detail/storage/ClusterStorageTableRow.tsx
pnaik1: @dpanshug , I’ve seen this pattern used in several places throughout our codebase. However, I’m open to refactoring it to useMemo.

Gkrumbach07: this should not be use memo, it is not really optimizing anything. useMemo should be reserved for when we have things that are at least O(n), but honestly JS does a good job with that and would could get away without memoizing that.

pnaik1:
Updated the PR accordingly

Gkrumbach07:
This all looks good. just need to still fix the modal overflow issue: https://github.com/opendatahub-io/odh-dashboard/pull/3529#issuecomment-2515714000

DaoDaoNoCode:
@pnaik1 Can you rebase the PR please? Want to see how it looks like on PF6

frontend/src/pages/projects/screens/detail/storage/clusterTableColumns.ts
```diff
@@ -0,0 +1,33 @@
...
+    label: 'Name',
+    field: 'name',
+    sortable: false,
+    width: 35,
+  },
+  {
+    label: 'Path format',
+    field: 'pathFormat',
+    sortable: false,
+    width: 25,
```
DaoDaoNoCode, line 21: There is a small change in the row width. when selecting between standard path format and custom. I guess it's because of the effect of the hidden id column.
Update widths could solve this problem
```suggestion
    width: 20,
```

frontend/src/pages/projects/screens/detail/storage/clusterTableColumns.ts
```diff
@@ -0,0 +1,33 @@
...
+  },
+  {
+    label: 'Mount path',
+    field: 'mountPath',
+    sortable: false,
+    info: {
+      popover:
+        'The directory within a container where a volume is mounted and accessible. Must consist of lowercase letters, numbers and hyphens (-). Use slashes (/) to indicate subdirectories.',
+    },
+    width: 40,
```
DaoDaoNoCode, line 31:
```suggestion
    width: 35,
```

DaoDaoNoCode (commented):
Left some comments based on my testing. I am not a fan of the way to save `id` in the table column and hide it just in order to auto-generate the name, there should be a better way to save and use this information. 🤔 WDYT? @Gkrumbach07

frontend/src/pages/projects/screens/detail/storage/ClusterStorageModal.tsx
```diff
@@ -130,28 +227,73 @@ const ClusterStorageModal: React.FC<ClusterStorageModalProps> = ({ existingPvc,
...
+                            ),
+                          ],
+                          loaded,
+                        }}
+                        addNotebookData={(
+                          notebookName?: string,
+                          mountPath?: MountPath,
+                          isUpdatedValue?: boolean,
+                          updatedExistingPvc?: boolean,
+                        ) =>
```
Gkrumbach07, line 262: why re write the types here

frontend/src/pages/projects/screens/detail/storage/ClusterStorageTableRow.tsx
```diff
@@ -0,0 +1,289 @@
...
+        clusterStorageName
+          ? {
+              value: defaultPathValue,
+              error: validateClusterMountPath(defaultPathValue, inUseMountPaths),
+            }
+          : undefined,
+      );
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [clusterStorageName, inUseMountPaths, obj.existingPvc, obj.isUpdatedValue, obj.newRowId]);
```
Gkrumbach07, line 85: why do this in the row. can you just do it in the component above where you are adding the row

Gkrumbach07 (changes requested):

Gkrumbach07:
> Left some comments based on my testing. I am not a fan of the way to save id in the table column and hide it just in order to auto-generate the name, there should be a better way to save and use this information. 🤔 WDYT? @Gkrumbach07

we should be autogenerating the name when we create it and not in the row

frontend/src/pages/projects/screens/detail/storage/BaseStorageModal.tsx
```diff
@@ -119,7 +119,7 @@ const BaseStorageModal: React.FC<BaseStorageModalProps> = ({
               editableK8sName={!existingPvc}
             />
           </StackItem>
-          {children}
+          {typeof children === 'function' ? children(createData.name) : children}
```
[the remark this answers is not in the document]
pnaik1: @Gkrumbach07 `getDisplayNameFromK8sResource` works perfectly fine when we have `existing cluster storage` but for creating new Cluster storage, this will return `undefined`

frontend/src/pages/projects/screens/detail/storage/ClusterStorageTableRow.tsx
pnaik1: @Gkrumbach07 since I am getting the `clusterStorageName` from [BaseStorageModal](https://github.com/opendatahub-io/odh-dashboard/pull/3529/files/18ab33f4ccab30a4b5ef5618f2df2a32b1cbc160#diff-4f31835a5a1bd8e7d3f245e9a812610197f3e1e9cca51f44e9560a227130057d) and the name changes upon typing , I prefered to keep it in the row  name  as this cannot be treated as default `default` to add while creating row

pnaik1:
@DaoDaoNoCode @Gkrumbach07 Thanks I have updated the pr accordingly
I have a question around this,
Since the rows are dynamic and are editable, there is a required for unique stable column, removing this it aligns the row based on `name`  which disturbs the flow
I see same think done here: https://github.com/opendatahub-io/odh-dashboard/blob/main/frontend/src/pages/projects/screens/spawner/storage/constants.ts#L6-L11

> > Left some comments based on my testing. I am not a fan of the way to save id in the table column and hide it just in order to auto-generate the name, there should be a better way to save and use this information. 🤔 WDYT? @Gkrumbach07
> 
> we should be autogenerating the name when we create it and not in the row

frontend/src/pages/projects/screens/detail/storage/ClusterStorageTableRow.tsx
```diff
@@ -0,0 +1,285 @@
...
+  if (!availableNotebooks.loaded) {
+    placeholderText = 'Loading workbenches';
+  } else if (selectOptions.length === 0) {
+    placeholderText = 'No existing workbench available';
+  } else {
+    placeholderText = 'Select a workbench';
+  }
+
+  return (
+    <Tr key={obj.newRowId}>
```
DaoDaoNoCode, line 146: The `style={{ verticalAlign: 'center' }}` is still not added. Can you also make sure it's added at https://github.com/opendatahub-io/odh-dashboard/blob/main/frontend/src/pages/projects/screens/spawner/storage/ClusterStorageTable.tsx#L88?

pnaik1: @DaoDaoNoCode , I tried adding this . It effects the layout when mount path error is displayed 
Screenshot 2024-12-10 at 9 38 20 PM

pnaik1:
@DaoDaoNoCode , I have updated the `SpawnPage` cluster storage table

Gkrumbach07:
There seems to be a rendering issue. you have some infinite render loop when you add a new row to the table

frontend/src/pages/projects/screens/detail/storage/ClusterStorageModal.tsx
```diff
@@ -1,161 +1,359 @@
...
+                          name: notebookName,
+                          mountPath: {
+                            ...row.mountPath,
+                            error: validateClusterMountPath(
+                              row.mountPath.value,
+                              getInUseMountPaths(
+                                notebookName,
+                                allNotebooks,
+                                existingPvc?.metadata.name,
+                              ),
```
pnaik1, line 318: @Gkrumbach07 I made this changes from your fix to a handle edge cases

Gkrumbach07: yeah np, i didnt look too much into the specifics when i posted it

PR #3530, opened 2024-11-28
Fix dryRun issue with permissionless user

lucferbux:
## Description

Closes https://issues.redhat.com/browse/RHOAIENG-15479

Fix an issue in which users without cluster admin roles would dry run rolebindings referencing a role that's not there.
Also, avoiding calling setUpTokenAuth twice in kserve.

## How Has This Been Tested?

1. Impersonate a regular user
2. Create a new project
3. Go to model serving section
4. Deploy a new kserve model
5. Select auth tokens
6. See there's an issue

These changes will fix that scenario

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/pages/modelServing/utils.ts
+++ b/frontend/src/pages/modelServing/utils.ts
@@ -121,8 +121,7 @@ export const setUpTokenAuth = async (
       ? Promise.all([
           createServiceAccountIfMissing(serviceAccount, namespace, opts),
           ...(role ? [createRoleIfMissing(role, namespace, opts)] : []),
-          createRoleBindingIfMissing(roleBinding, namespace, opts),
-        ])
+        ]).then(() => createRoleBindingIfMissing(roleBinding, namespace, opts))
       : Promise.resolve()
   )
     .then(() => createSecrets(fillData, deployedModelName, namespace, existingSecrets, opts))
@@ -160,7 +159,14 @@ export const createRoleBindingIfMissing = async (
 ): Promise<RoleBindingKind> =>
   getRoleBinding(namespace, rolebinding.metadata.name).catch((e) => {
     if (e.statusObject?.code === 404) {
-      return createRoleBinding(rolebinding, opts);
+      return createRoleBinding(rolebinding, opts).catch((error) => {
+        if (error.statusObject?.code === 404 && opts?.dryRun) {
+          // If dryRun is enabled and the user is not Cluster Admin it seems that there's a k8s error
+          // that raises a 404 trying to find the role, which is missing since it's a dryRun.
+          return Promise.resolve(rolebinding);
+        }
+        return Promise.reject(error);
+      });
     }
     return Promise.reject(e);
   });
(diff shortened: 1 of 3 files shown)


frontend/src/pages/modelServing/utils.ts
```diff
@@ -160,7 +160,15 @@ export const createRoleBindingIfMissing = async (
 ): Promise<RoleBindingKind> =>
   getRoleBinding(namespace, rolebinding.metadata.name).catch((e) => {
     if (e.statusObject?.code === 404) {
-      return createRoleBinding(rolebinding, opts);
+      return createRoleBinding(rolebinding, opts).catch((error) => {
+        if (error.statusObject?.code === 404 && opts?.dryRun) {
+          // If dryRun is enabled and the user is not Cluster Admin it seems that there's a k8s error
+          // that raises a 404 trying to find the role, which is missing from the dryRun
```
andrewballantyne, line 166: Missing from the Dry Run -- because we are dry running everything and the RoleBinding validates the existence of the Role on the cluster?

andrewballantyne (changes requested):
Not sure I understand the setup token auth stuff you changed... I'll need to dig into that a bit more. But I have a direct inquiry to the comment you wrote down about this solution. Let me know if I am on the right track.

frontend/src/pages/modelServing/utils.ts
lucferbux: Yup, that's right, I assume this is a k8s error, cause it only happens with permisionless users, when dryRunning, creating the rolebinding i think internally tries to get the role it's trying to assign, since it's not yet created it fails. My guess is that with cluster admins k8s does the correct check and it assumes it's dryRunned too.

lucferbux: It's funny, cause if you change the rolebinding to not use an specific role it works just fine, or if you run this with cluster admins, it's just with this specific scenario.

mturley: This is definitely strange. It doesn't make sense that the `createRoleBinding` util would return a 404 since it's a direct call to `k8sCreateResource` with the `RoleBindingModel`, the client never does a GET for the role - you mean that is happening internally in the k8s API? Seems bizarre that it would respond with 404 then and not 400.

It also doesn't make sense that this ever worked for admins tbh - I don't see why the API would be aware of the previous dryRun requests, they're all separate HTTP requests. So for some reason that internal check for the role existing isn't happening for admins? I wonder if admin permissions mean overriding checks like that or something? Confusing all around...

This solution seems fine to me based on the behavior we're seeing, I would love to understand it more but I suppose that doesn't block this PR.

emilys314:
i tested the PR image on my cluster with a non admin user and the issue was fixed. the extra isModelMesh is slightly confusing but i suppose it makes sense to avoid unnecessary calls

mturley (approved):
Pair reviewed and tested on a call with @lucferbux - LGTM

PR #3536, opened 2024-12-03
Cypress: Retry mechanism for e2e tests

FedeAlonso:
https://issues.redhat.com/browse/RHOAIENG-16485 

## Description
Add a retry  mechanism to the e2e Cypress tests.

## How Has This Been Tested?
I've forced the failure of the application.cy.ts test here 9a1c6762723ce4564cdf81746b2589a35dfe43af 
Then I've ran the cypress job (cypress/job/cypress-tests/59)
In the report, 3 attempts are shown:
image

Then I've fixed the test here  8e55ff23222efa539c2d43f8df9b2eed57b091d6 
And ran the cypress job (cypress/job/cypress-tests/60)
And the test was up and running fine again
image

## Test Impact
It's an "enhancement" 
 
## Request review criteria:

### DIFF
--- a/frontend/src/__tests__/cypress/cypress.config.ts
+++ b/frontend/src/__tests__/cypress/cypress.config.ts
@@ -138,7 +138,7 @@ export default defineConfig({
       // Apply retries only for tests in the "e2e" folder
       return {
         ...config,
-        retries: config.specPattern.includes('/e2e/') ? 2 : config.retries,
+        retries: !env.CY_MOCK && !env.CY_RECORD ? 2 : config.retries,
       };
     },
   },
--- a/frontend/src/__tests__/cypress/cypress.config.ts
+++ b/frontend/src/__tests__/cypress/cypress.config.ts
@@ -135,12 +135,11 @@ export default defineConfig({
         await mergeFiles(outputFile, inputFiles);
       });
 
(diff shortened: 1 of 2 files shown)


frontend/src/__tests__/cypress/cypress.config.ts
```diff
@@ -135,7 +135,11 @@ export default defineConfig({
         await mergeFiles(outputFile, inputFiles);
       });
 
-      return config;
+      // Apply retries only for tests in the "e2e" folder
+      return {
+        ...config,
+        retries: config.specPattern.includes('/e2e/') ? 2 : config.retries,
```
christianvogt, line 141: Our e2e is run when `const e2e = !env.CY_MOCK && !env.CY_RECORD`. Either create a new `const` or inline as follows:
```suggestion
        retries: !env.CY_MOCK && !env.CY_RECORD ? 2 : config.retries,
```

Do we want to include retries in open mode? If not, you can configure open and run mode separately:
See https://docs.cypress.io/app/guides/test-retries#Global-Configuration
```suggestion
        retries: !env.CY_MOCK && !env.CY_RECORD ? { runMode: 2, openMode: 0 } : config.retries,
```

PR #3538, opened 2024-12-03
Fetch correct cluster admin list for the notebook controller admin panel

DaoDaoNoCode:
JIRA: https://issues.redhat.com/browse/RHOAIENG-16434

## Description

Use `ResourceAccessReview` API as described [here](https://docs.openshift.com/container-platform/4.17/rest_api/authorization_apis/resourceaccessreview-authorization-openshift-io-v1.html) to fetch the cluster admin users and groups instead of arbitrarily push `cluster-admins` group to the list and fetch it. This could also avoid the error mentioned in the JIRA where we are fetching a non-exist user group.

## How Has This Been Tested?

1. Create some user groups
2. Bind them to the `cluster-admin` role
3. Add some users to those groups separately
4. Go to the Jupyter notebook controller admin panel, make sure those users are there and shown as `Admin`

## Test Impact

N/A

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/backend/src/routes/api/status/adminAllowedUsers.ts
+++ b/backend/src/routes/api/status/adminAllowedUsers.ts
@@ -4,6 +4,7 @@ import { getUserInfo } from '../../../utils/userUtils';
 import {
   getAdminUserList,
   getAllowedUserList,
+  getClusterAdminUserList,
   isUserAdmin,
   KUBE_SAFE_PREFIX,
 } from '../../../utils/adminUtils';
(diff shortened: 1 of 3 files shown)


Gkrumbach07:
I am not sure if this is a concern but

before a admin user would see this from `/status/{namespace}/allowedusers`. I am logged in as adminuser1 that does not have cluster admin privileges but has odh admin privileges
```json
[
    {
        "username": "adminuser1",
        "privilege": "Admin",
        "lastActivity": null
    },

]
```

but now they can see every user that has cluster admin role on the cluster. for example clusteradminuser1 is a cluster admin but is not part of the groups for cluster-admin or odh-admin
```json
[
    {
        "username": "clusteradminuser1",
        "privilege": "Admin",
        "lastActivity": null
    },
    {
        "username": "adminuser1",
        "privilege": "Admin",
        "lastActivity": null
    }
]
```

@andrewballantyne I remember you saying this is something we want to avoid, or maybe I am misremembering it.

DaoDaoNoCode:
IMO any users or groups that are bound with the `cluster-admin` role should have the same access. If we allowed users in the `cluster-admins` group to be viewed as Admin here, then the users who bind to the same role should have the same behavior.

andrewballantyne:
@Gkrumbach07 @DaoDaoNoCode it's interesting -- without a notebook, it's hard to say who is and who isn't going to be using the Dashboard's Jupyter Tile.

The whole feature is kinda problematic... I think we want to make sure anyone who has a notebook is reflected clearly as either an admin or a user -- after that it's a bit uncharted. Cluster-admins can be IT or SRE and not at all users of the app. It seems that would be undesirable.

I feel since the Jira didn't specifically say I want to see cluster-admins, and is just more concerned with admins being called users... is there anyway we can not show all cluster admins, and just the cluster admins that have had a notebook (aka last activity)?

DaoDaoNoCode:
@andrewballantyne I believe so. I will double check and update the PR.

DaoDaoNoCode:
@Gkrumbach07 @andrewballantyne Hi, I updated the PR (see my second commit), now it only shows the cluster admins that  have a Notebook resource in the notebook namespace (which means they ever try to use the KFNBC) in the admin panel and make sure the role column is `Admin` instead of `User`. If a cluster admin never touches the KFNBC, it won't be seen in the list. Please check it again and let me know if there is anything I need to update, thanks!

Gkrumbach07 (approved):
/lgtm

allowed users is not showing other cluster admins now
```json
[
    {
        "username": "adminuser1",
        "privilege": "Admin",
        "lastActivity": null
    },
    {
        "username": "kubeadmin",
        "privilege": "Admin",
        "lastActivity": null
    }
]
```

Issue #3521, opened 2024-11-26
[Bug]: RHOAIENG-16192 deleting a workbench also deletes envFrom configmaps and secrets not created by dashboard and not unique by name AND workbench


shalberd:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Deploy type

OpenDataHub core version (eg. `v1.6.0`)

### Version

main, v2.29.1-odh

### Current Behavior

Deleting a notebook / workbench currently has logic in it to also delete associated configmaps and secrets listed in the notebook container's envFrom section.
That deletion occurs by name and assumes that each secret and configmap is only associated to that one workbench / notebook and has a unique name in the namespace per workbench, either configmap or secret having been generated by odh dashboard.

This logic is a problem when we have configmaps and secrets as envFrom basis that have not been generated by dashboard.
Because those configmaps and secrets are often times created by CICD processes, not by dashboard, and linked up currently to the container envFrom section envFrom list via other processes (cronjob patch, ArgoCD).

In preparing for later being able to select existing secrets and configmaps from a [dropdown list in the GUI env section as well](https://github.com/opendatahub-io/odh-dashboard/issues/1380), I tagged the configmap and secret with label
`opendatahub.io/dashboard: 'true'` but **not** with label `opendatahub.io/managed: 'true'`.
In essence, dashboard-related: yes    .... managed by odh or dashboard: no

Preparation steps are described in "Steps to Reproduce".
After those steps, we can check this test config in the dashboard GUI as well. The refs show up in the form of the two new env variables in both workbenches jupyter and jupyter2:
![20241126 Workbench Edit GUI shows new env from CICD ConfigMap](https://github.com/user-attachments/assets/e37362c4-5a9c-4658-b991-8cd0eefb3fa2)
![20241126 Workbench Edit GUI shows new env from CICD Secret](https://github.com/user-attachments/assets/b745deea-7be6-4344-99d7-5a32310cd3c9)

**If you now delete one of the two workbenches**, it does not really matter, via Dashboard GUI .... the unique name per namespace; not workbench, and not managed by ODH, secret and configmap will no longer be there.

Which in the other remaining workbench will lead to problems, obviously, because it references the configmap and secret ...

### Expected Behavior

probable root cause:

configmaps and secrets to delete are deleted by name only as they appear in envFrom list, when workbench is deleted:

[...]

PR #3539, opened 2024-12-03
fix(frontend): prevent unintended deletion of shared ConfigMaps and Secrets during notebook deletion

Gkrumbach07:
https://issues.redhat.com/browse/RHOAIENG-16192

fixes https://github.com/opendatahub-io/odh-dashboard/issues/3521 (we will follow up with more flexibility in the future)

## Description

This PR addresses an issue where deleting a notebook would unintentionally delete ConfigMaps and Secrets that might be shared between multiple workbenches. The fix introduces prefix-based filtering to ensure only dashboard-generated resources are deleted.

- Modified deletion logic to only delete resources that start with these prefixes
- Updated tests to verify that custom resources are preserved

## Test Impact

updated delete test

1. create test resources
```yaml
# Create a shared ConfigMap
kind: ConfigMap
apiVersion: v1
metadata:
  name: shared-test-configmap
  labels:
data:
  STAGE: test

# Create a shared Secret
kind: Secret
apiVersion: v1
metadata:
  name: shared-test-secret
  labels:
data:
  USER_PW: bXlzYW1wbGVwYXNzd29yZDM0
type: Opaque
```

2. Create Notebook with secret and env var
3. Link manually created resources to notebook (shared-test-secret, shared-test-configmap)
```yaml
   envFrom:
            # Dashboard-generated (will be deleted with notebook)
            - secretRef:
                name: secret-abc123
            - configMapRef:
                name: configmap-xyz789
            
            # Shared/Custom resources (will NOT be deleted)
            - secretRef:
                name: shared-test-secret
            - configMapRef:
                name: shared-test-configmap
```
5. Test Deletion
   - Delete notebook-1 through the dashboard
   - Verify that:
      - notebook-1 is deleted
      - The shared ConfigMap and Secret remain
  
**Expected Results**
- Only resources prefixed with secret- or configmap- should be deleted
- Shared resources without these prefixes should remain intact
- Second notebook should continue to function with shared resources

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/api/k8s/configMaps.ts
+++ b/frontend/src/api/k8s/configMaps.ts
@@ -14,7 +14,7 @@ export const CONFIGMAP_PREFIX = 'configmap-';
 
 export const getGeneratedConfigMapName = (): string => `${CONFIGMAP_PREFIX}${genRandomChars()}`;
 export const isGeneratedConfigMapName = (name: string): boolean =>
-  Boolean(name.match(new RegExp(`^${CONFIGMAP_PREFIX}[a-z0-9]{6}$`)));
+  new RegExp(`^${CONFIGMAP_PREFIX}[a-z0-9]{6}$`).test(name);
 
 export const assembleConfigMap = (
   projectName: string,
--- a/frontend/src/api/k8s/configMaps.ts
+++ b/frontend/src/api/k8s/configMaps.ts
@@ -12,6 +12,10 @@ import { applyK8sAPIOptions } from '~/api/apiMergeUtils';
 
 export const CONFIGMAP_PREFIX = 'configmap-';
 
+export const getGeneratedConfigMapName = (): string => `${CONFIGMAP_PREFIX}${genRandomChars()}`;
+export const isGeneratedConfigMapName = (name: string): boolean =>
+  Boolean(name.match(new RegExp(`^${CONFIGMAP_PREFIX}[a-z0-9]{6}$`)));
+
 export const assembleConfigMap = (
   projectName: string,
   configMapData: Record<string, string>,
@@ -20,7 +24,7 @@ export const assembleConfigMap = (
   apiVersion: 'v1',
   kind: 'ConfigMap',
   metadata: {
-    name: configMapName || `${CONFIGMAP_PREFIX}${genRandomChars()}`,
+    name: configMapName || getGeneratedConfigMapName(),
     namespace: projectName,
     labels: {
       [KnownLabels.DASHBOARD_RESOURCE]: 'true',
--- a/frontend/src/api/k8s/configMaps.ts
+++ b/frontend/src/api/k8s/configMaps.ts
@@ -10,6 +10,8 @@ import { ConfigMapModel } from '~/api/models';
 import { genRandomChars } from '~/utilities/string';
 import { applyK8sAPIOptions } from '~/api/apiMergeUtils';
 
+export const CONFIGMAP_PREFIX = 'configmap-';
+
 export const assembleConfigMap = (
   projectName: string,
   configMapData: Record<string, string>,
@@ -18,7 +20,7 @@ export const assembleConfigMap = (
   apiVersion: 'v1',
   kind: 'ConfigMap',
   metadata: {
-    name: configMapName || `configmap-${genRandomChars()}`,
+    name: configMapName || `${CONFIGMAP_PREFIX}${genRandomChars()}`,
     namespace: projectName,
     labels: {
       [KnownLabels.DASHBOARD_RESOURCE]: 'true',
--- a/frontend/src/api/k8s/secrets.ts
+++ b/frontend/src/api/k8s/secrets.ts
@@ -13,6 +13,7 @@ import { translateDisplayNameForK8s } from '~/concepts/k8s/utils';
 import { applyK8sAPIOptions } from '~/api/apiMergeUtils';
 
 export const DATA_CONNECTION_PREFIX = 'aws-connection';
+export const SECRET_PREFIX = 'secret-';
 
 export const assembleSecret = (
   projectName: string,
@@ -26,7 +27,7 @@ export const assembleSecret = (
   const annotations: Record<string, string> = {};
 
   let stringData = data;
-  let name = `secret-${genRandomChars()}`;
+  let name = `${SECRET_PREFIX}${genRandomChars()}`;
 
   if (type === 'aws') {
     const { Name, ...secretBody } = data;
--- a/frontend/src/api/k8s/secrets.ts
+++ b/frontend/src/api/k8s/secrets.ts
@@ -17,7 +17,7 @@ export const SECRET_PREFIX = 'secret-';
 
 export const getGeneratedSecretName = (): string => `${SECRET_PREFIX}${genRandomChars()}`;
 export const isGeneratedSecretName = (name: string): boolean =>
-  Boolean(name.match(new RegExp(`^${SECRET_PREFIX}[a-z0-9]{6}$`)));
+  new RegExp(`^${SECRET_PREFIX}[a-z0-9]{6}$`).test(name);
 
 export const assembleSecret = (
   projectName: string,
--- a/frontend/src/api/k8s/secrets.ts
+++ b/frontend/src/api/k8s/secrets.ts
@@ -31,7 +31,7 @@ export const assembleSecret = (
   const annotations: Record<string, string> = {};
 
   let stringData = data;
-  let name = secretName || getGeneratedSecretName();
+  let name = getGeneratedSecretName();
 
   if (type === 'aws') {
     const { Name, ...secretBody } = data;
@@ -46,7 +46,7 @@ export const assembleSecret = (
     apiVersion: 'v1',
     kind: 'Secret',
     metadata: {
-      name,
+      name: secretName || name,
       namespace: projectName,
       annotations,
       labels,
--- a/frontend/src/api/k8s/secrets.ts
+++ b/frontend/src/api/k8s/secrets.ts
@@ -46,7 +46,7 @@ export const assembleSecret = (
     apiVersion: 'v1',
     kind: 'Secret',
     metadata: {
-      name: secretName || name,
+      name,
       namespace: projectName,
       annotations,
       labels,
--- a/frontend/src/api/k8s/secrets.ts
+++ b/frontend/src/api/k8s/secrets.ts
@@ -15,6 +15,10 @@ import { applyK8sAPIOptions } from '~/api/apiMergeUtils';
 export const DATA_CONNECTION_PREFIX = 'aws-connection';
 export const SECRET_PREFIX = 'secret-';
 
+export const getGeneratedSecretName = (): string => `${SECRET_PREFIX}${genRandomChars()}`;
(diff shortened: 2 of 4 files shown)


frontend/src/api/k8s/secrets.ts
```diff
@@ -26,7 +27,7 @@ export const assembleSecret = (
   const annotations: Record<string, string> = {};
 
   let stringData = data;
```
shalberd, line 29: in assembleSecret and assembleConfigmap, there is the metadata label constant named DASHBOARD_RESOURCE being set to true. But I think the label `opendatahub.io/managed: true` is set, too, I just do not know by which component (notebook controller possibly, maybe the operator itself ...) Do you see label `opendatahub.io/managed: true` in created env-related configmaps and secrets?

The types SecretKind and ConfigmapKind have a provision for labels, in this case, `opendatahub.io/dashboard: true` is set here in code i.e. that konwn labels contst DASHBOARD_RESOURCE.

The other const representing  `opendatahub.io/managed` has nothing do do with AWS related secrets, if you look around, you see all kinds of deployment objects in the main odh namespace and all sorts of other objects having that label.

I would argue against using a secret-prefix, as we cannot control users outside odh also using configmap-xxx or secret--- format.

frontend/src/pages/projects/notebook/DeleteNotebookModal.tsx
```diff
@@ -36,10 +43,16 @@ const DeleteNotebookModal: React.FC<DeleteNotebookModalProps> = ({ notebook, onC
           (envFrom) => !envFrom.secretRef?.name.includes(DATA_CONNECTION_PREFIX),
         );
         const configMapNames = nonDataConnectionVariables
-          .filter((envName): envName is ConfigMapRef => !!envName.configMapRef)
+          .filter(
+            (envName): envName is ConfigMapRef =>
+              !!envName.configMapRef && envName.configMapRef.name.startsWith(CONFIGMAP_PREFIX),
```
shalberd, line 48: I would not filter by a newly-created configmap or secret name prefix.
Instead, I would suggest by the odh-universal label kex `opendatahub.io/managed-by` which I think you call
https://github.com/opendatahub-io/odh-dashboard/blob/main/frontend/src/const.ts#L5 DATA_CONNECTION_AWS, but it is in fact just indicating that the secret or configmap are managed by opendatahub.

That label would be the ideal one to go for in filtering.

frontend/src/api/k8s/secrets.ts
Gkrumbach07: Yep this was the original solution. We went with this prefix solution to limit the number of api calls in the dashboard bc in the current context we were not already fetching the secret/config map. 

@andrewballantyne can chime in here becuase he was the one opting for this prefix solution over the label solution

Gkrumbach07: also to your side note...
> Small side note: currently, in the edit workbench section, shared configmaps and secrets also appear if they do not have the opendatahub.io/dashboard label. I think currently, all logic is exclusively by name and entry in the envFrom list, never by anything else like labels.

Yes that is true, we just try to fetch whatever is in envfrom, we do put that label on those secrets/configmaps on edit. do we use it, dont think so. probably just a artifact from an earlier version of the UI

frontend/src/pages/projects/notebook/DeleteNotebookModal.tsx
Gkrumbach07: > My main argument against prefix-based filtering remains: how can we assume non-dashboard, CICD type external processes do not create secrets that have configmap or secret not as a postfix, but as a prefix, then, they would still be deleted, would they not?

yep and that is the biggest issue here. 

for api call volume we would be fetching for every resource in env from which may be small yes, but could be huge.

frontend/src/api/k8s/secrets.ts
andrewballantyne: We are addressing the loss of configmaps/secrets not created by the Dashboard first by the same way we do it for DSPA secret.

This is not the final solution for addressing this, but we need to understand and plan our action to how we approach it. Dashboard is no longer in a place to address a system of resources and how one crafts or deletes them on the cluster like we were back a year or two ago. We'll need architectural design documents and walk through a general plan for all resources, not adhoc for how Workbenches handles it.

I'd like us to align on the current implementation we use for DSPA to avoid deleting secrets not generated. It's fair to say "configmap-" and "secret-" prefixes followed by a number is not ideal... but it's the easiest way to support backwards compatibility with what we have as we determine the way to move forward.

Given that in the system of injecting extra configmaps and secrets not through the UI is already on a custom adventure, it can just name them something different; and I'd argue people probably don't organically call their configmaps "configmap-<fixed length generated characters>" (I think we use 6 alphanumerical chars)

To address "but why not just make it more robust" -- we intend to. Just not with this initial effort. See my comment on the ticket, we will come back and address this more robustly in the future. But I'd prefer to align all of Dashboard on the same implementation then "we do it this way here, and that way there, and a different way when this applies". 

Hope that helps with what and why and where we plan to go after this.

frontend/src/pages/projects/notebook/DeleteNotebookModal.tsx
```diff
@@ -36,10 +43,16 @@ const DeleteNotebookModal: React.FC<DeleteNotebookModalProps> = ({ notebook, onC
...
+          )
           .map((data) => data.configMapRef.name);
         const secretNames = nonDataConnectionVariables
-          .filter((envName): envName is SecretRef => !!envName.secretRef)
+          .filter(
+            (envName): envName is SecretRef =>
+              !!envName.secretRef && envName.secretRef.name.startsWith(SECRET_PREFIX),
```
andrewballantyne, line 54: Lets make sure we also filter for the fixed length of characters (which defaults to 6 today)

See [DSPA secret logic](https://github.com/opendatahub-io/odh-dashboard/blob/main/frontend/src/concepts/pipelines/utils.ts#L31-L32)

andrewballantyne (changes requested):

frontend/src/pages/projects/notebook/DeleteNotebookModal.tsx
Gkrumbach07: sure can do that too

frontend/src/api/k8s/secrets.ts
shalberd: > to say "configmap-" and "secret-" prefixes followed by a number is not ideal... but it's the easiest way to support backwards compatibility with what we have as we determine the way to move forward.

Given that in the system of injecting extra configmaps and secrets not through the UI is already on a custom adventure, it can just name them something different; and I'd argue people probably don't organically call their configmaps "configmap-"

Yes, I can live with that for now :-) Thank you all for your feedback and effort. The way this is handled for now, I can just make sure my custom configmaps and secrets do not start with configmap- or secret- and I'll be ok.

frontend/src/pages/projects/notebook/DeleteNotebookModal.tsx
shalberd: It could be two-pronged possibly: slow-changing, non-ODH created secrets and configmaps and their info like labels only fetched on first start of dashboard ... and dashboard secrets only fetched by name live, in terms of call volume ...
But then again, that might not be optimal either. Plus, then the logic in dashboard would be more complex, too. I see what you mean by call volume to the API of openshift ...

frontend/src/pages/projects/notebook/DeleteNotebookModal.tsx
shalberd: yeah, good idea. Anyways, I am fine with this intermediate solution. Anything further would need to go into design docs and decision records architecture.

frontend/src/pages/projects/notebook/DeleteNotebookModal.tsx
```diff
@@ -36,10 +43,18 @@ const DeleteNotebookModal: React.FC<DeleteNotebookModalProps> = ({ notebook, onC
           (envFrom) => !envFrom.secretRef?.name.includes(DATA_CONNECTION_PREFIX),
         );
         const configMapNames = nonDataConnectionVariables
-          .filter((envName): envName is ConfigMapRef => !!envName.configMapRef)
+          .filter(
+            (envName): envName is ConfigMapRef =>
+              !!envName.configMapRef &&
+              Boolean(envName.configMapRef.name.match(new RegExp(`^${CONFIGMAP_PREFIX}.{6}$`))),
```
andrewballantyne, line 49: `.{6}` in regexp can be any 6 characters... they can be all dashes but the last character and still be valid k8s name but not what we are looking for.

The DSPA secret looks for `[a-z0-9]{6}` -- I think that is more accurate of a check.

andrewballantyne (changes requested):

shalberd:
fixes must-parts (not nice-to-have) parts of
fixes #3521

frontend/src/api/k8s/secrets.ts
```diff
@@ -26,7 +31,7 @@ export const assembleSecret = (
...
-  let name = `secret-${genRandomChars()}`;
+  let name = secretName || getGeneratedSecretName();
```
andrewballantyne, line 34: Hmm... can you update the return block below -- it does a weird mix of this too -- seems odd logic.

andrewballantyne:
@shalberd,  @Gkrumbach07 will be logging another ticket momentarily... we discussed it yesterday.

andrewballantyne (approved):
/hold

Fix the secret name, it's getting kinda bonkers how the code is written and we should try to clean that up imo. But otherwise tested and looks good to me.

frontend/src/api/k8s/secrets.ts
Gkrumbach07: done

Gkrumbach07:
follow up jira: https://issues.redhat.com/browse/RHOAIENG-16608

frontend/src/api/k8s/configMaps.ts
```diff
@@ -10,6 +10,12 @@ import { ConfigMapModel } from '~/api/models';
 import { genRandomChars } from '~/utilities/string';
 import { applyK8sAPIOptions } from '~/api/apiMergeUtils';
 
+export const CONFIGMAP_PREFIX = 'configmap-';
+
+export const getGeneratedConfigMapName = (): string => `${CONFIGMAP_PREFIX}${genRandomChars()}`;
+export const isGeneratedConfigMapName = (name: string): boolean =>
+  Boolean(name.match(new RegExp(`^${CONFIGMAP_PREFIX}[a-z0-9]{6}$`)));
```
DaoDaoNoCode, line 17: Though that might be the same result, just curious, would it be better to use ``new RegExp(`^${CONFIGMAP_PREFIX}[a-z0-9]{6}$`).test(name)`` instead of forcing to cast the type here?

DaoDaoNoCode (approved):
left one small comment, it lgtm overall.
/lgtm

frontend/src/api/k8s/configMaps.ts
Gkrumbach07: yeah its cleaner to use test. i made the switch can you readd the lgtm

andrewballantyne (approved):
/unhold

PR #3542, opened 2024-12-04
Add kserve raw feature flag

emilys314:
https://issues.redhat.com/browse/RHOAIENG-16488

## Description

Adds `disableKServeRaw` feature flag. defaulted to true for now

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/backend/src/types.ts
+++ b/backend/src/types.ts
@@ -34,6 +34,7 @@ export type DashboardConfig = K8sResourceCommon & {
       disableKServe: boolean;
       disableKServeAuth: boolean;
       disableKServeMetrics: boolean;
+      disableKServeRaw: boolean;
       disableModelMesh: boolean;
       disableAcceleratorProfiles: boolean;
       disableDistributedWorkloads: boolean;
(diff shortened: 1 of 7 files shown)


manifests/rhoai/shared/odhdashboardconfig/odhdashboardconfig.yaml
```diff
@@ -25,6 +25,7 @@ spec:
     disableKServe: false
     disableKServeAuth: false
     disableKServeMetrics: false
+    disableKServeRaw: true
```
christianvogt, line 28: We should not be adding feature flags to our custom resource definition or config any more until it is ready for release.
The rest of the code can stay and you can toggle the feature flag using the dev feature flags.

emilys314: Thanks, I removed it, should i remove it from `manifests/common/crd/odhdashboardconfigs.opendatahub.io.crd.yaml` as well?

christianvogt: @emilys314 good question. I'm inclined to say remove it. 

The reason I say this is because in general there's potential for a feature flag to change over the course of development and we don't want to release a flag early that will change or give the customer an ability to disable and then when we do release the feature, they won't get the new default.

PR #3546, opened 2024-12-04
[RHOAIENG-15773] Fix MR permissions management error (give dashboard ServiceAccount permission to get endpoints)

mturley:
Resolves critical bug [RHOAIENG-15773](https://issues.redhat.com/browse/RHOAIENG-15773).

## Description

Due to a change in the Model Registry operator ([PR here](https://github.com/opendatahub-io/model-registry-operator/pull/160/files#diff-ef231245a2fe05b23d3b1a452ddfacc56a03cd441b5bd2b7af46061466043eed), [issue here](https://issues.redhat.com/browse/RHOAIENG-15574)), the `registry-user-${registryname}` role that is generated for each model registry now has an additional permission: get endpoints.

The service account used by the dashboard to create rolebindings for managing model registry permissions does not have permission to get endpoints today. Because of this change, the Manage Permissions page's requests to create new rolebindings using this role are failing in the dashboard. This is due to a cluster API restriction that a user may not grant other users permissions that they don't have themselves.

We would like to discuss a longer-term solution to the constraint we've found here. It should not be possible to cause dashboard permission regressions by changing roles in another component (ideally, any user with rhoai admin permissions should have the same permissions any model registry user have). But for now, the short term fix is to add this new `get endpoints` permission to our ServiceAccount's ClusterRole.

## How Has This Been Tested?

Observe that with these manifests changed (via a devFlag on the cluster), the `odh-dashboard` ClusterRole has the changes from this PR's diff. Go to Model Registry Settings, click Manage Permissions on a registry, try to add a user, group or project, and make sure it succeeds.

## Test Impact

N/A, end to end tests would be nice for this case though once we are ready for that.

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/manifests/core-bases/base/cluster-role.yaml
+++ b/manifests/core-bases/base/cluster-role.yaml
@@ -219,3 +219,9 @@ rules:
       - delete
     resources:
       - accounts
+  - verbs:
+      - get
+    apiGroups:
+      - ''
+    resources:
+      - endpoints


andrewballantyne (approved):

andrewballantyne:
/hold

Double checking with the architects before we merge

andrewballantyne:
/unhold

No response, lets go ahead with this -- if we need to revert it, we have time.

PR #3547, opened 2024-12-04
Update typeahead select to make it auto select the only option as needed

DaoDaoNoCode:
JIRA: [RHOAIENG-16038](https://issues.redhat.com/browse/RHOAIENG-16038)

## Description

Add a hook in TypeaheadSelect to auto-select the only option and disable the field. This only happens when:
1. The field is required (e.g. When adding a cluster storage, the connected notebook field is not required, we cannot force to select the only workbench)
2. The field is not creatable

## How Has This Been Tested?

Verify all the components that use `TypeaheadSelect` is working as expected.
- If it's creatable, it should not be changed by this PR
- If it's not required, it should not be changed by this PR
- Otherwise, if there is only one option in the select menu list, it should be disabled and auto-select the only option

## Test Impact

Update the cypress test in workbench

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/pages/projects/screens/detail/storage/ClusterStorageTableRow.tsx
+++ b/frontend/src/pages/projects/screens/detail/storage/ClusterStorageTableRow.tsx
@@ -82,7 +82,7 @@ const ClusterStorageTableRow: React.FC<ClusterStorageTableRowProps> = ({
           <TypeaheadSelect
             shouldFocusToggleOnSelect
             placeholder={placeholderText}
-            isDisabled={!availableNotebooks.loaded || selectOptions.length === 0}
+            isDisabled={!availableNotebooks.loaded}
             selectOptions={selectOptions}
             selected={obj.name}
             onClearSelection={() => onNotebookSelect('')}
@@ -91,6 +91,7 @@ const ClusterStorageTableRow: React.FC<ClusterStorageTableRowProps> = ({
                 onNotebookSelect(selectedValue);
               }
             }}
+            previewDescription={false}
           />
         ) : (
           <>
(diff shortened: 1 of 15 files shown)


Gkrumbach07:
can you update the storage class select. it is disabling the options instead of hiding them

DaoDaoNoCode:
Sure, and I would prefer to hold this PR until https://github.com/opendatahub-io/odh-dashboard/pull/3529 gets merged. I will check the usage with the new connected workbench field.

DaoDaoNoCode:
/hold
There are still some issues with the cluster storage modal. Will check it next week.

DaoDaoNoCode:
/unhold
Should be ready for reviews now.

frontend/src/pages/projects/screens/detail/storage/ClusterStorageTableRow.tsx
```diff
@@ -61,7 +62,7 @@ const ClusterStorageTableRow: React.FC<ClusterStorageTableRowProps> = ({
 
   const selectOptions = availableNotebooks.notebooks.map((notebook) => ({
     value: notebook.metadata.name,
-    content: notebook.metadata.name,
+    content: getDisplayNameFromK8sResource(notebook),
```
Gkrumbach07, line 65: this is being done here https://github.com/opendatahub-io/odh-dashboard/pull/3584#pullrequestreview-2509549154

Gkrumbach07 (changes requested):

Gkrumbach07 (approved):
/lgtm

PR #3548, opened 2024-12-05
Add GitHub action to create Quay tags for release

manaswinidas:
Github action to create Quay tags for ODH release

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/.github/workflows/authorized-tag-creators.txt
+++ b/.github/workflows/authorized-tag-creators.txt
@@ -1,0 +1,1 @@
+manaswinidas
--- a/.github/workflows/authorized-tag-creators.txt
+++ b/.github/workflows/authorized-tag-creators.txt
@@ -1,1 +1,9 @@
 manaswinidas
+andrewballantyne
+lucferbux
+alexcreasy
+christianvogt
+mturley
+Gkrumbach07
+dgutride
+adnankhan666
--- a/.github/workflows/create-tag-release.yml
+++ b/.github/workflows/create-tag-release.yml
@@ -1,0 +1,48 @@
+name: Create new Quay tag from latest main-tag for ODH release
+on:
+  workflow_dispatch:
+    inputs:
+      existingTag:
+        description: Source tag
+        required: true
+        type: string
+      releaseTag:
+        description: Destination tag
+        required: true
+        type: string
+
+env:
+  QUAY_ODH_DASHBOARD_IMAGE_REPO: ${{ secrets.QUAY_ODH_DASHBOARD_IMAGE_REPO }}
+
+jobs:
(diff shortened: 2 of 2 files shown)


.github/workflows/authorized-tag-creators.txt
```diff
@@ -0,0 +1 @@
+manaswinidas
```
andrewballantyne, line 1: This is a list we should expand if it's dynamically read. 

Managers at a min (maybe a good start?), advisors should all be here (would make the most sense given their current responsibilities)... I don't have a problem with you being here, but I think if this is an actual grant list, we should definitely make sure it is prefilled with a good set of people so we are not locked out of anything.

andrewballantyne (changes requested):
Have you tested this? If so, how'd you go about testing it?

.github/workflows/authorized-tag-creators.txt
manaswinidas: I tested by pushing the changes to my fork and personal Quay repository with my secrets added to the repository. 

Ok, adding the advisors and managers. I didn't add you and Lucas coz you both already have access 😛

manaswinidas: Added myself coz I tested with the same.

andrewballantyne: Push the changes?

andrewballantyne: @manaswinidas can you take a screenshot of what you're seeing? I'm not seeing an update here. I think you're not pushing 🤔

manaswinidas: just added advisors and managers to the text file.. don't mind my previous comment I thought I wrote something weird

manaswinidas: About testing: here's a failed workflow run when I didn't add myself to the file: https://github.com/manaswinidas/odh-dashboard/actions/runs/12182131959/job/33980387355 and this is passing workflow run when I added myself to the file(this created a tag in my personal Quay repo on successful run): https://github.com/manaswinidas/odh-dashboard/actions/runs/12182092368/job/33980257364. Not sure whether you can access these links.

manaswinidas: And this is the workflow that I used there: https://github.com/manaswinidas/odh-dashboard/blob/main/.github/workflows/create-tag-release.yml

andrewballantyne (approved):
Lets see if this does what we are looking for. If this has issues we can just add more changes. It's not a product feature -- we can disable the github workflow if it goes wildly off the rails.

PR #3551, opened 2024-12-06
fix: add capability to update existing secret

olavtar:
https://issues.redhat.com/browse/NVPE-77

## Description

Added the ability to update the existing secret with the new API key.

## How Has This Been Tested?

- Tested by updating the secret. 
- Tested by removing the account and trying to update the secret.

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/backend/src/routes/api/integrations/nim/index.ts
+++ b/backend/src/routes/api/integrations/nim/index.ts
@@ -62,42 +62,11 @@ module.exports = async (fastify: KubeFastifyInstance) => {
         const enableValues = request.body;
 
         try {
-          const { createdNew } = await createNIMSecret(fastify, enableValues);
-          // If the secret already exists, ensure the account is created or retrieved
-          if (!createdNew) {
-            try {
-              const account = await getNIMAccount(fastify);
-              if (!account) {
-                // If the account does not exist, create it
-                const response = await createNIMAccount(fastify);
-                const isEnabled = isAppEnabled(response);
-                reply.send({
-                  isInstalled: true,
-                  isEnabled: isEnabled,
-                  canInstall: false,
-                  error: '',
-                });
-              } else {
-                // If the account exists, get its status
-                const isEnabled = isAppEnabled(account);
-                reply.send({
-                  isInstalled: true,
-                  isEnabled: isEnabled,
-                  canInstall: false,
-                  error: '',
-                });
-              }
-            } catch (accountError: any) {
-              const message =
-                accountError.response?.statusCode === 409
-                  ? 'NIM account already exists.'
-                  : `Failed to create or retrieve NIM account: ${accountError.response?.body?.message}`;
-              fastify.log.error(message);
-              reply.status(accountError.response?.statusCode || 500).send(new Error(message));
-            }
-          } else {
-            // If the secret is newly created, ensure the account is created
-            try {
+          await createNIMSecret(fastify, enableValues);
+          // Ensure the account exists
+          try {
+            const account = await getNIMAccount(fastify);
+            if (!account) {
               const response = await createNIMAccount(fastify);
               const isEnabled = isAppEnabled(response);
               reply.send({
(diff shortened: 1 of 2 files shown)


backend/src/routes/api/integrations/nim/nimUtils.ts
```diff
@@ -70,7 +70,7 @@ export const createNIMSecret = async (
   fastify: KubeFastifyInstance,
   namespace: string,
```
andrewballantyne, line 71: Missed this before -- lets not pass the namespace -- get it out of fastify in the function.

olavtar: Updated

backend/src/routes/api/integrations/nim/index.ts
```diff
@@ -61,36 +60,45 @@ module.exports = async (fastify: KubeFastifyInstance) => {
...
+              // Account not found after the secret was updated
+              fastify.log.error(`NIM account not found after updating secret.`);
+              reply.status(404).send(new Error('NIM account not found.'));
             }
-          });
+          }
+        } catch (e: any) {
+          if (e.response?.statusCode === 409) {
+            fastify.log.error(`NIM secret already exists, skipping creation.`);
+            reply.status(409).send(new Error(`NIM secret already exists, skipping creation.`));
```
andrewballantyne, line 94: You note:
```
            // Secret was updated, so we skip creating the NIM account but still need to get the status.
```
Is this even possible anymore?

backend/src/routes/api/integrations/nim/nimUtils.ts
```diff
@@ -68,10 +66,10 @@ export const createNIMAccount = async (
 
 export const createNIMSecret = async (
   fastify: KubeFastifyInstance,
-  namespace: string,
   enableValues: { [key: string]: string },
-): Promise<SecretKind> => {
+): Promise<{ secret: SecretKind; createdNew: boolean }> => {
   const { coreV1Api } = fastify.kube;
+  const { namespace } = fastify.kube;
```
andrewballantyne, line 72: Nit, you should really put these in the same line

andrewballantyne (changes requested):

backend/src/routes/api/integrations/nim/index.ts
emilys314: So the `e.response?.statusCode === 409` is still possible ..but from the `const response = await createNIMAccount(fastify);` because there is no longer the `.catch` for it. So if `createNIMAccount(fastify)` does fail the error message would be logged as the "Nim secret already exists,..." when it should be "Nim account already exists..."

So we also need to make sure NIMAccount api calls are still being caught properly

backend/src/routes/api/integrations/nim/index.ts
```diff
@@ -61,36 +60,67 @@ module.exports = async (fastify: KubeFastifyInstance) => {
...
+                  .status(accountError.response?.statusCode || 500)
+                  .send(
+                    new Error(
+                      `Failed to create NIM account, ${accountError.response?.body?.message}`,
+                    ),
+                  );
+              }
+            }
+          } else {
+            // Secret was updated, so we skip creating the NIM account but still need to get the status.
```
emilys314, line 93: One last thing, so if the `NIMSecret` already exists, we will skip creating the `NIMAccount.` Could there be a scenario were the `NIMSecret` exists, but the `NIMAccount` does not? In this case this code would never create the account. Like if there was a failure and the user is trying to redo it, or the account got deleted accidentally.

Just in case, I think it may be more defensive to also do a create/get for the account no matter if the secret was already existing or not?

olavtar: It is currently possible, but it should no longer be possible once the backend is updated to delete it along with all the associated resources. Should I go ahead and update the code anyway?

backend/src/routes/api/integrations/nim/index.ts
```diff
@@ -62,35 +61,72 @@ module.exports = async (fastify: KubeFastifyInstance) => {
...
-              reply
-                .status(e.response.statusCode)
-                .send(new Error(`Failed to create NIM secret, ${e.response?.body?.message}`));
+            } catch (accountError: any) {
+              const message = `Failed to create NIM account: ${accountError.response?.body?.message}`;
+              fastify.log.error(message);
+              reply.status(accountError.response?.statusCode || 500).send(new Error(message));
             }
-          });
+          }
```
emilys314, line 114: Okay cool, now that the `if (!createdNew) ...` supports the creationion of a new account as well, I think we can remove the entire `if ... else ...` and only keep the code block that was inside the if statement.

backend/src/routes/api/integrations/nim/index.ts
emilys314: With the latest update that checks if the account and secret exist, the 409 in here should no longer be possible again I believe?

olavtar: Yes, you are right.

andrewballantyne (approved):
I think this change is needed, please do it in a future PR.

Nothing else is wrong with this PR so I'm going to approve it; but it definitely could use with some wording update.

backend/src/routes/api/integrations/nim/nimUtils.ts
```diff
@@ -68,10 +64,9 @@ export const createNIMAccount = async (
 
 export const createNIMSecret = async (
```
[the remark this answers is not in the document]
olavtar: Ok, will update in the future PR. Thanks.

PR #3555, opened 2024-12-06
feat(15898): create MR secure db feature flag, skeleton for form elements

gitdallas:
closes https://issues.redhat.com/browse/RHOAIENG-15898

## Description
implement feature flag for securedb, which when enabled will show newly implemented "Add CA certificate to secure database connection" at the bottom of the create mr modal. checking that will give 4 ways that a certificate can be added.

this looks different than the mocks as "Note" was added to the file upload alert and we've since upgraded to pf6. @yih-wang 

before checking
image
after check
image
clicking existing
image
the searchable drop menu
image
the file upload
image

## How Has This Been Tested?
tested locally by triggering the new flag and checking the box at the bottom of the create mr form. check that form can be submitted if one of the first two radios are selected. if one of the last two are selected, you'll need to supply info before you can submit the form.

no real data yet, so that's about the extent of testing it.

## Test Impact
none yet, this is just skeleton stuff

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/pages/modelRegistrySettings/CreateModal.tsx
+++ b/frontend/src/pages/modelRegistrySettings/CreateModal.tsx
@@ -27,6 +27,7 @@ import FormSection from '~/components/pf-overrides/FormSection';
 import { AreaContext } from '~/concepts/areas/AreaContext';
 import { SupportedArea, useIsAreaAvailable } from '~/concepts/areas';
 import SearchSelector from '~/components/searchSelector/SearchSelector';
+import { PemFileUpload } from './PemFileUpload';
 
 type CreateModalProps = {
   onClose: () => void;
@@ -44,6 +45,7 @@ type SecureDBInfo = {
   radio: SecureDBRadios;
   nameSpace: string;
   configMap: string;
+  certificate: string;
   key: string;
 };
 
@@ -65,6 +67,7 @@ const CreateModal: React.FC<CreateModalProps> = ({ onClose, refresh }) => {
     radio: SecureDBRadios.CLUSTER_WIDE,
     nameSpace: '',
     configMap: '',
+    certificate: '',
     key: '',
   });
   const [searchValue, setSearchValue] = React.useState('');
@@ -202,10 +205,30 @@ const CreateModal: React.FC<CreateModalProps> = ({ onClose, refresh }) => {
     hasContent(password) &&
     hasContent(port) &&
     hasContent(username) &&
-    hasContent(database);
+    hasContent(database) &&
+    validSecureDBOptions();
+
+  const validSecureDBOptions = () => {
+    if (!addSecureDB) {
+      return true;
+    }
+    if ([SecureDBRadios.CLUSTER_WIDE, SecureDBRadios.OPENSHIFT].includes(secureDBInfo.radio)) {
+      return true;
+    }
+    if (secureDBInfo.radio === SecureDBRadios.EXISTING) {
+      return (
+        hasContent(secureDBInfo.configMap) &&
+        hasContent(secureDBInfo.key)
+      );
+    }
+    if (secureDBInfo.radio === SecureDBRadios.NEW) {
+      return hasContent(secureDBInfo.certificate);
+    }
+    return false;
+  }
 
   const handleSecureDBTypeChange = (type: SecureDBRadios) => {
-    setSecureDBInfo({ radio: type, nameSpace: '', key: '', configMap: '' });
+    setSecureDBInfo({ radio: type, nameSpace: '', key: '', configMap: '', certificate: '' });
   };
 
   return (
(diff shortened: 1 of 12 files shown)


yih-wang:
Looks good, thanks @gitdallas! 
The only comment I have is for the content indentation inside each radio button - the contents should be left aligned with the option title/description rather than the radio button, see the PF example [here](https://www.patternfly.org/components/forms/radio#with-description-and-body). So, for example, the **resource** and **key** fields should be aligned with `You can select the key of any configMap or Secret...` in the screen below.
image

frontend/src/pages/modelRegistrySettings/CreateModal.tsx
```diff
@@ -18,12 +23,30 @@ import NameDescriptionField from '~/concepts/k8s/NameDescriptionField';
...
+}
+
+type SecureDBInfo = {
+  radio: SecureDBRadios;
+  nameSpace: string;
+  configMap: string;
+  certificate: string;
+  key: string;
+};
+
```
manaswinidas, line 49: Wondering if these types should be here or in some `types` file? Maybe `frontend/src/concepts/modelRegistry/types.ts`?

manifests/rhoai/shared/odhdashboardconfig/odhdashboardconfig.yaml
```diff
@@ -28,6 +28,7 @@ spec:
     disableModelMesh: false
     disableDistributedWorkloads: false
     disableModelRegistry: false
+    disableModelRegistrySecureDB: true
```
manaswinidas, line 31: Also following [this thread](https://redhat-internal.slack.com/archives/C05SMJ09DD2/p1730922294428399), should we modify this file?

manaswinidas: asking about this in the relevant thread

manaswinidas: We have [an answer](https://redhat-internal.slack.com/archives/C05SMJ09DD2/p1733754222655599?thread_ts=1730922294.428399&cid=C05SMJ09DD2)

gitdallas:
@yih-wang - was due to a typo when moving to pf6, fixed will now look like:
image

frontend/src/pages/modelRegistrySettings/CreateModal.tsx
gitdallas: @manaswinidas - honestly i didn't want to do something too permanent as i'm sort of guessing on what some of the data will look like. also this (especially the `radio` part is very specific to this form and how it shows/hides/validates different things)

manifests/rhoai/shared/odhdashboardconfig/odhdashboardconfig.yaml
gitdallas: i added EA to the flag, big convo but i think that was what is expected out of it

yih-wang:
@gitdallas looks good, thank you!

manifests/rhoai/shared/odhdashboardconfig/odhdashboardconfig.yaml
manaswinidas: The [message](https://redhat-internal.slack.com/archives/C05SMJ09DD2/p1733754298033679?thread_ts=1730922294.428399&cid=C05SMJ09DD2) changed.. seems we don't need `EA` .. we don't need to add the flag to RHOAI OdhDashboardConfig during development .. we just need to add it at the end .. when the feature is going live and enable it then.

manifests/rhoai/shared/odhdashboardconfig/odhdashboardconfig.yaml
```diff
@@ -28,6 +28,7 @@ spec:
     disableModelMesh: false
     disableDistributedWorkloads: false
     disableModelRegistry: false
+    disableModelRegistrySecureDBEA: true
```
andrewballantyne, line 31: Remove this field for development features.

andrewballantyne: [Read more here](https://redhat-internal.slack.com/archives/C05SMJ09DD2/p1733754298033679?thread_ts=1730922294.428399&cid=C05SMJ09DD2)

manifests/rhoai/shared/odhdashboardconfig/odhdashboardconfig.yaml
gitdallas: removed EA, removed flag from odhdashboardconfig

frontend/src/pages/modelRegistrySettings/CreateModal.tsx
mturley: Yeah, we may want to consider reorganizing some of these types as part of API submit logic in [RHOAIENG-15900](https://issues.redhat.com/browse/RHOAIENG-15900) (cc @ppadti) but I don't think it needs to block this PR

mturley (changes requested):
Looking really good @gitdallas, nice work! Just a couple things that shouldn't be too hard to address, and we could also do them as followup PRs if you'd like to just get this in.

One other thing I notice when using this - when I check the checkbox or click a radio button that shows additional content at the bottom of the modal, it doesn't scroll down to show me there are more form fields which is potentially confusing and doesn't feel great. I wonder if we could add some automatic scrolling behavior similar to what we did with the "Add property" button for serving runtime params. Maybe in the click handlers for the controls that toggle expanded content (the checkbox and the latter two radio buttons), if they are being toggled on they should call `.scrollIntoView()` on a container around all the content being expanded (or something at the bottom of that content). We could instead have it always scroll to the bottom, but that would get weird if we added another form section below this one in the future.

frontend/src/pages/modelRegistrySettings/CreateModal.tsx
```diff
@@ -257,6 +363,147 @@ const CreateModal: React.FC<CreateModalProps> = ({ onClose, refresh }) => {
...
+                  />
+                  {secureDBInfo.radio === SecureDBRadios.EXISTING && (
+                    <>
+                      <FormGroup
+                        label="Resource"
+                        isRequired
+                        fieldId="existing-ca-resource"
+                        style={{ marginLeft: 'var(--pf-t--global--spacer--lg)' }}
+                      >
+                        <SearchSelector
```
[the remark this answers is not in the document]
gitdallas: hold off a bit i wanna refactor some more of the logic out of createmodal into the new section component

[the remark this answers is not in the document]
gitdallas: okay, ready for review @mturley

mturley (commented):
Looks like you still have a lint error here @gitdallas

mturley (approved):
LGTM! Thanks @gitdallas . I still think we should try adding the scrolling behavior from my earlier review but it can be part of one of the followup stories or maybe its own issue later.

gitdallas:
oh sorry, missed that.. at this point let's just add it with a later pr

PR #3558, opened 2024-12-09
feat: modify behaviours for the enable button for NIM.

yzhao583:
Jira: [https://issues.redhat.com/browse/NVPE-37](https://issues.redhat.com/browse/NVPE-37)
Jira: [https://issues.redhat.com/browse/NVPE-72](https://issues.redhat.com/browse/NVPE-72)

## Description

- Show [Skeleton](https://www.patternfly.org/components/skeleton/) when visibility of the enable button is determined.
- Disable the submit button if the API key is not provided by user.
- Hide the link to remove the NIM app from the enabled apps page when NIM is disabled.
- Hide the enable button after NIM app is enabled.

image
image
image

## How Has This Been Tested?

- Tested locally.

## Test Impact

- Go to the `Explore` page, select the `NVIDIA NIM` tile, and should see the [Skeleton](https://www.patternfly.org/components/skeleton/)when loading. If NIM is enabled, the `Enable` button should be hidden, if NIM is disabled but can be enabled, the `Enable` button should be displayed, if NIM cannot be enabled, the `Enable` button should be displayed but the button should be disabled, hover the button, the tooltip should be displayed.
- Click the `Enable` button, and the modal to enable NIM should be displayed. If no API key is provided, the `Submit` button should be disabled, otherwise, it should be enabled.
- Provide API key, and click the `Submit` button. If validation success, the modal should be closed and the `Enable` button on the `Explore` page should be hidden.
- Disable NIM and go to the `Enable` page -> click the `disable` on the top-right corner of the NIM tile, the link to remove NIM from this page should NOT be seen.

## Request review criteria:

If you have UI changes: 


frontend/src/pages/exploreApplication/EnableModal.tsx
```diff
@@ -53,6 +60,7 @@ const EnableModal: React.FC<EnableModalProps> = ({ selectedApp, shown, onClose }
       selectedApp.spec.shownOnEnabledPage = true;
       /* eslint-enable no-param-reassign */
 
+      setIsEnableValuesHasEmptyValue(true);
       onClose();
```
emilys314, line 64: Is this line necessary? It looks like once the app is enabled, it's going to close itself. 
Also I don't believe the value is being cleared here either, so it looks a little confusing to see code saying it's being set to empty

yzhao583: The `isEnableValueHasEmptyValue` is the variable to control the disability of the 'Submit' button, it need to be reset once the modal is closed.

frontend/src/components/OdhAppCard.tsx
```diff
@@ -156,18 +157,22 @@ const OdhAppCard: React.FC<OdhAppCardProps> = ({ odhApp }) => {
...
-          removeApplication();
-        }}
-      >
-        here
-      </Button>
-      .
+      {!isInternalRouteIntegrationsApp(odhApp.spec.internalRoute) ? (
+        <>
+          . To remove card click&nbsp;
+          <Button
```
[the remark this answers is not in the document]
yzhao583: This will be changed, UXD suggests using the `Skelton` component to show the loading status of the `Enable` button only, not the whole line.

frontend/src/pages/exploreApplication/EnableModal.tsx
emilys314: Gotcha, I see, the modal doesn't have conditional rendering and instead is returning undefined when it is hidden..

frontend/src/components/OdhAppCard.tsx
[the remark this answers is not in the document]
emilys314: Sorry, can you clarify one what you meant? I think the skeleton loading is good.

The code I referenced is for the "Application disabled" popover
image

As you can see in your screenshot the period `.` at the end of the sentence is now missing after the `here` hyperlink. The suggested change would be to add it back in the popover, so from `...click here` -> `...click here.`
It's a very minor issue but I think is worth fixing.

frontend/src/pages/exploreApplication/EnableModal.tsx
```diff
@@ -18,6 +19,7 @@ const EnableModal: React.FC<EnableModalProps> = ({ selectedApp, shown, onClose }
   const [postError, setPostError] = React.useState('');
   const [validationInProgress, setValidationInProgress] = React.useState(false);
   const [enableValues, setEnableValues] = React.useState<{ [key: string]: string }>({});
+  const [isEnableValuesHasEmptyValue, setIsEnableValuesHasEmptyValue] = React.useState(true);
```
emilys314, line 22: One last suggestion, I think you can change the from a `React.useState` into a `React.useMemo` 
```suggestion
  const isEnableValuesHasEmptyValue = React.useMemo(() => values(updatedValues).some((val) => isEmpty(val)), [updatedValues]);
```
 
This would reduce the need to keeping track of another state and potential mistakes with them getting out of sync. Then you don't need to manually set it to `true` or `false` as it will be done automatically

frontend/src/components/OdhAppCard.tsx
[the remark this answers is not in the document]
yzhao583: Oh! Sorry, I misunderstood. Thanks for the suggestion! Fixed.

frontend/src/pages/exploreApplication/EnableModal.tsx
yzhao583: Great suggestion! Thanks!

frontend/src/utilities/useEnableApplication.tsx
```diff
@@ -124,23 +129,18 @@ export const useEnableApplication = (
...
-                return;
-              }
-
-              if (response.isInstalled) {
-                setEnableStatus({
-                  status: EnableApplicationStatus.SUCCESS,
-                  error: response.error,
-                });
-                dispatchResults(undefined);
               }
```
emilys314, line 138: By removing this, it looks like you will now have an infinite loader waiting for the key to validate when creating a new key
image

You can see in the background the "Enable" button went away because it was validated and enabled, but the modal doesn't update

frontend/src/utilities/useEnableApplication.tsx
```diff
@@ -68,17 +63,27 @@ export const useEnableApplication = (
...
+                    ? EnableApplicationStatus.SUCCESS
+                    : EnableApplicationStatus.FAILED,
+                error:
+                  response.variablesValidationStatus === 'True' ? '' : 'Variables are not valid',
+              });
+              dispatchResults(
+                response.variablesValidationStatus === 'True'
+                  ? undefined
+                  : 'Variables are not valid',
+              );
```
emilys314, line 86: Also the first time I am enabled NIM with the modal, it will tell me it failed. I believe this is because the `response.variablesValidationStatus` is `'Unknown'`

frontend/src/pages/exploreApplication/useIntegratedAppStatus.ts
```diff
@@ -28,8 +29,9 @@ export const useIntegratedAppStatus = (app?: OdhApplication): FetchState<Integra
       isInstalled: false,
       isEnabled: false,
       canInstall: false,
+      variablesValidationStatus: '',
       error: '',
     },
-    { initialPromisePurity: true },
+    { refreshRate: 5000, initialPromisePurity: true },
```
andrewballantyne, line 35:
```suggestion
    { refreshRate: FAST_POLL_INTERVAL, initialPromisePurity: true },
```
Recommend you just use our environment variable for anything you want "quickly done" -- it's 3 seconds, not 5 -- but it's how we typically fetch for information on a polling system.

`POLL_INTERVAL` is 30s, so if you want a slower poll interval, this is also available.

cc @emilys314

andrewballantyne (changes requested):

frontend/src/types.ts
```diff
@@ -666,5 +666,6 @@ export type IntegrationAppStatus = {
   isInstalled: boolean;
   isEnabled: boolean;
   canInstall: boolean;
+  variablesValidationStatus: string;
```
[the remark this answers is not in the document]
emilys314: Yeah I wasn't sure about this. But the variable name specifies status, and the validation come directly from the cr yaml `.status` fields. This means the frontend needs to handle and know the different error codes that can happen in kubernetes.

There are more than 2 states it can be from testing. "Unknown" for example. 

There also is the option for the backend to handle the parse kubernetes validation states and return if it is valid as a boolean.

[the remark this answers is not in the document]
andrewballantyne: This CR is created by the backend, we should be able to know the status for this condition field... 🤔

frontend/src/utilities/useEnableApplication.tsx
yzhao583: @emilys314 The `setEnableStatus({ status: EnableApplicationStatus.INPROGRESS, error: '' });` turn the `status` to `Inprogress`, there is a "watcher" [here](https://github.com/opendatahub-io/odh-dashboard/blob/ca83e34385236beecede593b570de3a5b4279847/frontend/src/utilities/useEnableApplication.tsx#L66) to check the status of the application.

frontend/src/utilities/useEnableApplication.tsx
yzhao583: @emilys314 The logic here is to check if `response.variablesValidationStatus` is true. If yes, `dispatchResults`. If the `response.variablesValidationStatus` is `Unknown`, it should loop again to check the status of the application.

frontend/src/components/OdhAppCard.tsx
```diff
@@ -154,20 +155,24 @@ const OdhAppCard: React.FC<OdhAppCardProps> = ({ odhApp }) => {
...
-        variant="link"
-        onClick={() => {
-          hide();
-          removeApplication();
-        }}
-      >
-        here
-      </Button>
-      .
+      {!isInternalRouteIntegrationsApp(odhApp.spec.internalRoute) ? (
```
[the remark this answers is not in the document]
yzhao583: @andrewballantyne No we do not have a ticket for it now. We should do it as a separated task.

frontend/src/types.ts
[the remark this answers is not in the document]
yzhao583: @andrewballantyne @emilys314 As I tested, the `variablesValidationStatus ` could be ''True'', ''False'', `'Unknown'`, `''`, maybe ENUM is better?

frontend/src/pages/exploreApplication/useIntegratedAppStatus.ts
yzhao583: @andrewballantyne Thanks, will use `FAST_POLL_INTERVAL`

yzhao583: @andrewballantyne Not sure if it is my network issue, 3s is too fast for me. The situation is like, the first API call is triggered but it takes more than 3s to get a response, and now, the second API call is triggered. In this case, there is infinite loading.

frontend/src/utilities/useEnableApplication.tsx
emilys314: Oh okay so it depends on how you test it. If I just delete the `nvidia-nim-access` your PR will work fine and not give an error.

But if the key is invalid (deleted/incorrect/expired) then it will say it's erroring when I submit 
image

The response I get back is `{isInstalled: true, isEnabled: false, variablesValidationStatus: 'False', canInstall: true, error: ''}`

That's because the `getIntegrationAppEnablementStatus()` is polling the existing secret which is invalid via the GET. the POST `enableIntegrationApp()` quickly gets overwritten by the poller which says it's invalid

frontend/src/utilities/useEnableApplication.tsx
emilys314: Gotcha maybe I forgot to run the backend earlier, I don't see it anymore today

frontend/src/pages/exploreApplication/useIntegratedAppStatus.ts
emilys314: Yeah I believe you're correct Yu, the poller can start making new requests even when the previous hasn't finished. 

Is there a reason we need to do it that fast?

yzhao583: @emilys314 We use the poller here to check the enablement status of the application. If the application is enabled, we hide the `Enable` button on the UI, so it needs to be fast.

frontend/src/types.ts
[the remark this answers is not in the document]
yzhao583: Changed to ENUM

frontend/src/utilities/useEnableApplication.tsx
yzhao583: @emilys314 Good catch! I am having hard time fixing this.... Do you have any suggestion?

yzhao583: @emilys314 I have a solution but it is not very ideal.

frontend/src/pages/exploreApplication/useIntegratedAppStatus.ts
yzhao583: The interval is removed.

frontend/src/utilities/useEnableApplication.tsx
emilys314: Yeah I'm not sure the best solution either, it's a tricky thing to fix. I think the root cause is the nim account status isn't updating to `Unknown` or anything when a new key is set. It's still set to `False` until it's done validation.

On the UI side, there is no way to know if the key validation status is old or new. I think if the status could get set to pending or unknown quickly while it's loading, then it would work fine?

emilys314: @yzhao583 can you check with the backend to make sure the secret validation status is reset when a new nim api key is applied?

frontend/src/utilities/useEnableApplication.tsx
emilys314: Actually this still happens. (sorta related to the comment though) 
1. Submit an invalid API key and wait for the error
2. Submit a valid api key, wait for the error (due to the other comment i started)
3. Wait for the api key to validate behind the scenes
4. Submit the valid api key again, you have an infinite loader

frontend/src/utilities/useEnableApplication.tsx
swati-kale: cc: @TomerFi

TomerFi: @emilys314 
We're also working on a way to speed up the status reporting. Currently, we report it at the end of the reconciliation; we'll change this behaviour to report more frequently while the reconciliation is still running. Regardless, this will remain an async operation. While it will be faster, checking the timestamp is better, as you suggested.

yzhao583: @emilys314 @TomerFi Thanks for the suggestions, I will implement the workaround to compare the time stamp.

yzhao583: @emilys314 @TomerFi @swati-kale Implemented the Timestamp workaround, please review again, thanks!

emilys314:
It looks like there is a typescript error with the most recent rebasing to include Olga's pr in `frontend/src/__tests__/cypress/cypress/utils/nimUtils.ts` 

I think we may want to make the properties `variablesValidationStatus` and `variablesValidationTimestamp` optional in type `IntegrationAppStatus` in `frontend/src/types.ts`

yzhao583:
@emilys314 I changed variablesValidationTimestamp and variablesValidationStatus to optional

frontend/src/utilities/useEnableApplication.tsx
```diff
@@ -63,22 +62,39 @@ export const useEnableApplication = (
...
+              setLastVariablesValidationTimestamp(response.variablesValidationTimestamp || 'Unknown');
+              setEnableStatus({
+                status:
+                  response.variablesValidationStatus === VariablesValidationStatus.SUCCESS
+                    ? EnableApplicationStatus.SUCCESS
+                    : EnableApplicationStatus.FAILED,
+                error:
+                  response.variablesValidationStatus === VariablesValidationStatus.SUCCESS
+                    ? ''
+                    : 'Variables are not valid',
```
emilys314, line 91: I'm facing an issue where the first time I submit the api key, it will say it's invalid.
1. Delete the nim secret or change it to be invalid
2. Reload the page (important! otherwise the modal is saving the timestamp state after being closed because this modal isn't great)
3. Enter and submit new key
4. See invalid error

I believe this is happening because the initial value of the timestamp is set to `Unknown`, after the onSubmit POST, it will return the last timestamp which is stale. The watcher sees the timestamp changes and will check the status and see the stale invalid one.

frontend/src/pages/exploreApplication/useIntegratedAppStatus.ts
```diff
@@ -1,10 +1,13 @@
 import * as React from 'react';
-import { IntegrationAppStatus, OdhApplication } from '~/types';
+import { IntegrationAppStatus, OdhApplication, VariablesValidationStatus } from '~/types';
 import useFetchState, { FetchState, NotReadyError } from '~/utilities/useFetchState';
 import { getIntegrationAppEnablementStatus } from '~/services/integrationAppService';
 import { isIntegrationApp } from '~/utilities/utils';
+import { useAppSelector } from '~/redux/hooks';
 
 export const useIntegratedAppStatus = (app?: OdhApplication): FetchState<IntegrationAppStatus> => {
+  const forceUpdate = useAppSelector((state) => state.forceComponentsUpdate);
```
emilys314, line 9: I would avoid using `state.forceComponentsUpdate` in the future. I see it has some usage already merged in. It does work nice though. but some other hooks have a `refresh` function they return which can be used to update the state / status. It doesn't seem like an easy fix so I think it's fine for now

backend/src/routes/api/integrations/nim/index.ts
```diff
@@ -15,11 +23,27 @@ module.exports = async (fastify: KubeFastifyInstance) => {
...
+            });
           } else {
             // Not installed
             fastify.log.info(`NIM account does not exist`);
-            reply.send({ isInstalled: false, isEnabled: false, canInstall: true, error: '' });
+            reply.send({
+              isInstalled: false,
+              isEnabled: false,
+              variablesValidationStatus: VariablesValidationStatus.UNKNOWN,
+              variablesValidationTimestamp: 'Unknown',
```
emilys314, line 43: Nitpick: I'm not sure how I feel about having "Unknown" as the return value for the timestamp. It looks a little strange next to the other enum, and timestamp can have many values, not different enum states, also passing in "Unknown" is a magic string. In my opinion returning an empty value or no timestamp is a more universal sign that there is no timestamp

emilys314 (commented):
Also make sure to run the linter / formatter

frontend/src/pages/exploreApplication/useIntegratedAppStatus.ts
yzhao583: Thanks.

backend/src/routes/api/integrations/nim/index.ts
yzhao583: @emilys314 Thanks for your suggestion. Are you suggesting replacing `Unknown` with `''` ?

yzhao583: @emilys314 Replaced `Unknow` to `''`

frontend/src/utilities/useEnableApplication.tsx
yzhao583: @emilys314 Thanks! Fixed it.

emilys314:
Okay I just tested it again and everything seems to be working

/lgtm
/approve

PR #3559, opened 2024-12-09
Remove invalid jump link

ashley-o0o:
Closes: [RHOAIENG-16189](https://issues.redhat.com/browse/RHOAIENG-16189)

## Description
Removes invalid jump link to Data connections in the create workbench page. In doing so, also cleaned up some unused data connections code.

## How Has This Been Tested?
Tested locally on cluster.

- Navigate to your data science project
- Workbench tab, create workbench.
- The Data connections jump link should no longer exist.

## Test Impact
No tests changed

## Screenshots
Before:
Screenshot 2024-12-09 at 10 35 27 AM

After:
Screenshot 2024-12-09 at 10 34 28 AM

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/pages/projects/screens/spawner/SpawnerPage.tsx
+++ b/frontend/src/pages/projects/screens/spawner/SpawnerPage.tsx
@@ -44,7 +44,6 @@ import ContainerSizeSelector from './deploymentSize/ContainerSizeSelector';
 import EnvironmentVariables from './environmentVariables/EnvironmentVariables';
 import { getCompatibleAcceleratorIdentifiers } from './spawnerUtils';
 import { useNotebookEnvVariables } from './environmentVariables/useNotebookEnvVariables';
-import DataConnectionField from './dataConnection/DataConnectionField';
 import { useNotebookDataConnection } from './dataConnection/useNotebookDataConnection';
 import { useNotebookSizeState } from './useNotebookSizeState';
 import useDefaultStorageClass from './storage/useDefaultStorageClass';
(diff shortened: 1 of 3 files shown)


frontend/src/pages/projects/screens/spawner/SpawnerPage.tsx
```diff
@@ -335,9 +335,9 @@ const SpawnerPage: React.FC<SpawnerPageProps> = ({ existingNotebook }) => {
               />
             ) : (
               <FormSection
-                title={SpawnerPageSectionTitles[SpawnerPageSectionID.DATA_CONNECTIONS]}
-                id={SpawnerPageSectionID.DATA_CONNECTIONS}
-                aria-label={SpawnerPageSectionTitles[SpawnerPageSectionID.DATA_CONNECTIONS]}
+                title={SpawnerPageSectionTitles[SpawnerPageSectionID.CONNECTIONS]}
+                id={SpawnerPageSectionID.CONNECTIONS}
+                aria-label={SpawnerPageSectionTitles[SpawnerPageSectionID.CONNECTIONS]}
```
manaswinidas, line 340: I think the jump link should be hidden behind the feature flag like `Connections` instead of changing `Data Connections` to `Connections`. The change here is leading to inconsistencies when the feature flag `disableConnectionTypes=true`, for example, in the screenshot below when `disableConnectionTypes=true`, the form field is pointing to `Data Connections` (Use a data connection checkbox) while the jump links and from section is pointing to `Connections` which is a different field altogether.
Screenshot 2024-12-10 at 5 38 15 PM

manaswinidas (changes requested):

frontend/src/pages/projects/screens/spawner/SpawnerPage.tsx
ashley-o0o: As we're starting to clean up code from data connections, I have removed this code block instead of putting it behind the feature flag.

christianvogt:
@ashley-o0o 

It's ok to remove the the `Data connections` jump link however you're also removing tests which should stay until we convert the same to work work connection types.

Instead of trying to do the clean up of data connections here, we should keep the code around until we tackle the story to remove it so that we can update tests at the same time.

ashley-o0o:
@christianvogt okay I'll restore the test. Would you like me to comment that part out then? Otherwise the tests fail.

christianvogt:
@ashley-o0o the test should not fail if all you're doing is removing the jump link

PR #3464, opened 2024-11-12
Update 'Resource name' fields to meet UX guidelines: Model registries and Inference services

jeff-phillips-18:
Towards [RHOAIENG-14746](https://issues.redhat.com/browse/RHOAIENG-14746)

## Description
Updates create/edit modals to use `K8sNameDescriptionField` component in order to meet current UX guidelines.
Modals updated:

- Model Registries
- Project Serving (these share the same name)
  - Serving Runtimes (project “model servers”)
  - Inference Services (project models)

### Deploy model
image

### Edit model
image

### Model registry - Deploy model
image

## How Has This Been Tested?
- Navigate to the Model Serving section
  - Select a project
  - Select `Deploy model`
  - Enter the model deployment name
  - View the resource name via the `Edit resource name` link.
  - Change the resource name to something invalid (ie. contains a space)
  - Verify the submit button is disabled
  - Change the resource name to something valid
  - Verify the submit button is enabled
  - Save the deployment.
  - Verify the deployment is created with the correct Display name and Resource Name.

- Navigate to the Model Serving section
  - For an existing Model, select the kebab and choose `Edit`
  - Update the Name field
  - Notice the resource name is not editable
  - Save the model server.
  - Verify the model is updated with the correct Display name.

- Navigate to the Model Registry section
  - Select an existing model
  - For a version, select the kebab and then `Deploy`
  - Select a project
  - Verify the model deployment name is set
  - Enter data for the rest of the form such that `Deploy` is enabled
  - View the resource name via the `Edit resource name` link.
  - Change the resource name to something invalid (ie. contains a space)
  - Verify the `Deploy` button is disabled
  - Change the resource name to something valid
  - Verify the `Deploy` button is enabled
  - Save the deployment.
  - Navigate to the Model Serving section
  -  Find the new model created, select the kebab and choose `Edit`
  - Verify the deployment was created with the correct Display name and Resource Name.

- Navigate to a project that is not configured for kserve or model mesh
  - Select `Select single-model` from the Single-model serving platform card
  - Select `Deploy model`
  - Enter the model deployment name
  - Enter data for the rest of the form such that `Deploy` is enabled
  - View the resource name via the `Edit resource name` link.
  - Change the resource name to something invalid (ie. contains a space)
  - Verify the `Deploy` button is disabled
  - Change the resource name to something valid
  - Verify the `Deploy` button is enabled
  - Save the deployment.
  - Navigate to the Model Serving section
  -  Find the new model created, select the kebab and choose `Edit`
[...]

PR #3561, opened 2024-12-10
Fixing multiple models deploy 

YuliaKrimerman:
Closing issue 2 from https://issues.redhat.com/browse/RHOAIENG-16540

## Description

When selecting multiple models on a multi model server it was throwing and error when trying to deploy a second model.  Found on kserve cluster.
Before 
image
After 
Screenshot 2024-12-10 at 2 11 07 PM

## How Has This Been Tested?

2 Scenarios
1. Any cluster - In Data Science projects go to models, choose Multi-model serving platform. Add a model server.Deploy a model. Try to deploy a 2nd mode. Make sure that no error appears. 
2. Cluster that has NIM enabled- we need to make sure that the InferenceService gets the right name when deploying a kserve model or a NIM model.

## Test Impact

Fixed the failing test

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/api/k8s/__tests__/inferenceServices.spec.ts
+++ b/frontend/src/api/k8s/__tests__/inferenceServices.spec.ts
@@ -140,7 +140,7 @@ describe('assembleInferenceService', () => {
 
   it('should handle name and display name', async () => {
     const displayName = 'Llama model';
-    const resourceName = 'llama-model';
+    const resourceName = 'my-inference-service-test';
 
     const inferenceService = assembleInferenceService(
       mockInferenceServiceModalData({ name: displayName, servingRuntimeName: resourceName }),
--- a/frontend/src/api/k8s/__tests__/inferenceServices.spec.ts
+++ b/frontend/src/api/k8s/__tests__/inferenceServices.spec.ts
@@ -140,10 +140,10 @@ describe('assembleInferenceService', () => {
 
   it('should handle name and display name', async () => {
     const displayName = 'Llama model';
-    const resourceName = 'my-inference-service-test';
+    const resourceName = 'llama-model';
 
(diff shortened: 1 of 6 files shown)


frontend/src/api/k8s/__tests__/inferenceServices.spec.ts
```diff
@@ -140,7 +140,7 @@ describe('assembleInferenceService', () => {
 
   it('should handle name and display name', async () => {
     const displayName = 'Llama model';
-    const resourceName = 'llama-model';
+    const resourceName = 'my-inference-service-test';
```
mturley, line 143: This test seems weird - the resourceName should be a k8s-safe version of the displayName, so `llama-model` makes sense, but below we shouldn't be passing it as the `servingRuntimeName`. In our calls to `mockInferenceServiceModalData` we should pass `name: displayName, k8sName: resourceName` and for this test we shouldn't need to pass a `servingRuntimeName` at all.

mturley (changes requested):
Quick fix for the test here

frontend/src/api/k8s/__tests__/inferenceServices.spec.ts
YuliaKrimerman: Fixed

mturley (approved):

mturley (approved):
LGTM, tested on kserve-pm (modelmesh and kserve) and on `ai-dev04.kni.syseng` (borrowed from the NIM team) for the NIM form changes.

mturley:
FYI @jeff-phillips-18 , fixing some more InferenceService name stuff here that was caused by https://github.com/opendatahub-io/odh-dashboard/pull/3464. Sorry I overlooked these when talking through that PR with you.

PR #3563, opened 2024-12-10
Reduce Namespace Permissions

andrewballantyne:
https://issues.redhat.com/browse/RHOAIENG-16822

## Description

Reduces permissions for the SA around Namespaces. Removed deprecated/unused code to assist with the permissions reduction.

We only PATCH namespaces for model serving today.

## How Has This Been Tested?

On a cluster, should be fine to just deploy it with the DSC DevFlag as the code that is removed was never called.

## Test Impact

None

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/backend/src/utils/resourceUtils.ts
+++ b/backend/src/utils/resourceUtils.ts
@@ -1,14 +1,6 @@
 import * as _ from 'lodash';
 import createError from 'http-errors';
-import {
-  PatchUtils,
-  V1ConfigMap,
-  V1Namespace,
-  V1NamespaceList,
-  V1Role,
-  V1RoleBinding,
-  V1RoleBindingList,
-} from '@kubernetes/client-node';
+import { V1ConfigMap, V1Role, V1RoleBinding, V1RoleBindingList } from '@kubernetes/client-node';
 import {
   AcceleratorProfileKind,
   BuildPhase,
@@ -40,7 +32,6 @@ import {
 import { getComponentFeatureFlags } from './features';
 import { blankDashboardCR } from './constants';
 import { getIsAppEnabled, getRouteForApplication, getRouteForClusterId } from './componentUtils';
-import { createCustomError } from './requestUtils';
 import { getDetectedAccelerators } from '../routes/api/accelerators/acceleratorUtils';
 import { FastifyRequest } from 'fastify';
 import { fetchClusterStatus } from './dsc';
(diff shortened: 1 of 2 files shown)


Gkrumbach07 (approved):
/lgtm

can still switch model serving platform

shalberd:
How is the listing of data science projects in the GUI handled now that namespaces list is not part of the dashboard serviceaccount anymore?

andrewballantyne:
@shalberd hey, let me try to answer some of your inquires... 

> @andrewballantyne How is the listing of data science projects in the GUI handled now that namespaces list is not part of the dashboard serviceaccount anymore?

Nothing was ever listed like this by our service account. If you have access to seeing "all" the data science projects, that's because your user had K8s permissions to do so. This has not changed, your user is your user. 

The Service Account does the following:
* Some infrastructure fetching (who you are, groups, etc)
* Components (Enabled/Explore/Resources pages)
* Jupyter Tile
* Admin functionality

This trimmed the permissions the service account had but never used. Should have no impact what so ever unless you were logging into the service account to do stuff.

> Wondering mostly about namespace labels setting and so on, vs. projects and projectrequest and subjectaccessreview ...
> Is this achieved via openshift api calls now, what you call pass-through calls?

Our Projects API has always been used through the "pass-through API" directly to K8s using your user's token.

> I am trying to understand the new Auth CR in relation to operator v2 and dahboard in ODH, but cannot find it. Maybe I am also missing something in the docs.
> 
> Unclear to me what the operator does and what dashboard and what the logged-in dashboard user
> i.e. what is this for: https://github.com/opendatahub-io/opendatahub-operator/blob/main/config/crd/external/config.openshift.io_authentications.yaml

There are no changes planned today to "remove access" to features already accessed. The Auth Resource moves the ownership of the group configs away from the Dashboard and into the hands of the Operator (aka Platform). 

Platform will grant access to these groups what they need to do to do their role, effectively shifting our product admins (aka ODH Admins) away from using the service account *over time*... this will be a slow process. In the end, our `Admin functionality` will be removed from the list above of what our service account does... and the token of the admin will do it all.

In the end the goal will be simply more of a granular check of your admin permissions -- eg. If you want a partial admin (say a Notebook images only admin) you'd be able to craft a role and a binding for that user to do that functionality and get the UI to show up. 

This is and will be done through SelfSubjectAccessReview (SSAR) checks on the admin pages themselves. Again, this will happen over time and it's unclear what priority this tech debt will be given.

Let me know if you have any other questions.

PR #3567, opened 2024-12-11
fix(test): add wait in deploy model test

gitdallas:
fixes issue with a flakey deploy model test

## Description
When switching projects in the deploy version modal of the UI, I notice there are new network calls and some changes on the UI. This may have caused the test to fail if the option to select didn't exist yet.

I added a `wait` on the network call for getting projects before attempting to select.

I also changed increased the `timeout` of finding the project name to select to 5 seconds.

## How Has This Been Tested?
ran tests locally several times, passed

## Test Impact
hopefully won't intermittently fail anymore

## Request review criteria:


mturley (approved):
LGTM! Thanks

PR #3568, opened 2024-12-11
fix(secureDb): undo debugging boolean flip

gitdallas:
## Description
accidentally left a boolean flipped after debugging, undoing that.

## How Has This Been Tested?
secureDB section in create MR modal should not exist without the feature flag.

## Test Impact
none

## Request review criteria:


mturley (approved):
Oops 👀 that's what I get for not re-testing after a rebase

PR #3569, opened 2024-12-11
Add external routes to AppRoutes component

Gkrumbach07:
https://issues.redhat.com/browse/RHOAIENG-8476

## Description

This PR adds a system to handle external URL redirects in a centralized way, with the first implementation handling Pipeline SDK URLs.

WIP bc waiting for backend to confirm some things but could probably merge before them

## What's Added
- External routes system under `/external/*`
- Redirect handling with loading and error states
- Pipeline SDK URL support for experiments and runs

## How to Extend
1. Create a new redirect component in `frontend/src/pages/external/redirectComponents/`
2. Add your route to `ExternalRoutes.tsx`
3. Use `useRedirect` hook to handle the redirect logic

Example:
  ```tsx
  const [redirect, state] = useRedirect(() => '/foo');
 
  // With async path creation
  const [redirect, state] = useRedirect(async () => {
    const data = await fetchData();
    return `/bar/${data.id}`;
  });
 
  // With options
  const [redirect, state] = useRedirect(() => '/foobar', {
    navigateOptions: { replace: true },
    onComplete: () => console.log('Redirected'),
    onError: (error) => console.error(error)
  });
 
  // Usage
  const createRedirectPath = React.useCallback(() => '/some/path', []);
 
  const [redirect, { loaded, error }] = useRedirect(createRedirectPath);
 
  React.useEffect(() => {
    redirect();
  }, [redirect]);
 
 
  return (
  	<ApplicationsPage
  		loaded={loaded}
  		empty={!!error}
  		emptyStatePage={<RedirectErrorState fallbackUrl="/foo/bar"/>}
  	/>
  );
  ```

**Error state when parsing pipelinesSdk url fails**
_ex) `external/pipelinesSdk/gages-proj/#/some/unexpected/route`_
image

**Custom not found page when incorrect**
_ex) `external/someUndefinedRoute`_
image

## How Has This Been Tested?

Test these URLs locally:
For experiments
http://localhost:4010/external/pipelinesSdk/your-namespace/#/experiments/details/your-experiment-id
For runs
http://localhost:4010/external/pipelinesSdk/your-namespace/#/runs/details/your-run-id

The URLs will redirect to their respective internal routes while handling loading and error states.

## Test Impact

Added a test for the hook

## Request review criteria:

If you have UI changes: 

@kaedward @kywalker-rh 
Yall should look over wording and button styling for this.
- wording for both error messages and buttons (the description text comes from the error message which could be updated to if you want, but basically is generic we couldn't not match the path)
- where should the pipelines redirect one point to
   - they are given something like this when a run completes in the sdk
```bash
Experiment details: http://ml-pipeline-ui-kubeflow.apps.rmartine.dev.datahub.redhat.com/#/experiments/details/2b36adf0-cf2a-4c63-af8e-0d44c731b598
Run details: http://ml-pipeline-ui-kubeflow.apps.rmartine.dev.datahub.redhat.com/#/runs/details/f20096dd-669f-476f-ace3-c33fe7e0864a
```
   - should there be two buttons becuase we dont know if the user was trying to come from one or the other (pipelines or experiment)
- where should the not found button point to if anywhere

### DIFF
--- a/frontend/src/pages/external/ExternalRoutes.tsx
+++ b/frontend/src/pages/external/ExternalRoutes.tsx
@@ -1,0 +1,12 @@
+import React from 'react';
+import { Navigate, Route, Routes } from 'react-router-dom';
+import PipelinesSdkRedirect from './redirectComponents/PipelinesSdkRedirects';
+
+const ExternalRoutes: React.FC = () => (
+  <Routes>
+    <Route path="/pipelinesSdk/:namespace/*" element={<PipelinesSdkRedirect />} />
+    <Route path="*" element={<Navigate to="/not-found" replace />} />
+  </Routes>
+);
+
+export default ExternalRoutes;
--- a/frontend/src/pages/external/ExternalRoutes.tsx
+++ b/frontend/src/pages/external/ExternalRoutes.tsx
@@ -1,11 +1,12 @@
 import React from 'react';
-import { Navigate, Route, Routes } from 'react-router-dom';
+import { Route, Routes } from 'react-router-dom';
 import PipelinesSdkRedirect from './redirectComponents/PipelinesSdkRedirects';
+import ExternalRedirectNotFound from './redirectComponents/ExternalRedirectNotFound';
 
 const ExternalRoutes: React.FC = () => (
   <Routes>
     <Route path="/pipelinesSdk/:namespace/*" element={<PipelinesSdkRedirect />} />
-    <Route path="*" element={<Navigate to="/not-found" replace />} />
+    <Route path="*" element={<ExternalRedirectNotFound />} />
   </Routes>
 );
 
--- a/frontend/src/pages/external/ExternalRoutes.tsx
+++ b/frontend/src/pages/external/ExternalRoutes.tsx
@@ -3,6 +3,9 @@ import { Route, Routes } from 'react-router-dom';
 import PipelinesSdkRedirect from './redirectComponents/PipelinesSdkRedirects';
 import ExternalRedirectNotFound from './redirectComponents/ExternalRedirectNotFound';
 
+/**
+ * Be sure to keep this file in sync with the documentation in `docs/external-redirects.md`.
+ */
 const ExternalRoutes: React.FC = () => (
   <Routes>
     <Route path="/pipelinesSdk/:namespace/*" element={<PipelinesSdkRedirect />} />
--- a/frontend/src/pages/external/RedirectErrorState.tsx
+++ b/frontend/src/pages/external/RedirectErrorState.tsx
@@ -81,7 +81,7 @@ const RedirectErrorState: React.FC<RedirectErrorStateProps> = ({
         icon={ExclamationCircleIcon}
         titleText={title ?? 'Error redirecting'}
         variant={EmptyStateVariant.lg}
-        data-id="redirect-error"
+        data-testid="redirect-error"
       >
         {errorMessage && <EmptyStateBody>{errorMessage}</EmptyStateBody>}
         {fallbackUrl && (
--- a/frontend/src/pages/external/RedirectErrorState.tsx
+++ b/frontend/src/pages/external/RedirectErrorState.tsx
@@ -1,0 +1,106 @@
+import {
+  PageSection,
+  EmptyState,
+  EmptyStateVariant,
+  EmptyStateBody,
+  EmptyStateFooter,
+  EmptyStateActions,
+  Button,
+} from '@patternfly/react-core';
+import { ExclamationCircleIcon } from '@patternfly/react-icons';
+import React from 'react';
+import { useNavigate } from 'react-router-dom';
+import { EitherOrNone } from '~/typeHelpers';
+
+type RedirectErrorStateProps = {
+  title?: string;
+  errorMessage?: string;
+} & EitherOrNone<
+  {
+    fallbackUrl: string;
+    fallbackText: string;
+  },
+  {
+    actions: React.ReactNode | React.ReactNode[];
+  }
+>;
+
+/**
+ * A component that displays an error state with optional title, message and actions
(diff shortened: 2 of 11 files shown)


frontend/src/pages/external/ExternalRoutes.tsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import { Route, Routes } from 'react-router-dom';
+import PipelinesSdkRedirect from './redirectComponents/PipelinesSdkRedirects';
+import ExternalRedirectNotFound from './redirectComponents/ExternalRedirectNotFound';
+
+const ExternalRoutes: React.FC = () => (
+  <Routes>
+    <Route path="/pipelinesSdk/:namespace/*" element={<PipelinesSdkRedirect />} />
```
andrewballantyne, line 8: Are we not doing multiple entries under `external/` because the SDK needs multiple and doesn't want to do that?

Did we ever find out what happened with the `/#/` starting endpoints the SDK was using?

frontend/src/pages/external/RedirectErrorState.tsx
```diff
@@ -0,0 +1,106 @@
...
+  errorMessage?: string;
+} & EitherOrNone<
+  {
+    fallbackUrl: string;
+    fallbackText: string;
+  },
+  {
+    actions: React.ReactNode | React.ReactNode[];
+  }
+>;
```
andrewballantyne, line 26: Why the two states? It's usually odd to give such a large escape hatch 🤔 Not inherently problematic -- but what is the cost of making an Action?

andrewballantyne: Ah -- I see the comment, didn't read everything at first haha -- seems it's easy enough to do it. Either take an array of data and handle the rendering or give it to the caller... seems like little gain to just handle the two strings.

In the end, I think you want to control the rendering as we _always_ want to redirect, no? Not like support a full modal or something

frontend/src/pages/external/redirectComponents/PipelinesSdkRedirects.tsx
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    throw new Error('The URL format is invalid.');
+  }, [namespace, location.hash]);
+
+  const [redirect, { loaded, error }] = useRedirect(createRedirectPath);
+
+  React.useEffect(() => {
+    redirect();
+  }, [redirect]);
```
andrewballantyne, line 47: This model is often very weird. I'm still investigating what's going on here -- but typically you don't useEffect around a response from another hook -- typically that means you just do it in that custom hook.

frontend/src/utilities/useRedirect.ts
```diff
@@ -0,0 +1,89 @@
...
+
+  const navigate = useNavigate();
+  const [state, setState] = React.useState<RedirectState>({
+    loaded: false,
+    error: undefined,
+  });
+
+  const redirect = React.useCallback(async () => {
+    try {
+      const path = await createRedirectPath();
```
andrewballantyne, line 74: Should not `createRedirectPath` *always* return a promise since you're awaiting it here?

frontend/src/pages/external/redirectComponents/PipelinesSdkRedirects.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+  return (
+    <ApplicationsPage
+      loaded={loaded}
+      empty={!!error}
+      emptyStatePage={
```
andrewballantyne, line 53: Should this not be the error state variables?

frontend/src/pages/external/redirectComponents/PipelinesSdkRedirects.tsx
andrewballantyne: Why do you have a loading state if you invoke redirect immediately irrespective of it?

frontend/src/pages/external/ExternalRoutes.tsx
Gkrumbach07: its bc it would be the same entry and the router interprets the `/#/` as the hash string and not part of the path.

frontend/src/pages/external/redirectComponents/PipelinesSdkRedirects.tsx
Gkrumbach07: So you are saying I should just do the redirect in the hook itself. Thats fine i can make that change. And the loading state doesnt make sense here but in some cases it would. For this one i can get rid of it

frontend/src/pages/external/RedirectErrorState.tsx
Gkrumbach07: sure. I can switch it to just being the actions and no fallback url/text

frontend/src/utilities/useRedirect.ts
Gkrumbach07: i mean it is async but the function it is assigned to may not be. Does it hurt to await on a non async function?

I didnt want to create to hooks, one for async one for sync

Gkrumbach07: ah i see what you meant. this doesnt matter anymore bc we just navigate within the hook instead of returning a redirect object

rimolive:
@Gkrumbach07 This change lgtm for the DSP since it somewhat follows the pattern used in Kubeflow Pipelines:

Experiment details: https://ds-pipeline-dspa-pipelinesapi1.apps.dspv2.osp.rh-ods.com//#/experiments/details/c8ce99bc-3545-4f0f-b491-950258a55fab

When this PR is done, I can work to change KFP SDK to add new parameters to customize the experiments and runs page refirects.

DaoDaoNoCode (approved):
Tested on cluster, it can redirect to the pipeline experiment and run details pages well.
/lgtm

PR #3572, opened 2024-12-12
Support edit model registry after it's created

ppadti:
Closes: [RHOAIENG-15879](https://issues.redhat.com/browse/RHOAIENG-15879)

## Description
This PR aims to support edit model registry after it's created.

Screenshot 2024-12-12 at 9 48 09 PM

Loading state:

Screenshot 2024-12-12 at 9 48 36 PM

Error, when password fails to load: 

Screenshot 2024-12-12 at 9 51 36 PM

Screenshot 2024-12-12 at 10 08 02 PM

**Resource name field** 

while creating MR:
Screenshot 2024-12-13 at 1 17 48 PM

Screenshot 2024-12-13 at 1 18 20 PM

while editing MR:

Screenshot 2024-12-13 at 1 18 36 PM

## How Has This Been Tested?
1. Go to model registry settings.
2. Click on kebab action to see Edit model registry.

## Test Impact
Added cypress test

## Request review criteria:

If you have UI changes: 


ppadti:
@yih-wang could you please take a look at the screenshots?

mturley:
@ppadti there's a merge conflict here now that https://github.com/opendatahub-io/odh-dashboard/pull/3555 is merged. For now we can ignore those new secure DB fields for prefilling/editing since the submit logic for them isn't implemented yet.

yih-wang:
Oops sorry, I forgot to involve content review for the error message above...
@kaedward can you help take a look at the error message in this screen? This message will show up when the password is failed to be pulled from the OpenShift console because the source file in OpenShift is missing (been directly deleted in OpenShift, for example)
image

yih-wang:
Hey Pushpa, the screenshots look good, thank you! Can you add a screenshot to show how the first two fields look like?

ppadti:
> Hey @ppadti , the screenshots look good, thank you! Can you add a screenshot to show how the first two fields look like?

@yih-wang sorry, I missed that. Updated the screen shoots.

yih-wang:
Thank you @ppadti, looks good to me!

frontend/src/pages/modelRegistrySettings/CreateModal.tsx
```diff
@@ -245,16 +279,27 @@ const CreateModal: React.FC<CreateModalProps> = ({ onClose, refresh }) => {
...
+            ) : (
+              <PasswordInput
+                isRequired
+                type={showPassword ? 'text' : 'password'}
+                id="mr-password"
+                name="mr-password"
+                value={password}
+                onBlur={() => setIsPasswordTouched(true)}
+                onChange={(_e, value) => setPassword(value)}
+                validated={isPasswordTouched && !hasContent(password) ? 'error' : 'default'}
```
mturley, line 300: This is the same set of props that gets rendered inside ModelRegistryDatabasePassword. Can we just make the `editRegistry` prop of that component optional, and have its editing-specific behavior be conditional?

frontend/src/pages/modelRegistrySettings/ModelRegistryDatabasePassword.tsx
```diff
@@ -1,23 +1,70 @@
...
 }) => {
-  if (loadError) {
-    return <Alert variant="danger" isInline isPlain title="Error loading password" />;
+  const [dbPassword, passwordLoaded, passwordLoadError] = useFetchState(
+    React.useCallback(async () => {
+      const { databasePassword } = await getModelRegistryBackend(mr.metadata.name);
+      return databasePassword;
+    }, [mr]),
+    undefined,
+  );
```
mturley, line 31: To make this conditional, you can have the callback throw a `new NotReadyError();` if the `mr` is undefined. (`import { NotReadyError } from '~/utilities/useFetchState';`). That short-circuits `useFetchState` and makes it skip to the loaded = true state without any loadError value.

frontend/src/pages/modelRegistrySettings/ModelRegistryDatabasePassword.tsx
```diff
@@ -1,23 +1,70 @@
...
+
+  React.useEffect(() => {
+    if (dbPassword) {
+      setPassword(dbPassword);
+    }
+  }, [dbPassword, setPassword]);
```
mturley, line 37: Hmm. I don't love this because at first glance it looks like we're unnecessarily duplicating state, but we're not since `password` is our form input value and `dbPassword` is our existing value in the API. Can we rename `dbPassword` to `existingDbPassword` or `editingDbPassword`, something like that?

frontend/src/pages/modelRegistrySettings/ModelRegistryDatabasePassword.tsx
```diff
@@ -1,23 +1,70 @@
...
+
+  const hasContent = (value: string): boolean => !!value.trim().length;
+
+  if (!passwordLoaded && !passwordLoadError) {
```
mturley, line 41: This condition could also have `mr` in it to only apply to the editing case, but I think the NotReadyError may cover us there

mturley (changes requested):
A small code duplication nit, otherwise LGTM - although I'm having trouble testing it right now, [see thread here](https://redhat-internal.slack.com/archives/C05QB0A582E/p1734367449685839)

frontend/src/pages/modelRegistrySettings/ModelRegistryDatabasePassword.tsx
mturley: This conditional should probably also only run if we're in editing mode, i.e. `mr` is truthy. Otherwise I wonder if we'll run into a case where our password gets cleared on a re-render or something.

mturley (approved):
Perfect! LGTM, thanks @ppadti.

Note: didn't test end to end, but the network request looks correct and we will verify end to end as part of the epic verification. Approving to unblock other stories.

PR #3574, opened 2024-12-12
Add deployment mode select for kserve raw

emilys314:
Closes https://issues.redhat.com/browse/RHOAIENG-16487
Figma https://www.figma.com/design/hKimcyeu6pgAoqKH1XrqmY/Support-KServe-RawDeployment-mode?node-id=231-2365&node-type=canvas&t=KepUcCxzaanRkMU8-0

## Description

First part of adding kserve raw model deployments. This adds the deployment mode dropdown to single serving (kserve) modals. What is done is basically adding `...DeploymentMode: RawDeployment` and the corresponding labels for auth / routing. The backend model controller takes care of the rest.
- Advanced -> Serverless / old behaviour
- Standard -> KServe raw / new

Changing the deployment mode on an existing deployment is not supported so it will be disabled on edit. However the `assembleInferenceService()` does properly replace the labels and annotations. 

The deployment mode dropdown will also show up in NIM

image

## How Has This Been Tested?

1. Have a cluster with the latest backend controllers (I have DSC config dev flags or my cluster I can share)
```yaml
   kserve:
      # defaultDeploymentMode: RawDeployment
      devFlags:
        manifests:
          - contextDir: config
            sourcePath: overlays/odh
            uri: 'https://github.com/VedantMahabaleshwarkar/kserve/tarball/devflags'
          - contextDir: config
            sourcePath: ''
            uri: 'https://github.com/VedantMahabaleshwarkar/odh-model-controller/tarball/devflags'
      managementState: Managed
      serving:
        ingressGateway:
          certificate:
            type: OpenshiftDefaultIngress
        managementState: Managed
        name: knative-serving
```
3. add `?devFeatureFlags` to the URL and make sure the kserve raw is set to false
4. Go to a single model serving project, deploy a new or edit an existing model
5. The default selection should be advanced, but change it to standard (raw) 
6. Everything should work normally. In the console you can verify the inferenceService file, and check under the serverless tab on the sidebar it doesn't exist. Try inferencing the endpoint

(Public endpoint without auth doesn't work)

## Test Impact

cypress and jest tests added

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/__mocks__/mockInferenceServiceModalData.ts
+++ b/frontend/src/__mocks__/mockInferenceServiceModalData.ts
@@ -38,6 +38,7 @@ export const mockInferenceServiceModalData = ({
       },
     },
   },
+  isKServeRawDeployment,
 }: MockResourceConfigType): CreatingInferenceServiceObject => ({
   name,
   k8sName,
@@ -51,4 +52,5 @@ export const mockInferenceServiceModalData = ({
   tokenAuth,
   tokens,
   modelSize,
+  isKServeRawDeployment,
 });
--- a/frontend/src/__tests__/cypress/cypress/pages/modelServing.ts
+++ b/frontend/src/__tests__/cypress/cypress/pages/modelServing.ts
@@ -101,6 +101,10 @@ class InferenceServiceModal extends Modal {
     return this.find().findByTestId('inference-service-framework-selection');
   }
 
+  findDeploymentModeSelect() {
+    return this.find().findByTestId('deployment-mode-select');
+  }
+
   findExistingDataConnectionOption() {
     return this.find().findByTestId('existing-data-connection-radio');
   }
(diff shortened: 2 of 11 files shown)


yih-wang:
Hey @emilys314, is the `(default)` dynamically assigned to the option which is set as the default mode in the cluster settings?

emilys314:
> Hey @emilys314, is the `(default)` dynamically assigned to the option which is set as the default mode in the cluster settings?

In this PR, no it's not dynamic, but a future PR will make it dynamic. Unless we want to remove it completely

yih-wang:
> In this PR, no it's not dynamic, but a future PR will make it dynamic. Unless we want to remove it completely
I see. I didn’t include this default flag in my original mockup, but I think it could be a good idea to have it. We might just need to clarify what ‘default’ means in this context to ensure users understand. That said, I don’t think this is a blocker for the PR. I’ll think through this further and let you know if I decide to suggest any changes.

PR #3575, opened 2024-12-12
Update doc link

aduquett:
https://issues.redhat.com/browse/RHOAIENG-16973 , fix version is 2.17.0

## Description
We want to rename the "Enabling data science pipelines 2.0" section in the RHOAI docs to "Migrating to data science pipelines 2.0" to more accurately reflect the content/procedure. 
As a workaround for the 2.16 docs, I updated the title to "Migrating to data science pipelines 2.0" but I did not change the filename or id so that these links would not break. 
When the links are updated here, I can rename the file and update the id in the doc repo.

## How Has This Been Tested?
The links will be broken until the updated docs are published, but the preview links will be available when I know which release this change is implemented in.


Gkrumbach07 (approved):
/lgtm

PR #3577, opened 2024-12-13
Deprecate & Swap for Auth over DashboardConfig for group configs

andrewballantyne:
https://issues.redhat.com/browse/RHOAIENG-16519

## Description

A new Auth resource is being added by the Platform team. OdhDashboardConfig's group configurations are no longer needed.

When we go live with this the Auth resource, the OdhDashboardConfig groups will become read-only (see https://github.com/opendatahub-io/odh-dashboard/pull/3585), the Operator will copy over the items, and grant permissions to the admin users. This PR uses this functionality of new permissions to check if the admin can modify the Auth resource directly -- if so, pull & update there; if not fall back on todays logic (aka backwards compatibility). Do note, if the Auth resource cannot be modified by this user AND it exists, we will not be able to update the Groups Config from the UI as it will attempt to update the OdhDashboardConfig while it is read-only; known limitation but should be an untravelled path.

The infrastructure to get `isAdmin` uses the Auth resource if it exists, SA has been granted permissions to get the resource.

### Known Error States

If the CRD or the CR does not exist... 
Screenshot 2024-12-17 at 10 12 07 AM

How it happens and for who, are noted below:
* For `cluster-admin`s
    * You'll always succeed the SSAR check
    * CRD not existing or CR not existing breaks the flow
* For `product` admins
    * If you don't have directly provided access -- you fall back (you may not be able to update the resource, but you can read it)
    * If you do have direct access (RBAC granted), and the CRD or CR is missing, it breaks the flow

No matter what for the UI admin page, the resource can still be modified up until the CRD is merged with a no change allowed setting (see https://github.com/opendatahub-io/odh-dashboard/pull/3585)

## How Has This Been Tested?

Today, the work for the Operator is a WIP. I used a CRD from their code & setup a rolebinding for my user:

CRD
```yaml
kind: CustomResourceDefinition
apiVersion: apiextensions.k8s.io/v1
metadata:
  name: auths.services.platform.opendatahub.io
  annotations:
    controller-gen.kubebuilder.io/version: v0.16.1
spec:
  group: services.platform.opendatahub.io
  names:
    plural: auths
    singular: auth
    kind: Auth
    listKind: AuthList
  scope: Cluster
  versions:
    - name: v1alpha1
      served: true
      storage: true
      schema:
        openAPIV3Schema:
          description: Auth is the Schema for the auths API
          type: object
          properties:
            apiVersion:
              description: |-
                APIVersion defines the versioned schema of this representation of an object.
                Servers should convert recognized schemas to the latest internal value, and
                may reject unrecognized values.
                More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources
              type: string
            kind:
              description: |-
                Kind is a string value representing the REST resource this object represents.
                Servers may infer this from the endpoint the client submits requests to.
                Cannot be updated.
                In CamelCase.
                More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds
              type: string
            metadata:
              type: object
            spec:
              description: AuthSpec defines the desired state of Auth
              type: object
              required:
                - adminGroups
                - allowedGroups
              properties:
                adminGroups:
                  type: array
                  items:
                    type: string
                allowedGroups:
                  type: array
                  items:
                    type: string
            status:
              description: AuthStatus defines the observed state of Auth
              type: object
              properties:
                conditions:
                  type: array
                  items:
                    description: Condition contains details for one aspect of the current state of this API Resource.
                    type: object
                    required:
                      - lastTransitionTime
                      - message
                      - reason
                      - status
                      - type
                    properties:
                      lastTransitionTime:
                        description: |-
                          lastTransitionTime is the last time the condition transitioned from one status to another.
                          This should be when the underlying condition changed.  If that is not known, then using the time when the API field changed is acceptable.
                        type: string
                        format: date-time
                      message:
                        description: |-
                          message is a human readable message indicating details about the transition.
                          This may be an empty string.
                        type: string
                        maxLength: 32768
                      observedGeneration:
                        description: |-
                          observedGeneration represents the .metadata.generation that the condition was set based upon.
                          For instance, if .metadata.generation is currently 12, but the .status.conditions[x].observedGeneration is 9, the condition is out of date
                          with respect to the current state of the instance.
                        type: integer
                        format: int64
                        minimum: 0
                      reason:
                        description: |-
                          reason contains a programmatic identifier indicating the reason for the condition's last transition.
                          Producers of specific condition types may define expected values and meanings for this field,
                          and whether the values are considered a guaranteed API.
                          The value should be a CamelCase string.
                          This field may not be empty.
                        type: string
                        maxLength: 1024
                        minLength: 1
                        pattern: '^[A-Za-z]([A-Za-z0-9_,:]*[A-Za-z0-9_])?$'
                      status:
                        description: 'status of the condition, one of True, False, Unknown.'
                        type: string
                        enum:
                          - 'True'
                          - 'False'
                          - Unknown
                      type:
                        description: type of condition in CamelCase or in foo.example.com/CamelCase.
                        type: string
                        maxLength: 316
                        pattern: '^([a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*/)?(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])$'
                observedGeneration:
                  type: integer
                  format: int64
                phase:
                  type: string
          x-kubernetes-validations:
            - rule: self.metadata.name == 'auth'
              message: Auth name must be auth
      subresources:
        status: {}
      additionalPrinterColumns:
        - name: Ready
          type: string
          description: Ready
          jsonPath: '.status.conditions[?(@.type=="Ready")].status'
        - name: Reason
          type: string
          description: Reason
          jsonPath: '.status.conditions[?(@.type=="Ready")].reason'
  conversion:
    strategy: None
status:
  conditions:
    - type: NamesAccepted
      status: 'True'
      lastTransitionTime: '2024-12-11T13:05:40Z'
      reason: NoConflicts
      message: no conflicts found
    - type: Established
      status: 'True'
      lastTransitionTime: '2024-12-11T13:05:40Z'
      reason: InitialNamesAccepted
      message: the initial names have been accepted
  acceptedNames:
    plural: auths
    singular: auth
    kind: Auth
    listKind: AuthList
  storedVersions:
    - v1alpha1
```

Testing roles (will be handled by Platform more directly later):
```yaml
kind: ClusterRole
apiVersion: rbac.authorization.k8s.io/v1
metadata:
  name: temp-give-auth-access-role
rules:
  - verbs:
      - get
      - patch
    apiGroups:
      - services.platform.opendatahub.io
    resources:
      - auths
  - verbs:
      - get
      - list
    apiGroups:
      - user.openshift.io
    resources:
      - groups
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: temp-give-auth-access-role-binding
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: temp-give-auth-access-role
subjects:
  - apiGroup: rbac.authorization.k8s.io
    kind: Group
    name: odh-admins
```

## Test Impact

Updated the unit tests to use the new function and make sure the old one is not called.

## Request review criteria:

If you have UI changes: 


Gkrumbach07:
image
error when cr not found for both admin and cluster admin

image
when empty lists are in cr

patching works

/lgtm

andrewballantyne:
@Gkrumbach07 I had to rebase, conflict.

Can you also explain this comment more:
> error when cr not found for both admin and cluster admin

admin -- when role is applied? Or always... because admins have no permissions today and will not succeed on the RBAC SSAR check.

Gkrumbach07 (approved):
/lgtm

PR #3582, opened 2024-12-16
Add backend route for Configmaps and secrets for MR namespace

manaswinidas:
Closes: [RHOAIENG-15899](https://issues.redhat.com/browse/RHOAIENG-15899)

## Description

- Fetch config maps and secrets in model registry namespace and added hook for the same
- Lazy loading(Skeleton) when one clicks on the checkbox
- Added logic to disable/enable the first two options 
   - If the data from the new hook has no ConfigMap with the name `odh-trusted-ca-bundle` or that ConfigMap does not include the key `ca-bundle.crt`, disable the "Use cluster-wide CA bundle" radio button and add a tooltip "No certificate is available in the cluster-wide CA bundle". 
  - If the data from the new hook has no ConfigMap with the name `odh-trusted-ca-bundle` or that ConfigMap does not include the key `odh-ca-bundle.crt`, disable the "Use ${ODH_PRODUCT_NAME} CA bundle" radio button and add a tooltip "No certificate is available in the ${ODH_PRODUCT_NAME} CA bundle".
- Under "Choose from existing certificates":
   - In the Resource dropdown, replaced the mock options with the names of the secrets and configmaps fetched with hook.
   - Make the Key field disabled with the placeholder "Select a resource to view its available keys" until a Resource is selected. When a Resource is selected, populate the Key dropdown from the keys of that secret or configmap.

Screen Recording:

https://github.com/user-attachments/assets/3cf2a631-5f56-4bc6-a7eb-71a0ba553fbb

Screenshots:

Screenshot 2024-12-17 at 7 51 07 PM
Screenshot 2024-12-17 at 7 51 37 PM
Screenshot 2024-12-17 at 7 51 51 PM
Screenshot 2024-12-17 at 7 52 23 PM

## How Has This Been Tested?

- Enable `disableModelRegistrySecureDB` feature flag on your cluster to see the checkbox in the Create model registry modal
- Once you click on the checkbox, you should see a loading state while the config maps and secrets load
- Once that is loaded, some checkboxes may be disabled based on the data provided in the description of the radio buttons(first two radio buttons)
- Third radio button: The config maps and secrets should be populated in the dropdown and the second dropdown should populate according to the config maps/secrets name selected in first dropdown.. second dropdown should be disabled if first dropdown is not selected.

## Test Impact

Tests will be added in a later PR.

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/pages/modelRegistrySettings/CreateMRSecureDBSection.tsx
+++ b/frontend/src/pages/modelRegistrySettings/CreateMRSecureDBSection.tsx
@@ -1,5 +1,5 @@
 import React, { useState } from 'react';
-import { FormGroup, Radio, Alert, MenuItem, MenuGroup } from '@patternfly/react-core';
+import { FormGroup, Radio, Alert, MenuItem, MenuGroup, Tooltip } from '@patternfly/react-core';
 import SearchSelector from '~/components/searchSelector/SearchSelector';
 import { translateDisplayNameForK8s } from '~/concepts/k8s/utils';
 import { RecursivePartial } from '~/typeHelpers';
@@ -27,7 +27,6 @@ interface CreateMRSecureDBSectionProps {
   secureDBInfo: SecureDBInfo;
   modelRegistryNamespace: string;
   nameDesc: { name: string };
-  existingCertKeys: string[];
   existingCertConfigMaps: ConfigSecretItem[];
   existingCertSecrets: ConfigSecretItem[];
   setSecureDBInfo: (info: SecureDBInfo) => void;
@@ -37,12 +36,16 @@ export const CreateMRSecureDBSection: React.FC<CreateMRSecureDBSectionProps> = (
   secureDBInfo,
   modelRegistryNamespace,
   nameDesc,
-  existingCertKeys,
   existingCertConfigMaps,
   existingCertSecrets,
   setSecureDBInfo,
 }) => {
-  const [searchValue, setSearchValue] = useState('');
+  const [configSecretName, setConfigSecretName] = useState('');
+  const [key, setKey] = useState('');
+  const [existingCertKeys, setExistingCertKeys] = useState<string[]>([]);
+  const ODH_TRUSTED_BUNDLE = 'odh-trusted-ca-bundle';
+  const CA_BUNDLE_CRT = 'ca-bundle.crt';
+  const ODH_CA_BUNDLE_CRT = 'odh-ca-bundle.crt';
 
   const hasContent = (value: string): boolean => !!value.trim().length;
 
@@ -61,18 +64,37 @@ export const CreateMRSecureDBSection: React.FC<CreateMRSecureDBSectionProps> = (
   };
 
   const clusterWideCABundle = existingCertConfigMaps.find(
-    (configMap) =>
-      configMap.name === 'odh-trusted-ca-bundle' && configMap.keys.includes('ca-bundle.crt'),
+    (configMap) => configMap.name === ODH_TRUSTED_BUNDLE && configMap.keys.includes(CA_BUNDLE_CRT),
   );
 
   const isClusterWideCABundleAvailable = !!clusterWideCABundle;
 
   const openshiftCAbundle = existingCertConfigMaps.find(
     (configMap) =>
-      configMap.name === 'odh-trusted-ca-bundle' && configMap.keys.includes('odh-ca-bundle.crt'),
+      configMap.name === ODH_TRUSTED_BUNDLE && configMap.keys.includes(ODH_CA_BUNDLE_CRT),
   );
 
-  const isOpenshiftCABundleAvailable = !!openshiftCAbundle;
+  const isProductCABundleAvailable = !!openshiftCAbundle;
+
+  const getKeysByName = (
+    configMaps: ConfigSecretItem[],
+    secrets: ConfigSecretItem[],
+    targetName: string,
+  ): string[] => {
+    let keys: string[] = [];
+    // Search in ConfigMaps first
+    const configMap = configMaps.find((item) => item.name === targetName);
+    if (configMap) {
+      keys = configMap.keys;
+    }
+
+    // If not found, search in Secrets
+    const secret = secrets.find((item) => item.name === targetName);
+    if (secret) {
+      keys = secret.keys;
+    }
+    return keys;
+  };
 
   const handleSecureDBTypeChange = (type: SecureDBRType) => {
     const newInfo = {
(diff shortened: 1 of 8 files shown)


manaswinidas:
WIP coz I need to add tests and a few tooltips

frontend/src/pages/modelRegistrySettings/CreateMRSecureDBSection.tsx
```diff
@@ -58,6 +63,39 @@ export const CreateMRSecureDBSection: React.FC<CreateMRSecureDBSectionProps> = (
...
+    if (configMap) {
+      keys = configMap.keys;
+    }
+
+    // If not found, search in Secrets
+    const secret = secrets.find((item) => item.name === targetName);
+    if (secret) {
+      keys = secret.keys;
+    }
+    return keys;
```
mturley, line 96: Wouldn't this be a problem if we have a configmap and secret with the same name and the user is trying to select the secret? I think we can maybe pass only the list of ConfigSecretItems of the type being selected here.

frontend/src/pages/modelRegistrySettings/CreateMRSecureDBSection.tsx
```diff
@@ -72,51 +110,42 @@ export const CreateMRSecureDBSection: React.FC<CreateMRSecureDBSectionProps> = (
...
+    const newKeys = getKeysByName(existingCertConfigMaps, existingCertSecrets, selectedName);
+    setExistingCertKeys(newKeys); // Update keys for the second dropdown for existing certificate field
+
+    const newInfo = {
+      ...secureDBInfo,
+      configMap: selectedName,
+      key: '',
+    };
+
+    setSecureDBInfo({ ...newInfo, isValid: isValid(newInfo) });
```
mturley, line 124: I think we need to modify `SecureDBInfo` to include whether the selected resource is a ConfigMap or a Secret. That'll be needed for the submit logic.

frontend/src/pages/modelRegistrySettings/CreateMRSecureDBSection.tsx
```diff
@@ -25,22 +27,25 @@ interface CreateMRSecureDBSectionProps {
...
   secureDBInfo,
   modelRegistryNamespace,
   nameDesc,
-  existingCertKeys,
   existingCertConfigMaps,
   existingCertSecrets,
   setSecureDBInfo,
 }) => {
-  const [searchValue, setSearchValue] = useState('');
+  const [configSecretName, setConfigSecretName] = useState('');
```
mturley, line 43: I was a little confused by the name of this since it's the search input and not the selected value, can we rename it to something with `search` in the name?

frontend/src/pages/modelRegistrySettings/CreateMRSecureDBSection.tsx
```diff
@@ -25,22 +27,25 @@ interface CreateMRSecureDBSectionProps {
...
+  const [key, setKey] = useState('');
+  const [existingCertKeys, setExistingCertKeys] = useState<string[]>([]);
```
mturley, line 45: If we have `secureDBInfo` to tell us which secret or configmap is selected, we can derive the list of keys in it at render time and then we don't need this extra state for `existingCertKeys` I think?

frontend/src/pages/modelRegistrySettings/CreateMRSecureDBSection.tsx
```diff
@@ -25,22 +27,25 @@ interface CreateMRSecureDBSectionProps {
...
   modelRegistryNamespace,
   nameDesc,
-  existingCertKeys,
   existingCertConfigMaps,
   existingCertSecrets,
   setSecureDBInfo,
 }) => {
-  const [searchValue, setSearchValue] = useState('');
+  const [configSecretName, setConfigSecretName] = useState('');
+  const [key, setKey] = useState('');
```
mturley, line 44: Same here, this is the search input for the key and not the selected key so maybe we should rename it

ppadti:
> @yih-wang do you think the Skeleton should be a larger EmptyState with spinner? I wasn't sure how we should handle loading the data for these fields. I don't think we can just show the radio buttons immediately and skeletons for the dropdowns, because we need to load the data to know whether the first two radio buttons should be enabled. I think in another thread we had discussed having a loading state for the whole section like this.

I think, we had this discussion here in this issue: [RHOAIENG-15879](https://issues.redhat.com/browse/RHOAIENG-15879).

frontend/src/pages/modelRegistrySettings/CreateMRSecureDBSection.tsx
mturley: Dang, also we probably shouldn't have `isValid` in state but that's probably out of the scope of this PR. I missed that when reviewing the last PR.

mturley (changes requested):
Mostly looking great here! A couple of nitpicks 👀

mturley:
> I think, we had this discussion here in this issue: [RHOAIENG-15879](https://issues.redhat.com/browse/RHOAIENG-15879).

Ah thanks @ppadti, yeah so we should use an EmptyState with a large spinner for the whole section rather than a skeleton I think.

frontend/src/pages/modelRegistrySettings/CreateMRSecureDBSection.tsx
manaswinidas: Yes, that's right. Removed this state altogether after adding `resourceType` to `SecureDBInfo`

yih-wang:
> Ah thanks @ppadti, yeah so we should use an EmptyState with a large spinner for the whole section rather than a skeleton I think.

@mturley Seems you already have the answer! Yes let's use a spinner for loading the whole section under the checkbox.

yih-wang:
All look good. Awesome work! Thank you @manaswinidas!

frontend/src/pages/modelRegistrySettings/CreateMRSecureDBSection.tsx
```diff
@@ -102,6 +89,7 @@ export const CreateMRSecureDBSection: React.FC<CreateMRSecureDBSectionProps> = (
       nameSpace: '',
       key: '',
       configMap: '',
```
mturley, line 91: Total nit, we can keep it as-is if you like - the `configMap` key in SecureDBInfo is really the name of a configmap or a secret. I wonder if now that we have `resourceType` we should rename this to `resourceName`.

mturley (commented):
One tiny thing...

mturley (approved):
Will address the above nit in reviewing the next PR. LGTM! Thanks @manaswinidas !

PR #3584, opened 2024-12-17
show display name instead of pvc metadata name

pnaik1:
https://issues.redhat.com/browse/RHOAIENG-16239

## Description
Cluster storage is now showing the displayname before  and after creation of a workbench  

## How Has This Been Tested?

1. Create a PVC with displayName ( you can find the example in the jira issue)
2. Create a workbench 
3. Attach the created PVC, check if the displayed name of attached PVC is attached to workbench
4. Once created, go to cluster storage and check if the display name is still showing instead of metadata name

## Test Impact
N/A

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/pages/projects/screens/spawner/storage/AddExistingStorageField.tsx
+++ b/frontend/src/pages/projects/screens/spawner/storage/AddExistingStorageField.tsx
@@ -35,7 +35,7 @@ const AddExistingStorageField: React.FC<AddExistingStorageFieldProps> = ({
     () =>
       loaded
         ? storages.reduce((acc: TypeaheadSelectOption[], pvc) => {
-            if (!existingStorageNames?.includes(pvc.metadata.name)) {
+            if (!existingStorageNames?.includes(getDisplayNameFromK8sResource(pvc))) {
               acc.push({
                 value: pvc.metadata.name,
                 content: getDisplayNameFromK8sResource(pvc),
@@ -64,7 +64,7 @@ const AddExistingStorageField: React.FC<AddExistingStorageFieldProps> = ({
 
   if (!loaded) {
     placeholderText = 'Loading storages';
-  } else if (storages.length === 0) {
+  } else if (selectOptions.length === 0) {
     placeholderText = 'No existing storages available';
   } else {
     placeholderText = 'Select a persistent storage';
@@ -82,7 +82,6 @@ const AddExistingStorageField: React.FC<AddExistingStorageFieldProps> = ({
         selected={data.storage}
         onSelect={(_, storage) => {
           const pvc = storages.find((pvcData) => pvcData.metadata.name === storage);
-
           setData({
             ...data,
             pvc,
--- a/frontend/src/pages/projects/screens/detail/storage/ClusterStorageModal.tsx
+++ b/frontend/src/pages/projects/screens/detail/storage/ClusterStorageModal.tsx
@@ -64,6 +64,7 @@ const ClusterStorageModal: React.FC<ClusterStorageModalProps> = ({ existingPvc, 
     if (hasExistingNotebookConnections) {
       const addData = connectedNotebooks.map((connectedNotebook) => ({
         name: connectedNotebook.metadata.name,
+        notebookDisplayName: connectedNotebook.metadata.annotations?.['openshift.io/display-name'],
         mountPath: {
           value: existingPvc
             ? getNotebookPVCMountPathMap(connectedNotebook)[existingPvc.metadata.name]
--- a/frontend/src/pages/projects/screens/detail/storage/ClusterStorageTableRow.tsx
+++ b/frontend/src/pages/projects/screens/detail/storage/ClusterStorageTableRow.tsx
@@ -97,7 +97,7 @@ const ClusterStorageTableRow: React.FC<ClusterStorageTableRowProps> = ({
           />
         ) : (
           <>
-            {obj.name}{' '}
+            {obj.notebookDisplayName ?? obj.name}{' '}
             <Label isCompact color="green">
               Connected
             </Label>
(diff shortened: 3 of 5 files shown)


frontend/src/pages/projects/screens/spawner/storage/AddExistingStorageField.tsx
```diff
@@ -37,7 +37,7 @@ const AddExistingStorageField: React.FC<AddExistingStorageFieldProps> = ({
         ? storages.reduce((acc: TypeaheadSelectOption[], pvc) => {
             if (!existingStorageNames?.includes(pvc.metadata.name)) {
               acc.push({
-                value: pvc.metadata.name,
```
Gkrumbach07, line 40: this value should still the metadata name. only the content should change

Gkrumbach07 (changes requested):
@DaoDaoNoCode You noticed this, just fyi its being done here

frontend/src/pages/projects/screens/spawner/storage/AddExistingStorageField.tsx
pnaik1: The `value` here is passed to  https://github.com/opendatahub-io/odh-dashboard/blob/main/frontend/src/pages/projects/screens/spawner/storage/AddExistingStorageField.tsx#L89, which further checks this condition https://github.com/opendatahub-io/odh-dashboard/blob/main/frontend/src/pages/projects/screens/detail/storage/utils.ts#L26

so using `getDisplayNameFromK8sResource() will return `display-name` if present or fallback to `metadata.name`
do you want to handle it other way??

Gkrumbach07: well we dont want to use display name for the value becuase it is not a key. it can be the same as other rows

frontend/src/pages/projects/screens/spawner/service.ts
```diff
@@ -50,18 +48,13 @@ export const createPvcDataForNotebook = async (
...
   existingPvc: PersistentVolumeClaimKind | undefined,
-  dryRun?: boolean,
 ): Promise<{ volumes: Volume[]; volumeMounts: VolumeMount[] }> => {
   const volumes: Volume[] = [];
   const volumeMounts: VolumeMount[] = [];
 
   if (existingPvc && storageData.storageType === StorageType.EXISTING_PVC) {
-    if (isPvcUpdateRequired(existingPvc, storageData)) {
-      await updatePvc(storageData, existingPvc, projectName, { dryRun });
-    }
```
pnaik1, line 64: Removed these lines, since current implementation does not change the existingPVC data , its just a dropdown to select

frontend/src/pages/projects/screens/spawner/storage/AddExistingStorageField.tsx
```diff
@@ -35,7 +35,7 @@ const AddExistingStorageField: React.FC<AddExistingStorageFieldProps> = ({
     () =>
       loaded
         ? storages.reduce((acc: TypeaheadSelectOption[], pvc) => {
-            if (!existingStorageNames?.includes(pvc.metadata.name)) {
+            if (!existingStorageNames?.includes(getDisplayNameFromK8sResource(pvc))) {
```
Gkrumbach07, line 38: why are you using display name here now. shouldn't it still be metadata name if existingStorageNames did not change

frontend/src/pages/projects/screens/spawner/storage/ClusterStorageTable.tsx
```diff
@@ -97,7 +98,13 @@ export const ClusterStorageTable: React.FC<ClusterStorageTableProps> = ({
                 spaceItems={{ default: 'spaceItemsSm' }}
               >
                 <FlexItem>
-                  <Truncate content={row.name} />
+                  <Truncate
+                    content={
+                      row.existingPvc?.metadata.name === row.name
```
Gkrumbach07, line 103: what is this check for? couldnt you just do 

```tsx
row.existingPvc ? getDisplayNameFromK8sResource(row.existingPvc) : row.name

File: frontend/src/pages/projects/screens/spawner/storage/AddExistingStorageField.tsx (hunk 3 above)
Comment (pnaik1) [reply frontend/src/pages/projects/screens/spawner/storage/AddExistingStorageField.tsx:38]: when `display-name` and `metadata.name` are different incase of this example,existingStorageNames stores `display-name` not the `metadata.name`  so we need this condition here
```
`kind: PersistentVolumeClaim
apiVersion: v1
metadata:
  name: pvc-test-name
  namespace: dsp-wb-test
  labels:
  annotations:
    openshift.io/description: ''
    openshift.io/display-name: pvc-test-display-name
  finalizers:
    - kubernetes.io/pvc-protection
spec:
  accessModes:
    - ReadWriteOnce
  resources:
    requests:
      storage: 10Gi
  storageClassName: standard-csi
  volumeMode: Filesystem
status:
  phase: Pending 
  `
```

frontend/src/pages/projects/screens/spawner/storage/ClusterStorageTable.tsx
pnaik1: This condition is no longer needed I have reverted back to original one
(based on current implementation

frontend/src/pages/projects/screens/spawner/storage/AddExistingStorageField.tsx
pnaik1: https://github.com/opendatahub-io/odh-dashboard/pull/3584/files#diff-7de192649fe95b9006a1d1bc7f58c89a28e4a91ff6c6de83cdff80c0ce7203ddL104-R106

Gkrumbach07: image
image
This should be the display name. the dropdown shows the display name fine but then once selected, the workbench name is used

Gkrumbach07 (approved):
/lgtm

PR #3586, opened 2024-12-17
hopeful fix for flakey mr test

gitdallas:
## Description
thinking a possible reason for the failure (which i've never been able to reproduce locally) is that the dropdown menu gets clicked before it closes from the first selection.

originally tried a fix is to check for it to be closed before attempting to open it - but it was getting some "dom changed" error

this PR is now separating the two "unsupported platform" tests into their own so it doesn't have to deal with whatever issue is happening when flicking between two

## Request review criteria:


mturley (approved):
LGTM - hard to verify this really fixes the flake, but it's a harmless change if it doesn't and it seems logical that it would.

PR #3590, opened 2024-12-18
Prefill URI connections and fix prefill deploy model form from version 

manaswinidas:
Closes: [RHOAIENG-16146](https://issues.redhat.com/browse/RHOAIENG-16146)
Closes: [RHOAIENG-14136](https://issues.redhat.com/browse/RHOAIENG-14136)

## Description

Prefilling URI connections from `Register model` page(the dropdown should only contain connections filtered according to type):

https://github.com/user-attachments/assets/e87bcf7c-9991-4ce3-a49d-df5ffc538978

A few scenarios for prefilling connections:
1. If you have a registered model version registered with object storage and you have a matching connection, `Existing Connection` will be selected and prefilled with the matching connection and path. The connection selection field will only populate with the connection when there is only 1 recommended connection.(You can test this by deploying the version inside `md-test-s3` model on the `mdas` project - it has one matching connection and that should be pre-selected and you should see `Recommended` label beside it when you click on the dropdown to see more connections on the [psi-08](https://odh-dashboard-opendatahub.apps.ods-qe-psi-08.osp.rh-ods.com/) cluster)

https://github.com/user-attachments/assets/64676ced-4974-44e1-bc0c-98718014a8e7

2. If you have more than one recommended(matching) connection, the `Existing Connection` will be selected, but not be pre-filled - you can click on the dropdown and see two connections with `Recommended` labels.

Screenshot 2025-02-12 at 12 39 40 AM

3. If you have a registered model version registered with URI and you have a matching connection, `Existing connection` will be selected and prefilled with connection name. You should have only 1 matching URI connection to see this(You can test this by deploying the version inside `test-mr-md` model on the `mdas` project - it has one matching URI connection on the [psi-08](https://odh-dashboard-opendatahub.apps.ods-qe-psi-08.osp.rh-ods.com/) cluster). If you have multiple matching(`Recommended`) connections, `Existing Connection` is selected with no prefilled connections(Refer to point 2 screenshot)

https://github.com/user-attachments/assets/14aa4ba3-189d-46c2-930e-d4de872c8155

4. In case you deploy a model version on a project that has no matching connections, you should see an alert with `New Connection` pre-selected and prefilled with a few fields including bucket, region, endpoint and path.(Video 1)(You can test this by deploying the version inside `md-test-s3` model on the `mdas-new` project - it has no matching connections - you should see the prefilled connection type and s3/object storage fields with alert on the [psi-08](https://odh-dashboard-opendatahub.apps.ods-qe-psi-08.osp.rh-ods.com/) cluster). Same goes for URI field. (Video 2)(You can test this by deploying the version inside `test-mr-md` model on the `mdas-new` project - it has no matching connections - you should see the prefilled connection type and URI field with alert on the [psi-08](https://odh-dashboard-opendatahub.apps.ods-qe-psi-08.osp.rh-ods.com/) cluster)
Video 1: 

https://github.com/user-attachments/assets/1745ec5c-0d17-4152-9424-e7bedc45b8e9

Video 2:

https://github.com/user-attachments/assets/94a1895c-2dd9-42a9-9acf-191b3e1b9cbc

Unit tests, some Cypress tests and prefilling logic co-written with @ppadti 

## How Has This Been Tested?

Instructions added along with screen recordings in description.

## Test Impact

Added unit tests and fixing existing and adding new Cypress tests

## Request review criteria:

If you have UI changes: 


frontend/src/concepts/modelRegistry/utils.ts
```diff
@@ -32,7 +34,10 @@ export const uriToObjectStorageFields = (uri: string): ObjectStorageFields | nul
     const endpoint = searchParams.get('endpoint');
     const region = searchParams.get('defaultRegion');
     if (endpoint && bucket && path) {
-      return { endpoint, bucket, region: region || undefined, path };
+      return { s3Fields: { endpoint, bucket, region: region || undefined, path }, uri: null };
+    }
+    if (uri.startsWith('https:')) {
+      return { s3Fields: null, uri };
```
mturley, line 40: URI should be the catch-all case since it could use any protocol (such as the upcoming `oci://`). We should flip this logic and have all the S3-specific logic here only happen if the URI starts with `s3:`, else return this URI version.

Initially I was going to say we should try to keep this function entirely only related to S3 and only call it in the S3 case (so as not to propagate this "URI-split-to-separate-fields" logic that @andrewballantyne has stated we should avoid propagating), but that would lead to needing a `isS3URI()` check that we call pretty much everywhere we use a URI in order to determine if we need this function. I think it maybe makes sense to keep this pattern where we pass around this object of "URI and anything we need to know from parsing it", even if in the future we're only adding `isOCI` to that or something.

I do think however that we should factor out the `{ s3Fields: ObjectStorageFields | null; uri: string | null }` into a separate type since we repeat it. Is there ever a place anymore where we use `ObjectStorageFields` separate from this whole object? Maybe instead we can have a `StorageFields` type and the `s3Fields` can be nested in it. In fact, instead of `StorageFields` can we maybe call it `RegisteredModelLocation`? It's essentially an object representation of the location of a registered model (well, the location of an artifact of one of its versions, but `RegisteredModelVersionLocation` is getting long - and I do think we should maybe have "Registered" in the name if we're going to pass it around in model serving code, to distinguish it from how model locations are stored in connections there).

Please feel free to ignore everything here except the first paragraph if it becomes too much trouble for this PR.

frontend/src/pages/modelServing/screens/types.ts
```diff
@@ -91,6 +93,7 @@ export type InferenceServiceStorage = {
   type: InferenceServiceStorageType;
   path: string;
   dataConnection: string;
+  connectionType?: ModelLocationType | string;
```
mturley, line 96: As discussed on the call with @alexcreasy - If possible, we probably want to avoid adding model-registry-specific properties to these model serving types. `isRecommended` / `LabeledConnection` is sort of an exception because it's a generic feature we added to model serving - anything can make a connection recommended.

I know this complicates things though - I want to think through it a little but we could decide to just address it outside this PR. If we do, let's put a comment here and anywhere this is used saying it's temporary.

frontend/src/pages/modelServing/screens/projects/InferenceServiceModal/ConnectionSection.tsx
```diff
@@ -179,12 +204,39 @@ const NewConnectionField: React.FC<NewConnectionFieldProps> = ({
...
+            connectionTypes.find(
+              (t) => getResourceNameFromK8sResource(t) === data.storage.connectionType,
+            ),
+            ['URI'],
+          ),
+        );
+        setConnectionValues(getMRConnectionValues(data.storage.uri));
+      }
+    }
+  }, [data.storage.connectionType, connectionTypes, data.storage.uri, data.storage.awsData]);
```
mturley, line 238: Ok, so recapping part of our discussion with @alexcreasy - this useEffect exists to update state in response to a user action, which should ideally not be done in a useEffect (it should be part of whatever callback was triggered by the user action) to avoid unintended state updates. Let me try to think out loud about what we could maybe do:

In our case, we.... already broke that rule with `usePrefillDeployModalFromModelRegistry`.... for.... reasons. (short version: we'd otherwise have to lift all the form state out of the whole ManageInferenceServiceModal so we could change its defaults for our MR-specific rendering of it... which was untenable due to the already-existing mess of useEffects we didn't want to touch). But if possible we should keep our useEffect crimes encapsulated in that one existing hook.

So, what's stopping us from doing that? There's no way to call `setSelectedConnectionType` and `setConnectionValues` from outside this `ConnectionSection`. The traditional way to address that would be to "lift the state up" - take the `useState`s for these and bring them up into `ManageInferenceServiceModal`. That's.... not really practical here because we'd also have to lift all the other `enabledConnectionTypes` logic from NewConnectionField, etc.

So what's another way we could address this? Well, we could introduce props/logic that is not specific to model registry that allow any consumer to optionally pre-fill default values for `selectedConnectionType` and `connectionValues`. If `NewConnectionField` had props `initialConnectionTypeName` and `initialConnectionValues`, those could be conditionally used within the `React.useState()` call for setting the initial state values there (using the connection type name as the prop because this is where connection types are fetched and we can look up the matching object here). Then we could drill those new props down: `ConnectionSection` could have props like `initialNewConnectionTypeName` and `initialNewConnectionValues` (naming gets weird here maybe), and then those could be passed in where we render `ConnectionSection` in `ManageInferenceServiceModal`. Now we're back to where we have centralized MR-specific logic - we could have `usePrefillDeployModalFromModelRegistry` conditionally return these `initialNewConnectionType` and `initialNewConnectionValues` if you are prefilling from MR.

That's just one idea -- and if that causes too much trouble to resolve in this PR we can come back to it. But I do think we should make sure to not leave this tech debt for long.

mturley (commented):
A couple things I noted from our meeting @manaswinidas @alexcreasy and some resulting thoughts.

manaswinidas:
@yih-wang Can you check the above screenshots/screen recordings?

My doubts:
1. Can you confirm the behavior in point 3 of the PR description - when will we select `Existing URI` and `Existing storage`? Currently I'm selecting `Current URI` when there is one matching URI connection and `Existing connection` when there are multiple matching connections.
2. Is the positioning of the alert okay in this screenshot? Should the alert appear between the radio button and the Connection type?
Screenshot 2025-02-12 at 12 47 46 AM

frontend/src/pages/modelRegistry/screens/RegisterModel/utils.ts
```diff
@@ -100,6 +100,7 @@ export const registerVersion = async (
       author,
       modelFormatName: formData.sourceModelFormat,
       modelFormatVersion: formData.sourceModelFormatVersion,
+      modelLocationType: formData.modelLocationType,
       // TODO fill in the name of the data connection we used to prefill if we used one
       // TODO this should be done as part of https://issues.redhat.com/browse/RHOAIENG-9914
       // storageKey: 'TODO',
```
mturley, line 106: Hey @manaswinidas please don't kill me - this TODO is something we keep forgetting to come back to, we forgot about it when we did the referenced issue. Since you're modifying the behavior for pre-filling the registration form using a connection, can you either implement this or open a followup issue to implement it? The intended behavior is that if the user used the "Autofill from connection" link in the register model / version forms, the name of that connection goes in this `storageKey` value. The only place it is used is: when deploying this model, if there are no matching connections and the user has to create a new one, we use this `storageKey` as the default value for the name of that new connection. This is why nobody has reported this as a bug.... it never worked, and the user experience is fine without it......... in fact we could also just remove this and pretend it never happened.

Here's the code where the artifact's `storageKey` is used for this: https://github.com/manaswinidas/odh-dashboard/blob/prefill-uri-connections/frontend/src/pages/modelRegistry/screens/RegisteredModels/usePrefillDeployModalFromModelRegistry.ts#L54

frontend/src/pages/modelRegistry/screens/RegisterModel/utils.ts
```diff
@@ -100,6 +100,7 @@ export const registerVersion = async (
       author,
       modelFormatName: formData.sourceModelFormat,
       modelFormatVersion: formData.sourceModelFormatVersion,
+      modelLocationType: formData.modelLocationType,
```
mturley, line 103: Hey... so I came by here to leave that last comment (I scrolled past the TODO comments again when looking at this file to answer someone's question) and I noticed this line above --- `modelLocationType` isn't one of the properties of a ModelArtifact in [the swagger docs](https://editor.swagger.io/?url=https://raw.githubusercontent.com/opendatahub-io/model-registry/main/api/openapi/model-registry.yaml). I don't think adding this does anything.

frontend/src/pages/modelRegistry/screens/RegisterModel/utils.ts
yih-wang: > in fact we could also just remove this and pretend it never happened.
Nope, I didn’t see anything 🙃

yih-wang:
> Can you confirm the behavior in point 3 of the PR description - when will we select Existing URI and Existing storage? Currently I'm selecting Current URI when there is one matching URI connection and Existing connection when there are multiple matching connections.

@manaswinidas So regarding the URI case, I don't think we should add the `Current URI` option to the deploy modal, unless I missed any decision. @vconzola @kywalker-rh Can you help confirm if there was any decision about adding a `Current URI` option came up during the model deployment / ModelCar-specific deployment conversations?

IMO it should behave exactly the same way with S3 connection, i.e. 
1. If there's a matching connection for the URI this model contains, we should auto-select `Existing connection` and pre-fill the matching URI connection for the user;
2. If there're multiple matching connections for the URI, we should auto-select `Existing connection` but not pre-fill any connection for the user;
3. If there's not a matching connection, we should switch to `New connection` and prefill the URI address for the user. (work well in 4 - video2)

> Is the positioning of the alert okay in this screenshot? Should the alert appear between the radio button and the Connection type?

The alert position looks good to me.

yih-wang:
Overall the videos look really good, thank you @manaswinidas! 

Got a few nitpicks on some minor details:
- There're several places in the **registration form** > **autofill from connection** modal where we should update `data connection` to `connection` (to align with the new terminology). Also, the URI type cannot contain a bucket so the help text should be adjusted for the second field. image

- This alert will not be accurate when we reuse this deployment modal in the model catalog. It says `Model location info is available in the registry...` Do we want to simply omit the part `in the registry` and fudge where the flow starts, or is it better to conditionally change the alert to `Model location info is available in the catalog...` when user trigger deployment from the MC? @kaedward @mturley Any preference? image

frontend/src/pages/modelServing/screens/projects/kServeModal/ManageKServeModal.tsx
```diff
@@ -134,6 +135,14 @@ const ManageKServeModal: React.FC<ManageKServeModalProps> = ({
...
   });
 
+  const [connections, connectionsLoaded, connectionsLoadError] =
+    usePrefillDeployModalFromModelRegistry(
+      projectContext,
+      createDataInferenceService,
+      setCreateDataInferenceService,
+      registeredModelDeployInfo,
+    );
+
```
emilys314, line 145: Just an observation, but this seems confusing because traditionally `connections` would be of type `Connection` or an array of it. But having it a `LabeledConnection` seems confusing. 

I don't know if there is a better solution though to add extra properties like this though

manaswinidas: I followed what was done in #3074

frontend/src/pages/modelRegistry/screens/RegisterModel/utils.ts
manaswinidas: Removed this.. yes this is not being used anywhere. I maybe added this for testing and forgot to remove. Thanks for catching this!

frontend/src/pages/modelServing/screens/projects/kServeModal/ManageKServeModal.tsx
mturley: It may be worth opening a tech debt issue to see if we want to refactor to a different approach, but yeah the `LabeledConnection` was the existing approach that was lost when connection types were introduced. An alternative could be to maybe return a separate list of `recommendedConnections`, and when we render the connections dropdown check each to see if it's present in there to mark it recommended. But I'm not sure that's much better.

mturley:
@yih-wang 

> This alert will not be accurate when we reuse this deployment modal in the model catalog. It says `Model location info is available in the registry...` Do we want to simply omit the part `in the registry` and fudge where the flow starts, or is it better to conditionally change the alert to `Model location info is available in the catalog...` when user trigger deployment from the MC?

I'll defer to @kaedward 's opinion, but omitting that part and saying `Model location is available but no matching connection in the project` seems fine to me. If that's not acceptable, I don't think it would be too difficult to have this text be conditional, but for this PR I think we can let it stay as-is and implement that conditional text in https://issues.redhat.com/browse/RHOAIENG-19257.

Incidentally though, I think we may want to correct the grammar there - `but no matching connection in the project` should maybe say `but there are no matching connections in the project` or `but there is no matching connection in the project`?

mturley:
@manaswinidas thank you for outlining the test cases and including the videos. That helps a lot to understand the complex changes.

yih-wang:
> Incidentally though, I think we may want to correct the grammar there

@mturley hmmm, so... I checked the [content handoff doc](https://docs.google.com/document/d/1-YvEYVaVJXWKGGTsZTsVyxzKpaHYfSvGNzz6rIrlvDE/edit?tab=t.0#bookmark=id.vvgs4wwylmlk) and found there're several microcopies (in colorful rows) we forgot to refresh (possibly because the deployment pages are not available yet when the content is refreshing). I'm fine with not updating any content before summit and opening a separate ticket for tracking these changes. I'll defer to you...

frontend/src/pages/modelRegistry/screens/RegisterModel/utils.ts
manaswinidas: Followup issue: https://issues.redhat.com/browse/RHOAIENG-19921

manaswinidas:
@yih-wang I've updated the screen recording for point 3 and added a few other UX fixes to ensure uniformity. 
The `Deploy` button is disabled while the project is being selected for deploying the model version(you can see this in video 3).

Screenshot 2025-02-14 at 12 20 42 PM

mturley:
> @mturley hmmm, so... I checked the [content handoff doc](https://docs.google.com/document/d/1-YvEYVaVJXWKGGTsZTsVyxzKpaHYfSvGNzz6rIrlvDE/edit?tab=t.0#bookmark=id.vvgs4wwylmlk) and found there're several microcopies (in colorful rows) we forgot to refresh (possibly because the deployment pages are not available yet when the content is refreshing). I'm fine with not updating any content before summit and opening a separate ticket for tracking these changes. I'll defer to you...

oh no! agh, I wonder if this was me dropping the ball... we should probably open a ticket and not block summit, but it might be nice to try to get some of it done if we have time

frontend/src/pages/modelServing/screens/projects/InferenceServiceModal/ConnectionSection.tsx
manaswinidas: For posterity: tech debt-issue: https://issues.redhat.com/browse/RHOAIENG-19991(also added in the comments)

frontend/src/pages/modelServing/screens/types.ts
manaswinidas: For posterity: tech debt-issue: https://issues.redhat.com/browse/RHOAIENG-19991(also added in the comments)

frontend/src/concepts/modelRegistry/utils.ts
manaswinidas: Added a type in the same file and URI is a catch-all case now. Resolving this comment.

frontend/src/pages/modelRegistry/screens/RegisteredModels/usePrefillDeployModalFromModelRegistry.ts
```diff
@@ -1,35 +1,44 @@
...
 
   React.useEffect(() => {
-    if (registeredModelDeployInfo) {
+    const alert = {
+      type: AlertVariant.info,
+      title: "We've auto-switched to create a new connection and pre-filled the details for you.",
+      message:
+        'Model location info is available in the registry but there are no matching connections in the project. So we automatically switched the option to create a new connection and prefilled the information.',
+    };
+    if (registeredModelDeployInfo?.modelArtifactUri) {
```
alexcreasy, line 39: This section of nested conditionals is quite hard to grok, I think it might be worth seeing if this can be simplified, anyone who needs to enhance or fix bugs in this area in the future could struggle with updating this. 

I'll take a deeper look and see if I can suggest anything.

alexcreasy:
Tested on PSI 08 using local devmode - everything works as I'd expect, there's a few comments on the code to looke at.

frontend/src/pages/modelRegistry/screens/RegisteredModels/usePrefillDeployModalFromModelRegistry.ts
alexcreasy: This could be done after merging if it's going to hold things up, btw.

frontend/src/concepts/modelRegistry/utils.ts
```diff
@@ -20,21 +20,26 @@ export const objectStorageFieldsToUri = (fields: ObjectStorageFields): string |
...
-      .split('/')
-      .filter(Boolean);
-    const path = pathSplit.join('/');
-    const searchParams = new URLSearchParams(urlObj.search);
-    const endpoint = searchParams.get('endpoint');
-    const region = searchParams.get('defaultRegion');
-    if (endpoint && bucket && path) {
-      return { endpoint, bucket, region: region || undefined, path };
+    if (urlObj.toString().startsWith('s3:')) {
+      const [bucket, ...pathSplit] = `${urlObj.hostname}/${urlObj.pathname}`
```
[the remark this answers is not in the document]
mturley: This is just wrapping existing code in a conditional (and it looks like this is code I wrote), but for context: This is related to the comment above, which I could have phrased better:
```
// Some environments include the first token after the protocol (our bucket) in the pathname and some have it as the hostname
```
This refers to the behavior of `new URL()` which is not consistent in all browsers.

Let's say we have the URI `s3://test-bucket/demo-models/flan-t5-small-caikit?endpoint=http%3A%2F%2Fs3.amazonaws.com%2F&defaultRegion=us-east-1`. The "first token after the protocol (our bucket)" here is `test-bucket`. On some browsers we'll get:
```
urlObj.hostname = 'test-bucket'
urlObj.pathname = 'demo-models/flan-t5-small-caikit'
```
and on other browsers we'll get:
```
urlObj.hostname = undefined
urlObj.pathname = 'test-bucket/demo-models/flan-t5-small-caikit'
```
Joining them with a slash and splitting them again essentially turns both cases into the second case, so our result is consistent across browsers. It's stupid, I know.

@manaswinidas we don't need to change the existing code here you wrapped, but I do think we should put that comment line inside the `if` block so it's still next to this weird pathSplit line.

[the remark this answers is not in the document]
mturley: Also yeah I suppose this would also have been achieved with:
```ts
const [bucket, ...pathSplit] = [urlObj.hostname, urlObj.pathname.split('/')].filter(Boolean)
```

frontend/src/pages/modelRegistry/screens/ModelVersionDetails/ModelVersionDetailsView.tsx
```diff
@@ -141,51 +141,51 @@ const ModelVersionDetailsView: React.FC<ModelVersionDetailsViewProps> = ({
         ) : (
           <>
             <DescriptionList>
-              {storageFields && (
+              {storageFields?.s3Fields && (
                 <>
                   <DashboardDescriptionListGroup
                     title="Endpoint"
-                    isEmpty={modelArtifacts.size === 0 || !storageFields.endpoint}
+                    isEmpty={modelArtifacts.size === 0 || !storageFields.s3Fields.endpoint}
```
mturley, line 148: Now that `storageFields.s3Fields` will only be defined if there's an artifact with a URI starting with `s3:`, I think the `modelArtifacts.size === 0` case on all of these `isEmpty` props is unnecessary.

frontend/src/pages/modelRegistry/screens/components/DeployRegisteredModelModal.tsx
```diff
@@ -122,7 +124,7 @@ const DeployRegisteredModelModal: React.FC<DeployRegisteredModelModalProps> = ({
         isOpen
         onClose={() => onClose(false)}
         actions={[
-          <Button key="deploy" variant="primary" onClick={handleSubmit}>
+          <Button key="deploy" variant="primary" onClick={handleSubmit} isDisabled>
```
mturley, line 127: Wait, why is this button always disabled?

Edit: ah, I see - it's because this particular return of the Modal only happens when there's not a valid selected project, otherwise we'll render the `ManageKServeModal` or `ManageInferenceServiceModal`... ok. Can we add a comment about that?

mturley (changes requested):
A couple comments, but I think the only thing that should block merging is the change to TypeaheadSelect since it's a component shared across the dashboard that we want to be extra careful not to break. The other logic here works so let's not address that feedback it until after we merge this 😄

frontend/src/pages/modelRegistry/screens/components/DeployRegisteredModelModal.tsx
manaswinidas: Followed what was in #3074 .. this removed during the PF6 refactors somehow... I just reinstated this. I'll add a comment to remove any suspicions.

frontend/src/concepts/connectionTypes/utils.ts
```diff
@@ -222,6 +224,27 @@ export const getDefaultValues = (
...
+  if (typeof connectionValues !== 'string') {
+    connectionValues.map((connectionValue) => {
+      if (connectionValue.key !== 'Name') {
+        defaults[connectionValue.key] = connectionValue.value;
+      }
+      return defaults;
+    });
+    return defaults;
+  }
+  if (typeof connectionValues === 'string') {
```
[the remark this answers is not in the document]
manaswinidas: Removed the extra if

mturley (approved):
Tested on psi-08, working great, looking great - thank you so much @manaswinidas (and @ppadti and @alexcreasy) I know this one was painful

PR #3592, opened 2024-12-18
Add in proper handling of spaces and quotes in ISVC parsing

Xaenalt:
https://issues.redhat.com/browse/RHOAIENG-16900

## Description
This adds a rudimentary parser that should handle most cases for inferenceservice argument parsing.

Note, I'm not a JS expert, and wrote it with the help of my internal AI, so it's possible there's some degerenate edge case with this, but in testing it worked fine

## How Has This Been Tested?
Code used to test: 
[parseCommandLine.ts.txt](https://github.com/user-attachments/files/18186726/parseCommandLine.ts.txt)
(Note it misbehaves a bit on some of the negative tests, but the normal test cases it works fine on)

## Test Impact
Tested with the above tests, caveats noted

## Request review criteria:
- [ ] Someone with JS expertise looks this over to make sure it's sane

### DIFF
--- a/frontend/src/api/k8s/inferenceServices.ts
+++ b/frontend/src/api/k8s/inferenceServices.ts
@@ -15,7 +15,7 @@ import { ContainerResources } from '~/types';
 import { AcceleratorProfileFormData } from '~/utilities/useAcceleratorProfileFormState';
 import { AcceleratorProfileState } from '~/utilities/useReadAcceleratorState';
 import { getModelServingProjects } from './projects';
-import { assemblePodSpecOptions } from './utils';
+import { assemblePodSpecOptions, parseCommandLine } from './utils';
 
 export const assembleInferenceService = (
   data: CreatingInferenceServiceObject,
@@ -45,6 +45,8 @@ export const assembleInferenceService = (
   const dataConnectionKey = secretKey || dataConnection;
 
   const nonEmptyArgs = servingRuntimeArgs?.filter(Boolean) || [];
+  // Ensure that we properly handle separating args
+  const splitArgs: string[] = nonEmptyArgs.flatMap(parseCommandLine);
   const nonEmptyEnvVars = servingRuntimeEnvVars?.filter((ev) => ev.name) || [];
 
   const updateInferenceService: InferenceServiceKind = inferenceService
@@ -87,7 +89,7 @@ export const assembleInferenceService = (
                       path,
                     },
                   }),
-              args: nonEmptyArgs,
+              args: splitArgs,
               env: nonEmptyEnvVars,
             },
           },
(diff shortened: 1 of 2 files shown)


frontend/src/api/k8s/inferenceServices.ts
```diff
@@ -45,6 +45,13 @@ export const assembleInferenceService = (
   const dataConnectionKey = secretKey || dataConnection;
 
   const nonEmptyArgs = servingRuntimeArgs?.filter(Boolean) || [];
+  // If we have nonEmptyArgs, ensure that we properly handle separating args
+  let splitArgs: string[] = [];
+  if (nonEmptyArgs.length > 0) {
+    for (let i = 0; i < nonEmptyArgs.length; i++) {
+      splitArgs = splitArgs.concat(parseCommandLine(nonEmptyArgs[i]));
+    }
+  }
```
mturley, line 54: We can simplify this using [flatMap](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap).

```ts
const splitArgs = nonEmptyArgs.flatMap(parseCommandLine);
```

frontend/src/api/k8s/utils.ts
```diff
@@ -74,3 +74,24 @@ export const getshmVolume = (sizeLimit?: string): Volume => ({
   name: 'shm',
   emptyDir: { medium: 'Memory', ...(sizeLimit && { sizeLimit }) },
 });
+
+export const parseCommandLine = (input: string): string[] => {
+  const args: string[] = [];
+  const regex = /(?:[^\s"']+|"[^"]*"|'[^']*')+/g;
+  let match: RegExpExecArray | null;
+
+  while ((match = regex.exec(input)) !== null) {
```
mturley, line 83: Looking closer at this - what iterates this loop? does `regex.exec(input)` with the same input multiple times step to the next match? A quick glance at docs says yes - interesting, I haven't used it that way before.

I wonder if it would be simpler to use `regex.match` instead of `regex.exec` to just get back an array of all the matches, and then do a `forEach` or `reduce` on those? I'm nitpicking though, it's probably fine as-is.

Xaenalt: I'm fine with either solution, that's just the one that it gave me

mturley: If it works it works, but I may tinker with an alternative when I have more time to review. Did you still plan on adding / have time to add unit tests here?

mturley (approved):
LGTM. Thanks for helping out with this @Xaenalt !
Tests will be added in the followup PR https://github.com/opendatahub-io/odh-dashboard/pull/3633.

PR #3599, opened 2024-12-19
Fixed the 4 edge case for main page

YuliaKrimerman:
## Description

[RHOAIENG-16483](https://issues.redhat.com/browse/RHOAIENG-16483) 
## How Has This Been Tested?

This has to be tested on a cluster that has the backend components for model registry and serving disabled 
## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/__tests__/cypress/cypress/tests/mocked/home/homeAIFlows.cy.ts
+++ b/frontend/src/__tests__/cypress/cypress/tests/mocked/home/homeAIFlows.cy.ts
@@ -1,3 +1,4 @@
+import { mockDscStatus } from '~/__mocks__/mockDscStatus';
 import { homePage } from '~/__tests__/cypress/cypress/pages/home/home';
 
 describe('Home page AI Flows', () => {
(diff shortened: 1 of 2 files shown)


frontend/src/__tests__/cypress/cypress/tests/mocked/home/homeAIFlows.cy.ts
```diff
@@ -103,4 +103,13 @@ describe('Home page AI Flows', () => {
     homeAISection.getTrainFlowCard().find().click();
     homeAISection.findPipelinesTrainDescriptionText().should('not.exist');
   });
+
+  it('should hide the models card when model registry and model serving platform components are disabled regardless of feature flags', () => {
+    homePage.initHomeIntercepts({
+      disableModelRegistry: true,
+      disableModelServing: true,
+    });
```
mturley, line 111: We want to test the inverse case here - leave these features on, and instead disable the backend components. That's what was failing before. You can remove the `initHomeIntercepts` call here and instead call:
```ts
  cy.interceptOdh(
    'GET /api/dsc/status',
    mockDscStatus({
      installedComponents: {
        kserve: false,
        'model-mesh': false,
        'model-registry-operator': false,
      },
    }),
  );
```

mturley (changes requested):

frontend/src/__tests__/cypress/cypress/tests/mocked/home/homeAIFlows.cy.ts
YuliaKrimerman: Done

mturley (approved):
LGTM

PR #3600, opened 2024-12-19
Disabling the 'Deploy model' button if NIM is disabled

olavtar:
https://issues.redhat.com/browse/NVPE-66
https://issues.redhat.com/browse/NVPE-34

## Description

If NIM is disabled, previously deployed projects are prevented from re-deploying a NIM model by disabling the 'Deploy model' button on various screens.
image
image
image
image
image

## How Has This Been Tested?

Tested locally and verified various scenarios:

- Removed the Account CR and ensured that if it’s a NIM-enabled project, the 'Deploy model' button is disabled.
- Confirmed that for a KServe project, the 'Deploy model' button remains enabled.
- Verified that if the Account CR status is true, the 'Deploy model' button is enabled."

## Test Impact

Cypress tests were modified to accommodate. 

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/pages/modelServing/screens/global/InferenceServiceTableRow.tsx
+++ b/frontend/src/pages/modelServing/screens/global/InferenceServiceTableRow.tsx
@@ -11,6 +11,7 @@ import useIsAreaAvailable from '~/concepts/areas/useIsAreaAvailable';
 import { getDisplayNameFromK8sResource } from '~/concepts/k8s/utils';
 import { byName, ProjectsContext } from '~/concepts/projects/ProjectsContext';
 import { isProjectNIMSupported } from '~/pages/modelServing/screens/projects/nimUtils';
+import useServingPlatformStatuses from '~/pages/modelServing/useServingPlatformStatuses';
 import InferenceServiceEndpoint from './InferenceServiceEndpoint';
 import InferenceServiceProject from './InferenceServiceProject';
 import InferenceServiceStatus from './InferenceServiceStatus';
@@ -38,6 +39,8 @@ const InferenceServiceTableRow: React.FC<InferenceServiceTableRowProps> = ({
   const { projects } = React.useContext(ProjectsContext);
   const project = projects.find(byName(inferenceService.metadata.namespace)) ?? null;
   const isKServeNIMEnabled = project ? isProjectNIMSupported(project) : false;
+  const servingPlatformStatuses = useServingPlatformStatuses();
+  const isNIMAvailable = servingPlatformStatuses.kServeNIM.enabled;
 
   const [modelMetricsEnabled] = useModelMetricsEnabled();
   const kserveMetricsEnabled = useIsAreaAvailable(SupportedArea.K_SERVE_METRICS).status;
@@ -112,6 +115,7 @@ const InferenceServiceTableRow: React.FC<InferenceServiceTableRowProps> = ({
                 onClick: () => {
                   onEditInferenceService(inferenceService);
                 },
+                isDisabled: !isNIMAvailable && isKServeNIMEnabled,
               },
               { isSeparator: true },
               {
--- a/frontend/src/pages/modelServing/screens/global/ServeModelButton.tsx
+++ b/frontend/src/pages/modelServing/screens/global/ServeModelButton.tsx
@@ -68,16 +68,14 @@ const ServeModelButton: React.FC = () => {
   if (!project) {
     return (
       <Tooltip data-testid="deploy-model-tooltip" content="To deploy a model, select a project.">
-        <div>{deployButton}</div>
+        {deployButton}
       </Tooltip>
     );
   }
 
   if (!isNIMAvailable && isKServeNIMEnabled) {
     return (
-      <Tooltip content="NIM is not available. Contact your administrator.">
-        <div>{deployButton}</div>
-      </Tooltip>
+      <Tooltip content="NIM is not available. Contact your administrator.">{deployButton}</Tooltip>
     );
   }
 
--- a/frontend/src/pages/modelServing/screens/global/ServeModelButton.tsx
+++ b/frontend/src/pages/modelServing/screens/global/ServeModelButton.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { Button, Skeleton, Tooltip } from '@patternfly/react-core';
+import { Button, Tooltip } from '@patternfly/react-core';
 import { useParams } from 'react-router-dom';
 import ManageInferenceServiceModal from '~/pages/modelServing/screens/projects/InferenceServiceModal/ManageInferenceServiceModal';
 import { ModelServingContext } from '~/pages/modelServing/ModelServingContext';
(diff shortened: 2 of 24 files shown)


frontend/src/pages/modelServing/screens/projects/useIsNIMAvailable.ts
```diff
@@ -1,21 +1,31 @@
...
+export const useIsNIMAvailable = (): [boolean, boolean, Error | undefined] => {
   const isNIMModelServingAvailable = useIsAreaAvailable(SupportedArea.NIM_MODEL).status;
 
-  useEffect(() => {
-    const checkNIMServingRuntime = async () => {
-      const isNIMRuntimeAvailable = await isNIMServingRuntimeTemplateAvailable(dashboardNamespace);
-      setIsNIMAvailable(isNIMModelServingAvailable && isNIMRuntimeAvailable);
-    };
+  const fetchNIMAvailability = React.useCallback(async () => {
+    const components = await fetchComponents(false);
```
emilys314, line 11: Could you make the `fetchNIMAvailability` return early with `false` if `isNIMModelServingAvailable` is `false`? This way it will avoid the following API calls when we already know nim is not available

olavtar: Updated

frontend/src/pages/modelServing/screens/global/InferenceServiceTableRow.tsx
```diff
@@ -112,6 +115,7 @@ const InferenceServiceTableRow: React.FC<InferenceServiceTableRowProps> = ({
                 onClick: () => {
                   onEditInferenceService(inferenceService);
                 },
+                isDisabled: !isNIMAvailable && isKServeNIMEnabled,
```
emilys314, line 118: I just wanted to double check on disabling the edit button. Do we know if the model is still serving if nim is disabled? I saw the comments in Jira that if disabling nim breaks existing deployments, then we should disable the edit button. But I'm not 100% sure if this behaviour was confirmed or not

frontend/src/pages/modelServing/screens/global/ServeModelButton.tsx
```diff
@@ -58,7 +59,7 @@ const ServeModelButton: React.FC = () => {
           getProjectModelServingPlatform(project, servingPlatformStatuses).platform,
         )
       }
-      isAriaDisabled={!project || !templatesEnabled}
+      isAriaDisabled={!project || (!isNIMAvailable && isKServeNIMEnabled)}
```
emilys314, line 62:
```suggestion
      isAriaDisabled={!project || !templatesEnabled || (!isNIMAvailable && isKServeNIMEnabled)}
```

We should probably preserve the existing behaviour for if there are no serving runtime templates. (If there are none, deployments are not possible)

frontend/src/pages/modelServing/screens/global/ServeModelButton.tsx
```diff
@@ -67,7 +68,15 @@ const ServeModelButton: React.FC = () => {
...
-        {deployButton}
+        <div>{deployButton}</div>
+      </Tooltip>
+    );
+  }
+
+  if (!isNIMAvailable && isKServeNIMEnabled) {
+    return (
+      <Tooltip content="NIM is not available. Contact your administrator.">
+        <div>{deployButton}</div>
```
emilys314, line 79: nitpick: adding a `<div>` around the button is not needed
```suggestion
        {deployButton}
      </Tooltip>
    );
  }

  if (!isNIMAvailable && isKServeNIMEnabled) {
    return (
      <Tooltip content="NIM is not available. Contact your administrator.">
        {deployButton}
```

frontend/src/pages/modelServing/screens/projects/ModelServingPlatformButtonAction.tsx
```diff
@@ -17,35 +19,44 @@ const ModelServingPlatformButtonAction: React.FC<ModelServingPlatformButtonActio
...
+          'NIM is not available. Contact your administrator.'
+        ) : (
+          <Content component="p">{`At least one serving runtime must be enabled to ${
+            isProjectModelMesh ? 'add a model server' : 'deploy a model'
+          }. Contact your administrator.`}</Content>
+        )
       }
     >
-      {actionButton()}
+      <span>{actionButton}</span>
```
emilys314, line 59:
```suggestion
      {actionButton}
```

frontend/src/pages/modelServing/screens/projects/useIsNIMAvailable.ts
```diff
@@ -1,21 +1,33 @@
...
 
-export const useIsNIMAvailable = (): boolean => {
-  const { dashboardNamespace } = useDashboardNamespace();
-  const [isNIMAvailable, setIsNIMAvailable] = useState<boolean>(false);
+export const useIsNIMAvailable = (): [boolean, boolean, Error | undefined] => {
   const isNIMModelServingAvailable = useIsAreaAvailable(SupportedArea.NIM_MODEL).status;
 
-  useEffect(() => {
-    const checkNIMServingRuntime = async () => {
-      const isNIMRuntimeAvailable = await isNIMServingRuntimeTemplateAvailable(dashboardNamespace);
```
emilys314, line 13: It may still be worth it to still check if the NIMServingRuntime is available? since it's required for nim deployments I believe. Thoughts? (This would mostly be handling error edge cases)

frontend/src/pages/modelServing/screens/global/InferenceServiceTableRow.tsx
olavtar: @emilys314 I just confirmed that if NIM is disabled, redeployment is not possible, which is why the Edit button is disabled.

frontend/src/pages/modelServing/screens/global/ServeModelButton.tsx
olavtar: @emilys314 The account.status is our final source of truth because it is only set to true when all five conditions are met. While the template may be enabled, NIM is not considered available if any of the other conditions are not true.

The conditions that must be true are:

APIKeyValidation
ConfigMapUpdate
TemplateUpdate
SecretUpdate
AccountStatus
This ensures that deployments are only possible when the entire setup is in a valid state.

frontend/src/pages/modelServing/screens/projects/useIsNIMAvailable.ts
olavtar: See [above](https://github.com/opendatahub-io/odh-dashboard/pull/3600#discussion_r1902149067)

frontend/src/pages/modelServing/screens/projects/ModelServingPlatformButtonAction.tsx
olavtar: Updated

frontend/src/pages/modelServing/screens/global/ServeModelButton.tsx
olavtar: Updated

frontend/src/pages/modelServing/screens/global/ServeModelButton.tsx
emilys314: Thanks for the info. Although I think if this is not a nim project, the error can still occur.

So NIM takes it into account, but not kserve. so here i think we should keep it. Your right on the other comment though

olavtar: Updated

PR #3610, opened 2025-01-02
Fix Jupyter Tile env var readback

andrewballantyne:
https://issues.redhat.com/browse/RHOAIENG-17122

## Description

When we did a change for the selects in our app to select the first item and disable itself it had a side-effect to Jupyter tile where it triggered a re-selection and emptying of the secret/configmap details. This occurred in all env var groups but it seems due to poor reference mapping, we updated the values with in-memory old data and thus only ended up applying the last set state. 

Fix was to just not allow the selecting of the same dropdown item to reset the data. But I'm fairly certain this dropdown is obsolete in the code and maybe never did anything in the first place (unimplemented logic?). We'll have to swing back and clean this up later or more likely do a refactor of this page to a new design (like Workbenches).

## How Has This Been Tested?

Fire up Jupyter tile, add env vars (at least one) and then start the notebook.
Stop the notebook and refresh the page or "leave and return" to the page.

## Test Impact

Tests added to make sure response data is properly rendered. 

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/__mocks__/mockNotebookK8sResource.ts
+++ b/frontend/src/__mocks__/mockNotebookK8sResource.ts
@@ -3,6 +3,7 @@ import { KnownLabels, NotebookKind } from '~/k8sTypes';
 import { DEFAULT_NOTEBOOK_SIZES } from '~/pages/projects/screens/spawner/const';
 import {
   ContainerResources,
+  EnvironmentVariable,
   TolerationEffect,
   TolerationOperator,
   Volume,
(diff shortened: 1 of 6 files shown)


DaoDaoNoCode (commented):
I agree with this fix since the Jupyter tile will be refactored to workbench implementation in the future.
Tested and it works well now, the env variables will not be cleared.
/lgtm

Screenshot 2025-01-02 at 2 43 07 PM

andrewballantyne:
Self Approving. I'm the only advisor around this week -- next week I'm on PTO. If there is some problem with this PR, please look to fix it next week as a future PR.

PR #3623, opened 2025-01-08
Tests for mr secure db section

manaswinidas:
Tests for [RHOAIENG-15899](https://issues.redhat.com/browse/RHOAIENG-15899)

## Description

Added unit test for `useModelRegistryCertificateNames` hook and Cypress tests for MR secure DB section

## How Has This Been Tested?

Run the unit test file and Cypress tests using the commands from README

## Request review criteria:

Self checklist (all need to be checked):
- [X] The developer has manually tested the changes and verified that the changes work
- [X] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [X] The developer has added tests or explained why testing cannot be added (unit or cypress tests for related changes)

If you have UI changes: 

- [ ] Included any necessary screenshots or gifs if it was a UI change.
- [ ] Included tags to the UX team if it was a UI/UX change.

After the PR is posted & before it merges:
- [ ] The developer has tested their solution on a cluster by using the image produced by the PR to `main`

PR #3618, opened 2025-01-07
API submit logic and enabling feature flag by default

ppadti:
Closes: [RHOAIENG-15900](https://issues.redhat.com/browse/RHOAIENG-15900)

## Description
This PR aims to change the API objects created as part of creating/editing a ModelRegistry in order to include the selected or uploaded CA certificate and enabling feature flag by default.

## How Has This Been Tested?
1. make sure all the 4 options works correctly while creating/ editing model registry
2. Also test the creation and updation of MR works when the "Add CA certificate to secure database connection" is not checked.

## Test Impact
I will have followup PR to include cypress tests.

## Request review criteria:

Self checklist (all need to be checked):
- [x] The developer has manually tested the changes and verified that the changes work
- [x] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [x] The developer has added tests or explained why testing cannot be added (unit or cypress tests for related changes)

If you have UI changes: 

- [ ] Included any necessary screenshots or gifs if it was a UI change.
- [ ] Included tags to the UX team if it was a UI/UX change.

After the PR is posted & before it merges:
- [ ] The developer has tested their solution on a cluster by using the image produced by the PR to `main`

PR #3631, opened 2025-01-10
Add cypress tests to MR secure DB section

ppadti:
followup PR to https://github.com/opendatahub-io/odh-dashboard/pull/3623 and https://github.com/opendatahub-io/odh-dashboard/pull/3618

## Description
This PR aims to add cypress test for MR secureDB section and address few bugs mentioned [here](https://redhat-internal.slack.com/archives/C07C0447EAV/p1736425171390769)

## Test Impact
1. Run the cypress tests
2. Also test these scenarios:
  - When fetching certificate names fail, Create model registry button should be disabled
  - PEM upload file under "Upload new certificate" section is readOnly 
  - when the first radio option is disabled the form will switch to second option. And when both first and second options are disabled the form selects the third option.

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/pages/modelRegistrySettings/utils.ts
+++ b/frontend/src/pages/modelRegistrySettings/utils.ts
@@ -1,5 +1,5 @@
 import { RecursivePartial } from '~/typeHelpers';
-import { ModelRegistryKind } from '~/k8sTypes';
+import { ConfigSecretItem, ModelRegistryKind } from '~/k8sTypes';
 import {
   CA_BUNDLE_CRT,
   ODH_CA_BUNDLE_CRT,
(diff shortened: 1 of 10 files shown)


frontend/src/pages/modelRegistrySettings/utils.ts
```diff
@@ -51,3 +51,20 @@ export const constructRequestBody = (
 
   return mr;
 };
+
+export const isClusterWideCABundleEnabled = (
```
manaswinidas, line 55: that this may require unit tests but the coverage didn't catch this.. so we can maybe ignore.. the cypress tests cover this scenario.

manaswinidas (commented):
/lgtm

The tests for scenarios discussed are covered

mturley (approved):
LGTM, thanks guys!

PR #3633, opened 2025-01-10
Unit tests parsecommandline util

manaswinidas:
Follow up for #3592 

## Description

Add unit tests for parseCommandline util function

## How Has This Been Tested?

Ran the unit test for parseCommandline added in this PR.

## Test Impact

This PR only adds unit tests 

## Request review criteria:

If you have UI changes: 


Xaenalt:
Those tests look good, I think the only test to add in is to check a multi-line input string too, just in case, but I think for all our use cases this should be good for testing

mturley (approved):

mturley:
Following up with that additional test and a microcopy change in https://github.com/opendatahub-io/odh-dashboard/pull/3634.

PR #3646, opened 2025-01-15
Cypress e2e Test - Verify Non Admin Can Serve And Query A Model Using The UI

antowaddle:
https://issues.redhat.com/browse/RHOAIENG-17157

## Description
- Migrate ODS-2552 to Cypress
- The Test creates a Model via the UI as a non-Admin, an OC command (checkInferenceServiceState) then queries that the model loads (guidance from the Model Serving team is to wait up to 5 mins for Model to load). Finally, we query the UI to verify that the Model also shows in a 'Loaded' state
- Note that there are also changes to how we pass in AWS buckets, a few functions have been changed to allow multiple buckets to be passed in that exist in the test-variables file. Amendments have also been made to change tests that used the functions.
- A change has also been made to provisionProjectForPipelines and a new method added for provisioning data for models. Additional, I've also 
added createCleanProject in the place of createProject which seems to be linked to some of the issues in 
https://issues.redhat.com/browse/RHOAIENG-18077 

## How Has This Been Tested?
- Locally against a PSI cluster and ODH-Nightly
image

## Test Impact
- None - this is a test

## How to run?
Go to odh-dashboard/frontend/src/tests/cypress and run the command npx cypress open . This will open the Cypress UI where testSingleModelContributorCreation.cy.ts can be run.

Headless
Go to odh-dashboard/frontend/src/tests/cypress and run the command npx cypress run --spec "cypress/tests/e2e/dataScienceProjects/models/testSingleModelContributorCreation.cy.ts" --browser chrome

or Execute 'npx cypress run --env grepTags=@ODS-2552  --browser=chrome'

## Request review criteria:

If you have UI changes: 


FedeAlonso (changes requested):
I miss the information regarding BUCKET_3 in the test-variables.yml.example

frontend/src/__tests__/cypress/cypress/utils/pipelines.ts
```diff
@@ -16,20 +15,29 @@ import type {
...
+  cy.log(`Provisioning project with bucket key: ${bucketKey}`);
+
+  const bucketConfig = AWS_BUCKETS[bucketKey];
+
   // Provision a Project
-  createOpenShiftProject(projectName);
+  createCleanProject(projectName);
 
-  // Create a pipeline compatible Data Connection
+  // Create a pipeline-compatible Data Connection
```
FedeAlonso, line 31: Maybe we don't neeed the DSPA in this case

antowaddle: I've created a new util specifically for Model Serving, you're right the DSPA was not required and was superfluous.

antowaddle:
> I miss the information regarding BUCKET_3 in the test-variables.yml.example

@FedeAlonso I've added a reference to BUCKET_3 in the latest commit, good spot!

manosnoam (approved):
/lgtm

PR #3656, opened 2025-01-17
Add model registry name label when creating an inferenceservice

ppadti:
Closes: [RHOAIENG-16989](https://issues.redhat.com/browse/RHOAIENG-16989)

## Description
This PR aims to add model registry name label when creating an inference service

## How Has This Been Tested?
1. deploy a model from the model registry UI.
2. test that the deployed model won't show up in other model registry deployments.
3. Also, make sure that the model deployments which does not have label (modelregistry.opendatahub.io/name) will show up in deployments. As we are returning all ISVCs that either have modelregistry.opendatahub.io/name === registryName or do not have the modelregistry.opendatahub.io/name label.

## Test Impact
Added unit test and cypress test.

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/pages/modelRegistry/screens/ModelVersions/ModelVersions.tsx
+++ b/frontend/src/pages/modelRegistry/screens/ModelVersions/ModelVersions.tsx
@@ -29,7 +29,9 @@ const ModelVersions: React.FC<ModelVersionsProps> = ({ tab, ...pageProps }) => {
   const loadError = mvLoadError || rmLoadError;
   const loaded = mvLoaded && rmLoaded;
   const navigate = useNavigate();
-  const inferenceServices = useMakeFetchObject(useInferenceServices(undefined, rmId));
+  const inferenceServices = useMakeFetchObject(
+    useInferenceServices(undefined, rmId, undefined, preferredModelRegistry?.metadata.name),
+  );
 
   useEffect(() => {
     if (rm?.state === ModelState.ARCHIVED) {
--- a/frontend/src/pages/modelRegistry/screens/ModelVersions/ModelVersionsTable.tsx
+++ b/frontend/src/pages/modelRegistry/screens/ModelVersions/ModelVersionsTable.tsx
@@ -23,8 +23,10 @@ const ModelVersionsTable: React.FC<ModelVersionsTableProps> = ({
   isArchiveModel,
   refresh,
 }) => {
-  const { registeredModelId } = useParams();
-  const inferenceServices = useMakeFetchObject(useInferenceServices(undefined, registeredModelId));
+  const { mrName, registeredModelId } = useParams();
+  const inferenceServices = useMakeFetchObject(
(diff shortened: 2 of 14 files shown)


ppadti:
@mturley as we added a new label now, do we need to [update the architecture-decision-records repo](https://github.com/opendatahub-io/architecture-decision-records/blob/main/documentation/components/dashboard/k8sLabelsAndAnnotations.md) to describe this new label?

mturley:
@ppadti good point, yes we do. We can either do that in the scope of this story or open a new task, up to you.

frontend/src/pages/modelRegistry/screens/ModelVersions/ModelVersions.tsx
```diff
@@ -29,7 +29,9 @@ const ModelVersions: React.FC<ModelVersionsProps> = ({ tab, ...pageProps }) => {
   const loadError = mvLoadError || rmLoadError;
   const loaded = mvLoaded && rmLoaded;
   const navigate = useNavigate();
-  const inferenceServices = useMakeFetchObject(useInferenceServices(undefined, rmId));
+  const inferenceServices = useMakeFetchObject(
+    useInferenceServices(undefined, rmId, preferredModelRegistry?.metadata.name),
```
mturley, line 33: @ppadti you're passing the MR name as the 3rd argument here instead of the 4th argument, which I think would make it be interpreted as the version ID. You may need an `undefined` before it so it's in the 4th position.

frontend/src/pages/modelRegistry/screens/ModelVersions/ModelVersionsTable.tsx
```diff
@@ -23,8 +23,10 @@ const ModelVersionsTable: React.FC<ModelVersionsTableProps> = ({
   isArchiveModel,
   refresh,
 }) => {
-  const { registeredModelId } = useParams();
-  const inferenceServices = useMakeFetchObject(useInferenceServices(undefined, registeredModelId));
+  const { mrName, registeredModelId } = useParams();
+  const inferenceServices = useMakeFetchObject(
+    useInferenceServices(undefined, registeredModelId, mrName),
```
mturley, line 28: Same here, I think you need an `undefined` for the version ID so you push `mrName` into the 4th argument position.

mturley (changes requested):
A couple things with how the new arg is passed. I wonder if we should add a couple Cypress tests to make sure these are being called correctly as well? Just the places where the value of `useInferenceServices` gets passed down to ModelVersionRegisteredDeploymentsView via ModelVersionDetailsTabs, so that'd be the pages that render ModelVersionDetails, ArchiveModelVersionDetails and ModelVersionArchiveDetails (boy those 2 names are confusing).

ppadti:
> @ppadti good point, yes we do. We can either do that in the scope of this story or open a new task, up to you.

Yeah, I will create a PR for that.

mturley (approved):
LGTM, thanks @ppadti !

PR #3657, opened 2025-01-17
added warnings and tests for hardware profiles

rsun19:
[RHOAIENG-17841](https://issues.redhat.com/browse/RHOAIENG-17841)

## Description

Invalid inputs will now show warnings in the UI

Screenshot 2025-01-17 at 4 43 53 PM

## How Has This Been Tested?

You can directly insert hardware profiles in the cluster dashboard. Input invalid values as specified in the jira issue and mockups.

## Test Impact

I added unit tests and cypress tests

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/__mocks__/mockHardwareProfile.ts
+++ b/frontend/src/__mocks__/mockHardwareProfile.ts
@@ -1,11 +1,13 @@
 import { HardwareProfileKind } from '~/k8sTypes';
 import {
   Identifier,
+  IdentifierResourceType,
   NodeSelector,
   Toleration,
   TolerationEffect,
   TolerationOperator,
 } from '~/types';
+import { WarningNotification } from '~/concepts/hardwareProfiles/types';
 import { genUID } from './mockUtils';
 
 type MockResourceConfigType = {
@@ -19,11 +21,16 @@ type MockResourceConfigType = {
   nodeSelector?: NodeSelector;
   tolerations?: Toleration[];
   annotations?: Record<string, string>;
+  warning?: WarningNotification;
+  labels?: Record<string, string>;
 };
 
+/*
+The hardware profiles when mocked need to have unique names if they are in a list, or else they won't render properly.
+*/
 export const mockHardwareProfile = ({
   name = 'migrated-gpu',
-  namespace = 'test-project',
+  namespace = 'opendatahub',
   uid = genUID('service'),
   displayName = 'Nvidia GPU',
   identifiers = [
@@ -33,6 +40,7 @@ export const mockHardwareProfile = ({
       minCount: '2Gi',
       maxCount: '5Gi',
       defaultCount: '2Gi',
+      resourceType: IdentifierResourceType.MEMORY,
     },
     {
       displayName: 'CPU',
@@ -40,6 +48,7 @@ export const mockHardwareProfile = ({
       minCount: '1',
       maxCount: '2',
       defaultCount: '1',
+      resourceType: IdentifierResourceType.CPU,
     },
   ],
   description = '',
@@ -55,6 +64,7 @@ export const mockHardwareProfile = ({
     test: 'value',
   },
   annotations,
+  labels,
 }: MockResourceConfigType): HardwareProfileKind => ({
   apiVersion: 'dashboard.opendatahub.io/v1alpha1',
   kind: 'HardwareProfile',
@@ -66,6 +76,7 @@ export const mockHardwareProfile = ({
     resourceVersion: '1309350',
     uid,
     annotations,
+    labels,
   },
   spec: {
     identifiers,
--- a/frontend/src/__tests__/cypress/cypress/pages/hardwareProfile.ts
+++ b/frontend/src/__tests__/cypress/cypress/pages/hardwareProfile.ts
@@ -28,6 +28,16 @@ class HardwareProfileTableToolbar extends Contextual<HTMLElement> {
   }
 }
 
+class HardwareProfileWarningBanner extends Contextual<HTMLElement> {
+  findTitle(title: string) {
+    return this.find().contains(title);
+  }
+
+  findDescription(description: string) {
+    return this.find().contains(description);
+  }
+}
+
 class HardwareProfileRow extends TableRow {
   findDescription() {
     return this.find().findByTestId('table-row-title-description');
@@ -51,6 +61,10 @@ class HardwareProfileRow extends TableRow {
 
   findNodeSelectorTable() {
     return this.findExpandableSection().findByTestId('hardware-profile-node-selectors-table');
+  }
+
+  findWarningIconButton() {
+    return this.find().findByTestId('icon-warning');
   }
 
   findTolerationTable() {
@@ -77,8 +91,16 @@ class HardwareProfile {
     return this.findTable().find('thead').findByRole('button', { name });
   }
 
+  findHardwareProfileBanner() {
+    return new HardwareProfileWarningBanner(() => this.findHardwareProfileDisabledBanner());
+  }
+
   private findTable() {
     return cy.findByTestId('hardware-profile-table');
+  }
+
+  private findHardwareProfileDisabledBanner() {
+    return cy.findByTestId('hardware-profiles-error-alert');
   }
 
   getRow(name: string) {
@@ -103,6 +125,10 @@ class HardwareProfile {
 
   findClearFiltersButton() {
     return cy.findByTestId('clear-filters-button');
+  }
+
+  findRestoreDefaultHardwareProfileButton() {
+    return cy.findByTestId('restore-default-hardware-profile');
   }
 }
 
--- a/frontend/src/concepts/hardwareProfiles/types.ts
+++ b/frontend/src/concepts/hardwareProfiles/types.ts
@@ -1,0 +1,9 @@
+export type WarningNotification = {
+  title: string;
+  message: string;
+};
+
+export enum HardwareProfileWarningType {
+  HARDWARE_PROFILES_MISSING_CPU_MEMORY,
+  OTHER,
+}
--- a/frontend/src/pages/hardwareProfiles/HardwareProfileEnableToggle.tsx
+++ b/frontend/src/pages/hardwareProfiles/HardwareProfileEnableToggle.tsx
@@ -3,18 +3,21 @@ import { Switch } from '@patternfly/react-core';
 import useNotification from '~/utilities/useNotification';
 import { toggleHardwareProfileEnablement } from '~/api';
 import { HardwareProfileKind } from '~/k8sTypes';
+import { HardwareProfileWarningType } from '~/concepts/hardwareProfiles/types';
+import { validateProfileWarning } from './utils';
 
 type HardwareProfileEnableToggleProps = {
   hardwareProfile: HardwareProfileKind;
-  refreshHardwareProfiles: () => void;
 };
 
 const HardwareProfileEnableToggle: React.FC<HardwareProfileEnableToggleProps> = ({
   hardwareProfile,
-  refreshHardwareProfiles,
 }) => {
+  const hardwareProfileWarnings = validateProfileWarning(hardwareProfile);
   const { enabled } = hardwareProfile.spec;
-  const [isEnabled, setEnabled] = React.useState(enabled);
+  const warning = hardwareProfileWarnings.some(
+    (hardwareProfileWarning) => hardwareProfileWarning.type === HardwareProfileWarningType.OTHER,
+  );
   const [isLoading, setLoading] = React.useState(false);
   const notification = useNotification();
 
@@ -25,16 +28,11 @@ const HardwareProfileEnableToggle: React.FC<HardwareProfileEnableToggleProps> = 
       hardwareProfile.metadata.namespace,
       checked,
     )
-      .then(() => {
-        setEnabled(checked);
-        refreshHardwareProfiles();
-      })
       .catch((e) => {
         notification.error(
           `Error ${checked ? 'enable' : 'disable'} the hardware profile`,
           e.message,
         );
-        setEnabled(!checked);
       })
       .finally(() => {
         setLoading(false);
@@ -46,8 +44,8 @@ const HardwareProfileEnableToggle: React.FC<HardwareProfileEnableToggleProps> = 
       aria-label={enabled ? 'enabled' : 'stopped'}
       data-testid="enable-switch"
       id={`${hardwareProfile.metadata.name}-enable-switch`}
-      isChecked={isEnabled}
-      isDisabled={isLoading}
+      isChecked={enabled && !warning}
+      isDisabled={warning || isLoading}
       onChange={(_e, checked) => handleChange(checked)}
     />
   );
--- a/frontend/src/pages/hardwareProfiles/HardwareProfilesTableRow.tsx
+++ b/frontend/src/pages/hardwareProfiles/HardwareProfilesTableRow.tsx
@@ -1,6 +1,11 @@
 import * as React from 'react';
 import {
+  Button,
   Divider,
+  Icon,
+  List,
+  ListItem,
+  Popover,
   Stack,
   StackItem,
   Timestamp,
@@ -9,6 +14,7 @@ import {
 } from '@patternfly/react-core';
 import { ActionsColumn, ExpandableRowContent, Tbody, Td, Tr } from '@patternfly/react-table';
 import { useNavigate } from 'react-router-dom';
+import { ExclamationTriangleIcon } from '@patternfly/react-icons';
 import { relativeTime } from '~/utilities/time';
 import { TableRowTitleDescription } from '~/components/table';
 import HardwareProfileEnableToggle from '~/pages/hardwareProfiles/HardwareProfileEnableToggle';
@@ -16,26 +22,32 @@ import { HardwareProfileKind } from '~/k8sTypes';
 import NodeResourceTable from '~/pages/hardwareProfiles/nodeResource/NodeResourceTable';
 import NodeSelectorTable from '~/pages/hardwareProfiles/nodeSelector/NodeSelectorTable';
 import TolerationTable from '~/pages/hardwareProfiles/toleration/TolerationTable';
-import { isHardwareProfileOOTB } from '~/pages/hardwareProfiles/utils';
-import { HardwareProfileModel } from '~/api';
 import { useKebabAccessAllowed, verbModelAccess } from '~/concepts/userSSAR';
+import {
+  createHardwareProfileWarningTitle,
+  isHardwareProfileOOTB,
+  validateProfileWarning,
+} from '~/pages/hardwareProfiles/utils';
+import { HardwareProfileModel, updateHardwareProfile } from '~/api';
+import { useDashboardNamespace } from '~/redux/selectors';
+import { DEFAULT_HARDWARE_PROFILE_SPEC } from './const';
 
 type HardwareProfilesTableRowProps = {
   rowIndex: number;
   hardwareProfile: HardwareProfileKind;
   handleDelete: (cr: HardwareProfileKind) => void;
-  refreshHardwareProfiles: () => void;
 };
 
 const HardwareProfilesTableRow: React.FC<HardwareProfilesTableRowProps> = ({
   hardwareProfile,
   rowIndex,
   handleDelete,
-  refreshHardwareProfiles,
 }) => {
   const modifiedDate = hardwareProfile.metadata.annotations?.['opendatahub.io/modified-date'];
   const [isExpanded, setExpanded] = React.useState(false);
   const navigate = useNavigate();
+  const { dashboardNamespace } = useDashboardNamespace();
+  const hardwareProfileWarnings = validateProfileWarning(hardwareProfile);
 
   return (
     <Tbody isExpanded={isExpanded}>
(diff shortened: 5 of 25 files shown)


rsun19:
ahhh time to add more tests for the code coverage 🤣

frontend/src/__tests__/cypress/cypress/pages/hardwareProfile.ts
```diff
@@ -28,6 +28,16 @@ class HardwareProfileTableToolbar extends Contextual<HTMLElement> {
   }
 }
 
+class HardwareProfileWarningBanner extends Contextual<HTMLElement> {
+  findTitle() {
+    return this.find().find('.pf-v6-c-alert__title');
+  }
+
+  findDescription() {
+    return this.find().find('.pf-v6-c-alert__description');
```
bobbravo2, line 37:
```suggestion
    return this.findByTestId()
```
should be less brittle than relying on dom/xpath traversal in patternfly, if we can find/add a `data-test-id`

rsun19: Anyone have insight on how we can add test id's here for title and message

`<Alert
    isInline
    variant="warning"
    title={warningMessages.title}
    data-testid="hardware-profiles-error-alert"
  >
    <p>{warningMessages.message}</p>
  </Alert>`

rsun19: Just realized these classes aren't necessary, will remove them soon

frontend/src/utilities/useWatchHardwareProfiles.ts
```diff
@@ -0,0 +1,16 @@
+import { HardwareProfileKind } from '~/k8sTypes';
+import { CustomWatchK8sResult } from '~/types';
+import { groupVersionKind, HardwareProfileModel } from '~/api';
+import useK8sWatchResourceList from './useK8sWatchResourceList';
+
+export const useWatchHardwareProfiles = (
```
DaoDaoNoCode, line 6: If you are planning to use websocket to watch the hardware profiles, maybe also use it when fetching the hardware profile table. Otherwise, when user edits something on the cluster to something invalid, there is an error message on the nav item but nothing on the table because they are using different ways to fetch the hardware profiles.

Screenshot 2025-01-20 at 4 05 38 PM

frontend/src/pages/hardwareProfiles/utils.ts
```diff
@@ -1,4 +1,224 @@
...
+  },
+  {
+    displayName: 'Memory',
+    identifier: 'memory',
+    minCount: '2Gi',
+    maxCount: '8Gi',
+    defaultCount: '4Gi',
+    resourceType: IdentifierResourceType.MEMORY,
+  },
+];
```
DaoDaoNoCode, line 40: Move this into https://github.com/opendatahub-io/odh-dashboard/blob/main/frontend/src/pages/hardwareProfiles/const.ts and also you can replace https://github.com/opendatahub-io/odh-dashboard/blob/main/frontend/src/pages/hardwareProfiles/const.ts#L85-L102 with it at the same time.

frontend/src/pages/hardwareProfiles/utils.ts
```diff
@@ -1,4 +1,224 @@
...
+    isEnabled,
+  };
+};
+
+export const hardwareProfileWarning = (
+  hardwareProfiles: HardwareProfileKind[],
+): HardwareProfileKind[] =>
+  hardwareProfiles.map((profile) => {
+    let warningStatus = false;
+    const { identifiers } = profile.spec;
```
DaoDaoNoCode, line 107: There are so many `identifiers ?? []` below, you can actually do it here
```suggestion
    const identifiers = profile.spec.identifiers ?? [];
```

frontend/src/pages/hardwareProfiles/utils.ts
```diff
@@ -1,4 +1,224 @@
...
+      }
+      warningStatus = true;
+      parentWarningMessage =
+        'Select the restore button to restore the Default profile to its initial state.';
+    } else {
+      for (let identifier of identifiers ?? []) {
+        try {
+          if (identifier.minCount.toString().at(0) === '-') {
+            parentWarningMessage = `Minimum allowed ${
+              identifier.resourceType ?? 'resource label'
```
DaoDaoNoCode, line 130: I saw a lot of `identifier.resourceType ?? 'resource label'` in this file. Are you trying to say `identifier.displayName`? I believe you can replace all the places with `identifier.displayName` no matter what type is it. (And maybe add a <b> tag?)

frontend/src/pages/hardwareProfiles/utils.ts
```diff
@@ -1,4 +1,224 @@
 import { HardwareProfileKind } from '~/k8sTypes';
+import { Identifier, IdentifierResourceType } from '~/types';
+import { determineUnit, splitValueUnit } from '~/utilities/valueUnits';
+import { hasCPUandMemory } from './manage/ManageNodeResourceSection';
+
+export type WarningNotification = {
+  warningStatus: boolean;
```
DaoDaoNoCode, line 7: I don't see a value adding a `warningStatus` prop here. You are setting it to `true` whenever there is a `parentWarningMessage`. Why not use that as the condition?

frontend/src/pages/hardwareProfiles/utils.ts
```diff
@@ -1,4 +1,224 @@
...
+            warning: true,
+          };
+        }
+      }
+    }
+    return {
+      ...profile,
+      spec: {
+        ...profile.spec,
+        warning: {
```
DaoDaoNoCode, line 211: So as I said above, there is no need to use `warningStatus`, you can simply do
```
warning: parentWarningMessage ? {
title: xxx(your conditions)
message: parentWarningMessage
} : undefined
```

frontend/src/pages/hardwareProfiles/utils.ts
```diff
@@ -1,4 +1,224 @@
 import { HardwareProfileKind } from '~/k8sTypes';
+import { Identifier, IdentifierResourceType } from '~/types';
+import { determineUnit, splitValueUnit } from '~/utilities/valueUnits';
+import { hasCPUandMemory } from './manage/ManageNodeResourceSection';
+
+export type WarningNotification = {
+  warningStatus: boolean;
+  title: string;
+  message: string;
+  isDefaultProfile: boolean;
```
DaoDaoNoCode, line 10: This prop `isDefaultProfile` is also not very useful here. It's only used to decide whether to show a button to restore default or not when rendering the popover, you can just do `isHardwareProfileOOTB(profile)` there in the table row. No need to pass it around and make the type complicated. Keep it as simple as possible (P.S. `isDefaultProfile` should not be in `warning`, when structuring a type, we also need to consider the context and see whether we really need it or if it's reasonable enough to be here, just a suggestion)

frontend/src/pages/hardwareProfiles/utils.ts
```diff
@@ -1,4 +1,224 @@
...
+    warning: warningCount > 0 || !isEnabled,
+    title: generateWarningTitle(isEnabled, warningCount === hardwareProfiles.length, warningCount),
+    message: generateWarningMessage(
+      isEnabled,
+      warningCount === hardwareProfiles.length,
+      warningCount,
+    ),
+    isEnabled,
+  };
+};
```
DaoDaoNoCode, line 100: You can do all of these easily. There are `some` and `every` methods in array.
As for the `isEnabled` prop, you don't need to return it because you are not using it anywhere.
```suggestion
  const hasInvalid = hardwareProfiles.some((profile) => profile.spec.warning);
  const hasEnabled = hardwareProfiles.some((profile) => profile.spec.enabled);
  const allInvalid = hardwareProfiles.every((profile) => profile.spec.warning);
  return {
    warning: hasInvalid || !hasEnabled,
    title: generateWarningTitle(hasEnabled, allInvalid),
    message: generateWarningMessage(hasEnabled, allInvalid),
  };
```

frontend/src/pages/hardwareProfiles/utils.ts
```diff
@@ -1,4 +1,224 @@
...
+  if (warningCount > 0) {
+    return 'One or more hardware profiles are invalid';
+  }
+  return '';
+};
+
+export const generateWarningMessage = (
+  isEnabled: boolean,
+  isAllWarnings: boolean,
+  warningCount: number,
```
DaoDaoNoCode, line 62: You don't need warningCount here. There is no way warningCount could be `0` in this situation when showing the alert.

frontend/src/pages/hardwareProfiles/utils.ts
```diff
@@ -1,4 +1,224 @@
...
+    maxCount: '8Gi',
+    defaultCount: '4Gi',
+    resourceType: IdentifierResourceType.MEMORY,
+  },
+];
+
+export const generateWarningTitle = (
+  isEnabled: boolean,
+  isAllWarnings: boolean,
+  warningCount: number,
```
DaoDaoNoCode, line 45: Same here, no need to add warningCount here.
If the first 2 conditions are not met, directly return the third error title.

DaoDaoNoCode (changes requested):
Besides the comments I left, one important note is missing. You need to automatically disable the invalid hardware profiles and disable the toggle to prevent them from re-enable it before resolving the invalid values.

Screenshot 2025-01-20 at 4 53 03 PM

Other than that, it works well, good job! Most of my comments are about code quality because we want to make it reusable and easier to maintain in the future. :) Let me know if you have any questions with any of my comments and I am happy to help and walk you through it.

DaoDaoNoCode:
I will do another round of review after you address the comments I left above and test it overall again. 😄

rsun19:
Made requested changes

rsun19:
@DaoDaoNoCode made the requested changes. However ,for some methods, I had to fall back to `identifier.displayName` because for some reason `resourceType` is not allowed in the YAML for hardware profiles. Have you seen this before? Simply referring to `resourceType` worked a few days ago but now it disappears when trying to parse the hardware profiles

pnaik1:
> I had to fall back to `identifier.displayName` because for some reason `resourceType` is not allowed in the YAML for hardware profiles. Have you seen this before? Simply referring to `resourceType` worked a few days ago but now it disappears when trying to parse the hardware profiles

@rsun19 , Here's s possibility that you might be using old hardwareProfile CRD, Juntao added new field `resourceType` to CRD few days back, can you try applying the latest CRD?
 let me know if you need any help here

DaoDaoNoCode (commented):
LGTM most of the parts. There are still 2 major things.
1. The use of the types - We should try to avoid redundant types and make best use of the available/existing types, so it could be easier to maintain/add features in the future.
2. For "Incomplete" hardware profiles (Hardware profiles that miss CPU/Memory), we should not block them from enabling because that's still valid, just not recommended. If you want we can have a talk when you are online, so we can walk through all the comments together to make it work as expected.

frontend/src/pages/hardwareProfiles/utils.ts
```diff
@@ -1,4 +1,190 @@
...
+          identifier.displayName === IdentifierResourceType.CPU ||
+          identifier.resourceType === IdentifierResourceType.MEMORY ||
+          identifier.displayName === IdentifierResourceType.MEMORY
+        ) {
+          identifier = {
+            ...identifier,
+            warning: true,
+          };
+        }
+      }
```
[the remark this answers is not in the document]
rsun19: Ah ok that makes sense

rsun19:
@DaoDaoNoCode  Made requested changes. I have a few points though:

1. I have not tested yet with the `resourceType` added into the CRD, but I assume it would work the same because of the fallback. If we could sync on Monday to get that sorted out, that would be great.

2. Do you think there should be warning symbols on the sidebar and a banner on the top if there is a missing CPU/Memory in the default profile? I think we are missing guidance on this. If we do, we might need to figure out the warning message in that case. Currently, I just have the warning symbol for the row when this case is hit.

DaoDaoNoCode:
@rsun19 Hi, thanks for the updates. I am not sure what's the best way to show the warning icon in this situation, let's check with @vconzola today and sort it out together.

rsun19:
Made further changes.

Juntao and I talked with Vince, and we clarified 2 points:

1. The warning icon on the Navbar should only be shown when all hardware profiles are invalid
2. We should show warnings and messages for hardware profiles that are missing CPU/Memory

Gkrumbach07 (approved):
/lgtm

frontend/src/__mocks__/mockHardwareProfile.ts
```diff
@@ -77,5 +89,6 @@ export const mockHardwareProfile = ({
     tolerations,
     nodeSelectors,
     description,
+    warning,
```
andrewballantyne, line 92: Does the CRD have a warning field? If so, this feels too closer to controller land and not really what a spec should be managing 🤔

frontend/src/k8sTypes.ts
```diff
@@ -1253,7 +1254,8 @@ export type HardwareProfileKind = K8sResourceCommon & {
     enabled: boolean;
     description?: string;
     tolerations?: Toleration[];
-    identifiers?: Identifier[];
+    identifiers?: (Identifier & { warning?: boolean })[];
+    warning?: WarningNotification;
```
Gkrumbach07, line 1258: we do not want to add this warning as a part of the spec. we should just handle it ourselves and not store it. nothing is controlling this outside the UI so it is unreliable

frontend/src/pages/hardwareProfiles/utils.ts
```diff
@@ -1,4 +1,209 @@
...
+              message: `${parentWarningMessage} ${
+                isDefaultProfile
+                  ? HARDWARE_PROFILES_DEFAULT_WARNING_MESSAGE
+                  : 'Edit the profile to make the profile valid.'
+              }`,
+            }
+          : undefined,
+      },
+    };
+  });
```
Gkrumbach07, line 209: we dont want to be injecting the warnings into the spec.

instead lets create some new object type like `HardwareProfileValidation` type... better yet we can use zod and create a schema to validate this. 

 thwn we could do things like 

```tsx
export const isHardwareProfileValid = (data: HardwareProfileKind): boolean => {
  const schema = createHardwareProfileValidationSchema(data.selectedProfile);
  const result = schema.safeParse(data);
  return result.success;
};
```

frontend/src/types.ts
```diff
@@ -683,3 +683,8 @@ export type IntegrationAppStatus = {
   variablesValidationTimestamp?: string;
   error: string;
 };
+
+export type WarningNotification = {
```
Gkrumbach07, line 687: keep this type within the scope of hardware profiles. so move it under one of those folders in a `types.ts` folder

frontend/src/pages/hardwareProfiles/HardwareProfileEnableToggle.tsx
```diff
@@ -3,18 +3,21 @@ import { Switch } from '@patternfly/react-core';
...
   hardwareProfile,
-  refreshHardwareProfiles,
 }) => {
-  const { enabled } = hardwareProfile.spec;
-  const [isEnabled, setEnabled] = React.useState(enabled);
+  const hardwareProfileWarningMessage = hardwareProfileWarning(hardwareProfile);
+  const enabled =
+    hardwareProfile.spec.enabled &&
+    (typeof hardwareProfileWarningMessage === 'undefined' ||
+      hardwareProfileWarningMessage.message.includes(HARDWARE_PROFILES_MISSING_CPU_MEMORY_MESSAGE));
```
Gkrumbach07, line 20: i would keep is enabled separate from the warning. Just use warning to additionally disable the hardware profile

frontend/src/pages/hardwareProfiles/utils.ts
```diff
@@ -1,4 +1,229 @@
...
+  if (hasInvalid) {
+    return HardwareProfileBannerWarningTitles.SOME_INVALID;
+  }
+  if (allIncomplete) {
+    return HardwareProfileBannerWarningTitles.ALL_INCOMPLETE;
+  }
+  return HardwareProfileBannerWarningTitles.SOME_INCOMPLETE;
+};
+
+export const generateWarningMessage = (
```
Gkrumbach07, line 42: doesnt need to be exported if not used outside this file

frontend/src/pages/hardwareProfiles/utils.ts
```diff
@@ -1,4 +1,229 @@
...
+  ALL_DISABLED = 'All hardware profiles are disabled',
+  SOME_INVALID = 'One or more hardware profiles are invalid',
+  ALL_INCOMPLETE = 'All hardware profiles are incomplete',
+  SOME_INCOMPLETE = 'One or more hardware profiles are incomplete',
+}
 
 export const isHardwareProfileOOTB = (hardwareProfile: HardwareProfileKind): boolean =>
   hardwareProfile.metadata.labels?.['opendatahub.io/ootb'] === 'true';
+
+export const generateWarningTitle = (
```
Gkrumbach07, line 21: doesnt need to be exported unless used outside this file

frontend/src/pages/hardwareProfiles/utils.ts
```diff
@@ -1,4 +1,229 @@
...
+          isDefaultProfile === true ? 'default hardware' : 'hardware'
+        } profile`,
+        message: `${parentWarningMessage} ${
+          isDefaultProfile
+            ? HARDWARE_PROFILES_DEFAULT_WARNING_MESSAGE
+            : 'Edit the profile to make the profile valid.'
+        }`,
+      }
+    : undefined;
+};
```
Gkrumbach07, line 229: this function should basically validate the hardware profile. It looks like you are returning only one error becuase parentWarningMessage keeps getting overwritten.

I think this function should return an array of all the warnings with the profile. so...

```tsx
enum HardwareProfileWarningType {
   HARDWARE_PROFILES_DEFAULT_WARNING,
   HARDWARE_PROFILES_MISSING_CPU_MEMORY,
   OTHER
}
   
type HardwareProfileWarning = {
   title: string,
   message: string
   type: HardwareProfileWarningType
}

export const validateProfileWarning = ( hardwareProfile: HardwareProfileKind): HardwareProfileWarning[] => {
  // same func body but instead add to this array and return that
}

File: frontend/src/pages/hardwareProfiles/utils.ts [hunk 21]
```diff
@@ -1,4 +1,229 @@
...
+    (profile) => typeof hardwareProfileWarning(profile) !== 'undefined',
+  );
+  if (hardwareProfiles.length === 0 || (!hasInvalidIncludingIncompleteProfiles && hasEnabled)) {
+    return undefined;
+  }
+  return {
+    title: generateWarningTitle(hasEnabled, allInvalid, hasInvalid, allIncomplete),
+    message: generateWarningMessage(hasEnabled, allInvalid, hasInvalid, allIncomplete),
+  };
+};
```
Gkrumbach07, line 98: Now this function after updating the other func (hardwareProfileWarning) you can instead look for the specific warnings like
```tsx
const warnings = hardwareProfileWarning(profile);
const invalid === warnings.some(warning => warning.type === HARDWARE_PROFILES_MISSING_CPU_MEMORY)
```

frontend/src/pages/hardwareProfiles/HardwareProfileEnableToggle.tsx
```diff
@@ -3,18 +3,23 @@ import { Switch } from '@patternfly/react-core';
...
-  refreshHardwareProfiles,
 }) => {
+  const hardwareProfileWarnings = validateProfileWarning(hardwareProfile);
   const { enabled } = hardwareProfile.spec;
-  const [isEnabled, setEnabled] = React.useState(enabled);
+  const warning =
+    typeof hardwareProfileWarnings !== 'undefined' &&
+    !hardwareProfileWarnings.type.includes(
+      HardwareProfileWarningType.HARDWARE_PROFILES_MISSING_CPU_MEMORY,
+    );
```
Gkrumbach07, line 22: dont use this .includes here. its better to just do 

```suggestion
    !hardwareProfileWarnings.type === HardwareProfileWarningType.HARDWARE_PROFILES_MISSING_CPU_MEMORY
```

frontend/src/pages/hardwareProfiles/HardwareProfilesTableRow.tsx
```diff
@@ -55,13 +64,63 @@ const HardwareProfilesTableRow: React.FC<HardwareProfilesTableRowProps> = ({
...
+                      ) : (
+                        <List>
+                          {hardwareProfileWarnings.messages.map((message, index) => (
+                            <ListItem key={index}>{message}</ListItem>
+                          ))}
+                        </List>
+                      )}
+                      {hardwareProfileWarnings.type.includes(
+                        HardwareProfileWarningType.HARDWARE_PROFILES_DEFAULT_WARNING,
+                      ) && (
```
Gkrumbach07, line 90: same change here

frontend/src/concepts/hardwareProfiles/types.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export type WarningNotification = {
+  title: string;
+  message: string;
+};
+
+export enum HardwareProfileWarningType {
+  HARDWARE_PROFILES_DEFAULT_WARNING = HARDWARE_PROFILES_DEFAULT_WARNING_MESSAGE,
+  HARDWARE_PROFILES_MISSING_CPU_MEMORY = HARDWARE_PROFILES_MISSING_CPU_MEMORY_MESSAGE,
+}
```
Gkrumbach07, line 14: these are just types so use
```suggestion
export enum HardwareProfileWarningType {
  HARDWARE_PROFILES_DEFAULT_WARNING,
  HARDWARE_PROFILES_MISSING_CPU_MEMORY,
}
```

frontend/src/pages/hardwareProfiles/utils.ts
```diff
@@ -1,4 +1,307 @@
...
+  if (!complete) {
+    warningTypeArray.push(HardwareProfileWarningType.HARDWARE_PROFILES_MISSING_CPU_MEMORY);
+  }
+  if (isDefault) {
+    warningTypeArray.push(HardwareProfileWarningType.HARDWARE_PROFILES_DEFAULT_WARNING);
+  }
+  return warningTypeArray;
+};
+
+export const validateProfileWarning = (
```
Gkrumbach07, line 168: this func still is a bit confusing to me

I think we should follow more of a zod like flow, so

1. You have a schema, which in your case is this function
2. you pass a profile to be validated and it should return out all the errors/warnings
3. where you differ is that we just want a list of warnings
     - each warning will have a message and type at least
4.  this is all this function should do
5. then you use this function in generateWarningForHardwareProfiles which makes sense
6. however it feels like the type is being used as a message/title when it should be used as a actual type here. you never expose the message from validateProfileWarning

that generateWarningForHardwareProfiles only ever uses the type?

Gkrumbach07 (commented):
i wonder if just using zod is the way to go. it feels like we are reinventing the wheel now

frontend/src/pages/hardwareProfiles/HardwareProfileEnableToggle.tsx
rsun19: I used includes because the type properties aren't mutually exclusive

frontend/src/pages/hardwareProfiles/utils.ts
rsun19: > where you differ is that we just want a list of warnings
So are you suggesting we revert back to your previous suggestion of keeping it an array containing title, message, and type? 

>that generateWarningForHardwareProfiles only ever uses the type?
Are you saying that only `generateWarningForHardwareProfiles` should use it? I don't get the question

rsun19:
Sure, we can use Zod

frontend/src/pages/hardwareProfiles/HardwareProfileEnableToggle.tsx
Gkrumbach07: when not just check 
```tsx
hardwareProfileWarnings.some(hwp => hwp.messages.some(msg => msg.type === HardwareProfileWarningType.HARDWARE_PROFILES_MISSING_CPU_MEMORY))
```

frontend/src/pages/hardwareProfiles/HardwareProfileEnableToggle.tsx
```diff
@@ -3,18 +3,23 @@ import { Switch } from '@patternfly/react-core';
...
 
 const HardwareProfileEnableToggle: React.FC<HardwareProfileEnableToggleProps> = ({
   hardwareProfile,
-  refreshHardwareProfiles,
 }) => {
+  const hardwareProfileWarnings = validateProfileWarning(hardwareProfile);
   const { enabled } = hardwareProfile.spec;
-  const [isEnabled, setEnabled] = React.useState(enabled);
+  const warning =
+    typeof hardwareProfileWarnings !== 'undefined' &&
```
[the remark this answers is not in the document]
Gkrumbach07: I was thinking this function returns an array of all warnings. then you just work from there, dont store the title or some list of types

then if you need to, create funcs that take a list of warnings and give a title

frontend/src/pages/hardwareProfiles/utils.ts
Gkrumbach07: i think it should just be a list of messages/types objects. adding the title in there confuses things in my opinion

rsun19:
made requested changes

Gkrumbach07 (approved):
/lgtm
/unhold

PR #3676, opened 2025-01-23
Update the Serving runtime templates to use correct label to identify preinstalled resources

ppadti:
Closes: [RHOAIENG-10686](https://issues.redhat.com/browse/RHOAIENG-10686)

## Description
This PR aims to Update the Serving runtime templates to use `platform.opendatahub.io/part-of` label to identify preinstalled resources.

Screenshot 2025-01-28 at 2 21 51 PM

## How Has This Been Tested?
Make sure that the pre-installed label will display on serving runtime templates with `platform.opendatahub.io/part-of` label. 

## Test Impact
Updated cypress and unit tests.

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/__tests__/cypress/cypress/tests/mocked/customServingRuntimes/customServingRuntimes.cy.ts
+++ b/frontend/src/__tests__/cypress/cypress/tests/mocked/customServingRuntimes/customServingRuntimes.cy.ts
@@ -33,6 +33,18 @@ describe('Custom serving runtimes', () => {
     servingRuntimes.shouldBeSingleModel(true).shouldBeMultiModel(true);
   });
 
+  it('should test pre-installed label', () => {
+    servingRuntimes.getRowById('template-3').shouldHavePreInstalledLabel();
+    servingRuntimes.getRowById('template-3').find().findKebabAction('Edit').should('not.exist');
+    servingRuntimes.getRowById('template-3').find().findKebabAction('Delete').should('not.exist');
+    servingRuntimes.getRowById('template-3').find().findKebabAction('Duplicate').should('exist');
+
+    servingRuntimes.getRowById('template-2').shouldHavePreInstalledLabel(false);
+    servingRuntimes.getRowById('template-2').find().findKebabAction('Delete').should('exist');
+    servingRuntimes.getRowById('template-2').find().findKebabAction('Edit').should('exist');
+    servingRuntimes.getRowById('template-2').find().findKebabAction('Duplicate').should('exist');
+  });
+
   it('should display platform labels in table rows', () => {
     servingRuntimes.getRowById('template-1').shouldBeSingleModel(true);
     servingRuntimes.getRowById('template-2').shouldBeSingleModel(true);
--- a/frontend/src/api/k8s/__tests__/templates.spec.ts
+++ b/frontend/src/api/k8s/__tests__/templates.spec.ts
@@ -59,6 +59,9 @@ const createServingRuntime = (name: string): K8sDSGResource => ({
 describe('assembleServingRuntimeTemplate', () => {
   it('should assemble serving runtime template with templateName', () => {
     const servingRuntimeMock = JSON.stringify(createServingRuntime('template-1'));
+    const servingRuntimeTemplatesMock = mockServingRuntimeTemplateK8sResource({
+      platforms: [ServingRuntimePlatform.MULTI],
+    });
     const result = assembleServingRuntimeTemplate(
       servingRuntimeMock,
       namespace,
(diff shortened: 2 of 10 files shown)


frontend/src/__tests__/cypress/cypress/tests/mocked/customServingRuntimes/customServingRuntimes.cy.ts
```diff
@@ -33,6 +33,10 @@ describe('Custom serving runtimes', () => {
     servingRuntimes.shouldBeSingleModel(true).shouldBeMultiModel(true);
   });
 
+  it('should have pre-installed label for template-3', () => {
+    servingRuntimes.getRowById('template-3').shouldHavePreInstalledLabel();
```
gitdallas, line 37: could we add a negative test as well?

frontend/src/pages/modelServing/customServingRuntimes/CustomServingRuntimeTableRow.tsx
```diff
@@ -27,7 +27,7 @@ const CustomServingRuntimeTableRow: React.FC<CustomServingRuntimeTableRowProps>
 }) => {
   const navigate = useNavigate();
   const servingRuntimeName = getServingRuntimeNameFromTemplate(template);
-  const templateOOTB = isTemplateOOTB(template);
+  const templateOOTB = ownedByDSC(template);
```
gitdallas, line 30: should this be renamed to `ownedByDSC`?

frontend/src/api/k8s/__tests__/templates.spec.ts
```diff
@@ -59,32 +59,35 @@ const createServingRuntime = (name: string): K8sDSGResource => ({
 describe('assembleServingRuntimeTemplate', () => {
   it('should assemble serving runtime template with templateName', () => {
     const servingRuntimeMock = JSON.stringify(createServingRuntime('template-1'));
+    const servingRuntimeTemplatesMock = mockServingRuntimeTemplateK8sResource({
+      platforms: [ServingRuntimePlatform.MULTI],
+    });
+    delete servingRuntimeTemplatesMock.metadata.ownerReferences;
```
gitdallas, line 65: these `delete`s shouldn't be necessary right? having an empty `ownerReference` should return `false` from `ownedByDSC`

gitdallas (changes requested):
couple small nits/questions

frontend/src/api/k8s/__tests__/templates.spec.ts
mturley: I assume this is here because `assembleServingRuntimeTemplate` won't include an ownerReferences array which makes `toStrictEqual` fail below. You could also avoid this by just changing `mockServingRuntimeTemplateK8sResource` so that it doesn't have a default value of `[]` for `ownerReferences` - if they aren't passed, let them be undefined on the object

frontend/src/pages/modelServing/customServingRuntimes/CustomServingRuntimeTableRow.tsx
mturley: I don't feel strongly either way here - the template being owned by the DSC is what considers it out-of-the-box so this is technically correct, but it may be more readable changed to `templateOwnedByDSC`. I think I agree that the function would have been better named `isOwnedByDSC` but it already exists and we'd have to change usages of it in the connection types page - up to you if you want to do that refactor @ppadti

mturley (commented):
LGTM but I agree somewhat with @gitdallas 's comments here

frontend/src/__tests__/cypress/cypress/tests/mocked/customServingRuntimes/customServingRuntimes.cy.ts
mturley: +1

gitdallas (approved):

mturley:
/hold

@ppadti sorry... we should wait and rebase this on top of https://github.com/opendatahub-io/odh-dashboard/pull/3682 since it fixes a critical issue related to this util

frontend/src/pages/modelServing/customServingRuntimes/CustomServingRuntimeTableRow.tsx
```diff
@@ -46,7 +46,7 @@ const CustomServingRuntimeTableRow: React.FC<CustomServingRuntimeTableRowProps>
         <ResourceNameTooltip resource={template}>
           {getServingRuntimeDisplayNameFromTemplate(template)}
         </ResourceNameTooltip>
-        {templateOOTB && <Label>Pre-installed</Label>}
+        {templateOwnedByDSC && <Label data-testid="pre-installed-label">Pre-installed</Label>}
```
Griffin-Sullivan, line 49: Maybe a little out of scope but you could use the `PreInstalledName` const from https://github.com/opendatahub-io/odh-dashboard/blob/main/frontend/src/concepts/k8s/utils.ts#L5

frontend/src/concepts/k8s/utils.ts
```diff
@@ -4,7 +4,7 @@ import { genRandomChars } from '~/utilities/string';
 
 export const PreInstalledName = 'Pre-installed';
 
-export const ownedByDSC = (resource: K8sResourceCommon): boolean =>
+export const isOwnedByDSC = (resource: K8sResourceCommon): boolean =>
```
andrewballantyne, line 7: When you rebase this on top of the fix Griffin did... please rename this method to something like `isOOTB` as it's more important to know that this is "an out of the box" item, and not something owned by the operator/dsc (since it's not what we are after here)

ppadti: yeah sure, Thanks.

frontend/src/api/k8s/__tests__/templates.spec.ts
mturley, line 65: @ppadti this `delete` should no longer be necessary right?

frontend/src/api/k8s/__tests__/templates.spec.ts
```diff
@@ -59,32 +59,35 @@ const createServingRuntime = (name: string): K8sDSGResource => ({
...
+    expect(result).toStrictEqual(servingRuntimeTemplatesMock);
   });
   it('should assemble serving runtime template without templateName', () => {
     genRandomCharsMock.mockReturnValue('123');
     const servingRuntimeMock = JSON.stringify(createServingRuntime('template-123'));
+    const servingRuntimeTemplatesMock = mockServingRuntimeTemplateK8sResource({
+      name: 'template-123',
+      platforms: [ServingRuntimePlatform.MULTI],
+    });
+    delete servingRuntimeTemplatesMock.metadata.ownerReferences;
```
mturley, line 82: Same for this one

mturley (changes requested):

frontend/src/api/k8s/__tests__/templates.spec.ts
ppadti: yeah right, I did not updated the `mockServingRuntimeTemplateK8sResource` to remove `ownerReferences`. I will update the mock as well.

mturley (approved):
LGTM. Tested on ods-qe-psi-16. Thanks @ppadti !

PR #3681, opened 2025-01-23
updated regex and validation to properly catch non-compliant DNS-1035…

rsun19:
… labels

[RHOAIENG-15753](https://issues.redhat.com/browse/RHOAIENG-15753)

## Description

Previously, the dashboard did not catch invalid workbench names that started with numbers and had other characters in them as well. Now, the regex has been updated to catch them. Furthermore, I also reordered the `replace` order, since the trailing dashes were removed before non-compliant characters, which would fail a test like `$-123test`. The solution? We could either make the regex more complicated or simply reorder the `replace`. I chose the latter option.

For reference taken from the JIRA description:
```
 a DNS-1035 label must consist of lower case alphanumeric characters or '-', 
start with an alphabetic character, and end with an alphanumeric character 
(e.g. 'my-name', or 'abc-123', regex used for validation is '[a-z]([-a-z0-9]*[a-z0-9])?')"
```

## How Has This Been Tested?

You can try out some test cases manually. You can also run this command:

`npm run test:unit **/concepts/k8s/__tests__/utils.spec.ts`

## Test Impact

I added tests in `/concepts/k8s/__tests__/utils.spec.ts`

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/concepts/k8s/__tests__/utils.spec.ts
+++ b/frontend/src/concepts/k8s/__tests__/utils.spec.ts
@@ -7,6 +7,7 @@ import {
   translateDisplayNameForK8s,
   translateDisplayNameForK8sAndReport,
 } from '~/concepts/k8s/utils';
+import { K8_NOTEBOOK_RESOURCE_NAME_VALIDATOR } from '~/pages/projects/screens/spawner/const';
 
 describe('getDisplayNameFromK8sResource', () => {
   it('gets the display name when present', () => {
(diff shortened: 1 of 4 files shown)


pnaik1:
@rsun19 this works well for the  given use case
I doubt if we need to change validation for editing `resource name`
It allows to start with numbers  resulting in workbench which doesn't starts

Screenshot 2025-01-24 at 5 06 57 PM

rsun19:
Good catch! I will update this soon

rsun19:
@pnaik1 made requested changes

DaoDaoNoCode (commented):
I am not sure whether this is the best solution to make all the changes in the `K8sNameDescriptionField` component. It's a generic component and shouldn't be used to handle the specific use cases.
If you check https://github.com/opendatahub-io/odh-dashboard/blob/main/frontend/src/concepts/k8s/K8sNameDescriptionField/K8sNameDescriptionField.tsx#L21-L23, we use this hook everywhere when we need to use the name description field component and you can furthermore find https://github.com/opendatahub-io/odh-dashboard/blob/main/frontend/src/concepts/k8s/K8sNameDescriptionField/types.ts#L17, where you can set the regex you want to use for the component.
Thus, the ideal solution would be finding which workbench is using the component - in this case, it's https://github.com/opendatahub-io/odh-dashboard/blob/main/frontend/src/pages/projects/screens/spawner/SpawnerPage.tsx#L75-L79. Simply set the `regexp` there, and that should help with everything. If that doesn't meet your need, then you may need to consider changing the generic component, but still make it generic.
In this way, we can keep the purity of the generic component at maximum and also, make it won't affect other pages that are also using this component so it can reduce the chance of getting bugs and easier to debug in the future if anything happens.

rsun19:
made requested changes

Gkrumbach07 (approved):
/lgtm

PR #3682, opened 2025-01-24
Update ownedByDSC to use correct label for checking what is created by the operator

Griffin-Sullivan:
https://issues.redhat.com/browse/RHOAIENG-18622

## Description
Simply changing to check for the existence of the `platform.opendatahub.io/part-of` label. Also updating mocks to reflect this change. Personally wouldn't mind changing the name of the function since we aren't using the ownership to determine if something is created by the DSC. 

image

## How Has This Been Tested?
Manually tested. Should see the changes from [this PR](https://github.com/opendatahub-io/odh-dashboard/pull/3642) once it's working (correct label for pre-installed and no kebab actions for edit and delete if preinstalled). User created connection types should show the username as the creator and still allow editing and deleting in the kebab. 

## Test Impact
Updated mocks for data connection config maps. Wouldn't be a bad idea to add a line to the mock test to ensure there's no kebab actions for edit and delete but since that wasn't the point of this PR I'll leave it out.

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/concepts/k8s/utils.ts
+++ b/frontend/src/concepts/k8s/utils.ts
@@ -5,8 +5,7 @@ import { genRandomChars } from '~/utilities/string';
 export const PreInstalledName = 'Pre-installed';
 
 export const ownedByDSC = (resource: K8sResourceCommon): boolean =>
-  !!resource.metadata?.ownerReferences?.find((owner) => owner.kind === 'DataScienceCluster');
-
+  !!resource.metadata?.labels?.['platform.opendatahub.io/part-of'];
 export const isK8sDSGResource = (x?: K8sResourceCommon): x is K8sDSGResource =>
   x?.metadata?.name != null;
 export const getDisplayNameFromK8sResource = (resource: K8sDSGResource): string =>
--- a/frontend/src/__mocks__/mockConnectionType.ts
+++ b/frontend/src/__mocks__/mockConnectionType.ts
@@ -38,27 +38,17 @@ export const mockConnectionTypeConfigMapObj = ({
     namespace,
     resourceVersion: '173155965',
     creationTimestamp: '2024-08-29T00:00:00Z',
-    labels: { 'opendatahub.io/dashboard': 'true', 'opendatahub.io/connection-type': 'true' },
+    labels: {
+      'opendatahub.io/dashboard': 'true',
+      'opendatahub.io/connection-type': 'true',
+      ...(preInstalled && { 'platform.opendatahub.io/part-of': 'dashboard' }),
+    },
     annotations: {
       'openshift.io/display-name': displayName,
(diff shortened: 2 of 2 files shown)


frontend/src/concepts/k8s/utils.ts
```diff
@@ -5,8 +5,10 @@ import { genRandomChars } from '~/utilities/string';
 export const PreInstalledName = 'Pre-installed';
 
 export const ownedByDSC = (resource: K8sResourceCommon): boolean =>
-  !!resource.metadata?.ownerReferences?.find((owner) => owner.kind === 'DataScienceCluster');
-
+  !!Object.prototype.hasOwnProperty.call(
+    resource.metadata?.labels,
+    'platform.opendatahub.io/part-of',
+  );
```
andrewballantyne, line 11:
```suggestion
  !!resource.metadata?.labels?.['platform.opendatahub.io/part-of'];
```
I imagine we can just do a straight-forward truthy check -- technically this would not work if the value is `''` but I don't think that's a valid situation.

andrewballantyne:
@Griffin-Sullivan Can we also check the Serving Runtime templates to see if they are working too?

Griffin-Sullivan:
> @Griffin-Sullivan Can we also check the Serving Runtime templates to see if they are working too?

@andrewballantyne those are working fine. They are controlled by a label `opendatahub.io/ootb` so not affected by the bug if I understand correctly. See https://github.com/opendatahub-io/odh-dashboard/blob/main/frontend/src/pages/modelServing/customServingRuntimes/utils.ts#L18

frontend/src/concepts/k8s/utils.ts
```diff
@@ -5,8 +5,7 @@ import { genRandomChars } from '~/utilities/string';
 export const PreInstalledName = 'Pre-installed';
 
 export const ownedByDSC = (resource: K8sResourceCommon): boolean =>
```
zdtsw, line 7: probably rename the varible ownedByDSC?

ppadti: yeah, we are renaming this variable in this [PR](https://github.com/opendatahub-io/odh-dashboard/pull/3676/files#diff-c97e72b656ffecb9811bb9488007faa7ddf624b2999454aa361be902c635def8R7).

andrewballantyne: We'll likely want something different than `owned` when we rebase #3676 after this goes in @ppadti

andrewballantyne (approved):

frontend/src/concepts/k8s/utils.ts
ppadti: yeah sure, I will update it. Thanks

PR #3694, opened 2025-01-28
Use kserve without serverless

emilys314:
Closes https://issues.redhat.com/browse/RHOAIENG-16490

Depends on https://github.com/opendatahub-io/opendatahub-operator/pull/1565

## Description

This extends the defaultDeploymentMode by checking if serverless is available on the cluster. If it's not, we can default kserve to raw. 
- If raw is disabled, it should default to serverless still (even if both end up disabled somehow)
- raw is used if flag is enabled but serverless is set as the default  
- The auth warning for serverless is moved to below the deployment mode selector if it's visible

When serverless is removed but raw is enabled:
image

If autherino is not installed and serverless is selected:
image

If autherino is not installed and raw selected:
image

## How Has This Been Tested?

Overriding the dsc status in chrome networking to add the following:
`components.kserve.serverlessMode`  to `"serverlessMode": "Removed"`

Or use the latest operator code from main (or ask me for a cluster login with that)

## Test Impact

Added a cypress test 

## Request review criteria:

If you have UI changes: 

@yih-wang 


frontend/src/pages/modelServing/screens/projects/NIMServiceModal/NoAuthAlert.tsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import { Alert, AlertActionCloseButton, StackItem } from '@patternfly/react-core';
+
+export const NoAuthAlert: React.FC<{ setAlertVisible: (isVisible: boolean) => void }> = ({
```
christianvogt, line 4: Having an `onClose` callback makes more sense because the boolean param is never set to anything other than `false`.
```suggestion
export const NoAuthAlert: React.FC<{ onClose: () => void }> = ({
```

christianvogt:
@emilys314 do you need to wait on the dependent operator PR to merge first? If so, put this PR on hold until then.

yih-wang:
Looks good, thank you @emilys314!

emilys314:
/unhold

the opendatahub-operator PR got merged and it's working on my cluster

frontend/src/__tests__/cypress/cypress/tests/mocked/modelServing/servingRuntimeList.cy.ts
```diff
@@ -117,10 +123,16 @@ const initIntercepts = ({
   rejectServingRuntime = false,
   rejectConnection = false,
   requiredCapabilities = [],
+  DscComponents = {
+    modelregistry: {
+      registriesNamespace: 'odh-model-registries',
+    },
+  },
```
Griffin-Sullivan, line 130: why was this added?

Griffin-Sullivan (approved):
/lgtm tested on my cluster. Although I see some changes to NIM that I didn't test. Do we need to test that as well?

frontend/src/__tests__/cypress/cypress/tests/mocked/modelServing/servingRuntimeList.cy.ts
emilys314: I think was trying to not override the default value of `mockDscStatus` but if undefined is passed in javascript doesn't overwrite the default so i will remove this, thanks for the catch

Griffin-Sullivan (commented):
/lgtm

PR #3708, opened 2025-01-31
fix/12783-error-for-dup-model-name

gitdallas:
fises: https://issues.redhat.com/browse/RHOAIENG-12783 

## Description
model name field will check against existing names and give an error help text if it matches one, it will also prevent form submission
image

## How Has This Been Tested?
tested locally by using duplicate model name while trying to register a new model

## Test Impact
added a test to check that duplicate model name prevents form submission

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/__tests__/cypress/cypress/tests/mocked/modelRegistry/registerModel.cy.ts
+++ b/frontend/src/__tests__/cypress/cypress/tests/mocked/modelRegistry/registerModel.cy.ts
@@ -23,8 +23,10 @@ import {
   type ModelArtifact,
 } from '~/concepts/modelRegistry/types';
 import { mockModelRegistryService } from '~/__mocks__/mockModelRegistryService';
+import { mockRegisteredModelList } from '~/__mocks__/mockRegisteredModelsList';
 
 const MODEL_REGISTRY_API_VERSION = 'v1alpha3';
+const existingModelName = 'model1';
 
 const initIntercepts = () => {
   cy.interceptOdh(
@@ -62,6 +64,21 @@ const initIntercepts = () => {
       mockModelRegistryService({ name: 'modelregistry-sample-2' }),
     ]),
   );
+
+  cy.interceptOdh(
+    `GET /api/service/modelregistry/:serviceName/api/model_registry/:apiVersion/registered_models`,
+    {
+      path: { serviceName: 'modelregistry-sample', apiVersion: MODEL_REGISTRY_API_VERSION },
+    },
+    mockRegisteredModelList({
+      items: [
+        mockRegisteredModel({
+          id: '1',
+          name: existingModelName,
+        }),
+      ],
+    }),
+  ).as('getRegisteredModels');
 
   cy.interceptOdh(
     'POST /api/service/modelregistry/:serviceName/api/model_registry/:apiVersion/registered_models',
@@ -203,6 +220,23 @@ describe('Register model page', () => {
       .findFormField(FormFieldSelector.LOCATION_PATH)
       .type('demo-models/flan-t5-small-caikit');
     registerModelPage.findSubmitButton().should('be.enabled');
+  });
+
+  it('Disables submit if model name is duplicated', () => {
+    registerModelPage.findSubmitButton().should('be.disabled');
+    registerModelPage.findFormField(FormFieldSelector.MODEL_NAME).type('Test model name');
+    registerModelPage.findFormField(FormFieldSelector.VERSION_NAME).type('Test version name');
+    registerModelPage.findFormField(FormFieldSelector.LOCATION_TYPE_OBJECT_STORAGE).click();
+    registerModelPage
+      .findFormField(FormFieldSelector.LOCATION_ENDPOINT)
+      .type('http://s3.amazonaws.com/');
+    registerModelPage.findFormField(FormFieldSelector.LOCATION_BUCKET).type('test-bucket');
+    registerModelPage
+      .findFormField(FormFieldSelector.LOCATION_PATH)
+      .type('demo-models/flan-t5-small-caikit');
+    registerModelPage.findSubmitButton().should('be.enabled');
+    registerModelPage.findFormField(FormFieldSelector.MODEL_NAME).clear().type(existingModelName);
+    registerModelPage.findSubmitButton().should('be.disabled');
   });
 
   it('Creates expected resources on submit in object storage mode', () => {
--- a/frontend/src/__tests__/cypress/cypress/tests/mocked/modelRegistry/registerModel.cy.ts
+++ b/frontend/src/__tests__/cypress/cypress/tests/mocked/modelRegistry/registerModel.cy.ts
@@ -237,6 +237,7 @@ describe('Register model page', () => {
     registerModelPage.findSubmitButton().should('be.enabled');
     registerModelPage.findFormField(FormFieldSelector.MODEL_NAME).clear().type(existingModelName);
     registerModelPage.findSubmitButton().should('be.disabled');
+    registerModelPage.findModelNameError().contains('Model name already exists');
   });
 
   it('Creates expected resources on submit in object storage mode', () => {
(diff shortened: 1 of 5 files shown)


frontend/src/pages/modelRegistry/screens/RegisterModel/RegisterModel.tsx
```diff
@@ -36,12 +37,21 @@ const RegisterModel: React.FC = () => {
...
+  const [registeredModels] = useRegisteredModels();
+
+  const doesModelNameExist = (name: string) =>
+    registeredModels.items.some((model) => model.name === name);
+
+  const isModelNameValid =
+    isNameValid(formData.modelName) && !doesModelNameExist(formData.modelName);
+  const isSubmitDisabled =
+    isSubmitting ||
+    isRegisterModelSubmitDisabled(formData) ||
```
mturley, line 52: Hmm. We're slightly violating the abstraction of `isRegisterModelSubmitDisabled` here, ideally it should encapsulate the `doesModelNameExist` logic too. But we would have to either pass the `registeredModels` array into it or pass the `doesModelNameExist` callback into it.

Can we do the former? Maybe move `doesModelNameExist` to `frontend/src/pages/modelRegistry/screens/RegisterModel/utils.ts`, have it take the `registeredModels` as another arg, and call it from inside `isRegisterModelSubmitDisabled` and in the other places where we need it here.

I'm also wondering if we need to do this same validation for the version name... but maybe we can save that for a separate bug if it comes up.

frontend/src/pages/modelRegistry/screens/RegisterModel/RegisterModel.tsx
```diff
@@ -36,12 +37,21 @@ const RegisterModel: React.FC = () => {
   const [isSubmitting, setIsSubmitting] = React.useState(false);
   const [submitError, setSubmitError] = React.useState<Error | undefined>(undefined);
   const [formData, setData] = useRegisterModelData();
-  const isModelNameValid = isNameValid(formData.modelName);
-  const isSubmitDisabled = isSubmitting || isRegisterModelSubmitDisabled(formData);
-  const { modelName, modelDescription } = formData;
   const [registeredModelName, setRegisteredModelName] = React.useState<string>('');
   const [versionName, setVersionName] = React.useState<string>('');
   const [errorName, setErrorName] = React.useState<string | undefined>(undefined);
+  const [registeredModels] = useRegisteredModels();
```
mturley, line 43: I don't love this because technically we won't correctly validate the name until this fetch resolves, so to be fully correct we'd need to pull out the `isLoaded` from this fetch result and also block submit if it's not loaded yet. But this is probably fine because even if the user submits the form before this resolves (i.e. they type faster than their network, or there's an error fetching the registered models), they'll get the existing error message back, so this is still an improvement and it's very unlikely to be an issue.

Just thinking out loud.... I think it's fine to leave here, but we should be careful fetching without looking at the loading state in general.

mturley (changes requested):
Small nit about the abstraction

frontend/src/pages/modelRegistry/screens/RegisterModel/RegisterModel.tsx
gitdallas: we'd want to do more than block until loaded, as (if it were ever not loaded by the time the user went to submit) they would not know why we are blocking the submit. there would have to be some sort of indicator, and i'm not sure what that would look like.

frontend/src/pages/modelRegistry/screens/RegisterModel/RegisterModel.tsx
mturley: We could replace the `isNameValid` function with `validateName` and have it return an error message or undefined if no error, then we can just use its truthiness for whether to render the error and its value for the error?

But we don't necessarily have to do that to put `doesModelNameExist` logic inside `isRegisterModelSubmitDisabled`, we already call `isNameValid` both inside and outside of that function. I may be nitpicking too much based on the name of the function, it is only called in one place - I think the intention was just to make the form-level validation unit testable which we didn't do (yet)

frontend/src/pages/modelRegistry/screens/RegisterModel/RegisterModel.tsx
mturley: that's true. not worth implementing that indicator if they're never going to see it. We actually might be better served by showing a loading spinner on the page until the models are loaded - in fact that's easy here, we can just take `loaded` and `loadError` out of this hook call and pass them to `ApplicationsPage`, it already abstracts that rendering for us. That's probably what we should do.

frontend/src/pages/modelRegistry/screens/RegisterModel/RegisterModel.tsx
```diff
@@ -36,12 +42,17 @@ const RegisterModel: React.FC = () => {
...
-  const isModelNameValid = isNameValid(formData.modelName);
-  const isSubmitDisabled = isSubmitting || isRegisterModelSubmitDisabled(formData);
-  const { modelName, modelDescription } = formData;
   const [registeredModelName, setRegisteredModelName] = React.useState<string>('');
   const [versionName, setVersionName] = React.useState<string>('');
   const [errorName, setErrorName] = React.useState<string | undefined>(undefined);
+  const [registeredModels, registeredModelsLoaded, registeredModelsLoadError] =
+    useRegisteredModels();
+
+  const isModelNameValid = isNameValid(formData.modelName);
```
mturley, line 51: It looks like with the refactor, the submit disabled logic is working but the new error never renders. I think it's because `isModelNameValid` is what we're blocking rendering the error on and now you've split the duplicate check out of it.

Maybe we need a new `const shouldShowError = !isModelNameValid || isModelNameDuplicate` and use that for the `validated` prop and the conditional for the error below.

mturley (changes requested):

frontend/src/__tests__/cypress/cypress/tests/mocked/modelRegistry/registerModel.cy.ts
```diff
@@ -205,6 +222,23 @@ describe('Register model page', () => {
...
+      .type('http://s3.amazonaws.com/');
+    registerModelPage.findFormField(FormFieldSelector.LOCATION_BUCKET).type('test-bucket');
+    registerModelPage
+      .findFormField(FormFieldSelector.LOCATION_PATH)
+      .type('demo-models/flan-t5-small-caikit');
+    registerModelPage.findSubmitButton().should('be.enabled');
+    registerModelPage.findFormField(FormFieldSelector.MODEL_NAME).clear().type(existingModelName);
+    registerModelPage.findSubmitButton().should('be.disabled');
+  });
+
```
mturley, line 241: We should probably also have this test check that the error is rendered

frontend/src/pages/modelRegistry/screens/RegisterModel/utils.ts
```diff
@@ -140,12 +141,19 @@ const isSubmitDisabledForCommonFields = (formData: RegistrationCommonFormData):
...
-  !isNameValid(formData.modelName);
+  !isNameValid(formData.modelName) ||
+  isModelNameExisting(formData.modelName, registeredModels);
 
 export const isRegisterVersionSubmitDisabled = (formData: RegisterVersionFormData): boolean =>
   !formData.registeredModelId || isSubmitDisabledForCommonFields(formData);
 
 export const isNameValid = (name: string): boolean => name.length <= MR_CHARACTER_LIMIT;
+
+export const isModelNameExisting = (name: string, registeredModels: RegisteredModelList): boolean =>
```
manaswinidas, line 158: small nit: I think we should have a unit test for this function. Ideally we should have one for `isNameValid` too - but since we don't.. that's out of scope of this PR.

gitdallas: make mentioned this being in a tech debt list and that we'll end up refactoring form validation logic as part of a larger dashboard UX consistency effort

mturley: That may be a bad excuse to prevent writing a test though. I think @manaswinidas is right - it'd probably be a good start to create a unit tests file for this utils file even if we just put this one small test in there for now.

mturley: I'm skeptical we'll get to that form validation refactor any time soon, we should probably not use that as a reason to not write tests. Sorry for changing my position on that @gitdallas ... @manaswinidas keeping us to high standards here 😄

mturley (approved):
LGTM @gitdallas , thank youuuuu for your patience with nitpicks

PR #3717, opened 2025-02-04
disable "Disabled" NIM Link for Non-Admins with Tooltip

LinoyBitan1:
## Description

JIRA - [https://issues.redhat.com/browse/NVPE-136](https://issues.redhat.com/browse/NVPE-136)

This PR ensures that the "Disabled" NIM link is no longer clickable for non-admin users and provides a tooltip to inform them of the restriction.

**Before:**
Non-admin users could click on the "Disabled" NIM link, leading to confusion.
Before
**After:**
The link is now disabled for non-admin users and a tooltip provides context on why the action is unavailable.
image (13)

## How Has This Been Tested?

Tested locally by logging in as an admin and non-admin user.
Verified that:
- Admin users still have access to the link.
- Non-admin users see the disabled link with a tooltip.

## Test Impact

- Log in as a non-admin user.
- Navigate to the Enable page.
- Ensure that the "Disabled" NIM link is disabled and shows the tooltip when hovered over.

## Request review criteria:

If you have UI changes: 

- [] Included any necessary screenshots or gifs if it was a UI change.

### DIFF
--- a/frontend/src/components/OdhAppCard.tsx
+++ b/frontend/src/components/OdhAppCard.tsx
@@ -14,6 +14,7 @@ import {
   Label,
   MenuToggle,
   Popover,
+  Tooltip,
 } from '@patternfly/react-core';
 import { css } from '@patternfly/react-styles';
 import { EllipsisVIcon, ExclamationCircleIcon, ExternalLinkAltIcon } from '@patternfly/react-icons';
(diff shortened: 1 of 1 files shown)


frontend/src/components/OdhAppCard.tsx
```diff
@@ -204,9 +212,13 @@ const OdhAppCard: React.FC<OdhAppCardProps> = ({ odhApp }) => {
       headerIcon={<ExclamationCircleIcon />}
       alertSeverityVariant="danger"
     >
-      <Button variant="link" className="odh-card__disabled-text">
-        Disabled
-      </Button>
+      {!isAdmin ? (
+        <Tooltip content="To enable this application, contact your administrator.">
```
emilys314, line 216: So currently this is nesting the tooltip inside the popover. Instead this should return either the tooltip or the popover so like
```
!isAdmin ?
<Tooltip>...
: <Popover>...
```

LinoyBitan1: Thanks for the feedback I fixed the issue

PR #3724, opened 2025-02-04
Adds utils for access checks

andrewballantyne:
https://issues.redhat.com/browse/RHOAIENG-18732

## Description

Adds support for SelfSubjectAccessReviews (SSAR) on resources for admins to help with more granular flows/permissions out of the "admin".

Applies this logic to the Hardware Profiles screen.

## How Has This Been Tested?

There is a readme in the code that helps with setup for RBAC values on cluster.

Tested around using cluster-admin and impersonating another user with granted permissions.

## Test Impact

??

## Request review criteria:

If you have UI changes: 


frontend/src/api/useAccessReview.ts
```diff
@@ -29,17 +29,29 @@ const checkAccess = ({
...
+      ])
+      .catch((e) => {
+        // eslint-disable-next-line no-console
+        console.warn('SelfSubjectAccessReview failed', e);
+        return [true, true]; // if it critically fails, don't block SSAR checks; let it fail/succeed on future calls
+      })
+  );
 };
 
 export const useAccessReview = (
```
andrewballantyne, line 50: Do we remove this? Use the cache system for everything? Is this wise?

frontend/src/app/AppRoutes.tsx
```diff
@@ -148,6 +147,7 @@ const AppRoutes: React.FC = () => {
             <Route path="/groupSettings" element={<GroupSettingsPage />} />
           </>
         )}
+        <Route path="/hardwareProfiles/*" element={<HardwareProfileRoutes />} />
```
andrewballantyne, line 150: I'd like to find a way to improve this so that the Route isn't so exposed -- like hide it behind SSAR -- but this is not the place for it.

frontend/src/concepts/userSSAR/useAccessAllowed.ts
```diff
@@ -0,0 +1,35 @@
...
+ * Check if the user has access to the provided resource attributes.
+ * Consider using one of the other options.
+ *   - <AccessAllowed /> for simple rendering blocks
+ *   - useKebabAccessAllowed() for kebab options
+ * If you still want to use this directly, consider using verbModelAccess for resourceAttributes.
+ * @see AccessAllowed
+ * @see useKebabAccessAllowed
+ * @see verbModelAccess
+ */
+export const useAccessAllowed = (
```
andrewballantyne, line 15: This function is going to get mixed up with the `useAccessReview` hook if we leave them as-is... maybe we name all these items in this folder `...CachedAccessAllowed`?

andrewballantyne (commented):
Some WIP thoughts...

frontend/src/api/useAccessReview.ts
Gkrumbach07: maybe we can set it as deprecated in favor of the new cache system. I guess if you wanted to live outside the context of the cache system then this would be used

frontend/src/app/AppRoutes.tsx
Gkrumbach07: Maybe we can follow with another jira

something that wraps HardwareProfileRoute,  ClusterSettingsPage, etc... like how we have a helper for kebabs, we would have some HOC or what have you that does the check and either passes you through, or shows some custom 401 page

Gkrumbach07: nvm looks like you did that already

frontend/src/api/useAccessReview.ts
```diff
@@ -29,17 +29,29 @@ const checkAccess = ({
...
+        result.status ? result.status.allowed : true,
+        true,
+      ])
+      .catch((e) => {
+        // eslint-disable-next-line no-console
+        console.warn('SelfSubjectAccessReview failed', e);
+        return [true, true]; // if it critically fails, don't block SSAR checks; let it fail/succeed on future calls
+      })
+  );
 };
```
Gkrumbach07, line 48: loaded is never not true. is that intended

frontend/src/concepts/userSSAR/AccessReviewContext.tsx
```diff
@@ -0,0 +1,97 @@
+import React from 'react';
+import { checkAccess } from '~/api';
+import { AccessReviewResourceAttributes } from '~/k8sTypes';
+import useNamespaces from '~/pages/notebookController/useNamespaces';
+
+type AccessReviewCacheThingType = {
+  isLoading: boolean;
```
Gkrumbach07, line 7: should this be loaded instead to be the same as the hook and the checkAccess function

frontend/src/concepts/userSSAR/useKebabAccessAllowed.ts
```diff
@@ -0,0 +1,53 @@
+import { IAction } from '@patternfly/react-table';
+import { AccessReviewResourceAttributes } from '~/k8sTypes';
+import { useAccessAllowed } from '~/concepts/userSSAR/useAccessAllowed';
+
+const tooltipArgs = (text: string): Partial<IAction> => ({
+  isAriaDisabled: false,
+  isDisabled: false,
```
Gkrumbach07, line 7:
```suggestion
  isAriaDisabled: true,
```

i mean this is broken in pf6 so this is fine, but it should be this

frontend/src/pages/hardwareProfiles/HardwareProfiles.tsx
```diff
@@ -24,44 +26,60 @@ const HardwareProfiles: React.FC = () => {
...
+            <EmptyStateActions>
+              <Button
+                data-testid="display-hardware-modal-button"
+                variant={ButtonVariant.primary}
+                onClick={() => navigate('/hardwareProfiles/create')}
+              >
+                Add new hardware profile
+              </Button>
+            </EmptyStateActions>
+          </EmptyStateFooter>
```
Gkrumbach07, line 61: can we just wrap the footer with allowedToCreate

Gkrumbach07 (commented):
looks good, just a few things

frontend/src/concepts/userSSAR/AccessReviewContext.tsx
```diff
@@ -0,0 +1,97 @@
+import React from 'react';
+import { checkAccess } from '~/api';
+import { AccessReviewResourceAttributes } from '~/k8sTypes';
+import useNamespaces from '~/pages/notebookController/useNamespaces';
+
+type AccessReviewCacheThingType = {
```
Griffin-Sullivan, line 6: Maybe we could rename this? 
```suggestion
type AccessReviewCacheStatusType = {
```

frontend/src/concepts/userSSAR/AccessReviewContext.tsx
```diff
@@ -0,0 +1,97 @@
...
+        [key]: {
+          isLoading: true,
+          canAccess: false,
+        },
+      }));
+
+      // Determine access
+      return checkAccess({ group, resource, subresource, verb, name, namespace }).then(
+        ([allowed, loaded]) => {
+          setAccessReviewCache((oldValue) => ({
```
Griffin-Sullivan, line 71: Why would this be named `oldValue`?

frontend/src/concepts/userSSAR/useAccessAllowed.ts
Griffin-Sullivan: so would the expectation be to use this hook instead of useAccessReview if you just need to simply check access?

frontend/src/app/AppRoutes.tsx
andrewballantyne: The problem is this page needs to chunk our stuff up -- I think it might need to be something we do on the HardwareProfileRoutes object... I could look into this... that might work.

frontend/src/concepts/userSSAR/useAccessAllowed.ts
andrewballantyne: Yeah, well -- that's the question I'm stuck on. Do we have a "cached check and a not cached check" or do we do everything under the cache -- will that cause problems if you want to check your users access to something live -- instead of having a cached value...

Admins having it, makes sense... but honestly this could extend to _all access checks_ if we can decide if it's a good idea to cache everything.

frontend/src/api/useAccessReview.ts
andrewballantyne: SSAR cannot be assumed to have failed -- if we get it wrong, the UI never works on that cluster (theoretically) -- this is not security... this is conveienence. K8s still validates them on submit.

frontend/src/concepts/userSSAR/AccessReviewContext.tsx
andrewballantyne: I think everything should be `isLoading` and `hasAccess` -- price of using your code and the varying different ways other things have been written. I'll align them all on the `is` syntax as they are booleans.

frontend/src/concepts/userSSAR/useKebabAccessAllowed.ts
andrewballantyne: PF is aware of it https://github.com/patternfly/patternfly-react/issues/11358

frontend/src/api/useAccessReview.ts
```diff
@@ -29,17 +29,29 @@ const checkAccess = ({
...
-    resource: selfSubjectAccessReview,
-  });
+  return (
+    k8sCreateResource<SelfSubjectAccessReviewKind>({
+      model: SelfSubjectAccessReviewModel,
+      resource: selfSubjectAccessReview,
+    })
+      // TODO: TypeScript doesn't realize this can be inferred as this type and thinks it is boolean[]
+      .then((result): [allowed: boolean, loaded: boolean] => [
+        result.status ? result.status.allowed : true,
```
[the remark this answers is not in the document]
andrewballantyne: [Same answer](https://github.com/opendatahub-io/odh-dashboard/pull/3724#discussion_r1943197033)

frontend/src/concepts/userSSAR/AccessReviewContext.tsx
andrewballantyne: Hah, good catch -- didn't see that

frontend/src/concepts/userSSAR/AccessReviewContext.tsx
andrewballantyne: Suppose it could be `currentValue` https://react.dev/reference/react/useState#setstate -- it's the value that is currently in the useState object -- it's the "old or current value" because you're crafting a new value

Griffin-Sullivan: ok gotcha I was confused if it was any different from the part above where you are setting the loading state.

frontend/src/api/useAccessReview.ts
Gkrumbach07: regardless i dont think it can ever be not loaded so whats the point of having that value

frontend/src/concepts/userSSAR/AccessReviewContext.tsx
andrewballantyne: Set before network is made (we are loading, we don't know if we can access, so its false)
Set after network is made (with the values)

frontend/src/concepts/userSSAR/useAccessAllowed.ts
andrewballantyne: Granting permissions or revoking permissions is a wide guess... I'm not sure. 

If I reduced my permissions on the project page to a theoretical `view` state (we don't have this today) -- the UI won't update, even if the calls are remade because of the cache.

frontend/src/concepts/userSSAR/AccessReviewContext.tsx
Gkrumbach07: lol thats my sloppy naming carrying over

Gkrumbach07 (commented):
/lgtm

i will wait for you to resolve griffins comments before we push it through. im fine with adding tests in a later jira

Griffin-Sullivan (approved):
/lgtm 

thanks for getting this done!

Gkrumbach07 (approved):
looks good. needs a rebase

andrewballantyne:
Rebased -- hopefully no linter or testing issues 🙂 

@Gkrumbach07 LGTM again?

PR #3712, opened 2025-02-03
Model Catalog page skeletons and new feature flag `disableModelCatalog`

manaswinidas:
Closes: [RHOAIENG-18953](https://issues.redhat.com/browse/RHOAIENG-18953)

## Description

- Adds feature flag `disableModelCatalog` and routes with placeholder content 
- When the feature flag is enabled, check for the below screen when you navigate to `/modelCatalog`

Screenshot 2025-02-03 at 10 04 25 PM

- When the feature flag is enabled, check for the below screen when you navigate to `/modelCatalog/tempDetails`

Screenshot 2025-02-03 at 10 04 44 PM

- Check for the "not found" screen when the feature flag is disabled.

Screenshot 2025-02-03 at 10 03 32 PM

## How Has This Been Tested?

1. Add the feature flag `disableModelCatalog` to the CRD(You may have to scale down the `odh-dashboard` Deployment)
2. Enable the feature flag in `OdhDashboardConfig` and wait till the Navigation item `Model Catalog` appears - it will appear above `Model Registry`
3. Check for the above screens and routes - both when the feature flag is enabled and when it is not

## Test Impact

Added Cypress tests for the above scenarios

## Request review criteria:

Self checklist (all need to be checked):
- [x] The developer has manually tested the changes and verified that the changes work
- [x] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [x] The developer has added tests or explained why testing cannot be added (unit or cypress tests for related changes)

If you have UI changes: 

- [x] Included any necessary screenshots or gifs if it was a UI change.
- [ ] Included tags to the UX team if it was a UI/UX change.

After the PR is posted & before it merges:
- [ ] The developer has tested their solution on a cluster by using the image produced by the PR to `main`

PR #3718, opened 2025-02-04
Add types, mock data and context provider for model catalog

ppadti:
Closes:  [RHOAIENG-18954](https://issues.redhat.com/browse/RHOAIENG-18954)

## Description
This PR aims to add types, mock data and context provider for model catalog

## Test Impact
NA

## Request review criteria:

Self checklist (all need to be checked):
- [x] The developer has manually tested the changes and verified that the changes work
- [x] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [x] The developer has added tests or explained why testing cannot be added (unit or cypress tests for related changes)

If you have UI changes: 

- [ ] Included any necessary screenshots or gifs if it was a UI change.
- [ ] Included tags to the UX team if it was a UI/UX change.

After the PR is posted & before it merges:
- [ ] The developer has tested their solution on a cluster by using the image produced by the PR to `main`

PR #3730, opened 2025-02-05
Fix child routing for model catalog

manaswinidas:
Follow up #3712 

## Description

Render `ModelDetails` inside the core loader in a sub-route.

https://github.com/user-attachments/assets/19e54db4-8724-4e71-9857-3a1aef403a33

## How Has This Been Tested?

Same as #3712 but you should just see the details content inside the core loader content when you navigate to `/tempDetails`

## Test Impact

Tests added to #3712 already

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/pages/modelCatalog/ModelCatalogCoreLoader.tsx
+++ b/frontend/src/pages/modelCatalog/ModelCatalogCoreLoader.tsx
@@ -1,4 +1,5 @@
 import * as React from 'react';
+import { Outlet } from 'react-router';
 import { conditionalArea, SupportedArea } from '~/concepts/areas';
 import ApplicationsPage from '~/pages/ApplicationsPage';
 import TitleWithIcon from '~/concepts/design/TitleWithIcon';
@@ -23,14 +24,17 @@ const ModelCatalogCoreLoader: React.FC = conditionalArea(
   };
 
   return (
-    <ApplicationsPage
-      title={
-        <TitleWithIcon title="Model Catalog" objectType={ProjectObjectType.registeredModels} />
-      }
-      {...renderStateProps}
-      loaded
-      provideChildrenPadding
-    />
+    <>
+      <ApplicationsPage
+        title={
+          <TitleWithIcon title="Model Catalog" objectType={ProjectObjectType.registeredModels} />
+        }
+        {...renderStateProps}
+        loaded
+        provideChildrenPadding
+      />
+      <Outlet />
+    </>
   );
 });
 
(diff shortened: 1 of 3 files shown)


mturley (changes requested):
I don't think this is quite right - I think we need to replicate what we do with ModelRegistryCoreLoader and ModelRegistryRoutes, where the routes look like this:
```tsx
    <Route
      path={'/:modelRegistry?/*'}
      element={
        <ModelRegistryCoreLoader
          getInvalidRedirectPath={(modelRegistry) => modelRegistryUrl(modelRegistry)}
        />
      }
    >
      <Route index element={<ModelRegistry empty={false} />} />
      <Route path="registeredModels/:registeredModelId">
        <Route index element={<Navigate to={ModelVersionsTab.VERSIONS} replace />} />
        <Route
          path={ModelVersionsTab.VERSIONS}
          element={<ModelVersions tab={ModelVersionsTab.VERSIONS} empty={false} />}
        />
```
the CoreLoader does any logic to determine if it's safe to render the subroutes and, if so, renders an `<Outlet />` (which will later have the context provider around it), and the actual two pages are rendered by the child routes (in this case `<ModelRegistry>` and `<ModelVersions>` end up rendering the `<ApplicationsPage>` for their content). The CoreLoader's `<ApplicationsPage>` is just for its empty state when the prerequisites for even reaching the routes aren't met.

......but actually I don't think we have any dependent data to worry about (like for example ModelRegistryCoreLoader needs to make sure there are model registries before it renders the ModelRegistryContext for a single registry, ModelServingCoreLoader needs to make sure you have a project first, etc).... we just need a component that passes one context provider that doesn't get re-mounted when you switch pages so we're not fetching the catalog sources again on nav.

So I think we maybe just need ModelCatalogCoreLoader to just return the `<Outlet />` - and honestly you could just move the `<ModelCatalogContextProvider>` there around it now, no need to wait since https://github.com/opendatahub-io/odh-dashboard/pull/3718 is already merged - and then we need an index route under it with a new component for the ModelCatalog page. Then both that page and ModelDetailsPage will consume the context and separately check if they need to show the "Request access to model catalog" content, which ModelDetailsPage already has code for. So I don't think we need that in ModelCatalogCoreLoader.

hopefully that makes sense - really all that matters is we have a place to put one context provider that is a parent of both of these pages but not any other pages.

I kinda wonder if we can drop the CoreLoader entirely and just put the context provider in the tree as a child of the route, with the subroutes as its children... I wonder if we do that elsewhere... not that I can find.

manaswinidas:
Fixed in #3733

PR #3731, opened 2025-02-05
Add URI field advanced settings to restrict prefix

emilys314:
https://issues.redhat.com/browse/RHOAIENG-18724

## Description

If a user is making an OCI connection type, kserve requires you to start the uri with `oci://...`. This PR adds a way to restrict the types that a URI can start with (similar to file extensions). Once this is added, we can create an ootb OCI connection type that will take only OCI URIs

Admin connection types editing field
image

User creating a connection on a project
image

## How Has This Been Tested?

Create a new connection type, add a URI field, go to advanced settings and add some scheme types.
Use the connection type when making a connection to verify

## Test Impact

jest tests were added for new functionality (similar to specifiying file extension types)

## Request review criteria:

If you have UI changes: 

@kywalker-rh 

### DIFF
--- a/frontend/src/pages/connectionTypes/manage/advanced/UriAdvancedPropertiesForm.tsx
+++ b/frontend/src/pages/connectionTypes/manage/advanced/UriAdvancedPropertiesForm.tsx
@@ -1,0 +1,105 @@
+import * as React from 'react';
+import { Button, FormGroup } from '@patternfly/react-core';
+import { PlusCircleIcon } from '@patternfly/react-icons';
+import { UriField } from '~/concepts/connectionTypes/types';
+import {
+  URI_SCHEME_REGEX,
+  isDuplicateExtension,
+} from '~/concepts/connectionTypes/fields/fieldUtils';
+import { AdvancedFieldProps } from '~/pages/connectionTypes/manage/advanced/types';
+import ExpandableFormSection from '~/components/ExpandableFormSection';
+import FieldRegexValidationRow from './FieldRegexValidationRow';
+
+const UriAdvancedPropertiesForm: React.FC<AdvancedFieldProps<UriField>> = ({
+  properties,
+  onChange,
+  onValidate,
+}) => {
+  const displayedSchemes = React.useMemo(
+    () => [...(properties.schemes?.length ? properties.schemes : [''])],
+    [properties.schemes],
+  );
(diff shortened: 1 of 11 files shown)


Griffin-Sullivan (approved):
/lgtm 

Haven't tested. Some of the logic in the form is hard to follow, but looks correct to me.

frontend/src/pages/connectionTypes/manage/advanced/UriAdvancedPropertiesForm.tsx
```diff
@@ -0,0 +1,105 @@
...
+              if (!val && displayedSchemes.length === 1) {
+                onChange({ ...properties, schemes: [] });
+              } else {
+                onChange({
+                  ...properties,
+                  schemes: [
+                    ...displayedSchemes.slice(0, index),
+                    val,
+                    ...displayedSchemes.slice(index + 1),
+                  ],
```
Griffin-Sullivan, line 63: You could use `splice` instead
```suggestion
                  schemes: displayedSchemes.splice(index, 0, val)
```

frontend/src/pages/connectionTypes/manage/advanced/UriAdvancedPropertiesForm.tsx
```diff
@@ -0,0 +1,105 @@
...
+                });
+              }
+            }}
+            onRemove={() => {
+              const exts = [
+                ...displayedSchemes.slice(0, index),
+                ...displayedSchemes.slice(index + 1),
```
Griffin-Sullivan, line 70: Same down here

frontend/src/pages/connectionTypes/manage/advanced/UriAdvancedPropertiesForm.tsx
emilys314: So `splice(...)` mutates the original array, which can cause unintended side effects. So that's why  it wasn't originally used i think (I copy pasted a lot of it)
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice

Interestingly it looks like there is a new function called `toSpliced(...)` https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/toSpliced but that seems too new and might not work on older browsers

emilys314:
/hold
hold again because new figma designs again

PR #3733, opened 2025-02-06
Update Model catalog route

ppadti:
Follow up #3712 

## Description
This PR aims to update child routing for model catalog.

https://github.com/user-attachments/assets/e54a11f4-ca75-49e6-b917-7f54853193ed

## How Has This Been Tested?
Check that the details page is rendered when we navigate to `/tempDetails`.  

## Test Impact
tests are already added as part of #3712

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/pages/modelCatalog/ModelCatalogCoreLoader.tsx
+++ b/frontend/src/pages/modelCatalog/ModelCatalogCoreLoader.tsx
@@ -1,26 +1,51 @@
 import * as React from 'react';
+import { Outlet } from 'react-router';
 import { conditionalArea, SupportedArea } from '~/concepts/areas';
 import ApplicationsPage from '~/pages/ApplicationsPage';
 import TitleWithIcon from '~/concepts/design/TitleWithIcon';
 import { ProjectObjectType, typedEmptyImage } from '~/concepts/design/utils';
+import {
+  ModelCatalogContext,
+  ModelCatalogContextProvider,
+} from '~/concepts/modelCatalog/context/ModelCatalogContext';
 import EmptyModelCatalogState from './EmptyModelCatalogState';
 
+type ApplicationPageProps = React.ComponentProps<typeof ApplicationsPage>;
+
+type ApplicationPageRenderState = Pick<
+  ApplicationPageProps,
+  'emptyStatePage' | 'empty' | 'headerContent'
+>;
 const ModelCatalogCoreLoader: React.FC = conditionalArea(
   SupportedArea.MODEL_CATALOG,
   true,
 )(() => {
-  const renderStateProps = {
-    empty: true,
-    emptyStatePage: (
-      <EmptyModelCatalogState
-        testid="empty-model-catalog-state"
-        title="Request access to model catalog"
-        description="To request access to model catalog, contact your administrator."
-        headerIcon={() => <img src={typedEmptyImage(ProjectObjectType.registeredModels)} alt="" />}
-      />
-    ),
-    headerContent: null,
-  };
+  const { modelCatalogSources } = React.useContext(ModelCatalogContext);
+  let renderStateProps: ApplicationPageRenderState & { children?: React.ReactNode };
+  if (modelCatalogSources.length === 0) {
+    renderStateProps = {
+      empty: true,
+      emptyStatePage: (
+        <>
+          <EmptyModelCatalogState
+            testid="empty-model-catalog-state"
+            title="Request access to model catalog"
+            description="To request access to model catalog, contact your administrator."
+            headerIcon={() => (
+              <img src={typedEmptyImage(ProjectObjectType.registeredModels)} alt="" />
+            )}
+          />
+        </>
+      ),
+      headerContent: null,
+    };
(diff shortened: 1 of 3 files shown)


frontend/src/pages/modelCatalog/ModelCatalogRoutes.tsx
```diff
@@ -2,12 +2,15 @@ import * as React from 'react';
...
 import ModelCatalogCoreLoader from './ModelCatalogCoreLoader';
 import ModelDetailsPage from './ModelDetailsPage';
+import ModelCatalog from './screens/ModelCatalog';
 
 const ModelCatalogRoutes: React.FC = () => (
   <Routes>
-    <Route path={'/:modelCatalog?/*'} element={<ModelCatalogCoreLoader />} />
-    <Route path="tempDetails" element={<ModelDetailsPage />} />
-    <Route path="*" element={<Navigate to="." />} />
+    <Route path={'/:modelCatalogSource?/*'} element={<ModelCatalogCoreLoader />}>
```
manaswinidas, line 9: Should this be `/:modelCatalogSource?/*` or just `/:modelCatalog?/*`? Doesn't affect the route behavior tho 🤔

manaswinidas (changes requested):

frontend/src/pages/modelCatalog/ModelCatalogCoreLoader.tsx
```diff
@@ -1,26 +1,51 @@
...
+      ),
+      headerContent: null,
+    };
+  } else {
+    return (
+      <ModelCatalogContextProvider>
+        <Outlet />
+      </ModelCatalogContextProvider>
+    );
+  }
```
mturley, line 48: We can simplify this a little - this pattern came from other CoreLoaders where we had more complicated logic and multiple possible `renderStateProps` values. Here though there's no reason not to just `return <ApplicationsPage ...` right in this `if (modelCatalogSources.length === 0) {` condition. and then you can remove the `else` and just return the provider/outlet after the if block.

frontend/src/pages/modelCatalog/ModelCatalogRoutes.tsx
```diff
@@ -2,12 +2,15 @@ import * as React from 'react';
 import { Navigate, Route, Routes } from 'react-router-dom';
 import ModelCatalogCoreLoader from './ModelCatalogCoreLoader';
 import ModelDetailsPage from './ModelDetailsPage';
```
mturley, line 4: We should probably move the `ModelDetailsPage` into `./screens`

frontend/src/pages/modelCatalog/ModelCatalogRoutes.tsx
mturley: This was based on my feedback on an earlier PR. I guess it doesn't particularly matter, I'm being pedantic... but this route param refers to a specific object that may or may not be present (we're either at the page root or we're looking at something inside a model catalog source, which we could have multiple of later). For Summit this value will always be either missing or `"Red Hat"` (if it's ok to have spaces... if not we'll have to format that).

Example routes we'll have:
```
/modelCatalog (main catalog card view)
/modelCatalog/Red Hat/someModelIdentifierTBD (model details view)
```
I don't think we'll have just a `/modelCatalog/Red Hat` page, unless in the future we decide we want separate pages for each source. I'm making guesses since we only have the one source initially.

For `/:modelRegistry?/*` the route param is the name of a model registry. Here we are using the name of a catalog source. Being pedantic, there's only one catalog with data coming from multiple sources. I.e. we won't have a "catalog selector", we'll have one page combining data from multiple sources.

It is super duper just a nitpick though and I don't feel strongly about it - we can go either way here.

frontend/src/pages/modelCatalog/screens/ModelCatalog.tsx
```diff
@@ -0,0 +1,38 @@
...
+    <ApplicationsPage
+      title={
+        <TitleWithIcon title="Model Catalog" objectType={ProjectObjectType.registeredModels} />
+      }
+      {...renderStateProps}
+      loaded
+      provideChildrenPadding
+    />
+  );
+});
```
mturley, line 36: I think we can also just put these `renderStateProps` inline when we return `<ApplicationsPage`, but maybe we leave this alone so we can unblock @gitdallas for implementing this page in https://issues.redhat.com/browse/RHOAIENG-18956 and he can render this however makes sense for that story.

mturley (approved):
@ppadti this LGTM except some nitpicking, but since this is only part of your story and this change unblocks others let's merge it. Do you mind following up on that feedback in your next PR (if you agree with it)?

frontend/src/pages/modelCatalog/screens/ModelCatalog.tsx
ppadti: yeah, I think Dallas will replace this whole component here with the model catalog page that he will create as part of RHOAIENG-18956. Thats the reason, I just duplicated the content of core loader component here.

PR #3741, opened 2025-02-07
Attach pull secret to InferenceService when present

Griffin-Sullivan:
https://issues.redhat.com/browse/RHOAIENG-18660
## Description
When connection secret is type `dockerconfigjson` then we need to attach it as a `imagePullSecret` to the InferenceService to authenticate to private OCI registries. This PR makes this check and adds the secret to the isvc on submit for the KServe modal. 

## How Has This Been Tested?
Disclaimer: Everything has been tested in kserve raw mode only, although that shouldn't effect this
- Create public and private oci connections (will probably need to manually create these secrets since there is ongoing work to create an oci connection type)
- Deploy models using these connections
- Edit a model from private to public and ensure the `imagePullSecrets` is removed from the isvc.

## Test Impact

- Added a couple unit tests to ensure things are made correctly

## Request review criteria:

If you have UI changes: 


frontend/src/pages/modelServing/screens/projects/utils.ts
```diff
@@ -409,6 +413,10 @@ const createInferenceServiceAndDataConnection = async (
     storageUri = inferenceServiceData.storage.uri;
   }
 
+  if (connection?.type === 'kubernetes.io/dockerconfigjson') {
+    imagePullSecrets = [{ name: connection.metadata.name }];
+  }
```
emilys314, line 418: Creating works good,

(TBD based on later todays meeting) but when editing it will remove the `imagePullSecrets` because `connection` will be undefined since the URI type doesn't have a reference to the connection, so when you edit and save the deployment

emilys314: So yeah editing still needs to be fixed. So to reproduce:
1. deploy model with imagePullSecret
2. See it has the `imagePullSecrets`
3. Edit the deployed model
4. Change something, like the name or routing or replicas
5. See the inferenceService no longer has the  `imagePullSecrets`

This is because of a limitation in the previously implemented code with using `storageUri` for the model inference service. We didn't make any effort to look up the previous connection because it was impossible for URI only, since there was no reference. 

But now we do have a reference of it, inside the  `imagePullSecrets`, so on edit, we should search for the connection specified in the  `imagePullSecrets` and pull that into the modal

emilys314:
It looks like there is a failing CI test https://github.com/opendatahub-io/odh-dashboard/actions/runs/13292937586/job/37118195165?pr=3741#step:4:1115

This might be intermittent, or you can try rebasing to the latest code too

emilys314:
For record keeping, there is currently some bug that makes the inferenceService increment (probably not a UI issue but still worth investigating)

image

emilys314:
I tested creating and editing public and private oci models and everything seems to be working

/lgtm
/approve

PR #3743, opened 2025-02-07
Fetch ConfigMap in ModelCatalogContext provider and move provider to route level

YuliaKrimerman:
Closes [RHOAIENG-18959](https://issues.redhat.com/browse/RHOAIENG-18959)

## Description

Created the fetching of the configmap and verified that it's stringifies it in the way we need for all the models. 

## How Has This Been Tested?

Tested on the green cluster. To test it please add /tempDetails to the model catalog url and checkout the console is printing Model Catalog Context. 
Screenshot 2025-02-11 at 12 51 19 PM
## Test Impact

Tests will be added once the work on the Empty/Loading state is done

## Request review criteria:

If you have UI changes: 


mturley:
@YuliaKrimerman @ppadti  we need to remove the `<ModelCatalogContextProvider>` that is still in frontend/src/app/App.tsx, it probably makes sense for that to be in scope for this PR since data fetching is what makes that render location matter. I missed that when reviewing https://github.com/opendatahub-io/odh-dashboard/pull/3733, we have redundant context providers now (a provider has already been added to ModelCatalogCoreLoader).

@YuliaKrimerman you also have a merge conflict here

ppadti:
> @YuliaKrimerman @ppadti we need to remove the `<ModelCatalogContextProvider>` that is still in frontend/src/app/App.tsx, it probably makes sense for that to be in scope for this PR since data fetching is what makes that render location matter. I missed that when reviewing #3733, we have redundant context providers now (a provider has already been added to ModelCatalogCoreLoader).
> 
> Actually with the changes so far in this PR we now have that provider rendered in 3 places 😅 we can also revert the changes you have here to ModelCatalogRoutes. That'll resolve the merge conflict. The context provider rendered in ModelCatalogCoreLoader is what we want to keep - that will only be rendered when you're on one of these routes - the way this diff has it wrapped around the routes, it'll be rendered no matter which route you're on. Sorry, I could have been more clear in the issue description about that.

Sorry, I missed that...

YuliaKrimerman:
@mturley @ppadti Latest changes : removed the context from App, reverted changes in the ModelCatalogRoutes, added logging to show we're fetching the configmap. Here is a screenshot that we're parsing it correctly as well 
Screenshot 2025-02-10 at 11 07 29 AM

frontend/src/concepts/modelCatalog/context/ModelCatalogContext.tsx
```diff
@@ -1,29 +1,32 @@
...
 export const ModelCatalogContext = React.createContext<ModelCatalogContextType>({
-  modelCatalogSources: [],
+  modelCatalogSources: [[], false, undefined, () => Promise.resolve(undefined)],
 });
 
 export const ModelCatalogContextProvider: React.FC<ModelCatalogContextProviderProps> = ({
   children,
 }) => {
-  const modelCatalogSourcesMock = React.useMemo(() => [modelCatalogSourceMock({})], []);
+  const modelCatalogSources = useModelCatalogSources();
```
mturley, line 21: You can wrap this in `useMakeFetchObject`. This changes the array return value to an object with named properties, easier to deal with.

```ts
const modelCatalogSources = useMakeFetchObject(useModelCatalogSources());
```

frontend/src/concepts/modelCatalog/types.ts
```diff
@@ -39,3 +42,18 @@ export type ModelCatalogSource = {
...
+    name: string;
+    labels: DashboardLabels & {
+      'opendatahub.io/model-catalog': 'true';
+    };
+  };
+  data?: {
+    source: string;
+    models: string; // JSON string of CatalogModel[]
+  };
+};
```
mturley, line 59: You know what? I'm sorry - I think we can just delete this whole type. It's not used anywhere - I was about to suggest how we can use a type guard to have the value of `getConfigMap` use this type below, but the only thing that would do for us is protect against typos in `configMap.data?.modelCatalogSource` which you write on two lines - not worth the effort for a temporary workaround. Let's ditch this type.

frontend/src/pages/modelCatalog/ModelCatalogPage.tsx
```diff
@@ -0,0 +1,16 @@
...
+import { ModelCatalogContext } from '~/concepts/modelCatalog/context/ModelCatalogContext';
+
+const ModelCatalogPage: React.FC = () => {
+  const { modelCatalogSources } = React.useContext(ModelCatalogContext);
+  const [sources, loaded, error] = modelCatalogSources;
+
+  React.useEffect(() => {
+    // eslint-disable-next-line no-console
+    console.log('Loaded:', loaded);
+  }, [sources, loaded, error]);
```
mturley, line 11: You'll have to change how you consume this with the change to FetchStateObject - you'll just have `modelCatalogSources.data` (and `.loaded` and `.error`).

frontend/src/concepts/modelCatalog/useModelCatalogSources.ts
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    const configMap = await getConfigMap(dashboardNamespace, MODEL_CATALOG_SOURCE_CONFIGMAP);
+
+    if (!configMap.data?.modelCatalogSource) {
+      return [];
+    }
+
+    const source: ModelCatalogSource = JSON.parse(configMap.data.modelCatalogSource);
+    return [source];
```
mturley, line 30: JSON.parse can throw an exception, and we don't want invalid JSON in the configmap to crash the UI. You'll need to wrap these 2 lines in a `try { } catch(e) { }`, and in the catch block just return `[]`. I think it's ok to swallow errors like this for our temporary summit implementation.

mturley (changes requested):
Looking really good, just a few minor things left.

frontend/src/pages/modelCatalog/ModelCatalogPage.tsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import { ModelCatalogContext } from '~/concepts/modelCatalog/context/ModelCatalogContext';
+
+const ModelCatalogPage: React.FC = () => {
```
mturley, line 4: This page actually isn't rendered by any route - I think the page component was already created under `frontend/src/pages/modelCatalog/screens/ModelCatalog.tsx`, you can delete this file and use that one to test taking your data from context instead

mturley (changes requested):

mturley (approved):
LGTM! Nice work! Tested on ui-green cluster using this route: http://localhost:4010/modelCatalog/Red%20Hat/rhelai1/granite-8b-code-instruct/1%252E3%252E0
(that is valid with the configmap as it exists there currently)

mturley (approved):

PR #3746, opened 2025-02-07
Connection type admin view feature flag

ashley-o0o:
Closes: [RHOAIENG-19606](https://issues.redhat.com/browse/RHOAIENG-19606)

## Description
Re-add the connection type feature flag for admin view

## How Has This Been Tested?
tested locally

## Test Impact
Cypress test checking feature flag added back in

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/backend/src/utils/constants.ts
+++ b/backend/src/utils/constants.ts
@@ -73,6 +73,7 @@ export const blankDashboardCR: DashboardConfig = {
       disableConnectionTypes: false,
       disableStorageClasses: false,
       disableNIMModelServing: true,
+      disableAdminConnectionTypes: false,
       disableFineTuning: true,
     },
     notebookController: {
--- a/frontend/src/concepts/areas/const.ts
+++ b/frontend/src/concepts/areas/const.ts
@@ -34,6 +34,7 @@ export const allFeatureFlags: string[] = Object.keys({
   disableServingRuntimeParams: false,
   disableStorageClasses: false,
   disableNIMModelServing: true,
+  disableAdminConnectionTypes: false,
   disableFineTuning: true,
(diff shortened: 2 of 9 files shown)


emilys314:
Are we going to leave this disabled by default? Or will there be a followup to change the default to enabled, because I assume we want it enabled cc @christianvogt

andrewballantyne:
> Are we going to leave this disabled by default? Or will there be a followup to change the default to enabled, because I assume we want it enabled cc @christianvogt

This is a released feature -- we have to make sure its enabled by default right? The feature was enabled before hand... the goal is to not enable features that are in development, but this passed that gate in 2.16.

backend/src/utils/constants.ts
```diff
@@ -73,6 +73,7 @@ export const blankDashboardCR: DashboardConfig = {
       disableConnectionTypes: false,
       disableStorageClasses: false,
       disableNIMModelServing: true,
+      disableAdminConnectionTypes: true,
```
andrewballantyne, line 76:
```suggestion
      disableAdminConnectionTypes: false,
```

frontend/src/concepts/areas/const.ts
```diff
@@ -34,6 +34,7 @@ export const allFeatureFlags: string[] = Object.keys({
   disableServingRuntimeParams: false,
   disableStorageClasses: false,
   disableNIMModelServing: true,
+  disableAdminConnectionTypes: true,
```
andrewballantyne, line 37:
```suggestion
  disableAdminConnectionTypes: false,
```
?

andrewballantyne (commented):
One comment, let me test it with my cluster

andrewballantyne (approved):

PR #3749, opened 2025-02-11
Add model details page for model catalog

ppadti:
Closes: [RHOAIENG-18957](https://issues.redhat.com/browse/RHOAIENG-18957)

## Description
This PR aims update details page URL and render the CatalogModel object.
Screenshot 2025-02-10 at 7 25 15 PM

## How Has This Been Tested?
1. Verify that visiting the correct URL for the details page of a specific catalog model takes you to a page showing data for that model.
2. Make sure the details page renders the details correctly.

## Test Impact
Added unit tests and cypress test.

## Request review criteria:

If you have UI changes: 


frontend/src/pages/modelCatalog/ModelCatalogRoutes.tsx
```diff
@@ -1,14 +1,14 @@
...
-import ModelDetailsPage from './ModelDetailsPage';
+import ModelDetailsPage from './screens/ModelDetailsPage';
 import ModelCatalog from './screens/ModelCatalog';
 
 const ModelCatalogRoutes: React.FC = () => (
   <Routes>
     <Route path={'/:modelCatalogSource?/*'} element={<ModelCatalogCoreLoader />}>
       <Route index element={<ModelCatalog />} />
-      <Route path="tempDetails" element={<ModelDetailsPage />} />
+      <Route path=":sourceName/:repositoryName/:modelName/:tag" element={<ModelDetailsPage />} />
```
mturley, line 11: If we are handling `:sourceName` here, we should remove `:modelCatalogSource?/*` from the above root path and it should just be `path="/"`. It makes sense to do so I think, we only need that param in the root route if we intend to do something with it in ModelRegistryCoreLoader which we don't need to.

frontend/src/pages/modelCatalog/ModelCatalogCoreLoader.tsx
```diff
@@ -2,60 +2,45 @@ import * as React from 'react';
...
+          testid="empty-model-catalog-state"
+          title="Request access to model catalog"
+          description="To request access to model catalog, contact your administrator."
+          headerIcon={() => (
+            <img src={typedEmptyImage(ProjectObjectType.registeredModels)} alt="" />
+          )}
+        />
+      </ApplicationsPage>
     );
   }
```
mturley, line 39: @YuliaKrimerman this change will conflict with your PR https://github.com/opendatahub-io/odh-dashboard/pull/3743, but no matter which order these merge the result should be as it is in 3743 - this component renders only
```
      <ModelCatalogContextProvider>
        <Outlet />
      </ModelCatalogContextProvider>
```
and we'll rely on the two page components to handle the empty state (which they already do).

frontend/src/pages/modelCatalog/routeUtils.ts
```diff
@@ -0,0 +1,11 @@
+import { encodeParams } from './utils';
+
+export const modelCatalogUrl = (): string => `/modelCatalog`;
+
+export const modelDetailsUrl = (params: Record<string, string>): string => {
```
mturley, line 5: instead of `Record<string, string>`, can you use the types you'd expect to use here for this object (include sourceName, repositoryName, etc...) and lift that into an exported type `ModelDetailsRouteParams`? You'll need to either have all of the properties of that type optional, or use `Partial<ModelDetailsRouteParams>` in your function. See my next comment...

frontend/src/pages/modelCatalog/screens/ModelDetailsPage.tsx
```diff
@@ -0,0 +1,65 @@
...
+import { CatalogModel } from '~/concepts/modelCatalog/types';
+import EmptyModelCatalogState from '~/pages/modelCatalog/EmptyModelCatalogState';
+import { decodeParams, findModelFromModelCatalogSources } from '~/pages/modelCatalog/utils';
+import ModelDetailsView from './ModelDetailsView';
+
+const ModelDetailsPage: React.FC = conditionalArea(
+  SupportedArea.MODEL_CATALOG,
+  true,
+)(() => {
+  const params = useParams();
```
mturley, line 18: We can use generics here to narrow the type of these params rather than relying on them being `Record<string, string>`.

If you import your new type `ModelDetailsRouteParams` here and pass it to useParams:
```ts
const params = useParams<ModelDetailsRouteParams>();
```
then `params` will have the type `Partial<ModelDetailsRouteParams>`. You can then use that type for all the functions you're passing these params to instead of `Record<string, string>`.

Note that if all the properties of `ModelDetailsRouteParams` are optional, you should just be able to have your functions take that instead of `Partial<ModelDetailsRouteParams>`, because the two will be equivalent.

mturley (changes requested):
Looking great so far, a couple minor things

frontend/src/pages/modelCatalog/ModelCatalogCoreLoader.tsx
ppadti: @mturley you want me to update this here in this PR? or is this something will handle it in 3743?

frontend/src/pages/modelCatalog/utils.ts
```diff
@@ -0,0 +1,38 @@
...
+  );
+
+  return modelMatched || null;
+};
+
+export const encodeParams = (params: ModelDetailsRouteParams): ModelDetailsRouteParams =>
+  Object.fromEntries(
+    Object.entries(params).map(([key, value]) => [
+      key,
+      encodeURIComponent(value).replace(/\./g, '%252E'),
```
gitdallas, line 31: just curious why this needs the extra encoding of that replace

gitdallas (approved):
lgtm

alexcreasy (approved):
Nice work!

frontend/src/pages/modelCatalog/ModelCatalogCoreLoader.tsx
mturley: Sorry I just saw this - it doesn't particularly matter as long as we end up there when both are merged. I think we can merge this as-is since it's already approved and then @YuliaKrimerman will have to address the merge conflict...... sorry @YuliaKrimerman !!

mturley (approved):
LGTM! Good stuff!

frontend/src/pages/modelCatalog/utils.ts
mturley: Hmm, yeah @ppadti I'm also curious about this - we don't need to block merge, but maybe we can follow up with a loose PR if necessary

ppadti: yeah, I did this because `:tag ` in route param has decimal value and tag being the last route param, I was getting 404 error. Even when I encode `1.3-1732870892` to `1%2E3-1732870892`, after pasting this on url  - the browser automatically decoded that. So thought of doing double encoding. But after checking this with @manaswinidas I got to know that, it was just me who got that error. I am not sure why. Sorry, I can update it with a follow up PR.

Screenshot 2025-02-13 at 5 21 41 PM

PR #3753, opened 2025-02-11
Cypress e2e - Enhance Retries to run Before hooks

antowaddle:
https://issues.redhat.com/browse/RHOAIENG-17414

## Description
- Introduction of a new custom 'Before' method called 'retryableBefore'
- The new method ensures that when tests retry that the Before method is also executed which ensures there is a cleanup prior to the retry execution
- Adds a check also in the After Method, this check will now only execute if the before hook was executed

## How Has This Been Tested?
- Locally and via Jenkins

**This screenshot shows that the retry now includes the Before Hook**
image

**The below screenshots show both Smoke and Sanity executions.**
image
image

## How to Run?
- Execute the below commands and observe the output. Note, in order to see the retries work as expected you may need to inject a fault into a test, then run that test. When the test completes and fails, observe the video and the retries performing the Before method will be present.

_npx cypress run --env grepTags="@Smoke",skipTags="@Bug" --browser chrome 
npx cypress run --env grepTags="@Sanity",skipTags="@Bug" --browser chrome_ 

## Test Impact
- This affects Cypress e2e Tests executing on the pipeline with retries 

## Request review criteria:

If you have UI changes: 


manosnoam (approved):
/lgtm

PR #3755, opened 2025-02-11
fix: ensure NIM availability updates without requiring a manual refresh

olavtar:
https://issues.redhat.com/browse/NVPE-158

## Description

Ensure NIM availability updates without requiring a manual page refresh. `refreshRate` was added.

## How Has This Been Tested?

Tested locally.

## Test Impact

1. Enable NIM
2. Go to Data Science Project
3. Click on the existing or create a new project
4. Make sure the NIM card is showing without manually refreshing the page.

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/components/OdhAppCard.tsx
+++ b/frontend/src/components/OdhAppCard.tsx
@@ -30,6 +30,7 @@ import { SupportedArea, useIsAreaAvailable } from '~/concepts/areas';
 import { isInternalRouteIntegrationsApp } from '~/utilities/utils';
 import { deleteIntegrationApp } from '~/services/integrationAppService';
 import { useUser } from '~/redux/selectors';
+import useServingPlatformStatuses from '~/pages/modelServing/useServingPlatformStatuses';
 import { useQuickStartCardSelected } from './useQuickStartCardSelected';
 import SupportedAppTitle from './SupportedAppTitle';
 import BrandImage from './BrandImage';
@@ -51,6 +52,7 @@ const OdhAppCard: React.FC<OdhAppCardProps> = ({ odhApp }) => {
   const { dashboardConfig } = useAppContext();
   const dispatch = useAppDispatch();
   const { isAdmin } = useUser();
+  const servingPlatformStatuses = useServingPlatformStatuses();
 
   const onOpenKebab = () => {
     setIsOpen(!isOpen);
@@ -89,6 +91,10 @@ const OdhAppCard: React.FC<OdhAppCardProps> = ({ odhApp }) => {
         .then((response) => {
           if (response.success) {
             handleSuccess();
+            servingPlatformStatuses.refreshNIMAvailability().catch((e) =>
+              /* eslint-disable-next-line no-console */
+              console.error('Failed to refresh NIM availability:', e),
+            );
           } else {
             throw new Error(response.error);
           }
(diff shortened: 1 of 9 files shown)


emilys314:
So i have a concern if we add a 5 second refresh on a top app level context on a feature that most users won't use (since it requires a NIM key). 

I think the Jira and the scenario trying to be fixed is when the app is enabled / disabled for the tile to change visibility correctly. 

It would be preferable to add the `refresh()` from the `useFetchState` for when NIM is getting enabled or disabled in the modal. Worst case scenario, we can add the interval checking to only that page / tile

frontend/src/pages/modelServing/useServingPlatformStatuses.ts
```diff
@@ -11,7 +11,13 @@ const useServingPlatformStatuses = (): ServingPlatformStatuses => {
...
   const modelMeshInstalled = !!modelMeshStatus.requiredComponents?.[StackComponent.MODEL_MESH];
 
-  const { isNIMAvailable } = React.useContext(NIMAvailabilityContext);
+  const { isNIMAvailable, refresh } = React.useContext(NIMAvailabilityContext);
+  React.useEffect(() => {
+    refresh().catch((error) =>
+      /* eslint-disable-next-line no-console */
+      console.error('Failed to refresh NIM availability:', error),
+    );
+  });
```
emilys314, line 20:
```suggestion
  // eslint-disable-next-line react-hooks/exhaustive-deps
  }, []);
```

Not have `,[]` will cause an infinite loop probably. the `// eslinst-disable` will allow help you run it once on load 


(we have lint rules that want it all to run, technically you would have a `const [runOnce, setRunOnce] = React.useState(false)` and in the react effect set the run once and check if it's already run before running it, but that's verbose)

frontend/src/pages/modelServing/useServingPlatformStatuses.ts
```diff
@@ -11,7 +11,19 @@ const useServingPlatformStatuses = (): ServingPlatformStatuses => {
...
+  React.useEffect(() => {
+    if (!runOnce) {
+      refresh()
+        .then(() => {
+          setRunOnce(true);
+        })
+        // eslint-disable-next-line no-console
+        .catch((error) => console.error('Failed to refresh NIM availability:', error));
+    }
+  }, [refresh, runOnce]);
```
emilys314, line 26: So it turns out putting this in the hook itself is problematic. For example if you go view all the deployed models, the hook is imported and used in each row, this will lead to a `GET /api/integrations/nim` on each row. 

So if you really need to run the `refresh`, pull it out of the hook and into the pages that need it

image

emilys314: Also another idea, you can add a prop to `useServingPlatformStatuses()` for `shouldRefreshOnInit: boolean` or something.

So when using the hook you can optionally `  const servingPlatformStatuses = useServingPlatformStatuses(true OR false);` depending on where you are using it.

So on certain pages you can set it to `true`. but in places like the `Row` you can leave it `false`

olavtar: Updated with the prop addition.

frontend/src/components/OdhAppCard.tsx
```diff
@@ -89,6 +91,10 @@ const OdhAppCard: React.FC<OdhAppCardProps> = ({ odhApp }) => {
         .then((response) => {
           if (response.success) {
             handleSuccess();
+            servingPlatformStatuses.refreshNIMAvailability().catch((e) =>
+              /* eslint-disable-next-line no-console */
+              console.error('Failed to refresh NIM availability:', e),
+            );
```
emilys314, line 97: Oof, sorry I didn't catch this earlier, but this is NIM specific code in the `OdhAppCard.tsx` which should only have generic components or integration stuff. This is the same with `frontend/src/utilities/useEnableApplication.tsx`. I'm not sure the best way to fix this.

It wouldn't make sense for NIM to refresh for other apps being enabled and disabled

PR #3764, opened 2025-02-14
Patch DSC when Admin changes Default Deployment Mode

Griffin-Sullivan:
https://issues.redhat.com/browse/RHOAIENG-16492
## Description
Adds the ability for odh admins and cluster admins to edit the cluster settings dropdown for KServe raw which changes the default deployment mode between RawDeployment (Standard) and Serverless (Advanced).

## How Has This Been Tested?
Logging in as both privileged and unprivileged users to ensure cluster settings are unavailable.

Updating the setting and saving
Make sure DSC gets updated properly
Updating should work if `defaultDeploymentMode` is present or not present
After save, the button should be disabled
Reload after a few minutes should reflect proper change

## Test Impact
There's already a test `View KServe defaultDeploymentMode` for default deployment mode. Not sure if we want to expand the testing here. @emilys314 thoughts?

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/pages/clusterSettings/ClusterSettings.tsx
+++ b/frontend/src/pages/clusterSettings/ClusterSettings.tsx
@@ -22,7 +22,7 @@ import { useKServeDeploymentMode } from '~/pages/modelServing/useKServeDeploymen
 import { SupportedArea, useIsAreaAvailable } from '~/concepts/areas';
 import TitleWithIcon from '~/concepts/design/TitleWithIcon';
 import { ProjectObjectType } from '~/concepts/design/utils';
-import { toggleInstructLabState } from '~/api/';
+import { patchDefaultDeploymentMode, toggleInstructLabState } from '~/api/';
 import {
   DEFAULT_CONFIG,
   DEFAULT_PVC_SIZE,
@@ -35,7 +35,9 @@ import useDefaultDsc from './useDefaultDsc';
 import InstructLabSettings from './InstructLabSettings';
 
 const ClusterSettings: React.FC = () => {
-  const { defaultMode: defaultSingleModelDeploymentMode } = useKServeDeploymentMode();
+  const { defaultMode } = useKServeDeploymentMode();
+  const [defaultSingleModelDeploymentMode, setDefaultSingleModelDeploymentMode] =
+    React.useState<DeploymentMode>(defaultMode);
   const [dsc, dscLoaded, dscError, refreshDsc] = useDefaultDsc();
   const [loaded, setLoaded] = React.useState(false);
   const [saving, setSaving] = React.useState(false);
@@ -88,7 +90,8 @@ const ClusterSettings: React.FC = () => {
         },
         modelServingPlatformEnabled: modelServingEnabledPlatforms,
       }) ||
-      (isAreaFineTuningEnabled && instructLabEnabled !== instructLabDSCState),
+      (isAreaFineTuningEnabled && instructLabEnabled !== instructLabDSCState) ||
+      defaultDeploymentMode !== defaultSingleModelDeploymentMode,
     [
       clusterSettings,
       pvcSize,
@@ -100,6 +103,8 @@ const ClusterSettings: React.FC = () => {
       instructLabEnabled,
       instructLabDSCState,
       isAreaFineTuningEnabled,
+      defaultDeploymentMode,
+      defaultSingleModelDeploymentMode,
     ],
   );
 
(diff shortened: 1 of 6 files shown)


emilys314:
@Griffin-Sullivan Yeah I think expanding on the "View KServe defaultDeploymentMode" test is a good idea. You can just edit that one to include writing the DSC too if you'd like

frontend/src/pages/clusterSettings/ModelServingPlatformSettings.tsx
```diff
@@ -155,12 +161,12 @@ const ModelServingPlatformSettings: React.FC<ModelServingPlatformSettingsProps>
...
                         key: DeploymentMode.RawDeployment,
                         label: 'Standard (No additional dependencies)',
-                        isDisabled: true, // todo: allow admin to update dsc
+                        isDisabled: !allowedToPatchDSC,
                       },
                       {
                         key: DeploymentMode.Serverless,
                         label: 'Advanced (Serverless and Service Mesh)',
-                        isDisabled: !isServerlessAvailable || true, // todo: allow admin to update dsc
+                        isDisabled: !isServerlessAvailable && !allowedToPatchDSC,
```
emilys314, line 169:
```suggestion
                        isDisabled: !isServerlessAvailable || !allowedToPatchDSC,
```
This should be an `||` because if serverless is not available, you shouldn't be able to select it

frontend/src/pages/clusterSettings/ModelServingPlatformSettings.tsx
```diff
@@ -49,9 +50,13 @@ const ModelServingPlatformSettings: React.FC<ModelServingPlatformSettingsProps>
     modelMesh: { installed: modelMeshInstalled },
   } = useServingPlatformStatuses();
 
-  const [allowUpdate] = useAccessReview(accessReviewResource);
+  const [allowUpdate] = useAccessReview(accessReviewResource); // TODO: remove and use allowedToPatchDSC
```
emilys314, line 53: I think you can replace the usage of this on line 99 now and not wait on a todo

frontend/src/pages/clusterSettings/ModelServingPlatformSettings.tsx
```diff
@@ -133,7 +138,8 @@ const ModelServingPlatformSettings: React.FC<ModelServingPlatformSettingsProps>
             name="singleModelServingPlatformEnabledCheckbox"
             body={
               kServeInstalled &&
-              isRawAvailable && (
+              isRawAvailable &&
+              loadedAllowed && (
```
emilys314, line 142: Waiting for the request to load before using it is always good, but in this case it's conditionally rendering stuff, which causes layout shifting. 

I would say it's fine to remove it here, because `allowedToPatchDSC` is being used inside of the simple select options. So by the time the user will open that, it should have finished loading. (If it didn't, `allowedToPatchDSC` should be defaulting to false until it's loaded)

frontend/src/pages/clusterSettings/ClusterSettings.tsx
```diff
@@ -149,7 +156,15 @@ const ClusterSettings: React.FC = () => {
           },
         );
 
-    Promise.all([clusterSettingsPromise, instructLabPromise])
+    const defaultDeploymentModePromise = defaultDeploymentModeUnchanged
+      ? Promise.resolve()
+      : dsc &&
+        patchDefaultDeploymentMode(defaultDeploymentMode, dsc.metadata.name).then(() => {
+          setDefaultSingleModelDeploymentMode(defaultDeploymentMode);
```
emilys314, line 163: What is `setDefaultSingleModelDeploymentMode(defaultDeploymentMode);` doing here? It seems redundant

emilys314 (commented):
I tested it and it works! Although i left some comments. It's exciting though thank you

frontend/src/pages/clusterSettings/ClusterSettings.tsx
Griffin-Sullivan: Yeah it is redundant but needed for `isSettingsChanged`. `useKserveDeploymentMode` doesn't read and set the state of the deployment mode until the backend has refreshed everything (~ 2 minutes). Without this setDefault you can submit the settings changes, but after submit the button is not disabled again.

emilys314:
I tested on a cluster with only raw enabled and then another with both enabled and everything looks good!

/lgtm
/approve

PR #3770, opened 2025-02-18
Render loading/error states on catalog and details pages 

YuliaKrimerman:
Closes https://issues.redhat.com/browse/RHOAIENG-18960 

## Description

Added empty and loading states for both Model Catalog and Model Catalog Details page. 

https://github.com/user-attachments/assets/6a4f8b7c-83a8-4010-8a69-ebfae84d9c1f

## Test Impact

Added test cases to test all scenarios :

- if the configmap is missing (intercept responds 404, you can see other examples we talked about earlier for that), we show the empty state
-if there is a non-404 error loading the configmap, we show an error message
-if the configmap is present but has empty data (data: { modelCatalogSource: '[]' }), we show the empty state
-if the configmap is present but has malformed data / bad JSON (data: { modelCatalogSource: 'invalid JSON here' } , we show the empty state
-if the configmap is present and has good data (the intercept you have now), we show the actual model details / cards

## Request review criteria:

If you have UI changes: 


frontend/src/concepts/modelCatalog/useModelCatalogSources.ts
```diff
@@ -20,18 +20,34 @@ export const useModelCatalogSources = (): FetchState<State> => {
...
-    }
-
     try {
-      const source: ModelCatalogSource = JSON.parse(configMap.data.modelCatalogSource);
-      return [source];
-    } catch (e) {
-      // Swallow JSON parse errors and return empty array for temporary summit implementation
-      return [];
+      const configMap = await getConfigMap(dashboardNamespace, MODEL_CATALOG_SOURCE_CONFIGMAP);
+      /* eslint-enable @typescript-eslint/no-unsafe-member-access */
```
alexcreasy, line 25: Curious as to why this is necessary?

frontend/src/concepts/modelCatalog/useModelCatalogSources.ts
```diff
@@ -20,18 +20,34 @@ export const useModelCatalogSources = (): FetchState<State> => {
...
+
+      if (!configMap.data || !configMap.data.modelCatalogSource) {
+        return [];
+      }
+
+      try {
+        // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
+        const source = JSON.parse(configMap.data.modelCatalogSource) as ModelCatalogSource;
```
alexcreasy, line 33: You don't need this linter override here if you specify the type of source.
```suggestion
        const source: ModelCatalogSource = JSON.parse(configMap.data.modelCatalogSource);
```

frontend/src/concepts/modelCatalog/useModelCatalogSources.ts
```diff
@@ -20,18 +20,34 @@ export const useModelCatalogSources = (): FetchState<State> => {
...
+    } catch (e: unknown) {
+      // Check if error is a 404
+      if (
+        // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
+        'statusObject' in (e as object) &&
+        // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
+        isK8sStatus((e as { statusObject: unknown }).statusObject)
+      ) {
+        return [];
+      }
```
alexcreasy, line 48: You don't need the linter overrides here either, as long as you have enough guards to narrow the type down tightly enough you won't need the `as` keyword, which we generally want to avoid like the plague as that really has the effect of saying to the type system: "I've got this, kick back and chill for a bit", so we're losing compile time type safety. 

If you actually narrow the type through conditional statements rather than using `as`, you're proving to the compiler the code is type safe, so any unexpected runtime values will be handled.

Try something like this, it's not pretty, but it's type safe and avoids the linter overrides:

```suggestion
      // Check if error is a 404
      if (
        typeof e === 'object' &&
        e != null &&
        'statusObject' in e &&
        isK8sStatus(e.statusObject)
      ) {
        return [];
      }
```

alexcreasy (changes requested):
Great work! I'd remove the linter overrides though :)

I haven't tested this yet, I'll run it against a cluster this afternoon.

frontend/src/__tests__/cypress/cypress/tests/mocked/modelCatalog/modelCatalog.cy.ts
```diff
@@ -69,3 +69,108 @@ describe('Model Catalog core', () => {
...
+            models: [],
+          }),
+        },
+      },
+    );
+    modelCatalog.visit();
+    modelCatalog.findModelCatalogEmptyState().should('exist');
+  });
+
+  it('should show empty state when configmap has malformed data', () => {
```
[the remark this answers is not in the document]
YuliaKrimerman: @mturley I think he can answer this better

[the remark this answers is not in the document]
mturley: Hmm. Honestly it probably does make sense to render an error if there is malformed JSON... I'm not sure what I was thinking there 👀  maybe that it might be an ugly error? I think I'm just used to always making sure to catch JSON.parse exceptions, but in this case since it's in a fetch callback letting that error be thrown is probably a good idea so the admin messing with the configmap can see what they did wrong.

Sorry @YuliaKrimerman .... can you just remove the innermost `try/catch` around the JSON.parse? We can proceed as if it succeeded and let an exception there cause the fetch promise to be rejected. We should also update that test case to look for the rendered error.

frontend/src/__tests__/cypress/cypress/tests/mocked/modelCatalog/modelDetailsPage.cy.ts
```diff
@@ -54,3 +54,113 @@ describe('Model details page', () => {
...
+          message: 'Internal server error',
+          reason: 'InternalError',
+          code: 500,
+        },
+      },
+    );
+
+    modelDetailsPage.visit();
+    cy.contains('Unable to load model details').should('exist');
+  });
```
mturley, line 151: **EDIT --- we may not need any changes here. I think this test is passing because of the incorrect contents being rendered in emptyStatePage. See comments below before you try to change anything here.**

-----

Hang on... I'm confused about why this test is passing. You have the non-404 Status object like we need to add to the other non-404 test here, but (see my earlier comment) your current code is eating errors and returning `[]` if they have a statusObject even if it's not 404. This test is written correctly but it should be failing (and adding the `&& e.statusObject.code === 404` I mentioned should fix it).

I wonder if the problem is that you have two `cy.interceptK8s` calls for this configmap (one in `initIntercepts` and one in the test) and Cypress is using the first intercept for some reason? I wonder, does this test fail if you comment out the configmap intercept in `initIntercepts` so this is the only intercept?

Maybe we should add a param to `initIntercepts` like `includeConfigMap = true`, and then in these `describe` blocks around the tests that have their own configmap intercepts we should change the `beforeEach` so it passes true there and there is only ever one configmap intercept per test. We probably want to do that in the test files for both of these pages.

If that's not the issue, I'm not sure why this test would be passing...

frontend/src/concepts/modelCatalog/useModelCatalogSources.ts
```diff
@@ -20,18 +24,32 @@ export const useModelCatalogSources = (): FetchState<State> => {
...
+      const configMap = await getConfigMap(dashboardNamespace, MODEL_CATALOG_SOURCE_CONFIGMAP);
+      if (!configMap.data || !configMap.data.modelCatalogSource) {
+        return [];
+      }
+
+      try {
+        const parsed = JSON.parse(configMap.data.modelCatalogSource);
+        if (isModelCatalogSource(parsed)) {
+          return [parsed];
+        }
```
mturley, line 37: I think we either need to make `isModelCatalogSource` actually inspect the properties inside the object to make sure they match the type, or we need to remove it and not bother checking. I'd lean towards the latter.

A lot of the dashboard code makes the assumption that the API will return data that matches our TypeScript types - actually checking all the types at run time isn't practical. We are curating the contents of the configmap and not supporting manual edits to it, so it's essentially our API contract - it's editable by a cluster admin in case they really need to squeeze another model into the Summit demo, but normal users won't be editing it. I think it's safe to assume it's correctly formatted as long as it's valid JSON (which it is if JSON.parse didn't throw an exception).

I think it would be fine to get rid of `isModelCatalogSource` and just do this:

```ts
return parsed ? [parsed as ModelCatalogSource] : [];
```

I'm curious if @alexcreasy disagrees.

Alternatively, we could have `isModelCatalogSource` also check that `obj.models.length > 0`?

frontend/src/pages/modelCatalog/screens/ModelCatalog.tsx
```diff
@@ -15,16 +15,35 @@ const ModelCatalog: React.FC = conditionalArea(
 )(() => {
   const { modelCatalogSources } = React.useContext(ModelCatalogContext);
   const renderStateProps = {
-    empty: modelCatalogSources.data.length === 0,
+    empty: modelCatalogSources.data.length === 0 || !modelCatalogSources.data[0]?.models?.length,
```
mturley, line 18: This is kinda funky - ideally we shouldn't write anything outside of `useModelCatalogSources` that assumes we have only one source (since we'll be adding a second source soon for Neural Magic), so we don't want to do anything based on only looking at `data[0]`. I think we can assume that if we have sources they have models, so we can probably remove this line.

(Note: this is why I added the end of my comment above about `isModelCatalogSource` - if you do want to check that a source has models before trying to use it, we should do that check there instead).

frontend/src/pages/modelCatalog/screens/ModelCatalog.tsx
```diff
@@ -15,16 +15,35 @@ const ModelCatalog: React.FC = conditionalArea(
...
+        title={
+          modelCatalogSources.error
+            ? 'Unable to load model catalog'
+            : 'Request access to model catalog'
+        }
+        description={
+          modelCatalogSources.error
+            ? 'Refresh the page or try again later'
+            : 'To request access to model catalog, contact your administrator.'
+        }
```
mturley, line 31: We don't need to check anything related to `modelCatalogSources.error` in the `emptyStatePage` -- if there's an error, the `loadError` prop below will be true and we won't be rendering this empty state at all. Instead, ApplicationsPage will use its internal error rendering which will use your `errorMessage` prop below.

We can put the whole EmptyModelCatalogState back the way it was here - the `loadError` and `errorMessage` props of ApplicationsPage are sufficient for rendering errors on these pages.

frontend/src/pages/modelCatalog/screens/ModelCatalog.tsx
```diff
@@ -33,8 +52,6 @@ const ModelCatalog: React.FC = conditionalArea(
         <TitleWithIcon title="Model Catalog" objectType={ProjectObjectType.registeredModels} />
       }
       {...renderStateProps}
-      loaded={modelCatalogSources.loaded}
-      provideChildrenPadding
```
mturley, line 37: Why remove the `provideChildrenPadding` prop?

frontend/src/pages/modelCatalog/screens/ModelCatalog.tsx
```diff
@@ -15,16 +15,35 @@ const ModelCatalog: React.FC = conditionalArea(
 )(() => {
   const { modelCatalogSources } = React.useContext(ModelCatalogContext);
   const renderStateProps = {
```
mturley, line 17: I think we can take this whole `renderStateProps` object and move its stuff as directly passing props to ApplicationsPage below, like we have for ModelDetailsPage. It was lifted out in other pages because it was conditional / it would vary, but there's no reason for that here.

frontend/src/pages/modelCatalog/screens/ModelDetailsPage.tsx
```diff
@@ -42,18 +42,20 @@ const ModelDetailsPage: React.FC = conditionalArea(
           objectType={ProjectObjectType.registeredModels}
         />
       }
-      empty={model === null}
+      empty={Boolean(modelCatalogSources.error) || model === null}
```
mturley, line 45: Similar to the other page - ApplicationsPage renders errors before it renders empty. We're passing `loadError={modelCatalogSources.error}`, so if `modelCatalogSources.error` is defined, `emptyStatePage` is not going to be rendered regardless of the value of `empty`. So we can change this back to `model === null`.

frontend/src/pages/modelCatalog/screens/ModelDetailsPage.tsx
```diff
@@ -42,18 +42,20 @@ const ModelDetailsPage: React.FC = conditionalArea(
...
           testid="empty-model-catalog-state"
-          title="Details not found"
-          description="To request access to model catalog, contact your administrator."
+          title={modelCatalogSources.error ? 'Details not found' : 'Unable to load model details'}
+          description={modelCatalogSources.error?.message || 'Refresh the page or try again later'}
           headerIcon={() => (
-            <img src={typedEmptyImage(ProjectObjectType.registeredModels)} alt="" />
+            <img src={typedEmptyImage(ProjectObjectType.registeredModels, 'Error')} alt="" />
           )}
         />
```
mturley, line 54: Similarly here, we don't need to do any of this error rendering in emptyStatePage. You can put this all back as it was - it is only used if there are no model catalog sources.

Aha -- I know why your mysterious non-404 cypress test was passing when it shouldn't. You're rendering `'Unable to load model details'` here in the empty state even though there is no error. If you revert this stuff, that test will fail as it should. I'll edit the earlier comment.

mturley (changes requested):
@YuliaKrimerman this is great work - just a few things we need to tweak.

frontend/src/concepts/modelCatalog/useModelCatalogSources.ts
YuliaKrimerman: @mturley @alexcreasy So if I'm using return parsed ? [parsed as ModelCatalogSource] : []; then I will need to add the eslint ignore for "Do not use any type assertions"

YuliaKrimerman:
@mturley @alexcreasy Ready for re-review.

gitdallas (approved):

gitdallas (approved):
was going to mention that test is probably failing after rebase that gave the new details page (which removed the placeholder register button)... but looks like you did that at about the same time :P

lgtm

frontend/src/concepts/modelCatalog/useModelCatalogSources.ts
mturley: @alexcreasy please feel free to push back, but I chatted with @YuliaKrimerman about this and one way or another we need to tell TypeScript that this parsed JSON is a ModelCatalogSource. We debated having the `isModelCatalogSource` type guard vs using `as ModelCatalogSource` and turning off the `@typescript-eslint/consistent-type-assertions` for that line. IMO, if we aren't going to write a full type guard that inspects all the fields (not really practical), it's better to use the assertion and have the lint rule disabled so it stands out more in case we want to go through later and remove those lint disabled comments and inspect why we did this. The type guard that doesn't really check anything has the same behavior as a type assertion but is less obvious. IMO it doesn't matter a whole lot for this temporary configmap approach since it is curated by us - if it's valid JSON we can assume it has the right structure.

Now.... hm. I am looking for other examples of how we handle types with `JSON.parse`, and I see we do have type guards that only check the top level properties [like this one in kserve metrics](https://github.com/opendatahub-io/odh-dashboard/blob/main/frontend/src/concepts/metrics/kserve/utils.ts#L18-L24). I hadn't seen that before, maybe that's evidence we should go with the type assertion approach.

I also see places like [in storage classes here](https://github.com/opendatahub-io/odh-dashboard/blob/main/frontend/src/pages/storageClasses/utils.ts#L8-L10) where we do neither type guard nor type assertion - in fact, since `JSON.parse` returns `any`, all it takes is a type annotation on the variable you're assigning it to and TS is happy. Didn't occur to me before.

So @YuliaKrimerman maybe that's what we should do - remove the linter comment and just put a type annotation on `parsed`:

```ts
const parsed: ModelCatalogSource | undefined = JSON.parse(configMap.data.modelCatalogSource);
return parsed ? [parsed] : [];
```

mturley (approved):
LGTM, thanks @YuliaKrimerman !

frontend/src/concepts/modelCatalog/useModelCatalogSources.ts
alexcreasy: Yeah, i prefer going with the type annotation - whilst having the same net effect as an "as" statement in this case - it's preferable as we're not using a lint override.  Each time we put an override in the code it has the effect of making doing it more acceptable to do that for anyone else who sees them, ultimately they're a strong line of protection if used correctly, so this is why I prefer the type annotation over the "as" in this instance.

PR #3772, opened 2025-02-18
Enhancement to ODS-1218 'Verify That Usage Data Collection Can Be Set In Cluster Settings'

ConorOM1:
## Description
Enhancement for test ODS-1218 to improve the scenario for when a cluster is not torn down

## How Has This Been Tested?

- Tested against ODH-Nightly & RHOAI nightly

image

## How to Run
After exporting the test-variables.yml we have 2 different ways:

Using the UI
Go to odh-dashboard/frontend/src/tests/cypress and run the command npx cypress open . This will open the Cypress UI where testDataCollection.cy.ts can be run.

Headless
Go to odh-dashboard/frontend/src/tests/cypress and run the command npx cypress run --spec "cypress/tests/e2e/settings/clusterSettings/testDataCollection.cy.ts --browser chrome

## Test Impact:

- This is already a test

If you have UI changes: 


antowaddle (approved):
\lgtm

PR #3777, opened 2025-02-19
Add teacher/judge section to ilab form

DaoDaoNoCode:
JIRA: [RHOAIENG-19108](https://issues.redhat.com/browse/RHOAIENG-19108)

## Description

Screenshot 2025-02-19 at 4 35 59 PM
Screenshot 2025-02-19 at 4 36 14 PM

Added 2 similar sections to the form: Teacher model and Judge model. Here are the descriptions.
1. Added validation to these 2 sections. When the public endpoint is chosen, it doesn't require a token to be input. When the private endpoint is selected, it requires the token field not to be empty. Also, the endpoint field needs to follow the URI format.
2. I leave the links to the docs in the description as empty for now
3. When submitting, it will create 2 separate secrets for both teacher and judge. There are 3 fields in the secrets: `api_token`, 'endpoint' and `model_name`. When it's public, the `api_token` field will be set as empty string.
4. The name of these 2 secrets are not decided yet, thus, after creating the 2 secrets, the corresponding parameters in the pipeline run form are not decided, either. I added some TODO comments to it.

## How Has This Been Tested?

1. Test whether the submit button is disabled/enabled as described in the description by changing it between public/private and changing the format of the endpoint
2. Test whether submitting it will create 2 secrets

## Test Impact

Added unit tests for the validation
Added cypress tests for the 2 sections

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/concepts/pipelines/content/modelCustomizationForm/modelCustomizationFormSchema/validationUtils.ts
+++ b/frontend/src/concepts/pipelines/content/modelCustomizationForm/modelCustomizationFormSchema/validationUtils.ts
@@ -27,7 +27,7 @@ export const baseModelSchema = z.object({
 
 const teacherJudgeBaseSchema = z.object({
   endpoint: uriFieldSchemaBase(false),
-  modelName: z.string().min(1, 'Model name is required'),
+  modelName: z.string().trim().min(1, 'Model name is required'),
 });
 const teacherJudgePublicSchema = teacherJudgeBaseSchema.extend({
   endpointType: z.literal(ModelCustomizationEndpointType.PUBLIC),
@@ -35,7 +35,7 @@ const teacherJudgePublicSchema = teacherJudgeBaseSchema.extend({
 });
 const teacherJudgePrivateSchema = teacherJudgeBaseSchema.extend({
   endpointType: z.literal(ModelCustomizationEndpointType.PRIVATE),
-  apiToken: z.string().min(1, 'Token is required'),
+  apiToken: z.string().trim().min(1, 'Token is required'),
 });
 
 export const teacherJudgeModel = z.discriminatedUnion('endpointType', [
--- a/frontend/src/concepts/pipelines/content/modelCustomizationForm/modelCustomizationFormSchema/validationUtils.ts
+++ b/frontend/src/concepts/pipelines/content/modelCustomizationForm/modelCustomizationFormSchema/validationUtils.ts
@@ -1,29 +1,32 @@
 import { z } from 'zod';
 import { ModelCustomizationEndpointType, ModelCustomizationRunType } from './types';
 
-export const uriFieldSchema = z.string().refine(
-  (value) => {
-    if (!value) {
-      return true;
-    }
-    try {
-      return !!new URL(value);
-    } catch (e) {
-      return false;
-    }
-  },
-  { message: 'Invalid URI' },
-);
+export const uriFieldSchemaBase = (
+  isOptional: boolean,
+): z.ZodEffects<z.ZodString, string, string> =>
+  z.string().refine(
+    (value) => {
+      if (!value) {
+        return !!isOptional;
+      }
+      try {
+        return !!new URL(value);
+      } catch (e) {
+        return false;
+      }
+    },
+    { message: 'Invalid URI' },
+  );
 
 export const baseModelSchema = z.object({
   registryName: z.string(),
   name: z.string(),
   version: z.string(),
-  inputStorageLocationUri: uriFieldSchema,
+  inputStorageLocationUri: uriFieldSchemaBase(true),
 });
 
 const teacherJudgeBaseSchema = z.object({
-  endpoint: uriFieldSchema,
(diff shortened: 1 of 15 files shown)


dpanshug (changes requested):
it looks good to me, just a nit

DaoDaoNoCode:
Updated the PR, use `teacher-secret-${randomChars}` and `judge-secret-${randomChars}` as the secret names and set them to the `judge_secret` and `teacher_secret` field to the run parameters.
What's more, I added `dryRun` to the submit run function as well, so we can make sure we dry run everything before really creating a run, in case we really create some secrets but meet errors later when creating the run.
Now the question is - Is there a way to use the set function that's provided by the `useRunFormData` hook? You can see I am simply mapping the values when creating the run [here](https://github.com/opendatahub-io/odh-dashboard/pull/3777/files#diff-943f40a01a10cc31f4c8b28cb4aba014260bf4f07ae63a456bb17b5fcae031a8R58-R64) instead of setting them to the run form state because I cannot do the state setting and reading simultaneously. Any ideas? cc @Gkrumbach07

frontend/src/pages/pipelines/global/modelCustomization/FineTunePageFooter.tsx
```diff
@@ -33,8 +45,105 @@ const FineTunePageFooter: React.FC<FineTunePageFooterProps> = ({ isInvalid, onSu
   // TODO: translate data to `RunFormData`
   const [runFormData] = useRunFormData(null, {});
 
+  const onSubmit = async (dryRun: boolean) => {
+    const [teacherSecret, judgeSecret] = await createTeacherJudgeSecrets(
+      namespace,
+      data.teacher,
+      data.judge,
+      dryRun,
+    );
```
DaoDaoNoCode, line 54: In this way, the dry-run creation and the real creation will generate 2 different secret names. IMO it's fine because as long as we are setting the real created secret names to the parameters, it's OK. The dry run will only test whether there are any other problems or not.
However, if anyone who reviews the PR thinks that we need to use the same secret names for both dry-run creation and real creation, please let me know, I will think of another way to store the dry-run created names and reuse the names when we really create the secrets.

pnaik1:
The functionality and Zod validation work well. I'm  also waiting for the above comment so I can make changes in my [PR](https://github.com/opendatahub-io/odh-dashboard/pull/3785) accordingly

pnaik1:
@DaoDaoNoCode , I just notice that `start run` button is enabled even when we dont have any value in the `endpoint`
Can you look at the zod validation for this??
Screenshot 2025-02-25 at 6 53 25 PM

DaoDaoNoCode:
@pnaik1 Nice catch! I updated the zod validation functions.

Gkrumbach07:
@DaoDaoNoCode take a look at what Purva did for here taxonomy PR
https://github.com/opendatahub-io/odh-dashboard/pull/3785/files#diff-c01f9880ad8c80ab019ba783fb672667a3199b2daabc6be55a355bf9e53e9461

we should just do something like 

```tsx
const [runFormData] = useRunFormData(ilabPipelineVersion);
```
then with on submit you do something like

```tsx
handleSubmit({
   ...runFormData,
   ...translateIlabFormToPipelineInput(formData, judgeSecretName, teacherSecretName)
})
```

frontend/src/concepts/pipelines/content/modelCustomizationForm/modelCustomizationFormSchema/validationUtils.ts
```diff
@@ -1,36 +1,47 @@
...
-  endpointType: z.enum([
-    ModelCustomizationEndpointType.PUBLIC,
-    ModelCustomizationEndpointType.PRIVATE,
-  ]),
-  endpoint: uriFieldSchema,
-  username: z.string().min(1, 'Username is required'),
-  password: z.string().min(1, 'Password is required'),
+const teacherJudgeBaseSchema = z.object({
+  endpoint: uriFieldSchemaBase(false),
+  modelName: z.string().min(1, 'Model name is required'),
```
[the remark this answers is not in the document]
DaoDaoNoCode: Right, I will change it to prevent whitespace.

frontend/src/pages/pipelines/global/modelCustomization/FineTunePageFooter.tsx
```diff
@@ -33,8 +53,109 @@ const FineTunePageFooter: React.FC<FineTunePageFooterProps> = ({ isInvalid, onSu
   // TODO: translate data to `RunFormData`
   const [runFormData] = useRunFormData(null, {});
```
Gkrumbach07, line 54: can you put the ilabPipeline in instead of null

Gkrumbach07: also {} is optional so no need to have that

Gkrumbach07 (approved):
/lgtm

PR #3781, opened 2025-02-20
Add register model page for model catalog

ppadti:
Closes: [RHOAIENG-19060](https://issues.redhat.com/browse/RHOAIENG-19060) and [RHOAIENG-20245](https://issues.redhat.com/browse/RHOAIENG-20245)

## Description
This PR aims to add Register model page for model catalog. Also, fixed https://issues.redhat.com/browse/RHOAIENG-20245

1. Register model button is disabled with popover, when no registry present:

Screenshot 2025-02-20 at 2 03 11 PM

2. Register model page, when more than one registry present:

Screenshot 2025-02-20 at 2 07 21 PM

3. Register model flow from model catalog page:

https://github.com/user-attachments/assets/ddd7174f-88a0-43a7-bd19-2af402726319

4. You will see this error, if you try to register a model with unavailable model registry: 

Screenshot 2025-02-26 at 10 27 58 AM

NOTE: 
1. Now we won't be able to see the model name duplicate error helptext on onChange (but you will see duplicate error name on submit), as we need to consume ModelRegistryConext in modelCatalog - will create a new issue for that.

## How Has This Been Tested?
- Go to the model catalog, click a model, and see that there is a Register button in the top right corner.

 - Verify that the Register button is disabled with a tooltip if there are no model registries the user can access.
 
- With model registries available, click the Register button. See that you are taken to a register form that shows the correct model name,  prefills the correct URI and if **more than one model registries** are present you will be able to select it. If you have only **one model registry**, then the model registry will auto selected and the field will be disabled.

 - Verify that you are able to select a model registry, fill out the form and register a model. Verify that the registered model and version contain all the expected details based on the mockup. 
 
## Test Impact
Updated unit and added new cypress tests for register model and model details page.

## Request review criteria:

If you have UI changes: 


mturley (commented):
Hey @ppadti I'm sorry, I ran out of time to review things for the sprint - I'll get back to this ASAP. You've got some merge conflicts though.

yih-wang:
Thanks for the updates @ppadti! Look great!
The only comment is for the model registry selector, which might need an update to both this form and the MR main page - 
after consulting @angesDing, the latest ux guidelines of displaying a single option dropdown is not disabling the selector ([FYI](https://docs.google.com/document/d/1htTeVjOl-RjmJvLSJr6mw0SOnIBaWcZJ1fhJfuKotI0/edit?tab=t.0#bookmark=id.jx5ei0nlng6w)). So in the video, the field 'Model Registry' should be enabled, and the only option is selected by default.

ppadti:
> Thanks for the updates @ppadti! Look great! The only comment is for the model registry selector, which might need an update to both this form and the MR main page - after consulting @angesDing, the latest ux guidelines of displaying a single option dropdown is not disabling the selector ([FYI](https://docs.google.com/document/d/1htTeVjOl-RjmJvLSJr6mw0SOnIBaWcZJ1fhJfuKotI0/edit?tab=t.0#bookmark=id.jx5ei0nlng6w)). So in the video, the field 'Model Registry' should be enabled, and the only option is selected by default.

Thanks @yih-wang for taking a look at the screenshots. 
I think for the model registry selector to update to not to disable the dropdown when only one option available - this will require me to update [SimpleSelect](https://github.com/opendatahub-io/odh-dashboard/blob/main/frontend/src/components/SimpleSelect.tsx#L131) component - this is used in multiple places, so updating this might impact other components. I think this might be out of scope for this issue. Would it makes sense to create a separate issue to track and handle this? Let me know your thoughts!

yih-wang:
@ppadti I see. Yes I agree with you that this shouldn't belong to this issue then. cc @angesDing @simrandhaliw, we might need a consistency issue to address this change as a whole.

ppadti:
> i'm getting errors when i try to submit the form, is there something i am missing?
> 
> you mention we won't be able to see duplicate name error, but that's all i'm seeing: image
> 
> if i use `test` instead of `eder-mr` for the form, i get this: image
> 
> also kind of curious why in your recording, the MR looks disabled. i'm able to change it.

Sorry, I think the PR description was not clear enough to test this. Updated the description let me know if you need anything else.

frontend/src/concepts/modelRegistry/utils/updateTimestamps.ts
```diff
@@ -2,13 +2,15 @@ import {
...
   ModelRegistryMetadataType,
+  ModelRegistryCustomProperties,
 } from '~/concepts/modelRegistry/types';
 
 type MinimalModelRegistryAPI = Pick<ModelRegistryAPIs, 'patchRegisteredModel'>;
 
 export const bumpModelVersionTimestamp = async (
   api: ModelRegistryAPIs,
   modelVersionId: string,
+  customProperties: ModelRegistryCustomProperties,
```
mturley, line 13: Ok. @ppadti please don't kill me. Looking more at this I think it's not good for us to have to pass in the customProperties in `bumpModelVersionTimestamp` and `bumpRegisteredModelTimestamp`. If they are accidentally passed as `{}` for some reason or passed the wrong thing (model/version properties mixed up), we'll have a problem.

If it's practical, I think it would be better to refactor both of these utils so that they don't take `customProperties` or ids, but they take the entire current copy of the object they're updating, and then use the customProperties from it. So we'd have `bumpModelVersionTimestamp(api, modelVersion)` which uses `...modelVersion.customProperties`, `bumpRegisteredModelTimestamp(api, registeredModel) which uses `...registeredModel.customProperties`, and `bumpBothTimestamps(api, modelVersion, registeredModel)`.

That would require making sure everywhere we call all these we have the full model/version object in scope, but I think we do for the most part. The one place I think we don't is inside `createModelVersionForRegisteredModel` in `frontend/src/api/modelRegistry/custom.ts`. We only call that in one place though (the `registerVersion` util in `frontend/src/pages/modelRegistry/screens/RegisterModel/utils.ts`), and we could either just pass in the whole `registeredModel` there or lift out the `bump` so it happens in `registerVersion` and not in the underlying api util, whichever you think is cleaner.

I think this way we'll save ourselves some potential future headaches and also make the code a bit readable, right now it's not immediately clear why we'd be passing around customProperties like this.

frontend/src/pages/modelCatalog/screens/RegisterCatalogModel.tsx
```diff
@@ -0,0 +1,216 @@
...
+
+  const { preferredModelRegistry } = React.useContext(ModelRegistrySelectorContext);
+  const { modelCatalogSources } = React.useContext(ModelCatalogContext);
+
+  const [isSubmitting, setIsSubmitting] = React.useState(false);
+  const [submitError, setSubmitError] = React.useState<Error | undefined>(undefined);
+  const [formData, setData] = useRegisterCatalogModelData();
+  const [registeredModelName, setRegisteredModelName] = React.useState<string>('');
+  const [versionName, setVersionName] = React.useState<string>('');
+  const [errorName, setErrorName] = React.useState<string | undefined>(undefined);
```
mturley, line 55: Hey - it's ok not to block this PR on this since I know this is duplicated from the RegisterModel and RegisterVersion pages, but I think this state is a little confusing. I know it exists because when there's an error submitting we need to show the names of the resources that were attempted to be created even if the form has since changed, but with these names it looks like duplicated state that we might try to use expecting it to be the form values. I would love to rename these to something like `submittedRegisteredModelName` and `submittedVersionName`, and maybe `errorName` (and the `ErrorName` type) should be `registrationErrorType` or something.

We probably also want to see what we can deduplicate between these 3 pages... that definitely belongs in a followup tech debt issue though.

frontend/src/pages/modelCatalog/utils.ts
```diff
@@ -26,10 +26,7 @@ export const findModelFromModelCatalogSources = (
 
 export const encodeParams = (params: ModelDetailsRouteParams): ModelDetailsRouteParams =>
   Object.fromEntries(
-    Object.entries(params).map(([key, value]) => [
-      key,
-      encodeURIComponent(value).replace(/\./g, '%252E'),
-    ]),
+    Object.entries(params).map(([key, value]) => [key, encodeURIComponent(value)]),
```
mturley, line 29: Nice, this is way better, thanks.

frontend/src/pages/modelRegistry/screens/ModelRegistrySelector.tsx
```diff
@@ -32,12 +32,14 @@ type ModelRegistrySelectorProps = {
...
   onSelection: (modelRegistry: string) => void;
   primary?: boolean;
+  isCatalogModel?: boolean;
 };
 
 const ModelRegistrySelector: React.FC<ModelRegistrySelectorProps> = ({
   modelRegistry,
   onSelection,
   primary,
+  isCatalogModel,
```
mturley, line 42: I think this prop can just be called `isFullWidth`, there may be other forms in the future other than catalog where we want to use the ModelRegistrySelector.

frontend/src/pages/modelRegistry/screens/ModelVersionDetails/ModelVersionDetailsView.tsx
```diff
@@ -43,6 +44,7 @@ const ModelVersionDetailsView: React.FC<ModelVersionDetailsViewProps> = ({
   const modelArtifact = modelArtifacts.items.length ? modelArtifacts.items[0] : null;
   const { apiState } = React.useContext(ModelRegistryContext);
   const storageFields = uriToStorageFields(modelArtifact?.uri || '');
+  const [rm] = useRegisteredModelById(mv.registeredModelId);
```
mturley, line 47: We should pull in loaded, loadError and refresh from this fetch hook and make sure we handle them properly.

We can create composite variables for:
```ts
const loaded = modelArtifactsLoaded && registeredModelLoaded;
const loadError = modelArtifactsLoadError || registeredModelLoadError;
const refresh = () => {
  refreshModelArtifacts();
  refreshRegisteredModel();
};
```

and then use those any place we currently use `modelArtifactsLoaded`, `modelArtifactsLoadError` and `refreshModelArtifacts` here.

frontend/src/pages/modelRegistry/screens/RegisterModel/utils.ts
```diff
@@ -154,6 +162,14 @@ export const isRegisterModelSubmitDisabled = (
 export const isRegisterVersionSubmitDisabled = (formData: RegisterVersionFormData): boolean =>
   !formData.registeredModelId || isSubmitDisabledForCommonFields(formData);
 
+export const isRegisterCatalogModelSubmitDisabled = (
+  formData: RegisterCatalogModelFormData,
+): boolean =>
+  !formData.modelRegistry ||
+  !formData.modelName ||
+  isSubmitDisabledForCommonFields(formData) ||
+  !isNameValid(formData.modelName);
```
mturley, line 171: Since this form state extends the register model form state, I think this can just be
```ts
isRegisterModelSubmitDisabled(formData, []) || !formData.modelRegistry
```
and then we should add a comment explaining that the `[]` there is temporary and linking to the future issue for fixing the `isModelNameExisting` validation on the catalog registration page. Or if you prefer, have a `registeredModels` param here and pass in `[]` when calling `isRegisterCatalogModelSubmitDisabled`, and put the comment there.

frontend/src/pages/modelRegistry/screens/components/DeployRegisteredModelModal.tsx
```diff
@@ -51,6 +52,7 @@ const DeployRegisteredModelModal: React.FC<DeployRegisteredModelModalProps> = ({
   const [dataConnections] = useDataConnections(selectedProject?.metadata.name);
   const [connections] = useConnections(selectedProject?.metadata.name, true);
   const error = platformError || projectError;
+  const [rm] = useRegisteredModelById(modelVersion.registeredModelId);
```
mturley, line 55: Same deal here with the loading/error states. We should combine the loading/error here with `deployInfoLoaded` and `deployInfoError` below.

frontend/src/pages/modelRegistry/screens/components/DeployRegisteredModelModal.tsx
```diff
@@ -68,12 +70,21 @@ const DeployRegisteredModelModal: React.FC<DeployRegisteredModelModalProps> = ({
         modelRegistryApi.api,
         modelVersion.id,
         modelVersion.registeredModelId,
+        rm?.customProperties || {},
```
mturley, line 73: Also (and this applies to everywhere we'll pass in the objects to these `bump` utils) since they're not optional, we shouldn't fall back to a blank value like this. We can just narrow the type within the scope of calling these utils so we know they're defined, like by wrapping the call in a `if (rm)`. In this case, we can edit the early return a few lines above this (currently `if (!modelVersion.registeredModelId)`) so it returns early if `rm` isn't defined.

mturley (changes requested):
@ppadti this is looking awesome! tested and it's working great. just a few tweaks I think we need.

gitdallas (approved):

frontend/src/pages/modelCatalog/screens/RegisterCatalogModel.tsx
```diff
@@ -0,0 +1,227 @@
...
+        labels[label] = {
+          // eslint-disable-next-line camelcase
+          string_value: '',
+          metadataType: ModelRegistryMetadataType.STRING,
+        };
+      });
+      setData('modelCustomProperties', labels);
+      setData('versionCustomProperties', {
+        ...labels,
+        Licence: {
```
mturley, line 102: This is misspelled and should be `License` with an S

frontend/src/concepts/modelRegistry/utils/updateTimestamps.ts
```diff
@@ -43,9 +46,9 @@ export const bumpModelVersionTimestamp = async (
 
 export const bumpRegisteredModelTimestamp = async (
   api: MinimalModelRegistryAPI,
-  registeredModelId: string,
+  registeredModels: RegisteredModel,
```
mturley, line 49: This should be singular `registeredModel`

frontend/src/concepts/modelRegistry/utils/updateTimestamps.ts
```diff
@@ -78,11 +82,11 @@ export const bumpRegisteredModelTimestamp = async (
 
 export const bumpBothTimestamps = async (
   api: ModelRegistryAPIs,
-  modelVersionId: string,
-  registeredModelId: string,
+  registeredModels: RegisteredModel,
+  modelVersions: ModelVersion,
```
mturley, line 86: These should be singular `registeredModel` and `modelVersion`

frontend/src/concepts/modelRegistry/utils/updateTimestamps.ts
```diff
@@ -2,15 +2,17 @@ import {
...
+  RegisteredModel,
+  ModelVersion,
 } from '~/concepts/modelRegistry/types';
 
 type MinimalModelRegistryAPI = Pick<ModelRegistryAPIs, 'patchRegisteredModel'>;
 
 export const bumpModelVersionTimestamp = async (
   api: ModelRegistryAPIs,
-  modelVersionId: string,
+  modelVersions: ModelVersion,
```
mturley, line 13: This should be singular `modelVersion`

mturley (approved):
Thanks for these changes @ppadti ! This is in great shape.

I don't want to block merging this so I'll approve, but I notice a few places we should correct spelling. The only one that actually matters for Summit is the `License` property which falls under microcopy, but the plural names for singular objects could be confusing later.

PR #3786, opened 2025-02-20
Feat/18962 catalog model detail

gitdallas:
closes: https://issues.redhat.com/browse/RHOAIENG-18962

## Description
catalog model details page and tests. temporarily using the existing markdownViewer component for the model card. 
image

## How Has This Been Tested?
locally with green cluster

## Test Impact
modified/added a couple tests

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/pages/modelCatalog/screens/ModelDetailsPage.tsx
+++ b/frontend/src/pages/modelCatalog/screens/ModelDetailsPage.tsx
@@ -1,17 +1,32 @@
 import React from 'react';
 import { useParams } from 'react-router';
-import { Breadcrumb, BreadcrumbItem, Button, Divider } from '@patternfly/react-core';
+import {
+  Breadcrumb,
+  BreadcrumbItem,
+  Content,
+  ContentVariants,
+  Flex,
+  FlexItem,
+  Label,
+  Stack,
+  StackItem,
+} from '@patternfly/react-core';
 import { Link } from 'react-router-dom';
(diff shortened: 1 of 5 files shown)


frontend/src/pages/modelCatalog/screens/ModelDetailsPage.tsx
```diff
@@ -46,10 +61,33 @@ const ModelDetailsPage: React.FC = conditionalArea(
...
+                  {model && (
+                    <Label variant="outline" icon={<TagIcon />}>
+                      {getTagFromModel(model)}
+                    </Label>
+                  )}
+                </Flex>
+              </StackItem>
+              {model && (
+                <StackItem>
+                  <Content component={ContentVariants.small}>Provided by {model.provider}</Content>
```
mturley, line 85: Hmm. I would have expected this to use the `source` property of the ModelCatalogSource (currently "Red Hat" in our data), but I didn't realize we also had the model's `provider` property (currently "IBM" in our data). According to the mocks I think we do want these granite models to be "provided by Red Hat", but it probably makes sense to keep this behavior as `model.provider` and change the ConfigMap to say Red Hat instead of IBM. That way we have the flexibility for different "providers" inside one ModelCatalogSource, which is what we'll want when we introduce the "Third Party" source with a Neural Magic provider.

@pboyd do you think it makes sense to change our manifest configmap to have these models say `provider: Red Hat` instead of IBM?

mturley (approved):
Looking great - these pages are coming together!

PR #3799, opened 2025-02-24
initial module federation support

christianvogt:
https://issues.redhat.com/browse/RHOAIENG-19055

## Description

This PR adds support for module federation but has no default configuration. I will be preparing an on cluster demo alongside more instructions on running a local setup. But for now the intent is to ensure that regular dev and production envs continue to work as intended. I don't expect a reviewer to test module federation at this time. 

- Update dotenv to support multiline values
- Use the `MODULE_FEDERATION_CONFIG` env var JSON value to add module federation support
- Adds backend proxy which is only enabled when `MODULE_FEDERATION_CONFIG` is configured. This proxy is used to route frontend requests for both the federated modules and their API requests.
  - There are no default modules configured and therefore the proxy is not configured in production.
- Some config tweaks to support module federation
- Adds script `start:dev:mf` which will include download type definitions from modules. _I'm still experimenting with what is the best approach for working with types._

## How Has This Been Tested?

I tested this with a local instance of model registry running which was modified to expose some APIs. Then in ODH used the following configuration.
```
MODULE_FEDERATION_CONFIG='
[
  {
    "name": "modelRegistry",
    "remoteEntry": "/remoteEntry.js",
    "proxy": [
      {
        "path": "/model-registry/api",
        "pathRewrite": "/api"
      }
    ],
    "local": {
      "host": "localhost",
      "port": 9000
    },
    "service": {
      "name": "model-registry-ui-service",
      "namespace": "kubeflow",
      "port": 80
    }
  }
]
'
```

ODH was modified to consume the exposed API and render a portion of the model registry in the UI.

## Test Impact

N/A

## Request review criteria:

If you have UI changes: 


christianvogt:
I'm using multi line dotenv which doesn't work with the dockerfile which is also consuming dotenv files....

lucferbux (commented):
Changes so far look good. Do we wanna switch the target branch to a feature branch so we can work freely not breaking current config?

christianvogt:
Looks like the rebase has caused an issue I'll need to investigate.

frontend/src/index.tsx
```diff
@@ -1,39 +1 @@
-import React from 'react';
```
lucferbux, line 1: I'm sorry I don't understand this, why using bootstrap here and removing everything from index.tsx, I'm assuming there should be a requirement in module ferderation that the entrypoint should be bootstrap or smth like this, but it's not clear, can we add a comment somewhere to leave some context?

christianvogt: So having the single file like this for some reason is necessary. I can add a comment.

As for the SWC error, that's new to me. I'll rebase and look into it.

PR #3801, opened 2025-02-25
Create OOTB ConfigMap for curated model list data

alexcreasy:
Closes: [RHOAIENG-18955](https://issues.redhat.com/browse/RHOAIENG-18955)
## Description

Adds RHOAI only manifest for deploying ConfigMap for OOTP model list data and associated RBAC.

## How Has This Been Tested?

`kubectl apply` the manifest folder on an ODH cluster and verify the UI

## Request review criteria:

If you have UI changes: 


manifests/rhoai/shared/apps/model-catalog/model-catalog-configmap.yaml
```diff
@@ -0,0 +1,319 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
```
mturley, line 3: we'll need some labels here to make sure this is unmanaged / not reconciled by the operator if it's changed manually.
```yaml
  labels:
    opendatahub.io/managed: "false"
```
cc @andrewballantyne

andrewballantyne: As noted in [slack](https://redhat-internal.slack.com/archives/C07C0447EAV/p1740680228381989?thread_ts=1740671586.369199&cid=C07C0447EAV)... we need a change away from 
```
    opendatahub.io/managed: "false"
```

Drop the label entirely. Happy to explain more in slack.

@alexcreasy FYI

zdtsw: i think what you are looking for is the annoation not the label if you would like operator not manage your configmap which comes from manifests

manifests/rhoai/shared/apps/model-catalog/model-catalog-configmap.yaml
```diff
@@ -0,0 +1,366 @@
...
+  name: model-catalog-source-redhat
+  labels:
+    opendatahub.io/dashboard: "true"
+data:
+  modelCatalogSource: |
+{
```
mturley, line 9: @alexcreasy testing this, I get a parsing error trying to save it with the current spacing. I think the JSON needs to be indented past the `modelCatalogSource: |` line like:

```yaml
data:
  modelCatalogSource: |
    {
      "source": "Red Hat",
```
When I change it to this and save the configmap I get no error.

mturley (approved):
LGTM - tested by editing the configmap on ui-green with the contents here and checking that the role and rolebinding have the same contents (already tested those)

PR #3818, opened 2025-02-27
List only Pull OCI Connections in ConnectionSection

Griffin-Sullivan:
https://issues.redhat.com/browse/RHOAIENG-20228
## Description
Only allow Pull OCI Connections to show up in the Existing Connections part of the ConnectionSection in the deploy kserve model modal. 

## How Has This Been Tested?
* Create a random non-model serving connection type
* Create an instance for S3 connection, URI connection, OCI connection, and the random one you made in a project
* Go to single serving and deploy model
* Existing connections should only show S3, URI, and OCI
* Play around with OCI access type (none selected, pull selected, push & pull selected should show in modal, push only should not show in the modal)
* Make sure other connections deploy properly (S3 and URI)

## Test Impact
Edited utils test to check for OCI connections being serving compatible

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/concepts/connectionTypes/utils.ts
+++ b/frontend/src/concepts/connectionTypes/utils.ts
@@ -178,11 +178,33 @@ const modelServingCompatibleTypesMetadata: Record<
   },
 };
 
-export const isModelServingTypeCompatible = (
+export const isModelServingCompatible = (
   input: string[] | Connection | ConnectionTypeConfigMapObj,
-  type: ModelServingCompatibleTypes,
+  type?: ModelServingCompatibleTypes,
 ): boolean => {
+  if (!type) {
+    return getModelServingCompatibility(input).length > 0;
+  }
   if (isSecretKind(input) && isConnection(input)) {
+    if (type === ModelServingCompatibleTypes.OCI) {
+      const accessType =
+        input.stringData?.ACCESS_TYPE || window.atob(input.data?.ACCESS_TYPE ?? '');
+      if (!(accessType.includes('Pull') || accessType === '[]' || accessType === '')) {
+        return false;
+      }
+    }
+
+    const { managedType } = modelServingCompatibleTypesMetadata[type];
+    if (
+      managedType &&
+      !(
+        input.metadata.annotations['opendatahub.io/connection-type'] === managedType &&
+        input.metadata.labels['opendatahub.io/managed'] === 'true'
+      )
+    ) {
+      return false;
+    }
+
     return modelServingCompatibleTypesMetadata[type].envVars.every((envVar) =>
       Object.keys(input.data || input.stringData || []).includes(envVar),
     );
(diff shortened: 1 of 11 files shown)


frontend/src/concepts/connectionTypes/utils.ts
```diff
@@ -187,7 +187,16 @@ const getModelServingCompatibleTypes = (envVars: string[]): ModelServingCompatib
...
-  getConnectionModelServingCompatibleTypes(connection).length > 0;
+  getConnectionModelServingCompatibleTypes(connection).length > 0 &&
+  isPullOCIAccessType(connection);
+
+export const isPullOCIAccessType = (connection: Connection): boolean => {
+  const accessType = connection.data?.ACCESS_TYPE
+    ? window.atob(connection.data.ACCESS_TYPE)
+    : undefined;
+
+  return accessType === undefined || accessType.includes('Pull') || accessType === '[]';
```
Griffin-Sullivan, line 198: Would love some suggestions for other ways to do this besides `atob`. We have a util for decoding secrets, but it requires you to pass the connection type config map for it which could also get messy. But since I'm using `atob` it makes the check for an empty array a string :disappointed:

frontend/src/pages/modelRegistry/screens/RegisteredModels/usePrefillDeployModalFromModelRegistry.ts
```diff
@@ -20,13 +20,12 @@ const usePrefillDeployModalFromModelRegistry = (
   registeredModelDeployInfo?: RegisteredModelDeployInfo,
 ): [LabeledConnection[], boolean, Error | undefined] => {
   const [fetchedConnections, connectionsLoaded, connectionsLoadError] = useConnections(
-    projectContext ? undefined : createData.project,
+    projectContext ? projectContext.currentProject.metadata.name : createData.project,
```
Griffin-Sullivan, line 23: @manaswinidas @mturley we can wait to merge this until Manaswini makes her changes or if you want to test MR with this PR that works too.

manaswinidas: I can test this with MR once you rebase.

manaswinidas: This change doesn't affect Model registry prefilling. No regressions here.

frontend/src/concepts/connectionTypes/utils.ts
emilys314: Yeah it's pretty difficult to pass in the connection type alongside here in most places this is being consumed. In theory you could, and make it an optional arg. I don't think I like that though.

Another option could be to still use the `parseConnectionValues()` function from this utils file, it will still return the "[]", since if you use the function without giving a connection type, it just returns the `atob()` result. This feels better to me though, since the parsing logic is still in the same place, and not being reduplicated. 

```
export const isPullOCIAccessType = (connection: Connection): boolean => {
  const connectionEnvVars = parseConnectionSecretValues(connection);
  const accessType = connectionEnvVars.ACCESS_TYPE

  return accessType === undefined || accessType.includes('Pull') || accessType === '[]';
}
```

frontend/src/concepts/connectionTypes/utils.ts
```diff
@@ -210,7 +210,16 @@ const getModelServingCompatibleTypes = (envVars: string[]): ModelServingCompatib
   );
 
 export const isModelServingCompatibleConnection = (connection: Connection): boolean =>
```
emilys314, line 212: so only modifying this function doesn't cover showing model serving compatibility here sadly, it uses `getConnectionModelServingCompatibleTypes` 
image

I personally want to refactor this because it's hard to follow with all the different names

emilys314:
i tested it out locally on my cluster and everything works good

looks like https://github.com/opendatahub-io/odh-dashboard/pull/3836 just merged and created a conflict

PR #3823, opened 2025-02-27
fix: add NIM specific API key validation messages

olavtar:
https://issues.redhat.com/browse/NVPE-176

## Description

Added NIM-specific API key validation messages during validation in progress.
image

## How Has This Been Tested?

Tested locally.

## Test Impact

Tested by enabling the NIM application and verifying that the new message is displayed. Also tested a non-NIM application to ensure the general message 'Validating your entries' is shown correctly.

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/pages/exploreApplication/EnableModal.tsx
+++ b/frontend/src/pages/exploreApplication/EnableModal.tsx
@@ -148,7 +148,15 @@ const EnableModal: React.FC<EnableModalProps> = ({ selectedApp, shown, onClose }
                 variant="info"
                 title={
                   <div className="odh-enable-modal__progress-title">
-                    <Spinner size="md" /> Validating your entries
+                    <Spinner size="md" />
+                    {selectedApp.metadata.name === 'nvidia-nim' ? (
+                      <div>
+                        Contacting NVIDIA to validate the license key.
+                        <div>This can take a few minutes.</div>
+                      </div>
+                    ) : (
+                      'Validating your entries'
+                    )}
                   </div>
                 }
                 aria-live="polite"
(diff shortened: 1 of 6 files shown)


frontend/src/pages/exploreApplication/EnableModal.tsx
```diff
@@ -148,7 +148,15 @@ const EnableModal: React.FC<EnableModalProps> = ({ selectedApp, shown, onClose }
                 variant="info"
                 title={
                   <div className="odh-enable-modal__progress-title">
-                    <Spinner size="md" /> Validating your entries
+                    <Spinner size="md" />
+                    {selectedApp.metadata.name === 'nvidia-nim' ? (
```
emilys314, line 152: I don't think it's very scalable here to have specific text for different applications. Could the custom spinner message be more done generically? like `{ customMessage ? {customMessage} : "Validating your entries"}`

Maybe we could have the `useEnableApplication()` return a `validationMessage`? Or another option would be adding a `inProgressText` to the `OdhApplication` which would require modifying  `manifests/rhoai/shared/apps/nvidia-nim/nvidia-nim-app.yaml`

olavtar: I agree, adding it to the `OdhApplication` sounds like a good solution. Since the `useEnableApplication` hook might handle other IntegrationApps, it makes more sense to add it to OdhApplication instead.

emilys314:
Looks good to me

Although I'm slightly confused, I was testing the cluster (not locally on your branch) and it seemed to already have the message, but I don't think it was your changes since the `pre-line` thing doesn't seem to be there? 
image

olavtar:
I needed to see if it works, so I asked Mikhail to update the cluster resource for testing. That’s why you’re already seeing the message there.

PR #3827, opened 2025-02-28
Revert "chore(19957): set rhoai manifest disableModelCatalog flag false"

andrewballantyne:
Reverts opendatahub-io/odh-dashboard#3800

[PM make a call to revert the default state](https://redhat-internal.slack.com/archives/C07J39USZND/p1740759650546239?thread_ts=1740670757.356669&cid=C07J39USZND)


mturley (approved):
Made a note on https://issues.redhat.com/browse/RHOAIENG-19957 that we've reverted the change, and attached this PR there.

Thanks @andrewballantyne !

PR #3833, opened 2025-03-03
Cypress e2e - Button redirect Workaround

antowaddle:
https://issues.redhat.com/browse/RHOAIENG-20471

## Description
- Implement a workaround to the redirect issue we are hitting in [RHOAIENG-20471](https://issues.redhat.com//browse/RHOAIENG-20471)
- Intentionally left [testWorkbenchNegativeTests.cy.ts](https://github.com/opendatahub-io/odh-dashboard/pull/3833/files#diff-1fcbfc2b4bf8c5e57fa6bfad01fa52e3deb0b3f70b3ce4a9703e472d216bcce9) in failing state so that the fix can be tested (when available)
- Cleanup Test Titles

## How Has This Been Tested?
- Locally against a RHOAI and ODH, also crucially on Jenkins
image

## Test Impact
- None - this is a test

## How to run?
Go to odh-dashboard/frontend/src/tests/cypress and run the command npx cypress open . This will open the Cypress UI where all tests can be executed. 

Headless
Execute 'npx cypress run --env grepTags=@Sanity' or 'npx cypress run --env grepTags=@Smoke'

## Request review criteria:

If you have UI changes: 


frontend/src/__tests__/cypress/cypress/tests/e2e/dashboardNavigation/testManifestLinks.cy.ts
```diff
@@ -2,7 +2,7 @@ import * as yaml from 'js-yaml';
 import { isUrlExcluded } from '~/__tests__/cypress/cypress/utils/urlExtractor';
 import { retryableBefore } from '~/__tests__/cypress/cypress/utils/retryableHooks';
 
-describe('[Product bugs: RHOAIENG-9365,RHOAIENG-9187,RHOAIENG-16956, RHOAIENG-16959] Verify that all the URLs referenced in the Manifest directory are operational', () => {
+describe('[Dashboard bugs: RHOAIENG-9365,RHOAIENG-9187,RHOAIENG-16956, RHOAIENG-16959] Verify that all the URLs referenced in the Manifest directory are operational', () => {
```
andrewballantyne, line 5: What was wrong with `Product`? Just because Dashboard has a bug that breaks a test, doesn't mean it's not in the wider product 🤔 Did someone ask you not to use this naming?

frontend/src/__tests__/cypress/cypress/tests/e2e/dataScienceProjects/clusterStorage/testClusterStorageCreation.cy.ts
```diff
@@ -78,7 +78,7 @@ describe('Verify Cluster Storage - Creating, Editing and Deleting', () => {
 
       //Navigate to Cluster Storage and click to Add Storage
       cy.step('Navigate to Cluster Storage and click to create Cluster Storage');
-      projectDetails.findSectionTab('cluster-storages').click();
+      cy.visit(`projects/${projectName}?section=cluster-storages`);
```
andrewballantyne, line 81: Is this wise? I get you're working around quirks that happen where the cypress runner cannot click successfully.. but I feel this is concerning as are testing the cypress navigation and we will miss the user-experience of the navigation goes where we want it to...

frontend/src/__tests__/cypress/cypress/tests/e2e/dashboardNavigation/testManifestLinks.cy.ts
antowaddle: @andrewballantyne This is simply to distinguish between Bugs failing the tests that may not be Dashboard specific, we have some bugs for Model Serving and Pipeline teams.

frontend/src/__tests__/cypress/cypress/tests/e2e/dataScienceProjects/clusterStorage/testClusterStorageCreation.cy.ts
antowaddle: I explained this a little more on Slack, I've tested retries/plugin additions to solve this problem, none were successful. This is not ideal, but I want to reduce the noise that https://issues.redhat.com/browse/RHOAIENG-20471 is causing as it's caused 90% of all our failures over the past week.

frontend/src/__tests__/cypress/cypress/tests/e2e/dashboardNavigation/testManifestLinks.cy.ts
andrewballantyne: I'm wondering if there is value here -- When you log a bug, it could VERY MUCH look like Dashboard is doing something wrong... but we investigate it and find out it's actually the backend team, or cluster state, or you know anything other than a Dashboard fix. I think it's weird to come back to the test and update this string. 

Why not call everything product bugs & just point at the RHOAIENG issues -- the details can be found within'.

Or am I missing some nuance that would be helpful making the distinction?

frontend/src/__tests__/cypress/cypress/tests/e2e/dataScienceProjects/clusterStorage/testClusterStorageCreation.cy.ts
andrewballantyne: I understand the difficulties here -- but I feel uncomfortable with us making this broad based move. I'd have thought a "FIXME" comment or something would be included at each change...

Perhaps it's not doable/worth it? If we move over to this new model, it'll have consequences to our test stability/usefulness in this area. Refactors that shift location of things will not be able to use our tests to verify this. Now I get not having this flake at all, and maybe it's a cypress infrastructure problem... did we find anything online that even acknowledged this issue?

frontend/src/__tests__/cypress/cypress/tests/e2e/dashboardNavigation/testManifestLinks.cy.ts
antowaddle: I'm fine with either approach, it was brought up on one of the e2e calls that we should differentiate on issues for different teams. I'll go with calling everything a Product bug.

frontend/src/__tests__/cypress/cypress/tests/e2e/dataScienceProjects/clusterStorage/testClusterStorageCreation.cy.ts
antowaddle: I investigated this issue thoroughly on Thursday/Friday - see https://issues.redhat.com/browse/RHOAIENG-20471 I spent those days reading through forums trying out different plugins/retries, when I implemented any of them none solved this issue. 

This is 100% an issue relating to Jenkins, I cannot replicate this issue locally (via Cypress UI or heedlessly) and it was never an issue up until 2/3 weeks ago. The function is also used in Mocks and I've never encountered this issue there either. 

Similar to what I mentioned on Slack, I can spend more time to try to figure out if there's other workarounds but until then we will have broken builds every day. Even then, I will likely need to introduce some other hacky way to workaround this problem.

frontend/src/__tests__/cypress/cypress/tests/e2e/dashboardNavigation/testManifestLinks.cy.ts
antowaddle: Resolved in [8038ee4](https://github.com/opendatahub-io/odh-dashboard/pull/3833/commits/8038ee4b5450028fba2ab94e0142d1628fae8519)

frontend/src/__tests__/cypress/cypress/tests/e2e/dataScienceProjects/clusterStorage/testClusterStorageCreation.cy.ts
antowaddle: As per our meeting, adding this temporarily with a comment for it to be removed once we've found root cause.

PR #3838, opened 2025-03-03
FIX : Title and Project Selector Overflow on Data Science Pages 

claudialphonse78:
## Description

1. Closes https://issues.redhat.com/browse/RHOAIENG-12842 Project Selector Overflow onto the next line in the data science pipeline Page in case of a project name with a long name, added a fixed width calculated with calc
Screenshot 2025-03-03 at 5 42 42 PM

**Project Selector Text Overflow** 

2. Resolves Issue related to title overflow onto next line.
Screenshot 2025-03-03 at 10 02 00 PM

**Data Science Pipeline Text Overflow** 
 

## How Has This Been Tested?

Visually adding screenshots for reference.

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/concepts/projects/ProjectSelector.tsx
+++ b/frontend/src/concepts/projects/ProjectSelector.tsx
@@ -112,7 +112,11 @@ const ProjectSelector: React.FC<ProjectSelectorProps> = ({
             height: 'var(--pf-t--global--icon--size--font--2xl)',
           }}
         />
-        <Flex spaceItems={{ default: 'spaceItemsSm' }} alignItems={{ default: 'alignItemsCenter' }} style={{width: 'calc(50% - 1rem / 2)'}}>
+        <Flex
+          spaceItems={{ default: 'spaceItemsSm' }}
+          alignItems={{ default: 'alignItemsCenter' }}
+          style={{ width: 'calc(50% - 1rem / 2)' }}
+        >
           <FlexItem>
             <Bullseye>{selectorLabel}</Bullseye>
           </FlexItem>
--- a/frontend/src/concepts/projects/ProjectSelector.tsx
+++ b/frontend/src/concepts/projects/ProjectSelector.tsx
@@ -115,12 +115,12 @@ const ProjectSelector: React.FC<ProjectSelectorProps> = ({
         <Flex
           spaceItems={{ default: 'spaceItemsSm' }}
           alignItems={{ default: 'alignItemsCenter' }}
-          style={{ width: 'calc(50% - 1rem / 2)' }}
+          flex={{ default: 'flex_2' }}
         >
           <FlexItem>
             <Bullseye>{selectorLabel}</Bullseye>
           </FlexItem>
-          <FlexItem style={{ maxWidth: '90%' }}>{selector}</FlexItem>
+          <FlexItem style={{ maxWidth: 'min(90%, 50vw)' }}>{selector}</FlexItem>
         </Flex>
       </Flex>
     );
--- a/frontend/src/concepts/projects/ProjectSelector.tsx
+++ b/frontend/src/concepts/projects/ProjectSelector.tsx
@@ -112,11 +112,11 @@ const ProjectSelector: React.FC<ProjectSelectorProps> = ({
             height: 'var(--pf-t--global--icon--size--font--2xl)',
           }}
         />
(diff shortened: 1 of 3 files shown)


DaoDaoNoCode (commented):
It looks good! Can you fix 2 more places?
1. On the project list, when the name is too long, it will make it horizontally scrollable. We should truncate the long name instead
Screenshot 2025-03-03 at 5 33 43 PM

2. Add `Truncate` to [this line](https://github.com/opendatahub-io/odh-dashboard/blob/main/frontend/src/pages/modelServing/screens/global/EmptyModelServing.tsx#L40) so it make this page not scrollable horizontally
Screenshot 2025-03-03 at 5 40 47 PM

One more thing:
There is still a slight layout issue if you shrink the window for just a little bit more, see the icon, label and the selector are not on the same line again
Screenshot 2025-03-03 at 5 42 01 PM

claudialphonse78:
1. Added the truncation on the name field for the table 
Table title truncate fix

**Text truncation on the name field in the data science table** 

2. the alignment issue on shrinking the width of the screen

https://github.com/user-attachments/assets/51661907-ad66-4533-b0bf-c5c03155ecf8

DaoDaoNoCode (approved):
Nice work!
/lgtm

frontend/src/concepts/projects/ProjectSelector.tsx
```diff
@@ -112,11 +112,15 @@ const ProjectSelector: React.FC<ProjectSelectorProps> = ({
...
+        <Flex
+          spaceItems={{ default: 'spaceItemsSm' }}
+          alignItems={{ default: 'alignItemsCenter' }}
+          flex={{ default: 'flex_2' }}
+        >
           <FlexItem>
             <Bullseye>{selectorLabel}</Bullseye>
           </FlexItem>
-          <FlexItem>{selector}</FlexItem>
+          <FlexItem style={{ maxWidth: 'min(90%, 50vw)' }}>{selector}</FlexItem>
```
Gkrumbach07, line 123: you should be able to do this without having to do any tricky max width stuff

```tsx
          <FlexItem flex={{ default: 'flex_1' }}>{selector}</FlexItem>
```

frontend/src/pages/projects/screens/detail/ProjectDetails.tsx
```diff
@@ -56,9 +56,9 @@ const ProjectDetails: React.FC = () => {
...
         <Flex spaceItems={{ default: 'spaceItemsSm' }} alignItems={{ default: 'alignItemsCenter' }}>
           <HeaderIcon type={ProjectObjectType.projectContext} sectionType={SectionType.general} />
-          <FlexItem>
+          <FlexItem style={{ maxWidth: 'calc(100% - var(--pf-t--global--spacer--2xl))' }}>
             <ResourceNameTooltip resource={currentProject} wrap={false}>
-              {displayName}
+              <Truncate content={displayName} />
             </ResourceNameTooltip>
           </FlexItem>
         </Flex>
```
Gkrumbach07, line 64: you should be able to set the width properly with just flex props and truncate

```tsx
<Flex
          flex={{ default: 'flex_1' }}
        >
          <HeaderIcon type={ProjectObjectType.projectContext} sectionType={SectionType.general} />
          <FlexItem flex={{ default: 'flex_1' }}>
            <ResourceNameTooltip resource={currentProject} wrap={false}>
              <Truncate content={displayName}>{displayName}</Truncate>
            </ResourceNameTooltip>
          </FlexItem>
        </Flex>
```

Gkrumbach07 (changes requested):
One of the coding practices we try to abide by is to avoid custom styling and to work within patternfly. You will notice when a custom style is needed we try to add a reason why we did it and possibly associate a patternfly ticket to it. In this case we may be able to avoid the custom styles

frontend/src/concepts/projects/ProjectSelector.tsx
claudialphonse78: Thanks for the feedback! I’ll keep this in mind moving forward and avoid custom styling unless absolutely necessary.

Gkrumbach07: The main reason is that the closer we are to patternfly, the easier we can upgrade to new patternfly versions and fixes... which is why we try to paste those PF github issues in the comments

Gkrumbach07 (approved):
/lgtm

PR #3841, opened 2025-03-03
Feat/20078 model detail fine tune popover

gitdallas:
https://issues.redhat.com/browse/RHOAIENG-20078

## Description
image
image

"Tune this model" action item on the model catalog detail page, popover will have option to register or link to learn more

## How Has This Been Tested?
tested locally on green cluster with disableFineTuning flag flipped

## Test Impact
added a test

## Request review criteria:

If you have UI changes: 


mturley:
Can you take a look here @yih-wang ?

frontend/src/pages/modelCatalog/screens/ModelDetailsPage.tsx
```diff
@@ -44,6 +49,7 @@ const ModelDetailsPage: React.FC = conditionalArea(
   const decodedParams = decodeParams(params);
   const { modelRegistryServices, modelRegistryServicesLoaded, modelRegistryServicesLoadError } =
     React.useContext(ModelRegistrySelectorContext);
+  const tuningAvailable = !useIsAreaAvailable(SupportedArea.FINE_TUNING).status;
```
mturley, line 52: This value is the reverse of what we want - the `!` here makes this true of tuning is not available. You can see and change the values of the feature flags by putting `?devFeatureFlags` at the end of the URL and clicking the override link that appears, it was disabled on ui-green. I just enabled it for us.

frontend/src/pages/modelCatalog/screens/ModelDetailsPage.tsx
```diff
@@ -113,33 +182,12 @@ const ModelDetailsPage: React.FC = conditionalArea(
...
-            onClick={() => navigate(registerCatalogModel(params))}
-          >
-            Register model
-          </Button>
-        ))
+        loaded && (
+          <ActionList>
+            {tuningAvailable && fineTuneActionItem}
+            {registerModelButton()}
+          </ActionList>
```
mturley, line 189: The spacing between these buttons is more than I would expect. Looking at the [action list PF docs](https://www.patternfly.org/components/action-list#action-list-single-group), I think we need an `<ActionListGroup>` in here even though there's only one group, and then maybe `<ActionListItem>`s around the buttons (when I edit that docs example to have the buttons as direct children of the ActionList I get this same extra spacing). We probably want that for both of the ActionLists here (this one and the one in the popover).

mturley (changes requested):
Looking good, just a couple details

frontend/src/pages/modelCatalog/screens/ModelDetailsPage.tsx
gitdallas: think that may have actually been a sloppy lint fix of some sort, 😕 i was using the flag.

mturley (approved):
Good stuff, thanks @gitdallas

yih-wang:
Weird... I thought I left my `LGTM` but I wasn't. Anyway it looks good and is merged ;P

PR #3842, opened 2025-03-04
Registered model link pipeline run and artifacts

dpanshug:
[RHOAIENG-19015](https://issues.redhat.com/browse/RHOAIENG-19015)
## Description

Pipeline run created with model info stored in artifacts will display the link of pipeline run/artifacts with registered model 

1. Pipeline run list
Screenshot 2025-03-04 at 1 11 55 PM

2. Pipeline run details
Screenshot 2025-03-04 at 1 09 39 PM
Screenshot 2025-03-04 at 1 09 50 PM

3. Artifact list
Screenshot 2025-03-04 at 1 10 47 PM

4. Artifact details
Screenshot 2025-03-04 at 1 10 02 PM

## How Has This Been Tested?
1. Run a pipeline with model registered info ([registered_model_pipeline.yaml](https://drive.google.com/file/d/1z_XbrSA_wZl9cUinBKOX_YNEca6r_-1R/view?usp=drive_link))
2. Check the pipeline run list, details, artifact list, artifact details page.

## Request review criteria:

If you have UI changes: 

@yannnz

### DIFF
--- a/frontend/src/concepts/pipelines/content/PipelineRunTypeLabel.tsx
+++ b/frontend/src/concepts/pipelines/content/PipelineRunTypeLabel.tsx
@@ -6,8 +6,13 @@ import { PipelineRecurringRunKF, PipelineRunKF } from '~/concepts/pipelines/kfTy
 type PipelineRunTypeLabelProps = {
   run: PipelineRunKF | PipelineRecurringRunKF;
   isCompact?: boolean;
+  isModelRegistered?: boolean;
 };
-const PipelineRunTypeLabel: React.FC<PipelineRunTypeLabelProps> = ({ run, isCompact }) => (
+const PipelineRunTypeLabel: React.FC<PipelineRunTypeLabelProps> = ({
+  run,
+  isCompact,
+  isModelRegistered,
+}) => (
   <>
     {run.recurring_run_id ? (
       <>
@@ -24,6 +29,14 @@ const PipelineRunTypeLabel: React.FC<PipelineRunTypeLabelProps> = ({ run, isComp
             {PipelineRunLabels.ONEOFF}
           </Label>
         </Tooltip>
+      </>
+    )}
+    {isModelRegistered && (
+      <>
+        {' '}
+        <Label color="green" isCompact={isCompact}>
+          Model registered
+        </Label>
       </>
     )}
   </>
--- a/frontend/src/concepts/pipelines/content/pipelinesDetails/PipelineDetailsTitle.tsx
+++ b/frontend/src/concepts/pipelines/content/pipelinesDetails/PipelineDetailsTitle.tsx
@@ -8,12 +8,14 @@ type RecurringRunTitleProps = {
   run: PipelineRunKF;
   statusIcon?: boolean;
   pipelineRunLabel?: boolean;
+  isRegistered?: boolean;
 };
 
 const PipelineDetailsTitle: React.FC<RecurringRunTitleProps> = ({
   run,
   statusIcon,
   pipelineRunLabel,
+  isRegistered,
 }) => {
   const { icon, label, color } = computeRunStatus(run);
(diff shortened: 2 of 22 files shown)


DaoDaoNoCode:
The functionality works great, however, I am a little bit worried about the performance. I got a lot of MLMD calls because of `usePipelineRunArtifactModelData`, you will generate 4 more API calls for each row (1 to fetch the version and 3 others to fetch the MLMD context, executions and artifacts). We should look into reusing the existing result of the API calls and try not to re-fetch the data that's already fetched.
Back to the performance part, I found we made a lot of calls to `getContextByTypeAndName`, I am wondering if we could change that to `getContextsByType` and do the filter by name locally, but that needs more investigation.

Screenshot 2025-03-04 at 2 23 50 PM

DaoDaoNoCode:
Following the previous comment, I have tested that we can do just one request to `getContextsByType` and do a filtering by `result.getContextsList()?.find((r) => r.getName() === name)` to get the real context that we want to fetch instead of making a single `getContextByTypeAndName` every time when you want to fetch a context for a single run.
I think we should look into making MLMD React context provider for the page and expose some filter functions. We need to do as many client filters as we can to prevent API call numbers.
Any thoughts? cc @Gkrumbach07

Gkrumbach07:
@dpanshug @DaoDaoNoCode Yeah i thought we could get away without some sort of caching but it seems to be a little too much. 

I am thinking we create some kind of MLMD context as suggested by @DaoDaoNoCode, however fetching all the run contexts may be an issue due to the response being paginated and us having to fetch all runs that we may not need. Implementation wise we could go a lot of different directions. I know Christian played around with some sort of caching at the network level, which could be a option but may be too complex for our summit timeline. 

I say we implement some short term caching, maybe just dropping a object of type
```tsx
type MLMDContextCache = {
   artifacts: Record<string, Artifact> // key is run id
   executions: ... // same thing
   ... anything else
}
```

then fill that context with this hook
https://gist.github.com/Gkrumbach07/f9e1d86daa267219fa73f2c69c634751

that uses what juntao was saying. I tested it and all it needs is three calls: get contexts, get artifacts, get executions

yannnz:
How does the model version links look like for the use case that users don't have access to click to links's pages?

This is the [design link](https://www.figma.com/proto/Ieu1d5p9LN1OOe51PDX9IF/Pipelines-Summit-2025?page-id=0%3A1&node-id=292-15929&viewport=-2034%2C5963%2C0.19&t=hALg5Ma1V3Acfuf7-1&scaling=min-zoom&content-scaling=fixed&starting-point-node-id=169%3A13422) for this case. 

Just FYI the content of the tooltip has been updated.

Screenshot 2025-03-06 at 20 37 52

kaedward:
@yannnz I think that using a plural here could be confusing, especially if the more common use case is that only one model will be registered. I can see the user seeing "models registered" and thinking "wait, more than 1? Where are the others?" If we could have this be dynamic based on the number of models registered (1 or more than 1), I think it could work.

Gkrumbach07:
@yannnz > How does the model version links look like for the use case that users don't have access to click to links's pages?

We are not doing that for the summit deliverable (talking between features i.e. pipelines -> model registry). We want to just be working with the data that is given to us in the artifact

dpanshug:
@Gkrumbach07 @DaoDaoNoCode I've optimized the hook more, let me know what you think

Gkrumbach07 (approved):
@dpanshug this looks good. just one thing i forgot, which we can follow up on in another jira is that this should probably be behind the Model registry feature flag, otherwise we are linking them to a area that is not enabled

i will drop that jira here momentarily

/lgtm

Gkrumbach07:
follow up: https://issues.redhat.com/browse/RHOAIENG-21246

PR #3853, opened 2025-03-04
Fix selecting all projects for global model page

emilys314:
Closes https://issues.redhat.com/browse/RHOAIENG-20051

## Description

The `preferredProject` is a thing in the `ProjectsContext` to so a user doesn't always have to reselect their project. It's basically the last project you visited. The global model serving page used `useSyncPreferredProject` and would set the preferred project to the current project you are on (the project name in the url). 
If you tried to set `preferredProject` to something else (like null all projects), it would be overwritten by the current page you are on, never changing. The `preferredProject` logic navigation is used when you nav to the global `/modelServing` without any project, so this is why it was only happening here.

image

## How Has This Been Tested?

1. go to `/modelServing` (it should be defaulting to "All projects")
2. select a project
3. Select "All projects"
4. It should have gone to "All projects" (the before it would be stuck at the project you are at)

## Test Impact

No changes, it's a really small hook change, and there are no existing tests for the hook

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/concepts/projects/ProjectSelectorNavigator.tsx
+++ b/frontend/src/concepts/projects/ProjectSelectorNavigator.tsx
@@ -19,7 +19,8 @@ const ProjectSelectorNavigator: React.FC<ProjectSelectorProps> = ({
     <ProjectSelector
       {...projectSelectorProps}
       onSelection={(projectName) => {
-        updatePreferredProject(projects.find(byName(projectName)) || null);
+        const match = projectName ? projects.find(byName(projectName)) ?? null : null;
+        updatePreferredProject(match);
         navigate(getRedirectPath(projectName));
       }}
       namespace={namespace ?? ''}
--- a/frontend/src/concepts/projects/ProjectSelectorNavigator.tsx
+++ b/frontend/src/concepts/projects/ProjectSelectorNavigator.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react';
 import { useNavigate, useParams } from 'react-router-dom';
-import { ProjectsContext } from '~/concepts/projects/ProjectsContext';
+import { byName, ProjectsContext } from '~/concepts/projects/ProjectsContext';
 import ProjectSelector from './ProjectSelector';
 
 type ProjectSelectorProps = {
@@ -13,15 +13,13 @@ const ProjectSelectorNavigator: React.FC<ProjectSelectorProps> = ({
 }) => {
   const navigate = useNavigate();
   const { namespace } = useParams();
-  const { updatePreferredProject } = React.useContext(ProjectsContext);
+  const { projects, updatePreferredProject } = React.useContext(ProjectsContext);
 
   return (
     <ProjectSelector
       {...projectSelectorProps}
       onSelection={(projectName) => {
-        if (!projectName) {
-          updatePreferredProject(null);
-        }
+        updatePreferredProject(projects.find(byName(projectName)) || null);
         navigate(getRedirectPath(projectName));
       }}
       namespace={namespace ?? ''}
--- a/frontend/src/pages/modelServing/ModelServingContext.tsx
+++ b/frontend/src/pages/modelServing/ModelServingContext.tsx
@@ -21,7 +21,6 @@ import { useContextResourceData } from '~/utilities/useContextResourceData';
 import { useDashboardNamespace } from '~/redux/selectors';
 import { DataConnection } from '~/pages/projects/types';
 import useDataConnections from '~/pages/projects/screens/detail/data-connections/useDataConnections';
-import useSyncPreferredProject from '~/concepts/projects/useSyncPreferredProject';
 import { byName, ProjectsContext } from '~/concepts/projects/ProjectsContext';
 import { conditionalArea, SupportedArea } from '~/concepts/areas';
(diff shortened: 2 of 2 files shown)


frontend/src/concepts/projects/useSyncPreferredProject.tsx
andrewballantyne, line 1: This solution got me thinking there was something funky going on -- I wonder if it's the preference selection in the Model Serving context. 

I would think if we just remove [this line](https://github.com/opendatahub-io/odh-dashboard/blob/main/frontend/src/pages/modelServing/ModelServingContext.tsx#L84), we end up with a working solution without touching this file. 

Based on debugging -- it's the "[on click](https://github.com/opendatahub-io/odh-dashboard/blob/main/frontend/src/concepts/projects/ProjectSelectorNavigator.tsx#L21-L26)" triggering -- then this hook undoing it. Which appears to be the Context updating after the hook and resetting the project to the specific name.

Some logging experience gets you (before any change):
```
// From "all projects" to "a project"
>> onClick; a-project
>> ModelServingContext;
  static namespace: a-project ,
  preferred namespace: undefined
>> useSyncPreferred useEffect; // inside the if statement
  preferred namespace: undefined

// from "a project" to "all projects"
>> onClick; ''
>> ModelServingContext;
  static namespace: a-project ,
  preferred namespace: undefined
>> useSyncPreferred useEffect; 
  preferred namespace: undefined
```

After removal of the context `useSyncPreferredProject` line:
```
// From "all projects" to "a project"
>> onClick; a-project
>> ModelServingContext;
  static namespace: a-project ,
  preferred namespace: undefined

// from "a project" to "all projects"
>> onClick; ''
>> ModelServingContext;
  static namespace: undefined ,
  preferred namespace: undefined  
```
The preference switch is not invoked. Not sure why we had it in the first place, but the code did shift quite a bit after the fact and possibly I just made this gap in the project dropdown refactor.

andrewballantyne (changes requested):

frontend/src/concepts/projects/useSyncPreferredProject.tsx
emilys314: Yep that's what I saw was happening, with the hook undoing the context update.

Removing `useSyncPreferredProject(project);` from `ModelServingContext.tsx` also works, I tested that out before. But it won't be updating the preferred project anymore when you select a project in the global model serving. I assumed we still wanted the preferred project to be updated so i chose to add the useRef

andrewballantyne: It does though -- right? I updated it and moved to DSP and it still had the same project 🤔 the on-click action registers it...

andrewballantyne: Do you know of a gap here? It does feel like there _should_ be a gap here -- but I can't find it.

emilys314: With the hook commented out in the model serving context
1. In projects select project 1 (set preferred)
2. Go to modelServing (see project 1 is preselected)
3. In modelServing change selection to project 2
4. Leave the modelServing
5. Come back, it will preselect project 1 instead of project 2 which was the latest

The hook only sets the preferred project for changing to "All projects" / null
```
onSelection={(projectName) => {
        if (!projectName) {
          updatePreferredProject(null);
        }
        navigate(getRedirectPath(projectName));
      }}
```

emilys314: I guess something like this will also work, and we can remove the hook
```
<ProjectSelector
      {...projectSelectorProps}
      onSelection={(projectName) => {
        if (!projectName) {
          updatePreferredProject(null);
        } else {
          const newSelect = projects.find((p) => p.metadata.name === projectName);
          if (newSelect) {
            updatePreferredProject(newSelect);
          }
        }
        navigate(getRedirectPath(projectName));
      }}
      namespace={namespace ?? ''}
    />
```

andrewballantyne: I'd prefer for that to be the use-case -- less "use sync" hooks the better imo -- they are handy, but they cause a weird double loop.

Make sure you use the selector utility in the Project Context file `byName` instead

frontend/src/concepts/projects/ProjectSelectorNavigator.tsx
```diff
@@ -13,15 +13,13 @@ const ProjectSelectorNavigator: React.FC<ProjectSelectorProps> = ({
...
+  const { projects, updatePreferredProject } = React.useContext(ProjectsContext);
 
   return (
     <ProjectSelector
       {...projectSelectorProps}
       onSelection={(projectName) => {
-        if (!projectName) {
-          updatePreferredProject(null);
-        }
+        updatePreferredProject(projects.find(byName(projectName)) || null);
```
andrewballantyne, line 22:
```suggestion
        updatePreferredProject(projectName ? projects.find(byName(projectName)) : null);
```
I favour this ever so slightly -- Projects are sorta not well managed and looping over 500 projects looking for `''` seems unnecessary -- so lets just check if it's a value before looking.

andrewballantyne (approved):
/hold 

LGTM -- unhold if you disagree with the requested change

andrewballantyne (approved):

PR #3855, opened 2025-03-05
cache cypress build for improved test execution performance

christianvogt:
https://issues.redhat.com/browse/RHOAIENG-20961

## Description

Separates the cypress build from each cypress test job and caches the result.

This saves approximately 2mins per cypress test job. 
If we can minimize the time it takes to run the longest running cypress tests, we can benefit more. eg the pipelines cypress tests take over 10 minutes.

An additional 2-3 mins is saved by running unit tests in parallel to cypress tests instead of sequentially.

## How Has This Been Tested?

PR checks

## Test Impact

N/A

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -129,8 +129,34 @@ jobs:
             echo "test-groups=[\"default\"]" >> "$GITHUB_OUTPUT"
           fi
 
+  Cypress-Setup:
+    needs: [Setup]
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Restore cache
+        uses: actions/cache@v4
+        with:
+          path: |
+            ~/.cache/Cypress
+            ${{ github.workspace }}/node_modules
+            ${{ github.workspace }}/backend/node_modules
+            ${{ github.workspace }}/frontend/node_modules
+          key: ${{ runner.os }}-${{ env.NODE_VERSION }}-modules-${{ hashFiles('**/package-lock.json') }}
+      - name: Cypress build cache
+        uses: actions/cache@v4
+        id: cypress-build-cache
+        with:
+          path: |
+            ${{ github.workspace }}/frontend/public-cypress
+          key: ${{ runner.os }}-${{ env.NODE_VERSION }}-cypress-build-${{ github.sha }}
+      - name: Build frontend for cypress
+        if: steps.cypress-build-cache.outputs.cache-hit != 'true'
+        run: npm run cypress:server:build:coverage
+        working-directory: ./frontend
+
   Cypress-Mock-Tests:
-    needs: [Setup, Get-Test-Groups, Unit-Tests, Lint]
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -129,8 +129,34 @@ jobs:
...
+            ${{ github.workspace }}/backend/node_modules
+            ${{ github.workspace }}/frontend/node_modules
+          key: ${{ runner.os }}-${{ env.NODE_VERSION }}-modules-${{ hashFiles('**/package-lock.json') }}
+      - name: Cypress build cache
+        uses: actions/cache@v4
+        id: cypress-build-cache
+        with:
+          path: |
+            ${{ github.workspace }}/frontend/public-cypress
+          key: ${{ runner.os }}-${{ env.NODE_VERSION }}-cypress-build-${{ hashFiles('**/package-lock.json') }}
```
christianvogt, line 152: @antowaddle what should I use instead of `${{ hashFiles('**/package-lock.json') }}` to ensure that the cache is tied to this specific run?

christianvogt: Maybe `github.sha` makes the most sense because it'll invalidate this cache anytime we have new changes which is exactly what can affect the build and we need to rebuild.

christianvogt:
Rearranged the job dependency tree to hopefully save a bit more time to run more jobs in parallel.

christianvogt:
@antowaddle this is now ready for review

antowaddle:
@christianvogt 
Can you experiment again with the setup? Currently the entire pipeline takes between 18-22 mins, this change is actually increasing that to [25m 4s](https://github.com/opendatahub-io/odh-dashboard/actions/runs/13684795042/usage)
image

Can the Unit tests run in parallel with the Cypress Mock tests? I agree that we should only be building the Dashboard once, but we need to figure out how by doing this to not increase the overall pipeline run time.

antowaddle (changes requested):
@christianvogt 
Can you experiment again with the setup? Currently the entire pipeline takes between 19-22 mins, this change is actually increasing that to [25m 4s](https://github.com/opendatahub-io/odh-dashboard/actions/runs/13684795042/usage)
image

Can the Unit tests run in parallel with the Cypress Mock tests? I agree that we should only be building the Dashboard once, but we need to figure out how by doing this to not increase the overall pipeline run time.

christianvogt:
@antowaddle you cannot use the overall time reported by github because sometimes a sequential job is waiting to be picked up. You need to look at the theoretical best time by adding up the times from each job block.

If you look at the historical runs you can see that the action for this PR is the shortest of any other job at 18mins and that was even before making the unit tests run in parallel.

image

christianvogt:
@antowaddle i reran the job and it took 13min 33secs :) Note that `cypress-setup` job took 30 mins because it worked from cache whereas it would typically take just over 2.xx mins to build. If it was a new commit, it would have been closer to 16 mins.

image

antowaddle (approved):
/lgtm

PR #3858, opened 2025-03-05
Model registry ilab fine tune button on version details, and kebab action on model versions table

YuliaKrimerman:
Closes [RHOAIENG-19001](https://issues.redhat.com/browse/RHOAIENG-19001) and [RHOAIENG-17782](https://issues.redhat.com/browse/RHOAIENG-17782)

## Description

Enabled Fine tune button on the version details page as well as a kebab action
 

https://github.com/user-attachments/assets/e371bffb-ab53-45bb-8e12-472c4ee6b4e9

https://github.com/user-attachments/assets/9d7c45b2-7e87-47d7-892f-788966738cfc

## How Has This Been Tested?

On shared green cluster. Make sure you have the feature flag on and the configmap.  Go to the version details page and verify that you see the LAB tune button that opens a modal, you can choose juntao as a project option and it will take to to the ilab form.  same on the kebab action for the version 

## Test Impact

Will add tests

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/concepts/modelRegistry/hooks/useModelVersionTuningData.ts
+++ b/frontend/src/concepts/modelRegistry/hooks/useModelVersionTuningData.ts
@@ -1,5 +1,5 @@
 import { useContext } from 'react';
-import type { ModelVersion } from '~/concepts/modelRegistry/types';
+import type { ModelVersion, RegisteredModel } from '~/concepts/modelRegistry/types';
 import useModelArtifactsByVersionId from '~/concepts/modelRegistry/apiHooks/useModelArtifactsByVersionId';
 import { getDisplayNameFromK8sResource } from '~/concepts/k8s/utils';
 import { getServerAddress } from '~/pages/modelRegistry/screens/utils';
@@ -8,17 +8,9 @@ import { ModelRegistrySelectorContext } from '~/concepts/modelRegistry/context/M
 export const useModelVersionTuningData = (
   modelVersionId: string | null,
   modelVersion: ModelVersion | null,
+  registeredModel: RegisteredModel | null,
 ): {
-  tuningData: {
-    modelRegistryName: string;
-    modelRegistryDisplayName: string;
-    registeredModelId: string;
-    registeredModelName: string;
-    modelVersionId: string;
-    modelVersionName: string;
-    inputModelLocationUri: string;
-    outputModelRegistryApiUrl: string;
-  } | null;
+  tuningData: Record<string, string> | null;
   loaded: boolean;
   loadError: Error | null;
 } => {
@@ -39,15 +31,15 @@ export const useModelVersionTuningData = (
 
   return {
     tuningData:
-      modelVersionId && modelVersion && inputModelLocationUri
+      modelVersion && registeredModel
         ? {
             modelRegistryName: registryService?.metadata.name || '',
             modelRegistryDisplayName,
             registeredModelId: modelVersion.registeredModelId,
-            registeredModelName: modelVersion.name,
+            registeredModelName: registeredModel.name,
             modelVersionId: modelVersion.id,
             modelVersionName: modelVersion.name,
-            inputModelLocationUri,
+            inputModelLocationUri: inputModelLocationUri || '',
             outputModelRegistryApiUrl,
           }
         : null,
--- a/frontend/src/concepts/modelRegistry/hooks/useModelVersionTuningData.ts
+++ b/frontend/src/concepts/modelRegistry/hooks/useModelVersionTuningData.ts
@@ -1,0 +1,44 @@
+import useModelArtifactsByVersionId from '~/concepts/modelRegistry/apiHooks/useModelArtifactsByVersionId';
+import { getDisplayNameFromK8sResource } from '~/concepts/k8s/utils';
+import { getServerAddress } from '~/pages/modelRegistry/screens/utils';
+import { useModelRegistryServices } from '~/concepts/modelRegistry/apiHooks/useModelRegistryServices';
+import { ModelRegistrySelectorContext } from '../context/ModelRegistrySelectorContext';
+import { useContext } from 'react';
+import type { ModelVersion } from '~/concepts/modelRegistry/types';
+
+export const useModelVersionTuningData = (
+  modelVersionId: string | null,
+  modelVersion: ModelVersion | null,
+) => {
+  const { preferredModelRegistry, modelRegistryServices } = useContext(ModelRegistrySelectorContext);
+  const registryService = modelRegistryServices.find(
+    (s) => s.metadata.name === preferredModelRegistry?.metadata?.name
+  );
+
+  const [artifacts, loaded, loadError] = useModelArtifactsByVersionId(modelVersionId || undefined);
+
+  const inputModelLocationUri = artifacts?.items?.[0]?.uri;
+  const modelRegistryDisplayName = registryService 
+    ? getDisplayNameFromK8sResource(registryService)
+    : '';
+  const outputModelRegistryApiUrl = registryService 
+    ? getServerAddress(registryService)
+    : '';
+
+  return {
+    tuningData: modelVersionId && modelVersion && inputModelLocationUri
+      ? {
+          modelRegistryName: registryService?.metadata.name || '',
+          modelRegistryDisplayName,
+          registeredModelId: modelVersion.registeredModelId,
+          registeredModelName: modelVersion.name,
+          modelVersionId: modelVersion.id,
+          modelVersionName: modelVersion.name,
+          inputModelLocationUri,
+          outputModelRegistryApiUrl,
+        }
+      : null,
+    loaded,
+    loadError,
+  };
+}; 
--- a/frontend/src/concepts/modelRegistry/hooks/useModelVersionTuningData.ts
+++ b/frontend/src/concepts/modelRegistry/hooks/useModelVersionTuningData.ts
@@ -1,44 +1,58 @@
+import { useContext } from 'react';
+import type { ModelVersion } from '~/concepts/modelRegistry/types';
 import useModelArtifactsByVersionId from '~/concepts/modelRegistry/apiHooks/useModelArtifactsByVersionId';
 import { getDisplayNameFromK8sResource } from '~/concepts/k8s/utils';
 import { getServerAddress } from '~/pages/modelRegistry/screens/utils';
-import { useModelRegistryServices } from '~/concepts/modelRegistry/apiHooks/useModelRegistryServices';
-import { ModelRegistrySelectorContext } from '../context/ModelRegistrySelectorContext';
-import { useContext } from 'react';
-import type { ModelVersion } from '~/concepts/modelRegistry/types';
+import { ModelRegistrySelectorContext } from '~/concepts/modelRegistry/context/ModelRegistrySelectorContext';
 
 export const useModelVersionTuningData = (
   modelVersionId: string | null,
   modelVersion: ModelVersion | null,
-) => {
-  const { preferredModelRegistry, modelRegistryServices } = useContext(ModelRegistrySelectorContext);
+): {
+  tuningData: {
+    modelRegistryName: string;
+    modelRegistryDisplayName: string;
+    registeredModelId: string;
+    registeredModelName: string;
+    modelVersionId: string;
+    modelVersionName: string;
+    inputModelLocationUri: string;
+    outputModelRegistryApiUrl: string;
+  } | null;
+  loaded: boolean;
+  loadError: Error | null;
+} => {
+  const { preferredModelRegistry, modelRegistryServices } = useContext(
+    ModelRegistrySelectorContext,
+  );
   const registryService = modelRegistryServices.find(
-    (s) => s.metadata.name === preferredModelRegistry?.metadata?.name
+    (s) => s.metadata.name === preferredModelRegistry?.metadata.name,
   );
 
   const [artifacts, loaded, loadError] = useModelArtifactsByVersionId(modelVersionId || undefined);
 
-  const inputModelLocationUri = artifacts?.items?.[0]?.uri;
-  const modelRegistryDisplayName = registryService 
+  // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+  const inputModelLocationUri = artifacts?.items[0]?.uri;
+  const modelRegistryDisplayName = registryService
     ? getDisplayNameFromK8sResource(registryService)
     : '';
-  const outputModelRegistryApiUrl = registryService 
-    ? getServerAddress(registryService)
-    : '';
+  const outputModelRegistryApiUrl = registryService ? getServerAddress(registryService) : '';
 
   return {
-    tuningData: modelVersionId && modelVersion && inputModelLocationUri
-      ? {
-          modelRegistryName: registryService?.metadata.name || '',
-          modelRegistryDisplayName,
-          registeredModelId: modelVersion.registeredModelId,
-          registeredModelName: modelVersion.name,
-          modelVersionId: modelVersion.id,
-          modelVersionName: modelVersion.name,
-          inputModelLocationUri,
-          outputModelRegistryApiUrl,
-        }
-      : null,
+    tuningData:
+      modelVersionId && modelVersion && inputModelLocationUri
+        ? {
+            modelRegistryName: registryService?.metadata.name || '',
+            modelRegistryDisplayName,
+            registeredModelId: modelVersion.registeredModelId,
+            registeredModelName: modelVersion.name,
+            modelVersionId: modelVersion.id,
+            modelVersionName: modelVersion.name,
+            inputModelLocationUri,
+            outputModelRegistryApiUrl,
+          }
+        : null,
     loaded,
-    loadError,
+    loadError: loadError || null,
   };
-}; 
+};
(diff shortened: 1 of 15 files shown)


frontend/src/pages/modelRegistry/screens/ModelVersionDetails/ModelVersionDetailsHeaderActions.tsx
```diff
@@ -32,13 +36,27 @@ const ModelVersionsDetailsHeaderActions: React.FC<ModelVersionsDetailsHeaderActi
...
   const navigate = useNavigate();
   const [isOpenActionDropdown, setOpenActionDropdown] = React.useState(false);
   const [isArchiveModalOpen, setIsArchiveModalOpen] = React.useState(false);
   const [isDeployModalOpen, setIsDeployModalOpen] = React.useState(false);
+  const [isLabTuneModalOpen, setIsLabTuneModalOpen] = React.useState(false);
   const tooltipRef = React.useRef<HTMLButtonElement>(null);
 
+  const isKServeOciEnabled = useIsAreaAvailable(SupportedArea.K_SERVE_OCI);
+  const isFineTuningEnabled = useIsAreaAvailable(SupportedArea.FINE_TUNING);
+  // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
```
gitdallas, line 48: seeing a lot of these, but i'm imagining they are just there as some sort of debugging residual? like maybe you hard coded the booleans originally?

YuliaKrimerman: Nope, the lint is complaining that those are unnecessary because they are always truthy and after trying different stuff that's the only way I was able to make it not complain

gitdallas: oh i see, you should actually be using .status at the end of those useIsAreaAvaialble() calls, i did this earlier as well

frontend/src/pages/modelRegistry/screens/ModelVersionDetails/ModelVersionDetailsHeaderActions.tsx
```diff
@@ -50,6 +67,18 @@ const ModelVersionsDetailsHeaderActions: React.FC<ModelVersionsDetailsHeaderActi
       >
         Deploy
       </Button>
+      {/* eslint-disable-next-line @typescript-eslint/no-unnecessary-condition */}
+      {isLabTuneEnabled && (
+        <Button
```
ppadti, line 72: we can remove this eslint part here as well. 
Also can we have `<ActionListItem>` [like mentioned in this PR](https://github.com/opendatahub-io/odh-dashboard/pull/3841#discussion_r1980218357) to reduce the spacing between buttons.

ppadti (changes requested):

frontend/src/pages/modelRegistry/screens/ModelVersionDetails/ModelVersionDetailsHeaderActions.tsx
YuliaKrimerman: I added it like this, and tried without the action List but it still looks the same 

`        <ActionList>
          <ActionListItem
            id="lab-tune-button"
            aria-label="Lab tune version"
          >
            <Button variant="secondary" onClick={() => setIsLabTuneModalOpen(true)}>
              Lab tune
            </Button>
          </ActionListItem>
        </ActionList>`

gitdallas: You'll want to wrap in an ActionItemGroup

YuliaKrimerman: I don't see that type in the code, but I tried wrapping in ActionListGroup and that didn't help aither

gitdallas: Yea I meant that lol.  Should work if you have both items in a single group.  It's list>group>item.. so if you have something right under list it treats it as group spacing

frontend/src/pages/modelRegistry/screens/ModelVersionDetails/ModelVersionDetailsHeaderActions.tsx
```diff
@@ -67,17 +67,16 @@ const ModelVersionsDetailsHeaderActions: React.FC<ModelVersionsDetailsHeaderActi
...
-          aria-label="Lab tune version"
-          variant={ButtonVariant.secondary}
-          onClick={() => setIsLabTuneModalOpen(true)}
-          data-testid="lab-tune-button"
-        >
-          Lab tune
-        </Button>
+      {isFineTuningEnabled && (
+        <ActionListGroup>
+          <ActionList>
```
gitdallas, line 72: Looks like you have a list inside of a group.

frontend/src/pages/modelRegistry/screens/ModelVersionDetails/ModelVersionDetailsHeaderActions.tsx
YuliaKrimerman: I think it's just disliking me - Tried `      <ActionList>
        <ActionListGroup>
            <ActionListItem data-testid="lab-tune-button" aria-label="Lab tune version">
              <Button variant="secondary" onClick={() => setIsLabTuneModalOpen(true)}>
                Lab tune
              </Button>
            </ActionListItem>
        </ActionListGroup>
      </ActionList> ` And still the same :(

gitdallas: Hmm.. thought there were two action components? Each should be in the same group

frontend/src/pages/modelRegistry/screens/ModelVersionDetails/ModelVersionDetailsHeaderActions.tsx
YuliaKrimerman: Screenshot 2025-03-07 at 2 49 18 PM
This is as closest as I got, and that's after wrapping the hollow button in a group of their own and then wrapping again with the filled button

gitdallas: Wrong thread 😜 but they look appropriately spaced there?

YuliaKrimerman: it's Friday, have mercy! But I think yes( biased of course) @yih-wang what do you think ?

yih-wang:
@YuliaKrimerman Forgive me... I just realized I had a typo in the mockup. It should be "LAB" instead of "Lab". And there's also an [in-progress discussion](https://redhat-internal.slack.com/archives/C08BHBUF9R9/p1741296044820309) about whether it should be "LAB-tune" or "LAB tune". If there's not a conclusion before this PR is merged, we can defer it to the overall microcopy update ticket.

frontend/src/__tests__/cypress/cypress/tests/mocked/application.cy.ts
```diff
@@ -76,12 +76,12 @@ describe('Application', () => {
       dataScienceStackComponentMap[DataScienceStackComponent.MODEL_REGISTRY],
     );
     row1.find().findByText('Kubeflow Model Registry').should('exist');
-    row1.find().findByText('1.14.0').should('exist');
+    row1.find().findByText('v0.2.13').should('exist');
```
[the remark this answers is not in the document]
mturley: I'm concerned about this change being part of your PR in case it disrupts something else. Can you try reversing the changes to this file and then rebasing?

frontend/src/pages/modelRegistry/screens/ModelVersionDetails/ModelVersionDetailsHeaderActions.tsx
```diff
@@ -32,75 +38,119 @@ const ModelVersionsDetailsHeaderActions: React.FC<ModelVersionsDetailsHeaderActi
...
+        </ActionListItem>
+      </ActionListGroup>
+      {isLabTuneModalOpen ? (
+        <StartRunModal
+          onCancel={() => setIsLabTuneModalOpen(false)}
+          onSubmit={(selectedProject) => {
+            navigate(
+              `${getModelCustomizationPath(selectedProject)}?${new URLSearchParams(
+                tuningData!,
+              ).toString()}`,
```
mturley, line 130: This isn't how we want to pass the params - you want to use router history state, not URL params.

```ts
navigate(`${getModelCustomizationPath(selectedProject)}`, { state: tuningData });
```

You can see if it worked correctly by making sure the params are visible in the ilab form. with your current version they show as `-`.

frontend/src/pages/modelRegistry/screens/ModelVersions/ModelVersionsTableRow.tsx
```diff
@@ -157,14 +181,28 @@ const ModelVersionsTableRow: React.FC<ModelVersionsTableRowProps> = ({
...
             />
           ) : null}
+          {tuningModelVersionId && tuningData && (
+            <StartRunModal
+              onCancel={() => setTuningModelVersionId(null)}
+              onSubmit={(selectedProject) => {
+                navigate(
+                  `${getModelCustomizationPath(selectedProject)}?${new URLSearchParams(
+                    tuningData,
+                  ).toString()}`,
```
mturley, line 199: Same deal here - use the `state` param instead.

frontend/src/__tests__/cypress/cypress/tests/mocked/modelRegistry/modelVersionDetails.cy.ts
```diff
@@ -615,5 +626,21 @@ describe('Model version details', () => {
...
+        mockK8sResourceList([
+          mockProjectK8sResource({
+            k8sName: 'data-science-project',
+            displayName: 'Data Science Project',
+            isDSProject: true,
+          }),
+        ]),
+      );
+      modelVersionDetails.findLabTuneButton().click();
+      modelVersionDetails.findStartRunModal().should('exist');
```
mturley, line 643: Ideally these tests should go all the way to submitting the modal and checking that the params made it into the ilab form as expected. Since the params are passed in router state we can't look in the URL for them, but we can at least check that we can find the text on the page for some of them: model registry name, model name, version name, URI.

frontend/src/__tests__/cypress/cypress/tests/mocked/modelRegistry/modelVersions.cy.ts
```diff
@@ -271,4 +271,51 @@ describe('Model Versions', () => {
...
+      },
+      mockModelVersion({ id: '1', name: 'model version' }),
+    ).as('getTuningData');
+
+    modelRegistry.visit();
+    const registeredModelRow = modelRegistry.getRow('Fraud detection model');
+    registeredModelRow.findName().contains('Fraud detection model').click();
+
+    const modelVersionRow = modelRegistry.getModelVersionRow('model version');
+    modelVersionRow.findKebabAction('LAB tune').click();
```
mturley, line 319: Same here, we should submit the modal and check for the params showing up on the ilab form. It's important to test from both pages since they get the info differently.

frontend/src/concepts/modelRegistry/hooks/useModelVersionTuningData.ts
```diff
@@ -0,0 +1,57 @@
...
+  loadError: Error | null;
+} => {
+  const { preferredModelRegistry, modelRegistryServices } = useContext(
+    ModelRegistrySelectorContext,
+  );
+  const registryService = modelRegistryServices.find(
+    (s) => s.metadata.name === preferredModelRegistry?.metadata.name,
+  );
+
+  const [artifacts, loaded, loadError] = useModelArtifactsByVersionId(modelVersionId || undefined);
```
mturley, line 32: Hmm. This is fine for now, but this means on the version details page we are fetching the artifacts twice (they also get fetched in ModelVersionDetailsView). We probably eventually want to refactor so these are lifted out to the page level or the context level. I'll make a note to follow up with a task for that, but this works fine for now.

mturley (changes requested):
Looking great, a couple of small things.

frontend/src/__tests__/cypress/cypress/tests/mocked/modelRegistry/modelVersionDetails.cy.ts
YuliaKrimerman: I was trying to make that, but the I couldn't make the test load the correct project and enable the submit button, and actually get to the final page. Any intercepts that you think I need for that? Same for the test below.

frontend/src/concepts/modelRegistry/hooks/useModelVersionTuningData.ts
```diff
@@ -0,0 +1,57 @@
...
+  const outputModelRegistryApiUrl = registryService ? getServerAddress(registryService) : '';
+
+  return {
+    tuningData:
+      modelVersionId && modelVersion && inputModelLocationUri
+        ? {
+            modelRegistryName: registryService?.metadata.name || '',
+            modelRegistryDisplayName,
+            registeredModelId: modelVersion.registeredModelId,
+            registeredModelName: modelVersion.name,
```
mturley, line 47: `registeredModelName` here needs to be set to the model name, not the version name. The hook needs a `registeredModel` passed into it, which means you need to pass that down from ModelVersionDetails into ModelVersionDetailsHeaderActions so it's in scope there, and also pass it down from ModelVersionListView into ModelVersionsTable -> ModelVersionsTableRow so it's in scope there.

Maybe another case for future refactoring so the model is in context.... 🤔

mturley (changes requested):
One more issue with the tuning data

frontend/src/concepts/modelRegistry/hooks/useModelVersionTuningData.ts
YuliaKrimerman: done

frontend/src/concepts/modelRegistry/hooks/useModelVersionTuningData.ts
```diff
@@ -0,0 +1,49 @@
...
+import { getDisplayNameFromK8sResource } from '~/concepts/k8s/utils';
+import { getServerAddress } from '~/pages/modelRegistry/screens/utils';
+import { ModelRegistrySelectorContext } from '~/concepts/modelRegistry/context/ModelRegistrySelectorContext';
+
+export const useModelVersionTuningData = (
+  modelVersionId: string | null,
+  modelVersion: ModelVersion | null,
+  registeredModel: RegisteredModel | null,
+): {
+  tuningData: Record<string, string> | null;
```
mturley, line 13: This was useful for you to have as the full type:
```ts
tuningData: {
    modelRegistryName: string;
    modelRegistryDisplayName: string;
    registeredModelId: string;
    registeredModelName: string;
    modelVersionId: string;
    modelVersionName: string;
    inputModelLocationUri: string;
    outputModelRegistryApiUrl: string;
  } | null;
```

I imagine you changed it to `Record<string, string>` because that's what the navigate function expects for state - I would rather us instead have this strongly typed so you're enforcing what you return here, and then where you call navigate you can do:
```ts
if (tuningData) {
  const navigateState: Record<string, string> = tuningData;
  navigate(getModelCustomizationPath(selectedProject), { state: tuningData });
}
```
Your strict object should be assignable to a Record like this, and wrapping it in `if (tuningData)` ensures we're not taking the user to the ilab page without params (I should have caught this before).

frontend/src/concepts/modelRegistry/hooks/useModelVersionTuningData.ts
```diff
@@ -39,15 +31,15 @@ export const useModelVersionTuningData = (
...
         ? {
             modelRegistryName: registryService?.metadata.name || '',
             modelRegistryDisplayName,
             registeredModelId: modelVersion.registeredModelId,
-            registeredModelName: modelVersion.name,
+            registeredModelName: registeredModel.name,
             modelVersionId: modelVersion.id,
             modelVersionName: modelVersion.name,
-            inputModelLocationUri,
+            inputModelLocationUri: inputModelLocationUri || '',
```
mturley, line 42: Why the change here? Ilab won't work if we don't pass a URI. I think it was correct before where we should only return tuningData if it has what we need. Same with registryService actually, let's make sure everything is defined before we return anything here.

mturley (changes requested):
One more round 😄

frontend/src/pages/modelRegistry/screens/ModelVersionDetails/ModelVersionDetailsHeaderActions.tsx
manaswinidas: I didn't realise [RHOAIENG-17782](https://issues.redhat.com/browse/RHOAIENG-17782) is addressed here since the ticket wasn't assigned. Feel free to close #3896 once this PR goes in.

mturley (approved):
LGTM, good stuff @YuliaKrimerman !

manaswinidas:
@YuliaKrimerman PR description not updated

PR #3860, opened 2025-03-06
Link pipeline run from MR version details page

ppadti:
Closes: [RHOAIENG-19017](https://issues.redhat.com/browse/RHOAIENG-19017)

## Description
This PR aims to add link to pipeline run from MR version details page.

https://github.com/user-attachments/assets/e9e65a48-5b31-4208-a681-829c1723db7b

## How Has This Been Tested?
 - Register a version with a reference to a pipelinerun. You can do this by actually running the InstructLab pipeline, or you can manually add customProperties referencing a pipelinerun. See the issue description for the relevant properties.
 - On the version details page, verify that you see the "Registered from" link as shown in the screenshot above and the customProperties referencing a pipelinerun should not present in the properties table, as we filtering them out.
 - Click the link to the run and verify that you are taken to the run details page.
 - Look at a version that was not registered from a pipeline, and verify that the "Registered from" section is not shown.

## Test Impact
Added cypress and unit tests.

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/pages/modelRegistry/screens/ModelVersionDetails/ModelVersionDetailsView.tsx
+++ b/frontend/src/pages/modelRegistry/screens/ModelVersionDetails/ModelVersionDetailsView.tsx
@@ -10,12 +10,18 @@ import {
   Spinner,
   Alert,
 } from '@patternfly/react-core';
+import { Link } from 'react-router';
 import { ModelVersion } from '~/concepts/modelRegistry/types';
 import DashboardDescriptionListGroup from '~/components/DashboardDescriptionListGroup';
 import EditableTextDescriptionListGroup from '~/components/EditableTextDescriptionListGroup';
 import { EditableLabelsDescriptionListGroup } from '~/components/EditableLabelsDescriptionListGroup';
 import ModelPropertiesDescriptionListGroup from '~/pages/modelRegistry/screens/ModelPropertiesDescriptionListGroup';
-import { getLabels, mergeUpdatedLabels } from '~/pages/modelRegistry/screens/utils';
+import {
+  getLabels,
+  getProperties,
+  isPipelineRunExist,
+  mergeUpdatedLabels,
+} from '~/pages/modelRegistry/screens/utils';
 import useModelArtifactsByVersionId from '~/concepts/modelRegistry/apiHooks/useModelArtifactsByVersionId';
 import { ModelRegistryContext } from '~/concepts/modelRegistry/context/ModelRegistryContext';
 import ModelTimestamp from '~/pages/modelRegistry/screens/components/ModelTimestamp';
@@ -26,6 +32,9 @@ import {
   bumpRegisteredModelTimestamp,
 } from '~/concepts/modelRegistry/utils/updateTimestamps';
 import useRegisteredModelById from '~/concepts/modelRegistry/apiHooks/useRegisteredModelById';
+import { globalPipelineRunDetailsRoute } from '~/routes';
+import { ProjectObjectType, typedObjectImage } from '~/concepts/design/utils';
+import { pipelineRunSpecificKeys } from './const';
 
 type ModelVersionDetailsViewProps = {
   modelVersion: ModelVersion;
@@ -44,6 +53,7 @@ const ModelVersionDetailsView: React.FC<ModelVersionDetailsViewProps> = ({
   const modelArtifact = modelArtifacts.items.length ? modelArtifacts.items[0] : null;
   const { apiState } = React.useContext(ModelRegistryContext);
   const storageFields = uriToStorageFields(modelArtifact?.uri || '');
+  const filteredProperties = getProperties(mv.customProperties);
   const [registeredModel, registeredModelLoaded, registeredModelLoadError, refreshRegisteredModel] =
     useRegisteredModelById(mv.registeredModelId);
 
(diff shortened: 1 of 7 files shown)


ppadti:
@yih-wang could you please take a look the screen recording?

yih-wang:
@ppadti The run name and the project name are supposed to be in bold font (see the image below). Otherwise looks good! cc @yannnz 
image

yannnz:
Looks good. Just wondering how the link looks like if the user doesn't have access to click to that link's page. 

This is the [design link](https://www.figma.com/proto/Ieu1d5p9LN1OOe51PDX9IF/Pipelines-Summit-2025?page-id=0%3A1&node-id=363-19435&viewport=-2464%2C5937%2C0.19&t=YceczsbSBYh3KKEj-1&scaling=scale-down-width&content-scaling=fixed&starting-point-node-id=169%3A13422). The content for the popover has just been updated.
Screenshot 2025-03-06 at 20 48 53

ppadti:
> Looks good. Just wondering how the link looks like if the user doesn't have access to click to that link's page.
> 
> This is the [design link](https://www.figma.com/proto/Ieu1d5p9LN1OOe51PDX9IF/Pipelines-Summit-2025?page-id=0%3A1&node-id=363-19435&viewport=-2464%2C5937%2C0.19&t=YceczsbSBYh3KKEj-1&scaling=scale-down-width&content-scaling=fixed&starting-point-node-id=169%3A13422). The content for the popover has just been updated.
> 
> Screenshot 2025-03-06 at 20 48 53

Thanks @yannnz for taking a look at it. We have this issue: [RHOAIENG-21130](https://issues.redhat.com/browse/RHOAIENG-21130) to handle this behavior

frontend/src/pages/modelRegistry/screens/utils.ts
```diff
@@ -176,3 +176,21 @@ export const isValidHttpUrl = (value: string): boolean => {
     return false;
   }
 };
+
+export const filterCustomProperties = (
```
gitdallas, line 180: the way i was filtering on the story i'm working on is just adding a check to updating the `acc` on line 49 like:

`    if (key === '_lastModified' || /^_registeredFromCatalog/.test(key)) {
`

i wonder if instead of getting props and then filtering them, we have different functions for getting various types of props (like we already do for props/labels), maybe `getProperties` just gets the standard custom props that we currently show in the prop list on the details pages (does the filtering itself, like i mention above).

thoughts?

mturley: (@manaswinidas responding to your confusion on Slack - if it helps @gitdallas is referring to [RHOAIENG-20077](https://issues.redhat.com/browse/RHOAIENG-20077) that he's working on - the two will conflict because they are adding the same "Registered from" section and similar customProperties filtering logic, just this PR is for linking to a pipelinerun and that issue is for linking back to the model catalog)

I do like the type safety we could get out of `getRegisteredFromCatalogProps, getRegisteredFromPipelineProps`. I think it sets us up better for future refactoring.

> for `getProps`, would it be okay to filter out `/^_/`? otherwise add as needed... `/^_registeredFrom/` or both `/^_registeredFromCatalog/` and `/^_registeredFromPipeline/`

Let's not filter out everything that starts with _ in case a user wants to use that for some reason - the naming convention for these temporary properties we're using is `_registeredFrom*` so I think we can filter out based on that.

I think maybe if we're implementing `getRegisteredFromPipelineProps` in this PR that util could take the bare `mv.customProperties` (not the return value of `getProperties`) and then we could filter out all `_registeredFrom*` keys as part of the existing `getProperties` util (i.e. actually have it only return the properties that should be in the Properties table). That would mean we don't need this `filterCustomProperties` function.

frontend/src/pages/modelRegistry/screens/ModelVersionDetails/ModelVersionDetailsView.tsx
```diff
@@ -142,7 +152,48 @@ const ModelVersionDetailsView: React.FC<ModelVersionDetailsViewProps> = ({
...
+                data-testid="registered-from"
+              >
+                <FlexItem data-testid="pipeline-run-link">
+                  Run{' '}
+                  {
+                    <Link
+                      style={{ fontWeight: 'var(--pf-t--global--font--weight--body--bold)' }}
+                      to={globalPipelineRunDetailsRoute(
+                        filteredProperties[pipelineRunSpecificKeys[0]].string_value,
+                        filteredProperties[pipelineRunSpecificKeys[1]].string_value,
```
mturley, line 169: I don't think we should reference these by numeric index, there's probably a safer / more readable way to do this. Like @gitdallas said in his review, maybe we can make a `getRegisteredFromPipelineProperties` function (or however you want to name it) that takes all the customProperties, picks these out as plain strings and returns a typed object so we could reference for example `registeredFromPipelineProperties.project`.

frontend/src/pages/modelRegistry/screens/ModelVersionDetails/const.ts
```diff
@@ -7,3 +7,9 @@ export enum ModelVersionDetailsTabTitle {
   DETAILS = 'Details',
   DEPLOYMENTS = 'Deployments',
 }
+
+export const pipelineRunSpecificKeys: string[] = [
+  '_registeredFromPipelineProject',
+  '_registeredFromPipelineRunId',
+  '_registeredFromPipelineRunName',
+];
```
mturley, line 15: If we're simply filtering out by `_registeredFrom*` and we're writing a `getRegisteredFromPipelineProperties` util, maybe we can just reference these property keys directly inside that util and not expose them anywhere else (we could remove this export). Then we can drive everything from the return values of that util and it'll probably be easier to refactor later to use different properties.

frontend/src/pages/modelRegistry/screens/utils.ts
```diff
@@ -176,3 +176,21 @@ export const isValidHttpUrl = (value: string): boolean => {
...
+      filteredCustomProperties[key] = customProperties[key];
+    }
+  });
+  return filteredCustomProperties;
+};
+
+export const isPipelineRunExist = (
+  customProperties: ModelRegistryCustomProperties,
+  keys: string[],
+): boolean => keys.every((key) => key in customProperties);
```
mturley, line 196: If we're moving to `getRegisteredFromPipelineProps`, we could have it return `null` if all the required properties aren't present and then we wouldn't need `isPipelineRunExist`.

mturley (changes requested):
Working great @ppadti . I'm on the fence here -- we could just merge this and have @gitdallas address this feedback in his PR for https://issues.redhat.com/browse/RHOAIENG-20077 - but if you're up for it I think some changes here would contain the hackiness a bit and set us up a bit better moving forward.

gitdallas (approved):
i think i should be okay taking on that refactoring as part of 20077.

PR #3861, opened 2025-03-06
Deploy models with OCI from model registry

manaswinidas:
Closes: [RHOAIENG-19258](https://issues.redhat.com/browse/RHOAIENG-19258)

## Description

The Deploy model project dropdown should show the info alert and only show projects that have either no model serving platform enabled or single-serving platform enabled(not any modelmesh/multi-serving platform projects). To test this, change any project's model serving platform to multi-model. You should not find this multi-model project in the dropdown.
Screenshot 2025-03-06 at 3 56 07 PM

### Screen recordings/Scenarios to test:
1. Try deploying any Granite model registered from Model catalog, all have OCI URIs containing `oci://registry.redhat.io`, you will find "Current URI" with the read-only OCI - URI. (You can test this by deploying the version inside "granite-8b-code-base-from-catalog" model on the "mdas" project or any project - a project without any connection and/or project without a model serving platform selected("mdas-new" project) - even if there are no connections and model-serving platform selected 
 - you should see "Current URI" preselected and read-only URI filled in [ui-green](https://console-openshift-console.apps.ui-green.dev.datahub.redhat.com/) cluster)
 
https://github.com/user-attachments/assets/fd9a4626-2a4b-483f-8e31-b3cb8fbef342

2. Try deploying any model registered using an OCI - URI but not containing`oci://registry.redhat.io`, you will find "Existing connection" with the fields filled. When you click on the connection name dropdown, you should see the "Recommended" label as shown in the screenshot.(You can test this by deploying the version inside "oci-model" model on the "mdas" project - you should see "Existing connection" preselected and fields filled [ui-green](https://console-openshift-console.apps.ui-green.dev.datahub.redhat.com/) cluster)

https://github.com/user-attachments/assets/7fc4533d-54b9-4dc4-b9f3-4b0bd73bebf4

Screenshot 2025-03-06 at 5 57 18 PM

3. Try deploying any model registered using an OCI - URI but not containing`oci://registry.redhat.io` and not containing any matching connection, you will find "Create connection" with the "Model URI" field filled. (You can test this by deploying the version inside "oci-model" model on the "mdas-new" project - you should see "Create connection" preselected and "Model URI" field filled [ui-green](https://console-openshift-console.apps.ui-green.dev.datahub.redhat.com/) cluster)

https://github.com/user-attachments/assets/400ae062-04b7-4748-bd98-9942ce61fe95

4. Additionally, check all the same test cases as described in https://github.com/opendatahub-io/odh-dashboard/pull/3590 (S3 and URI ones) to make sure they haven't regressed.

## How Has This Been Tested?

Testing instructions added in description along with screenshots/screen recordings

## Test Impact

Added unit tests, Cypress tests are on the way.

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/concepts/modelRegistry/utils.ts
+++ b/frontend/src/concepts/modelRegistry/utils.ts
@@ -10,6 +10,7 @@ export type ObjectStorageFields = {
 export type RegisteredModelLocation = {
   s3Fields: ObjectStorageFields | null;
   uri: string | null;
+  ociUri: string | null;
 } | null;
 
 export const objectStorageFieldsToUri = (fields: ObjectStorageFields): string | null => {
@@ -38,11 +39,18 @@ export const uriToStorageFields = (uri: string): RegisteredModelLocation => {
       const endpoint = searchParams.get('endpoint');
       const region = searchParams.get('defaultRegion');
       if (endpoint && bucket && path) {
-        return { s3Fields: { endpoint, bucket, region: region || undefined, path }, uri: null };
+        return {
+          s3Fields: { endpoint, bucket, region: region || undefined, path },
+          uri: null,
+          ociUri: null,
+        };
       }
       return null;
     }
-    return { s3Fields: null, uri };
+    if (uri.startsWith('oci:')) {
+      return { s3Fields: null, uri: null, ociUri: uri };
+    }
+    return { s3Fields: null, uri, ociUri: null };
   } catch {
     return null;
   }
--- a/frontend/src/pages/modelRegistry/screens/components/DeployRegisteredModelModal.tsx
+++ b/frontend/src/pages/modelRegistry/screens/components/DeployRegisteredModelModal.tsx
@@ -21,6 +21,7 @@ import useDataConnections from '~/pages/projects/screens/detail/data-connections
 import { bumpBothTimestamps } from '~/concepts/modelRegistry/utils/updateTimestamps';
 import useConnections from '~/pages/projects/screens/detail/connections/useConnections';
 import useRegisteredModelById from '~/concepts/modelRegistry/apiHooks/useRegisteredModelById';
+import { isRedHatRegistryUri } from '~/pages/modelRegistry/screens/utils';
 
 interface DeployRegisteredModelModalProps {
   modelVersion: ModelVersion;
@@ -168,7 +169,12 @@ const DeployRegisteredModelModal: React.FC<DeployRegisteredModelModalProps> = ({
         registeredModelDeployInfo={registeredModelDeployInfo}
         projectContext={{ currentProject: selectedProject, connections }}
         projectSection={projectSection}
-        existingUriOption={registeredModelDeployInfo.modelArtifactUri}
+        existingUriOption={
+          registeredModelDeployInfo.modelArtifactUri &&
+          isRedHatRegistryUri(registeredModelDeployInfo.modelArtifactUri)
+            ? registeredModelDeployInfo.modelArtifactUri
+            : undefined
+        }
       />
     );
   }
--- a/frontend/src/pages/modelRegistry/screens/components/DeployRegisteredModelModal.tsx
+++ b/frontend/src/pages/modelRegistry/screens/components/DeployRegisteredModelModal.tsx
@@ -121,6 +121,7 @@ const DeployRegisteredModelModal: React.FC<DeployRegisteredModelModalProps> = ({
         ) : isOciModel ? (
           <FormSection title="Model deployment">
             <Alert
+              data-testid="oci-deploy-kserve-alert"
               variant="info"
               isInline
               title="This model uses an OCI storage location which supports deploying to only the single-model serving platform. Projects using the multi-model serving platform are excluded from the project selector."
--- a/frontend/src/pages/modelRegistry/screens/components/DeployRegisteredModelModal.tsx
+++ b/frontend/src/pages/modelRegistry/screens/components/DeployRegisteredModelModal.tsx
@@ -47,11 +47,8 @@ const DeployRegisteredModelModal: React.FC<DeployRegisteredModelModalProps> = ({
     selectedProject,
     servingPlatformStatuses,
   );
-  const { loaded: projectDeployStatusLoaded, error: projectError } =
-    useProjectErrorForRegisteredModel(selectedProject?.metadata.name, platform);
   const [dataConnections] = useDataConnections(selectedProject?.metadata.name);
   const [connections] = useConnections(selectedProject?.metadata.name, true);
-  const error = platformError || projectError;
   const [registeredModel, registeredModelLoaded, registeredModelLoadError, refreshRegisteredModel] =
     useRegisteredModelById(modelVersion.registeredModelId);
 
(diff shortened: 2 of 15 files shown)


manaswinidas:
@yih-wang Can you have a look at the screen shots and the screen recordings in the PR description?

yih-wang:
Thanks for the clear scenarios and recordings @manaswinidas! I got several comments, but not sure any of them should belong to this PR though... 
- I realized we need loading states for the connection part. It makes me confused when seeing
[**00:09 of the 2nd video**] </br>image
and [**00:13 of the 3rd video**] until everything rendered correctly... </br>image
Maybe we should show a spinner for the whole connection section before query ends. @simrandhaliw @vconzola WDYT?

- I don't see how `recommends` working here. But reading through [RHOAIENG-19258](https://issues.redhat.com/browse/RHOAIENG-19258) 
> model URIs starting with "oci:" will cause connections to be labeled as recommended only if:
> - The connection is of the OCI connection type (not the plain URI type) AND
> - The connection's OCI_HOST value is a substring of the model URI (e.g. if the connection's OCI_HOST is quay.io and the model we're deploying has the URI "oci://quay.io/foo/bar..." that connection will be recommended). It's important to do this as just a substring check since we cannot make assumptions about the structure of OCI URIs, so we shouldn't try to parse / split them.

And seeing relevant threads [in Figma](https://www.figma.com/design/97akqulXiGIGtvxkdPLjMm?node-id=4953-6224#1142133808) mentioned
> Additionally, and now that you are mentioning "registry host", it is quite possible to omit the registry host. Thus, users can enter only the "user_or_org/repository[tag_or_digest]" part.

I'm wondering whether the recommend strategy here makes sense for OCI. @simrandhaliw might be a question for you.

manaswinidas:
@yih-wang Yes, "Recommended" works here. Sorry I missed that while recording - in the second scenario, if you click on the Connection name dropdown you see this:

Screenshot 2025-03-06 at 5 57 18 PM

simrandhaliw:
+1 regarding @yih-wang's first comment. It was strange to see the Connections section change states like that while scrolling to it. I'm not sure how we currently handle such form sections that might not have been loaded with the overall form. I think we use spinners per section, but it would be good to get confirmation from the dev team on this.

frontend/src/concepts/modelRegistry/utils.ts
```diff
@@ -10,6 +10,7 @@ export type ObjectStorageFields = {
 export type RegisteredModelLocation = {
   s3Fields: ObjectStorageFields | null;
   uri: string | null;
+  ociUri: string | null;
 } | null;
```
mturley, line 14: Maybe we should rename `uri` here to `genericUri` so it's a little more readable - as is I can imagine someone being confused why OCI models don't have their URI in the `uri` property.

I don't think we should block this PR on this since we already have [RHOAIENG-19991](https://issues.redhat.com/browse/RHOAIENG-19991) to follow up on tech debt for the prefilling behavior, but we may want to add to that tech debt ticket or open another - or ignore me...

... but a thought here: `RegisteredModelLocation` might be cleaner as a discriminated union with a `type` field and various other fields. I get why `ociUri` and `uri` are different properties, but it's a little awkward.

In the future, I could see this structured like:

```ts
export enum RegisteredModelLocationType {
  S3 = 'S3',
  OCI_URI = 'OCI_URI',
  GENERIC_URI = 'GENERIC_URI',
}

export type RegisteredModelLocation =
  | ({
      type: RegisteredModelLocationType.S3,
    } & ObjectStorageFields)
  | {
      type: RegisteredModelLocationType.OCI_URI | RegisteredModelLocationType.GENERIC_URI
      uri: string;
    }
  | null;
```

and then wherever we consumed it if we narrowed the type with `if (location.type === RegisteredModelLocationType.S3) {` it would allow us to access `location.bucket` but not `location.uri`, etc. Might make consuming it a little cleaner and the type would be more accurate - for example, right now it's not an error to have both `s3Fields` and `ociUri` which isn't valid.

This maybe gets weird because we already have `ModelLocationType` (ObjectStorage or URI) for the registration forms in MR, and we wouldn't necessarily want to reuse that since OCI URI shouldn't be its own option there... hm. We could reuse `ModelLocationType` and not have separate types for OCI_URI and GENERIC_URI, but in the URI case the type could also have an `isOCI` boolean?

Anyway - not for this PR 👀  I added a note about this to the description of [RHOAIENG-19991](https://issues.redhat.com/browse/RHOAIENG-19991) in case we want to consider it later.

frontend/src/pages/modelRegistry/screens/components/DeployRegisteredModelModal.tsx
```diff
@@ -153,6 +168,7 @@ const DeployRegisteredModelModal: React.FC<DeployRegisteredModelModalProps> = ({
         registeredModelDeployInfo={registeredModelDeployInfo}
         projectContext={{ currentProject: selectedProject, connections }}
         projectSection={projectSection}
+        existingUriOption={registeredModelDeployInfo.modelArtifactUri}
```
mturley, line 171: We only want to pass `existingUriOption` in our special registry.redhat.io OCI URI case, otherwise it should be undefined. That's the only prefilling case where the user doesn't need a connection. Always passing it like this means that all models being deployed from the registry get this extra option which isn't valid in other cases, for example it exposes our nonsense S3 URIs:

Screenshot 2025-03-06 at 6 38 17 PM

We might want an `isRedHatRegistryUri` util for our `.includes('oci://registry.redhat.io/')` so we aren't duplicating that magic string, and then we can use it here and in usePrefillDeployModalFromModelRegistry.

frontend/src/pages/modelServing/screens/projects/InferenceServiceModal/ProjectSelector.tsx
```diff
@@ -50,11 +57,19 @@ const ProjectSelector: React.FC<ProjectSelectorProps> = ({
...
+                    key: project.metadata.name,
+                    value: project.metadata.name,
+                    label: getDisplayNameFromK8sResource(project),
+                  }))
+                : projects.map((project) => ({
+                    key: project.metadata.name,
+                    value: project.metadata.name,
+                    label: getDisplayNameFromK8sResource(project),
+                  }))
+            }
```
mturley, line 72: Nit: this could be deduplicated.

```tsx
options={(isOciModel ? kserveProjects : projects).map((project) => ({
  key: project.metadata.name,
  value: project.metadata.name,
  label: getDisplayNameFromK8sResource(project),
}))}
```

mturley (changes requested):
@manaswinidas this is awesome - nice work. Just one minor thing we do need to address, and a tech debt rant we definitely don't.

frontend/src/pages/modelServing/screens/projects/InferenceServiceModal/ProjectSelector.tsx
manaswinidas: Thanks. Addressed.

frontend/src/pages/modelRegistry/screens/components/DeployRegisteredModelModal.tsx
manaswinidas: Ugh, I was supposed to get to this after writing the tests 😞 . Completely slipped my mind. Thanks for pointing this out.

frontend/src/concepts/modelRegistry/utils.ts
manaswinidas: Good suggestion 💡  . I attempted to do this within this PR. Seems I have to make changes in several places transcending several files. Thanks for adding this to the issue. This may need an issue of it's own tho(attached supporting screenshot that made me come to this conclusion) and these are just the linting issues.

Screenshot 2025-03-07 at 12 04 31 PM

mturley: Yeah it'd be a big change for maybe not enough benefit. We can move it to its own lower priority issue, I'll put that on my todo list

mturley:
@yih-wang @simrandhaliw as far as I can tell the behavior you're seeing with loading happening after form sections render is existing behavior not caused by this PR. Not sure if that was happening from when connection types were first introduced or if it was caused by https://github.com/opendatahub-io/odh-dashboard/pull/3590. I opened this followup bug: [RHOAIENG-21301](https://issues.redhat.com/browse/RHOAIENG-21301)

cc @emilys314

mturley (approved):
LGTM - nice to get this functionality working! For posterity, we'll follow up on the related work in these tickets: [RHOAIENG-21274](https://issues.redhat.com/browse/RHOAIENG-21274), [RHOAIENG-19991](https://issues.redhat.com/browse/RHOAIENG-19991), [RHOAIENG-21301](https://issues.redhat.com/browse/RHOAIENG-21301)

PR #3867, opened 2025-03-06
Add OCI Deployment E2E Test

Griffin-Sullivan:
https://issues.redhat.com/browse/RHOAIENG-18663
## Description
Added a new test for deploying a model with an OCI connection.
Updated the OOTB Connection Types test to include OCI.

## How Has This Been Tested?
Run this branch locally with `npm run start:dev:ext`
setup testing config
Run `npm run cypress:open` and run the tests in the PR.

Ran on Jenkins
image

## Test Impact
Yes...

## Request review criteria:
Run the tests yourself. I might need to add disabling the OCI feature flag :thinking: or do we instead hold this until OCI is enabled?

If you have UI changes: 

### DIFF
--- a/frontend/src/__tests__/cypress/cypress/pages/connections.ts
+++ b/frontend/src/__tests__/cypress/cypress/pages/connections.ts
@@ -45,6 +45,10 @@ class ConnectionModal extends Modal {
     return cy.findByText('S3 compatible object storage - v1');
   }
 
+  findOciConnectionType() {
+    return cy.findByText('OCI compliant registry - v1');
+  }
+
   findSubmitButton() {
     return this.findFooter().findByTestId('data-connection-submit-button');
   }
@@ -83,6 +87,26 @@ class ConnectionModal extends Modal {
 
   findBucketInput() {
     return this.find().findByTestId('field AWS_S3_BUCKET');
+  }
+
+  findOciAccessType() {
+    return this.find().findByTestId('field ACCESS_TYPE');
+  }
+
+  findOciPullSecretOption() {
+    return cy.findByText('Pull secret');
+  }
+
+  findOciSecretDetails() {
+    return this.find().findByTestId('field .dockerconfigjson');
+  }
+
+  uploadSecretDetails(filePath: string) {
+    this.findOciSecretDetails().get('input[type=file]').selectFile(filePath, { force: true });
+  }
+
+  findOciRegistryHost() {
+    return this.find().findByTestId('field OCI_HOST');
   }
 }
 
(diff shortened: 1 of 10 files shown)


frontend/src/__tests__/cypress/cypress/tests/e2e/dataScienceProjects/models/testDeployOCIModel.cy.ts
```diff
@@ -0,0 +1,89 @@
...
+  modelServingSection,
+} from '~/__tests__/cypress/cypress/pages/modelServing';
+import { checkInferenceServiceState } from '~/__tests__/cypress/cypress/utils/oc_commands/modelServing';
+
+const projectName = 'test-oci-deployment';
+const connectionName = 'test-oci-connection';
+const secretDetailsFile = Cypress.env('OCI_SECRET_DETAILS_FILE');
+const OCIRegistryHost = 'quay.io';
+const modelDeploymentURI = Cypress.env('OCI_MODEL_URI');
+const modelDeploymentName = 'test-oci-model-deployment';
```
manosnoam, line 22: Maybe some of this can go to Fixture resources ?

Griffin-Sullivan: Sure I don't mind doing that. I did it on a previous PR, but I'm just wondering why we are doing that approach? It seems each fixture only works for one file, so I don't see the difference between fixture or const

manosnoam: That is true, but the best practice would be to distinguish between test data and test logic. 
This allows us to control and reuse test data later, for example, to know if the same resource name could or could not be used in another test.

frontend/src/__tests__/cypress/cypress/pages/connections.ts
```diff
@@ -84,6 +88,26 @@ class ConnectionModal extends Modal {
...
+    return this.find().findByTestId('field ACCESS_TYPE');
+  }
+
+  findOciPullSecretOption() {
+    return cy.findByText('Pull secret');
+  }
+
+  findOciSecretDetails() {
+    return this.find().findByTestId('field .dockerconfigjson');
+  }
```
emilys314, line 102: I know this is the existing pattern, but it seems like a lot of copy pasting for getting connection fields? All the connection fields would follow the testId pattern of `'field <env var name>'`. so you could have 1 function `findFieldInput(envVar) =>  this.find().findByTestId('field ${envVar}');`

But then I guess you would have to rewrite the same env var in each test when looking for the field

emilys314: Similarly with `findOpenVinoOnnx() {
    return this.find().findByTestId('onnx - 1');
  }`

emilys314 (commented):
Just my nitpick observations, not required changes

frontend/src/__tests__/cypress/cypress/pages/connections.ts
Griffin-Sullivan: yeah I had this same thought. I know there's quite a few places in the tests where we could make the page objects more organized but I just opted to do what previous tests were doing. I'm open to changing if you think we should.

manosnoam (approved):
/lgtm

FedeAlonso:
There's some missing info in the test-variables.yml.examples
https://github.com/opendatahub-io/odh-dashboard/tree/main/frontend/src/__tests__/cypress/cypress/tests/e2e#test-variables

PR #3869, opened 2025-03-06
fix: prevent useEffect from overwriting model server size selection

olavtar:
https://issues.redhat.com/browse/NVPE-179
https://issues.redhat.com/browse/NVPE-180

## Description

Prevent useEffect from overwriting model server size selection, allowing users to switch options in the dropdown.

## How Has This Been Tested?

Tested locally.

## Test Impact

- Verified that the **NIM default model server size** is initially set to `'Custom'` with predefined default values that can be modified.  
- Confirmed that users can successfully switch from `'Custom'` to any other available options in the dropdown.  
- Ensured that in **edit mode**, the initial value is retained, and if modified, it correctly displays **NIM Defaults** instead of the generic defaults.  

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/pages/modelServing/screens/projects/NIMServiceModal/ManageNIMServingModal.tsx
+++ b/frontend/src/pages/modelServing/screens/projects/NIMServiceModal/ManageNIMServingModal.tsx
@@ -21,7 +21,7 @@ import {
 import { requestsUnderLimits, resourcesArePositive } from '~/pages/modelServing/utils';
 import useCustomServingRuntimesEnabled from '~/pages/modelServing/customServingRuntimes/useCustomServingRuntimesEnabled';
 import DashboardModalFooter from '~/concepts/dashboard/DashboardModalFooter';
-import { ServingRuntimeEditInfo } from '~/pages/modelServing/screens/types';
+import { ModelServingSize, ServingRuntimeEditInfo } from '~/pages/modelServing/screens/types';
 import ServingRuntimeSizeSection from '~/pages/modelServing/screens/projects/ServingRuntimeModal/ServingRuntimeSizeSection';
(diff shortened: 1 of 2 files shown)


frontend/src/pages/modelServing/screens/projects/NIMServiceModal/ManageNIMServingModal.tsx
```diff
@@ -373,6 +377,7 @@ const ManageNIMServingModal: React.FC<ManageNIMServingModalProps> = ({
             servingRuntimeSelected={servingRuntimeSelected}
             podSpecOptionState={podSpecOptionsState}
             infoContent="Select CPU and memory resources large enough to support the NIM being deployed."
+            nimDefaults={NIM_CUSTOM_DEFAULTS}
```
emilys314, line 380: It should be possible to not have a nim specific argument for this component. Before it didn't, is it possible to just remove this line and the changes in `<ServingRuntimeSizeSection/>`?

emilys314 (commented):
I tested it out and everything does work good

frontend/src/pages/modelServing/screens/projects/NIMServiceModal/ManageNIMServingModal.tsx
olavtar: I'm not sure I follow. We were setting the 'Custom' defaults for NIM before.

emilys314: Sorry, i'm referring to having `nimDefaults={NIM_CUSTOM_DEFAULTS}` added. 

The previous code set the defaults without needing this line.

I don't want defaults specifically for nim, i'd rather use the existing code to set the custom defaults or a generic custom defaults

PR #3870, opened 2025-03-06
Improve hardware profiles permission state

andrewballantyne:
https://issues.redhat.com/browse/RHOAIENG-21069

## Description

Shifts the page & navigation to `create` Hardware Profiles to avoid it accidentally listing/visible to basic users.
Add support for the patch toggle (disable the toggle)

UX will need to comment more on the effort for where we will go and proper guidance for read-only values (like the toggle). https://issues.redhat.com/browse/RHOAIENG-21129 will cover the future of those features and come with documentation for the team.

Also spent some time investigating the performance issues -- apparently we didn't cache the values at all fast enough and made a ton of quick "while loading" calls before the state settled. I wrestled with it, tried to also fix the not unloading properly stuff and then ended up in a spot where it's a bit of a mess. I have put all these into the first commit. The 2nd one reverts everything down to a smaller state (and takes a win from the improving cache responsiveness).

## How Has This Been Tested?

Using the Hardware Profiles page (needs to be enabled on your cluster or via dev flags query).

## Test Impact

N/A

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/concepts/userSSAR/AccessReviewContext.tsx
+++ b/frontend/src/concepts/userSSAR/AccessReviewContext.tsx
@@ -1,5 +1,4 @@
 import React from 'react';
-import { useLocation } from 'react-router-dom';
 import { checkAccess } from '~/api';
 import { AccessReviewResourceAttributes } from '~/k8sTypes';
 import useNamespaces from '~/pages/notebookController/useNamespaces';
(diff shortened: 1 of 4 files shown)


Gkrumbach07 (approved):
this looks good. rerunning the failed (timed out) test.
/lgtm

PR #3882, opened 2025-03-07
stashed home hint banner for hardware profiles

rsun19:
https://issues.redhat.com/browse/RHOAIENG-21273

## Description

Screenshot 2025-03-13 at 12 19 21 PM

Stashed the home hint banner for hardware profiles for future use. It no longer appears.

## How Has This Been Tested?

set feature flag for hardware profiles and visit the home page. the banner should not appear.

## Test Impact

I reverted the cypress tests that enforce their disappearance

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/pages/home/Home.tsx
+++ b/frontend/src/pages/home/Home.tsx
@@ -8,7 +8,7 @@ import ProjectsSection from './projects/ProjectsSection';
 import { useAIFlows } from './aiFlows/useAIFlows';
 import { useResourcesSection } from './resources/useResourcesSection';
 import { useEnableTeamSection } from './useEnableTeamSection';
-import HomeHint from './HomeHint';
+import LandingPageHomeHint from './LandingPageHomeHint';
 
 const Home: React.FC = () => {
   const { status: projectsAvailable } = useIsAreaAvailable(SupportedArea.DS_PROJECTS_VIEW);
@@ -18,7 +18,7 @@ const Home: React.FC = () => {
 
   return (
     <div data-testid="home-page">
-      <HomeHint />
+      <LandingPageHomeHint />
       {!projectsAvailable && !aiFlows && !resourcesSection && !enableTeamSection ? (
         <PageSection
           hasBodyWrapper={false}
--- a/frontend/src/pages/home/HomeHint.tsx
+++ b/frontend/src/pages/home/HomeHint.tsx
@@ -4,28 +4,29 @@ import {
   Card,
   CardBody,
   CardHeader,
+  Content,
   Flex,
   FlexItem,
   PageSection,
-  Content,
 } from '@patternfly/react-core';
 import { TimesIcon } from '@patternfly/react-icons';
 import { useBrowserStorage } from '~/components/browserStorage';
-import TitleWithIcon from '~/concepts/design/TitleWithIcon';
-import { ProjectObjectType } from '~/concepts/design/utils';
-import ExternalLink from '~/components/ExternalLink';
-import { SupportedArea, useIsAreaAvailable } from '~/concepts/areas';
 
-const HomeHint: React.FC = () => {
+type HomeHintProps = {
+  title: React.ReactNode;
+  body: React.ReactNode;
+  isDisplayed: boolean;
+  homeHintKey: string;
+  image?: React.ReactNode;
+};
+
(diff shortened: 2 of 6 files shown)


dpanshug (changes requested):
Can you add a TODO comment to the commented-out code that needs to be restored in the future?

frontend/src/pages/home/Home.tsx
```diff
@@ -18,7 +17,6 @@ const Home: React.FC = () => {
 
   return (
     <div data-testid="home-page">
-      <HomeHint />
```
Gkrumbach07, line 21: please bring back the old home hint

Gkrumbach07 (changes requested):

frontend/src/pages/home/Home.tsx
rsun19: should i just comment it out then?

rsun19: other than that not sure how else to stash it

Gkrumbach07: so there was an old home hint that i link to in the PR in the jira. so we should being that bake but still keep this hardware profile one. So you will end with 
- HardwareProfileHomeHint
- WHateverTheOldHomeHint
- probably a BaseHomeHint

rsun19: ok i am so confused. can we do a quick 5 min sync whenever you're free?

rsun19:
fixed

frontend/src/pages/home/LandingPageHomeHint.tsx
```diff
@@ -0,0 +1,97 @@
...
+              padding: 'var(--pf-t--global--spacer--xs)',
+              borderRadius: 'var(--pf-t--global--border--radius--small)',
+            }}
+          />
+          <FlexItem>
+            <Content>
+              <Content component="p" data-testid="hint-body-text">
+                {ODH_PRODUCT_NAME} has a new landing page. You can access applications that are
+                enabled for your organization, such as Jupyter, from the{' '}
+                <Button
```
DaoDaoNoCode, line 77: Cannot we replace this `Button` component with `Link` component simply? It will have only 2 props: `to='/enabled'` and `data-testid="home-page-hint-navigate"`

DaoDaoNoCode:
Also, I am wondering if we could make a generic component for the home hint here. I saw so many custom styles and they are duplicated in different components. Unless all the hints are very different, we should look into making a generic component and pass necessary props to it like title, body, image...

rsun19:
> Also, I am wondering if we could make a generic component for the home hint here. I saw so many custom styles and they are duplicated in different components. Unless all the hints are very different, we should look into making a generic component and pass necessary props to it like title, body, image...

if we do it this way, we need to find a way to stash the props. Maybe I can do this?

```
<homehint with old props />
{/* <homehint with hardware profile props /> */}
```

my concern is that the commented out props might not look as clean, but that's just my two cents

DaoDaoNoCode:
@rsun19 No, I mean we can still keep `LandingPageHomeHint` and `HardwareProfilesHomeHint` components, which will wrap the generic component I mentioned.
like
```
const LandingPageHomeHint = () => <HomeHint {...landingPageProps} />
```
and in `HomeHint`, you can do all the state management there like `hintHidden`, and receive all the props you need to render the `Card` component.
Let me know if that makes sense!

rsun19:
@DaoDaoNoCode let me know if this is what you mean

DaoDaoNoCode:
@rsun19 Exactly!

frontend/src/pages/home/HomeHint.tsx
```diff
@@ -4,28 +4,28 @@ import {
...
+  body: React.ReactNode;
+  isDisplayed: boolean;
+  image?: React.ReactNode;
+};
+
+const HomeHint: React.FC<HomeHintProps> = ({ title, image, body, isDisplayed }) => {
   const [hintHidden, setHintHidden] = useBrowserStorage<boolean>(
     'odh.dashboard.landing.hint',
     false,
   );
```
Gkrumbach07, line 26: this key needs to chnage based on the type of home hint. becuase if we change the home hint then we will want users that have hidden the hint before to see the new one. probably passing some key or id in shoudl do the trick and slap it on the end of the key

Gkrumbach07 (approved):
/lgtm

PR #3889, opened 2025-03-10
Organize Models by group

andrewballantyne:
After being asked what API we support I noticed that we lack good organization in the way our API models work.

Reorganize the location of all the models to match their `group`.


emilys314:
looks good to me

frontend/src/api/k8s/hardwareProfiles.ts
```diff
@@ -33,7 +33,7 @@ export const assembleHardwareProfile = (
   namespace: string,
   visibility: string[] = [],
 ): HardwareProfileKind => ({
-  apiVersion: `${HardwareProfileModel.apiGroup}/${HardwareProfileModel.apiVersion}`,
+  apiVersion: kindApiVersion(HardwareProfileModel),
```
andrewballantyne, line 36: Converted direct usage -- it seems we were using `const model = { ... } satisfies K8sCommonModel` to get around this issue with our linter not appreciating the type can have a missing `apiGroup`.

andrewballantyne:
Added lgtm based on Gage's comment

PR #3892, opened 2025-03-10
remove linting from unit test coverage command

christianvogt:
## Description

Linting was being run twice:
 - once as a separate job
 - once again alongside unit tests

This PR removes the lint command from the unit test script but also makes it a requirement to pass before starting any test jobs.

## How Has This Been Tested?

PR check

## Test Impact

N/A

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -45,7 +45,7 @@ jobs:
         run: npm run lint:frontend:backend
 
   Unit-Tests:
-    needs: Setup
+    needs: [Setup, Lint]
     runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


antowaddle (commented):
I'm just thinking about this, if the linting fails we don't want anything else to run. Perhaps we would want the linting to run first and then based on the results to execute the Unit Tests/Mocks in parallel? We'll probably save another 30 seconds to 1 minute.

christianvogt:
Updated to make linting a sequential step in the checks which will prevent unit and cypress tests from executing if it fails.

PR #3893, opened 2025-03-10
Show error if only Push access

emilys314:
Closes https://issues.redhat.com/browse/RHOAIENG-20982

## Description

This changes a lot of the connection form validation from valid to error state. I then made it a `boolean | string` for allowing messages. Then in the validation callback being passed to the form, i added a custom one that can be added by the parent of the connection form. 

image

File padding alignment fix
image

## How Has This Been Tested?

Create an oci connection and select only the push type, it should error

## Test Impact

test added

## Request review criteria:

If you have UI changes: 

@simrandhaliw 

### DIFF
--- a/frontend/src/concepts/connectionTypes/ConnectionTypeForm.tsx
+++ b/frontend/src/concepts/connectionTypes/ConnectionTypeForm.tsx
@@ -112,6 +112,9 @@ type Props = Pick<
   connectionValues?: {
     [key: string]: ConnectionTypeValueType;
   };
+  connectionErrors?: {
+    [key: string]: boolean | string;
+  };
 };
 
 const ConnectionTypeForm: React.FC<Props> = ({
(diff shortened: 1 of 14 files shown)


frontend/src/concepts/connectionTypes/ConnectionTypeForm.tsx
```diff
@@ -112,6 +112,9 @@ type Props = Pick<
   connectionValues?: {
     [key: string]: ConnectionTypeValueType;
   };
+  connectionErrors?: {
+    [key: string]: boolean | string;
```
emilys314, line 116: I could change this into an `Error` class which basically does the same thing. But it's kinda heavyweight I feel compared to this. 

The 2 different types is a little unusual I think. but it allows just general error states that already exist, or a custom message to be provided. 

Lastly the validation could be updated fully in all instances of connections, but i want to reduce the scope of changes to reduce chance for bugs

Griffin-Sullivan (approved):
/lgtm works well for me and didn't find any regressions

Griffin-Sullivan (approved):
/lgtm

PR #3898, opened 2025-03-11
Fix: Prevent API key from being removed when closing input early

LinoyBitan1:
JIRA - [https://issues.redhat.com/browse/NVPE-151](https://issues.redhat.com/browse/NVPE-151)

## Description

This PR fixes an issue where the `api_key` was being removed when closing the input window before it finished. The key was missing in the resulting Secret. Now, the `api_key` is handled correctly and is not removed unintentionally.

**Before:**
- When the input window was closed before it finished, a Secret was created without any data keys, resulting in a missing `api_key` in the account.
image

**After:**
- The API key is now properly preserved, even when the input window is closed prematurely.
image

## How Has This Been Tested?

Tested locally.

## Test Impact

following these steps:
1. Enabled NVIDIA.
2. Closed the input window before it finished.
3. Verified that the API key is now included in the Secret and no longer missing.

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/pages/exploreApplication/EnableModal.tsx
+++ b/frontend/src/pages/exploreApplication/EnableModal.tsx
@@ -50,7 +50,12 @@ const EnableModal: React.FC<EnableModalProps> = ({ selectedApp, shown, onClose }
   };
 
   const handleClose = React.useCallback(() => {
-    setEnableValues({});
+    // Clear only the values, keeping the keys intact
+    const resetValues: { [key: string]: string } = {};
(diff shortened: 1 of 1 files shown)


emilys314:
So the PR does seem to address the immediate problem, the key in the yaml will still exist.

However if you reopen the enable modal it will still have the loading state, and it will fail
image

Vince mentioned closing the modal should cancel the operation, so i would think the modal is fully reset. this is probably beyond the scope of the actual bug. Also it would be more work. thoughts?

PR #3900, opened 2025-03-11
fix: add BUCKET_1 to test-variables

FedeAlonso:
https://issues.redhat.com/browse/RHOAIENG-21409 

## Description
After some time out and trying to execute the tests locally from a fresh install it wasn't working. It was because the tests needed the config info of BUCKET_1 which was not present in the "test-variables.yml.example".

## How Has This Been Tested?
Now the tests work

## Test Impact
Not really 


antowaddle (approved):
/lgtm

PR #3901, opened 2025-03-11
Quarantine ODS-1877 'Verify That Usage Data Collection Can Be Set In Cluster Settings'

ConorOM1:
## Description
Quarantine `Verify That Usage Data Collection Can Be Set In Cluster Settings` due to a product bug https://issues.redhat.com/browse/RHOAIENG-18495


antowaddle (approved):
/lgtm

PR #3906, opened 2025-03-12
fix: infinite loop when re-entering same invalid key

mtalvi:
Jira - https://issues.redhat.com/browse/NVPE-185
## Description

This PR handles the following issue:
On the first try, when an invalid API key is submitted, the backend returns an error and the hook correctly updates the UI to show the error. However, on a **second attempt** with the same invalid key from the first attempt, we get into an **infinite loop** - the polling function keeps running (continuously showing the “Validating your entries” spinner). 
## How Has This Been Tested?

Tested locally.
## Test Impact

1.  Try to enable NIM with an invalid key.  Dashboard should show after a few seconds that the key is invalid.
2.  Try again with the same invalid key.  Dashboard should instantly show  that the validation failed with this key.
3. Valid key or a new invalid key process and logic should not change.

https://github.com/user-attachments/assets/8cee9840-2a0f-4287-bc30-6f5b157f7272

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/utilities/useEnableApplication.tsx
+++ b/frontend/src/utilities/useEnableApplication.tsx
@@ -31,6 +31,8 @@ export const useEnableApplication = (
   }>({ status: EnableApplicationStatus.IDLE, error: '' });
   const [lastVariablesValidationTimestamp, setLastVariablesValidationTimestamp] =
     React.useState<string>('');
+  // Add a counter for tracking the number of responses with the same timestamp
+  const sameTimestampCountRef = React.useRef<number>(0);
   const dispatch = useAppDispatch();
 
   const dispatchResults = React.useCallback(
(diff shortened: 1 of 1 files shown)


emilys314:
So 2 issues I see with this implementation
1. The time limit is arbitrarily set based on an assumption that I don't think we can accurately assume.
2. This is a generic odhApp component, and the changes would affect other components. They probably won't be facing the same issue, and this update would negatively affect those.

andrewballantyne:
We are discussing this [on slack](https://redhat-internal.slack.com/archives/C07CZDA6DE1/p1741730651459129) -- I do not believe the client doing a random timeout is good for the health of the client and the support of more integrations. I'm going to close this until we determine a path forward.

PR #3868, opened 2025-03-06
Add helperText property to FileField

emilys314:
Closes https://issues.redhat.com/browse/RHOAIENG-20559

## Description

Add a warning text for the oci pull secret. This is only done behind the scenes with a custom property. I'm not sure if adding an entire field in the admin page is necessary since this is the only use case so far

image

## How Has This Been Tested?

Update your oci oobt configmap with the one in this PR

## Test Impact

Added a unit test

## Request review criteria:

Self checklist (all need to be checked):
- [X] The developer has manually tested the changes and verified that the changes work
- [X] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [X] The developer has added tests or explained why testing cannot be added (unit or cypress tests for related changes)

If you have UI changes: 

- [X] Included any necessary screenshots or gifs if it was a UI change.
- [X] Included tags to the UX team if it was a UI/UX change.
@simrandhaliw 

After the PR is posted & before it merges:
- [ ] The developer has tested their solution on a cluster by using the image produced by the PR to `main`

PR #3907, opened 2025-03-12
OCI connections microcopy updates

Griffin-Sullivan:
https://issues.redhat.com/browse/RHOAIENG-20231
## Description
Microcopy updates for OCI ModelCars on Connections and serving modals. https://github.com/opendatahub-io/odh-dashboard/pull/3868 covers part of the microcopy that is missing here.

## How Has This Been Tested?
Checking screens against figma and microcopy doc
Screenshot From 2025-03-13 10-41-09
Screenshot From 2025-03-13 10-41-20
Screenshot From 2025-03-13 10-41-42
Screenshot From 2025-03-13 10-41-58
Screenshot From 2025-03-13 10-42-22

## Test Impact
Update a test with the right modal title

## Request review criteria:

If you have UI changes: 


frontend/src/pages/modelServing/screens/projects/InferenceServiceModal/ConnectionOciPathField.tsx
```diff
@@ -42,36 +43,45 @@ const ConnectionOciPathField: React.FC<ConnectionOciPathFieldProps> = ({
...
+      <FormSection title="OCI storage location">
         <FormHelperText>
           <HelperText>
             <HelperTextItem>
-              Example model path structure is Base URL / Registry organization / Model Name:Model
-              Version
-            </HelperTextItem>
-            <HelperTextItem>
-              Example: registry.connect.redhat.com/redhat-test/test-mymodel:v3.2
+              Provide the location of the model. This is not part of your connection instance.
```
Griffin-Sullivan, line 50: The spacing on this help text looks off since there's such a far gap from the title. Maybe I shouldn't use the FormSection title and just add a normal title with Helper text?

Griffin-Sullivan:
@simrandhaliw OCI microcopy is ready for review!

frontend/src/concepts/connectionTypes/ConnectionTypeForm.tsx
```diff
@@ -176,6 +177,16 @@ const ConnectionTypeForm: React.FC<Props> = ({
...
+        {connectionTypeName === 'oci-v1' && (
+          <Alert
+            variant="warning"
+            style={{ marginTop: 'var(--pf-t--global--spacer--lg)' }}
+            title="Deploying from an OCI connection"
+          >
+            Only models that are packaged as ModelCars can be deployed using OCI connections. Ensure
+            that that your model is in ModelCar format before deploying or the deployment will fail.
+          </Alert>
+        )}
```
emilys314, line 189: Ah, so this is going to appear in every instance of the connectionTypeForm, even if it's not model serving specific. Which might not be ideal. It's tricky though since it's right in the middle of the form. you could pass the alert in as a prop. But maybe it's fine if it's always shown??

Also it would probably be better to change `connectionTypeName === 'oci-v1'` to `isModelServingCompatible(connectionType, ModelServingCompatibleTypes.OCI)`. Unless we only want to show it for the ootb connection type?

simrandhaliw:
@vconzola, just to confirm, we decided that we won’t display the modelcar's yellow warning alert in the 'Create Connection' modal, but instead, we will show it within the Connection section (for both 'Existing Connection' and 'Create Connection' options) when OCI connections are selected. So, the yellow warning in the first screenshot above should be removed, right?

frontend/src/pages/modelServing/screens/projects/InferenceServiceModal/ConnectionOciPathField.tsx
emilys314: Yeah it does look like a large gap from the title. Does this have to be a <HelperText/>? Or can that be removed and this become regular html text?

emilys314:
> Also, the info alert we display for 'Secret Details' is missing (see designs [here](https://www.figma.com/design/97akqulXiGIGtvxkdPLjMm/Data-connections-enhancement?node-id=5539-113679&t=PBc5AY7SrdqgCESh-4)). The microcopy for that might actually change a bit. I’ll send an update as soon as I hear back from Katie.

@simrandhaliw that part is here https://github.com/opendatahub-io/odh-dashboard/pull/3868

Griffin-Sullivan:
Ok thanks I'll fix the warning and look into the italics. I'm also going to add the warning for the model URI when it doesn't match the registry host here since it's basically microcopy.

frontend/src/pages/modelServing/screens/projects/InferenceServiceModal/ConnectionOciPathField.tsx
Griffin-Sullivan: I guess it doesn't have to be I was just using it to match the rest of the form. Let me try some other ways around it

vconzola:
@Griffin-Sullivan Just so we're all on the same page, the final microcopy for the warning to be shown within the Connection section (for both 'Existing Connection' and 'Create Connection' options) when OCI connections are selected is:
"**Model requirement for OCI connections**
OCI connections can be used to deploy only models packaged as ModelCars. Deploying models packaged in different formats will result in failure."

vconzola:
@Griffin-Sullivan The warning alert lgtm. Not sure about other microcopy changes. @simrandhaliw will have to weigh in on those.

Griffin-Sullivan:
> except for the 'View connection type details' link in the Create connection section. It should be styled similarly to the 'info icon + View details' link displayed in the Existing connection section.

@simrandhaliw I can do that too wasn't sure if we wanted to since it wasn't mentioned in the figma. I'll update now!

Griffin-Sullivan:
Screenshot From 2025-03-13 11-21-46
Screenshot From 2025-03-13 11-21-59

@simrandhaliw here's the button for new connections

simrandhaliw:
Thank you! LGTM!

frontend/src/concepts/connectionTypes/ConnectionTypeForm.tsx
```diff
@@ -128,6 +129,7 @@ const ConnectionTypeForm: React.FC<Props> = ({
   onChange,
   onValidate,
   connectionErrors,
+  Alert = undefined,
```
emilys314, line 132: Very minor nitpick, but should this start with lowercase since it's a prop variable? I guess it's a component prop, but I'm not sure which priority. Thoughts?

Also imo you don't have to init an optional prop to undefined, the `?` will make it for you, it saves code clutter

frontend/src/pages/modelServing/screens/projects/InferenceServiceModal/ConnectionOciPathField.tsx
```diff
@@ -8,19 +8,25 @@ import {
...
+  isNewConnection?: boolean;
 };
 
 const ConnectionOciPathField: React.FC<ConnectionOciPathFieldProps> = ({
   ociHost,
   modelUri,
   setModelUri,
+  isNewConnection = false,
 }) => {
+  const [isOciHostInModelUri, setIsOciHostInModelUri] = React.useState(true);
```
emilys314, line 28: Actually I wonder if we should make this a React.useMemo that checks on the fly. Because if you enter the oci URI, but then update the registry path, the 2 can get out of sync. Thoughts? (That does mean the onblur / onChange logic changes from what we talked about)

image

frontend/src/concepts/connectionTypes/ConnectionTypeForm.tsx
Griffin-Sullivan: Yeah I went back and forth but landed on capital. I don't really care though. And yeah I can do ?

frontend/src/pages/modelServing/screens/projects/InferenceServiceModal/ConnectionOciPathField.tsx
Griffin-Sullivan: ok I can switch back to useMemo but that means the warnings show up as you type / edit. Unless there's a way to fix it and use Memo?

frontend/src/concepts/connectionTypes/ConnectionTypeForm.tsx
emilys314: You don't have to change it then

frontend/src/pages/modelServing/screens/projects/InferenceServiceModal/ConnectionOciPathField.tsx
emilys314: I don't think there is a clean / simple way to do that with just useMemo. you can resolve the convos as you see fit.

Griffin-Sullivan:
@simrandhaliw just realized when I added that view details button it also changed it for the Create Connection modal. Is that how it should be?
image

frontend/src/pages/modelServing/screens/projects/InferenceServiceModal/ConnectionOciPathField.tsx
Griffin-Sullivan: Decided to keep it as is but added the `onFocus` in case you tab / click on the uri path again it will take the warning away

emilys314: oo ok nice

emilys314: So I'm seeing an issue with the validation, when i focus and then blur, it seems to be acting differently 
image

I think this might be due to the prefix that gets added `oci://`, 

Also I kinda thought you would validate on focus, not remove the error. Because if t's in the errors state and you go click inside of it, the error goes away, which is slightly confusing because a user might think it's valid again even if they didn't change the input

Griffin-Sullivan: Updated this to remove the on focus

emilys314: There is still an issue when you repeatedly focus and blur the text input. I think it's from the oci:// prefix behind the scenes. you may want you validation to ignore this, or change it to `includes()`? 

image

simrandhaliw:
@Griffin-Sullivan Yes, that update LGTM! Any modal that previously had the link buttons below the dropdown for 'View connection details' or 'View connection type details' should now ideally be replaced with the info icon + 'View details' button on the same line as the dropdown.

PR #3871, opened 2025-03-07
Update reliantAreas of Fine tuning feature

manaswinidas:
Closes: [RHOAIENG-21244](https://issues.redhat.com/browse/RHOAIENG-21244)

## Description

When both pipelines and model registry feature flags are turned OFF("Model catalog" is absent coz it is reliant on "Model registry")

Screenshot 2025-03-07 at 1 32 01 PM

When both model registry and pipelines feature flags are turned ON

Screenshot 2025-03-07 at 1 36 53 PM

When model registry is turned OFF but pipelines feature flag is ON
Screenshot 2025-03-07 at 1 34 13 PM

## How Has This Been Tested?

Try editing the feature flags as stated in the Description

## Test Impact

N/A only reliant area changes

## Request review criteria:

Self checklist (all need to be checked):
- [x] The developer has manually tested the changes and verified that the changes work
- [x] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [x] The developer has added tests or explained why testing cannot be added (unit or cypress tests for related changes)

If you have UI changes: 

- [x] Included any necessary screenshots or gifs if it was a UI change.
- [ ] Included tags to the UX team if it was a UI/UX change.

After the PR is posted & before it merges:
- [ ] The developer has tested their solution on a cluster by using the image produced by the PR to `main`

PR #3908, opened 2025-03-12
Add feature flag for project scoped items

manaswinidas:
Closes: [RHOAIENG-21018](https://issues.redhat.com/browse/RHOAIENG-21018)

## Description

Just adding a feature flag to unblock stories in [RHOAIENG-18463](https://issues.redhat.com/browse/RHOAIENG-18463)

Added global icon that will be used in other stories

## Test Impact

Needs to be tested with the upcoming stories of the epic

## Request review criteria:

If you have UI changes: 


frontend/src/concepts/areas/const.ts
```diff
@@ -73,6 +74,14 @@ export const SupportedAreasStateMap: SupportedAreasState = {
   [SupportedArea.DS_PROJECTS_VIEW]: {
     featureFlags: ['disableProjects'],
   },
+  [SupportedArea.DS_PROJECT_SCOPED]: {
+    featureFlags: ['disableProjectScoped'],
+    reliantAreas: [
```
manaswinidas, line 79: @mturley Does this section look good?

mturley: Hmm, this means the feature will only work if all 3 reliantAreas are enabled - do we want this to work for workbenches and hardware profiles if model serving is disabled for example? If so it might be fine to just not have any reliantAreas here

manaswinidas: Yes, we need it to work with whatever component is available. Removing the `reliantAreas`

manaswinidas: On second thoughts, `reliantAreas` only needs one to be true as per https://github.com/opendatahub-io/odh-dashboard/blob/92357499cc533936aa7d9e24a77f2dbde333f28e/frontend/src/concepts/areas/types.ts#L16

manaswinidas: I think we don't need a ProjectScoped if all three are turned OFF, although workbenches will always be turned ON by default coz I guess it doesn't have a feature flag.

manaswinidas: Also, based on manual test for https://github.com/opendatahub-io/odh-dashboard/pull/3871, it was working for the other reliantAreas if one was absent.

mturley: Ok interesting, I wasn't aware of reliantAreas working that way.... hmm, I wonder if we've made assumptions that all reliant areas are required in other places. Like SupportedArea.FINE_TUNING should really require all of its reliantAreas.... something to come back to...

mturley (approved):
LGTM - thanks @manaswinidas !

PR #3919, opened 2025-03-13
Update OCI Model URI onChange to not to reset the connection type

ppadti:
Closes: [RHOAIENG-21274](https://issues.redhat.com/browse/RHOAIENG-21274)

## Description
In this PR I have removed the useEffect and refactored a bit. 
When we deploy a model from model registry. When there is no matching connection in a project and we pre-fill the information (model with s3 connection). If we switch the connection type to OCI and try to enter value to Model URI  field, now the connection type will not switch it back to s3. 

https://github.com/user-attachments/assets/edf749b7-b323-4c3c-a09a-45e9bf21cd85

Also, added loading state to the connection section in the Deploy model Modal for both model registry and model deployment

https://github.com/user-attachments/assets/629fa1d4-3f2c-46cd-b39b-3db3a025251b

https://github.com/user-attachments/assets/aabd98bb-b2f5-465a-9551-8aab88407cdc

**Note:**
We do have existing issue/issue:
When we deploy a model (created with uri-connection type) from model registry, we will prefill the uri field. But if we switch it to OCI - then the Model uri field of OCI will also be pre-filled with same uri. 

https://github.com/user-attachments/assets/3d7c7d3b-757e-4fc0-9e16-34dc187df0a2

## How Has This Been Tested?
1. Navigate to Model Registry
2. Select a registered model
3. Select 'Deploy' via the kebab menu for the version
4. Deploy to a project with no matching connections
5. Switch the connection type to OCI and try to enter a value to Model URI field, the connection type will not switch back to s3 or uri(based on the model location of the model). Evrything will work as expected.

## Test Impact
Added cypress tests and will be adding unit test to the `usePrefillDeployModalFromModelRegistry` (may be we can do it in follow up PR as well)

## Request review criteria:

If you have UI changes: 


frontend/src/__tests__/cypress/cypress/pages/modelServing.ts
```diff
@@ -226,6 +226,10 @@ class InferenceServiceModal extends Modal {
     return this.find().findByTestId('field AWS_S3_BUCKET');
   }
 
+  findBaseURL() {
```
Griffin-Sullivan, line 227: I have a PR adding some page objects that might be useful for you if you want to align with me https://github.com/opendatahub-io/odh-dashboard/pull/3867

Griffin-Sullivan (commented):
image
After rebasing your PR locally, if I go to deploy a model and create connection I'm unable to type in the Model URI text input

ppadti:
@yih-wang Could you please take a look at the screen recording(2nd and 3rd one) for loading state of connection section. For now I have added the skeleton. Let me know your thoughts one this.

Griffin-Sullivan:
Ok thanks I tested and seems to be working

yih-wang:
Thanks @ppadti, the loading states look good to me!

gitdallas:
this isn't regarding your changes, but seeing the deploy modal project popover
image

wonder if we can remove the `appendTo: 'inline'` popper props on the `SimpleSelect` in `ProjectSelector` to make it look like this:
image

@mturley @yih-wang

yih-wang:
Sure @gitdallas! That bug has been bugging us for a while 😅  It’d be great if we could tackle it here with a quick fix!

mturley:
@ppadti it looks like with the loading state here you're also closing https://issues.redhat.com/browse/RHOAIENG-21301

mturley (approved):
LGTM! Thanks @ppadti !

PR #3930, opened 2025-03-17
Releasing ODH 2.31.2

dpanshug:


manifests/odh/params.env
```diff
@@ -1,4 +1,4 @@
 # Injected variables from the Operator
-odh-dashboard-image=quay.io/opendatahub/odh-dashboard:main
+odh-dashboard-image=quay.io/opendatahub/odh-dashboard:sha256:a2e03b85b28a1e265754e49fd9a46c9c9ee45a01c47bd9736212f754a29b4570
```
lucferbux, line 2: Shouldn't be ˋ@sha256...ˋ? @dpanshug the image is valid, checked against quay.

lucferbux (changes requested):
Only one small nit

manifests/odh/params.env
manaswinidas: Even I think it's `quay.io/opendatahub/odh-dashboard@sha256:a2e03b85b28a1e265754e49fd9a46c9c9ee45a01c47bd9736212f754a29b4570` or `:` and `@` can be used interchangeably?

PR #3937, opened 2025-03-17
limit heading level within model card

gitdallas:
https://issues.redhat.com/browse/RHOAIENG-21744

## Description
downgrading heading levels in model card markdown so that nothing has a higher level than h3. this prevents the a11y issue of having a heading as high or higher than those above it (h2)
image

## How Has This Been Tested?
locally, green cluster

## Test Impact
none, just styling

## Request review criteria:

If you have UI changes: 

- [x] Included tags to the UX team if it was a UI/UX change. @yih-wang gave a thumbs up on the visual in slack

### DIFF
--- a/frontend/src/pages/modelCatalog/screens/ModelDetailsView.tsx
+++ b/frontend/src/pages/modelCatalog/screens/ModelDetailsView.tsx
@@ -33,7 +33,9 @@ const ModelDetailsView: React.FC<ModelDetailsViewProps> = ({ model }) => (
           <h2>Model card</h2>
           {!model.readme && <p className={text.textColorDisabled}>No model card</p>}
         </Content>
-        {model.readme && <MarkdownView data-testid="model-card-markdown" markdown={model.readme} />}
+        {model.readme && (
+          <MarkdownView data-testid="model-card-markdown" markdown={model.readme} maxHeading={3} />
+        )}
       </SidebarContent>
       <SidebarPanel>
         <DescriptionList isFillColumns>
--- a/frontend/src/utilities/markdown.ts
+++ b/frontend/src/utilities/markdown.ts
@@ -2,14 +2,27 @@ import DOMPurify from 'dompurify';
 import { Converter } from 'showdown';
 
 export const markdownConverter = {
-  makeHtml: (markdown: string): string => {
-    const unsafeHtml = new Converter({
+  makeHtml: (markdown: string, maxHeading = 1): string => {
+    const converter = new Converter({
       tables: true,
       openLinksInNewWindow: true,
       strikethrough: true,
       emoji: true,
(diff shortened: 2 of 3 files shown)


frontend/src/pages/modelCatalog/screens/ModelDetailsView.tsx
```diff
@@ -33,7 +33,9 @@ const ModelDetailsView: React.FC<ModelDetailsViewProps> = ({ model }) => (
           <h2>Model card</h2>
           {!model.readme && <p className={text.textColorDisabled}>No model card</p>}
         </Content>
-        {model.readme && <MarkdownView data-testid="model-card-markdown" markdown={model.readme} />}
+        {model.readme && (
+          <MarkdownView data-testid="model-card-markdown" markdown={model.readme} maxHeading={3} />
```
mturley, line 37: Nit - max/min-based naming like `maxHeading` is potentially confusing here because maxHeading=3 numerically implies that the allowed headings will be 1-3 when it really means 3-6.. but I'm not sure what a better name would be. `startHeadingsAt` or `initialHeading` or something maybe? max is fine, I'm overthinking it

frontend/src/utilities/markdown.ts
```diff
@@ -2,14 +2,27 @@ import DOMPurify from 'dompurify';
...
+    });
+    let unsafeHtml = converter.makeHtml(markdown);
+
+    if (maxHeading > 1) {
+      // shift headings down (prevent them from competeing with page headings)
+      const shift = maxHeading - 1;
+      unsafeHtml = unsafeHtml.replace(
+        /<(\/?)h(\d)(.*?)>/gi,
+        (match, closingSlash, level, attributes) => {
+          const newLevel = Math.min(parseInt(level) + shift, 4); // cap at h4
```
mturley, line 21: Why cap at h4? Aren't they valid up to h6?

mturley (commented):
We could merge as-is, but one nit and one question

frontend/src/utilities/markdown.ts
gitdallas: just to be in-line with the current `ALLOWED_TAGS` in that file

frontend/src/pages/modelCatalog/screens/ModelDetailsView.tsx
gitdallas: idk, i guess i'm open to change but to me h1 is the top heading so would be at the top and whatever is closest to that would be 'max', `startHeadingsAt` or `initialHeading` would be confusing to me.

maybe something like `maxHeadingSize` or something, but i doubt this will be used by anything else

mturley (approved):
Ship it 🚀

PR #3946, opened 2025-03-18
Filter out OCI connections for model mesh

emilys314:
Closes https://issues.redhat.com/browse/RHOAIENG-21893

## Description

Kyle also mentioned we shouldn't have it show up in model mesh. OCI isn't compatible with model mesh / we don't want to support it

## How Has This Been Tested?

Create a connection in a project, try to deploy a model mesh server, it shouldn't show up. Same with the connection type

## Test Impact

Updated the modelmesh cypress test to assert the correct connection types are showing

## Request review criteria:

If you have UI changes: 


Griffin-Sullivan (commented):
/lgtm 

tested locally and works. Thanks for fixing!

PR #3952, opened 2025-03-19
update test.yml with cache lookup-only

christianvogt:
## Description

- Use cache `lookup-only` option to skip restoring the cache unless needed. Saves ~20 seconds on the `Setup` job if the node modules cache is hit; which it often is unless a change in package.json is made.
- Generate cache keys and store in job output for reuse across jobs.
- Run `actions/setup-node` per job to ensure consistent node version.

## How Has This Been Tested?

git hub action

## Test Impact

N/A

## Request review criteria:

If you have UI changes: 


antowaddle (approved):
/lgtm

PR #3958, opened 2025-03-20
Convert Hardware Profile manage form validation to use Zod schema validation

dpanshug:
[RHOAIENG-20560](https://issues.redhat.com/browse/RHOAIENG-20560)

## Description
Added zod validation for `ManageHardwareProfile` form

## How Has This Been Tested?
Try to create hardware profile without name, the submit button should be disabled.

## Test Impact
Added unit test for the validation.

## Request review criteria:

If you have UI changes: 


dpanshug:
@Gkrumbach07 Need your review on this, are there any more validations required for it?

frontend/src/pages/hardwareProfiles/manage/validationUtils.ts
```diff
@@ -0,0 +1,37 @@
+import { z } from 'zod';
+import { IdentifierResourceType, TolerationEffect, TolerationOperator } from '~/types';
+
+const identifierSchema = z.object({
+  displayName: z.string(),
+  identifier: z.string(),
+  resourceType: z.nativeEnum(IdentifierResourceType),
```
pnaik1, line 7: @dpanshug resourceType is `optional` here; this will eventually solve the `create hardware profile` button  been disabled when resourceType is `Other`

pnaik1 (changes requested):

frontend/src/pages/hardwareProfiles/manage/validationUtils.ts
```diff
@@ -0,0 +1,37 @@
...
+  tolerations: z.array(tolerationSchema).optional(),
+  nodeSelector: nodeSelectorSchema.optional(),
+  name: z
+    .string()
+    .trim()
+    .min(1)
+    .max(253)
+    .regex(/^[a-z0-9]([-a-z0-9]*[a-z0-9])?$/),
+  description: z.string().optional(),
+  visibility: z.array(z.any()),
```
pnaik1, line 36: Can we also add zod validation for `visibility`

pnaik1 (changes requested):

frontend/src/pages/hardwareProfiles/manage/validationUtils.ts
dpanshug: I think if its not optional its already in validation, do you mean to add few more parameters to it?

dpanshug: It can only take string of array as that's the type of the data required. Changing it to `z.array(z.string())`

frontend/src/pages/hardwareProfiles/manage/validationUtils.ts
```diff
@@ -0,0 +1,37 @@
...
+  effect: z.nativeEnum(TolerationEffect).optional(),
+  tolerationSeconds: z.number().optional(),
+});
+
+export const nodeSelectorSchema = z.record(z.string().min(1), z.string().min(1));
+
+export const manageHardwareProfileValidationSchema = z.object({
+  displayName: z.string().trim().min(1, 'Display name is required'),
+  enabled: z.boolean(),
+  identifiers: z.array(identifierSchema),
```
Gkrumbach07, line 26: We need to validate that there are at least a memory and cpu resource

Gkrumbach07: we have some of these validation utils already used in `frontend/src/pages/hardwareProfiles/utils.ts`

We should merge those validation rules into your schema. Then update the helpers in that util file to work off of your schema. Also just do general cleanup around there so it all feels unified.

Gkrumbach07 (changes requested):
This looks good but we need ot have this validation on the modals in the form as well. 

Think of those modals as separate forms that have their own schema. You already have it split up in validationUtils, just use it for those modals now.

frontend/src/pages/hardwareProfiles/manage/validationUtils.ts
dpanshug: The current form works without CPU and Memory identifier, are you sure we want to add validation for having atleast CPU and meemory resource?

Gkrumbach07 (changes requested):
There are still usages of the utils and schemas from `frontend/src/pages/hardwareProfiles/utils.ts` that use `hardwareProfileWarningSchema`

Let me know if you have questions on what I mean by combine these

frontend/src/pages/hardwareProfiles/utils.ts
```diff
@@ -152,12 +152,10 @@ export const validateProfileWarning = (
   const warnings = hardwareProfileWarningSchema.safeParse({ isDefault: false, value: identifiers });
```
Gkrumbach07, line 152: lets remove this isDefault becuase that is deprecated logic now. We dont have "default" hardware profiles anymore

frontend/src/concepts/hardwareProfiles/validationUtils.ts
```diff
@@ -185,204 +181,7 @@ export const isHardwareProfileConfigValid = (data: HardwareProfileConfig): boole
...
-            type: HardwareProfileWarningType.OTHER,
-          },
-        });
-      }
-    }
-  });
+export const hardwareProfileWarningSchema = z.object({
+  isDefault: z.boolean(),
+  value: z.array(identifierSchema),
+});
```
Gkrumbach07, line 187: why still have this one at all anymore. can we just use the other one?

frontend/src/pages/hardwareProfiles/utils.ts
```diff
@@ -152,12 +152,10 @@ export const validateProfileWarning = (
...
-      if ('params' in issue && typeof issue.params !== 'undefined') {
-        warningMessages.push({
-          message: issue.message,
-          type: issue.params.type,
-        });
-      }
+      warningMessages.push({
+        message: issue.message,
+        type: HardwareProfileWarningType.OTHER,
+      });
```
Gkrumbach07, line 158: I think we can get rid of this HardwareProfileWarning in favor of having custom typed zod errors. we should be able to set a. custom error type in the schema. doing that would make this function simply do a safe parse of the whole hardware profile

frontend/src/pages/hardwareProfiles/utils.ts
```diff
@@ -137,7 +137,7 @@ export const createHardwareProfileWarningTitle = (hardwareProfile: HardwareProfi
   return `${complete ? 'Invalid' : 'Incomplete'} hardware profile`;
 };
 
-export const createIdentifierWarningMessage = (message: string, isDefault: boolean): string =>
+export const createIdentifierWarningMessage = (message: string, isDefault?: boolean): string =>
```
Gkrumbach07, line 140: we dont need is default warnings anymore

frontend/src/pages/hardwareProfiles/manage/validationUtils.ts
```diff
@@ -0,0 +1,155 @@
...
+    enabled: z.boolean(),
+    identifiers: z.array(identifierSchema),
+    tolerations: z.array(tolerationSchema).optional(),
+    nodeSelector: nodeSelectorSchema.optional(),
+    name: z
+      .string()
+      .trim()
+      .min(1)
+      .max(253)
+      .regex(/^[a-z0-9]([-a-z0-9]*[a-z0-9])?$/),
```
Gkrumbach07, line 138: where does this regex come from? can we store it in a const

frontend/src/pages/hardwareProfiles/manage/validationUtils.ts
```diff
@@ -0,0 +1,155 @@
...
+    }
+  } catch (e) {
+    ctx.addIssue({
+      code: z.ZodIssueCode.custom,
+      message: createIdentifierWarningMessage(
+        `The resource count for ${
+          identifier.resourceType ?? identifier.displayName
+        } has an invalid unit.`,
+      ),
+      params: { type: HardwareProfileWarningType.OTHER },
```
Gkrumbach07, line 111: instead of params can we put the custom type in code... we probably dont need an other type anymore and just use the missing cpu type as thats the only specific one

frontend/src/concepts/hardwareProfiles/validationUtils.ts
dpanshug: If we're removing isDefault then it's fine removing this, and continue with the other one.

frontend/src/pages/hardwareProfiles/utils.ts
Gkrumbach07: you are right. i dont think we can add it there. I was thinking of a user case we have that looks like this. We are fine then

```tsx
 code: z.ZodIssueCode.custom,
params: { code: ValidationErrorCodes.DUPLICATE },
```

Gkrumbach07: however we dont need to have a custom type called other. we should chnage it to be more specific on what type of error it i

frontend/src/pages/hardwareProfiles/manage/validationUtils.ts
```diff
@@ -0,0 +1,160 @@
...
+  const checkDecimal = (value: number, fieldName: string) => {
+    if (!Number.isInteger(value)) {
+      ctx.addIssue({
+        code: z.ZodIssueCode.custom,
+        message: createIdentifierWarningMessage(
+          `${fieldName} for ${
+            identifier.resourceType ?? identifier.displayName
+          } cannot be a decimal.`,
+        ),
+        path: [HardwareProfileWarningType.CANNOT_BE_DECIMAL],
```
Gkrumbach07, line 56: path is not used like this. you should use param to add the cusotm type

```tsx
 params: { code: HardwareProfileWarningType.CANNOT_BE_DECIMAL },
```

Then path should be what specific item has an error. in this specific case it would be something like `["identifiers", index, "minValue" | "minValue" | "defaultValue"]`
I would pass that path in as part of check decimal

Gkrumbach07: do this for all the custom errors

Gkrumbach07 (changes requested):

frontend/src/concepts/hardwareProfiles/types.ts
```diff
@@ -9,8 +9,11 @@ export type WarningNotification = {
 };
 
 export enum HardwareProfileWarningType {
-  HARDWARE_PROFILES_MISSING_CPU_MEMORY,
-  OTHER,
+  HARDWARE_PROFILES_MISSING_CPU_MEMORY = 'hardware_profiles_missing_cpu_memory',
+  CANNOT_BE_NEGATIVE = 'cannot_be_negative',
+  CANNOT_BE_DECIMAL = 'cannot_be_decimal',
+  INVALID_UNIT = 'invalid_unit',
+  OTHER = 'other',
```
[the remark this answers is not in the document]
dpanshug: We still use that in `validateProfileWarning` function for the edge case if params doesn't exists.

frontend/src/pages/hardwareProfiles/manage/validationUtils.ts
Gkrumbach07: You are not using path now. we wont be able to assign the error to that specific field without it

Gkrumbach07 (approved):
/lgtm

PR #3968, opened 2025-03-24
Prefill visibility correctly when updating or duplicating a hardware profile

DaoDaoNoCode:
JIRA: [RHOAIENG-21961](https://issues.redhat.com/browse/RHOAIENG-21961)

## Description

Instead of using only `visibility` as `string[]` to control the radio selections, I made it an object type `{isUnlimited: boolean, features: string[]}`, so it can control the radio selection and the features separately. In that way, when updating or duplicating an existing hardware profile, we can successfully choose the correct radio button and prefill the value.

Screenshot 2025-03-24 at 10 47 36 AM

## How Has This Been Tested?

1. Create a hardware profile or update a hardware profile 
2. In the Visibility section, choose "Limited visibility" and choose some features
3. Save the hardware profile
4. Edit/Duplicate the hardware profile you just saved
5. The Visibility section should be pre-filled with the features you selected just now

## Test Impact

Added unit tests to verify the annotations are set as expected

## Request review criteria:

If you have UI changes: 


frontend/src/api/k8s/hardwareProfiles.ts
```diff
@@ -31,7 +32,7 @@ export const assembleHardwareProfile = (
   hardwareProfileName: string,
   data: HardwareProfileKind['spec'],
   namespace: string,
-  visibility: string[] = [],
+  visibility: HardwareProfileVisibility = { isUnlimited: true, features: [] },
```
Gkrumbach07, line 35: So instead of passing this isUnlimited object, we should instead just use the same array as before

the difference is that in the section and form we should always interpret an empty feature array as unlimited, however we will have some local state to track which option they are on for the visibility form section

so something like 

```tsx
const VisSectionComp({data, setData}) => {
   const [limitedOptionSelected, set:LimitedOptionSelected] = useState(false)

   const isLimitedSelected = limitedOptionSelected || data.length > 0

   return (
      <Radio selected={isLimitedSelected}> Limited option </Radio>
      <Radio selected={!isLimitedSelected}> Unlimited option </Radio>
   )
}
```
So basically the state is still just an array, but to be able to support a situation where they have nothing selected for the limited option, we store that state in the section's local state and not in the form state.

Gkrumbach07: I see. I think you can still do this technically but just saving the prev limited option. SO you would add another state var for that and when you toggle switching you can pull from that in addition to setting limitedOptionSelected to true. this keeps our form state clean

Gkrumbach07 (approved):
Code looks good

nananosirova:
Tested locally: 

https://github.com/user-attachments/assets/c6adab28-2b08-4ebb-a250-3cfaaa3be058

/lgtm

PR #3969, opened 2025-03-24
Fixing workbench image pull event success bug

katieperry4:
Closes: [RHOAIENG-21741](https://issues.redhat.com/browse/RHOAIENG-21741)

## Description
- Removed code that assumed steps were successful if later steps succeeded. 
- Also removed the 'pod created' step as it was never being sent by Kubernetes and therefore the event never succeeded. (Might be added again in the [follow up ticket.](https://issues.redhat.com/browse/RHOAIENG-22056)
- Added additional check for 'ose-oauth-proxy' for clarity in differentiating between oauth and notebook events.
- Fixed a typo.

## How Has This Been Tested?
I ran `npm run test` with 6 test failures due to the removal of the 'pod created' event, otherwise all other tests passed.

## Test Impact
No additional tests were written, but the unit tests were temporarily modified to handle the removal of the 'pod created' event and will be reverted/reinstated during the [follow up ticket](https://issues.redhat.com/browse/RHOAIENG-22056).

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/concepts/__tests__/StartNotebookModal.spec.tsx
+++ b/frontend/src/concepts/__tests__/StartNotebookModal.spec.tsx
@@ -5,12 +5,11 @@ import StartNotebookModal from '~/concepts/notebooks/StartNotebookModal';
 import {
   mockCompletedStates,
   mockFailedStates,
-  //mockInitialStates,
   mockInProgressStates,
 } from '~/concepts/__tests__/mockNotebookStates';
 
 describe('Start Notebook modal', () => {
-  //TODO: RHOAIENG-22056 uncomment this test once 'pod created' event is reintroduced AND uncomment the mockInitialStates import
+  //TODO: RHOAIENG-22056 uncomment this test once 'pod created' event is reintroduced
   // it('should show initial notebook startup status', async () => {
   //   const mockData = mockInitialStates;
   //   render(
@@ -52,7 +51,6 @@ describe('Start Notebook modal', () => {
   //   expect(steps[13]).toHaveTextContent('Server started');
   // });
 
-  //TODO: RHOAIENG-22056 increment .toHaveLength after reintroducting the 'pod created' event
   it('should show failed notebook startup status', async () => {
     const mockData = mockFailedStates;
     render(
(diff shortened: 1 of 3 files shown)


frontend/src/concepts/__tests__/StartNotebookModal.spec.tsx
```diff
@@ -5,52 +5,54 @@ import StartNotebookModal from '~/concepts/notebooks/StartNotebookModal';
 import {
   mockCompletedStates,
   mockFailedStates,
-  mockInitialStates,
+  //mockInitialStates,
```
Griffin-Sullivan, line 8: You can just remove this line. Your IDE will probably yell at you when you uncomment everything to bring this import back.

frontend/src/concepts/__tests__/StartNotebookModal.spec.tsx
```diff
@@ -154,10 +159,11 @@ describe('Start Notebook modal', () => {
     const stepper = screen.getByTestId('notebook-startup-steps');
     expect(stepper).toBeTruthy();
     const steps = screen.getAllByRole('listitem');
-    expect(steps).toHaveLength(14);
-    expect(screen.getAllByTestId('step-status-Success')).toHaveLength(14);
+    expect(steps).toHaveLength(13);
+    expect(screen.getAllByTestId('step-status-Success')).toHaveLength(13);
   });
 
+  //TODO: RHOAIENG-22056 increment both .toHaveLength after reintroducting the 'pod created' event
```
andrewballantyne, line 166: Move these comments next to the line you modified... easier to trace and follow through

frontend/src/utilities/notebookControllerUtils.ts
```diff
@@ -356,14 +356,8 @@ export const getNotebookEventStatus = (
         };
     }
   }
-
+  //for events not related to oauth
```
andrewballantyne, line 359:
```suggestion
  // For notebook events
```
Best to call out what they are -- it's a 2 container system -- although you are technically correct, it's a bit more helpful to say they are notebook related to run home the goal of this -- honestly an "else" between the two blocks is probably better lol... but no need to do that here.

andrewballantyne (commented):
Can you test successful deploys and see how that looks too. We may not have time to do both separately for a given release -- so lets make sure our state we are left with looks good to succeed (it may not be and we may need to pivot)

Griffin-Sullivan (approved):
/lgtm 

I would remove the comment in the import but not a necessity. Tested on my cluster and the Pod created event is no longer there. 

image

katieperry4:
@andrewballantyne @Griffin-Sullivan I made the changes suggested and I did try it out on a workbench that succeeds and all looks okay to me!

frontend/src/utilities/notebookControllerUtils.ts
```diff
@@ -553,18 +547,6 @@ export const useNotebookProgress = (
     }
   });
 
-  // Some events are not always received, mark all steps prior to the last completed step complete.
```
andrewballantyne, line 556: This is the line I was talking about in standup @katieperry4

katieperry4: ahh makes sense! I'll get going on the next ticket to fix that~

PR #3973, opened 2025-03-25
Fix for when modelRegistryNamespace is undefined and no DSC status in response

YuliaKrimerman:
Closes https://issues.redhat.com/browse/RHOAIENG-19890

## Description

This PR improves error handling when the Model Registry namespace is not configured(Case 1) or when the DSC status returns an empty object(case 2). Previously, we were using fallback empty strings (modelRegistryNamespace || '') which masked the underlying issue and could lead to unexpected behavior.
Key changes:
Removed fallback empty strings - We no longer use modelRegistryNamespace || '' or registriesNamespace || '' in component logic, instead properly handling the undefined case.
Component-specific error handling - Each component that requires the Model Registry namespace now shows a specific error page when the namespace is undefined:
ModelRegistriesPermissions.tsx - Shows an error page when trying to manage permissions
CreateModal.tsx - Shows an error page when trying to create/edit a Model Registry
ModelRegistrySelectorContext.tsx - Sets an error state in the context when namespace is undefined
Improved empty DSC status handling - Added proper error handling in AreaContext.tsx to detect when fetchDscStatus returns an empty object ({}), showing a clear error page without the navbar instead of crashing the application.
Case 1 
Screenshot 2025-03-25 at 2 36 13 PM
Screenshot 2025-03-25 at 2 36 20 PM

Case 2 - it was agreed that when dsc status is undefined the whole app is not usable anymore, and we just show this error page. Note the error message in the screenshot is mocked and in reality will be the error message from the fetchDscStatus
Screenshot 2025-03-25 at 12 08 46 PM

## How Has This Been Tested?

Not too many useful ways to test except for in code. 

## Test Impact

No test impact 

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/__mocks__/mockDscStatus.ts
+++ b/frontend/src/__mocks__/mockDscStatus.ts
@@ -137,7 +137,7 @@ export const mockDscStatus = ({
   installedComponents: Object.values(StackComponent).reduce(
     (acc, component) => ({
       ...acc,
-      [component]: installedComponents[component] ?? true,
+      [component]: installedComponents[component] ?? false,
     }),
     {},
   ),
--- a/frontend/src/__mocks__/mockDscStatus.ts
+++ b/frontend/src/__mocks__/mockDscStatus.ts
@@ -33,7 +33,10 @@ export const mockDscStatus = ({
       ],
     },
   },
-  installedComponents,
+  installedComponents = Object.values(StackComponent).reduce(
+    (acc, component) => ({ ...acc, [component]: true }),
+    {},
+  ),
   conditions = [],
   phase = 'Ready',
 }: MockDscStatus): DataScienceClusterKindStatus => ({
@@ -134,7 +137,7 @@ export const mockDscStatus = ({
   installedComponents: Object.values(StackComponent).reduce(
     (acc, component) => ({
       ...acc,
-      [component]: installedComponents?.[component] ?? false,
+      [component]: installedComponents[component] ?? true,
     }),
     {},
   ),
--- a/frontend/src/concepts/areas/AreaContext.tsx
+++ b/frontend/src/concepts/areas/AreaContext.tsx
@@ -1,11 +1,13 @@
 import * as React from 'react';
-import { Alert, Bullseye, Spinner } from '@patternfly/react-core';
+import { Bullseye, Page, Spinner } from '@patternfly/react-core';
 import useFetchDscStatus from '~/concepts/areas/useFetchDscStatus';
 import useFetchDsciStatus from '~/concepts/areas/useFetchDsciStatus';
 import {
   DataScienceClusterInitializationKindStatus,
   DataScienceClusterKindStatus,
 } from '~/k8sTypes';
+import ApplicationsPage from '~/pages/ApplicationsPage';
+import RedirectErrorState from '~/pages/external/RedirectErrorState';
 
 type AreaContextState = {
   /**
@@ -34,11 +36,22 @@ const AreaContextProvider: React.FC<AreaContextProps> = ({ children }) => {
   const loaded = loadedDsc && loadedDsci;
 
   const contextValue = React.useMemo(() => ({ dscStatus, dsciStatus }), [dscStatus, dsciStatus]);
-  if (error) {
+
+  if (error || (dscStatus && Object.keys(dscStatus).length === 0)) {
     return (
-      <Alert isInline variant="danger" title="Problem loading component state">
-        {error.message}
-      </Alert>
+      <Page>
+        <ApplicationsPage
+          loaded
+          empty={false}
+          loadError={error}
+          loadErrorPage={
+            <RedirectErrorState
+              title="Could not load component state"
+              errorMessage={error?.message}
+            />
+          }
+        />
+      </Page>
     );
   }
 
--- a/frontend/src/concepts/areas/AreaContext.tsx
+++ b/frontend/src/concepts/areas/AreaContext.tsx
@@ -37,7 +37,7 @@ const AreaContextProvider: React.FC<AreaContextProps> = ({ children }) => {
 
   const contextValue = React.useMemo(() => ({ dscStatus, dsciStatus }), [dscStatus, dsciStatus]);
 
-  if (error || !dscStatus || Object.keys(dscStatus).length === 0) {
+  if (error || (loaded && (!dscStatus || Object.keys(dscStatus).length === 0))) {
     return (
       <Page>
         <ApplicationsPage loaded empty={false}>
--- a/frontend/src/concepts/areas/AreaContext.tsx
+++ b/frontend/src/concepts/areas/AreaContext.tsx
@@ -37,20 +37,15 @@ const AreaContextProvider: React.FC<AreaContextProps> = ({ children }) => {
 
   const contextValue = React.useMemo(() => ({ dscStatus, dsciStatus }), [dscStatus, dsciStatus]);
 
-  if (error || (dscStatus && Object.keys(dscStatus).length === 0)) {
+  if (error || !dscStatus || Object.keys(dscStatus).length === 0) {
     return (
       <Page>
-        <ApplicationsPage
-          loaded
-          empty={false}
-          loadError={error}
-          loadErrorPage={
-            <RedirectErrorState
-              title="Could not load component state"
-              errorMessage={error?.message}
-            />
-          }
-        />
+        <ApplicationsPage loaded empty={false}>
+          <RedirectErrorState
+            title="Could not load component state"
+            errorMessage={error?.message}
+          />
+        </ApplicationsPage>
       </Page>
     );
   }
@@ -65,5 +60,4 @@ const AreaContextProvider: React.FC<AreaContextProps> = ({ children }) => {
 
   return <AreaContext.Provider value={contextValue}>{children}</AreaContext.Provider>;
 };
-
 export default AreaContextProvider;
--- a/frontend/src/concepts/modelRegistry/apiHooks/useModelRegistryServices.ts
+++ b/frontend/src/concepts/modelRegistry/apiHooks/useModelRegistryServices.ts
@@ -67,9 +67,7 @@ export const useModelRegistryServices = (
     { ...accessReviewResource, namespace: namespace || '' },
     !!namespace,
   );
-  const [rulesReviewStatus, rulesReviewLoaded, refreshRulesReview] = useRulesReview(
-    namespace || '',
-  );
+  const [rulesReviewStatus, rulesReviewLoaded, refreshRulesReview] = useRulesReview(namespace);
 
   const serviceNames = React.useMemo(() => {
     if (!rulesReviewLoaded) {
@@ -83,19 +81,17 @@ export const useModelRegistryServices = (
       .flatMap((rule) => rule.resourceNames || []);
   }, [rulesReviewLoaded, rulesReviewStatus]);
 
-  const callback = React.useCallback<FetchStateCallbackPromise<ServiceKind[]>>(() => {
-    if (!namespace) {
-      return Promise.reject(new NotReadyError('No registries namespace could be found'));
-    }
-
-    return listServicesOrFetchThemByNames(
-      allowList,
-      accessReviewLoaded,
-      rulesReviewLoaded,
-      namespace,
-      serviceNames,
-    );
-  }, [allowList, accessReviewLoaded, rulesReviewLoaded, serviceNames, namespace]);
+  const callback = React.useCallback<FetchStateCallbackPromise<ServiceKind[]>>(
+    () =>
+      listServicesOrFetchThemByNames(
+        allowList,
+        accessReviewLoaded,
+        rulesReviewLoaded,
+        namespace,
+        serviceNames,
+      ),
+    [allowList, accessReviewLoaded, rulesReviewLoaded, serviceNames, namespace],
+  );
 
   const [modelRegistryServices, isLoaded, error] = useFetchState(callback, [], {
     initialPromisePurity: true,
--- a/frontend/src/concepts/modelRegistry/apiHooks/useModelRegistryServices.ts
+++ b/frontend/src/concepts/modelRegistry/apiHooks/useModelRegistryServices.ts
@@ -35,9 +35,13 @@ const listServicesOrFetchThemByNames = async (
   allowList: boolean,
   accessReviewLoaded: boolean,
   rulesReviewLoaded: boolean,
-  namespace: string,
+  namespace: string | undefined,
   serviceNames?: string[],
 ): Promise<ServiceKind[]> => {
+  if (!namespace) {
+    throw new NotReadyError('No registries namespace could be found');
+  }
+
   if (!accessReviewLoaded || !rulesReviewLoaded) {
     throw new NotReadyError('Access review or Rules review not loaded');
   }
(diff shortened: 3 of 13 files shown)


frontend/src/concepts/areas/AreaContext.tsx
```diff
@@ -34,11 +36,22 @@ const AreaContextProvider: React.FC<AreaContextProps> = ({ children }) => {
...
+          empty={false}
+          loadError={error}
+          loadErrorPage={
+            <RedirectErrorState
+              title="Could not load component state"
+              errorMessage={error?.message}
+            />
+          }
+        />
+      </Page>
```
mturley, line 54: In the case where the dscStatus is empty/missing but there is no `loadError` here, the `loadErrorPage` doesn't render and you get a blank page. Since this entire return statement is for the error/missing case and we don't need to conditionally show or not show an error, I don't think we need to use the `loadError`/`loadErrorPage` functionality of ApplicationsPage. Instead we can just pass the RedirectErrorState to it as children:

```tsx
<ApplicationsPage loaded empty={false}>
  <RedirectErrorState
    title="Could not load component state"
    errorMessage={error?.message}
  />
</ApplicationsPage>
```

frontend/src/concepts/modelRegistry/context/ModelRegistrySelectorContext.tsx
```diff
@@ -43,30 +43,50 @@ const EnabledModelRegistrySelectorContextProvider: React.FC<React.PropsWithChild
...
+  const namespaceForHook = React.useMemo(
+    () => modelRegistryNamespace || '',
+    [modelRegistryNamespace],
+  );
+  const {
+    modelRegistryServices = [],
+    isLoaded,
+    error: servicesError,
+    refreshRulesReview,
+  } = useModelRegistryServices(namespaceForHook);
```
mturley, line 67: You're still passing an empty string for the namespace here into `useModelRegistryServices`, which is causing it to make API requests that will fail. It's not as serious an issue anymore now that we don't try to use the responses from those requests, but we still should just not make the requests in the first place if there's no namespace.

The way to avoid that is to remove your `namespaceForHook` and just let `modelRegistryNamespace` be possibly undefined and pass it in, and change `useModelRegistryServices` so it takes `string | undefined` and properly handles the undefined case.

To do that, make sure the hooks called inside `useModelRegistryServices` also prevent their requests when the namespace is undefined:
* For `useAccessReview`, it already has an optional second argument you can pass called `shouldRunCheck`, so you can just pass `!!namespace` for that.
* For `useRulesReview`, we can edit it to make `namespace` a `string | undefined` and then wrap the contents of its `refreshRulesReview` function with `if (namespace) { ... }`.
* The rest of the function should behave properly - `listServicesOrFetchThemByNames` already throws a `NotReadyError` if it doesn't have what it needs from the access/rules reviews which is what we want in this case.

frontend/src/concepts/modelRegistry/context/ModelRegistrySelectorContext.tsx
```diff
@@ -43,30 +43,50 @@ const EnabledModelRegistrySelectorContextProvider: React.FC<React.PropsWithChild
...
-    () => ({
+  const contextValue = React.useMemo(() => {
+    const error = !modelRegistryNamespace
+      ? new Error('No registries namespace could be found')
+      : servicesError;
+    return {
       modelRegistryServicesLoaded: isLoaded,
       modelRegistryServicesLoadError: error,
-      modelRegistryServices,
+      modelRegistryServices: modelRegistryNamespace ? modelRegistryServices : [],
```
mturley, line 76: I don't think you need this condition, `modelRegistryServices` should already be `[]` if the useModelRegistryServices hook fails or isn't ready

frontend/src/pages/modelRegistry/ModelRegistryCoreLoader.tsx
```diff
@@ -55,6 +58,22 @@ const ModelRegistryCoreLoader: React.FC<ModelRegistryCoreLoaderProps> =
...
+          loaded
+          empty={false}
+          loadError={new Error('No registries namespace could be found')}
+          loadErrorPage={
+            <RedirectErrorState
+              title="Could not load component state"
+              errorMessage="No registries namespace could be found"
+            />
+          }
+        />
```
mturley, line 73: You aren't actually rendering what you pass to `new Error` here, you honestly could just do the same thing I suggested for AreaContext and omit `loadError` and `loadErrorPage` and instead pass the RedirectErrorState as children.

frontend/src/pages/modelRegistry/ModelRegistryCoreLoader.tsx
```diff
@@ -55,6 +58,22 @@ const ModelRegistryCoreLoader: React.FC<ModelRegistryCoreLoaderProps> =
...
+      );
+    }
+
     if (modelRegistryServicesLoadError) {
```
mturley, line 77: Interesting - in ModelRegistrySelectorContext you have `modelRegistryServicesLoadError` possibly be a new Error with your "No registries namespace could be found" message, but it doesn't get rendered here since you have the above special rendering for the missing namespace. I think maybe we should show the ApplicationsPage version of the error screen no matter the cause of the error - we don't need to show "Could not load component state", it's still a "Model registry load error".

Let's get rid of the `if (!modelRegistryNamespace) {` and replace the `<Bullseye><Alert>` version here with your `ApplicationsPage` with `RedirectErrorState` child, have its title be "Model registry load error", `loadError` be `modelRegistryServicesLoadError`, and the `errorMessage` of the `RedirectErrorState` can be `modelRegistryServicesLoadError.message`. Then you don't even need to consume AreaContext here - we'll hit this nice error screen simply because of what you did in ModelRegistrySelectorContext to make the services fetches fail gracefully.

frontend/src/pages/modelRegistrySettings/CreateModal.tsx
```diff
@@ -276,6 +278,22 @@ const CreateModal: React.FC<CreateModalProps> = ({ onClose, refresh, modelRegist
...
+        empty={false}
+        loadError={new Error('No registries namespace could be found')}
+        loadErrorPage={
+          <RedirectErrorState
+            title="Could not load component state"
+            errorMessage="No registries namespace could be found"
+          />
+        }
+      />
+    );
```
mturley, line 294: Same rendering change here - we don't need `loadError` / `loadErrorPage` if we just pass the rendered error as children, it's kinda weird to not use the value you pass to `new Error()`.

frontend/src/pages/modelRegistrySettings/ModelRegistriesPermissions.tsx
```diff
@@ -48,6 +49,22 @@ const ModelRegistriesManagePermissions: React.FC = () => {
...
+        loaded
+        empty={false}
+        loadError={new Error('No registries namespace could be found')}
+        loadErrorPage={
+          <RedirectErrorState
+            title="Could not load component state"
+            errorMessage="No registries namespace could be found"
+          />
+        }
+      />
```
mturley, line 64: Same deal here with the rendering change

frontend/src/pages/modelRegistrySettings/ModelRegistrySettings.tsx
```diff
@@ -8,30 +8,48 @@ import {
...
+
   const [modelRegistries, mrloaded, loadError, refreshModelRegistries] =
     useModelRegistriesBackend();
   const roleBindings = useContextResourceData<RoleBindingKind>(useModelRegistryRoleBindings());
-  const { refreshRulesReview } = React.useContext(ModelRegistrySelectorContext);
   const loaded = mrloaded && roleBindings.loaded;
 
   const refreshAll = React.useCallback(
-    () => Promise.all([refreshModelRegistries(), refreshRulesReview()]),
-    [refreshModelRegistries, refreshRulesReview],
```
mturley, line 32: We can't remove this `refreshRulesReview()` call. That's what updates the registry selector after you create/delete a registry. Good call adding the `roleBindings.refresh` here though, although I'm curious what made you think to do that?

frontend/src/pages/modelRegistrySettings/ModelRegistrySettings.tsx
```diff
@@ -8,30 +8,48 @@ import {
...
+        empty={false}
+        loadError={new Error('No registries namespace could be found')}
+        loadErrorPage={
+          <RedirectErrorState
+            title="Could not load component state"
+            errorMessage="No registries namespace could be found"
+          />
+        }
+      />
+    );
```
mturley, line 51: Same rendering change here.

mturley (changes requested):
A handful of small things here, this is looking great.

Also:
* You missed a spot where we pass `modelRegistryNamespace || ''`: in ModelRegistriesPermissions, when we call `useModelRegistryNamespaceCR` we pass it in that way. We should modify `useModelRegistryNamespaceCR` so it can take an undefined namespace, and if so its callback can `return Promise.reject(new NotReadyError('...'));` like it currently does when MR is disabled.
* In model catalog, the details page for a model refuses to render when the MR services fetch fails, because it uses that data to determine whether to show the Register button. Ideally though, model catalog should still work even if MR is broken. Can we instead change `frontend/src/pages/modelCatalog/screens/ModelDetailsPage.tsx` so that it doesn't include `modelRegistryServicesLoadError` in the `loadError` it passes to ApplicationsPage, and instead have the `registerModelButton` function return null if `modelRegistryServicesLoadError` exists?

frontend/src/concepts/modelRegistry/apiHooks/useModelRegistryServices.ts
```diff
@@ -56,9 +60,16 @@ export type ModelRegistryServicesResult = {
...
+export const useModelRegistryServices = (
+  namespace: string | undefined,
+): ModelRegistryServicesResult => {
+  const [allowList, accessReviewLoaded] = useAccessReview(
+    { ...accessReviewResource, namespace: namespace || '' },
+    !!namespace,
+  );
+  const [rulesReviewStatus, rulesReviewLoaded, refreshRulesReview] = useRulesReview(
+    namespace || '',
+  );
```
mturley, line 72: `namespace || ''` here is still kicking the can -- we still need to change `useRulesReview` so it works similarly to `useAccessReview`, the namespace passed to it can be undefined and it has an `if (shouldRunCheck) {` before it calls `checkAccess`. (although in that case it could just be `if (namespace) {` since that's the only thing that should stop it from its check unlike useAccessReview).

frontend/src/concepts/modelRegistry/apiHooks/useModelRegistryServices.ts
```diff
@@ -72,17 +83,19 @@ export const useModelRegistryServices = (namespace: string): ModelRegistryServic
...
-        rulesReviewLoaded,
-        namespace,
-        serviceNames,
-      ),
-    [allowList, accessReviewLoaded, rulesReviewLoaded, serviceNames, namespace],
-  );
+  const callback = React.useCallback<FetchStateCallbackPromise<ServiceKind[]>>(() => {
+    if (!namespace) {
+      return Promise.reject(new NotReadyError('No registries namespace could be found'));
+    }
```
mturley, line 89: You don't need this since you added the NotReadyError inside `listServicesOrFetchThemByNames`, you can let that function handle this error case

frontend/src/pages/modelCatalog/screens/ModelDetailsPage.tsx
```diff
@@ -66,38 +66,47 @@ const ModelDetailsPage: React.FC = conditionalArea(
     [modelCatalogSources, decodedParams],
   );
 
-  const registerModelButton = (isSecondary = false) =>
```
mturley, line 69: You don't want to get rid of the `isSecondary` argument here. It was making this register button be secondary in the details page header but primary in the ilab info popover, we still need that behavior.

frontend/src/pages/modelCatalog/screens/ModelDetailsPage.tsx
```diff
@@ -115,7 +124,7 @@ const ModelDetailsPage: React.FC = conditionalArea(
       footerContent={
         <ActionList>
           <ActionListGroup>
-            <ActionListItem>{registerModelButton(true)}</ActionListItem>
+            {registerModelButton()}
```
mturley, line 127: Referring to this - we should still pass `true` here.

frontend/src/pages/modelRegistrySettings/CreateModal.tsx
```diff
@@ -280,17 +280,9 @@ const CreateModal: React.FC<CreateModalProps> = ({ onClose, refresh, modelRegist
...
-        loadError={new Error('No registries namespace could be found')}
-        loadErrorPage={
-          <RedirectErrorState
-            title="Could not load component state"
-            errorMessage="No registries namespace could be found"
-          />
-        }
-      />
+      <ApplicationsPage loaded empty={false}>
+        <RedirectErrorState title="Could not load component state" errorMessage={error?.message} />
```
mturley, line 284: In this specific case you don't want to use `error?.message`, that error comes from submitting the form so it'll never be defined here. You can pass "No registries namespace could be found" for the errorMessage.

mturley (changes requested):
A couple followups

frontend/src/__mocks__/mockDscStatus.ts
```diff
@@ -134,7 +137,7 @@ export const mockDscStatus = ({
   installedComponents: Object.values(StackComponent).reduce(
     (acc, component) => ({
       ...acc,
-      [component]: installedComponents?.[component] ?? false,
+      [component]: installedComponents[component] ?? true,
```
mturley, line 140: This should not default to true, it should stay as it was. The only change to this file should be the default value you added on line 36, if the user passes in installedComponents and that default is not used we shouldn't default everything they excluded to true.

mturley (changes requested):

frontend/src/concepts/areas/AreaContext.tsx
```diff
@@ -34,11 +36,17 @@ const AreaContextProvider: React.FC<AreaContextProps> = ({ children }) => {
   const loaded = loadedDsc && loadedDsci;
 
   const contextValue = React.useMemo(() => ({ dscStatus, dsciStatus }), [dscStatus, dsciStatus]);
-  if (error) {
+
+  if (error || !dscStatus || Object.keys(dscStatus).length === 0) {
```
mturley, line 40: I just noticed that if the network is slow and the DSC status isn't loaded yet, we show this error screen momentarily instead of the loading screen. Can you change this condition to:
```ts
if (error || (loaded && (!dscStatus || Object.keys(dscStatus).length === 0))) {
```

mturley (changes requested):
In addition to the change below: when the MR namespace is not in the DSC status, the model catalog details page just gets stuck in loading state because its `loaded` value remains false when the MR services fetch fails. Can you change the `const loaded` declaration on line 56 of ModelDetailsPage to:
```ts
// Allow the page to render if MR services can't be loaded, just hide the Register Model button
  const loaded =
    (modelRegistryServicesLoaded || !!modelRegistryServicesLoadError) && modelCatalogSources.loaded;
```

mturley (approved):
Tested 6 permutations: as both an admin and non-admin user, with the DSC status fetch returning `{}`, `null`, and full contents excluding the registriesNamespace. All work as expected.

LGTM! Good stuff!

PR #3974, opened 2025-03-25
Outdated image workbench alerts

rsun19:
https://issues.redhat.com/browse/RHOAIENG-348

## Description

Added the update and latest image labels, plus the update image modal

Screenshot 2025-03-25 at 2 19 18 PM
Screenshot 2025-03-25 at 2 19 53 PM

## How Has This Been Tested?

Create an outdated image version by adding this annotation: `'opendatahub.io/image-tag-outdated': 'true',`. There should be a label that appears and lets you update it through a modal.

## Test Impact

Added cypress tests

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/api/k8s/notebooks.ts
+++ b/frontend/src/api/k8s/notebooks.ts
@@ -10,7 +10,13 @@ import {
 } from '@openshift/dynamic-plugin-sdk-utils';
 import * as _ from 'lodash-es';
 import { NotebookModel } from '~/api/models';
-import { K8sAPIOptions, KnownLabels, NotebookKind } from '~/k8sTypes';
+import {
+  ImageStreamKind,
+  ImageStreamSpecTagType,
+  K8sAPIOptions,
+  KnownLabels,
+  NotebookKind,
+} from '~/k8sTypes';
 import { usernameTranslate } from '~/utilities/notebookControllerUtils';
 import { EnvironmentFromVariable, StartNotebookData } from '~/pages/projects/types';
 import { ROOT_MOUNT_PATH } from '~/pages/projects/pvc/const';
@@ -109,6 +115,8 @@ export const assembleNotebook = (
         'opendatahub.io/username': username,
         'opendatahub.io/accelerator-name': selectedAcceleratorProfile?.metadata.name || '',
         'opendatahub.io/hardware-profile-name': selectedHardwareProfile?.metadata.name || '',
+        'notebooks.opendatahub.io/last-image-version-git-commit-selection':
+          image.imageVersion?.annotations?.['opendatahub.io/notebook-build-commit'] ?? '',
       },
       name: notebookId,
       namespace: projectName,
@@ -322,6 +330,57 @@ export const mergePatchUpdateNotebook = (
       opts,
     ),
   );
+
+export const patchNotebookImage = (
+  existingNotebook: NotebookKind,
+  imageStream: ImageStreamKind,
+  imageVersion: ImageStreamSpecTagType,
+  opts?: K8sAPIOptions,
+): Promise<NotebookKind> => {
+  const imageUrl = `${imageStream.status?.dockerImageRepository ?? ''}:${imageVersion.name}`;
+  const imageSelection = `${imageStream.metadata.name}:${imageVersion.name}`;
+  const patches: Patch[] = [
+    {
+      op: 'replace',
+      path: '/metadata/annotations/notebooks.opendatahub.io~1last-image-selection',
+      value: imageSelection,
+    },
+    {
+      op: 'replace',
+      path: '/spec/template/spec/containers/0/image',
+      value: imageUrl,
+    },
+    {
+      op: 'replace',
+      path: '/spec/template/spec/containers/0/env/1/value',
+      value: imageUrl,
+    },
+  ];
+
+  patches.push({
+    op: existingNotebook.metadata.annotations?.[
+      'notebooks.opendatahub.io/last-image-version-git-commit-selection'
+    ]
+      ? 'replace'
+      : 'add',
+    path: '/metadata/annotations/notebooks.opendatahub.io~1last-image-version-git-commit-selection',
+    value: imageVersion.annotations?.['opendatahub.io/notebook-build-commit'],
+  });
+
+  return k8sPatchResource<NotebookKind>(
+    applyK8sAPIOptions(
+      {
+        model: NotebookModel,
+        queryOptions: {
+          name: existingNotebook.metadata.name,
+          ns: existingNotebook.metadata.namespace,
+        },
+        patches,
+      },
+      opts,
+    ),
+  );
+};
 
 export const deleteNotebook = (notebookName: string, namespace: string): Promise<K8sStatus> =>
   k8sDeleteResource<NotebookKind, K8sStatus>({
--- a/frontend/src/pages/projects/notebook/NotebookImagePackageDetails.tsx
+++ b/frontend/src/pages/projects/notebook/NotebookImagePackageDetails.tsx
@@ -11,12 +11,12 @@ import { ImageVersionDependencyType } from '~/pages/projects/screens/spawner/typ
 
 type NotebookPackageDetailsProps = {
   dependencies: ImageVersionDependencyType[];
-  title?: string;
+  title?: React.ReactNode;
 };
 
 const NotebookImagePackageDetails: React.FC<NotebookPackageDetailsProps> = ({
   dependencies,
-  title,
+  title = <DescriptionListTerm>Packages</DescriptionListTerm>,
 }) => {
   if (dependencies.length === 0) {
     return null;
@@ -25,7 +25,7 @@ const NotebookImagePackageDetails: React.FC<NotebookPackageDetailsProps> = ({
   return (
     <DescriptionList>
       <DescriptionListGroup>
-        <DescriptionListTerm>{title || 'Packages'}</DescriptionListTerm>
+        {title}
         <DescriptionListDescription>
           {dependencies.map(getNameVersionString).map((pkg) => (
             <div key={pkg}>
--- a/frontend/src/pages/projects/notebook/NotebookImagePackageDetails.tsx
+++ b/frontend/src/pages/projects/notebook/NotebookImagePackageDetails.tsx
@@ -16,7 +16,7 @@ type NotebookPackageDetailsProps = {
 
 const NotebookImagePackageDetails: React.FC<NotebookPackageDetailsProps> = ({
   dependencies,
-  title = <DescriptionListTerm>Packages</DescriptionListTerm>,
+  title,
 }) => {
   if (dependencies.length === 0) {
     return null;
@@ -25,7 +25,7 @@ const NotebookImagePackageDetails: React.FC<NotebookPackageDetailsProps> = ({
   return (
     <DescriptionList>
       <DescriptionListGroup>
-        {title}
+        <DescriptionListTerm>{title || 'Packages'}</DescriptionListTerm>
         <DescriptionListDescription>
           {dependencies.map(getNameVersionString).map((pkg) => (
             <div key={pkg}>
--- a/frontend/src/pages/projects/screens/detail/notebooks/NotebookImageDisplayName.tsx
+++ b/frontend/src/pages/projects/screens/detail/notebooks/NotebookImageDisplayName.tsx
@@ -11,12 +11,18 @@ import {
   ContentVariants,
   HelperText,
   HelperTextItem,
+  Button,
 } from '@patternfly/react-core';
 import React from 'react';
-import { ExclamationCircleIcon, InfoCircleIcon } from '@patternfly/react-icons';
+import {
+  CheckCircleIcon,
+  ExclamationCircleIcon,
+  ExclamationTriangleIcon,
+  InfoCircleIcon,
+} from '@patternfly/react-icons';
 import { ProjectObjectType, typedObjectImage } from '~/concepts/design/utils';
 import { SupportedArea, useIsAreaAvailable } from '~/concepts/areas';
-import { NotebookImageAvailability } from './const';
+import { NotebookImageAvailability, NotebookImageStatus } from './const';
 import { NotebookImage } from './types';
 
 type NotebookImageDisplayNameProps = {
@@ -25,6 +31,7 @@ type NotebookImageDisplayNameProps = {
   loaded: boolean;
   loadError?: Error;
   isExpanded?: boolean;
+  onUpdateImageClick: () => void;
 };
 
 export const NotebookImageDisplayName = ({
@@ -33,8 +40,11 @@ export const NotebookImageDisplayName = ({
   loaded,
   loadError,
   isExpanded,
+  onUpdateImageClick,
 }: NotebookImageDisplayNameProps): React.JSX.Element => {
   const isProjectScopedAvailable = useIsAreaAvailable(SupportedArea.DS_PROJECT_SCOPED).status;
+  const [isPopoverVisible, setIsPopoverVisible] = React.useState(false);
+
   // if there was an error loading the image, display unknown WITHOUT a label
   if (loadError) {
     return (
@@ -55,21 +65,9 @@ export const NotebookImageDisplayName = ({
         title: string;
         body: React.ReactNode;
         variant: AlertProps['variant'];
+        footer?: React.ReactNode;
       } => {
-    if (notebookImage.imageAvailability === NotebookImageAvailability.DISABLED) {
-      return {
-        title: 'Notebook image disabled',
-        body: (
-          <p>
-            The <b>{notebookImage.imageDisplayName}</b> notebook image has been disabled. This
-            workbench can continue using the image, but disabled images are not available for
-            selection when creating new workbenches.
-          </p>
-        ),
-        variant: 'info',
-      };
-    }
-    if (notebookImage.imageAvailability === NotebookImageAvailability.DELETED) {
+    if (notebookImage.imageStatus === NotebookImageStatus.DELETED) {
       const unknownBody = (
         <p>
           An <b>unknown</b> notebook image has been deleted. To run this workbench, select a new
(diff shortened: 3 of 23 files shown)


xianli123:
Thanks @rsun19 . The screenshots look good.

frontend/src/concepts/notebooks/getImageStatus.ts
```diff
@@ -0,0 +1,27 @@
...
+): ImageStreamStatus | NotebookImageAvailability | undefined => {
+  if (imageData && imageData.imageAvailability === NotebookImageAvailability.DELETED) {
+    return NotebookImageAvailability.DELETED;
+  }
+  if (imageData && imageData.imageAvailability === NotebookImageAvailability.DISABLED) {
+    return NotebookImageAvailability.DISABLED;
+  }
+  if (
+    imageData &&
+    imageData.imageVersion.annotations?.['opendatahub.io/image-tag-outdated'] === 'true'
```
Gkrumbach07, line 16: I talked with the IDE team about this yesterday and we need to change a few things

What is correct
- latest check
- outdated
- disabled

What needs updating maybe from `getNotebookImageData`
- deleted

Deleted state right is for when we cant find the imagestream _tag_ by comparing the tag names. What we need to do in addition to this is to compare the git hashes of the imagestream tags. The reasoning for this, on upgrade of our product, these image stream tags could be updated for things like security reasons, but they keep the same tag name. So what is changing is the git commit hash. So a workbenches image stream version is also deleted if the version it last used was upgraded to a new git commit hash

So to handle this we need to do the following
1. start tracking the git commit hash like we do with the last image selected
```tsx
  'notebooks.opendatahub.io/last-image-selection': string; // the last image they selected
  'notebooks.opendatahub.io/last-image-version-git-commit-selection': string; // the git commit hash of the image version they selected
```
2. Then in `getNotebookImageData`, we need to say a image version is deleted by
   1. from notebook resource
   2. get last imagestream
   3. if found get image version
   (we are already doing this)
   4. if image version found, we compare the tags got commit hash with our saved `notebooks.opendatahub.io/last-image-version-git-commit-selection`
   5. If they differ, then it is also deleted

frontend/src/api/k8s/notebooks.ts
```diff
@@ -29,6 +35,97 @@ import { getImageStreamDisplayName } from '~/pages/projects/screens/spawner/spaw
 import { k8sMergePatchResource } from '~/api/k8sUtils';
 import { getshmVolume, getshmVolumeMount } from '~/api/k8s/utils';
 
+export const assembleNotebookImage = (
```
Gkrumbach07, line 38: This is very similer to the assembleNotebook function. what is the reason for this one

frontend/src/concepts/notebooks/getImageStatus.ts
```diff
@@ -0,0 +1,27 @@
+import { NotebookImageAvailability } from '~/pages/projects/screens/detail/notebooks/const';
+import { NotebookImageData } from '~/pages/projects/screens/detail/notebooks/types';
+import { ImageStreamStatus } from './types';
+
+export const getImageStatus = (
```
Gkrumbach07, line 5: Can we instead just upgrade `NotebookImageAvailability` to include the updated state instead of having both NotebookImageAvailability and ImageStreamStatus

Gkrumbach07: id like to merge this function into getNotebookImageData becuase it is only ever used right next to it so makes sense to combine

frontend/src/concepts/notebooks/types.ts
```diff
@@ -0,0 +1,4 @@
+export enum ImageStreamStatus {
```
Gkrumbach07, line 1: should be merged into notebookavailability enum

frontend/src/pages/projects/screens/detail/notebooks/NotebookList.tsx
```diff
@@ -27,7 +27,7 @@ const NotebookList: React.FC = () => {
   const projectName = currentProject.metadata.name;
   const isNotebooksEmpty = notebooks.length === 0;
 
-  useRefreshInterval(FAST_POLL_INTERVAL, () =>
+  useRefreshInterval(FAST_NOTEBOOK_START_STOP_POLL_INTERVAL, () =>
```
Gkrumbach07, line 30: dont create a new poll interval, just reuse FAST_POLL_INTERVAL

Gkrumbach07: fast is the fasted we want to _poll_, if we need something faster we should instead look to websockets. In this case we should just stick to the fast poll interval

frontend/src/pages/projects/screens/detail/notebooks/getNotebookImageVersionData.ts
```diff
@@ -0,0 +1,132 @@
+import { ImageStreamKind, ImageStreamSpecTagType, NotebookKind } from '~/k8sTypes';
+import { PodContainer } from '~/types';
+
+export const getCurrentAndLatestNotebookImageVersionData = (
```
Gkrumbach07, line 4: this duplicates a lot of functions already done in getNotebookImageData.  we should look to reuse that here

frontend/src/pages/projects/screens/detail/notebooks/NotebookUpdateImageCard.tsx
```diff
@@ -0,0 +1,88 @@
...
+  imageSoftwareVersions: string | undefined,
+): React.ReactNode => {
+  const imageNameSoftwareVersionJSON = getImageNameSoftwareVersionJSON(imageSoftwareVersions);
+  return (
+    <>
+      {imageNameSoftwareVersionJSON.map((value, index) => (
+        <p key={index}>{`${value.name} ${value.version}`}</p>
+      ))}
+    </>
+  );
```
Gkrumbach07, line 43: these should be in a util file

frontend/src/pages/projects/screens/detail/notebooks/NotebookTableRow.tsx
```diff
@@ -62,7 +65,11 @@ const NotebookTableRow: React.FC<NotebookTableRowProps> = ({
       requests: {},
     },
   };
+  const { dashboardNamespace } = useNamespaces();
+  const [images, imagesLoaded, imagesLoadError] = useImageStreams(dashboardNamespace, true);
```
Gkrumbach07, line 69: for each notebook we are fetching all the imagestreams, this is a lot of extra calls. we should try to not fetch all theses images until we enter the modal to reduce the number of calls

Gkrumbach07: or we should upgrade useNotebookImage becuase that is already fetching all images so look at that route as well

Gkrumbach07 (changes requested):

rsun19:
Made requested changes

frontend/src/api/k8s/notebooks.ts
```diff
@@ -30,7 +34,7 @@ import { k8sMergePatchResource } from '~/api/k8sUtils';
 import { getshmVolume, getshmVolumeMount } from '~/api/k8s/utils';
 
 export const assembleNotebook = (
-  data: StartNotebookData,
+  data: UpdateNotebookData,
```
Gkrumbach07, line 37: why did the type need to change for this? It looks like the type are the same expect for this type `NotebookUpdateImageSpecOptions ` which im not sure is needed?

Gkrumbach07 (changes requested):

frontend/src/pages/projects/screens/detail/notebooks/types.ts
```diff
@@ -8,6 +8,10 @@ export type NotebookImage =
       imageDisplayName: string;
       tagSoftware: string;
       dependencies: ImageVersionDependencyType[];
+      imageStream: ImageStreamKind;
+      imageVersion: ImageStreamSpecTagType;
+      imageStatus?: NotebookImageAvailability;
```
Gkrumbach07, line 13: you have two keys now here. one for image status and one for image availability. both are of type `NotebookImageAvailability`. this should just be one field

frontend/src/api/k8s/notebooks.ts
rsun19: The main motivation for including it is because there is no natural way to retrieve information about complete hardware or accelerator profiles directly from the yaml files, only the name. Thus, I am relaxing the types because the patch will treat this omission as an actual omission if not included.

rsun19: The `assembleNotebook` function only uses the name regardless, nothing else

frontend/src/pages/projects/screens/detail/notebooks/types.ts
rsun19: I added both because the `ENABLED` field is still used. If this isn't important, I can just remove this `Enabled` field and only use image availability

frontend/src/api/k8s/notebooks.ts
rsun19: I'm thinking there are 3 solutions to this:
1. Relax the types
2. Only accept ap/hwp names in `assemble notebook` instead of the whole thing
3. Retrieve the entire ap/hwp and pass it into the functions without relaxing the type

frontend/src/pages/projects/screens/detail/notebooks/NotebookUpdateImageCard.tsx
```diff
@@ -0,0 +1,70 @@
...
+};
```
Gkrumbach07, line 26: we have a component that does this already 

NotebookImagePackageDetails

frontend/src/api/k8s/notebooks.ts
Gkrumbach07: id say go with option 2.

rsun19:
Made changes

frontend/src/pages/projects/screens/detail/notebooks/NotebookUpdateImageModalFooter.tsx
```diff
@@ -0,0 +1,163 @@
...
+      dataConnectionType: undefined,
+      dataConnectionEnabled: !!notebook.spec.template.spec.containers.at(0)?.envFrom,
+      dataConnectionCategory: undefined,
+      outcome: TrackingOutcome.submit,
+      success: true,
+    };
+    fireFormTrackingEvent('Workbench image updated', tep);
+    notebooks.find((x) => x.notebook.metadata.name === notebook.metadata.name)?.refresh();
+    onModalClose();
+  };
```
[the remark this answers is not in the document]
rsun19: For instance I put `undefined` as some of the properties too; am unsure of where the data could be pulled from

frontend/src/pages/projects/screens/detail/notebooks/NotebookUpdateImageModalFooter.tsx
```diff
@@ -0,0 +1,163 @@
...
+      imageName: imageStream.metadata.name,
+      projectName: notebook.metadata.namespace,
+      notebookName: notebook.metadata.name,
+      dataConnectionType: undefined,
+      dataConnectionEnabled: !!notebook.spec.template.spec.containers.at(0)?.envFrom,
+      dataConnectionCategory: undefined,
+      outcome: TrackingOutcome.submit,
+      success: true,
+    };
+    fireFormTrackingEvent('Workbench image updated', tep);
```
rsun19, line 68: @Gkrumbach07 and is this the correct form tracking event name?

frontend/src/pages/projects/screens/detail/notebooks/NotebookUpdateImageModalFooter.tsx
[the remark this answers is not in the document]
Gkrumbach07: this is a tracking even so it doesnt need to include all of those other params if they were not changed. i would expect stuff about the image and the notebook identifying details

so we dont need any of those data connection ones, you can just omit them

frontend/src/pages/projects/screens/detail/notebooks/NotebookUpdateImageModalFooter.tsx
Gkrumbach07: Yep that works great

frontend/src/api/k8s/notebooks.ts
```diff
@@ -323,6 +331,65 @@ export const mergePatchUpdateNotebook = (
...
+      path: '/metadata/annotations/notebooks.opendatahub.io~1last-image-version-git-commit-selection',
+      value: imageVersion.annotations?.['opendatahub.io/notebook-build-commit'],
+    });
+  } else {
+    patches.push({
+      op: 'add',
+      path: '/metadata/annotations/notebooks.opendatahub.io~1last-image-version-git-commit-selection',
+      value: imageVersion.annotations?.['opendatahub.io/notebook-build-commit'],
+    });
+  }
```
Gkrumbach07, line 376:
```suggestion
    patches.push({
      op: existingNotebook.metadata.annotations?.['notebooks.opendatahub.io/last-image-version-git-commit-selection'] ? 'replace' : 'add',
      path: '/metadata/annotations/notebooks.opendatahub.io~1last-image-version-git-commit-selection',
      value: imageVersion.annotations?.['opendatahub.io/notebook-build-commit'],
  });
```

Gkrumbach07: you can simplify the logic

frontend/src/pages/projects/notebook/NotebookImagePackageDetails.tsx
```diff
@@ -25,7 +27,7 @@ const NotebookImagePackageDetails: React.FC<NotebookPackageDetailsProps> = ({
   return (
     <DescriptionList>
       <DescriptionListGroup>
-        <DescriptionListTerm>{title || 'Packages'}</DescriptionListTerm>
```
Gkrumbach07, line 28: you are just adding title back when using this component, why do you need includeTitle

I am guessing becuase the styling is different. Instead of having that `includeTitle` prop, open up the title prop to accept more than just string

`title?: string;` -> `title?: React.ReactNode;`

Then you can pass in your title component that has the correct heading size

frontend/src/pages/projects/screens/detail/notebooks/NotebookImageDisplayName.tsx
```diff
@@ -11,25 +11,45 @@ import {
...
   notebookImage: NotebookImage | null;
+  notebook: NotebookKind;
   loaded: boolean;
   loadError?: Error;
   isExpanded?: boolean;
 };
 
 export const NotebookImageDisplayName = ({
   notebookImage,
+  notebook,
```
Gkrumbach07, line 40: I have an idea. I want to try and make this component specific to itself... meaning we have notebook related requiremtns for this when it is notebook _image_ component. What we can do pull out the modal related stuff 
```tsx
setIsPopoverVisible(false);
              if (notebookImage.latestImageVersion) {
                setIsModalOpen(true);
              } else {
                navigate(
                  `/projects/${currentProject.metadata.name}/spawner/${notebook.metadata.name}`,
                );
              }
            }}
```

And instead have that on that in the notebookRow. then in this component we just expose a function called something like `onUpdateImageClick` which you then would put that code you have for handling the modal there

so props would now be 

```tsx
type NotebookImageDisplayNameProps = {
  notebookImage: NotebookImage | null;
  onUpdateImageClick?: () => void;
  loaded: boolean;
  loadError?: Error;
  isExpanded?: boolean;
};
```
you can even make it optional and only render that button when that function is defined... then other people can use this `NotebookImageDisplayName ` without having to pass in a notebook and care about the update functionality

frontend/src/pages/projects/screens/detail/notebooks/NotebookImageDisplayName.tsx
```diff
@@ -143,26 +194,52 @@ export const NotebookImageDisplayName = ({
...
       )}
+      {isModalOpen &&
+        notebookImage.imageStatus !== NotebookImageStatus.DELETED &&
+        notebookImage.latestImageVersion && (
+          <NotebookUpdateImageModal
+            notebookImageVersionData={{
+              currentImageVersion: notebookImage.imageVersion,
+              latestImageVersion: notebookImage.latestImageVersion,
+              imageStream: notebookImage.imageStream,
+            }}
```
Gkrumbach07, line 238: why break up notebookImage, just pass in the whole notebook image as `NotebookImage `

frontend/src/pages/projects/screens/detail/notebooks/utils.ts
```diff
@@ -21,3 +21,21 @@ export const extractAcceleratorResources = (
...
+};
+
+export const getNameImageSoftwareVersion = (
+  imageSoftwareVersion: string | undefined,
+): string | undefined => {
+  const imageNameSoftwareVersionJSON = getImageNameSoftwareVersionJSON(imageSoftwareVersion);
+  return `${imageNameSoftwareVersionJSON.at(0)?.name ?? ''} ${
+    imageNameSoftwareVersionJSON.at(0)?.version ?? ''
+  }`;
+};
```
Gkrumbach07, line 41: there are already functions that exist for this

getImageVersionSoftwareString
and 
getImageVersionDependencies

frontend/src/pages/projects/screens/detail/notebooks/NotebookUpdateImageCard.tsx
```diff
@@ -0,0 +1,61 @@
...
+        ? `(${imageVersion.annotations['opendatahub.io/notebook-build-commit']})`
+        : ''}
+    </CardHeader>
+    <CardBody>
+      <Stack hasGutter>
+        <StackItem>
+          <Title headingLevel="h4">Software</Title>
+          {getNameImageSoftwareVersion(
+            imageVersion.annotations?.['opendatahub.io/notebook-software'],
+          )}
```
Gkrumbach07, line 47: use getImageVersionSoftwareString

frontend/src/pages/projects/screens/detail/notebooks/NotebookUpdateImageCard.tsx
```diff
@@ -0,0 +1,61 @@
...
+        </StackItem>
+        <StackItem>
+          <Title headingLevel="h4">Packages</Title>
+          <NotebookImagePackageDetails
+            dependencies={getImageNameSoftwareVersionJSON(
+              imageVersion.annotations?.['opendatahub.io/notebook-python-dependencies'],
+            )}
```
Gkrumbach07, line 54: use getImageVersionDependencies

frontend/src/pages/projects/screens/detail/notebooks/useNotebookImageData.ts
```diff
@@ -196,8 +250,30 @@ export const getImageAvailability = (imageStream: ImageStreamKind): NotebookImag
...
 ): NotebookImageData[0] => ({
-  imageAvailability: NotebookImageAvailability.DELETED,
+  imageStatus: NotebookImageStatus.DELETED,
   imageDisplayName,
 });
 
+const getImageStatus = (
+  imageVersion: ImageStreamSpecTagType,
+  imageStatus: NotebookImageAvailability | NotebookImageStatus,
+):
```
Gkrumbach07, line 260: get image status should just return the status. right now it is returning both status or availability

frontend/src/pages/projects/screens/detail/notebooks/NotebookUpdateImageModalFooter.tsx
```diff
@@ -0,0 +1,163 @@
...
+  const handleStart = () => {
+    setError(undefined);
+    setCreateInProgress(true);
+  };
+
+  const updateNotebookImagePromise = async (dryRun: boolean) => {
+    if (dryRun) {
+      return patchNotebookImage(notebook, imageStream, latestImageVersion, { dryRun });
+    }
+    return patchNotebookImage(notebook, imageStream, latestImageVersion);
```
Gkrumbach07, line 92: you dont need to dry run when there is only one api call

frontend/src/pages/projects/screens/detail/notebooks/NotebookImageDisplayName.tsx
rsun19: I broke it up because it avoids a lot of type-checking later down the line since `latestImageVersion` can be undefined, and `currentImageVersion` can only be there if the image is not deleted.

rsun19: I'll refactor to pass in `NotebookImage` for now, it's just going to have a lot of checking

rsun19:
Made changes

frontend/src/pages/projects/screens/detail/notebooks/NotebookTableRow.tsx
```diff
@@ -248,6 +266,16 @@ const NotebookTableRow: React.FC<NotebookTableRowProps> = ({
           }}
         />
       ) : null}
+      {isModalOpen &&
+        notebookImage &&
+        notebookImage.imageStatus !== NotebookImageStatus.DELETED &&
+        notebookImage.latestImageVersion && (
```
DaoDaoNoCode, line 272: I just feel the requirements are a little bit duplicated here. The `isModalOpen` will only be set to `true` when these requirements are met above in `onUpdateImageClick` above. Thus, these checks are not necessary here IMO.

rsun19: Is this a better fix? I don't think I can remove any more checks after this without the compiler yelling at me

```
{isModalOpen && notebookImage && (
        <NotebookUpdateImageModal
          notebookImage={notebookImage}
          notebook={obj.notebook}
          onModalClose={onModalClose}
        />
      )}
```

rsun19: since the notebookImage can be null in the hook

DaoDaoNoCode: Yeah look good, or you can change the type of `notebookImage` in `NotebookUpdateImageModal` to accept null, and early return `null` there if the image doesn't exist. It all depends on you.

frontend/src/pages/projects/notebook/NotebookImagePackageDetails.tsx
```diff
@@ -11,12 +11,12 @@ import { ImageVersionDependencyType } from '~/pages/projects/screens/spawner/typ
...
 type NotebookPackageDetailsProps = {
   dependencies: ImageVersionDependencyType[];
-  title?: string;
+  title?: React.ReactNode;
 };
 
 const NotebookImagePackageDetails: React.FC<NotebookPackageDetailsProps> = ({
   dependencies,
-  title,
+  title = <DescriptionListTerm>Packages</DescriptionListTerm>,
```
Gkrumbach07, line 19: You want to make title still default to just `Packages` and then keep this usage
```tsx
        <DescriptionListTerm>{title}</DescriptionListTerm>
```

reason is that if you put a custom title in that is just a string, you loose any formatting.

frontend/src/pages/projects/screens/detail/notebooks/NotebookImageDisplayName.tsx
```diff
@@ -163,26 +178,61 @@ export const NotebookImageDisplayName = ({
...
+              shouldClose={() => setIsPopoverVisible(false)}
+              isVisible={isPopoverVisible}
             >
-              {notebookImage.imageAvailability}
-            </Label>
-          </Popover>
-        </FlexItem>
+              <Label
+                // eslint-disable-next-line @typescript-eslint/no-empty-function
+                onClick={() => {}}
```
Gkrumbach07, line 221:
```suggestion
                onClick={(e) => e.preventDefault()}
```

frontend/src/pages/projects/screens/detail/notebooks/useNotebookImageData.ts
```diff
@@ -185,8 +221,19 @@ const getNotebookImageNoInternalRegistryNoSHA = (
...
+    (image) =>
+      image.spec.tags &&
+      image.spec.tags.some(
+        (imageTags) =>
+          imageTags.annotations?.['opendatahub.io/notebook-build-commit'] ===
+          notebook.metadata.annotations?.[
+            'notebooks.opendatahub.io/last-image-version-git-commit-selection'
+          ],
+      ),
+  );
```
Gkrumbach07, line 234: this is duplicated, maybe put it into a helper function

Gkrumbach07 (changes requested):
just a few small things

rsun19:
addressed comments

frontend/src/pages/projects/screens/detail/notebooks/NotebookImageDisplayName.tsx
Gkrumbach07: you are still using  onClick={() => {}}

Gkrumbach07 (approved):
/lgtm

PR #3978, opened 2025-03-26
Cypress rhoaieng 20572

antowaddle:
## Request review criteria:

If you have UI changes: 


frontend/src/__tests__/cypress/cypress/tests/e2e/settings/hardwareProfiles/testModelServingTolerations.cy.ts
```diff
@@ -0,0 +1,158 @@
...
+        deleteOpenShiftProject(projectName, { timeout: 300000 });
+      }
+    });
+  });
+
+  it(
+    'Verify Model Serving Creation using Hardware Profiles and applying Tolerations',
+    // TODO: Add the below tags once this feature is enabled in 2.20+
+    //  { tags: ['@Sanity', '@SanitySet2', '@Dashboard'] },
+    { tags: ['@Featureflagged', '@HardwareProfileModelServing', '@HardwareProfiles'] },
```
manosnoam, line 104: @antowaddle maybe we should add a common flag `Dev` to indicate test or feature is still in development.
When moved from development to production, remove that `Dev` tag.
In addition, we'll add `--env skipTags=@Dev` when running on CI, so it will skip it like with Bug tags.

manosnoam (commented):
/lgtm
Minor comment

frontend/src/__tests__/cypress/cypress/tests/e2e/settings/hardwareProfiles/testModelServingTolerations.cy.ts
antowaddle: @manosnoam The Tag I've currently used to indicate that a feature is behind a feature flag is '@Featureflagged', this has been applied also to the other Hardware Profile tests. I believe this tag is descriptive enough for folks to know the test is behind a feature flag, what do you think?

manosnoam: yeah, it makes sense. I just wanted a more common tag like "Bug" (as we don't use "AutomationBug" or "ProductBug"). I would think a "Dev" tag could cover these development stages.

manosnoam (approved):

PR #3984, opened 2025-03-27
RHOAIENG-22380 Track Register Model in Catalog

pilhuhn:
https://issues.redhat.com/browse/RHOAIENG-22380

## Description
Add analytics tracking for 'Register Catalog Model' and 'Start LAB tune' actions

## How Has This Been Tested?
Manual clicking through with watching the console log

## Test Impact
We decided that analytics need no automated testing.

## Request review criteria:

### DIFF
--- a/frontend/src/pages/pipelines/global/modelCustomization/startRunModal/StartRunModal.tsx
+++ b/frontend/src/pages/pipelines/global/modelCustomization/startRunModal/StartRunModal.tsx
@@ -1,6 +1,9 @@
 import * as React from 'react';
 import { Modal } from '@patternfly/react-core/deprecated';
 import {
+  Alert,
+  Content,
+  ContentVariants,
   Form,
   FormGroup,
   FormHelperText,
@@ -11,9 +14,6 @@ import {
   ModalVariant,
   Stack,
   StackItem,
-  Alert,
-  Content,
-  ContentVariants,
 } from '@patternfly/react-core';
 import { Link } from 'react-router-dom';
 import ProjectSelector from '~/concepts/projects/ProjectSelector';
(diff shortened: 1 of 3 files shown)


pilhuhn:
@andrewballantyne

andrewballantyne:
/assign @mturley 

Please do a review here -- let me know if you have any questions. We'll all have to get ramped up on this ecosystem of tracking events _soonish_. cc @dgutride

andrewballantyne:
> @andrewballantyne for my review should I just be making sure this doesn't break any existing logic? If I should test the tracking events, how do I do that / is there documentation?

The goal is to best understand if we are getting the right events in the right spot. We do have [a confluence page on segment tracking](https://spaces.redhat.com/display/RHODS/Add+Segment+tracking+in+the+UI) -- let me know if you're still not sure what to do for a review.

frontend/src/pages/pipelines/global/modelCustomization/startRunModal/StartRunModal.tsx
```diff
@@ -47,10 +50,17 @@ const StartRunModal: React.FC<StartRunModalProps> = ({
         <DashboardModalFooter
           onSubmit={() => {
             if (selectedProject) {
+              fireFormTrackingEvent(eventName, { outcome: TrackingOutcome.submit });
               onSubmit(selectedProject);
```
mturley, line 54: This modal submission does not start the Lab Tune. It prompts for a project to use, validates that project, and then submitting it takes the user to the model customization form where they still need to fill out more details before starting the actual tune (`frontend/src/pages/pipelines/global/modelCustomization/ModelCustomizationForm.tsx`). The cancel event here still probably makes sense, but we probably want to exclude the submit event here or perhaps split this into two events and have this one be something like "Redirected to Lab Tune form from registry".

mturley (changes requested):
Changes to RegisterCatalogModel LGTM, but the StartRunModal events may not make sense.

frontend/src/pages/pipelines/global/modelCustomization/startRunModal/StartRunModal.tsx
pilhuhn: Yes, you are right, that should be a better event name.

mturley (approved):
LGTM!

PR #3985, opened 2025-03-27
Adding associated events to workbench status modal

katieperry4:
Closes: [RHOAIENG-21741](https://issues.redhat.com/browse/RHOAIENG-21741) and [RHOAIENG-22056](https://issues.redhat.com/browse/RHOAIENG-22056)

## Description
-Removed code that assumed notebook modal events had succeeded based on the most recently successful event.
-Added an additional check for ‘ose-oauth-proxy’ for clarity in differentiating between notebook and oauth events.
-Created a new type that associated events with event milestones -> If the container started, then the image events must have succeeded and so on.
-Added a snippet that loops through the associated events and marks off events based on milestone successes.
-Added optional events and event filtering
-Fixed a typo.

## How Has This Been Tested?
-Tested with npm run test
-Tested locally with successful and broken workbenches.

-If you want to test, open a data science project
	-Navigate to the workbenches page and start a workbench.
	-Click the button in the status column to view the modal

Successful:
Screenshot 2025-03-27 at 9 09 40 AM

Failed:
Screenshot 2025-03-28 at 1 23 40 PM

## Test Impact
Modified unit tests to handle the changes made in the number of events and error text

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/concepts/__tests__/StartNotebookModal.spec.tsx
+++ b/frontend/src/concepts/__tests__/StartNotebookModal.spec.tsx
@@ -76,10 +76,7 @@ describe('Start Notebook modal', () => {
     expect(stepper).toBeTruthy();
     const steps = screen.getAllByRole('listitem');
     expect(steps).toHaveLength(14);
-    /* TODO: This finds the POD_PROBLEM: 'There was a problem with the pod'
-      once text and location of error is finalized with UX set the error location index and error text
-    */
-    //expect(steps[3]).toHaveTextContent('Failed to scale-up');
+    expect(steps[1]).toHaveTextContent('There was a problem with the pod');
   });
 
   it('should show in progress notebook startup status', async () => {
--- a/frontend/src/concepts/__tests__/StartNotebookModal.spec.tsx
+++ b/frontend/src/concepts/__tests__/StartNotebookModal.spec.tsx
@@ -77,7 +77,7 @@ describe('Start Notebook modal', () => {
     const steps = screen.getAllByRole('listitem');
     expect(steps).toHaveLength(14);
     /* TODO: This finds the POD_PROBLEM: 'There was a problem with the pod'
-      once text and location of error is finalized with UX set the error location and error text
+      once text and location of error is finalized with UX set the error location index and error text
     */
     //expect(steps[3]).toHaveTextContent('Failed to scale-up');
   });
--- a/frontend/src/concepts/__tests__/StartNotebookModal.spec.tsx
+++ b/frontend/src/concepts/__tests__/StartNotebookModal.spec.tsx
@@ -34,21 +34,20 @@ describe('Start Notebook modal', () => {
     const stepper = screen.getByTestId('notebook-startup-steps');
     expect(stepper).toBeTruthy();
     const steps = screen.getAllByRole('listitem');
-    expect(steps).toHaveLength(14);
+    expect(steps).toHaveLength(13);
     expect(steps[0]).toHaveTextContent('Server requested');
     expect(steps[1]).toHaveTextContent('Pod created');
     expect(steps[2]).toHaveTextContent('Pod assigned');
-    expect(steps[3]).toHaveTextContent('PVC attached');
-    expect(steps[4]).toHaveTextContent('Interface added');
-    expect(steps[5]).toHaveTextContent('Pulling workbench image');
-    expect(steps[6]).toHaveTextContent('Workbench image pulled');
-    expect(steps[7]).toHaveTextContent('Workbench container created');
-    expect(steps[8]).toHaveTextContent('Workbench container started');
-    expect(steps[9]).toHaveTextContent('Pulling oauth proxy');
-    expect(steps[10]).toHaveTextContent('Oauth proxy pulled');
-    expect(steps[11]).toHaveTextContent('Oauth proxy container created');
-    expect(steps[12]).toHaveTextContent('Oauth proxy container started');
-    expect(steps[13]).toHaveTextContent('Server started');
+    expect(steps[3]).toHaveTextContent('Interface added');
+    expect(steps[4]).toHaveTextContent('Pulling workbench image');
+    expect(steps[5]).toHaveTextContent('Workbench image pulled');
+    expect(steps[6]).toHaveTextContent('Workbench container created');
+    expect(steps[7]).toHaveTextContent('Workbench container started');
+    expect(steps[8]).toHaveTextContent('Pulling oauth proxy');
+    expect(steps[9]).toHaveTextContent('Oauth proxy pulled');
+    expect(steps[10]).toHaveTextContent('Oauth proxy container created');
+    expect(steps[11]).toHaveTextContent('Oauth proxy container started');
+    expect(steps[12]).toHaveTextContent('Server started');
   });
 
   it('should show failed notebook startup status', async () => {
@@ -76,8 +75,11 @@ describe('Start Notebook modal', () => {
     const stepper = screen.getByTestId('notebook-startup-steps');
     expect(stepper).toBeTruthy();
     const steps = screen.getAllByRole('listitem');
-    expect(steps).toHaveLength(15);
-    expect(steps[1]).toHaveTextContent('Failed to scale-up');
+    expect(steps).toHaveLength(14);
+    /* TODO: This finds the POD_PROBLEM: 'There was a problem with the pod'
+      once text and location of error is finalized with UX set the error location and error text
+    */
+    //expect(steps[3]).toHaveTextContent('Failed to scale-up');
   });
 
   it('should show in progress notebook startup status', async () => {
@@ -181,7 +183,7 @@ describe('Start Notebook modal', () => {
     const stepper = screen.getByTestId('notebook-startup-steps');
     expect(stepper).toBeTruthy();
     const steps = screen.getAllByRole('listitem');
-    expect(steps).toHaveLength(14);
-    expect(screen.getAllByTestId('step-status-Pending')).toHaveLength(14);
+    expect(steps).toHaveLength(13);
+    expect(screen.getAllByTestId('step-status-Pending')).toHaveLength(13);
   });
 });
--- a/frontend/src/types.ts
+++ b/frontend/src/types.ts
@@ -637,23 +637,18 @@ export const ProgressionStepTitles = {
 
 export const AssociatedSteps = {
   [ProgressionStep.NOTEBOOK_CONTAINER_STARTED]: [
-    ProgressionStep.SERVER_REQUESTED,
-    ProgressionStep.POD_CREATED,
-    ProgressionStep.POD_ASSIGNED,
     ProgressionStep.PVC_ATTACHED,
     ProgressionStep.INTERFACE_ADDED,
     ProgressionStep.PULLING_NOTEBOOK_IMAGE,
     ProgressionStep.NOTEBOOK_IMAGE_PULLED,
-    ProgressionStep.NOTEBOOK_CONTAINER_CREATED,
   ],
   [ProgressionStep.OAUTH_CONTAINER_STARTED]: [
     ProgressionStep.PULLING_OAUTH,
     ProgressionStep.OAUTH_PULLED,
     ProgressionStep.OAUTH_CONTAINER_CREATED,
-    ProgressionStep.OAUTH_CONTAINER_STARTED,
   ],
   [ProgressionStep.POD_ASSIGNED]: [ProgressionStep.POD_CREATED],
-  [ProgressionStep.SERVER_STARTED]: [ProgressionStep],
+  [ProgressionStep.SERVER_STARTED]: Object.values(ProgressionStep),
 };
 
 export type NotebookProgressStep = {
--- a/frontend/src/types.ts
+++ b/frontend/src/types.ts
@@ -605,39 +605,44 @@ export enum ProgressionStep {
   SERVER_REQUESTED = 'SERVER_REQUESTED',
   POD_CREATED = 'POD_CREATED',
   POD_ASSIGNED = 'POD_ASSIGNED',
+  POD_PROBLEM = 'POD_PROBLEM',
   PVC_ATTACHED = 'PVC_ATTACHED',
   INTERFACE_ADDED = 'INTERFACE_ADDED',
   PULLING_NOTEBOOK_IMAGE = 'PULLING_NOTEBOOK_IMAGE',
   NOTEBOOK_IMAGE_PULLED = 'NOTEBOOK_IMAGE_PULLED',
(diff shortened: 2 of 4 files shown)


frontend/src/utilities/notebookControllerUtils.ts
```diff
@@ -357,13 +359,8 @@ export const getNotebookEventStatus = (
...
   }
 
+  // For notebook-related events
   switch (event.reason) {
-    case 'SuccessfulCreate':
-      return {
-        step: ProgressionStep.POD_CREATED,
-        status: EventStatus.SUCCESS,
-        timestamp,
-      };
```
andrewballantyne, line 366: Whoops, bring this back 😄

frontend/src/utilities/notebookControllerUtils.ts
```diff
@@ -577,6 +562,18 @@ export const useNotebookProgress = (
     }
   }
 
+  // If milestone steps are completed, mark off associated steps
+  Object.entries(AssociatedSteps).forEach(([key, values]) => {
+    if (progressSteps.find((p) => p.step === key)?.status === EventStatus.SUCCESS) {
+      values.forEach((value) => {
+        const currentStep = progressSteps.find((p) => p.step === value);
+        if (currentStep && currentStep.status !== EventStatus.SUCCESS) {
```
andrewballantyne, line 570:
```suggestion
        if (currentStep) {
```
Setting it to success doesn't matter if it's already there or not... (however if you were after something with this line, you can jump to the point with optional chaining)
```suggestion
        if (currentStep?.status !== EventStatus.SUCCESS) {
```

frontend/src/types.ts
```diff
@@ -635,6 +635,27 @@ export const ProgressionStepTitles = {
...
+  [ProgressionStep.NOTEBOOK_CONTAINER_STARTED]: [
+    ProgressionStep.SERVER_REQUESTED,
+    ProgressionStep.POD_CREATED,
+    ProgressionStep.POD_ASSIGNED,
+    ProgressionStep.PVC_ATTACHED,
+    ProgressionStep.INTERFACE_ADDED,
+    ProgressionStep.PULLING_NOTEBOOK_IMAGE,
+    ProgressionStep.NOTEBOOK_IMAGE_PULLED,
+    ProgressionStep.NOTEBOOK_CONTAINER_CREATED,
+  ],
```
andrewballantyne, line 648:
```suggestion
  [ProgressionStep.NOTEBOOK_CONTAINER_STARTED]: [
    ProgressionStep.PVC_ATTACHED,
    ProgressionStep.INTERFACE_ADDED,
    ProgressionStep.PULLING_NOTEBOOK_IMAGE,
    ProgressionStep.NOTEBOOK_IMAGE_PULLED,
    ProgressionStep.NOTEBOOK_CONTAINER_CREATED,
  ],
```
This kind of separation is better because the "pod" and "server requested" are technically not tied to the Notebook Container (pod is the wrapper, and server requested is always available)

frontend/src/types.ts
```diff
@@ -635,6 +635,27 @@ export const ProgressionStepTitles = {
...
+  [ProgressionStep.OAUTH_CONTAINER_STARTED]: [
+    ProgressionStep.PULLING_OAUTH,
+    ProgressionStep.OAUTH_PULLED,
+    ProgressionStep.OAUTH_CONTAINER_CREATED,
+    ProgressionStep.OAUTH_CONTAINER_STARTED,
+  ],
```
andrewballantyne, line 654:
```suggestion
  [ProgressionStep.OAUTH_CONTAINER_STARTED]: [
    ProgressionStep.PULLING_OAUTH,
    ProgressionStep.OAUTH_PULLED,
    ProgressionStep.OAUTH_CONTAINER_CREATED,
  ],
```
Your 'key' doesn't need to be in the dependencies -- its the check.

frontend/src/types.ts
```diff
@@ -635,6 +635,27 @@ export const ProgressionStepTitles = {
...
+  [ProgressionStep.POD_ASSIGNED]: [ProgressionStep.POD_CREATED],
+  [ProgressionStep.SERVER_STARTED]: [ProgressionStep],
```
andrewballantyne, line 656: `[ProgressionStep]` is an array wrapping an object (the enum) -- you won't be able to match items in the `progressStep[].step` against this as they will be a string

```suggestion
  [ProgressionStep.SERVER_STARTED]: Object.values(ProgressionStep),
```
This will give you a "dynamic" list of all items -- ie "when I hit Server started, everything is success" -- it'll catch itself, but that's not really a problem, it'll already be success and there is no harm setting it to the same value.

andrewballantyne (changes requested):
Code Changes needed to get this closer to the desired solution -- I however ran into a couple awkward scenarios... I'll be documenting them on another review post in a moment

frontend/src/utilities/notebookControllerUtils.ts
```diff
@@ -310,7 +311,8 @@ export const getNotebookEventStatus = (
 ): NotebookProgressStep => {
```
andrewballantyne, line 311: Modify the `NotebookProgressStep` object -- we need to make `step` required.

In any place that used `description` instead of `step` (because it was optional) -- go ahead and use description instead -- because step will always be defined now.

frontend/src/types.ts
```diff
@@ -635,6 +635,27 @@ export const ProgressionStepTitles = {
   [ProgressionStep.SERVER_STARTED]: 'Server started',
 };
 
+export const AssociatedSteps = {
```
andrewballantyne, line 638: Make sure we type this variable, our new array variable, and if you could also do the `ProgressionStepTitles` one as well.

`Record<ProgressionStep, ...whatever the type is...>` -- that way when we add a new enum value, they can keep us honest to filling out things. If you have any questions about this, lemme know.

andrewballantyne (changes requested):
Per our testing session -- here are the things we discussed for you to also change.

I think once we are done these last sets of changes, we should be good to go! 🤞

frontend/src/types.ts
```diff
@@ -605,38 +605,68 @@ export enum ProgressionStep {
...
+    ProgressionStep.PULLING_OAUTH,
+    ProgressionStep.OAUTH_PULLED,
+    ProgressionStep.OAUTH_CONTAINER_CREATED,
+  ],
+  [ProgressionStep.POD_ASSIGNED]: [ProgressionStep.POD_CREATED],
+  [ProgressionStep.SERVER_STARTED]: Object.values(ProgressionStep),
+};
+
+export const OptionalSteps: Record<string, ProgressionStep[]> = {
+  Steps: [
```
andrewballantyne, line 660: Probably don't need the object -- this should just be an array -- I think my typing point may have lead you this way -- sorry! Rework it into an array and type it appropriately

frontend/src/types.ts
```diff
@@ -605,38 +605,68 @@ export enum ProgressionStep {
...
   [ProgressionStep.NOTEBOOK_CONTAINER_STARTED]: 'Workbench container started',
   [ProgressionStep.PULLING_OAUTH]: 'Pulling oauth proxy',
   [ProgressionStep.OAUTH_PULLED]: 'Oauth proxy pulled',
   [ProgressionStep.OAUTH_CONTAINER_CREATED]: 'Oauth proxy container created',
+  [ProgressionStep.OAUTH_PROBLEM]: 'There was a problem with Oauth',
   [ProgressionStep.OAUTH_CONTAINER_STARTED]: 'Oauth proxy container started',
   [ProgressionStep.SERVER_STARTED]: 'Server started',
 };
 
+export const AssociatedSteps: Record<string, ProgressionStep[]> = {
```
andrewballantyne, line 644:
```suggestion
export const AssociatedSteps: Record<Partial<ProgressionStep>, ProgressionStep[]> = {
```
Does this work?

katieperry4: Yeah I fought with that for a bit so we could either stick with 'string' or I could do something like this (it didn't enjoy the Partial):
`export type AssociatedStepKeys =
  | ProgressionStep.NOTEBOOK_CONTAINER_STARTED
  | ProgressionStep.OAUTH_CONTAINER_STARTED
  | ProgressionStep.POD_ASSIGNED
  | ProgressionStep.SERVER_STARTED;

export const AssociatedSteps: Record<AssociatedStepKeys, ProgressionStep[]> = {`
But then we'd need to update the AssociatedStepKeys if there are any changes there but it also might be better than just plain 'string'

andrewballantyne: Hmmm... okay, I'll look -- I'm not sure we want to tie the type to the specific use-case, more define the structure... TS sometimes has issues with this... lemme see what I can do when I check this out.

andrewballantyne:
```suggestion
export const AssociatedSteps: { [key in ProgressionStep]?: ProgressionStep[] } = {
```
Record is such a useful way to write this block -- but effectively it seems it cannot handle partial keys -- looks TS cares more about "not which keys" but "which keys will have values" -- so you have to use the optional `?` indicator to set this up -- this is a slight step forward so that `foo` key cannot be one in this map -- as it requires it to be part of the Enum -- so if we ever remove an Enum value, TS will catch that we are still using it here -- which might happen anyways, as we key off the enum -- but at least it will limit it to only values that will "ever work" and we don't have to assume all `string`s are valid.

frontend/src/concepts/__tests__/StartNotebookModal.spec.tsx
```diff
@@ -76,8 +75,11 @@ describe('Start Notebook modal', () => {
     const stepper = screen.getByTestId('notebook-startup-steps');
     expect(stepper).toBeTruthy();
     const steps = screen.getAllByRole('listitem');
-    expect(steps).toHaveLength(15);
-    expect(steps[1]).toHaveTextContent('Failed to scale-up');
+    expect(steps).toHaveLength(14);
+    /* TODO: This finds the POD_PROBLEM: 'There was a problem with the pod'
+      once text and location of error is finalized with UX set the error location and error text
+    */
+    //expect(steps[3]).toHaveTextContent('Failed to scale-up');
```
andrewballantyne, line 82: Add the test now -- no biggie on pushing stuff -- sometimes UX doesn't get to it right away, we can do best effort.

frontend/src/types.ts
```diff
@@ -605,38 +605,68 @@ export enum ProgressionStep {
   SERVER_REQUESTED = 'SERVER_REQUESTED',
   POD_CREATED = 'POD_CREATED',
   POD_ASSIGNED = 'POD_ASSIGNED',
+  POD_PROBLEM = 'POD_PROBLEM',
```
andrewballantyne, line 608:
```suggestion
  POD_PROBLEM = 'POD_PROBLEM',
  POD_CREATED = 'POD_CREATED',
  POD_ASSIGNED = 'POD_ASSIGNED',
```
A pod problem will prevent the pod from existing most likely... scaling, not enough resources, etc -- all of which prevent a pod from existing. It's an awkward case because there is no "we are starting to create a pod" event like there is for "starting to pull an image"

frontend/src/types.ts
```diff
@@ -605,38 +605,68 @@ export enum ProgressionStep {
...
   PVC_ATTACHED = 'PVC_ATTACHED',
   INTERFACE_ADDED = 'INTERFACE_ADDED',
   PULLING_NOTEBOOK_IMAGE = 'PULLING_NOTEBOOK_IMAGE',
   NOTEBOOK_IMAGE_PULLED = 'NOTEBOOK_IMAGE_PULLED',
   NOTEBOOK_CONTAINER_CREATED = 'NOTEBOOK_CONTAINER_CREATED',
+  NOTEBOOK_PROBLEM = 'NOTEBOOK_PROBLEM',
```
andrewballantyne, line 614:
```suggestion
  NOTEBOOK_CONTAINER_PROBLEM = 'NOTEBOOK_CONTAINER_PROBLEM',
```
Sorry for misleading you -- looks better if you just mirror the other naming conventions


(do the same for OAuth Problem)

frontend/src/utilities/notebookControllerUtils.ts
```diff
@@ -310,7 +312,8 @@ export const getNotebookEventStatus = (
...
   const timestamp = new Date(getEventTimestamp(event)).getTime();
 
-  if (event.message.includes('oauth-proxy')) {
+  // For Oauth-related events
+  if (event.message.includes('oauth-proxy') || event.message.includes('ose-oauth-proxy')) {
```
andrewballantyne, line 316: Funnily enough -- this change actually didn't do anything. It doesn't hurt to keep it -- but my bad (again lol)

frontend/src/utilities/notebookControllerUtils.ts
```diff
@@ -485,11 +497,11 @@ export const useNotebookStatus = (
 
   // Parse the last event
   const lastItem = filteredEvents[filteredEvents.length - 1];
-  const { step, description, status } = getNotebookEventStatus(lastItem, gracePeriod);
+  const { step, status } = getNotebookEventStatus(lastItem, gracePeriod);
 
   return [
     {
-      currentEvent: step ? ProgressionStepTitles[step] : description ?? '',
+      currentEvent: ProgressionStepTitles[step],
```
andrewballantyne, line 504:
```suggestion
      currentEvent: description || ProgressionStepTitles[step],
```
Best have the descriptor over the step name

frontend/src/utilities/notebookControllerUtils.ts
```diff
@@ -520,7 +532,7 @@ export const useNotebookProgress = (
   const progressSteps: NotebookProgressStep[] = Object.values(ProgressionStep).map((step) => ({
     step: ProgressionStep[step],
     percentile: 0,
-    status: isRunning ? EventStatus.SUCCESS : EventStatus.PENDING,
+    status: EventStatus.PENDING,
```
andrewballantyne, line 535: Guess the `success` status here is kinda pointless now that we have the related steps -- good catch

andrewballantyne (changes requested):
Once you do the last set of changes, we can test this -- looking very close!

frontend/src/utilities/notebookControllerUtils.ts
katieperry4: yeah this was the one that was steamrolling all the events into success so even the optional ones were showing up. So we had the other block that assumed previous ones were successful based on the most recently successful event and then this one also forcing all events into success if the server started lol

andrewballantyne (approved):
Very smooth now. 🎉 

https://github.com/user-attachments/assets/7d0c8561-a890-4988-8487-4168cdf900ee

I do see funky stuff with the grace period -- but that's not your effort related (you never touched it) ... so I'll keep investigating that and log something new if I see what's up

PR #3986, opened 2025-03-27
added optimizations to fetch mlmd contexts in a single call

rsun19:
https://issues.redhat.com/browse/RHOAIENG-21973

## Description

Added the `useMlmdContextsByType` hook, which allows for fewer MLMD calls.

## How Has This Been Tested?

Go through the code and see the optimization, where one single call is made instead of many individual ones. Alternatively, you can go to the `Runs` section and inspect network calls. `GetContextsByType` is called only once, but before, `GetContextByTypeAndName` was called many more times, resulting in substantially more MLMD calls.

## Test Impact

Added unit tests

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/concepts/pipelines/apiHooks/mlmd/useMlmdContextsByType.ts
+++ b/frontend/src/concepts/pipelines/apiHooks/mlmd/useMlmdContextsByType.ts
@@ -29,10 +29,10 @@ const getMlmdContextsByType = async (
 export const useMlmdContextsByType = (
   type?: MlmdContextTypes,
   refreshRate?: number,
-): FetchState<Context[] | null> => {
+): FetchState<Context[]> => {
   const { metadataStoreServiceClient } = usePipelinesAPI();
 
-  const call = React.useCallback<FetchStateCallbackPromise<Context[] | null>>(async () => {
+  const call = React.useCallback<FetchStateCallbackPromise<Context[]>>(async () => {
     if (!type) {
       return Promise.reject(new NotReadyError('No context type'));
     }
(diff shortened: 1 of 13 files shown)


frontend/src/concepts/pipelines/apiHooks/mlmd/useMlmdContextsByType.ts
```diff
@@ -0,0 +1,47 @@
...
+  const call = React.useCallback<FetchStateCallbackPromise<Context[] | null>>(async () => {
+    if (!type) {
+      return Promise.reject(new NotReadyError('No context type'));
+    }
+
+    const context = await getMlmdContextsByType(metadataStoreServiceClient, type);
+    return context;
+  }, [metadataStoreServiceClient, type]);
+
+  return useFetchState(call, null, {
```
DaoDaoNoCode, line 44: If you set the default value as `[]` instead of `null` here, it could save you some effort with the type. You don't need to worry about null value anymore.

frontend/src/concepts/pipelines/content/tables/pipelineRun/useMetricColumns.ts
```diff
@@ -52,10 +54,15 @@ export const useMetricColumns = (
   runs: RunWithMetrics[];
   metricsColumnNames: string[];
   runArtifactsLoaded: boolean;
+  contextError: Error | undefined;
   runArtifactsError: Error | undefined;
   metricsNames: Set<string>;
 } => {
-  const [runArtifacts, runArtifactsLoaded, runArtifactsError] = useGetArtifactsByRuns(runs);
+  const [context, , contextError] = useMlmdContextsByType(MlmdContextTypes.RUN);
```
DaoDaoNoCode, line 61: Nit: for naming convention, we'd better call it `contexts` and `contextsError` here, since it returns the array.

Gkrumbach07 (approved):
looks good thanks

/lgtm

PR #3991, opened 2025-03-28
Fix issue on run form when optional string parameter does not have a default value

DaoDaoNoCode:
JIRA: https://issues.redhat.com/browse/RHOAIENG-22654

## Description

This is because we didn't check the type of the defaultValue and always convert it to a string (`String(undefined) === 'undefined'`). The way to fix it is just to check the value before converting it, if it's undefined, set the value as empty string.

## How Has This Been Tested?

1. Go to the global runs page
2. Click on the `Create run` button on the table toolbar (the blue button)
3. In the run form, choose a pipeline that contains optional undefined parameter, e.g. InstructLab pipeline
4. Check the parameter fields, make sure there are no `undefined` string prefilled in the form

## Test Impact

Added a case in the cypress test to double verify the change

## Request review criteria:

If you have UI changes: 


frontend/src/components/ParamsDefaultFields.tsx
```diff
@@ -72,7 +73,7 @@ const ParamsDefaultFields: React.FC<ParamsDefaultFieldsProps> = ({
         <TextInput
           data-testid={inputProps.id}
           {...inputProps}
-          value={String(inputProps.value)}
+          value={inputProps.value === undefined ? '' : String(inputProps.value)}
```
caponetto, line 76:
```suggestion
          value={String(inputProps.value ?? '')}
```
Another option to consider.

FedeAlonso (approved):

Gkrumbach07 (approved):
I wonder if doing this is better ` value={inputProps.value?.toString()}` because then we are not setting it to a value that it isn't but doesn't really matter

PR #3993, opened 2025-03-28
Removes the unused /api/group-config endpoint

andrewballantyne:
https://issues.redhat.com/browse/RHOAIENG-16988

## Description

Removes the `api/group-config` endpoint now that we are using the Auth resource directly from the client.

## How Has This Been Tested?

Testing the User Management page (change things & save)

## Test Impact

Updated any tests that still had use of the fallback code & updated the mocked tests to support the new API.

## Request review criteria:

If you have UI changes: 


emilys314:
I tested the PR image on my cluster and it all seems to be working. That said I'm not 100% confident checking all the scenarios

I checked the admin page as an admin and it worked correctly
image


I also checked as a non admin and it expectedly said i didn't have access to navigate
image

frontend/src/utilities/NavData.tsx
```diff
@@ -233,13 +233,17 @@ const useModelRegisterySettingsNav = (): NavDataHref[] =>
...
+  useAreaCheck<NavDataHref>(
+    SupportedArea.USER_MANAGEMENT,
+    [
+      {
+        id: 'settings-group-settings',
+        label: 'User management',
+        href: '/groupSettings',
+      },
+    ],
+    verbModelAccess('update', AuthModel),
```
emilys314, line 245: What if somebody have read only access to the AuthModel? And they might want to see the page but not change things?

Then would we want to update the verb `update` to something like `get`? Idk if this is a valid use case

PR #3997, opened 2025-04-01
fix the infinitely loading state on execution and artifact page

pnaik1:
https://issues.redhat.com/browse/RHOAIENG-16338

## Description
Fix the infinite loading state when Pipeline run fails to fetch
Screenshot 2025-04-01 at 6 29 41 PM

Screenshot 2025-04-01 at 6 26 21 PM

## How Has This Been Tested?

1. Check the execution detail page of which the pipeline run is deleted
2. Do same for artifact detail page

## Test Impact
Updated and added cypress test

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/__tests__/cypress/cypress/pages/pipelines/artifacts.ts
+++ b/frontend/src/__tests__/cypress/cypress/pages/pipelines/artifacts.ts
@@ -98,6 +98,11 @@ class ArtifactDetails {
     cy.testA11y();
   }
 
+  shouldFailToLoadRun() {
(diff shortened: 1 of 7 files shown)


DaoDaoNoCode (commented):
Code looks good! Just wondering if we could extract this link part out to a single component so it can be reused in both pages you changed. Also, it will be easier to early return the null or `Unknow` or error status if it's an independent component instead inline, which increases the readability (I feel the logic is getting to complex here)

PR #4001, opened 2025-04-01
Add default storage class handling in ModelCustomizationForm (#3995)

Gkrumbach07:
* Integrate useDefaultStorageClass hook to set default storage class
* Update storage class state based on pipeline input and default value
* Ensure storage class is set correctly on component mount

https://issues.redhat.com/browse/RHOAIENG-22858

## Description

sets default storage class to...
- first the default coming from the pipeline
- then if a preferred storage class a available it sets it to that

## How Has This Been Tested?

1. make sure storage classes dont have any metadata set (that annotation called sc-config)
2. open ilab form; before it would of said no storage class found, now it should just not show the storage class and be empty

## Test Impact

tests exist

## Request review criteria:

If you have UI changes: 


frontend/src/pages/pipelines/global/modelCustomization/ModelCustomizationForm.tsx
```diff
@@ -140,6 +144,54 @@ const ModelCustomizationForm: React.FC = () => {
     }),
   );
 
+  const [defaultStorageClass, defaultStorageClassLoaded, defaultStorageClassError] =
+    useDefaultStorageClass();
+  const isStorageClassesAvailable = useIsAreaAvailable(SupportedArea.STORAGE_CLASSES).status;
+  const preferredStorageClass = usePreferredStorageClass();
+  const openshiftDefaultStorageClass = useOpenshiftDefaultStorageClass(true);
```
andrewballantyne, line 151: Any reason this can't be a simple "one stop shop" for storage classes -- this seems like a lot of granular checks for what should be a straight forward check, no?

Gkrumbach07: PVCs benefit from having their storage class auto selected if we dont provide a storage class. We can not provide undefined as an input to the pipeline so this is sort of ilab specific logic. I can move it into reusable hook, but this would only be used for ilab

Gkrumbach07:
/retest 

pls work

pnaik1:
Can we get rid of these line https://github.com/opendatahub-io/odh-dashboard/blob/main/frontend/src/pages/pipelines/global/modelCustomization/trainingHardwareSection/TrainingHardwareSection.tsx#L53-L60 now??

DaoDaoNoCode (approved):
Tested, works well
/lgtm

shalberd:
Hi @Gkrumbach07 very nice, thank you. What is the difference between useAdminStorageClass and useStorageClass ts functions?
Why did you end up splitting that in two helpers? Just out of interest, no criticism.

PR #4002, opened 2025-04-01
Refactor to support deploying models directly from model catalog

mturley:
Resolves https://issues.redhat.com/browse/RHOAIENG-19257

## Description

Several refactors to turn the existing deploy-from-model-registry behavior into generic deploy-prefilled-model logic and leverage it from both registry and catalog.

https://github.com/user-attachments/assets/8405ec78-14e6-4471-956f-ad7c26e1943c

Note: In this recording you can see the model deploy form is visible with an empty name field before the model name is prefilled - this appears to be an existing bug that also occurs when deploying from model registry. Will open a ticket.

* Renames/moves various files from ~/pages/modelRegistry to ~/pages/modelServing:
  * `DeployRegisteredModelModal` -> `DeployPrefilledModelModal`
    * I did this as a `git mv` since most of its contents remain the same and this makes the diff more clear on what parts are abstracted out and shared between deploying from both registry and catalog.
    * The registry-specific replacement for this file is the new file `DeployRegisteredVersionModal`, which has a new name simply to preserve this `git mv` in the diff, but it's also a slightly more accurate name anyway.
    * Instead of pulling model serving related resources from ModelRegistryContext, this component now wraps itself in a `ModelServingContextProvider` using the selected project and pulls its data from that. This allows it to be rendered outside the model registry page and also allows us to remove the model-serving-specific data fetching from ModelRegistryContext, so now those resources are only fetched after clicking Deploy in both registry and catalog.
  * `usePrefillDeployModalFromModelRegistry` -> `usePrefillModelDeployModal`
    * This hook was mostly not registry-specific anyway, its logic now applies to prefilling the deploy model from any source by passing in `ModelDeployPrefillInfo` which may or may not contain `modelRegistryInfo` under it (not used in this hook, only used by `getCreateInferenceServiceLabels` in the `Manage*Modal` components)
  * `useProjectErrorForRegisteredModel` -> `useProjectErrorForPrefilledModel`
    * Wasn't really MR specific. Also, it now consumes ModelServingContext instead of fetching things itself.
  * `useLabeledConnections` (same name, just moved to serving since it's no longer just for MR)
* Takes the old `RegisteredModelDeployInfo` type which was previously defined in `useRegisteredModelDeployInfo`, renames it to `ModelDeployPrefillInfo`, moves it next to its usage in `usePrefillModelDeployModal`, moves its MR-specific properties into a nested `modelRegistryInfo` sub-object.
  * Also renames `useRegisteredModelDeployInfo` -> `useRegisteredModelDeployPrefillInfo` while it stays under ~/pages/modelRegistry, to be consistent with the naming of the new `ModelDeployPrefillInfo` type
  * Also renames its property `modelLocationType` to `connectionTypeName` to be more accurate and to avoid confusion with the `modelLocationType` we use for radio buttons in the MR register model form.
  * I considered renaming the `modelArtifactStorageKey` property to `registeredFromConnectionName` and moving it under `modelRegistryInfo`, but I didn't want to cause more conflicts with [RHOAIENG-19921](https://issues.redhat.com/browse/RHOAIENG-19921) so I'll leave that up to @manaswinidas if she wants to do that.
* Creates a new `DeployCatalogModelModal` which is fairly trivial - unlike the `DeployRegisteredVersionModal`, we don't need to fetch anything extra (there's no need for a `useCatalogModelDeployPrefillInfo` hook), we just declare a `ModelDeployPrefillInfo` object and pass it into `DeployPrefilledModelModal` and everything works 😄
  * Adds the new Deploy button on the catalog model details page which is the primary button, so the register button is now secondary regardless of where it is rendered (removes the `isSecondary` argument there)
  * After deploying, navigates to the top-level model deployments page (previously model serving page) with the target project selected
* To avoid introducing more usage of the magic string `'oci-v1'`, adds new utils `getModelServingConnectionTypeName` and `uriToConnectionTypeName`, and replaces existing hard-coded usages of these OOTB serving-compatible connection names.
* The `~/pages/modelServing/screens/projects/InferenceServiceModal/ProjectSelector.tsx` had some MR-specific props for the sole purpose of adding URL params upon navigating to a project page for setting a platform. Replaces them with a new `projectLinkExtraUrlParams` prop so we can limit the MR-specific code to `DeployRegisteredVersionModal`

## How Has This Been Tested?

Tested manually on ui-green cluster, made sure existing deployment from model registry also works as expected

## Test Impact

Fixed existing unit tests and cypress tests, added new cypress tests for deploying from catalog (more limited than tests for deployment from MR since we only have OCI models in catalog). Refactored out the shared intercepts between modelVersionDeploy.cy.ts and modelCatalogDeploy.cy.ts into new `initDeployPrefilledModelIntercepts` util.

## Request review criteria:

If you have UI changes: 


frontend/src/concepts/modelRegistry/utils.ts
```diff
@@ -69,6 +76,20 @@ export const uriToStorageFields = (uri: string): RegisteredModelLocation => {
   }
 };
 
+export const uriToConnectionTypeName = (uri?: string): string => {
```
manaswinidas, line 79: This needs a unit test

manaswinidas (changes requested):
/lgtm just missing a small unit test

Works fine in `ui-green`.. I don't think we need custom logic with this as all of the URIs will have `registry.redhat.io` and "Current URI" will be preselected.

frontend/src/concepts/modelRegistry/utils.ts
mturley: Done

manaswinidas (approved):
/lgtm

PR #4003, opened 2025-04-01
Support for adding additional unmanaged model catalog sources (second configmap)

YuliaKrimerman:
closes [RHOAIENG-22483](https://issues.redhat.com/browse/RHOAIENG-22483)
## Description

Added a support for additional configmap when the Red Hat source and another one for unmanaged that will have a Third party as a source. We need to make sure we will see both models in 2 different sections on the model Catalog. Make sure 

## How Has This Been Tested?

When testing explore the below scenarios : If it doesn't exist, there's no error rendered, we just see the existing red hat source as expected
If it exists but is malformed, there's an error rendered
If it exists and is empty (sources: [] like in the manifest) we just see the existing red hat source as expected
If it exists and has a source in it, we see two source sections in the UI

## Test Impact

Added test cases and fixed failing ones to mitigate working woth 2 configmaps
## Request review criteria:

Screenshot 2025-03-26 at 2 59 06 PM
Screenshot 2025-03-26 at 3 07 10 PM
Screenshot 2025-03-26 at 3 07 42 PM

If you have UI changes: 

### DIFF
--- a/frontend/src/__tests__/cypress/cypress/tests/mocked/modelCatalog/modelCatalog.cy.ts
+++ b/frontend/src/__tests__/cypress/cypress/tests/mocked/modelCatalog/modelCatalog.cy.ts
@@ -23,7 +23,6 @@ import type { ModelCatalogSource } from '~/concepts/modelCatalog/types';
 type HandlersProps = {
   modelRegistries?: K8sResourceCommon[];
   catalogSources?: ModelCatalogSource[];
-  disableFineTuning?: boolean;
   disableModelCatalogFeature?: boolean;
   hasUnmanagedSourcesConfigMap?: boolean;
   unmanagedSources?: ModelCatalogSource[];
@@ -34,7 +33,6 @@ const initIntercepts = ({
   modelRegistries = [mockModelRegistryService({ name: 'modelregistry-sample' })],
   managedSources = [mockModelCatalogSource({})],
   unmanagedSources = [],
-  disableFineTuning = false,
   disableModelCatalogFeature = false,
   hasUnmanagedSourcesConfigMap = true,
 }: HandlersProps) => {
@@ -44,7 +42,6 @@ const initIntercepts = ({
     mockDscStatus({
       installedComponents: {
         'model-registry-operator': true,
-        'data-science-pipelines-operator': true,
       },
     }),
   );
@@ -53,7 +50,6 @@ const initIntercepts = ({
     'GET /api/config',
     mockDashboardConfig({
       disableModelCatalog: disableModelCatalogFeature,
-      disableFineTuning,
     }),
   );
 
@@ -193,7 +189,7 @@ describe('Model Catalog loading states', () => {
     cy.contains('Unable to load model catalog').should('exist');
   });
 
-  it('should show empty state when configmap has empty data', () => {
+  it('should show empty state when configmap has empty sources', () => {
     cy.interceptK8s(
       {
         model: ConfigMapModel,
@@ -208,12 +204,12 @@ describe('Model Catalog loading states', () => {
           namespace: 'opendatahub',
         },
         data: {
-          modelCatalogSources: '',
+          modelCatalogSources: JSON.stringify({ sources: [] }),
         },
       },
     );
     modelCatalog.visit();
-    cy.contains('Unable to load model catalog').should('exist');
+    cy.contains('Request access to model catalog').should('exist');
   });
 
   it('should show empty state when configmap has malformed data', () => {
--- a/frontend/src/__tests__/cypress/cypress/tests/mocked/modelCatalog/modelCatalog.cy.ts
+++ b/frontend/src/__tests__/cypress/cypress/tests/mocked/modelCatalog/modelCatalog.cy.ts
@@ -22,17 +22,23 @@ import type { ModelCatalogSource } from '~/concepts/modelCatalog/types';
 
 type HandlersProps = {
   modelRegistries?: K8sResourceCommon[];
-  catalogModels?: ModelCatalogSource[];
+  catalogSources?: ModelCatalogSource[];
   disableFineTuning?: boolean;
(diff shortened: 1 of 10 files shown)


manifests/rhoai/shared/apps/model-catalog/model-catalog-unmanaged-sources.yaml
```diff
@@ -0,0 +1,15 @@
...
+  name: model-catalog-unmanaged-sources
+  labels:
+    opendatahub.io/dashboard: 'true'
+    opendatahub.io/managed: 'false'
+data:
+  modelCatalogSources: |
+    {
+      "sources": [
+          ]
+        }
```
gitdallas, line 13: what's this? seems like duplicate `]}`?

mturley: +1 - since we have no long contents we could even just make this value a single line string.

```yaml
data:
  modelCatalogSources: '{ "sources" : [] }'
```

frontend/src/concepts/modelCatalog/useModelCatalogSources.ts
```diff
@@ -27,22 +28,28 @@ export const useModelCatalogSources = (): FetchState<State> => {
       throw new NotReadyError('Model catalog feature is not enabled');
     }
 
-    try {
-      const configMap = await getConfigMap(dashboardNamespace, MODEL_CATALOG_SOURCES_CONFIGMAP);
-      if (!configMap.data || !configMap.data.modelCatalogSources) {
-        return [];
+    const [successes, fails] = await allSettledPromises([
+      getConfigMap(dashboardNamespace, MODEL_CATALOG_SOURCES_CONFIGMAP),
+      getConfigMap(dashboardNamespace, 'model-catalog-unmanaged-sources'),
```
mturley, line 33: `'model-catalog-unmanaged-sources'` should be a constant like `MODEL_CATALOG_SOURCES_CONFIGMAP` is

manifests/rhoai/shared/apps/model-catalog/model-catalog-rbac.yaml
```diff
@@ -13,7 +13,7 @@ rules:
       - configmaps
     resourceNames:
       - model-catalog-sources
----
+      - model-catalog-unmanaged-sources
```
mturley, line 16: You removed the `---` separating the Role and RoleBinding here, so this won't work without adding that back

mturley (changes requested):
A couple things here. Feel free to disagree with my suggestions about organizing intercepts, but I think centralizing things in `initIntercepts` with options passed for any conditional intercepts helps us keep things more readable.

frontend/src/__tests__/cypress/cypress/tests/mocked/modelCatalog/modelCatalog.cy.ts
```diff
@@ -32,25 +53,47 @@ const initIntercepts = ({ disableModelCatalogFeature = false }: HandlersProps) =
...
   it('Model Catalog Disabled in the cluster and URLs should not exist', () => {
     initIntercepts({
       disableModelCatalogFeature: true,
+      hasUnmanagedSourcesConfigMap: false,
     });
 
-    modelCatalog.landingPage();
-
-    cy.findByRole('button', { name: 'Models' }).click();
+    cy.visit('/');
```
mturley, line 96: I think you still need the `cy.findByRole('button', { name: 'Models' }).click();` here right? since it has to open the models menu in the nav to check if its children are there (even though it'll be not there either way, this test isn't testing exactly what you intend).

Also, why remove the `modelCatalog.landingPage()` call? It does the `visitWithLogin()` and has its `wait` command that I think are maybe necessary

frontend/src/__tests__/cypress/cypress/tests/mocked/modelCatalog/modelCatalog.cy.ts
```diff
@@ -161,12 +204,12 @@ describe('Model Catalog loading states', () => {
...
         },
         data: {
-          modelCatalogSources: '',
+          modelCatalogSources: JSON.stringify({ sources: [] }),
         },
       },
     );
     modelCatalog.visit();
-    modelCatalog.findModelCatalogEmptyState().should('exist');
+    cy.contains('Request access to model catalog').should('exist');
```
mturley, line 212: This also should still use `modelCatalog.findModelCatalogEmptyState()`, no need to hard-code this string again

mturley (changes requested):
A couple things I missed in the last review too

YuliaKrimerman:
@gitdallas I love to live dangerously

frontend/src/__tests__/cypress/cypress/tests/mocked/modelCatalog/modelCatalog.cy.ts
mturley: I'm so sorry @YuliaKrimerman 😅 with `landingPage` added back here you don't need `cy.visit('/')`, landingPage calls that via its `visitWithLogin('/')`

frontend/src/__tests__/cypress/cypress/tests/mocked/modelCatalog/modelCatalog.cy.ts
```diff
@@ -209,7 +210,7 @@ describe('Model Catalog loading states', () => {
       },
     );
     modelCatalog.visit();
-    cy.contains('Request access to model catalog').should('exist');
+    modelCatalog.findModelCatalogEmptyState();
```
mturley, line 213: @YuliaKrimerman you still need the `.should('exist')` 🙈 I meant to turn the line back to how it was, sorry

mturley (approved):
LGTM! Thanks @YuliaKrimerman !

PR #4008, opened 2025-04-02
Update zod validation for ilab form

DaoDaoNoCode:
JIRA: [RHOAIENG-22034](https://issues.redhat.com/browse/RHOAIENG-22034)

## Description

Implement a consistent form validation using a hook called `useZodFormValidation`. More details are described in the comments in the hook.
Updated `BaseModelSection`, `TeacherModelSection`, `JudgeModelSection`, `TaxonomySection`, `HardwareSection` and `HyperparemeterSection` to use the hook to do Zod validation separately in the section.

## How Has This Been Tested?

Go to ilab form, and make sure nothing is broken and validation works as expected. There is no UX change.

## Test Impact

Added unit tests for `BaseModelSection`, `TeacherModelSection`, `JudgeModelSection` and `TaxonomySection`.
For `HardwareSection`, its validation will be mostly handled by the HardwareProfileSelect component
For `HyperparemeterSection`, it doesn't worth the effort to write the unit tests, I updated the cypress test to validate the whole form overall.

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/__tests__/cypress/cypress/pages/pipelines/modelCustomizationForm.ts
+++ b/frontend/src/__tests__/cypress/cypress/pages/pipelines/modelCustomizationForm.ts
@@ -67,24 +67,6 @@ class ModelCustomizationFormGlobal {
   findEmptyErrorState() {
     return cy.findByTestId('error-empty-state-body');
   }
-
-  findExpandableSectionButton() {
-    return cy.findByTestId('hyperparameters-expandable').findByRole('button');
-  }
-
-  findNumericInputPlusButton(name: string) {
-    return cy.findByTestId(name).findByRole('button', {
-      name: 'Plus',
-    });
-  }
-
-  findLongNumberInput(name: string) {
-    return cy.findByTestId(name);
-  }
-
-  findRadioInput(name: string) {
-    return cy.findByTestId(name);
-  }
 }
 
 class TeacherModelSection {
@@ -94,6 +76,14 @@ class TeacherModelSection {
 
   findModelNameInput() {
     return cy.findByTestId('teacher-model-name-input');
+  }
+
+  findTokenInput() {
+    return cy.findByTestId('teacher-token-input');
+  }
+
+  findPrivateRadioButton() {
+    return cy.findByTestId('teacher-section-authenticated-endpoint-radio');
   }
 }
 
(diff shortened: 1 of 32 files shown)


frontend/src/__tests__/cypress/cypress/pages/pipelines/modelCustomizationForm.ts
```diff
@@ -164,6 +168,26 @@ class BaseModelSection {
...
+  }
+
+  findLongNumberInput(name: string) {
+    return cy.findByTestId(name);
+  }
+
+  findRadioInput(name: string) {
+    return cy.findByTestId(name);
+  }
 }
```
rsun19, line 191: Just a question: why are we separating these functions, if they're doing the same thing? If all we're doing is finding the test id, wouldn't this be sufficient?

```
findHyperparameterField(name: string) {
   return cy.findByTestId(name);
}
```

frontend/src/concepts/pipelines/content/modelCustomizationForm/modelCustomizationFormSchema/validationUtils.ts
```diff
@@ -98,15 +105,24 @@ export const pipelineParameterSchema = z.record(z.string(), z.any()).superRefine
...
 });
 
+export const trainingNodeSchema = z
+  .number()
+  .refine((val) => val > 0, { message: 'Number must be greater than 0' });
+
+export const storageClassSchema = z
+  .string()
+  .trim()
+  .min(1, { message: 'storage class is required' });
```
rsun19, line 115:
```suggestion
  .min(1, { message: 'Storage class is required' });
```
All of our error messages seem to be capitalized. Should we follow this existing convention?

frontend/src/__tests__/cypress/cypress/pages/pipelines/modelCustomizationForm.ts
DaoDaoNoCode: This is for readability in the test cases. Normally, we will use the following:
```
findRadioInput(name: string) {
  return this.find().findByTestId(name);
}
```
to scope the result within the section in case there are duplicated test id across the page. In this case, we forgot to add the scope. I will update that to add `find()` method to each sections so we don't use `cy.findByTestId()` for specific fields.

rsun19 (commented):
The code lgtm so far. I will look at it again after the taxonomy section is done, then I will start testing it.

DaoDaoNoCode:
@rsun19 OK, it's ready for testing now.

rsun19:
> @rsun19 OK, it's ready for testing now.

I will take a second look in about an hour 🙂

Gkrumbach07 (approved):

PR #4009, opened 2025-04-02
feat(RegisterModel): 13054 change read only fields to basic text

gitdallas:
closes: https://issues.redhat.com/browse/RHOAIENG-13054

## Description
make read only fields at the top of create registry and create version just plain text instead of disabled text fields.

create registry:
image

create version:
image

create version from model:
image

## How Has This Been Tested?
local deploy, green cluster

## Test Impact
none, mostly a styling change

## Request review criteria:

If you have UI changes: 


manaswinidas (approved):
/lgtm small nit: should we have a small test for changes in `RegisterVersion.tsx`? If the other tests pass as is, maybe that's not needed 🤔

gitdallas:
yea previously it would just make a disabled text field, so it's the same text and the same inability to change

mturley (approved):
LGTM, thanks @gitdallas

PR #4010, opened 2025-04-02
fix: infinite loop

olavtar:
https://issues.redhat.com/browse/NVPE-237
https://issues.redhat.com/browse/NVPE-242

## Description

This PR fixes an infinite loop that occurred during the enablement process by annotating the API key secret with a timestamp. It also corrects the logic used to compare timestamps, ensuring the flow proceeds as expected.

## How Has This Been Tested?

Tested locally.

## Test Impact

Manually tested by entering the same incorrect API key twice to verify that the backend error message is displayed correctly without triggering an infinite loop.
Also tested with a valid API key to ensure the enablement process works as expected.

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/backend/src/routes/api/integrations/nim/nimUtils.ts
+++ b/backend/src/routes/api/integrations/nim/nimUtils.ts
@@ -109,10 +109,22 @@ export const manageNIMSecret = async (
   } catch (e: any) {
     if (e.response?.statusCode === 409) {
       // Secret already exists, so update it (replace)
+      // Add annotation to force validation
+      const updatedSecret = {
+        ...nimSecret,
+        metadata: {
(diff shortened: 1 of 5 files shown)


emilys314:
I seem to be getting an infinite loop of "In progress" when i enter an invalid key. 

The nim account does get updated though so it seems like this isn't breaking out of the loop correctly when the key validation fails

emilys314:
> I seem to be getting an infinite loop of "In progress" when i enter an invalid key.
> 
> The nim account does get updated though so it seems like this isn't breaking out of the loop correctly when the key validation fails

For documentation, when you close the modal and open it a second time, this will cause an infinite loop after submitting a wrong key

PR #4014, opened 2025-04-03
Permission fix for project users

ashley-o0o:
Closes: [RHOAIENG-18968](https://issues.redhat.com/browse/RHOAIENG-18968)

## Description
Corrected a permission bug that always blocked the user from seeing the actions toggle. 

When logged in as a non-admin user, you should now be able to see the Actions dropdown option for projects you’ve created. It should still be hidden for projects created by the Admin.

I’ve also created a new file for this permission check to help with future usage

## How Has This Been Tested?
Tested locally as a non-admin user

## Test Impact
No tests changed

## Screenshots
Before (Actions dropdown hidden):
Screenshot 2025-04-03 at 1 07 39 PM

After (Actions available):
Screenshot 2025-04-03 at 1 07 01 PM

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/__tests__/cypress/cypress/utils/mockUsers.ts
+++ b/frontend/src/__tests__/cypress/cypress/utils/mockUsers.ts
@@ -108,6 +108,7 @@ const setUserConfig = (userConfig: UserConfig = {}, isAllowed = true) => {
     },
   );
 
+  const PROJECT_ADMIN_RESOURCES = ['projects', 'rolebindings'];
   const EDIT_VERBS = ['*', 'create', 'delete', 'deletecollection', 'update', 'patch'];
   cy.interceptK8s('POST', SelfSubjectAccessReviewModel, (req) => {
     const { verb, group, resource, namespace } = req.body.spec.resourceAttributes;
@@ -124,8 +125,8 @@ const setUserConfig = (userConfig: UserConfig = {}, isAllowed = true) => {
             : // self provisioner capabilities grant access to project creation
             resource === 'projectrequests'
             ? isSelfProvisioner
-            : // only project admins can create rolebindings
-            resource === 'rolebindings' && EDIT_VERBS.includes(verb)
+            : // project admins
+            PROJECT_ADMIN_RESOURCES.includes(resource) && EDIT_VERBS.includes(verb)
             ? isProjectAdmin
             : isProductAdmin
             ? // product admins are getting direct access to resources in the deployment namespace (but importantly, not other projects)
--- a/frontend/src/__tests__/cypress/cypress/utils/mockUsers.ts
+++ b/frontend/src/__tests__/cypress/cypress/utils/mockUsers.ts
@@ -122,7 +122,7 @@ const setUserConfig = (userConfig: UserConfig = {}, isAllowed = true) => {
           isClusterAdmin
             ? true
             : // self provisioner capabilities grant access to project creation
-            resource === 'projectrequests' || resource === 'projects'
+            resource === 'projectrequests'
             ? isSelfProvisioner
             : // only project admins can create rolebindings
             resource === 'rolebindings' && EDIT_VERBS.includes(verb)
--- a/frontend/src/__tests__/cypress/cypress/utils/mockUsers.ts
+++ b/frontend/src/__tests__/cypress/cypress/utils/mockUsers.ts
@@ -122,7 +122,7 @@ const setUserConfig = (userConfig: UserConfig = {}, isAllowed = true) => {
           isClusterAdmin
             ? true
             : // self provisioner capabilities grant access to project creation
-            resource === 'projectrequests'
+            resource === 'projectrequests' || resource === 'projects'
             ? isSelfProvisioner
             : // only project admins can create rolebindings
             resource === 'rolebindings' && EDIT_VERBS.includes(verb)
--- a/frontend/src/pages/projects/screens/projects/useProjectTableRowItems.ts
+++ b/frontend/src/pages/projects/screens/projects/useProjectTableRowItems.ts
@@ -1,8 +1,8 @@
 import * as React from 'react';
 import { useNavigate } from 'react-router-dom';
 import { TooltipProps } from '@patternfly/react-core';
-import { useAccessReview } from '~/api';
-import { AccessReviewResourceAttributes, ProjectKind } from '~/k8sTypes';
+import { ProjectKind } from '~/k8sTypes';
+import { useProjectAccessReview } from '~/concepts/projects/accessChecks';
 
 type KebabItem = {
   title?: string;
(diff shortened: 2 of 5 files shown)


frontend/src/pages/projects/screens/projects/useProjectTableRowItems.ts
```diff
@@ -11,11 +11,6 @@ type KebabItem = {
   onClick?: () => void;
   tooltipProps?: TooltipProps;
 };
-const accessReviewResource: AccessReviewResourceAttributes = {
-  group: 'rbac.authorization.k8s.io',
-  resource: 'rolebindings',
-  verb: 'create',
-};
```
emilys314, line 18: I'm not 100% sure on the move away from checking access to modify the role bindings. 

Part of the editing as an admin is to grant roles to other users, I think this is different than modifying the project itself? I think right now they are both given in tandem so it still works?

emilys314 (commented):
Everything does seem to work now 

image

frontend/src/pages/projects/screens/projects/useProjectTableRowItems.ts
ashley-o0o: This is what @andrewballantyne had suggested doing. Still learning about how this works exactly, but he wanted to get away from using the rolebindings here.

emilys314: Okay cool thanks, I'm not that familiar with this area either so

frontend/src/__tests__/cypress/cypress/utils/mockUsers.ts
```diff
@@ -122,7 +122,7 @@ const setUserConfig = (userConfig: UserConfig = {}, isAllowed = true) => {
           isClusterAdmin
             ? true
             : // self provisioner capabilities grant access to project creation
-            resource === 'projectrequests'
+            resource === 'projectrequests' || resource === 'projects'
```
andrewballantyne, line 125: Was this to make a test pass?

Griffin-Sullivan (approved):
/lgtm

Screenshot From 2025-04-08 11-13-38

One thing that stands out to me is we have the mock test 
`it('Should allow actions for provisioning users'` that should have covered this. Might want to look into it more. There's also another test that ensure non project admins don't see the edit button.

andrewballantyne (changes requested):
These mock changes seem a little weird -- going to investigate. 

Tiny suggestion to add a comment to the new hook to help run home the value of it.

frontend/src/__tests__/cypress/cypress/utils/mockUsers.ts
andrewballantyne, line 125: @Griffin-Sullivan Ever so close... you spotted it and didn't investigate enough 🙂 

This is actually _wrong_ -- because the test is bad.

* Self Provisioning is the act of "can I create a project".
* Being an admin of a project is the permissions to Delete, Create, Update, and add more people (via adding permissions) to a project
* **_Typically_** one would not make someone an admin of a project if it was being created FOR them... since they can "delete" the project and then be SOL because they can't recreate

However, and this is a big however. Permissions are granular. K8s allows additive-only permissions (aka you cannot remove permissions without going to the source of the add and removing it) but one can grant access to Update & Patch on a Project on top of a contributor so they can add display name and description changes. It's likely not done a ton, but it **IS** technically possible -- so we gotta handle it.

This all said, there are some changes you need to do:
* Move this check down two lines
  ```suggestion
              resource === 'projectrequests'
            ? isSelfProvisioner
            : // project admins
            PROJECT_ADMIN_RESOURCES.includes(resource) && EDIT_VERBS.includes(verb)
  ```
  & add:
  ```
    const PROJECT_ADMIN_RESOURCES = ['projects', 'rolebindings'];
  ```
* Delete the test `Should not allow actions for non-provisioning users` in `ProjectDetails.cy.ts` -- it's making an assumption that is fundamentally not true
* I'd encourage you to make sure you write a mocked test to validate this state is sane for project admins, product admins, and a "project edit" users (aka basic users invited to contribute on a project).
  You can see the helper functions at the top of this file
  
Let me know if you want to talk about this review or the changes needed. Happy to detail them out better.

andrewballantyne (changes requested):
Yup, good spot by Griffin -- we do have a bug here that the tests are hiding after your change. Suggestions on what to do included.

frontend/src/pages/projects/screens/projects/useProjectTableRowItems.ts
```diff
@@ -69,11 +51,11 @@ const useProjectTableRowItems = (
     },
     {
       title: 'Edit permissions',
-      isAriaDisabled: !allowCreate || !allowCreateLoaded,
+      isAriaDisabled: !allowUpdate || !allowUpdateLoaded,
       onClick: () => {
         navigate(`/projects/${project.metadata.name}?section=permissions`);
       },
-      ...noPermissionToolTip(allowCreate, allowCreateLoaded),
+      ...noPermissionToolTip(allowUpdate, allowUpdateLoaded),
```
andrewballantyne, line 58: During our conversation, this actually is accurate to being needing create against Role Bindings -- because that's what this tab does...

andrewballantyne (changes requested):

frontend/src/pages/projects/screens/projects/useProjectTableRowItems.ts
```diff
@@ -69,7 +59,7 @@ const useProjectTableRowItems = (
     },
     {
       title: 'Edit permissions',
-      isAriaDisabled: !allowCreate || !allowCreateLoaded,
+      isAriaDisabled: !allowUpdate || !allowUpdateLoaded,
```
andrewballantyne, line 62: You'll need to revert this.

frontend/src/pages/projects/screens/projects/useProjectTableRowItems.ts
```diff
@@ -25,27 +23,19 @@ const useProjectTableRowItems = (
...
+    'update',
+    project.metadata.name,
     shouldRunCheck,
   );
-  const [allowCreate, allowCreateLoaded] = useAccessReview(
-    {
-      ...accessReviewResource,
-      namespace: project.metadata.name,
-    },
+  const [allowCreate, allowCreateLoaded] = useProjectPermissionsAccessReview(
```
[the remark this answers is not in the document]
ashley-o0o: Agreed, `allowPermissions` makes more sense

andrewballantyne (approved):
Thanks for walking down this wild path of getting this done. This is definitely at a good merge spot. I'm going to log another issue because there are additional things that need to happen overall here to sync things up better 🙂 (all outside the issue you were addressing but in the same space)

For posterity, the followup: https://issues.redhat.com/browse/RHOAIENG-23534

PR #4021, opened 2025-04-07
Add Model Serving Global mock test for all projects

Griffin-Sullivan:
https://issues.redhat.com/browse/RHOAIENG-22512
## Description
As part of the model serving refactor planning, we decided to audit our test coverage and fill in gaps. One major area we are lacking is in the Global Model Serving page when filtering by All Projects. This test adds a row for this table for each serving type we support and makes sure the page works correctly.

## How Has This Been Tested?
Ran locally

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/__mocks__/mockNimResource.ts
+++ b/frontend/src/__mocks__/mockNimResource.ts
@@ -45,11 +45,20 @@ export const mockNimImages = (): ConfigMapKind =>
     },
   });
 
-export const mockNimInferenceService = (): InferenceServiceKind => {
+type NimInferenceService = {
+  namespace?: string;
+  displayName?: string;
+};
+
+export const mockNimInferenceService = ({
+  displayName = 'Test Name',
+  namespace = 'test-project',
+}: NimInferenceService = {}): InferenceServiceKind => {
   const inferenceService = mockInferenceServiceK8sResource({
     name: 'test-name',
     modelName: 'test-name',
-    displayName: 'Test Name',
+    displayName,
+    namespace,
     kserveInternalLabel: true,
     resources: {
       limits: { cpu: '16', memory: '64Gi' },
@@ -124,10 +133,25 @@ export const mockNvidiaNimImagePullSecret = (): SecretKind => {
   return secret;
 };
 
-export const mockNimProject = (hasAllModels: boolean): ProjectKind => {
+type NimProjectType = {
+  hasAllModels?: boolean;
+  k8sName?: string;
+  displayName?: string;
+  enableNIM?: boolean;
+};
+
+export const mockNimProject = ({
+  hasAllModels,
+  k8sName,
+  displayName,
+  enableNIM,
+}: NimProjectType): ProjectKind => {
   const project = mockProjectK8sResource({
     hasAnnotations: true,
     enableModelMesh: hasAllModels ? undefined : false,
+    k8sName,
+    displayName,
+    enableNIM,
   });
   if (project.metadata.annotations != null) {
     project.metadata.annotations['opendatahub.io/nim-support'] = 'true';
--- a/frontend/src/__tests__/cypress/cypress/tests/mocked/modelServing/modelServingGlobal.cy.ts
+++ b/frontend/src/__tests__/cypress/cypress/tests/mocked/modelServing/modelServingGlobal.cy.ts
@@ -17,6 +17,7 @@ import {
   kserveModal,
   kserveModalEdit,
   modelServingGlobal,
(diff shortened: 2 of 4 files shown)


katieperry4:
Screenshot 2025-04-09 at 2 02 20 PM
Screenshot 2025-04-09 at 2 03 08 PM

I read through the changes and ran it locally 😀  everything makes sense as far as I can tell and all the tests are working. I messed with them a bit and they broke when I broke the mock a little which was good!

I do have a question though 😅 , in the last edited chunk of modelServingGlobal.cy.ts:
```    
    // Open each modal and make sure it is the correct one
    modelServingGlobal.getModelRow('KServe Model').findKebabAction('Edit').click();
    // KServe Modal has section at the bottom for configuring params where as Model Mesh does not
    kserveModalEdit.findConfigurationParamsSection().should('exist');
    kserveModalEdit.findCancelButton().click();

    modelServingGlobal.getModelRow('Model Mesh Model').findKebabAction('Edit').click();
    inferenceServiceModalEdit.findConfigurationParamsSection().should('not.exist');
    inferenceServiceModalEdit.findCancelButton().click();

    // NIM Modal is essentially the same layout as KServe so we check the runtime is correct
    modelServingGlobal.getModelRow('NIM Model').findKebabAction('Edit').click();
    kserveModalEdit.findServingRuntimeTemplate().should('have.text', 'NVIDIA NIM');
    kserveModalEdit.findCancelButton().click();
```
To me it makes sense to check the NIM modal based on the serving runtime, but why not check the other ones that way? Or chain on the serving runtime check to the others in addition to checking for the configuring params section? I haven't done too much with Cypress yet so I'm not 100% sure how this works so maybe checking for the params section is more solid 🤷 


But besides my question I think it looks good!
/lgtm

Griffin-Sullivan:
@katieperry4 That's a good question. The reason I'm checking the params section on the modal is because that's what the KServe Modal has that the Model Mesh Modal does not (easy way for me to see that the right modal opened). For NIM we are basically using the KServe Modal so the only way I can distinguish that the NIM modal is open is by checking the runtime is NVIDIA NIM. 

I'm not checking run times for the other models since they are just using a standard mocked runtime (I think OpenVINO for both) so there's nothing there to verify really.

frontend/src/__mocks__/mockNimResource.ts
```diff
@@ -45,11 +45,20 @@ export const mockNimImages = (): ConfigMapKind =>
...
+
+export const mockNimInferenceService = ({
+  displayName = 'Test Name',
+  namespace = 'test-project',
+}: NimInferenceService): InferenceServiceKind => {
   const inferenceService = mockInferenceServiceK8sResource({
     name: 'test-name',
     modelName: 'test-name',
-    displayName: 'Test Name',
+    displayName: displayName || 'Test Name',
```
andrewballantyne, line 60: You double defaulted? Why? Once in the signature of the function & one more here...

frontend/src/__tests__/cypress/cypress/pages/modelServing.ts
```diff
@@ -111,6 +111,10 @@ class InferenceServiceModal extends Modal {
     return this.findFooter().findByTestId('modal-submit-button');
   }
 
+  findCancelButton(): Cypress.Chainable<JQuery<HTMLElement>> {
+    return this.findFooter().findByTestId('modal-cancel-button');
+  }
+
```
andrewballantyne, line 117: This is already implemented in the base `Modal` that this class extends -- why is that not working?

frontend/src/__mocks__/mockNimResource.ts
```diff
@@ -45,11 +45,20 @@ export const mockNimImages = (): ConfigMapKind =>
...
-export const mockNimInferenceService = (): InferenceServiceKind => {
+type NimInferenceService = {
+  namespace?: string;
+  displayName?: string;
+};
+
+export const mockNimInferenceService = ({
+  displayName = 'Test Name',
+  namespace = 'test-project',
+}: NimInferenceService): InferenceServiceKind => {
```
andrewballantyne, line 56:
```suggestion
}: NimInferenceService = {}): InferenceServiceKind => {
```

Seems if you did this you'd avoid a bunch of unnecessary changes

andrewballantyne (commented):
Nits -- testing

andrewballantyne:
/un-lgtm
/approve

@Griffin-Sullivan merge conflict 🙂

emilys314:
NIM will update tests themselves

PR #4026, opened 2025-04-08
introduce extension points to ODH

christianvogt:
https://issues.redhat.com/browse/RHOAIENG-23185

## Description

This PR introduces extension points mechanism to ODH.
It re-uses the `@openshift/dynamic-plugin-sdk` where possible. It doesn't utilize it entirely because I did not want to introduce a build aspect to this mechanism. Also by not utilizing the existing library as is, we can benefit from creating code refs in typescript and benefit from the types through dynamic import support for code refs.

There was some code copied and changed from the dynamic-plugin-sdk (eg. `useResolvedExtensions`) because internally it uses a symbol which we cannot reproduce.

How to use:
- Create new extension point files in the `src/plugins/extension-points/` directory. 
- Re-export the extension points from the `src/plugins/extension-points/index.ts` file.
- Add extensions to `src/plugins/extensions`.
- Use `useExtensions` or `useResolvedExtensions` from `src/plugins/` and not from `@openshift/dynamic-plugin-sdk`

Note:
The change to babel config is for jest. It prevents dynamic imports from being transformed. This is useful for our test case which validates the declared extensions to ensure they are adhering to the code ref pattern. However this may interfere with actual unit tests that require dynamic imports in the future.

## How Has This Been Tested?

- Tested with an example extension point locally.

This PR does not introduce any extension points but does introduce a new context. The application should function exactly as before.

## Test Impact

- Added unit tests

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/plugins/__tests__/extensions.spec.ts
+++ b/frontend/src/plugins/__tests__/extensions.spec.ts
@@ -1,0 +1,95 @@
+import { Extension } from '@openshift/dynamic-plugin-sdk';
+import extensionDeclarations from '~/plugins/extensions';
+import { isCodeRef } from '~/plugins/internal/coderefs';
+import { visitDeep } from '~/plugins/internal/objects';
+
+describe('extensions', () => {
+  it('should be valid', () => {
+    expect(validateExtensions(extensionDeclarations)).toBe(true);
+  });
+});
+
+// Tests the test function
+describe('validateExtensions', () => {
+  it('should validate a valid code ref', () => {
+    expect(
+      validateExtensions([
+        {
+          type: 'test',
+          properties: {
+            firstLevel: {
+              props: {
+                // This is an invalid import used for testing purposes.
+                // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+                // @ts-ignore
+                content: () => import('./MyModule'),
+              },
+            },
+            bool: true,
+          },
+        },
+      ]),
+    ).toBe(true);
+  });
+
+  it('should validate a valid code ref with a named export', () => {
+    expect(
+      validateExtensions([
+        {
+          type: 'test',
+          properties: {
+            firstLevel: {
+              props: {
+                // This is an invalid import used for testing purposes.
+                // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+                // @ts-ignore
+                content: () => import('./MyModule').then((module) => module.namedExport),
+              },
+            },
+            bool: true,
+          },
+        },
+      ]),
+    ).toBe(true);
+  });
+
+  it('should validate an invalid code ref', () => {
+    expect(
+      validateExtensions([
+        {
+          type: 'test',
+          properties: {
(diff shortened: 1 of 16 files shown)


frontend/src/plugins/__tests__/extensions.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+    expect(
+      validateExtensions([
+        {
+          type: 'test',
+          properties: {
+            firstLevel: {
+              props: {
+                // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+                // @ts-ignore
+                content: () => import('./Test'),
```
andrewballantyne, line 24: This looks weird -- tests using `./Test` -- but then it's not a test file -- it's just a placeholder. Looking at it in my IDE is definitely better -- but I worry it's a weird optics for a test file

frontend/src/plugins/extension-points/index.ts
```diff
@@ -0,0 +1,24 @@
+import type { CodeRef } from '@openshift/dynamic-plugin-sdk';
+
+// eslint-disable-next-line @typescript-eslint/ban-types
+export type ComponentCodeRef<Props = {}> = CodeRef<{ default: React.ComponentType<Props> }>;
```
andrewballantyne, line 4:
```
import { AnyObject } from '~/typeHelpers';
```

```suggestion
export type ComponentCodeRef<Props = AnyObject> = CodeRef<{ default: React.ComponentType<Props> }>;
```

frontend/src/plugins/internal/promise.ts
andrewballantyne, line 1: What about what we already have? https://github.com/opendatahub-io/odh-dashboard/blob/main/frontend/src/utilities/allSettledPromises.ts

andrewballantyne (changes requested):
Neat :bender-neat: -- I'll have to incorporate effort here in our Model Serving Refactor PoC as we are intending to break apart most things into effective isolated modules

Do we have a name for what we will call these? Mico-frontends? Plugins? (you named the folder this, maybe we embrace it?)

frontend/src/plugins/__tests__/extensions.spec.ts
christianvogt: That's a fair point. Something like `ExampleComponent` is probably a better name to use.

frontend/src/plugins/internal/promise.ts
christianvogt: Everything in the `internal` dir was straight copied from dynamic-plugin-sdk for ease of integration simply because of the need to support a customized `useResolvedExtensions`. I can look to simplify things here but I figured this was going to eventually get cleaned up.

christianvogt:
Applied the requested changes and rebased.
@andrewballantyne let me know how your trial goes with this.

andrewballantyne (approved):

andrewballantyne (approved):

PR #4031, opened 2025-04-09
added pass through patch api method for storageclasses

rsun19:
https://issues.redhat.com/browse/RHOAIENG-22900

## Description

Migrated the patch endpoint to the frontend.

## How Has This Been Tested?

First, play around with the storage class screen and make sure that all of the buttons and functionality work. Then follow the below steps.

Tested on Ilab cluster. I will keep the role and rolebindings if you want to test there.

Basically, you need to create a new role with the following permissions:

```
rules:
  - verbs:
      - update
      - patch
      - get
      - list
      - watch
    apiGroups:
      - storage.k8s.io
    resources:
      - storageclasses
```

Then, bind it to the `admin user`. 

### NOTE: You do not need to do the above steps I just mentioned if you are using the ilab cluster. It is already created and applied

Next, you need to add both the `admin user` and the `regular user` to the `odh-admins` group to gain access to the storage classes screen. 

After, you need to spin down the operator, and remove the storage class permissions from the `admingroupcluster-role` because it is binded to `odh-admins` group.

**Note that in the ilab cluster, you can spin this down by going to `Deployments` -> `opendatahub-operator-controller-manager`**

Now test with the `admin user` and `regular user`. It should work. 

## Test Impact

Added cypress and unit tests

## Request review criteria:

If you have UI changes: 


frontend/src/api/k8s/storageClasses.ts
```diff
@@ -1,9 +1,57 @@
...
+
+export const getStorageClass = (name: string): Promise<StorageClassKind> =>
+  k8sGetResource<StorageClassKind>({
+    model: StorageClassModel,
+    queryOptions: { name },
+  });
+
+export const updateStorageClassConfig = async (
+  name: string,
+  config: Partial<Omit<StorageClassConfig, 'lastModified'>> | undefined,
```
DaoDaoNoCode, line 25: There is no need to use both `Partial` and `Omit` here IMO, the `Partial` has already made all the props in the type to be optional.

DaoDaoNoCode (commented):
You can also remove the type definition in the backend folder [here](https://github.com/opendatahub-io/odh-dashboard/blob/main/backend/src/types.ts#L632-L638)

frontend/src/api/k8s/storageClasses.ts
```diff
@@ -1,9 +1,57 @@
...
+  opts?: K8sAPIOptions,
+): Promise<StorageClassConfig> => {
+  const oldStorageClassResource = await getStorageClass(name);
+  const patches: Patch[] = [
+    {
+      op: 'replace',
+      path: '/metadata/annotations/opendatahub.io~1sc-config',
+      value: config
+        ? JSON.stringify({
+            ...(oldStorageClassResource.metadata.annotations?.['opendatahub.io/sc-config'] &&
```
DaoDaoNoCode, line 35: Nit: For all the 'opendatahub.io/sc-config', we can replace it with the type [here](https://github.com/opendatahub-io/odh-dashboard/blob/main/frontend/src/k8sTypes.ts#L64)

frontend/src/api/k8s/storageClasses.ts
```diff
@@ -1,9 +1,57 @@
...
+              JSON.parse(oldStorageClassResource.metadata.annotations['opendatahub.io/sc-config'])),
```
DaoDaoNoCode, line 36: There is one different behavior after you made this change.
**Before:**
When you have an invalid storage class config (e.g. There is a parse error when you parse the annotation, you can do this by editing the annotation to remove one of the quote symbols from the props), the storage class will automatically reset to default values and remove other props like description when you load the storage class settings page. Because it goes to [this path](https://github.com/opendatahub-io/odh-dashboard/blob/main/frontend/src/pages/storageClasses/StorageClassesContext.tsx#L77-L93)
**After:**
It won't auto-update the invalid storage class anymore, and shows the corrupted alert instead, because there could be a parse error here and the patch won't work, so it's not triggering the update call.
Screenshot 2025-04-14 at 3 08 26 AM

I think the behavior looks actually more reasonable after you made this update, just want to point it out and see what others think. You can also give your opinions here. Maybe we should still handle the parse error but in a different way?
cc @Gkrumbach07 to raise attention to this behavior change.

rsun19:
The new logic adds linear space in the worst case, but that is unlikely to be reached unless the user deliberately intends to create invalid storage classes. It is minimal and does not have effects outside of the file.

rsun19:
addressed comments

frontend/src/pages/storageClasses/StorageClassesContext.tsx
```diff
@@ -68,8 +67,22 @@ export const StorageClassContextProvider: React.FC<StorageClassContextProviderPr
...
-    const updateRequests = Object.entries(storageClassConfigs).reduce(
-      (acc: Promise<ResponseStatus>[], [name, config], index) => {
+    const updateRequests = storageClasses.reduce(
+      (acc: Promise<StorageClassConfig>[], storageClass, index) => {
+        const { name } = storageClass.metadata;
+        let config;
+        if (storageClass.metadata.annotations?.[MetadataAnnotation.OdhStorageClassConfig]) {
+          try {
+            config = JSON.parse(
+              storageClass.metadata.annotations[MetadataAnnotation.OdhStorageClassConfig] || '',
```
DaoDaoNoCode, line 77: `|| ''` is not necessary here.

DaoDaoNoCode:
@rsun19 Great work! I tested everything and it works as expected, just need to address my comment here https://github.com/opendatahub-io/odh-dashboard/pull/4031#pullrequestreview-2763230005 then I can approve it.

rsun19:
Just removed the types

Gkrumbach07:
we should be go to with this one.  going to test again that the cluster is getting the right permissions

Gkrumbach07:
ok this works. tested on a fresh 1.28 odh operator deployment with this pr image. admin users with no other perms other than being part of the dashboard admin group have access to update. It is not calling the backend /storage-class endpoint anymore

Gkrumbach07 (approved):
/lgtm

PR #4035, opened 2025-04-09
Fix dependent legacy hardware profile migration from edit page

Gkrumbach07:
https://issues.redhat.com/browse/RHOAIENG-22843

## Description

- accelerator profiles now correctly create two hardware profiles on migration instead of just the one being edited

## How Has This Been Tested?

- create an accelerator profile
- go to hardware profile manage page
- find legacy accelerator profile in expandable table section
- edit one of the two accelerator profiles
- submit form
- confirm modal and migrate
- two new hardware profile should be created

## Test Impact

added a mocked cypress test

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/__mocks__/mockHardwareProfile.ts
+++ b/frontend/src/__mocks__/mockHardwareProfile.ts
@@ -60,9 +60,7 @@ export const mockHardwareProfile = ({
       effect: TolerationEffect.NO_SCHEDULE,
     },
   ],
-  nodeSelector = {
-    test: 'value',
-  },
+  nodeSelector,
   annotations,
   labels,
 }: MockResourceConfigType): HardwareProfileKind => ({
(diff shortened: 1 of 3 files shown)


rsun19:
code looks good and works, will come back after tests are fixed

rsun19:
> Not sure if I tested it correctly, but I am always getting this error no matter how I change the node resources, it keeps using the old value and throw this error. This happens on the ilab shared cluster when you edit the legacy accelerator profile. Not sure if it's related to this ticket.

I tested on my own cluster and it worked. Not sure why yours is different though

rsun19:
Actually, I am facing this issue as well on the ilab cluster trying to do QA contact validation

DaoDaoNoCode:
@rsun19 I checked the code, looks like it's always using the old values to create the dependent hardware profile when migration https://github.com/opendatahub-io/odh-dashboard/blob/main/frontend/src/pages/hardwareProfiles/migration/utils.ts#L163-L191. When there are some invalid values in the old values, there is no way to migrate.

PR #4036, opened 2025-04-10
Verify components versions in About dialog

manosnoam:
https://issues.redhat.com/browse/RHOAIENG-21403

## Description
Created e2e/applications/testAboutDialog.cy.ts to verify that the listed component versions in the new table of the About Dialog correspond to the actual components' version installed.

In addition, added new functionality to E2E, which is required for testAboutDialog:

`SoftAssert` - It allows the verification of multiple steps, like every row in the versions table, without failing immediately if the version is mismatched. 
Instead, it runs any test steps that uses soft assert (e.g., verify all versions), report it, and only at the end, mark the test as failed if any of the SoftAssert steps did not pass. 

A usage example for soft assert via ` softTrue(...)` in an E2E step:

```
cy.step(`Verify ${productName} version in About dialog`);
      getVersionFromCsv(odhCsv as { spec: { version: string } }).then((version) => {
        aboutDialog
          .findProductVersion()
          .invoke('text')
          .should('match', /\d/) // Wait until text contains at least one digit
          .then((uiVersion) => {
            softTrue(
              version.includes(uiVersion),
              `Expected version '${version}' to include '${uiVersion}'`,
            );
          });
      });
```

## How Has This Been Tested?
Tested locally and via Jenkins (with all other tests), and on multiple clusters:
image

## Test Impact
`e2e/applications/testAboutDialog.cy.ts`

## Request review criteria:


frontend/src/__tests__/cypress/cypress/tests/e2e/applications/explore/testEnabledISVs.cy.ts
```diff
@@ -4,7 +4,7 @@ import { getOcResourceNames } from '~/__tests__/cypress/cypress/utils/oc_command
 import { filterRhoaiIfHidden } from '~/__tests__/cypress/cypress/utils/appCheckUtils';
 import { retryableBefore } from '~/__tests__/cypress/cypress/utils/retryableHooks';
 
-const applicationNamespace = Cypress.env('TEST_NAMESPACE');
+const applicationNamespace = Cypress.env('ODH_NAMESPACE');
```
antowaddle, line 7: The naming here suggests this is an ODH specific namespace but these tests also execute on RHOAI. Is there any specific requirement/need to change this? 'TEST_NAMESPACE' is probably not idea but if we're changing it ideally it should be something more generic.

frontend/src/__tests__/cypress/cypress/tests/e2e/applications/testAboutDialog.cy.ts
```diff
@@ -0,0 +1,101 @@
...
+  it(
+    'Verify RHODS About Dialog contains correct information',
+    { tags: ['@Smoke', '@SmokeSet1', '@Dashboard', '@RHOAIENG-21403'] },
+    () => {
+      // Login and open the About dialog
+      cy.step(`Login to ${productName}`);
+      cy.visitWithLogin('/', HTPASSWD_CLUSTER_ADMIN_USER);
+
+      cy.step('Open the About dialog');
+      appChrome.visit();
```
antowaddle, line 36: Typically this is put inside a .navigate page object, can you make that change here please?

frontend/src/__tests__/cypress/cypress/tests/e2e/applications/testAboutDialog.cy.ts
```diff
@@ -0,0 +1,101 @@
...
+
+      cy.step(`Verify ${productName} version in About dialog`);
+      getVersionFromCsv(odhCsv as { spec: { version: string } }).then((version) => {
+        aboutDialog
+          .findProductVersion()
+          .invoke('text')
+          .should('match', /\d/) // Wait until text contains at least one digit
+          .then((uiVersion) => {
+            softTrue(
+              version.includes(uiVersion),
```
antowaddle, line 59: Is there a more concrete way to find the version? The addition of Soft assertion is good but I'm concerned that this test will fail when maybe there's an issue.

frontend/src/__tests__/cypress/cypress/tests/e2e/applications/testAboutDialog.cy.ts
```diff
@@ -0,0 +1,101 @@
...
+                return aboutDialog.getComponentReleasesText(component);
+              })
+              .then((texts) => {
+                getResourceVersionByName(component).then((version) => {
+                  if (version.length === 0) {
+                    return;
+                  }
+                  const text = texts.join(' ');
+                  softTrue(
+                    version.some((v) => text.includes(v)),
```
antowaddle, line 92: Again, is this something that can be verified to find the actual version and compare it against the UI?

antowaddle (changes requested):
Thanks for this PR @manosnoam, I've added some comments.

Additionally, can you also execute the test against both RHOAI and ODH?

frontend/src/__tests__/cypress/cypress/tests/e2e/applications/testAboutDialog.cy.ts
manosnoam: This is similar to what we're already doing in `tests/mocked/applications/application.cy.ts`.

frontend/src/__tests__/cypress/cypress/tests/e2e/applications/testAboutDialog.cy.ts
manosnoam: Yes, I checked it and it fails **as expected** when the CSV version is not similar to the version in UI.

Regarding the `.should('match,'/\d/)`, we need to wait because otherwise, the value is not propagated and will be left as literal 'Unknown.' Waiting for the value of the version to include any number is required before verifying it is equal to the CSV version.

frontend/src/__tests__/cypress/cypress/tests/e2e/applications/testAboutDialog.cy.ts
manosnoam: Yes, this is exactly it. Verifying that every version in the UI table matches the actual version of the corresponding component k8s resource.

frontend/src/__tests__/cypress/cypress.config.ts
```diff
@@ -21,6 +21,7 @@ const resultsDir = `${env.CY_RESULTS_DIR || 'results'}/${env.CY_MOCK ? 'mocked'
 
 export default defineConfig({
   experimentalMemoryManagement: true,
+  watchForFileChanges: false,
```
christianvogt, line 24: We want to keep watching for file changes when using `open` mode. By default `run` mode does not watch for file changes, while `open` mode does.

frontend/src/__tests__/cypress/cypress/pages/aboutDialog.ts
```diff
@@ -19,6 +19,10 @@ export class AboutDialog {
     return cy.findByTestId('about-text');
   }
 
+  findImageByAltText(altText: string): Chainable<JQuery<HTMLElement>> {
+    return cy.get(`img[alt*="${altText}"]`);
+  }
```
christianvogt, line 24: should use `cy.findByAltText`.

manosnoam: Why not by Image too ?
Shall I add it to a more common lib ?

frontend/src/__tests__/cypress/cypress.config.ts
manosnoam: I turned it off since it consumes a lot of memory when just saving tiny changes, which we don't always need to verify on the spot.

christianvogt: `cypress open` watches by default while `cypress run` does not. So this only affects locally execution with `open`. 

You can parameterize this if you want for when running e2e locally. But `open` is used typically during development which is why we want to listen for changes so that the test can reload.

frontend/src/__tests__/cypress/cypress/pages/aboutDialog.ts
christianvogt: I suppose if you want to specifically ensure that this is an `img` tag then this can stay. I don't know of another way to assert the element is an image.

BTW using `cy.get` has the entire page as the context. We should be searching just in the modal itself or a sub section of the modal.

FedeAlonso:
As we disussed yesterday, there are some variables missing from the `test-variables.yml.example`:
```
PRODUCT: RHODS
ODH_DASHBOARD_PROJECT_NAME: Red Hat OpenShift AI
MONITORING_NAMESPACE: redhat-ods-monitoring
OPERATOR_NAMESPACE: redhat-ods-operator
NOTEBOOKS_NAMESPACE: rhods-notebooks
```

frontend/src/__tests__/cypress/cypress.config.ts
manosnoam: ack. I changed it into an optional env var.
Now it will disable watching only if env variable `CY_WATCH=false`, otherwise it is true.

christianvogt: @manosnoam that won't work because right now the default can be `true` or `false` depending on whether the mode is `open` or `run`.
When we are running mock tests in CI, watching is turned off while when we are running locally in open, watching is on.

If you want to control it via env variable, you need to omit the optional altogether if the env variable isn't supplied or otherwise check the mode and set the default accordingly.

manosnoam:
Can you give it a try?

I tested it with false and true, and also ommited, and only when explicitly
setting to false, it did not watch for ongoing changes.

frontend/src/__tests__/cypress/cypress.config.ts
christianvogt: @manosnoam I never tried running cypress `run` mode with `watchForFileChanges === true`. It's quite possible their documentation is wrong and `run` mode never considers `watchForFileChanges`. if that's the case then this change will be fine to include.

manosnoam: Here's the tests I tried:

- **Not passing the env at runtime**: env.CY_WATCH will be undefined, and undefined !== 'false' will be true. So, watchForFileChanges will be set to **true**.
- Passing the env at runtime as **CY_WATCH=false**: env.CY_WATCH will be 'false', and 'false' !== 'false' will be false. So, watchForFileChanges will be set to **false**.
- Passing the env at runtime as **CY_WATCH=true**: env.CY_WATCH will be 'true', and 'true' !== 'false' will be true. So, watchForFileChanges will be set to **true**.
- Passing the env at runtime as CY_WATCH=bla: env.CY_WATCH will be 'bla', and 'bla' !== 'false' will be true. So, watchForFileChanges will be set to **true**.

frontend/src/__tests__/cypress/cypress/tests/e2e/applications/testAboutDialog.cy.ts
```diff
@@ -0,0 +1,101 @@
...
+      aboutDialog.show();
+
+      cy.step(`Verify product '${productName}' in About dialog`);
+      aboutDialog
+        .findText()
+        .invoke('text')
+        .should((text) => {
+          const simplifiedText = text.replace(/[^\w\s]/g, '');
+          expect(simplifiedText).to.contain(productName);
+        });
```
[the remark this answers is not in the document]
manosnoam: Done. Also added comment: 
`// Remove special characters from the text before comparison`
Since for RHOAI we're adding trademarks icon.

frontend/src/__tests__/cypress/cypress/tests/e2e/applications/explore/testEnabledISVs.cy.ts
manosnoam: renamed to APPLICATIONS_NAMESPACE as in test variables.

frontend/src/__tests__/cypress/cypress/pages/aboutDialog.ts
manosnoam: Yes, it's specificly for the image. For RHOAI and ODH the image alt is according to product name.

frontend/src/__tests__/cypress/cypress/tests/e2e/applications/testAboutDialog.cy.ts
```diff
@@ -0,0 +1,101 @@
...
+        .findTable()
+        .should('exist')
+        .then(() => {
+          Object.entries(dataScienceStackComponentMap).forEach(([, component]) => {
+            cy.step(`Verify versions in About dialog's table for component: ${component}`);
+            cy.wrap(null)
+              .then(() => {
+                return aboutDialog.getComponentReleasesText(component);
+              })
+              .then((texts) => {
```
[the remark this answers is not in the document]
manosnoam: Done

frontend/src/__tests__/cypress/cypress/tests/e2e/applications/testAboutDialog.cy.ts
```diff
@@ -0,0 +1,101 @@
...
+      ).then((channel) => {
+        aboutDialog.findChannel().should('contain.text', channel);
+      });
+
+      cy.step(`Verify ${HTPASSWD_CLUSTER_ADMIN_USER.USERNAME} access level`);
+      aboutDialog.isAdminAccessLevel();
+
+      aboutDialog
+        .findTable()
+        .should('exist')
```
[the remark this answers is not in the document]
manosnoam: Done

frontend/src/__tests__/cypress/cypress/tests/e2e/applications/testAboutDialog.cy.ts
manosnoam: Removed `appChrome.visit();` - as you and Fede noted it's only for Mock

frontend/src/__tests__/cypress/cypress.config.ts
christianvogt: I just want to make sure that when we use `run` mode that it doesn't make our CI checks any longer because `watchForFileChanges` is set to `true`.

manosnoam: I see. In that case I added an extra safety:
`watchForFileChanges: env.CY_WATCH ? env.CY_WATCH !== 'false' : undefined`
I verified that it keeps the default (false for run, true for open), unless explicitly setting `CY_WATCH` for cypress open

manosnoam:
> Thanks for this PR @manosnoam, I've added some comments.
> 
> Additionally, can you also execute the test against both RHOAI and ODH?

Good point, I had to make few changes for ODH too, and verified on multiple clusters:

image

frontend/src/__tests__/cypress/cypress.config.ts
antowaddle: Do we still need to change this @manosnoam ?

FedeAlonso (approved):

frontend/src/__tests__/cypress/cypress.config.ts
manosnoam: I believe so, yes. It's a tiny change, which is quite usefull to debug E2E. Verified many times.

PR #4042, opened 2025-04-10
Add Model Catalog section to homepage

nananosirova:
https://issues.redhat.com/browse/RHOAIENG-22047
## Description

- Added a new model catalog section on the homepage behind the feature flag `disableModelCatalog`
- Added a dismissible card in front of the model catalog cards in the new section
- Ensured a limited number of cards (4) are displayed as per design while taking hint width into account
- Added a footer displaying the model count and link to the model catalog page
- Loading spinner on the whole section (including header and footer) while data is fetching
- Hide section entirely if no available models
- Mocked cypress tests, checked responsiveness across breakpoints

Notes: 
- Used alternative image as opposed to the main one because the latter had rendering issues in Firefox 

- Responsiveness
Screen Recording 2025-04-14 at 10 35 09 AM (1)

- Footer link and hint closing
Screen Recording 2025-04-14 at 10 35 46 AM

## How Has This Been Tested?

- Tested locally in Chrome and Firefox, in light/dark modes

## Test Impact

Tested responsiveness, links, card display, mocked cypress tests

## Request review criteria:

If you have UI changes: 


frontend/src/pages/home/modelCatalog/ModelCatalogSection.tsx
```diff
@@ -0,0 +1,124 @@
...
+import { MAX_SHOWN_MODELS, MIN_CARD_WIDTH } from '~/concepts/modelCatalog/const';
+
+const ModelCatalogSection: React.FC = conditionalArea(
+  SupportedArea.MODEL_CATALOG,
+  true,
+)(() => {
+  const { modelCatalogSources } = React.useContext(ModelCatalogContext);
+  const { data, loaded } = modelCatalogSources;
+
+  const models = React.useMemo(
```
DaoDaoNoCode, line 23: There is no need to use `useMemo` here because the `models` are not used as any dependencies. So there is no need to memoize it.

frontend/src/pages/home/modelCatalog/ModelCatalogSection.tsx
```diff
@@ -0,0 +1,124 @@
...
+      data.flatMap((sourceModels: ModelCatalogSource) =>
+        sourceModels.models.map((vals: CatalogModel) => ({ source: sourceModels.source, ...vals })),
+      ),
+    [data],
+  );
+
+  const [visibleCardCount, setVisibleCardCount] = React.useState<number>(MAX_SHOWN_MODELS);
+  const numCards = Math.min(models.length, visibleCardCount);
+
+  const shownModels = React.useMemo(
```
DaoDaoNoCode, line 34: Same here, there is no need to use `useMemo` here.

frontend/src/pages/home/modelCatalog/ModelCatalogSectionFooter.tsx
```diff
@@ -0,0 +1,42 @@
...
+            <Content component="small">
+              {shownModelCount < totalModelCount
+                ? `Showing ${shownModelCount} of all models`
+                : 'Showing all models'}
+            </Content>
+          </Content>
+        ) : null}
+      </FlexItem>
+      <FlexItem>
+        <Button
```
DaoDaoNoCode, line 29: Use `Link` from `react-router-dom` here instead of `Button` component to make it an anchor. The link button should only be used when we want to use the link styles but want it to be a `<button>` instead of an anchor.

frontend/src/pages/home/modelCatalog/ModelCatalogHint.tsx
```diff
@@ -0,0 +1,73 @@
...
+  isHidden,
+  setHidden,
+}) => {
+  if (isHidden) {
+    return null;
+  }
+
+  const hintActions = (
+    <Button
+      icon={<TimesIcon style={{ fontSize: '0.75em' }} />}
```
rsun19, line 24: @DaoDaoNoCode is styling necessary here? If so, my pr would need to change too...

frontend/src/pages/home/modelCatalog/ModelCatalogHint.tsx
```diff
@@ -0,0 +1,73 @@
...
+    marginLeft: '-2em',
+  };
+
+  const cardBodyStyle: React.CSSProperties = {
+    display: 'flex',
+    alignItems: 'center',
+    paddingLeft: 0,
+    paddingTop: 0,
+    paddingRight: '4em',
+  };
```
rsun19, line 47: I wonder if it's possible to do this section without all of these custom styles, and use `split` and `flex` instead. I will try to do this and report back 😂

frontend/src/pages/home/modelCatalog/ModelCatalogHint.tsx
DaoDaoNoCode: Hmm, not really. @nananosirova Let's remove the `fontSize` here. We should try to avoid using custom styles unless it's necessary. The mockup cannot be pixel-perfect so we should use what PF provides us as much as possible and follow their guidelines. If there are some cases when you need to write a bunch of custom styles, we should contact the UX designers first to check the best solution with them.

frontend/src/pages/home/modelCatalog/ModelCatalogSectionFooter.tsx
```diff
@@ -0,0 +1,33 @@
...
+}> = ({ shownModelCount, totalModelCount }) => (
+  <Flex
+    gap={{ default: 'gapMd' }}
+    alignItems={{ default: 'alignItemsCenter' }}
+    data-testid="model-catalog-footer"
+  >
+    <FlexItem>
+      {shownModelCount ? (
+        <Content>
+          <Content component="small">
```
DaoDaoNoCode, line 17: Duplicated `Content` component here.

frontend/src/pages/home/modelCatalog/ModelCatalogHint.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+  const cardBodyStyle: React.CSSProperties = {
+    paddingLeft: 0,
+    paddingTop: 0,
+    paddingRight: '4em',
+  };
+
+  return (
+    <Card style={{ height: '100%' }} data-testid="model-catalog-hint">
```
DaoDaoNoCode, line 48: Small advice: We could check the PF docs to see if there is already an available prop to use before adding custom styles.
```suggestion
    <Card isFullHeight data-testid="model-catalog-hint">
```

frontend/src/pages/home/modelCatalog/ModelCatalogHint.tsx
```diff
@@ -0,0 +1,71 @@
...
+    backgroundImage: `url(${modelCatalogImg})`,
+    backgroundPosition: 'left',
+    marginLeft: '-2em',
+  };
+
+  const cardBodyStyle: React.CSSProperties = {
+    paddingLeft: 0,
+    paddingTop: 0,
+    paddingRight: '4em',
+  };
```
DaoDaoNoCode, line 45: For consistency, I would suggest you use inline styles directly instead of creating a variable here unless the styles need to be reused by another component. In that way, we might create a CSS class.

DaoDaoNoCode (commented):
I suggest removing some of the styles here. Reasons:
1. Some of the styles make no difference on the layout at all after getting rid of them, the PF layout component probably already covered that, or the default (inherit) styles are the same as your settings. In this way, removing them can keep these components clean. It could be hard to debug the layout bugs in the future if we add too many custom styles
2. For the close button on the card component, let's just use what PF provides us. I know you want to make it pixel-perfect and look exactly the same as the mockups. However, the design cannot always perfectly match when we use PF to build the components. If that happens, we should just adopt what PF provides us because the PF team is responsible for the styles and we shouldn't need to add any custom styles unless it's necessary. That said, as long as the styles look alright, we don't need to add any extra custom styles to make them perfectly match. If you have any questions about the styles, feel free to ask in the channel or ask the designer. If it's a PF issue, raise a question in  **#list-patternfly** channel

PR #4044, opened 2025-04-10
Migrate accelerator profile apis

rsun19:
https://issues.redhat.com/browse/RHOAIENG-23069

## Description

Migrated the create, update, and delete apis to the frontend.

## How Has This Been Tested?

Go to the accelerator profiles page and make sure everything works first. Then,

Tested on Ilab cluster. I will keep the role and rolebindings if you want to test there.

Basically, you need to create a new role with the following permissions:

```
rules:
  - verbs:
      - create
      - get
      - list
      - watch
      - patch
      - update
      - delete
    apiGroups:
      - dashboard.opendatahub.io
    resources:
      - acceleratorprofiles
```

Then, bind it to the `admin user`. 

### NOTE: You do not need to do the above steps I just mentioned if you are using the ilab cluster. It is already created and applied

Next, you need to add both the `admin user` and the `regular user` to the `odh-admins` group to gain access to the accelerator profiles screen. 

After, you need to spin down the operator, and remove the storage class permissions from the `admingroup-role` because it is binded to `odh-admins` group.

**Note that in the ilab cluster, you can spin this down by going to `Deployments` -> `opendatahub-operator-controller-manager`**

Now test with the `admin user` and `regular user`. It should work. 

## Test Impact

Added unit tests

## Request review criteria:

If you have UI changes: 


frontend/src/pages/acceleratorProfiles/screens/manage/ManageAcceleratorProfileFooter.tsx
```diff
@@ -33,34 +31,31 @@ export const ManageAcceleratorProfileFooter: React.FC<ManageAcceleratorProfileFo
...
+                </Button>
+              </StackItem>
+            </Stack>,
+          );
         }
+        navigate(redirectPath);
       })
       .catch((err) => {
-        setErrorMessage(err.message);
+        setErrorMessage(err.message || 'Could not create accelerator profile');
```
DaoDaoNoCode, line 58: There should always be an error message for the k8s call, so we can remove the content after `||`. Same below.

DaoDaoNoCode (commented):
Left one small nit, the other parts lgtm

rsun19:
Addressed comments

Gkrumbach07 (approved):
/lgtm

this looks good. perms should be good by now

PR #4051, opened 2025-04-14
Remove superfluous secureRoute usages

andrewballantyne:
https://issues.redhat.com/browse/RHOAIENG-16091

## Description

Cleanup of superfluous secureRoute usages -- the only use-case for this utility is to help secure jupyter tile endpoints... other usages are invalid.

## How Has This Been Tested?

With the nodejs backend -- locally via `:ext` or hosting an image. 

Testing the corresponding areas of the app that match for the routes modified.

## Test Impact

N/A

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/backend/src/routes/api/connection-types/index.ts
+++ b/backend/src/routes/api/connection-types/index.ts
@@ -1,7 +1,7 @@
 import { V1ConfigMap } from '@kubernetes/client-node';
 import { FastifyReply, FastifyRequest } from 'fastify';
 import { KubeFastifyInstance, RecursivePartial } from '../../../types';
-import { secureAdminRoute, secureRoute } from '../../../utils/route-security';
+import { secureAdminRoute } from '../../../utils/route-security';
 import {
   getConnectionType,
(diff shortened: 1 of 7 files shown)


andrewballantyne:
> I've just reviewed it and test it locally, one quick question @andrewballantyne shouldn't `secureRoute` be there so non-rhods users couldn't access to certain resources? All of the endpoints seem safe, just want to double check since it's been a while since I don't check this part of the dashboard.

Nope, never was the intent of the `secureRoute` endpoint -- its there to product Jupyter tile because of the weird nature it exists under. https://issues.redhat.com/browse/RHOAIENG-22788 will make that clearer with a code refactor and endpoints being under `api/jupyter/*` and the `secureRoute` next to it

Griffin-Sullivan (approved):
Tested as a cluster-admin while running backend and frontend locally. You can hit most of these endpoints by going to the cluster settings page. Then you just need to go to the home page for /docs and connection type settings for /connection-types. Everything works as expected!

PR #4056, opened 2025-04-14
Update default connections radio

ashley-o0o:
Closes: [RHOAIENG-22646](https://issues.redhat.com/browse/RHOAIENG-22646)

## Description
This update changes the default behavior for the connection radio buttons when deploying a model. Before, it defaults to existing connection, even if there aren’t any connections to choose from. Now, if there are no connections available, the option won’t show at all and you will only have to option the create a new connection. If there are connections available, it will continue to default to existing. 

## How Has This Been Tested?
Tested locally

- Navigate to your data science projects, and to the Models tab. 
- Ensure that you have no existing connections, then click deploy model. 
- The existing radio option shouldn't be visible, and you should only see the option for Connection type.
- If you go back and create a connection and then return to deploy model, existing connection radio will be selected by default, and you will still be able to create a new connection.

## Test Impact
No tests changed

## Screenshots 
Before:
Screenshot 2025-04-14 at 11 18 44 AM

After: 
Screenshot 2025-04-15 at 1 50 34 PM

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/__tests__/cypress/cypress/tests/mocked/modelRegistry/modelVersion/modelVersionDeploy.cy.ts
+++ b/frontend/src/__tests__/cypress/cypress/tests/mocked/modelRegistry/modelVersion/modelVersionDeploy.cy.ts
@@ -266,7 +266,6 @@ describe('Deploy model version', () => {
     cy.findByText('The format of the source model is').should('not.exist');
 
     // Validate connection section
-    kserveModal.findNewConnectionOption().should('be.checked');
     kserveModal.findModelURITextBox().should('have.value', 'test.io/test/private:test');
     kserveModal
       .findConnectionFieldInput('ACCESS_TYPE')
--- a/frontend/src/__tests__/cypress/cypress/tests/mocked/modelServing/runtime/servingRuntimeList.cy.ts
+++ b/frontend/src/__tests__/cypress/cypress/tests/mocked/modelServing/runtime/servingRuntimeList.cy.ts
@@ -1082,7 +1082,7 @@ describe('Serving Runtime List', () => {
       });
     });
 
-    it('Successfully submit KServe Modal on edit', () => {
+    it.only('Successfully submit KServe Modal on edit', () => {
       initIntercepts({
         projectEnableModelMesh: false,
         disableKServeConfig: true,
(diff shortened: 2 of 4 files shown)


ashley-o0o:
@kywalker-rh @simrandhaliw , Thoughts on this UI change? Basically disallowing users from seeing the existing connections option if there aren't any available

Griffin-Sullivan (commented):
It seems like just using the NewConnectionField on it's own doesn't work out of the box. We'll need to make some changes. In my testing if you try to deploy a model with a new connection in a project with no connections then it doesn't work.

The ISVC is made with blank fields for the connections. There's probably something we need to set on the create data to make it so the connection is actually created and used in the deployment

simrandhaliw:
I like this change—I think I may have even advocated for it in the past. That said, there’s a difference between hiding the 'Existing connections' radio option when there are truly no existing connections in the user’s project, versus hiding it when there are connections but none meet a specific constraint (like in model serving, where we only show compatible ones). If we hide the radio option in the latter case, we also end up hiding the reason why no connections are being displayed, which could be confusing.

So maybe what you’re showing in the screencaps—i.e. hiding the 'Existing connections' option—could work well for the first case. But for the second case I mentioned, maybe we could instead automatically select the 'Create connections' radio option to streamline things for the user, while still keeping the 'Existing connections' option visible. That way, we guide the user but also make it clear why no existing connections are available.

@kywalker-rh thoughts?

Griffin-Sullivan:
@ashley-o0o now that @simrandhaliw mentions it I think our logic will break if a user has a connection in the project that is not model serving compatible. It will probably show the existing connection like before with no options to select anything

kywalker-rh:
@simrandhaliw I like your suggestion!

If something exists but is incompatible and hidden it will confuse users, so we should definitely show any connections if they exist. If they are incompatible with a certain action they should be disabled and we should guide the user to the action that helps them move forward, just as Sim suggests. Thanks for working on this!

cc: @ashley-o0o

Griffin-Sullivan:
@kywalker-rh @simrandhaliw Showing non-serving compatible connections in the serving modal, even if disabled, would be an anti-pattern to what we've done before. There's quite a bit of logic we use to filter connections based on serving compatibility (we even show this information in the connections tab under projects). If the concern is that the user won't understand why they have to make a connection because they already have one in their project, then maybe we should add an alert that says there were no serving-compatible connections found and they need to follow the form to create one.

@andrewballantyne @emilys314 thoughts?

Griffin-Sullivan (approved):
Tested and everything works for me!

PR #4060, opened 2025-04-14
Update ModelCatalogCard component with a truncated option

nananosirova:
https://issues.redhat.com/browse/RHOAIENG-22048

## Description

Extended the existing ModelCatalogCard component for the truncated scenario:
- Model name: middle is truncated after 15 characters, tooltips for accessing the full name
- Model description: truncate if longer than 2 lines, tooltips for accessing the full paragraph
- Model labels: only display first two labels 

modelcatalogcard2

## How Has This Been Tested?
- Tested manually with different values on the model catalog page
- Making sure the current usage of ModelCatalogCard by existing components isn't disrupted by the changes and existing model catalog tests work.

## Test Impact

- The truncated state for ModelCatalogCard will be used in the homepage model catalog section and thus will be thoroughly tested once it's integrated with the section along with the featured models in [RHOAIENG-22050](https://issues.redhat.com/browse/RHOAIENG-22050).

## Request review criteria:

If you have UI changes: 


frontend/src/concepts/modelCatalog/content/ModelCatalogCard.tsx
```diff
@@ -0,0 +1,112 @@
...
+              {truncateNameEnabled ? (
+                <Tooltip content={model.name} position="top">
+                  <span>{`${model.name.substring(
+                    0,
+                    TRUNCATED_LEFT_EDGE_LENGTH,
+                  )}${ELLIPSIS}${model.name.slice(-TRUNCATED_RIGHT_EDGE_LENGTH)}`}</span>
+                </Tooltip>
+              ) : (
+                <span>{model.name}</span>
+              )}
```
DaoDaoNoCode, line 85: Consider using the `Truncate` component [from the middle](https://www.patternfly.org/components/truncate#middle) instead of creating the truncate by yourself. We need to avoid building wheels if there is an available PF component.
Reasons:
1. We did some extra work that PF has already done
2. This might not align with the PF design guidelines
3. If there are future updates on the component, we will possibly miss them. And that increases the workload to maintain it

frontend/src/concepts/modelCatalog/content/ModelCatalogCard.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+const TRUNCATED_LEFT_EDGE_LENGTH = 15;
+const TRUNCATED_RIGHT_EDGE_LENGTH = Math.floor(TRUNCATED_LEFT_EDGE_LENGTH / 2);
+const ELLIPSIS = '...';
+const MAX_LENGTH = TRUNCATED_LEFT_EDGE_LENGTH + ELLIPSIS.length + TRUNCATED_RIGHT_EDGE_LENGTH;
+
+export const ModelCatalogCard: React.FC<ModelCatalogCardProps> = ({
+  model,
+  source,
+  truncate = false,
```
DaoDaoNoCode, line 41: I don't see this prop is used for now, is it saved for future use?

frontend/src/concepts/modelCatalog/content/ModelCatalogCard.tsx
```diff
@@ -1,31 +1,34 @@
...
+import TruncatedText from '~/components/TruncatedText';
 
-export const ModelCatalogCard: React.FC<{ model: CatalogModel; source: string }> = ({
-  model,
-  source,
-}) => (
+export const ModelCatalogCard: React.FC<{
+  model: CatalogModel;
+  source: string;
+  truncate?: boolean;
```
DaoDaoNoCode, line 30: The `truncate` prop is unnecessary here IMO, we should always use `Truncate` for the name and use `TruncatedText` for the description no matter whether the title and description are overflown or not. They have logic to control the truncate behavior and we don't need to worry about that.

rsun19: Are we also going to be truncating the names for the model catalog section? Because I think this prop here depends on what page the user is on.

frontend/src/concepts/modelCatalog/content/ModelCatalogLabels.tsx
```diff
@@ -3,14 +3,27 @@ import { Label, LabelGroup } from '@patternfly/react-core';
...
+
+type ModelCatalogLabelsProps = {
   labels?: string[];
   tasks?: string[];
   showNonILabLabels?: boolean;
-}> = ({ labels = [], tasks = [], showNonILabLabels = false }) => (
-  <LabelGroup data-testid="model-catalog-label-group">
-    {getILabLabels(labels).map((l) => {
-      switch (l) {
+  truncate?: boolean;
```
DaoDaoNoCode, line 12: Instead of passing the `truncate` and set it as `TRUNCATED_MAX_LABEL_NUM` in the component, why not just pass a prop named `numLabels`? Then you can pass `numLabels={TRUNCATED_MAX_LABEL_NUM}` when using it and don't need to pass anything if you don't want the labels to be truncated. Also, it provides more flexibility because we can define the truncate number now for future use

DaoDaoNoCode (approved):
/lgtm

PR #4061, opened 2025-04-15
Fix (2.16.2): consistency of NIM status between explore and enabled pages  + Enable button spinner for non-admin user

mtalvi:
Related to the following:
https://issues.redhat.com/browse/NVPE-229
https://issues.redhat.com/browse/NVPE-233
## Description

In OpenShift AI 2.16, when enabling NVIDIA NIM from the Explore page, the app appeared to be enabled immediately without actually waiting for the real backend response. Not only that, but in the Enabled page it didn't show up or was shown as disabled. After my fix, NIM now appears on the Enabled page only after it is truly enabled, and users no longer get false-positive UI states. In addition, the "Enble" button disappears after enabling. In addition, in case the user is a non-admin user, there won't be an option for the user to enable NIM. (There won't be an "Enable" button)
This created an inconsistent and confusing user experience.
## How Has This Been Tested?

Tested locally.
## Test Impact

1. Go to the Explore page and choose NIM.
2. Click the Enable button.
3. You should get success message and the button should disappear if key is valid.
4. Go to Enabled page and make sure NIM is there as enabled.

Regarding the second issue:
1. Log in with a non-admin user.
2. Go to the Explore page and choose NIM.
3. Enable button should not be seen.

https://github.com/user-attachments/assets/efff5e1d-4d97-4123-b7bb-5f0b3fba6b02

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/pages/exploreApplication/EnableModal.tsx
+++ b/frontend/src/pages/exploreApplication/EnableModal.tsx
@@ -22,6 +22,17 @@ type EnableModalProps = {
   onClose: () => void;
 };
 
+// Poll /api/integrations/nim to confirm backend enablement
+const fetchNimIntegrationStatus = async (): Promise<boolean> => {
+  try {
+    const res = await fetch('/api/integrations/nim');
+    const data = await res.json();
+    return data?.isEnabled === true;
+  } catch {
+    return false;
+  }
+};
+
 const EnableModal: React.FC<EnableModalProps> = ({ selectedApp, shown, onClose }) => {
   const [postError, setPostError] = React.useState('');
   const [validationInProgress, setValidationInProgress] = React.useState(false);
@@ -33,6 +44,7 @@ const EnableModal: React.FC<EnableModalProps> = ({ selectedApp, shown, onClose }
     enableValues,
     selectedApp.spec.internalRoute,
   );
+
   const focusRef = (element: HTMLElement | null) => {
     if (element) {
       element.focus();
@@ -40,11 +52,10 @@ const EnableModal: React.FC<EnableModalProps> = ({ selectedApp, shown, onClose }
   };
 
   const updateEnableValue = (key: string, value: string): void => {
-    const updatedValues = {
-      ...enableValues,
+    setEnableValues((prev) => ({
+      ...prev,
       [key]: value,
-    };
-    setEnableValues(updatedValues);
+    }));
   };
 
   const onDoEnableApp = () => {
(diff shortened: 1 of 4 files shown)


frontend/src/pages/exploreApplication/EnableModal.tsx
```diff
@@ -22,6 +22,17 @@ type EnableModalProps = {
   onClose: () => void;
 };
 
+// Poll /api/integrations/nim to confirm backend enablement
+const fetchNimIntegrationStatus = async (): Promise<boolean> => {
```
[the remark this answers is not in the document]
mtalvi: I understand what you are saying. Maybe I can generalize the logic by using the internalRoute (from the selectedApp) to determine the backend route dynamically, since that route is already stored in the CR (custom resource) for every integration tile.  It shouldn't affect other apps unless they implement the same pattern.

mtalvi:
I removed all NIM-specific logic and made the code generic by relying on internalRoute from the CR.
I also moved the polling for isEnabled into useIntegratedAppStatus, so that logic is now encapsulated and reusable.
Regarding versions: I'm not sure if there’s a planned 2.16.3 release, but I know a 2.19.1 version is coming. I reached out to mmikhail (on Slack) and asked for a cluster running 2.19 so I can test whether the bugs fixed in this PR occur there as well.

frontend/src/pages/exploreApplication/EnableModal.tsx
```diff
@@ -114,18 +143,18 @@ const EnableModal: React.FC<EnableModalProps> = ({ selectedApp, shown, onClose }
         </Button>,
       ]}
     >
-      {enable.description ? enable.description : null}
-      {enable.link ? (
+      {enable.description && <div>{enable.description}</div>}
```
emilys314, line 146: Very minor nitpick but the following helps reduce clutter
```suggestion
      {enable.description && {enable.description}}
```

frontend/src/pages/exploreApplication/EnableModal.tsx
```diff
@@ -53,31 +54,58 @@ const EnableModal: React.FC<EnableModalProps> = ({ selectedApp, shown, onClose }
...
-      // TODO: Disable rule below temporarily. Refactor to notify the owner and avoid modifying the object directly.
-      /* eslint-disable no-param-reassign */
-      selectedApp.spec.isEnabled = true;
-      selectedApp.spec.shownOnEnabledPage = true;
-      /* eslint-enable no-param-reassign */
+    if (
+      validationInProgress &&
+      validationStatus === EnableApplicationStatus.SUCCESS &&
+      selectedApp.spec.internalRoute
+    ) {
```
emilys314, line 61: I think this would break functionality for non internalRoute apps

So you may need to give a way for them to handle the success too like
```suggestion
    if (
      validationInProgress &&
      validationStatus === EnableApplicationStatus.SUCCESS
    ) {
      if (selectedApp.spec.internalRoute) {...your changes}
      else { ...old code}
```

PR #4067, opened 2025-04-17
Fix resetting issue when editing a legacy hardware profile

DaoDaoNoCode:
JIRA: [RHOAIENG-22859](https://issues.redhat.com/browse/RHOAIENG-22859)

## Description

The problem in the ticket is because the editing page for the legacy hardware profiles keeps resetting to the old value every few seconds. The value could just be reset when the user migrates the hardware profile.
Tracing it back, I found the `existingHardwareProfile` that's passed to the editing page is not reference-stable. Thus it keeps triggering the initial form value setting hook.
Then tracing it back more, I found the `refresh` function in `useApplicationSettings` is not using `React.useCallback`, thus it's not reference stable. However, it's used as the dependency when we create the `migratedHardwareProfiles`.
Make it reference-stable fixes the issue.

## How Has This Been Tested?

1. Go to the hardware profile settings page
2. Edit a legacy hardware profile
3. Change the identifier sizes in the "Resource requests and limits" section
4. Submit and confirm migration
5. Check the hardware profile identifier section after migration, make sure the values are updated

## Test Impact

N/A

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/app/useApplicationSettings.tsx
+++ b/frontend/src/app/useApplicationSettings.tsx
@@ -17,27 +17,30 @@ export const useApplicationSettings = (): {
   const [dashboardConfig, setDashboardConfig] = React.useState<DashboardConfigKind | null>(null);
   const setRefreshMarker = useTimeBasedRefresh();
 
-  const refresh = () =>
-    fetchDashboardConfig(true)
-      .then((config) => {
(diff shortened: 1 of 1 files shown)


Gkrumbach07 (approved):
nice catch

/lgtm

PR #4069, opened 2025-04-17
Remove page style overrides in App.tsx

jenny-s51:
https://issues.redhat.com/browse/RHOAIENG-23580

## Description

Before:
Screenshot 2025-04-10 at 1 22 08 PM

After:
Screenshot 2025-04-17 at 1 58 32 PM

## How Has This Been Tested?

1. Run npm install && npm run dev:start:ext in the frontend folder (after connecting to a cluster)
2. Load up localhost & wait for the home page to show up
3. CTRL+C the previous npm terminal
4. Wait for timeout — see related image

## Test Impact

N/A, no changes to functionality. 

## Request review criteria:

If you have UI changes: 


frontend/src/app/App.tsx
```diff
@@ -127,7 +127,7 @@ const App: React.FC = () => {
               masthead={
                 <Header onNotificationsClick={() => setNotificationsOpen(!notificationsOpen)} />
               }
-              sidebar={isAllowed ? <NavSidebar /> : undefined}
+              sidebar={isAllowed ? <NavSidebar /> : null}
```
andrewballantyne, line 130: Why was this change necessary?

frontend/src/app/App.tsx
```diff
@@ -82,7 +82,7 @@ const App: React.FC = () => {
 
     // Default error handling for other cases
     return (
-      <Page>
+      <Page className="pf-m-no-sidebar">
```
andrewballantyne, line 85: Is there really no way to avoid adding a class name? This seems to be breaking the idea of React PF

frontend/src/concepts/pipelines/content/pipelinesDetails/pipelineRun/PipelineRunDrawerRightContent.tsx
```diff
@@ -37,12 +37,7 @@ const PipelineRunDrawerRightContent: React.FC<PipelineRunDrawerRightContentProps
   return (
     // TODO: Revisit below approach, either to further to look into what caused the content to not render or
     // to restructure the code -- see https://issues.redhat.com/browse/RHOAIENG-23537
-    <div
-      className="pf-v6-c-drawer__panel"
-      data-testid="pipeline-run-drawer-right-content"
-      // TODO: look into removing this inline style; PF Drawers should handle height/scrolling by default
-      style={{ height: '100%', overflowY: 'auto' }}
-    >
+    <div className="pf-v6-c-drawer__panel-main" data-testid="pipeline-run-drawer-right-content">
```
andrewballantyne, line 40: Wasn't this a PF component before? Are we stuck with the `div` + class name?

andrewballantyne (changes requested):

frontend/src/app/App.tsx
jenny-s51: Great question @andrewballantyne - in other use cases, we could achieve this with `sidebar={null}`. However, when the server is shut down, the React component cannot re-render, so in this scenario we do need to apply the classname.

andrewballantyne: Why can't the react cycle re-render? Losing a server connection shouldn't mean we lose React render cycles...

frontend/src/app/App.tsx
jenny-s51: Leaving it `undefined` wouldn't trigger the same styling - when `isAllowed` is false, we want to explicitly set the no sidebar modifier when there is no sidebar, and ensure `pf-m-no-sidebar` is properly applied. 

This aligns with the [expected behavior in the PF Page component](https://github.com/patternfly/patternfly-react/commit/217aad3241e79b02ab8f1f74fd1d8022860a138b#diff-33f100e4656af51b92b0c2540a3ad8066a7384595d7b8d399fdbd448515c46ccR25-R341).

andrewballantyne: Sounds like poor react usage 🤔 

Why is "nothing provided" and "explicitly nothing provided" not treated the same way -- this seems like an unnessary JavaScript-ism... in modern JavaScript `null` and `undefined` have little value when no default is available... I think we should take this back to PF as a bug.

If there was a default Sidebar, `undefined` would assume "use the default" whereas `null` would say "no, I explicitly want nothing -- avoid default" -- but since a default sidebar isn't the case, it seems like a bad configuration on the PF component.

With all that said -- wouldn't this negate the need for the className on the page object?

The goal here isn't to just "do what PF says" -- we have a hack in place already, keep the hack until we get something that doesn't require hacks -- swapping hacks is not worth the change unless you're making a better hack -- and this seems subjective imo.

frontend/src/concepts/pipelines/content/pipelinesDetails/pipelineRun/PipelineRunDrawerRightContent.tsx
jenny-s51: Yes, this was previously a PatternFly component, and we're currently using the div + classname as a workaround. This PR focuses on removing the page style overrides, while the changes to this component are tracked separately here: https://issues.redhat.com/browse/RHOAIENG-23537

I’d be glad to incorporate it into this PR, but it will likely require some refactoring of the Drawer component, which is outside the current scope. We could also consider addressing it as a follow-up. What do you think, @andrewballantyne ?

frontend/src/app/App.tsx
jenny-s51: Thank you for your feedback here @andrewballantyne , really appreciate your clarity on this. That sounds like a good plan - I will revert this change in the meantime, check in with the PF team, and circle back on the implementation.

andrewballantyne: thanks -- sorry for all the noise on your PR 😅

frontend/src/concepts/pipelines/content/pipelinesDetails/pipelineRun/PipelineRunDrawerRightContent.tsx
andrewballantyne: I think if we revert these two, it's better imo -- do that with the other issue

frontend/src/app/App.tsx
christianvogt: From PF :(
```
  /** Sidebar component for a side nav, recommended to be a PageSidebar. If set to null, the page grid layout
   * will render without a sidebar.
   */
  sidebar?: React.ReactNode;
```

frontend/src/app/App.tsx
```diff
@@ -82,7 +82,7 @@ const App: React.FC = () => {
 
     // Default error handling for other cases
     return (
-      <Page>
+      <Page sidebar={null}>
```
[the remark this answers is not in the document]
jenny-s51: Thanks for your insight @andrewballantyne ! You make a really great point and there’s definitely an opportunity for improvement here. 

I'll clarify this behavior with the PF team to see if we can figure out a more intuitive implementation.

frontend/src/concepts/pipelines/content/pipelinesDetails/pipelineRun/PipelineRunDrawerRightContent.tsx
jenny-s51: That sounds like a good idea @andrewballantyne - I'm okay with closing this PR and moving issue 23580 over to 23537 as a sub-task. I’ll confer with the PF enablement team and anyone else interested in these discussions regarding the improvements to the Page implementation and the Drawer structure, and we can aim to include both revisions in one PR.

frontend/src/app/App.tsx
jenny-s51: Thank you for your reviews and feedback @andrewballantyne @christianvogt

jenny-s51:
Closing this the per discussion in above commend thread.

These changes will be handled with https://issues.redhat.com/browse/RHOAIENG-23537 which has been updated to include https://issues.redhat.com/browse/RHOAIENG-23580.

frontend/src/concepts/pipelines/content/pipelinesDetails/pipelineRun/PipelineRunDrawerRightContent.tsx
andrewballantyne: That works for me -- the only real value add here I think is the CSS to explicit `null` call to the sidebar for Page... which I like, but don't like at the same time haha.

Delaying the whole aspect works for me if we can get better results out of PF.

PR #4070, opened 2025-04-18
fix: dynamically select default storage class for PVC creation

antowaddle:
https://issues.redhat.com/browse/RHOAIENG-24129

## Description
- Modified the logic to dynamically fetch and use the cluster's default storage class for PVC creation
- Updated related test cases to work with dynamically selected storage classes
- Ensures PVC creation is more reliable across different cluster configurations

## How Has This Been Tested?
- Tests executed against the e2e Cluster along with the Cloud cluster it was initially failing on
image

## Test Impact
- None this is a test

## How to run?
Go to odh-dashboard/frontend/src/tests/cypress and run the command npx cypress open . This will open the Cypress UI where the workbenches.cy.ts test can be executed.

## Request review criteria:

If you have UI changes: 


antowaddle:
@FedeAlonso can you review this? The change is passing on ODH/RHOAI clusters, however, it's failing due to insufficient resources on Cloud Clusters. I can see the following applied to the Workbench.
`dsp-wb-test-storage - This was the existing PVC that we were using in the test
<intentionally-not-adding-this> - This was the default storage class available in the cluster`

Do we want the test to only select the default storage class available in the cluster rather than currently which adds 2?

frontend/src/__tests__/cypress/cypress/utils/oc_commands/storageClass.ts
```diff
@@ -69,7 +69,7 @@ export const getStorageClassConfig = (scName: string): Cypress.Chainable<Command
  * @returns Result Object of the operation
  */
 export const getOpenshiftDefaultStorageClass = (): Cypress.Chainable<CommandLineResult> => {
```
manosnoam, line 71: If I recall, the default storage class in GCP is ssd-csi, but maybe for UI tests we should avoid the logic of identifying the storage class, and choose whatever is available in the dropbox ?

FedeAlonso: This is used for the backend, so no dropdowns

FedeAlonso (approved):
/lgtm /approved

FedeAlonso (approved):
/lgtm /approved

PR #4072, opened 2025-04-18
Use new Top-level properties instead of Custom properties

YuliaKrimerman:
Closes [RHOAIENG-21062](https://issues.redhat.com/browse/RHOAIENG-21062) 

## Description

We've updated the model registration workflow to use top-level properties instead of custom properties for catalog model information. This change improves code clarity and maintainability by using proper model fields rather than storing this data in custom properties.
Key changes include:
Modified the RegisterCatalogModel component to use top-level properties (modelSourceKind, modelSourceGroup, modelSourceId, modelSourceName)
Updated the registerModel and registerVersion functions to handle these top-level properties
Adapted the API calls to convert these top-level properties back to custom properties since the API still expects them in that format
Added ESLint disable comments to handle snake_case property names required by the API contract
This refactoring provides a cleaner interface for components while maintaining compatibility with the API's expectations.

## How Has This Been Tested?

Verifying the model registration workflow from catalog to registry
Ensuring that models registered from the catalog can be properly viewed in the model registry
Confirming that links between catalog and registry models work correctly
Also attaching screenshot showing that now the Artifact call is using our new Top level properties and version call doesn't  
Screenshot 2025-04-18 at 9 56 23 AM

## Test Impact

Tests were updated to reflect the new data structure for model registration. Specifically:
Unit tests for the registerModel and registerVersion functions
Integration tests covering the end-to-end registration flow
Validation of proper catalog links in the model registry UI

## Request review criteria:

If you have UI changes: 


pnaik1:
Hello @YuliaKrimerman , I am getting this error while registering a model from model catalog, is there something else I need to look at??

Screenshot 2025-04-22 at 12 54 32 PM

frontend/src/pages/modelRegistry/screens/RegisterModel/utils.ts
```diff
@@ -100,29 +170,100 @@ export const registerVersion = async (
...
+        string_value: formData.catalogModelTag,
+      };
+    }
+
+    const artifactData: CreateModelArtifactData = {
       name: `${formData.versionName}`,
-      description: formData.versionDescription,
-      customProperties: {},
+      description: formData.versionDescription || '',
+      customProperties: artifactCustomProperties,
```
ppadti, line 236: I don't think we need to update the customProperties of modelArtifact, right?

ppadti (changes requested):
@YuliaKrimerman even I was having the same problem as Purva mentioned above, may be its because we are passing few extra properties in registerVersion.

YuliaKrimerman:
@ppadti I addressed the comments and fixed the registration failure, please review and I will work on fixing the tests

frontend/src/pages/modelRegistry/screens/RegisterModel/utils.ts
```diff
@@ -100,29 +163,100 @@ export const registerVersion = async (
...
-      // storageKey: 'TODO',
+      // Set top-level pipeline source properties
+      modelSourceGroup: formData.modelSourceGroup,
+      modelSourceId: formData.modelSourceId,
+      modelSourceName: formData.modelSourceName,
+      // Set top-level catalog properties
+      catalogSourceName: formData.catalogSourceName,
+      catalogRepositoryName: formData.catalogRepositoryName,
+      catalogModelName: formData.catalogModelName,
+      catalogModelTag: formData.catalogModelTag,
```
mturley, line 242: These don't need to be separated by pipeline and catalog, we'll simply include the `modelSource*` properties - and make sure to include them all including `modelSourceKind` and `modelSourceClass`, refer to what's added in [the backend PR](https://github.com/kubeflow/model-registry/pull/838/files#diff-76e892700d0c69e6da9e12e5f92028b640e18a02d359d3e7fc4a32854a10dc9d).

If we're bundling up `additionalArtifactProperties` in formData as I mentioned, we can remove all these lines and instead just spread that here:

```ts
...(formData.additionalArtifactProperties || {})
```

frontend/src/pages/modelRegistry/screens/RegisterModel/utils.ts
```diff
@@ -100,29 +163,100 @@ export const registerVersion = async (
...
       uri:
         formData.modelLocationType === ModelLocationType.ObjectStorage
           ? objectStorageFieldsToUri({
               endpoint: formData.modelLocationEndpoint,
               bucket: formData.modelLocationBucket,
               region: formData.modelLocationRegion,
               path: formData.modelLocationPath,
-            }) || '' // We'll only hit this case if required fields are empty strings, so form validation should catch it.
```
mturley, line 122: I think this comment was helpful in explaining why we have the fallback `|| ''` since we usually want to avoid it but there's a valid reason here.

frontend/src/pages/modelRegistry/screens/components/ModelLabels.tsx
```diff
@@ -17,7 +17,7 @@ const ModelLabels: React.FC<ModelLabelsProps> = ({ name, customProperties }) =>
   const [isLabelModalOpen, setIsLabelModalOpen] = React.useState(false);
   const [searchValue, setSearchValue] = React.useState('');
 
-  const allLabels = getLabels(customProperties);
+  const allLabels = getLabels(customProperties || {});
```
mturley, line 20: We can revert this too if we don't make customProperties optional.

frontend/src/pages/modelRegistry/screens/components/ModelLabels.tsx
```diff
@@ -102,7 +102,7 @@ const ModelLabels: React.FC<ModelLabelsProps> = ({ name, customProperties }) =>
     </Modal>
   ) : null;
 
-  if (Object.keys(customProperties).length === 0) {
+  if (Object.keys(customProperties || {}).length === 0) {
```
mturley, line 105: ^

frontend/src/pages/modelRegistry/screens/utils.ts
```diff
@@ -133,7 +146,13 @@ export const filterModelVersions = (
...
           mv.name.toLowerCase().includes(searchLower) ||
           (mv.description && mv.description.toLowerCase().includes(searchLower)) ||
-          getLabels(mv.customProperties).some((label) => label.toLowerCase().includes(searchLower))
+          (mv.customProperties &&
+            getLabels(mv.customProperties).some((label) =>
+              label.toLowerCase().includes(searchLower),
+            )) ||
+          (mv.modelSourceName && mv.modelSourceName.toLowerCase().includes(searchLower)) ||
+          (mv.modelSourceId && mv.modelSourceId.toLowerCase().includes(searchLower)) ||
+          (mv.modelSourceGroup && mv.modelSourceGroup.toLowerCase().includes(searchLower))
```
mturley, line 155: Interesting - it might be a good idea to support searching based on the model source, but since it's only present on the artifact we can't do it here on the version filtering without some additional lookups that will be painful. Best to remove these lines for now.

frontend/src/pages/modelRegistry/screens/utils.ts
```diff
@@ -172,15 +191,25 @@ export const filterRegisteredModels = (
...
           rm.name.toLowerCase().includes(searchLower) ||
           (rm.description && rm.description.toLowerCase().includes(searchLower)) ||
-          getLabels(rm.customProperties).some((label) => label.toLowerCase().includes(searchLower));
+          (rm.customProperties &&
+            getLabels(rm.customProperties).some((label) =>
+              label.toLowerCase().includes(searchLower),
+            )) ||
+          (rm.modelSourceName && rm.modelSourceName.toLowerCase().includes(searchLower)) ||
+          (rm.modelSourceId && rm.modelSourceId.toLowerCase().includes(searchLower)) ||
+          (rm.modelSourceGroup && rm.modelSourceGroup.toLowerCase().includes(searchLower));
```
mturley, line 200: Same here, we can't really filter models by these without looking up artifacts for every version which isn't feasible.

frontend/src/pages/modelRegistry/screens/utils.ts
```diff
@@ -172,15 +191,25 @@ export const filterRegisteredModels = (
...
-            getLabels(mv.customProperties).some((label) =>
-              label.toLowerCase().includes(searchLower),
-            ),
+            (mv.customProperties &&
+              getLabels(mv.customProperties).some((label) =>
+                label.toLowerCase().includes(searchLower),
+              )) ||
+            (mv.modelSourceName && mv.modelSourceName.toLowerCase().includes(searchLower)) ||
+            (mv.modelSourceId && mv.modelSourceId.toLowerCase().includes(searchLower)) ||
+            (mv.modelSourceGroup && mv.modelSourceGroup.toLowerCase().includes(searchLower)),
```
mturley, line 212: Same for versions

frontend/src/pages/modelRegistry/screens/ModelVersions/ModelDetailsView.tsx
```diff
@@ -48,17 +48,17 @@ const ModelDetailsView: React.FC<ModelDetailsViewProps> = ({
             }
           />
           <EditableLabelsDescriptionListGroup
-            labels={getLabels(rm.customProperties)}
+            labels={getLabels(rm.customProperties || {})}
```
mturley, line 51: A few more places I missed where we can remove `|| {}`

frontend/src/pages/modelRegistry/screens/ModelVersions/ModelDetailsView.tsx
```diff
@@ -48,17 +48,17 @@ const ModelDetailsView: React.FC<ModelDetailsViewProps> = ({
...
             isArchive={isArchiveModel}
-            allExistingKeys={Object.keys(rm.customProperties)}
+            allExistingKeys={Object.keys(rm.customProperties || {})}
```
mturley, line 53: ^

frontend/src/pages/modelRegistry/screens/ModelVersions/ModelDetailsView.tsx
```diff
@@ -48,17 +48,17 @@ const ModelDetailsView: React.FC<ModelDetailsViewProps> = ({
...
             title="Labels"
             contentWhenEmpty="No labels"
             onLabelsChange={(editedLabels) =>
               apiState.api
                 .patchRegisteredModel(
                   {},
                   {
-                    customProperties: mergeUpdatedLabels(rm.customProperties, editedLabels),
+                    customProperties: mergeUpdatedLabels(rm.customProperties || {}, editedLabels),
```
mturley, line 61: ^

frontend/src/pages/modelRegistry/screens/ModelVersions/ModelDetailsView.tsx
```diff
@@ -67,7 +67,7 @@ const ModelDetailsView: React.FC<ModelDetailsViewProps> = ({
           />
           <ModelPropertiesDescriptionListGroup
             isArchive={isArchiveModel}
-            customProperties={rm.customProperties}
+            customProperties={rm.customProperties || {}}
```
mturley, line 70: ^

frontend/src/pages/modelRegistry/screens/RegisterModel/useRegisterModelData.ts
```diff
@@ -19,6 +19,13 @@ export type RegistrationCommonFormData = {
   modelLocationURI: string;
   versionCustomProperties?: ModelRegistryCustomProperties;
   modelCustomProperties?: ModelRegistryCustomProperties;
+  modelSourceGroup?: string;
+  modelSourceId?: string;
+  modelSourceName?: string;
+  catalogSourceName?: string;
+  catalogRepositoryName?: string;
+  catalogModelName?: string;
+  catalogModelTag?: string;
```
mturley, line 28: Since we aren't mapping these to editable form fields and they're just being passed along with the form data similar to the `versionCustomProperties` and `modelCustomProperties`, I think we can simply add one property here for the bundle of them. In fact it doesn't even need to be unique to the model source, it's just additional properties for us to slap on the artifact - so you could add:

```ts
additionalArtifactProperties?: Partial<ModelArtifact>;
```

frontend/src/pages/modelRegistry/screens/RegisterModel/useRegisterModelData.ts
```diff
@@ -47,6 +54,13 @@ const registrationCommonFormDataDefaults: RegistrationCommonFormData = {
   modelLocationURI: '',
   modelCustomProperties: {},
   versionCustomProperties: {},
+  modelSourceGroup: '',
+  modelSourceId: '',
+  modelSourceName: '',
+  catalogSourceName: '',
+  catalogRepositoryName: '',
+  catalogModelName: '',
+  catalogModelTag: '',
```
mturley, line 63: We can remove these - anything you add to `RegistrationCommonFormData` is inherited here.

mturley (changes requested):
Apologies for the big review - I think we need to rethink a few things about the approach here.

> Adapted the API calls to convert these top-level properties back to custom properties since the API still expects them in that format  

We want to totally remove the use of customProperties for these links and it's ok to not be backwards compatible with those properties.

frontend/src/concepts/modelRegistry/types.ts
```diff
@@ -90,6 +95,8 @@ export type ModelRegistryBase = {
   createTimeSinceEpoch: string;
   lastUpdateTimeSinceEpoch: string;
   customProperties: ModelRegistryCustomProperties;
+  modelSourceKind?: string;
+  modelSourceClass?: string;
```
YuliaKrimerman, line 99: I know I need to remove those from here as well, will do once push the final commit with tests

frontend/src/concepts/modelRegistry/utils.ts
```diff
@@ -113,3 +114,80 @@ export const filterArchiveModels = (registeredModels: RegisteredModel[]): Regist
...
+ * @param modelSourceName - The name of model source (maps to catalog model name)
+ * @param modelSourceId - The ID of model source (maps to catalog model tag)
+ * @returns CatalogModelDetailsParams object or null if not a catalog source
+ */
+export const modelSourcePropertiesToCatalogParams = (
+  modelSourceKind?: string,
+  modelSourceClass?: string,
+  modelSourceGroup?: string,
+  modelSourceName?: string,
+  modelSourceId?: string,
```
mturley, line 132: To make this easier, maybe we should factor out a type from ModelArtifact for `ModelSourceProperties` that contains these, and then ModelArtifact can inherit them with `&` like it does with ModelRegistryBase. Then you can just have this function take a `ModelSourceProperties` object, and passing the whole ModelArtifact into this function will be valid since it's a superset of that type.

frontend/src/concepts/modelRegistry/utils.ts
```diff
@@ -113,3 +114,80 @@ export const filterArchiveModels = (registeredModels: RegisteredModel[]): Regist
...
+    return null;
+  }
+
+  return {
+    sourceName: modelSourceClass || '',
+    repositoryName: modelSourceGroup || 'default',
+    modelName: modelSourceName || '',
+    tag: modelSourceId || '',
+  };
+};
```
mturley, line 144: If you require the properties you need to be defined up front you don't need the fallbacks to `''` (also `'default'` isn't a valid repository name).

```ts
if (
  modelSourceKind !== ModelSourceKind.CATALOG ||
  !modelSourceClass ||
  !modelSourceGroup ||
  !modelSourceName ||
  !modelSourceId
) {
  return null;
}
```

that'll narrow the types to defined when you return them.

frontend/src/concepts/modelRegistry/utils.ts
```diff
@@ -113,3 +114,80 @@ export const filterArchiveModels = (registeredModels: RegisteredModel[]): Regist
...
+ */
+export const catalogParamsToModelSourceProperties = (
+  params: CatalogModelDetailsParams,
+): {
+  modelSourceKind: ModelSourceKind;
+  modelSourceClass?: string;
+  modelSourceGroup?: string;
+  modelSourceName?: string;
+  modelSourceId?: string;
+} => ({
```
mturley, line 159: You could reuse the `ModelSourceProperties` type here for the return type.

frontend/src/concepts/modelRegistry/utils.ts
```diff
@@ -113,3 +114,80 @@ export const filterArchiveModels = (registeredModels: RegisteredModel[]): Regist
...
+ * @param modelSourceId - The ID of model source (run ID)
+ * @param modelSourceName - The name of model source (run name)
+ * @returns PipelineRunRef object or null if not a DSP source
+ */
+export const modelSourcePropertiesToPipelineRunRef = (
+  modelSourceKind?: string,
+  modelSourceGroup?: string,
+  modelSourceId?: string,
+  modelSourceName?: string,
+): { project: string; runId: string; runName: string } | null => {
```
mturley, line 180: You could take `ModelSourceProperties` for your argument here. We could also make a type like `PipelineRunReference` or something for the returned object here if we're repeating those 3 properties a lot, but that doesn't matter quite as much.

Edit: Ah, we already had that and it was called `PipelineModelCustomProps` - we probably want to bring that back just with a new name.

frontend/src/concepts/modelRegistry/utils.ts
```diff
@@ -113,3 +114,80 @@ export const filterArchiveModels = (registeredModels: RegisteredModel[]): Regist
...
+  if (modelSourceKind !== ModelSourceKind.DSP) {
+    // For backward compatibility, also check if the fields exist without explicit DSP kind
+    if (!modelSourceGroup || !modelSourceId || !modelSourceName) {
+      return null;
+    }
+  }
```
mturley, line 186: Backwards compatibility doesn't make sense here - if these properties exist they'll either exist with the right kind, or we can ignore them. This can just be one `if` that requires all of these, and that'll also narrow the type so you don't have to use `|| ''` below.

```ts
if (
  modelSourceKind !== ModelSourceKind.DSP ||
  !modelSourceGroup ||
  !modelSourceId ||
  !modelSourceName
) {
  return null;
}
```

frontend/src/pages/modelCatalog/screens/RegisterCatalogModel.tsx
```diff
@@ -109,24 +106,19 @@ const RegisterCatalogModel: React.FC = () => {
...
-        }, {});
+      // Use the utility function to convert catalog params to model source properties
+      const sourceProperties = catalogParamsToModelSourceProperties(decodedParams);
+      // Set additional artifact properties in one object
+      setData('additionalArtifactProperties', {
+        ...sourceProperties,
+        modelSourceName: model.name,
+        modelSourceId:
+          model.artifacts?.map((artifact) => artifact.tags && artifact.tags[0])[0] ?? '',
+      });
```
mturley, line 117: I don't think you need to define modelSourceName and modelSourceId again here right? The values you get here will be the same as the ones in `decodedParams` which are already included in `sourceProperties`.

frontend/src/pages/modelRegistry/screens/ModelVersionDetails/ModelVersionDetailsView.tsx
```diff
@@ -233,12 +207,12 @@ const ModelVersionDetailsView: React.FC<ModelVersionDetailsViewProps> = ({
                   <>
                     <DashboardDescriptionListGroup
                       title="URI"
-                      isEmpty={!modelArtifact?.uri}
+                      isEmpty={!modelArtifact || !modelArtifact.uri}
```
mturley, line 210: Not sure why we need any of these changes (from this line to the end of the file)? These sections of the page aren't related to the issue and the use of the `?.` operator being removed here was fine.

frontend/src/pages/modelRegistry/screens/ModelVersionDetails/ModelVersionRegisteredFromLink.tsx
```diff
@@ -0,0 +1,174 @@
...
+import { globalPipelineRunDetailsRoute, getCatalogModelDetailsRoute } from '~/routes';
+import { ModelArtifact } from '~/concepts/modelRegistry/types';
+import {
+  modelSourcePropertiesToCatalogParams,
+  modelSourcePropertiesToPipelineRunRef,
+} from '~/concepts/modelRegistry/utils';
+
+/**
+ * Props for the ModelVersionRegisteredFromLink component
+ */
```
mturley, line 19: Heh, looks like Cursor is going to give us a lot of unnecessary comments... my preference would be to remove this since it's obvious but it doesn't really matter, we may need to let that one go if we're starting to use AI

frontend/src/pages/modelRegistry/screens/ModelVersionDetails/ModelVersionRegisteredFromLink.tsx
```diff
@@ -0,0 +1,174 @@
...
+    modelArtifact.modelSourceName,
+    modelArtifact.modelSourceId,
+  );
+
+  const registeredFromPipelineDetails = modelSourcePropertiesToPipelineRunRef(
+    modelArtifact.modelSourceKind,
+    modelArtifact.modelSourceGroup,
+    modelArtifact.modelSourceId,
+    modelArtifact.modelSourceName,
+  );
```
mturley, line 47: If you change these utilities to use a single argument with a `ModelSourceProperties` type, you can just pass in the `modelArtifact` here since it'll be assignable to that type.

frontend/src/pages/modelRegistry/screens/ModelVersionDetails/ModelVersionRegisteredFromLink.tsx
```diff
@@ -0,0 +1,174 @@
...
+
+  // If no source information is available, don't render anything
+  if (!registeredFromCatalogDetails && !registeredFromPipelineDetails) {
+    return null;
+  }
+
+  // Generate the catalog URL if we have catalog details
+  const catalogModelUrl = registeredFromCatalogDetails
+    ? getCatalogModelDetailsRoute(registeredFromCatalogDetails)
+    : '';
```
mturley, line 57: I think we can move this down into the `if (registeredFromCatalogDetails)` block, that way it doesn't need to fall back to `''`.

frontend/src/pages/modelRegistry/screens/ModelVersionDetails/ModelVersionRegisteredFromLink.tsx
```diff
@@ -0,0 +1,174 @@
...
+
+  // For pipeline source, look up the project
+  const registeredPipelineProject =
+    registeredFromPipelineDetails &&
+    projects.find((p) => p.metadata.name === registeredFromPipelineDetails.project);
```
mturley, line 62: Same here, this can move into the `if (registeredFromPipelineDetails)` block and then we can drop the `registeredFromPipelineDetails &&`. No need to define these outside the if blocks if they aren't used anywhere else.

frontend/src/pages/modelRegistry/screens/ModelVersionDetails/ModelVersionRegisteredFromLink.tsx
```diff
@@ -0,0 +1,174 @@
...
+
+    // If we also have a catalog URL, link to the catalog
+    const renderRunLink = catalogModelUrl ? (
+      <Link
+        to={catalogModelUrl}
+        style={{ fontWeight: 'var(--pf-t--global--font--weight--body--bold)' }}
+        data-testid="catalog-model-link"
+      >
+        {runName}
+      </Link>
```
mturley, line 76: This is held over from the earlier changes and doesn't make sense - this if block only cares about rendering the link to the pipeline, we don't need it to check or render anything related to the catalog. The rendering inside `if (registeredFromPipelineDetails) {` can basically be copy-pasted from what we had before in `ModelVersionPipelineDescription` on main.

frontend/src/pages/modelRegistry/screens/ModelVersionDetails/ModelVersionRegisteredFromLink.tsx
```diff
@@ -0,0 +1,174 @@
...
+        <FlexItem>
+          <b>{renderProject}</b>
+        </FlexItem>
+        {catalogModelUrl && (
+          <FlexItem data-testid="catalog-model-info">
+            <span style={{ fontWeight: 'var(--pf-t--global--font--weight--body--bold)' }}>
+              (from Model Catalog)
+            </span>
+          </FlexItem>
+        )}
```
mturley, line 115: Same here - it doesn't make sense to render details about both pipeline and catalog, a model will be registered from either one or the other.

frontend/src/pages/modelRegistry/screens/ModelVersionDetails/ModelVersionRegisteredFromLink.tsx
```diff
@@ -0,0 +1,174 @@
...
+        {!registeredPipelineProject && !catalogModelUrl && (
```
mturley, line 116: Here too - we don't need to care about `catalogModelUrl`. All this stuff can be how it was in the original `ModelVersionPipelineDescription`.

frontend/src/pages/modelRegistry/screens/ModelVersionDetails/ModelVersionDetailsView.tsx
```diff
@@ -155,21 +141,9 @@ const ModelVersionDetailsView: React.FC<ModelVersionDetailsViewProps> = ({
           >
             <InlineTruncatedClipboardCopy testId="model-version-id" textToCopy={mv.id} />
           </DashboardDescriptionListGroup>
-          {Object.values(pipelineCustomProperties).every((value) => !!value) && (
+          {modelArtifact && (
```
mturley, line 144: We need to not render the DashboardDescriptionListGroup unless we are fully rendering one of the links inside it - with this logic, if the artifact exists but does not have the model source properties we'll get a "Registered from" title with empty contents.

Let's just render the component itself and then conditionally use `DashboardDescriptionListGroup` inside it.

```tsx
{modelArtifact && (
  <ModelVersionRegisteredFromLink modelArtifact={modelArtifact} />
)}
```

frontend/src/pages/modelRegistry/screens/ModelVersionDetails/ModelVersionRegisteredFromLink.tsx
```diff
@@ -0,0 +1,174 @@
...
+      <b>{runName}</b>
+    );
+
+    const renderProject = registeredPipelineProject
+      ? registeredPipelineProject.metadata.annotations?.['openshift.io/display-name'] ??
+        registeredPipelineProject.metadata.name
+      : project;
+
+    return (
+      <Flex
```
mturley, line 94: Based on my comment in ModelVersionDetailsView, let's wrap this returned Flex in `<DashboardDescriptionListGroup title="Registered from">`.

frontend/src/pages/modelRegistry/screens/ModelVersionDetails/ModelVersionRegisteredFromLink.tsx
```diff
@@ -0,0 +1,174 @@
...
+          color: 'var(--pf-global--primary-color--100)',
+          textDecoration: 'underline',
+        }}
+      >
+        <span style={{ fontWeight: 'var(--pf-t--global--font--weight--body--bold)' }}>
+          {modelArtifact.modelSourceName || ''} ({modelArtifact.modelSourceId || ''})
+        </span>{' '}
+        in Model catalog
+      </a>
+    );
```
mturley, line 167: Same here, let's wrap this returned link in `<DashboardDescriptionListGroup title="Registered from">`.

Also, I'm not sure why we're switching to use a raw `<a>` here with inline styles. We can just render this how we do today in ModelVersionDetailsView on main with a Link component. Additionally you're using the `modelArtifact` details directly here, but you have them nicely converted for you in `registeredFromCatalogDetails` so you can just use those properties.

```tsx
return (
  <DashboardDescriptionListGroup title="Registered from">
    <Link to={catalogModelDetailsUrl} data-testid="registered-from-catalog">
      <span style={{ fontWeight: 'var(--pf-t--global--font--weight--body--bold)' }}>
        {registeredFromCatalogDetails.modelName} ({registeredFromCatalogDetails.tag})
      </span>
    </Link>{' '}
    in Model catalog
  </DashboardDescriptionListGroup>
);
```

frontend/src/pages/modelRegistry/screens/ModelVersionDetails/ModelVersionRegisteredFromLink.tsx
```diff
@@ -0,0 +1,174 @@
...
+  }
+
+  // Fallback case - shouldn't reach here given the earlier checks
+  return null;
```
mturley, line 171: Once you change the `if (registeredFromCatalogDetails && catalogModelUrl) {` to `if (registeredFromCatalogDetails) {` and define `catalogModelUrl` inside it, I bet you can remove this line and TS will be happy that all cases are satisfied. (If not it's fine to leave it)

frontend/src/pages/modelRegistry/screens/ModelVersionDetails/const.ts
```diff
@@ -8,14 +8,11 @@ export enum ModelVersionDetailsTabTitle {
...
-export type PipelineModelCustomProps = {
-  project: string;
-  runId: string;
-  runName: string;
+export type ModelVersionPipelineDescriptionProps = {
+  sourceInfo: {
+    project: string;
+    runId: string;
+    runName: string;
+  };
```
mturley, line 16: Ok, we do repeat `project`, `runId` and `runName` a fair amount. We can reuse the type formerly named `PipelineModelCustomProps`, with the new name `PipelineRunReference` or whatever we want to call it.

frontend/src/pages/modelRegistry/screens/RegisterModel/utils.ts
```diff
@@ -100,10 +103,12 @@ export const registerVersion = async (
   }
 
   try {
-    modelArtifact = await apiState.api.createModelArtifactForModelVersion({}, modelVersion.id, {
+    const artifactCustomProperties: ModelRegistryCustomProperties = {};
```
mturley, line 106: You can remove `artifactCustomProperties`, it's not doing anything here

frontend/src/pages/modelRegistry/screens/utils.ts
```diff
@@ -99,22 +100,25 @@ export const getCustomPropString = <
...
+  }
+
+  // Return empty values if no valid catalog information is available
+  return {
+    sourceName: '',
+    repositoryName: '',
+    modelName: '',
+    tag: '',
+  };
+};
```
mturley, line 121: Hmmm.... this function is redundant with your new `modelSourcePropertiesToCatalogParams` and no longer used except in your tests. I think you can remove it

frontend/src/pages/modelRegistry/screens/utils.ts
```diff
@@ -132,12 +136,12 @@ export const filterModelVersions = (
       case SearchType.KEYWORD:
         return (
           mv.name.toLowerCase().includes(searchLower) ||
-          (mv.description && mv.description.toLowerCase().includes(searchLower)) ||
+          mv.description?.toLowerCase().includes(searchLower) ||
```
mturley, line 139: Same deal here, I'm not sure why we're removing the `?.` operator in unrelated code here, let's revert the changes from this line to the end of the file.

mturley (changes requested):
Getting a lot closer, some followups here

frontend/src/concepts/modelRegistry/types.ts
```diff
@@ -92,19 +97,35 @@ export type ModelRegistryBase = {
...
+  runName: string;
 };
 
+export type ModelSourceProperties = {
+  modelSourceKind?: ModelSourceKind;
+  modelSourceClass?: string;
+  modelSourceGroup?: string;
+  modelSourceName?: string;
+  modelSourceId?: string;
+  pipelineRunReference?: PipelineRunReference;
```
mturley, line 112: @YuliaKrimerman this isn't a property in the MR API, PipelineRunReference was just for convenience in passing around project/runId/runName where you already were. We need to remove it here. It doesn't appear to be used anywhere anyway

frontend/src/pages/modelRegistry/screens/ModelVersionDetails/ModelVersionRegisteredFromLink.tsx
```diff
@@ -0,0 +1,134 @@
...
+      );
+    }
+
+    return null;
+  };
+
+  const content = renderContent();
+  if (!content) {
+    return null;
+  }
```
mturley, line 127: It could have been a little simpler to just have a `let content = null;` at the top and have each of the if cases set it instead of returning, and then you wouldn't need to wrap it in a renderContent function. But that's such a tiny nit, please feel free to ignore it, this is fine.

frontend/src/pages/modelRegistry/screens/ModelVersionDetails/const.ts
```diff
@@ -8,14 +10,7 @@ export enum ModelVersionDetailsTabTitle {
...
-  '_registeredFromPipelineProject',
-  '_registeredFromPipelineRunId',
-  '_registeredFromPipelineRunName',
-];
-
-export type PipelineModelCustomProps = {
-  project: string;
-  runId: string;
-  runName: string;
+export type ModelVersionPipelineDescriptionProps = {
```
mturley, line 13: It looks like this isn't used anywhere now, I think it can be removed.

frontend/src/pages/modelRegistry/screens/utils.ts
```diff
@@ -100,21 +101,19 @@ export const getCustomPropString = <
...
+  if (artifact.modelSourceKind === ModelSourceKind.CATALOG && artifact.modelSourceName) {
+    return {
+      sourceName: artifact.modelSourceClass || '',
+      repositoryName: artifact.modelSourceGroup || '',
+      modelName: artifact.modelSourceName,
+      tag: artifact.modelSourceId || '',
+    };
+  }
+  return null;
+};
```
mturley, line 116: I think this function is redundant with your `modelSourcePropertiesToCatalogParams` util, you can still just pass that function an artifact and it'll work because an artifact contains the `ModelSourceProperties` (In typescript terms, ModelArtifact is assignable to ModelSourceProperties). You can get rid of `getCatalogModelDetailsProps` and in the places where it was used, just drop in `modelSourcePropertiesToCatalogParams`.

mturley (changes requested):
A couple of tiny things... this is looking great!!

mturley (approved):
LGTM! Thanks so much for your patience here @YuliaKrimerman

PR #4046, opened 2025-04-11
fix(RHOAIENG-21303): added default pull access type connection, test cases and recommended connection type 

claudialphonse78:
closes [RHOAIENG-21303](https://issues.redhat.com/browse/RHOAIENG-21303)

## Description

Scenarios Included:
1. When Deploying any model registered using an OCI - URI but not containingoci://registry.redhat.io, you will find "Existing connection" with the fields filled. When you click on the connection name dropdown, you should see the "Recommended" label as shown in the screenshot.(You can test this by deploying the version inside "oci-model" model on the "mdas" project - you should see "Existing connection" preselected and fields filled [ui-green](https://console-openshift-console.apps.ui-green.dev.datahub.redhat.com/) cluster)

The recommended label will now be based on whether the connection includes a Pull request selected in the connection type as seen in the screenshot below even though we have another OCI connection present prior to this where even if the OCI connection did not have a Pull type as selected it was recommended.

Screenshot 2025-04-11 at 4 13 09 PM

**OCI connection without Pull request checked**

Screenshot 2025-04-11 at 4 13 28 PM

Fig: **OCI connection without Pull request checked and included in the recommended tag**

Screenshot 2025-04-11 at 9 54 09 PM

https://github.com/user-attachments/assets/caa5a6c6-d171-4a89-9f0f-a714e909b2f7

2. When deploying any model registered using an OCI - URI but not containing oci://registry.redhat.io and not containing any matching connection, you will find "Create connection" with the "Model URI" field filled. (You can test this by deploying the version inside "oci-model" model on the "mturley test" project - you should see "Create connection" preselected and "Model URI" field filled [ui-green](https://console-openshift-console.apps.ui-green.dev.datahub.redhat.com/) cluster)

Now the pull connection will be preselected by default.

https://github.com/user-attachments/assets/dd304aba-2b8b-4d52-8618-e22ae1896138

## How Has This Been Tested?

1. Test case for handling the preselection of the pull request on OCI create connection

## Test Impact

Cypress tests added.

## Request review criteria:

Self checklist (all need to be checked):
- [x] The developer has manually tested the changes and verified that the changes work
- [x] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [x] The developer has added tests or explained why testing cannot be added (unit or cypress tests for related changes)

If you have UI changes: 

- [x] Included any necessary screenshots or gifs if it was a UI change.
- [ ] Included tags to the UX team if it was a UI/UX change.

[...]

PR #4077, opened 2025-04-21
Load data before rendering Deploy form, when deploying from model catalog or registry

ppadti:
Closes: [RHOAIENG-23292](https://issues.redhat.com/browse/RHOAIENG-23292)

## Description
This PR aims to load the data fully before rendering deploy form, when deploying a model from model catalog and model registry.
1. Model catalog

https://github.com/user-attachments/assets/370cd733-5983-492f-9b8f-2403dd4a6817

2. Model Registry with kServe

https://github.com/user-attachments/assets/c6ccd7c6-7c91-4e30-b357-a5059e410e9e

3. Model registry with model mesh

https://github.com/user-attachments/assets/d746b286-fb38-4b55-9ca8-2a3b9a701135

## How Has This Been Tested?
1. deploy a model from model catalog, make sure spinner is displayed and data is loaded before the form renders.
2. Same goes for model registry.(check for both single and multi model platform)
3. Also, check for model deployment page  - this should work as expected.

## Test Impact
Added cypress test

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/concepts/modelCatalog/content/ModelCatalogLabels.tsx
+++ b/frontend/src/concepts/modelCatalog/content/ModelCatalogLabels.tsx
@@ -14,19 +14,19 @@ export const ModelCatalogLabels: React.FC<{
       switch (label) {
         case ReservedILabLabel.LabBase:
           return (
-            <Label data-testid="model-catalog-label" color="yellow" variant="filled">
+            <Label key={l} data-testid="model-catalog-label" color="yellow" variant="filled">
               LAB starter
             </Label>
           );
         case ReservedILabLabel.LabTeacher:
           return (
-            <Label data-testid="model-catalog-label" color="purple" variant="filled">
+            <Label key={l} data-testid="model-catalog-label" color="purple" variant="filled">
               LAB teacher
             </Label>
           );
         case ReservedILabLabel.LabJudge:
           return (
-            <Label data-testid="model-catalog-label" color="orange" variant="filled">
+            <Label key={l} data-testid="model-catalog-label" color="orange" variant="filled">
               LAB judge
             </Label>
           );
--- a/frontend/src/concepts/modelCatalog/content/ModelCatalogLabels.tsx
+++ b/frontend/src/concepts/modelCatalog/content/ModelCatalogLabels.tsx
@@ -14,19 +14,19 @@ export const ModelCatalogLabels: React.FC<{
       switch (label) {
         case ReservedILabLabel.LabBase:
           return (
-            <Label key={l} data-testid="model-catalog-label" color="yellow" variant="filled">
+            <Label key={label} data-testid="model-catalog-label" color="yellow" variant="filled">
               LAB starter
             </Label>
           );
         case ReservedILabLabel.LabTeacher:
           return (
-            <Label key={l} data-testid="model-catalog-label" color="purple" variant="filled">
+            <Label key={label} data-testid="model-catalog-label" color="purple" variant="filled">
               LAB teacher
             </Label>
           );
         case ReservedILabLabel.LabJudge:
           return (
-            <Label key={l} data-testid="model-catalog-label" color="orange" variant="filled">
+            <Label key={label} data-testid="model-catalog-label" color="orange" variant="filled">
               LAB judge
             </Label>
           );
--- a/frontend/src/pages/modelServing/screens/projects/InferenceServiceModal/ManageInferenceServiceModal.tsx
+++ b/frontend/src/pages/modelServing/screens/projects/InferenceServiceModal/ManageInferenceServiceModal.tsx
@@ -57,7 +57,6 @@ const ManageInferenceServiceModal: React.FC<ManageInferenceServiceModalProps> = 
     useK8sNameDescriptionFieldData({ initialData: editInfo });
   const [actionInProgress, setActionInProgress] = React.useState(false);
   const [error, setError] = React.useState<Error | undefined>();
-  const [isLoading, setIsLoading] = React.useState(!!modelDeployPrefillInfo);
 
   const currentProjectName = projectContext?.currentProject.metadata.name || '';
   const currentServingRuntimeName = projectContext?.currentServingRuntime?.metadata.name || '';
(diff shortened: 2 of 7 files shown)


ppadti:
@yih-wang could you please take a look the the screen recordings?

frontend/src/pages/modelServing/screens/projects/kServeModal/ManageKServeModal.tsx
```diff
@@ -185,6 +186,29 @@ const ManageKServeModal: React.FC<ManageKServeModalProps> = ({
...
+    kServeNameDesc.name,
+    servingRuntimeTemplates,
+    projectTemplates,
+    projectContext,
+    connectionsLoaded,
+    modelDeployPrefillInfo,
+    podSpecOptionsState.acceleratorProfile.loaded,
+    podSpecOptionsState.hardwareProfile.profilesLoaded,
+  ]);
+
```
manaswinidas, line 211: `useEffect` scares me now after we have fixed multiple issues with prefilling info, can we wrap everything into a `useMemo` and have a `useEffect` on it? Something like:
```suggestion
  const isDataReady = React.useMemo(() => {
    if (!modelDeployPrefillInfo) {
      return false;
    }

    return (
      !!kServeNameDesc.name &&
      Array.isArray(servingRuntimeTemplates) &&
      servingRuntimeTemplates.length > 0 &&
      (!projectContext || connectionsLoaded) &&
      (podSpecOptionsState.acceleratorProfile.loaded ||
        podSpecOptionsState.hardwareProfile.profilesLoaded)
    );
  }, [
    kServeNameDesc.name,
    servingRuntimeTemplates,
    projectContext,
    connectionsLoaded,
    modelDeployPrefillInfo,
    podSpecOptionsState.acceleratorProfile.loaded,
    podSpecOptionsState.hardwareProfile.profilesLoaded,
  ]);

  React.useEffect(() => {
    setIsLoading(!isDataReady);
  }, [isDataReady]);
```

manaswinidas (changes requested):
Works fine just small performance/readability improvements

frontend/src/pages/modelServing/screens/projects/InferenceServiceModal/ManageInferenceServiceModal.tsx
```diff
@@ -100,6 +101,21 @@ const ManageInferenceServiceModal: React.FC<ManageInferenceServiceModalProps> =
     return isConnectionValid;
   };
 
+  React.useEffect(() => {
+    if (createData.name && createData.name !== inferenceServiceNameDesc.name) {
+      setInferenceServiceNameDesc('name', createData.name);
```
mturley, line 106: I don't see why we need to add this additional `setInferenceServiceNameDesc` - I thought the issue here was just the spinner rendering and the form state was already getting set eventually?

frontend/src/pages/modelServing/screens/projects/InferenceServiceModal/ManageInferenceServiceModal.tsx
```diff
@@ -100,6 +101,21 @@ const ManageInferenceServiceModal: React.FC<ManageInferenceServiceModalProps> =
...
+    setIsLoading(false);
+  }, [inferenceServiceNameDesc.name, createData.name, setInferenceServiceNameDesc]);
+
+  React.useEffect(() => {
+    if (modelDeployPrefillInfo) {
+      const isDataReady = !!inferenceServiceNameDesc.name && (!projectContext || connectionsLoaded);
+
+      setIsLoading(!isDataReady);
+    }
+  }, [inferenceServiceNameDesc.name, projectContext, connectionsLoaded, modelDeployPrefillInfo]);
```
mturley, line 117: It would be great to not have `isLoading` be state if we can avoid it. Can we derive it at render time as a regular variable and prevent needing this useEffect?

```ts
const isDataReady = !modelDeployPrefillInfo || (!!inferenceServiceNameDesc.name && (!projectContext || connectionsLoaded));
const isLoading = !isDataReady;
```

frontend/src/pages/modelServing/screens/projects/kServeModal/ManageKServeModal.tsx
mturley, line 211: Same here, we can avoid using state for `isLoading` if its value is just derived from other things we know at render time. It can just be a regular variable with no useEffect.

```ts
const isDataReady = !modelDeployPrefillInfo ||
  !!kServeNameDesc.name &&
  Array.isArray(servingRuntimeTemplates) &&
  servingRuntimeTemplates.length > 0 &&
  (!projectContext || connectionsLoaded) &&
  (podSpecOptionsState.acceleratorProfile.loaded ||
    podSpecOptionsState.hardwareProfile.profilesLoaded)
```

mturley (changes requested):
I think we can avoid these useEffects, and I have a question about the new state setter call.

frontend/src/pages/modelServing/screens/projects/kServeModal/ManageKServeModal.tsx
mturley: Ah - this is similar to what @manaswinidas was saying with the useMemo - we can useMemo this as well if we want, I think it's most important to avoid introducing useState/useEffect here if we can avoid them. But I don't think useMemo gains us much since React must compare the values of all its dependencies every render anyway and the logic inside is about the same amount of boolean comparison, so we may as well just not memoize it IMO. I don't feel strongly about that part though.

manaswinidas:
I think this may fix the tests failing in https://github.com/opendatahub-io/odh-dashboard/pull/4046 cc @claudialphonse78 🤔

vconzola:
@ppadti I agree with @yih-wang about not displaying the hardware profile field until the label data has been fetched and is ready too.

frontend/src/concepts/modelCatalog/content/ModelCatalogLabels.tsx
```diff
@@ -14,19 +14,19 @@ export const ModelCatalogLabels: React.FC<{
...
           return (
-            <Label data-testid="model-catalog-label" color="purple" variant="filled">
+            <Label key={label} data-testid="model-catalog-label" color="purple" variant="filled">
               LAB teacher
             </Label>
           );
         case ReservedILabLabel.LabJudge:
           return (
-            <Label data-testid="model-catalog-label" color="orange" variant="filled">
+            <Label key={label} data-testid="model-catalog-label" color="orange" variant="filled">
```
mturley, line 29: Ah - thanks for this

frontend/src/pages/modelServing/screens/projects/InferenceServiceModal/ManageInferenceServiceModal.tsx
mturley: Interesting. The whole way we sync state with these components is a bit wonky, but this approach makes sense with what we have. Thanks

mturley (approved):
Good stuff, thanks @ppadti !

PR #4086, opened 2025-04-23
[Tech debt] Add eslint no-non-null-assertion rule to config

ikeola13:
Issue: https://issues.redhat.com/browse/RHOAIENG-23621

## Description

- added no-non-null-assertion to eslint config in both frontend and backend package.json files
- fixed resulting linting errors

## How Has This Been Tested?

Existing tests were ran and all passed.

## Test Impact

No additional tests were added.

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/pages/modelRegistrySettings/CreateModal.tsx
+++ b/frontend/src/pages/modelRegistrySettings/CreateModal.tsx
@@ -206,7 +206,7 @@ const CreateModal: React.FC<CreateModalProps> = ({ onClose, refresh, modelRegist
         kind: 'ModelRegistry',
         metadata: {
           name: nameDesc.k8sName.value || translateDisplayNameForK8s(nameDesc.name),
-          namespace: modelRegistryNamespace!,
+          namespace: modelRegistryNamespace || '',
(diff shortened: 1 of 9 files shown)


frontend/src/pages/modelRegistrySettings/CreateModal.tsx
```diff
@@ -206,7 +206,7 @@ const CreateModal: React.FC<CreateModalProps> = ({ onClose, refresh, modelRegist
         kind: 'ModelRegistry',
         metadata: {
           name: nameDesc.k8sName.value || translateDisplayNameForK8s(nameDesc.name),
-          namespace: modelRegistryNamespace!,
+          namespace: modelRegistryNamespace || '',
```
mturley, line 209: @ikeola13 we should avoid falling back to `''` wherever possible. Instead can you put an early return at the top of this function if modelRegistryNamespace isn't defined? If we don't have it we can't submit.

mturley (changes requested):
One tiny thing, sorry...

mturley (approved):

PR #4094, opened 2025-04-23
Allow the pipeline version of a recurring run to be optional

caponetto:
## Description

See https://issues.redhat.com/browse/RHOAIENG-19285

In this PR:
- Make `pipeline_version_reference.pipeline_version_id` optional
- Populate `pipeline_version_id` with the latest version if not provided when fetching recurring runs
- Change pipeline version selection area of recurring runs in the Run Page according to UX
- Add jest and cypress tests

## How Has This Been Tested?

Visit the pipeline run page to explore the new user experience. The page for scheduled runs should display a new radio group for selecting the pipeline version, either the latest or a fixed version. The scheduled run list and details should always load the latest version if users choose to use that option.

Be sure to be using DSPO v2.12.0+

https://github.com/user-attachments/assets/0d028cd9-2347-41c9-84a1-93c346e27bb5

## Test Impact

Added jest and cypress tests to cover the changes.

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/concepts/pipelines/apiHooks/usePipelineRecurringRunById.ts
+++ b/frontend/src/concepts/pipelines/apiHooks/usePipelineRecurringRunById.ts
@@ -6,21 +6,49 @@ import useFetchState, {
 } from '~/utilities/useFetchState';
 import { usePipelinesAPI } from '~/concepts/pipelines/context';
 import { PipelineRecurringRunKF } from '~/concepts/pipelines/kfTypes';
+import { K8sAPIOptions } from '~/k8sTypes';
 
 const usePipelineRecurringRunById = (
   pipelineRecurringRunByIdId?: string,
 ): FetchState<PipelineRecurringRunKF | null> => {
   const { api } = usePipelinesAPI();
 
+  const fetchLatestVersionId = React.useCallback(
+    async (pipelineId: string, opts: K8sAPIOptions) => {
+      const response = await api.listPipelineVersions(opts, pipelineId, {
+        sortField: 'created_at',
+        sortDirection: 'desc',
+      });
+      return response.pipeline_versions?.[0].pipeline_version_id;
+    },
+    [api],
+  );
+
   const call = React.useCallback<FetchStateCallbackPromise<PipelineRecurringRunKF | null>>(
-    (opts) => {
+    async (opts) => {
       if (!pipelineRecurringRunByIdId) {
         return Promise.reject(new NotReadyError('No pipeline recurring run id'));
       }
+      const recurringRun = await api.getPipelineRecurringRun(opts, pipelineRecurringRunByIdId);
 
-      return api.getPipelineRecurringRun(opts, pipelineRecurringRunByIdId);
+      if (recurringRun.pipeline_version_reference.pipeline_version_id) {
+        return recurringRun;
+      }
+
+      return {
+        ...recurringRun,
+        // eslint-disable-next-line camelcase
+        pipeline_version_reference: {
+          ...recurringRun.pipeline_version_reference,
+          // eslint-disable-next-line camelcase
+          pipeline_version_id: await fetchLatestVersionId(
+            recurringRun.pipeline_version_reference.pipeline_id,
+            opts,
+          ),
+        },
+      };
     },
-    [api, pipelineRecurringRunByIdId],
+    [api, fetchLatestVersionId, pipelineRecurringRunByIdId],
   );
 
   return useFetchState(call, null);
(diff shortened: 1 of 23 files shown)


frontend/src/concepts/pipelines/content/createRun/contentSections/PipelineVersionRadioGroup.tsx
```diff
@@ -0,0 +1,229 @@
...
+                        </>
+                      }
+                    >
+                      <Button
+                        style={{ paddingLeft: 0 }}
+                        icon={<OutlinedQuestionCircleIcon />}
+                        iconPosition="right"
+                        data-testid="view-latest-version-button"
+                        variant="link"
+                        disabled
```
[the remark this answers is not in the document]
caponetto: Probably a leftover. Good catch!

frontend/src/concepts/pipelines/apiHooks/usePipelineRecurringRunById.ts
```diff
@@ -6,21 +6,49 @@ import useFetchState, {
...
+    async (pipelineId: string, opts: K8sAPIOptions) => {
+      const response = await api.listPipelineVersions(opts, pipelineId, {
+        sortField: 'created_at',
+        sortDirection: 'desc',
+      });
+      return response.pipeline_versions?.[0].pipeline_version_id;
+    },
+    [api],
+  );
+
```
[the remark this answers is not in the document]
caponetto: In this case, I didn't use that hook to avoid making an unnecessary request.

frontend/src/concepts/pipelines/content/createRun/contentSections/RunTypeSection.tsx
```diff
@@ -29,7 +33,12 @@ export const RunTypeSection: React.FC<RunTypeSectionProps> = ({ data, isDuplicat
       To create a schedule that executes recurring runs,{' '}
       <Link
         to={createRecurringRunRoute(namespace, experiment?.experiment_id)}
-        state={{ locationData: data }}
+        state={{
+          locationData: data,
+          contextData: {
+            versionToUse: PipelineVersionToUse.LATEST,
+          },
```
DaoDaoNoCode, line 40: This is not correct, because we need to carry over what we set for the pipeline version on the create-run page to the create-schedule page. If we are doing this, when switching between the 2 pages, we will lose the context of what pipeline version the user chose before and always select the latest radio. I think we could remove the `contextData` here because it's only used to create run/schedules from a specific version/experiment. The `locationData` contains everything we have including the pipeline version, we should rely on that and always set to `PipelineVersionToUse.PROVIDED` if the `locationVersion` is set.

frontend/src/concepts/pipelines/content/createRun/RunPage.tsx
```diff
@@ -79,6 +83,15 @@ const RunPage: React.FC<RunPageProps> = ({
     [isSchedule, triggerType],
   );
 
+  const versionToUseData = React.useMemo<PipelineVersionToUse>(() => {
+    if (Object.values(PipelineVersionToUse).includes(contextVersionToUse)) {
+      return contextVersionToUse;
+    }
```
DaoDaoNoCode, line 89: Is this really needed? I think the `return` statement below should be able to cover all the situations. Could you give me an example to help me understand any scenarios I missed? 🤔

DaoDaoNoCode (commented):
Besides the 2 comments I left, I found there is a bug that's not related to this ticket, but might be good to fix here (Or you can create a ticket if you like). When uploading a new version on the create run page, it will always navigate to the version details page, which is inconsistent with the behavior when we import a new pipeline or create a new experiment on the create run page.
Like what we are doing for the import pipeline button, we should set a prop to avoid auto-navigation after creation in some situations, like here.

frontend/src/concepts/pipelines/content/createRun/RunPage.tsx
caponetto: This is to prioritize the prop that comes from the context, if provided, rather than trying to resolve the value through location or context versions. I figured it could be useful but at the moment there's no specific reason to have it. I'll remove it.

caponetto:
> Besides the 2 comments I left, I found there is a bug that's not related to this ticket, but might be good to fix here (Or you can create a ticket if you like). When uploading a new version on the create run page, it will always navigate to the version details page, which is inconsistent with the behavior when we import a new pipeline or create a new experiment on the create run page. Like what we are doing for the import pipeline button, we should set a prop to avoid auto-navigation after creation in some situations, like here.

I agree! I'll create a new ticket to avoid increasing the scope of this one.

Gkrumbach07:
@guimou is this PR ready to go then?

Gkrumbach07 (approved):
/lgtm

tested and works. the latest pipeline version is also displayed in the table view

PR #4099, opened 2025-04-24
Update SA Permissions Part Two

katieperry4:
## Description

Quick fix adding 'create' back into rbac.authorization.k8s.io for the cluster role.
After it's removal, attempting to create a standalone jupyter workbench if `opendatahub-image-pullers` didn't exist or was deleted, would result in an error. The SA account needs `create` to create a rolebinding if one doesn't exist.

## How Has This Been Tested?

Tested on my cluster
If you want to try it:
-Point your dsc dashboard devflags at this branch: `https://github.com/katieperry4/odh-dashboard/tarball/SA-Fix`
-Find the opendatahub-image-pullers rolebinding in the opendatahub namespace
-Delete it
-Open ODH
-Try to launch a standalone workbench

If you want to see the broken version (if this hasn't merged yet) point the devFlags at main
-Delete the opendatahub-image-pullers rolebinding in the opendatahub namespace
-Try to launch a standalone workbench, you'll see an error 

## Test Impact

No changes

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/manifests/core-bases/base/cluster-role.yaml
+++ b/manifests/core-bases/base/cluster-role.yaml
@@ -135,6 +135,7 @@ rules:
     verbs:
       - list
       - get
+      - create
     resources:
       - rolebindings
       - clusterrolebindings
--- a/manifests/core-bases/base/cluster-role.yaml
+++ b/manifests/core-bases/base/cluster-role.yaml
@@ -132,6 +132,8 @@ rules:
       - namespaces
   - apiGroups:
       - rbac.authorization.k8s.io
+      # The rbac.authorization.k8s.io permissions (list, get, create, patch, and delete on rolebindings, clusterrolebindings, and roles) are needed so the dashboard can handle user access and permissions automatically. 
+      # This includes tasks like assigning roles, updating existing bindings, and creating or removing access rules.
     verbs:
       - list
       - get
--- a/manifests/core-bases/base/cluster-role.yaml
+++ b/manifests/core-bases/base/cluster-role.yaml
@@ -134,6 +134,7 @@ rules:
       - rbac.authorization.k8s.io
       # The rbac.authorization.k8s.io permissions (list, get, create, patch, and delete on rolebindings, clusterrolebindings, and roles) are needed so the dashboard can handle user access and permissions automatically. 
       # This includes tasks like assigning roles, updating existing bindings, and creating or removing access rules.
+      # Ex. the standalone Jupyter workbench requires the ability to create an image puller role if one doesn't exist.
     verbs:
       - list
       - get
--- a/manifests/core-bases/base/cluster-role.yaml
+++ b/manifests/core-bases/base/cluster-role.yaml
@@ -136,6 +136,8 @@ rules:
       - list
       - get
       - create
+      - patch
+      - delete
     resources:
       - rolebindings
       - clusterrolebindings
--- a/manifests/core-bases/base/role.yaml
+++ b/manifests/core-bases/base/role.yaml
@@ -35,6 +35,7 @@ rules:
     verbs:
       - get
       - update
+      - delete
     resources:
       - cronjobs
   - apiGroups:


manifests/core-bases/base/cluster-role.yaml
```diff
@@ -135,6 +135,7 @@ rules:
     verbs:
       - list
       - get
+      - create
```
andrewballantyne, line 138: Just asking because I don't have it in front of me -- did we trace where this usage was and make sure it was only create needed? Do we need update?

jiridanek: > - [concepts/pipeline/elyra/utils.ts](https://github.com/opendatahub-io/odh-dashboard/blob/e02e1f67881d32a3ddb4bf9a740ecee9958c65d3/frontend/src/concepts/pipelines/elyra/utils.ts#L183) refer to needing to be able to update the service account?

Yes, elyra for sure, we have

* https://issues.redhat.com/browse/RHOAIENG-580
* https://issues.redhat.com/browse/RHOAIENG-23204

in progress, but for now Dashboard is still responsible for managing that one.

katieperry4: Okay so we're reverting the `rbac.authorization.k8s.io` to include `patch` and `delete`? It doesn't seem like `update` itself is being used.
(Just checking before I go ahead and do it)
[concepts/pipeline/elyra/utils.ts](https://github.com/opendatahub-io/odh-dashboard/blob/e02e1f67881d32a3ddb4bf9a740ecee9958c65d3/frontend/src/concepts/pipelines/elyra/utils.ts#L183) - uses `patch` 
[src/api/k8s/roleBindings.ts](https://github.com/opendatahub-io/odh-dashboard/blob/e02e1f67881d32a3ddb4bf9a740ecee9958c65d3/frontend/src/api/k8s/roleBindings.ts#L131) - uses `patch`
[backend/src/utils/resourceUtils.ts ](https://github.com/opendatahub-io/odh-dashboard/blob/e02e1f67881d32a3ddb4bf9a740ecee9958c65d3/backend/src/utils/resourceUtils.ts#L709) - just uses `create` and `delete` to update

emilys314: Would it make sense to try and just create the `opendatahub-image-pullers` RoleBinding through these manifest files as well? Or does the dynamic name of it prevent this

jiridanek: > Or does the dynamic name of it prevent this

It does, name is specified in DSCi, so I guess the rhods-operator _could_ somehow propagate it to the manifests... but it sounds more entangled than current implementation is, to me personally

emilys314:
I was able to confirm the standalone workbench is now working on a fresh cluster from cluster bot

katieperry4:
Took some time but I was able to run the e2e dashboard tests on Jenkins using a clean cluster. There was [one test failure](https://github.com/opendatahub-io/odh-dashboard/blob/main/frontend/src/__tests__/cypress/cypress/tests/e2e/applications/testAboutDialog.cy.ts), but I tested that locally on a cluster with permissions from: before these new SA changes, pointed at main, pointed at this PR and the same test failed.
Otherwise I think we're good to go!

manifests/core-bases/base/role.yaml
```diff
@@ -35,6 +35,7 @@ rules:
     verbs:
       - get
       - update
+      - delete
```
jiridanek, line 38: would it make sense to add single-line comments here explaining _why_ this is needed?

I imagine even if some of it goes out of date (as it invariably will), it may still prove useful for future perms/code archeologists. Or is the git-blame that should lead them to this PR sufficient hint to future generations of engineers?

katieperry4: I think it might be hard to keep the comments accurate considering any changes throughout the application and I don't want to mislead any future people, but I'm open to adding them if people think they'd be helpful 😃 
@emilys314 thoughts?

emilys314: Yeah it's pretty subjective, I don't have strong opinions. 

If any need a comment i'd say it's the `
rbac.authorization.k8s.io` though considering how powerful it is

manifests/core-bases/base/cluster-role.yaml
```diff
@@ -132,9 +132,14 @@ rules:
       - namespaces
   - apiGroups:
       - rbac.authorization.k8s.io
+      # The rbac.authorization.k8s.io permissions (list, get, create, patch, and delete on rolebindings, clusterrolebindings, and roles) are needed so the dashboard can handle user access and permissions automatically. 
+      # This includes tasks like assigning roles, updating existing bindings, and creating or removing access rules.
```
emilys314, line 136: Sorry if this is nitpicking but specifying where the dashboard automatically handles permissions is most useful here. So saying something about how the standalone jupyter workbench need to dynamically create a image pull role if it's missing or something like that

PR #4105, opened 2025-04-25
[wip] Model serving refactor poc

andrewballantyne:


andrewballantyne (commented):
Best to ignore everything under `concepts/modelServing` -- the "old way" is still part of this PoC 😞

PR #4106, opened 2025-04-28
Prefill name while deploying model versions from model registry

manaswinidas:
Closes: [RHOAIENG-19921](https://issues.redhat.com/browse/RHOAIENG-19921)

## Description

Prefilled "Connection name" in case of URI(try deploying the version in "mdas-uri-test" model to "Purva" project in `ui-green` cluster:
Screenshot 2025-04-24 at 7 28 15 PM

Prefilled "Connection name" in case of S3(try deploying the version in "mdas-test" model to "Purva" project in `ui-green` cluster):
Screenshot 2025-04-28 at 5 11 10 PM

## How Has This Been Tested?

1. Register any model by auto-filling from a connection belonging to a specific project - does not apply to old models
2. Try deploying the same registered model to another project without a matching connection and check whether the "Connection name" field gets prefilled(applies only to S3 and URI connections as we don't have registering using OCI connections yet)

## Test Impact

Added Cypress test and fixed unit tests

## Request review criteria:

If you have UI changes: 


frontend/src/pages/modelServing/screens/projects/usePrefillModelDeployModal.ts
```diff
@@ -100,7 +100,7 @@ const usePrefillModelDeployModal = (
         if (recommendedConnections.length === 0) {
           setCreateData('storage', {
             awsData: prefilledAWSData,
-            dataConnection: '',
```
mturley, line 103: @manaswinidas in the existing ModelDeployPrefillInfo type, can we either rename `modelArtifactStorageKey` to something like `initialConnectionName` or move it into the `modelRegistryInfo` sub-object there? The current name is model registry specific - I would have renamed it in #4002 but didn't want to cause conflicts with your branch here.

mturley (changes requested):
LGTM but one small thing

frontend/src/pages/modelServing/screens/projects/usePrefillModelDeployModal.ts
manaswinidas: Renamed `modelArtifactStorageKey` to `initialConnectionName`

PR #4119, opened 2025-04-29
added unit tests for useModelVersionsByRegisteredModel

ikeola13:
Issue: https://issues.redhat.com/browse/RHOAIENG-24227

## Description

added unit test for useModelVersionsByRegisteredModel.ts

## How Has This Been Tested?

Ran unit test and passed
Screenshot 2025-05-01 at 10 33 38

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/concepts/modelRegistry/apiHooks/__tests__/useModelVersionsByRegisteredModel.spec.ts
+++ b/frontend/src/concepts/modelRegistry/apiHooks/__tests__/useModelVersionsByRegisteredModel.spec.ts
@@ -1,0 +1,152 @@
+import * as React from 'react';
+import { renderHook } from '@testing-library/react';
+import useModelVersionsByRegisteredModel from '~/concepts/modelRegistry/apiHooks/useModelVersionsByRegisteredModel';
+
+// Mock the useContext hook
+jest.mock('react', () => ({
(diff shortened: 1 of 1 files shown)


frontend/src/concepts/modelRegistry/apiHooks/__tests__/useModelVersionsByRegisteredModel.spec.ts
```diff
@@ -0,0 +1,170 @@
...
+    mockUseFetchState.mockImplementation(actualUseFetchState);
+  });
+
+  it('should return initial state', () => {
+    const renderResult = testHook(useModelVersionsByRegisteredModel)();
+    expect(renderResult).hookToStrictEqual(standardUseFetchState(defaultEmptyState));
+    expect(renderResult).hookToHaveUpdateCount(1);
+  });
+
+  it('should reject when registeredModelId is not provided', () => {
```
YuliaKrimerman, line 57: Small thing here - "should reject when registeredModelId is not provided" is not exactly what the test is doing. The test actually checks that the API isn't called, but doesn't verify the rejection behavior. You can either rename it to something like"should not call API when registeredModelId is not provided"
Or adding actual rejection verification. Whatever you're comfortable with.

PR #4121, opened 2025-04-29
Model catalog homepage: fix issue of tooltip not showing during mouseover

nananosirova:
https://issues.redhat.com/browse/RHOAIENG-24491

## Description

The `Truncate` component was wrapped in a `Link`, which interfered with the `Tooltip` event listeners inside `Truncate`, which in turn caused the `Tooltip` not to show up. Switching `Link` with `Button` doesn't seem to have that problem. 

https://github.com/user-attachments/assets/f80d6cc9-d6f4-4709-b0e9-fca181b264c3

_Note: After the above changes, there might still be cases when the text is truncated but the tooltip isn't displayed on mouseover. In `Truncate`, the logic for adding truncation to the text and changing the display state of the tooltip are separate, therefore there are cases when the text is truncated but the display state for the tooltip is not switched to 'visible' on mouseover._

## How Has This Been Tested?

Navigate to the homepage:
- In the model catalog section, get one of the model names to truncate by either resizing the screen or manually changing the name to be longer 
- Mouseover the name, the full name should be displayed on the tooltip

## Test Impact

Added tests, and have existing tests checking link functionality.

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/__tests__/cypress/cypress/pages/home/homeModelCatalog.ts
+++ b/frontend/src/__tests__/cypress/cypress/pages/home/homeModelCatalog.ts
@@ -13,12 +13,20 @@ export class HomeModelCatalog extends Contextual<HTMLElement> {
     return this.find().findByTestId('model-catalog-card-gallery');
   }
 
+  getModelCatalogCardName(): Cypress.Chainable<JQuery<HTMLElement>> {
+    return this.find().findByTestId('model-catalog-card-name');
+  }
+
   getModelCatalogCardDescription(): Cypress.Chainable<JQuery<HTMLElement>> {
     return this.find().findByTestId('model-catalog-card-description');
   }
 
+  getModelCatalogCardNameTooltip(): Cypress.Chainable<JQuery<HTMLElement>> {
+    return this.getModelCatalogCardName().get('.pf-v6-c-tooltip');
+  }
+
   getModelCatalogCardDescriptionTooltip(): Cypress.Chainable<JQuery<HTMLElement>> {
-    return this.find().get('.pf-v6-c-tooltip');
+    return this.getModelCatalogCardDescription().get('.pf-v6-c-tooltip');
   }
 
   getModelCatalogFooter(): Cypress.Chainable<JQuery<HTMLElement>> {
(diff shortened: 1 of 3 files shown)


rsun19:
> Note: After the above changes, there might still be cases when the text is truncated but the tooltip isn't displayed on mouseover. In Truncate, the logic for adding truncation to the text and changing the display state of the tooltip are separate, therefore there are cases when the text is truncated but the display state for the tooltip is not switched to 'visible' on mouseover.

@nananosirova  Looked good when I tested this, however can you give an example case where this occurs? Such as what was the model name and description and the circumstances surrounding it?

nananosirova:
@rsun19 it doesn't seem to show up in circumstances where the name is only truncated by a few letters, the tooltip (popover) state doesn't seem to catch on that it's truncated, i.e the `isTruncated` state in `Truncate` doesn't switch to `true`.  `Truncate` seems to work as expected when I put the `Truncate` position to `position="end"` though, it seems to happen for `middle` and `start`

rsun19:
@nananosirova this seems to be a bug in the truncate component itself. I would bring this up to @Gkrumbach07, see what he thinks about this

Screenshot 2025-05-06 at 1 44 01 PM

I don't think this is something we can fix, so... 

/lgtm

Gkrumbach07 (approved):
Yeah truncate has had issues in the past. this seems like a edge case and maybe follow up with PF so best not block this

@jenny-s51 Do you recal this happening before. I thought something like this happened in pipelines and you had a fix for it

jenny-s51:
@Gkrumbach07 After investigating the issue, this indeed is a PF bug on an edge case, really great find @nananosirova @rsun19!

Opened https://github.com/patternfly/patternfly-react/issues/11811 in Patternfly to track the resolution of this bug.

Yes, PF Topology has its own [`truncateMiddle` util](https://github.com/patternfly/react-topology/blob/b88e9eb7ad15aa52bb33a92ea295633f144c446a/packages/module/src/utils/truncate-middle.ts#L19) (not from PF React) which it applies by default on the node labels @Gkrumbach07.

PR #4130, opened 2025-04-30
Deprecate useFetchState and useMakeFetchObject, add new useFetch hook, remove useContextResourceData, related refactoring and using new pattern to fix RHOAIENG-15593

mturley:
Resolves https://issues.redhat.com/browse/RHOAIENG-15593

## Description

In order to fix the bug, we needed to be able to pass additional properties like `hasNonDashboardItems` along with the array data when fetching InferenceServices and ServingRuntimes, which wasn't possible with the structure of `useContextResourceData<T>` (which limited the data to `T[]`, so it couldn't be used to store a wrapping object). Initially the fix for this was to replace the `useContextResourceData` usages for InferenceServices and ServingRuntimes with `useMakeFetchObject` and a `refreshRate` option passed down into `useFetchState` (since `useContextResourceData` also provided a refetch interval). After discussing with @andrewballantyne , we decided we should also fully deprecate `useFetchState` and replace it with a new hook that returns the fetch state object without requiring us to use something like `useContextResourceData` or `useMakeFetchObject` at all. We determined this branch was a good candidate for rolling that change out since it was already modifying related code.

### Full summary of the other changes, mostly generated by Cursor 😄 

# Codebase Changes Summary

## 1. New useFetch Hook Implementation
- Created new `useFetch.ts` file implementing a modern and flexible data fetching hook
- Returns `FetchStateObject` type instead of an array for better type safety
- Added comprehensive test coverage in `useFetch.spec.ts`

## 2. Deprecation of useFetchState
- `useFetchState` marked as deprecated in favor of `useFetch`
- Old implementation maintained as wrapper around `useFetch` for backward compatibility
- Main difference: `useFetch` returns `{ data, loaded, error, refresh }` object instead of array

## 3. Removed Deprecated Utilities
- Deleted `useContextResourceData.tsx`
- Deleted `useRestructureContextResourceData.tsx`
- `useMakeFetchObject` marked as deprecated

## 4. Changes in ModelServingContext and ProjectDetailsContext
- Since `useInferenceServices` and `useServingRuntimes` now no longer require a wrapping `useContextResourceData` since they use the new `useFetch` to directly return a FetchStateObject, made similar updates to all the other data fetching hooks in these two context providers for consistency and to provide more examples of the simplification that comes from replacing useFetchState with useFetch.

## 5. Usage of the new patterns to implement the bug fix for https://issues.redhat.com/browse/RHOAIENG-15593
- Now that `useContextResourceData` is gone and we can more flexibly pass around object data, the `useInferenceServices` and `useServingRuntimes` hooks now return an object wrapping the items array so that it can also include a `hasNonDashboardItems` boolean. The hooks no longer use `labelSelector` to filter by our dashboard label and instead do this filtering in JS to detect items that don't have that label.
- Having this boolean present in the fetched data allows us to change the logic in `ModelServingPlatformSelectButton` so it will not allow un-setting the serving platform for a project if that project contains non-dashboard ServingRuntimes or InferenceServices.

## How Has This Been Tested?

Tested the bug functionality, and also sanity checked various data fetching across the app

## Test Impact

- Added new tests for useFetch based on the old useFetchState tests
- Added new test case for the non-dashboard item filtering behavior

## Request review criteria:

If you have UI changes: 


pnaik1:
great work mike!,  I noticed a blank page appears when non-dashboard items are present 
shouldn't this display a empty page with disabled change button??
otherwise all functionality works well, 
Screenshot 2025-05-16 at 1 02 14 PM

mturley:
@pnaik1 thanks!

That's an interesting case actually - I suspect what you did is deployed a model with the dashboard and then edited the InferenceService to remove the dashboard label? If so, that broken table you end up with is actually already present on main. It happens when the dashboard finds servingruntimes but no inferenceservices in a kserve project. You can see the correct behavior if you also remove the dashboard label from the servingruntime.

However, I think this case should be impossible to hit - maybe a user is creating servingruntimes and inference services by hand, and they are in a kserve project, created their servingruntime and have not yet created their inferenceservice. My last commit makes this view accurately hit its empty state in that case - we should consider the view empty if there are no inferenceservices for kserve projects, even if there are servingruntimes. So, with my latest changes the same test case you used should also work as expected.

This is ready for re-review!

PR #4137, opened 2025-05-01
RHOAIENG-24706: Part 1: Replace Deprecated Modal Instances with PF6 Modal

srtanish1992:
**JIRA:** https://issues.redhat.com/browse/RHOAIENG-24706

## Description

This is the first PR to replace deprecated Modal component with PF6 Modal. Below are the major changes implemented in the PR,

> In PF6, we have separate ModalBody, ModalHeader, ModalFooter components. In the deprecated version, we pass header and footer/actions as props. So we just have to use these new components instead of props.

## How Has This Been Tested?

1. Manually tested the UI for the affected areas of the dashboard to make sure that the new Modal loads correctly with all the right content.

## Test Impact

I have just made sure that the tests which are already present are passing. I have fixed the cypress mock tests which failed.

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/concepts/pipelines/content/experiment/CreateExperimentModal.tsx
+++ b/frontend/src/concepts/pipelines/content/experiment/CreateExperimentModal.tsx
@@ -7,8 +7,11 @@ import {
   Stack,
   StackItem,
   TextInput,
+  Modal,
+  ModalBody,
+  ModalHeader,
+  ModalFooter,
 } from '@patternfly/react-core';
-import { Modal } from '@patternfly/react-core/deprecated';
 import { usePipelinesAPI } from '~/concepts/pipelines/context';
 import useCreateExperimentData from '~/concepts/pipelines/content/experiment/useCreateExperimentData';
 import { ExperimentKF } from '~/concepts/pipelines/kfTypes';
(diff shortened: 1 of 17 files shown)


rsun19:
Tested most of the changes, looks good. I will give another update later

DaoDaoNoCode:
What's the scope of the files we are replacing the `Modal` component? I see there are still a lot of components that are using the deprecated Modal.

frontend/src/concepts/pipelines/content/tables/pipelineRun/CustomMetricsColumnsModal.tsx
```diff
@@ -44,54 +47,47 @@ export const CustomMetricsColumnsModal: React.FC<CustomMetricsColumnsModalProps>
...
-            Select up to 10 metrics that will display as columns in the table. Drag and drop column
-            names to reorder them.
-          </StackItem>
+      variant="small"
+      isOpen
+      onClose={onClose}
+    >
+      <ModalHeader
+        title="Customize metrics columns"
+        description={
```
nananosirova, line 58: Is there a reason why the header description is not in `ModalBody`?

DaoDaoNoCode: This is how PF designed the component https://www.patternfly.org/components/modal#modalheader

srtanish1992:
> ~What's the scope of the files we are replacing the `Modal` component? I see there are still a lot of components that are using the deprecated Modal.~
> 
> EDIT: I see the following PRs. That makes sense. My only suggestion is to use `<ModalFooter>` to wrap everything in the `DashboardModalFooter` component after all the updates, because it makes more sense, and it can help prevent from importing `ModalFooter` every time when you need to use the `DashboardModalFooter`.

I like the suggestion. I will create a follow-up PR once all the updated have been merged.

frontend/src/concepts/pipelines/content/experiment/CreateExperimentModal.tsx
```diff
@@ -81,59 +134,11 @@ const CreateExperimentModal: React.FC<CreateExperimentModalProps> = ({ onClose }
           }}
         >
           Create experiment
-        </Button>,
+        </Button>
```
DaoDaoNoCode, line 137: Just one thing, could you please replace this part with the `DashboardModalFooter` here? There might be some modals still using `actions` instead of `footer`, please make sure to replace that part as well to make sure it's consistent.

DaoDaoNoCode (commented):
Just left 2 comments about the same thing, other parts lgtm!

frontend/src/concepts/pipelines/content/experiment/CreateExperimentModal.tsx
srtanish1992: Is it okay if I do this change and moving `ModalFooter` inside `DashboardModalFooter` in a follow-up PR since both the changes are in the same area?

DaoDaoNoCode: Yeah, that works!

DaoDaoNoCode (approved):
/lgtm

PR #4138, opened 2025-05-01
init monorepo and refactor navigation to extensions

christianvogt:

## Description

Introduces a monorepo into the `frontend/packages` directory. To not cause too much disruption, no restructuring was done and the `frontend/src` dir was added as a package as well and not moved. 

- Introduces `navigation` and `status-provider` extensions.
- All of the left nav has been migrated to extensions.
- Hardware profiles has a special use case which adds status to the navigation item and therefore the `status-provider` extension was created.
- Had to fix up some bad dependencies in the cypress directory which were causing issues for some reason. Moved types to mocks so that mocks won't import from cypress.
- Updated `mockUsers` because a test use to falsely pass but now it was failing. The default setup for a project admin should be that they have access to their project (or by default any project we are testing with) except the deployment namespace.
- Introduced `useAccessReviewExtensions` hook to make it easier to filter extensions based on access review checks.
- Introduced `HookNotify` which is a useful component to render hooks from extensions for which their return values must be aggregated. Since extensions are dynamic, we shouldn't render the extensions hooks in a loop, so instead we can render them as components and get a callback for their return value.

## How Has This Been Tested?

Ad hoc in the UI to ensure navigation continues to work.
Cypress tests already cover navigation and feature flagged nav items.

## Test Impact

Added some unit tests.

## Request review criteria:

If you have UI changes: 


emilys314:
What's the difference between `frontend/packages/plugin-core` vs `frontend/src/plugins`? 

Is `packages/plugin-core` basically a customized / internal `@openshift/dynamic-plugin-sdk`? and `frontend/src/plugins` are odh declaring plugins it can consume?

christianvogt:
@emilys314 yes the intent of `plugin-core` was just to be the customized dynamic-plugin-sdk bits. Ideally it doesn't import from anything else. I put the extensions there for now.

The `src/` dir will eventually be broken up better into individual feature packages and shared components. But I didn't want to do too much at once.

frontend/src/app/navigation/ExtensibleNav.tsx
```diff
@@ -0,0 +1,35 @@
...
+type Props = {
+  label: string;
+};
+
+export const ExtensibleNav: React.FC<Props> = ({ label }) => {
+  const extensions = useExtensions<NavExtension>(isNavExtension);
+  const topLevelExtensions = React.useMemo(
+    () =>
+      extensions
+        .filter((e) => isNavSectionExtension(e) || !e.properties.section)
```
emilys314, line 21:
```suggestion
  const extensions = useExtensions<NavExtension>(isNavSectionExtension);
```

Shouldn't you just directly pull the extensions you want?

christianvogt: Top level items are not just sections but also links which are not part of a section such as the `Home` link.

It's certainly possible that I make a new type guard function and do it in one shot instead of getting the extensions and then filtering again.

emilys314:
I tested the nav bar items on the left and they all worked. I also deployed the image on my cluster and logged in as a non admin user, and correctly wasn't able to access the admin pages

frontend/config/webpack.common.js
```diff
@@ -36,7 +36,12 @@ module.exports = (env) => {
         {
           test: /\.(tsx|ts|jsx|js)?$/,
           exclude: [/node_modules/, /__tests__/, /__mocks__/],
-          include: [SRC_DIR, COMMON_DIR],
+          include: [
+            SRC_DIR,
+            COMMON_DIR,
+            // path.resolve(RELATIVE_DIRNAME, '../packages'),
+            // /node_modules\/(@odh-dashboard|odh-dashboard-frontend)\//,
```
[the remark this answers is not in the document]
christianvogt: Fixed.

PR #4141, opened 2025-05-01
Cleaning up SimpleSelect Component

katieperry4:
Closes [RHOAIENG-16059](https://issues.redhat.com/browse/RHOAIENG-16059)

## Description
There were some inconsistencies in how the simpleSelect component was being used
-value not always being set -> now it's required
-placeholder and toggleLabel being used incorrectly -> usages fixed
-The simpleSelectOptions could have any prop passed into it -> added type to enforce strict usage of SimpleSelectOption to prevent passing in other props -> stopped exporting SimpleSelectOption -> only using SimpleSelectOptionStrict

- **Placeholder** is for when there's no selection and is optional, defaults to 'Select...'.

- **ToggleLabel** overrides value and is optional, just for if you want something other than value's label. Without toggleLabel, value will match with the key from `options` and use `label`.

- **Value** was made to be required but allows for `undefined` since there isn't always a selected value (like before an option is chosen). Arguably I could change it to `value: string` and in each usage where `value` might be `undefined` do something like `value={theoreticalValue || ''}` to set it to an empty string (aka undefined). I landed on `value: string | undefined` for now but if there are strong preferences for the other, I can make those changes.

There were two instances of simpleSelect with custom checkmarks in: [frontend/src/pages/learningCenter/LearningCenterToolbar.tsx](https://github.com/opendatahub-io/odh-dashboard/blob/main/frontend/src/pages/learningCenter/LearningCenterToolbar.tsx) that had little grey checkmarks left of the selected option, whereas in the rest of the app the checkmarks are blue and to the right of the selected option. So I updated it for consistency, but if there was a specific reason for having these selects be different, I can revert it.
**Previously:**
Screenshot 2025-05-01 at 10 33 18 AM 
Screenshot 2025-05-01 at 10 33 09 AM

**Now:**
Screenshot 2025-05-01 at 11 00 47 AM
Screenshot 2025-05-01 at 10 32 57 AM

## How Has This Been Tested?

Tested locally and with `npm run test`

if you want to test this:
- Go mess with the 44 usages of this component and see if they're still working. 
- They should be since all the tests passed 🤞 

## Test Impact

No tests changed

## Request review criteria:

If you have UI changes: 


frontend/src/components/SimpleSelect.tsx
```diff
@@ -33,13 +33,18 @@ export type SimpleSelectOption = {
...
   key: string;
   label: string;
-  options: SimpleSelectOption[];
+  options: SimpleSelectOptionStrict[];
 };
 
+// Prevents addition of anything outside what's defined in SimpleSelectOption
+type Strict<T> = T & { [K in Exclude<keyof T, keyof SimpleSelectOption>]: never };
+
+export type SimpleSelectOptionStrict = Strict<SimpleSelectOption>;
```
emilys314, line 42: Hi I don't think this is doing what you intended it to do. (I hope I'm not missing something but here is my explanation:)

I think your intent is to prevent creating options with properties that don't match up / are not defined in the `Option` type. But the original type is already doing this. The following code using the 2 types will give the same typescript error `Object literal may only specify known properties, and 'additionalProp' does not exist in type 'SimpleSelectOption'`:
```
const original = {
  key: 'key',
  label: 'label',
  additionalProp: 'asdf',
} satisfies SimpleSelectOption;

const originalStrict = {
  key: 'key',
  label: 'label',
  additionalProp: 'asdf',
} satisfies SimpleSelectOptionStrict;
```

In this case, I am adding an extra prop that doesn't exist in the `SimpleSelectOptionStrict` type. But this isn't being caught because it is going throuh the `.map` so the typeguard isn't sufficient here. 
image

In this case, I am directly using the type, and typescript can still directly typeguard the non strict version
image

If you look closely at the IDE, and what typescript evaluate the type to, it would be `type SimpleSelectOptionStrict = SimpleSelectOption & {}`
And the ` <base_type> & {}` is saying ' you can have properties of base_type & non null values'. And any object is a non null value, so the `& {}` will always be satisfied by typescript 

image

I would recommend just leaving the original `SimpleSelectionOption`

frontend/src/components/SimpleSelect.tsx
```diff
@@ -33,13 +33,18 @@ export type SimpleSelectOption = {
...
+
 type SimpleSelectProps = {
-  options?: SimpleSelectOption[];
+  options?: SimpleSelectOptionStrict[];
   groupedOptions?: SimpleGroupSelectOption[];
-  value?: string;
+  value: string | undefined;
```
emilys314, line 47: very minor nitpick, but all the other properties for this type are using `...?: ` and not `...: .. | undefined`. So i'd personally prefer this to not be changed. (The 2 different styles are quite subjective though, i just match what exists usually)
```suggestion
  value?: string;
```

frontend/src/components/SimpleSelect.tsx
katieperry4: Ahhhhhhh okay I see what you mean! I'll revert back to `SimpleSelectOption`

emilys314:
seems to be working good for me!

image

frontend/src/pages/acceleratorProfiles/screens/manage/tolerations/TolerationFields.tsx
```diff
@@ -49,7 +49,7 @@ const TolerationFields: React.FC<TolerationFieldsProps> = ({ toleration, onUpdat
         <SimpleSelect
           isFullWidth
           options={effectDropdownOptions}
-          value={toleration.effect || ''}
+          value={toleration.effect}
```
emilys314, line 52: A few lines up on line `42` you missed a spot to remove the `|| ''`

emilys314 (commented):
everything else looks good to me

frontend/src/pages/acceleratorProfiles/screens/manage/tolerations/TolerationFields.tsx
katieperry4: good catch thank you!

PR #4145, opened 2025-05-02
refactor AppRoutes to extensions

christianvogt:

## Description

- created a new extension point for defining routes
- Migrated all routes found in `AppRoutes` to extensions. These routes were not organized into separate packages at this time.

The `/notebookController/*` route needs to be feature flagged. However if we use dynamic flags to govern the enablement of a route, when navigating directly to the route, the page will flash the `NotFound` page first until the flag loads. To work around this, I have enabled the route always but done the check in the route and if it fails, the user is redirected back to the root page.

## How Has This Been Tested?

Ad hoc
Cypress tests

## Test Impact

App routes are covered by cypress tests.

## Request review criteria:

If you have UI changes: 


frontend/src/app/AppRoutes.tsx
```diff
@@ -1,94 +1,41 @@
...
+  const dynamicRoutes = React.useMemo(
+    () =>
+      routeExtensions.map((routeExtension) => (
+        <Route
+          key={routeExtension.uid}
+          path={routeExtension.properties.path}
+          element={
+            <LazyCodeRefComponent
+              key={routeExtension.uid}
+              component={routeExtension.properties.component}
```
andrewballantyne, line 31: Did you have a problem here with the `.default` aspect of the code ref? @emilys314 was experiencing this with the [initial code](https://redhat-internal.slack.com/archives/C069KSM8T9N/p1746046883693909?thread_ts=1746046259.375289&cid=C069KSM8T9N) we had from your first PR

emilys314: Yes i had this issue when using `useResolvedExtensions` and rendering the extension directly into `< ... />`

But with `useExtensions` and putting it into `<LazyCodeRefComponent />` there was not this issue anymore

andrewballantyne: Ah gotcha -- LazyCodeRefComponent manages that aspect I guess -- good to note!

frontend/package-lock.json
lucferbux, line 1: ugh, I recreate the `package-lock.json` here https://github.com/opendatahub-io/odh-dashboard/pull/4232 it might cause some conflicts, if that was the same goal we might be good if nothing was added.

frontend/src/app/AppRoutes.tsx
```diff
@@ -1,92 +1,41 @@
...
+              fallback={fallback}
+            />
+          }
+        />
+      )),
+    [routeExtensions],
```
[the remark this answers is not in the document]
lucferbux: Heh, I see it doesn't check the whole context of the database, just to confirm, this is already covered by `isExtensionInUse` right @christianvogt (I'm sorry to call copilot, but i was wondering if it improved with the new addition of the other agent in issues)

frontend/src/pages/notebookController/NotebookController.tsx
```diff
@@ -1,12 +1,18 @@
...
+import { useCheckJupyterEnabled } from '~/utilities/notebookControllerUtils';
 import NotebookServerRoutes from './screens/server/NotebookServerRoutes';
 import NotebookControllerTabs from './screens/admin/NotebookControllerTabs';
 import { NotebookControllerContextProvider } from './NotebookControllerContext';
 import ValidateNotebookNamespace from './ValidateNotebookNamespace';
 
 const NotebookController: React.FC = () => {
   const { isAdmin } = useUser();
+  const isJupyterEnabled = useCheckJupyterEnabled();
+  if (!isJupyterEnabled) {
```
lucferbux, line 13: Is there any other way to pull this workaround? I'm not agains it, I mean we would need to deprecate this anyway, but still. either way we can keep it.

frontend/src/app/AppRoutes.tsx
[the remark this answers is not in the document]
christianvogt: The extensions returned by the hook are already filtered against the flags.

frontend/src/pages/notebookController/NotebookController.tsx
christianvogt: We need an additional check that is async to conditionally enable / disable extension contributions. But we don't have anything other than feature flags / areas at the moment.
This can be added in the future but for now this seems like a simple work around.

frontend/package-lock.json
christianvogt: i rebased the pr and updated the package-lock

christianvogt:
had to update the markdown test as for some unrelated reason the generated html has different ordering of attributes which cause string comparisons to fail

frontend/src/plugins/extensions/routes.ts
```diff
@@ -0,0 +1,244 @@
...
+    type: 'app.route',
+    properties: {
+      path: '/distributedWorkloads/*',
+      component: () => import('~/pages/distributedWorkloads/GlobalDistributedWorkloadsRoutes'),
+    },
+  },
+  {
+    type: 'app.route',
+    flags: {
+      required: [ADMIN_USER],
```
andrewballantyne, line 157: FWIW, naturally we will follow up with proper work later, this has to go away in order to support future growth. BYO OIDC will cause some issues with the concept of "admin".

We'll be migrating away from "isAdmin" in the future in favour of SSAR checks

PR #4148, opened 2025-05-05
Cypress --env option to set tests timeout: CY_TEST_TIMEOUT_SECONDS  

manosnoam:
https://issues.redhat.com/browse/RHOAIENG-24885

## Description
This PR adds a new environment variable, `CY_TEST_TIMEOUT_SECONDS`, to set a global timeout for `it` tests in Cypress. The value is specified in seconds and is converted to milliseconds for compatibility with Cypress's `testTimeout` setting.

Usage example: 
`npm run --prefix frontend cypress:run -- --env CY_TEST_TIMEOUT_SECONDS=30`

Only those tests that exceeded 30 seconds will stop due to the timeout limit:

image

In addition, this PR fixes a problem when `grepTags` includes tags that do not exist in the actual test files, but e2e.ts was still executing those tests `Before` and `After` steps, instead of skipping those steps.

## How Has This Been Tested?
Tested by running Cypress with various values for `CY_TEST_TIMEOUT_SECONDS` to ensure the timeout is applied correctly to all `it` tests.

## Test Impact
The implementation has been verified to work without affecting existing tests. 

## Request review criteria:

### DIFF
--- a/frontend/src/__tests__/cypress/cypress/support/e2e.ts
+++ b/frontend/src/__tests__/cypress/cypress/support/e2e.ts
@@ -363,3 +363,17 @@ after(() => {
   });
 });
 
+// Override Mocha's global after() to skip spec-level after hooks when using grepTags
+const origAfter = (window as Window & typeof globalThis).after;
+(window as Window & typeof globalThis).after = function afterOverride(
+  nameOrFn: string | Mocha.Func,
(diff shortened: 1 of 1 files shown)


Griffin-Sullivan (commented):
/lgtm haven't tested myself

FedeAlonso (approved):

PR #4165, opened 2025-05-06
Update isRHOAI to check product name in DSC

nananosirova:
https://issues.redhat.com/browse/RHOAIENG-16249

## Description

The isRHOAI check does a namespace check against the static naming of `redhat-ods-applications`. This will not work if the customer changes this in the DSCI.

We should check against the DSC and check the product name. 

_Note: I checked the release name definition in rhods-operator: https://github.com/red-hat-data-services/rhods-operator/blob/main/pkg/cluster/const.go_ 

## How Has This Been Tested?

It's been tested locally for both opendatahub and self-managed rhoai platform types. 
From the homepage, opened the About Modal and verified the release name on the modal, and the value in backend in debug mode.
Added unit tests as well.

## Test Impact

Unit tests.

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/backend/src/types.ts
+++ b/backend/src/types.ts
@@ -1028,6 +1028,9 @@ export type DataScienceClusterKindStatus = {
   conditions: K8sCondition[];
   installedComponents: { [key in ComponentNames]?: boolean };
   phase?: string;
+  release?: {
+    name: string;
+  };
 };
 
 export type DataScienceClusterKind = K8sResourceCommon & {
@@ -1312,3 +1315,9 @@ export type AuthKind = K8sResourceCommon & {
     allowedGroups: string[];
   };
 };
+
+export enum OdhPlatformType {
+  OPEN_DATA_HUB = 'Open Data Hub',
+  SELF_MANAGED_RHOAI = 'OpenShift AI Self-Managed',
+  MANAGED_RHOAI = 'OpenShift AI Cloud Service',
+} // Reference: https://github.com/red-hat-data-services/rhods-operator/blob/main/pkg/cluster/const.go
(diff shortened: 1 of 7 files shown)


srtanish1992:
I have verified `Open Data Hub and OpenShift AI Self-Managed` cases locally. I am able to see the right boolean value for `isRHOAI` locally. 

I think we should also update this front end check so that we display the right text content in the `AboutDialog.tsx (search for isRHOAI)` - https://github.com/opendatahub-io/odh-dashboard/blob/0fac5dc5b3b392943095cb4492e18903481a3b80/frontend/src/app/App.tsx#L69

I will work with you to see if we can manually test `OpenShift AI Cloud Service` test case.

backend/src/types.ts
```diff
@@ -1312,3 +1315,9 @@ export type AuthKind = K8sResourceCommon & {
     allowedGroups: string[];
   };
 };
+
+export enum OdhPlatformType {
+  OPEN_DATA_HUB = 'Open Data Hub',
+  SELF_MANAGED_RHOAI = 'OpenShift AI Self-Managed',
+  MANAGED_RHOAI = 'OpenShift AI Cloud Service',
+}
```
Gkrumbach07, line 1323: can you add a comment here referencing where you got the values for future reference

https://github.com/red-hat-data-services/rhods-operator/blob/main/pkg/cluster/const.go

Gkrumbach07 (changes requested):
looks good. 

++ on what @srtanish1992  said

srtanish1992:
I manually tested the latest changes and I see the right content for `About` section. I tested `Open Data Hub and Self Managed RHOI`.

Gkrumbach07 (approved):

PR #4167, opened 2025-05-06
init

emilys314:
## Request review criteria:

If you have UI changes: 


frontend/packages/kserve/src/extensions.ts
```diff
@@ -0,0 +1,74 @@
...
+// eslint-disable-next-line import/no-extraneous-dependencies
+import { NamespaceApplicationCase } from '@odh-dashboard/internal/pages/projects/types';
+
+const extensions: (ModelServingPlatform | ModelServingPlatformCard)[] = [
+  {
+    type: 'model-serving.platform',
+    properties: {
+      id: 'kserve',
+      name: 'KServe',
+      isInstalled: () => Promise.resolve(true),
```
emilys314, line 17: Probably should extract these 4 functions into a dedicated file

frontend/packages/kserve/src/extensions.ts
```diff
@@ -0,0 +1,74 @@
...
+      isEnabled: (project: ProjectKind) =>
+        project.metadata.labels?.['modelmesh-enabled'] === 'false',
```
emilys314, line 19: Not sure if we should fix now or later, this also requires matching up the backend api call. 

For the interim we could make kserve check modelmesh=false OR kserve=true

frontend/packages/kserve/src/extensions.ts
```diff
@@ -0,0 +1,74 @@
...
+      enable: (project: ProjectKind) =>
+        addSupportServingPlatformProject(
+          project.metadata.name,
+          NamespaceApplicationCase.KSERVE_PROMOTION,
+        ),
+      disable: (project: ProjectKind) =>
+        addSupportServingPlatformProject(
+          project.metadata.name,
+          NamespaceApplicationCase.RESET_MODEL_SERVING_PLATFORM,
```
emilys314, line 28: Maybe we could entirely move away from this backend endponit? And directly update the project

frontend/packages/modelServing/src/ProjectModelsContext.tsx
```diff
@@ -0,0 +1,58 @@
...
+  models: [],
+});
+
+type ProjectModelsProviderProps = {
+  children: React.ReactNode;
+};
+
+export const ProjectModelsProvider: React.FC<ProjectModelsProviderProps> = ({ children }) => {
+  const { availablePlatforms } = React.useContext(ModelServingContext);
+  const { currentProject } = React.useContext(ProjectDetailsContext); // TODO: this should refresh from a websocket
```
emilys314, line 29: This should be a websocket and should rerender when the project changes. This is how the old model serving updates the page

frontend/packages/modelServing/src/ProjectModelsContext.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+  const [servingPlatform, setServingPlatform] = React.useState<
+    ProjectModelsContextType['platform']
+  >(getActiveServingPlatform(currentProject, availablePlatforms ?? []));
+  const setModelServingPlatform = React.useCallback(
```
emilys314, line 34: I think it's still good to update the page immediately for faster UI, and not have to always wait on the project websocket data to update

frontend/packages/modelServing/src/components/projectDetails/NoModelsView.tsx
```diff
@@ -0,0 +1,34 @@
...
+    () => cards.find((c) => c.properties.platform === platform.properties.id),
+    [cards, platform],
+  );
+
+  return (
+    <EmptyDetailsView
+      allowCreate
+      iconImage={typedEmptyImage(ProjectObjectType.modelServer)}
+      // imageAlt={isProjectModelMesh ? 'No model servers' : 'No deployed models'}
+      title={selectedPlatformCard?.properties.title}
```
emilys314, line 26: This is actualy different text data that needs to be updated

frontend/packages/modelServing/src/components/projectDetails/ModelsProjectDetailsTab.tsx
```diff
@@ -0,0 +1,85 @@
...
+    setModelServingPlatform,
+    models,
+  } = React.useContext(ProjectModelsContext);
+
+  const isLoading = !project || !models || !availablePlatforms;
+  const hasModels = models && models.length > 0;
+
+  const activePlatform = React.useMemo(
+    () =>
+      availablePlatforms && availablePlatforms.length === 1
```
emilys314, line 30: Also need to make sure we will be setting the active platform when preselecting the 1 platform. Previously it's done with clicking the Deploy button

frontend/packages/modelServing/src/ModelServingContext.tsx
```diff
@@ -0,0 +1,28 @@
...
+export const ModelServingContext = React.createContext<ModelServingContextType>({
+  availablePlatforms: [],
+});
+
+type ModelServingProviderProps = {
+  children: React.ReactNode;
+};
+
+export const ModelServingProvider: React.FC<ModelServingProviderProps> = ({ children }) => {
+  const platforms = useExtensions<ModelServingPlatform>(isModelServingPlatform);
```
emilys314, line 18: We also need to be checking if the platforms are enabled

frontend/packages/modelServing/src/extensions.ts
```diff
@@ -3,18 +3,20 @@ import type { ProjectDetailsTab } from '@odh-dashboard/plugin-core/extension-poi
 // eslint-disable-next-line import/no-extraneous-dependencies
 import { SupportedArea } from '@odh-dashboard/internal/concepts/areas/types';
```
emilys314, line 4:
```suggestion
// eslint-disable-next-line import/no-extraneous-dependencies
import { SupportedArea } from '@odh-dashboard/internal/concepts/areas/types';
```

This type is broken now for some reason?

emilys314:
I got a little wacky and the platform will update as the project changes (as it was before) but i also made it optimistically update the state.

The 2 would fight eachother but the useRef solutin should work

frontend/packages/modelServing/src/extensions.ts
```diff
@@ -3,18 +3,20 @@ import type { ProjectDetailsTab } from '@odh-dashboard/plugin-core/extension-poi
...
+      component: () => import('./components/projectDetails/ModelsProjectDetailsTab'),
     },
     flags: {
       required: [SupportedArea.PLUGIN_MODEL_SERVING],
     },
-  },
+  } satisfies ProjectDetailsTab,
 ];
 
+const extensions: ProjectDetailsTab[] = [...projectDetailsTabs];
```
andrewballantyne, line 20: Not sure if you need a type here -- but if you're going to do it, maybe `typeof variable` so you don't have to update both spots when things change.

frontend/packages/modelServing/src/concepts/ModelServingContext.tsx
```diff
@@ -0,0 +1,29 @@
...
+  children: React.ReactNode;
+};
+
+export const ModelServingProvider: React.FC<ModelServingProviderProps> = ({ children }) => {
+  const [platforms] = useResolvedExtensions(isModelServingPlatformExtension);
+
+  const contextValue = React.useMemo(
+    () => ({ availablePlatforms: platforms } satisfies ModelServingContextType),
+    [platforms],
+  );
```
andrewballantyne, line 24:
```
function useMemo<T>(factory: () => T, deps: DependencyList): T;
```
FWIW, Memo is a generic -- you can just provide the type to it instead of satisfies.

frontend/packages/modelServing/src/concepts/modelServingPlatforms.ts
```diff
@@ -0,0 +1,81 @@
...
+    // This prevents the effect from overriding an optimistic update.
+    if (newPlatform.current !== undefined) {
+      return;
+    }
+    const p = getActiveServingPlatform(project, platforms);
+    if (p?.properties.id !== activePlatform?.properties.id) {
+      setTmpActivePlatform(p);
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [project, platforms, activePlatform?.properties.id, newPlatform.current]);
```
andrewballantyne, line 41: I think you need to call out what is going on when you disable this...

`newPlatform.current` is also a useRef, you don't have to dependency track it -- the linter should know better.

frontend/packages/modelServing/src/components/projectDetails/SelectPlatformView.tsx
```diff
@@ -0,0 +1,106 @@
...
+                      <Content component={ContentVariants.h2}>
+                        {p.properties.enableCardText.title}
+                      </Content>
+                    </CardTitle>
+                    <CardBody>{p.properties.enableCardText.description}</CardBody>
+                    <CardFooter>
+                      <Bullseye>
+                        <Button
+                          variant="secondary"
+                          onClick={async () => {
```
andrewballantyne, line 75: Async?

frontend/packages/modelServing/src/extension-points/index.ts
```diff
@@ -1,11 +1,43 @@
...
+      title: string;
+      description: string;
+      selectText: string;
+      enabledText: string;
+    };
+    deployedModelsView: {
+      startHintTitle: string;
+      startHintDescription: string;
+      deployButtonText: string;
+    };
```
andrewballantyne, line 25: This seems unruly -- Curious why not separate extensions -- the overlapping "id" need?

frontend/packages/modelServing/src/components/projectDetails/NoModelsView.tsx
```diff
@@ -0,0 +1,21 @@
...
+    iconImage={typedEmptyImage(ProjectObjectType.modelServer)}
+    // imageAlt={isProjectModelMesh ? 'No model servers' : 'No deployed models'}
+    title={platform.properties.deployedModelsView.startHintTitle}
+    description={platform.properties.deployedModelsView.startHintDescription}
+    createButton={<Button>{platform.properties.deployedModelsView.deployButtonText}</Button>}
+    // footerExtraChildren={
+    //   deployingFromRegistry stuff
+    // }
+  />
+);
```
andrewballantyne, line 21: Hmmm... I think we'll need to consider what the goals are for shared components like this... 

I wonder if these should be extensions consumed by the Project-area and then passed in... seems a bit funky if you want to do something special with your empty state.

Ideally we don't import the component from the base into it like this. Maybe we need to have some idea of shared components.

@christianvogt do you have thoughts on how to manage this?

andrewballantyne (changes requested):
Some reviews on the code -- let me know if you want to talk more. I have not ran it fwiw

frontend/packages/modelServing/src/components/projectDetails/ModelsProjectDetailsTab.tsx
```diff
@@ -0,0 +1,88 @@
...
+  );
+};
+
+const WithContext: React.FC = () => (
+  <ModelServingProvider>
+    <ProjectModelsProvider>
+      <ModelsProjectDetailsTab />
+    </ProjectModelsProvider>
+  </ModelServingProvider>
+);
```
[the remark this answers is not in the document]
emilys314: I wanted the 2 different ones to separate project specific things, vs odh wide model serving stuff. 

So deploying models would be for the projects one. An admin enabling and disabling platforms or configuring cluster wide settings would be the the ModelServing one

frontend/src/pages/modelServing/ModelServingContext.tsx
```diff
@@ -75,7 +75,11 @@ const ModelServingContextProvider = conditionalArea<ModelServingContextProviderP
   const { dashboardNamespace } = useDashboardNamespace();
   const navigate = useNavigate();
   const { projects, preferredProject } = React.useContext(ProjectsContext);
-  const project = projects.find(byName(namespace)) ?? null;
+  const project = React.useMemo(
+    // memoize so project data changes rerender (when a serving platform changes)
+    () => projects.find(byName(namespace)) ?? null,
+    [namespace, projects],
+  );
```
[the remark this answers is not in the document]
andrewballantyne: Wait -- we are mutating projects in the `ProjectContext`? 😨 That's hugely sounding like a horrible pattern... 

Projects should 100% be filtered on each render if things changed -- which means you don't need a memo, because the individual Projects are not changing references if they are not changing data -- the list per se may never be stable 🤔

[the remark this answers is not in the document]
emilys314: > Wait -- we are mutating projects in the ProjectContext? 😨 That's hugely sounding like a horrible pattern...
I don't think we mutate the project as consumers of this variable. I meant the nested values in `const { projects } = React.useContext(ProjectsContext);` are getting changed from the websocket. 

Sorry, I'm not fully sure what you mean in the 2nd part of your comment. If the only thing that changes in this context is a nested property in the `project` object, react isn't triggering a rerender, which causes. 
^ that's the more immediate explanation fix

Isn't the `const project = projects.find` breaking the react dependency chain? if `projects` is something like `[project1_objref1, project2_objref1]`, so `project` would become the ref `project1_objref1`. If a consumer is directly using `projects` then react will notify the state has been updated. But `project` is just a js obj so only if the ref is changed, then the consumer will rerender. In this case the ref isn't being changed when nested properties are changing. I may be missing something though

frontend/packages/modelServing/src/components/projectDetails/ModelsProjectDetailsTab.tsx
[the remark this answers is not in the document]
emilys314: The 2 contexts will mostly be together, not always. 

For example if an admin is changing the default deployment mode of kserve, this only requires the `ModelServingProvider`. There are no project specific operations here. Or if an admin is entering an API key for NIM, this is stored on the cluster. None of this is specific to one cluster.

frontend/src/pages/modelServing/ModelServingContext.tsx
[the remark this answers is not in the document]
andrewballantyne: I think you're casually mentioning a very serious issue with data management in our Project Context if I am following what you're saying lol.

`project1_objref1` only exists until it gets updated -- never should an existing reference's data be updated. That's bad reactive coding 😞 

Maybe we need a bug to chase this down 🤔

frontend/packages/modelServing/src/components/projectDetails/ModelsProjectDetailsTab.tsx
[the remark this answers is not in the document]
andrewballantyne: 🤔 Contexts only really have good value storing data for multiple interactions / navigations / sharing between components -- unless I am missing something, the contexts in those use-cases are for 1 dropdown, 1 text field, and never reused by anyone else without unloading the context first (navigating)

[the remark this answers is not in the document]
emilys314: That's fair, yeah for that to have value, then it would have to be a top level context. I'll remove it / combine the 2

frontend/src/pages/modelServing/ModelServingContext.tsx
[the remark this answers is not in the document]
emilys314: Yeah I think this is a bad react data management issue

frontend/packages/modelServing/src/components/projectDetails/ModelsProjectDetailsTab.tsx
```diff
@@ -0,0 +1,85 @@
...
+  const activePlatform = React.useMemo(
+    () =>
+      availablePlatforms && availablePlatforms.length === 1
+        ? availablePlatforms[0]
+        : projectPlatform,
+    [availablePlatforms, projectPlatform],
+  );
+
+  return (
+    <>
```
Griffin-Sullivan, line 34: Do you need this to wrap?

frontend/packages/modelServing/src/components/projectDetails/ModelsProjectDetailsTab.tsx
```diff
@@ -0,0 +1,85 @@
...
+          !isLoading && (
+            <SelectPlatformView
+              platforms={availablePlatforms}
+              setModelServingPlatform={setModelServingPlatform}
+            />
+          )
+        }
+      >
+        {activePlatform &&
+          (hasModels ? <>View models table</> : <NoModelsView platform={activePlatform} />)}
```
Griffin-Sullivan, line 73: I assume the view models table here is temporary?

frontend/packages/modelServing/src/components/projectDetails/ModelsProjectDetailsTab.tsx
```diff
@@ -0,0 +1,85 @@
...
+      </DetailsSection>
+    </>
+  );
+};
+
+const WithContext: React.FC = () => (
+  <ModelServingProvider>
+    <ModelsProjectDetailsTab />
+  </ModelServingProvider>
+);
```
Griffin-Sullivan, line 83: What is this for?

frontend/packages/modelServing/src/components/projectDetails/ModelsProjectDetailsTab.tsx
```diff
@@ -0,0 +1,85 @@
+import React from 'react';
+// eslint-disable-next-line import/no-extraneous-dependencies
```
Griffin-Sullivan, line 2: You should add a lint rule to ignore everything

frontend/packages/modelServing/src/extension-points/index.ts
```diff
@@ -1,11 +1,43 @@
...
   }
 >;
-export const isModelServingPlatform = (extension: Extension): extension is ModelServingPlatform =>
-  extension.type === 'model-serving.platform';
+export const isModelServingPlatformExtension = (
+  extension: Extension,
+): extension is ModelServingPlatformExtension => extension.type === 'model-serving.platform';
+
+// export type ModelServingPlatformCard = Extension<
```
Griffin-Sullivan, line 32: Why is this commented?

frontend/packages/modelServing/src/components/projectDetails/ModelsProjectDetailsTab.tsx
emilys314: Yep

frontend/packages/modelServing/src/components/projectDetails/ModelsProjectDetailsTab.tsx
emilys314: It's so when using the ` <ModelsProjectDetailsTab />` that the context will be defined `= React.useContext(ModelServingContext);`. If the component isn't a child of the provider, it will be returning undefined values

Griffin-Sullivan (commented):
/lgtm

frontend/packages/modelServing/src/components/projectDetails/ModelsProjectDetailsTab.tsx
emilys314: Fixed

frontend/packages/kserve/src/extensions.ts
```diff
@@ -0,0 +1,52 @@
...
+    properties: {
+      id: 'modelmesh-dummy',
+      name: 'ModelMesh (dummy)',
+      manage: () =>
+        import('./managePlatform').then(() => ({
+          isInstalled: () => Promise.resolve(true),
+          isEnabled: () => false,
+          enable: () => Promise.resolve(''),
+          disable: () => Promise.resolve(''),
+        })),
```
andrewballantyne, line 35: If you replace this with:
```
      manage: () =>
        new Promise((resolve) => {
          resolve({ foo: true });
        }),
```
There is no typescript error. I think we have a TypeScript problem here... this object is already invalid and I was curious if it was properly typing things... turns out no 😢 

@christianvogt FYI

frontend/packages/kserve/src/managePlatform.ts
```diff
@@ -0,0 +1,29 @@
...
+  addSupportServingPlatformProject(
+    project.metadata.name,
+    NamespaceApplicationCase.KSERVE_PROMOTION,
+  );
+
+const disable = (project: ProjectKind): Promise<string> =>
+  addSupportServingPlatformProject(
+    project.metadata.name,
+    NamespaceApplicationCase.RESET_MODEL_SERVING_PLATFORM,
+  );
```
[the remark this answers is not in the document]
andrewballantyne: @emilys314 so you did half the solution here? You breaking up the other half of the solution to a new ticket?

frontend/packages/modelServing/src/components/projectDetails/ModelsProjectDetailsTab.tsx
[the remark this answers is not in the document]
andrewballantyne: Seems you didn't break this out to a new file -- any reason why?

frontend/packages/kserve/src/managePlatform.ts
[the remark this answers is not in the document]
emilys314: So yeah idk, with the current backend api to call `addSupportServingPlatformProject`, the platform extension would have to change from the `enable: () => Promise` into a `NamespaceEnum: NamespaceApplicationCase`. I can add this for now, but that doesn't seem like the permanent solution either

That logic is pretty specific to the app, and it's hardcoded with the `/backend/src` api with the corresponding enums, so it isn't very modular. Any other implementation (like specifying the strings to check in the labels / annotations) requires updating this flow. 

When should we update this whole operation? Including the `model-mesh=false` stuff?

frontend/src/pages/modelServing/ModelServingContext.tsx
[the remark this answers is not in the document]
andrewballantyne: Do we have an issue for this?

frontend/packages/modelServing/src/concepts/modelServingPlatforms.ts
emilys314: The ref was used instead of a useState because `newPlatform` isn't being used for any UI rendering, it's an internal state variable. It would technically save some rerenders from happening

So the ref was helping with an edge case if you switch back and forth between platforms really fast. It's not very clear so I changed it to return a loader which will disable the button to prevent this instead

frontend/packages/modelServing/src/components/projectDetails/ModelsProjectDetailsTab.tsx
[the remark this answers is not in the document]
emilys314: The reason why: I wasn't sure what to name the new separate file lol and then forgot about it. Updated

frontend/packages/modelServing/src/extension-points/index.ts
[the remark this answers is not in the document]
emilys314: The `plugin-core` was having stuff under `src` so i copied that. But imo this makes more sense so updated

frontend/packages/kserve/src/extensions.ts
christianvogt: @andrewballantyne i changed the code to what you suggested and I get an error. The fact that the object includes a property named `disable` doesn't make it invalid, the type still satisfies the api contract.
image

frontend/src/pages/modelServing/ModelServingContext.tsx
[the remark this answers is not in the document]
emilys314: https://issues.redhat.com/browse/RHOAIENG-25408

frontend/packages/modelServing/src/concepts/modelServingPlatforms.ts
andrewballantyne: I get you changed this -- but for posterity...

I think using a ref like this is a weird mindset. Attempting to avoid re-renders usually makes wonky code. Just let it happen -- React's goal is to optimize out re-renders.

Besides, if you were able to get what you wanted, and you stopped the re-render, the hook would never fire off because updating a ref doesn't allow hooks to trigger as no re-render happens.

frontend/packages/kserve/src/extensions.ts
andrewballantyne: Oh -- interesting... must be an issue with my IDE or something.

andrewballantyne (approved):

PR #4170, opened 2025-05-07
add project search in model registry deploy model

pnaik1:
https://issues.redhat.com/browse/RHOAIENG-24694

## Description
Added project search to deploy model modal

https://github.com/user-attachments/assets/ca9c2222-60df-4072-9056-e35e3d9185cc

## How Has This Been Tested?

- Deploy a model either from model registry or directly from model catalog
- Verify you can now see a project search to search the project by name

## Test Impact
Updated the cypress test

## Request review criteria:

If you have UI changes: 

cc @yannnz

### DIFF
--- a/frontend/src/__tests__/cypress/cypress/pages/modelCatalog/modelCatalogDeployModal.ts
+++ b/frontend/src/__tests__/cypress/cypress/pages/modelCatalog/modelCatalogDeployModal.ts
@@ -5,13 +5,17 @@ class ModelCatalogDeployModal extends Modal {
     super('Deploy model');
   }
 
-  findProjectSelector() {
-    return this.find().findByTestId('deploy-model-project-selector');
+  private findProjectSelector() {
+    return this.find().findByTestId('deploy-model-project-selector-toggle');
   }
 
   selectProjectByName(name: string) {
     this.findProjectSelector().click();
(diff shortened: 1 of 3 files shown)


YuliaKrimerman:
@pnaik1 @yih-wang When the search yields no results, should we display a “No projects found” message for better UX?

yih-wang:
I don't see how it looks like from the video. But what you suggested makes sense to me @YuliaKrimerman. Lets be consistent with the project selector that is used elsewhere (screenshot for no results pasted below)
image

manaswinidas (commented):
+1 to what Yulia said , it can be either of the two. I'll leave it to @yih-wang ("No matching results" or "No results found")
Screenshot 2025-05-09 at 3 55 15 PM
Screenshot 2025-05-09 at 3 59 19 PM

yih-wang:
@manaswinidas It's interesting that we have two different messages for a similar case. cc @kaedward for review and suggestions.

manaswinidas:
"No matching results" is for Projects drop-down and "No results found" is what we use for project-scoped image streams/hardware profiles/accelerator profiles/serving runtimes. Should they all be same? Or we can just stick to "No matching results" like the projects dropdown.

PR #4175, opened 2025-05-07
Hardware guard deletion: show a dialog when deleting the *last* CPU or memory for a hardware profile

jrenee42:
for: https://issues.redhat.com/browse/RHOAIENG-22692

## Description
show a dialog when deleting the *last* CPU or memory for a hardware profile.

previously, the warning happened on the page *after* the deletion happened.  
Now; the warning is still there; but there is a confirmation dialog that pops up if and only if the user is trying to delete the last CPU or memory resource.

if there are multiple cpu or memory resources; then the deletion just happens without a dialog.

Note:  normally hardware profiles are not visible to the user; the feature flag for hiding the hardware profiles has to be unchecked. see this video: (it has audio)

https://github.com/user-attachments/assets/da2c5440-bbae-459a-9739-58d8be34a06d

## How Has This Been Tested?
tested this manually, see the video up above.  tested while editing and while creating a hardware profile.  the dialog *only* shows if the item being deleted is the only cpu or memory in the profile; the dialog does not show/get used at all if there are multiple of the cpus or memories, or if the type is not cpu or memory.

Originally, thought was to add a generic deletion dialog; but a deletion modal already exists for more destructive operations; and no where else needed this.  so is not necessary. 

## Test Impact
 no tests needed; this is very minor. manual testing is sufficient.

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/pages/hardwareProfiles/nodeResource/DeleteNodeResourceModal.tsx
+++ b/frontend/src/pages/hardwareProfiles/nodeResource/DeleteNodeResourceModal.tsx
@@ -1,6 +1,14 @@
 import * as React from 'react';
-import { Alert, Button, Stack, StackItem } from '@patternfly/react-core';
-import { Modal } from '@patternfly/react-core/deprecated';
+import {
+  Alert,
+  Button,
+  Stack,
+  StackItem,
+  Modal,
+  ModalHeader,
+  ModalBody,
+  ModalFooter,
+} from '@patternfly/react-core';
 import { Identifier } from '~/types';
 import { CPU_MEMORY_MISSING_WARNING } from '~/pages/hardwareProfiles/const';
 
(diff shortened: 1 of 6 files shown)


frontend/src/pages/hardwareProfiles/nodeResource/DeleteNodeResourceModal.tsx
```diff
@@ -0,0 +1,63 @@
+import * as React from 'react';
+import { Alert, Button, Stack, StackItem } from '@patternfly/react-core';
+import { Modal } from '@patternfly/react-core/deprecated';
```
srtanish1992, line 3: Let's not use the deprecated Modal. Replace it with the PF6 Modal, you can check out my [PR](https://github.com/opendatahub-io/odh-dashboard/pull/4168/files) on how to use the latest version.

jrenee42: done

Gkrumbach07 (approved):
/lgtm
/approve

thanks Jill!

PR #4184, opened 2025-05-08
Add preview URI links to the specific artifact types

DaoDaoNoCode:
JIRA: [RHOAIENG-11412](https://issues.redhat.com/browse/RHOAIENG-11412)

## Description

I made some changes to the Artifact URI links behavior.

For metrics type, including `system.Metrics`, `system.ClassificationMetrics`, and `system.SlicedClassificationMetrics`, we don't show the URI anymore because it doesn't link to any S3 storage, so it's not a link, and it makes no sense to show them. I made a dash for that field.

For model type, which is 'system.Model', it shows only a downloadable link, which will trigger a download when clicked on.

For all the other types, we show both icons for previewing and downloading. Clicking on the first icon will open a new tab to preview the content that's returned by the S3 storage, and clicking on the second icon will download the resource.

Screenshot 2025-05-08 at 12 33 12 PM

## How Has This Been Tested?

1. Create some pipeline runs that will generate different artifact types. (Or you can check ilab shared cluster to see the artifacts list)
2. See the corresponding URI field to different types of artifact types, make sure they are the same as described above
3. Try to click on the external link icon, see if it opens a new tab
4. Try to click on the download icon, see if it downloads the resource

## Test Impact

Modify tests according to the hook changes

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/concepts/pipelines/content/artifacts/ArtifactUriLink.tsx
+++ b/frontend/src/concepts/pipelines/content/artifacts/ArtifactUriLink.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import { Button, Flex, FlexItem, Truncate } from '@patternfly/react-core';
-import { ExternalLinkAltIcon } from '@patternfly/react-icons';
+import { DownloadIcon, ExternalLinkAltIcon } from '@patternfly/react-icons';
 import { ArtifactType } from '~/concepts/pipelines/kfTypes';
 import { useArtifactStorage } from '~/concepts/pipelines/apiHooks/useArtifactStorage';
 import { Artifact } from '~/third_party/mlmd';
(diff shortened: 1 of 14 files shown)


DaoDaoNoCode:
@yannnz Hi, could you check the screenshot above? Thanks!

frontend/src/concepts/pipelines/content/artifacts/ArtifactUriLink.tsx
```diff
@@ -11,58 +11,99 @@ interface ArtifactUriLinkProps {
...
+  const handlePreview = () => {
+    getStorageObjectRenderUrl(artifact)
+      .then((url) => {
+        if (url) {
+          window.open(url, '_blank');
+        }
+      })
+      .catch((error) => {
+        // eslint-disable-next-line no-console
+        console.error('Failed to get the storage URL:', error);
```
Gkrumbach07, line 48: do you think we should have notifications for these errors instead of silencing them

frontend/src/concepts/pipelines/content/artifacts/ArtifactUriLink.tsx
```diff
@@ -11,58 +11,99 @@ interface ArtifactUriLinkProps {
...
-      spaceItems={{ default: 'spaceItemsNone' }}
+      spaceItems={{ default: 'spaceItemsSm' }}
     >
       <FlexItem>
         <Button
           isLoading={loading}
           variant="link"
           isInline
-          icon={<ExternalLinkAltIcon />}
+          icon={isDownloadableOnly ? <DownloadIcon /> : <ExternalLinkAltIcon />}
```
Gkrumbach07, line 74: why make this button be one or the other. can you just do 
```tsx
<ButtonForDownload />
{!downloadOnly && <ButtonForPreview />}
```

frontend/src/concepts/pipelines/content/compareRuns/metricsSection/markdown/useFetchMarkdownMaps.ts
```diff
@@ -39,18 +39,27 @@ const useFetchMarkdownMaps = (
...
           const { run } = path;
           let sizeBytes: number | undefined;
-
-          const url = await getStorageObjectUrl(path.linkedArtifact.artifact).catch(
-            () => undefined,
-          );
+          let url: string | undefined;
+          if (
+            path.linkedArtifact.artifact.getType() === ArtifactType.MODEL ||
+            path.linkedArtifact.artifact.getType() === ArtifactType.HTML
```
Gkrumbach07, line 45: why is HTML download and not preview

Gkrumbach07: or should markdown be here

Gkrumbach07: actually i think only markdown should be here and everything else we wont use. This output is used as a src in a iframe and that needs to be the download url, not preview.

frontend/src/concepts/pipelines/content/pipelinesDetails/pipelineRun/artifacts/ArtifactVisualization.tsx
```diff
@@ -46,18 +46,18 @@ export const ArtifactVisualization: React.FC<ArtifactVisualizationProps> = ({ ar
     if (artifactType === ArtifactType.MARKDOWN || artifactType === ArtifactType.HTML) {
       const uri = memoizedArtifact.getUri();
       if (uri) {
-        const downloadArtifact = async () => {
-          await getStorageObjectUrl(memoizedArtifact)
-            .then((url) => setDownloadedArtifactUrl(url))
+        const renderArtifact = async () => {
+          await getStorageObjectRenderUrl(memoizedArtifact)
```
Gkrumbach07, line 50: this should be download

yannnz:
Hi @DaoDaoNoCode, does the update of removing the link apply to all metric types? Where can the users find the metrics if they are no in s3? Otherwise, the updates look good. Thanks.

frontend/src/concepts/pipelines/content/artifacts/ArtifactUriLink.tsx
DaoDaoNoCode: Yes, I will change it to a notification.

frontend/src/concepts/pipelines/content/artifacts/ArtifactUriLink.tsx
DaoDaoNoCode: The icon links to the text, and they are in the same button, so the primary action for different types is different, Thus, we need to do it in this way.

frontend/src/concepts/pipelines/content/compareRuns/metricsSection/markdown/useFetchMarkdownMaps.ts
DaoDaoNoCode: Yes, you are right. It's a mistake, so I will change the model type to the markdown type.

frontend/src/concepts/pipelines/content/pipelinesDetails/pipelineRun/artifacts/ArtifactVisualization.tsx
DaoDaoNoCode: The download URL will trigger an automatic download when you put it into an iframe, only the render URL can successfully render the HTML/Markdown resources.

DaoDaoNoCode:
@yannnz Hi, the metrics are not in S3. We are getting the metrics data from the pipeline run details and parsing them at the client side, and it doesn't need to fetch any data from the S3 storage.

frontend/src/concepts/pipelines/content/compareRuns/metricsSection/markdown/useFetchMarkdownMaps.ts
```diff
@@ -39,18 +39,27 @@ const useFetchMarkdownMaps = (
...
+          let url: string | undefined;
+          if (
+            path.linkedArtifact.artifact.getType() === ArtifactType.MARKDOWN ||
+            path.linkedArtifact.artifact.getType() === ArtifactType.HTML
+          ) {
+            url = await getStorageObjectRenderUrl(path.linkedArtifact.artifact).catch(
+              () => undefined,
+            );
+          } else {
+            url = await getStorageObjectDownloadUrl(path.linkedArtifact.artifact).catch(
```
Gkrumbach07, line 51: you dont need this part. as it will just auto download if put in a iframe

DaoDaoNoCode: Yep you are right, removed.

Gkrumbach07 (approved):

PR #4187, opened 2025-05-09
fix: clusterStorage bug during provision

FedeAlonso:
https://issues.redhat.com/browse/RHOAIENG-25189
## Description
Fix ClusterStorage provision bug found during the nightly exectution.

## How Has This Been Tested?
image
Command used:
`$ npx cypress run --browser chrome --project src/__tests__/cypress --spec src/__tests__/cypress/cypress/tests/e2e/storageClasses/clusterStorage.cy.ts`
It does not make sense to run a pipeline in this case, as for checking that the error is not happening anymore there's a setup needed
So I've tested both cases: With the setup (green) and a regular execution (green too)

## Test Impact
Total impact as it's a fix for a test.

## Request review criteria:
Run the tests locally


antowaddle (approved):
lgtm

PR #1256, opened 2023-05-18
Make the .npm directory writable by the root group

LaVLaS:
## Description
Addresses #1123 to make sure that when the container is running with a random UID in OpenShift, the `/opt/app-root/src/.npm` folder is writable by any user in the `root` group

I can reproduce this locally with a v2.9.2 image by running the container with a UID different from the `default` user ID (`1001`)
```
podman run --rm --user 65536 quay.io/opendatahub/odh-dashboard:v2.9.2 
````

## How Has This Been Tested?
Deployed into my dev cluster that was experiencing the issue with the quay.io/opendatahub/odh-dashboard:v2.9.2 image and patched in the quay.io/opendatahub/odh-dashboard:pr-1256 image with no issues

## Test Impact
No dashboard functionality was modified

## Request review criteria:
Deploy the PR container in OpenShift with `make deploy` to ensure that there are no CrashLoopErrors

- [ ] The commits have meaningful messages (squashes happen on merge by the bot).
- [ ] Included any necessary screenshots or gifs if it was a UI change.
- [ ] Included tags to the UX team if it was a UI/UX change.
- [ ] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [ ] The developer has manually tested the changes and verified that the changes work
- [ ] The developer has added tests or explained why testing cannot be added (unit tests & storybook for related changes)

PR #4190, opened 2025-05-09
Refactor dockerfile

lucferbux:
## Description

Refactor Dockerfile to optimize production dependency installation and clean up unnecessary COPY commands

## How Has This Been Tested?

1. Building it with podman
2. Create a new Openshift cluster
3. Install the operator
4. Modify the dashboard image to `quay.io/lferrnan/odh-dashboard:fix-dockerfile`

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,6 +23,10 @@ RUN npm ci --omit=optional
 
 RUN npm run build
 
+# Install only production dependencies
+# This is needed to remove the dev dependencies that were installed in the previous step
+RUN cd backend && npm prune --omit=dev
+
 FROM ${BASE_IMAGE} as runtime
 
 WORKDIR /usr/src/app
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -23,6 +23,10 @@ RUN npm ci --omit=optional
 
 RUN npm run build
 
+# Install only production dependencies
+# This is needed to remove the dev dependencies that were installed in the previous step
+RUN cd backend && npm prune --omit=dev
```
andrewballantyne, line 28: The goal is to not copy over the build tooling while building images, but post building the source?

You cut out the omit optional, cache clean, & the chmod effort. The drop of the omit optional & I guess the cache clean are done in the lines above 🤔 But what was the value of the `chmod`?

lucferbux: I'm glad that you asked, I didn't add that into the description, apologies, let me add more context.
That change was done here https://github.com/opendatahub-io/odh-dashboard/pull/1256 due to issues with the default user and openshift user's randomness.
We already fixed that back when we introduced user id's, you can check that is fixed by first running an old version:
`podman run --rm --user 65536 quay.io/opendatahub/odh-dashboard:v2.9.2 ``
Which will induce the error
And then do the exact same thing with this current build:
`podman run --rm --user 65536 quay.io/lferrnan/odh-dashboard:fix-dockerfile`

PR #4206, opened 2025-05-13
Fix Artifacts are fetched twice on the model registry version details page

YuliaKrimerman:
Closes 

## Description
This PR implements prop drilling for model artifacts data to eliminate duplicate network requests. Previously, the useModelArtifactsByVersionId hook was called twice: once in ModelVersionsDetailsHeaderActions (via useModelVersionTuningData) and once directly in ModelVersionDetailsView. This caused unnecessary network requests and potential inconsistencies in the data.
Changes made:

- Moved useModelArtifactsByVersionId hook call to the top-level ModelVersionDetails component
- Created a combined refresh function that handles both model version and artifacts updates
- Updated all child components to receive and use artifacts data via props
Updated archive-related components to maintain consistency
## How Has This Been Tested?
Verified model artifacts data is displayed correctly in all views
Confirmed refresh functionality works properly
Tested archive/restore flows
Verified lab-tune and deploy actions work as expected
Checked that all modals open and close correctly
## Test Impact
No new tests were added as this is a refactoring that maintains existing functionality
Existing tests should continue to pass as the component behavior remains the same
Note: The changes are purely architectural and don't affect the UI or user experience, hence no screenshots are included.
## Request review criteria:

If you have UI changes: 


manaswinidas (commented):
Works fine.

1. Tried archiving/restoring
2. Modals load fine
3. Deploy and LAB-tune buttons work fine
4. Verified model artifacts data displayed correctly 
5. Confirmed refresh functionality works properly

manaswinidas (approved):
existing tests should pass and this is good to go. There is a merge conflict that needs to be fixed

frontend/src/pages/modelRegistry/screens/ModelVersionDetails/ModelVersionDetailsHeaderActions.tsx
```diff
@@ -53,15 +66,79 @@ const ModelVersionsDetailsHeaderActions: React.FC<ModelVersionsDetailsHeaderActi
...
+  const { inputModelLocationUri, modelRegistryDisplayName, outputModelRegistryApiUrl } =
+    React.useMemo(() => {
+      const registryService = modelRegistryServices.find(
+        (s) => s.metadata.name === preferredModelRegistry?.metadata.name,
+      );
+      return {
+        inputModelLocationUri: modelArtifacts.items[0]?.uri,
+        modelRegistryDisplayName: registryService
+          ? getDisplayNameFromK8sResource(registryService)
+          : '',
```
mturley, line 78: It's kind of a shame to lose the encapsulation we had from the `useModelVersionTuningData` hook here, and we've now duplicated all the logic from it because it is still used elsewhere (in ModelVersionsTableRow). Can we instead change `useModelVersionTuningData` into a util `getModelVersionTuningData` that takes the artifacts as a param, and then use that here instead of duplicating its logic? We'll have to also replace the other usage of it in ModelVersionsTableRow so it relies on the useModelArtifactsByVersionId call that it already makes outside its usage of useModelVersionTuningData (looks like we had a double fetch there as well). We'll have to just pull the loaded, loadError out of that existing useModelArtifactsByVersionId call instead (the one on line 54 in ModelVersionsTableRow).

mturley (changes requested):
I think we can avoid duplicating logic here

PR #4209, opened 2025-05-14
Fix model serving service account when isvc and sr have different names

Griffin-Sullivan:
https://issues.redhat.com/browse/RHOAIENG-23800
## Description
When you have an ISVC and a SR that have different names, the token created in the kserve modal will not display. I had to revert one change from a recent PR that will need testing from the NIM team.

## How Has This Been Tested?
1. Select single model serving
2. Create an ISVC and a SR with different names
3. Go to edit the model
4. Require token authentication
5. Redeploy
6. Edit the model again
7. The service account input field should be visible

## Test Impact
Going to add tests in a separate PR

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/k8sTypes.ts
+++ b/frontend/src/k8sTypes.ts
@@ -562,6 +562,9 @@ export type InferenceServiceKind = K8sResourceCommon & {
   };
 };
 
+export const isInferenceServiceKind = (obj: K8sResourceCommon): obj is InferenceServiceKind =>
+  obj.kind === 'InferenceService';
+
 export type RoleBindingSubject = {
   kind: string;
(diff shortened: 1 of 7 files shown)


romartin:
Hey @Griffin-Sullivan ! Just one notice, since the slack conversation, I am now deploying the ISVC and a SR using the same name, but the issue still persist in our cluster..

emilys314:
I tested out the kserve flow with different names and it all works now. The normal expected UI flows with the same names still works too.

image

emilys314:
Everything with NIM seems to be working good too

/lgtm
/approve

PR #4216, opened 2025-05-15
show basic table

emilys314:
https://issues.redhat.com/browse/RHOAIENG-24563

## Description

Basic table of the models has been setup

The row rendering is in another story so.

image

## How Has This Been Tested?

1. Deploy a model with kserve
2. Toggle the extension flag and see the table

## Request review criteria:

If you have UI changes: 


frontend/packages/modelServing/extension-points/index.ts
```diff
@@ -1,6 +1,18 @@
...
 // eslint-disable-next-line import/no-extraneous-dependencies
 import { NamespaceApplicationCase } from '@odh-dashboard/internal/pages/projects/types';
+import { K8sResourceCommon } from '@openshift/dynamic-plugin-sdk-utils';
+// eslint-disable-next-line import/no-extraneous-dependencies
+import { SortableData } from '@odh-dashboard/internal/components/table/types';
+import { ModelServingPlatform } from 'concepts/modelServingPlatforms';
+
+//// Types for the model serving platform extension
+
+export type Deployment<T extends K8sResourceCommon = K8sResourceCommon> = {
```
emilys314, line 11: Should we put extensible types in the extension-points as well? I'm not sure if there is a better place

frontend/packages/modelServing/extension-points/index.ts
```diff
@@ -1,6 +1,18 @@
...
+  model: T;
```
emilys314, line 12: data type is definitely a wip

frontend/packages/modelServing/src/concepts/modelDeployments.ts
```diff
@@ -0,0 +1,35 @@
...
+import useFetchState, {
+  FetchState,
+  FetchStateCallbackPromise,
+  NotReadyError,
+} from '@odh-dashboard/internal/utilities/useFetchState';
+// eslint-disable-next-line import/no-extraneous-dependencies
+import { listInferenceService } from '@odh-dashboard/internal/api/index';
+import { Deployment } from '../../extension-points';
+
+export const useDeployedModels = (namespace: string): FetchState<Deployment[] | undefined> => {
```
emilys314, line 12: WIP: working on making this more generic

frontend/packages/modelServing/src/components/projectDetails/DeploymentsTable.tsx
```diff
@@ -0,0 +1,94 @@
...
+  TableRowTitleDescription,
+} from '@odh-dashboard/internal/components/table/index';
+// eslint-disable-next-line import/no-extraneous-dependencies
+import ResourceTr from '@odh-dashboard/internal/components/ResourceTr';
+import { ActionsColumn, Td } from '@patternfly/react-table';
+import { useResolvedExtensions } from '@odh-dashboard/plugin-core';
+import { ModelServingPlatform } from '../../concepts/modelServingPlatforms';
+import { Deployment, isModelServingDeploymentsTableExtension } from '../../../extension-points';
+
+const genericColumns: SortableData<Deployment>[] = [
```
Griffin-Sullivan, line 15: Might be worth adding inference endpoint? I feel like we can expect every deployment to have this

frontend/src/pages/projects/ProjectDetailsContext.tsx
```diff
@@ -69,10 +69,7 @@ const ProjectDetailsContextProvider: React.FC = () => {
   const { dashboardNamespace } = useDashboardNamespace();
   const { namespace } = useParams<{ namespace: string }>();
   const { projects } = React.useContext(ProjectsContext);
-  const project = React.useMemo(
-    () => projects.find(byName(namespace)) ?? null,
-    [namespace, projects],
-  );
+  const project = projects.find(byName(namespace)) ?? null;
```
Griffin-Sullivan, line 72: Why not useMemo?

emilys314: Turns out it's not needed. I had added the useMemo in the previous PR

frontend/packages/kserve/extensions.ts
```diff
@@ -27,28 +33,36 @@ const extensions: ModelServingPlatformExtension[] = [
...
+  //       selectText: 'Select multi-model',
+  //       enabledText: 'Multi-model serving enabled',
+  //     },
+  //     deployedModelsView: {
+  //       startHintTitle: 'Start by deploying a model',
+  //       startHintDescription: 'Each model is deployed on its own model server',
+  //       deployButtonText: 'Deploy model',
+  //     },
+  //   },
+  // },
```
Griffin-Sullivan, line 65: Do we want to rip this out of the PR?

frontend/packages/modelServing/src/concepts/ModelServingContext.tsx
```diff
@@ -44,7 +45,7 @@ export const ModelServingProvider: React.FC<ModelServingProviderProps> = ({ chil
     activePlatformError,
   } = useActiveServingPlatform(currentProject, availablePlatforms);
 
-  const [deployedModels] = React.useState<string[]>([]);
+  const [deployedModels] = useDeployedModels(currentProject, activePlatform);
```
emilys314, line 48: Going to make this a hook that the model serving provides. 

But the challenge is the platforms are possibly not defined in here. So calling a hook conditionally won't work. 

I think this requires another context that will be used and rendered only when the platform is fully defined.

frontend/packages/kserve/extensions.ts
emilys314: We probably should, it's not a good habit to include this. But it's nice and convenient lol. I'll remove it

Griffin-Sullivan:
Maybe we should setup a github team for zaffre?

frontend/packages/modelServing/extension-points/index.ts
```diff
@@ -11,6 +33,14 @@ export type ModelServingPlatformExtension = Extension<
       enabledLabel: string;
       enabledLabelValue: string;
     };
+    deployments: {
+      watch: CodeRef<
+        (
+          project: ProjectKind,
+          opts?: K8sAPIOptions,
+        ) => [Deployment[] | undefined, boolean, Error | undefined]
```
christianvogt, line 41: Why not make the result always an array?
```suggestion
        ) => [Deployment[], boolean, Error | undefined]
```

frontend/packages/kserve/extensions.ts
```diff
@@ -27,26 +35,11 @@ const extensions: ModelServingPlatformExtension[] = [
...
-        selectText: 'Select multi-model',
-        enabledText: 'Multi-model serving enabled',
-      },
-      deployedModelsView: {
-        startHintTitle: 'Start by deploying a model',
-        startHintDescription: 'Each model is deployed on its own model server',
-        deployButtonText: 'Deploy model',
-      },
+      platform: KSERVE_ID,
+      columns: () => import('./src/deploymentsTable').then((m) => m.columns),
```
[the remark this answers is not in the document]
emilys314: so columns is of type `SortableData[]` which contains an optional `sortable: boolean | ((a: T, b: T, keyField: string) => number);` so in theory it's not just json, but it can contain JS code. I think this would only happen if you explicitly added it, so as of right now I don't think it has to be a code ref

I'll look into putting the renderer in the columns data structure though

frontend/packages/modelServing/extension-points/index.ts
```diff
@@ -1,6 +1,28 @@
...
+// eslint-disable-next-line import/no-extraneous-dependencies
+import { SortableData } from '@odh-dashboard/internal/components/table/types';
+// eslint-disable-next-line import/no-extraneous-dependencies
+import { K8sAPIOptions, ProjectKind } from '@odh-dashboard/internal/k8sTypes';
+
+//// Types for the model serving platform extension
+
+export type Deployment<
+  M extends K8sResourceCommon = K8sResourceCommon,
+  S extends K8sResourceCommon = K8sResourceCommon,
```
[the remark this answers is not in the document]
emilys314: We can't assume a specific kind of resource kind for future model serving platforms `InferenceService` and `ServingRuntime` are specific to kserve based ones. While we have only had model serving platforms based off kserve, it seems unlikely that it would only be kserve based. For example llm-d has `InferencePool` and `ModelService`, and no `InferenceService` or kserve specific ones. And llama stack is probably also very different

frontend/packages/kserve/extensions.ts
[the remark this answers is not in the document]
emilys314: Updated

PR #4221, opened 2025-05-16
Fix for default character limit message

ashley-o0o:
Closes: [RHOAIENG-24109](https://issues.redhat.com/browse/RHOAIENG-24109)

## Description
Removes the static character limit message and updates it to follow the existing pattern. Now when the user exceeds the character limit, the message will pop up with a warning icon, and the create run button will be disabled. 

## How Has This Been Tested?
Tested locally

- Navigate to Data science projects > Runs
- Click create run 
- Scroll to Run details, and put a name or description that’s larger than 255 characters to see the warning.

## Test Impact
No tests changed

## Screenshots
Before:
Screenshot 2025-05-14 at 9 43 32 AM

After:

https://github.com/user-attachments/assets/67fe8319-b538-41b8-8115-cfd56b57d195 

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/concepts/k8s/NameDescriptionField.tsx
+++ b/frontend/src/concepts/k8s/NameDescriptionField.tsx
@@ -31,6 +31,7 @@ type NameDescriptionFieldProps = {
   nameHelperText?: React.ReactNode;
   hasNameError?: boolean;
   onNameChange?: (value: string) => void;
+  onValidationChange?: (isValid: boolean) => void;
 };
 
 /**
@@ -50,9 +51,9 @@ const NameDescriptionField: React.FC<NameDescriptionFieldProps> = ({
   disableK8sName,
   maxLengthName,
   maxLengthDesc,
-  nameHelperText,
   hasNameError,
   onNameChange,
+  onValidationChange,
 }) => {
   const k8sName = React.useMemo(() => {
     if (showK8sName) {
@@ -61,6 +62,12 @@ const NameDescriptionField: React.FC<NameDescriptionFieldProps> = ({
 
     return '';
   }, [showK8sName, data.name]);
+
+  React.useEffect(() => {
+    const isNameValid = !hasNameError && (!maxLengthName || data.name.length <= maxLengthName);
+    const isDescValid = !maxLengthDesc || data.description.length <= maxLengthDesc;
+    onValidationChange?.(isNameValid && isDescValid);
+  }, [data.name, data.description, maxLengthName, maxLengthDesc, hasNameError, onValidationChange]);
 
   return (
     <Stack hasGutter>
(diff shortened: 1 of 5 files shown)


katieperry4:
When both fields are 254 characters or less everything is still enabled:
Screenshot 2025-05-19 at 3 32 24 PM

But when either/both have 255+ characters the `create run` button is disabled and the errors show correctly:
Screenshot 2025-05-19 at 3 32 31 PM
Screenshot 2025-05-19 at 3 32 39 PM
/lgtm

frontend/src/pages/projects/types.ts
```diff
@@ -10,9 +10,9 @@ export type UpdateObjectAtPropAndValue<T> = <K extends keyof T>(
 ) => void;
 
 export type NameDescType = {
-  name: string;
+  name?: string;
   k8sName?: string;
-  description: string;
+  description?: string;
```
Griffin-Sullivan, line 15: Why make these optional?

frontend/src/concepts/k8s/NameDescriptionField.tsx
```diff
@@ -50,18 +51,26 @@ const NameDescriptionField: React.FC<NameDescriptionFieldProps> = ({
...
+      return translateDisplayNameForK8s(data.name ?? '');
     }
 
     return '';
   }, [showK8sName, data.name]);
 
+  React.useEffect(() => {
+    const isNameValid =
+      !hasNameError && (!maxLengthName || (data.name ?? '').length <= maxLengthName);
+    const isDescValid = !maxLengthDesc || (data.description ?? '').length <= maxLengthDesc;
```
Griffin-Sullivan, line 70: I would expect these to use `useMemo` and then reuse this variables in all of the places below

Griffin-Sullivan (changes requested):

frontend/src/concepts/k8s/NameDescriptionField.tsx
emilys314: Yeah a `useEffect` should almost always be avoided unless there are network calls happening

emilys314:
Waitttt the `NameDescriptionField` component the pipelines page is using is deprecated. That's why the behaviour is inconsistent. 

We should replace it with the new `K8sNameDescriptionField.tsx`, it should fix the issues too I think

frontend/src/pages/projects/types.ts
ashley-o0o: Should be fixed now. They were causing an error, but it was because I was using a deprecated component (Emily mentioned below). Updating to the correct component solved the issue I was having.

frontend/src/concepts/k8s/K8sNameDescriptionField/K8sNameDescriptionField.tsx
```diff
@@ -53,14 +57,27 @@ const K8sNameDescriptionField: React.FC<K8sNameDescriptionFieldProps> = ({
...
   nameHelperText,
+  maxLengthName,
+  maxLengthDesc,
   hideDescription,
 }) => {
   const [showK8sField, setShowK8sField] = React.useState(false);
 
   const { name, description, k8sName } = data;
 
+  const isValid = React.useMemo(() => {
```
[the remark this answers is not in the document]
ashley-o0o: So fully revert the file?

frontend/src/concepts/pipelines/content/createRun/RunPage.tsx
```diff
@@ -111,11 +113,16 @@ const RunPage: React.FC<RunPageProps> = ({
     <div data-testid={testId}>
       <PageSection hasBodyWrapper={false} isFilled>
         <GenericSidebar sections={jumpToSections} titles={runPageSectionTitles} maxWidth={175}>
-          <RunForm isDuplicated={!!duplicateRun} data={formData} onValueChange={onValueChange} />
+          <RunForm
+            isDuplicated={!!duplicateRun}
+            data={formData}
```
emilys314, line 118: `formData` doesn't contain a `k8sName.value` property, so the "Resource name" field isn't doing anything I believe? I'm not sure if a pipeline run even uses a k8s name, I can't find any references. Maybe @Gkrumbach07 would know more about this.

We may want to either hide the k8s name completely or somehow associate the `k8sName.value` to a property?

image
image

emilys314: Also duplicating it will lock in a resource name, when it should be editable. Because it's a new resource

image

emilys314 (commented):
Sorry this is feeling quite complicated

PR #4222, opened 2025-05-18
testManifestLinks: Validate multiple HTTP status code responses

manosnoam:
https://issues.redhat.com/browse/RHOAIENG-9235

## Description
Add support for common successful responses (201, 202, 204), redirects (301, 302, 307, 308), and rate limiting (429) in testManifestLinks.cy.ts.

## How Has This Been Tested?
In Jenkins and locally, the test duration is now 15 - 30 seconds, while previously it was 40 - 100 seconds:

image

## Test Impact
E2E test `dashboardNavigation/testManifestLinks.cy.ts`

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/__tests__/cypress/cypress.config.ts
+++ b/frontend/src/__tests__/cypress/cypress.config.ts
@@ -1,5 +1,5 @@
-import path from 'path';
-import fs from 'fs';
+import * as path from 'path';
+import * as fs from 'fs';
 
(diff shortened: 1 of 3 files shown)


FedeAlonso:
This PR adds a little bit of complexity. I think that there should be an easier way to support different HTTP codes. Also a 429 is not a valid code as it could be masking a 404 (when the resource does not exist) if the 429 is at a domain level.

PR #4237, opened 2025-05-20
easier to use feature flags via a button in the header

jrenee42:
For: https://issues.redhat.com/browse/RHOAIENG-25405

## Description
Feature Flags are frustrating to use; you have to add a parameter to the url and remember to format it *just right* (do you have any other params? did you spell it right?).

now, can just access it via a button on the header.

the banner *still works*.  but the banner only pops up if the user puts in the parameter in the url.  if only the new button is used, the banner never pops up.

and the button is *only* present in dev mode.
I also added tooltips to most of the buttons.  not to the light/dark mode buttons; as adding tooltips there changed how they looked. (i'm new here, i don't know what everything means yet)

demo:

https://github.com/user-attachments/assets/ad1e5719-f023-4819-b9f7-2858e9ae4c34

## How Has This Been Tested?

I tested this in my local dev environment, and changed the 'app_env' variable to 'development2' and then the the flag button did NOT show up.

I found a bug in which when you revert the flags back to the default; sometimes the ui doesn't pick up on it immediately (the hardware profiles/accelerator profiles swapping, for example).  but i was able to repro the same bug in main with the. banner; so this is an existing issue.  and it's just that the ui doesn't always immediately read the flags; the flags are being removed from localstorage. (this is a syncing error).

the change is mostly just a button and menu to call up existing functionality, no new functionality or code is written, it's just refactored so that both the banner *and* the button can use it.

to test, click on the new button and turn flags on and off; and also reset them to defaults too.

## Test Impact

I added a new flag to store in localstorage so that the banner comes up *only* when explicitly asked for via the param.  (without this, when you change any flag then the banner shows).  so adjusted jest tests to add in that flag.  I initially had an error in the refactoring that the tests even caught!  jest tests are passing now.

## Request review criteria:

If you have UI changes: 

## Summary by CodeRabbit

- **New Features**
  - Introduced a Feature Flag Launcher in the header for managing feature flags via a dropdown and modal interface (development mode only).
  - Added tooltips to header toolbar elements for improved accessibility.
  - Banner visibility for developer feature flags is now controlled and reflected in the UI.

- **Bug Fixes**
  - Updated class names and import paths for consistency and maintainability.

- **Refactor**
  - Extended and unified types for feature flag-related components and props.
  - Simplified and consolidated modal logic for editing feature flags.

- **Tests**
  - Added comprehensive tests for the Feature Flag Launcher component.
  - Updated and improved tests for feature flag hooks and banners.


frontend/src/app/FeatureFlags/FeatureFlagLauncher.tsx
```diff
@@ -0,0 +1,86 @@
...
+  const [isOpen, setIsOpen] = React.useState(false);
+  const [isModalOpen, setModalOpen] = React.useState(false);
+
+  const toggle = (
+    <MenuToggle
+      aria-label="Feature Flag Launcher"
+      variant="plain"
+      onClick={() => setIsOpen(!isOpen)}
+      isExpanded={isOpen}
+      style={{ width: 'auto' }}
```
rsun19, line 28: Is the width necessary? When I removed the style it didn't look like it made a difference.

frontend/src/app/FeatureFlags/FeatureFlagLauncher.tsx
```diff
@@ -0,0 +1,86 @@
...
+  return (
+    <>
+      <Dropdown
+        isOpen={isOpen}
+        onOpenChange={(isOpenChange) => setIsOpen(isOpenChange)}
+        onSelect={() => setIsOpen(false)}
+        toggle={(toggleRef) => (
+          <Tooltip content="Feature Flags" triggerRef={toggleRef} position="bottom">
+            {React.cloneElement(toggle, { ref: toggleRef })}
+          </Tooltip>
```
rsun19, line 43: Is there a reason why we're cloning the element as opposed to directly putting it inside?

frontend/src/app/App.tsx
```diff
@@ -28,7 +28,7 @@ import { AccessReviewProvider } from '~/concepts/userSSAR';
 import { ExtensibilityContextProvider } from '~/plugins/ExtensibilityContext';
 import useFetchDscStatus from '~/concepts/areas/useFetchDscStatus';
 import { OdhPlatformType } from '~/types';
-import useDevFeatureFlags from './useDevFeatureFlags';
+import useDevFeatureFlags from './FeatureFlags/useDevFeatureFlags';
```
DaoDaoNoCode, line 31: Please follow the camel-case convention of the folder name, so here the folder name should be `featureFlags` starting with a lower-case `f`.

frontend/src/app/FeatureFlags/FeatureFlagLauncher.tsx
jrenee42: i will check.  a lot of this is cursor-generated

frontend/src/app/FeatureFlags/FeatureFlagLauncher.tsx
jrenee42: i'll check.  (was cursor-generated code)

jrenee42: cursor did this in order to attach the ref to the already-declared component.  
instead i removed the initial declaration and just put it inline.

frontend/src/app/featureFlags/FeatureFlagModal.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+import { allFeatureFlags } from '~/concepts/areas/const';
+import { isFeatureFlag } from '~/concepts/areas/utils';
+import { FeatureFlagLauncherProps } from '~/app/featureFlags/FeatureFlagLauncher';
+
+type Props = FeatureFlagLauncherProps & { onClose: () => void };
+
+const FeatureFlagModal: React.FC<Props> = ({
+  dashboardConfig,
+  devFeatureFlags = {},
```
DaoDaoNoCode, line 21: Since you are setting it below, I think there is no need to set the value here? 🤔

DaoDaoNoCode (commented):
Left some small comments regarding to code, I will test more tomorrow.

frontend/src/app/App.tsx
```diff
@@ -133,7 +133,12 @@ const App: React.FC = () => {
                 isManagedSidebar
                 isContentFilled
                 masthead={
-                  <Header onNotificationsClick={() => setNotificationsOpen(!notificationsOpen)} />
+                  <Header
+                    dashboardConfig={dashboardConfig.spec.dashboardConfig}
+                    {...devFeatureFlagsProps}
+                    devFeatureFlags={devFeatureFlagsProps.devFeatureFlags ?? {}}
```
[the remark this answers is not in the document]
jrenee42: cursor ...... (getting used to cursor and the code.  good question!)

PR #4238, opened 2025-05-20
'valueFrom' env vars bug fix

katieperry4:
Closes [RHOAIENG-15595](https://issues.redhat.com/browse/RHOAIENG-15595)

## Description

If you had an env variable that pointed to a `Secret` or `ConfigMap` using `valueFrom` then the env value shown in the model deployment would be empty. Upon trying to edit that empty value it would add `value` next to  `valueFrom` and break the whole thing.
Per the ticket, the input now shows the `valueFrom` but is disabled to prevent editing; with a tooltip that explains the value can be edited from the OCP.
Screenshot 2025-05-20 at 3 44 14 PM

**Previously:**
Screenshot 2025-05-20 at 3 44 21 PM

**Now:**
Screenshot 2025-05-20 at 3 56 11 PM

**The tooltip:**
Screenshot 2025-05-20 at 3 56 17 PM

## How Has This Been Tested?

Tested locally and with `npm run test`

**To test this:**
- Go to deploy a kserve model.
- Add a regular env variable. Ex key: `test-variable-key` value: `test-variable-value`.
- Deploy the model.
- Find the deployed inference service in OCP.
- Go to the yaml.
- in `spec.predictor.model.env` add:
```
        - name: console-variable-update
          valueFrom:
            secretKeyRef:
              key: fake-secret-key
              name: fake-secret
```
- Hit save.
- Go back to ODH (might need to refresh the page) and edit the model.
- Find the 'additional environment variables' section and notice that the new envVar is there.
- The key input should be enabled and can be edited but the value input should be disabled and populated with the `valueFrom` with a tooltip showing the content of the input and explaining the value can only be edited from OCP.

## Test Impact

Added unit and cypress tests

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/concepts/connectionTypes/utils.ts
+++ b/frontend/src/concepts/connectionTypes/utils.ts
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { KnownLabels, SecretKind, ServingContainer } from '~/k8sTypes';
+import { KnownLabels, SecretKind } from '~/k8sTypes';
 import { getDisplayNameFromK8sResource, translateDisplayNameForK8s } from '~/concepts/k8s/utils';
 import { K8sNameDescriptionFieldData } from '~/concepts/k8s/K8sNameDescriptionField/types';
 import {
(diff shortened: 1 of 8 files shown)


frontend/src/concepts/connectionTypes/utils.ts
```diff
@@ -446,6 +446,9 @@ export const validateEnvVarName = (name: string): string | undefined => {
   return undefined;
 };
 
+export const isValueFromEnvVar = (envVar: NonNullable<ServingContainer['env']>[number]): boolean =>
```
emilys314, line 449: This doesn't belong in `connectionTypes/utils.ts`. 

(Also there does exist a `EnvironmentVariable` type that could be used to make things simpler. But what currently exists does work too)

andrewballantyne:
PO 👍

frontend/src/concepts/connectionTypes/utils.ts
katieperry4: I put it in there because
```
export const validateEnvVarName = (name: string): string | undefined => {
  if (!name) {
    return undefined;
  }
  if (STARTS_WITH_DIGIT_REGEX.test(name)) {
    return 'Must not start with a digit.';
  }
  if (!VALID_ENV_VARNAME_REGEX.test(name)) {
    return "Must consist of alphabetic characters, digits, '_', '-', or '.'";
  }
  return undefined;
};
```
is right above it in that `util` file 😅 but I'll look at moving it and look into that type too

emilys314: Ah interesting haha. The `validateEnvVarName` could probably be put in a better location too then

katieperry4: they're moved! I didn't end up using the `EnvironmentVariable` type since the `valueFrom` structure is different from what the env section is using:
```
export type EnvironmentVariable = EitherNotBoth<
  { value: string | number },
  { valueFrom: Record<string, unknown> }
> & {
  name: string;
};
```
vs
```
servingRuntimeEnvVars?: {
    name: string;
    value?: string;
    valueFrom?: {
        secretKeyRef?: {
            name: string;
            key: string;
        };
    };
}[] | undefined
```

I could try and convert them but not sure it's worth it for a one line check to see if `valueFrom` is there or not?

emilys314: sounds good!

emilys314:
looks good!
/lgtm
/approve

image

PR #4241, opened 2025-05-21
test : Fix multiModelAdminCreation

FedeAlonso:
https://issues.redhat.com/browse/RHOAIENG-26100

## Description
In this PR there's a fix for the wrong detection of the model served spinner in the e2e tests
Also, we've added some logs in order to try to debug what happens when (sometimes) the response from the modelServingService is not well formed (which makes the test to fail randomly)

## How Has This Been Tested?
Locally against the e2e cluster

## Test Impact
Test fix

## Request review criteria:

If you have UI changes: 


manosnoam (approved):
/lgtm

PR #4242, opened 2025-05-21
Add min/max selector to advanced deploy model

ashley-o0o:
Closes [RHOAIENG-24712](https://issues.redhat.com/browse/RHOAIENG-24712)

## Description
Changes the Advanced deployment option to require a minimum and maximum number of model servers to deploy. 
The numbers are capped at 99, and the inputs won't allow the min to be higher than the max. 

## How Has This Been Tested?
To test this:

- Go to your project and to models > deploy model
- Scroll to deployment mode and ensure 'Advanced' is selected 
- Modify the min and max numbers and deploy the model. The numbers can be equal, but min can't be higher than max and vice versa 
- Go to your OpenShift console, and select your project
- Under your project, search for 'InferenceServices' and open the yaml
- Under `spec:`, you should see your maxReplicas and minReplicas.

## Test Impact
Added additional cypress tests to check the functionality of the advanced selectors.

## Screenshots
Before:
Screenshot 2025-05-21 at 9 49 26 AM

After:

https://github.com/user-attachments/assets/33284ca0-c2ff-413f-b1e0-db0ff9e7eeef

- - - -
Screenshot 2025-05-21 at 9 47 01 AM

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/components/ReplicaSection.tsx
+++ b/frontend/src/components/ReplicaSection.tsx
@@ -1,17 +1,21 @@
 import * as React from 'react';
-import { FormGroup, Popover, Icon } from '@patternfly/react-core';
+import { FormGroup, Popover, Icon, Flex, FlexItem } from '@patternfly/react-core';
 import { OutlinedQuestionCircleIcon } from '@patternfly/react-icons';
 import NumberInputWrapper from '~/components/NumberInputWrapper';
 import { normalizeBetween } from '~/utilities/utils';
 
 const MIN_SIZE = 0;
 const MAX_SIZE = 999;
+const ADVANCED_MAX_SIZE = 99;
 
 type ReplicaSectionProps = {
   value: number;
   onChange: (value: number) => void;
   infoContent?: string;
   isRequired?: boolean;
+  isRaw: boolean;
+  maxValue: number;
+  onMaxChange?: (value: number) => void;
 };
 
 const ReplicaSection: React.FC<ReplicaSectionProps> = ({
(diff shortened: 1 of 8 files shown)


ashley-o0o:
@kywalker-rh @vconzola Thoughts on how this looks?

katieperry4:
I ran the updated cypress tests locally and tested the changes:
Screenshot 2025-05-21 at 1 37 20 PM
Screenshot 2025-05-21 at 1 37 47 PM
All working!

I was going to ask why the `formGroup` `label` had it's own styling:
`label={<span style={{ fontWeight: 'normal' }}>Maximum replicas</span>}` 
but I tried removing that for minimum replicas and I see why 😅  it doesn't look quite right having it be bold and I can't find anything in patternfly docs about changing the weight of the label
Screenshot 2025-05-21 at 1 54 18 PM
/lgtm

andrewballantyne:
PO is happy to see this 🙂 

I can give you a better review when I get a moment between things. Wanted to come through and 👍  the visuals asap.

frontend/src/components/ReplicaSection.tsx
```diff
@@ -1,52 +1,109 @@
 import * as React from 'react';
-import { FormGroup, Popover, Icon } from '@patternfly/react-core';
+import { FormGroup, Popover, Icon, Flex, FlexItem } from '@patternfly/react-core';
 import { OutlinedQuestionCircleIcon } from '@patternfly/react-icons';
 import NumberInputWrapper from '~/components/NumberInputWrapper';
 import { normalizeBetween } from '~/utilities/utils';
 
 const MIN_SIZE = 0;
 const MAX_SIZE = 999;
+const ADVANCED_MAX_SIZE = 99;
```
emilys314, line 9: Hm this is a little funny looking, I don't think the jira explicitly talked about making the kserve raw limit 99 as well? I wonder if that would be fine to do, it may have been an oversight when making the jira. It would be nice to double check if they should be the same or different

frontend/src/components/ReplicaSection.tsx
```diff
@@ -1,52 +1,109 @@
...
 
 type ReplicaSectionProps = {
   value: number;
   onChange: (value: number) => void;
   infoContent?: string;
   isRequired?: boolean;
+  isRaw: boolean;
```
emilys314, line 16: I think it would be a good idea to avoid platform specific implementation details (especially this component is under a generic `/src/components` section. 

Would it be possible to rename this to something more generic? Like showMinMax or something

frontend/src/components/ReplicaSection.tsx
```diff
@@ -1,52 +1,109 @@
...
-        }
-      }}
-    />
-  </FormGroup>
-);
+  isRaw = true,
+  maxValue,
+  onMaxChange,
+}) => {
+  const maxLimit = isRaw ? MAX_SIZE : maxValue;
```
emilys314, line 30: Also the naming is a little confusing to follow with 2 different things being called `max...`. Would it make sense to rename `MIN_SIZE` and `MAX_SIZE` to somehting like ceiling / floor or lowerLimit / upperLimit?

frontend/src/components/ReplicaSection.tsx
emilys314: cc @andrewballantyne

andrewballantyne: Replicas don't need to be more than 99 -- min or max or just straight replicas. 

A replica is a pod -- if you _need_ 100+ pods, I think you're doing something really weird. The CLI / YAML access is always an override option if you do indeed need it.

andrewballantyne: Do make sure you test CLI usage as part of all our PRs to make sure your logic handles weirdness from that side.

ashley-o0o: I'll just set the max for all to be 99. I think the jira only specified the max size being 99 for advanced, so I assumed the existing would stay the same. But I agree having them be the same makes more sense

emilys314:
The naming changes look nice!

Everything is working with standard and advanced deployment modes. Editing the yaml also updates the corresponding values in the UI

/lgtm
/approve

image

PR #4244, opened 2025-05-21
test: support not waiting for oc delete

FedeAlonso:
https://issues.redhat.com/browse/RHOAIENG-26127
## Description
Support not waiting to `oc delete project <projectname>` 

## How Has This Been Tested?
Locally:
image

## Test Impact
Full
Jenkins job: job/dashboard-tests/1615

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/__tests__/cypress/cypress/utils/projectChecker.ts
+++ b/frontend/src/__tests__/cypress/cypress/utils/projectChecker.ts
@@ -20,7 +20,7 @@ export const createCleanProject = (projectName: string): void => {
   verifyOpenShiftProjectExists(projectName).then((exists) => {
     if (exists) {
       cy.log(`Project ${projectName} already exists. Deleting it.`);
-      deleteOpenShiftProject(projectName, { wait: false });
+      deleteOpenShiftProject(projectName, { wait: true });
     }
     cy.log(`Creating project ${projectName}`);
     createAndVerifyProject(projectName);


frontend/src/__tests__/cypress/cypress/utils/oc_commands/project.ts
```diff
@@ -31,25 +31,34 @@ export const createOpenShiftProject = (
...
+  options: { timeout?: number; wait?: boolean } = {},
 ): Cypress.Chainable<CommandLineResult> => {
-  const { timeout } = options;
-  const ocCommand = `oc delete project ${projectName}`;
-  return cy
-    .exec(ocCommand, { failOnNonZeroExit: false, ...(timeout && { timeout }) })
-    .then((result) => {
-      if (result.code !== 0) {
-        cy.log(`ERROR deleting ${projectName} Project
+  const { timeout, wait = true } = options;
```
andrewballantyne, line 43:
```suggestion
  const { timeout, wait = false } = options;
```
Why not default this to false -- waiting for OpenShift Project deletions seems a bit weird; maybe a given test would care because of some watching delete functionality -- but I can't imagine this is important and just gives a quirk to the timeout need

FedeAlonso: The `oc delete` project command has the `wait` value set to True by default, so I wanted to mimic that behavior by making the developer consciously choose to set it to False if they want to.
Also, I think setting it to False by default could have negative consequences, since I’ve only updated the cases where I know it matters (like in the test teardowns), but I haven’t reviewed every single place where that function is called.
On top of that, if at some point calling oc delete returns an error or an unexpected value, it won’t be easy to debug.

andrewballantyne: I'm not sure I agree with that assessment, but this is in your wheelhouse -- I'll retract my request.

PR #4249, opened 2025-05-22
RHOAIENG-25531: Storage Classes: Define and Store Access Mode Mappings and Configuration

srtanish1992:
**JIRA story:** https://issues.redhat.com/browse/RHOAIENG-25531

## Description

1. Created a map of provisioner and supported access modes by that provisioner. I have referenced https://docs.redhat.com/en/documentation/openshift_container_platform/4.18/html/storage/understanding-persistent-storage#pv-access-modes_understanding-persistent-storage and https://github.com/openshift/console/blob/1b1c0b8832b523f958e3a4a196677e3c227b8166/frontend/public/components/storage/shared.ts#L33. 
2. Create default access mode settings for new storage class/config corrupted/config doesn't exist based on the mapping created.
3. Set all the supported access modes to false and RWO to true for default settings.

## How Has This Been Tested?

1. Go to openshift console.
2. Add a new storage class. Select `kubernetes.io/aws-ebs` provisioner. 
4. Go to the dashboard and storage classes page.
5. You should see the newly created storage class.
6. Now go to openshift console again and check the yaml file. You should see storage class config annotation and the annotation should contain a new field i.e. `accessModeSettings` with the right access modes and their values.
7. Repeat the above steps for different provisioners with different access modes combinations.

Screenshot 2025-05-22 at 5 00 41 PM

## Test Impact

Added unit tests for new utility functions.

## Request review criteria:

If you have UI changes: 


frontend/src/pages/storageClasses/utils.ts
```diff
@@ -37,3 +38,38 @@ export const isValidConfigValue = (
...
+): AccessMode[] => {
+  const provisionerString = String(provisionerParameter);
+  if (isStorageProvisioner(provisionerString)) {
+    // Here, provisionerString is confirmed to be of type StorageProvisioner (which is a string enum value)
+    // and a valid key for provisionerAccessModes.
+    return provisionerAccessModes[provisionerString];
+  }
+
+  // If it's a string not in the enum
+  return [];
```
srtanish1992, line 60: @DaoDaoNoCode @Gkrumbach07 
Can there be a case where the provisioner is invalid? For now, I am returning and empty array and that will make the `accessModeSettings` as en empty object. In this case, we can display no access modes found error message for this provisioner on the UI side. Open to suggestions.

frontend/src/pages/storageClasses/utils.ts
```diff
@@ -37,3 +38,38 @@ export const isValidConfigValue = (
...
+};
+
+export const getDefaultAccessModeSettings = (
+  supportedAccessModes: AccessMode[],
+): Partial<Record<AccessMode, boolean>> => {
+  const initialSettings: Partial<Record<AccessMode, boolean>> = {};
+  return supportedAccessModes.reduce((currentSettings, mode) => {
+    // AccessMode.ROW should be set to true, and all other supported access modes should be set to false
+    const newSetting = mode === AccessMode.RWO;
```
srtanish1992, line 69: @DaoDaoNoCode As per the latest decision in the slack thread, I am setting `RWO` as true and other modes as false.

frontend/src/pages/storageClasses/utils.ts
DaoDaoNoCode: I think you won't meet that situation. The provisioner cannot be changed to any strings

Screenshot 2025-05-22 at 5 16 48 PM

frontend/src/pages/storageClasses/utils.ts
DaoDaoNoCode: Yep, that's great 👍

frontend/src/pages/storageClasses/StorageClassesContext.tsx
DaoDaoNoCode, line 1: Do we also need to check the `accessModeSettings` in the config here? If the `config` exists, while there is no `accessModeSettings`, we also update it with the default values.

DaoDaoNoCode:
All the other parts look great! Just one comment, please check and let me know what you think. Thanks!

frontend/src/k8sTypes.ts
```diff
@@ -56,6 +57,9 @@ export type StorageClassConfig = {
   isDefault: boolean;
   lastModified: string;
   description?: string;
+  accessModeSettings: {
+    [key in AccessMode]?: boolean;
+  };
```
Gkrumbach07, line 62: small nit, use record<>

maybe this can be a linter. will bring it up to team

frontend/src/pages/storageClasses/utils.ts
Gkrumbach07: just wondering if we want to provide a UX that it is an unrecognized provider somewhere in admin edit page? @DaoDaoNoCode you can make the call on that, maybe a follow up jira after this goes out

frontend/src/pages/storageClasses/StorageClassesContext.tsx
Gkrumbach07: I dont think we do @DaoDaoNoCode, this update function is only for storage classes that lack a config annotation. The new logic all sits under `if (!config) {`

frontend/src/k8sTypes.ts
srtanish1992: implemented the change to use Record

frontend/src/k8sTypes.ts
```diff
@@ -56,6 +56,7 @@ export type StorageClassConfig = {
   isDefault: boolean;
   lastModified: string;
   description?: string;
+  accessModeSettings: Record<string, boolean>;
```
DaoDaoNoCode, line 59: Can you change it to `Record<AccessMode, boolean>`? 🤔

DaoDaoNoCode:
@srtanish1992 Can you fix the test?

srtanish1992:
> fixed lint issue

@DaoDaoNoCode i just pushed a fix for the lint error in https://github.com/opendatahub-io/odh-dashboard/pull/4249/commits/b829121e133757bb3b749b45848ee4d326954a4c

srtanish1992:
> @srtanish1992 Can you fix the test?

@DaoDaoNoCode i just pushed a fix for the lint error in https://github.com/opendatahub-io/odh-dashboard/pull/4249/commits/b829121e133757bb3b749b45848ee4d326954a4c

DaoDaoNoCode (approved):
/lgtm
Thanks!

PR #4260, opened 2025-05-27
Fix duplicate name issue with workbench images

ppadti:
Closes: [RHOAIENG-25734](https://issues.redhat.com/browse/RHOAIENG-25734)

## Description
This PR aims to fix duplicate name issue with workbench images. We have introduced an annotation to identify whether a workbench image is project-scoped or global-scoped. 
Also, minor changes around the dropdown items of search selector.

## How Has This Been Tested?
disbaleProjectScoped feature flag should be on.
1. Create same workbench image in opendatahub and any namespace in which, you are going to create a workbench.
2. Make sure the edit page and the table will show the correct label (Project-scoped/global-scoped).
3. Also, turn off the feature flag after creating a workbench with the project-scoped image, go to edit page - you will see the placeholder text in image selector and update workbench button will be disabled.

## Test Impact
Added cypress test.

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/__tests__/cypress/cypress/tests/mocked/projects/tabs/workbench.cy.ts
+++ b/frontend/src/__tests__/cypress/cypress/tests/mocked/projects/tabs/workbench.cy.ts
@@ -1035,31 +1035,33 @@ describe('Workbench page', () => {
       ],
       notebooks: [
         mockNotebookK8sResource({
-          imageDisplayName: 'Project-scoped test image',
-          displayName: 'Test project-scoped notebook',
-          lastImageSelection: 'test-10stream:1.2',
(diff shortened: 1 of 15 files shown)


frontend/src/__tests__/cypress/cypress/tests/mocked/projects/tabs/workbench.cy.ts
```diff
@@ -1317,6 +1319,128 @@ describe('Workbench page', () => {
     });
   });
 
+  it('Edit workbench with project-scoped imgaes', () => {
```
manaswinidas, line 1322: typo: imgaes -> images

manaswinidas (changes requested):
Works fine- duplicate name workbench images in different namespaces is refelcting correctly when selected while edit/create, just a small nit

PR #4254, opened 2025-05-22
Add Serving Runtime Version to Admin Serving Runtime Table

katieperry4:
Closes [RHOAIENG-26114](https://issues.redhat.com/browse/RHOAIENG-26114)

## Description

A continuation of [RHOAIENG-25675](https://issues.redhat.com/browse/RHOAIENG-25675) <- where I added the serving runtime version to the model deployment table. 
In this PR I added the runtime version to the runtime templates in the settings > serving runtimes table.

**Before:**
Screenshot 2025-05-22 at 3 14 09 PM

**After:**
Screenshot 2025-05-27 at 2 53 49 PM

## How Has This Been Tested?

Tested locally and with `npm run test`

To test this:
- Log in as a user thats allowed to see this page
- Go to settings > serving runtimes and notice that there are no version labels
- Duplicate any of the existing serving runtimes
- Now to the OCP search in the `opendatahub` namespace for `template` resources
- You can find the resource name for the templates from the popups in settings > serving runtimes
- It'll be generated, something like `template-r6vhw3` open that, go to the yaml
- Go to `objects[0].metadata.annotations`
- Add `opendatahub.io/runtime-version: v0.5.1` to the annotations in the yaml
- Go back to the dashboard
- Label should be there 👍 
- If you want to see if with the `Pre-installed` label go back to your duplicated serving runtime and in the regular `metadata.annotations` add `platform.opendatahub.io/part-of: modelcontroller`

## Test Impact

Added a cypress test

## Request review criteria:

Self checklist (all need to be checked):
- [x] The developer has manually tested the changes and verified that the changes work
- [x] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [x] The developer has added tests or explained why testing cannot be added (unit or cypress tests for related changes)

If you have UI changes: 

- [x] Included any necessary screenshots or gifs if it was a UI change.
- [ ] Included tags to the UX team if it was a UI/UX change.

After the PR is posted & before it merges:
- [ ] The developer has tested their solution on a cluster by using the image produced by the PR to `main`

PR #4279, opened 2025-05-29
Add version for serving runtime dropdown

Griffin-Sullivan:
https://issues.redhat.com/browse/RHOAIENG-25676
## Description
This adds the Runtime Version Label to the dropdown when selecting your serving runtime in both multi and single serving modals. It replaces the SimpleSelect with the SearchSelector full time (something we only did when there were project specific templates). If there are no project specific templates you'll see the normal ones under global.

## How Has This Been Tested?
Create a Serving Runtime Template and add the annotation under the Serving Runtime 
```
opendatahub.io/runtime-version: 'v0.5.1'
```
Go to deploy single and multi models
In the modal go to select a template and see the version label next to your template item
Once selected it should show the label in the selector's place holder.

## Test Impact
Added a couple mock tests
Created a new base ServingModal that is inherited by the ServingRuntimeModal and InferenceServiceModal to host anything shared between the two (in this case the template selector)
Updated tests that relied on the SimpleSelect that was removed

## Request review criteria:

If you have UI changes: 
With my changes we will now always show the search selector for serving runtimes. Before it would change behavior if you had project specific runtimes defined.
Screenshot From 2025-05-29 14-18-10
Before:
image

Here's what it looks like when you have project specific templates too: 
image

### DIFF
--- a/frontend/src/__tests__/cypress/cypress/pages/modelServing.ts
+++ b/frontend/src/__tests__/cypress/cypress/pages/modelServing.ts
@@ -116,41 +116,25 @@ class ModelServingGlobal {
 
 class ServingRuntimeGroup extends Contextual<HTMLElement> {}
 
-class InferenceServiceModal extends Modal {
-  k8sNameDescription = new K8sNameDescriptionField('inference-service');
-
-  constructor(private edit = false) {
-    super(`${edit ? 'Edit' : 'Deploy'} model`);
-  }
-
-  findConnectionType(name: string | RegExp) {
-    return this.findExistingConnectionSelect()
-      .findByRole('button', { name: 'Typeahead menu toggle' })
-      .findSelectOption(name);
-  }
-
-  findSubmitButton() {
-    return this.findFooter().findByTestId('modal-submit-button');
-  }
-
-  findModelNameInput() {
-    return this.k8sNameDescription.findDisplayNameInput();
-  }
-
-  findSpinner() {
-    return this.find().findByTestId('spinner');
-  }
-
-  findServingRuntimeSelect() {
-    return this.find().findByTestId('inference-service-model-selection');
-  }
-
+class ServingModal extends Modal {
   findServingRuntimeTemplateSearchSelector() {
     return this.find().findByTestId('serving-runtime-template-selection-toggle');
   }
 
   findServingRuntimeTemplateSearchInput() {
     return cy.findByTestId('serving-runtime-template-selection-search').find('input');
+  }
+
+  findGlobalScopedTemplateOption(name: string) {
+    return this.getGlobalScopedServingRuntime()
+      .find()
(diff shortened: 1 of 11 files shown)


Griffin-Sullivan:
Waiting on https://github.com/opendatahub-io/odh-dashboard/pull/4254 first

emilys314:
How are you creating a project scoped serving runtime? Do you have to manually create it in kubernetes? Or is it one that's created after deploying a model

Griffin-Sullivan:
You have to manually create one in openshift and specify the namespace to your project. The last screenshot here renders (without the version obviously) even on main. so the logic is if you have project connections it will show SearchSelector and not SimpleSelect at the moment.

Griffin-Sullivan:
In that case I will go ahead and transfer all of our test cases to the new component! Anyone else let me know if we are making the wrong choice here!

andrewballantyne:
@Griffin-Sullivan looking good from a PO perspective... 

However, the spacing between the name & label looks different between Project & Global entries
image

frontend/src/__tests__/cypress/cypress/pages/modelServing.ts
```diff
@@ -116,35 +116,7 @@ class ModelServingGlobal {
...
-
-  findSpinner() {
-    return this.find().findByTestId('spinner');
-  }
-
-  findServingRuntimeSelect() {
-    return this.find().findByTestId('inference-service-model-selection');
-  }
-
+class ServingModal extends Modal {
```
Griffin-Sullivan, line 119: Yeah this diff looks weird. The stuff above was not removed, but I did create a new "base" ServingModal to include anything with the template selection since it's shared in both multi-model modal and single-model modal

frontend/src/pages/modelServing/screens/projects/ServingRuntimeModal/ServingRuntimeTemplateSection.tsx
```diff
@@ -281,92 +286,69 @@ const ServingRuntimeTemplateSection: React.FC<ServingRuntimeTemplateSectionProps
...
+        isDisabled={isEditing}
+        dataTestId="serving-runtime-template-selection"
+        onSearchChange={(newValue) => setSearchServingRuntime(newValue)}
+        onSearchClear={() => setSearchServingRuntime('')}
+        searchValue={searchServingRuntime}
+        toggleContent={
+          data.servingRuntimeTemplateName ? (
+            <Flex gap={{ default: 'gapSm' }}>
+              <FlexItem>
+                {isEditing ? data.servingRuntimeTemplateName : servingRuntimeDisplayName}{' '}
```
katieperry4, line 300: is the `{' '}` necessary here with the `gap={{ default: 'gapSm' }}` it looks like it doesn't change anything when I remove it

PR #4280, opened 2025-05-29
Remove provider specification from Jupyter app manifest

Gkrumbach07:
https://issues.redhat.com/browse/RHOAIENG-25601

## Description

Sets the provider of the 'Start basic workbench' tile to an empty string. It cannot be fully removed as the CRD requires that value. Our code does a check for a defined value there so it will act as undefined when rendered.

## How Has This Been Tested?

go to explore section and make sure 'by Jupyter' is gone

## Test Impact

none - microcopy

## Request review criteria:

If you have UI changes: 


DaoDaoNoCode (approved):
/lgtm

PR #4282, opened 2025-05-29
kserve deployment status on models table row

emilys314:
Towards https://issues.redhat.com/browse/RHOAIENG-24565

## Description

Adds the status of the kserve model to the table. It should be a 1:1 equivalent of the existing code.

The common functionality such as the Status icon component and the kserve logic to determine messages has been moved into `/frontend/src/concepts/...` to hopefully make things easier when we update the model serving status for both the new and old versions.

This new code is also using a webhook for watching the pods of the inference services. This makes the status update faster, and also only one api call is done to get the entire list, instead of an individual fetch call for each model.

image
image

## How Has This Been Tested?

1. turn on the `Model Serving Plugin` flag
2. Go to a project > models tab
  - view the existing models show the correct status
  - deploy a new model and watch it update
  - (i recommend to have the old and new versions side by side to compare)

## Test Impact

No tests have been added
The existing cypress tests for all model status stuff should transfer over
It may be good to write unit tests though 

## Request review criteria:

If you have UI changes: 


frontend/packages/modelServing/src/components/projectDetails/DeploymentsTable.tsx
```diff
@@ -101,6 +60,14 @@ const DeploymentsTable: React.FC<{
     [platformColumns],
   );
 
+  if (!tableExtensionLoaded) {
+    return (
+      <Bullseye>
+        <Spinner />
+      </Bullseye>
+    );
```
emilys314, line 68: More loaders...

frontend/packages/modelServing/src/concepts/extensionUtils.ts
```diff
@@ -23,11 +23,15 @@ export const usePlatformExtension = <T extends Extension>(
 export const useResolvedPlatformExtension = <T extends Extension>(
   extensionPredicate: ExtensionPredicate<T>,
   platform: ModelServingPlatform,
-): ResolvedExtension<T> | undefined => {
-  const [resolvedExtensions] = useResolvedExtensions<T>(extensionPredicate);
+): [ResolvedExtension<T> | undefined | null, boolean, unknown[]] => {
```
emilys314, line 26: Should i update this to return `undefined` when `!loaded` and return `null` when `!resolvedExtensions.find`?

frontend/packages/kserve/src/deploymentStatus.ts
```diff
@@ -0,0 +1,30 @@
...
+  );
+  const modelPodStatus = deploymentPod ? checkModelStatus(deploymentPod) : null;
+
+  const state = modelPodStatus?.failedToSchedule
+    ? InferenceServiceModelState.FAILED_TO_LOAD
+    : getInferenceServiceModelState(inferenceService);
+
+  const message = modelPodStatus?.failedToSchedule
+    ? modelPodStatus.failureMessage || 'Insufficient resources'
+    : getInferenceServiceStatusMessage(inferenceService);
```
emilys314, line 27: maybe this stuff should be moved into the `/frontend/src/concepts/modelServingKserve/kserveStatusutils.ts` too

frontend/packages/modelServing/src/concepts/extensionUtils.ts
Griffin-Sullivan: Not sure about the null but I would expect `loaded` to stay boolean

Griffin-Sullivan: based on below then maybe null is good :man_shrugging:

Griffin-Sullivan (commented):
Do we have some unit tests we could bring over as well or do we need to write them?

Griffin-Sullivan (approved):

frontend/packages/modelServing/src/concepts/extensionUtils.ts
andrewballantyne: I think in TS, we should start making some judgement calls on if we want to talk "undefined" or "null"

- "undefined" << Nothing, through accident or omission
- "null" << Nothing, intentionally

I think `undefined` is better suited for things like `const foo = (bar: boolean, baz: number)` needing to say `baz` is optional... I'd not expect it to be null.

Likewise inversely, a response from a function should always be `null` -- or `void` (`undefined` in JS). Returning `undefined` seems like a misunderstanding of how to perceive this function response.

frontend/src/concepts/modelServingKServe/kserveStatusUtils.ts
andrewballantyne, line 1: Ummm... not sure there is an idea of "concepts" anymore for anything that becomes a plugin... once we transition, we should look at moving all of this to the package itself so it can live and breathe as an additional "part" or "package" 🙂 to the greater whole of the ODH Dashboard.

andrewballantyne (commented):
Some minor code comments @emilys314 -- sorry it was late to the party.

frontend/packages/kserve/src/api.ts
```diff
@@ -0,0 +1,65 @@
...
+} from '@odh-dashboard/internal/k8sTypes';
+import useK8sWatchResourceList from '@odh-dashboard/internal/utilities/useK8sWatchResourceList';
+import { groupVersionKind } from '@odh-dashboard/internal/api/k8sUtils';
+import {
+  InferenceServiceModel,
+  ServingRuntimeModel,
+} from '@odh-dashboard/internal/api/models/kserve';
+import { CustomWatchK8sResult } from '@odh-dashboard/internal/types';
+
+export const useWatchInferenceServices = (
```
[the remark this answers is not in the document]
emilys314: Correct polling was used before.

This helps solve issues with people complaining about status updates being slow basically. As well as general responsiveness with all the other data shown in the table

Although I'm not sure if we have any guidelines on using polling vs websocket watching - but the watch seems like a better user experience overall

frontend/packages/modelServing/src/concepts/extensionUtils.ts
emilys314: I can definitely make the changes, but I'm having a hard to with fully understanding never returning `undefined`. 

There are a good number of React hooks that will return `undefined` as a response. Like `const a = useRef(); // a === undefined`. And same with useState. (Although there is a caveat with no value is being passed in here)

Also in javascript the `Array.find(...)` will return `undefined` even if if you pass in values. `const b = ['a','b','c'].find((value) => value === 'z'); // b === undefined`

I feel that returning `undefined` follows the spirit of omission of data, the value isn't initialized yet because the calculations / async calls are still in progress. `null` is being returned when the hook says 'hey we looked at all the options, we know it's not there'.

andrewballantyne: Not sure I fully agree with your assessment.

`const a = useRef()` is both returning `{ current: undefined }` and returning what you pass to `useRef` -- that's not the same as "return undefined" as a path through your custom logic. 

`Array.find` returns `undefined` because it literally didn't find the item. It sliced the result. `array[1000]` will say "it's implicitly nothing" with `undefined`, as it's not in the content. It's not computing a result. `undefined` is omission, missing, not there... not an explicit computational result.

In the end, the idea of both wanting `undefined` and `null` as a data result seems more of a bug in the function signature, than otherwise usefulness for the caller.

We have `isLoaded` boolean -- `undefined` while that is false, and `null` when it's true and not found something -- seems like a weird JS-ism... and we can force cleaner code with TS. 

I really like JS, but most people treat undefined and null too interchangeably, and I feel a code-smell when I see them. I think some of that is at fault to the JS engine and allowing `==` to work for null or undefined. The one exception to "always use ===". 

Anyways, I've said my peace. If you don't agree after this, so be it. The ownership of this codebase is no longer in my hands anymore. I'm just sharing my knowledge and experience.

emilys314: Interestingly the official typescript codebase doesn't use `null` at all https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines#null-and-undefined

But yeah returning `... | undefined | null` does feel like a code smell for sure. I do like the sound of a general rule of not having `... | undefined` in in return types. (And maybe only allowing undefined through the usage of optional chaining? `foo?: MyType`)

I think the most important part seems like consistency within the code base. Which is hard to enforce without eslint rules

I'll can change this to return just one of them, so `[ResolvedExtension<T> | null, boolean, unknown[]]`

PR #4284, opened 2025-05-29
refactor import paths to use subpath imports

christianvogt:

## Description

- update to use [node subpath imports](https://nodejs.org/api/packages.html#subpath-imports)
- update Typescript and jest versions to support subpath imports 
- We use the typescript paths compiler option to allow for alias `~/` to reference the source code root. Consequently, an equivalent alias is needed in order to instruct webpack how to resolve paths with this prefix. By using subpath imports, both webpack and typescript can understand a common configuration.
- refactored all usage of `~/` in import statements to `#~/`

Package.json imports:
```
  "imports": {
    "#~/*": [
      "./*",
      "./*.ts",
      "./*.d.ts",
      "./*.tsx",
      "./*/index.ts",
      "./*/index.tsx"
    ]
  },
```

After testing with webstorm it was found that auto completion of imports was providing a relative path instead of one starting with `#~/`. Therefore the path alias in the tsconfig was added to maintain IDE support.

## How Has This Been Tested?

test automation
ad hoc

## Test Impact

N/A

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/tsconfig.json
+++ b/frontend/src/tsconfig.json
@@ -1,0 +1,8 @@
+{
+  "extends": "../tsconfig.json",
+  "compilerOptions": {
+    "paths": {
+      "#~/*": ["./*"]
+    }
+  }
+}


christianvogt:
@andrewballantyne thanks for testing with webstorm. As a result we decided that including a work around for IDE auto completion was necessary through tsconfig paths.

andrewballantyne (approved):
Reviewed this in a meeting with Christian and we got it working for WebStorm (last commit). Should work for VS Code / Cursor already (as that's what Christian uses).

Ran the app and looked like everything was loading fine too... so here we go!

PR #4287, opened 2025-05-30
Update k8sMergePatchResource, replace undefined value with null to make it removable

DaoDaoNoCode:
JIRA: [RHOAIENG-26263](https://issues.redhat.com/browse/RHOAIENG-26263)

## Description

This issue is because we use `k8sMergePatchResource` when we update the notebook. However, when it gets a field set as `undefined`, it won't do anything to that field based on JSON merge-patch rules. Only when it's set to `null` will it remove that field. The way to fix is to update the merge patch function to replace the `undefined` fields with `null` so we can make sure it works.

## How Has This Been Tested?

1. Create a hardware profile with the node selector set (HP1) and a hardware profile with the empty node selector (HP2)
2. Create a notebook with HP1
3. Verify the notebook resource contains the node selector in `spec.template.spec`
4. Update the notebook, change the hardware profile to HP2
5. Verify the notebook resource doesn't contain the `spec.template.spec.nodeSelector` field anymore
6. Update the notebook back to HP1
7. Verify the node selector field is added back
8. Verify all the other fields are updated as expected and not broken

## Test Impact

Added a unit test to verify the utility function to replace the `undefined` with `null` in an object.

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/api/k8sUtils.ts
+++ b/frontend/src/api/k8sUtils.ts
@@ -8,19 +8,6 @@ import {
   K8sResourceUpdateOptions,
 } from '@openshift/dynamic-plugin-sdk-utils';
 
-// Utility type: requires identifier fields, rest is partial
-export type K8sResourcePatchIdentifier = {
-  apiVersion: string;
-  kind: string;
-  metadata: {
-    name: string;
-    namespace?: string;
-    [key: string]: unknown;
-  };
-};
-
-export type K8sMergePatch<T> = K8sResourcePatchIdentifier & Partial<T>;
-
 export const addOwnerReference = <R extends K8sResourceCommon>(
   resource: R,
   owner?: K8sResourceCommon,


frontend/src/api/k8sUtils.ts
```diff
@@ -45,27 +58,58 @@ export const groupVersionKind = (model: K8sModelCommon): K8sGroupVersionKind =>
...
   fetchOptions = {},
-}: K8sResourceUpdateOptions<TResource>): Promise<TUpdatedResource> => {
-  if (!resource.metadata?.name) {
+}: K8sResourceUpdateOptions<TPatch>): Promise<TUpdatedResource> => {
+  if (!resource.metadata.name) {
     return Promise.reject(new Error('Resource payload name not specified'));
   }
 
+  // Convert all undefined values to null for merge patch
+  const resourceWithNulls = deepReplaceUndefinedWithNull(resource);
```
Gkrumbach07, line 103: this may be undesired behavior for _all_ merge patches we do. if the user expects the standard behavior then this would break that. I would suggest just fixing it at the hardware profile level

Gkrumbach07 (changes requested):

Gkrumbach07 (approved):

PR #4297, opened 2025-06-02
Custom Model Server Size Limits Bug

katieperry4:
Closes [RHOAIENG-17786](https://issues.redhat.com/browse/RHOAIENG-17786)

## Description

Before you could set the model server size requests to more than the limit
Screenshot 2025-06-02 at 4 34 00 PM

Now the +/- buttons enable/disable and the input prevents setting the request to more than the limit and the limit to less than the request:
Screenshot 2025-06-02 at 4 35 19 PM
(arrow keys also work)

## How Has This Been Tested?

Ran `npm run test` and added to a cypress test

To Test:
- Go to deploy a KServe model
- Set the model server size to `custom`
- Mess with the +/- buttons and inputs 
- Requested should never be more than the limit and the limit should never be less than the requested
- Still kinda awkward if you have ex. 4 in the CPU request and 5 in the CPU limit and then try to type '12' in the limit, the onChange in `NumberInputWrapper` runs and you'll end up with '42' since it sets the limit to 4 (the request) and then registers the 2

## Test Impact

Added to a cypress test

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/components/ValueUnitField.tsx
+++ b/frontend/src/components/ValueUnitField.tsx
@@ -66,13 +66,16 @@ const ValueUnitField: React.FC<ValueUnitFieldProps> = ({
             })
           }
           onChange={(value) => {
-            const newValue = value ?? 0;
-            if (!Number.isNaN(newValue)) {
+            if (value === undefined) {
+              onChange(`${''}${currentUnitOption.unit}`);
+              return;
+            }
+            if (!Number.isNaN(value)) {
               if (
-                (minAsNumber === undefined || newValue >= minAsNumber) &&
-                (maxAsNumber === undefined || newValue <= maxAsNumber)
+                (minAsNumber === undefined || Number(value) >= minAsNumber) &&
+                (maxAsNumber === undefined || Number(value) <= maxAsNumber)
               ) {
-                onChange(`${newValue}${currentUnitOption.unit}`);
+                onChange(`${value}${currentUnitOption.unit}`);
               }
             }
           }}


katieperry4:
@emilys314 my main hesitation on changing from `onChange` to `onBlur` is that it happens on `NumberInputWrapper` [here](https://github.com/opendatahub-io/odh-dashboard/blob/8fb3aaf96cae586ee8281bf76b3290bfc1685bde/frontend/src/components/NumberInputWrapper.tsx#L33-L55) which is a few components deep and ends up being used in a bunch of places. Making that change could effect a lot of other components behavior 

CPU/Memory Field -> ValueUnitField -> NumberInputWrapper -> NumberInput
I could theoretically make my own `NumberInputWrapper` to use in `ValueUnitField` since `ValueUnitField` is only used in `CPUField`, `MemoryField`, and `PVSizeField` so that's an option 🤷‍♀️

emilys314:
> @emilys314 my main hesitation on changing from `onChange` to `onBlur` is that it happens on `NumberInputWrapper` [here](https://github.com/opendatahub-io/odh-dashboard/blob/8fb3aaf96cae586ee8281bf76b3290bfc1685bde/frontend/src/components/NumberInputWrapper.tsx#L33-L55) which is a few components deep and ends up being used in a bunch of places. Making that change could effect a lot of other components behavior
> 
> CPU/Memory Field -> ValueUnitField -> NumberInputWrapper -> NumberInput I could theoretically make my own `NumberInputWrapper` to use in `ValueUnitField` since `ValueUnitField` is only used in `CPUField`, `MemoryField`, and `PVSizeField` so that's an option 🤷‍♀️

Gotcha that makes sense

in that case
/lgtm
/approve

PR #4303, opened 2025-06-03
Add the link to the Stop workbench modal when navigating from the admin tab.

d0w:
Issue: https://issues.redhat.com/browse/RHOAIENG-25556 

## Description

Within the /notebookController route -> administration tab, made it so that when you click on stop workbench, the modal that comes up has been changed slightly to reflect UX requirements by changing some text and adding a link to open the notebook within the stop workbench modal.

If stopping one or stopping all workbenches but there is only one workbench to stop:
Screenshot 2025-06-03 at 3 36 21 PM

If stopping all workbenches and there are multiple:
Screenshot 2025-06-03 at 3 39 48 PM

## How Has This Been Tested?

1. Head to Applications -> Enabled -> Launch application (start basic workbench).
    a. You should have a workbench created by any user first.
3. Go to administration panel and click on the 3 dots on the rightmost column, click stop workbench
4. View the new modal and click on the link to go to the notebook IDE.
5. Click on "Stop all workbenches" and if there are more than 1 to stop, there will be no link and the text reflects it.

## Test Impact

No tests added since primarily a text change. No new reusable functions added either. You can test this by creating multiple accounts with different permissions (developer and admin), creating workbenches through the Applications -> Enabled tab, through a data science project tab, or other, and view the page as an admin. You should be able to stop all the workbenches for any user and also open the notebook from within the close workbench modal.

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/pages/notebookController/screens/admin/NotebookAdminControl.tsx
+++ b/frontend/src/pages/notebookController/screens/admin/NotebookAdminControl.tsx
@@ -15,6 +15,7 @@ import ApplicationsPage from '#~/pages/ApplicationsPage';
 import StopServerModal from '#~/pages/notebookController/screens/server/StopServerModal';
 import { Notebook } from '#~/types';
 import { ODH_PRODUCT_NAME } from '#~/utilities/const';
+import useRouteForNotebook from '#~/pages/projects/notebook/useRouteForNotebook.ts';
 import { columns } from './data';
 import StopAllServersButton from './StopAllServersButton';
 import UserTableCellTransform from './UserTableCellTransform';
@@ -24,6 +25,9 @@ import { NotebookAdminContext } from './NotebookAdminContext';
 const NotebookAdminControl: React.FC = () => {
   const [users, loaded, loadError] = useAdminUsers();
   const { serverStatuses, setServerStatuses } = React.useContext(NotebookAdminContext);
+  const [selectedNotebook, setSelectedNotebook] = React.useState<Notebook | undefined>(undefined);
+  const [currentNotebookLink, currentNotebookLinkLoaded, currentNotebookLinkError] =
+    useRouteForNotebook(selectedNotebook?.metadata.name, selectedNotebook?.metadata.namespace);
 
   const onNotebooksStop = React.useCallback(
     (didStop: boolean) => {
@@ -44,6 +48,14 @@ const NotebookAdminControl: React.FC = () => {
         .filter((notebook): notebook is Notebook => !!notebook),
     [serverStatuses],
   );
+
+  React.useEffect(() => {
+    if (notebooksToStop.length === 1) {
+      setSelectedNotebook(notebooksToStop[0]);
+    } else {
+      setSelectedNotebook(undefined);
+    }
+  }, [notebooksToStop]);
 
   return (
     <ApplicationsPage
@@ -107,11 +119,11 @@ const NotebookAdminControl: React.FC = () => {
           />
         </StackItem>
       </Stack>
-      {notebooksToStop.length ? (
+      {(currentNotebookLinkLoaded && notebooksToStop.length === 1) || notebooksToStop.length > 1 ? (
         <StopServerModal
           notebooksToStop={notebooksToStop}
           onNotebooksStop={onNotebooksStop}
-          link="https://google.com"
+          link={currentNotebookLinkError || !currentNotebookLink ? '#' : currentNotebookLink}
         />
       ) : null}
     </ApplicationsPage>
--- a/frontend/src/pages/notebookController/screens/admin/NotebookAdminControl.tsx
+++ b/frontend/src/pages/notebookController/screens/admin/NotebookAdminControl.tsx
@@ -111,7 +111,7 @@ const NotebookAdminControl: React.FC = () => {
         <StopServerModal
           notebooksToStop={notebooksToStop}
           onNotebooksStop={onNotebooksStop}
-          link="#"
+          link="https://google.com"
         />
       ) : null}
     </ApplicationsPage>
--- a/frontend/src/pages/notebookController/screens/server/StopServerModal.tsx
+++ b/frontend/src/pages/notebookController/screens/server/StopServerModal.tsx
@@ -117,7 +117,7 @@ const StopServerModal: React.FC<StopServerModalProps> = ({
       isOpen
       onClose={onClose}
     >
-      <ModalHeader title={`Stop ${textToShow}`} />
+      <ModalHeader title={`Stop ${textToShow}?`} />
       <ModalBody>
         <Stack hasGutter>
           <StackItem>Any unsaved changes to the {getWorkbenchName()} will be lost.</StackItem>
--- a/frontend/src/utilities/notebookControllerUtils.ts
+++ b/frontend/src/utilities/notebookControllerUtils.ts
@@ -190,6 +190,19 @@ export const validateNotebookNamespaceRoleBinding = async (
   );
 };
 
+export const getNotebookRedirectLink = async (
+  notebookNamespace: string,
+  routeName: string,
+): Promise<string | undefined> => {
+  try {
+    const route = await getRoute(notebookNamespace, routeName);
+    return `https://${route.spec.host}/notebook/${notebookNamespace}/${routeName}`;
+  } catch (e) {
+    // console.warn('Unable to get the route for notebook redirect link', e);
+    return undefined;
+  }
+};
+
 export const useNotebookRedirectLink = (): (() => Promise<string>) => {
   const { currentUserNotebook, currentUserNotebookLink } =
     React.useContext(NotebookControllerContext);


frontend/src/pages/notebookController/screens/admin/NotebookActions.tsx
```diff
@@ -34,10 +41,10 @@ const NotebookActions: React.FC<ServerStatusProps> = ({ data }) => {
   return (
     <>
       <ActionsColumn items={rowActions} />
-      {showModal && (
+      {showModal && notebookLinkLoaded && (
         <StopServerModal
           notebooksToStop={notebooksToStop}
-          link="#"
+          link={notebookLinkError || !notebookLink ? '#' : notebookLink}
```
rsun19, line 47:
```suggestion
          link={notebookLink}
```
I dont think we need to show the error in the link when populating it.

Also, I think we can design this component better by changing the type to `link?: string;` in stop server modal to accept undefined. So we remove the check for hardcoded `'#'` and check for undefined instead (fix the checking function), and in `NotebookServer.tsx`, we can do

```
<StopServerModal
                  notebooksToStop={notebooksToStop}
                  isDeleting={isDeleting}
                  link={link !== '#' ? link : undefined}
                  onNotebooksStop={onNotebooksStop}
                />
```

frontend/src/pages/notebookController/screens/admin/StopAllServersButton.tsx
```diff
@@ -42,14 +49,14 @@ const StopAllServersButton: React.FC<StopAllServersButtonProps> = ({ users }) =>
       >
         Stop all workbenches ({serverCount})
       </Button>
-      {showModal && (
+      {showModal && loaded ? (
         <StopServerModal
           notebooksToStop={notebooksToStop}
-          link="#"
+          link={loadError || !routeLink ? '#' : routeLink}
```
rsun19, line 55: no need for load error

frontend/src/pages/notebookController/screens/admin/NotebookActions.tsx
rsun19: Ah i see, i was probably too tired to read the ternary correctly lol

frontend/src/pages/notebookController/screens/admin/NotebookActions.tsx
```diff
@@ -34,10 +41,10 @@ const NotebookActions: React.FC<ServerStatusProps> = ({ data }) => {
   return (
     <>
       <ActionsColumn items={rowActions} />
-      {showModal && (
+      {showModal && notebookLinkLoaded && (
         <StopServerModal
           notebooksToStop={notebooksToStop}
-          link="#"
+          link={!!notebookLinkError || !notebookLink ? '#' : notebookLink}
```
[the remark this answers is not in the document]
d0w: > didn't we agree to pass in undefined as opposed to '#'?

Just changed those.

> and then change
The change is the `!!link` portion to assert that link is not undefined/null. I kept the other checks in there for backwards compatibility.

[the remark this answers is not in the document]
d0w: You're right. I didn't realize there were no other references to the modal. Just made the change

frontend/src/pages/notebookController/screens/admin/StopAllServersButton.tsx
```diff
@@ -42,14 +49,14 @@ const StopAllServersButton: React.FC<StopAllServersButtonProps> = ({ users }) =>
       >
         Stop all workbenches ({serverCount})
       </Button>
-      {showModal && (
+      {showModal && loaded ? (
```
rsun19, line 52:
```suggestion
      {showModal && loaded && (
```
we don't need to use ternary here

rsun19 (commented):
Everything else looks good

frontend/src/pages/notebookController/screens/admin/NotebookActions.tsx
```diff
@@ -2,14 +2,21 @@ import * as React from 'react';
...
 
 const NotebookActions: React.FC<ServerStatusProps> = ({ data }) => {
-  const notebooksToStop = data.notebook ? [data.notebook] : [];
+  const notebookToStop = data.notebook || null;
+  const notebooksToStop = notebookToStop ? [notebookToStop] : [];
+
+  const [notebookLink, notebookLinkLoaded, notebookLinkError] = useRouteForNotebook(
+    notebookToStop?.metadata.name,
+    notebookToStop?.metadata.namespace,
+  );
```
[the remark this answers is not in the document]
d0w: Good point. The initial issue was that you'd have to wait for an api call on demand, so the modal would either delay the popup or you would see the "open workbench" text change with the modal open after the api request resolved. Though, looking at components that use NotebookRouteLink.tsx, I see that we don't mind the flicker as long as it's handled on the UI side with a fade in.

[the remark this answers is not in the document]
DaoDaoNoCode: Yeah, or probably considering using `Skeleton` component in the modal for the loading links. So user won't get a delay when opening the modal.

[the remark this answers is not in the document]
d0w: @DaoDaoNoCode and I discussed this and agreed that we should just keep it as-is since this page doesn't get too much traffic anyways.

frontend/src/pages/notebookController/screens/admin/useRouteForNotebook.ts
```diff
@@ -0,0 +1,38 @@
+import * as React from 'react';
+import { getRoute } from '#~/api';
+
+// separate from useRouteForNotebook within projects directory. This one does not poll
+const useRouteForNotebook = (
```
Gkrumbach07, line 5: it may be better to reuse the current useRouteForNotebook with polling but add an option to make polling optional. While you aren there you can simplify the polling to use our useFetch hook which has a built in option to enable polling

Gkrumbach07 (changes requested):
very nice! just one chnage

frontend/src/pages/notebookController/screens/admin/NotebookActions.tsx
```diff
@@ -34,10 +41,10 @@ const NotebookActions: React.FC<ServerStatusProps> = ({ data }) => {
   return (
     <>
       <ActionsColumn items={rowActions} />
-      {showModal && (
+      {showModal && notebookLinkLoaded && (
         <StopServerModal
           notebooksToStop={notebooksToStop}
-          link="#"
+          link={!!notebookLinkError || !notebookLink ? undefined : notebookLink}
           isDeleting={isDeleting}
```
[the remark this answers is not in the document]
DaoDaoNoCode: @d0w Same here, we shouldn't block users from stopping the notebook if the link is not loaded.

frontend/src/concepts/notebooks/apiHooks/useRouteForNotebook.ts
```diff
@@ -0,0 +1,66 @@
...
+  );
+
+  const initialState: NotebookRouteState = {
+    route: null,
+    error: null,
+    loaded: false,
+  };
+
+  const { data, error } = useFetch<NotebookRouteState>(fetchRoute, initialState, {
+    refreshRate: pollInterval || undefined,
```
DaoDaoNoCode, line 59: `pollInterval` is undefined if it's not there. I think there is no need to do the check 🤔

DaoDaoNoCode (commented):
Left some comments, I think I need to double check the usage of `useFetch` since it was refactored before. I will look into it and probably leave some more comments in the `useRouteForNotebook` hook later.

frontend/src/concepts/notebooks/apiHooks/useRouteForNotebook.ts
d0w: @DaoDaoNoCode implemented those changes. Please let me know about the usage of `useFetch` in this context. `useFetch` was used instead of a setTimeout loop within the `useRouteForNotebook` hook itself.

Gkrumbach07 (approved):
very nice (☞ຈل͜ຈ)☞

/lgtm
/approve

Gkrumbach07:
@DaoDaoNoCode looks good to me, i will wait for u to put the approval on since ill be out after today

frontend/src/concepts/notebooks/apiHooks/useRouteForNotebook.ts
```diff
@@ -0,0 +1,46 @@
+import * as React from 'react';
+import { getRoute } from '#~/api';
+import useFetch, { FetchStateCallbackPromise, FetchStateObject } from '#~/utilities/useFetch';
+import { K8sAPIOptions } from '#~/k8sTypes';
+
+type NotebookRouteState = {
+  route: string | null;
+};
```
DaoDaoNoCode, line 8: We can totally remove this type. Return `<string | null>` as the result of the network call

PR #4313, opened 2025-06-05
Cypress logs Improvements: timestamp and log level

FedeAlonso:
https://issues.redhat.com/browse/RHOAIENG-26712

## Description
Add timestamp and log level to Cypress logs

## How Has This Been Tested?
dashboard-tests/1718
image

## Test Impact
full

## Request review criteria:

If you have UI changes: 


frontend/src/__tests__/cypress/cypress.config.ts
```diff
@@ -97,18 +98,15 @@ export default defineConfig({
         },
         log(message) {
           // eslint-disable-next-line no-console
-          console.log(message);
-          return null;
+          return logToConsole('INFO', message, console.log);
```
srtanish1992, line 101: We can create a enum for for level like
```
enum Level {
INFO = 'info',
ERROR = 'error',
TABLE = 'table'
}
```

bobbravo2 (commented):
This is a great improvement for out-of-order logs! 

Couple of ideas to share

1. Can we use a LogLevel enum instead of raw strings for logToConsole(level)?  
That should help typescript help us ensure future compatibility of this interface, while making extensibility a bit more explicit. 

props @srtanish1992 
```typescript
enum Level {
INFO = 'info',
ERROR = 'error',
TABLE = 'table'
}
``` 

2. Do we need both level and `consoleMethod` as part of the function signature? I kinda feel like having a signature of `logToConsole(LogLevel.LOG, message)` would be cleaner, and we could push the `LogLevel.TABLE` abstraction down to `logToConsole`. (2 also assumes 1, that we create a new LogLevel enum in typescript, hence the capitals instead of magick strings)

srtanish1992:
@FedeAlonso I have completed the review. I have few comments. Let me know when you address them and I will take a look again.

frontend/src/__tests__/cypress/cypress.config.ts
FedeAlonso: Fixed in [ea957dc](https://github.com/opendatahub-io/odh-dashboard/pull/4313/commits/ea957dc5d7c56a94593b7517dad0f5215b316aeb)

frontend/src/__tests__/cypress/cypress/utils/logger.ts
```diff
@@ -0,0 +1,23 @@
+export const LogLevel = {
+  INFO: { value: 'info', method: console.log },
```
bobbravo2, line 2:
```suggestion
  // test code, logs to stdout, not a browser console
  // eslint-disable-next-line no-console
  INFO: { value: 'info', method: console.log },
```

FedeAlonso: The call to the linter is done  in cypress.config.ts 
Also, I've discovered that when doing the console.log, everything's logged in the stdout too, the issue is that everything's a little bit mixed because of the parallelism (And I'm fixing that in another PR)

bobbravo2 (approved):

frontend/src/__tests__/cypress/cypress/utils/logger.ts
srtanish1992, line 2: To get rid of the lint issue we can do `method: logMethod` and define logMethod like below if we don't wan't to use `eslint-disable-next-line no-console`.

```
function logInfo(message) {
    console.log(message);
}
```

FedeAlonso: You were right, I needed to fix correctly how to skip the lint issue

FedeAlonso:
After latest changes:

PR #4314, opened 2025-06-05
Pipeline creation feedback

jrenee42:
for: https://issues.redhat.com/browse/RHOAIENG-21524

## Description
Adds a modal to the page that pops up when the user is creating a pipeline server.  it gives status updates for all the pods that are being created.

updated video as of 6-9 2:50pm: 

https://github.com/user-attachments/assets/65b1fce0-21ae-483a-be07-313fd4befc85

 @acrago  please take a look

note;  the modal pops up from both the Pipelines page (as above in the video) and in the pipelines tile in the data science project  : 
Screenshot 2025-06-05 at 2 52 21 PM

## How Has This Been Tested?
I tested this manually; by creating a pipeline server over and over (i kept deleting it from my data science project)
also added unit tests

## Test Impact
added tests for the modal itself, and for determining the status

the issue is that the status returned by the pods is either: 
"True", "False", or "Unknown"
and the desired statuses (re the figma) is:
pending, in progress, success, warning, or error

I spent * a lot* of time starting pipeline servers and looking at logs.

findings:
  - a lot of the pods are dependent on other pods starting.  so they will keep saying 'failure' ('False') until they don't. 
     - every time i tested, all the failures resolved to "True" (success) within 2 minutes
     - when it is deploying, a pod's condition (status) will look like this:
     
```
lastTransitionTime: "2025-06-05T18:30:51Z"
message: "Component [ds-pipeline-dspa] is deploying."
reason: "Deploying"
status: "False"
type: "APIServerReady"
```

before a pod gets to the deployment stage (above); the condition can look like this:

```
lastTransitionTime: "2025-06-05T18:30:51Z"
message : "Component's replica [ds-pipeline-dspa] has failed to create. Reason: [FailedCreate]. Message: [pods \"ds-pipeline-dspa-66b958d84d-\" is forbidden: error looking up service account brand-new-one/ds-pipeline-dspa: serviceaccount \"ds-pipeline-dspa\" not found]"
reason: "FailingToDeploy"
status: "False"
type: "APIServerReady"
```

so the code checks the last transition time above.  if the time is within the last 3 minutes (a short-term range); the status becomes pending
if the status is between 3 and 5 (between a short and long term range)e minutes; then the status is warning;
if it over the 'long term ' threshold, then and only then it becomes an error.

all the status determination is done in the utlis class.  

and an extra 'isStarting' is added to the pipelines fetch; to determine the difference between starting a new server and connecting to an already existing server.

## Request review criteria:

If you have UI changes: 

## Summary by CodeRabbit

- **New Features**
  - Introduced an interactive modal to monitor pipeline server startup, showing progress and event logs.
  - Added status icons and refined status mapping for pipeline components.
  - Provided real-time Kubernetes event logs and detailed condition statuses within modals.
  - Added hooks to watch pipeline server pods and related Kubernetes events.
  - Added a new utility to categorize timestamps for status evaluation.

- **Bug Fixes**
  - Improved user feedback during pipeline server initialization and unavailability.

- **Refactor**
  - Replaced manual event log rendering with a dedicated reusable component.
  - Unified modal styling using new SCSS styles and CSS classes.
  - Enhanced context and hooks to propagate pipeline server starting state.

- **Tests**
  - Added extensive tests for the startup status modal and status mapping utilities.

- **Style**
  - Added new SCSS styles for modals and spinner layouts to ensure consistent UI.


acrago:
@jrenee42 - The modal itself looks great, however I'm hesitant to automatically display the modal upon creation. My fear is that the user will think they are required to keep the modal open in order for the server to configure. In the original design, a message is displayed on the page with a link to the modal that they can open if they want information on what's happening

frontend/src/concepts/pipelines/content/K8sStatusIcon.tsx
```diff
@@ -0,0 +1,48 @@
...
+      case 'in-progress':
+        return <InProgressIcon style={{ color: BlackColor.var }} className="odh-u-spin" />;
+      case 'success':
+        return <CheckCircleIcon style={{ color: SuccessColor.var }} />;
+      case 'warning':
+        return <ExclamationTriangleIcon style={{ color: WarningColor.var }} />;
+      case 'error':
+        return <ExclamationCircleIcon style={{ color: DangerColor.var }} />;
+      default:
+        return <PendingIcon style={{ color: BlackColor.var }} />;
```
d0w, line 37: it might be viable to use an enum here. maybe even defined in a separate types file so that if other components need to check or use the same statuses, they can assert the Status type and use the enum values.

jrenee42:
> @jrenee42 - The modal itself looks great, however I'm hesitant to automatically display the modal upon creation. My fear is that the user will think they are required to keep the modal open in order for the server to configure. In the original design, a message is displayed on the page with a link to the modal that they can open if they want information on what's happening

so i'm changing the 'in progress' description to as follows: 
'  The pipeline server is currently being initialized. This process may take a few minutes. Closing this dialog will not affect the pipeline server creation; this just shows the status.'

jrenee42:
> This description does provide further detail, so you can add this to the modal description section:
> 
> > The pipeline server is currently being initialized. This process may take a few minutes. Closing this dialog will not affect the pipeline server creation.
> 
> I'll update the designs to reflect that. But for the functionality, we don't have anything in the product that automatically opens the status modal, so this action would be inconsistent. It is common to have a button that to open a status modal, so I recommend we just stick with that. The [designs](https://www.figma.com/design/DH2Fdtm8sbWt3rdrkHtliq/Indicator-that-pipeline-server-is-starting?node-id=93-3404&t=AOw93KDEiKoblPEy-1) show the "Initializing pipeline server" as a blue link which the user can click and the modal will pop up.
> 
> So to summarize, when the user creates the pipeline server, they should see the initializing message on the page, and can optionally open the modal- but we won't automatically open the modal for them.

@acrago  i'm worried the user won't know to click the link.  how about text that says: "The Pipeline Server is initializing; click for more details"

frontend/src/concepts/pipelines/content/StartingStatusModal.tsx
```diff
@@ -0,0 +1,163 @@
...
+  );
+
+  const successDesc = (
+    <span data-testid="successDescription">
+      The pipeline server has been successfully initialized and is ready to use.
+    </span>
+  );
+  const inProgressDesc = (
+    <span data-testid="inProgressDescription">
+      The pipeline server has been successfully initialized and is ready to use.
```
bobbravo2, line 111:
```suggestion
      The pipeline server is currently initializing.
```

acrago:
@jrenee42 Typically "Click here" type of messaging isn't reflective of good UX for a few reasons. The purpose of this project was to provide more messaging to the user that the pipeline server is initializing because previously it was just a spinning wheel with no indication of what's happening. The modal with the progress + event log was an additional "nice to have" that I included, but the main purpose was to let the user know what's happening when they finish configuring a server. 
Plus, blue text is a standard for links. PatternFly has set the expectations for users that blue text = clickable. If I were to make changes to this design, I would have to take it back to the UX team because it's already been through rounds of review and approval.

frontend/src/concepts/pipelines/content/StartingStatusModal.tsx
jrenee42: good catch.  fixed

jrenee42:
> @jrenee42 Typically "Click here" type of messaging isn't reflective of good UX for a few reasons. The purpose of this project was to provide more messaging to the user that the pipeline server is initializing because previously it was just a spinning wheel with no indication of what's happening. The modal with the progress + event log was an additional "nice to have" that I included, but the main purpose was to let the user know what's happening when they finish configuring a server. Plus, blue text is a standard for links. PatternFly has set the expectations for users that blue text = clickable. If I were to make changes to this design, I would have to take it back to the UX team because it's already been through rounds of review and approval.

ok; updated

acrago:
@jrenee42 Awesome! The copy in the modal description hasn't been updated, the messaging should be:

> The pipeline server is currently being initialized. This process may take a few minutes. Closing this dialog will not affect the pipeline server creation.

frontend/src/concepts/pipelines/EnsureAPIAvailability.tsx
```diff
@@ -1,22 +1,94 @@
...
+  };
+  return (
+    <>
+      {getMainComponent()}
+      {showModal && (
+        <StartingStatusModal onClose={() => setShowModal(false)} conditionLog={conditionLog} />
+      )}
+    </>
+  );
 };
```
Gkrumbach07, line 92: There looks like there is a lot of mixing of logic in this component. we can do something like this. 

- no need to add new prop for isPipeline
- we can move the condition log state into the modal bc thats the only one that uses it
```tsx
  const conditionLog = React.useMemo(() => {
    if (!pipelinesServer.crStatus?.conditions) {
      return [];
    }
    return [...pipelinesServer.crStatus.conditions].reverse();
  }, [pipelinesServer.crStatus?.conditions]);
```
- buttons have a loading state
- avoid using classNames when you can. this same UX can be accomplished with the Flex component

_suggested component_
```tsx
const EnsureAPIAvailability: React.FC<EnsureAPIAvailabilityProps> = ({ children }) => {
  const { apiAvailable, pipelinesServer } = usePipelinesAPI();
  const [showModal, setShowModal] = React.useState(false);

  // if not ready, show the modal
  if (pipelinesServer.crReady) {
    return (
      <>
        <Bullseye style={{ minHeight: 150 }}>
          <Flex
            direction={{ default: 'column' }}
            justifyContent={{ default: 'justifyContentCenter' }}
          >
            <FlexItem>
              <Button
                data-testid="open-pipeline-status-link"
                variant="link"
                isLoading
                onClick={() => {
                  setShowModal(true);
                }}
              >
                Initializing Pipeline Server
              </Button>
            </FlexItem>
            <FlexItem>This may take a while</FlexItem>
          </Flex>
        </Bullseye>
        {showModal && <StartingStatusModal onClose={() => setShowModal(false)} />}
      </>
    );
  }

  // if not available, show the spinner
  if (!apiAvailable) {
    return (
      <Bullseye style={{ minHeight: 150 }} data-testid="pipelines-api-not-available">
        <Spinner />
      </Bullseye>
    );
  }

  return children;
};
```

frontend/src/concepts/pipelines/content/utils.tsx
```diff
@@ -107,3 +109,25 @@ export const isPipelineRun = (resource: PipelineCoreResourceKF): resource is Pip
...
+      case 'shortRange':
+        return 'pending';
+      case 'mediumRange':
+        return 'warning';
+      case 'longRange':
+        return 'error';
+    }
+  }
+  return 'pending';
+};
```
Gkrumbach07, line 133: so the biggest issue we have with getting the true component status for the DSPA is that we dont have a good enum defined in the schema. it is just a string for reason.

What we do know is that `status === 'True'` means success which you have

What concerns me is the time range being used to show if it is a loading, a warning, or an error and that it is based on arbitrary times that dont relate to the status. We should not use time to determine this. It may result in a weird looking modal that jumps between statuses, but that is actually what is happening. Pods are spinning up and failing bc other components may not be ready, so we should accurately depict that.

I would say we hunt down what the pipelines codebase uses for reason. maybe there is a enum ion there and we can link it for reference later in case they change something. relying on plain strings to be the same can get risky. At least we should use an enum like

```tsx
// link to where this enum is defined in the backend if possible
enum DSPAConditionReasons {DEPLOYING: 'Deploying', ...}
```

Also `Unknown` status needs to show this icon
Screenshot 2025-06-11 at 3 21 09 PM

frontend/src/concepts/pipelines/content/StartingStatusModal.tsx
```diff
@@ -0,0 +1,157 @@
...
+import { getStatusFromCondition } from '#~/concepts/pipelines/content/utils.tsx';
+import K8sStatusIcon from '#~/concepts/pipelines/content/K8sStatusIcon.tsx';
+import { K8sCondition } from '#~/k8sTypes.ts';
+
+const PROGRESS_TAB = 'Progress';
+const EVENT_LOG_TAB = 'Events log';
+
+const notReadySection = (
+  <div>
+    <p style={{ textAlign: 'center' }} />
```
Gkrumbach07, line 29: no custom styles

frontend/src/concepts/pipelines/content/StartingStatusModal.tsx
```diff
@@ -0,0 +1,157 @@
...
+} from '@patternfly/react-core';
+import { usePipelinesAPI } from '#~/concepts/pipelines/context';
+import { getStatusFromCondition } from '#~/concepts/pipelines/content/utils.tsx';
+import K8sStatusIcon from '#~/concepts/pipelines/content/K8sStatusIcon.tsx';
+import { K8sCondition } from '#~/k8sTypes.ts';
+
+const PROGRESS_TAB = 'Progress';
+const EVENT_LOG_TAB = 'Events log';
+
+const notReadySection = (
```
Gkrumbach07, line 27: why is this pulled out, it is used once

frontend/src/concepts/pipelines/content/StartingStatusModal.tsx
```diff
@@ -0,0 +1,157 @@
...
+// (this component only exists when it is open): eslint rules do not allow this component to exist if it is not open
+const StartingStatusModal: React.FC<StartingStatusModalProps> = ({ onClose, conditionLog }) => {
+  const { pipelinesServer } = usePipelinesAPI();
+  const [activeTab, setActiveTab] = React.useState<string>(PROGRESS_TAB);
+  const isServerReadyAndCompletelyDone = pipelinesServer.crStatus?.conditions?.some(
+    (c) => c.type === 'Ready' && c.status === 'True',
+  );
+
+  const spinner = (
+    <div className="pf-v6-u-display-flex">
```
Gkrumbach07, line 56: dont use classnames

reason is that they are tied to Patternfly 6. when we upgrade to the next patternfly, someone has to go through every one of these and fix them. better to not use them

frontend/src/concepts/pipelines/content/StartingStatusModal.tsx
```diff
@@ -0,0 +1,157 @@
...
+          })}
+        </Stack>
+      </StackItem>
+      <StackItem />
+      {!isServerReadyAndCompletelyDone && notReadySection}
+    </Stack>
+  );
+
+  const renderLogs = () => (
+    <Panel style={{ overflowY: 'auto', height: '100%' }}>
```
Gkrumbach07, line 84: no cusotm styling

frontend/src/concepts/pipelines/content/StartingStatusModal.tsx
```diff
@@ -0,0 +1,157 @@
...
+      <PanelMain>
+        <List isPlain isBordered data-id="event-logs">
```
Gkrumbach07, line 86: data-testid

frontend/src/concepts/pipelines/context/PipelinesContext.tsx
```diff
@@ -178,6 +181,8 @@ type UsePipelinesAPI = PipelineAPIState & {
     timedOut: boolean;
     compatible: boolean;
     name: string;
+    crStatus: DSPipelineKind['status'];
+    isStarting: boolean | undefined;
```
Gkrumbach07, line 185: you can use isCrReady to get the "isStarting"
```tsx
crReady: isCRReady,
```

frontend/src/concepts/pipelines/content/K8sStatusIcon.tsx
Gkrumbach07: it is used as input and is used elsewhere. it should then be a enum

frontend/src/utilities/useFetchState.ts
```diff
@@ -31,6 +31,7 @@ export type FetchState<Type> = [
   refresh: FetchStateRefreshPromise<Type>,
 ];
 
+export type FetchStateWithStarting<Type> = [...FetchState<Type>, isStarting?: boolean | undefined];
```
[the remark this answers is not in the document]
jrenee42: then how do we know to tell the difference between when the pipeline is starting vs. when an existing pipeline is being accessed/connected to?

frontend/src/concepts/pipelines/context/usePipelineNamespaceCR.ts
```diff
@@ -57,18 +60,20 @@ const usePipelineNamespaceCR = (namespace: string): FetchState<State> => {
...
+  const pipelineApiServerReady = dspaLoaded(state);
+
   React.useEffect(() => {
-    setIsStarting(resourceLoaded && !hasStatus);
-  }, [hasStatus, resourceLoaded]);
+    setIsStarting(resourceLoaded && !pipelineApiServerReady);
+  }, [pipelineApiServerReady, resourceLoaded]);
 
-  return state;
+  return [...state, isStarting];
```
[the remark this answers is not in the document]
jrenee42: that still doesn't differentiate between when a pipeline server is starting/being initialized for the first time vs when an existingt pipeline server is being connected to.

frontend/src/pages/pipelines/global/pipelines/PipelinesView.tsx
```diff
@@ -30,6 +30,7 @@ const PipelinesView: React.FC = () => {
   if (!loaded && !initialLoaded) {
     return (
       <Bullseye>
+        Pipeline is loading
```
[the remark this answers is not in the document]
jrenee42: true, but not everything is in the mocks; this is for when the pipeline is just loading (not when it is initializing).  this way, the user can tell the difference.

frontend/src/concepts/pipelines/context/usePipelineNamespaceCR.ts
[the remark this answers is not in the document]
jrenee42: because APIServerReady is not the same as all the pods being ready.
(which is what dspaLoaded checks)

frontend/src/concepts/pipelines/content/StartingStatusModal.tsx
Gkrumbach07: no use the `Flex` component

frontend/src/concepts/pipelines/EnsureAPIAvailability.tsx
Gkrumbach07: wont it just pull the conditions each time u open the modal

frontend/src/concepts/pipelines/context/usePipelineNamespaceCR.ts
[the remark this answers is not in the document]
Gkrumbach07: isCRReady is specifically telling you if the CR has a ready condition (all pods being ready)

you can do this if else logic

```
if !isCRReady && crExists => show modal

else if !apiServerReady => show spinner
```
this should then only show the modal when the pipeline server is not ready

frontend/src/utilities/useFetchState.ts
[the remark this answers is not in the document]
Gkrumbach07: see https://github.com/opendatahub-io/odh-dashboard/pull/4314#discussion_r2140918370

frontend/src/pages/pipelines/global/pipelines/PipelinesView.tsx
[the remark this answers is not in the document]
jrenee42: ah; ok.  yeah, agreed the flash/jittery part is sub-optimal

frontend/src/concepts/pipelines/content/StartingStatusModal.tsx
Gkrumbach07: it is not a big component so it should be fine inline.

frontend/src/concepts/pipelines/content/StartingStatusModal.tsx
```diff
@@ -0,0 +1,157 @@
...
+
+  const renderProgress = () => (
+    <Stack hasGutter>
+      <StackItem>
+        <Stack hasGutter>
+          {pipelinesServer.crStatus?.conditions?.map((condition, index) => {
+            const containerStatus = getStatusFromCondition(condition);
+            return (
+              <StackItem key={`${condition.type}-${index}`}>
+                <div className="pipeline-status-modal__status-item">
```
Gkrumbach07, line 69: what are these classnames. are you using them for tests? if so then use data-testid

frontend/src/concepts/pipelines/content/StartingStatusModal.tsx
Gkrumbach07: generally we always want to use the patternfly components and props as much as possible bc classnames are a pain to migrate on upgrade

frontend/src/concepts/pipelines/content/utils.tsx
jrenee42: but it's *expected* to fail first; because the pod and resources it needs isn't ready yet.
as opposed to failing because something is wrong.

when the status says 'failure' and ' is deploying'...that feels wrong.
the issue is that the data we are getting is not user friendly, and this code makes it friendly (it doesn't show any fast failures; because they aren't real failures; in that the pod will (very soon) be successful and those early 'failures' are expected.

frontend/src/concepts/pipelines/context/usePipelineNamespaceCR.ts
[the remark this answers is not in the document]
jrenee42: do you mean crExists--> there are actual conditions?  because then we could be telling the user that the pod is starting up/initializing when really it's just starting up/connecting and then we'll get a flash of the link in that case.  

we need a way to tell when the pipeline server has been initialized from scratch vs. when it is connecting to an already existing pipeline server.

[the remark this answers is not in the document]
jrenee42: `isDspaAllReady` added to` usePipelineNamespaceCR` per our conversation (over google meet).  and isStarting removed also per conversation

frontend/src/concepts/pipelines/content/utils.tsx
jrenee42: i spoke to @HumairAK  about this; and we came up with logic to determine the status.  the "Fail" reason is a bug; see https://issues.redhat.com/browse/RHOAIENG-27727 ; and the reasons are now an enum in the code as well.

frontend/src/utilities/time.ts
```diff
@@ -215,3 +215,25 @@ export const convertSecondsToPeriodicTime = (seconds: number): string => {
...
+  const diffMinutes = (now - timestampMs) / (1000 * 60);
+
+  if (diffMinutes > mediumTimeRangeMinuteLimit) {
+    return 'longRange';
+  }
+  if (diffMinutes > shortTimeRangeMinuteLimit) {
+    return 'mediumRange';
+  }
+  return 'shortRange';
+};
```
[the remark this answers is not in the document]
jrenee42: used suggestion

frontend/src/concepts/pipelines/content/StartingStatusModal.tsx
jrenee42: those are gone; using flex-items now.

frontend/src/concepts/pipelines/content/StartingStatusModal.tsx
```diff
@@ -0,0 +1,157 @@
...
+import { getStatusFromCondition } from '#~/concepts/pipelines/content/utils.tsx';
+import K8sStatusIcon from '#~/concepts/pipelines/content/K8sStatusIcon.tsx';
+import { K8sCondition } from '#~/k8sTypes.ts';
+import './StartingStatusModal.scss';
+
+const PROGRESS_TAB = 'Progress';
+const EVENT_LOG_TAB = 'Events log';
+
+const notReadySection = (
+  <div className="pipeline-status-modal__not-ready-section">
```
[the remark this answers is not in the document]
jrenee42: removed

frontend/src/concepts/pipelines/content/StartingStatusModal.tsx
jrenee42: removed; using flex items

frontend/src/concepts/pipelines/content/K8sStatusIcon.tsx
```diff
@@ -0,0 +1,62 @@
...
+    icon: ExclamationCircleIcon,
+    statusColor: 'danger',
+  },
+};
+
+type StatusIconProps = {
+  status: StatusType;
+};
+
+const K8sStatusIcon: React.FC<StatusIconProps> = ({ status }) => {
```
Gkrumbach07, line 51: not really a k8s status. it is more specifically a pipeline server condition status. we should rename so people dont end up using this outside of that specific CRD's conditions

frontend/src/concepts/pipelines/context/usePipelineEvents.ts
```diff
@@ -0,0 +1,85 @@
...
+  const pod0Events = useWatchPipelineServerEvents(namespace, podUids[0] ?? '');
+  const pod1Events = useWatchPipelineServerEvents(namespace, podUids[1] ?? '');
+  const pod2Events = useWatchPipelineServerEvents(namespace, podUids[2] ?? '');
+  const pod3Events = useWatchPipelineServerEvents(namespace, podUids[3] ?? '');
+  const pod4Events = useWatchPipelineServerEvents(namespace, podUids[4] ?? '');
+  const pod5Events = useWatchPipelineServerEvents(namespace, podUids[5] ?? '');
+  const pod6Events = useWatchPipelineServerEvents(namespace, podUids[6] ?? '');
+  const pod7Events = useWatchPipelineServerEvents(namespace, podUids[7] ?? '');
+  const pod8Events = useWatchPipelineServerEvents(namespace, podUids[8] ?? '');
+  const pod9Events = useWatchPipelineServerEvents(namespace, podUids[9] ?? '');
```
Gkrumbach07, line 52: dont do this. there might not always be the same number of pipeline server related pods. make it dynamic instead

frontend/src/concepts/pipelines/content/StartingStatusModal.tsx
```diff
@@ -0,0 +1,193 @@
...
+    (c) => c.type === 'Ready' && c.status === 'True',
+  );
+
+  const [pods] = useWatchPodsForPipelineServerEvents(namespace);
+  const podUids = pods.map((onePod) => onePod.metadata.uid);
+
+  // Use the custom hook to get all events
+  const allEvents = useWatchMultiplePodEvents(
+    namespace,
+    podUids.filter((uid): uid is string => uid !== undefined),
```
Gkrumbach07, line 52: it might make sense to combine this map and filter into one podUId and put it behind useMemo. Otherwise useWatchMultiplePodEvents will have to rerun each time.

frontend/src/concepts/pipelines/content/utils.tsx
```diff
@@ -107,3 +109,36 @@ export const isPipelineRun = (resource: PipelineCoreResourceKF): resource is Pip
 export const isPipelineRecurringRun = (
   resource: PipelineCoreResourceKF,
 ): resource is PipelineRecurringRunKF => 'recurring_run_id' in resource && !('run_id' in resource);
+
+// workaround until https://issues.redhat.com/browse/RHOAIENG-27727 is fixed
+// after the above ticket is resolved, once 'failingTodeploy' is detected then it is an automatic fail
+// without any need for a timeout.
+// but we should still timeout everything else
```
Gkrumbach07, line 116: can you create a followup jira that is blocked by https://issues.redhat.com/browse/RHOAIENG-27727. so we can track it

frontend/src/k8sTypes.ts
```diff
@@ -126,6 +127,12 @@ export type ModelServingProjectLabels = {
   [KnownLabels.MODEL_SERVING_PROJECT]: 'true' | 'false';
 };
 
+/* not using K8sDSPAConditionReason because these conditions are not specific to the Data Science Pipelines Operator;
+// they are used all over the code.
+// for example: other values:
+//  'DBConnecting' (used in trustyai utils)
+//  'Unschedulable' (used in kserve status utils)
+//  'TriggeredScaleUp', 'NotTriggerScaleUp', 'FailedScheduling', 'BackOff' (used in notebook controller utils) */
```
Gkrumbach07, line 135: this comment is not needed here

Gkrumbach07 (changes requested):

frontend/src/concepts/pipelines/content/utils.tsx
jrenee42: added the followup ticket into the comment as well

jrenee42:
@Gkrumbach07  those two changes are made:  just one watch getting all the events that are now filtered, and using Flexitems in EnsureAPIAvailability.  the watch code is very clean actually.

frontend/src/concepts/pipelines/content/StartingStatusModal.tsx
```diff
@@ -0,0 +1,198 @@
...
+
+  // Combine map and filter operations into a single useMemo
+  const podUids = React.useMemo(
+    () =>
+      pods.map((onePod) => onePod.metadata.uid).filter((uid): uid is string => uid !== undefined),
+    [pods],
+  );
+
+  // Use the custom hook to get all events
+  const allEvents = useWatchAllPodEventsAndFilter(namespace, podUids);
```
Gkrumbach07, line 58: i would move these hook calls into `useWatchAllPodEventsAndFilter`

frontend/src/concepts/notebooks/StartNotebookModal.tsx
```diff
@@ -193,25 +186,12 @@ const StartNotebookModal: React.FC<StartNotebookModalProps> = ({
   };
 
   const renderLogs = () => (
-    <Panel style={{ overflowY: 'auto', height: '100%', padding: SmallSpacerSize.var }}>
+    <Panel className="odh-modal__scrollable-panel">
```
bobbravo2, line 189: Would it be simpler for us to use this: https://www.patternfly.org/components/panel/#panel the `isScrollable` prop?

Gkrumbach07:
/hold

another comment

frontend/src/concepts/notebooks/StartNotebookModal.tsx
jrenee42: sure, updated

PR #4316, opened 2025-06-05
Edge cases for storage class access mode settings

DaoDaoNoCode:
JIRA: [RHOAIENG-26111](https://issues.redhat.com/browse/RHOAIENG-26111)

## Description

Cover the following edge cases:
- No annotation/No `accessModeSettings` field -> Patch the annotation/the field when loading the page without informing the users
- `accessModeSetting` field is present
    * The `ReadWriteOnly` key in the settings is not correct (set to false/random string/missing) -> Patch the `"ReadWriteOnly": true` to the `accessModeSettings` field and keep other fields unchanged.
    * Field value is not an object/Field value is empty/Missing supported keys (e.g. supports RWO but the key is not there at all)/Having unsupported keys (e.g. doesn’t support RWX but there is an RWX key in the object) -> Mark as invalid, don’t auto patch, show a warning for users to patch it. When users edit the field, it will check the default values in the field and patch on save. 
    Screenshot 2025-06-05 at 5 15 04 PM

## How Has This Been Tested?

Test all the invalid scenarios that you can imagine by editing the annotations and the `accessModeSettings` field. Make sure it covers all the situations.

## Test Impact

Added unit tests and a cypress test to cover the invalid situations.

## Request review criteria:

If you have UI changes: 


frontend/src/pages/storageClasses/utils.ts
```diff
@@ -26,10 +31,18 @@ export const getPossibleStorageClassAccessModes = (
...
   );
   const selectedStorageClassConfig = storageClass ? getStorageClassConfig(storageClass) : undefined;
-  const adminSupportedAccessModes = getAccessModeSettings(
-    openshiftSupportedAccessModes,
-    selectedStorageClassConfig?.accessModeSettings || { [AccessMode.RWO]: true },
-  );
+
+  // RWO is always supported
+  const adminSupportedAccessModes: AccessMode[] = [AccessMode.RWO];
+  for (const accessMode of openshiftSupportedAccessModes) {
```
srtanish1992, line 37: we can make it shorter and cleaner.

```
// RWO is always supported
const adminSupportedAccessModes: AccessMode[] = [
  AccessMode.RWO,
  ...openshiftSupportedAccessModes.filter(
    accessMode =>
      accessMode !== AccessMode.RWO &&
      selectedStorageClassConfig?.accessModeSettings?.[accessMode] === true
  )
];
```

frontend/src/pages/storageClasses/constants.tsx
```diff
@@ -67,8 +67,7 @@ export const columns: SortableData<StorageClassKind>[] = [
     sortable: false,
     info: {
       popoverProps: { headerContent: 'Default' },
-      popover:
-        'The default storage class is automatically selected for OpenShift AI users when creating new cluster storage.',
+      popover: `The default storage class is automatically selected for ${ODH_PRODUCT_NAME} users when creating new cluster storage.`,
```
srtanish1992, line 70: Thank you for making these changes!

PR #4327, opened 2025-06-09
Fix ODH Dashboard release notes and PR creation

Griffin-Sullivan:
## Request review criteria:

If you have UI changes: 

## Summary by CodeRabbit

- **Chores**
	- Updated release workflow to use version-specific branch names and tags for pull request and release creation, following a new `-odh-release` naming convention.

### DIFF
--- a/.github/workflows/release-odh-dashboard.yml
+++ b/.github/workflows/release-odh-dashboard.yml
@@ -116,7 +116,7 @@ jobs:
           git add manifests/odh/params.env frontend/.env
           git commit -m "Edit version info for ${{ github.event.inputs.RELEASE_VERSION }}"
           git push origin ${{ github.event.inputs.RELEASE_VERSION }}-PR
-          gh pr create --base main --head ${{ github.event.inputs.RELEASE_VERSION }}-PR \
+          gh pr create --base ${{ github.event.inputs.RELEASE_VERSION }}-odh-release --head ${{ github.event.inputs.RELEASE_VERSION }}-PR \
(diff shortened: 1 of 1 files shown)


.github/workflows/release-odh-dashboard.yml
```diff
@@ -134,8 +134,9 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
-          gh release create "${{ github.event.inputs.RELEASE_VERSION }}" \
+          gh release create "${{ github.event.inputs.RELEASE_VERSION }}-odh" \
               --repo="$GITHUB_REPOSITORY" \
-              --title="${{ github.event.inputs.RELEASE_VERSION }}" \
+              --title="${{ github.event.inputs.RELEASE_VERSION }}-odh" \
+              --target="${{ github.event.inputs.RELEASE_VERSION }}-odh-release" \
```
manaswinidas, line 140: what do we do about the previous tag? Do we leave it as it is coz it can be changed later? I think the previous tag is "auto" by default.

manaswinidas (changes requested):
Works fine. Just one small question.

.github/workflows/release-odh-dashboard.yml
manaswinidas: Yeah I think this can be changed later after publishing the release notes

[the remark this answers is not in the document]
Griffin-Sullivan: would still need someone to merge it which we don't wait for

PR #4334, opened 2025-06-10
fix: PVC resizing doesn't trigger a redeployment

mtalvi:
Related to the following:
https://issues.redhat.com/browse/NVPE-287
## Description

When a user deploys a NIM model with a PVC that's too small, the deployment fails during model download. Resizing a PVC from the UI doesn't trigger a redeployment of the serving.
Fixes Implemented:

- Trigger pod restart and force model re-download after PVC resize:
- Injected a single annotation (runtimes.opendatahub.io/force-redeploy) into the PVC metadata during resizing.
- This annotation is updated with a timestamp to ensure a meaningful change on every resize.
- The annotation causes the InferenceService to reconcile and restart the pod.
- Since the PVC metadata has changed, NIM performs a fresh download into the volume, resolving any stale or corrupted cache state.

### Important:
This solution is applicable only to our current operational way pf work, which utilizes a dedicated inference service per Persistent Volume Claim (PVC) and maintains isolated PVCs that are not shared across multiple inference services.

## How Has This Been Tested?

Tested locally.
## Test Impact

1. Make sure NIM is enabled.
2. Deploy a NIM model with insufficient storage and wait for the InferenceService to fail. (e.g. llama-3.1-8b-instruct with 1 GiB NVIDIA NIM storage size)
3. Resize the PVC via edit in the UI for 30 GiB, verify the new storage is available and wait for the deployment status to be loaded.  
## Request review criteria:

If you have UI changes: 

## Summary by CodeRabbit

- **New Features**
  - PersistentVolumeClaims now include a timestamp annotation indicating when they were created or updated.

### DIFF
--- a/frontend/src/pages/modelServing/screens/projects/NIMServiceModal/ManageNIMServingModal.tsx
+++ b/frontend/src/pages/modelServing/screens/projects/NIMServiceModal/ManageNIMServingModal.tsx
@@ -330,7 +330,6 @@ const ManageNIMServingModal: React.FC<ManageNIMServingModalProps> = ({
               'runtimes.opendatahub.io/force-redeploy': new Date().toISOString(),
             }),
           );
-          //promises.push(updatePvc(updatePvcData, pvc, namespace, { dryRun: false }, false, true)); // Pass forceRedeploy: true for NIM
         }
         return Promise.all(promises);
       })


frontend/src/pages/modelServing/screens/projects/NIMServiceModal/ManageNIMServingModal.tsx
```diff
@@ -326,6 +328,38 @@ const ManageNIMServingModal: React.FC<ManageNIMServingModalProps> = ({
...
+            .then((latestISvc) => {
+              const restartAnnotationPatch = [
+                {
+                  op: 'add',
+                  path: '/metadata/annotations/runtimes.opendatahub.io~1force-redeploy',
+                  value: new Date().toISOString(),
+                },
+                {
+                  op: 'add',
+                  path: '/metadata/annotations/runtimes.opendatahub.io~1redeploy-token',
```
TomerFi, line 347: What's the `redeploy-token` annotation used for?

mtalvi: That was my initial thought too but while testing the redeployment behavior after resizing a PVC, I initially followed the suggestion to use a single force-redeploy annotation to trigger a new revision of the InferenceService. This indeed restarted the pod, but the new revision failed with a consistent tokenizer-related error traced to the model's runtime initialization (JSONDecodeError during tokenizer loading).

After inspecting the logs, it’s clear that the model is still attempting to load from a previously cached state (e.g., /tmp/LLM-*), which appears to be in a corrupted or incomplete state due to the original deployment failing from insufficient PVC size.

To fully resolve this and ensure the NIM container perceives a clean state and re-downloads the model assets, I reintroduced a second annotation (redeploy-token) with a random value. This effectively changes the pod context and forces the runtime to bypass cached state.

mtalvi:
I was able to come up with a solution that includes a single annotation and updated the PR description as well. @TomerFi  thanks for the help. @emilys314 could you please review?

frontend/src/api/k8s/pvcs.ts
```diff
@@ -38,6 +38,7 @@ export const assemblePvc = (
       annotations: {
         'openshift.io/display-name': pvcName.trim(),
         ...(description && { 'openshift.io/description': description }),
+        'runtimes.opendatahub.io/force-redeploy': new Date().toISOString(), // still 1 annotation
```
emilys314, line 41: I know there are other parts of the app using the assemble / create / update pvc functions. I'm not sure how this would affect things. It may be a good idea to have this run conditionally based on an optional argument? (So that only NIM is applying this annotation)

Thoughts?

mtalvi: Agreed! You're totally right. I updated the code.

frontend/src/api/k8s/pvcs.ts
```diff
@@ -20,10 +20,21 @@ export const assemblePvc = (
   namespace: string,
   editName?: string,
   hideFromUI?: boolean,
+  forceRedeploy?: boolean, // New optional parameter for NIM use case
```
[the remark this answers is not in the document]
mtalvi: Done! 
BTW, I also added a video to the PR.

PR #4346, opened 2025-06-11
Add model metrics for KServe refactor

Griffin-Sullivan:
https://issues.redhat.com/browse/RHOAIENG-26522
## Description
This adds the link to the metrics page for a Kserve model in the refactor.

## How Has This Been Tested?
Go to a project with a model deployed
Enable the model serving refactor plugin
Make sure helper popover has right content and link takes you to the existing metrics page.

## Request review criteria:

If you have UI changes: 


katieperry4 (commented):
Working as expected with and without metrics enabled~
/lgtm

frontend/packages/kserve/extensions.ts
```diff
@@ -37,6 +37,8 @@ const extensions: (
         startHintDescription: 'Each model is deployed on its own model server',
         deployButtonText: 'Deploy model',
       },
+      useIsMetricsSupported: () =>
+        import('./src/metrics').then((m) => m.useIsKServeMetricsSupported),
```
emilys314, line 41: I think this should be its own extension. My latest reasoning being, if a new platform is being added, I try to picture how the different features would get implemented, and if any of the extensions can be reused. 

So for example the platform card would be a different chunk of implementation than the platform base / platform selection card. Also there is the chance a platform would just copy the kserve metrics extension, but it would have to implement its own `model-serving.platform`

Also then there is no resolving to do for this extension, similar to what i did to the the watch deployments

frontend/packages/kserve/src/metrics.ts
```diff
@@ -0,0 +1,7 @@
+import { SupportedArea } from '@odh-dashboard/internal/concepts/areas/types';
+import { useIsAreaAvailable } from '@odh-dashboard/internal/concepts/areas/index';
+
+export const useIsKServeMetricsSupported = (): boolean => {
+  const kserveMetricsSupported = useIsAreaAvailable(SupportedArea.K_SERVE_METRICS).status;
+  return kserveMetricsSupported;
+};
```
emilys314, line 7: I wonder if we could remove this hook and use the built in feature flag plugin functionality like for the tab. This would also be one less code resolve to bother with
```
  {
    type: 'app.project-details/tab',
    properties: {
      id: 'model-server', // same value as ProjectSectionID.MODEL_SERVER
      title: 'Models',
      component: () => import('./src/ModelsProjectDetailsTab'),
     // or  ?     reliantAreas: [SupportedArea.MODEL_SERVING],
    },
    flags: {
      required: [PLUGIN_MODEL_SERVING],
    },
```

katieperry4:
Changes look good and everything is still working! 
/lgtm

frontend/packages/modelServing/src/ModelsProjectDetailsTab.tsx
```diff
@@ -13,7 +13,7 @@ import {
 const ModelsProjectDetailsTab: React.FC = () => {
   const { currentProject } = React.useContext(ProjectDetailsContext);
 
-  const availablePlatforms = useExtensions(isModelServingPlatformExtension);
+  const [availablePlatforms] = useResolvedExtensions(isModelServingPlatformExtension);
```
Griffin-Sullivan, line 16: I don't think I need this actually might be a leftover change from last rework

emilys314: Yeah reverting it to `useExtensions(...)` makes sense to me, i think you changed it because there was a codeRef for the metrics, but it's not there anymore

ashley-o0o:
One small request if you wouldn’t mind. If you’re able to rebase with main, and update [this displayName](https://github.com/opendatahub-io/odh-dashboard/blob/main/frontend/packages/modelServing/src/components/overview/DeployedModelsSection.tsx#L65) to link to the model metrics page too?

emilys314:
/lgtm
/approve

I tested it out and it looks good

PR #4350, opened 2025-06-11
Add toast notification upon successful datascience pipeline configuration

d0w:
Issue: https://issues.redhat.com/browse/RHOAIENG-14423

## Description

Added toast notifications for the pipeline configuration process. One toast gets created when the custom resources are first created; a success or error toast is posted once the custom resource reports that the pipeline api service is ready or after a timeout period respectively. This works regardless of what page the user navigates to, what project they switch to (pipeline context independent), but does not work after page refresh.

**Videos updated 7-1 to integrate well with the added pipeline server statuses. There is no more info notification since there is already a loading indication.**

https://github.com/user-attachments/assets/d189ded9-7497-42bb-b3c0-df575ea0c1bd

No notification is shown if the modal (from the same project) is opened. If this pipeline server status modal is opened for another project, the notifications will show for the original project's pipeline server.

https://github.com/user-attachments/assets/07a75ce0-79db-4490-aefe-af949e67d811

<strike>Right after form submission</strike>:
Screenshot 2025-06-12 at 4 21 26 PM

Screenshot 2025-06-12 at 4 23 52 PM

Screenshot 2025-06-12 at 4 22 01 PM

## How Has This Been Tested?

1. Data science pipelines -> pipelines -> configure pipeline server (must have no server configured previously)
2. Fill in the details and submit the form
3. View the pending toast and wait for the success or error toast (you can switch pages during this time to confirm it works outside of the pipeline page).

- If you fill correct details, you should get a success notification once the pipeline server is ready.
- If you fill in incorrect details, you should get an error notification after a timeout period with the toast showing an error message (typically 5 min).
- Both success/error toasts should provide a link that takes you to the pipelines page for that given project.

### Testing methodology
Tested correct details, incorrect details, slightly correct details (correct endpoint, incorrect credentials). Each should provide a toast for the intended consequence and the proper links
- success -> success
- wrong key / wrong endpoint -> error

### Test suite changes
No unit tests added since uses NotificationWatcherContext which already has tests. Some cypress tests were added but do not cover full functionality of this change since the callback function relies on FAST_POLL_INTERVAL which is an environment variable set to 99999 in some test contexts such as cypress dev server.

## Request review criteria:

If you have UI changes: 

## Summary by CodeRabbit

- **New Features**
  - Added real-time notifications to monitor the readiness of newly created pipeline servers, including automatic status updates and navigation actions upon completion or error.
- **Bug Fixes**
  - Improved toast notification testing with new identifiers and helper methods for better test reliability and clarity.


frontend/src/__tests__/cypress/cypress/tests/mocked/pipelines/pipelines.cy.ts
```diff
@@ -146,6 +154,42 @@ describe('Pipelines', () => {
...
+    //     mockDataSciencePipelineApplicationK8sResource({
+    //       namespace: projectName,
+    //     }),
+    //   ]),
+    // ).as('pollReadyDSPA2');
+    //
+    // // cy.wait('@pollUnreadyDSPA');
+    // cy.wait('@pollReadyDSPA2');
+    //
+    // toastNotifications.findToastNotification(1).should('exist');
```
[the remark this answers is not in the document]
d0w: @rsun19 yea, i left that comment in on accident. it was meant to test the notifications but doesnt work on dev since the poll rate environment variable gets set to 99999. If you think we should still add the test case (if that env variable doesn't get in any other env for ex), then let me know.

frontend/src/concepts/pipelines/content/configurePipelinesServer/ConfigurePipelinesServerModal.tsx
```diff
@@ -94,6 +105,57 @@ export const ConfigurePipelinesServerModal: React.FC<ConfigurePipelinesServerMod
...
+
+                  // repoll
+                  return {
+                    status: NotificationResponseStatus.REPOLL,
+                  };
+                } catch (e) {
+                  return {
+                    status: NotificationResponseStatus.ERROR,
+                    title: `Error configuring pipeline server for ${pipelineNamespace}`,
+                    message: <>{e instanceof Error ? e.message : 'Unknown error'}</>,
```
srtanish1992, line 148: we can simplify to `e instanceof Error ? e.message : 'Unknown error'`

d0w: just made this change

frontend/src/__tests__/cypress/cypress/tests/mocked/pipelines/pipelines.cy.ts
```diff
@@ -124,6 +125,13 @@ describe('Pipelines', () => {
       expect(interceptions).to.have.length(2); // 1 dry-run request and 1 actual request
     });
 
+    toastNotifications
+      .findToastNotification(0)
+      .should(
+        'have.text',
+        'Info alert:Waiting on pipeline server resources for test-project-name...',
```
srtanish1992, line 132: @d0w Can we add tests for below cases?

1. When there is a error while pipeline configuration.
2. When the pipeline configuration is successful.
3. While the pipeline configuration is in progress, you navigate to a different page and you still see the success notification. And the link for pipeline takes the user to the right pipeline.

d0w: @srtanish1992 I tried this but couldn't get it to work due to the way the FAST_POLL_INTERVAL environment variable gets set in test servers. The register notification function I used doesn't make a call to the callback function until after 1 set interval (rather than invoking it first then waiting). So the tests I wrote would never pass simply since the code would never make the call to check for the notification.

frontend/src/concepts/pipelines/content/configurePipelinesServer/ConfigurePipelinesServerModal.tsx
```diff
@@ -92,8 +104,62 @@ export const ConfigurePipelinesServerModal: React.FC<ConfigurePipelinesServerMod
...
+              callbackDelay: FAST_POLL_INTERVAL || 3000,
+              callback: async (signal: AbortSignal) => {
+                try {
+                  // check if polling for too long
+                  if (
+                    Date.now() - new Date(obj.metadata.creationTimestamp || Date.now()).getTime() >
+                    SERVER_TIMEOUT
+                  ) {
+                    throw Error(`${pollingNamespace} pipeline server creation timed out`);
+                  }
```
DaoDaoNoCode, line 122: We have https://github.com/opendatahub-io/odh-dashboard/blob/main/frontend/src/concepts/pipelines/context/usePipelineNamespaceCR.ts#L14 to detect the time out, try to reuse this function. You can refer to the usage in `PipelinesContext` file.

jrenee42:
also:  do we want the toast showing when the modal showing the pipeline details is showing?  i would think not because then the user gets 2 sets of notifications which will be confusing @acrago  please comment

acrago:
@jrenee42 Is there a timer for how long the toast is displayed? If there is, then we'll just let the toast show for as long as it typically does, unless the user closes it out themselves

frontend/src/concepts/pipelines/content/configurePipelinesServer/ConfigurePipelinesServerModal.tsx
```diff
@@ -92,8 +108,54 @@ export const ConfigurePipelinesServerModal: React.FC<ConfigurePipelinesServerMod
...
             onBeforeClose();
+
+            const pollingNamespace = obj.metadata.namespace;
+            notification.info(`Waiting on pipeline server resources for ${pollingNamespace}...`);
+            registerNotification({
+              callbackDelay: FAST_POLL_INTERVAL || 3000,
+              callback: async (signal: AbortSignal) => {
+                try {
+                  const response = await listPipelinesCR(pollingNamespace, { signal });
+                  const isCRReady = dspaLoaded([response[0], true]);
```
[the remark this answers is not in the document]
jrenee42: exactly, plus don't want the toasts showing if the dialog is up

are you testing for failures too?  you can induce a failure by putting in an incorrect password in the initialization. 
I also adjusted the timeouts (temporarily in my new branch) so i don't have to wait five minutes each time (made it 100 seconds instead of 300; you can make it even shorter if you like (while testing, of course)). my new branch is here: https://github.com/jrenee42/odh-dashboard/tree/pipeline-failure-feedback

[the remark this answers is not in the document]
d0w: This checks for timeout failure using the logic from `usePipelineNamespaceCR.ts`'s server timed out function only. I mimicked what the pipeline configuration page was doing before, which was only showing failure after a 5 minute timeout but will adjust it to match your changes if the statuses update from events other than success or timeout.

frontend/src/__tests__/cypress/cypress/tests/mocked/pipelines/pipelines.cy.ts
srtanish1992: @d0w apologies for the late reply. I will take a look at this today. 
@FedeAlonso Tagging you to share your experience. Can we add the tests I have recommended above as cypress mock?

d0w: I didn't write the registerNotification function but to my understanding, the invoke function it uses calls itself according to the callback's response and there is some logic that uses a map to keep track of all the current timeouts if there are multiple notifications watchers registered. The code uses a setTimeout() to make this wait happen and uses the reference to the setTimeout to keep track of all notification watchers, and this doesn't call the function until after the timeout.

jrenee42 (commented):
unfortunately the toasts show up on top of the modal: 
Screenshot 2025-06-27 at 3 15 05 PM


also; there are 2 similar toasts/errors on the page on an unsuccessful load: (i clipped the video to show the highlights)

https://github.com/user-attachments/assets/c9cb028e-f0b0-4f19-b1ac-4c2b1c8c8eb8

if there could be some simple elegant way to pass the reference to the pipeline to the dialog; for the dialog to stop it when it is shown; then show it again after it is closed that would be best.

frontend/src/__tests__/cypress/cypress/tests/mocked/pipelines/pipelines.cy.ts
christianvogt: Without looking too closely at the details, shouldn't the app make a call to the callback which provides the status immediately and then start the poll if the response is inadequate? Firstly this would provide for a more responsive app if the request can be immediately fulfilled. Second this would allow a mock to be prepared and return a value immediately without having to wait for a poll interval.

d0w: I'll have to see what the implications are for adjusting the behavior to call first then wait. Changing it to call the callback immediately first does seem more reasonable.

frontend/src/concepts/pipelines/content/configurePipelinesServer/ConfigurePipelinesServerModal.tsx
```diff
@@ -92,8 +109,67 @@ export const ConfigurePipelinesServerModal: React.FC<ConfigurePipelinesServerMod
...
       .then((spec) => {
         createPipelinesCR(namespace, spec)
-          .then(() => {
+          .then((obj: DSPipelineKind) => {
             onBeforeClose();
+
+            const pollingNamespace = obj.metadata.namespace;
+            notification.info(`Waiting on pipeline server resources for ${pollingNamespace}...`);
+            registerNotification({
+              callbackDelay: FAST_POLL_INTERVAL || 3000,
```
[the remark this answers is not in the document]
d0w: The latest commit i added makes it so that new notifications for server configuration in the **same** namespace as the modal are not triggered when the modal is open. Also, since the loading indicator already tells the user that the pipeline server is starting, I've removed the info notification (purple) which should get rid of the overlap from the first photo.

Here are some caveats:
- Notifications triggered before will still show when the modal is opened (e.g. notifications from other unrelated events)
- Notifications not relevant to that namespace's pipeline server will still show when the modal is opened
    - This applies to success notifications for pipeline servers from other namespaces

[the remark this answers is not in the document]
jrenee42: sounds good; will take a look!

frontend/src/concepts/pipelines/content/configurePipelinesServer/ConfigurePipelinesServerModal.tsx
```diff
@@ -115,6 +115,7 @@ export const ConfigurePipelinesServerModal: React.FC<ConfigurePipelinesServerMod
               callbackDelay: FAST_POLL_INTERVAL,
               callback: async (signal: AbortSignal) => {
                 try {
+                  console.log('POLL');
```
jrenee42, line 118: extra console.log.  and there is a linter error too

frontend/src/concepts/pipelines/context/usePipelineNamespaceCR.ts
```diff
@@ -30,7 +30,7 @@ export const hasServerTimedOut = (
   }
 
   // If we are here, and 5 mins have past, we are having issues
-  return Date.now() - new Date(createTime).getTime() > SERVER_TIMEOUT;
+  return Date.now() - new Date(createTime).getTime() > 10000;
```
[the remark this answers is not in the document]
d0w: sorry accidentally left these in while testing

frontend/src/concepts/pipelines/content/configurePipelinesServer/ConfigurePipelinesServerModal.tsx
d0w: removed it

frontend/src/__tests__/cypress/cypress/tests/mocked/pipelines/pipelines.cy.ts
d0w: @srtanish1992 @christianvogt I've changed the code to invoke the callback immediately and then start polling. I've also added in tests to test the success notification for this PR.

frontend/src/concepts/pipelines/content/configurePipelinesServer/ConfigurePipelinesServerModal.tsx
```diff
@@ -92,8 +107,78 @@ export const ConfigurePipelinesServerModal: React.FC<ConfigurePipelinesServerMod
...
+                    };
+                  }
+
+                  if (serverLoaded && serverAllReady) {
+                    // If we're viewing the StartingStatusModal in the same namespace, we don't need to show the notification
+                    if (startingStatusModalOpenRef?.current === pollingNamespace) {
+                      return {
+                        status: NotificationResponseStatus.STOP,
+                      };
+                    }
```
Gkrumbach07, line 144: why do we not show the notification still. i think its fine to show no matter what. then we wouldnt need to have this tracked startingStatusModalOpenRef

d0w: As shown here: https://github.com/opendatahub-io/odh-dashboard/pull/4350#pullrequestreview-2966657860, the notifications show on top of the modal (when zoomed in more).

From @acrago 
> If the user happens to click the progress link within that 5 seconds, the toast can just stay present behind the modal until it clears itself @jrenee42 @d0w

Are we are still okay with the notification showing over the modal?

frontend/src/concepts/notificationWatcher/NotificationWatcherContext.tsx
```diff
@@ -63,50 +63,49 @@ export const NotificationWatcherContextProvider: React.FC<
     (itemToInvoke: NotificationWatcherItem, signal: AbortSignal) => {
       const callbackDelay = itemToInvoke.callbackDelay ?? POLL_INTERVAL;
 
-      const timeoutId = setTimeout(() => {
-        timeoutIdsMapRef.current.delete(itemToInvoke);
-
-        itemToInvoke
```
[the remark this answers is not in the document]
d0w: To address this issue: https://github.com/opendatahub-io/odh-dashboard/pull/4350#discussion_r2155429756

Instead of the function waiting for the timeout then invoking the callback, the callback is invoked first, then the timeout occurs to start the polling.

Gkrumbach07 (approved):

PR #4363, opened 2025-06-16
Reduce confusion on wording around partial resources

andrewballantyne:
@manosnoam here is your suggested update to wording.


manosnoam (approved):
/lgtm

PR #4413, opened 2025-06-25
RHOAIENG-25339: Update application of hardware profiles in all workloads

srtanish1992:
**JIRA:** https://issues.redhat.com/browse/RHOAIENG-25339

## Description

1. Setting the `opendatahub.io/hardware-profile-name` annotation to an empty string for a accelerator/legacy hardware profile so that the mutating webhook understands that this is not a hardware profile. 
2. Setting the tolerations and node selectors to undefined so that the webhook takes care of setting them.
3. Targeting workloads like Notebook and Serving runtime.
4. Is dependent on https://github.com/opendatahub-io/opendatahub-operator/pull/2130

## How Has This Been Tested?

**NOTE:** I have deployed an image created using https://github.com/opendatahub-io/opendatahub-operator/pull/2130 to a cluster. Please reach out to me to get access.

Testing scenarios - repeat them for Notebook (create and edit) and Model deployment (Multi model serving):

1. Create a hardware profile with tolerations and node selectors.
2. Edit a notebook and select the newly created hardware profile.
3. Go to the openshift console and find the Hardware profile resource. 
4. You should see the tolerations and node selectors set correctly under `spec`
---------------------------------------------
1. Create a accelerator profile with tolerations.
2. Edit a notebook and select the newly created accelerator profile.
3. Go to the openshift console and find the accelerator profile resource.
4. You should see the tolerations set correctly under `spec`
---------------------------------------------
1. Migrate a accelerator profile to an hardware profile.
2. This is a regression test to make sure the migration continues to work as expected.
---------------------------------------------
1. When editing notebook, swap between different hardware profiles and make sure the right properties are set for the Notebook especially tolerations and node selectors.
2. When editing notebook, swap between different accelerator profiles and make sure the right properties are set for the Notebook especially tolerations.
3. When editing notebook, swap between different accelerator and hardware profiles and make sure the right properties are set for the Notebook especially tolerations and node selectors.

## Test Impact

I have added unit tests for the new changes.

## Request review criteria:

If you have UI changes: 


frontend/src/api/k8s/notebooks.ts
```diff
@@ -94,6 +94,9 @@ export const assembleNotebook = (
     volumeMounts.push(getshmVolumeMount());
   }
 
+  const isLegacyHardwareProfile =
+    !!selectedAcceleratorProfile || !selectedHardwareProfile?.metadata.uid;
```
srtanish1992, line 98: Initially, a label ('opendatahub.io/is-legacy-profile': 'true') was added to profiles during migration. The problem was that this label was permanent, causing the system to forever treat these migrated profiles as "legacy," which was incorrect.

The new approach is:

1. It checks if the selected hardware profile is a temporary configuration created in the user interface (which won't have a unique ID, or uid, from the cluster yet).
2. It also checks if the configuration is coming directly from an old Accelerator Profile.

@Gkrumbach07 @nananosirova Let me know what you guys think about it.

Gkrumbach07: I think this works for now. I want to revist this next sprint and see if we can do it without this trick. I have been thinking about a way to do it and I think I found a elegant ish way. 

But to not hold this up, we can go with this for a MVP

PR #4419, opened 2025-06-26
Fix token issue when ISVC and SR are different names (#4209)

emilys314:
Fixes https://issues.redhat.com/browse/RHOAIENG-27682

Cherry pick of
https://issues.redhat.com/browse/RHOAIENG-23800
Description

When you have an ISVC and a SR that have different names, the token created in the kserve modal will not display. I had to revert one change from a recent PR that will need testing from the NIM team.
How Has This Been Tested?

    Select single model serving
    Create an ISVC and a SR with different names
    Go to edit the model
    Require token authentication
    Redeploy
    Edit the model again
    The service account input field should be visible

### DIFF
--- a/frontend/src/k8sTypes.ts
+++ b/frontend/src/k8sTypes.ts
@@ -557,6 +557,9 @@ export type InferenceServiceKind = K8sResourceCommon & {
   };
 };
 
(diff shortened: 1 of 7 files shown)


andrewballantyne (approved):
hah, 2.19 didn't have the updated OWNERS it seems -- approving for Griffin

PR #4421, opened 2025-06-26
Add stopping status to model deployments

Griffin-Sullivan:
https://issues.redhat.com/browse/RHOAIENG-27147
## Description
Adds the stopping status to the model serving row. Also reworked how some of the other states are determined since it was only based on the annotation being present before.

## How Has This Been Tested?
Deploy a model (notice the first deploy you will get unknown status for a sec)
wait for it to run
stop the model
start the model
stop in the middle of starting, etc

Need more testing on what happens after editing

## Test Impact
Only updated the unit tests so far. We have an e2e test coming soon and maybe we need some mock tests not sure.

## Request review criteria:

If you have UI changes: 

## Summary by CodeRabbit

* **New Features**
  * Introduced new hooks to manage and fetch model serving and pod statuses, providing more granular control and visibility into model lifecycle states.
  * Enhanced status indicators for models, including explicit "starting," "stopping," and "stopped" states in UI components.

* **Bug Fixes**
  * Improved the logic for disabling action toggles based on model status, allowing for conditional disabling during the starting state.

* **Tests**
  * Added comprehensive tests for the new model status hook.
  * Updated and simplified existing tests to reflect new status management logic.

* **Refactor**
  * Centralized and streamlined model status management across components for consistency and maintainability.
  * Updated components to use new status props and hooks, removing deprecated properties and logic.

### DIFF
--- a/frontend/src/concepts/modelServing/ModelStatusIcon.tsx
+++ b/frontend/src/concepts/modelServing/ModelStatusIcon.tsx
@@ -6,9 +6,9 @@ import {
   OffIcon,
   OutlinedQuestionCircleIcon,
   PlayIcon,
+  SyncAltIcon,
 } from '@patternfly/react-icons';
 import { InferenceServiceModelState } from '#~/pages/modelServing/screens/types';
-import { InferenceServiceKind } from '#~/k8sTypes';
 
 type ModelStatusIconProps = {
   state: InferenceServiceModelState;
@@ -16,8 +16,9 @@ type ModelStatusIconProps = {
   bodyContent?: string;
   isCompact?: boolean;
   onClick?: LabelProps['onClick'];
-  inferenceService?: InferenceServiceKind;
   isStarting?: boolean;
+  isStopping?: boolean;
+  isStopped?: boolean;
 };
 
 export const ModelStatusIcon: React.FC<ModelStatusIconProps> = ({
@@ -26,8 +27,9 @@ export const ModelStatusIcon: React.FC<ModelStatusIconProps> = ({
   bodyContent = '',
   isCompact,
   onClick,
-  inferenceService,
   isStarting,
+  isStopping,
+  isStopped,
 }) => {
   const statusSettings = React.useMemo((): {
     label: string;
@@ -37,13 +39,20 @@ export const ModelStatusIcon: React.FC<ModelStatusIconProps> = ({
     message?: string;
   } => {
     // Highest-priority: service explicitly stopped
-    const isStopped = inferenceService?.metadata.annotations?.['serving.kserve.io/stop'] === 'true';
     if (isStopped) {
       return {
         label: 'Stopped',
         color: 'grey',
         icon: <OffIcon />,
         message: 'Offline and not using resources. Restart to use model.',
+      };
+    }
+
+    if (isStopping) {
+      return {
+        label: 'Stopping',
+        color: 'grey',
+        icon: <SyncAltIcon className="odh-u-spin" />,
       };
     }
     // Show 'Starting' for optimistic updates or for loading/pending states from the backend.
@@ -82,7 +91,7 @@ export const ModelStatusIcon: React.FC<ModelStatusIconProps> = ({
           icon: <OutlinedQuestionCircleIcon />,
         };
     }
-  }, [state, defaultHeaderContent, inferenceService, isStarting]);
+  }, [state, defaultHeaderContent, isStarting, isStopping, isStopped]);
 
   return (
     <Popover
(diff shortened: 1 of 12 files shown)


frontend/src/pages/modelServing/useModelPodStatus.ts
```diff
@@ -0,0 +1,22 @@
...
+ * Returns { data: pod, loaded, error, refresh }.
+ * If no pod is found, data is null.
+ */
+const useModelPodStatus = (
+  namespace: string,
+  inferenceServiceName: string,
+): FetchStateObject<PodKind | null, Error> => {
+  const fetchPod = React.useCallback(async () => {
+    const pods = await getPodsForKserve(namespace, inferenceServiceName);
+    return pods.length > 0 ? pods[0] : null;
```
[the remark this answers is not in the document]
Griffin-Sullivan: I'm not sure about which one will be 0th, but Andrew did mention we should only need to be checking for the existence of one of the pods. I think 0th might be the old pod if you edit which isn't a bad thing in case you submit a bad deployment.

frontend/src/pages/modelServing/utils.ts
```diff
@@ -371,14 +374,22 @@ export const isOciModelUri = (modelUri?: string): boolean => !!modelUri?.include
...
+  const isRunning =
+    state === InferenceServiceModelState.LOADED || state === InferenceServiceModelState.STANDBY;
+
   return {
     inferenceService,
-    isStopped: status,
-    isRunning: !status,
-    isStopping: false,
-    isStarting: false,
+    isStopped: status && !modelPod,
```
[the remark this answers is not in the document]
Griffin-Sullivan: yeah I can update this and add comments to make it more readable

Griffin-Sullivan:
@ashley-o0o the first comment is definitely something that flakes. The check for the label and the button use different state objects so they can get out of sync for small amounts of time. might need to rewrite it so they use the same state object.

The second comment I'm not sure how we can avoid that on refresh. It's flipping to stopped since we make that check first for a few good reasons, but changes to stopping once it's fetched all the data it needs.

frontend/src/pages/modelServing/useModelStatus.ts
```diff
@@ -0,0 +1,87 @@
...
+    }
+    if (pollingInterval) {
+      return () => {
+        clearInterval(pollingInterval);
+        setPollingInterval(null);
+      };
+    }
+    return undefined;
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [isStopping, refreshModelPodStatus, refresh, modelPodStatus]);
```
Griffin-Sullivan, line 52: adding pollingInterval here breaks functionality

andrewballantyne:
> Screencast.From.2025-07-09.09-46-58.mp4

Looking good -- I'm not seeing any exposed routes -- something we'd need to fetch... we should make sure we test with external routes (the route checkbox in the modal checked on)

frontend/src/concepts/modelServing/ModelStatusIcon.tsx
```diff
@@ -6,18 +6,19 @@ import {
...
 import { InferenceServiceModelState } from '#~/pages/modelServing/screens/types';
-import { InferenceServiceKind } from '#~/k8sTypes';
 
 type ModelStatusIconProps = {
   state: InferenceServiceModelState;
   defaultHeaderContent?: string;
   bodyContent?: string;
   isCompact?: boolean;
   onClick?: LabelProps['onClick'];
-  inferenceService?: InferenceServiceKind;
```
emilys314, line 19: The removal of the inferenceServiceKind is nice , should help with adding this to the refactor. plus the folder is 'modelServing' not 'kserve'

frontend/src/pages/modelServing/useModelStatus.ts
```diff
@@ -0,0 +1,87 @@
...
+  React.useEffect(() => {
+    if (isStopping) {
+      const interval = setInterval(async () => {
+        await refreshModelPodStatus();
+        if (!modelPodStatus) {
+          setIsStopping(false);
+        }
+      }, FAST_POLL_INTERVAL);
+
+      setPollingInterval(interval);
```
emilys314, line 37: Nice this works pretty good. 

Slight nit / observation: this is kinda inherit to the JS naming, but I do find it slightly confusing to read. It's easy to mix up `pollingInterval` would be a duration, like `FAST_POLL_INTERVAL` when it's actually a unique id of a JS polling handle.. Not sure there is a solution to this since it's JS naming

emilys314 (commented):
looks good!

emilys314:
> > Screencast.From.2025-07-09.09-46-58.mp4
> 
> Looking good -- I'm not seeing any exposed routes -- something we'd need to fetch... we should make sure we test with external routes (the route checkbox in the modal checked on)

Oh good callout, when the model is stopped, it seems like it's stuck in a loader forever?
image
image

/remove-lgtm

PR #4502, opened 2025-07-21
feat: implement dynamic notebook image selection with backend fallback

antowaddle:
- Replace hardcoded 'jupyter' image validation with dynamic selection
- Add selectNotebookImageWithBackendFallback utility function:
  - Try UI selection first (createSpawnerPage.findNotebookImage)
  - Fall back to backend query if UI element not found
  - Uses opendatahub.io/notebook-image-name annotation for display name
- Update workbench tests to use scoped validation (notebookRow.shouldHaveNotebookImageName)
- Centralize all OC commands in imageStreams.ts utility per Cypress E2E rules
- Handle environment variables (APPLICATIONS_NAMESPACE) internally in utilities
- Fix async/sync mixing issues in Cypress command chaining

Resolves: RHOAIENG-30224

## Request review criteria:

If you have UI changes: 


antowaddle:
> @antowaddle looks good, what happens if the image selection/version dropdown itself is broken ? Do we catch that ?

@ConorOM1 if the Image dropdown itself is broken the test will simply fail telling us that the element is not selectable or that the testID for the dropdown is not available. The key logic change here is simply adding a fallback in the case that our default image is not selectable and then adding dynamic validation of the selected image.

manosnoam (approved):

PR #4533, opened 2025-07-23
Port start / stop models to the plugin

Griffin-Sullivan:
https://issues.redhat.com/browse/RHOAIENG-28897
## Description

Porting over the start / stop models feature to the model serving plugin

Note: UX fixes will be implemented in future PR 
## How Has This Been Tested?
Deploy a model in a separate non-plugin window and make sure model goes into starting
Ensure model mesh rows in global serving don't have the start / stop toggle
Start models and make sure they go into starting > started
Stop models and make sure they go into stopping > stopped

## Test Impact
Old e2e test and mocked tests should still work

## Request review criteria:

If you have UI changes: 


frontend/packages/modelServing/extension-points/index.ts
```diff
@@ -12,10 +12,12 @@ import type { ModelServingPodSpecOptionsState } from '@odh-dashboard/internal/co
 import type { K8sResourceCommon } from '@openshift/dynamic-plugin-sdk-utils';
 import type { InferenceServiceModelState } from '@odh-dashboard/internal/pages/modelServing/screens/types';
 import type { ComponentCodeRef } from '../../plugin-core/src/extension-points/types';
+import { ToggleState } from '@odh-dashboard/internal/components/StateActionToggle';
 
 export type DeploymentStatus = {
   state: InferenceServiceModelState;
   message?: string;
+  stoppedStates?: ToggleState;
```
Griffin-Sullivan, line 20: hate this name and not sure if it makes sense to live under here

Griffin-Sullivan:
Just noticed when I clicked start it went into unknown for a second so still needs a little work to figure out why it does that since it's flaky

frontend/packages/modelServing/src/components/deployments/DeploymentsTableRow.tsx
```diff
@@ -107,8 +131,26 @@ export const DeploymentRow: React.FC<{
...
           />
         </Td>
+        <Td dataLabel="State toggle">
+          {resolvedStartStopActionExtension && (
+            <StateActionToggle
+              currentState={deployment.status?.stoppedStates ?? {
+                isRunning: true,
+                isStopped: false,
+                isStarting: false,
+                isStopping: false,
```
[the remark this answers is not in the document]
Griffin-Sullivan: Decided to not render the button until we have deployment status

frontend/packages/modelServing/extension-points/index.ts
Griffin-Sullivan: maybe we removed the layer for `stoppedStatus` and just union with the ToggleState so it becomes 
```suggestion
  export type DeploymentStatus = {
      state: InferenceServiceModelState;
      message?: string;
      isStarting: boolean;
      isStopping: boolean;
      isRunning: boolean;
      isStopped: boolean;
```

frontend/packages/modelServing/extension-points/index.ts
```diff
@@ -195,3 +197,18 @@ export type ModelServingMetricsExtension<D extends Deployment = Deployment> = Ex
 export const isModelServingMetricsExtension = <D extends Deployment = Deployment>(
   extension: Extension,
 ): extension is ModelServingMetricsExtension<D> => extension.type === 'model-serving.metrics';
+
+export type ModelServingStartStopAction<D extends Deployment = Deployment> = Extension<
+  'model-serving.deployments-table/start-stop-action',
+  {
+    platform: D['modelServingPlatformId'];
+    patchDeploymentStoppedStatus: CodeRef<
+      (deployment: D, isStopped: boolean) => Promise<D['model']>
```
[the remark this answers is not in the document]
Griffin-Sullivan: I could at least rename to be more open ended?

frontend/packages/kserve/src/deploymentStatus.ts
```diff
@@ -20,5 +22,22 @@ export const getKServeDeploymentStatus = (
...
+    // ISVC doesn't have annotation and state is LOADED
+    isRunning: state === InferenceServiceModelState.LOADED && !isStopped,
+    // ISVC has annotation and there are no pods
+    isStopped: isStopped && !deploymentPod,
+    // ISVC doesn't have annotation and state is PENDING, LOADING, STANDBY or UNKNOWN
+    isStarting:
+      (state === InferenceServiceModelState.PENDING ||
+        state === InferenceServiceModelState.LOADING ||
+        state === InferenceServiceModelState.STANDBY ||
+        state === InferenceServiceModelState.UNKNOWN) &&
```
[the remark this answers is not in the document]
Griffin-Sullivan: Note this was added because of the comment above about flakily going into "Unknown" status for a split second

frontend/packages/modelServing/src/components/overview/DeployedModelsSection.tsx
```diff
@@ -56,6 +56,9 @@ const DeployedModelCard: React.FC<{ deployment: Deployment }> = ({ deployment })
               <ModelStatusIcon
                 state={deployment.status?.state ?? InferenceServiceModelState.UNKNOWN}
                 bodyContent={deployment.status?.message}
+                isStarting={deployment.status?.stoppedStates?.isStarting}
+                isStopping={deployment.status?.stoppedStates?.isStopping}
+                isStopped={deployment.status?.stoppedStates?.isStopped}
```
emilys314, line 61: Whenever I see this I wonder if we can condense this into 1 input parameter, similar to the `state` in this component.

it seems inefficient and wordy to pass each boolean separately, when they would all be mutually exclusive like an enum?

frontend/packages/modelServing/src/components/deployments/DeploymentsTableRow.tsx
```diff
@@ -50,9 +55,28 @@ export const DeploymentRow: React.FC<{
     isModelServingAuthExtension,
     deployment,
   );
+  const [resolvedStartStopActionExtension] = useResolvedDeploymentExtension(
+    isModelServingStartStopAction,
+    deployment,
+  );
```
emilys314, line 61: I don't think it's necessary to use the resolved version of this extension hook because this is an async promise that it's returning.

So because resolving it is also a promise, you can just chain the 2 together like so (syntax may be off)
```suggestion
  const tartStopActionExtension = useDeploymentExtension(
    isModelServingStartStopAction,
    deployment,
  );
  
  const onStart = () => resolvedStartStopActionExtension?.properties.patchDeploymentStoppedStatus().then((resolvedFunction) => resolvedFunction(deployment));
```

This way the network call to fetch the function is only done when you are actually using the function, on the onClick / onStart

emilys314: This would also allow you to determine instantly if the StartStopActionExtension exists because it moves to the nonResolved hook

frontend/packages/modelServing/src/components/overview/DeployedModelsSection.tsx
Griffin-Sullivan: True I think this is leftover from how painful it was without refactor code :laughing: 

I can change the components to be more flexible with how to pass it

frontend/packages/modelServing/src/components/deployments/DeploymentsTableRow.tsx
Griffin-Sullivan: I tried this way but couldn't remember the syntax to get the function to resolve. Thanks!

frontend/packages/kserve/src/deploymentsTable.ts
```diff
@@ -0,0 +1,33 @@
...
+  {
+    field: 'servingRuntime',
+    label: 'Serving runtime',
+    sortable: false,
+    cellRenderer: (deployment: KServeDeployment) =>
+      deployment.server?.metadata.annotations?.['opendatahub.io/template-display-name'] ?? '-',
+  },
+];
+
+export const patchDeploymentStoppedStatus = (
```
katieperry4, line 16: Why is there a `patchDeploymentStoppedStatus` in `deploymentsTable` and in `deploymentStatus`? The one in `deploymentsTable` is being used in 
```
 {
    type: 'model-serving.deployments-table/start-stop-action',
    properties: {
      platform: KSERVE_ID,
      patchDeploymentStoppedStatus: () =>
        import('./src/deploymentsTable').then((m) => m.patchDeploymentStoppedStatus),
    },
  },
```
in `kserve>extensions.ts` but the one in `deploymentStatus` I don't see being used anywhere?

Griffin-Sullivan: Lol one of the files got deleted when I rebased and I think I duplicated it by accident

ashley-o0o:
Overall everything seems to be working well! One small thing I noticed when testing is that with stopped models, the model name isn't a hyperlink.  

Refactor enabled: 
Screenshot 2025-07-28 at 3 13 04 PM

Old code:
Screenshot 2025-07-28 at 3 13 16 PM

PR #4582, opened 2025-07-30
Fix editing isvc to maintain extraneous UI properties

Griffin-Sullivan:
https://issues.redhat.com/browse/RHOAIENG-30257
## Description
Adding a fix for maintaining non-dashboard properties on an ISVC when editing. Maybe there's more that I need to keep? I'm also not sure we need to do this for serving runtimes?

## How Has This Been Tested?
Create a project via OC
Create a ServingRuntime & InferenceService with as much bare bones as possible
Add a custom annotation & label (doesn't matter what)
Include dashboard label (related verification is [RHOAIENG-26146](https://issues.redhat.com/browse/RHOAIENG-26146))
Deploy resources into the project via OC
Go to the UI & edit the resource
Make no change if possible, but maybe change replicas if needed
Verify the resource still mains your custom annotation & label
Repeat for a property in the spec. See https://issues.redhat.com/browse/RHOAIENG-18976

## Test Impact
We should probably add a mock test for this

## Request review criteria:

If you have UI changes: 


frontend/src/api/k8s/inferenceServices.ts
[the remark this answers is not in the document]
Griffin-Sullivan: Yeah it seems to be kinda funky doing it this way

frontend/src/api/k8s/inferenceServices.ts
```diff
@@ -225,18 +204,18 @@ export const assembleInferenceService = (
   if (!isModelMesh && podSpecOptions) {
     const { resources } = podSpecOptions;
 
-    updateInferenceService.spec.predictor.model = {
-      ...updateInferenceService.spec.predictor.model,
+    updatedInferenceService.spec.predictor.model = {
+      ...updatedInferenceService.spec.predictor.model,
       resources,
```
emilys314, line 209: The resources doesn't preserve the `claims` field, everything else is working though
image

frontend/src/api/k8s/inferenceServices.ts
emilys314, line 1: Also I think adding some unit tests for the new functionality should be fairly straightforward and valuable

emilys314 (commented):
Besides the one resources property everything seems to work good

Also there seems to be a CI test failure

frontend/src/api/k8s/inferenceServices.ts
```diff
@@ -133,89 +146,55 @@ export const assembleInferenceService = (
...
+    name: format.name,
+    ...(format.version && { version: format.version }),
+  };
+  model.runtime = servingRuntimeName;
+
+  if (uri) {
+    model.storageUri = uri;
+    delete model.storage;
+  } else {
+    delete model.storageUri;
```
Griffin-Sullivan, line 175: hmmm I don't think this is working actually...

Griffin-Sullivan: Nevermind this is because the ConnectionSection had a bug where selecting a new existing connection didn't clear the URI data

frontend/src/__tests__/cypress/cypress/tests/mocked/modelServing/modelServingGlobal.cy.ts
```diff
@@ -469,7 +469,8 @@ describe('Model Serving Global', () => {
     inferenceServiceModalEdit.findSubmitButton().should('be.disabled');
     inferenceServiceModalEdit.findConnectionNameInput().type('Test Name');
     inferenceServiceModalEdit.findConnectionFieldInput('URI').type('/');
-    inferenceServiceModalEdit.findSubmitButton().click().should('be.disabled');
+    inferenceServiceModalEdit.findConnectionFieldInput('URI').blur();
+    inferenceServiceModalEdit.findSubmitButton().should('be.disabled');
```
Griffin-Sullivan, line 473: I'm not sure why I hit this? Did I mess something up with validation? You have to blur the field for the submit button to disable or else it will be enabled

Griffin-Sullivan: Same behavior in my PR too so not sure what changed but from user perspective it's the same and I updated the test to be more reliable

emilys314 (approved):

PR #4737, opened 2025-08-20
Add support for persisting chat context

divyanshiGupta:
## Request review criteria:

If you have UI changes: 


frontend/packages/llama-stack-modular-ui/frontend/src/app/Chatbot/hooks/useChatbotMessages.ts
```diff
@@ -106,7 +110,7 @@ const useChatbotMessages = ({
         setIsMessageSendButtonDisabled(false);
       }
     },
-    [isRawUploaded, modelId, selectedSourceSettings, systemInstruction],
+    [isRawUploaded, messages, modelId, selectedSourceSettings, systemInstruction],
```
DaoDaoNoCode, line 113: Might not related to the PR but since you made the change here - I don't see any benefits of using the `React.useCallback` on this function, maybe we could get rid of it.
It's called [here](https://github.com/opendatahub-io/odh-dashboard/blob/main/frontend/packages/llama-stack-modular-ui/frontend/src/app/Chatbot/ChatbotMain.tsx#L135-L139), which only triggers when you click on the send button, thus, I don't see a value to memoize the callback.

divyanshiGupta:
Converting this PR to draft since the BFF apis are being refactored by this PR https://github.com/opendatahub-io/odh-dashboard/pull/4741 and I will be replacing `/api/v1/query` with `/genai/v1/responses` and updating its usage accordingly.

frontend/packages/llama-stack-modular-ui/frontend/src/app/services/llamaStackService.ts
```diff
@@ -35,11 +35,14 @@ export const getModels = (): Promise<LlamaModel[]> => {
  * @returns Promise<LlamaModel[]> - Array of models matching the specified type
  * @throws Error - When the API request fails or returns an error response
  */
+//TODO: This does not work as expected. It returns all models, not just the ones of the specified type.
+//Should be fixed by updating the API to return only the models of the specified type.
+//Leaving this here as a reminder for now as it is not being used anywhere.
 export const getModelsByType = (modelType: LlamaModelType): Promise<LlamaModel[]> => {
-  const url = `${URL_PREFIX}/api/v1/models?model_type=${modelType}`;
+  const url = `${URL_PREFIX}/genai/v1/models?model_type=${modelType}`;
```
divyanshiGupta, line 42: @Schimuneck can you confirm if we are going to have this filtering capability in models endpoint later?

Schimuneck: Still not clear. We will not have for now.

frontend/packages/llama-stack-modular-ui/frontend/src/app/utilities/axios.ts
```diff
@@ -1,4 +1,8 @@
 // eslint-disable-next-line no-restricted-imports
 import axios from 'axios';
 
-export default axios.create();
+export default axios.create({
+  headers: {
+    Authorization: `Bearer ${process.env.AUTH_TOKEN || 'dummy-token'}`,
```
rohitkrai03, line 6: I don't think env will have AUTH_TOKEN in browser. Maybe the backend works right now with dummy-token but we'll probably have to figure out how to get actual user auth token from the application context. We had some conversation about this on [slack](https://redhat-internal.slack.com/archives/C07C0447EAV/p1755179645078049) but not sure of the exact way forward yet.

PR #4755, opened 2025-08-25
Ensure Archive model and Archive model version work correctly

manaswinidas:
Closes: [RHOAIENG-32742](https://issues.redhat.com/browse/RHOAIENG-32742)

## Description

When there are deployments associated with any model/model version - the archive buttons should be disabled
Screenshot 2025-08-25 at 9 01 00 PM

Screenshot 2025-08-25 at 8 56 04 PM

Screenshot 2025-08-25 at 8 56 10 PM

Screenshot 2025-08-25 at 8 56 37 PM

Screenshot 2025-08-25 at 8 56 53 PM

## How Has This Been Tested?

1. Run the upstream KF MR module
2. The archive model header action and archive model kebab action along with archive model version header action and archive model version kebab action should be consistent

## Test Impact

Added unit tests for the newly added deployment util

## Request review criteria:

If you have UI changes: 

## Summary by CodeRabbit

- New Features
  - Deployment-aware detection added to drive archive gating and UI behavior across lists, versions, and details.

- Bug Fixes
  - Archive actions stay disabled while deployment data loads to avoid UI flicker.
  - Tooltip messaging clarified when archiving is blocked by deployments.

- Refactor
  - Row rendering centralized via a deployment-aware wrapper; pages wrapped with a deployments provider.
  - Header actions simplified (deployment flag removed); rows accept a loading-aware prop.

- Tests
  - Comprehensive tests added for deployment detection and loading-state behavior.


yih-wang:
@manaswinidas Should be "Deployed **model versions**..." - we've refreshed all "versions" to "model versions" at some point.

packages/model-registry/upstream/frontend/src/app/pages/modelRegistry/screens/ModelVersionDetails/ModelVersionDetails.tsx
```diff
@@ -25,6 +26,8 @@ import ModelVersionSelector from './ModelVersionSelector';
 import ModelVersionDetailsTabs from './ModelVersionDetailsTabs';
 import ModelVersionsDetailsHeaderActions from './ModelVersionDetailsHeaderActions';
 import { MRDeployButton } from '~/odh/components/MRDeployButton';
+import { KnownLabels } from '~/odh/k8sTypes';
```
mturley: @manaswinidas as coderabbit says, this import is duplicated.

packages/model-registry/upstream/frontend/src/app/pages/modelRegistry/screens/ModelVersionDetails/ModelVersionDetails.tsx
```diff
@@ -33,7 +36,7 @@ type ModelVersionsDetailProps = {
   'breadcrumb' | 'title' | 'description' | 'loadError' | 'loaded' | 'provideChildrenPadding'
 >;
 
-const ModelVersionsDetails: React.FC<ModelVersionsDetailProps> = ({ tab, ...pageProps }) => {
+const ModelVersionsDetailsContent: React.FC<ModelVersionsDetailProps> = ({ tab, ...pageProps }) => {
```
mturley, line 38: I was about to suggest we avoid renaming this and instead change the default export to something like ModelVersionsDetailsWrapper, but this is the same impact so it's fine

packages/model-registry/upstream/frontend/src/app/pages/modelRegistry/screens/RegisteredModels/RegisteredModelTable.tsx
```diff
@@ -4,40 +4,65 @@ import { ModelVersion, RegisteredModel } from '~/app/types';
...
+        );
+      }}
+    />
+  );
+};
+
+const RegisteredModelTable: React.FC<RegisteredModelTableProps> = (props) => {
+  // Use deployment context from parent (RegisteredModelListView)
+  return <RegisteredModelTableContent {...props} />;
+};
```
mturley, line 66: This is a pretty considerable diff, and the RegisteredModelTable wrapper isn't doing anything. I wonder if we can change this so we don't have it touching every line of the rendered `<Table>`. Can we move the hook calls and `hasRegisteredModelDeployment` stuff into your wrapper, and maybe just have the wrapper component create and pass in an alternate rowRenderer function that will be used instead if present?

packages/model-registry/upstream/frontend/src/app/pages/modelRegistry/screens/ModelVersionDetails/ModelVersionDetails.tsx
```diff
@@ -142,4 +145,19 @@ const ModelVersionsDetails: React.FC<ModelVersionsDetailProps> = ({ tab, ...page
   );
 };
 
+const ModelVersionsDetails: React.FC<ModelVersionsDetailProps> = (props) => {
+  const { preferredModelRegistry } = React.useContext(ModelRegistrySelectorContext);
+  const { modelVersionId: mvId } = useParams();
+  
+  const labelSelectors = React.useMemo(() => {
+    return mvId ? { [KnownLabels.MODEL_VERSION_ID]: mvId } : undefined;
+  }, [mvId]);
```
mturley, line 153: ~~I think the labelSelectors have to always also look for the label with the model registry name, to prevent showing the wrong thing if the user has multiple registries and deployments from the same model/version ids that exist in another registry. `[KnownLabels.MODEL_REGISTRY_NAME]: preferredModelRegistry`. We need to make sure that's included everywhere we are using MRDeploymentsContextProvider.~~

Huh, interesting... this isn't necessary because that filter gets applied all the way down in the kserve package. [Slack thread about it](https://redhat-internal.slack.com/archives/C07BFD5J4CB/p1756928453068389). tech debt for another day...

mturley, line 153: Don't we also need to filter on KnownLabels.REGISTERED_MODEL_ID here in the labelSelectors?

packages/model-registry/upstream/frontend/src/odh/utils/deploymentUtils.ts
```diff
@@ -0,0 +1,81 @@
...
+      // Conservative approach: assume deployment exists during loading to prevent flicker
+      // This keeps the archive button disabled until we're sure there are no deployments
+      return { hasDeployment: true, loaded };
+    }
+
+    const hasDeployment = deployments.some(deployment => {
+      const isInferenceService = deployment.model?.kind === 'InferenceService';
+      const modelVersionId = deployment.model?.metadata?.labels?.[KnownLabels.MODEL_VERSION_ID];
+      return isInferenceService && modelVersionId === mvId;
+    });
```
mturley, line 24: Hmm, don't we also need to check that it matches the `KnownLabels.REGISTERED_MODEL_ID`?

Also, I'm a little confused about why this filtering is necessary. Don't we already have the filtering we need achieved by the labelSelectors we passed into MRDeploymentsContextProvider, which is what affects the deployments we get from `useDeploymentsState`? Maybe I'm missing something there.

packages/model-registry/upstream/frontend/src/odh/utils/deploymentUtils.ts
```diff
@@ -0,0 +1,81 @@
...
+  }, [deployments, loaded]);
+
+  return {
+    hasModelVersionDeployment,
+    hasRegisteredModelDeployment,
+    hasRegisteredModelDeploymentByVersionIds,
+    loaded,
+    deployments
+  };
+};
```
mturley, line 81: If this hook is indeed necessary, I wonder if these two functions `hasRegisteredModelDeployment` and `hasRegisteredModelDeploymentByVersionIds` are redundant. They mostly have the same behavior, maybe we could factor out the duplicated logic and have `hasRegisteredModelDeployment` mostly just get a list of version ids to pass into the current contents of `hasRegisteredModelDeploymentByVersionIds`.

But yeah in general I think we might be filtering by these labels twice. Not sure if I'm missing some piece where we couldn't filter by version id in the context provider or something.

packages/model-registry/upstream/frontend/src/odh/utils/deploymentUtils.ts
```diff
@@ -0,0 +1,81 @@
+import * as React from 'react';
+import { useDeploymentsState } from '../hooks/useDeploymentsState';
+import { KnownLabels } from '../k8sTypes';
+import { ModelVersion } from '../../app/types';
+
+export type DeploymentDetectionResult = {
```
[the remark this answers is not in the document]
manaswinidas: Thanks.

mturley (approved):
Pushed a commit to use isAriaDisabled instead of isDisabled on the ArchiveButtonDropdownItem so the tooltip will work properly. Otherwise looks good - we may want to look at cleaning up some of this logic in the future but the tests are pretty comprehensive and this will get us where we need to be :)

PR #4765, opened 2025-08-26
Add deployments tab to Model details page

manaswinidas:
Closes: [RHOAIENG-30977](https://issues.redhat.com/browse/RHOAIENG-30977)

## Description

- Add new "Deployments" tab to model details page that displays deployments filtered by registered model ID (similar to model version deployments) but with an extra version name column
- Implement dynamic extension system for model-level detail tabs
- Add custom filtering (name, project) and inline pagination controls
- Add alert explaining filtered deployment view in both model details deployments tab and version details deployments tab
- Remove deploy buttons from both model and model version deployment views for consistent read-only experience - adhering to [UX](https://www.figma.com/design/i7k7FnqqVi47JAUwmNQuCc/Model-Registry-2025?node-id=1625-138445&t=yyxCHNLn30sUd5ei-0)

Screenshot 2025-08-26 at 9 39 11 PM

## How Has This Been Tested?

1. Run MR KF plugin
2. Click on any model and you can now see the Deployments tab with models filtered by registered model ID
3. Ensure that you see the alert on both model details deployments tab and model version deployments tab

## Test Impact

Will be covered in another ticket

## Request review criteria:

If you have UI changes: 


manaswinidas:
@yih-wang can you have a look at the screenshot in the PR description?

packages/model-serving/modelRegistry/DeploymentsTab.tsx
emilys314, line 1: Seems like a lot of copy pasting from the `GlobalDeploymentsTable.tsx`

packages/model-serving/modelRegistry/ModelDeploymentsTab.tsx
emilys314, line 1: I'm a little confused with the naming of the 2 tabs, they seem pretty similar

packages/model-serving/modelRegistry/DeploymentsTab.tsx
manaswinidas: Yes, we are essentially filtering the deployments by a certain label - the registered model ID. And adding one more "Version name" column in addition to the ones we have.

yih-wang:
Two small comments:
- Could we make the alert span the full width, similar to the version deployments page? (sorry for not catching this in the mockup :(
- The version name should be an inline link that goes to the deployments tab of that version. Didn’t see it in the screenshot, but also not sure if that’s meant to be in scope for this issue.

Otherwise looks good, thanks @manaswinidas!

frontend/packages/model-registry/upstream/frontend/src/odh/DetailsTabExtensionRoutes.tsx
```diff
@@ -0,0 +1,24 @@
...
+}) => {
+  return tabExtensions.map((extension) => (
+    <Route
+      key={extension.properties.id}
+      path={extension.properties.id}
+      element={
+        <ModelVersions
+          tab={extension.properties.id}
+          empty={false}
+        />
```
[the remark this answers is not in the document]
manaswinidas: So the empty={false} prop is actually necessary. The ModelVersionsProps extends ApplicationsPage props, and empty is a required prop from that component that wasn't included in the Omit list.
The component is correctly passing empty={false} as required by the ApplicationsPage interface. The import path fix I made was correct, and the empty prop should remain as it's required by the component's type definition.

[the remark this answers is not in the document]
manaswinidas: resolve this comment

packages/model-serving/modelRegistry/DeploymentsTab.tsx
manaswinidas: now both of the files reuse `GlobalDeploymentsTable`

packages/model-serving/src/components/global/GlobalDeploymentsTable.tsx
```diff
@@ -99,7 +192,52 @@ const GlobalDeploymentsTable: React.FC<{ deployments: Deployment[]; loaded: bool
...
+          }}
+          emptyTableView={
+            <DashboardEmptyTableView
+              onClearFilters={() => {
+                setFilterData(initialModelServingFilterData);
+                setPage(1);
+              }}
+            />
+          }
+        />
```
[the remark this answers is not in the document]
Griffin-Sullivan: I wonder if we should have a `ModelRegistryDeploymentsTable` where we can do MR specific things that don't affect the Global and Projects ones we have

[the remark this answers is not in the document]
mturley: I haven't had a chance to dig in here too much yet, but on the topic of reusing components for this table, I do want to make sure we're paying attention to the existing deployments table on the version details page (separate from this model details page). If we factor out a `ModelRegistryDeploymentsTable` it should maybe work for both places and just behave differently depending on whether it is passed a specific version id (show deployments for that version, which was introduced in https://github.com/opendatahub-io/odh-dashboard/pull/4598 and refactored in https://github.com/opendatahub-io/odh-dashboard/pull/4727) or not (show deployments for all versions of a model, this PR). That's rendered in `frontend/packages/modelServing/modelRegistry/DeploymentsTab.tsx` which currently just directly renders GlobalDeploymentsTable

packages/model-registry/upstream/frontend/src/app/pages/modelRegistry/screens/ModelVersions/ModelVersionsHeaderActions.tsx
```diff
@@ -35,28 +40,79 @@ const ModelVersionsHeaderActions: React.FC<ModelVersionsHeaderActionsProps> = ({
...
+                >
+                  Actions
+                </MenuToggle>
+              )}
+            >
+              <DropdownList>
+                <ArchiveButtonDropdownItem setIsArchiveModalOpen={setIsArchiveModalOpen} />
+              </DropdownList>
+            </Dropdown>
+          )}
```
mturley, line 115: This seems like a large diff. It looks like the `<Dropdown>` props are duplicated exactly in both cases here and only its children are different depending on whether there is a version and the modal is available. Can we not just render the whole `<DeployModalExtension>` as a child of the DropdownList right before the ArchiveButtonDropdownItem and have it only render DropdownItems? or does that cause some loading problem?

packages/model-registry/upstream/frontend/src/app/pages/modelRegistry/screens/ModelVersions/ModelVersionsTabs.tsx
```diff
@@ -27,6 +30,64 @@ const ModelVersionsTabs: React.FC<ModelVersionsTabProps> = ({
...
 }) => {
   const navigate = useNavigate();
+  const tabExtensions = useExtensions(isModelRegistryDetailsTabExtension);
+  const { registeredModelId: rmId } = useParams();
+  const { preferredModelRegistry } = React.useContext(ModelRegistrySelectorContext);
+  
+  const modelDetailsTabs = [
+    <Tab
+      key={ModelVersionsTab.OVERVIEW}
+      eventKey={ModelVersionsTab.OVERVIEW}
```
mturley, line 40: This is another large diff, but based on @rsun19 's previous PRs around tab extensions I guess there's not much we can do about it for now. It would be nice to eventually make this same refactor in kubeflow (just lift the existing tabs out into an array) so that this midstream diff is just adding an array element.

packages/model-registry/upstream/frontend/src/app/pages/modelRegistry/screens/RegisteredModels/RegisteredModelTableRow.scss
```diff
@@ -0,0 +1,20 @@
...
+
+/* Alternative: Target by content if needed */
+.pf-v6-c-dropdown__menu-item[aria-disabled="true"]:has-text("View model information"),
+.pf-v6-c-dropdown__menu-item[aria-disabled="true"]:has-text("Latest version actions") {
+  font-size: var(--pf-t--global--font--size--body--sm) !important;
+  color: var(--pf-t--global--text--color--subtle) !important;
+  font-weight: var(--pf-t--global--font--weight--body--bold) !important;
+  text-transform: uppercase;
+  letter-spacing: 0.5px;
+}
```
mturley, line 20: Is it possible for us to use PF utility classNames to add these styles instead of CSS? I wonder if this will still look right in both Material and PF themes. We should try to avoid custom CSS if possible.

packages/model-registry/upstream/frontend/src/app/pages/modelRegistry/screens/RegisteredModels/RegisteredModelTableRow.tsx
```diff
@@ -61,7 +67,22 @@ const RegisteredModelTableRow: React.FC<RegisteredModelTableRowProps> = ({
...
+        navigate(`${rmUrl}/deployments`);
+      },
+    },
+
+    { isSeparator: true },
+
+    {
+      title: 'Latest version actions',
+      isDisabled: true,
+    },
```
mturley, line 82: Hmm, a disabled action as a group header isn't ideal since PF now supports [DropdownGroups](https://www.patternfly.org/components/menus/dropdown#with-groups-of-items), but I guess we're stuck with these IAction objects for now and refactoring that is out of scope here. Something else to come back to some day.

mturley: Hmm - it looks like the archive and restore actions that also exist upstream should still be labeled "Latest version actions" now, right? so maybe this separator/title should be added upstream instead? Not a blocker, but something we might want to port over at some point.

packages/model-serving/src/components/global/GlobalDeploymentsTable.tsx
```diff
@@ -49,10 +51,89 @@ const projectColumn = (projects: ProjectKind[]): DeploymentsTableColumn => ({
...
+  if (registeredModelId && versionId) {
+    const registryPath = preferredModelRegistryName || '';
+    return `/model-registry/${registryPath}/registeredModels/${registeredModelId}/versions/${versionId}`;
+  }
+
+  return null;
+};
+
+const createVersionColumn = (
+  preferredModelRegistryName?: string,
```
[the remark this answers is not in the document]
mturley: +1, we can probably achieve this with minimal changes to this file. I don't know if we can use the existing `tableExtensions` behavior, since it has some behavior we may not want (I'm looking at the code below commented `Only add platform columns if all platforms have the same columns`), but we can do something similar to pass in additional custom columns

packages/model-serving/src/components/global/GlobalDeploymentsTable.tsx
```diff
@@ -81,6 +162,14 @@ const GlobalDeploymentsTable: React.FC<{ deployments: Deployment[]; loaded: bool
...
   );
 
+  const paginatedDeployments = React.useMemo(() => {
+    if (showAlert) {
+      const startIndex = (page - 1) * perPage;
+      return filteredDeployments.slice(startIndex, startIndex + perPage);
+    }
+    return filteredDeployments;
+  }, [filteredDeployments, page, perPage, showAlert]);
+
```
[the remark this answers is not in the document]
mturley: +1

mturley (changes requested):
+1 to other comments, and a couple other small things.

packages/model-registry/upstream/frontend/src/app/pages/modelRegistry/screens/RegisteredModels/RegisteredModelTableRow.tsx
manaswinidas: Deploy and LAB-tune are the "latest version" actions and these only exist midstream 😅

PR #4785, opened 2025-09-01
Quarantine RHOAIENG-33410

antowaddle:
## Description
- Quarantine testClusterStorageAccessModes due to https://issues.redhat.com/browse/RHOAIENG-33410


antowaddle:
> I think you can add the Maintain tag under `describe` (line 18) just once, and it will skip all its tests

@manosnoam I haven't tested this before, adding the Maintain to the 'It' is a documented/known way for the tests to skip.

manosnoam:
> ` // If options is an object with tags, store them on Cypress.testTags. if (typeof options === 'object' && options.tags) { Cypress.testTags[name] = options.tags; }
> 

Yep, I added it. As far as I remember, it should handle Suite tags too. 
Anyway, no reason not to merge, just FYI, we should allow skipping the whole suite, not just single tests.

PR #4763, opened 2025-08-26
Refactor resource handling in assembleNotebook function by simplifyin…

Gkrumbach07:
https://issues.redhat.com/browse/RHOAIENG-31937

## Description

Resource requests and and limits are no longer hardcoded to just be cpu and memory for jupyter spawner

## How Has This Been Tested?

create a jupyter notebook through the spawner (not workbench) with a hardware profile with a gpu resource set and confirm the resulting notebook is updated and has the correct  resource request and limit

## Test Impact

backend change so none

## Request review criteria:

Self checklist (all need to be checked):
- [x] The developer has manually tested the changes and verified that the changes work
- [x] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [x] The developer has added tests or explained why testing cannot be added (unit or cypress tests for related changes)
- [x] The code follows our [Best Practices](/docs/best-practices.md) (React coding standards, PatternFly usage, performance considerations)

If you have UI changes: 

- [ ] Included any necessary screenshots or gifs if it was a UI change.
- [ ] Included tags to the UX team if it was a UI/UX change.

After the PR is posted & before it merges:
- [ ] The developer has tested their solution on a cluster by using the image produced by the PR to `main`

## Summary by CodeRabbit

- **Refactor**
  - Streamlined how resource settings are applied to notebook containers to reduce duplication and improve consistency in backend processing. This change is internal and does not alter current behavior or user workflows. Existing resource requests and limits continue to be honored as before. Improves clarity and reduces potential for configuration mismatches. No action required from users, and no visible changes in the UI or performance are expected.

PR #4797, opened 2025-09-02
Prevent number of accelerators count reset in standalone workbench creation

nananosirova:
https://issues.redhat.com/browse/RHOAIENG-33476

## Description

The number of accelerators count resets to 0 when creating standalone workbenches, preventing GPU resources from being properly assigned to workbench pods. The `useAcceleratorProfileFormState` hook was overriding user selections with initial state values whenever the `initialState` changed, which happens during form processing. Added an `isInitialized` state to ensure the form initialization only happens once when the component first loads, preventing additional resets of user-selected values.
The following is also needed for this to work:  [PR #4763](https://github.com/opendatahub-io/odh-dashboard/pull/4763)

https://github.com/user-attachments/assets/95ba534c-0b5b-4c6a-bb72-c176747d6692

## How Has This Been Tested?

Manually, see above.

## Request review criteria:

If you have UI changes: 

## Summary by CodeRabbit

- Bug Fixes
  - Prevents the Accelerator Profile form from re-initializing after first load, avoiding unintended overwrites and visual flicker.
  - Ensures the form populates only once when data is available, improving stability across reloads and edits.
- Tests
  - Adds end-to-end coverage verifying accelerator profile selection, GPU count preservation, and GPU-related fields are included when starting a server.
- Notes
  - No changes to user-facing APIs; includes a testability attribute for the GPU count input.

### DIFF
--- a/frontend/src/utilities/useAcceleratorProfileFormState.ts
+++ b/frontend/src/utilities/useAcceleratorProfileFormState.ts
@@ -153,11 +153,13 @@ const useAcceleratorProfileFormState = (
     useExistingSettings: false,
   });
 
+  const initializedRef = React.useRef(false);
   React.useEffect(() => {
-    if (loaded) {
+    if (loaded && !initializedRef.current) {
       setFormData('profile', initialState.acceleratorProfile);
       setFormData('count', initialState.count);
       setFormData('useExistingSettings', initialState.unknownProfileDetected);
+      initializedRef.current = true;
     }
   }, [loaded, initialState, setFormData]);
 
--- a/frontend/src/__tests__/cypress/cypress/pages/notebookServer.ts
+++ b/frontend/src/__tests__/cypress/cypress/pages/notebookServer.ts
@@ -136,6 +136,10 @@ class NotebookServer {
   findNotebookVersion(version: string) {
     return cy.get(`[data-id="${version}"]`);
   }
+
+  findNumberOfAcceleratorsInput() {
+    return cy.findByTestId('number-of-accelerators').find('input');
+  }
 }
 
 export const notebookServer = new NotebookServer();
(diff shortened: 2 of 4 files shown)


frontend/src/utilities/useAcceleratorProfileFormState.ts
```diff
@@ -153,13 +153,15 @@ const useAcceleratorProfileFormState = (
...
   });
 
+  const [isInitialized, setIsInitialized] = React.useState(false);
   React.useEffect(() => {
-    if (loaded) {
+    if (loaded && !isInitialized) {
       setFormData('profile', initialState.acceleratorProfile);
       setFormData('count', initialState.count);
       setFormData('useExistingSettings', initialState.unknownProfileDetected);
+      setIsInitialized(true);
```
srtanish1992, line 162: Instead of of using `useState`, I would suggest we use `useRef` to avoid re-renders when we call `setIsInitialized` and we don't need to include the ref in the dependency array.

```
const initializedRef = React.useRef(false);

  React.useEffect(() => {
    if (loaded && !initializedRef.current) {
      setFormData('profile', initialState.acceleratorProfile);
      setFormData('count', initialState.count);
      setFormData('useExistingSettings', initialState.unknownProfileDetected);
      initializedRef.current = true;
    }
  }, [loaded, setFormData]);
```

srtanish1992:
@nananosirova I have left few comments. Let me know once you address them.
I also think we should add a cypress test to cover this bug.

frontend/src/utilities/useAcceleratorProfileFormState.ts
```diff
@@ -153,11 +153,13 @@ const useAcceleratorProfileFormState = (
     useExistingSettings: false,
   });
 
+  const initializedRef = React.useRef(false);
```
srtanish1992, line 156: nit: can you add space between const and hook?

Gkrumbach07 (approved):
this looks good

PR #4830, opened 2025-09-05
Add root level dependency installation and npm caching to gen-ai frontend workflow

rohitkrai03:
## Description
Tests fail in CI when trying to reuse utils or components from the internal package because CI is only installing npm packages at Gen AI level and not at root level.

- Added root level npm dependency installation to the workflow.
- Added npm caching.

## Summary by CodeRabbit

- Chores
  - Standardized Node.js version across build pipelines for consistent behavior.
  - Implemented dependency caching to speed up installs and reduce CI time.
  - Added root-level dependency installation to ensure all workspace dependencies are present before frontend tasks.
  - Retained existing frontend install/test/build flow while improving the surrounding pipeline.
  - Overall, improves build reliability, performance, and consistency across environments.
  - No user-facing functionality changes.

### DIFF
--- a/.github/workflows/gen-ai-frontend-build.yml
+++ b/.github/workflows/gen-ai-frontend-build.yml
@@ -15,6 +15,8 @@ on:
       - '!DOCKERFILE*'
       - '!**.gitignore'
       - '!**.md'
+env:
+  NODE_VERSION: 20
 jobs:
   test-and-build:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


christianvogt:
This seems fine. 
@antowaddle we should talk about the caching of of npm modules across workflows. The recommendation is to cache `~/.npm` and not `node_modules`. We should align all workflows on the same mechanism and cache key so all benefit from the same.

PR #4897, opened 2025-09-17
Add hardware profile binding state labels for notebook and model serving tables

nananosirova:
https://issues.redhat.com/browse/RHOAIENG-22846

## Description

Added labels for updated/deleted/disabled hardware profiles in the workbenches and KServe model deployments tables. 

- Added an 'Deleted' label when a hardware profile is deleted after being selected for a notebook/model deployment:
   - Shows as deleted when the Notebook/InferenceService contains the hardware profile name in the annotation but can't fetch the hardware profile with that name

- Added a 'Disabled' label when a hardware profile is disabled after being selected for a notebook/model deployment:
   - Shows as disabled when the Notebook/InferenceService contains the hardware profile name in the annotation and can fetch the hardware profile with that name, but it has the disabled annotation set to true. 
- Added an 'Updated' label when a hardware profile is updated after being selected for a notebook/model deployment
   - Shows as updated when the Notebook/InferenceService contains the hardware profile name and that hardware profile can be fetched, but the hardware profile snapshot attached to the resource through the snapshot annotation is different than the newly fetched hardware profile values. 
- Clicking the label opens a popover with more information
- Added a popover to the hardware profile name with more details. 

## How Has This Been Tested?

Tested manually. 

## Test Impact

Enable the hardware profiles feature. 
Create three workbenches/model deployments and make sure they started.
For each:
- Delete one of the hardware profiles bound to the workbench/model deployment
- Update one of them
- Disable one of them

Check the labels.

## Request review criteria:

If you have UI changes: 


frontend/src/pages/hardwareProfiles/useWatchHardwareProfile.ts
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
+import { useK8sWatchResource } from '@openshift/dynamic-plugin-sdk-utils';
+import { HardwareProfileKind } from '#~/k8sTypes';
+import { groupVersionKind, HardwareProfileModel } from '#~/api';
+import { CustomWatchK8sResult } from '#~/types';
+
+const useWatchHardwareProfile = (
```
Gkrumbach07, line 7: i would just use a regular GET and not watch here. was there a reason to use watch

Gkrumbach07: also u should then use useFetch

frontend/src/concepts/hardwareProfiles/HardwareProfileTableColumn.tsx
```diff
@@ -0,0 +1,90 @@
...
+  const isProjectScoped = useIsAreaAvailable(SupportedArea.DS_PROJECT_SCOPED).status;
+  const [bindingStateInfo, bindingStateLoaded, loadError] =
+    useHardwareProfileBindingState(resource);
+
+  const hardwareProfile = bindingStateInfo?.profile;
+
+  if (loadError && bindingStateInfo?.state !== HardwareProfileBindingState.DELETED) {
+    return (
+      <HelperText>
+        <HelperTextItem variant="error">Custom</HelperTextItem>
```
Gkrumbach07, line 39: This is not custom its just an error. I would put a error icon and then a popover to show the load error for the content of the popover. a popover can have alertSeverityVariant you can set to error

frontend/src/concepts/hardwareProfiles/HardwareProfileTableColumn.tsx
```diff
@@ -0,0 +1,90 @@
...
+  const displayName = hardwareProfile ? getHardwareProfileDisplayName(hardwareProfile) : 'Custom';
+  const resourceType = resourceTypeOf(resource);
+  return (
+    <>
+      <Flex
+        spaceItems={{ default: 'spaceItemsSm' }}
+        alignItems={{ default: 'alignItemsCenter' }}
+        data-testid="hardware-profile-table-column"
+      >
+        {bindingStateInfo?.state !== HardwareProfileBindingState.DELETED && (
```
Gkrumbach07, line 55: so if it is deleted we can still show the HardwareProfileDetailsPopover

it would really just use the container resources but this would give the user visibility that we have here
Screenshot 2025-09-30 at 4 25 34 PM


now u have on hover for it to show a general deleted message, but we could also have the resource exisiting settings as well. 

this was not in the mocks so i wont block the pr on this. but ill tag @vconzola and maybe we can circle back in a later release on this

Gkrumbach07 (changes requested):

frontend/src/pages/hardwareProfiles/useWatchHardwareProfile.ts
Gkrumbach07: wouldn't the labels just show up on page load? what is happening in the background here

frontend/src/concepts/hardwareProfiles/HardwareProfileTableColumn.tsx
nananosirova: Updated

Gkrumbach07 (approved):

PR #4906, opened 2025-09-17
Add models list view in assets page

divyanshiGupta:
https://issues.redhat.com/browse/RHOAIENG-33359
https://issues.redhat.com/browse/RHOAIENG-34076
https://issues.redhat.com/browse/RHOAIENG-34402

## Description

https://github.com/user-attachments/assets/976a7282-5f42-4b9b-a101-4fc2859b3e18

## Request review criteria:

If you have UI changes: 

## Summary by CodeRabbit

- New Features
  - Revamped Models tab: filterable/searchable table with model name, endpoints, use case, status, and per-model "Try in playground" action.
  - Namespace-aware loading and model list (includes new model entries and endpoints); model count shown on the Models tab.
  - Deep-linking: selecting "Try in playground" preselects the model in the Chatbot playground.
  - Improved empty/error states and deployment navigation.

- Tests
  - Updated tests to reflect namespace-specific model lists and new datasets.

### DIFF
--- a/packages/gen-ai/frontend/src/app/AIAssets/AIAssetsModelsTab.tsx
+++ b/packages/gen-ai/frontend/src/app/AIAssets/AIAssetsModelsTab.tsx
@@ -1,14 +1,46 @@
 import * as React from 'react';
 import { Bullseye, Content, ContentVariants, Spinner } from '@patternfly/react-core';
+import { Namespace } from 'mod-arch-core';
 import { useNavigate } from 'react-router-dom';
 import ModelsEmptyState from '~/app/EmptyStates/NoData';
-import useFetchLlamaModels from '~/app/hooks/useFetchLlamaModels';
-import { GenAiContext } from '~/app/context/GenAiContext';
+import { LlamaModel, AIModel } from '~/app/types';
+import AIModelsTable from './components/AIModelsTable';
 
-const AIAssetsModelsTab: React.FC = () => {
+type AIAssetsModelsTabProps = {
+  models: AIModel[];
+  playgroundModels: LlamaModel[];
+  namespace?: Namespace;
+  loaded: boolean;
+  error?: Error;
+};
+
+const AIAssetsModelsTab: React.FC<AIAssetsModelsTabProps> = ({
+  models,
+  playgroundModels,
+  namespace,
+  loaded,
+  error,
+}) => {
   const navigate = useNavigate();
-  const { namespace } = React.useContext(GenAiContext);
-  const { data: models, loaded } = useFetchLlamaModels(namespace?.name);
+  // Determine which AI Assets models are available in the playground
+  const modelsWithPlaygroundStatus = React.useMemo(() => {
+    const playgroundModelIds = new Set(playgroundModels.map((model: LlamaModel) => model.id));
+
+    return models.map((model: AIModel) => ({
+      ...model,
+      playgroundStatus: playgroundModelIds.has(model.model_name) ? 'available' : 'not-available',
+    }));
+  }, [models, playgroundModels]);
+
+  const handleTryInPlayground = (model: AIModel) => {
+    // Navigate to playground with the selected model
+    const playgroundUrl = `/gen-ai/playground/${namespace?.name}`;
+    navigate(playgroundUrl, {
+      state: {
+        model: model.model_name,
+      },
+    });
+  };
 
   if (!loaded) {
     return (
@@ -18,7 +50,7 @@ const AIAssetsModelsTab: React.FC = () => {
     );
   }
 
-  if (models.length === 0) {
+  if (error || modelsWithPlaygroundStatus.length === 0) {
     return (
       <ModelsEmptyState
         title="To begin you must deploy a model"
@@ -47,16 +79,14 @@ const AIAssetsModelsTab: React.FC = () => {
         }
         actionButtonText="Go to Deployments"
         handleActionButtonClick={() => {
-          navigate('/modelServing');
+          navigate(`/modelServing/${namespace?.name}`);
         }}
       />
     );
   }
+
   return (
-    <>
-      {/* TODO: Add list of models */}
-      List goes here
-    </>
+    <AIModelsTable models={modelsWithPlaygroundStatus} onTryInPlayground={handleTryInPlayground} />
   );
 };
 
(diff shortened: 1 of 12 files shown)


packages/gen-ai/frontend/src/app/services/llamaStackService.ts
```diff
@@ -156,3 +168,50 @@ export const getLSDstatus = (project: string): Promise<LlamaStackDistributionMod
...
+          null;
+
+        return {
+          id: `model-${index}`,
+          name: item.model_name,
+          description: item.description || 'No description available',
+          internalEndpoint,
+          externalEndpoint,
+          useCase: item.usecase || 'General purpose',
+          playgroundStatus: 'available', // All models are available for playground
```
DaoDaoNoCode, line 201: I thought this would be whether the model itself (the `InferenceService` CR) is available. If not, we cannot use it. You can check `Model deployment` page to see some similar fields.

packages/gen-ai/frontend/src/app/services/llamaStackService.ts
```diff
@@ -156,3 +168,50 @@ export const getLSDstatus = (project: string): Promise<LlamaStackDistributionMod
...
+          deploymentName: item.model_name,
+          // Fields from BFF API response
+          modelName: item.model_name,
```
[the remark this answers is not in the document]
DaoDaoNoCode: Though it's the fields from the BFF, we should either:
1. Handle it here and pass new props for the components to use
2. Leave it unhandled and handle it at a component level

Handling it and still passing the original data to the component could be confusing, you will see a lot of duplicated fields when you are using this `AIModel` type, and it could be a headache for the following developers who will work on it.

divyanshiGupta:
@varshaprasad96 The designs have a status column for the models but in the api response for /api/v1/aa/models?namespace= i dont see a status key. Is that something we are adding or all the models returned will have an active status by default?

packages/gen-ai/frontend/src/app/AIAssets/components/AIModelTableRow.tsx
```diff
@@ -0,0 +1,180 @@
...
+                Internal Endpoint URL
+              </div>
+            </div>
+          }
+          bodyContent={
+            <div style={{ padding: '0 16px 16px 16px' }}>
+              <ClipboardCopy
+                hoverTip="Copy"
+                clickTip="Copied"
+                style={{ fontFamily: 'monospace' }}
```
DaoDaoNoCode, line 108: Don't change the font here, we should use whatever PF provides

packages/gen-ai/frontend/src/app/AIAssets/AIAssetsModelsTab.tsx
```diff
@@ -2,23 +2,46 @@ import * as React from 'react';
...
-  if (!loaded) {
+    return models.map((model: AIModel) => ({
+      ...model,
+      playgroundStatus: playgroundModelIds.has(model.model_name) ? 'available' : 'not-available',
+    }));
+  }, [models, playgroundModels]);
+
+  const handleTryInPlayground = (model: AIModel) => {
+    // Navigate to playground with the selected model
+    const playgroundUrl = `/gen-ai/playground/${namespace?.name}?model=${encodeURIComponent(model.model_name)}`;
```
DaoDaoNoCode, line 32: Based on the discussion here https://redhat-internal.slack.com/archives/C099MEPGF43/p1757526668691629?thread_ts=1757523105.673619&cid=C099MEPGF43. The preferred way could be using `{state: {modelName: model.model_name}}` to pass the data between 2 pages because we don't want to bookmark the data a lot. We only want to use it for selection when navigates it.
In the Chatbot, it only needs to create `const location = useLocation()`, read it from `location.state` and select it if it exists.

pnaik1 (changes requested):

packages/gen-ai/frontend/src/app/AIAssets/components/AIModelTableRow.tsx
```diff
@@ -0,0 +1,180 @@
...
+          <Button variant={ButtonVariant.link}>View</Button>
+        </Popover>
+      ) : (
+        <Button
+          variant={ButtonVariant.link}
+          icon={<PlusCircleIcon color="var(--pf-t--global--palette--blue-600)" />}
+          isDisabled // TODO: Enable when external endpoint creation api is implemented
+        >
+          Create
+        </Button>
```
DaoDaoNoCode, line 155: From the discussion https://redhat-internal.slack.com/archives/C099MEPGF43/p1758209924990989?thread_ts=1758204451.054829&cid=C099MEPGF43, this will not be supported for 3.0. We can show something else instead of showing a disabled create button.

pnaik1:
I see sorting isnt working, Can you please fix that

packages/gen-ai/frontend/src/app/AIAssets/types.ts
```diff
@@ -0,0 +1,30 @@
...
+}
+
+export interface AIModel extends AAModelResponse {
+  id: string;
+  playgroundStatus: string;
+  // Parse endpoints into usable format
+  internalEndpoint?: string;
+  externalEndpoint?: string;
+}
+
```
DaoDaoNoCode, line 18: Please put these 2 types into /app/types.ts for consistency

varshaprasad96:
> @varshaprasad96 The designs have a status column for the models but in the api response for /api/v1/aa/models?namespace i dont see a status key. Is that something we are adding or all the models returned will have an active status by default?

@divyanshiGupta I'll create a PR to address this issue. We can get the status of model from existing infsvc pods. Should be straightforward. Will modify the model list response.

pnaik1 (changes requested):

packages/gen-ai/frontend/src/app/AIAssets/components/AIModelsTable.tsx
```diff
@@ -0,0 +1,254 @@
...
+          variant={EmptyStateVariant.sm}
+        >
+          <EmptyStateBody>No models match your filter criteria.</EmptyStateBody>
+          <EmptyStateFooter>
+            <Button variant={ButtonVariant.link} onClick={onClearFilters}>
+              Clear filters
+            </Button>
+          </EmptyStateFooter>
+        </EmptyState>
+      }
```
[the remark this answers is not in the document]
divyanshiGupta: It doesnt seem to have a prop for custom empty state body text that is why I couldnt use this comp.

pnaik1:
@divyanshiGupta nice work, the UI works well
looking at the design i see its also displaying number of models in the model tab

https://github.com/user-attachments/assets/c70e15ac-19c9-4c3d-a6b0-8406e11afd1b

PR #4962, opened 2025-09-23
[WIP] Migrate to gateway api based manifests

jctanner:
https://issues.redhat.com/browse/RHOAIENG-32517

## Description
This changes ingress for the dashboard from an OCP route to a gateway based HTTProute and other related changes.

## Summary by CodeRabbit

- New Features
  - Support BYOIDC header-based authentication, including group-based admin detection.
  - Added Gateway API HTTPRoute for the dashboard.

- Changes
  - Switched to kube-rbac-proxy with updated health checks for improved reliability.
  - Enforced HTTPS to the backend service.
  - Extended permissions to allow token and access reviews for smoother auth flows.
  - Deployment scaled to a single replica.

- Chores
  - Introduced configuration for kube-rbac-proxy and updated kustomization to use new resources and images.

### DIFF
--- a/backend/src/routes/api/status/adminAllowedUsers.ts
+++ b/backend/src/routes/api/status/adminAllowedUsers.ts
@@ -69,7 +69,7 @@ export const getAllowedUsers = async (
   const { namespace } = request.params;
   const userInfo = await getUserInfo(fastify, request);
   const currentUser = userInfo.userName;
-  const isAdmin = await isUserAdmin(fastify, currentUser, namespace);
+  const isAdmin = await isUserAdmin(fastify, currentUser, namespace, request);
   if (!isAdmin) {
     // Privileged call -- return nothing
     fastify.log.warn(
--- a/backend/src/routes/api/status/statusUtils.ts
+++ b/backend/src/routes/api/status/statusUtils.ts
@@ -20,7 +20,7 @@ export const status = async (
   const { server } = currentCluster;
 
   const { userName, userID } = await getUserInfo(fastify, request);
-  const isAdmin = await isUserAdmin(fastify, userName, namespace);
+  const isAdmin = await isUserAdmin(fastify, userName, namespace, request);
   const isAllowed = isAdmin ? true : await isUserAllowed(fastify, userName);
 
   if (!kubeContext && !kubeContext.trim()) {
--- a/backend/src/utils/adminUtils.ts
+++ b/backend/src/utils/adminUtils.ts
@@ -8,10 +8,37 @@ import { getAllGroupsByUser, getGroup } from './groupsUtils';
 import { flatten, uniq } from 'lodash';
 import { getNamespaces } from '../utils/notebookUtils';
 import { getAuth } from './resourceUtils';
+import { BYOIDC_GROUPS_HEADER } from './constants';
 
 const SYSTEM_AUTHENTICATED = 'system:authenticated';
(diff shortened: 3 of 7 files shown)


manifests/core-bases/base/httproute.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: gateway.networking.k8s.io/v1
+kind: HTTPRoute
+metadata:
+  name: odh-dashboard
+spec:
+  parentRefs:
+    - name: odh-gateway
```
jstourac, line 7: Looks like this is wrong as the default name odh operator uses is: https://github.com/opendatahub-io/opendatahub-operator/blob/4e2ab4b21afbb5b2d8c7df46b038456c36fbb690/internal/controller/services/gateway/gateway_support.go#L24 . Or am I missing something?

Also, I wonder whether this isn't something that should be configurable or at least - shouldn't we rather read this value from some odh/rhoai operator config (e.g. DSCI or DSC) or somewhere instead having this hardcoded? Or shall we rather search for it? Or are we truly okay to have this hardcoded?

jstourac:
```suggestion
    - name: data-science-gateway
```

backend/src/routes/api/status/adminAllowedUsers.ts
andrewballantyne, line 1: Not needed, covered by SSAR checks

backend/src/routes/api/status/statusUtils.ts
andrewballantyne, line 1: Not needed, covered by SSAR checks

backend/src/utils/adminUtils.ts
andrewballantyne, line 1: Not needed, covered by SSAR checks

andrewballantyne (commented):
All the comments that say
> Not needed, covered by SSAR checks
Are covered via the SSAR checks done in https://github.com/opendatahub-io/odh-dashboard/pull/5028

I do believe we can just revert these / not include them if @lucferbux takes over the PR.

manifests/core-bases/base/httproute.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: gateway.networking.k8s.io/v1
+kind: HTTPRoute
+metadata:
+  name: odh-dashboard
+spec:
+  parentRefs:
+    - name: data-science-gateway
```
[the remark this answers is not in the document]
andrewballantyne: [Another example](https://github.com/opendatahub-io/odh-dashboard/pull/4962#discussion_r2394151881)

backend/src/utils/userUtils.ts
```diff
@@ -89,12 +89,64 @@ export const getUser = async (
...
+    userName = userHeader;
+  }
+
+  // Use email as userID if available, otherwise fall back to extracted username
+  const userID = emailHeader || userName;
+
+  return {
+    userName,
+    userID,
+  };
```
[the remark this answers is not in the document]
lucferbux: Yes, we already are using [SelfSubjectReview](https://kubernetes.io/docs/reference/kubernetes-api/authorization-resources/self-subject-rules-review-v1/) (which is the default when you use `kubectl auth whoami`) [in BFF already](https://github.com/opendatahub-io/odh-dashboard/blob/main/packages/model-registry/upstream/bff/internal/integrations/kubernetes/token_k8s_client.go#L220) it might be a viable solution.

PR #4973, opened 2025-09-24
Disable Model Mesh v3.0

andrewballantyne:
https://issues.redhat.com/browse/RHOAIENG-32586

## Description

Disables Model Mesh feature flag at the server -- so all flows lock down and cannot be enabled.

No way to get the Model Mesh card to show up in projects & without NIM installed, it also selects KServe on new project creation.

Settings page was also updated to remove Model Mesh configuration flag:
image

## How Has This Been Tested?

I installed Model Mesh & KServe

My DSC Status for Installed Components
```
  installedComponents:
    codeflare: false
    kserve: true
    feastoperator: false
    trustyai: true
    ray: false
    kueue: false
    data-science-pipelines-operator: true
    workbenches: true
    model-registry-operator: false
    model-mesh: true
    dashboard: true
    llamastackoperator: false
    trainingoperator: false
```

## Test Impact

N/A -- feature flag disabled -- might need to remove tests -- looking

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/backend/src/utils/resourceUtils.ts
+++ b/backend/src/utils/resourceUtils.ts
@@ -517,6 +517,29 @@ export const initializeWatchedResources = (fastify: KubeFastifyInstance): void =
   consoleLinksWatcher = new ResourceWatcher<ConsoleLinkKind>(fastify, fetchConsoleLinks);
 };
 
+/**
+ * Sometimes we need to lockout a feature while we look to remove it more properly. This function
+ * can help align the feature to being disabled.
+ */
+const applyFeatureLockouts = (config: DashboardConfig): DashboardConfig => ({
+  ...config,
+  spec: {
+    ...config.spec,
+    dashboardConfig: {
+      ...config.spec.dashboardConfig,
+      // Apply Feature Lockouts Directly below
+      // Feature flags noted below are removable from the CRD at the earliest convenience
+      // Do note, update the CRD is a backwards incompatible step and needs an up-version
+      //---------------------------------------
+
+      /**
+       * Model Mesh is removed in v3.0
+       */
+      disableModelMesh: true,
+    },
+  },
+});
+
 const FEATURE_FLAGS_HEADER = 'x-odh-feature-flags';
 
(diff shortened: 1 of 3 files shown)


frontend/src/__tests__/cypress/cypress/tests/e2e/settings/servingRuntimes/testMultiModelServingRuntimeCreation.cy.ts
andrewballantyne, line 1: @antowaddle FYI -- not sure if we have tickets you want to update here in this space for any "quarantined" items

Griffin-Sullivan:
@vconzola ah ok then no we would keep the card I think cause we still want to use the check box for enabling and disabling serving from the UI. The dropdown for serverless vs raw as the default will be removed though.

frontend/src/__tests__/cypress/cypress/tests/e2e/settings/servingRuntimes/testMultiModelServingRuntimeCreation.cy.ts
antowaddle: https://issues.redhat.com/browse/RHOAIENG-32764 is against this test, but that's also failing other tests which I don't believe we are removing. 

Image

emilys314:
I tested the PR image on my cluster with model mesh enabled (serverless and service mesh but not authorino)

It all looks good
- Creating a new project will not have the model mesh option platform
- An existing model mesh project will will be 'converting' into a single serving one
- The admin cluster settings are not showing model mesh

/lgtm
(no approve yet because I'm slightly confused with the other convo but i'm fine with merging whenever)

vconzola:
> For "card" components themselves, are you asking if the llm-d setting will be underneath the "Model serving platforms" one? Or a new one underneath it for llmd? I was assuming the llmd setting is inside of this existing one
I think it's OK if the llm-d setting is in the same card as the single-model serving platform setting. We'll probably want to redo the content, however.

PR #5072, opened 2025-10-02
Deprecate notebookSizes, modelServerSizes, notebookTolerationSettings, AP and HWP feature flags and add validations

nananosirova:
https://issues.redhat.com/browse/RHOAIENG-35025

## Description

Since hardware profiles are going GA, we need to deprecate `notebookSizes` and `modelServerSizes` container size fields and the `notebookTolerationSettings` field from `OdhDashboardConfig` and add validations. 

## How Has This Been Tested?

In the console:
- Ensure that the `odh-dashboard-config` CR has the fields `notebookSizes`, `modelServerSizes` and `notebookController.notebookTolerationSettings` set. If not, [here's an example you can create.](https://gist.github.com/nananosirova/10c48f480cdc85835a063636c7a655f8) 
- Set `devFlags` in the DSC to:
```
devFlags:
  manifests:
    - contextDir: manifests
      sourcePath: ''
      uri: 'https://github.com/nananosirova/odh-dashboard/tarball/RHOAIENG-35025'
```
- Verify that the `OdhDashboardConfig` CRD was been updated
- Try to modify `notebookSizes`. Modifying it without setting it to `[]` or removing it completely yields this error:
image
- Try to modify `modelServerSizes`. Modifying it without setting it to `[]` or removing it completely yields this error:
image
- Try to modify `notebookTolerationSettings`. Modifying it without setting it to `{}` or removing it completely yields this error:
image
- Changing any other field yields this warning:
image

In a cluster with RHOAI installed:
- Set the devFlags in the DSC as shown above
- Delete the pre-existing `odh-dashboard-config` CR if present. A new one should get created. 
- Verify that the fields `notebookSizes`, `modelServerSizes` and `notebookTolerationSettings` are missing in the newly created `odh-dashboard-config`

## Test Impact

Tested with above. 

## Request review criteria:

If you have UI changes: 

## Summary by CodeRabbit

* **Chores**
  * Marked legacy notebook and model-server size settings, and accelerator/hardware-disable flags, as deprecated; validations now prevent changes or enforce removal.  
  * Removed predefined notebook/model-server size blocks from shared configuration; toleration settings deprecated with validations to preserve unchanged behavior only.

* **Documentation**
  * Added deprecation notices and guidance to use Hardware Profiles instead of notebookSizes, modelServerSizes, accelerator/hardware-disable flags, and notebook toleration settings.

### DIFF
--- a/manifests/common/crd/odhdashboardconfigs.opendatahub.io.crd.yaml
+++ b/manifests/common/crd/odhdashboardconfigs.opendatahub.io.crd.yaml
@@ -73,8 +73,10 @@ spec:
                       type: boolean
                     disableAcceleratorProfiles:
                       type: boolean
+                      description: 'DEPRECATED: Accelerator Profiles are replaced by Hardware Profiles. This field will be removed in a future version.'
                     disableHardwareProfiles:
                       type: boolean
+                      description: "DEPRECATED: Hardware Profiles are GA and cannot be disabled via this field. This field will be removed in a future version."
                     disableDistributedWorkloads:
                       type: boolean
                     disableModelCatalog:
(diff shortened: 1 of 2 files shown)


manifests/common/crd/odhdashboardconfigs.opendatahub.io.crd.yaml
```diff
@@ -140,8 +150,12 @@ spec:
                                 type: string
                               memory:
                                 type: string
-                # Hardware Profiles going GA will help remove these
+                  x-kubernetes-validations:
+                    - rule: self == oldSelf
```
Gkrumbach07, line 154: we should do this for all these 
```
self == oldSelf || self == null
```

this will allow it to be removed also

nananosirova: It allows it to be removed this way, seems like validation runs only when the field is present, and removal is allowed cause those fields aren't set as required.
Also adding `self == null` throws an error

PR #5083, opened 2025-10-03
Remove OCP Route logic & rely on gateway sharing with Workbenches

andrewballantyne:
https://issues.redhat.com/browse/RHOAIENG-34819

## Description

Drops OCP Routing for 3.0. Leans on the gateway for shared routes between Workbenches & Dashboard.

Pretty much simplifies it to the new utility addition:
```
/**
 * In v3.0, the accessing of a Workbench will be assuming a shared gateway route with Dashboard.
 * Leveraging browser feature of "same-origin" on the use of slash (/) links.
 */
export const getRoutePathForWorkbench = (
  workbenchNamespace: string,
  workbenchName: string,
): string => `/notebook/${workbenchNamespace}/${workbenchName}`;
```

Both Jupyter Tile & Project Workbenches can leverage this functionality and backend doesn't need the route api anymore.

## How Has This Been Tested?

Using clusters provided that have the gateway -- but wasn't fully tested as the Notebook Controller was not updated on this cluster to handle that.

## Test Impact

Cleaned up some tests around the fetching of notebook OCP routes.

## Request review criteria:

If you have UI changes: 


frontend/src/pages/notebookController/NotebookLogoutRedirect.tsx
```diff
@@ -54,25 +39,13 @@ const NotebookLogoutRedirect: React.FC = () => {
...
     if (namespace && notebookName && namespace !== workbenchNamespace) {
-      if (error) {
-        notification.error(`Error when logging out ${notebookName}`, error.message);
-        navigate(`/projects/${namespace}`);
-      }
-      if (loaded && notebookRoute) {
+      if (notebookRoute) {
         const location = new URL(notebookRoute);
+        // TODO: how do we handle logout? We share a token login state
         window.location.href = `${location.origin}/oauth/sign_out`;
```
andrewballantyne, line 45: @lphiri Do we have a path forward for what we are doing with logout? Hopefully the Notebook just doesn't have the action -- but if it does, we will invoke an oauth signout path that is just not going to work and we will get a 404 page likely.

PR #5087, opened 2025-10-06
Show model id in model info popover

divyanshiGupta:
https://issues.redhat.com/browse/RHOAIENG-35536

## Description

This PR updates model info popover to show model id instead of model name

Screenshot 2025-10-06 at 6 41 12 PM

## Request review criteria:

If you have UI changes: 

## Summary by CodeRabbit

- New Features
  - Model rows now include an info popover with model details and a one-click copy for the Model ID.

- Bug Fixes
  - API token section only appears for external models when a token is actually available, preventing empty token blocks.

- Style
  - Updated the Model column label from “Model deployment name” to “Model”.
  - Refined model title rendering in the table for a clearer, more consistent presentation.

### DIFF
--- a/packages/gen-ai/frontend/src/app/AIAssets/components/AIModelTableRow.tsx
+++ b/packages/gen-ai/frontend/src/app/AIAssets/components/AIModelTableRow.tsx
@@ -2,14 +2,14 @@ import * as React from 'react';
 import { Button, Truncate, Label, ButtonVariant } from '@patternfly/react-core';
 import { Td, Tr } from '@patternfly/react-table';
 import { CheckCircleIcon, ExclamationCircleIcon, PlusCircleIcon } from '@patternfly/react-icons';
-import { ResourceNameTooltip, TableRowTitleDescription, TruncatedText } from 'mod-arch-shared';
 import { useNavigate } from 'react-router-dom';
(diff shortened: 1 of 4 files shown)


rohitkrai03 (approved):
/lgtm

divyanshiGupta:
> I got a question, so we are not going to show the users what k8s kind of the model is such as `InferenceService` or `LLMInferenceService`? Then it would be hard for the users to locate the model on the cluster.

@varshaprasad96 Danielle wdyt about this?

Daniel-Warner-X:
> > I got a question, so we are not going to show the users what k8s kind of the model is such as `InferenceService` or `LLMInferenceService`? Then it would be hard for the users to locate the model on the cluster.
> 
> @varshaprasad96 Danielle wdyt about this?

The AI available endpoints view is for engineers to access models. For that they need the model id, endpoint, and access token. @DaoDaoNoCode what would be the reason they would also need to know what kind of k8s service it is?

DaoDaoNoCode:
@Daniel-Warner-X Sometimes they will need to find the corresponding k8s resource on the cluster to see the spec (e.g. if they want to go deeper to see the content of a model), they will need to know the type of the resource so they can locate it on the cluster. That's my 2 cents, but I have no problem with this solution if you think that information is not that important.

PR #5088, opened 2025-10-06
Fix gen-ai routes

divyanshiGupta:
https://issues.redhat.com/browse/RHOAIENG-35576

## Description

In latest main - playground and assets pages were not loading for me. Upon digging deeper I saw a PR https://github.com/opendatahub-io/odh-dashboard/pull/4983 for refactoring nav routes. It changed frontend route from /gen-ai to /gen-ai-studio but didn't update the app mount point and specific route definitions for these pages and hence the pages were not loading.

## Request review criteria:

If you have UI changes: 

## Summary by CodeRabbit

* **Refactor**
  * Standardized navigation to use centralized route definitions and switched routing to a dynamic Gen AI root path.
  * Updated the “Try in Playground” action in the AI Assets toolbar to use the revised playground path.

* **Chores**
  * Removed unused internal URL constants to streamline configuration.

Impact: More consistent URLs and more maintainable routing with no visible behavior changes.

### DIFF
--- a/packages/gen-ai/frontend/src/app/utilities/const.ts
+++ b/packages/gen-ai/frontend/src/app/utilities/const.ts
@@ -1,18 +1,10 @@
 import { asEnumMember, DeploymentMode } from 'mod-arch-core';
 
 const URL_PREFIX = process.env.URL_PREFIX || '/gen-ai';
-const PLAYGROUND_URL_PREFIX = `${URL_PREFIX}/playground`;
-const ASSETS_URL_PREFIX = `${URL_PREFIX}/assets`;
 
 const DEPLOYMENT_MODE =
   asEnumMember(process.env.DEPLOYMENT_MODE, DeploymentMode) || DeploymentMode.Federated;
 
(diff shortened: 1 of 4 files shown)


packages/gen-ai/frontend/src/app/utilities/const.ts
```diff
@@ -1,18 +1,10 @@
 import { asEnumMember, DeploymentMode } from 'mod-arch-core';
 
 const URL_PREFIX = process.env.URL_PREFIX || '/gen-ai';
```
rohitkrai03, line 3: Shouldn't we update this as well?

rohitkrai03 (changes requested):

packages/gen-ai/frontend/src/app/utilities/const.ts
divyanshiGupta: No, as this is used for apis and frontend url prefix is different from api url prefix

PR #5109, opened 2025-10-08
add wiring for llm-d install

varshaprasad96:
## Description
This PR modifies the install logic to also trigger install for LLMInferenceServices in addition to KServe based Raw deployments.
This needs to be tested after the latest version of LLS is downstream, which has logic to send in the auth tokens with the header. 

## Request review criteria:

If you have UI changes: 


packages/gen-ai/bff/internal/integrations/kubernetes/token_k8s_client.go
```diff
@@ -1023,11 +1022,50 @@ server:
...
+			if displayName, exists := targetLLMSVC.Annotations[DisplayNameAnnotation]; exists {
+				metadata["display_name"] = displayName
+			}
+			if description, exists := targetLLMSVC.Annotations[InferenceServiceDescriptionAnnotation]; exists {
+				metadata["description"] = description
+			}
+		}
+
+		// All models are LLM models using vllm-inference
+		providerID := "vllm-inference"
```
akram, line 1056: we need a constant here; and the value must match with the vllm provider's provider id created in the configmap

packages/gen-ai/bff/internal/integrations/kubernetes/token_k8s_client.go
```diff
@@ -1023,11 +1022,50 @@ server:
...
+		modelType := "llm"
+
+		// Use the actual model name from LLMInferenceService spec instead of service name
```
akram, line 1059: we need a `model_provider_id` field

packages/gen-ai/bff/internal/integrations/kubernetes/token_k8s_client.go
```diff
@@ -1068,6 +1106,7 @@ func (kc *TokenKubernetesClient) getModelDetailsFromServingRuntime(ctx context.C
 	providerID := "vllm-inference"
```
akram, line 1106: the constant also here

packages/gen-ai/bff/internal/integrations/kubernetes/token_k8s_client.go
```diff
@@ -1098,6 +1137,54 @@ func (kc *TokenKubernetesClient) findInferenceServiceByModelName(ctx context.Con
...
+		if llmSvc.Name == modelName {
+			kc.Logger.Info("found LLMInferenceService by model name", "modelName", modelName, "llmSvcName", llmSvc.Name, "namespace", namespace)
+			return &llmSvc, nil
+		}
+	}
+
+	return nil, fmt.Errorf("LLMInferenceService with model name '%s' not found in namespace %s", modelName, namespace)
+}
+
+// extractEndpointFromLLMInferenceService extracts the endpoint URL from LLMInferenceService status.addresses
```
akram, line 1161: that is for llm-d right? maybe we should have a more generic way using deployment status ; so we can treat vllm and llm-d the same way.

Checking the endpoint address may be a bit confusing and low level.

packages/gen-ai/bff/internal/integrations/kubernetes/token_k8s_client.go
```diff
@@ -1098,6 +1137,54 @@ func (kc *TokenKubernetesClient) findInferenceServiceByModelName(ctx context.Con
...
+		return "", fmt.Errorf("LLMInferenceService '%s' first address has no URL", llmSvc.Name)
+	}
+
+	endpointURL := firstAddress.URL.String()
+
+	// Add /v1 suffix if not present for LLMInferenceService compatibility
+	if !strings.HasSuffix(endpointURL, "/v1") {
+		endpointURL = endpointURL + "/v1"
+	}
+
```
akram, line 1182: I have seen this in other locations, this is confusing, we need to find a better way; 
either we fix it here, or do a follow up PR and discuss it

akram (changes requested):

packages/gen-ai/bff/internal/integrations/kubernetes/token_k8s_client.go
varshaprasad96: `/v1` is a OpenAI schema which is to be added to be able to reach the URL that is not present in KServe by default

packages/gen-ai/bff/internal/integrations/kubernetes/token_k8s_client.go
varshaprasad96: Yes - this is for llm-d. As of now, we have only use InferenceService and LLMInferenceService as the source of truth to get the URL.

packages/gen-ai/bff/internal/integrations/kubernetes/token_k8s_client.go
varshaprasad96: `provider_model_id` is required only when the internal (actual) model_id is different than the provided `model_id`. If it's null, it uses the `model_id`: https://github.com/llamastack/llama-stack/blob/5d711d4bcb1e7e84ca5325f88f3943b1fb274d79/llama_stack/core/routing_tables/models.py#L101

In our case - we are using the exact model id extracting from Ifsvc or llminfsvc spec. Which is why we don't need to repeat `provider_model_id`.

PR #5119, opened 2025-10-08
enable backend to share code within the monorepo

christianvogt:
https://issues.redhat.com/browse/RHOAIENG-35612

## Description

The goal here is to support code sharing across backend and frontend so we can stop duplicating configurations and type definitions. It is also a step to being able to switch more commands over to turbo to benefit from it's speed.

- Introduces [tsdown](https://tsdown.dev/) to build `backend` and dependent library `app-config`
  - Faster than tsc and easy to configure
- Several tsconfig changes to better support faster builds and libraries
  - Bonus: found a lot of improper imports from packages across package boundaries which had to get fixed up
- Unfortunately because of CPaaS we still are not using turborepo to build the application. Therefore added a work around to the root package.json to also build `app-config` for the `build` command
- Updated the backend to depend on `@odh-dashboard/app-config` (`packages/app-config`) as a test for sharing code in the monorepo with the backend
- Introduced building to the `app-config` package (`turbo run build`)

Note: Dockerfiles for CPaaS and konflux have been updated to remove omitting optional dependencies already.

## How Has This Been Tested?

- Tested builds (`npm run build`) from the root and from the `backend` and `packages/app-config` dirs
- Tested local dev environment
- Tested backend `debug` command in vscode to ensure we can still hit breakpoints
- TODO test build image and deploy

## Test Impact

N/A

## Request review criteria:

If you have UI changes: 

## Summary by CodeRabbit

* **Refactor**
  * Centralized module federation configuration management through a shared app-config package.
  * Optimized TypeScript build system for faster compilation and improved maintainability.
  * Reorganized backend utilities for better code organization and reduced duplication.


lucferbux:
@christianvogt I've reviewed the PR and it seems ok, it would be nice if we can do a rundown today cause I don't have the full context, I've tested it and it works. There's a couple of conflicts so after rebasing and doing some other checks I would guess this is ready to go.

christianvogt:
I too need to now build and test the image as the final check.

PR #5317, opened 2025-10-30
Access Deployment Wizard W/O Project

katieperry4:
Closes [RHOAIENG-37347](https://issues.redhat.com/browse/RHOAIENG-37347)

## Description

Added a new route (which can be changed, I just removed the namespace from the existing route) so now it's:
`/ai-hub/deployments/deploy` vs `/ai-hub/deployments/{project-name}/deploy`
^probably need to change it bc it'll be weird if you for some reason had a project named deploy

If you go into the wizard without a project it won't load existing connections or pvcs so you'll have to pick a new connection:
Screenshot 2025-10-30 at 4 03 21 PM

Added a project selector to the 2nd step (only shows if you enter w/o a project) otherwise it just shows as text like before:
Screenshot 2025-10-30 at 4 05 17 PM
Screenshot 2025-10-30 at 4 05 22 PM

If you select a project and navigate back to step 1, the existing connections and pvcs will load in if the project has them which is fun:
Screenshot 2025-10-30 at 4 05 49 PM

Added validation so you can't get past step 2 without a project, so the next button stays disabled if you haven't picked one
Shows up in the summary per usual and the deployment works the same too
Screenshot 2025-10-30 at 4 06 00 PM

## How Has This Been Tested?

Tested locally

To test:
- navigate to `/ai-hub/deployments/deploy`
- fill in the model location and type as usual, notice the lack of existing connections
- in step 2 select a project, poke around make sure everything is as it should be, runtimes, replicas, etc
- go back to step 1 and see the existing connections popping in the select, and make sure AAA and MaaS checkbox show as usual, summary page is normal
- deploy
- edit should be unchanged since you always enter with a project from edit

## Test Impact

Added unit tests for the project selector and updated one of the cypress tests to enter without a project (I changed like 4 lines it seemed too minor to write an entire test just for that)

## Request review criteria:

If you have UI changes: 

## Summary by CodeRabbit

* **New Features**
  * Enter the deployment wizard without a pre-selected project and choose the target project inside the wizard.
  * New direct route to the deployment flow.

* **Bug Fixes**
  * Validation now requires a selected project before final deploy.
  * Deploy button can open the wizard even when no project is selected.

* **Tests**
  * Added and updated E2E and unit tests for project-selection and deployment flows.

### DIFF
--- a/frontend/src/__tests__/cypress/cypress/tests/mocked/modelServing/modelServingDeploy.cy.ts
+++ b/frontend/src/__tests__/cypress/cypress/tests/mocked/modelServing/modelServingDeploy.cy.ts
@@ -1572,9 +1572,9 @@ describe('Model Serving Deploy Wizard', () => {
     });
 
     it('deploy create', () => {
-      cy.visitWithLogin(`/modelServing/test-project/deploy`);
+      cy.visitWithLogin(`/modelServing/deploy`);
       cy.findByTestId('app-page-title').contains('Deploy a model');
-      cy.url().should('include', '/ai-hub/deployments/test-project/deploy');
+      cy.url().should('include', '/ai-hub/deployments/deploy');
     });
   });
 });
--- a/packages/model-serving/src/GlobalModelsRoutes.tsx
+++ b/packages/model-serving/src/GlobalModelsRoutes.tsx
@@ -4,7 +4,6 @@ import useModelMetricsEnabled from '@odh-dashboard/internal/pages/modelServing/u
 import ProjectsRoutes from '@odh-dashboard/internal/concepts/projects/ProjectsRoutes';
 import GlobalModelsPage from './components/global/GlobalModelsPage';
 import GlobalMetricsRoutes from './components/metrics/GlobalMetricsRoutes';
-import { ModelDeploymentWizardPage } from './components/deploymentWizard/ModelDeploymentWizardPage';
 
 const GlobalModelsRoutes: React.FC = () => {
   const [modelMetricsEnabled] = useModelMetricsEnabled();
@@ -13,7 +12,6 @@ const GlobalModelsRoutes: React.FC = () => {
     <ProjectsRoutes>
       <Route index element={<GlobalModelsPage />} />
       {modelMetricsEnabled && <Route path="metrics/*" element={<GlobalMetricsRoutes />} />}
-      <Route path="deploy" element={<ModelDeploymentWizardPage />} />
       <Route path="*" element={<Navigate to="." />} />
     </ProjectsRoutes>
   );
--- a/packages/model-serving/src/components/deploymentWizard/useDeploymentWizard.ts
+++ b/packages/model-serving/src/components/deploymentWizard/useDeploymentWizard.ts
@@ -115,8 +115,8 @@ export const useModelDeploymentWizard = (
 
   // loaded state
   const modelDeploymentLoaded = React.useMemo(() => {
-    return modelFormatState.loaded && hardwareProfileConfig.profilesLoaded;
-  }, [modelFormatState.loaded, hardwareProfileConfig.profilesLoaded]);
+    return hardwareProfileConfig.profilesLoaded;
+  }, [hardwareProfileConfig.profilesLoaded]);
 
   // Step 3: Advanced Options - Individual Fields
   const modelAvailabilityFormData = React.useMemo(


packages/model-serving/src/components/deploymentWizard/steps/ModelDeploymentStep.tsx
```diff
@@ -35,7 +39,7 @@ export const ModelDeploymentStepContent: React.FC<ModelDeploymentStepProps> = ({
           nameHelperTextAbove="Name this deployment. This name is also used for the inference service created when the model is deployed."
         />
         <ModelServingHardwareProfileSection
-          project={projectName}
+          project={projectName ?? undefined}
```
emilys314: nitpick, but why is the `?? undefined` needed? Isn't the type already `string | undefined`? Same with the other places

I prefer to avoid having an excess amount of them because it makes the code cluttered and harder to read

packages/model-serving/src/concepts/ModelDeploymentsContext.tsx
```diff
@@ -151,8 +151,8 @@ export const ModelDeploymentsProvider: React.FC<ModelDeploymentsProviderProps> =
 
   return (
     <ModelDeploymentsContext.Provider value={contextValue}>
-      {
-        // the only way to dynamically call hooks (useWatchDeployments) is to render them in dynamic components
+      {projects &&
+        projects.length > 0 &&
```
emilys314: You can just pass in `undefined` for `projects` here I think, since in the Watcher it can make it undefined anyways
```
  // If there's only 1 project, scope the call to that project, otherwise call without project scoping
  const projectToScope = projects.length === 1 ? projects[0] : undefined;
```

vconzola:
Hi @katieperry4  - I'm not sure I fully understand the scope of this PR. These are only changes for accessing the wizard without a Project correct? Any flow changes (skipping or hiding the first wizard step) will be addressed separately, correct?

katieperry4:
@vconzola Yeah this is just so you can even see the wizard if there's no project passed in, step stuff / prefilling with catalog and registry info comes in the following tickets 👍

packages/model-serving/src/concepts/ModelDeploymentsContext.tsx
```diff
@@ -151,29 +155,26 @@ export const ModelDeploymentsProvider: React.FC<ModelDeploymentsProviderProps> =
...
+        if (!deploymentWatchersLoaded) {
+          return null;
+        }
+
+        return (
+          <PlatformDeploymentWatcher
+            key={platformId}
+            platformId={platformId}
+            watcher={watcher}
+            projects={projects ?? []}
```
emilys314: The infinite rerender might be caused by the ` ?? []`, the `[]` is a new object reference on every rerender. 


```suggestion
            projects={projects}
```

packages/model-serving/src/components/deploymentWizard/fields/ProjectSection.tsx
```diff
@@ -1,33 +1,63 @@
...
-            className="pf-v5-u-ml-xs"
+export const isValidProject = (project: ProjectKind | null): boolean => {
+  return !!project?.metadata.name;
+};
+
+export const useProjectSection = (
+  initialProject?: ProjectKind | null,
+): {
+  initialProject?: ProjectKind | null;
+  project: ProjectKind | null;
```
emilys314: I was wondering about the decision to use `project: ProjectKind` instead of `projectName: string`. It seems like the data that is passed into the wizard is just the `projectName: string`, so I'm wondering if we really need the full `ProjectKind` object.

In the end it doesn't really matter, and no change is required, but I think all usages of `project` end up being `project.metadata.name`

emilys314: Oh so there are 2 `ProjectSelector` components. The one i was looking at was `src/concepts/projects/ProjectSelector.tsx` and it doesn't require the whole project.

You are currently using `src/pages/modelServing/screens/projects/InferenceServiceModal/ProjectSelector.tsx`

emilys314:
> If you go into the wizard without a project it won't load existing connections or pvcs so you'll have to pick a new connection:
Screenshot 2025-10-30 at 4 03 21 PM

Oh yeah I just realized that, I don't know how to feel about that

packages/model-serving/src/components/deploymentWizard/utils.ts
emilys314: Currently you are still able to navigate to `/ai-hub/deployments/<namespace>/deploy`, it acts like no project is selected.

We should remove this route, it's in `src/GlobalModelsRoutes.tsx`

packages/model-serving/extensions/odh.ts
```diff
@@ -70,6 +70,16 @@ const extensions: (
       path: '/ai-hub/deployments/:namespace?/*',
     },
   },
+  {
+    type: 'app.route',
+    properties: {
+      path: '/ai-hub/deployments/deploy',
```
emilys314: Can confirm that if you have a project called "deploy", then it will override and take you to the wizard 😅

emilys314 (commented):
The only required thing is to just removed the `/ai-hub/deployments/<namespace>/deploy` from `src/GlobalModelsRoutes.tsx` then i can approve

emilys314 (approved):
looks good, i just tested it again and there is no loader and the old route is gone

PR #5343, opened 2025-11-04
Sync `packages/model-registry` with upstream

manaswinidas:
## Request review criteria:

If you have UI changes: 

## Summary by CodeRabbit

* **New Features**
  * Added active filters UI and a global "Reset all filters" action in the model catalog.
  * Introduced category display names for model catalog filters.

* **Bug Fixes**
  * Fixed Server URL clipboard copy to exclude protocol prefix for model registry details.

* **Chores**
  * Updated Node.js runtime to v22 and bumped various frontend/backend dependencies.
  * Updated tracked upstream commit for the model-registry subtree.
  * Refined minimum RPS filter to use per-replica values and removed the total RPS column.

### DIFF
--- a/packages/model-registry/package.json
+++ b/packages/model-registry/package.json
@@ -22,7 +22,7 @@
     "repo": "https://github.com/kubeflow/model-registry.git",
     "src": "clients/ui",
     "target": "upstream",
-    "commit": "9cc66a1704b03dceb7b677af477949dc92e849ba"
+    "commit": "47a5c581b77a2c3fe0843eedfd39f1613fffebce"
   },
   "module-federation": {
     "name": "modelRegistry",


mturley (changes requested):
Looks like we have a problem related to https://github.com/kubeflow/model-registry/pull/1817 that's only appearing when this hits ODH - worked correctly upstream in the PF theme, not sure if it was an issue of resolving conflicts? The search bar width is small and the "Clear all filters" link shows up even if there are no filters

Image

We should troubleshoot as part of this PR just in case it's a conflict resolution issue but if it's not we can open a separate bug

ppadti:
> Looks like we have a problem related to [kubeflow/model-registry#1817](https://github.com/kubeflow/model-registry/pull/1817) that's only appearing when this hits ODH - worked correctly upstream in the PF theme, not sure if it was an issue of resolving conflicts? The search bar width is small and the "Clear all filters" link shows up even if there are no filters
> 
> Image
> We should troubleshoot as part of this PR just in case it's a conflict resolution issue but if it's not we can open a separate bug

I can still see the `reset all filters` issue upstream with both patternfly & mui theme on main, but the search bar width issue seems to just midstream.
Screenshot 2025-11-06 at 12 49 10 PM

manaswinidas:
@mturley The issue with the search bar exists in the [latest build](https://data-science-gateway.apps.dash-e2e-odh.osp.rh-ods.com/ai-hub/catalog) - so it's not a sync problem. And the "reset all filters" issue exists upstream - we need to fix it there.

Screenshot 2025-11-06 at 9 56 48 PM

PR #5350, opened 2025-11-05
restrict imports from frontend package

christianvogt:
Part of https://issues.redhat.com/browse/RHOAIENG-27202

## Description

As we work towards having clearer boundaries for reusing code from the frontend package, we need to put in place restrictions and guidance for developers to do the right thing. This adds eslint warnings to inform devs about poor dependencies. These aren't errors right now because of the number of warnings that need to be fixed up.

Packages should not be importing from the frontend/src/pages directory. In general old code from this directory should be migrated to their respective feature package or made into reusable components / utils instead of creating bidirectional dependencies. 

I've also restricted some of the exports from the frontend/src directory so they won't be allowed to be imported at all from other packages.

The frontend should not be importing from feature packages. While it's possible for the need of a old feature to be referencing new package code, that can be dealt with on an exception basis. 

In order to have both a warning and error cases for the same eslint rule, I needed to duplicate the rule through the use of a custom eslint plugin.

## How Has This Been Tested?

`npm run lint`
CI

## Test Impact

N/A

## Request review criteria:

If you have UI changes: 

## Summary by CodeRabbit

* **Chores**
  * Added a warn-level lint rule enforcing import boundaries with allowlist and negation support, plus a custom ESLint plugin and central restriction config to apply it.
  * Adjusted lint overrides for mock files and updated imports in tests to use the new module aliasing.

* **Refactor**
  * Explicitly exposed AppContext and AppLauncher in package exports and tightened wildcard exports for internal paths.

### DIFF
--- a/frontend/src/.eslintrc.js
+++ b/frontend/src/.eslintrc.js
@@ -10,6 +10,30 @@ module.exports = require('@odh-dashboard/eslint-config')
           prefix: '#~',
         },
       ],
+      '@odh-dashboard/no-restricted-imports': [
+        'warn',
+        {
+          patterns: [
+            {
+              group: [
+                '!@odh-dashboard/app-config',
+                '!@odh-dashboard/app-config/**',
+                '!@odh-dashboard/plugin-core',
+                '!@odh-dashboard/plugin-core/**',
+                '!@odh-dashboard/jest-config',
+                '!@odh-dashboard/jest-config/**',
+                '!@odh-dashboard/tsconfig',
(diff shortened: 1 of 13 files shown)


jenny-s51 (commented):
Lgtm @christianvogt , tested this locally with `npm run lint`. Linter flagged existing violations as expected (feast and model-serving imports). 

Also tested in the `frontend` directory with a direct import from `model-serving` and it was correctly flagged by the new eslint rule.

Left a comment above.

PR #4976, opened 2025-09-24
Remove 'disableFineTuning' field from odhdashboardconfigs CRD definition

Gkrumbach07:
Signed-off-by: Gage Krumbach <gkrumbach@gmail.com>

https://issues.redhat.com/browse/RHOAIENG-34910

## Description

only removing the feature flag from the crd. full removal in 3.1

## How Has This Been Tested?

verify crd no longer has disbalefinetuning flag

## Test Impact

none, just a crd change. there are also no e2e tests for this flag

## Request review criteria:

Self checklist (all need to be checked):
- [x] The developer has manually tested the changes and verified that the changes work
- [x] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [x] The developer has added tests or explained why testing cannot be added (unit or cypress tests for related changes)
- [x] The code follows our [Best Practices](/docs/best-practices.md) (React coding standards, PatternFly usage, performance considerations)

If you have UI changes: 

- [ ] Included any necessary screenshots or gifs if it was a UI change.
- [ ] Included tags to the UX team if it was a UI/UX change.

After the PR is posted & before it merges:
- [ ] The developer has tested their solution on a cluster by using the image produced by the PR to `main`

## Summary by CodeRabbit

- Chores
  - Removed the deprecated configuration option "disableFineTuning" from the dashboard enablement settings in the CRD. Manifests or configurations that still include this field will fail validation on apply/update—please remove it. No other schema or behavior changes were made.

PR #5352, opened 2025-11-05
Restore and deprecate disableFineTuning field in OdhDashboardConfig CRD

Gkrumbach07:
https://issues.redhat.com/browse/RHOAIENG-38112

## Description

This PR fixes a breaking API change where the `disableFineTuning` field was incorrectly removed from the OdhDashboardConfig CRD without proper API upversioning.

**The Problem:**
- Removing fields from a CRD is a breaking change for existing resources
- Any deployed resources with `disableFineTuning` set would fail validation
- This violates Kubernetes API compatibility guidelines

**The Solution:**
Instead of removing the field, it is now properly marked as DEPRECATED following the established pattern:
- Added deprecation notice in the field description
- Added validation rule to prevent modifications: `!has(self.disableFineTuning) || self.disableFineTuning == oldSelf.disableFineTuning`
- Follows the same pattern as `disableAcceleratorProfiles` and `disableHardwareProfiles`

The field will remain in the CRD until it can be safely removed during a proper API version upgrade (e.g., from v1alpha to v1).

**Related:**
- Original removal PR: #4976
- Slack discussion: https://redhat-internal.slack.com/archives/C05SMJ09DD2/p1762359435915959

## How Has This Been Tested?

- ✅ Verified CRD YAML syntax is valid
- ✅ Confirmed no linter errors
- ✅ Verified the field follows the same deprecation pattern as other deprecated fields
- ✅ Checked that backend/frontend code still references this field properly

## Test Impact

This is a manifest-only change that restores backward compatibility. Testing will verify:
- Existing resources with `disableFineTuning` set continue to work
- Validation rule prevents modifications to the deprecated field
- CRD can be applied successfully

## Request review criteria:

- [x] The developer has added tests or explained why testing cannot be added (unit or cypress tests for related changes) - **N/A: This is a CRD manifest change; validation happens at the Kubernetes API level**
- [x] The code follows our [Best Practices](/docs/best-practices.md) (React coding standards, PatternFly usage, performance considerations) - **N/A: Manifest change only**

If you have UI changes: 

- [ ] Included any necessary screenshots or gifs if it was a UI change. - **N/A: No UI changes**
- [ ] Included tags to the UX team if it was a UI/UX change. - **N/A: No UI changes**

## Summary by CodeRabbit

* **Chores**
  * Marked the dashboard setting "disableFineTuning" as deprecated and scheduled for removal in a future release.
  * Added validation to ensure the deprecated setting is not changed.
  * System now consistently applies the deprecated flag where feature lockouts are enforced.

### DIFF
--- a/backend/src/utils/resourceUtils.ts
+++ b/backend/src/utils/resourceUtils.ts
@@ -534,6 +534,11 @@ const applyFeatureLockouts = (config: DashboardConfig): DashboardConfig => ({
        * Model Mesh is removed in v3.0
        */
       disableModelMesh: true,
+
+      /**
+       * Fine Tuning feature is no longer supported
+       */
+      disableFineTuning: true,
     },
   },
 });


andrewballantyne (approved):
Design effort fits, just need to test before and after the PR effort. 

/hold 

Until tested

PR #5362, opened 2025-11-05
Infinite Loading State Fix for deployments table in MR

YuliaKrimerman:
## Description

Fixes an issue where users with proper permissions could not view deployments from the Model Registry deployments tab. The page showed infinite loading skeletons with "0 - 0 of 0" results.

Root Causes:
1. Missing ClusterRole permissions for servingruntimes and llminferenceservices resources
2. Overly strict loading condition requiring ALL resources (InferenceServices, ServingRuntimes, Pods) to load before showing deployments, even though ServingRuntimes and Pods are optional metadata
3. Incorrect filtering logic that removed deployments in system namespaces when using Model Registry label selectors

## How Has This Been Tested?

Tested on cluster: https://api.ods-qe-psi-16.osp.rh-ods.com:6443

Test Steps:
1. Verified user could list projects and InferenceServices via oc auth can-i commands
2.. Navigated to Model Registry deployments tab as ldap-user1
3. Confirmed deployments now visible in table (granite-7b-redhat-lab-140d-v in knative-serving namespace)
4. Verified no infinite loading states
Screenshot 2025-11-05 at 2 55 09 PM

## Test Impact

No new tests added. Changes are to existing loading logic and RBAC manifest

## Request review criteria:

If you have UI changes: 

## Summary by CodeRabbit

* **Bug Fixes**
  * Deployments now properly load and display when permission-related errors occur, eliminating indefinite loading states.
  * Improved deployment filtering for various project scoping configurations, ensuring consistent visibility across different setup scenarios.

### DIFF
--- a/packages/model-serving/src/concepts/ModelDeploymentsContext.tsx
+++ b/packages/model-serving/src/concepts/ModelDeploymentsContext.tsx
@@ -65,7 +65,7 @@ const PlatformDeploymentWatcher: React.FC<PlatformDeploymentWatcherProps> = ({
       const deploymentNamespace = deployment.model.metadata.namespace;
       return deploymentNamespace && projectNames.has(deploymentNamespace);
     });
-  }, [allDeployments, projects]);
+  }, [allDeployments, projects, labelSelectors]);
 
   React.useEffect(() => {
     onStateChange(platformId, { deployments: filteredDeployments, loaded, error });
--- a/frontend/src/__tests__/cypress/cypress/tests/mocked/modelRegistry/modelVersion/modelVersionArchive.cy.ts
+++ b/frontend/src/__tests__/cypress/cypress/tests/mocked/modelRegistry/modelVersion/modelVersionArchive.cy.ts
@@ -219,12 +219,8 @@ describe('Archiving version', () => {
 
   it('Cannot archive version that has a deployment from versions table', () => {
     cy.interceptK8sList(ProjectModel, mockK8sResourceList([mockProjectK8sResource({})]));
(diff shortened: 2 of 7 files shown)


packages/model-serving/src/concepts/ModelDeploymentsContext.tsx
```diff
@@ -55,7 +55,11 @@ const PlatformDeploymentWatcher: React.FC<PlatformDeploymentWatcherProps> = ({
 
   // Filter deployments to only include those from the specified projects
   const filteredDeployments = React.useMemo(() => {
-    if (!allDeployments || projects.length === 1) {
+    if (
+      !allDeployments ||
+      projects.length === 1 ||
+      (labelSelectors && Object.keys(labelSelectors).length > 0)
```
mturley: This means if any `labelSelectors` are passed at all, we will not filter deployments based on the `projects` passed in. I think there are some views that use this context and only want to show deployments from certain projects, and I'm not sure if any of them are using labelSelectors. I'm also generally a little confused about the root cause here - without this fix, can the user see all deployments across all their projects on the Deployments page, but the same fetching doesn't work when we apply labelSelectors?

Not sure if @emilys314 has some context here, I get the sense that we need a different fix.

emilys314: With this
> Incorrect filtering logic that removed deployments in system namespaces when using Model Registry label selectors

You should just change the input params of the `ModelDeploymentsContext` so that `projects = undefined` so that there is no project scoping

PR #5521, opened 2025-11-19
Apply unified hook to model serving

nananosirova:
https://issues.redhat.com/browse/RHOAIENG-38062

## Description

- Replaced the hardware profile logic in model serving with the unified hook `useAssignHardwareProfile` to standardize hardware profile management across different resource types. It's already in use for workbenches.
- Moved the loading of project-scoped hardware profiles from `ProjectsDetailsContext` to `HardwareProfilesContext` as wizard isn't wrapped by `ProjectDetailsContext`. This in turn ensures that both global and project-scoped hardware profiles are handled.

## How Has This Been Tested?

Need to regression test, have the latest RHOAI cluster. 
Under projects, navigate to Deployments tab and create deployments.
- Check that a global-scoped hardware profile binds successfully
- Check that a project-scoped hardware profile binds successfully
- Check that customization of requests and limits is successful

Other:
- Verify whether loading project-scoped hardware profiles in `HardwareProfilesContext` is the best way to go

Note: NIM isn't in wizard, not fully supported as plugin so it's not using the unified hook

## Test Impact

Added unit tests.

## Request review criteria:

If you have UI changes: 

## Summary by CodeRabbit

* **New Features**
  * Visibility presets (workbench, model-serving) and configurable hardware-profile path constants exposed; new hook returns profile options and an apply-to-resource function for deployments and model-serving flows.
* **Bug Fixes**
  * More consistent namespace-aware profile resolution and clearer loaded/error reporting when assigning profiles.
* **Refactor**
  * APIs and component props simplified to be prop-driven and callback-based for hardware profiles.
* **Tests**
  * Consolidated, extended, and parameterized tests covering notebooks and inference services.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/frontend/src/concepts/hardwareProfiles/useHardwareProfileConfig.ts
+++ b/frontend/src/concepts/hardwareProfiles/useHardwareProfileConfig.ts
@@ -131,15 +131,9 @@ export const useHardwareProfileConfig = (
     useExistingSettings: false,
   });
 
-  let profiles = dashboardProfiles;
-  let profilesLoaded = dashboardProfilesLoaded;
-  let profilesLoadError = dashboardProfilesLoadError;
-
-  if (projectScopedProfiles.length > 0) {
-    profiles = [...dashboardProfiles, ...projectScopedProfiles];
-    profilesLoaded = dashboardProfilesLoaded && projectScopedProfilesLoaded;
-    profilesLoadError = dashboardProfilesLoadError || projectScopedProfilesLoadError;
-  }
+  const profiles = [...dashboardProfiles, ...projectScopedProfiles];
+  const profilesLoaded = dashboardProfilesLoaded && projectScopedProfilesLoaded;
+  const profilesLoadError = dashboardProfilesLoadError || projectScopedProfilesLoadError;
 
   const isFormDataValid = React.useMemo(() => isHardwareProfileConfigValid(formData), [formData]);
   const { kueueFilteringState } = useKueueConfiguration(currentProject);


frontend/src/utilities/useWatchMultiNamespaceHardwareProfiles.ts
```diff
@@ -0,0 +1,66 @@
...
+import { POLL_INTERVAL } from './const';
+import { allSettledPromises } from './allSettledPromises';
+
+export const useWatchMultiNamespaceHardwareProfiles = (
+  namespaces: string[],
+): CustomWatchK8sResult<HardwareProfileKind[]> => {
+  const [combinedData, setCombinedData] = React.useState<HardwareProfileKind[]>([]);
+  const [isLoaded, setIsLoaded] = React.useState(false);
+  const [error, setError] = React.useState<Error>();
+  const namespacesString = React.useMemo(() => namespaces.toSorted().join(','), [namespaces]);
```
jrenee42: clearer to rename this 'allNamespacesString'

frontend/src/utilities/useWatchMultiNamespaceHardwareProfiles.ts
```diff
@@ -0,0 +1,66 @@
...
+        setError(
+          new Error(
+            `Failed to fetch hardware profiles from ${failures.length} namespace(s):\n${failures
+              .map(({ reason }) => `  - ${reason.message}`)
+              .join('\n')}`,
+          ),
+        );
+      }
+    } else {
+      setError(undefined);
```
jrenee42: ~~better to have a message here; so that if this happens then the user/debugger has more information;~~
~~something like: your previous message; and stating that the~~
...Oh; this is marking the watcher as not having any errors and succeeding.  
it was unclear.  please add a comment stating that it succeeded

jrenee42 (commented):
first round review; a few comments (adding some clarification, one potential error that is an easy fix)

frontend/src/__tests__/cypress/cypress/tests/mocked/modelServing/modelServingLlmd.cy.ts
```diff
@@ -195,15 +195,14 @@ describe('Model Serving LLMD', () => {
...
       row
         .findDescriptionListItem('Model server size')
         .next('dd')
-        .should('contain.text', '1 CPUs, 4GiB Memory requested');
+        .should('contain.text', '1 CPUs, 2GiB Memory requested');
       row
         .findDescriptionListItem('Model server size')
         .next('dd')
-        .should('contain.text', '2 CPUs, 8GiB Memory limit');
+        .should('contain.text', '1 CPUs, 2GiB Memory limit');
```
srtanish1992: Were the old test values incorrect, or does this represent a change in behavior?

frontend/src/concepts/hardwareProfiles/utils.ts
```diff
@@ -310,7 +315,7 @@ export const applyHardwareProfileConfig = <T extends K8sResourceCommon>(
 
   if (!useExistingSettings && selectedProfile) {
     if (resources) {
-      set(result, paths.containerResourcesPath, resources);
+      set(result, paths?.containerResourcesPath || '', resources);
```
srtanish1992: Instead of setting an empty string we can skip setting result if it undefined something like below logic, but let me know if setting to an empty string is intentional and missing a use case.
```
if (!useExistingSettings && selectedProfile && paths) {
  if (resources && paths.containerResourcesPath) {
    set(result, paths.containerResourcesPath, resources);
  }
}
```

frontend/src/pages/hardwareProfiles/useHardwareProfilesByFeatureVisibility.ts
```diff
@@ -1,23 +1,37 @@
...
   } = React.useContext(ProjectDetailsContext);
 
+  const [givenProjectProfiles, givenProjectProfilesLoaded, givenProjectProfilesError] =
+    projectHardwareProfiles || [];
+  const projectProfiles = givenProjectProfiles || projectProfilesFromContext;
+  const projectProfilesLoaded =
+    givenProjectProfilesLoaded ||
+    (currentProject.metadata.name ? projectProfilesFromContextLoaded : true);
+  const projectProfilesError = givenProjectProfilesError || projectProfilesFromContextError;
+
```
DaoDaoNoCode: When we start to find the prop drilldowns like this, it's time to take a step back to rethink the architecture because there are so many overlaps with the `projectProfilesFromContext` and `givenProjectProfiles`. And if you trace them down, they are actually fetched mostly in the same way and use the same data, we only need the minimum data we need, and let the hook finish the logic perfectly. The callers will never need to worry about getting the extra hardware profiles and passing it into this hook then, and they never should.

frontend/src/pages/hardwareProfiles/useHardwareProfilesByFeatureVisibility.ts
```diff
@@ -1,22 +1,60 @@
...
+  // Only watch if we actually need to fetch (pass undefined to disable the watch)
+  const namespaceToWatch = shouldFetchProfiles ? namespace : undefined;
+  const fetchedProjectProfilesResult = useWatchHardwareProfiles(namespaceToWatch);
+
+  const [projectProfiles, projectProfilesLoaded, projectProfilesError] = shouldUseContextProfiles
+    ? contextProjectProfiles
+    : shouldFetchProfiles
+    ? fetchedProjectProfilesResult
+    : [[], true, undefined];
```
DaoDaoNoCode: It's important to make the result reference stable because we will need to use the data later in the dependency arrays. This could prevent a lot of re-renders.


```suggestion
const [fetchedProfiles, fetchedProfilesLoaded, fetchedProfilesError] =
    useWatchHardwareProfiles(namespaceToWatch);

  // Determine which project profiles to use with stable references
  // Destructure the tuple to get stable references to the individual elements
  const projectProfilesResult = React.useMemo<
    [HardwareProfileKind[], boolean, Error | undefined]
  >(() => {
    if (shouldUseContextProfiles) {
      return contextProjectProfiles;
    }
    if (shouldFetchProfiles) {
      return [fetchedProfiles, fetchedProfilesLoaded, fetchedProfilesError];
    }
    // Stable empty result
    return [[], true, undefined];
  }, [
    shouldUseContextProfiles,
    shouldFetchProfiles,
    contextProjectProfiles,
    fetchedProfiles,
    fetchedProfilesLoaded,
    fetchedProfilesError,
  ]);

  const [projectProfiles, projectProfilesLoaded, projectProfilesError] = projectProfilesResult;
```

packages/model-serving/src/components/deploymentWizard/useDeploymentWizard.ts
```diff
@@ -35,6 +36,7 @@ export type UseModelDeploymentWizardState = WizardFormData & {
 export const useModelDeploymentWizard = (
   initialData?: InitialWizardFormData,
   initialProjectName?: string | undefined,
+  existingDeployment?: Deployment,
```
emilys314: Overall the changes look good.

I'd prefer if we didn't pass in the `Deployment` here, because settling on a specific resource kind can be difficult.

But the hook requires a resource to work so that is quite difficult

packages/kserve/src/hardware.ts
```diff
@@ -1,15 +1,35 @@
...
+  return [
+    name,
+    existingContainerResources,
+    existingTolerations,
+    existingNodeSelector,
+    MODEL_SERVING_VISIBILITY,
+    kserveDeployment.model.metadata.namespace,
+    hardwareProfileNamespace,
+  ];
 };
```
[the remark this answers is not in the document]
DaoDaoNoCode: @nananosirova This is valid, can you address it?

DaoDaoNoCode:
For project-scoped hardware profile, it shows me `Use existing settings` instead of the selected hardware profile.

Screenshot 2025-12-10 at 4 23 36 PM

PR #5573, opened 2025-12-01
fixing text wrapping in the pipelines project status

jrenee42:
for: https://issues.redhat.com/browse/RHOAIENG-36147

## Description
adjust css and combine text blocks so that the text flows naturally and is responsive.
if there is no css; then the text is left aligned (not center aligned) ,and when the window is wide then the text is too 'wide'.
patternfly doesn't have the capability to center it without using the patternfly classes; so using inline styling.

## How Has This Been Tested?
expanded (widened and narrowed ) the browser window.  watch below (with sound!  shows with and without the changes)

https://github.com/user-attachments/assets/f36faddd-8baa-427b-93f2-aa28adc3b1ea

## Test Impact
this is cosmetic, no impacts to tests at all

## Request review criteria:

If you have UI changes: 

## Summary by CodeRabbit

* **Style**
  * Improved layout and centering of status messages in the API availability interface for clearer visual presentation.
  * Consolidated multi-line status text into single, centered blocks to reduce clutter and improve readability.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/frontend/src/concepts/pipelines/EnsureAPIAvailability.tsx
+++ b/frontend/src/concepts/pipelines/EnsureAPIAvailability.tsx
@@ -17,6 +17,7 @@ type CenteredColumnContainerProps = {
   children: React.ReactNode;
 };
 
+// using inline styles since the Flex component does not support text alignment or max width
 const CenteredColumnContainer: React.FC<CenteredColumnContainerProps> = ({ children }) => (
   <Flex
     direction={{ default: 'column' }}


srtanish1992:
manually tested and works as expected
/lgtm

frontend/src/concepts/pipelines/EnsureAPIAvailability.tsx
```diff
@@ -35,9 +35,13 @@ const EnsureAPIAvailability: React.FC<EnsureAPIAvailabilityProps> = ({
   const pipelineServerName = getPipelineServerName(project);
   const defaultConnectingText = (
     <Flex direction={{ default: 'column' }} alignItems={{ default: 'alignItemsCenter' }}>
-      <FlexItem>The {pipelineServerName} connection is being established.</FlexItem>
-      <FlexItem>The process should take less than five minutes. When the server is ready,</FlexItem>
-      <FlexItem>you will be able to create and import pipelines.</FlexItem>
+      <FlexItem style={{ textAlign: 'center', maxWidth: '600px' }}>
```
DaoDaoNoCode: Can we apply the style to the parent `Flex` container so we don't need to create duplicated styles for every `FlexItem`? Haven't verified if that works, but that's my first thought.

rsun19: +1, plus you can use the `pf-v6-u-text-align-center` class from Patternfly instead of the CSS style.

rsun19: @Gkrumbach07 is this true?

jrenee42:
ok; container for custom styles added and in use.  just waiting to hear back about the patternfly class question. @DaoDaoNoCode

rsun19:
>ok; container for custom styles added and in use. just waiting to hear back about the patternfly class question. >@DaoDaoNoCode

He's on vacation for the rest of the week; best to ask Gage

frontend/src/concepts/pipelines/EnsureAPIAvailability.tsx
Gkrumbach07: > I noticed the codebase uses a mix of inline style, PF classnames, and SCSS files. What’s the recommended guidance or priority for choosing between these approaches?

dont do any of them unless you absolutely have to. many PF components have ways to set styling without inline style={} so use that first. Otherwise inline styles with a comment on why or a SCSS file with a comment.

jrenee42: ok; will use the inline styles then since the Flex box doesn't have the styles we need

PR #6199, opened 2026-02-08
Dashboard Air-Gapped NIM Support Implementation

mtalvi:
https://issues.redhat.com/browse/NVPE-394
## Description

All relevant info is [here.](https://docs.google.com/document/d/1aDB_p6EawPYyoFSzxiXk72Slsu8eCVh3KFwGznlRFdI/edit?tab=t.0)
This implementation adds support for air-gapped NIM deployments in the OpenShift AI Dashboard by allowing custom container registries and imagePullSecrets to be configured through the `odh-nim-account-cm` ConfigMap. In air-gapped mode, NGC secrets are created with dummy placeholder values to maintain controller compatibility, while actual image pulling uses the custom imagePullSecret.

**Assumptions and blockers:**
@trujillm  PRs have been merged:

1. [DSC API changes](https://github.com/opendatahub-io/opendatahub-operator/pull/3115): Added `spec.components.kserve.nim.airGapped` boolean flag

2. [odh-model-controller](https://github.com/opendatahub-io/odh-model-controller/pull/677) changes: Skip external API calls when `airGapped: true `and populate `odh-nim-account-cm` ConfigMap

## Request review criteria:

If you have UI changes: 

## Summary by CodeRabbit

* **New Features**
  * Air-gapped deployment support: propagate image pull secrets into serving runtimes, auto-fill dummy offline credentials, and update enable modal flow/labels/progress.
  * New account-configuration hook exposing registry and imagePullSecret; model registry resolution now prefers per-model → account → default.
  * UI: propagate selected model and registry info through serving/runtime selection.

* **Bug Fixes**
  * PVC selector avoids unintended clears when no compatible PVCs exist.

* **Tests**
  * Extensive coverage for air-gapped flows, model parsing, hook behavior, and UI interactions.


frontend/src/pages/exploreApplication/EnableModal.tsx
```diff
@@ -31,6 +34,14 @@ const EnableModal: React.FC<EnableModalProps> = ({ selectedApp, onClose }) => {
...
   const debounceTimeoutRef = React.useRef<ReturnType<typeof setTimeout> | null>(null);
 
+  // Detect air-gapped mode for NIM
+  const isNIMApp = selectedApp.spec.internalRoute === '/api/integrations/nim';
+  const accountConfig = useNIMAccountConfig();
+  const isAirGapped =
+    isNIMApp &&
+    !accountConfig.loading &&
+    !!(accountConfig.registry || accountConfig.imagePullSecret);
+
```
trujillm: General concern here is the silent fallback to nvcr.io when ConfigMap fetch fails in air-gapped mode

The air-gapped detection in this PR is based entirely on the odh-nim-account-cm ConfigMap fields (registry and imagePullSecret). It does not read the actual spec.components.kserve.nim.airGapped flag from the DSC, which is the source of truth that the odh-model-controller uses.

If the ConfigMap fetch fails (RBAC issue, ConfigMap not yet created, etc.), registry and imagePullSecret stay undefined, and the air-gapped detection silently falls back to the normal flow across all three consuming components:

1. useNIMAccountConfig.ts (lines 41-48) - The catch block stores the error but registry and imagePullSecret stay undefined. The comment says "this is not an error condition" but in air-gapped mode it is.

2. NIMModelListSection.tsx (line 121 in the diff) - accountConfig.registry is undefined, so it falls back to nvcr.io:
`const registry = result.modelInfo.registry || accountConfig.registry || 'nvcr.io';`

3. ManageNIMServingModal.tsx (line 23 in the diff) - isAirGapped is false because both fields are undefined:
`const isAirGapped = !!(accountConfig.registry || accountConfig.imagePullSecret);`

4. EnableModal.tsx (lines 19-22 in the diff) - Same. The user sees the NGC API key prompt instead of the "Air-Gapped Configuration Detected" screen.

5. None of the three consuming components check accountConfig.error.

 I would suggest using the DSC airGapped flag (spec.components.kserve.nim.airGapped) as the primary air-gapped detection instead of relying on the ConfigMap fields. The backend already has **backend/src/utils/dsc.ts** which can fetch DSC data. Then use the ConfigMap for the registry and imagePullSecret values, and if the ConfigMap fetch fails while airGapped is true, show an error message telling the user the odh-nim-account-cm ConfigMap is missing or inaccessible and needs to be created by a cluster admin.
This way the air-gapped detection is reliable (tied to the DSC flag, same source the controller uses) and the user gets clear feedback if the ConfigMap isn't set up yet.

andrewballantyne (changes requested):

emilys314:
Closing since we are adding NIM to the wizard and this PR won't be needed and will be outdated

PR #6734, opened 2026-03-17
fix(e2e): handle other catalog sources when verifying disable in testSourceEnableDisable

manaswinidas:
Fix for [e2e test failure today](https://jenkins-csb-rhods-opendatascience.dno.corp.redhat.com/job/cypress/job/dashboard-tests/4462/Test_20Reports/) 

The test previously assumed disabling the two Red Hat AI sources would always produce an empty catalog state. On clusters with additional catalog sources still enabled, this assertion fails. The test now checks for remaining enabled sources and expects either cards (if others exist) or empty state (if none remain).

[Thread for context](https://redhat-internal.slack.com/archives/C05SMJ09DD2/p1773732232081309)

This is the exact test that was failing(see screenshot) and the PR attempts to fix it 
Screenshot 2026-03-17 at 1 34 53 PM

## Request review criteria:

If you have UI changes: 

## Release Notes

* **Tests**
  * Improved test reliability for source enable/disable workflows with enhanced state verification
  * Added server-side response validation for configuration update operations
  * Extended test data coverage with additional source configurations
  * Refined testing utilities for more accurate catalog state detection


packages/cypress/cypress/utils/oc_commands/modelCatalog.ts
```diff
@@ -386,6 +386,76 @@ export const waitForModelCatalogCards = (
...
+      if command -v yq >/dev/null 2>&1; then
+        ${getYaml} | yq -o=json | jq '${buildJqFilter()}'
+      else
+        ${getYaml} | python3 -c 'import sys, yaml, json; json.dump(yaml.safe_load(sys.stdin), sys.stdout)' 2>/dev/null | jq '${buildJqFilter()}'
+      fi
+    `.trim();
+  };
+
+  const userCmd = buildCommand('model-catalog-sources');
+  const defaultCmd = buildCommand('model-catalog-default-sources');
```
FedeAlonso: Does this work on both Upstream and Downstream?

manaswinidas: yes

packages/cypress/cypress/utils/oc_commands/modelCatalog.ts
```diff
@@ -386,6 +386,76 @@ export const waitForModelCatalogCards = (
...
+export const hasOtherEnabledCatalogSources = (
+  excludeSourceIds: string[],
+): Cypress.Chainable<boolean> => {
+  const namespace = getModelRegistryNamespace();
+  const excludeFilter = excludeSourceIds.map((id) => `"${id}"`).join(', ');
+
+  const buildJqFilter = (): string =>
+    `[.catalogs[] | select(.enabled == true) | select(.id | IN(${excludeFilter}) | not)] | length`;
+
+  const buildCommand = (configmap: string): string => {
```
antowaddle: Is there a simpler way to do this? My concern here is that the pythons commands additionally may also fail when executed on CI/Jenkins.

antowaddle (changes requested):
Can we also see the results from the local/Jenkins run here too when we have?

mturley (changes requested):

packages/cypress/cypress/utils/oc_commands/modelCatalog.ts
```diff
@@ -386,6 +400,76 @@ export const waitForModelCatalogCards = (
...
+ * @returns A Cypress chainable that resolves with true if at least one other source is enabled.
+ */
+export const hasOtherEnabledCatalogSources = (
+  excludeSourceIds: string[],
+): Cypress.Chainable<boolean> => {
+  const namespace = getModelRegistryNamespace();
+  const excludeFilter = excludeSourceIds.map((id) => `"${id}"`).join(', ');
+
+  const buildJqFilter = (): string =>
+    `[.catalogs[] | select(.enabled == true) | select(.id | IN(${excludeFilter}) | not)] | length`;
```
mturley: The issue feels significant, not sure if it's necessary to rewrite in TS but it may be cleaner.

🤖 **🔴 Critical: Incorrect enabled-by-default semantics**

The `jq` filter `select(.enabled == true)` only counts sources that explicitly have `enabled: true`. According to the model catalog conventions, sources with a missing `.enabled` field should be treated as **enabled by default**. This means your helper could return `false` on clusters that actually have other enabled sources (just using implicit defaults).

**Suggested fix:** Rewrite to use `js-yaml` (already a dependency) instead of the shell pipeline:

```typescript
const parseAndCount = (yamlContent: string): boolean => {
  const parsed = yaml.load(yamlContent) as { 
    catalogs: Array<{ id: string; enabled?: boolean }> 
  };
  const otherEnabled = parsed.catalogs.filter(
    (c) => (c.enabled !== false) && !excludeSourceIds.includes(c.id),
  );
  cy.log(`Other enabled sources found: ${otherEnabled.length}`);
  return otherEnabled.length > 0;
};

const userCmd = `oc get configmap model-catalog-sources -n ${namespace} -o jsonpath='{.data.sources\\.yaml}'`;
const defaultCmd = `oc get configmap model-catalog-default-sources -n ${namespace} -o jsonpath='{.data.sources\\.yaml}'`;

return execWithOutput(userCmd, 30).then((userResult: CommandLineResult) => {
  if (userResult.code === 0 && userResult.stdout.trim()) {
    return cy.wrap(parseAndCount(userResult.stdout));
  }
  return execWithOutput(defaultCmd, 30).then((defaultResult: CommandLineResult) => {
    if (defaultResult.code !== 0 || !defaultResult.stdout.trim()) {
      const maskedStderr = maskSensitiveInfo(
        [userResult.stderr, defaultResult.stderr].filter(Boolean).join('\n')
      );
      throw new Error(`Failed to read catalog sources configmaps: ${maskedStderr}`);
    }
    return cy.wrap(parseAndCount(defaultResult.stdout));
  });
});
```

This also fixes:
- ✅ Portability: no dependency on `yq`/`python3`/`jq` in CI
- ✅ Error handling: fails fast instead of assuming "no other sources"
- ✅ Simpler code: single language instead of shell + jq

packages/cypress/cypress/utils/oc_commands/modelCatalog.ts
```diff
@@ -386,6 +400,76 @@ export const waitForModelCatalogCards = (
...
+
+    return execWithOutput(defaultCmd, 30).then((defaultResult: CommandLineResult) => {
+      const defaultCount = parseInt(defaultResult.stdout.trim(), 10);
+      if (defaultResult.code === 0 && !Number.isNaN(defaultCount)) {
+        cy.log(`Other enabled sources (default configmap): ${defaultCount}`);
+        return cy.wrap(defaultCount > 0);
+      }
+
+      cy.log('Could not determine other enabled sources, assuming none');
+      return cy.wrap(false);
```
mturley: Feels maybe significant but I'm not sure if Claude is right about "Critical".

🤖 **🔴 Critical: Silent failure handling**

Lines 446-447 log "assuming none" and return `false` when parsing fails. This masks infrastructure errors and could cause misleading test failures.

**Fix:** This would be addressed by the TypeScript rewrite suggestion above, which throws an error with masked stderr instead of assuming "no other sources".

packages/cypress/cypress/tests/e2e/modelCatalog/testSourceEnableDisable.cy.ts
```diff
@@ -53,23 +59,29 @@ describe('Verify Model Catalog Source Enable/Disable', () => {
...
 
       cy.step('Navigate to catalog');
       modelCatalog.visit();
 
-      cy.step('Wait for model catalog to show empty state');
-      waitForModelCatalogEmptyState();
-
-      cy.step('Verify model catalog shows empty state');
-      modelCatalog.findModelCatalogEmptyState().should('exist');
+      cy.step('Wait for catalog to reflect disabled sources');
```
mturley, line 86: This one feels significant. I think `waitForModelCatalogAfterDisable` is missing a check.

🤖 **🔴 Critical: Missing assertion for specific source absence**

The test waits for UI stabilization but doesn't explicitly verify that the two disabled sources are no longer visible. If other sources are present, the test passes as long as *some* cards appear—even if the disabled sources are still incorrectly showing.

**Suggested addition** after line 83:

```typescript
cy.step('Wait for catalog to reflect disabled sources');
waitForModelCatalogAfterDisable([testData.redhatAiSourceId, testData.redhatAiSourceId2]);

// Add this:
cy.step('Verify disabled sources are not visible');
modelCatalog.findModelCatalogCards().should('exist').then(($cards) => {
  const visibleCards = $cards.toArray();
  const visibleSourceNames = visibleCards.map(card => 
    Cypress.$(card).text()
  );
  expect(visibleSourceNames.join(',')).not.to.include(testData.sourceName);
  expect(visibleSourceNames.join(',')).not.to.include(testData.sourceName2);
});
```

packages/cypress/cypress/tests/e2e/modelCatalog/testSourceEnableDisable.cy.ts
```diff
@@ -53,23 +59,29 @@ describe('Verify Model Catalog Source Enable/Disable', () => {
       cy.step('Navigate back to Model catalog settings');
       modelCatalogSettings.visit();
```
mturley, line 64: This one feels like it could be ignored.

🤖 **🟡 Medium: Scope network intercepts to specific sources**

The wildcard intercept `'**/source_configs/*'` could match unrelated PATCH requests. More specific matching would be safer:

```typescript
cy.intercept('PATCH', `**/source_configs/${testData.redhatAiSourceId}`).as('toggleSource1');
cy.intercept('PATCH', `**/source_configs/${testData.redhatAiSourceId2}`).as('toggleSource2');

cy.step(`Disable the ${testData.sourceName} source`);
modelCatalogSettings.findEnableToggle(testData.redhatAiSourceId).click();
cy.wait('@toggleSource1').its('response.statusCode').should('eq', 200);
// ...

cy.step(`Disable the ${testData.sourceName2} source`);
modelCatalogSettings.findEnableToggle(testData.redhatAiSourceId2).click();
cy.wait('@toggleSource2').its('response.statusCode').should('eq', 200);
```

packages/cypress/cypress/utils/oc_commands/modelCatalog.ts
```diff
@@ -386,6 +400,76 @@ export const waitForModelCatalogCards = (
...
+    });
+  });
+};
+
+/**
+ * Poll until the model catalog UI has stabilized after disabling sources.
+ * If other sources are still enabled, waits for cards to be present (no empty state).
+ * If no other sources remain, waits for the empty state to appear.
+ * @param disabledSourceIds The source IDs that were just disabled
+ * @param maxAttempts Maximum number of attempts (default: 20)
```
mturley: **📝 Minor: JSDoc parameter mismatch**

The JSDoc mentions `maxAttempts` and `pollIntervalMs` parameters that don't exist in the function signature. Remove these from the doc comment.

mturley (changes requested):
Hey, @manaswinidas apologies for the big AI review dump, please let me know if any of these aren't accurate. It seems like there might be some real issues here though. A few things from old reviews are unaddressed too, so some of this might be repetitive.

🤖 The core fix is sound, but I'd strongly recommend addressing the inline comments before merging to avoid potential false negatives and ensure the test actually validates what it intends to. The TypeScript rewrite is not just a style preference—it fixes real semantic bugs and makes the code more portable.

manaswinidas:
dashboard-e2e-tests/400

Screenshot 2026-04-24 at 12 38 24 PM

packages/cypress/cypress/tests/e2e/modelCatalog/testSourceEnableDisable.cy.ts
```diff
@@ -19,13 +19,20 @@ describe('Verify Model Catalog Source Enable/Disable', () => {
...
+        verifyModelCatalogSourceEnabled(testData.redhatAiSourceId, true);
+        verifyModelCatalogSourceEnabled(testData.redhatAiSourceId2, true);
       });
   });
 
   after(() => {
     cy.step('Re-enable model catalog sources via configmap');
     enableModelCatalogSource(testData.redhatAiSourceId);
     enableModelCatalogSource(testData.redhatAiSourceId2);
+    enableModelCatalogSource(testData.redhatAiSourceId3);
```
FedeAlonso, line 37: Why this is being enabled in the teardown but not in the test setup?

packages/cypress/cypress/tests/e2e/modelCatalog/testSourceEnableDisable.cy.ts
```diff
@@ -19,13 +19,20 @@ describe('Verify Model Catalog Source Enable/Disable', () => {
       .fixture('e2e/modelCatalog/testSourceEnableDisable.yaml', 'utf8')
       .then((yamlContent: string) => {
         testData = yaml.load(yamlContent) as ModelCatalogSourceTestData;
+      })
+      .then(() => {
+        enableModelCatalogSource(testData.redhatAiSourceId);
```
ConorOM1, line 24: why are we enabling the sources in the before hook ?

packages/cypress/cypress/tests/e2e/modelCatalog/testSourceEnableDisable.cy.ts
```diff
@@ -19,13 +19,20 @@ describe('Verify Model Catalog Source Enable/Disable', () => {
...
+        enableModelCatalogSource(testData.redhatAiSourceId2);
+        verifyModelCatalogSourceEnabled(testData.redhatAiSourceId, true);
```
ConorOM1, line 27: adding a check here is good, but we should add the new sourceId3 also

ConorOM1 (commented):
@manaswinidas looks good ! Added two small comments

packages/cypress/cypress/tests/e2e/modelCatalog/testSourceEnableDisable.cy.ts
manaswinidas: The `retryableBefore` ensures sources are enabled at the start of the test because if a previous run failed mid-test (after disabling sources but before the after cleanup ran), the sources would be left disabled. Or I shouldn't worry about this ?

packages/cypress/cypress/tests/e2e/modelCatalog/testSourceEnableDisable.cy.ts
manaswinidas: Addressed

PR #6822, opened 2026-03-24
[WIP] fix(hardwareProfiles): omit resources when user has not customized them

DaoDaoNoCode:
https://issues.redhat.com/browse/RHOAIENG-54386

## Description

When deploying a model or creating a workbench, the Dashboard always sends explicit resource requests/limits from the auto-selected hardware profile (defaulting to `defaultCount` for both requests and limits). This overrides the platform webhook defaults (which set requests=`defaultCount`, limits=`maxCount`), causing a behavioral difference between Dashboard and GitOps/CLI deployments.

### Root Cause

`applyHardwareProfileConfig` always wrote resources to the CR spec whenever a hardware profile was selected, even when the user never explicitly customized them. Additionally, `getContainerResourcesFromHardwareProfile` set limits to `defaultCount` instead of `maxCount`, diverging from the webhook behavior.

### Fix

Uses `isEqual` comparison against profile defaults (requests=`defaultCount`, limits=`maxCount`) at each decision point to determine whether resources should be written:

- **`getContainerResourcesFromHardwareProfile`**: limits now use `maxCount` (matching webhook behavior)
- **`applyHardwareProfileConfig`**: only writes resources to the CR when they differ from profile defaults; returns the CR unchanged when `useExistingSettings` is true
- **`assemblePodSpecOptions`**: only includes resources in `podSpecOptions` when they differ from profile defaults
- **`useHardwareProfileConfig` (edit path)**: detects whether existing CR resources match profile defaults to decide between showing the profile as selected vs "Use existing settings"

### Additional fixes

- **Broken `initialSelectGuard`**: replaced with a same-profile check that correctly prevents SimpleSelect auto-select from resetting customized resources
- **"Use existing settings" not shown on edit**: `allowExistingSettings` now uses `isEditing` so the option is always available when editing, allowing the user to switch between profiles and their existing settings freely
- **Edit with customized resources**: when existing CR resources differ from profile defaults, the form initializes with "Use existing settings" selected instead of the profile
- **Existing resources restored on switch-back**: an `existingResources` ref captures the initial CR resources so switching back to "Use existing settings" restores them
- **Table popover showing wrong values**: `NotebookTableRow` reads `formData.resources` (always populated with merged CR values); `DeploymentHardwareProfileCell` reads CR resources directly via `getExistingResources` for display

### Behavior summary

| Scenario | Resources on CR | Annotation on CR |
|----------|----------------|------------------|
| Create, no customization | Omitted (webhook sets defaults) | Written |
| Create, user customizes | Written | Written |
| Edit, resources match defaults | Omitted (webhook handles) | Written |
| Edit, resources customized | Written (preserved) | Written |
| Edit, "Use existing settings" | CR unchanged | CR unchanged |

### Files Changed

| File | Change |
|------|--------|
| `useHardwareProfileConfig.ts` | `initialized` ref for one-time init; `isEqual` for edit-path customization detection; branches `useExistingSettings`/`selectedProfile` |
| `HardwareProfileFormSection.tsx` | Same-profile check replaces broken guard; `existingResources` ref; `allowExistingSettings={isEditing}` |
| `utils.ts` | `getContainerResourcesFromHardwareProfile` limits use `maxCount`; `applyHardwareProfileConfig` early return for `useExistingSettings`, `isEqual` check; `assemblePodSpecOptions` conditional resources |
| `useModelServingPodSpecOptionsState.ts` | Mirrors `isEqual` resource omission (deprecated modelmesh path) |
| `NotebookTableRow.tsx` | Reads `formData.resources` for display |
| `DeploymentHardwareProfileCell.tsx` | Reads CR resources directly via `getExistingResources` for display |

## How Has This Been Tested?

- `npm run type-check` passes (all 17 packages)
- `npm run lint` passes (0 errors)
- Unit tests: 307 tests pass across 19 test suites covering `useHardwareProfileConfig`, `useAssignHardwareProfile`, `useModelServingHardwareProfileState`, `applyHardwareProfileConfig`, `assemblePodSpecOptions`, `HardwareProfileSelect`, `ManageNIMServingModal`, and `inferenceServices`
- Manually tested create/edit flows for model deployments and workbenches on a live cluster

## Test Impact

- **`useHardwareProfileConfig.spec.ts`**: Added tests for limits=maxCount on auto-selected profiles, edit path with customized vs default resources, "Use existing settings" behavior, GPU identifier merging with maxCount limits
- **`utils.spec.ts`**: Added tests for resource omission when matching defaults, GPU resource handling, `useExistingSettings` early return preserving CR state
- **`useAssignHardwareProfile.spec.ts`**: Updated expectations for `podSpecOptions.resources` being `undefined` when not customized
- **`useModelServingHardwareProfileState.spec.ts`**: Updated for conditional resource inclusion
- **`ManageNIMServingModal.spec.tsx`**: Updated mock formData
- **Cypress mocked tests**: Updated `notebookServer.cy.ts`, `modelServingDeploy.cy.ts`, `modelServingLlmd.cy.ts`, `workbench.cy.ts` to verify resources are omitted from request bodies when not customized

## Request review criteria:

### DIFF
--- a/frontend/src/concepts/hardwareProfiles/HardwareProfileFormSection.tsx
+++ b/frontend/src/concepts/hardwareProfiles/HardwareProfileFormSection.tsx
@@ -43,6 +43,7 @@ const HardwareProfileFormSection: React.FC<HardwareProfileFormSectionProps<PodSp
   } = useHardwareProfilesByFeatureVisibility(visibleIn, project);
 
   const [isExpanded, setIsExpanded] = React.useState(false);
+  const existingResources = React.useRef<ContainerResources | undefined>(undefined);
 
   React.useEffect(() => {
     if (initialHardwareProfile && hasValidationErrors) {
@@ -50,20 +51,32 @@ const HardwareProfileFormSection: React.FC<HardwareProfileFormSectionProps<PodSp
     }
   }, [initialHardwareProfile, hasValidationErrors]);
 
+  React.useEffect(() => {
+    if (formData.resources && !existingResources.current) {
+      existingResources.current = formData.resources;
+    }
+  }, [formData.resources]);
+
   const onProfileSelect = (profile?: HardwareProfileKind) => {
-    // if no profile provided, use existing settings
     if (!profile) {
       setFormData('selectedProfile', undefined);
       setFormData('useExistingSettings', true);
+      if (existingResources.current) {
+        setFormData('resources', existingResources.current);
+      }
       return;
     }
 
-    // Reset customization when changing profiles
-    const newResources = getContainerResourcesFromHardwareProfile(profile);
+    if (
+      formData.selectedProfile?.metadata.name === profile.metadata.name &&
+      formData.selectedProfile.metadata.namespace === profile.metadata.namespace
+    ) {
+      return;
+    }
 
     setFormData('selectedProfile', profile);
     setFormData('useExistingSettings', false);
-    setFormData('resources', newResources);
+    setFormData('resources', getContainerResourcesFromHardwareProfile(profile));
   };
 
   return (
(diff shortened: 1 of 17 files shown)


DaoDaoNoCode:
Follow-up pushed in `43a240b28`.

- Added the missing auto-selection unit coverage for `selectedByUser === false`
- Updated mocked KServe and LLMd deploy assertions so auto-selected profiles now explicitly omit hardware-profile annotations and resources until the user makes a selection
- Re-ran the targeted mocked model-serving specs locally and both passed

PR #6823, opened 2026-03-24
fix(model-catalog): remove min-width constraint causing search bar overflow

DaoDaoNoCode:
https://issues.redhat.com/browse/RHOAIENG-54566

## Description

The search bar in the Model Catalog view overflows horizontally when a user types into it.

### Root Cause

Two issues compound to cause the overflow:

1. **Parent flex chain doesn't propagate growth:** `Flex`, `ToolbarToggleGroup`, and `ToolbarGroup` all have default `flex-grow: 0`, so the `ToolbarItem` cannot grow to fill the available toolbar width — even though the toolbar itself has hundreds of pixels of space.

2. **Hardcoded `min-width: 600px`:** An inline `<style>` block forces `min-width: 600px` on an inner element via the `.toolbar-fieldset-wrapper` class, preventing the search input from ever shrinking below 600px.

### Fix

* Add `style={{ flex: 1 }}` to `Flex`, `ToolbarToggleGroup`, and `ToolbarGroup` so the parent chain allows the search `ToolbarItem` to grow to the full toolbar width
* Add `minWidth: 0` to the `ToolbarItem` style so the flex child can shrink below its content size
* Replace the inner CSS rules: swap the fragile nested `> div:first-child > div > div { min-width: 600px }` selector with two targeted class-based rules that make the search input container and its inner `pf-v6-c-input-group__item` fill available space with `flex: 1 1 0; min-width: 0`

The search input now fills the available toolbar width and shrinks gracefully — no overflow regardless of input length or viewport size.

Before:
Screenshot 2026-03-25 at 4 27 28 PM

After:
Screenshot 2026-03-25 at 4 27 47 PM

## How Has This Been Tested?

* Validated with Playwright MCP against the live federated dev server at `http://localhost:4010/ai-hub/catalog`:
  * Idle: search bar fills full width, placeholder fully visible
  * Long text: stays contained within toolbar, no overflow
* Ran `npm run test:type-check` in `packages/model-registry/upstream/frontend` — clean
* Ran `npm run test:unit` in `packages/model-registry/upstream/frontend` — all tests pass

## Test Impact

No behavior change — this is a CSS-only fix. Existing tests remain unaffected.

- [x] The developer has added tests or explained why testing cannot be added (CSS-only fix validated visually with Playwright)

If you have UI changes:


DaoDaoNoCode:
Follow-up pushed in `b39e584e7`.

- Added the test coverage requested by the ticket: one unit regression for the search toolbar container and one mocked Cypress regression for overflow on the catalog page
- Added page-object helpers for the new mocked assertion
- Updated the model-registry Jest setup so the new unit test runs in this package

Philip-Carneiro (changes requested):
This should not be done in midstream, but in the upstream.
This will work fine in the upstream (as this works today) and need to be tested via script to import this to midstream and test it.

Also this change should be going not only in this file, but in the `clients/ui/frontend/src/app/pages/mcpCatalog/screens/McpCatalogSourceLabelSelector.tsx`

To test it when create the PR, make a temporary branch in midstream and then cd packages/model-registry and run `npm run update-subtree -- --pr=<url>` with the URL of your upstream PR

DaoDaoNoCode:
Closing in favor of https://github.com/kubeflow/model-registry/pull/2567

PR #6829, opened 2026-03-24
fix(storage): correct storage percentage for shared and block storage

DaoDaoNoCode:
https://issues.redhat.com/browse/RHOAIENG-5352

## Description

Fixes incorrect storage percentage display caused by two separate issues:

1. **Unit mismatch**: `bytesAsRoundedGiB(inUseInBytes)` was divided by `parseFloat(getPvcTotalSize(obj))`, which stripped the unit suffix — so `1TiB` was treated as `1` instead of `1024` GiB, producing percentages like "10984% full".

2. **Shared storage (NFS/CephFS)**: On clusters using NFS-backed PersistentVolumes, `kubelet_volume_stats_used_bytes` reports the **entire NFS share usage** (e.g. 686 GiB), not per-PVC usage. Dividing by the PVC capacity (100 GiB) gives "686% full" even though the underlying filesystem is only ~57% utilized.

Screenshot 2026-04-07 at 6 24 13 PM
Screenshot 2026-04-07 at 6 24 46 PM

**Root cause confirmed on a live disconnected cluster:**

| Data Source | Value | What it represents |
|---|---|---|
| `pvc.spec.resources.requests.storage` | 20 GiB | What the user requested |
| `pvc.status.capacity.storage` | 100 GiB | What the PV declares |
| `kubelet_volume_stats_capacity_bytes` | 1198 GiB | Entire NFS share size |
| `kubelet_volume_stats_used_bytes` | 686 GiB | Entire NFS share used |
| `df -h` inside pod | 1.2T / 687G used / 58% | Actual filesystem view |

**Implementation:**

- Fetch both `kubelet_volume_stats_used_bytes` and `kubelet_volume_stats_capacity_bytes` in a **single Prometheus query** using a regex `__name__` selector — no additional HTTP requests vs the previous implementation
- When Prometheus-reported capacity exceeds PVC capacity (shared storage detected), use the Prometheus capacity as the denominator so the percentage matches what `df -h` shows inside the pod
- For block storage (EBS, gp3, etc.), Prometheus capacity ≈ PVC capacity, so behavior is unchanged
- Centralize the percentage calculation in `getPvcPercentageUsed()` so the warning icon and storage bar stay in sync
- Fix `getFullStatusFromPercentage` to use `>= 100` instead of `=== 100` — the error state was unreachable for over-capacity volumes
- Add `min-width` to the progress bar CSS so it never collapses in narrow table columns
- Migrate `usePrometheusQuery` from deprecated `useFetchState`/`FetchState` to `useFetch`/`FetchStateObject`
- Replace manual `useEffect` polling with built-in `refreshRate` option
- Remove deprecated `useMakeFetchObject` wrapper in `distributedWorkloads.ts`

**Files changed:**

- `frontend/src/api/prometheus/pvcs.ts` — single-query dual-metric fetch with `PVCUsageInfo` return type
- `frontend/src/api/prometheus/usePrometheusQuery.ts` — migrate to `useFetch`/`FetchStateObject`
- `frontend/src/api/prometheus/distributedWorkloads.ts` — drop `useMakeFetchObject` wrapper
- `frontend/src/pages/projects/utils.ts` — `getEffectiveCapacityGiB()` + updated `getPvcPercentageUsed()`
- `frontend/src/pages/projects/components/StorageSizeBars.tsx` — use effective capacity for max label
- `frontend/src/pages/projects/screens/detail/storage/StorageWarningStatus.tsx` — pass capacity to percentage calc
- `frontend/src/pages/projects/screens/detail/storage/utils.ts` — `>= 100` fix
- `frontend/src/components/ProgressBarWithLabels.scss` — `min-width: 3rem` safety net
- `frontend/src/api/prometheus/__tests__/pvcs.spec.ts` — updated for single-query + PVCUsageInfo
- `frontend/src/api/prometheus/__tests__/usePrometheusQuery.spec.ts` — updated for FetchStateObject
- `frontend/src/pages/projects/__tests__/utils.spec.ts` — NFS, block, threshold tests

## How Has This Been Tested?

- `cd frontend && npx jest --runTestsByPath src/api/prometheus/__tests__/pvcs.spec.ts src/api/prometheus/__tests__/usePrometheusQuery.spec.ts src/pages/projects/__tests__/utils.spec.ts` — 22 tests pass
- Validated the regex `__name__` Prometheus query returns both metrics in a single request on a live cluster (via direct Thanos query and via the dashboard backend `/api/prometheus/pvc` proxy)
- Verified on a live disconnected cluster with NFS-backed PVCs: progress bar shows ~57% (686 GiB / 1198 GiB) instead of 686%
- Verified block storage PVCs continue to work correctly (single query, same request count as before)
- `npm run lint` — passes
- `npm run type-check` — passes

## Test Impact

- 22 unit tests covering: shared storage (NFS) with Prometheus capacity fallback, block storage, TiB/GiB unit conversion, missing capacity, zero capacity, `getEffectiveCapacityGiB`, `getFullStatusFromPercentage` thresholds, `usePrometheusQuery` object return, and `usePVCFreeAmount` single-query behavior
- No Cypress changes needed — the Prometheus query format is mocked at the intercept level

## Summary by CodeRabbit

* **Bug Fixes**
  * More accurate storage usage percentages (uses effective capacity, supports Prometheus-derived capacity, tighter NaN/loaded handling, 2-decimal formatting).
  * Storage warning thresholds updated to classify >=100% as error; tooltip “Max” and progress values corrected for missing/effective capacities.
* **Style**
  * Progress bar minimum width increased for more consistent layout.
* **Tests**
  * Added/updated unit, hook, and e2e tests for PVC capacity/percentage, Prometheus parsing, and status thresholds.

### DIFF
--- a/frontend/src/api/prometheus/pvcs.ts
+++ b/frontend/src/api/prometheus/pvcs.ts
@@ -1,29 +1,48 @@
 import * as React from 'react';
 import { PersistentVolumeClaimKind } from '#~/k8sTypes';
+import { PrometheusQueryResponse } from '#~/types';
 import { POLL_INTERVAL } from '#~/utilities/const';
 import usePrometheusQuery from './usePrometheusQuery';
 
+export type PVCUsageInfo = {
+  usedInBytes: number | typeof NaN;
+  capacityInBytes: number | typeof NaN;
+};
+
+type PVCMetricResult = PrometheusQueryResponse<{ metric: { __name__: string } }>;
+
+const PROM_FETCH_OPTIONS = { refreshRate: POLL_INTERVAL };
+
+const findMetricValue = (result: PVCMetricResult | null, metricName: string): string | undefined =>
+  result?.data.result.find((r) => r.metric.__name__ === metricName)?.value[1];
+
 export const usePVCFreeAmount = (
   pvc: PersistentVolumeClaimKind,
-): [bytesInUse: number | typeof NaN, loaded: boolean, error: Error | undefined] => {
-  const [result, loaded, loadError, refetch] = usePrometheusQuery(
-    '/api/prometheus/pvc',
-    `namespace=${pvc.metadata.namespace}&query=kubelet_volume_stats_used_bytes{persistentvolumeclaim='${pvc.metadata.name}'}`,
+): [info: PVCUsageInfo, loaded: boolean, error: Error | undefined] => {
+  const pvcName = pvc.metadata.name;
+  const { namespace } = pvc.metadata;
+
+  const query =
+    pvcName && namespace
+      ? `namespace=${namespace}&query={__name__=~"kubelet_volume_stats_used_bytes|kubelet_volume_stats_capacity_bytes",persistentvolumeclaim='${pvcName}'}`
+      : undefined;
+
+  const {
+    data: result,
+    loaded,
+    error,
+  } = usePrometheusQuery<PVCMetricResult>('/api/prometheus/pvc', query, PROM_FETCH_OPTIONS);
+
+  const info: PVCUsageInfo = React.useMemo(
+    () => ({
+      usedInBytes: parseInt(findMetricValue(result, 'kubelet_volume_stats_used_bytes') || '', 10),
+      capacityInBytes: parseInt(
+        findMetricValue(result, 'kubelet_volume_stats_capacity_bytes') || '',
+        10,
+      ),
+    }),
+    [result],
   );
 
-  // TODO mturley we can probably get rid of this and just pass a refreshRate in fetchOptions to usePrometheusQuery
-  React.useEffect(() => {
-    const interval = setInterval(() => {
-      refetch();
-    }, POLL_INTERVAL);
-
-    return () => {
-      clearInterval(interval);
-    };
-  }, [refetch]);
-
-  const value = result?.data.result[0]?.value[1];
-  const usedInBytes = parseInt(value || '', 10);
-
-  return [usedInBytes, loaded, loadError];
+  return [info, loaded, error];
 };
--- a/frontend/src/api/prometheus/pvcs.ts
+++ b/frontend/src/api/prometheus/pvcs.ts
@@ -41,12 +41,11 @@ export const usePVCFreeAmount = (
     error,
   } = usePrometheusQuery<PVCMetricResult>('/api/prometheus/pvc', query, PROM_FETCH_OPTIONS);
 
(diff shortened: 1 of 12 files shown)


bobbravo2 (approved):
/lgtm

DaoDaoNoCode:
@bobbravo2 Sorry it's still in progress, will check it in these 2 days.

frontend/src/api/prometheus/pvcs.ts
```diff
@@ -1,29 +1,53 @@
...
 
-  const value = result?.data.result[0]?.value[1];
-  const usedInBytes = parseInt(value || '', 10);
+  const info: PVCUsageInfo = React.useMemo(
+    () => ({
+      usedInBytes: findMetricValue(result, 'kubelet_volume_stats_used_bytes'),
+      capacityInBytes: findMetricValue(result, 'kubelet_volume_stats_capacity_bytes'),
+    }),
+    [result],
+  );
```
nananosirova, line 49: nit: `result` is not a stable dependency. 
It might be worth doing the following instead:
```typescript
const usedInBytes = findMetricValue(result, 'kubelet_volume_stats_used_bytes');
const capacityInBytes = findMetricValue(result, 'kubelet_volume_stats_capacity_bytes');
const info: PVCUsageInfo = React.useMemo(
  () => ({ usedInBytes, capacityInBytes }),
  [usedInBytes, capacityInBytes],
);
```

nananosirova (approved):
/lgtm

Tested on a disconnected cluster with PVs with NFS server backend and on a connected AWS cluster with EBS block storage. 

The issue doesn't seem specific to disconnected environments but any cluster where PVs are backed by a shared filesystem instead of getting their own block storage. On shared storage, we get situations where the PVC requests 3gb, the bound PV declares 100gb, but the actual NFS filesystem is 799gb. To get the storage percentage for the PVC, the dashboard queries Prometheus for `kubelet_volume_stats_used_bytes`, which is exposed to Prometheus by kubelet, which in turn gets that stat by calling [`statfs()`](https://man7.org/linux/man-pages/man2/statfs.2.html), and in this case that's the entire shared NFS filesystem. The old implementation was using this value as the currently used storage but keeping the denominator as the PVC capacity, which is why we were getting those >100% percentages. This only worked for block storage because there the filesystem capacity happens to be equal to the PVC capacity.

This PR fixes this by also querying `kubelet_volume_stats_capacity_bytes` from Prometheus and using that as the denominator when it's greater than the PVC capacity, and therefore handles both storage backend types properly.

nananosirova:
/lgtm
thank you!

PR #6840, opened 2026-03-25
fix(RHOAIENG-53846): Show admin-aware error state for unavailable model registry

kanishka-commits:
https://redhat.atlassian.net/browse/RHOAIENG-53846

## Description

Fixes issue where all users saw the same error when the model registry was unavailable.

Adds role-based messaging:
- **Admins**: See troubleshooting message with settings link  
- **Non-admins**: See “contact your administrator” message  

Uses existing admin-detection pattern from `ModelRegistryCoreLoader`.

## Changes
- Added admin check and custom error page (`frontend/.../ModelRegistry.tsx`)  
- Added `unavailableErrorPage` prop  
- Introduced `OdhModelRegistry` wrapper  
- Updated routes to use wrapper  

## How Has This Been Tested?

- TypeScript type-checking passes for both `frontend` and `packages/model-registry` projects
- ESLint passes for `frontend`
- Verified the admin detection pattern matches the existing usage in `ModelRegistryCoreLoader` and `OdhModelRegistryCoreLoader`

## Test Impact

No existing tests cover the model registry error state. Manual testing is required to verify:
1. As admin: navigate to an unavailable model registry → see admin-specific message with settings link
2. As non-admin: navigate to an unavailable model registry → see "contact your administrator" message with "Who's my administrator?" popover
3. Verify existing "no registries" empty state behavior is unchanged

If you have UI changes:

## Summary by CodeRabbit

* **New Features**
  * Model registry can display custom error/unavailable pages via the empty-state UI.
  * New settings extension allows configuring a settings URL and page title for the model registry.
  * Updated default registry landing behavior to use an alternate registry view as the index route.


kanishka-commits (commented):
## Code Review

Thank you for this PR! The feature is well-intentioned and addresses a real UX gap. However, I've identified several issues that should be addressed before merging.

---

### 🔴 **High Priority Issues**

#### 1. Code Duplication
The `UnavailableErrorPage` component is implemented **twice** with slight variations:
- `frontend/src/pages/modelRegistry/screens/ModelRegistry.tsx` (lines 51-78)
- `packages/model-registry/upstream/frontend/src/odh/components/OdhModelRegistry.tsx` (lines 20-46)

**Suggestion**: Extract this into a shared component to ensure consistency and reduce maintenance burden. Any future changes would need to be made in two places otherwise.

#### 2. Inconsistent Settings Link Handling
- **Frontend version**: Uses `modelRegistrySettingsRoute()` helper function
- **Packages version**: Hardcoded string `/settings/model-resources-operations/model-registry`

**Suggestion**: Use the same route helper in both places. If that's not possible in the packages layer, please document why the hardcoded URL is necessary.

#### 3. Inconsistent Terminology
- **Frontend**: "Go to **AI registry settings**"
- **Packages**: "Go to **Model registry settings**"

**Suggestion**: Standardize the terminology. Which is the correct product term?

---

### 🟡 **Medium Priority Issues**

#### 4. Confusing Prop Manipulation Logic
In `packages/model-registry/.../ModelRegistry.tsx` (lines 59-61):
```typescript
loadError={hasCustomErrorPage ? undefined : loadError}
loaded={hasCustomErrorPage ? true : loaded}
empty={hasCustomErrorPage ? true : pageProps.empty}
emptyStatePage={hasCustomErrorPage ? unavailableErrorPage : pageProps.emptyStatePage}
```

This hijacks the empty state mechanism to display an error page, which is clever but not immediately obvious to future maintainers.

**Suggestion**: Add a comment explaining this pattern, or consider if there's a more explicit/conventional way to handle custom error pages.

#### 5. Missing Spinner Component
In `OdhModelRegistry.tsx` (line 62):
```typescript
return <Bullseye>Loading...</Bullseye>;
```

**Suggestion**: Use PatternFly's `Spinner` component for better UX consistency:
```typescript
import { Spinner } from '@patternfly/react-core';
return <Bullseye><Spinner /></Bullseye>;
```

---

### ✅ **What's Good**

1. ✅ Follows existing pattern from `ModelRegistryCoreLoader` and `OdhModelRegistryCoreLoader`
2. ✅ Includes `data-testid` for testing (`model-registry-unavailable-error`)
3. ✅ Clear PR description with context and rationale
4. ✅ Well-organized imports
5. ✅ Solves a real UX problem with appropriate permission-based messaging

---

### 📝 **Testing Recommendation**

While manual testing is acceptable, consider adding at least basic unit tests for:
- Admin detection logic
- Correct error message rendering based on permissions
- Link/action visibility based on user role

---

### **Recommendation**
**Request Changes** - primarily to address code duplication and inconsistencies. Once these are resolved, this will be a solid improvement to the user experience.

Let me know if you'd like help with any of the suggested refactoring!

manaswinidas (changes requested):
These changes need to be done in `packages/model-registry` rather than the dead code in `frontend` folder - can you try this this fix there - also this has to be done with the help of extensions like what we already have for admin aware states in model regsitry and model catalog. (More info about extensions can be found in [this doc](https://github.com/opendatahub-io/odh-dashboard/blob/main/docs/extensibility.md))

kanishka-commits:
## Review Comments Addressed

I've reviewed and addressed the feedback from the code reviews:

### ✅ Fixed Issues:

1. **CodeRabbit's feedback** - Added loading state handling in `frontend/src/pages/modelRegistry/screens/OdhModelRegistry.tsx`:
   - Now destructures `[isAdmin, isAdminLoaded]` from `useAccessAllowed`
   - Returns `null` while loading to prevent rendering with incorrect admin state
   
2. **Kanishka's feedback** - Replaced "Loading..." text with PatternFly Spinner component:
   - Updated `packages/model-registry/upstream/frontend/src/odh/components/OdhModelRegistry.tsx`
   - Now uses `<Spinner />` inside `<Bullseye>` for consistent loading UX

### 📝 Changes Made:

**frontend/src/pages/modelRegistry/screens/OdhModelRegistry.tsx:**
```typescript
const [isAdmin, isAdminLoaded] = useAccessAllowed(verbModelAccess('create', ModelRegistryModel));

if (!isAdminLoaded) {
  return null;
}
```

**packages/model-registry/upstream/frontend/src/odh/components/OdhModelRegistry.tsx:**
```typescript
if (!loaded) {
  return (
    <Bullseye>
      <Spinner />
    </Bullseye>
  );
}
```

These changes ensure the admin permission check completes before rendering the error page and provide a better loading experience.

**Note:** The `frontend` folder changes address the immediate feedback, but as @manaswinidas mentioned, the primary implementation is in the `packages/model-registry` folder which already uses the extension-based approach.

PR3MM:
Verified and error messages looks good to me 

**ADMIN**
image


**NON ADMIN**
image



**One suggestion** -  add cypress test which will cover the fix ,without a test, someone can change ModelRegistry, OdhModelRegistry, or shared ApplicationsPage behavior and break the fix

PR3MM (changes requested):
Verified unavailable registry UI after scaling deployment/test to 0 in odh-model-registries. Check Screenshot, still the  UnavailableModelRegistry path shows same non-admin “contact your administrator” for an admin user.



image

kanishka-commits (commented):
### Tested and Verified:

**ADMIN**
Image

**NON ADMIN**
Image

PR3MM:
verified manually 
SCR-20260518-mrbj


/lgtm

packages/model-registry/extensions.ts
```diff
@@ -8,6 +8,9 @@ import type {
 const CATALOG_SETTINGS_PAGE_TITLE = 'Model catalog settings';
 const CATALOG_SETTINGS_URL = '/settings/model-resources-operations/model-catalog';
 
+const REGISTRY_SETTINGS_PAGE_TITLE = 'Model registry settings';
+const REGISTRY_SETTINGS_URL = '/settings/model-resources-operations/model-registry';
```
manaswinidas, line 15: why aren't we using the one that's already declared instead of reinitializing ?

packages/model-registry/upstream/frontend/src/odh/const.ts
```diff
@@ -6,3 +6,6 @@ export const DEPLOY_BUTTON_TOOLTIP = {
 };
 
 export const DEPLOYMENTS_TAB_EXTENSION_ID = 'deployments';
+
+export const REGISTRY_SETTINGS_PAGE_TITLE = 'Model registry settings';
+export const REGISTRY_SETTINGS_URL = '/settings/model-resources-operations/model-registry';
```
manaswinidas, line 11: declared twice here - need a way to consolidate this and reduce code duplication

packages/model-registry/upstream/frontend/src/__tests__/cypress/cypress/tests/mocked/modelRegistry/modelRegistry.cy.ts
```diff
@@ -196,7 +196,7 @@ describe('Model Registry core', () => {
     // Navigate to the unavailable registry (app would redirect to first registry; we go there directly)
     cy.visit(modelRegistryUrl(unavailableRegistryName));
 
-    modelRegistry.findUnavailableModelRegistryState().should('exist');
+    modelRegistry.findUnavailableModelRegistry().should('exist');
```
manaswinidas: There should be an admin level and a non-admin level test

packages/model-registry/upstream/frontend/src/__tests__/cypress/cypress/pages/modelRegistry.ts
```diff
@@ -90,10 +90,14 @@ class ModelRegistry {
     return cy.findByTestId('empty-model-registries-state');
   }
 
-  findUnavailableModelRegistryState() {
+  findUnavailableModelRegistry() {
     return cy.findByTestId('unavailable-model-registry');
   }
 
+  findWhosMyAdministratorLink() {
```
manaswinidas, line 97: where is this used ?

packages/model-registry/upstream/frontend/src/__tests__/cypress/cypress/pages/modelRegistry.ts
```diff
@@ -90,10 +90,14 @@ class ModelRegistry {
     return cy.findByTestId('empty-model-registries-state');
   }
 
-  findUnavailableModelRegistryState() {
+  findUnavailableModelRegistry() {
```
manaswinidas: why was this renamed ?

packages/model-registry/upstream/frontend/src/__tests__/cypress/cypress/pages/modelRegistry.ts
```diff
@@ -90,10 +90,14 @@ class ModelRegistry {
...
     return cy.findByTestId('unavailable-model-registry');
   }
 
+  findWhosMyAdministratorLink() {
+    return cy.findByTestId('whos-my-admin-link');
```
manaswinidas, line 98: use `findWhosMyAdministratorLink()` link instead ?

packages/model-registry/upstream/frontend/src/odh/components/OdhModelRegistryCoreLoader.tsx
```diff
@@ -1,28 +1,87 @@
...
 import { ProjectObjectType, typedEmptyImage, WhosMyAdministrator } from 'mod-arch-shared';
-import { useResolvedExtensions } from '@odh-dashboard/plugin-core';
+import { useResolvedExtensions, useExtensions } from '@odh-dashboard/plugin-core';
 import EmptyModelRegistryState from '~/app/pages/modelRegistry/screens/components/EmptyModelRegistryState';
 import ModelRegistryCoreLoader from '~/app/pages/modelRegistry/ModelRegistryCoreLoader';
-import { isAdminCheckExtension } from '~/odh/extension-points';
+import { isAdminCheckExtension, isRegistrySettingsUrlExtension } from '~/odh/extension-points';
+import { REGISTRY_SETTINGS_PAGE_TITLE, REGISTRY_SETTINGS_URL } from '~/odh/const';
+
+type OdhUnavailableModelRegistryProps = {
```
manaswinidas: Extract the unavailable body text into a shared constant or compose `OdhUnavailableModelRegistry` in it's own file on top of `UnavailableModelRegistry` as a thin wrapper to avoid message drift, rather than introducing it here as in the core loader.

packages/model-registry/upstream/frontend/src/app/pages/modelRegistry/ModelRegistryRoutes.tsx
```diff
@@ -9,7 +9,6 @@ import {
   isModelRegistryDetailsTabExtension,
 } from '~/odh/extension-points';
 import ModelRegistry from './screens/ModelRegistry';
-
```
manaswinidas, line 10: is this intentional ? this file has only one new line as diff

manaswinidas (changes requested):

packages/model-registry/extensions.ts
kanishka-commits: `extensions.ts` can’t import from `upstream/.../const.ts` since the tsconfig there excludes the `upstream` folder.

So the duplicate constant is intentional- we already do the same thing for catalog settings (`CATALOG_SETTINGS_PAGE_TITLE` is defined in both places too).

The upstream value is really just a fallback for when the extension isn’t loaded.

packages/model-registry/upstream/frontend/src/odh/const.ts
kanishka-commits: Same as above- the tsconfig module boundary prevents sharing here. This follows the existing catalog settings pattern too.

packages/model-registry/upstream/frontend/src/app/pages/modelRegistry/ModelRegistryRoutes.tsx
kanishka-commits: Not intentional- the blank line was accidentally removed during a prior commit when the fix was moved from `frontend/` to `packages/model-registry/`. Added it back so this file drops out of the diff.

packages/model-registry/upstream/frontend/src/__tests__/cypress/cypress/tests/mocked/modelRegistry/modelRegistry.cy.ts
kanishka-commits: Added a Jest unit test with both admin and non-admin test cases (file:`OdhUnavailableModelRegistry.spec.tsx`).

packages/model-registry/upstream/frontend/src/__tests__/cypress/cypress/pages/modelRegistry.ts
kanishka-commits: Used in the Cypress test- `modelRegistry.findWhosMyAdministratorLink().should('exist')` (replaced the raw `cy.findByTestId `call in commit f2ccf63367336080b1607ac071315c22fbc48fd6 ).

packages/model-registry/upstream/frontend/src/__tests__/cypress/cypress/pages/modelRegistry.ts
kanishka-commits: Reverted- no reason to rename it.

packages/model-registry/upstream/frontend/src/odh/components/OdhModelRegistryCoreLoader.tsx
kanishka-commits: Extracted to `OdhUnavailableModelRegistry.tsx`. It imports `UNAVAILABLE_TITLE` and `getUnavailableBodyText` from the upstream `UnavailableModelRegistry` so the non-admin text is defined in one place.

packages/model-registry/upstream/frontend/src/__tests__/cypress/cypress/pages/modelRegistry.ts
kanishka-commits: Done- updated the test to use the page object method.

dpanshug:

packages/model-registry/upstream/frontend/src/odh/components/OdhModelRegistryCoreLoader.tsx
```diff
@@ -48,16 +60,27 @@ const OdhModelRegistryCoreLoader: React.FC<OdhModelRegistryCoreLoaderProps> = ({
...
+        registryDisplayName={registryDisplayName}
+        isAdmin={isAdmin}
+        settingsUrl={settingsUrl}
+        settingsTitle={settingsTitle}
+      />
+    );
+    return renderUnavailablePage;
+  };
+
   if (adminCheckExtensionsLoaded && adminCheckExtensions.length > 0) {
```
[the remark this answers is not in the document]
kanishka-commits: This is pre-existing code that was there before this PR. This PR didn't introduce this pattern; just added `unavailableStatePage` following the same pattern used by `emptyStatePage`. Fixing this is scope creep.

packages/model-registry/upstream/frontend/src/odh/components/OdhModelRegistryCoreLoader.tsx
```diff
@@ -48,16 +60,27 @@ const OdhModelRegistryCoreLoader: React.FC<OdhModelRegistryCoreLoaderProps> = ({
...
+              Go to <b>{settingsTitle}</b>
             </Link>
           )
         }
       />
     );
   };
 
-  // If an admin check extension is provided and loaded, use it
+  const createUnavailableStatePage = (isAdmin: boolean) => {
```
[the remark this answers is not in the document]
kanishka-commits: `createEmptyStatePage` was already using this pattern before the PR. The implementation was kept the same for consistency. Refactoring it to use `useMemo` can be handled in a separate optimization PR.

packages/model-registry/upstream/frontend/src/odh/components/OdhModelRegistryCoreLoader.tsx
```diff
@@ -48,16 +60,27 @@ const OdhModelRegistryCoreLoader: React.FC<OdhModelRegistryCoreLoaderProps> = ({
...
+        isAdmin={isAdmin}
+        settingsUrl={settingsUrl}
+        settingsTitle={settingsTitle}
+      />
+    );
+    return renderUnavailablePage;
+  };
+
   if (adminCheckExtensionsLoaded && adminCheckExtensions.length > 0) {
     const AdminCheckComponent = adminCheckExtensions[0].properties.component.default;
```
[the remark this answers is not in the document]
kanishka-commits: Pre-existing code, not introduced by this PR.

packages/model-registry/upstream/frontend/src/odh/components/OdhUnavailableModelRegistry.tsx
```diff
@@ -0,0 +1,61 @@
...
+    >
+      <EmptyStateBody>
+        {isAdmin
+          ? getAdminBodyText(registryDisplayName)
+          : getUnavailableBodyText(registryDisplayName)}
+      </EmptyStateBody>
+      <EmptyStateFooter>
+        <EmptyStateActions>
+          {isAdmin ? (
+            <Link to={settingsUrl} data-testid="registry-settings-link">
```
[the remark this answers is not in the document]
kanishka-commits: The existing `createEmptyStatePage` uses the same bare `<Link>` pattern. Keeping consistent; can address both together in a follow-up.

packages/model-registry/upstream/frontend/src/odh/components/OdhUnavailableModelRegistry.tsx
```diff
@@ -0,0 +1,61 @@
...
+const getAdminBodyText = (registryDisplayName: string): string =>
+  `The ${registryDisplayName} registry is currently unavailable. Check the registry configuration in settings to troubleshoot the issue.`;
+
+const OdhUnavailableModelRegistry: React.FC<OdhUnavailableModelRegistryProps> = ({
+  registryDisplayName,
+  isAdmin,
+  settingsUrl,
+  settingsTitle,
+}) => (
+  <PageSection hasBodyWrapper={false} isFilled data-testid="unavailable-model-registry">
```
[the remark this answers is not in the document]
kanishka-commits: Only one component renders at a time- `unavailableStatePage` replaces the default. No ambiguity in practice.

kanishka-commits:

manaswinidas (changes requested):
@kanishka-commits We are doing some refactoring to extensions - can you follow https://github.com/opendatahub-io/odh-dashboard/pull/7792 and make any changes if necessary ?

kanishka-commits:
@manaswinidas 
Checked #7792 - it introduces `createExtensionGuard` in plugin-core, but the existing model-registry guards still use the manual pattern. I can adopt it for `isRegistrySettingsUrlExtension` if preferred, though it would currently be the only guard in the package using the new approach. It may be cleaner to migrate them together in a follow-up.
Let me know what you suggest.

manaswinidas:
@kanishka-commits We are already migrating existing extensions here: https://github.com/opendatahub-io/odh-dashboard/pull/7874 - based on what goes in first - this may be the odd one out

kanishka-commits:
@manaswinidas
Thanks for the heads-up! I’ve updated `isRegistrySettingsUrlExtension` to use `createExtensionGuard` from `plugin-core`, similar to #7792. Once #7874 lands, we can align the remaining extension guards as needed- whichever PR merges later can easily adapt.

kanishka-commits:
@manaswinidas 
I've pushed the changes, kindly review. Thanks!

manaswinidas (changes requested):
@kanishka-commits there's merge conflict - the other extensions refactor went in first 😓  sorry

packages/model-registry/upstream/frontend/src/__tests__/cypress/cypress/tests/mocked/modelRegistry/modelRegistry.cy.ts
```diff
@@ -201,8 +201,7 @@ describe('Model Registry core', () => {
     cy.contains('The Unavailable Registry Example registry is currently unavailable').should(
       'be.visible',
     );
-    cy.findByTestId('whos-my-admin-link').should('exist');
-    // View details button should not be present when registry is unavailable
+    modelRegistry.findWhosMyAdministratorLink().should('exist');
```
manaswinidas, line 204: Can we add a similar cypress tests for admins ? I mentioned [this before](https://github.com/opendatahub-io/odh-dashboard/pull/6840#discussion_r3258309879) - we need a Cypress test to test the extension loading (which I don't think the Jest unit test does)

manaswinidas (changes requested):

packages/model-registry/upstream/frontend/src/__tests__/cypress/cypress/tests/mocked/modelRegistry/modelRegistry.cy.ts
kanishka-commits: I spent some time looking through the repo but couldn't find an existing pattern for mocking the extension system (`useResolvedExtensions` / `plugin-core`) in Cypress. The existing admin/non-admin Cypress tests rely on mocking the BFF API (`/api/v1/user`), which is a different path from the `AdminCheck` extension used here.

In Cypress component tests, the extension system isn't wired up, so `useResolvedExtensions` always returns `[]`, meaning only the non-admin path gets exercised. The Jest unit test already covers both admin and non-admin rendering.

I'm happy to add a Cypress admin test once we have a pattern for mocking `plugin-core`, but I think that's better handled as a separate effort rather than in this PR.

packages/model-registry/extensions.ts
```diff
@@ -10,12 +10,24 @@ import type {
...
+
+type RegistrySettingsUrlExtension = Extension<
```
ppadti: Why are we declaring this `RegistrySettingsUrlExtension `type here rather than importing it from `@mf/modelRegistry/extension-points`
And to import successfully you need to re-genarate the mf-types and then import `RegistrySettingsUrlExtension` here

packages/model-registry/upstream/frontend/src/odh/components/OdhModelRegistryCoreLoader.tsx
```diff
@@ -48,15 +59,27 @@ const OdhModelRegistryCoreLoader: React.FC<OdhModelRegistryCoreLoaderProps> = ({
...
+            <Link to={settingsUrl}>
+              Go to <b>{settingsTitle}</b>
             </Link>
           )
         }
       />
     );
   };
 
+  const createUnavailableStatePage = (isAdmin: boolean) => {
```
ppadti: `createUnavailableStatePage` creates a new closure on every render, use `React.useMemo` here.
we will followup to fix the `createEmptyStatePage`

ppadti (changes requested):
Whenever we add a new extension-point - we should always regenerate the mf-types

packages/model-registry/extensions.ts
kanishka-commits: Addressed in 4211a12cf7ffe91e4c6693a0e715fd218a0a4e9c.

packages/model-registry/upstream/frontend/src/odh/components/OdhModelRegistryCoreLoader.tsx
kanishka-commits: Addressed in 4211a12cf7ffe91e4c6693a0e715fd218a0a4e9c.

ppadti (approved):
Thanks @kanishka-commits for working on this - as @manaswinidas mentioned we should add cypress test to avoid any future regression - So could you create a SPIKE ticket and do the investigation on how can we cover the admin case? 

Tested this PR locally - verified both admin and non-admin case

Image

Image

/lgtm

manaswinidas (approved):
/approve

PR3MM (approved):

PR #6847, opened 2026-03-25
fix: use annotation instead of label to hide unsaved model-serving connections

dpanshug:
https://issues.redhat.com/browse/RHOAIENG-54163

## Description

When deploying a predictive model in the dashboard with "Create a connection to this location" unchecked, the dashboard sets `opendatahub.io/dashboard: "false"` on the connection secret. The model-serving controller (`odh-model-controller`) requires this label to be `"true"` to include the secret in `storage-config`. As a result, KServe's pod mutator webhook rejects the pod with:

```
admission webhook "inferenceservice.kserve-webhook-server.pod-mutator" denied the request:
specified storage key secret-eo6wli not found in storage secret storage-config
```

**Root cause:** The `opendatahub.io/dashboard` label was being used for two separate concerns:
1. Whether the connection should be visible in the dashboard connections list (UI concern)
2. Whether the model-serving controller should include it in `storage-config` (functional concern)

**Fix:** Instead of setting the `opendatahub.io/dashboard` label to `"false"`, we now keep it as `"true"` (so the controller works) and add an `opendatahub.io/connection-hidden: "true"` annotation to hide the connection from the dashboard UI.

Changes:
- **`packages/model-serving/src/concepts/connectionUtils.ts`**: Replace `dashboard: 'false'` label mutation with `connection-hidden: 'true'` annotation
- **`frontend/src/pages/projects/screens/detail/connections/useConnections.ts`**: Filter out connections with the `connection-hidden` annotation (unless `includeDashboardFalse` is true, e.g. in the deployment wizard)

## How Has This Been Tested?

In-session validation skipped due to resource limits; rely on CI/local verification.

**Manual test steps:**
1. Create a new predictive model deployment
2. Copy valid S3 connection field values into the form (don't select "Use existing connection")
3. Uncheck "Create a connection to this location"
4. Fill out the rest of the form and deploy
5. Verify the predictor pod scales to 1 successfully
6. Verify the hidden connection does NOT appear in the project's connections list
7. Verify that checking the checkbox still creates a visible, saved connection

## Test Impact

No existing unit tests cover `handleConnectionCreation` in the model-serving package. The `useConnections` hook filtering change is additive and backward-compatible. CI tests should pass without modification.

## Request review criteria:

If you have UI changes:
- N/A — This is a logic-only change, no visual UI changes.

## Summary by CodeRabbit

* **Refactor**
  * Hidden connections are now excluded from dashboard and selection lists when the dashboard filter is off.
* **Behavior**
  * Connections created without being saved are marked hidden so they don't appear in the dashboard by default; other creation, update, deletion and dry‑run behaviors are unchanged.
* **Tests**
  * End-to-end checks updated to reflect hidden-flag and dashboard visibility behavior.

### DIFF
--- a/frontend/src/pages/projects/screens/detail/connections/useConnections.ts
+++ b/frontend/src/pages/projects/screens/detail/connections/useConnections.ts
@@ -23,9 +23,13 @@ const useConnections = (
       const labelSelector = includeDashboardFalse ? '' : `${LABEL_SELECTOR_DASHBOARD_RESOURCE}`;
 
       const secrets = await getSecretsByLabel(labelSelector, namespace, opts);
-      const connections = secrets.filter((secret) => isConnection(secret));
-
-      return connections;
+      return secrets
+        .filter(isConnection)
+        .filter(
+          (connection) =>
+            includeDashboardFalse ||
+            connection.metadata.annotations['opendatahub.io/connection-hidden'] !== 'true',
+        );
     },
     [namespace, includeDashboardFalse],
   );
--- a/packages/cypress/cypress/tests/mocked/modelServing/modelServingDeploy.cy.ts
+++ b/packages/cypress/cypress/tests/mocked/modelServing/modelServingDeploy.cy.ts
@@ -1678,8 +1678,11 @@ describe('Model Serving Deploy Wizard', () => {
       expect(interception.request.body.metadata.name).to.satisfy(isGeneratedSecretName);
       expect(interception.request.body.metadata.namespace).to.equal('test-project');
       expect(interception.request.body.metadata.labels['opendatahub.io/dashboard']).to.equal(
-        'false',
+        'true',
       );
+      expect(
+        interception.request.body.metadata.annotations['opendatahub.io/connection-hidden'],
+      ).to.equal('true');
       expect(
         interception.request.body.metadata.annotations['opendatahub.io/connection-type-protocol'],
       ).to.equal('uri');
--- a/packages/model-serving/src/components/deploymentWizard/fields/ModelLocationInputFields.tsx
+++ b/packages/model-serving/src/components/deploymentWizard/fields/ModelLocationInputFields.tsx
@@ -289,9 +289,15 @@ export const ModelLocationInputFields: React.FC<ModelLocationInputFieldsProps> =
   customTypeOptions,
   customTypeKey,
 }) => {
-  const filteredConnections = React.useMemo(() => {
-    return connections.filter((c) => c.metadata.labels['opendatahub.io/dashboard'] === 'true');
-  }, [connections]);
+  const filteredConnections = React.useMemo(
+    () =>
+      connections.filter(
+        (c) =>
+          c.metadata.labels['opendatahub.io/dashboard'] === 'true' &&
+          c.metadata.annotations['opendatahub.io/connection-hidden'] !== 'true',
+      ),
+    [connections],
+  );
   const pvcNameFromUri: string | undefined = React.useMemo(() => {
     // Get the PVC name from the URI if it's a PVC URI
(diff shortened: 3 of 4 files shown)


katieperry4 (commented):
Just tried it out!
It makes a new generated connection with the new annotation and the dashboard label as true:

Image



the inference service references just the same as before:

Image



the connections tab, still filters it out so we don't see the generated connection in the list:

Image




it does however still show up in the existing connection dropdown in the wizard:

Image 

so I think we'd need to add the filter there too

katieperry4:
- makes the secret with the new annotation
- still adds the owner reference to the secret and the correct name in the inference service
- the generated secret is hidden in the connections tab and in the dropdown in the wizard
- still hides the generated connections with the old dashboard=false label from old deployments


https://github.com/user-attachments/assets/931a994a-a7a7-47ed-bc5e-ee4dd7efdccc



/lgtm

frontend/src/pages/projects/screens/detail/connections/useConnections.ts
```diff
@@ -23,9 +23,13 @@ const useConnections = (
...
       const secrets = await getSecretsByLabel(labelSelector, namespace, opts);
-      const connections = secrets.filter((secret) => isConnection(secret));
-
-      return connections;
+      return secrets
+        .filter(isConnection)
+        .filter(
+          (connection) =>
+            includeDashboardFalse ||
+            connection.metadata.annotations['opendatahub.io/connection-hidden'] !== 'true',
```
emilys314, line 31: Is this always going to remove connections where `connection-hidden = true`? 

@katieperry4 Aren't there some use cases where we would want to fetch all of them, even hidden ones?

This should be gated behind the `includeDashboardFalse` flag. (The name is kinda bad)

emilys314 (commented):
My brain is trying to process the implications of this

frontend/src/pages/projects/screens/detail/connections/useConnections.ts
katieperry4: ah you're right it should be behind `includeDashboardFalse`, I looked though the uses of `useConnections` -> `useServingConnections` and we do need all of them in [one](https://github.com/opendatahub-io/odh-dashboard/blob/3ef417a8178f79c3fd5615cbd110ef586dd764b7/packages/model-serving/src/components/deploymentWizard/fields/ModelLocationInputFields.tsx#L55) instance
```
  const [connections, connectionsLoaded] = useServingConnections(projectName, true, false);
```

dpanshug: @emilys314 i'm not getting the concern here. It's already gated behind includeDashboardFalse - when true, the || short-circuits and all connections (including hidden) pass through. 

The [deployment wizard](https://github.com/opendatahub-io/odh-dashboard/blob/3ef417a8178f79c3fd5615cbd110ef586dd764b7/packages/model-serving/src/components/deploymentWizard/fields/ModelLocationInputFields.tsx#L55) is the one caller that passes true to get all of them.

emilys314: You're right @dpanshug, when `includeDashboardFalse` is true, it will always include them. When it's false, it relies on the other boolean check.

It was a little confusing to follow with all the double negative names but this is good

emilys314: Okay so so the default of `includeDashboardFalse` is `false`. That means the default is to filter out hidden connections. 
When this is `true` we want all the connections (including the dashboard: false)

`true` will always return `true` so it gets them all
`false` will then depend on the annotation, only getting the enabled ones

emilys314 (approved):

PR #6850, opened 2026-03-25
Add the relevant MLflow links in the Runs page

nananosirova:
https://redhat.atlassian.net/browse/RHOAIENG-37533
## Description
https://github.com/user-attachments/assets/e1a01a92-2e97-480c-b9d2-d7422eafdcdc

- When the MLflow feature flag is enabled and pipeline runs have MLflow metadata (plugins_output), the Compare runs button on the pipeline runs page navigates to the MLflow compare runs page instead of the KFP compare runs page.
- When the MLflow feature flag is disabled, both links fall back to the existing KFP routes (no behavioral change)

Note: Currently behind the `mlflowPipelines` dev flag

## How Has This Been Tested?

This needs to be verified with a cluster with a KFP API server image that supports `plugins_input`/`plugins_output`
- Enable the `mlflow` feature flag (will later be `mlflowPipelines`)
- Navigate to the Runs page, and select some runs
- Click on the compare runs button, and verify it redirects to mlflow compare runs page
- Click on an experiment in the experiments column in the runs page

## Test Impact

Added mock tests.

## Request review criteria:

If you have UI changes: 

## Summary by CodeRabbit

* **New Features**
  * "Compare runs" now routes to MLflow's comparison view when MLflow is enabled and selected runs include MLflow identifiers.

* **Improvements**
  * Workspace query handling for comparison routes improved to append parameters correctly.

* **Tests**
  * Added end-to-end tests covering navigation for both MLflow-enabled and MLflow-disabled scenarios.

### DIFF
--- a/frontend/src/concepts/pipelines/content/tables/pipelineRun/PipelineRunTable.tsx
+++ b/frontend/src/concepts/pipelines/content/tables/pipelineRun/PipelineRunTable.tsx
@@ -5,7 +5,7 @@ import { Button, Skeleton, Tooltip } from '@patternfly/react-core';
 import { TableVariant, Td } from '@patternfly/react-table';
 import { ColumnsIcon } from '@patternfly/react-icons';
 
-import { TableBase, getTableColumnSort, useCheckboxTable } from '#~/components/table';
+import { getTableColumnSort, TableBase, useCheckboxTable } from '#~/components/table';
 import { ExperimentKF, PipelineRunKF, StorageStateKF } from '#~/concepts/pipelines/kfTypes';
 import { getPipelineRunColumns } from '#~/concepts/pipelines/content/tables/columns';
 import useIsMlflowPipelinesAvailable from '#~/concepts/mlflow/hooks/useIsMlflowPipelinesAvailable';
@@ -25,6 +25,7 @@ import SimpleMenuActions from '#~/components/SimpleMenuActions';
 import { ArchiveRunModal } from '#~/pages/pipelines/global/runs/ArchiveRunModal';
 import { RestoreRunModal } from '#~/pages/pipelines/global/runs/RestoreRunModal';
 import { compareRunsRoute, createRunRoute } from '#~/routes/pipelines/runs';
+import { mlflowCompareRunsRoute } from '#~/routes/pipelines/mlflow';
 import {
   ExperimentContext,
   useContextExperimentArchivedOrDeleted,
@@ -37,7 +38,7 @@ import useMlflowExperiments from '#~/concepts/mlflow/hooks/useMlflowExperiments'
 import { CustomMetricsColumnsModal } from './CustomMetricsColumnsModal';
 import { UnavailableMetricValue } from './UnavailableMetricValue';
 import { useMetricColumns } from './useMetricColumns';
-import { filterByMlflowExperiment } from './utils';
+import { filterByMlflowExperiment, getMlflowExperimentId, getMlflowRunId } from './utils';
 
 type PipelineRunTableProps = {
   runs: PipelineRunKF[];
@@ -101,15 +102,17 @@ const PipelineRunTable: React.FC<PipelineRunTableProps> = ({
   const [isArchiveModalOpen, setIsArchiveModalOpen] = React.useState(false);
   const [isRestoreModalOpen, setIsRestoreModalOpen] = React.useState(false);
   const [isCustomColModalOpen, setIsCustomColModalOpen] = React.useState(false);
-  const selectedRuns = selectedIds.reduce((acc: PipelineRunKF[], selectedId) => {
-    const selectedRun = runs.find((run) => run.run_id === selectedId);
-
-    if (selectedRun) {
-      acc.push(selectedRun);
-    }
-
-    return acc;
-  }, []);
+  const selectedRuns = React.useMemo(
+    () =>
+      selectedIds.reduce((acc: PipelineRunKF[], selectedId) => {
+        const selectedRun = runs.find((run) => run.run_id === selectedId);
+        if (selectedRun) {
+          acc.push(selectedRun);
+        }
+        return acc;
+      }, []),
+    [selectedIds, runs],
+  );
   const restoreButtonTooltipRef = React.useRef(null);
   const archivedExperiments = selectedRuns.reduce<ExperimentKF[]>((acc, selectedRun) => {
     const currentExperiment = allExperiments.find(
@@ -117,7 +120,6 @@ const PipelineRunTable: React.FC<PipelineRunTableProps> = ({
     );
 
     if (currentExperiment && currentExperiment.storage_state === StorageStateKF.ARCHIVED) {
-      // Create a Set to track unique experiment_id
       if (
         !acc.some(
           (selectedExperiment) =>
@@ -134,8 +136,31 @@ const PipelineRunTable: React.FC<PipelineRunTableProps> = ({
   const { isExperimentArchived: isContextExperimentArchived } =
     useContextExperimentArchivedOrDeleted();
   const createRunHref = createRunRoute(namespace, experiment?.experiment_id);
-  const compareRunsHref = compareRunsRoute(namespace, selectedIds, experiment?.experiment_id);
-  const isCompareDisabled = selectedIds.length === 0 || selectedIds.length > 10;
+  const hasSelectedRuns = selectedIds.length > 0;
+
+  const { compareRunsHref, isCompareDisabled } = React.useMemo(() => {
+    const validRunIds = selectedRuns.map(getMlflowRunId).filter((id): id is string => !!id);
+    const validExpIds = selectedRuns.map(getMlflowExperimentId).filter((id): id is string => !!id);
+    const allHaveMlflow =
+      hasSelectedRuns &&
+      validRunIds.length === selectedIds.length &&
+      validExpIds.length === selectedIds.length;
+    const href =
+      isMlflowAvailable && allHaveMlflow
+        ? mlflowCompareRunsRoute(namespace, validRunIds, [...new Set(validExpIds)])
+        : compareRunsRoute(namespace, selectedIds, experiment?.experiment_id);
+    return {
+      compareRunsHref: href,
+      isCompareDisabled: !hasSelectedRuns || selectedIds.length > 10,
+    };
+  }, [
+    selectedRuns,
+    selectedIds,
+    hasSelectedRuns,
+    isMlflowAvailable,
+    namespace,
+    experiment?.experiment_id,
+  ]);
   const primaryToolbarAction = React.useMemo(() => {
     if (runType === PipelineRunType.ARCHIVED) {
       return (
--- a/frontend/src/concepts/pipelines/content/tables/pipelineRun/PipelineRunTableRow.tsx
+++ b/frontend/src/concepts/pipelines/content/tables/pipelineRun/PipelineRunTableRow.tsx
@@ -21,6 +21,7 @@ import {
   globalArchivedPipelineRunsRoute,
   globalPipelineRunsRoute,
 } from '#~/routes/pipelines/runs';
+import { mlflowCompareRunsRoute } from '#~/routes/pipelines/mlflow';
 import { ArchiveRunModal } from '#~/pages/pipelines/global/runs/ArchiveRunModal';
 import PipelineRunTableRowExperiment from '#~/concepts/pipelines/content/tables/pipelineRun/PipelineRunTableRowExperiment';
 import PipelineRunTableRowMlflowExperiment from '#~/concepts/pipelines/content/tables/pipelineRun/PipelineRunTableRowMlflowExperiment';
@@ -38,7 +39,7 @@ import {
   FilterOptions,
   LEGACY_EXPERIMENT_FILTER_PARAM,
 } from '#~/concepts/pipelines/content/tables/usePipelineFilter';
-import { isPipelineRunRegistered } from './utils';
+import { getMlflowExperimentId, getMlflowRunId, isPipelineRunRegistered } from './utils';
 
 type PipelineRunTableRowProps = {
   checkboxProps: Omit<React.ComponentProps<typeof CheckboxTd>, 'id'>;
@@ -150,7 +151,13 @@ const PipelineRunTableRow: React.FC<PipelineRunTableRowProps> = ({
       {
         title: 'Compare runs',
         onClick: () => {
-          navigate(compareRunsRoute(namespace, [run.run_id], contextExperiment?.experiment_id));
+          const mlflowRunId = getMlflowRunId(run);
+          const mlflowExpId = getMlflowExperimentId(run);
+          if (mlflow?.isAvailable && mlflowRunId && mlflowExpId) {
+            navigate(mlflowCompareRunsRoute(namespace, [mlflowRunId], [mlflowExpId]));
+          } else {
+            navigate(compareRunsRoute(namespace, [run.run_id], contextExperiment?.experiment_id));
+          }
         },
       },
       ...(!version ? [] : [duplicateAction]),
@@ -162,10 +169,13 @@ const PipelineRunTableRow: React.FC<PipelineRunTableRowProps> = ({
         onClick: () => setIsArchiveModalOpen(true),
       },
     ];
+    // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [
     runType,
-    run.state,
     run.run_id,
+    run.state,
+    run.plugins_output,
+    run.plugins_input,
     version,
     navigate,
     namespace,
@@ -176,6 +186,7 @@ const PipelineRunTableRow: React.FC<PipelineRunTableRowProps> = ({
     api,
     refreshAllAPI,
     notification,
+    mlflow?.isAvailable,
   ]);
 
   return (
--- a/frontend/src/concepts/pipelines/content/tables/pipelineRun/PipelineRunTableRow.tsx
+++ b/frontend/src/concepts/pipelines/content/tables/pipelineRun/PipelineRunTableRow.tsx
@@ -169,6 +169,8 @@ const PipelineRunTableRow: React.FC<PipelineRunTableRowProps> = ({
         onClick: () => setIsArchiveModalOpen(true),
       },
     ];
+    // `run` is used in closures but only specific fields affect the result;
+    // listing them individually avoids re-creating actions on unrelated run changes.
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [
     runType,
--- a/frontend/src/concepts/pipelines/content/tables/pipelineRun/__tests__/utils.spec.ts
+++ b/frontend/src/concepts/pipelines/content/tables/pipelineRun/__tests__/utils.spec.ts
@@ -213,6 +213,18 @@ describe('getMlflowExperimentId', () => {
     });
     expect(getMlflowExperimentId(run)).toBe('7');
   });
+
+  it('returns experiment_id from plugins_output even when plugin state is FAILED (current behavior)', () => {
+    const run = buildMockRunKF({
+      plugins_output: {
+        mlflow: {
+          entries: { experiment_id: { value: 'failed-exp' } },
+          state: PluginStateKF.PLUGIN_FAILED,
+        },
+      },
+    });
+    expect(getMlflowExperimentId(run)).toBe('failed-exp');
+  });
 });
 
 describe('getMlflowRunId', () => {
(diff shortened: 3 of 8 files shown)


frontend/src/concepts/pipelines/content/tables/pipelineRun/PipelineRunTableRow.tsx
```diff
@@ -162,10 +169,13 @@ const PipelineRunTableRow: React.FC<PipelineRunTableRowProps> = ({
         onClick: () => setIsArchiveModalOpen(true),
       },
     ];
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
DaoDaoNoCode, line 174: This `eslint-disable` is new and the reason isn't immediately obvious to the next reader. The closure passes the full `run` object to `getMlflowRunId(run)` and `getMlflowExperimentId(run)`, but the deps only list the specific properties that matter (`run.plugins_output`, `run.plugins_input`). That's an intentional optimization to avoid re-computing actions on every render when unrelated `run` fields change.

nit: a short comment explaining *why* the disable is there would help future maintainers. Something like:

```suggestion
    // `run` is used in closures but only specific fields affect the result;
    // listing them individually avoids re-creating actions on unrelated run changes.
    // eslint-disable-next-line react-hooks/exhaustive-deps
```

DaoDaoNoCode (commented):
This PR adds MLflow-aware routing for the Compare Runs button and per-row kebab action. When MLflow is enabled and all selected runs have MLflow metadata, the user navigates to the MLflow embedded compare-runs page; otherwise it falls back to KFP. The utility functions, route helper, and `withWorkspace` fix are all clean. Good unit test coverage on the new utils and route function.

Local checks: lint (0 errors, 129 pre-existing warnings) -- type-check -- unit tests for the new mlflow.spec.ts pass; utils.spec.ts has a pre-existing `location is not defined` failure on main that is unrelated to this PR. CI is all green.

Prior discussion: nananosirova asked about UX for mixed MLflow/non-MLflow runs (silent fallback vs disabled button + tooltip). caponetto suggested an Alert banner. No resolution from UX yet -- the current silent fallback is a reasonable default while that decision is pending.

Left a few inline comments -- main things are the eslint-disable justification and a missing Cypress test for the per-row kebab action.

frontend/src/concepts/pipelines/content/tables/pipelineRun/PipelineRunTable.tsx
```diff
@@ -134,8 +136,31 @@ const PipelineRunTable: React.FC<PipelineRunTableProps> = ({
...
+        ? mlflowCompareRunsRoute(namespace, validRunIds, [...new Set(validExpIds)])
+        : compareRunsRoute(namespace, selectedIds, experiment?.experiment_id);
+    return {
+      compareRunsHref: href,
+      isCompareDisabled: !hasSelectedRuns || selectedIds.length > 10,
+    };
+  }, [
+    selectedRuns,
+    selectedIds,
+    hasSelectedRuns,
```
DaoDaoNoCode, line 159: nit: `hasSelectedRuns` is derived from `selectedIds.length > 0`, so it's redundant in the dependency array since `selectedIds` is already listed. Removing it would make the deps slightly cleaner, though it doesn't affect correctness.

frontend/src/concepts/pipelines/content/tables/pipelineRun/PipelineRunTableRow.tsx
```diff
@@ -150,7 +151,13 @@ const PipelineRunTableRow: React.FC<PipelineRunTableRowProps> = ({
       {
         title: 'Compare runs',
         onClick: () => {
```
DaoDaoNoCode, line 153: The per-row kebab "Compare runs" action now has non-trivial branching logic (MLflow vs KFP routing), but I don't see a corresponding Cypress mock test for it. The new Cypress tests in `pipelineRuns.cy.ts` only cover the toolbar button. It would be good to add a test that opens the kebab on a single row and verifies the "Compare runs" action navigates to the MLflow compare-runs page when MLflow metadata is present, and falls back to KFP when it isn't.

frontend/src/concepts/pipelines/content/tables/pipelineRun/PipelineRunTable.tsx
```diff
@@ -134,8 +136,31 @@ const PipelineRunTable: React.FC<PipelineRunTableProps> = ({
   const { isExperimentArchived: isContextExperimentArchived } =
     useContextExperimentArchivedOrDeleted();
   const createRunHref = createRunRoute(namespace, experiment?.experiment_id);
-  const compareRunsHref = compareRunsRoute(namespace, selectedIds, experiment?.experiment_id);
-  const isCompareDisabled = selectedIds.length === 0 || selectedIds.length > 10;
+  const hasSelectedRuns = selectedIds.length > 0;
+
+  const { compareRunsHref, isCompareDisabled } = React.useMemo(() => {
```
DaoDaoNoCode, line 141: Nice improvement wrapping `selectedRuns` and the compare route computation in `useMemo`. One observation: the `allHaveMlflow` check at line 144 compares array lengths against `selectedIds.length`, which works correctly. Just want to confirm: is `selectedRuns.length` always guaranteed to equal `selectedIds.length`? If a selected ID doesn't match any run (e.g., the runs list refreshed while checkboxes were still checked), `selectedRuns` could be shorter than `selectedIds`, and `allHaveMlflow` would be false even if all *available* selected runs have MLflow data. That seems like the safer behavior (fall back to KFP), so this is fine -- just flagging it as an edge case to be aware of.

frontend/src/routes/pipelines/mlflow.ts
```diff
@@ -18,7 +18,8 @@ const withWorkspace = (basePath: string, namespace?: string): string => {
   if (!namespace) {
     return basePath;
   }
-  return `${basePath}?${WORKSPACE_QUERY_PARAM}=${encodeURIComponent(namespace)}`;
+  const separator = basePath.includes('?') ? '&' : '?';
```
DaoDaoNoCode, line 21: Good fix. The old version would produce `?runs=...?workspace=ns` when the basePath already had query params. Using `includes('?')` to choose the separator is the right approach.

frontend/src/concepts/pipelines/content/tables/pipelineRun/PipelineRunTable.tsx
nananosirova: `hasSelectedRuns` seems to be required by `react-hooks/exhaustive-deps` since it's referenced inside the memo closure

frontend/src/concepts/pipelines/content/tables/pipelineRun/PipelineRunTableRow.tsx
nananosirova: Committed!

frontend/src/concepts/pipelines/content/tables/pipelineRun/PipelineRunTableRow.tsx
nananosirova: Added the tests for this case, thank you

bobbravo2 (commented):
@caponetto I like the fallback suggestion of an alert banner. @acrago we need a UX decision, otherwise this will need to be merged, and a follow up story logged in Jira. If the suggested fallback sounds good, please let us know. thanks!

acrago:
@bobbravo2, @nananosirova, @caponetto - The current implementation can move forward with the proposed solution! I want to flag that UX was not able to fully complete the design review/exploration for this flow due to an engineering blocker. I let @nananosirova know a couple weeks ago, but I know there's a time crunch to get this out. Just wanted to note that there may need to be future refinement!

Gkrumbach07:

frontend/src/concepts/pipelines/content/tables/pipelineRun/PipelineRunTable.tsx
```diff
@@ -134,8 +136,31 @@ const PipelineRunTable: React.FC<PipelineRunTableProps> = ({
...
+    const validRunIds = selectedRuns.map(getMlflowRunId).filter((id): id is string => !!id);
+    const validExpIds = selectedRuns.map(getMlflowExperimentId).filter((id): id is string => !!id);
+    const allHaveMlflow =
```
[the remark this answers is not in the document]
nananosirova: This is intentional. Comparing against `selectedIds.length` means that if any selected run can't be resolved (for example the list refreshed while checkboxes were still checked), we fall back to KFP rather than silently comparing a partial set in MLflow. If we compared against `selectedRuns.length`, we'd navigate to MLflow with fewer runs than the user selected, which would be confusing. @DaoDaoNoCode already reviewed this edge case and confirmed the current approach is the safer behavior.

frontend/src/concepts/pipelines/content/tables/pipelineRun/utils.ts
```diff
@@ -36,3 +36,25 @@ export const isPipelineRunRegistered = (artifact: Artifact[]): boolean => {
...
+    return String(outputId);
+  }
+  const inputId = run.plugins_input?.mlflow?.experiment_id;
+  if (inputId != null) {
+    return String(inputId);
+  }
+  return undefined;
+};
+
+export const getMlflowRunId = (run: PipelineRunKF): string | undefined => {
```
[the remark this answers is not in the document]
nananosirova: The asymmetry is intentional since `root_run_id` only exists after the backend creates the MLflow run (there's nothing in `plugins_input` to fall back to), but I added a comment to make that clear.

Gkrumbach07:

DaoDaoNoCode (commented):
Full re-review including all prior comment threads. Prior review items from DaoDaoNoCode (eslint-disable comment, kebab tests, hasSelectedRuns dep) and odh-dashboard-agent (asymmetry documentation) are addressed in commits 2-4. See inline comments for outstanding and new findings.

frontend/src/concepts/pipelines/content/tables/pipelineRun/utils.ts
```diff
@@ -36,3 +36,27 @@ export const isPipelineRunRegistered = (artifact: Artifact[]): boolean => {
...
+// root_run_id is only populated in plugins_output after the backend creates the MLflow run.
+// Unlike experiment_id, there is no plugins_input equivalent to fall back to.
+export const getMlflowRunId = (run: PipelineRunKF): string | undefined => {
+  const outputEntry = run.plugins_output?.mlflow?.entries;
+  const outputId = outputEntry?.root_run_id?.value;
+  if (outputId != null) {
+    return String(outputId);
+  }
+  return undefined;
+};
```
DaoDaoNoCode, line 62: **Outstanding: `getMlflowRunId` / `getMlflowExperimentId` read `plugins_output.entries` without checking `state`**

The prior agent review flagged this as Major but got no reply. After checking the codebase, I can confirm this is a **real gap but a codebase-wide pattern** — `PipelineRunTableRowMlflowExperiment.tsx` also reads `plugins_output.mlflow.entries.experiment_id.value` without checking `state`. So this PR is consistent with existing behavior.

However, the risk is real: if `state === PLUGIN_FAILED` but `root_run_id` is partially populated, the Compare button routes to MLflow for a broken tracking run. The fix is straightforward:
```ts
const output = run.plugins_output?.mlflow;
if (output?.state !== PluginStateKF.PLUGIN_SUCCEEDED) {
  return undefined;
}
return output.entries?.root_run_id?.value != null ? String(output.entries.root_run_id.value) : undefined;
```

**Recommendation**: address as a follow-up (since the existing column code has the same gap), but the test gap should be closed now — add a test with `state: PLUGIN_FAILED` + `root_run_id` present to document the current behavior.

frontend/src/concepts/pipelines/content/tables/pipelineRun/__tests__/utils.spec.ts
```diff
@@ -161,3 +163,92 @@ describe('filterByMlflowExperiment', () => {
...
+          entries: { experiment_id: { value: 7 } },
+          state: PluginStateKF.PLUGIN_SUCCEEDED,
+        },
+      },
+    });
+    expect(getMlflowExperimentId(run)).toBe('7');
+  });
+});
+
+describe('getMlflowRunId', () => {
```
DaoDaoNoCode, line 230: **Outstanding (minor): test gap for non-SUCCEEDED plugin state**

All `getMlflowRunId` and `getMlflowExperimentId` tests use `PluginStateKF.PLUGIN_SUCCEEDED`. Whether the functions should return values when `state === PLUGIN_FAILED` is a design decision (see comment on `utils.ts`), but either way the current behavior should be locked in with a test:

```ts
it("returns root_run_id even when plugin state is FAILED (current behavior)", () => {
  const run = buildMockRunKF({
    plugins_output: {
      mlflow: {
        entries: { root_run_id: { value: "partial-id" } },
        state: PluginStateKF.PLUGIN_FAILED,
      },
    },
  });
  // Current behavior: returns the value regardless of state
  expect(getMlflowRunId(run)).toBe("partial-id");
});
```

This makes the implicit contract explicit and will catch it if a state guard is added later.

frontend/src/concepts/pipelines/content/tables/pipelineRun/utils.ts
nananosirova: Created a followup ticket: https://redhat.atlassian.net/browse/RHOAIENG-65491
and added the test!

frontend/src/concepts/pipelines/content/tables/pipelineRun/__tests__/utils.spec.ts
nananosirova: Created the test, and have a followup ticket: https://redhat.atlassian.net/browse/RHOAIENG-65491

nananosirova:

nananosirova:

DaoDaoNoCode (commented):
Re-review after commit `02d3eb19` ("Address comments"). All prior outstanding items from my May 29 review are resolved — see inline thread-by-thread status.

frontend/src/concepts/pipelines/content/tables/pipelineRun/utils.ts
DaoDaoNoCode: Looks good. PLUGIN_FAILED tests are in the latest diff and RHOAIENG-65491 covers the state-guard addition. Both outstanding items from my May 29 review are resolved.

frontend/src/concepts/pipelines/content/tables/pipelineRun/__tests__/utils.spec.ts
DaoDaoNoCode: Confirmed resolved — both `getMlflowRunId` and `getMlflowExperimentId` now have a PLUGIN_FAILED test case that documents the current behavior.

packages/cypress/cypress/tests/mocked/pipelines/runs/pipelineRuns.cy.ts
```diff
@@ -449,6 +479,129 @@ describe('Pipeline runs', () => {
...
+        it('compare runs button navigates to MLflow when dev flag is enabled', () => {
+          cy.interceptOdh('GET /api/config', mockDashboardConfig({ mlflowPipelines: true }));
+          interceptMlflowStatus();
+          interceptDSPAMlflowIntegration(projectName);
+          pipelineRunsGlobal.visit(projectName, 'active');
+
+          activeRunsTable.getRowByName(mockActiveRuns[0].display_name).findCheckbox().click();
+          activeRunsTable.getRowByName(mockActiveRuns[1].display_name).findCheckbox().click();
+
+          pipelineRunsGlobal.findCompareRunsButton().click();
```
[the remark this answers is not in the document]
DaoDaoNoCode: Confirmed — this is a real flake risk. The other three MLflow tests in this section (`mixed metadata`, `per-row kebab MLflow`, `per-row kebab KFP fallback`) all have `cy.wait("@mlflowStatus")` after `visit()` and before interacting with the table. This test skips it. Without the wait, `isMlflowAvailable` may still be `false` when the checkboxes are clicked, so the compare button holds the KFP href and `verifyRelativeURL` fails intermittently. Please add `cy.wait("@mlflowStatus")` after line 487 to match the pattern.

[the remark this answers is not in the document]
DaoDaoNoCode: @nananosirova ^

frontend/src/concepts/pipelines/content/tables/pipelineRun/PipelineRunTable.tsx
```diff
@@ -25,6 +25,13 @@ import SimpleMenuActions from '#~/components/SimpleMenuActions';
...
 import { RestoreRunModal } from '#~/pages/pipelines/global/runs/RestoreRunModal';
 import { compareRunsRoute, createRunRoute } from '#~/routes/pipelines/runs';
+import {
+  getMlflowExperimentId,
+  getMlflowRunId,
+  mlflowCompareRunsRoute,
+} from '#~/routes/pipelines/mlflow';
+import useIsAreaAvailable from '#~/concepts/areas/useIsAreaAvailable';
+import { SupportedArea } from '#~/concepts/areas';
 import {
```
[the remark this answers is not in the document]
DaoDaoNoCode: Outdated — this was on an earlier version of the code.

frontend/src/concepts/pipelines/content/tables/pipelineRun/PipelineRunTableRow.tsx
DaoDaoNoCode: Tests are in the latest diff — two new Cypress tests cover the per-row kebab for both MLflow and KFP fallback. This one is done.

frontend/src/concepts/pipelines/content/tables/pipelineRun/PipelineRunTable.tsx
DaoDaoNoCode: Yep, the current behavior is intentional — author confirmed comparing against `selectedIds.length` is the safer fallback (avoids navigating to MLflow with fewer runs than selected). No action needed.

frontend/src/routes/pipelines/mlflow.ts
DaoDaoNoCode: Agreed, clean fix. Nothing to do here.

frontend/src/concepts/pipelines/content/tables/pipelineRun/PipelineRunTable.tsx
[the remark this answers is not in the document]
DaoDaoNoCode: Already resolved — author acknowledged the intentional behavior, and DaoDaoNoCode confirmed the existing approach (compare against selectedIds.length) is the safer fallback. Follow-up state guard is tracked in RHOAIENG-65491.

packages/cypress/cypress/tests/mocked/pipelines/runs/pipelineRuns.cy.ts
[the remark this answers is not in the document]
DaoDaoNoCode: @nananosirova DaoDaoNoCode already flagged this one — just adding a +1 that this is a real flake risk. The other MLflow tests in this section all have `cy.wait('@mlflowStatus')` after visit, this test is the only one missing it. Should be a one-liner fix.

frontend/src/concepts/pipelines/content/tables/pipelineRun/__tests__/utils.spec.ts
```diff
@@ -161,3 +163,116 @@ describe('filterByMlflowExperiment', () => {
     expect(result[0].recurring_run_id).toBe('sa');
   });
 });
+
+describe('getMlflowExperimentId', () => {
```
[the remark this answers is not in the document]
DaoDaoNoCode: Nice to have but not blocking — the downstream `.filter(id => !!id)` in PipelineRunTable correctly excludes empty strings, so the behavior is safe. Can be a follow-up if the team wants to lock that in.

frontend/src/routes/pipelines/__tests__/mlflow.spec.ts
```diff
@@ -0,0 +1,30 @@
+import { mlflowCompareRunsRoute } from '#~/routes/pipelines/mlflow';
+
+describe('mlflowCompareRunsRoute', () => {
+  it('builds a compare URL with runs, experiments, and workspace', () => {
```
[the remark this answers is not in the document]
DaoDaoNoCode: Not blocking — the test is correct for the current implementation and would catch regressions. Asserting individual params is a nice pattern but overkill for this case.

packages/cypress/cypress/tests/mocked/pipelines/runs/pipelineRuns.cy.ts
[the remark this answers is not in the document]
DaoDaoNoCode: Scratch my earlier comment — this is already fixed in commit 212b67c ("Add cy.wait for mlflowStatus to prevent flake"). The `cy.wait('@mlflowStatus')` is present in the current diff. Thread can be closed.

DaoDaoNoCode:
@nananosirova Looks like there is still a test fail that you need to fix before merging

DaoDaoNoCode (approved):
/lgtm

PR #6884, opened 2026-03-25
Handle legacy visibility annotation values in hardware profiles

nananosirova:
https://issues.redhat.com/browse/RHOAIENG-37048

## Description

Hardware profiles created in RHOAI 2.x with DSP (pipelines) visibility have the annotation `opendatahub.io/dashboard-feature-visibility: '["pipelines"]'`. After the `PIPELINES` enum value was removed from `HardwareProfileFeatureVisibility` ([RHOAIENG-34215](https://redhat.atlassian.net/browse/RHOAIENG-34215)), these profiles have no visibility in the 3.x UI:

- The visibility column shows `-` instead of a meaningful label
image

- The edit modal shows "Limited visibility" with an empty dropdown
image

- The profile cannot be selected for workbenches or model deployment

Root cause: Multiple components parsed the `opendatahub.io/dashboard-feature-visibility` annotation and used the raw values without validating them against the current enum. When `"pipelines"` is no longer a recognized value, the UI treats these profiles as having empty/no visibility.

Fix: Added inline filtering of annotation values against the current `HardwareProfileFeatureVisibility` enum in all locations that read the annotation:
- `useHardwareProfilesByFeatureVisibility`: filters parsed values to only recognized enum members before checking workbench/model-serving eligibility; treats profiles with no recognized values as "visible everywhere"
- `ManageHardwareProfile`: filters parsed values when prefilling the edit modal, so legacy values like `"pipelines"` are silently dropped
- `HardwareProfilesTableRow`: displays _All features_ (italicized) instead of `-` when no recognized visibility values remain

This ensures that profiles with only legacy annotation values (e.g. `["pipelines"]`) become "visible everywhere" rather than unusable after an upgrade from 2.x to 3.0.

## How Has This Been Tested?

Manual testing:
1. Create a hardware profile with the annotation `opendatahub.io/dashboard-feature-visibility: '["pipelines"]'` (simulating a 2.x profile)
2. Navigate to Settings > Hardware profiles: verify the visibility column shows "All features" instead of "-"
3. Click Edit on that profile: verify the visibility section shows "All features" (not "Limited visibility" with an empty dropdown)
4. Navigate to create a workbench: verify the profile appears in the hardware profile selector
5. Navigate to deploy a model: verify the profile appears in the hardware profile selector
6. Create a profile with `'["pipelines", "model-serving"]'`:  verify it shows "Model serving" in the table and is only available for model serving

## Request review criteria:

## Summary by CodeRabbit

* **Bug Fixes**
  * Show "All features" (italicized) instead of a placeholder when feature visibility is empty.
  * Only recognize and validate supported feature-visibility values when determining profile visibility.
  * Ensure consistent filtering behavior for empty, legacy, and mixed visibility annotations.

* **Tests**
  * Added tests for legacy and mixed visibility annotation scenarios to verify correct filtering.

### DIFF
--- a/packages/cypress/cypress/tests/mocked/hardwareProfiles/hardwareProfiles.cy.ts
+++ b/packages/cypress/cypress/tests/mocked/hardwareProfiles/hardwareProfiles.cy.ts
@@ -794,5 +794,6 @@ describe('hardware profiles - legacy feature visibility annotations', () => {
 
     const mixedRow = hardwareProfile.getRow('Mixed Profile');
     mixedRow.findFeatureLabel('model-serving').should('exist');
+    mixedRow.findAllFeaturesText().should('not.exist');
   });
 });
--- a/packages/cypress/cypress/tests/mocked/hardwareProfiles/hardwareProfiles.cy.ts
+++ b/packages/cypress/cypress/tests/mocked/hardwareProfiles/hardwareProfiles.cy.ts
@@ -751,3 +751,48 @@ describe('hardware profiles - ordering', () => {
     hardwareProfile.findRows().eq(3).should('contain', 'Alpha One Profile');
   });
 });
+
+describe('hardware profiles - legacy feature visibility annotations', () => {
+  beforeEach(() => {
+    asProductAdminUser();
+  });
+
+  it('should show "All features" for a profile whose only visibility values are legacy', () => {
+    cy.interceptOdh(
+      'GET /api/config',
+      mockDashboardConfig({
+        modelServerSizes: [],
+        notebookSizes: [],
+      }),
+    );
+    cy.interceptK8sList(
+      { model: HardwareProfileModel, ns: 'opendatahub' },
+      mockK8sResourceList([
+        mockHardwareProfile({
+          name: 'legacy-profile',
+          displayName: 'Legacy Profile',
+          annotations: {
+            'opendatahub.io/dashboard-feature-visibility': JSON.stringify(['pipelines']),
+          },
+        }),
+        mockHardwareProfile({
+          name: 'mixed-profile',
+          displayName: 'Mixed Profile',
+          annotations: {
+            'opendatahub.io/dashboard-feature-visibility': JSON.stringify([
+              'pipelines',
+              'model-serving',
+            ]),
+          },
+        }),
+      ]),
+    );
+    hardwareProfile.visit();
+
+    const legacyRow = hardwareProfile.getRow('Legacy Profile');
+    legacyRow.findAllFeaturesText().should('have.text', 'All features');
+
+    const mixedRow = hardwareProfile.getRow('Mixed Profile');
+    mixedRow.findFeatureLabel('model-serving').should('exist');
+  });
+});
--- a/frontend/src/pages/hardwareProfiles/HardwareProfilesTableRow.tsx
+++ b/frontend/src/pages/hardwareProfiles/HardwareProfilesTableRow.tsx
@@ -26,6 +26,7 @@ import TolerationTable from '#~/pages/hardwareProfiles/toleration/TolerationTabl
 import { useKebabAccessAllowed, verbModelAccess } from '#~/concepts/userSSAR';
 import {
   createHardwareProfileWarningTitle,
+  filterRecognizedVisibility,
   getHardwareProfileDescription,
   getHardwareProfileDisplayName,
   isDefaultHardwareProfile,
@@ -58,9 +59,10 @@ const HardwareProfilesTableRow: React.FC<HardwareProfilesTableRowProps> = ({
   const useCases: HardwareProfileFeatureVisibility[] = React.useMemo(() => {
     if (hardwareProfile.metadata.annotations?.['opendatahub.io/dashboard-feature-visibility']) {
       try {
-        return JSON.parse(
+        const raw: string[] = JSON.parse(
           hardwareProfile.metadata.annotations['opendatahub.io/dashboard-feature-visibility'],
         );
+        return filterRecognizedVisibility(raw);
       } catch (error) {
         return [];
       }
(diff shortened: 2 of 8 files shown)


DaoDaoNoCode (commented):
## Code Review

The fix is logically correct and the root cause is well understood. A few things worth addressing before merge:

---

### 1. Recognition filter is copy-pasted in two places

This block appears identically in both `useHardwareProfilesByFeatureVisibility.ts` and `ManageHardwareProfile.tsx`:

```ts
visibleIn.filter((v) =>
  Object.values(HardwareProfileFeatureVisibility).some(
    (ev: HardwareProfileFeatureVisibility) => ev === v,
  ),
)
```

CodeRabbit already flagged this. Consider extracting it into `utils.ts`:

```ts
export const getRecognizedVisibility = (
  raw: string | undefined,
): HardwareProfileFeatureVisibility[] => {
  if (!raw) return [];
  try {
    const parsed: string[] = JSON.parse(raw);
    const valid = Object.values(HardwareProfileFeatureVisibility) as string[];
    return parsed.filter((v): v is HardwareProfileFeatureVisibility => valid.includes(v));
  } catch {
    return [];
  }
};
```

Then both call sites become a single `getRecognizedVisibility(annotation)` call and the error handling is consolidated too.

---

### 2. Two different validators for "is this value recognized?"

- `HardwareProfilesTableRow` filters via `HardwareProfileFeatureVisibilityTitles[v]` (titles record lookup)
- `filterHardwareProfileByFeatureVisibility` and `ManageHardwareProfile` filter via `Object.values(HardwareProfileFeatureVisibility)`

Both are correct today, but they diverge if a new enum value is added without a matching entry in `HardwareProfileFeatureVisibilityTitles`. A shared utility (point 1) eliminates this risk.

---

### 3. Missing test coverage on the changed UI paths

Codecov flags 3 uncovered lines:

- **`ManageHardwareProfile.tsx` (50%)** — no test for the prefill path when the annotation contains only legacy values like `["pipelines"]`. The `catch` block (`setVisibility([])`) is also untested.
- **`HardwareProfilesTableRow.tsx` (0%)** — no test for the `<i>All features</i>` branch when `validUseCases.length === 0` (i.e., annotation present but all values are legacy).

---

### 4. Nit: explicit type annotation on `ev` is redundant

```ts
// TypeScript already infers ev as HardwareProfileFeatureVisibility from Object.values(...)
(ev: HardwareProfileFeatureVisibility) => ev === v
```

This is minor, but removing it (or using `valid.includes(v)` as in the utility above) reads more cleanly.

DaoDaoNoCode (commented):
## Code Review (updated — reviewing the latest force-push)

All four issues from the previous review have been addressed. Here's what changed and what's left.

---

### What was fixed

**DRY violation → resolved.**
`filterRecognizedVisibility(visibleIn: string[]): HardwareProfileFeatureVisibility[]` is now a shared utility in `utils.ts`. All three call sites (`HardwareProfilesTableRow`, `ManageHardwareProfile`, `filterHardwareProfileByFeatureVisibility`) use it.

**Two sources of truth → resolved.**
`HardwareProfilesTableRow` now applies `filterRecognizedVisibility` at parse time (inside the `useMemo`), so `useCases` only ever contains known enum values. The secondary `validUseCases` filter that used `HardwareProfileFeatureVisibilityTitles` is gone. `HardwareProfileFeatureVisibilityTitles` is now used only for display (mapping enum → label text), not for validation — which is the right separation.

**Redundant type annotation → resolved.**
The utility uses a proper type predicate `(v): v is HardwareProfileFeatureVisibility` instead of an explicit `ev: HardwareProfileFeatureVisibility` annotation on a `.some()` callback.

**Missing utility tests → resolved.**
`utils.spec.ts` now has 4 tests for `filterRecognizedVisibility` covering: all-valid, all-unrecognized, mixed, and empty input.

---

### What's still missing

**No test for `ManageHardwareProfile` prefill with a legacy annotation.**
The `try/catch` block in `prefillFormData` and the `else` path (`setVisibility([])`) are still untested at the component level. The risk is lower now that `filterRecognizedVisibility` itself is well-tested, but a test for the happy path (prefill strips `"pipelines"` and leaves the form in the right state) would close the coverage gap Codecov flagged.

**No test for `HardwareProfilesTableRow` rendering `<i>All features</i>` for legacy-only annotations.**
The branch where `filterRecognizedVisibility` returns `[]` and the row renders `<i>All features</i>` is still untested.

---

### Summary

The implementation is clean and correct. The shared utility is well-designed and solves the original bug at the right layer. The two remaining gaps are both in component-level test coverage — the core logic is solid. LGTM with those tests as a follow-up if not done inline.

packages/cypress/cypress/tests/mocked/hardwareProfiles/hardwareProfiles.cy.ts
```diff
@@ -751,3 +751,48 @@ describe('hardware profiles - ordering', () => {
...
+        }),
+      ]),
+    );
+    hardwareProfile.visit();
+
+    const legacyRow = hardwareProfile.getRow('Legacy Profile');
+    legacyRow.findAllFeaturesText().should('have.text', 'All features');
+
+    const mixedRow = hardwareProfile.getRow('Mixed Profile');
+    mixedRow.findFeatureLabel('model-serving').should('exist');
```
DaoDaoNoCode, line 796: nit: This verifies the mixed profile shows the `model-serving` label, but doesn't confirm the legacy `pipelines` value was actually stripped. Adding one more assertion would make this airtight:

```suggestion
    mixedRow.findFeatureLabel('model-serving').should('exist');
    mixedRow.findAllFeaturesText().should('not.exist');
```

Not blocking -- the unit tests already cover the filtering logic.

DaoDaoNoCode (commented):
Re-reviewed after the two latest commits (Apr 24). All four issues from DaoDaoNoCode's first review are resolved: the DRY violation is gone with the shared `filterRecognizedVisibility` utility, the dual validation source is eliminated, the type predicate is clean, and `utils.spec.ts` has solid coverage for the new function. The Cypress mock test covers the table row rendering path for both legacy-only and mixed visibility annotations. One nit inline on the Cypress test, nothing blocking.

Local checks: type-check pass, unit tests pass (219/219 in hardwareProfile suite), CI all green.

DaoDaoNoCode (commented):
Re-reviewed after the latest commit. The `findAllFeaturesText().should('not.exist')` assertion is now in the Cypress test — that was the last open nit from my previous pass.

Local checks: type-check ✅  unit tests ✅ (219/219 in the hardwareProfile suite).

The `Cypress-Mock-Tests (pipelines, ...)` job shows a failure in CI, but this PR doesn't touch any pipelines code — looks like a pre-existing flake. Worth `/retest`ing before merge.

Everything from the previous review rounds is resolved. LGTM.

DaoDaoNoCode (approved):
/lgtm

DaoDaoNoCode:
Waiting for the tests to pass

DaoDaoNoCode (approved):
/lgtm

PR #6892, opened 2026-03-26
Fix legacy HWP tolerations not injected into ISVCs

nananosirova:
https://issues.redhat.com/browse/RHOAIENG-50153

## Description

The dashboard controller isn't injecting tolerations from legacy Hardware Profiles (HWPs) into ISVCs in RHOAI 2.25. When creating an ISVC using a legacy HWP, the tolerations are not applied to the ISVC predictor spec.

Root cause: In `assembleInferenceService()`, the condition `if (!podSpecOptions.selectedHardwareProfile)` skips injecting tolerations whenever any hardware profile is selected. This is correct for native HWPs (where the backend controller handles injection via the `opendatahub.io/hardware-profile-name` annotation), but wrong for legacy HWPs; there is no backend controller that injects tolerations for legacy profiles.

Fix: Added a legacy HWP detection check based on the absence of `metadata.uid`. Legacy HWPs are always in-memory objects (created by `useMigratedHardwareProfiles`), never persisted to the K8s API server, so they never have a `uid`. When a legacy HWP is detected, tolerations are injected directly into the ISVC predictor spec.

## How Has This Been Tested?

Manual testing:
1. Ensure an AcceleratorProfile exists on the cluster (this will appear as a legacy HWP in the dashboard)
2. Go to Model Serving and deploy a model using KServe single-model serving
3. Select the legacy hardware profile
4. Deploy the model
5. Verify the resulting ISVC has the expected tolerations in `spec.predictor.tolerations` by running:
   ```
   oc get inferenceservice <name> -n <namespace> -o jsonpath='{.spec.predictor.tolerations}'
   ```

## Test Impact

- 1 new unit test added covering the legacy HWP toleration injection scenario

## Request review criteria:

### DIFF
--- a/frontend/src/api/k8s/__tests__/inferenceServices.spec.ts
+++ b/frontend/src/api/k8s/__tests__/inferenceServices.spec.ts
@@ -602,6 +602,8 @@ describe('assembleInferenceService', () => {
         },
       ],
     });
+    // mockHardwareProfile uses a destructuring default for uid, so passing
+    // undefined doesn't clear it. We must delete it explicitly.
     delete (legacyHardwareProfile.metadata as Record<string, unknown>).uid;
 
     const podSpecOptions = mockModelServingPodSpecOptions({
--- a/frontend/src/api/k8s/__tests__/inferenceServices.spec.ts
+++ b/frontend/src/api/k8s/__tests__/inferenceServices.spec.ts
@@ -590,6 +590,45 @@ describe('assembleInferenceService', () => {
     expect(resultKServe.spec.predictor.nodeSelector).toBeUndefined();
   });
 
+  it('should inject tolerations for legacy hardware profiles', () => {
+    const legacyHardwareProfile = mockHardwareProfile({
+      name: 'legacy-hwp',
+      uid: undefined,
+      tolerations: [
+        {
+          key: 'nvidia.com/gpu',
+          operator: TolerationOperator.EXISTS,
+          effect: TolerationEffect.NO_SCHEDULE,
+        },
+      ],
+    });
+    delete (legacyHardwareProfile.metadata as Record<string, unknown>).uid;
+
+    const podSpecOptions = mockModelServingPodSpecOptions({
+      selectedHardwareProfile: legacyHardwareProfile,
+      tolerations: legacyHardwareProfile.spec.scheduling?.node?.tolerations,
+      nodeSelector: undefined,
+      resources: {
+        requests: { cpu: '1', memory: '2Gi' },
+        limits: { cpu: '2', memory: '4Gi' },
+      },
+    });
+
+    const result = assembleInferenceService(
+      mockInferenceServiceModalData({}),
+      undefined,
+      undefined,
+      false,
+      undefined,
+      undefined,
+      podSpecOptions,
+    );
+
+    expect(result.spec.predictor.tolerations).toEqual(
+      legacyHardwareProfile.spec.scheduling?.node?.tolerations,
+    );
+  });
+
   it('should set pod specs for accelerator profiles', () => {
     const gpuTolerations = [
       {
--- a/frontend/src/api/k8s/__tests__/inferenceServices.spec.ts
+++ b/frontend/src/api/k8s/__tests__/inferenceServices.spec.ts
@@ -607,7 +607,7 @@ describe('assembleInferenceService', () => {
     const podSpecOptions = mockModelServingPodSpecOptions({
       selectedHardwareProfile: legacyHardwareProfile,
       tolerations: legacyHardwareProfile.spec.scheduling?.node?.tolerations,
-      nodeSelector: undefined,
+      nodeSelector: { 'nvidia.com/gpu.present': 'true' },
       resources: {
         requests: { cpu: '1', memory: '2Gi' },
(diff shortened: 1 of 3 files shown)


frontend/src/api/k8s/inferenceServices.ts
```diff
@@ -230,7 +230,10 @@ export const assembleInferenceService = (
 
   if (!isModelMesh && podSpecOptions) {
     const { tolerations, resources, nodeSelector } = podSpecOptions;
-    if (!podSpecOptions.selectedHardwareProfile) {
+    const isLegacyHardwareProfile =
+      podSpecOptions.selectedHardwareProfile &&
+      !podSpecOptions.selectedHardwareProfile.metadata.uid;
```
DaoDaoNoCode: This definition of `isLegacyHardwareProfile` only checks `!metadata.uid`, but the other one up at line 151-153 uses `!!selectedAcceleratorProfile || !metadata.uid`. In practice they produce the same result because the runtime hook (`useModelServingPodSpecOptionsState`) never sets both `selectedAcceleratorProfile` and `selectedHardwareProfile` at the same time. But it's confusing for future readers and fragile if that invariant ever changes.

Worth aligning the two -- either extract a shared helper or at least use the same condition in both places.

nit: not blocking since it's correct today, but would prevent a subtle future bug.

frontend/src/api/k8s/__tests__/inferenceServices.spec.ts
```diff
@@ -590,6 +590,45 @@ describe('assembleInferenceService', () => {
...
+      uid: undefined,
+      tolerations: [
+        {
+          key: 'nvidia.com/gpu',
+          operator: TolerationOperator.EXISTS,
+          effect: TolerationEffect.NO_SCHEDULE,
+        },
+      ],
+    });
+    delete (legacyHardwareProfile.metadata as Record<string, unknown>).uid;
```
DaoDaoNoCode, line 607: The `uid: undefined` passed to `mockHardwareProfile` on line 596 doesn't actually clear the uid because destructuring defaults kick in (`uid = genUID('service')`). The explicit `delete` here is the real mechanism that removes it. This works, but it's a bit fragile -- if someone sees `uid: undefined` above and thinks the `delete` is redundant, removing it would silently break the test.

nit: consider adding a brief comment explaining why the `delete` is necessary, e.g.:
```suggestion
    // mockHardwareProfile uses a destructuring default for uid, so passing
    // undefined doesn't clear it. We must delete it explicitly.
    delete (legacyHardwareProfile.metadata as Record<string, unknown>).uid;
```

frontend/src/api/k8s/__tests__/inferenceServices.spec.ts
```diff
@@ -590,6 +590,45 @@ describe('assembleInferenceService', () => {
...
+      undefined,
+      false,
+      undefined,
+      undefined,
+      podSpecOptions,
+    );
+
+    expect(result.spec.predictor.tolerations).toEqual(
+      legacyHardwareProfile.spec.scheduling?.node?.tolerations,
+    );
```
DaoDaoNoCode, line 631: The test verifies tolerations are injected for legacy HWPs, which is the core fix. Two additional things worth covering:

1. **nodeSelector injection** -- the test passes `nodeSelector: undefined`, so the nodeSelector path in the fix (line 240-242 in the source) is never exercised for the legacy case. Adding a non-undefined `nodeSelector` and asserting it appears on the result would cover the full fix.

2. **`legacy-hardware-profile-name` annotation** -- the annotation block (lines 158-161 in the source) sets `opendatahub.io/legacy-hardware-profile-name` for legacy HWPs, which is how the serving config remembers which legacy profile was used. An assertion on that annotation here would give confidence the full legacy path works end-to-end.

Neither is blocking, but they'd make the test more robust.

DaoDaoNoCode (commented):
Good targeted fix for the legacy HWP toleration injection bug. The root cause analysis is solid and the fix correctly addresses the issue by detecting legacy HWPs via missing `metadata.uid`. All CI checks pass (the E2E failure on `dash-e2e-int` appears pre-existing).

Left a couple of comments inline -- the main thing is an inconsistency between the two `isLegacyHardwareProfile` definitions in `assembleInferenceService`, and a few suggestions on the test coverage.

DaoDaoNoCode (commented):
All three comments from the first review are addressed. The `isLegacyHardwareProfile` inconsistency is resolved by extracting a shared `isLegacyHardwareProfileSelected` helper in `utils.ts` that both call sites now use. The `delete` cast has a clear explanatory comment. The test now covers nodeSelector injection and asserts the `legacy-hardware-profile-name` annotation value. Unit tests pass locally (85/85). The Cypress failure (`ClusterStorage - Add cluster storage with multiple workbench connections`) is unrelated to this PR.

Local checks: lint pass, type-check pass, unit tests pass (5 suites, 85 tests)

DaoDaoNoCode:
I can see the tolerations after fix. Nice job!

Screenshot 2026-04-29 at 5 24 20 PM

Previously:
Screenshot 2026-04-29 at 5 24 55 PM

DaoDaoNoCode (approved):
/lgtm

PR #6905, opened 2026-03-26
fix(e2e): pass contributor credentials in non-admin cluster settings …

rsun19:
https://redhat.atlassian.net/browse/RHOAIENG-55219

## Description

The E2E test "Test User - should not have access rights to view the Cluster Settings tab" in `testAdminClusterSettings.cy.ts` was intermittently failing with:

```
Timed out retrying after 10000ms: Unable to find an element by: [data-testid="not-found-page"]
```

**Root cause:** `clusterSettings.visit(false)` calls `cy.visitWithLogin('/settings/cluster/general')` without passing credentials, which defaults to `HTPASSWD_CLUSTER_ADMIN_USER`. This caused `oc login` to switch back to an admin user before visiting the page. The admin user can access cluster settings, so the expected 404 page never appeared.

**Fix:** Replace `clusterSettings.visit(false)` with `cy.visitWithLogin('/settings/cluster/general', LDAP_CONTRIBUTOR_USER)` to explicitly pass the contributor credentials.

**Why it was flaky (not always failing):** Race condition between the `oc login` admin switch completing and the page load timing. Sometimes the backend still served the contributor context if the page loaded before the admin token fully propagated.

## How Has This Been Tested?

- Type-check passes (`tsc --noEmit`)
- ESLint passes with no warnings
- The fix must be validated by CI running the `@ci-dashboard-regression-tags` E2E suite against a live cluster, as this test requires multiple LDAP users and `oc login` switching

## Test Impact

This PR fixes an existing flaky E2E test. No new tests are needed — the existing test should now pass reliably.

## Request review criteria:

If you have UI changes:


jenny-s51 (approved):
/lgtm

antowaddle (approved):
/lgtm

PR #6918, opened 2026-03-26
fix(model-serving): preserve OCI storageUri and imagePullSecrets on edit (RHOAIENG-53682)

mturley:
https://issues.redhat.com/browse/RHOAIENG-53682

## Description

When editing a KServe deployment that uses an unsaved OCI connection, the `storageUri` and `imagePullSecrets` fields were being dropped from the InferenceService after save.

### Root cause: KServe mutating webhook

The **KServe mutating admission webhook** (`inferenceservice.kserve-webhook-server.defaulter`) strips `storageUri` from `spec.predictor.model` when the secret currently referenced by the InferenceService's `imagePullSecrets` has been deleted — even if the PUT payload includes a valid `storageUri` and references a new, existing secret. This is a KServe bug tracked as [RHOAIENG-61105](https://redhat.atlassian.net/browse/RHOAIENG-61105); `storageUri` is independent of `imagePullSecrets` and should not be stripped. Workaround revert tracked as [RHOAIENG-61106](https://redhat.atlassian.net/browse/RHOAIENG-61106).

The dashboard triggered this by performing the following sequence on every edit of an unsaved OCI connection:
1. **Delete** the old generated secret (e.g. `secret-abc`)
2. **Create** a new secret with a new generated name (e.g. `secret-xyz`)
3. **PUT** the InferenceService with `imagePullSecrets` updated to `[{name: "secret-xyz"}]`

At step 3, the webhook checks the IS's *current* `imagePullSecrets` reference (`secret-abc`), finds it deleted, and strips `storageUri` from the spec.

### CLI reproduction of the KServe webhook bug

These steps reproduce the bug entirely from the CLI, independent of the dashboard. Replace `$NAMESPACE` with your target namespace.

```bash
# 1. Create an OCI pull secret with dashboard labels (required by the ODH admission webhook)
cat <<EOF | oc apply -f -
apiVersion: v1
kind: Secret
metadata:
  name: repro-oci-secret
  namespace: $NAMESPACE
  labels:
  annotations:
    opendatahub.io/connection-type-ref: oci-v1
    opendatahub.io/connection-type-protocol: oci
type: kubernetes.io/dockerconfigjson
data:
  .dockerconfigjson: eyJhdXRocyI6eyJxdWF5LmlvIjp7InVzZXJuYW1lIjoiZmFrZSIsInBhc3N3b3JkIjoiZmFrZSJ9fX0=
EOF

# 2. Create an InferenceService with storageUri + imagePullSecrets (OCI pattern)
cat <<EOF | oc create -f -
apiVersion: serving.kserve.io/v1beta1
kind: InferenceService
metadata:
  name: repro-oci-bug
  namespace: $NAMESPACE
  annotations:
    openshift.io/display-name: repro-oci-bug
    opendatahub.io/connections: repro-oci-secret
  labels:
spec:
  predictor:
    imagePullSecrets:
    - name: repro-oci-secret
    model:
      modelFormat:
        name: vLLM
      runtime: repro-oci-bug
      storageUri: oci://quay.io/example/model:latest
EOF

# 3. Verify storageUri is present
oc get inferenceservice repro-oci-bug -n $NAMESPACE \
  -o jsonpath='{.spec.predictor.model.storageUri}'
# Expected output: oci://quay.io/example/model:latest

# 4. Create a second secret (simulating what the dashboard does on edit)
cat <<EOF | oc apply -f -
apiVersion: v1
kind: Secret
metadata:
  name: repro-oci-secret-2
  namespace: $NAMESPACE
  labels:
  annotations:
    opendatahub.io/connection-type-ref: oci-v1
    opendatahub.io/connection-type-protocol: oci
type: kubernetes.io/dockerconfigjson
data:
  .dockerconfigjson: eyJhdXRocyI6eyJxdWF5LmlvIjp7InVzZXJuYW1lIjoiZmFrZSIsInBhc3N3b3JkIjoiZmFrZSJ9fX0=
EOF

# 5. DELETE the old secret (this is what the dashboard did before the fix)
oc delete secret repro-oci-secret -n $NAMESPACE

# 6. PUT the IS with imagePullSecrets changed to the new secret
oc get inferenceservice repro-oci-bug -n $NAMESPACE -o json | \
  python3 -c "
import json, sys
obj = json.load(sys.stdin)
obj['spec']['predictor']['imagePullSecrets'] = [{'name': 'repro-oci-secret-2'}]
obj['metadata']['annotations']['opendatahub.io/connections'] = 'repro-oci-secret-2'
json.dump(obj, open('/tmp/repro.json', 'w'))
" && oc replace -f /tmp/repro.json

# 7. Verify storageUri is stripped
oc get inferenceservice repro-oci-bug -n $NAMESPACE \
  -o jsonpath='{.spec.predictor.model.storageUri}'
# Expected output: (empty) ← BUG: webhook stripped storageUri

# Cleanup
oc delete inferenceservice repro-oci-bug -n $NAMESPACE
oc delete secret repro-oci-secret-2 -n $NAMESPACE
```

### Fix (workaround for [RHOAIENG-61105](https://redhat.atlassian.net/browse/RHOAIENG-61105))

**`connectionUtils.ts` (handleConnectionCreation)** — Instead of deleting the old generated secret inline before creating the new one, return a `cleanup` function alongside the created secret. The cleanup function encapsulates the old secret deletion logic but lets the caller control when it runs.

**`utils.ts` (deployModel)** — Call the returned `cleanup` function **after** the InferenceService has been updated to reference the new secret. This ensures the old secret still exists when the webhook processes the IS update, avoiding the bug.

This workaround should be reverted when [RHOAIENG-61105](https://redhat.atlassian.net/browse/RHOAIENG-61105) is fixed upstream (tech debt issue for revert: [RHOAIENG-61106](https://redhat.atlassian.net/browse/RHOAIENG-61106)).

## How Has This Been Tested?
- Reproduced the root cause via CLI (see reproduction steps above)
- Manually tested the fix on a live cluster: created OCI deployments, stopped them, edited them, and verified `storageUri` and `imagePullSecrets` persist after save
- Verified with `oc get` that the InferenceService spec is correct after each edit
- Lint and type-check pass cleanly

## Test Impact
- Added unit tests for `handleConnectionCreation` in `connectionUtils.spec.ts` covering:
  - Cleanup function is returned when old and new secret names differ
  - Old secret is not deleted until cleanup is explicitly called
  - Cleanup deletes the old generated secret
  - Non-generated secrets are not cleaned up
  - No cleanup function when there is no old connection
  - Cleanup function behavior during dry run

## Request review criteria:

### DIFF
--- a/packages/kserve/src/deployUtils.ts
+++ b/packages/kserve/src/deployUtils.ts
@@ -327,18 +327,27 @@ export const applyConnectionData = (
       delete result.metadata.annotations['opendatahub.io/connection-path'];
     }
   }
-  if (
-    modelLocationData.additionalFields.modelUri &&
-    isModelServingCompatible(
-      modelLocationData.connectionTypeObject ?? [],
-      ModelServingCompatibleTypes.OCI,
-    )
-  ) {
+  const isOci = isModelServingCompatible(
+    modelLocationData.connectionTypeObject ?? [],
+    ModelServingCompatibleTypes.OCI,
+  );
+
+  if (modelLocationData.additionalFields.modelUri && isOci) {
     result.spec.predictor.model = {
       ...result.spec.predictor.model,
       storageUri: modelLocationData.additionalFields.modelUri,
     };
   }
+
+  // Update imagePullSecrets to reference the connection secret for OCI connections,
+  // or clear stale imagePullSecrets when switching away from OCI
+  const connectionSecretName = secretName ?? createConnectionData.nameDesc?.name;
+  if (isOci && connectionSecretName) {
+    result.spec.predictor.imagePullSecrets = [{ name: connectionSecretName }];
+  } else {
+    delete result.spec.predictor.imagePullSecrets;
+  }
+
   return result;
 };
 
--- a/packages/kserve/src/deployUtils.ts
+++ b/packages/kserve/src/deployUtils.ts
@@ -327,10 +327,16 @@ export const applyConnectionData = (
       delete result.metadata.annotations['opendatahub.io/connection-path'];
     }
   }
-  const isOci = isModelServingCompatible(
-    modelLocationData.connectionTypeObject ?? [],
-    ModelServingCompatibleTypes.OCI,
-  );
+  // Determine OCI status from connectionTypeObject when available, falling back to
+  // the modelUri prefix when connectionTypeObject hasn't been resolved yet (e.g.
+  // during edit before connection types finish loading).
+  const isOci =
+    isModelServingCompatible(
+      modelLocationData.connectionTypeObject ?? [],
+      ModelServingCompatibleTypes.OCI,
+    ) ||
+    (!modelLocationData.connectionTypeObject &&
+      !!modelLocationData.additionalFields.modelUri?.startsWith('oci://'));
 
   if (modelLocationData.additionalFields.modelUri && isOci) {
     result.spec.predictor.model = {
@@ -344,7 +350,7 @@ export const applyConnectionData = (
   const connectionSecretName = secretName ?? createConnectionData.nameDesc?.name;
   if (isOci && connectionSecretName) {
     result.spec.predictor.imagePullSecrets = [{ name: connectionSecretName }];
-  } else {
+  } else if (!isOci) {
     delete result.spec.predictor.imagePullSecrets;
   }
 
--- a/packages/kserve/src/deployUtils.ts
+++ b/packages/kserve/src/deployUtils.ts
@@ -327,16 +327,10 @@ export const applyConnectionData = (
       delete result.metadata.annotations['opendatahub.io/connection-path'];
     }
   }
-  // Determine OCI status from connectionTypeObject when available, falling back to
-  // the modelUri prefix when connectionTypeObject hasn't been resolved yet (e.g.
-  // during edit before connection types finish loading).
-  const isOci =
-    isModelServingCompatible(
-      modelLocationData.connectionTypeObject ?? [],
-      ModelServingCompatibleTypes.OCI,
-    ) ||
-    (!modelLocationData.connectionTypeObject &&
-      !!modelLocationData.additionalFields.modelUri?.startsWith('oci://'));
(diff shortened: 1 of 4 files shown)


mturley:
@emilys314 I started fresh here and found that the original AI fix's root cause was entirely wrong. It seems like a bug in a kserve webhook. I opened that here: https://redhat.atlassian.net/browse/RHOAIENG-61105

This PR works around it in the dashboard, and I also opened a followup issue to revert this workaround when the real bug is fixed: https://redhat.atlassian.net/browse/RHOAIENG-61106

Should be good to go now. Full repro steps in the updated description.

emilys314:
> @emilys314 I started fresh here and found that the original AI fix's root cause was entirely wrong. It seems like a bug in a kserve webhook. I opened that here: https://redhat.atlassian.net/browse/RHOAIENG-61105
> 
> This PR works around it in the dashboard, and I also opened a followup issue to revert this workaround when the real bug is fixed: https://redhat.atlassian.net/browse/RHOAIENG-61106
> 
> Should be good to go now. Full repro steps in the updated description.

Oo interesting

emilys314:
I responded in the original slack thread that found the issue https://redhat-internal.slack.com/archives/C03UGJY6Z1A/p1778251117036339?thread_ts=1773684587.899709&cid=C03UGJY6Z1A

I'd prefer us not to have to merge in a temporary fix to workaround a backend issue, so I'd like to see how fast they can be with fixing the bug you created for them. 

Also on a similar note, this wouldn't be happening if we were reusing the same secret on redeployment. This seems like part of the root cause that caused this bug to surface in the UI,

mturley:
@emilys314 I'll close this PR. Do you think we should make a change to reuse the secret though? That could be a separate tech debt ticket if it's necessary.

emilys314:
> @emilys314 I'll close this PR. Do you think we should make a change to reuse the secret though? That could be a separate tech debt ticket if it's necessary.

Yep I made this jira ticket already https://redhat.atlassian.net/browse/RHOAIENG-61570

PR #7065, opened 2026-04-03
Fix dev proxy auth header forwarding in federated modules

mturley:
https://issues.redhat.com/browse/RHOAIENG-56705

## Description

Each federated module's webpack dev server captures the kubeconfig/oc token at startup as static proxy headers. This overwrites any dynamically forwarded auth headers from the host backend proxy, breaking dev impersonation (`DEV_IMPERSONATE_USER`/`DEV_IMPERSONATE_PASSWORD`).

This PR replaces the static `headers` approach with an `onProxyReq` callback that dynamically forwards the `Authorization` header from each incoming request when present, falling back to the startup-captured token for standalone dev mode.

**Modules fixed:**
- gen-ai (uses `oc whoami`)
- maas
- automl
- autorag
- eval-hub (both proxy entries)
- mlflow

**Not included:**
- `packages/notebooks/upstream/` — mirrors upstream kubeflow/notebooks; fix submitted upstream in https://github.com/kubeflow/notebooks/pull/1013
- `packages/model-registry/upstream/` — already fixed upstream in https://github.com/kubeflow/model-registry/pull/2544

## How Has This Been Tested?

Tested manually in the model-registry module (upstream fix) against an ODH cluster:
1. Started the host backend with `DEV_IMPERSONATE_USER` and `DEV_IMPERSONATE_PASSWORD` set
2. Verified the bug: with static headers, the impersonated user's token was overwritten by the startup token
3. Verified the fix: with `onProxyReq`, the incoming auth header from the host proxy is forwarded correctly, and impersonation works as expected

The same pattern is applied consistently to all other modules.

## Test Impact

This is a dev-only webpack config change (not shipped in production builds). No automated tests are applicable — webpack dev server proxy configuration is validated through manual dev workflow testing.

## Request review criteria:

### DIFF
--- a/packages/automl/frontend/config/webpack.dev.js
+++ b/packages/automl/frontend/config/webpack.dev.js
@@ -28,33 +28,54 @@ const DEPLOYMENT_MODE = process.env._DEPLOYMENT_MODE;
 const AUTH_METHOD = process.env._AUTH_METHOD;
 const BASE_PATH = DEPLOYMENT_MODE === 'kubeflow' ? '/automl/' : PUBLIC_PATH;
 
+// Get the kubeconfig token at startup as a fallback for standalone dev mode.
+const getKubeconfigToken = () => {
+  try {
+    const token = execSync(
+      "kubectl config view --raw --minify --flatten -o jsonpath='{.users[].user.token}'",
+    )
+      .toString()
+      .trim();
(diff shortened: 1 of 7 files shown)


lucferbux (approved):
/lgtm
/approve

After reviewing, this is doing more than it should but it does no harm, we can approve it and if anyone wants to tweak it later they can.

PR #7122, opened 2026-04-08
Add lint, type-check, and CI workflows for model-registry frontend and BFF

ppadti:
Closes: https://redhat.atlassian.net/browse/RHOAIENG-37600 partially

## Description

- This PR aims to add lint, type-check, and CI workflows for model-registry frontend and BFF.

- The two es-lint disable comment in  `packages/model-registry/upstream/frontend/src/app/types.ts` and  `packages/model-registry/upstream/frontend/src/app/modelCatalogTypes.ts` will be fixed in folloup as well.

- Will add the follow-up PR add cypress test and unit test to CI workflows

## How Has This Been Tested?
cd packages/model-registry/upstream/frontend
Run these:
`npm run test:lint`
`npm run test:type-check `

in BFF:
`make lint`
`make test`
`make build`

## Test Impact
NA

## Request review criteria:

If you have UI changes: 

## Summary by CodeRabbit

* **Chores**
  * Updated development tooling for the model registry package to enable improved local development, dashboard plugin support, and in-editor code editing for contributors. This is a developer-only change; it does not affect runtime behavior, public APIs, or exports.


packages/model-registry/upstream/frontend/src/app/types.ts
```diff
@@ -1,4 +1,5 @@
 import { APIOptions } from 'mod-arch-core';
+// eslint-disable-next-line no-relative-import-paths/no-relative-import-paths
```
manaswinidas, line 2: any reasons for these ?

packages/model-registry/upstream/frontend/src/app/modelCatalogTypes.ts
```diff
@@ -1,4 +1,5 @@
 import { APIOptions } from 'mod-arch-core';
+// eslint-disable-next-line no-relative-import-paths/no-relative-import-paths, import/order
```
manaswinidas, line 2: is this for reducing blast radius when it comes to merge conflicts while syncing?

manaswinidas (changes requested):

packages/model-registry/upstream/frontend/src/app/types.ts
ppadti: yeah, this was breaking the cypress tests present in packages/cypress folder because upstream ~ alias is not matching with the midstream one, so whenever the cypress looks for this path - that was failing. Is it okay if we do it in the next phase of this issue, where we add cypress to the CI?

packages/model-registry/upstream/frontend/src/app/modelCatalogTypes.ts
ppadti: same reason.

manaswinidas (commented):
/lgtm

.github/workflows/model-registry-bff-tests.yml
```diff
@@ -0,0 +1,49 @@
...
+        with:
+          go-version: '1.25.0'
+
+      - name: Lint
+        working-directory: packages/model-registry/upstream/bff
+        run: make lint
+
+      - name: Build (BFF)
+        working-directory: packages/model-registry/upstream/bff
+        run: make build
```
mturley, line 37: Ah - was going to say I don't think we're running `make test`, but it looks like `make build` also runs the tests via `build: fmt vet test`

packages/model-registry/upstream/frontend/src/app/modelCatalogTypes.ts
mturley: Are we planning to keep this issue open until we do the followup or open a new issue?

packages/model-registry/package.json
```diff
@@ -74,8 +74,11 @@
   },
   "devDependencies": {
     "@odh-dashboard/eslint-config": "*",
+    "@odh-dashboard/plugin-core": "*",
     "@odh-dashboard/tsconfig": "*",
     "@odh-dashboard/contract-tests": "*",
+    "@openshift/dynamic-plugin-sdk": "*",
+    "@patternfly/react-code-editor": "*",
```
mturley, line 81: Why are these new dependencies needed? for type resolution? seems harmless but I want to follow up and understand this change

mturley (approved):
LGTM. Thanks so much @ppadti

ppadti:
> Are we planning to keep this issue open until we do the followup or open a new issue?

Yeah, will keep that issue open - as we need to add cypress and unit test to the CI - that disabled es-lint will be removed along with that.

packages/model-registry/package.json
ppadti: Yeah, without these it was failing with "Cannot find module" errors, as the extensions are importing these.

ppadti: Now I am seeing this as you pointed it out - this is bit inconsistent from how other packages are handling the dependencies. Will fix it in follow-up. Thanks @mturley

PR #7152, opened 2026-04-09
Surface errors from responses API in Playground UI

toklumpp:
https://issues.redhat.com/browse/RHOAIENG-49767

## Description

Implements structured error handling for the Gen AI Playground responses API, replacing generic error messages with clear, actionable feedback. Errors from OGX are categorized by subsystem, mapped to user-friendly microcopy, and surfaced with appropriate retry guidance.

### Backend (BFF)

**Streaming error handling** — Previously the BFF only handled `stream.Err()` for transport-level errors. OGX produces two additional streaming error paths that were silently dropped:
- `response.failed` events: error details were lost because `ResponseData` had no error field
- `type: "error"` SSE events: filtered out entirely by `supportedEventTypes`

All three error paths are now intercepted in the streaming loop across `handleStreamingResponse`, `handleStreamingResponseAsync`, and `streamWithoutModeration`, producing consistent `{error: {code, message,
component, retriable}}` SSE events.

**Error constants** — Defined `OGXErr*` constants mirroring OGX's `_VALID_RESPONSE_ERROR_CODES` (sourced from `ogx/providers/inline/responses/builtin/responses/streaming.py`). Only codes OGX actually returns are handled — speculative OpenAI API codes (`tool_not_found`, `model_not_found`, `resource_not_found`, etc.) not confirmed to be returned were removed.

**Component constants** — `ComponentBFF`, `ComponentOGX`, `ComponentRAG`, `ComponentModel` identify which subsystem caused the error. The frontend uses these to select microcopy and classify errors as partial (warning) vs full (danger) failures.

**Guardrail error responses** — `guardrailViolationResponse` and `guardrailServiceUnavailableResponse` (renamed from `serviceUnavailableResponse`) now use `FrontendErrorResponse` with `component: "guardrails"` and `retriable: false`, so the frontend can match them to guardrail-specific microcopy instead of falling through to "Something went wrong".

**Other fixes:**
- `type: "error"` event codes parsed as HTTP status ints for correct `isRetriable` fallback
- Removed unused MCP error constants (`ErrCodeToolNotFound`, `ErrCodeInvalidConfig`)

### Frontend

- `ChatbotErrorAlert` component with expandable inline alerts matching UXD design
- `errorClassifier.ts` maps error codes to user-friendly messages via `${component}:${code}` template keys
- Microcopy templates for error scenarios the BFF and OGX actually produce today — no speculative entries
- Mock error triggers (`MOCK:` prefix in dev mode) for testing error UI without a backend
- `ApiErrorClass` for structured error propagation through streaming and non-streaming paths

## How Has This Been Tested?

**BFF error paths verified against local OGX (all confirmed with curl and in browser):**
- `response.failed` (`server_error`) — verified in browser: "Server error" with retry
- `type: "error"` (`404`, nonexistent model) — verified via curl: structured error with correct code/component
- `stream.Err()` (`CONNECTION_FAILED`, OGX down) — verified in browser: "Couldn't reach the server" with retry
- `guardrail_service_unavailable` — verified in browser: "Guardrail check was not applied"
- Non-streaming equivalents of above
- OGX error response formats traced through source to confirm which codes are returned and which are not

**Mock error triggers tested in browser:**
- `MOCK:server_error`, `MOCK:rate_limit_exceeded`, `MOCK:connection_failed`, `MOCK:timeout`
- `MOCK:guardrail_input_violation`, `MOCK:guardrail_output_violation`, `MOCK:guardrail_service_unavailable`
- `MOCK:stream_lost`, `MOCK:stream_timeout`

## Test Impact

- BFF: Added tests for `buildStreamingErrorEvent`, `extractStreamingError`, `isRetriable`, `ResolveComponent`, OGX error code constants, and component constructor tests. Removed test cases for speculative error codes.
- Frontend: Added `errorClassifier.spec.ts` (32 tests), `ChatbotErrorAlert.spec.tsx`, `useChatbotMessages.errors.spec.ts`. Removed speculative microcopy templates, mock scenarios, and test assertions for codes OGX doesn't return. Updated `ChatbotMessagesList.spec.tsx` to remove standalone loading dots tests.

## Request review criteria:

If you have UI changes:

Screenshot_21-4-2026_101024_localhost
Screenshot_21-4-2026_101034_localhost
Screenshot_21-4-2026_101159_localhost
Screenshot_21-4-2026_101213_localhost
Screenshot_21-4-2026_102343_localhost
Screenshot_21-4-2026_102446_localhost


packages/gen-ai/bff/internal/integrations/llamastack/response_errors.go
```diff
@@ -0,0 +1,205 @@
...
+// ErrorPattern represents a pattern to match against error messages
+type ErrorPattern struct {
+	Pattern  *regexp.Regexp
+	Category ResponseErrorCategory
+}
+
+// Common error patterns from LlamaStack, vLLM, and other components
+var errorPatterns = []ErrorPattern{
+	// Model configuration errors
+	{regexp.MustCompile(`(?i)max_tokens.*invalid|max_tokens.*exceed|max_tokens.*too (large|small)`), CategoryInvalidModelConfig},
```
NickGagan: Would recommend against parsing messages in errors to get user friendly messages, since this can be quite brittle and difficult to maintain.

Generally, you should only rely on error codes, providing the error details in the UI as a code snippet for more specific details would be enough

packages/gen-ai/bff/internal/integrations/llamastack/response_errors.go
```diff
@@ -0,0 +1,205 @@
...
+)
+
+// ErrorPattern represents a pattern to match against error messages
+type ErrorPattern struct {
+	Pattern  *regexp.Regexp
+	Category ResponseErrorCategory
+}
+
+// Common error patterns from LlamaStack, vLLM, and other components
+var errorPatterns = []ErrorPattern{
```
NickGagan: Were all of these tested to make sure they get returned from LLS?

packages/gen-ai/bff/internal/integrations/llamastack/response_errors.go
```diff
@@ -0,0 +1,205 @@
...
+		}
+	}
+
+	// Default to generic error
+	return CategoryGenericError
+}
+
+// GetUserFriendlyErrorMessage returns a user-friendly error message based on category and error code
+// This provides guidance to users on how to fix the issue
+func GetUserFriendlyErrorMessage(category ResponseErrorCategory, errorCode string) string {
```
NickGagan: Same as above, all of these were real errors you were able to get from LLS?

NickGagan (changes requested):
Just did a read through to unblock you, UI side looks good at first glance! Can do another pass once this is more cleaned up.

Something to address is scoping down the errors to ones that for sure get propagated from LLS and ones that we can parse from a code rather than applying regex to the message, seems like a lot of these are AI generated. Then as a follow up once this is merged is to raise issues in the upstream to see if we can get more specific errors for the spots that don't have coverage.

danijett-design:
Looks good from UX perspective

packages/gen-ai/bff/internal/api/lsd_helpers.go
```diff
@@ -43,40 +47,61 @@ func (app *App) getDefaultStatusCodeForLlamaStackClientError(errorCode string) i
...
+	}
+
+	// Determine component from error code
+	var component string
+	switch errorCode {
+	case "tool_not_found", "tool_error", "mcp_error":
+		component = "mcp"
+	case "resource_not_found", "vector_store_not_found":
+		component = "rag"
+	case "content_policy_violation", "content_blocked", "guardrail_violation":
```
NickGagan: You can drop the guardrails ones, we don't use LLS for this. Rest make sense to me!

Lucifergene:
Nice work on the error handling architecture — the BFF-to-frontend contract is clean and the microcopy system is well thought out.

One thing I noticed: when a streaming error occurs mid-stream (`lsd_responses_handler.go:482-489`), the error SSE event only includes `message` and `code`:

```go
errorData := map[string]interface{}{
    "error": map[string]interface{}{
        "message": errorMessage,
        "code":    errorCode,
    },
}
```

But the frontend's `ApiError` contract expects `component` and `retriable` as well. The `normalizeToApiError` fallback handles this gracefully (defaults to `component: 'bff'`, `retriable: false`), so nothing breaks — but it means streaming errors like connection lost or timeouts won't show the retry button even when they're transient.

Would be worth adding `component` and `retriable` to the streaming error event to stay consistent with the non-streaming error path. Should be a quick change.

NickGagan (approved):
Thanks @jharan1 @toklumpp!

## Changes Made During Review

- Fixed unsafe `strconv.Atoi` error swallowing in 3 streaming handlers that caused incorrect retriability classification for non-numeric error codes.
- Removed redundant copied state using PatternFly's built-in tooltip system instead (deleted 14 lines of unnecessary state management).
- Moved mock error support from frontend to BFF layer. 
- Centralized error component constants into a single source of truth to prevent drift across TypeScript types, display logic, and Go backend.
- Removed DOM manipulation workaround that used `useEffect` to remove `lang`
attribute from CodeBlock elements. This should be an upstream fix if there are A11Y concerns, this is too hacky of a workaround to put in the product.

PR #7163, opened 2026-04-09
fix(gen-ai): distinguish MaaS vs namespace models with the same model_id

NickGagan:
Issue: https://redhat.atlassian.net/browse/RHOAIENG-57527

## Description

Two AIModels sharing the same model_id but with different model_source_types (e.g. a namespace deployment and a MaaS model for the same underlying model) were incorrectly treated as duplicates in several places:

- AIModelTableRow: enabledModel lookup only compared modelId, causing both rows to show "Try in playground" when only one source type was in the playground.
- ChatbotConfigurationModal: preSelectedModels used a Set of modelIds, so opening the Configure playground modal pre-selected both source types when only one was in the playground.
- ChatbotConfigurationTable: useCheckboxTableBase key was model_name, so selecting one row also selected the other if names matched.
- SubscriptionDropdown: MaaS subscription lookup could match a namespace model sharing the same base model_id.

Fix by introducing isMaasLlamaModelId (checks the "maas-" provider prefix in the LlamaStack model ID) and isPlaygroundModelMatchForAIModel (matches a LlamaModel to an AIModel accounting for source type). These are used consistently across all four call sites.

## How Has This Been Tested?

Registered a maas model and llm-d model in playground & inferenced with them.

https://github.com/user-attachments/assets/eb65f358-b7ef-454e-9faa-655100098fcf

## Test Impact

Needed to update the provider ids in the mock tests and added two tests to prevent this from regressing.

## Request review criteria:

If you have UI changes: 

- [x]  Included tags to the UX team if it was a UI/UX change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed model-matching so models with the same ID from different sources are correctly distinguished, preventing incorrect "Try in playground" vs "Add to playground" behavior and spurious MaaS launch tracking.
  * Subscription dropdown now only appears for MaaS-compatible models; non-MaaS models show no options.
  * Selection, select-all, and row identity fixes prevent distinct models that share a name from being collapsed, preserving intended selections.

### DIFF
--- a/packages/gen-ai/frontend/src/app/AIAssets/components/AIModelTableRow.tsx
+++ b/packages/gen-ai/frontend/src/app/AIAssets/components/AIModelTableRow.tsx
@@ -33,6 +33,7 @@ import {
 } from '~/app/types';
 import ChatbotConfigurationModal from '~/app/Chatbot/components/chatbotConfiguration/ChatbotConfigurationModal';
 import { genAiAiAssetsTabRoute, genAiChatPlaygroundRoute } from '~/app/utilities/routes';
+import { isPlaygroundModelMatchForAIModel } from '~/app/utilities/utils';
 import useAiAssetVectorStoresEnabled from '~/app/hooks/useAiAssetVectorStoresEnabled';
 import { GenAiContext } from '~/app/context/GenAiContext';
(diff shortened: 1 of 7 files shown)


pnaik1:
/lgtm 
this looks good

jharan1 (approved):

PR #7182, opened 2026-04-10
fix(autorag/eval): restrict local uploads to json

daniduong:
https://redhat.atlassian.net/browse/RHOAIENG-58656

## Description
Restricts the file upload component in AutoRAG evaluation configuration to only accept JSON files. This adds validation to the dropzone by specifying the accepted file type as `application/json` with `.json` extension.

This prevents users from attempting to upload incompatible file formats that would fail later in the process, providing immediate feedback at the upload stage.

## How Has This Been Tested?
- Manually tested the file upload component in the AutoRAG evaluation configuration
- Verified that only JSON files can be selected/dropped
- Confirmed that non-JSON files are rejected by the dropzone

## Test Impact
Testing focused on UI behavior - the dropzone now properly restricts file types to JSON only. No additional automated tests added as this is a UI constraint configuration change.

## Request review criteria:

If you have UI changes: 

## Summary by CodeRabbit

* **Bug Fixes**
  * File upload now restricts selection to JSON files only.

### DIFF
--- a/packages/autorag/frontend/src/app/components/configure/AutoragEvaluationSelect.tsx
+++ b/packages/autorag/frontend/src/app/components/configure/AutoragEvaluationSelect.tsx
@@ -49,6 +49,7 @@ function AutoragEvaluationSelect(): React.JSX.Element {
         }}
         onClear={() => field.onChange('')}
         fileUploadProps={{
+          dropzoneProps: { accept: { 'application/json': ['.json'] } },
           filenamePlaceholder: 'Drag and drop or browse from...',
           // @ts-expect-error: bypass ts error to allow icon
           browseButtonText: (


jefho-rh:
Thanks @daniduong! Works well on my end. No issues with AI review. 

Just a comment. When I drag a non-JSON file to the dropbox, it looks like I can still technically do it. But after dropping the file in, nothing happens. Do you think we need some sort of error message here?

image

---------
✅ Manual testing looks good. I am no longer able to upload any non-json files from S3 or locally

https://github.com/user-attachments/assets/736943ca-0e45-4b24-86d6-f766ce93f2bf

✅ All tests passing


Suite | Status
-- | --
Lint | ✅ Pass
Type-check | ✅ Pass
Unit Tests | ✅ Pass
Contract Tests | ✅ Pass
AutoRAG BFF Tests | ✅ Pass
Cypress Mock Tests | ✅ Pass
Quality Gates | ✅ Pass


/lgtm

daniduong:
@jefho-rh
> Just a comment. When I drag a non-JSON file to the dropbox, it looks like I can still technically do it. But after dropping the file in, nothing happens. Do you think we need some sort of error message here?

good catch! we can probably do this in a followup, as I realize the input file dropzone also has the same behaviour so we can address both at the same time, but this would probably be like a minor sev, since we already tell them what files are allowed
FYI @MatthewAThompson

PR #7202, opened 2026-04-13
Trim response to just needed body (#7190)

andrewballantyne:
https://redhat.atlassian.net/browse/RHOAIENG-57750


nananosirova (approved):
/lgtm

PR #7214, opened 2026-04-13
fix(maas): add empty state with Create tier button when no tiers exis…

andrewballantyne:
Merging our 3.3.2-fixes & 3.3.0-fixes branches together.

3.3.2 was a misunderstanding and poor coordination on my part. It'll be impossible to handle z-streams this way, so we need to reinforce our process. 

```
* c42a026c9 (HEAD -> v3.3.0-fixes, upstream/v3.3.0-fixes) Trim response to just needed body (#7190) (#7202)
| * 0934c675c (upstream/v3.3.2-fixes, v3.3.2-fixes) fix(maas): add empty state with Create tier button when no tiers exist (#6915)
|/
* b3e0bb2f1 update rate limit policies when predicate (#6218) (#6230)
```


rsun19 (commented):
/lgtm

PR #7188, opened 2026-04-10
fix(e2e): shorten model registry test fixture prefixes to avoid DNS limit

mturley:
## Description
KServe constructs DNS hostnames by concatenating `{resourceName}-predictor-{namespace}`. The previous test fixture prefixes (`test-object-model` + `test-deploy-project`) combined with the UUID suffix produced hostnames exceeding the 63-character DNS label limit, causing the InferenceService to fail with `ReconcileFailed`.

Shortened `objectStorageModelName` from `test-object-model` to `test-obj-model` and `deployProjectNamePrefix` from `test-deploy-project` to `test-deploy-prj`, bringing the combined hostname from 72 to 58 characters.

**Note:** This is a test-only fix. The underlying product issue remains — there is no UI-side validation warning users when their model name + namespace would exceed the KServe hostname limit. A follow-up issue should be opened to add client-side validation in the deployment wizard.

## How Has This Been Tested?
- Verified the math: worst-case hostname `test-obj-model-XXXXXX-v10-predictor-test-deploy-prj-XXXXXX` = 58 chars (under 63 limit)
- Ran `testRegistryDeployModel` E2E test against a live RHOAI cluster — the previous prefixes caused `ReconcileFailed` due to DNS label length

## Test Impact
This PR fixes the `testRegistryDeployModel` E2E test fixture data. No new tests needed.

## Request review criteria:
Self checklist (all need to be checked):
- [x] The developer has manually tested the changes and verified that the changes work
- [x] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [x] The developer has added tests or explained why testing cannot be added (unit or cypress tests for related changes)
- [x] The code follows our [Best Practices](/docs/best-practices.md) (React coding standards, PatternFly usage, performance considerations)
After the PR is posted & before it merges:
- [ ] The developer has tested their solution on a cluster by using the image produced by the PR to `main`

## Summary by CodeRabbit

* **Tests**
  * Updated test configuration values to ensure test infrastructure consistency.

PR #7231, opened 2026-04-14
Fix flaky modelserving e2e tests

sridarna:
closes https://redhat.atlassian.net/browse/RHOAIENG-57922
## Description

This pr resolves the failing Model Serving / smokeSet3 cypress E2E tests.

- Add select selectSingleServingModelButtonIfExists to  testDeployLLMDServing.
- Add checkInferenceServiceState before calling the inference endpoint to make sure model is running.
- Reduce project name length, as the test is failing with fails to create NewRawKubeReconciler for predictor since project name must be no more than 63 characters.
created ticket https://redhat.atlassian.net/browse/RHOAIENG-57819 to handle this better.

## How Has This Been Tested?

Run in Jenkins against rhoai cluster
/job/dashboard/job/dashboard-e2e-tests/207/Test_20Reports/
image

## Test Impact

N/A
## Request review criteria:

If you have UI changes: 


packages/cypress/cypress/tests/e2e/dataScienceProjects/models/testDeployOCIModel.cy.ts
```diff
@@ -216,6 +216,12 @@ describe(
 
         // Submit the changes
         modelServingWizardEdit.findSubmitButton().click();
+        modelServingSection.findModelServerDeployedName(modelDeploymentName);
+
+        cy.step('Verify that the Model is running');
+        cy.get<string>('@resourceName').then((resourceName) => {
```
antowaddle: Can we extract this out? We don't typically use cy.get directly in test files.

antowaddle (changes requested):
@sridarna small change here, also, note I updated the label as it contained an '@' which is not required. Could we delete the previous label?

packages/cypress/cypress/tests/e2e/dataScienceProjects/models/testDeployOCIModel.cy.ts
sridarna: Updated

sridarna:
> @sridarna small change here, also, note I updated the label as it contained an '@' which is not required. Could we delete the previous label?

Removed the wrong tag

sridarna:
Tested out new changes in Jenkins. The tests are passing
/job/dashboard/job/dashboard-e2e-tests/208/Test_20Reports/

packages/cypress/cypress/tests/e2e/dataScienceProjects/models/testDeployLLMDServing.cy.ts
```diff
@@ -119,7 +120,9 @@ describe('A user can deploy an LLMD model', () => {
         .findResourceNameInput()
         .should('be.visible')
         .invoke('val')
-        .as('resourceName');
+        .then((val) => {
+          resourceName = val as string;
+        });
```
emilys314, line 125: Just curious, how different is the `.as('resourceName');` compare to setting a js variable for practicality? 

`resourceName = val` seems much more familiar to me

emilys314 (commented):
Seems good to me

packages/cypress/cypress/tests/e2e/dataScienceProjects/models/testDeployLLMDServing.cy.ts
sridarna: I made that change to address review comment https://github.com/opendatahub-io/odh-dashboard/pull/7231#pullrequestreview-4106921852
to avoid the cy.get

FedeAlonso (approved):
/lgtm

mturley:
Related: https://github.com/opendatahub-io/odh-dashboard/pull/7188

packages/cypress/cypress/tests/e2e/dataScienceProjects/models/testDeployLLMDServing.cy.ts
emilys314: Okay makes sense, so the cypress variables are unconventional in our code base, got it. makes sense

PR #7178, opened 2026-04-10
Remove toast notification for MCP deployment and sort it by Created

ppadti:
Closes: https://redhat.atlassian.net/browse/RHOAIENG-57566

## Description
This PR aims to remove toast notification for MCP deployment and sort it by Created

https://github.com/user-attachments/assets/7cf6ec3e-f08f-4e4d-be28-e8f6ca08b677

## How Has This Been Tested?

1. Go to MCP details page and click on deployment
2. Fill the required fields and click on Deploy button
3. We will not see the toast notification and Deployment table is sorted by creation timestamp

## Test Impact
Added cypress test

## Request review criteria:

Self checklist (all need to be checked):
- [x] The developer has manually tested the changes and verified that the changes work
- [x] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [x] The developer has added tests or explained why testing cannot be added (unit or cypress tests for related changes)
- [ ] The code follows our [Best Practices](/docs/best-practices.md) (React coding standards, PatternFly usage, performance considerations)

If you have UI changes: 

- [x] Included any necessary screenshots or gifs if it was a UI change.
- [x] Included tags to the UX team if it was a UI/UX change.

After the PR is posted & before it merges:
- [ ] The developer has tested their solution on a cluster by using the image produced by the PR to `main`

## Summary by CodeRabbit

* **Tests**
  * Added test utilities for MCP Deployments page interactions, including methods to access deployment table rows and creation time details.
  * Added test case validating MCP Deployments default sort order by "Created" timestamp (newest-first).

PR #7131, opened 2026-04-08
feat(mcp-deployments): microcopy and UX updates for MCP server deployments

YuliaKrimerman:
https://issues.redhat.com/browse/RHOAIENG-51726
Microcopy doc https://docs.google.com/document/d/1-UvuMMreVLxX1IPMdeU8KCrbJ7qASz0SwhJsTTEeJmc/edit?tab=t.0

## Description

Microcopy and UX improvements across the MCP server deployment experience:

**Deploy MCP server modal (`McpDeployModal`):**
- Removed the **Reset** button from the modal footer (Close + Deploy/Save only)
- Updated OCI image popover text to clarify the field is read-only from the catalog
- 
Screenshot 2026-04-08 at 11 10 13 AM

- Renamed "Configuration (YAML)" label to "YAML configuration"
- Updated YAML popover and helper text to be more concise and actionable
- 
Screenshot 2026-04-08 at 11 01 50 AM

**MCP server deployments page (`McpDeploymentsPage`):**
- Added page heading with RocketIcon in a teal circular background
- Reordered table columns: **Name → MCP server → Created → Status → Service** (was Server → Name → ...)
- Renamed "Server" column to "MCP server"
- Updated kebab menu order: **Edit → Publish as AI asset → Delete** (added "Publish as AI asset" action)
- Updated status tooltips: Available → "The pod and its containers are healthy and running." / Unavailable → "At least 1 container in the pod failed."
- Updated filter placeholder to "Filter by name or MCP server" and added `serverName` to filter matching
- 
Screenshot 2026-04-08 at 11 55 32 AM
Screenshot 2026-04-08 at 11 55 49 AM
Screenshot 2026-04-08 at 11 56 26 AM
Screenshot 2026-04-08 at 11 59 40 AM

**Delete confirmation modal (`DeleteMcpDeploymentModal`):**
- Uses `displayName` (bold) in the body copy instead of raw `deployment.name`
- Simplified button label from "Delete MCP server deployment" to "Delete"
- Removed placeholder text and helper text from the confirmation input
- Added required indicator (`*`) after the confirmation prompt

**Publish as AI asset:**
- No confirmation modal — clicking "Publish as AI asset" immediately shows a success toast notification
- Toast includes the deployment display name, and a link to the AI asset endpoints MCP tab

**Shared `DeleteModal` component:**
- Added `confirmationRequiredIndicator` prop to optionally append `*` to the confirmation prompt
[...]

PR #6999, opened 2026-04-01
test(cypress): add mocked MCP deployment delete flow

Taj010:
For https://redhat.atlassian.net/browse/RHOAIENG-56331
## Description

Adds Cypress coverage for deleting an MCP server deployment from AI hub → MCP server deployments: open the row kebab, choose Delete, confirm by typing the deployment name, and assert the DELETE request. Also covers cancel (no DELETE) and a failed delete (inline error, modal stays open).

## How Has This Been Tested?

Ran the mocked Cypress spec locally
` cd packages/cypress && npm run run:mock -- --spec cypress/tests/mocked/mcpCatalog/mcpDeploymentsDelete.cy.ts
` and confirmed all three scenarios pass: successful delete (204), cancel, and delete failure (forceNetworkError + inline alert).

## Request review criteria:

Self checklist (all need to be checked):
- [X] The developer has manually tested the changes and verified that the changes work
- [X] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [X] The developer has added tests or explained why testing cannot be added (unit or cypress tests for related changes)
- [X] The code follows our [Best Practices](/docs/best-practices.md) (React coding standards, PatternFly usage, performance considerations)

If you have UI changes: 

- [ ] Included any necessary screenshots or gifs if it was a UI change.
- [ ] Included tags to the UX team if it was a UI/UX change.

After the PR is posted & before it merges:
- [ ] The developer has tested their solution on a cluster by using the image produced by the PR to `main`

## Summary by CodeRabbit

* **Tests**
  * Added end-to-end tests covering MCP server deployment deletion: confirms modal UI, requires typing the deployment name to enable submit, validates successful deletion, verifies cancel prevents deletion, and shows inline error on failure.
* **Chores**
  * Added a test page object for the MCP deployments UI to support the new tests.

PR #7235, opened 2026-04-14
Move MCP deployments downstream and use ProjectsContext for namespace selection

mturley:
https://issues.redhat.com/browse/RHOAIENG-56566
https://issues.redhat.com/browse/RHOAIENG-55983

**WIP - needs testing, an agent did this and I haven't looked at it thoroughly yet**

## Description

MCP deployments is a downstream-only feature — it doesn't exist in the upstream kubeflow/model-registry repo (MCP **catalog** is upstream; MCP **deployments** is not). All MCP deployment files were living in `packages/model-registry/upstream/frontend/src/` but belong in `packages/model-registry/src/`.

This PR switches the MCP deployments namespace selector from the upstream `NamespaceSelectorFieldWrapper` to the host's `ProjectsContext` via a new `ProjectSelectorField` component ([RHOAIENG-56566](https://issues.redhat.com/browse/RHOAIENG-56566)). `ProjectsContext` is provided by `@odh-dashboard/internal`, which transitively pulls in `@patternfly/react-styles` CSS imports. Files in `upstream/frontend/src/` are compiled by the federated webpack, which can't handle those CSS imports (88 webpack errors). Moving the files downstream to `src/` means they're compiled by the host webpack instead, which has full CSS support.

This made the file move ([RHOAIENG-55983](https://issues.redhat.com/browse/RHOAIENG-55983)) a prerequisite for the ProjectsContext change — so both are done together in this PR.

### What moved
- **Core odh components** → `src/odh/` (McpDeploymentsWrapper, McpDeployModal, McpDeployModalExtension, McpDeployButton, extension-points)
- **Types and API** → `src/types/`, `src/api/`
- **Hooks** → `src/hooks/` (useMcpServerDeployAvailable, useMcpServerConverter)
- **Pages** → `src/pages/mcpDeployments/` (all page components, tests, utils)
- **Utils** → `src/utils/` (mcpServerYaml)
- **Mocks** → `src/__mocks__/`

### New components
- **McpDeploymentsCoreLoader** (`src/odh/McpDeploymentsCoreLoader.tsx`) — new layout route component that handles redirect to preferred project, invalid namespace detection, and "no projects" empty state. Uses both `ProjectsContext` and mod-arch-core's `updatePreferredNamespace` for dual-context sync.

### URL routing change
The MCP deployments URL changed from query-param-based (`/ai-hub/mcp-deployments?namespace=X`) to path-param-based (`/ai-hub/mcp-servers/deployments/:namespace`). `McpDeploymentsRoutes` was rewritten to use a nested route with `McpDeploymentsCoreLoader` as a layout route and `/:namespace` as a path param. `McpDeploymentsPage` was updated to use `useParams` instead of `useSearchParams` for namespace selection.

### Import rewriting
- `~/app/...` imports → relative paths to moved files or `../../upstream/frontend/src/...` for files that stay upstream
- `~/concepts/k8s/K8sNameDescriptionField` → `@odh-dashboard/internal/concepts/k8s/K8sNameDescriptionField/K8sNameDescriptionField`
- `~/odh/components/NamespaceSelectorFieldWrapper` → replaced with `ProjectSelectorField` using `ProjectsContext`
- `~/odh/components/OdhDevFeatureFlagOverridesProvider` → removed (not needed for MCP deployments)

### Extension registration changes
- MCP deployment extensions (tab route, redirect, deploy modal hook) moved from upstream `odh/extensions.ts` to root `extensions.ts`
- `McpDeployButton` decoupled from upstream via a new `mcp-catalog.mcp-server/deploy-button` extension point, so `McpServerDetailsPage.tsx` (upstream) loads it dynamically instead of importing directly

### McpDeployModal migrated to ContentModal wrapper
The `McpDeployModal` was migrated from direct PatternFly `Modal`/`ModalBody`/`ModalFooter`/`ModalHeader` imports to the project's `ContentModal` wrapper to satisfy the `@odh-dashboard/no-restricted-imports` lint rule. This changes the footer button layout from a custom `Split` to a standard `ActionList` row (Deploy, Close). This aligns with the project-standard modal pattern.

### Unused providers removed from McpDeploymentsWrapper
Removed `AppContext.Provider`, `NotificationContextProvider`, and `NotificationListener` from `McpDeploymentsWrapper`. These were carried over from the upstream wrapper pattern (`McpCatalogWrapper`, `ModelRegistryWrapper`), but nothing in the MCP deployments code consumes them — no page calls `useAppContext`, and toast notifications were removed in #7178.

### Upstream DeleteModal duplicate data-testid fix
The upstream `DeleteModal.tsx` had a duplicate `data-testid` on its confirm button (`delete-modal-submit-button` and `delete-modal-confirm-button` on the same element). This was harmless on `main` because `upstream/` is excluded from `tsc`, but our downstream files import `DeleteModal` via relative paths, pulling it into type-check and surfacing TS17001. Fixed by removing the stale `delete-modal-submit-button` attribute, keeping `delete-modal-confirm-button` (which matches the real upstream repo).

### Config changes
- `tsconfig.json` — added `paths` for `mod-arch-core`, `mod-arch-shared`, `mod-arch-kubeflow` (resolved from `upstream/frontend/node_modules/`)
- `jest.config.ts` — full config with mod-arch module mappings, `transformIgnorePatterns`, `moduleDirectories`, `testPathIgnorePatterns`
- `package.json` — added `@odh-dashboard/plugin-core` dependency, updated `mcpDeploymentTypes` export path, added `src/__mocks__/*` to `mocks/*` export

### Lint fixes (pre-existing upstream issues now under stricter downstream rules)
- `react/jsx-no-constructed-context-values` — wrapped context value in `useMemo`
- `@typescript-eslint/no-unnecessary-condition` — removed `?? 3` fallback in phase sort comparator; to make this safe, added a `validatePhase` function in `mcpDeploymentService.ts` that normalizes unknown API phases to `PENDING` at the data layer
- `no-restricted-syntax` — removed `isOpen` prop from modals, use conditional rendering
- `react-hooks/rules-of-hooks` — placed the new `useMemo` (from `jsx-no-constructed-context-values` fix above) before early returns

### Changes preserved from rebase

This refactor was originally based on [`c44d764`](https://github.com/opendatahub-io/odh-dashboard/commit/c44d764e92524e4ef86be7f55eb58a89836b6c75) and was rebased onto current `main`. Three commits landed on `main` after the original base that modified files this PR moves. All changes from those commits have been verified as correctly preserved in the moved files:
- [#6999](https://github.com/opendatahub-io/odh-dashboard/pull/6999) — Added `mockMcpDeploymentList` mock factory and `McpDeploymentList` import ✅
- [#7131](https://github.com/opendatahub-io/odh-dashboard/pull/7131) — Microcopy and UX updates: `Tooltip` → `Popover` on status labels, `tooltip` → `popoverBody` field rename, column reorder (Name first), "MCP server" label, `getDeploymentDisplayName` in delete modal, `confirmationRequiredIndicator`, updated OCI image / YAML help text, removed reset button / `initialYaml` state, `useMemo` on table row actions, `serverName` added to search filter ✅
- [#7178](https://github.com/opendatahub-io/odh-dashboard/pull/7178) — `defaultSortColumn={2}` for Created column, newest-first sort comparator, removed toast notifications, added cypress sort test ✅

## How Has This Been Tested?
- `npx turbo run type-check --filter=@odh-dashboard/model-registry` — passes clean
- `npx turbo run type-check --filter=@odh-dashboard/cypress` — passes clean
- `npx turbo run lint --filter=@odh-dashboard/model-registry` — passes (0 errors, 0 warnings)
- Upstream unit tests — 50 suites pass, 624 tests pass
- TODO: Build verification, browser testing

## Test Impact
All existing MCP deployment tests were moved and updated with correct import paths. No test logic was changed.

## Request review criteria:
If you have UI changes:


packages/model-registry/src/pages/mcpDeployments/McpDeploymentsTableColumns.ts
```diff
@@ -31,7 +31,7 @@ export const mcpDeploymentColumns: SortableData<McpDeployment>[] = [
   {
     field: 'status',
     label: 'Status',
-    sortable: (a, b) => (phaseOrder[a.phase] ?? 3) - (phaseOrder[b.phase] ?? 3),
+    sortable: (a, b) => phaseOrder[a.phase] - phaseOrder[b.phase],
```
mturley, line 34: The `@typescript-eslint/no-unnecessary-condition` lint rule flagged the `?? 3` fallback here because `phaseOrder` is typed as `Record<McpDeploymentPhase, number>` — so TypeScript sees the lookup as always defined.

But `McpDeployment.phase` comes from an API response, and APIs can return values outside the enum (e.g. a new phase added server-side before the UI is updated). If `phase` is an unexpected string, `phaseOrder[a.phase]` returns `undefined` and the subtraction produces `NaN`, which silently breaks sorting.

A few options to handle this without fighting the lint rule:

1. **Type guard at the boundary** — validate `phase` when deserializing the API response, defaulting unknown values to a known phase (e.g. `PENDING`). This is the cleanest fix if there's a central place where `McpDeployment` objects are constructed from API data.

2. **Widen the type** — change `McpDeployment.phase` from `McpDeploymentPhase` (enum) to `string`, and update `phaseOrder` to `Partial<Record<string, number>>`. Then the `??` fallback is no longer flagged as unnecessary. This is more honest about what the API actually guarantees, but may require downstream type narrowing.

3. **Restore the fallback with a lint-disable** — if options 1 or 2 would require significant refactoring, adding an inline `// eslint-disable-next-line` with a comment explaining the API boundary concern is acceptable. Could open a follow-up ticket to address it properly later.

Any of these work — just want to make sure we don't silently break sorting if the API ever returns a phase value we don't expect.

mturley: Replying to my AI-self: let's go with option 1.

mturley (commented):
AI self-review

packages/model-registry/src/odh/McpDeploymentsWrapper.tsx
```diff
@@ -8,37 +9,35 @@ import {
...
-            </NotificationContextProvider>
-          </OdhDevFeatureFlagOverridesProvider>
+          <NotificationContextProvider>
+            <NotificationListener>
+              <McpDeploymentsRoutes />
+            </NotificationListener>
+          </NotificationContextProvider>
         </BrowserStorageContextProvider>
       </ThemeProvider>
     </AppContext.Provider>
```
mturley: Are `AppContext.Provider`, `NotificationContextProvider`, and `NotificationListener` still needed here?

- **AppContext**: Nothing in the moved downstream MCP deployment code calls `useAppContext`. It's provided because the wrapper was cloned from the upstream pattern (`McpCatalogWrapper`, `ModelRegistryWrapper`), but those wrappers serve pages that _do_ consume it (e.g. `useUser`, `RegisterModel`). The MCP deployments pages don't.

- **NotificationContextProvider + NotificationListener**: The toast notifications that used to be in `McpDeployModal` were removed in #7178. No file under `src/pages/mcpDeployments/` or `src/odh/components/` calls `useNotification` or interacts with `NotificationContext`. The `NotificationListener` bridge (which forwards mod-arch-core notifications to the host's Redux store via `CustomEvent`) has no notifications to forward.

Since this is downstream-only code compiled by the host webpack, these providers add context layers that nothing consumes. Could these three be removed to simplify the wrapper?

mturley (commented):
More AI self-review

mturley:
This is getting ridiculous. Runaway AI is doing the wrong thing. Closing this again and I'll follow up with something sensible when I can.

PR #7239, opened 2026-04-14
fix(RHOAIENG-14563): translate K8s error messages to user-friendly text in deploy modals

Philip-Carneiro:
https://redhat.atlassian.net/browse/RHOAIENG-14563

## Description

When deploying a model using the KServe or NIM modals, Kubernetes error messages were shown directly to the user in the error alert. These messages contain internal K8s resource references (e.g., `servingruntimes.serving.kserve.io "test-model" already exists`) that are confusing and unhelpful for end users.

This PR introduces a `translateModelServingError` utility function that intercepts K8s error messages and translates them into clear, user-friendly messages. Specifically:

- **Duplicate name errors** are converted to: `A model deployment with the name "X" already exists. Please choose a different Model deployment name.`
- **Other K8s resource references** (e.g., `servingruntimes.serving.kserve.io` or `inferenceservices.serving.kserve.io`) are replaced with `model deployment` to remove jargon.
- **Unrelated errors** pass through unchanged.

Additionally, both `ManageKServeModal` and `ManageNIMServingModal` have been refactored to use the `ContentModal` component instead of manually composing `Modal` + `ModalBody` + `ModalFooter` + `DashboardModalFooter`, reducing duplication and aligning with project patterns.

The new wizard footer (`WizardFooterWithDisablingNext.tsx`) also applies error translation in its `DeployErrorAlert` component to ensure consistent user-friendly error messages across both legacy modals and the new deployment wizard.

Before:
before

After:
after

## How Has This Been Tested?

- Manually triggered duplicate name deployment errors in both KServe and NIM modals and verified the translated message is displayed
- Verified that non-duplicate K8s errors (e.g., forbidden) also have resource references replaced
- Verified unrelated errors pass through unchanged
- All existing unit tests pass
- 6 new unit tests added for `translateModelServingError` covering all translation branches
- Lint and TypeScript compilation pass with no errors

## Test Impact

- **Unit tests**: 6 new tests added in `utils.spec.ts` for `translateModelServingError`
- **Cypress tests**: No new E2E tests (UI behavior unchanged, only error message text differs)
- **Manual testing**: Duplicate name scenario, forbidden error scenario, generic error passthrough

## Request review criteria:

If you have UI changes:

---
Generated with Claude Code 🤖 

## Summary by CodeRabbit

* **Bug Fixes**
  * Standardized model-serving error messages to surface friendly duplicate-name guidance and rewrite KServe-specific references to "model deployment".

* **Refactor**
  * Replaced multiple modal implementations with a unified modal component across model-serving flows; submit/cancel behavior and error handling centralized.
  * Error displays now show translated, user-facing messages.

* **Tests**
  * Added and updated tests/mocks to validate translated error messages and the revised modal submit/cancel behavior.


packages/model-serving/src/components/generic/WizardFooterWithDisablingNext.tsx
```diff
@@ -13,6 +13,8 @@ import {
   AlertActionLink,
 } from '@patternfly/react-core';
 import { K8sStatusError } from '@odh-dashboard/internal/api/errorUtils';
+// eslint-disable-next-line @odh-dashboard/no-restricted-imports
```
manaswinidas: **Blocking:** Lint-suppressed `@odh-dashboard/internal` deep import into a package. Move `translateModelServingError` to a shared package so both consumers can import cleanly.

frontend/src/pages/modelServing/screens/projects/utils.ts
```diff
@@ -728,3 +728,21 @@ export const getModelPathFromUri = (uri: string): string => {
     return '';
   }
 };
```
manaswinidas: **Blocking:** `/g` flag on a module-level const regex makes `.test()` stateful — `lastIndex` advances between calls. If called twice in a render cycle (React strict mode), the second call silently returns wrong results. Drop `/g`; `.test()` doesn't need it and `.replace()` on these messages only has one occurrence.

frontend/src/pages/modelServing/screens/projects/kServeModal/ManageKServeModal.tsx
```diff
@@ -261,8 +254,9 @@ const ManageKServeModal: React.FC<ManageKServeModalProps> = ({
     onClose(submitted);
   };
 
-  const setErrorModal = (e: Error) => {
-    setError(e);
+  const setErrorModal = (e: unknown) => {
+    const msg = e instanceof Error ? e.message : String(e || 'Unknown error');
+    setError(new Error(translateModelServingError(msg)));
     setActionInProgress(false);
```
manaswinidas, line 263: **DRY:** This `setErrorModal` is identical in `ManageNIMServingModal.tsx`. Extract to a shared helper in `utils.ts`.

packages/model-serving/src/components/generic/WizardFooterWithDisablingNext.tsx
```diff
@@ -148,12 +150,21 @@ export const ModelDeploymentFooter: React.FC<DeploymentFooterProps> = ({
   );
 };
 
+const getErrorMessage = (error: Error | null | undefined): string => {
+  if (!error) {
+    return '';
+  }
+  return error.message;
+};
+
```
manaswinidas: **Nit:** `getErrorMessage` is just `error?.message ?? ''`. Inline it.

frontend/src/pages/modelServing/screens/projects/__tests__/utils.spec.ts
```diff
@@ -613,3 +614,55 @@ describe('getPVCNameFromURI', () => {
...
+describe('translateModelServingError', () => {
+  it('should return a friendly message when a duplicate name error includes a quoted name', () => {
+    const k8sError = 'servingruntimes.serving.kserve.io "test-model" already exists';
+    expect(translateModelServingError(k8sError)).toBe(
+      'A model deployment with the name "test-model" already exists. Please choose a different Model deployment name.',
+    );
+  });
+
+  it('should return a friendly message for inferenceservices duplicate name errors', () => {
+    const k8sError = 'inferenceservices.serving.kserve.io "my-deployment" already exists';
```
manaswinidas: **Nit (repetitive):** This and the `inferenceservices` test below hit the same code path — only the regex alternation differs. Combine into `it.each`.

frontend/src/pages/modelServing/screens/projects/nim/NIMServiceModal/__tests__/ManageNIMServingModal.spec.tsx
```diff
@@ -493,9 +491,9 @@ describe('ManageNIMServingModal', () => {
       expect(mockOnClose).toHaveBeenCalledWith(false);
     });
 
-    it('handles form submission', () => {
+    it('renders submit button', () => {
```
manaswinidas: **Tautological:** Renamed from `handles form submission` but still just asserts the button exists. Either test actual submission (click → verify callback) or remove.

frontend/src/pages/modelServing/screens/projects/nim/NIMServiceModal/__tests__/ManageNIMServingModal.spec.tsx
```diff
@@ -476,12 +473,13 @@ describe('ManageNIMServingModal', () => {
         jest.fn(),
       ]);
       render(<ManageNIMServingModal onClose={mockOnClose} projectContext={mockProjectContext} />);
-      expect(screen.getByTestId('modal-footer')).toBeInTheDocument();
+      expect(screen.getByTestId('modal-submit-button')).toBeInTheDocument();
+      expect(screen.getByTestId('modal-submit-button')).toBeDisabled();
     });
```
manaswinidas, line 481: **Incomplete:** The companion test `enables submit button when form is valid` never asserts `not.toBeDisabled()`. Without that, it doesn't verify the enabled state.

frontend/src/pages/modelServing/screens/projects/nim/NIMServiceModal/__tests__/ManageNIMServingModal.spec.tsx
```diff
@@ -533,6 +530,7 @@ describe('ManageNIMServingModal', () => {
         return { status: true };
       });
```
manaswinidas, line 535: **Unnecessary:** Comment relocations throughout this file (`// Mock serving runtime params as enabled` moved a few lines) are pure diff noise. Avoid cosmetic changes in unrelated code.

manaswinidas (changes requested):
Good UX fix. Three merge-blocking issues (cross-boundary import, stateful regex, DRY) plus some test quality nits.

manaswinidas:
I'll leave the subsequent reviews to @mturley @YuliaKrimerman @emilys314 (I don't want to step on your toes 😅 )

frontend/src/pages/modelServing/screens/projects/utils.ts
```diff
@@ -728,3 +728,21 @@ export const getModelPathFromUri = (uri: string): string => {
...
+      const nameMatch = message.match(/"([^"]+)"/);
+      const name = nameMatch?.[1];
+      return name
+        ? `A model deployment with the name "${name}" already exists. Please choose a different Model deployment name.`
+        : 'A model deployment with this name already exists. Please choose a different Model deployment name.';
+    }
+    return message;
+  }
+  K8S_SERVING_RESOURCE_PATTERN.lastIndex = 0;
+  return message.replace(K8S_SERVING_RESOURCE_PATTERN, 'model deployment');
```
mturley: @Philip-Carneiro string-matching on server-provided strings like error messages is an antipattern because they could change and we don't want to maintain frontend logic that depends on backend internal logic. Is there a way we can detect this error more generically? What's the HTTP error code here? I would expect a k8s request with an existing name to give a [409 Conflict](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/409) response.

It looks like the same error code is already checked later in the file to decide whether to render the "force update" button as well.

Here's claude's full summary on this:

----

## Suggestion: Use HTTP status code instead of string matching

The `translateModelServingError` function currently matches error message strings with regex (`/already exists/i`, `K8S_SERVING_RESOURCE_PATTERN`). This is fragile — K8s error message wording isn't part of the API contract and could change across versions.

The error objects thrown by K8s API calls are `K8sStatusError` instances (defined in `frontend/src/api/errorUtils.ts`) that carry structured data:

```typescript
class K8sStatusError extends Error {
  statusObject: K8sStatus; // includes: code (409), reason ("AlreadyExists"), message, details
}
```

There's already precedent in this codebase for checking the status code instead — see `frontend/src/concepts/pipelines/content/configurePipelinesServer/utils.ts`:

```typescript
if (e.statusObject.code === 409 && e.statusObject.reason === 'AlreadyExists') { ... }
```

And there's an existing utility `getGenericErrorCode(error)` in `errorUtils.ts` that extracts the status code from either `K8sStatusError` or `AxiosError`.

**Suggested approach:** Instead of `translateModelServingError(message: string)`, accept the full error object and check `error instanceof K8sStatusError && error.statusObject.code === 409` for the duplicate-name case. The resource name in the message can be based on the user's form input.

mturley (changes requested):

frontend/src/api/errorUtils.ts
```diff
@@ -38,3 +38,24 @@ export const getGenericErrorCode = (error: unknown): number | undefined => {
   }
   return undefined;
 };
+
+const K8S_SERVING_RESOURCE_PATTERN = /(?:servingruntimes|inferenceservices)\.serving\.kserve\.io/i;
```
emilys314: Sorry for more difficulty with this PR. Interpretting error messages is difficult which is why we don't. But it's very valuable to know which resource is having the issue. 

The main "deployment" is the inferenceservice and would better qualify as the "model deployment"

The servingRuntime is a config for the deployment, not the deployment itself..

frontend/src/api/errorUtils.ts
```diff
@@ -38,3 +38,24 @@ export const getGenericErrorCode = (error: unknown): number | undefined => {
...
+export const translateModelServingError = (error: unknown): string => {
+  if (
+    error instanceof K8sStatusError &&
+    error.statusObject.code === 409 &&
+    error.statusObject.reason === 'AlreadyExists'
+  ) {
+    const name = error.statusObject.details?.name;
+    return name
+      ? `A model deployment with the name "${name}" already exists. Please choose a different Model deployment name.`
+      : 'A model deployment with this name already exists. Please choose a different Model deployment name.';
```
emilys314: nitpick, but you have the first `model deployment` as all lowercase, and the second one has the M as uppercase `Model deployment`

YuliaKrimerman:
@Philip-Carneiro Needs rebase and comments resolution please

YuliaKrimerman:
@Philip-Carneiro this has merge conflicts, please take a look.

Philip-Carneiro:
/retest

These failing test are not related to my PR:
AutoRAG
Feature Store

frontend/src/api/errorUtils.ts
```diff
@@ -25,3 +25,40 @@ export const getGenericErrorCode = (error: unknown): number | undefined => {
...
+
+  let message = error instanceof Error ? error.message : String(error || 'Unknown error');
+  for (const [pattern, replacement] of K8S_RESOURCE_REPLACEMENTS) {
+    message = message.replace(pattern, replacement);
+  }
+  return message;
+};
+
+export const createModelServingError = (error: unknown): Error =>
+  new Error(translateModelServingError(error));
```
mturley: These are model-serving related utils, and you’re having to import them from deep in the internal code with an eslint-disable comment (in `WizardFooterWithDisablingNext`, which is in the `model-serving` package). I think it makes more sense to put them somewhere in that `model-serving` package, especially because I don't think we need to use them outside that package (see my other comments).

frontend/src/api/errorUtils.ts
```diff
@@ -25,3 +25,40 @@ export const getGenericErrorCode = (error: unknown): number | undefined => {
...
+const K8S_RESOURCE_REPLACEMENTS: [RegExp, string][] = [
+  [/inferenceservices\.serving\.kserve\.io/i, 'model deployment'],
+  [/servingruntimes\.serving\.kserve\.io/i, 'serving runtime'],
+];
+
+const getResourceDisplayName = (kind?: string): string => {
+  if (kind === 'inferenceservices') {
+    return 'model deployment';
+  }
+  return 'resource';
```
mturley: This is fairly hard-coded logic and also excludes other resource types we can create in the wizard like servingruntimes or llminferenceservices. we may want to use a map similar to your `K8S_RESOURCE_REPLACEMENTS`, or maybe just drop this entirely and lift your use of `K8S_RESOURCE_REPLACEMENTS` higher in the `translateModelServingError` function so it is also used for the `resourceName`. Also, `resourceName` and `displayName` here read like it’s the name of the resource rather than its type, we may want different variable names.

frontend/src/pages/modelServing/screens/projects/__tests__/utils.spec.ts
```diff
@@ -574,3 +579,129 @@ describe('getPVCFromURI', () => {
...
+    const uri = 'not a uri';
+    expect(getPVCNameFromURI(uri)).toEqual('');
+  });
+  it('should return an empty string if the URI is not a PVC URI', () => {
+    const uri = 'http://pvc-1/model-path';
+    expect(getPVCNameFromURI(uri)).toEqual('');
+  });
+});
+
+describe('translateModelServingError', () => {
```
mturley: The function this is testing isn’t in the file this tests file is colocated with (`pages/modelServing/screens/projects/utils`), we should maybe leave this test file alone and colocate the tests with your new util

frontend/src/pages/modelServing/screens/projects/kServeModal/ManageKServeModal.tsx
```diff
@@ -341,12 +331,30 @@ const ManageKServeModal: React.FC<ManageKServeModalProps> = ({
   };
 
   return (
-    <Modal variant="medium" isOpen onClose={() => onBeforeClose(false)}>
-      <ModalHeader
-        title={editInfo ? 'Edit model' : 'Deploy model'}
-        description="Configure properties for deploying your model"
-      />
-      <ModalBody>
+    <ContentModal
```
mturley: ManageKServeModal is the old model deployment modal, which is deprecated and no longer maintained. Now that the wizard is used, this file is dead code that will be removed, and we shouldn’t change it.

frontend/src/pages/modelServing/screens/projects/utils.ts
```diff
@@ -702,3 +703,24 @@ export const getPVCFromURI = (
...
 };
+
+export const getModelPathFromUri = (uri: string): string => {
+  try {
+    const url = new URL(uri);
+    return url.pathname.replace(/^\//, '');
+  } catch {
+    return '';
+  }
+};
```
mturley: It seems like you added this util and its tests but didn't use it anywhere, maybe it's a remnant of an earlier version? we can probably get rid of it and its tests

frontend/src/pages/modelServing/screens/projects/utils.ts
```diff
@@ -702,3 +703,24 @@ export const getPVCFromURI = (
...
+
+export { translateModelServingError, createModelServingError };
+export { getPVCNameFromURI, isPVCUri };
```
mturley: Why re-export? also it seems like `getPVCNameFromURI` and `isPVCUri` are also not used in your changes and this re-exporting is just making it possible to add tests for it in the `frontend/src/pages/modelServing/screens/projects/__tests__/utils.spec.ts` file?

It also is strange to re-export `translateModelServingError, createModelServingError` here to put tests in that file for the same reason. They should go in `packages/model-serving/src/shared` somewhere along with tests there, and be imported where needed using `from '@odh-dashboard/model-serving/shared'`.

packages/model-serving/src/components/generic/WizardFooterWithDisablingNext.tsx
```diff
@@ -13,6 +13,8 @@ import {
   AlertActionLink,
 } from '@patternfly/react-core';
 import { K8sStatusError } from '@odh-dashboard/k8s-core';
+// eslint-disable-next-line @odh-dashboard/no-restricted-imports
+import { translateModelServingError } from '@odh-dashboard/internal/api/errorUtils';
```
mturley: Moving `translateModelServingError` to the model-serving package will mean you don't need the `eslint-disable-next-line @odh-dashboard/no-restricted-imports`.

frontend/src/pages/modelServing/screens/projects/nim/NIMServiceModal/ManageNIMServingModal.tsx
```diff
@@ -389,12 +378,29 @@ const ManageNIMServingModal: React.FC<ManageNIMServingModalProps> = ({
   };
 
   return (
-    <Modal variant="medium" isOpen onClose={() => onBeforeClose(false)}>
-      <ModalHeader
-        title={`${editInfo ? 'Edit' : 'Deploy'} model with NVIDIA NIM`}
-        description="Configure properties for deploying your model using an NVIDIA NIM."
-      />
-      <ModalBody>
+    <ContentModal
```
mturley: I think we may need to not bother modifying ManageNIMServingModal either, because it is being removed soon as NIM is being integrated into the serving wizard. [Thread here](https://redhat-internal.slack.com/archives/C069KSM8T9N/p1785864354402969)

frontend/src/pages/modelServing/screens/projects/utils.ts
```diff
@@ -702,3 +703,24 @@ export const getPVCFromURI = (
...
+
+export const handleModelServingError = (
+  e: unknown,
+  setError: (error: Error | undefined) => void,
+  setActionInProgress: (inProgress: boolean) => void,
+): void => {
+  setError(createModelServingError(e));
+  setActionInProgress(false);
+};
```
mturley: It looks like your `handleModelServingError` function is only used in ManageKServeModal and ManageNIMServingModal, so if we're not modifying those we can just get rid of it. which also means we can get rid of `createModelServingError` since it's only used in this function.

mturley (changes requested):
Sorry to tear this apart after all this time @Philip-Carneiro ! I think we can remove most of these changes and limit the PR's scope to the changes in `WizardFooterWithDisablingNext`, and the util required for it, which should live in the model-serving package (and its tests as well).

mturley (approved):
/approve
/lgtm

Thanks @Philip-Carneiro !

Image

PR #7252, opened 2026-04-15
fix: flaky argo workflow and v1 pipeline import tests in `pipelines.cy.ts`

caponetto:
## Description

https://redhat.atlassian.net/browse/RHOAIENG-58263

The Cypress mock test "uploads fails with argo workflow" is flaky because `onSubmit` can fire with a stale closure where `isArgoWorkflow=false`, bypassing client-side argo detection and calling `submitAction()` against an un-intercepted upload endpoint that hangs forever. This PR scopes `pipelineVersionImportModal.findSubmitButton()` to the modal footer (matching the pattern used by `pipelineImportModal`), and adds defensive `mockUploadPipeline`/`mockUploadVersion` intercepts to the three affected tests so the request returns immediately if the race occurs, preventing the permanent loading state that causes the 10s timeout.

## How Has This Been Tested?

By checking if CI passes 5x in a row without failing.

## Test Impact

N/A

## Request review criteria:

If you have UI changes: 

## Summary by CodeRabbit

* **Tests**
  * Improved pipeline import test reliability with enhanced mocking and UI assertion coverage for import workflows.

### DIFF
--- a/packages/cypress/cypress/tests/mocked/pipelines/pipelines.cy.ts
+++ b/packages/cypress/cypress/tests/mocked/pipelines/pipelines.cy.ts
@@ -849,7 +849,11 @@ describe('Pipelines', () => {
     pipelineImportModal.shouldBeOpen();
     pipelineImportModal.fillPipelineName('New pipeline');
     pipelineImportModal.fillPipelineDescription('New pipeline description');
+    // Safety net: intercept the upload endpoint so the test never hangs if the
+    // client-side argo detection is bypassed due to a stale-closure race.
+    pipelineImportModal.mockUploadPipeline({}, projectName);
     pipelineImportModal.uploadPipelineYaml(argoWorkflowPipeline);
+    pipelineImportModal.findSubmitButton().should('be.enabled');
     pipelineImportModal.submit();
 
     pipelineImportModal.findImportModalError().should('exist');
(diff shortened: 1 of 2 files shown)


srtanish1992:
@caponetto I ran the tests locally and after the latest changes, there is one test `fails to import a v1 pipeline` failing locally repeteadly.

**From Claude:**

The argo workflow tests (import + version) look correct — the should('be.enabled') wait and mockUploadPipeline/mockUploadVersion safety nets are the right approach since argo detection is a client-side check that blocks the API call entirely.

However, the "fails to import a v1 pipeline" test is broken by the mockUploadPipeline safety net. Unlike argo, v1 detection is server-side — the upload always goes to the API, which must reject with InvalidInputError for the v1 error message to appear. The mock intercept returns success instead, so the modal closes and the test fails.

**Suggested fix:** Remove mockUploadPipeline from the v1 test. Keep the should('be.enabled') assertion (it's harmless). The v1 test was never affected by the original race condition since there's no client-side gate.

packages/cypress/cypress/tests/mocked/pipelines/pipelines.cy.ts
```diff
@@ -1107,7 +1112,11 @@ describe('Pipelines', () => {
     pipelineVersionImportModal.selectPipelineByName('Test pipeline');
     pipelineVersionImportModal.fillVersionName('Argo workflow version');
     pipelineVersionImportModal.fillVersionDescription('Argo workflow version description');
+    // Safety net: intercept the upload endpoint so the test never hangs if the
+    // client-side argo detection is bypassed due to a stale-closure race.
```
bobbravo2, line 1116: Stale closure race... new ai band name? 😝

bobbravo2 (approved):
/lgtm
Thank you! 🙏

PR #7259, opened 2026-04-15
fix(model-registry): scope in-cluster K8s client to SAR-only interface in MCP deployment auth

manaswinidas:
https://issues.redhat.com/browse/RHOAIENG-57223

## Description

In `mcp_deployment_auth.go`, the `initSARClient()` function created a full `kubernetes.Interface` client using `rest.InClusterConfig()` and stored it as a **package-level variable** (`sarClient`). While only used for SAR checks, this pattern poses a preventive privilege escalation risk:

- `sarClient` was a full `kubernetes.Interface` capable of any operation the pod SA is allowed to do
- As a package-level variable, it was accessible to any code added to the `handlers` package
- If someone copies this pattern or reuses the client, they'd perform K8s operations as the pod's SA rather than the requesting user

**Fix:** Replace the broad `kubernetes.Interface` with `authorizationv1client.SubjectAccessReviewInterface` — the narrowest possible interface that can only perform SubjectAccessReview operations:

- Removed the `kubernetes` import (full clientset), replaced with `k8s.io/client-go/kubernetes/typed/authorization/v1`
- Renamed `sarClient`/`sarClientOnce` → `sarReviews`/`sarReviewsOnce` to reflect the narrower scope
- `initSARReviews()` now creates only an `AuthorizationV1Client` and stores only the `SubjectAccessReviewInterface`
- `requireMcpDeploymentAccess()` calls `reviews.Create(...)` directly instead of chaining through `client.AuthorizationV1().SubjectAccessReviews().Create(...)`

Lazy initialization with `sync.Once` is preserved for performance.

## How Has This Been Tested?

Ran all 27 MCP deployment handler tests — all pass without modification:

```
go test ./internal/redhat/handlers/ -v -run "TestMcpDeployment"
```

Also ran `go vet ./internal/redhat/handlers/` and `go build ./...` — both clean.

## Test Impact

No test changes needed. The existing SAR-specific tests (`TestMcpDeploymentRequireAccessSkipsSARWhenAuthNotInternal`, `TestMcpDeploymentListUserTokenSkipsSAR`) use `AuthMethod: config.AuthMethodUser`, which causes `requireMcpDeploymentAccess` to return `true` immediately without touching the SAR client. The refactoring only changed internals below that early-return path.

## Request review criteria:

### DIFF
--- a/packages/model-registry/upstream/bff/internal/redhat/handlers/mcp_deployment_auth.go
+++ b/packages/model-registry/upstream/bff/internal/redhat/handlers/mcp_deployment_auth.go
@@ -21,8 +21,8 @@ func requireMcpDeploymentAccess(app *api.App, w http.ResponseWriter, r *http.Req
 		return false
 	}
 	if client == nil {
-		app.Logger().Warn("K8s client unavailable, skipping MCP access check")
-		return true
+		app.ServerError(w, r, fmt.Errorf("kubernetes client factory returned nil client without error"))
+		return false
 	}
 
 	identity, _ := r.Context().Value(constants.RequestIdentityKey).(*k8s.RequestIdentity)
--- a/packages/model-registry/upstream/bff/internal/redhat/handlers/mcp_deployment_auth.go
+++ b/packages/model-registry/upstream/bff/internal/redhat/handlers/mcp_deployment_auth.go
@@ -1,87 +1,40 @@
 package handlers
 
 import (
-	"context"
 	"fmt"
 	"log/slog"
 	"net/http"
-	"sync"
-	"time"
-
-	authv1 "k8s.io/api/authorization/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/client-go/kubernetes"
-	"k8s.io/client-go/rest"
 
 	"github.com/kubeflow/model-registry/ui/bff/internal/api"
-	"github.com/kubeflow/model-registry/ui/bff/internal/config"
 	"github.com/kubeflow/model-registry/ui/bff/internal/constants"
 	k8s "github.com/kubeflow/model-registry/ui/bff/internal/integrations/kubernetes"
 )
 
-const (
-	mcpServerSARAPIGroup = "mcp.x-k8s.io"
-	mcpServerSARResource = "mcpservers"
-	sarRequestTimeout    = 30 * time.Second
-)
-
-var (
-	sarClientOnce sync.Once
-	sarClient     kubernetes.Interface
-)
-
-// initSARClient lazily initializes a typed K8s clientset using in-cluster config.
-// Returns nil when not running in a cluster (e.g. federated/dev mode).
-func initSARClient() kubernetes.Interface {
-	sarClientOnce.Do(func() {
-		cfg, err := rest.InClusterConfig()
-		if err != nil {
-			return
-		}
-		sarClient, _ = kubernetes.NewForConfig(cfg)
-	})
-	return sarClient
-}
-
-// requireMcpDeploymentAccess performs a SubjectAccessReview to verify that
-// the requesting user has the given verb on mcpservers in the target namespace.
+// requireMcpDeploymentAccess performs a permission check to verify that the
+// requesting user has the given verb on mcpservers in the target namespace.
+// The check is delegated to KubernetesClientInterface.CanVerbMcpServersInNamespace,
+// which uses SAR (internal auth) or SSAR (user-token auth) as appropriate.
 func requireMcpDeploymentAccess(app *api.App, w http.ResponseWriter, r *http.Request, namespace, verb string) bool {
-	if app.Config().AuthMethod != config.AuthMethodInternal {
-		return true
-	}
-
-	client := initSARClient()
-	if client == nil {
-		app.Logger().Warn("SAR client unavailable in internal auth mode, skipping access check")
-		return true
-	}
-
-	identity, ok := r.Context().Value(constants.RequestIdentityKey).(*k8s.RequestIdentity)
-	if !ok || identity == nil {
-		app.BadRequest(w, r, fmt.Errorf("missing RequestIdentity in context"))
+	client, err := app.KubernetesClientFactory().GetClient(r.Context())
+	if err != nil {
+		app.ServerError(w, r, fmt.Errorf("failed to get Kubernetes client for MCP access check: %w", err))
 		return false
 	}
-
-	ctx, cancel := context.WithTimeout(r.Context(), sarRequestTimeout)
-	defer cancel()
-
-	sar := &authv1.SubjectAccessReview{
-		Spec: authv1.SubjectAccessReviewSpec{
-			User:   identity.UserID,
-			Groups: identity.Groups,
-			ResourceAttributes: &authv1.ResourceAttributes{
-				Verb:      verb,
-				Group:     mcpServerSARAPIGroup,
-				Resource:  mcpServerSARResource,
-				Namespace: namespace,
-			},
-		},
+	if client == nil {
+		app.Logger().Warn("K8s client unavailable, skipping MCP access check")
+		return true
 	}
 
-	response, err := client.AuthorizationV1().SubjectAccessReviews().Create(ctx, sar, metav1.CreateOptions{})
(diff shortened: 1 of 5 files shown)


packages/model-registry/upstream/bff/internal/redhat/handlers/mcp_deployment_auth.go
```diff
@@ -26,21 +27,27 @@ const (
...
-// initSARClient lazily initializes a typed K8s clientset using in-cluster config.
+// initSARReviews lazily initializes a SubjectAccessReview client using
+// in-cluster config. Only the narrow SubjectAccessReviewInterface is retained,
+// preventing reuse for other K8s operations.
 // Returns nil when not running in a cluster (e.g. federated/dev mode).
-func initSARClient() kubernetes.Interface {
-	sarClientOnce.Do(func() {
+func initSARReviews() authorizationv1client.SubjectAccessReviewInterface {
+	sarReviewsOnce.Do(func() {
 		cfg, err := rest.InClusterConfig()
```
ppadti: Could we move this MCP SAR into `internal/integrations/kubernetes/` — for example add `CanVerbMcpServersInNamespace(ctx, identity, ns, verb) (bool, error)` on KubernetesClientInterface? Then `requireMcpDeploymentAccess` would only: (1) call app.KubernetesClientFactory().GetClient to get the shared k8s helper used elsewhere, and (2) call CanVerbMcpServersInNamespace on it and map true/false (and errors) to allow/deny. That removes initSARReviews, the separate InClusterConfig path, and returning / caching authorizationv1client.SubjectAccessReviewInterface from `mcp_deployment_auth.go`.

ppadti (changes requested):

packages/model-registry/upstream/bff/internal/redhat/handlers/mcp_deployment_auth.go
```diff
@@ -26,21 +27,27 @@ const (
...
 		if err != nil {
 			return
 		}
-		sarClient, _ = kubernetes.NewForConfig(cfg)
+		authClient, err := authorizationv1client.NewForConfig(cfg)
+		if err != nil {
+			return
+		}
+		sarReviews = authClient.SubjectAccessReviews()
```
Philip-Carneiro: The error from `NewForConfig` is silently swallowed. When this fails, `sarReviews` stays `nil`, and `requireMcpDeploymentAccess` will log "SAR client unavailable" and skip the access check entirely. Without logging the actual error here, it becomes very difficult to diagnose why SAR checks are being skipped in production.

suggested fix:
```go
		authClient, err := authorizationv1client.NewForConfig(cfg)
		if err != nil {
			slog.Error("failed to create authorization client for SAR checks", slog.Any("error", err))
			return
		}
```

The `sync.Once` closure has no access to `app.Logger()`, but using the `slog` default logger (or accepting a logger parameter) would provide critical observability for production debugging. The original code also silently discarded this error (`sarClient, _ = kubernetes.NewForConfig(cfg)`), but since this PR is explicitly improving error handling, this is the right time to fix it.

Philip-Carneiro (changes requested):
Echoing @ppadti's existing review comment: `initSARReviews()` creates a separate `InClusterConfig` path that bypasses the `KubernetesClientFactory` used by the rest of the BFF. The `KubernetesClientInterface` already defines several SAR-related permission methods (`CanListServicesInNamespace`, `CanAccessServiceInNamespace`, etc.). A method like `CanVerbMcpServersInNamespace(ctx, identity, ns, verb) (bool, error)` on `KubernetesClientInterface` would:

1. Eliminate the separate `rest.InClusterConfig()` call and package-level `sync.Once` state
2. Make the SAR check fully testable through existing mock client infrastructure
3. Align with the project's established pattern for permission checks

This PR's narrowing of the interface type is still a valid improvement over the status quo. If the architectural refactor is out of scope for this ticket, consider filing a follow-up issue.

packages/model-registry/upstream/bff/internal/redhat/handlers/mcp_deployment_auth.go
manaswinidas: Done. Moved the SAR logic into `KubernetesClientInterface.CanVerbMcpServersInNamespace` with SAR on `InternalKubernetesClient` and SSAR on `TokenKubernetesClient`. `requireMcpDeploymentAccess` now just calls `app.KubernetesClientFactory().GetClient` and delegates to the interface method. `initSARReviews`, the separate `InClusterConfig` path, and all package-level state are removed.

packages/model-registry/upstream/bff/internal/redhat/handlers/mcp_deployment_auth.go
manaswinidas: Addressed by the same refactor — `initSARReviews` and the `sync.Once` closure no longer exist. The SAR/SSAR logic now lives in `CanVerbMcpServersInNamespace` on the K8s client implementations, where errors are returned to the caller rather than swallowed.

packages/model-registry/upstream/bff/internal/redhat/handlers/mcp_deployment_auth.go
```diff
@@ -1,87 +1,40 @@
...
 
-	identity, ok := r.Context().Value(constants.RequestIdentityKey).(*k8s.RequestIdentity)
-	if !ok || identity == nil {
-		app.BadRequest(w, r, fmt.Errorf("missing RequestIdentity in context"))
-		return false
-	}
-
-	ctx, cancel := context.WithTimeout(r.Context(), sarRequestTimeout)
-	defer cancel()
+	identity, _ := r.Context().Value(constants.RequestIdentityKey).(*k8s.RequestIdentity)
```
ppadti: We need to return error if identity is nil 

```suggestion
	identity, ok := r.Context().Value(constants.RequestIdentityKey).(*k8s.RequestIdentity)
	if !ok || identity == nil {
		app.BadRequest(w, r, fmt.Errorf("missing RequestIdentity in context"))
		return false
	}
```

packages/model-registry/upstream/bff/internal/redhat/handlers/mcp_deployment_auth.go
```diff
@@ -1,87 +1,40 @@
...
+	if err != nil {
+		app.ServerError(w, r, fmt.Errorf("failed to get Kubernetes client for MCP access check: %w", err))
+		return false
 	}
-
-	client := initSARClient()
 	if client == nil {
-		app.Logger().Warn("SAR client unavailable in internal auth mode, skipping access check")
+		app.Logger().Warn("K8s client unavailable, skipping MCP access check")
 		return true
```
ppadti: Sorry, This is my mistake of returning true, when the client is nil - I think we should return an error instead

```suggestion
		app.ServerError(w, r, fmt.Errorf("kubernetes client factory returned nil client without error"))
		return false
```

ppadti (changes requested):

ppadti:
@manaswinidas Could you please rebase and test the BFF tests with `make test` - few tests are failing after the last commit

packages/model-registry/upstream/bff/internal/redhat/handlers/mcp_deployment_auth.go
ppadti: @manaswinidas this review comment is not fixed yet

packages/model-registry/upstream/bff/internal/redhat/handlers/mcp_deployment_auth.go
ppadti: same here

packages/model-registry/upstream/bff/internal/redhat/handlers/mcp_deployment_auth.go
manaswinidas: The force-push reverted the changes 😓 - should have checked after rebase and force-push

ppadti (approved):
Thanks @manaswinidas 
/lgtm

One small changes we need to do here is that we need add access check for user_token as well. That is out of scope for this issue - will create an issue to track that.

manaswinidas (commented):
/approve

Self- approve with team lgtm

PR #7280, opened 2026-04-16
Fix genai playground null error

Lucifergene:
## Request review criteria:

If you have UI changes: 

## Summary by CodeRabbit

* **Tests**
  * Enhanced test coverage for model utility functions, including edge cases for null/undefined model data and various configuration scenarios.

* **Bug Fixes**
  * Improved robustness of model utility functions to gracefully handle missing or null model data with appropriate fallbacks.

### DIFF
--- a/packages/gen-ai/frontend/src/app/utilities/__tests__/utils.spec.ts
+++ b/packages/gen-ai/frontend/src/app/utilities/__tests__/utils.spec.ts
@@ -1,8 +1,11 @@
 /* eslint-disable camelcase */
-import type { AIModel, MaaSModel } from '~/app/types';
+import type { AIModel, LlamaModel, MaaSModel } from '~/app/types';
 import {
   convertMaaSModelToAIModel,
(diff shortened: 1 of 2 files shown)


NickGagan (commented):
Thanks for the quick work! I think the safer fix here would be to update the BFFs to return an empty array instead of null when no data is coming back. Should also check whether any of the APIs we rely on have an incorrect contract, might be a bug in MaaS API if they are returning null instead of an empty array: https://github.com/opendatahub-io/models-as-a-service/blob/c7895772bf202b3f547c6f72e95413faf7998fa6/maas-api/openapi3.yaml#L639.

Lucifergene:
@NickGagan If we’re touching the BFF, I’d strongly recommend addressing this as part of the broader contract validation quality improvements. Fixing this in isolation won’t be sustainable in the long run.

We can discuss this next week and plan the overall effort together.

For now, let’s hold off on this PR.

PR #7284, opened 2026-04-16
feat: add automated dependency update tooling and validation

jenny-s51:
## Description

Adds automated dependency management tooling for odh-dashboard:

**Dependabot configuration (.github/dependabot.yml)**

- Covers all npm packages in the monorepo with weekly version update schedules
- Groups security updates separately from version bumps
- Excludes upstream-synced packages to avoid conflicts with upstream lockfiles

**Dependency validation (.github/workflows/dependency-validation.yml)**

- Runs `npm audit --omit=dev` on head and base, comparing advisory IDs (not counts) to detect truly new vulnerabilities
- Triggers on `package-lock.json` changes only
- Supports skip-audit label bypass with re-trigger on label add/remove

**Dependabot auto-merge (.github/workflows/dependabot-auto-merge.yml)**

- Automatically applies lgtm + approved labels to Dependabot patch PRs when all CI checks pass

## Prerequisites

- [x] [openshift/release#77926](https://github.com/openshift/release/pull/77926) merged (adds `dependabot` to `trusted_apps`)

## How Has This Been Tested?

CI workflow config only — no runtime code changed. Validated against existing Dependabot PR [#7264](https://github.com/opendatahub-io/odh-dashboard/pull/7264) and Prow merge behavior.

## Self checklist

- [x] Manually tested and verified
- [x] Testing instructions provided
- [x] Tests added or explained (CI config — no testable logic)
- [x] Follows Best Practices

## Jira

Closes: [RHOAIENG-57877](https://redhat.atlassian.net/browse/RHOAIENG-57877), [RHOAIENG-57878](https://redhat.atlassian.net/browse/RHOAIENG-57878)
Epic: [RHOAIENG-57872](https://redhat.atlassian.net/browse/RHOAIENG-57872)

## Summary by CodeRabbit

* **Chores**
  * Weekly automated dependency updates enabled across frontend workspaces with grouped PRs for production/dev and separate security and PatternFly groups.
  * Dependabot exclusions for upstream-synced packages documented.
  * PRs labeled "dependencies" exempt from stale/close automation.

* **New Features**
  * Auto-label/approve and conditional auto-merge for patch-level dependency updates after successful checks, with scheduled sweep fallback.
  * New dependency-validation run that audits vulnerabilities per package and enforces override consistency.


.github/workflows/dependabot-auto-merge.yml
```diff
@@ -0,0 +1,89 @@
...
+  # workflow_run fires deterministically when the named workflow finishes.
+  workflow_run:
+    workflows: ["Dependency Validation"]
+    types:
+      - completed
+
+permissions:
+  contents: read
+  pull-requests: write
+  statuses: read
```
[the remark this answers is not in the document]
jenny-s51: According to Claude, this suggestion would make sense if this workflow had multiple jobs. But since auto-merge is the only job, the current structure is correct and simpler.

.github/workflows/dependabot-auto-merge.yml
```diff
@@ -0,0 +1,89 @@
...
+
+          echo "ready=true" >> "$GITHUB_OUTPUT"
+
+      - uses: actions/checkout@v4
+        if: steps.checks.outputs.ready == 'true'
+
+      - name: Fetch Dependabot metadata
+        if: steps.checks.outputs.ready == 'true'
+        id: metadata
+        uses: dependabot/fetch-metadata@v2
```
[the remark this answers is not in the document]
jenny-s51: According to Claude, every workflow in this repo currently uses unpinned tag references, not SHA-pinned. If the we want SHA pinning, it should be done as a repo-wide change, not as a single file change here.

.github/workflows/dependency-validation.yml
```diff
@@ -0,0 +1,279 @@
...
+            let body = header + limited.join('\n') + '\n';
+            if (changes.length > 50) {
+              body += '\n*...and ' + (changes.length - 50) + ' more changes.*\n';
+            }
+            fs.writeFileSync('dep-summary.md', body);
+          }
+          "
+
+      - name: Post PR comment
+        uses: actions/github-script@v7
```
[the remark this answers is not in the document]
jenny-s51: Same reasoning as the previous SHA-pinning finding.

.github/workflows/dependency-validation.yml
```diff
@@ -0,0 +1,279 @@
...
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.pull_request.base.sha }}
+          path: base
+
+      - name: Generate dependency diff summary
+        id: diff
+        env:
+          CHANGED_DIRS: ${{ needs.detect-dirs.outputs.dirs }}
+        run: |
```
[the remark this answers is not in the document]
jenny-s51: The dependency-summary job intentionally has different gating than the audit job - the two jobs serve different purposes and should have independent gating.

.github/workflows/dependency-validation.yml
```diff
@@ -0,0 +1,279 @@
...
+      - "packages/*/upstream/workspaces/frontend/package.json"
+      - "packages/*/upstream/workspaces/frontend/package-lock.json"
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+  pull-requests: write
```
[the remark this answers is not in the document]
jenny-s51: Ok - fixed

.github/workflows/dependency-validation.yml
```diff
@@ -0,0 +1,309 @@
...
+          AUDIT_DIR: ${{ matrix.dir }}
+          BASE_HIGH: ${{ steps.audit-base.outputs.high }}
+          BASE_CRITICAL: ${{ steps.audit-base.outputs.critical }}
+          HEAD_HIGH: ${{ steps.audit.outputs.high }}
+          HEAD_CRITICAL: ${{ steps.audit.outputs.critical }}
+        run: |
+          echo "::error::PR introduces new high/critical vulnerabilities in ${AUDIT_DIR}."
+          echo "Base: ${BASE_HIGH} high, ${BASE_CRITICAL} critical"
+          echo "Head: ${HEAD_HIGH} high, ${HEAD_CRITICAL} critical"
+          exit 1
```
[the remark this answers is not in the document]
jenny-s51: Validated with Claude - keeping count-based comparison is recommended since it correctly catches regressions without false positives from dependency reshuffling in grouped updates.

.github/dependabot.yml
```diff
@@ -9,3 +9,171 @@ updates:
...
+
+  - package-ecosystem: npm
+    directory: /packages/mlflow/frontend
+    schedule:
+      interval: weekly
+      day: monday
+    open-pull-requests-limit: 5
+    target-branch: "main"
+    labels:
+      # Keep in sync with exempt-pr-labels in .github/workflows/stale.yml
```
christianvogt, line 167: This comment is relevant to all mentions in this file for this label but there's only a single comment here.

.github/workflows/dependency-validation.yml
```diff
@@ -0,0 +1,267 @@
...
+              echo "$NEW_ADVISORIES"
+              echo 'ADVISORY_EOF'
+            } >> "$GITHUB_OUTPUT"
+            echo "Found $COUNT new advisory ID(s) in ${AUDIT_DIR}"
+          else
+            echo "has_new=false" >> "$GITHUB_OUTPUT"
+            echo "No new advisories introduced in ${AUDIT_DIR}."
+          fi
+
+      - name: Post bypass comment when ok-to-skip-audit is set
```
christianvogt: This comment will remain in the history even if the label "ok-to-skip-audit" is removed.

Maybe just stick to keeping the audit in the logs. Or consider adding the comment only once the PR merges?

.github/workflows/dependency-validation.yml
```diff
@@ -0,0 +1,267 @@
...
+          NEW_ADVISORIES: ${{ steps.evaluate.outputs.advisories }}
+        run: |
+          echo "::error::PR introduces ${NEW_COUNT} new advisory ID(s) in ${AUDIT_DIR} not present on the base branch:"
+          echo "$NEW_ADVISORIES"
+          echo ""
+          echo "If this is acceptable (e.g. not reachable, already tracked),"
+          echo "add the 'ok-to-skip-audit' label to bypass this check."
+          exit 1
+
+  override-consistency:
```
christianvogt: With NPM, package overrides in monorepo packages are ignored. Only the root override matters.
Is the purpose here to ensure that any monorepo package override is consistent in the root package.json?

Does it also check standalone packages such as `packages/maas/frontend/package.json` ? if so does it compare the monorepo - because that seems unnecessary as they are independent.

What is the the overall purpose here? An npm audit will catch issues regardless of overrides.

.github/dependabot.yml
jenny-s51: Thank you - updated and applied to all mentions in the file

.github/workflows/dependency-validation.yml
jenny-s51: Updated, does both now - the audit is in the logs and also added it as a post-merge comment, that's a great way to keep the history clean. Thanks @christianvogt.

.github/workflows/dependency-validation.yml
jenny-s51: Ah, yep, that makes a lot of sense, thank you for clarifying here that npm ignores those. It was meant as an extra safety net to catch mismatched override versions, but since that doesn't matter here, I've removed it.

.github/workflows/dependabot-label.yml
```diff
@@ -0,0 +1,25 @@
+name: Dependabot auto-label
+
+on:
+  pull_request_target:
+    types: [opened]
```
christianvogt: **Bug: stale label after Dependabot updates a grouped PR**

This only triggers on `opened`, but Dependabot can force-push updates to an existing grouped PR (which fires `synchronize`). If a group initially contained only patches and Dependabot later adds a minor bump, the `version-update:semver-patch` label persists and `dependabot-auto-merge.yml` will auto-approve a minor update without human review.

Suggest:
```yaml
types: [opened, synchronize]
```

The step should also remove the previous version-update:* label before applying the new one, since the update type can change between pushes.

.github/workflows/audit-bypass-notice.yml
```diff
@@ -0,0 +1,91 @@
...
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          REPO: ${{ github.repository }}
+          HEAD_SHA: ${{ github.event.pull_request.head.sha }}
+        run: |
+          RUN_ID=$(gh run list \
+            --repo "$REPO" \
+            --workflow dependency-validation.yml \
+            --json databaseId,headSha,conclusion \
+            --jq "[.[] | select(.headSha == \"$HEAD_SHA\" and .conclusion == \"success\")] | .[0].databaseId")
```
christianvogt, line 33: `gh run list` defaults to 20 items. In a busy repo, if >20 Dependency Validation runs occurred between when this PR's validation ran and when the PR was merged, the target run won't be found. The step silently exits with `found=false` and no bypass comment is posted — the paper trail is silently lost.

Fix: add `--limit 100` (or `--paginate` with `--jq` if the list could be very long).

.github/workflows/dependency-validation.yml
```diff
@@ -0,0 +1,195 @@
...
+        env:
+          AUDIT_DIR: ${{ matrix.dir }}
+        run: |
+          if [ ! -d "base/${AUDIT_DIR}" ]; then
+            echo "Directory base/${AUDIT_DIR} does not exist on the base branch (new package); treating as zero advisories"
+            mkdir -p "base/${AUDIT_DIR}"
+            touch "base/${AUDIT_DIR}/base-ids.txt"
+          else
+            cd "base/${AUDIT_DIR}"
+            npm install --ignore-scripts
```
christianvogt, line 118: Recent change stopped installing modules as a postinstall to npm install.
You need to also run `npm run install:modules`

.github/workflows/audit-bypass-notice.yml
```diff
@@ -0,0 +1,92 @@
...
+permissions:
+  contents: read
+  pull-requests: write
+  actions: read
+
+jobs:
+  post-bypass-notice:
+    if: >
+      github.event.pull_request.merged == true &&
+      contains(github.event.pull_request.labels.*.name, 'ok-to-skip-audit')
```
christianvogt, line 17: **[Info] audit-bypass-notice.yml still missing repository guard**
Unlike the other four workflows which all now have `github.repository == 'opendatahub-io/odh-dashboard'` guards, `audit-bypass-notice.yml` doesn't. It will execute (harmlessly) on forks. For consistency:

```
    if: >
      github.repository == 'opendatahub-io/odh-dashboard' &&
      github.event.pull_request.merged == true &&
      contains(github.event.pull_request.labels.*.name, 'ok-to-skip-audit')
```

.github/workflows/dependabot-label.yml
```diff
@@ -0,0 +1,38 @@
+name: Dependabot auto-label
+
+on:
+  pull_request_target:
+    types: [opened, synchronize]
```
christianvogt, line 5: **[Info] dependabot-label.yml synchronize trigger causes label churn**
Adding synchronize to the trigger types means every Dependabot rebase removes and re-adds version-update labels, creating noisy PR timeline events. The label removal step is defensive:

```
for label in version-update:semver-patch version-update:semver-minor version-update:semver-major; do
  if [ "$label" != "$UPDATE_TYPE" ]; then
    gh pr edit "$PR_URL" --remove-label "$label" 2>/dev/null || true
  fi
done
```
This is functionally correct (handles the rare case where a rebase changes the update type), but the common case (type stays the same) will still produce remove+add events for the same label. Consider checking if the correct label already exists before modifying.

jenny-s51:
Thank you @christianvogt - added the repo guard and the label check. Did another `/review` pass on the changes as well and caught a couple more things that I've just pushed here:

* Used `GH_TOKEN` for consistency across workflows
* Excluded tide from the check verification in auto-merge (since tide won't pass until the labels exist).

PR #7292, opened 2026-04-17
feat(RHOAIENG-29749): validate database name in model registry settings

manaswinidas:
https://issues.redhat.com/browse/RHOAIENG-29749

## Description

Adds client-side validation to the **Database** name field in the Model Registry create/edit modal, rejecting names that contain `?` characters.

The Model Registry backend ([`kubeflow/model-registry`](https://github.com/kubeflow/model-registry/blob/main/internal/datastore/embedmd/service.go)) appends `?charset=utf8mb4` to MySQL DSNs. If the database name itself contains `?`, the resulting connection string is malformed and the Model Registry fails to start. The backend already validates and rejects this (see [RHOAIENG-29726](https://issues.redhat.com/browse/RHOAIENG-29726)), but the user gets no feedback until after submission. This PR adds matching frontend validation to surface the error inline before form submission.

Screenshot 2026-04-17 at 12 15 51 PM

**Changes:**
- `const.ts`: Added `INVALID_DATABASE_CHARS` regex constant (`/[?]/`)
- `CreateModal.tsx`: Imported the constant, added `hasDatabaseInvalidChars` helper, updated `canSubmit()` to block submission, and added `validated` state + error helper text on the Database input
- `const.spec.ts`: Unit tests for the regex constant

## How Has This Been Tested?

- Unit tests: `cd frontend && npx jest --testPathPattern="modelRegistrySettings/__tests__/const" --no-coverage` (3 tests passing)
- Manual testing steps:
  1. Navigate to **Settings > Model Registry Settings**
  2. Click **Create model registry**, select **External database**
  3. In the **Database** field, type `mydb?tls=preferred` → red error border appears with message: *Database name must not contain "?" characters*; **Create** button is disabled
  4. Clear the field and blur → message changes to: *Database cannot be empty*
  5. Type a valid name like `model-registry` → no error, Create button enables (with other required fields filled)

## Test Impact

Added unit tests for the `INVALID_DATABASE_CHARS` constant in `frontend/src/pages/modelRegistrySettings/__tests__/const.spec.ts`.

## Request review criteria:

If you have UI changes:

## Summary by CodeRabbit

* **New Features**
  * Enhanced validation for database names in model registry settings with improved field-level validation feedback and error messaging.
  * Error messages now provide specific guidance for different validation failures: empty names and names with invalid characters.
  * Form submission is blocked for invalid database names, improving data integrity and user experience during configuration.


frontend/src/pages/modelRegistrySettings/__tests__/const.spec.ts
```diff
@@ -0,0 +1,18 @@
+import { INVALID_DATABASE_CHARS } from '#~/pages/modelRegistrySettings/const';
+
+describe('INVALID_DATABASE_CHARS', () => {
```
Taj010: The suite validates INVALID_DATABASE_CHARS in isolation. To make it stronger, we could add a small test that renders CreateModal (external DB path), enters a value with ?, and asserts that mr-database-error is present —similar to what CodeRabbit suggested.

frontend/src/pages/modelRegistrySettings/const.ts
```diff
@@ -8,6 +8,8 @@ export const DEFAULT_MYSQL_PORT = '3306';
 export const DEFAULT_POSTGRES_PORT = '5432';
 export const DEFAULT_DATABASE_NAME = 'model-registry';
 
+export const INVALID_DATABASE_CHARS = /[?]/;
```
Taj010: Nit: `/\?/ `is more commonly used for a single literal than `/[?]/` (a character class). Same behavior; mainly style.

Taj010 (commented):
Tested it locally - looks good, just some small suggestions

frontend/src/pages/modelRegistrySettings/__tests__/const.spec.ts
manaswinidas: Good point! The regex unit test covers the validation logic itself. Adding a full `CreateModal` component test is a larger effort since no test file exists for it yet — I'd rather track that as a follow-up rather than expand this PR's scope. The wiring between the regex check and the PF `validated` prop is straightforward glue code. Happy to create a follow-up ticket for `CreateModal` test coverage if you think it's worth it.

Taj010: Gotcha, we can leave the component test for later and pick it up when we add more checks like this, or if we decide we need broader test coverage here.

frontend/src/pages/modelRegistrySettings/CreateModal.tsx
```diff
@@ -391,6 +392,8 @@ const CreateModal: React.FC<CreateModalProps> = ({ onClose, refresh, modelRegist
 
   const hasContent = (value: string): boolean => !!value.trim().length;
 
+  const hasDatabaseInvalidChars = (value: string): boolean => INVALID_DATABASE_CHARS.test(value);
```
ppadti: would it make sense to move it to `utils.ts` and add few tests to it?

ppadti (changes requested):

frontend/src/pages/modelRegistrySettings/utils.ts
```diff
@@ -105,6 +105,8 @@ export const isOpenshiftCAbundleEnabled = (existingCertConfigMaps: ConfigSecretI
  * @param value - The port value to validate
  * @returns true if the port is valid, false otherwise
  */
+export const hasDatabaseInvalidChars = (value: string): boolean => /\?/.test(value);
```
ppadti: Thank you for adding this here.
One small comment: The comments present above this utils are for `isValidPort` util.

ppadti (changes requested):

.github/workflows/sync-branches.yml
```diff
@@ -0,0 +1,24 @@
...
+  sync:
+    runs-on: ubuntu-latest
+    
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v4
+      with:
+        ref: odh-release-poc
+        fetch-depth: 0
+        # Uses automatic GITHUB_TOKEN - no manual secrets needed!
```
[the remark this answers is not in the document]
manaswinidas: this was incorrectly rebased - ignore this

frontend/src/pages/modelRegistrySettings/CreateModal.tsx
```diff
@@ -586,14 +591,14 @@ const CreateModal: React.FC<CreateModalProps> = ({ onClose, refresh, modelRegist
                             value={database}
                             onBlur={() => setIsDatabaseTouched(true)}
                             onChange={(_e, value) => setDatabase(value)}
-                            validated={
-                              isDatabaseTouched && !hasContent(database) ? 'error' : 'default'
-                            }
+                            validated={isDatabaseTouched && hasDatabaseError ? 'error' : 'default'}
```
ppadti: This is the existing behaviour but bit confusing - when we enter all the details and finally enters the database name with ? it will make the update/create button disabled without any error, because the error will be shown when the user leaves the field and not onChnage

https://github.com/user-attachments/assets/82ac95f3-ee94-4cfa-bf16-f81c228222fa

ppadti (changes requested):

ppadti (approved):
/lgtm
Tested this locally and the database name validation works fine.

manaswinidas (commented):
/approve 

Self-approve with team lgtm

PR #7184, opened 2026-04-10
fix(cypress): fix correctness bugs, flakiness sources, and injection risk across test suite

christianvogt:
## Description

Systematic review of the Cypress test suite (`packages/cypress/`) identified verified correctness bugs, flakiness sources, and an injection risk across 12 files. Each issue was validated by reading the actual source code, checking Cypress/Chai documentation, and deep-reviewing every proposed fix before implementation.

### Correctness Bugs (9)
- **modelServingNim.cy.ts**: Missing `]` in CSS attribute selector — `querySelectorAll` throws `SyntaxError`
- **clusterStorage.ts**: `date-testid` typo instead of `data-testid` — selector never matches, `.should('not.exist')` passes vacuously
- **storageClass.ts**: Wrong condition guard (`expectedDescription` instead of `expectedAccessModes`) — access modes never validated
- **imageStreams.ts**: `Promise.all()` on Cypress chainables (not real Promises) — `imageInfos` can be incomplete; replaced with `cy.wrap().each()`
- **createConnectionType.cy.ts**: `.not('be.checked')` is jQuery element filtering (no-op), not a Cypress assertion — changed to `.should('not.be.checked')`
- **connectionTypes.cy.ts**: Redundant `cy.visitWithLogin` before intercepts registered; misleading test title
- **modelRegistrySettings.cy.ts**: `containSubset({})` always passes any object — added actual expected fields
- **featureDataSources.cy.ts** / **featureEntities.cy.ts**: `total_count: 1` with 2 mock projects

### Flakiness Fixes (2 files)
- **notebookServer.cy.ts**: `isRunning: false` intercept registered after stop button click — moved before the click
- **projectList.cy.ts**: `SelfSubjectAccessReview` intercept registered after `visit()` — moved before visit

### Security
- **modelRegistry.ts**: SQL escaping added to cleanup path (matching existing read path), plus shell double-quote escaping

## How Has This Been Tested?

- TypeScript type-check passes (`npx tsc --noEmit -p packages/cypress/tsconfig.json`)
- No linter errors introduced
- Mocked Cypress tests pass locally
- Each fix was validated by reading surrounding test context, Cypress/Chai docs, and verifying the fix does not change test semantics beyond correcting the identified bug

## Test Impact

This PR **only** modifies Cypress test files, page objects, and test utilities. No application source code is changed. The fixes correct silent test failures, reduce flakiness, and improve test reliability.

## Request review criteria:

Self checklist (all need to be checked):
[...]

PR #7319, opened 2026-04-20
fix(cypress): fix access mode verification in verifyStorageClassConfig

srtanish1992:
https://redhat.atlassian.net/browse/RHOAIENG-58950

## Description

The condition guard fix in #7184 changed `if (expectedDescription !== undefined)` to `if (expectedAccessModes !== undefined)` in `verifyStorageClassConfig`, which was correct -- but it exposed two pre-existing bugs in the access mode assertion that were previously hidden (dead code):
1. **`should('equal')` vs `should('deep.equal')`**: Chai's `equal` uses strict `===` reference equality, which always fails when comparing two different object instances even with identical contents. Changed to `deep.equal` for proper structural comparison.
2. **Missing `ReadWriteOncePod: false` in test expectation**: The storage class is provisioned with all four access mode keys (`ReadWriteOnce: true, ReadWriteMany: false, ReadWriteOncePod: false, ReadOnlyMany: false`). When the edit modal saves, it preserves the full state object including `ReadWriteOncePod: false`. The test expectation was missing this key.

This was caught by the nightly E2E run. It was not caught in #7184's CI because the smart test selection only ran `@ci-dashboard-regression-tags` (the `packages/cypress` package has no `e2eCiTags` and no `frontend/src/` files were changed to trigger `@StorageClassesCI`).

## How Has This Been Tested?

Ran the test locally and verified that it's passing successfully.

Screenshot 2026-04-20 at 11 27 48 AM

## Test Impact

This PR *is* the test fix. No additional tests needed -- the existing E2E test "An admin user can edit the access mode of a storage class" will now pass with the corrected assertion.

## Request review criteria:

If you have UI changes: 

## Release Notes

* **Tests**
  * Enhanced automated testing for storage class configurations with improved validation of access modes and more robust configuration assertion accuracy.


FedeAlonso:
Let's wait to the test-e2e to run, and if the tag "StorageClassesCI" passes, then let's merge it!

FedeAlonso:
image
It works :)

FedeAlonso (approved):

PR #7345, opened 2026-04-22
Fix federated mode theme defaulting to OS preference instead of odh default

nananosirova:
https://redhat.atlassian.net/browse/RHOAIENG-59002

## Description

By default, ODH dashboard’s theme is set to light mode. If a user's OS theme preference is set to dark mode, MLflow renders in dark mode and persists this to localStorage (_mlflow_dark_mode_toggle_enabled: true). This causes a mismatch in federated mode. 

The fix adds a mount-time `useEffect` in `ThemeProvider` that explicitly writes `_mlflow_dark_mode_toggle_enabled` to localStorage to match ODH's theme. This ensures the key exists before the MLflow remote reads it via `loadRemote`. Subsequent theme changes are already handled by `setAllThemes`.

The change is idempotent for existing users since they already have the key in localStorage from prior `setAllThemes` calls.

## How Has This Been Tested?

1. Cleared all localStorage keys (`odh.dashboard.ui.theme` and `_mlflow_dark_mode_toggle_enabled`).
2. Set OS to dark mode.
3. Loaded the dashboard, verified both ODH and MLflow render in light mode (ODH's default).
4. Toggled the theme to dark, verified both ODH and MLflow switch to dark.
5. Refreshed, verified both stay in dark.
6. Toggled back to light, refreshed, both stay in light.

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/app/ThemeContext.tsx
+++ b/frontend/src/app/ThemeContext.tsx
@@ -20,6 +20,11 @@ export const ThemeProvider: React.FC<ThemeProviderProps> = ({ children }) => {
   const [odhTheme, setOdhTheme] = useBrowserStorage<string>('odh.dashboard.ui.theme', 'light');
   const [, setMlflowTheme] = useBrowserStorage<boolean>(MLFLOW_DARK_MODE_KEY, odhTheme === 'dark');
 
+  React.useEffect(() => {
+    setMlflowTheme(odhTheme === 'dark');
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
+
   const setAllThemes = React.useCallback(
     (theme: string) => {
       setMlflowTheme(theme === 'dark');


frontend/src/app/ThemeContext.tsx
```diff
@@ -20,6 +20,11 @@ export const ThemeProvider: React.FC<ThemeProviderProps> = ({ children }) => {
   const [odhTheme, setOdhTheme] = useBrowserStorage<string>('odh.dashboard.ui.theme', 'light');
   const [, setMlflowTheme] = useBrowserStorage<boolean>(MLFLOW_DARK_MODE_KEY, odhTheme === 'dark');
 
+  React.useEffect(() => {
+    setMlflowTheme(odhTheme === 'dark');
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
DaoDaoNoCode, line 26: nit: you can drop the eslint-disable by including the real dependencies. `setMlflowTheme` is stable (its underlying `useCallback` deps are all stable), and if `odhTheme` changes post-mount, the duplicate write is harmless since `setAllThemes` already handles it. This avoids the lint suppression without changing behavior.

```suggestion
  }, [odhTheme, setMlflowTheme]);
```

DaoDaoNoCode (commented):
This fixes a real race condition where MLflow reads `_mlflow_dark_mode_toggle_enabled` from localStorage before ODH has written it, causing MLflow to fall back to `prefers-color-scheme` and potentially rendering in dark mode while ODH is in light mode. The approach of seeding localStorage on mount is sound -- `useBrowserStorage` only returns the default in-memory but never writes it, so the explicit `setMlflowTheme` call is necessary.

Local checks: lint pass, type-check pass, unit tests pass in CI, all Cypress mock tests pass. The E2E failure is a model-registry "storage and URI" test timing out on a create button -- unrelated to this change (infra flake).

One comment inline about the dependency array.

frontend/src/app/ThemeContext.tsx
DaoDaoNoCode: This AI review is invalid.

DaoDaoNoCode (approved):
/lgtm

PR #7385, opened 2026-04-24
test: prevent false-positive duplicate-name check in argo workflow version test

caponetto:
## Description

https://redhat.atlassian.net/browse/RHOAIENG-58263

## Problem

The Cypress mock test **"uploads fails with argo workflow"** in `pipelines.cy.ts` is flaky. It intermittently fails with:
```
AssertionError: Timed out retrying after 10000ms: expected '<button.pf-v6-c-button.pf-m-primary.pf-m-progress>' to be 'enabled'
```

## Root Cause

The `initIntercepts` function registers a generic intercept for the pipeline versions endpoint (`GET .../pipelines/:pipelineId/versions`) that returns mock versions **regardless of query parameters**.

When the user types a version name, `PipelineImportBase` triggers a debounced (500ms) duplicate-name check that calls `listPipelineVersions` with a `?filter=display_name==...` query param. The generic intercept catches this filtered request and returns mock versions. The application code (non-kubernetes path) trusts the server-side filter and only checks `duplicateVersions.length` — since the mock returns a non-empty list, `hasDuplicateName` is set to `true`, permanently disabling the submit button.

The test is flaky rather than always broken because it's a race between the 500ms debounce completing and Cypress reaching the `should('be.enabled')` assertion:
- **Passes**: Cypress outruns the debounce and clicks submit before the check completes.
- **Fails**: The debounce resolves first, the false positive disables the button, and `should('be.enabled')` times out.

## Fix

Add a more-specific `cy.intercept` after `initIntercepts` that matches version-list requests containing a `filter` query param and returns an empty list. Since Cypress matches the most recently registered intercept first, this takes priority over the generic one for filtered requests, while the unfiltered table-population requests continue to work as before.

## How Has This Been Tested?

Through CI runs.

## Test Impact

N/A

## Request review criteria:

If you have UI changes: 

## Summary by CodeRabbit

* **Tests**
  * Improved test reliability for Argo workflow pipeline uploads by fixing version lookup handling in the test suite.

### DIFF
--- a/packages/cypress/cypress/tests/mocked/pipelines/pipelines.cy.ts
+++ b/packages/cypress/cypress/tests/mocked/pipelines/pipelines.cy.ts
@@ -1099,6 +1099,18 @@ describe('Pipelines', () => {
 
   it('uploads fails with argo workflow', () => {
     initIntercepts({});
+
+    // Return empty results for filtered version requests (duplicate-name check)
+    // so the generic initIntercepts mock doesn't cause a false-positive.
+    cy.intercept(
+      {
+        method: 'GET',
+        pathname: `/api/service/pipelines/${projectName}/dspa/apis/v2beta1/pipelines/${initialMockPipeline.pipeline_id}/versions`,
+        query: { filter: /.*/ },
+      },
+      { pipeline_versions: [], total_size: 0 },
+    );
+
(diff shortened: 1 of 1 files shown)


DaoDaoNoCode (approved):
/lgtm

PR #7400, opened 2026-04-27
test(autorag): add AutoRAG Cypress E2E

jefho-rh:
https://redhat.atlassian.net/browse/RHOAIENG-58757
https://redhat.atlassian.net/browse/RHOAIENG-58393

## Description
Add comprehensive Cypress test coverage for AutoRAG with self-provisioned infrastructure support:

**E2E Tests** (`packages/cypress/cypress/tests/e2e/autorag/`):
- Optimization run — full flow: project provisioning, DSPA setup, OGX secret, document upload, run submission with max RAG patterns. Supports both external OGX and self-provisioned infrastructure (models, Milvus, OGX Distribution). Automatically detects and uses the correct CRD (OGXServer or legacy LlamaStackDistribution).
- Metric variations — same flow with different optimization metrics (answer correctness, faithfulness)
- Experiments & run management — listing, stopping, and deleting runs

**Page Objects** (`packages/cypress/cypress/pages/autorag/`):
- Split into `experimentsPage.ts`, `configurePage.ts`, `resultsPage.ts` with shared helpers (`submitRunSetup()`, `verifyResultsInteraction()`, `waitForRunCompletion()`)

**Self-Provisioned Infrastructure** (`packages/cypress/cypress/utils/oc_commands/`):
- `autoragInfra.ts` — orchestrates deployment of Milvus standalone, LLM model (vLLM CPU via KServe InferenceService), OGX Distribution (auto-detects OGXServer vs LlamaStackDistribution CRD, discovers image from operator CSV or controller-manager deployment), and credentials secret. Includes full cleanup utilities.
- `ogxNetworkPolicy.ts` — creates/removes NetworkPolicies for cross-namespace OGX access with stale policy cleanup
- `ogxSecret.ts` — creates K8s secrets with OGX connection credentials (shell-escaped, log-suppressed)

**Source Changes**:
- Added `data-testid` attributes to AutoRAG components (configure page, create form, evaluation select, run-in-progress state, action buttons, PatternDetailsModal)
- Refactored `autoX.ts` feature flag utilities from AutoML-specific to generic `isFeatureFlagEnabled`/`setFeatureFlag` functions (reused by both AutoML and AutoRAG)
- Shell-escaped AWS credentials in `s3Cleanup.ts` to prevent injection

**Test Data & Utilities**:
- Test document and evaluation data for document upload
- YAML fixtures for all test configurations
- K8s resource fixtures for LLM serving runtime, inference service, and Milvus standalone deployment
- OGX secret creation utility (`ogxSecret.ts`)
- S3 cleanup utility (`s3Cleanup.ts`) for test teardown
- Shared pipeline provisioning utilities (`autoXPipelines.ts`)
- Added `OGX_URL` and `OGX_API_KEY` to Cypress test config and `test-variables.yml.example`

image

Screenshot 2026-04-28 at 3 53 01 PM

---
Jenkins Test Run(`job/dashboard/job/dashboard-e2e-tests/1066/`)
image

## Self-Provisioned Infrastructure Mode

The optimization E2E test (`testAutoragOptimization.cy.ts`) supports two modes:

### External OGX (default when `OGX_URL` is set)
Uses a pre-existing OGX instance. Set `OGX_URL` and optionally `OGX_API_KEY` in `test-variables.yml`.

### Self-Provisioned (when `OGX_URL` is empty/unset)
Deploys the full stack programmatically in the test namespace:
1. **OGX operator** — ensures the OGX component (auto-detected as `ogx` or legacy `llamastackoperator`) is `Managed` in the DataScienceCluster (restores original state in teardown)
2. **LLM model** — deploys a vLLM CPU serving runtime + InferenceService (configurable via `CYPRESS_AUTORAG_VLLM_CPU_IMAGE` and `CYPRESS_AUTORAG_LLM_MODEL_URI`)
3. **Milvus** — deploys standalone Milvus with embedded etcd (configurable via `CYPRESS_AUTORAG_MILVUS_IMAGE`)
4. **OGX Distribution** — creates a ConfigMap with `run.yaml` config and the appropriate CRD (`OGXServer` on newer clusters, `LlamaStackDistribution` on legacy clusters) using the operator's discovered image (overridable via `CYPRESS_AUTORAG_OGX_IMAGE`)
5. **Credentials** — creates a K8s secret pointing to the in-cluster OGX service URL

All resources are cleaned up independently in the `after()` hook (each cleanup is resilient to partial failures).

### OGX / LlamaStack CRD Compatibility

The infrastructure code auto-detects which CRDs are active on the cluster:

| Component | New (OGX) | Legacy (LlamaStack) |
|-----------|-----------|---------------------|
| DSC component | `ogx` | `llamastackoperator` |
| Readiness condition | `OGXReady` | `LlamaStackOperatorReady` |
| Distribution CRD | `OGXServer` (`ogx.io/v1beta1`) | `LlamaStackDistribution` (`llamastack.io/v1alpha1`) |
| Operator deployment | `ogx-k8s-operator-controller-manager` | `llama-stack-k8s-operator-controller-manager` |
| CSV image key | `odh_ogx_core_image` | `odh_llama_stack_core_image` |

Detection logic: try OGX first, fall back to LlamaStack. Works on both ODH and RHOAI clusters regardless of operator version.

### Disconnected cluster support
All container images default to public registries but can be overridden via `CYPRESS_AUTORAG_*` env vars for air-gapped environments.

## CI Concurrency
The E2E tests are safe for parallel execution — no `@NonConcurrent` tag required:

- Each run creates a unique project via `generateTestUUID()`, so multiple CI runners on the same cluster won't collide
- All resources (DSPA secret, OGX secret, uploaded documents, pipeline run) are scoped to that unique project
- The `after()` hook deletes the project on completion
- No shared mutable state between test runs

## Connected & Disconnected Cluster Support

The E2E tests work on both **connected** and **disconnected** (air-gapped) OpenShift clusters.

The DSPA (Data Science Pipelines Application) needs an S3-compatible endpoint for object storage. On connected clusters, the DSPA can reach the external S3 endpoint directly. On disconnected clusters, the DSPA pods cannot reach external endpoints, so an in-cluster S3 service (e.g., MinIO) is required.

The `provisionProjectForAutoX()` utility in `autoXPipelines.ts` handles this via the `CYPRESS_DSPA_S3_ENDPOINT` environment variable:

- **Connected clusters (default):** Uses the external S3 `ENDPOINT` from `test-variables.yml` (e.g., `https://s3.us-east-1.amazonaws.com`) for both the Cypress-side file upload and the DSPA manifest. No additional configuration needed.
- **Disconnected clusters:** Set `CYPRESS_DSPA_S3_ENDPOINT` to the in-cluster S3 service URL. This overrides only the DSPA's S3 endpoint so pipeline pods can reach object storage internally, while the data connection secret still uses the external endpoint for Cypress-side operations (file upload via the dashboard UI).

### Mock Tests (Results Page)

**HEADLESS**
```bash
cd packages/autorag/frontend
npm run test:cypress-ci -- --spec "**/autoragResults.cy.ts"
```

### E2E Submission Flow

**Prerequisites:**
- Connected to OpenShift cluster (`oc whoami` to verify)
- `test-variables.yml` configured with OGX URL and API key (or leave `OGX_URL` empty for self-provisioned mode)

#### Setting up test-variables.yml

Copy the example file and fill in your cluster credentials:

```bash
cd packages/cypress
cp test-variables.yml.example test-variables.yml
```

Refer to `test-variables.yml.example` for the required fields and fill in your cluster-specific values (users, S3 credentials, OGX URL/API key, etc.).

**Important:** To test locally with your component changes, set `ODH_DASHBOARD_URL` to your local frontend:
```yaml
ODH_DASHBOARD_URL: "http://localhost:4010"
```

#### Running E2E Tests

**Terminal 1:** Start AutoRAG federated frontend + BFF
```bash
cd packages/autorag
make dev-start-federated
```

**Terminal 2:** Start main dashboard (backend + frontend)
```bash
npm run dev
```

**Terminal 3:** Open Cypress and select test
```bash
cd frontend
export CY_TEST_CONFIG=<absolute-path-to>/packages/cypress/test-variables.yml
npm run cypress:open
```

## How Has This Been Tested?
- E2E tests run against a live RHOAI cluster with both external OGX and self-provisioned infrastructure modes
- Lint and type-check pass across the full monorepo

## Test Impact
New tests only — no existing tests modified.
- 3 E2E test files covering experiments management, metric variations, and full optimization runs
- 1 mock test file covering results page UI interactions
- 3 page object files with reusable helpers
- Infrastructure provisioning utilities for self-contained E2E runs

## Request review criteria:

If you have UI changes:
- N/A — test-only changes with minor `data-testid` additions to source components


packages/autorag/frontend/src/app/components/run-results/PatternDetailsModal.tsx
```diff
@@ -490,7 +490,11 @@ const PatternDetailsModal: React.FC<PatternDetailsModalProps> = ({
                     >
                       <DropdownList>
                         {patterns.map((pattern, i) => (
-                          <DropdownItem key={i} value={i}>
+                          <DropdownItem
+                            key={i}
+                            value={i}
+                            data-testid={`pattern-option-${pattern.name}`}
+                          >
```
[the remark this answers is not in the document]
jefho-rh: Pattern names come from backend API data and are expected to be well-formed identifiers

packages/cypress/cypress/fixtures/resources/yaml/automl_dspa.yaml
```diff
@@ -0,0 +1,50 @@
...
+    injectDefaultScript: true
+  database:
+    disableHealthCheck: false
+    mariaDB:
+      deploy: true
+      pipelineDBName: mlpipeline
+      pvcSize: 10Gi
+      username: mlpipeline
+  dspVersion: v2
+  podToPodTLS: false
```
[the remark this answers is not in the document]
jefho-rh: This is a test fixture for provisioning ephemeral E2E test projects that are deleted in the after() hook. podToPodTLS: false is intentional to simplify DSPA setup in short-lived test namespaces. Enabling TLS would require certificate provisioning that adds complexity without security benefit in this context, since these namespaces exist only for the duration of a test run.

packages/cypress/cypress/tests/e2e/autorag/testAutoragMetricVariations.cy.ts
```diff
@@ -0,0 +1,123 @@
...
+import type { AutoragTestData } from '../../../types';
+import {
+  autoragExperimentsPage,
+  autoragConfigurePage,
+  autoragResultsPage,
+} from '../../../pages/autorag';
+
+const uuid = generateTestUUID();
+
+describe('AutoRAG Metric Variations E2E', { testIsolation: false }, () => {
```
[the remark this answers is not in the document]
jefho-rh: testIsolation: false is an intentional and necessary pattern for E2E tests that share expensive cluster setup (project creation, DSPA provisioning, LlamaStack secret). Each describe block creates a unique project via generateTestUUID(), so there's no cross-suite collision. The inter-test dependency within a single describe is by design — the second test reuses a document already uploaded to S3 to avoid redundant uploads and save ~2 minutes of test time.

packages/cypress/cypress/tests/e2e/autorag/testAutoragOptimization.cy.ts
```diff
@@ -0,0 +1,68 @@
...
+import { provisionProjectForAutoX } from '../../../utils/autoXPipelines';
+import { createLlamaStackSecret } from '../../../utils/oc_commands/llamaStackSecret';
+import { retryableBefore } from '../../../utils/retryableHooks';
+import { generateTestUUID } from '../../../utils/uuidGenerator';
+import type { AutoragTestData } from '../../../types';
+import { autoragConfigurePage, autoragResultsPage } from '../../../pages/autorag';
+
+const uuid = generateTestUUID();
+
+describe('AutoRAG Optimization E2E', { testIsolation: false }, () => {
```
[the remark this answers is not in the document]
jefho-rh: this is a multi-step E2E flow (submit → wait for completion → verify results) that intentionally shares state within a single describe. Each describe block is independently runnable with its own unique project. Splitting into a single it would create an unreviewably long test.

packages/cypress/cypress/utils/autoXPipelines.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  const dspaReplacements = {
+    DSPA_SECRET_NAME: dspaSecretName,
+    NAMESPACE: projectName,
+    AWS_S3_BUCKET: bucketConfig.NAME,
+    AWS_REGION: bucketConfig.REGION,
+    AWS_S3_HOST: host,
+    AWS_S3_SCHEME: scheme,
+  };
+  createDSPA(dspaReplacements, 'resources/yaml/automl_dspa.yaml');
```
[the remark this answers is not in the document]
jefho-rh: The dspaReplacements object uses a plain object literal (not the DspaReplacements type) because it passes additional fields (AWS_S3_HOST, AWS_S3_SCHEME) that the standard type doesn't include. The createDSPA function accepts a generic replacements object for template substitution, so strict typing isn't required here.

packages/cypress/cypress/utils/oc_commands/ogxSecret.ts
```diff
@@ -0,0 +1,26 @@
...
+  secretName: string,
+  baseUrl: string,
+  apiKey: string,
+): void => {
+  const escapeShellSingleQuote = (s: string) => s.replace(/'/g, "'\\''");
+
+  cy.exec(
+    `oc create secret generic ${secretName} -n ${namespace} ` +
+      `--from-literal=LLAMA_STACK_CLIENT_BASE_URL='${escapeShellSingleQuote(baseUrl)}' ` +
+      `--from-literal=LLAMA_STACK_CLIENT_API_KEY='${escapeShellSingleQuote(apiKey)}'`,
```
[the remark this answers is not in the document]
jefho-rh: namespace and secretName come from hardcoded YAML fixture files checked into the repo (e.g., cy-lls-secret, cy-e2e-autorag-opt-<uuid>), not from user input. These values are controlled by test authors.

packages/autorag/frontend/src/bootstrap.tsx
```diff
@@ -19,6 +20,16 @@ import App from '~/app/App';
...
 
+const queryClient = new QueryClient({
+  defaultOptions: {
+    queries: {
+      refetchOnWindowFocus: false,
+      retry: 1,
+      staleTime: 30000,
+    },
+  },
+});
```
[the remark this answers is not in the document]
jefho-rh: QueryClientProvider is needed for the standalone dev/test bootstrap. Other modules with BFFs follow the same pattern. Each standalone app wraps in its own QueryClientProvider since it runs independently of the host dashboard.

packages/cypress/cypress/tests/e2e/autorag/testAutoragExperimentsAndRunManagement.cy.ts
```diff
@@ -0,0 +1,109 @@
...
+        provisionProjectForAutoX(projectName, testData.dspaSecretName, testData.awsBucket);
+
+        const llamaStackUrl = Cypress.env('LLAMA_STACK_URL') as string;
+        const llamaStackApiKey = (Cypress.env('LLAMA_STACK_API_KEY') as string) || '';
+        createLlamaStackSecret(
+          projectName,
+          testData.llamaStackSecretName,
+          llamaStackUrl,
+          llamaStackApiKey,
+        );
```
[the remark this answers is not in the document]
jefho-rh: LLAMA_STACK_URL is loaded from test-variables.yml via testConfig.ts and set as a Cypress env var. If it's missing, oc create secret --from-literal=...=undefined fails immediately with a clear oc error.

packages/cypress/cypress/tests/e2e/autorag/testAutoragMetricVariations.cy.ts
```diff
@@ -0,0 +1,123 @@
...
+        provisionProjectForAutoX(projectName, testData.dspaSecretName, testData.awsBucket);
+
+        const llamaStackUrl = Cypress.env('LLAMA_STACK_URL') as string;
+        const llamaStackApiKey = (Cypress.env('LLAMA_STACK_API_KEY') as string) || '';
+        createLlamaStackSecret(
+          projectName,
+          testData.llamaStackSecretName,
+          llamaStackUrl,
+          llamaStackApiKey,
+        );
```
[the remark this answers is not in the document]
jefho-rh: LLAMA_STACK_URL is loaded from test-variables.yml via testConfig.ts and set as a Cypress env var. If it's missing, oc create secret --from-literal=...=undefined fails immediately with a clear oc error.

packages/cypress/cypress/tests/e2e/autorag/testAutoragOptimization.cy.ts
```diff
@@ -0,0 +1,68 @@
...
+      provisionProjectForAutoX(projectName, testData.dspaSecretName, testData.awsBucket);
+
+      const llamaStackUrl = Cypress.env('LLAMA_STACK_URL') as string;
+      const llamaStackApiKey = (Cypress.env('LLAMA_STACK_API_KEY') as string) || '';
+      createLlamaStackSecret(
+        projectName,
+        testData.llamaStackSecretName,
+        llamaStackUrl,
+        llamaStackApiKey,
+      );
```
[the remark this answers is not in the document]
jefho-rh: LLAMA_STACK_URL is loaded from test-variables.yml via testConfig.ts and set as a Cypress env var. If it's missing, oc create secret --from-literal=...=undefined fails immediately with a clear oc error.

packages/cypress/cypress/utils/oc_commands/ogxSecret.ts
```diff
@@ -0,0 +1,26 @@
...
+    { failOnNonZeroExit: true, log: false },
```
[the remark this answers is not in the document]
jefho-rh: retryableBefore creates a fresh project via createCleanProject() each time, so the secret never pre-exists within a single test run. If the project already exists from a prior failed run, createCleanProject deletes and recreates it. So oc create won't hit a conflict in practice.

packages/cypress/cypress/utils/oc_commands/s3Cleanup.ts
```diff
@@ -0,0 +1,45 @@
...
+    `oc run ${podName} -n ${namespace} ` +
+      `--image=amazon/aws-cli:latest ` +
+      `--restart=Never --rm --attach ` +
+      `--env=AWS_ACCESS_KEY_ID=${AWS_BUCKETS.AWS_ACCESS_KEY_ID} ` +
+      `--env=AWS_SECRET_ACCESS_KEY=${AWS_BUCKETS.AWS_SECRET_ACCESS_KEY} ` +
+      `--env=AWS_DEFAULT_REGION=${bucketConfig.REGION} ` +
+      `-- s3 rm s3://${bucketConfig.NAME}/ --recursive ` +
+      `--endpoint-url ${bucketConfig.ENDPOINT} ` +
+      `--exclude '*' --include '${prefix}'`,
+    { failOnNonZeroExit: false, log: false, timeout: 120000 },
```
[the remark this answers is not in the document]
jefho-rh: This pod runs in an ephemeral test namespace that's deleted immediately after by deleteOpenShiftProject(). The amazon/aws-cli is the official AWS CLI image. Pinning to a digest would prevent automatic security patches and add maintenance burden for a test-only cleanup utility that runs for a few seconds.

packages/cypress/cypress/utils/testConfig.ts
```diff
@@ -84,6 +84,10 @@ const NGC_API_KEY = testConfig?.NGC_API_KEY;
 const OCI_SECRET_VALUE = testConfig?.OCI_SECRET_VALUE;
 const OCI_MODEL_URI = testConfig?.OCI_MODEL_URI;
 
+// LlamaStack connection settings
+const LLAMA_STACK_URL = testConfig?.LLAMA_STACK_URL;
+const LLAMA_STACK_API_KEY = testConfig?.LLAMA_STACK_API_KEY;
```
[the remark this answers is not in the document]
jefho-rh: LlamaStack settings are only needed for AutoRAG E2E tests, which always run with CY_TEST_CONFIG pointing to test-variables.yml. Adding a process.env fallback would silently mask misconfiguration. The existing pattern (returning undefined when not configured) is consistent with how NGC_API_KEY, OCI_SECRET_VALUE, and other optional config keys are handled in this same file.

packages/autorag/frontend/src/__tests__/cypress/cypress/tests/mocked/run-results/autoragResults.cy.ts
```diff
@@ -0,0 +1,306 @@
...
+
+      autoragResultsPage.findTab('embedding').should('exist').click();
+      autoragResultsPage.findPatternDetailsModal().should('contain.text', 'cosine');
+
+      autoragResultsPage.findTab('retrieval').should('exist').click();
+      autoragResultsPage.findPatternDetailsModal().should('contain.text', 'window');
+
+      autoragResultsPage.findTab('generation').should('exist').click();
+      autoragResultsPage.findPatternDetailsModal().should('contain.text', '{document}');
+
```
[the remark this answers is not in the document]
jefho-rh: PatternFly v6 tabs unmount inactive tab content by default. Only the active panel is rendered in the DOM. So .should('contain.text', 'milvus') after clicking the Vector store tab is already scoped to the active panel's content. There's no risk of matching text from an inactive tab.

packages/autorag/frontend/src/__tests__/cypress/cypress/tests/mocked/run-results/autoragResults.cy.ts
```diff
@@ -0,0 +1,306 @@
...
+      // Open selector and switch to a different pattern
+      autoragResultsPage.findPatternSelectorDropdown().click();
+      autoragResultsPage.findPatternSelectorOption(sortedPatterns[1].name).click();
+
+      autoragResultsPage.findPatternDetailsModal().should('be.visible');
+      autoragResultsPage
+        .findPatternSelectorDropdown()
+        .invoke('text')
+        .should('match', new RegExp(sortedPatterns[1].name.replace(/(\D)(\d)/, '$1.?$2')));
+    });
```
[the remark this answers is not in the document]
jefho-rh: Pattern names are backend-generated identifiers following a strict naming convention (e.g., milvus_recursive_bge_large). They don't contain regex metacharacters like (, ), or +. The .replace(/(\D)(\d)/, '$1.?$2') is intentionally flexible to handle minor formatting differences in the display name (e.g., bge-large2 vs bge-large 2).

packages/cypress/cypress/tests/e2e/autorag/testAutoragMetricVariations.cy.ts
```diff
@@ -0,0 +1,126 @@
...
+import { retryableBefore } from '../../../utils/retryableHooks';
+import { generateTestUUID } from '../../../utils/uuidGenerator';
+import type { AutoragTestData } from '../../../types';
+import {
+  autoragExperimentsPage,
+  autoragConfigurePage,
+  autoragResultsPage,
+} from '../../../pages/autorag';
+
+const uuid = generateTestUUID();
```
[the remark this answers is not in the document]
jefho-rh: generateTestUUID() is a shared utility used by all E2E suites in the repo (AutoML, pipelines, etc.). Its behaviour isn't introduced or changed by this PR. In practice, collision is avoided because each spec uses a unique projectNamePrefix (cy-e2e-autorag-opt, cy-e2e-autorag-metrics, cy-e2e-autorag-mgmt), so full project names are distinct even with the same UUID suffix.

packages/cypress/cypress/utils/autoXPipelines.ts
```diff
@@ -0,0 +1,81 @@
...
+ * Parse an S3 endpoint URL into host and scheme components.
+ * e.g. "https://s3.us-west.cloud-object-storage.appdomain.cloud/" → { host: "s3.us-west...", scheme: "https" }
+ */
+const parseS3Endpoint = (endpoint: string): { host: string; scheme: string } => {
+  try {
+    const url = new URL(endpoint);
+    return { host: url.host, scheme: url.protocol.replace(':', '') };
+  } catch {
+    return { host: endpoint.replace(/\/+$/, ''), scheme: 'https' };
+  }
```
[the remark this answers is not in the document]
jefho-rh: DSPA_S3_ENDPOINT is an optional override only used for disconnected clusters. It's set intentionally by operators who know their in-cluster endpoint. When unset, Cypress.env() returns undefined and ?? correctly falls back to bucketConfig.ENDPOINT. An explicitly empty string (CYPRESS_DSPA_S3_ENDPOINT="") would fail immediately at DSPA creation with a clear error about an invalid endpoint. The catch fallback for schemeless URLs (e.g., minio.ns.svc:9000) defaulting to https is intentional. The DSPA template separates host and scheme, so this preserves the host while using a safe default scheme.

packages/cypress/cypress/tests/e2e/autorag/testAutoragOptimization.cy.ts
```diff
@@ -0,0 +1,128 @@
...
+
+    // Restore operator to previous state if we changed it
+    if (selfProvisioned && !operatorWasManaged) {
+      resetOgxOperator();
+    }
+  });
+
+  it(
+    'Can create and submit a full AutoRAG optimization run',
+    { tags: ['@AutoRAG', '@AutoRAGRegression'] },
```
FedeAlonso:
```suggestion
    { tags: ['@AutoRAG', '@AutoRAGRegression', '@AutoRAGOptimization'] },
```

packages/cypress/cypress/pages/autorag/configurePage.ts
```diff
@@ -0,0 +1,261 @@
...
+  /**
+   * Common setup for submitting an AutoRAG run.
+   *
+   * Handles: login, wait for DSPA, navigate to experiments, create run,
+   * fill name/description, select OGX secret, select S3 connection,
+   * upload document, and select first available models + vector store.
+   *
+   * After this, optionally configure metric/patterns, then call `submitRun()`.
+   */
+  submitRunSetup(testData: AutoragTestData, projectName: string, uuid: string) {
```
sridarna: This is not suitable in Pages folder. These steps should be in e2e file.
If these steps are same in multiple e2e tests. You could include the UI flow in 1 or 2 e2e tests. For others could you able to create from CLI comands?

packages/cypress/cypress/pages/autorag/configurePage.ts
```diff
@@ -0,0 +1,261 @@
...
+
+    cy.step('Navigate to AutoRAG experiments page');
+    autoragExperimentsPage.visit(projectName);
+
+    cy.step('Wait for pipeline server to be fully ready and click Create run');
+    // Pipeline server startup can take up to 2 minutes on first provisioning
+    cy.findByTestId('app-page-title', { timeout: 120000 }).should('be.visible');
+    // Use header button if runs already exist, otherwise empty state button
+    cy.get('body').then(($body) => {
+      if ($body.find('[data-testid="autorag-header-create-run-button"]').length) {
```
sridarna: We need to create page object for each data-testid

packages/cypress/cypress/pages/autorag/configurePage.ts
```diff
@@ -0,0 +1,261 @@
...
+   */
+  submitRun() {
+    cy.step('Submit the form');
+    this.findCreateRunButton().click();
+
+    cy.step('Verify redirect to results page');
+    cy.url().should('include', '/gen-ai-studio/autorag/results/');
+
+    cy.step('Verify the run is in progress');
+    autoragResultsPage.findRunInProgressMessage().should('be.visible');
```
sridarna: verifications should not be defined in pages folder. these go into the e2e flow.

packages/autorag/frontend/src/__tests__/cypress/cypress/pages/autoragResults.ts
```diff
@@ -0,0 +1,117 @@
+class AutoragResultsPage {
```
sridarna: This is not the right place. please add it here packages/cypress/cypress/pages

packages/autorag/frontend/src/__tests__/cypress/cypress/tests/mocked/run-results/autoragResults.cy.ts
```diff
@@ -0,0 +1,306 @@
+/* eslint-disable camelcase -- BFF API uses snake_case field names */
```
sridarna: The mocks tests should be added here packages/cypress/cypress/tests/mocked.
I guess the CI checks pick from here.

packages/cypress/cypress/pages/autorag/resultsPage.ts
```diff
@@ -0,0 +1,234 @@
...
+   * | Pattern information  | pattern_information  | Key-value fields + scores    |
+   * | Vector store         | vector_store         | Settings key-value list      |
+   * | Chunking             | chunking             | Settings key-value list      |
+   * | Embedding            | embedding            | Settings key-value list      |
+   * | Retrieval            | retrieval            | Settings key-value list      |
+   * | Generation           | generation           | Settings key-value list      |
+   * | Sample Q&A           | sample_qa            | Conditional (eval results)   |
+   */
+  verifyResultsInteraction() {
+    cy.step('Verify leaderboard has at least one pattern row');
```
sridarna: The validations are not suitable in Pages. These should be moved in e2e file

packages/cypress/cypress/tests/e2e/autorag/testAutoragExperimentsAndRunManagement.cy.ts
```diff
@@ -0,0 +1,120 @@
...
+        isAutoragEnabled().then((wasEnabled) => {
+          autoragWasEnabled = wasEnabled;
+        }),
+      )
+      .then(() => setAutoragEnabled(true))
+      .then(() => {
+        provisionProjectForAutoX(projectName, testData.dspaSecretName, testData.awsBucket);
+        allowOgxAccess(projectName);
+
+        const ogxUrl = Cypress.env('OGX_URL') as string | undefined;
```
sridarna: Have we added OGX_URL in test-variables.yml file ?

packages/cypress/cypress/tests/e2e/autorag/testAutoragExperimentsAndRunManagement.cy.ts
```diff
@@ -0,0 +1,120 @@
...
+  after(() => {
+    if (!autoragWasEnabled) {
+      setAutoragEnabled(false);
+    }
+    removeOgxAccess(projectName);
+    deleteS3TestFiles(projectName, testData.awsBucket, `*${uuid}*`);
+    deleteOpenShiftProject(projectName, { wait: false, ignoreNotFound: true });
+  });
+
+  it(
```
sridarna, line 89: This E2E flow validates small use case, Which should be achieved through the mock tests

packages/cypress/cypress/tests/e2e/autorag/testAutoragExperimentsAndRunManagement.cy.ts
```diff
@@ -0,0 +1,120 @@
...
+      autoragResultsPage.findRunsTable().should('contain.text', testData.runName);
+    },
+  );
+
+  it(
+    'Can stop a running experiment from the results page',
+    { tags: ['@AutoRAG', '@AutoRAGRegression'] },
+    () => {
+      cy.step('Click on the run to go to results page');
+      autoragResultsPage.findRunsTable().contains(testData.runName).click();
```
sridarna, line 104: If this test depends on the above created run, then we should not create a  new test. If the above tests fails to create/submit run. Then these tests also fails.
It is correct to combine both into one as this validates a real scenario.

packages/cypress/cypress/tests/e2e/autorag/testAutoragExperimentsAndRunManagement.cy.ts
```diff
@@ -0,0 +1,120 @@
...
+      cy.step('Verify run is in progress');
+      autoragResultsPage.findRunInProgressMessage().should('be.visible');
+
+      cy.step('Click stop button and confirm');
+      autoragResultsPage.findStopRunButton().click();
+      autoragResultsPage.findStopRunModal().should('be.visible');
+      autoragResultsPage.findConfirmStopRunButton().click();
+
+      cy.step('Verify run status shows as canceling, canceled, or failed');
+      autoragResultsPage
```
sridarna: Didn't add a wait here. Add a step that validates from backend that the run is stopped. then validate the UI

packages/cypress/cypress/tests/e2e/autorag/testAutoragMetricVariations.cy.ts
```diff
@@ -0,0 +1,137 @@
...
+
+      cy.step('Select answer_correctness optimization metric');
+      autoragConfigurePage.findOptimizationMetricSelect().click();
+      autoragConfigurePage.findMetricOption('answer_correctness').click();
+
+      cy.step('Set max RAG patterns');
+      autoragConfigurePage.setMaxRagPatterns(testData.maxRagPatterns);
+
+      autoragConfigurePage.submitRun();
+
```
sridarna, line 102: After submit, can we add few validations from backend that a resource is created and succeeded?

packages/cypress/cypress/tests/e2e/autorag/testAutoragMetricVariations.cy.ts
```diff
@@ -0,0 +1,137 @@
...
+      cy.step('Click Next');
+      autoragConfigurePage.findNextButton().click();
+
+      cy.step('Select S3 connection');
+      autoragConfigurePage.findSecretSelector().click();
+      autoragConfigurePage.findSecretSelector().type(testData.s3SecretName);
+      autoragConfigurePage.findSelectOption(new RegExp(testData.s3SecretName, 'i')).click();
+
+      cy.step('Browse and select the document uploaded by the first test');
+      const uploadFileName = `${testData.documentFile.replace('.txt', '')}-${uuid}.txt`;
```
sridarna: declare all the constant at one place.

packages/cypress/cypress/tests/e2e/autorag/testAutoragMetricVariations.cy.ts
```diff
@@ -0,0 +1,137 @@
...
+      autoragConfigurePage.findFileExplorerSearch().type(uploadFileName);
+      autoragConfigurePage
+        .findFileExplorerTable()
+        .contains('td', uploadFileName)
+        .should('be.visible')
+        .click();
+      autoragConfigurePage.findFileExplorerSelectBtn().click();
+
+      cy.step('Upload evaluation dataset JSON');
+      const evalFileName = `${testData.evaluationFile.replace('.json', '')}-${uuid}-faith.json`;
```
sridarna: declare all the constant at one place.

packages/cypress/cypress/tests/e2e/autorag/testAutoragMetricVariations.cy.ts
```diff
@@ -0,0 +1,137 @@
...
+      autoragConfigurePage
+        .findEvaluationFileInput()
+        .selectFile(
+          { contents: `resources/autorag/${testData.evaluationFile}`, fileName: evalFileName },
+          { force: true },
```
sridarna: create a variable for file location at top and use the variable here.

packages/cypress/cypress/tests/e2e/autorag/testAutoragOptimization.cy.ts
```diff
@@ -0,0 +1,128 @@
...
+    'Can create and submit a full AutoRAG optimization run',
+    { tags: ['@Smoke', '@SmokeSet4', '@AutoRAG', '@AutoRAGCI'] },
+    () => {
+      autoragConfigurePage.submitRunSetup(testData, projectName, uuid);
+
+      cy.step('Set max RAG patterns to minimize run time');
+      autoragConfigurePage.setMaxRagPatterns(testData.maxRagPatterns);
+
+      autoragConfigurePage.submitRun();
+    },
```
sridarna, line 118: Please add validation that the run is submitted successfully and a resource is created in backend. 
Include validation in UI. If it updates status.

packages/cypress/cypress/tests/e2e/autorag/testAutoragOptimization.cy.ts
```diff
@@ -0,0 +1,128 @@
...
+  );
+
+  it(
+    'Verify optimization run completes with leaderboard',
+    { tags: ['@AutoRAG', '@AutoRAGRegression'] },
+    () => {
+      cy.step('Wait for run to complete and verify leaderboard');
+      autoragResultsPage.waitForRunCompletion();
```
sridarna: A e2e test should be validating a small use case. Here , there are no user steps performed in UI

packages/cypress/cypress/tests/e2e/autorag/testAutoragOptimization.cy.ts
```diff
@@ -0,0 +1,128 @@
...
+    },
+  );
+
+  it(
+    'Can interact with results page (leaderboard, pattern details, download)',
+    { tags: ['@AutoRAG', '@AutoRAGRegression'] },
+    () => {
+      autoragResultsPage.verifyResultsInteraction();
+    },
```
sridarna, line 135: A e2e test should be validating a small use case. Here , there are no user steps performed in UI

packages/cypress/cypress/utils/oc_commands/autoragInfra.ts
```diff
@@ -0,0 +1,849 @@
...
+    ).then((result) => {
+      const state = result.stdout.trim().replace(/'/g, '');
+      if (state === 'Managed') {
+        cy.log(`OGX operator (${componentName}) already Managed, skipping`);
+        return;
+      }
+
+      cy.log(`OGX operator (${componentName}) is "${state}", patching to Managed`);
+      cy.exec(
+        `oc patch ${DSC_RESOURCE} --type=merge -p '{"spec":{"components":{"${componentName}":{"managementState":"Managed"}}}}'`,
```
sridarna: We should not update the components in doc from e2e test

packages/cypress/cypress/utils/oc_commands/autoragInfra.ts
```diff
@@ -0,0 +1,849 @@
...
+/**
+ * Reset the OGX operator to Removed state.
+ * Only call this if the operator was not Managed before the test started.
+ */
+export const resetOgxOperator = (): void => {
+  detectDscComponentName().then((componentName) => {
+    cy.log(`Resetting OGX operator (${componentName}) to Removed`);
+    cy.exec(
+      `oc patch ${DSC_RESOURCE} --type=merge -p '{"spec":{"components":{"${componentName}":{"managementState":"Removed"}}}}'`,
+      { failOnNonZeroExit: false },
```
sridarna: We should not update the components in doc from e2e test

sridarna:
@jefho-rh This is definitely long and many changes in one PR. i.e include both the mock tests and e2e tests. Should have been created separately. 
Added comments, Please review them. 
Most of the E2E tests perform small use case. please combine them where ever possible as a valid user behavior.
I think you added the UI flow in Pages. This is not correct, Please move them to e2e file.
Setting up component names to Managed and Removed should not be part of the e2e.
We are configuring the state of components in Jenkins job itself.

packages/cypress/cypress/pages/autorag/configurePage.ts
jefho-rh: Resolved by moving flow logic out of page objects and into a shared utility file(`utils/autoragTestFlows.ts`)  since all three AutoRAG E2E tests share the same configure + submit flow.

packages/cypress/cypress/pages/autorag/configurePage.ts
jefho-rh: Resolved, page object now only has find* methods for each data-testid

packages/cypress/cypress/pages/autorag/configurePage.ts
jefho-rh: Resolved by moving logic out of page objects and into a shared utility file(`utils/autoragTestFlows.ts`)  since all three AutoRAG E2E tests share the same configure + submit flow.

packages/autorag/frontend/src/__tests__/cypress/cypress/pages/autoragResults.ts
jefho-rh: Resolved by removing this from PR. Mock tests will be added in a subsequent separate PR.

packages/autorag/frontend/src/__tests__/cypress/cypress/tests/mocked/run-results/autoragResults.cy.ts
jefho-rh: Resolved by removing this from PR. Mock tests will be added in a subsequent separate PR.

packages/cypress/cypress/pages/autorag/resultsPage.ts
jefho-rh: Resolved by moving flow logic out of page objects and into a shared utility file(utils/autoragTestFlows.ts) since all three AutoRAG E2E tests share the same configure + submit flow.

packages/cypress/cypress/tests/e2e/autorag/testAutoragExperimentsAndRunManagement.cy.ts
jefho-rh: OGX_URL and OGX_API_KEY are added in the test-variables.yml file and are already documented in test-variables.yml.example.

packages/cypress/cypress/tests/e2e/autorag/testAutoragExperimentsAndRunManagement.cy.ts
jefho-rh: Resolved by removing this from PR. Mock tests will be added in a subsequent separate PR.

packages/cypress/cypress/tests/e2e/autorag/testAutoragExperimentsAndRunManagement.cy.ts
jefho-rh: Resolved by combining the submit + stop flow into a single it() block in the experiments/run-management test.
For the optimization test, split into two it() blocks with different purposes: a CI-scoped test (configure + submit only) and a regression-scoped test (wait for completion + verify results). The CI test will get @AutoRAGCI tags once the OGX operator is enabled on the CI cluster.

packages/cypress/cypress/tests/e2e/autorag/testAutoragMetricVariations.cy.ts
jefho-rh: Resolved by adding `verifyAutoRAGRunSubmitted()` which verifies the run appears in the runs table in the experiment page. AutoRAG runs are managed by the KFP server(not direct K8s Pipeline runs CRs), so we validate via the UI that the run was created successfully.

packages/cypress/cypress/tests/e2e/autorag/testAutoragMetricVariations.cy.ts
jefho-rh: Resolved internally by `configureAutoRAGRun()` in the shared utility using a `RESOURCE_PATH` constant

packages/cypress/cypress/tests/e2e/autorag/testAutoragMetricVariations.cy.ts
jefho-rh: Resolved internally by `configureAutoRAGRun()` in the shared utility using a `RESOURCE_PATH` constant

packages/cypress/cypress/tests/e2e/autorag/testAutoragMetricVariations.cy.ts
jefho-rh: Resolved internally by `configureAutoRAGRun()` in the shared utility using a `RESOURCE_PATH` constant

packages/cypress/cypress/tests/e2e/autorag/testAutoragOptimization.cy.ts
jefho-rh: Resolved by adding verifyAutoRAGRunSubmitted() which verifies the run appears in the runs table in the experiment page. AutoRAG runs are managed by the KFP server(not direct K8s Pipeline runs CRs), so we validate via the UI that the run was created successfully.

packages/cypress/cypress/tests/e2e/autorag/testAutoragOptimization.cy.ts
jefho-rh: Resolved. The optimization test now has two it() blocks with distinct purposes: (1) configure + submit (fast, CI-suitable), and (2) wait for completion + verify results interaction (regression only, ~30 min). These are separated by tag so CI runs only the submit test while regression runs both.

packages/cypress/cypress/tests/e2e/autorag/testAutoragOptimization.cy.ts
jefho-rh: Resolved The optimization test now has two it() blocks with distinct purposes: (1) configure + submit (fast, CI-suitable), and (2) wait for completion + verify results interaction (regression only, ~30 min). These are separated by tag so CI runs only the submit test while regression runs both.

packages/cypress/cypress/utils/oc_commands/autoragInfra.ts
jefho-rh: Resolved by removing ensureOgxOperator() and resetOgxOperator() which patched the DSC. The test now uses isOgxOperatorManaged() as a read-only precondition check and fails with a clear error if the operator isn't Managed.

packages/cypress/cypress/utils/oc_commands/autoragInfra.ts
jefho-rh: Resolved by removing ensureOgxOperator() and resetOgxOperator() which patched the DSC. The test now uses isOgxOperatorManaged() as a read-only precondition check and fails with a clear error if the operator isn't Managed.

packages/cypress/cypress/tests/e2e/autorag/testAutoragExperimentsAndRunManagement.cy.ts
jefho-rh: Resolved by adding verifyAutoRAGRunSubmitted() which verifies the run appears in the runs table in the experiment page. AutoRAG runs are managed by the KFP server(not direct K8s Pipeline runs CRs), so we validate via the UI that the run was created successfully.

jefho-rh:
Thanks for the review @sridarna. Appreciate the thorough feedback. 

I wanted to address some of your most common feedback and provide more context:

1. **Updating components from tests**
I had originally added the OGX self-provisioning since the ODH CI Cluster hasn't migrated to OGX from Llamastack yet. When the migration happens, I will add the AutoRAGCI tag in the future since the ODH CI Cluster is not provisioned to enable AutoRAG to successfully complete testing currently.
2. **Backend validation for runs**
AutoRAG runs are managed through the KFP pipeline server API, not as direct K8s resources, so oc get pipelinerun can't query them. We validate run creation and status through the UI. Deeper backend validation would require calling the KFP REST API through the DSPA proxy.
3. **Mocked tests were in the wrong location**
I agree with your sentiments. It's better to separate this PR for E2E tests only. I have removed the mocked tests and will add them in a subsequent PR.

packages/cypress/cypress/fixtures/resources/autorag/llm-inference-service.yaml
```diff
@@ -0,0 +1,22 @@
...
+  annotations:
+    serving.kserve.io/deploymentMode: RawDeployment
+spec:
+  predictor:
+    model:
+      modelFormat:
+        name: vLLM
+      runtime: autorag-vllm-cpu-runtime
+      storageUri: '{{LLM_MODEL_URI}}'
+    resources:
```
antowaddle: Where did the resources come from? Can they be decreased to speed up the test?

packages/cypress/cypress/pages/autorag/index.ts
```diff
@@ -0,0 +1,5 @@
+/* eslint-disable no-barrel-files/no-barrel-files -- test page objects only */
```
antowaddle: Why is this needed?

packages/cypress/cypress/pages/autorag/resultsPage.ts
```diff
@@ -0,0 +1,138 @@
...
+  findManageColumnsButton() {
+    return cy.findByTestId('manage-columns-button');
+  }
+
+  findManageColumnsModal() {
+    return cy.findByRole('dialog');
+  }
+
+  findManageColumnsCancelButton() {
+    return cy.findByRole('dialog').findByRole('button', { name: 'Cancel' });
```
antowaddle: Can we create testIDs for these buttons? We try to avoid hard coded names like this.

packages/cypress/cypress/pages/autorag/resultsPage.ts
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+  findManageColumnsSaveButton() {
+    return cy.findByRole('dialog').findByRole('button', { name: 'Save' });
```
antowaddle: Can we create testIDs for these buttons? We try to avoid hard coded names like this.

packages/cypress/cypress/pages/autorag/resultsPage.ts
```diff
@@ -0,0 +1,138 @@
...
+    return cy.findByTestId('confirm-stop-run-button');
+  }
+
+  // Pattern details modal
+  findPatternDetailsModal() {
+    return cy.findByTestId('pattern-details-modal');
+  }
+
+  findPatternDetailsModalCloseButton() {
+    return this.findPatternDetailsModal().findByRole('button', { name: 'Close' });
```
antowaddle, line 87: Can we create testIDs for these buttons? We try to avoid hard coded names like this.

packages/cypress/cypress/tests/e2e/autorag/testAutoragExperimentsAndRunManagement.cy.ts
```diff
@@ -0,0 +1,122 @@
...
+      cleanupAutoragInfrastructure(projectName, testData.ogxSecretName);
+    }
+    removeOgxAccess(projectName);
+    deleteS3TestFiles(projectName, testData.awsBucket, `*${uuid}*`);
+    deleteOpenShiftProject(projectName, { wait: false, ignoreNotFound: true });
+  });
+
+  it(
+    'Can submit a run, verify it in experiments list, and stop it',
+    { tags: ['@AutoRAG', '@AutoRAGRegression'] },
```
antowaddle: Please add the FeatureFlagged tag here.

packages/cypress/cypress/tests/e2e/autorag/testAutoragExperimentsAndRunManagement.cy.ts
```diff
@@ -0,0 +1,122 @@
...
+      autoragResultsPage.findStopRunModal().should('be.visible');
+      autoragResultsPage.findConfirmStopRunButton().click();
+
+      verifyAutoragRunStopped(projectName);
+
+      cy.step('Verify run status shows as canceling, canceled, or failed');
+      autoragResultsPage
+        .findRunStatusLabel(80000)
+        .invoke('text')
+        .should('match', /CANCEL|FAIL/i);
```
antowaddle: Avoid hard coded labels, can we create testIDs here?

packages/cypress/cypress/tests/e2e/autorag/testAutoragMetricVariations.cy.ts
```diff
@@ -0,0 +1,131 @@
...
+      cleanupAutoragInfrastructure(projectName, testData.ogxSecretName);
+    }
+    removeOgxAccess(projectName);
+    deleteS3TestFiles(projectName, testData.awsBucket, `*${uuid}*`);
+    deleteOpenShiftProject(projectName, { wait: false, ignoreNotFound: true });
+  });
+
+  it(
+    'Can submit a run with answer_correctness metric',
+    { tags: ['@AutoRAG', '@AutoRAGRegression'] },
```
antowaddle: Please add FeatureFlagged tags here.

packages/cypress/cypress/tests/e2e/autorag/testAutoragMetricVariations.cy.ts
```diff
@@ -0,0 +1,131 @@
...
+    () => {
+      configureAutoragRun(testData, projectName, uuid);
+
+      cy.step('Select answer_correctness optimization metric');
+      autoragConfigurePage.findOptimizationMetricSelect().click();
+      autoragConfigurePage.findMetricOption('answer_correctness').click();
```
antowaddle, line 96: Again, add a testID here.

packages/cypress/cypress/tests/e2e/autorag/testAutoragMetricVariations.cy.ts
```diff
@@ -0,0 +1,131 @@
...
+        .type(`{selectall}${testData.maxRagPatterns}`);
+
+      submitAutoragRun();
+      verifyAutoragRunSubmitted(projectName, testData.runName);
+    },
+  );
+
+  it(
+    'Can submit a run with faithfulness metric',
+    { tags: ['@AutoRAG', '@AutoRAGRegression'] },
```
antowaddle: Add FeatureFlagged tag here too.

packages/cypress/cypress/tests/e2e/autorag/testAutoragMetricVariations.cy.ts
```diff
@@ -0,0 +1,131 @@
...
+    () => {
+      configureAutoragRun(
+        { ...testData, runName: `${testData.runName}-faith` },
+        projectName,
+        faithUuid,
+      );
+
+      cy.step('Select faithfulness optimization metric');
+      autoragConfigurePage.findOptimizationMetricSelect().click();
+      autoragConfigurePage.findMetricOption('faithfulness').click();
```
antowaddle, line 120: Again, create a testID here please.

packages/cypress/cypress/tests/e2e/autorag/testAutoragOptimization.cy.ts
```diff
@@ -0,0 +1,127 @@
+import yaml from 'js-yaml';
```
antowaddle, line 1: This seems like a lot of code to enable a feature - is there any more straight forward way to enable this feature?

packages/cypress/resources/autorag/autorag-test-document.txt
```diff
@@ -0,0 +1,17 @@
+Red Hat OpenShift AI Platform Overview
```
antowaddle, line 1: Do we need this?

antowaddle (changes requested):
Thanks @jefho-rh for this contribution.

I've added some comments, please review them. 

Note - I've added the label test:AutoRAGRegression which will run this test on this PR also. You will see the Cypress-e2e tests kicking off once the main CI tests run (Unit, Contract, Mock tests). 

Please also consider adding some CI specific tags for this module in package.json so that e2e tests will run on every change to this module. See https://redhat-internal.slack.com/archives/C05SMJ09DD2/p1780399386351379

packages/cypress/resources/autorag/autorag-test-document.txt
jefho-rh: Yupp, we need this as the test document uploaded to S3 during the configure flow. AutoRAG requires a document file for its RAG pipeline to process.

packages/cypress/cypress/tests/e2e/autorag/testAutoragOptimization.cy.ts
jefho-rh: Unfortunately not 😞.  The setup complexity comes from AutoRAG's infrastructure requirements. It needs a pipeline server, an OGX instance with registered models and vector store providers, and associated secrets.

packages/cypress/cypress/pages/autorag/index.ts
jefho-rh: You're right. On second thought it's not needed. I originally wanted to centralize a place for easy import of page files instead of individual files. Will remove

packages/cypress/cypress/pages/autorag/resultsPage.ts
jefho-rh: I think for this close button it's rendered internally by PF v6 Modal so there's no prop to add a data-testid.

packages/cypress/cypress/fixtures/resources/autorag/llm-inference-service.yaml
jefho-rh: The resources (1 CPU / 8Gi request, 4 CPU / 12Gi limit) were sized for running vLLM with Llama 3.2 1B on CPU. We can experiment with lowering them. Definitely open to suggestions if you've seen smaller resource profiles work for CPU-based vLLM serving.

packages/cypress/cypress/tests/e2e/autorag/testAutoragMetricVariations.cy.ts
jefho-rh: These should already be using data-testids.  `findMetricOption('answer_correctness')` resolves to `cy.findByTestId('metric-option-answer_correctness')`

packages/cypress/cypress/tests/e2e/autorag/testAutoragMetricVariations.cy.ts
jefho-rh: These should already be using data-testids.  `findMetricOption('faithfulness')` resolves to `cy.findByTestId('metric-option-answer_faithfulness)`

jefho-rh:
Thanks for the review @antowaddle, I've addressed your comments and have added the appropriate `FeatureFlagged` tags.

jefho-rh:
@antowaddle the PR has been updated. The provisioning is much simpler now. The tests self-provision everything they need: a lightweight pgvector instance (from quay.io), OGX Distribution, DSPA, secrets, and NetworkPolicy. Everything is cleaned up in the after() hook. The only cluster prerequisite is that the OGX operator is managed in the DSC.

It looks like the OGX operator is now managed on CI cluster, just tested and verified the tests are passing on the RHOAI nightly cluster. PR should be ready for re-review. Please let me know if there are any other concerns.

Thank you!

image

antowaddle:
> @antowaddle the PR has been updated. The provisioning is much simpler now. The tests self-provision everything they need: a lightweight pgvector instance (from quay.io), OGX Distribution, DSPA, secrets, and NetworkPolicy. Everything is cleaned up in the after() hook. The only cluster prerequisite is that the OGX operator is managed in the DSC.
> 
> It looks like the OGX operator is now managed on CI cluster, just tested and verified the tests are passing on the RHOAI nightly cluster. PR should be ready for re-review. Please let me know if there are any other concerns.
> 
> Thank you!
> 
> image

Thanks @jefho-rh, would you mind please also running this on our ODH-Nightly cluster? Thanks

jefho-rh:
@antowaddle I've just ran the tests and it looks like they are also passing on the ODH nightly-cluster(Run 1476)

image

nickmazzi:
Thanks @jefho-rh! Ran it locally yesterday to work on my branch and worked well.

/lgtm
/approve

nikhils-2:
E2E Tests passed: 

Screenshot 2026-06-25 at 2 33 01 PM

PR #7403, opened 2026-04-28
RHOAIENG-60183: LlamaStack Upgrade Path for 1.0.0

Lucifergene:
Jira: https://redhat.atlassian.net/browse/RHOAIENG-60183

## Description

Full migration of the Gen AI BFF, frontend, and test infrastructure from **LlamaStack** to **OGX 1.0.0**, aligned with the upstream project rename and breaking API changes.

### Commits

| Commit | Summary |
|--------|---------|
| `0551f27` | Align BFF with LlamaStack 1.0.0 breaking changes (API renames, struct updates, Go SDK upgrade) |
| `6ceae31` | Remove deprecated safety API (Eval, DatasetIO, Scoring, Safety) from test config |
| `8f45a08` | Upgrade BFF to OGX 1.0.0 (Go module rename, env var/header renames, `ogx run` command) |
| `4afe539` | Rename CRD from `LlamaStackDistribution` (v1alpha1) to `OGXServer` (v1beta1) |
| `7890c63` | Switch vector_io from Milvus to FAISS, align full config with upstream OGX 1.0.0 |
| `721ecbf` | Upgrade golangci-lint to v2.12.2 for Go 1.25 support |
| `c9554a3` | Switch file_processors from `inline::auto` to `inline::pypdf` for midstream compatibility |
| `3754764` | Move citation processing from frontend to BFF (strip markers, resolve filenames, annotate) |
| `3524498` | Update Dockerfile Go toolset from 1.24 to 1.25 for CI compatibility |
| `c8d530c` | Fix OGXServer CR command (`ogx run`), mock config path, and Cypress DSC condition fallback (agentic review fixes) |

### Key Decisions

**1. CRD Rename: `LlamaStackDistribution` → `OGXServer`**
All K8s client code, RBAC rules, API groups, and CRD references updated from `llamastack.rhoai.io/v1alpha1/LlamaStackDistribution` to `ogx.io/v1beta1/OGXServer`. Go module replaced from `llamastack-k8s-operator` to `ogx-k8s-operator`.

**2. Default vector store: Milvus → FAISS**
OGX 1.0.0 changed the inline Milvus provider to create collections with BM25 sparse vector fields. Milvus Lite (the embedded mode we use) doesn't implement the required gRPC method (`AllocTimestamp`), so file embeddings silently fail — files report `status: completed` but zero bytes are indexed, and RAG search returns nothing. FAISS works out of the box. External managed Milvus instances (`remote::milvus`) are unaffected.

**3. File processor: `inline::pypdf`**
Upstream OGX 1.0.0 introduced `inline::auto` which selects the right parser based on file type. However, the midstream `odh-ogx-core:rhoai-v3.5-ea1-latest` image does not include `markitdown` (a dependency of `inline::auto`), causing pod crashes. Using `inline::pypdf` instead, which works in both midstream and upstream images.

**4. Removed HuggingFace offline env vars**
`HF_HUB_OFFLINE=1`, `TRANSFORMERS_OFFLINE=1`, `HF_DATASETS_OFFLINE=1` were preventing the server from downloading embedding models at startup, breaking file uploads. The OGX 1.0.0 distribution image no longer ships cached models.

**5. Responses persistence: removed `agent_state`**
Upstream OGX 1.0.0 removed `agent_state` from the responses provider. The old Agents API was replaced by the Responses API.

**6. New storage stores: `prompts` and `connectors`**
SQL-backed stores added for the new Prompts API and Connectors API in 1.0.0.

**7. RAG configuration parameters**
Added `file_search_params`, `annotation_prompt_params` to fine-tune RAG behavior. Simplified to `enable_annotations: true` only.

**8. Citation processing moved to BFF**
OGX 1.0.0 changed the citation marker format from `【n:m†filename】` to raw UUIDs (`<|uuid|>`). Instead of keeping fragile regex matching in the frontend, citation processing now happens in the BFF:
- BFF requests `file_search` call results via the `Include` parameter
- `processResponseCitations` strips `<|uuid|>` markers from response text
- Resolves UUIDs to filenames using the file_search results map
- Populates `FileCitationAnnotation` objects on the response output
- Frontend simply reads pre-processed annotations (no marker parsing)
- Mock-based tests (`TestMockRAGCitationPipeline`) validate the full strip+annotate pipeline

**9. Frontend DSC feature flag: support both old and new keys**
`customCondition` now checks for **either** `llamastackoperator` OR `ogxoperator` in the DSC status, accepting `Managed` or `Unmanaged` states. This ensures Gen AI Studio works during the transition period.

**10. Go version: 1.24 → 1.25**
Required by the `ogx-k8s-operator` Go module. Updated in `go.mod`, CI workflows, Dockerfiles, and the CI build image (`ubi9/go-toolset:1.24` → `1.25`).

**11. golangci-lint: v2.1.0 → v2.12.2**
v2.1.0 was built with Go 1.24 and panics on Go 1.25 code.

### Agentic Review

A comprehensive agentic code review was run across the full PR diff covering:
- **Critical pass**: Enum completeness (OGX_OPERATOR consumers), race conditions, LLM trust boundary (citation processing)
- **Upstream cross-reference**: Verified commands and config paths against the `ogx-k8s-operator` source — caught deprecated `ogx stack run` (should be `ogx run`) and inconsistent mock config path
- **Cypress E2E alignment**: Added `OGXOperatorReady` fallback in the DSC condition check to match the frontend's dual-key approach
- **Nick's review items**: Verified `ogx run` command, `omitempty` on config arrays, Makefile version (confirmed 1.0.0)
- **Frontend review**: Confirmed DSC flag OR condition, citation annotation types, local dev override is unstaged

Fixes applied in commit `c8d530c`.

### Verification

Verified end-to-end on `crimson-odh` cluster with the **official midstream OGX 1.0.0 images**:
- **Operator**: `quay.io/opendatahub/odh-ogx-k8s-operator:odh`
- **Distribution**: `quay.io/opendatahub/odh-ogx-core:rhoai-v3.5-ea1-latest` (version `1.0.0+rhaiv.0`)

Tested:
- ✅ BFF build, lint, unit tests, record mode
- ✅ Frontend lint, type-check, jest tests
- ✅ OGXServer CR creation, status, deletion (full CRUD)
- ✅ Chat inference (llama-32-1b-instruct via vLLM)
- ✅ RAG end-to-end (file upload → FAISS vector store → file search with citations)
- ✅ Citation rendering (markers stripped, filenames resolved, Source Cards displayed)
- ✅ MCP server discovery and tool listing
- ✅ Model listing

### Known Limitations

- The RHOAI 3.5 EA1 DSC does not yet have the `ogxoperator` field (upstream PR merged May 13, not yet in an EA image). The frontend OR logic handles this gracefully.
- The `odh-dashboard-modules` ClusterRole on the cluster doesn't include RBAC for `ogx.io/ogxservers` yet (managed by the RHOAI operator).
- OGX 1.0.0 requires full `provider_id/model_id` format for API calls — the BFF currently passes short model IDs which works for most flows.
- The midstream `odh-ogx-core` image is missing `markitdown`, so `inline::auto` file processing is not available. Using `inline::pypdf` as fallback.
- Streaming delta events may contain raw citation markers during streaming; the final cleaned text and annotations are sent via the `response.completed` event.

## Summary by CodeRabbit

* **New Features**
  * Added support for RAG file citations in responses, enabling users to see which source files contributed to answers.
  * Introduced file processing capabilities for document indexing.

* **Refactor**
  * Migrated from LlamaStack to OGX infrastructure.
  * Replaced Milvus vector store with Faiss for improved performance.

* **Chores**
  * Updated Go toolchain to version 1.25.
  * Updated linting and analysis tool versions.
  * Added Python dependency support for document processing and package management.


packages/gen-ai/bff/internal/integrations/kubernetes/k8smocks/token_k8s_client_mock.go
```diff
@@ -552,35 +565,15 @@ providers:
...
-  telemetry:
-  - provider_id: meta-reference
-    provider_type: inline::meta-reference
-    config:
-      service_name: "${env.OTEL_SERVICE_NAME:=\u200B}"
-      sinks: ${env.TELEMETRY_SINKS:=console,sqlite}
-      sqlite_db_path: /opt/app-root/src/.llama/distributions/rh/trace_store.db
-      otel_exporter_otlp_endpoint: ${env.OTEL_EXPORTER_OTLP_ENDPOINT:=}
+  datasetio: []
+  scoring: []
```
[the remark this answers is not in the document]
Lucifergene: Thanks for flagging this. Leaving these as-is intentionally.

The `Providers` struct retains `DatasetIO` and `Scoring` fields because they *can* be populated conditionally by callers (same pattern as `Safety`). The mock emits them as empty arrays to match Go's YAML marshalling behavior — when `NewDefaultLlamaStackConfig()` is serialized, omitted slice fields marshal as `[]` in YAML, not as absent keys. Removing them from the mock would make the mock output diverge from what `config.ToYAML()` actually produces.

The canonical fixture at `testdata/llamastack/config.yaml` also includes `datasetio: []` and `scoring: []` (lines 78-79), so the mock is consistent with it.

Evidence:
- `testdata/llamastack/config.yaml` lines 78-79: `datasetio: []` / `scoring: []` present
- `llamastack_config.go` `Providers` struct: fields exist with `yaml:"datasetio"` / `yaml:"scoring"` tags — Go marshals empty slices as `[]`
- `TestDefaultConfig_Serialization` (line ~1455) verifies round-trip serialization and would break if mock diverged from actual marshal output

packages/gen-ai/bff/internal/integrations/kubernetes/testdata/test_llama_stack_config.yaml
```diff
@@ -3,12 +3,10 @@ version: "2"
 distro_name: rh
 apis:
 - responses
-- datasetio
+- file_processors
 - files
 - inference
 - safety
```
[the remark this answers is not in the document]
Lucifergene: Keeping this as-is. The fixture intentionally represents a **post-guardrails production config**, not a bare default.

The tests that load this fixture (`TestGetModelProviderInfo_LoadFromYAML`, `TestGetModelProviderInfo_EnvVarCleaning`, `TestGetModelProviderInfo_MaaSDetection`) are testing model extraction and provider-info parsing. They need a config that looks like what production actually produces — which *does* include safety when guardrails are configured via `AddGuardrailsToConfig()`.

The conditional-safety design is already tested separately:
- `TestDefaultConfig_APIsAndProviders` (line ~1322) asserts the bare default has **no** safety API, no safety providers
- `TestAddGuardrailsToConfig_InjectsSafetyAPI` (line ~1360) tests that safety is injected only when guardrails are present
- `TestAddGuardrailsToConfig_EmptyGuardrails` (line ~1395) tests that empty guardrails don't inject safety

Removing safety from this fixture would make it less realistic (production configs *will* have safety after guardrails are applied) and wouldn't improve coverage — the conditional path is already fully exercised by the tests above.

packages/gen-ai/bff/internal/api/lsd_responses_handler_test.go
```diff
@@ -649,13 +649,15 @@ var _ = Describe("LlamaStackCreateResponseHandler", func() {
...
+	// Moderation tests use NewMockLlamaStackClient() directly because in production,
+	// moderation is served by Guardrails Orchestrator / NeMo Guardrails (not LLS).
+	// The safety API was removed from LLS defaults in 1.0.0 (RHAISTRAT-1362).
 	It("should pass input moderation with clean input and InputShieldID", func() {
 		t := GinkgoT()
 		payload := CreateResponseRequest{
 			Input:         "Hello, how are you?",
 			Model:         testutil.GetTestLlamaStackModel(),
-			InputShieldID: testutil.GetTestLlamaStackShieldModel(),
+			InputShieldID: "test-shield",
```
NickGagan: You can probably get rid of any input shield tests BTW, we handle that outside of LLS with nemo guardrails

packages/gen-ai/bff/internal/integrations/kubernetes/llamastack_config.go
```diff
@@ -872,12 +869,16 @@ func CreateShieldsFromGuardrails(guardrails []models.GuardrailInput) []Shield {
...
-// AddGuardrailsToConfig adds safety providers and shields based on the guardrails configuration
+// AddGuardrailsToConfig adds safety providers and shields based on the guardrails configuration.
+// It also ensures the "safety" API is in the APIs list (it is not included by default
+// since the upstream 1.0.0 distribution removed safety from the default config).
 func (c *LlamaStackConfig) AddGuardrailsToConfig(guardrails []models.GuardrailInput) {
 	if len(guardrails) == 0 {
 		return
 	}
 
+	c.ensureAPI("safety")
```
NickGagan: Same as above, can probably get rid of this

packages/gen-ai/bff/Makefile
```diff
@@ -34,7 +34,7 @@ MLFLOW_PORT ?= 5001
 MLFLOW_DATA ?= $(shell pwd)/.mlflow
 
 # Llama Stack test server configuration (used by mock/test mode only)
-TEST_LLAMA_STACK_VERSION ?= 0.7.0
```
NickGagan, line 39: Should this be 0.8.0?

NickGagan (changes requested):
Thanks for doing this! Just a few small things from testing this on my own 0.8.0 image.
1. I noticed that eval, datasetio, scoring, and safety were all still included in the config as empty arrays. You may need to update the `Providers` type as well, or something else I'm missing. Pod was crashing until I removed them from the config.
2. There is a line in `token_k8s_client.go` for creating the LSD CR `Command: []string{"/bin/sh", "-c", "llama stack run /etc/llama-stack/config.yaml"},`. You'll need to update this with `ogx stack run /etc/llama-stack/config.yaml` or something simliar.
3. There are supposed to be operator changes as well: [Doc link](https://docs.google.com/document/d/1_L6cHUDipuRj5bYgJrKc_7Vtla4FxlVQGbavXNCih6E/edit?tab=t.da259suna39m#heading=h.gcchg4o9zdha). This means that we'll need to update:
- The CR that is being created on install to use the OGX one
- The feature flag condition for GenAI Studio to check the ogx field in DSC and not the llama stack one

One extra comment below + the previous ones for removing the shields test should probably be done as well.

NickGagan (approved):
Thanks for this! I am going to approve this since the OGX operator changes have made their way downstream and E2E is already broken anyways. I don't think the OGX core image downstream has been updated, but I tested with the midstream and that was fine
Made a few updates:
1. Switch back to Milvus from FAISS, see: https://redhat.atlassian.net/browse/RHOAIENG-62563. I made a PR to the ogx-distribution repo in odh to lock one of the deps.
2. DSC check from `ogxoperator` to `ogx`
3. Fix some merge conflicts

I tested this on a RHOAI 3.5EA1 cluster that has OGX installed. Patched the images for odh-dashboard and gen-ai to the ones created by this PR + the ogx image to `quay.io/opendatahub/odh-ogx-core:latest`. Validated that:
1. I could see Gen AI Studio in the nav with OGX set to `Managed`
2. Create an OGX deployment from AAE
3. Inference with the model
4. Use inline RAG

E2E test failures are happening in other PRs. This is unrelated.

@Lucifergene can you do a few follow ups please:
1. Create an issue for fixing in disconnected envs (after EA1). Ideally we would add back those env variables, but can't right now since the midstream OGX image doesn't install the model.
2. Do a double check of the LLS mock tests, they pass but want to make sure they're accurate of what we do in Gen AI Studio.
3. See if the ogx additions to [manifests/core-bases/base/sa-rbac/cluster-role.yaml](https://github.com/opendatahub-io/odh-dashboard/pull/7403/files/e478cf7f768633e26b7a7f09a99f144b9bbfc0ac#diff-c8b55511ef6e974e716c72d2d727f2317b58eb1e9fd97d6b67dba314b1c8e328) are still needed
4. Confirm why the go upgrade in [.github/workflows/cypress-e2e-test.yml](https://github.com/opendatahub-io/odh-dashboard/pull/7403/files/e478cf7f768633e26b7a7f09a99f144b9bbfc0ac#diff-fede99341cb900e411fa809986270876fdd42b31b56b8fd8f5a1f44eef491473) was needed

manifests/core-bases/base/sa-rbac/cluster-role.yaml
```diff
@@ -216,13 +216,13 @@ rules:
...
   - apiGroups:
-      - llamastack.io
+      - ogx.io
     verbs:
       - get
       - list
       - watch
     resources:
-      - llamastackdistributions
+      - ogxservers
```
andrewballantyne, line 225: I imagine this is just overreach stuff and these roles are not needed...

frontend/src/concepts/areas/types.ts
```diff
@@ -121,6 +121,7 @@ export enum DataScienceStackComponent {
   TRUSTY_AI = 'trustyai',
   WORKBENCHES = 'workbenches',
   LLAMA_STACK_OPERATOR = 'llamastackoperator',
+  OGX_OPERATOR = 'ogx',
```
andrewballantyne, line 124: This seems duplication 🤔

.github/workflows/cypress-e2e-test.yml
```diff
@@ -1110,7 +1110,7 @@ jobs:
         if: needs.get-test-tags.outputs.bff_packages != '[]'
         uses: actions/setup-go@v5
         with:
-          go-version: '1.24'
+          go-version: '1.25'
```
andrewballantyne, line 1113: This is generic -- was this needed?

andrewballantyne (commented):
General comments -- can I get some context?

frontend/src/concepts/areas/types.ts
andrewballantyne: "Just because OGX has a mix and match of things in the midstream and downstream" @NickGagan noted this to me in DMs... fair enough.

manifests/core-bases/base/sa-rbac/cluster-role.yaml
andrewballantyne: This is not increasing the overreach -- so I guess we can go forward here. Lets make sure we clean this up immediately post EA1. I'm not okay with this cluster-role sitting here.

NickGagan: We can do a follow up and see if these are still needed. Most likely this was done just to have parity with what was already there for LLS

andrewballantyne (approved):
Worked with @NickGagan to investigate the stuff outside of the geni-ai folder. He tested it, I commented on the other items.

We are moving to golang 1.25 overall -- so even a bit outside of the intention here I think it's fine. Hopefully our github actions work out with this new requirement.

PR #7417, opened 2026-04-29
fix: Show connected resources for URI-v1 connections

YuliaKrimerman:
https://issues.redhat.com/browse/RHOAIENG-59657

## Description

URI-v1 connections always show `--` in the **Connected Resources** column of the Connections table, even when deployed models are actively using them.

**Root cause:** `useInferenceServicesForConnection` only matched connections to InferenceServices via `storage.key` (S3), `imagePullSecrets` (OCI), or PVC URI hostname. When a model is deployed with a URI-v1 connection, `assembleInferenceService` sets `model.storageUri` directly and **deletes** `model.storage` — so `storage.key` is never present, and none of the existing matchers apply to URI connections.

**Fix:** Added two new matching strategies to `useInferenceServicesForConnection`:
1. **Annotation match** — checks the `opendatahub.io/connections` annotation (`MetadataAnnotation.ConnectionName`) on the InferenceService, which the kserve package deploy flow sets.
2. **URI value match** — decodes the connection's `data.URI` (base64) and compares it to the InferenceService's `storageUri`. This handles the `ManageKServeModal` deploy flow which doesn't set the annotation. Follows the same pattern used in `useLabeledConnections.ts` and `useRegistryConnections.ts`.

A new `isConnectionWithUri` type guard was introduced to safely narrow `Connection | PersistentVolumeClaimKind` without relying on `isSecretKind` (which fails on listed K8s items because the list API omits `kind` from individual items).

## How Has This Been Tested?

https://github.com/user-attachments/assets/11b56c4e-4452-493f-8b98-bd5e56c4a914

- Manually tested on a live cluster (ODH Dashboard local dev, `npm run start:dev:ext`) against a project with URI-v1 connections and deployed InferenceServices — Connected Resources column correctly shows the model name.
- Added 3 new unit tests covering URI storageUri comparison, annotation-based matching, and the undefined connection edge case.
- Existing `ConnectionsTable` tests continue to pass.

## Test Impact

Added 3 unit tests to `frontend/src/pages/projects/__tests__/useInferenceServicesForConnection.spec.ts`:
- `should match URI-v1 connections via storageUri value comparison`
- `should match connections via the opendatahub.io/connections annotation`
- `should return empty array when no connection is provided`

## Request review criteria:

If you have UI changes:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved inference service-to-connection association, including better handling of uri-v1 URIs and annotation-based matching.
  * Passing an undefined connection now returns an empty result instead of unexpected matches.

* **Tests**
  * Expanded test coverage to verify uri-v1 matching and the undefined-connection guard.

### DIFF
--- a/frontend/src/pages/projects/useInferenceServicesForConnection.ts
+++ b/frontend/src/pages/projects/useInferenceServicesForConnection.ts
@@ -1,8 +1,37 @@
 import * as React from 'react';
-import { InferenceServiceKind, PersistentVolumeClaimKind } from '#~/k8sTypes';
+import { InferenceServiceKind, MetadataAnnotation, PersistentVolumeClaimKind } from '#~/k8sTypes';
 import { Connection } from '#~/concepts/connectionTypes/types';
 import { ProjectDetailsContext } from '#~/pages/projects/ProjectDetailsContext';
 import { getPVCNameFromURI } from '#~/pages/modelServing/screens/projects/utils';
+
+const isConnectionWithUri = (
+  conn: Connection | PersistentVolumeClaimKind,
+): conn is Connection & { data: { URI: string } } => {
+  const { annotations } = conn.metadata;
+  return (
+    !!annotations &&
+    ('opendatahub.io/connection-type' in annotations ||
+      'opendatahub.io/connection-type-ref' in annotations) &&
+    'data' in conn &&
+    conn.data != null &&
+    typeof conn.data === 'object' &&
+    'URI' in conn.data &&
+    typeof conn.data.URI === 'string'
+  );
+};
+
+const getDecodedConnectionUri = (
+  connection: Connection | PersistentVolumeClaimKind,
+): string | undefined => {
+  if (!isConnectionWithUri(connection)) {
+    return undefined;
+  }
+  try {
+    return atob(connection.data.URI);
+  } catch {
+    return undefined;
+  }
+};
 
 export const useInferenceServicesForConnection = (
   connection?: Connection | PersistentVolumeClaimKind,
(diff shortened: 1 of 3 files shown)


frontend/src/pages/projects/useInferenceServicesForConnection.ts
```diff
@@ -17,12 +46,17 @@ export const useInferenceServicesForConnection = (
...
+    return (
+      inferenceService.metadata.annotations?.[MetadataAnnotation.ConnectionName] ===
+        connectionName ||
       inferenceService.spec.predictor.model?.storage?.key === connectionName ||
       inferenceService.spec.predictor.imagePullSecrets?.[0]?.name === connectionName ||
-      getPVCNameFromURI(inferenceService.spec.predictor.model?.storageUri ?? '') ===
-        connection.metadata.name,
-  );
+      getPVCNameFromURI(storageUri ?? '') === connection.metadata.name ||
+      (connectionUri != null && storageUri === connectionUri)
```
emilys314: So historically, this file is looking at the metadata of the resource in known spots to find a secret with the same name as this connection. This was done pre connections api. 

The reason why the URI wasn't support was because there was originally no secret being set onto the inferenceService. Only the URI was set. The issue was if you have multiple connections that all have the same URI, you have no way of knowing which is the actual connection being "used". So URI was omitted. 

With the connections API that was added with 3.0 https://github.com/opendatahub-io/architecture-decision-records/blob/main/architecture-decision-records/operator/ODH-ADR-Operator-0009-connection-api.md there is now a `opendatahub.io/connections` annotation on each resource that is using it. 

You change will still have the issue of thinking multiple connections could be on that resource, because it's only looking if the URI matches.

We only need the change `inferenceService.metadata.annotations?.[MetadataAnnotation.ConnectionName] ===
        connectionName`, the `(connectionUri != null && storageUri === connectionUri)` is not needed and will introduce issues

Here in this screenshot I have a model deployed using connection A, but the other 2 are showing up as connection as well
Image

frontend/src/pages/projects/useInferenceServicesForConnection.ts
```diff
@@ -1,9 +1,38 @@
...
+    !!annotations &&
+    ('opendatahub.io/connection-type' in annotations ||
+      'opendatahub.io/connection-type-ref' in annotations) &&
+    'data' in conn &&
+    conn.data != null &&
+    typeof conn.data === 'object' &&
+    'URI' in conn.data &&
+    typeof conn.data.URI === 'string'
+  );
+};
```
emilys314: Side note: there is already a `isConnection` function in `frontend/src/concepts/connectionTypes/utils.ts` 

(but I think `isConnectionWithUri` needs to be removed anyways based on the other comment)

YuliaKrimerman:
Opened https://github.com/opendatahub-io/odh-dashboard/pull/7443

PR #643, opened 2022-10-07
Limit notebook endpoint continued

andrewballantyne:
Minor addition on #587 -- This is done due to last min timezone issues.

Adds a layer of security to make sure we don't warn unnecessarily on non-admin access of the notebook endpoints. 

First commit is #587 -- which will rebase onto this next week and cleanup the rest of the comments.

PR #7420, opened 2026-04-29
RHOAIENG-38577: Can't stop a basic workbench on a cluster with custom OIDC provider

srtanish1992:
https://redhat.atlassian.net/browse/RHOAIENG-38577

## Description

When an admin stops a basic workbench, the frontend sends `PATCH /api/notebooks` with `{ state: "stopped", username: "<oidc-user>" }`. The backend security guard `requestSecurityGuardNotebook` in `route-security.ts` calls `getOpenshiftUser()`, which fetches `user.openshift.io/v1/users/<username>`. On clusters with custom OIDC providers (Keycloak, Azure AD, etc.), this API has **no entries for BYOIDC users** -- they authenticate via the external IdP and OpenShift does not create `User` CRs for them. The call always fails, resulting in a **403 "Wrong username"** error.

This was introduced in PR #643 (Oct 2022) as a defense-in-depth check to validate that the target user exists before allowing an admin to operate on their notebook. At the time, all clusters used OpenShift's built-in OAuth, so every user had a `User` CR. BYOIDC support came later and this guard was never updated.

**Working screen recording:**

https://github.com/user-attachments/assets/7bbf6fc6-4696-4133-9ddd-523168e79da4

### Fix
Remove the `getOpenshiftUser` check from `requestSecurityGuardNotebook`. The `testAdmin()` check alone is sufficient security because:

- Only admins can reach this code path (enforced by `testAdmin(..., true)`)
- If no notebook exists for the given username, the K8s patch in `stopNotebook` will return a 404 from the API
- Admins already have cluster-wide access to all notebooks via other routes
- The fix is scoped to the notebook admin path only -- no other resources or guards are affected

Also removes the now-unused `getOpenshiftUser` function from `userUtils.ts`.

## How Has This Been Tested?

Tested on a cluster with custom (keycloak) OIDC provider. Reach out to get access to the cluster for testing.

## Test Impact

No test changes required. There are no unit, mock, or e2e tests covering the `requestSecurityGuardNotebook` function or the `getOpenshiftUser` call. The existing Cypress tests for notebook stop/start intercept `PATCH /api/notebooks` at the HTTP level and mock the response, so they do not exercise the backend security guard. Manual testing on a BYOIDC cluster is recommended to verify the fix end-to-end.

## Request review criteria:

If you have UI changes: 

## Summary by CodeRabbit

* **Refactor**
  * Removed redundant username validation from admin security checks.
  * Consolidated user identity retrieval into existing lookup flows.
  * No expected user-facing changes; admin access flow streamlined.

### DIFF
--- a/backend/src/utils/route-security.ts
+++ b/backend/src/utils/route-security.ts
@@ -1,5 +1,5 @@
 import { FastifyReply, FastifyRequest } from 'fastify';
-import { getOpenshiftUser, getUserInfo, usernameTranslate } from './userUtils';
+import { getUserInfo, usernameTranslate } from './userUtils';
 import { createCustomError } from './requestUtils';
 import { isUserAdmin } from './adminUtils';
 import { getNamespaces } from './notebookUtils';
@@ -38,25 +38,6 @@ const testAdmin = async (
   }
 
   return false; // Not admin, no bypassing
-};
-
-const requestSecurityGuardNotebook = async (
-  fastify: KubeFastifyInstance,
-  request: OauthFastifyRequest,
-  username: string,
-): Promise<void> => {
-  // Check first admin to not give away if a user does not exist to regular users
-  await testAdmin(fastify, request, true);
-
-  try {
-    await getOpenshiftUser(fastify, username);
-  } catch (e) {
-    throw createCustomError(
-      'Wrong username',
-      'Request invalid against a username that does not exist.',
-      403,
-    );
-  }
 };
 
 const requestSecurityGuard = async (
@@ -191,7 +172,7 @@ const handleSecurityOnRouteData = async (
       request.params.name,
     );
   } else if (isRequestNotebookAdmin(request)) {
-    await requestSecurityGuardNotebook(fastify, request, request.body.username ?? '');
+    await testAdmin(fastify, request, true);
   } else if (isRequestNotebookEndpoint(request)) {
     // Endpoint has self validation internal
   } else {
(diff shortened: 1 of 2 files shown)


DaoDaoNoCode (commented):
Clean, well-scoped fix. The `getOpenshiftUser` call was fundamentally incompatible with BYOIDC clusters (Keycloak, Azure AD) since those clusters don't create `User` CRs in the `user.openshift.io` API. Removing it is the right call -- the `testAdmin` check is the real authorization gate here, and the K8s API itself will 404 if someone passes a username that doesn't map to an existing notebook.

Local checks: lint (warnings only, pre-existing) pass, type-check pass, CI all green.

backend/src/utils/route-security.ts
```diff
@@ -43,20 +43,8 @@ const testAdmin = async (
...
-  try {
-    await getOpenshiftUser(fastify, username);
-  } catch (e) {
-    throw createCustomError(
-      'Wrong username',
-      'Request invalid against a username that does not exist.',
-      403,
-    );
-  }
 };
```
DaoDaoNoCode: nit: `requestSecurityGuardNotebook` is now just a pass-through to `testAdmin`. You could inline it at the call site (line 182) and drop the function entirely, which would make the security dispatch logic slightly easier to follow. Not blocking though -- keeping it as a named function is also fine for readability if you prefer the symmetry with `requestSecurityGuard`.

srtanish1992: Inlined it.

DaoDaoNoCode (approved):
/lgtm

PR #7427, opened 2026-04-30
chore: feature store frontend clean up

PR3MM:
Fixes https://redhat.atlassian.net/browse/RHOAIENG-58679

### Description

Cleanup sweep of the Feature Store package covering grammar errors, a misspelled filename, hard-coded route strings, stale TODOs/FIXMEs, copy-paste errors in data-testid and aria-label values, and dead code.

### How Has This Been Tested?

- cd packages/feature-store && npm run type-check && npm run lint && npm run test-unit — all 450 unit tests passing; lint 0 errors (existing restricted-import warnings unchanged).
- No cluster / image smoke test from this environment.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed UI text grammar and corrected default feature-store view selection to "Overview" for unknown paths

* **Tests**
  * Updated test selectors to use consistent data-testid attributes and added/updated specs for routing and popover links

* **Refactor**
  * Replaced hardcoded feature-store URLs with route-builder functions and improved accessibility labels

* **Chores**
  * Removed outdated docs references, cleaned commented code, and added small testability hooks (data-testid props)


packages/feature-store/src/apiHooks/useFeatureStoreObject.tsx
```diff
@@ -10,7 +10,7 @@ export function getFeatureStoreObjectFromPath(pathname: string): FeatureStoreObj
     // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
     return featureStoreObject as FeatureStoreObject;
   }
-  return FeatureStoreObject.ENTITIES; // TODO: change default to overview page once we have it
+  return FeatureStoreObject.OVERVIEW;
```
kanishka-commits, line 15: Since you're already changing the default, it would be good to also check that the URL segment is a valid `FeatureStoreObject` before casting it.

Right now, any random URL segment (for example, `/develop-train/feature-store/bogus-page`) is treated like a valid enum value. This can cause problems later in maps or lookups.



```suggestion
const VALID_OBJECTS = new Set<string>(Object.values(FeatureStoreObject));

export function getFeatureStoreObjectFromPath(pathname: string): FeatureStoreObject {
  const segment = pathname.split('/')[3];
  if (segment && VALID_OBJECTS.has(segment)) {
    return segment as FeatureStoreObject;
  }
  return FeatureStoreObject.OVERVIEW;
 }
```

packages/feature-store/src/routes.ts
```diff
@@ -25,6 +25,19 @@ export const featureServiceRoute = (
 export const featureRoute = (featureName: string, featureStoreProject: string): string =>
   `${featureStoreRootRoute()}/features/${featureStoreProject}/${featureName}`;
 
+/** Features list for a project, with optional query */
```
kanishka-commits, line 28: The new function added in `routes.ts` does not have a unit test yet.
Codecov shows only 14% coverage for this change (6 lines are not covered).
Adding a small test in `routes.spec.ts` should be enough to fix this.

kanishka-commits (commented):
This is a clean, low-risk cleanup PR with no blocking issues.

Here are a few non-blocking suggestions:

packages/feature-store/src/apiHooks/useFeatureStoreObject.tsx
PR3MM: good catch, implemented in [8f9bdb0](https://github.com/opendatahub-io/odh-dashboard/pull/7427/commits/8f9bdb0d8db8eb723c25d8d5a7edacd9cf2857dc)

packages/feature-store/src/routes.ts
PR3MM: added tests in [8f9bdb0](https://github.com/opendatahub-io/odh-dashboard/pull/7427/commits/8f9bdb0d8db8eb723c25d8d5a7edacd9cf2857dc)

claudialphonse78:
Verified the route changes on metrics and lineage page redirect for features 

https://github.com/user-attachments/assets/6d6e6ac6-3f75-43f0-8f3d-a2041334c05c

/lgtm
/approve

PR #7433, opened 2026-04-30
feat(nim-serving): add NIMService watch-deployments extension to unified deployments table

YuliaKrimerman:
https://issues.redhat.com/browse/RHAISTRAT-60274

## Description

Adds `ModelServingPlatformWatchDeploymentsExtension` and `model-serving.platform` registration to the existing `@odh-dashboard/nim-serving` package, enabling NIMService deployments (`apps.nvidia.com/v1alpha1`) to appear in the unified model deployments table.

## How Has This Been Tested?

- Manually tested against a live OpenShift cluster (zaffre) with the NIM Operator installed, NIM Account configured, and a `NIMService` resource deployed in a NIM-enabled project (`opendatahub.io/nim-support: true`).

Screenshot 2026-04-29 at 12 54 39 PM

- Confirmed NIMService deployments appear in the global model deployments table with correct name, platform label ("NVIDIA NIM"), status (Stopped), and Start action.
- Verified no console errors or 404s from the NIMService watch API.
- `tsc --noEmit` passes clean on `packages/nim-serving/` and `packages/model-serving-backport/`.
- `eslint .` passes with 0 errors on both packages.

## Request review criteria:

If you have UI changes:

## Summary by CodeRabbit

* **New Features**
  * Added NVIDIA NIM deployment monitoring and management platform, providing real-time visibility into deployment status, service health, and pod metrics. Fully integrated with KServe inference services for comprehensive tracking and management of NIM-based model serving deployments.


packages/model-serving-backport/extensions.ts
emilys314, line 1: We should leave this untouched until the new NIM feature is released, otherwise this leaves NIM in a not working state

packages/nim-serving/src/deploymentEndpoints.ts
emilys314, line 1: This looks like it will work, but it's duplicating the existing kserve inference endpoint code. If we had to update the logic then we have to remember to do it in 2 places. 

We are allowed to import from the kserve package, because the NIM we are implementing built upon kserve. We can just re-export the original one in `packages/kserve/src/deploymentEndpoints.ts`. (Or we could also import directly in the `packages/nim-serving/src/deployments.ts`

packages/nim-serving/src/api/deployments/useWatchDeployments.ts
```diff
@@ -0,0 +1,79 @@
...
+  opts?: K8sAPIOptions,
+): [NIMDeployment[] | undefined, boolean, Error[] | undefined] => {
+  const [nimServices, nimServicesLoaded, nimServicesError] = useWatchNIMServices(
+    project,
+    labelSelectors,
+    opts,
+  );
+
+  const [inferenceServices, inferenceServicesLoaded, inferenceServicesError] =
+    useWatchInferenceServices(project, opts);
```
emilys314, line 26: sided though: I suppose that this could also be re-used and import from the kserve package. but that that doesn't feel as necessary

emilys314: Another side thought: This will be opening up a 2nd websocket for the exact same resources which isn't efficient.. I wonder if the browser is smart enough to de-dupe it. But fixing this isn't simple. Same with the pods

packages/nim-serving/src/api/deployments/status.ts
```diff
@@ -0,0 +1,40 @@
...
+ * The NIMService operator reconciles into an InferenceService, so we reuse the
+ * KServe status utilities against that child resource.
+ */
+export const getNIMDeploymentStatus = (
+  inferenceService: InferenceServiceKind | undefined,
+  deploymentPods: PodKind[],
+): DeploymentStatus | undefined => {
+  if (!inferenceService) {
+    return undefined;
+  }
```
emilys314, line 29: We can probably be more precise with the conditions, but it's always hard to understand all the states..

If the infsvc is missing, should we return a FAILED_TO_LOAD? or I wonder if this could also be a `LOADING` state

packages/model-serving-backport/extensions.ts
emilys314: Especially if the one in the `nim-serving` package drops the `backport:` components

packages/nim-serving/extensions.ts
```diff
@@ -12,6 +27,51 @@ const extensions: AreaExtension[] = [
       reliantAreas: [SupportedArea.NIM_MODEL],
     },
   },
+  {
+    type: 'model-serving.platform',
+    properties: {
```
emilys314, line 32: It's probably best to not have this in the new package, it will need to be deleted when we release. 

We are trying to move away from the "Select a platform" logic for a project. That's because it limits you to only one platform and it's the old way of doing it. 

KServe still has it because it had to be compatible with the legacy NIM, so kserve would show up as "Single model serving platform". the llmd-serving package better shows the target state, with how it doesn't have this extension but it still works and is fully integrated into the wizard

emilys314 (commented):
Haven't tested this yet

mturley:
@YuliaKrimerman the package-lock file has entries from packages/nim from before your rebase, and some changes to license fields in the notebooks package (I think that's an npm version weirdness thing). We should reset to the lockfile from main.

mturley:
Working on an actual review, but we should probably add some tests here for the new hooks/utils as well @YuliaKrimerman

emilys314:
What was the yaml you used to create the NIMService?

YuliaKrimerman:
@emilys314 I didn't create a new one - it was the one that was there in your project - should I create a new one as well? I thought that just seeing the NIM service(even an old one) is good enough for the testing

emilys314:
> @emilys314 I didn't create a new one - it was the one that was there in your project - should I create a new one as well? I thought that just seeing the NIM service(even an old one) is good enough for the testing

That one was the legacy one I believe (inferenceservice + servingruntime)

I did create yesterday a NIMService `meta-llama-3-2-1b-instruct` though which can be used to check

packages/nim-serving/src/api/watch.ts
```diff
@@ -0,0 +1,64 @@
...
+): CustomWatchK8sResult<PodKind[]> =>
+  useK8sWatchResourceList<PodKind[]>(
+    {
+      isList: true,
+      groupVersionKind: groupVersionKind(PodModel),
+      namespace: project?.metadata.name,
+      selector: {
+        matchExpressions: [
+          {
+            key: 'serving.kserve.io/inferenceservice',
```
emilys314: This is an example `Pod` for the NIMService deployment. (It's not healthy but it's assembled)
```
kind: Pod
apiVersion: v1
metadata:
  annotations:
    openshift.io/required-scc: nonroot
    openshift.io/scc: nonroot
    security.openshift.io/validated-scc-subject-type: serviceaccount
  name: meta-llama3-1b-instruct-pod
  namespace: emily
  ownerReferences:
    - apiVersion: apps.nvidia.com/v1alpha1
      kind: NIMCache
      name: meta-llama3-1b-instruct
      uid: 8f58b2ef-15eb-4644-a61f-2c18a06e2ae9
      controller: true
      blockOwnerDeletion: true
  labels:
    app: k8s-nim-operator
    app.kubernetes.io/managed-by: k8s-nim-operator
    app.kubernetes.io/name: meta-llama3-1b-instruct
spec:
  nodeSelector:
    feature.node.kubernetes.io/pci-10de.present: 'true'
  restartPolicy: Always
  serviceAccountName: nim-cache-sa
  imagePullSecrets:
    - name: odh-nim-account-pull
  ...
status:
  phase: Pending
  conditions:
    - type: PodScheduled
      status: 'False'
      lastProbeTime: null
      lastTransitionTime: '2026-05-05T17:50:55Z'
      reason: Unschedulable
      message: '0/2 nodes are available: 2 node(s) didn''t match Pod''s node affinity/selector. preemption: 0/2 nodes are available: 2 Preemption is not helpful for scheduling.'
  qosClass: BestEffort
```

There is no `serving.kserve.io/inferenceservice` label present. We should change this to `app: k8s-nim-operator` AND / OR `app.kubernetes.io/managed-by: k8s-nim-operator`

packages/nim-serving/src/deploymentStatus.ts
```diff
@@ -0,0 +1,47 @@
...
+      message: 'Waiting for NIM Operator to provision InferenceService',
+    };
+  }
+
+  const deploymentPod = deploymentPods.find(
+    (pod) =>
+      pod.metadata.labels?.['serving.kserve.io/inferenceservice'] ===
+      inferenceService.metadata.name,
+  );
+  const modelPodStatus = deploymentPod ? checkModelPodStatus(deploymentPod) : null;
```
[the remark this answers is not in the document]
emilys314: This is an interesting thought, if there is a re-rollout out, there could be multiple pods.

The first match could make the status flicker if: on one render it shows the new one, and on another render it shows the 2nd

YuliaKrimerman:
Screenshot 2026-05-06 at 1 28 59 PM
@mturley @emilys314 Verified that the new one that Emily created is there as well

emilys314 (commented):
It shows up!

Can you fix the alignment of the row? It is missing an expanded section extension I think

Image

packages/nim-serving/src/api/watch.ts
emilys314: So yeah we need to change the pod selector otherwise this will never find the corresponding NIMService's Pods to merge this PR

packages/nim-serving/src/api/deployments/useWatchDeployments.ts
mturley: Should we open a followup tech debt issue for this?

packages/nim-serving/src/api/watch.ts
```diff
@@ -0,0 +1,59 @@
...
+  PodKind,
+  ProjectKind,
+} from '@odh-dashboard/internal/k8sTypes';
+import useK8sWatchResourceList from '@odh-dashboard/internal/utilities/useK8sWatchResourceList';
+import { groupVersionKind } from '@odh-dashboard/internal/api/k8sUtils';
+import { InferenceServiceModel } from '@odh-dashboard/internal/api/models/kserve';
+import type { CustomWatchK8sResult } from '@odh-dashboard/internal/types';
+import { NIMServiceModel, type NIMServiceKind } from '../types';
+
+export const useWatchNIMServices = (
```
mturley: To fit with the file structure of `src/api/<object>/[constants|hooks|k8s|utils|types|etc].ts` we should probably reorganize things a bit here since we're working with distinct objects here (NIMServices, InferenceServices, Pods). We probably want directories `src/api/[nimservices|inferenceservices|pods]` with a `watch.ts` under each, or I guess maybe just `src/api/deployments` since they're a bit intermingled. Either way, `types.ts` and `watch.ts` in each api subdirectory matching whichever object that directory is for.

packages/nim-serving/src/pages/deployments/NIMServingDetails.tsx
```diff
@@ -0,0 +1,5 @@
+import React from 'react';
+
+const NIMServingDetails: React.FC = () => <>NIM service</>;
```
mturley, line 3: This also probably fits in something like `src/pages/deployments/*` to match with `src/pages/projectSettings` (already present) and `src/pages/deploymentWizard` (added in https://github.com/opendatahub-io/odh-dashboard/pull/7466). Generally I think we should organize components based on which page they extend, where applicable, unless they are shared by multiple pages.

packages/nim-serving/src/api/deployments/useWatchDeployments.ts
```diff
@@ -0,0 +1,88 @@
...
+  useWatchNIMServices,
+  useWatchInferenceServices,
+  useWatchNIMDeploymentPods,
+} from './api/watch';
+import { getNIMDeploymentStatus } from './deploymentStatus';
+import { NIM_ID } from '../extensions';
+
+export type { NIMDeployment };
+
+export const useWatchDeployments = (
```
mturley, line 13: Feels like the stuff in this file also belongs somewhere like `src/api/deployments/watch.ts`

packages/nim-serving/src/api/deployments/status.ts
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+ * Derive deployment status from the InferenceService created by the NIM Operator.
+ * The NIMService operator reconciles into an InferenceService, so we reuse the
+ * KServe status utilities against that child resource.
+ *
+ * When the InferenceService hasn't been created yet (operator is still reconciling),
+ * we return LOADING rather than undefined so the UI shows a meaningful state.
+ */
+export const getNIMDeploymentStatus = (
```
mturley, line 19: Same here, somewhere in `src/api/deployments`, or we could split up deployments from deploymentStatus - up to you, I could see having everything related to inferenceservices/nimservices/pods/deployments/status/etc combined in `src/api/deployments` or broken into a few different directories. Whatever feels easier to read/navigate

packages/nim-serving/src/api/nimservices/types.ts
```diff
@@ -0,0 +1,92 @@
...
+import { NIM_ID } from '../extensions';
+
+export const NIMServiceModel: K8sModelCommon = {
+  apiVersion: 'v1alpha1',
+  apiGroup: 'apps.nvidia.com',
+  kind: 'NIMService',
+  plural: 'nimservices',
+};
+
+export type NIMServiceKind = K8sResourceCommon & {
```
mturley, line 16: Types somewhere in `src/api` alongside the objects they are for

packages/nim-serving/src/types.ts
```diff
@@ -0,0 +1,92 @@
...
+  status?: {
+    state?: string;
+    conditions?: K8sCondition[];
+    availableReplicas?: number;
+  };
+};
+
+export type NIMDeployment = Deployment<NIMServiceKind, InferenceServiceKind>;
+
+export const isNIMDeployment = (deployment: Deployment): deployment is NIMDeployment =>
```
mturley: is* helpers somewhere in a `utils.ts` file in `src/api`

packages/nim-serving/src/api/watch.ts
```diff
@@ -0,0 +1,59 @@
...
+  project?: ProjectKind,
+  opts?: K8sAPIOptions,
+): CustomWatchK8sResult<PodKind[]> =>
+  useK8sWatchResourceList<PodKind[]>(
+    {
+      isList: true,
+      groupVersionKind: groupVersionKind(PodModel),
+      namespace: project?.metadata.name,
+      selector: {
+        'app.kubernetes.io/managed-by': 'k8s-nim-operator',
```
mturley: Nit: we have some hard-coded strings that could go in a `constants.ts` file somewhere. Looks like the managed-by label is already hard-coded across the repo, but `k8s-nim-operator` is something we could establish as a new constant. It's not related to a specific object in `api/` so we could just have a `nim-serving/src/constants.ts`

packages/nim-serving/src/deployments.ts
```diff
@@ -0,0 +1,88 @@
...
+  const filteredNIMServices = React.useMemo(
+    () => (filterFn ? nimServices.filter(filterFn) : nimServices),
+    [nimServices, filterFn],
+  );
+
+  const inferenceServiceByNIMName = React.useMemo(() => {
+    const byName = new Map<string, InferenceServiceKind>();
+    inferenceServices.forEach((is) => {
+      is.metadata.ownerReferences?.forEach((ref) => {
+        if (ref.kind === 'NIMService' && ref.apiVersion.startsWith('apps.nvidia.com/')) {
```
mturley: Nit: `'NIMService'` here could be using `NIMServiceModel.kind` instead of a raw string, and `'apps.nvidia.com'` could be using `NIMServiceModel.apiGroup`. Also this check could also just be its own little `isNIMServiceRef` util

packages/nim-serving/src/types.ts
```diff
@@ -0,0 +1,92 @@
...
+  deployment.modelServingPlatformId === NIM_ID;
+
+/**
+ * Check if an InferenceService is owned by a NIMService
+ * by inspecting its ownerReferences.
+ */
+export const isNIMOwned = (inferenceService: InferenceServiceKind): boolean =>
+  inferenceService.metadata.ownerReferences?.some(
+    (ref) => ref.kind === 'NIMService' && ref.apiVersion.startsWith('apps.nvidia.com/'),
```
mturley: Nit: Same here, we can import and use `NIMServiceModel` to get these strings

mturley (changes requested):
Some moderate nits about folder organization and super minor nits about string constants. Mostly LGTM, haven't tested though

packages/nim-serving/src/api/watch.ts
emilys314: Yeah the same managed-by will most likely be needed for Yulia's next PR to filter out InferenceServices from kserve that have this label too

packages/nim-serving/src/deployments.ts
emilys314: Yeah a generic ~`isNIMServiceRef = (resource: K8sResourceCommon)` would handy

edit: wait there already is a `isNIMOwned` that we can reuse

packages/nim-serving/src/types.ts
emilys314: I kinda like having the is* near this type, otherwise the `utils.ts` will get very long 

Although there is an argument to keep type files clean of any JS and only TS types, but then we'd have to move `NIMServiceModel` and also this file is importing a string from ../extensions so this typescript file is tainted.

packages/nim-serving/src/api/deployments/useWatchDeployments.ts
```diff
@@ -0,0 +1,86 @@
...
+
+  const errors = React.useMemo(
+    () =>
+      [nimServicesError, inferenceServicesError, deploymentPodsError].filter(
+        (error): error is Error => Boolean(error),
+      ),
+    [nimServicesError, inferenceServicesError, deploymentPodsError],
+  );
+
+  return [deployments, effectivelyLoaded, errors];
```
emilys314: We should memoize the array result, otherwise it provides a new reference every re-render

packages/nim-serving/extensions.ts
emilys314, line 1: I'm still seeing the table alignment issue for the NIMServices

Image

packages/nim-serving/extensions.ts
```diff
@@ -26,6 +38,27 @@ const extensions: (AreaExtension | ProjectDetailsSettingsCardExtension)[] = [
...
+    },
+  },
+  {
+    type: 'model-serving.deployed-model/serving-runtime',
+    properties: {
+      platform: NIM_ID,
+      ServingDetailsComponent: () => import('./src/pages/deployments/NIMServingDetails'),
+    },
+    flags: {
+      required: [SupportedArea.NIM_MODEL],
```
emilys314: You should change the feature flags to `SupportedArea.NIM_WIZARD` since that is what we are hiding it all behind. Otherwise model serving users's browsers will be fetching and running this nim code even if it's not enabled
```suggestion
      required: [SupportedArea.NIM_WIZARD],
    },
  },
  {
    type: 'model-serving.deployed-model/serving-runtime',
    properties: {
      platform: NIM_ID,
      ServingDetailsComponent: () => import('./src/pages/deployments/NIMServingDetails'),
    },
    flags: {
      required: [SupportedArea.NIM_WIZARD],
```

packages/nim-serving/extensions.ts
YuliaKrimerman: This is caused by NIM rows not having a form-data extension, so hardwareProfilePaths is undefined and the expand toggle <Td> isn't rendered, creating a column count mismatch. Registering a full form-data extension requires implementing ~10 extractors (replicas, model format, hardware config, etc.), do we want to do this as part of this PR or should we maybe open up a follow up story ?

packages/nim-serving/src/types.ts
mturley: Makes sense to colocate the type guards with the types actually, that sounds good to me

packages/nim-serving/extensions.ts
mturley: Maybe we can create a stub form-data extension for now that just triggers the toggle with some placeholders, and then follow up? That's data we need for this story, right? and isn't NIM using the same mechanisms as kserve in terms of the replicas/format/etc data? I wonder if we can reuse existing code for that (my limited knowledge of NIM is getting in the way here)

emilys314: So yeah we do have this story which is hooking up the misc stuff, this would probably fall under it so that's fine if we do it in a follow up PR / story https://redhat.atlassian.net/browse/RHOAIENG-60282

But I'd like to at least fix the alignment issue when there is no extension for the platform. That seems fairly easy and it would just be showing an empty cell as a placeholder I think

emilys314 (approved):
Looks good!

Image

PR #7436, opened 2026-04-30
feat(nim-serving): NIM settings card for project API key management

mturley:
https://redhat.atlassian.net/browse/RHOAIENG-60271
https://redhat.atlassian.net/browse/RHOAIENG-60272

## Description

Adds a NIM settings card to the Project Settings tab, allowing users to enter an NVIDIA API key per-project. The card creates a Secret and Account CR in the project namespace, polls for validation status, and supports remove/replace-key flows.

https://github.com/user-attachments/assets/b52f085e-e9bb-4505-b26a-e056abcab760

### What's included

- **New extension point** `app.project-details/settings-card` in `plugin-core` — allows packages to register cards on the project settings tab
- **Settings tab visibility** — now shows when any settings card extensions are active (not just when bias metrics is available)
- **NIM settings card** (in `packages/nim-serving/`) with full state machine:
  - **No account**: "Add personal API key" button opens API key modal
  - **Pending**: info alert with spinner — "Contacting NVIDIA to validate the license key." (shown on card only; modal shows "Validating" loading button instead)
  - **Error**: danger alert with error details on card + "Remove" / "Replace key" buttons; inline field validation error in modal
  - **Ready**: inline success helper text on card + "Remove" / "Replace key" buttons
- **API key modal** that automatically detects create vs replace based on account status:
  - Submit button shows "Validating" with spinner during validation (no inline alert)
  - On success: shows "API key validated" screen with success message and Close button
  - On error: shows inline field validation error under the input
- **K8s resource management**: creates Secret (`nvidia-nim-secrets`) → Account; on delete, explicitly removes both Account and Secret (no ownerReference cascade)
- **Dry-run validation**: all multi-step K8s operations (create, update) dry-run first to validate before committing — prevents orphaned resources on validation failures
- **Polling hook** (`useNIMAccountStatus`) that derives status from Account conditions, built on the project's `useFetch` hook with `refreshRate`
- **Shared `NIMAccountStatusAlerts` component** for status display on the card
- **Unit tests** for K8s utilities and the status hook (24 tests)
- Gated behind the `nimWizard` feature flag

### Changes since initial review

- **Permission gate**: Settings tab visibility gated behind `allowCreate` — users without write permissions no longer see settings cards
- **Namespace prop**: `NIMSettingsCard` accepts `namespace` as a prop via typed `ProjectDetailsSettingsCardProps` extension point, avoiding cross-package context import
- **No cross-package dependency**: `NIMAccountModel` and `listNIMAccounts` are defined in `packages/nim-serving/` for the new code, with the old NIM code in `frontend/src/` keeping its own copies — avoids the host app depending on `@odh-dashboard/nim-serving`
- **Dry-run validation**: `createNIMResources` dry-runs both secret and account creation in parallel before creating either; `updateNIMSecretAndRevalidate` dry-runs the replace before committing
- **Callback memoization**: `handleSubmit` and `handleClose` in the API key modal wrapped in `useCallback`; `handleRemoveConfirm` in the settings card wrapped in `useCallback`
- **Unmount cleanup**: Delete polling uses `setInterval` with a ref, cleaned up via `useEffect` return — prevents state updates on unmounted components
- **Input guards**: Replace mode guarded with explicit error if `existingSecretName` is missing; API key trimmed before use
- **Deterministic account selection**: `getNIMAccount` uses `accounts.find(a => a.metadata.name === NIM_ACCOUNT_NAME)` instead of `accounts[0]`
- **Folder organization**: `src/` reorganized into `api/accounts/` (k8s, utils, hooks, constants with colocated tests) and `pages/projectSettings/` (UI components) — structure accommodates future resource groups like `api/deployments/`
- **Explicit resource cleanup**: `deleteNIMResources` deletes both the Account and the Secret (account first, then secret); both delete operations treat 404 as success so concurrent or repeated deletes don't surface spurious errors
- **LOADING status**: new status returned before the initial fetch completes, preventing a flash of the NOT_FOUND UI
- **Stale error clearing**: `deleteError` is reset when the remove modal is closed
- **Retry after validation failure**: if account already exists when retrying, the modal automatically uses the replace flow instead of trying to create a duplicate
- **Simplified modal state**: removed `isReplacing` prop — the modal determines create vs replace from `accountStatus`; replaced `ApiKeyModalState` enum with a simple boolean
- **Await refresh in modal**: `refresh` prop typed as `Promise<unknown>` and awaited in submit handler to keep the submit lock until status is fetched
- **Create-or-replace resilience**: `createNIMResources` handles a pre-existing secret (409 Conflict) by falling back to fetch + replace; `updateNIMSecretAndRevalidate` handles a missing secret (404 Not Found) by creating it instead of failing — both flows are now idempotent against stale or externally-modified secret state

### Status derivation logic

Status is derived from the Account CR's `status.conditions`:
- **LOADING**: initial fetch has not yet completed — prevents flashing the "Add personal API key" button before the real state is known
- **NOT_FOUND**: no account exists
- **PENDING**: account exists but validation/reconciliation is not yet complete (either no conditions yet, or key is validated but `AccountStatus` is not yet `True`) and no conditions have `status: False`
- **READY**: `AccountStatus` condition is `True`
- **ERROR**: any condition has `status: False` (e.g. `APIKeyValidation: False` for a bad key, or `AccountStatus: False` for other failures), or PENDING has exceeded the 5-minute timeout

### Validation timeouts

The PENDING state has a 5-minute timeout based on **server-side timestamps** (`lastAccountCheck` or `creationTimestamp`), not client-side timers. This means the timeout is detected correctly even if the user reloads the page after the time has passed. After 5 minutes, PENDING transitions to ERROR — with the operator's condition messages if available, or a fallback message suggesting the user replace the key or remove NIM.

### Validation state architecture

The `useNIMAccountStatus` hook is owned by the settings card and its results are passed down to the modal as props. This means:
- The card and modal always show consistent status (single source of truth)
- If the user closes the modal mid-validation, the card continues showing the pending/error/ready state — no state is lost
- The pending spinner is hidden on the card while the modal is open (the modal's "Validating" button is sufficient)
- The modal stays reactive to status changes even after entering an error state — if the operator recovers, the modal transitions back to validating

When replacing a key, the hook enters a "waiting for revalidation" mode: it overrides the derived status to PENDING until the operator's `lastTransitionTime` on the `AccountStatus` condition changes (indicating the operator has re-checked the key) or a 30-second timeout elapses. This prevents the UI from briefly showing the old "success" state before the operator processes the new key.

### Secret data keys

The NIM Account operator expects `api_key` as the secret data key, while NIM deployments (ServingRuntime/InferenceService) expect `NGC_API_KEY`. To use a single secret for both purposes, we include both keys with the same value. The secret uses a static well-known name (`nvidia-nim-secrets`) so models can reference it directly.

This is a temporary workaround — once the Account controller is updated to use `NGC_API_KEY`, we can drop the `api_key` entry.

### Replace key implementation

Replacing a key reuses `createOrReplaceSecret` — the same function used during initial creation. It tries to create the secret; on 409 (already exists), it fetches the existing secret to get its `resourceVersion`, spreads in the new `stringData`, and PUTs it back. The `runtimes.opendatahub.io/nim-force-validation` annotation is set on the secret (not the Account CR), matching the old NIM implementation.

### Initial state — no NIM account
Screenshot 2026-05-01 at 2 34 47 PM

### Modal — entering API key
Screenshot 2026-04-30 at 2 50 43 PM

### Modal - failed to create resources (immediate error before validating starts)
In this example, RBAC isn't set up correctly for the user to create NIM accounts
Screenshot 2026-05-01 at 4 34 48 PM

### Modal — validating
Screenshot 2026-05-01 at 3 32 31 PM

### Modal — validation succeeded
Screenshot 2026-05-01 at 3 20 53 PM

### Modal — validation failed (inline error)
Screenshot 2026-05-01 at 3 22 41 PM

### Card — validating (pending), modal closed
Screenshot 2026-04-30 at 4 42 26 PM

### Card — validation failed
Screenshot 2026-05-01 at 3 28 04 PM

### Card — API key saved (ready)
Screenshot 2026-05-01 at 2 37 29 PM

### Card — validation timeout
Screenshot 2026-05-01 at 3 30 25 PM

### Remove confirmation
Screenshot 2026-05-01 at 3 23 45 PM

## How Has This Been Tested?
- Unit tests for `nimK8sUtils.ts` (resource assembly, create/update/delete flows, dry-run validation, status helpers) and `useNIMAccountStatus.ts` (status derivation, polling, timeouts)
- Manual testing on a cluster with the NIM operator installed: verified secret + Account CR creation, validation polling, error states, replace key flow, remove flow, and timeout behavior

## Test Impact
- Added 24 unit tests across 2 test files in `packages/nim-serving/src/api/accounts/__tests__/`
- Cypress mock tests not yet added (will follow in a subsequent PR)

## Request review criteria:
If you have UI changes:

## Release Notes

* **New Features**
  * Added NVIDIA NIM settings interface to project settings, allowing users to manage personal API keys.
  * Implemented API key validation with real-time status feedback (pending, success, error states).
  * Users can now add, replace, and remove API keys with a streamlined modal experience.
  * Added status alerts to guide users through the license key validation process.


mturley:
/hold
needs more testing

mturley:
Held because we need to figure out RBAC, regular users don't have permission to create NIM accounts in their own project

mturley:
cc @vconzola , screenshots ready for review

vconzola:
@vconzola

> 1. For this state, "Modal - failed to create resources (immediate error before validating starts)", if RBAC isn't set up correctly resubmitting is never going to work, so I suggest we only have Close as a primary button.

This is just one example of why it might fail at that stage, the UI code can't know whether it was RBAC or something else that prevented it submitting. might have been that the user disconnected from wifi or something. We will fix this RBAC issue, I don't think it'll be possible to hit that error on release, and the only other way we'd hit this is if there's some unexpected unknown error that could be transient. The pattern elsewhere is to let them retry

> 2. For the cases where the validation fails, why do we need two actions Remove and Replace? As a user I would not know which one to choose. Is the bad key saved so it needs to be removed or replaced? If not, couldn't we just re-display the Add button? If we keep both buttons, we probably should have tooltips on them explaining what they do.

Hmm, fair point. Maybe tooltips would help. In this case yes, the bad key is saved so the user needs to choose whether to remove it (give up and undo everything) or replace it (try again with a different key).

> 3. The validation failure alerts should probably be dismissable.

This is tricky because the validation is async on the backend, so the alert is shown when errors are present on the server, not just right after a user request failed. if I make them dismissable they'll just appear again after reloading the page, unless I persist that dismissal which feels hacky. also if you dismiss that, it's unclear that there is still a bad key present that you probably want to remove, this is more of a status display than a failed request. it's more similar to a failed deploy of a model.

> 4. For the success case, are the Remove/Replace options consistent with what we're doing with tokens/secrets for sources in model catalog?

I'm not sure I've seen the latest UX for tokens for model catalog, that's just for the new HF private models strat right? in the current model catalog there are no tokens/secrets. it's a similar mechanism though, yeah, do you have those designs handy?

> 5. For the remove confirmation, instead of having them enter "NVIDIA NIM" (I agree that's weird.), how about just "Remove" or "Remove key"?

Sure, can do that

vconzola:
@mturley 
> @vconzola
> 
> > 1. For this state, "Modal - failed to create resources (immediate error before validating starts)", if RBAC isn't set up correctly resubmitting is never going to work, so I suggest we only have Close as a primary button.
> 
> This is just one example of why it might fail at that stage, the UI code can't know whether it was RBAC or something else that prevented it submitting. might have been that the user disconnected from wifi or something. We will fix this RBAC issue, I don't think it'll be possible to hit that error on release, and the only other way we'd hit this is if there's some unexpected unknown error that could be transient. The pattern elsewhere is to let them retry

OK, if the RBAC issue is getting fixed, then good enough.
> 
> > 2. For the cases where the validation fails, why do we need two actions Remove and Replace? As a user I would not know which one to choose. Is the bad key saved so it needs to be removed or replaced? If not, couldn't we just re-display the Add button? If we keep both buttons, we probably should have tooltips on them explaining what they do.
> 
> Hmm, fair point. Maybe tooltips would help. In this case yes, the bad key is saved so the user needs to choose whether to remove it (give up and undo everything) or replace it (try again with a different key).

Yeah let's add tooltips (can be a follow-up issue if need be). Just take a swing at the text (or I can). I've asked Katie to do a complete content review before this is released.
> 
> > 3. The validation failure alerts should probably be dismissable.
> 
> This is tricky because the validation is async on the backend, so the alert is shown when errors are present on the server, not just right after a user request failed. if I make them dismissable they'll just appear again after reloading the page, unless I persist that dismissal which feels hacky. also if you dismiss that, it's unclear that there is still a bad key present that you probably want to remove, this is more of a status display than a failed request. it's more similar to a failed deploy of a model.

Fair enough.
> 
> > 4. For the success case, are the Remove/Replace options consistent with what we're doing with tokens/secrets for sources in model catalog?
> 
> I'm not sure I've seen the latest UX for tokens for model catalog, that's just for the new HF private models strat right? in the current model catalog there are no tokens/secrets. it's a similar mechanism though, yeah, do you have those designs handy?

I'll try to dig them up.
> 
> > 5. For the remove confirmation, instead of having them enter "NVIDIA NIM" (I agree that's weird.), how about just "Remove" or "Remove key"?
> 
> Sure, can do that

Thanks.

mturley:
@vconzola I pushed changes to the PR to add tooltips and remove the input box in the "remove" modal:

Screenshot 2026-05-04 at 3 21 00 PM
Screenshot 2026-05-04 at 3 21 06 PM
Screenshot 2026-05-04 at 3 23 01 PM

frontend/src/pages/projects/screens/detail/ProjectDetails.tsx
```diff
@@ -192,7 +196,7 @@ const ProjectDetails: React.FC = () => {
                   },
                 ]
               : []),
-            ...(biasMetricsAreaAvailable && allowCreate
+            ...((biasMetricsAreaAvailable && allowCreate) || hasSettingsCards
               ? [
```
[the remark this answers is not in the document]
emilys314: +1 we don't want to show a user buttons that will fail due to insufficient permissions

packages/nim-serving/src/NIMSettingsCard.tsx
```diff
@@ -0,0 +1,193 @@
...
+const NIM_DESCRIPTION =
+  'NVIDIA NIM, part of NVIDIA AI Enterprise, is a set of easy-to-use microservices designed ' +
+  'for secure, reliable deployment of high-performance AI model inferencing across the cloud, ' +
+  'data center and workstations. Supporting a wide range of AI models, including open-source ' +
+  'community and NVIDIA AI Foundation models, it ensures seamless, scalable AI inferencing, ' +
+  'on-premises or in the cloud, leveraging industry standard APIs.';
+
+const NIMSettingsCard: React.FC = () => {
+  const { currentProject } = React.useContext(ProjectDetailsContext);
+  const namespace = currentProject.metadata.name;
```
emilys314: Oo what if we passed in the ProjectKind (or namespace string) in as a prop? That would prevent us from needing the cross package import


```suggestion
const NIMSettingsCard: React.FC = ({project}) => {
  const namespace = project.metadata.name;
```

packages/nim-serving/src/nimK8sUtils.ts
```diff
@@ -0,0 +1,160 @@
+import {
+  k8sCreateResource,
+  k8sDeleteResource,
+  k8sPatchResource,
+} from '@openshift/dynamic-plugin-sdk-utils';
+import { NIMAccountKind, SecretKind, K8sCondition } from '@odh-dashboard/internal/k8sTypes';
+import { NIMAccountModel } from '@odh-dashboard/internal/api/models/odh';
+import { SecretModel } from '@odh-dashboard/internal/api/models';
+import { createSecret, getSecret, replaceSecret } from '@odh-dashboard/internal/api/k8s/secrets';
+import { listNIMAccounts } from '@odh-dashboard/internal/api/k8s/nimAccounts';
```
emilys314: We should move the listNIMAccounts into the `nim-serving` package

packages/nim-serving/src/nimK8sUtils.ts
```diff
@@ -0,0 +1,160 @@
+import {
+  k8sCreateResource,
+  k8sDeleteResource,
+  k8sPatchResource,
+} from '@openshift/dynamic-plugin-sdk-utils';
+import { NIMAccountKind, SecretKind, K8sCondition } from '@odh-dashboard/internal/k8sTypes';
+import { NIMAccountModel } from '@odh-dashboard/internal/api/models/odh';
```
emilys314: Probably a good idea to move these into the nim package as well

packages/nim-serving/src/nimK8sUtils.ts
```diff
@@ -0,0 +1,160 @@
...
+    await k8sDeleteResource<SecretKind>({
+      model: SecretModel,
+      queryOptions: { name: secretName, ns: namespace },
+    });
+    throw e;
+  }
+
+  await k8sPatchResource<SecretKind>({
+    model: SecretModel,
+    queryOptions: { name: secretName, ns: namespace },
```
emilys314: There may be a edge case bug?
- try create the NIM account
- it fails, so deletes the secret
- try patching the secret. But it's deleted so it would fail?

packages/nim-serving/src/nimK8sUtils.ts
```diff
@@ -0,0 +1,160 @@
...
+      name: secretName,
+    },
+  },
+});
+
+export const createNIMResources = async (
+  namespace: string,
+  apiKey: string,
+): Promise<NIMAccountKind> => {
+  const secret = await createSecret(assembleNIMSecret(namespace, apiKey));
```
emilys314: Dry runs are probably a good idea to have for all our api calls

(I'm not sure what our guidelines are, but k8s passthrough makes dry runs fairly easy, so it seems like an easy win)

packages/nim-serving/src/NIMApiKeyModal.tsx
```diff
@@ -0,0 +1,178 @@
...
+  startRevalidation,
+  accountStatus,
+}) => {
+  const [apiKey, setApiKey] = React.useState('');
+  const [showKey, setShowKey] = React.useState(false);
+  const [isCreating, setIsCreating] = React.useState(false);
+  const [submitted, setSubmitted] = React.useState(false);
+  const [createError, setCreateError] = React.useState<string>();
+
+  const handleSubmit = async () => {
```
emilys314: Best to make this a callback so it doesn't recreate the reference on every render to the child

packages/nim-serving/src/NIMSettingsCard.tsx
```diff
@@ -0,0 +1,193 @@
...
+
+  const [apiKeyModalState, setApiKeyModalState] = React.useState(ApiKeyModalState.CLOSED);
+  const [isDeleteModalOpen, setIsDeleteModalOpen] = React.useState(false);
+  const [isDeleting, setIsDeleting] = React.useState(false);
+  const [deleteError, setDeleteError] = React.useState<Error>();
+
+  const existingSecretName = nimAccount?.spec.apiKeySecret.name;
+  const isApiKeyModalOpen = apiKeyModalState !== ApiKeyModalState.CLOSED;
+
+  const handleRemoveConfirm = async () => {
```
emilys314: It's a good idea to make this a callback

packages/nim-serving/src/NIMSettingsCard.tsx
```diff
@@ -0,0 +1,193 @@
...
+    setIsDeleting(true);
+    setDeleteError(undefined);
+    try {
+      await deleteNIMResources(namespace);
+      const pollUntilGone = async (retries = 10): Promise<void> => {
+        const result = await refresh();
+        if (result && retries > 0) {
+          await new Promise((resolve) => {
+            setTimeout(resolve, 1000);
```
emilys314: This might continue polling even when the component gets unmounted.

emilys314 (commented):
I haven't tested it yet

packages/nim-serving/src/api/accounts/utils.ts
emilys314, line 1: nitpick but also it would be nice to start categorizing these files into different folders. Otherwise the root src folder of this package will become chaotic quickly

packages/nim-serving/src/nimK8sUtils.ts
mturley: We'd need the existing NIM feature to import it from nim-serving then right, since we don't want to break it? Not sure if that'll be a problem. If so I guess we could just duplicate it for now.

emilys314: Ah. True. 

if we don't do it now, we will have to do it after the feature release. Duplicating also works. I'm fine either way

frontend/src/pages/projects/screens/detail/ProjectDetails.tsx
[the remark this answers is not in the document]
mturley: Fixed

packages/nim-serving/src/NIMSettingsCard.tsx
mturley: Good call, done

packages/nim-serving/src/nimK8sUtils.ts
mturley: Done -- however, if we move rather than copy these, the old NIM code requires us to import from `@odh-dashboard/nim-serving` in the main app. I added that to package.json and saw it's the first time we'd be importing from one of the packages from the host app directly. Do you see that as a problem? It works fine for now since nim-serving isn't federated

packages/nim-serving/src/nimK8sUtils.ts
mturley: Good catch. I added dry runs and cleanup, so this should be covered.

packages/nim-serving/src/nimK8sUtils.ts
mturley: Done

packages/nim-serving/src/NIMApiKeyModal.tsx
mturley: Done

packages/nim-serving/src/NIMSettingsCard.tsx
mturley: Done

packages/nim-serving/src/NIMSettingsCard.tsx
mturley: Done. I switched to using a setInterval with a useEffect to kill it on unmount. This is kind of an awkward pattern, I wonder if later on it's worth trying to abstract out a sort of "wait and poll to see if the delete went through" hook/util, but probably fine for now

packages/nim-serving/src/api/accounts/utils.ts
mturley: Good call, I did some reorganization

packages/nim-serving/src/api/accounts/utils.ts
```diff
@@ -0,0 +1,188 @@
...
+  await replaceSecret(updatedSecret, { dryRun: true });
+  await replaceSecret(updatedSecret);
+};
+
+export const deleteNIMResources = async (namespace: string): Promise<void> => {
+  await k8sDeleteResource<NIMAccountKind>({
+    model: NIMAccountModel,
+    queryOptions: { name: NIM_ACCOUNT_NAME, ns: namespace },
+  });
+};
```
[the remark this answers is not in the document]
mturley: Addressed

packages/nim-serving/src/pages/projectSettings/NIMSettingsCard.tsx
```diff
@@ -0,0 +1,205 @@
...
+          setIsDeleteModalOpen(false);
+        } else if (retries === 0) {
+          stopPollingDeleteStatus(new Error('NIM resources were not deleted in time.'));
+        }
+        retries -= 1;
+      } catch (e) {
+        stopPollingDeleteStatus(e instanceof Error ? e : new Error('Failed to remove NIM.'));
+      }
+    }, 1000);
+  }, [namespace, refresh, stopPollingDeleteStatus]);
```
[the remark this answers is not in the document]
mturley: Addressed

packages/nim-serving/src/hooks/useNIMAccountStatus.ts
```diff
@@ -0,0 +1,149 @@
...
+
+  const {
+    data: nimAccount,
+    loaded,
+    refresh,
+  } = useFetch(fetchCallback, null, {
+    refreshRate: pollRate,
+  });
+
+  const derived = loaded ? deriveStatus(nimAccount) : deriveStatus(null);
```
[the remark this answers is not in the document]
mturley: Addressed

packages/nim-serving/src/nimK8sUtils.ts
emilys314: Ah yeah we haven't been doing that to avoid circular dependencies. Model-serving imports dashboard, dashboard imports nim-serving, nim-serving imports model-serving, and repeat

So i'd revert that and just do a cross package import (from odh-dashboard) for now then

mturley:
Fixing tests etc after addressing comments, will push and unhold shortly

mturley:
@emilys314 this one should be ready for re-review. I addressed the feedback and did some other cleanup based on some things that revealed.

frontend/package.json
```diff
@@ -65,6 +65,7 @@
   },
   "dependencies": {
     "@module-federation/runtime": "^0.15.0",
+    "@odh-dashboard/nim-serving": "*",
```
emilys314: Yeah not sure about this. Best to remove the circular dependency I think, just to be safe

packages/cypress/cypress/tests/mocked/projects/projectDetails.cy.ts
```diff
@@ -28,11 +28,11 @@ import {
 import type { InferenceServiceKind, ServingRuntimeKind } from '@odh-dashboard/internal/k8sTypes';
 import { DataScienceStackComponent } from '@odh-dashboard/internal/concepts/areas/types';
 import { deleteProjectModal, editProjectModal, projectDetails } from '../../../pages/projects';
+import { NIMAccountModel } from '@odh-dashboard/nim-serving/k8s';
```
emilys314: This should be fine to import from `nim-serving`, because model serving / nim doesn't import from anything from the cypress package

packages/nim-serving/src/pages/projectSettings/NIMSettingsCard.tsx
```diff
@@ -0,0 +1,200 @@
...
+              <StackItem>
+                <NIMAccountStatusAlerts status={status} errorMessages={errorMessages} />
+              </StackItem>
+            )}
+            {actionButtons && <StackItem>{actionButtons}</StackItem>}
+          </Stack>
+        );
+      }
+      default:
+        return null;
```
emilys314, line 184: If you don't have permission, there is simply no footer. We may want to add the button but put it in a disabled state with a popover saying why? 

Or Some text saying you need permission. 

Although idk how likely it is to be able to access this page if you are a project viewer in the first place, since you need `allowCreate = true` in `frontend/src/pages/projects/screens/detail/ProjectDetails.tsx`

Image

packages/nim-serving/src/api/accounts/utils.ts
```diff
@@ -0,0 +1,196 @@
...
+    generateName: NIM_SECRET_GENERATE_NAME,
+    namespace,
+    labels: {
+      'opendatahub.io/managed': 'true',
+    },
+  },
+  type: 'Opaque',
+  stringData: {
+    [NIM_API_KEY_DATA_KEY]: apiKey,
+    [NGC_API_KEY_DATA_KEY]: apiKey,
```
emilys314: very very very very minor nitpick: but can you add a comment on each key just to say which one is used for what? It's an unusual pattern so I just want to make sure it's documented in the code

packages/nim-serving/src/api/accounts/utils.ts
```diff
@@ -0,0 +1,196 @@
...
+    },
+  },
+});
+
+export const createNIMResources = async (
+  namespace: string,
+  apiKey: string,
+): Promise<NIMAccountKind> => {
+  const secretData = assembleNIMSecret(namespace, apiKey);
+  const accountData = assembleNIMAccount(namespace, '');
```
emilys314: Need to pass in the secret name

```suggestion
  const accountData = assembleNIMAccount(namespace, NIM_SECRET_GENERATE_NAME);
```

packages/nim-serving/src/api/accounts/utils.ts
```diff
@@ -0,0 +1,196 @@
...
+      name: secretName,
+    },
+  },
+});
+
+export const createNIMResources = async (
+  namespace: string,
+  apiKey: string,
+): Promise<NIMAccountKind> => {
+  const secretData = assembleNIMSecret(namespace, apiKey);
```
emilys314: We might want to `await` so the account will have the secret ready when looking

packages/nim-serving/src/api/accounts/utils.ts
```diff
@@ -0,0 +1,196 @@
...
+  NGC_API_KEY_DATA_KEY,
+  NIM_FORCE_VALIDATION_ANNOTATION,
+} from './constants';
+
+export const assembleNIMSecret = (namespace: string, apiKey: string): SecretKind => ({
+  apiVersion: 'v1',
+  kind: 'Secret',
+  metadata: {
+    name: '',
+    generateName: NIM_SECRET_GENERATE_NAME,
```
emilys314: We need a stable and well known secret for models to refer to.
```suggestion
    name: NIM_SECRET_NAME,
```

emilys314 (commented):
Hm I seem to get an error when submitting the key 

Image

packages/nim-serving/src/api/accounts/utils.ts
mturley: I think @andrewballantyne had said don't hard code the name of the secret, and we can have other code look it up by looking at the account. but I agree it would be simpler to just have a known name. want me to do that?

packages/nim-serving/src/pages/projectSettings/NIMSettingsCard.tsx
mturley: @emilys314 pushed a change, here's how it looks now if for some reason you reach this tab but don't have permission to create nim accounts. Tested by temporarily modifying our temp role.

Screenshot 2026-05-07 at 1 40 48 PM

packages/nim-serving/src/api/accounts/utils.ts
mturley: Done

packages/nim-serving/src/api/accounts/utils.ts
mturley: Done

packages/nim-serving/src/api/accounts/utils.ts
mturley: Done

packages/nim-serving/src/api/accounts/utils.ts
mturley: Done

packages/nim-serving/src/api/accounts/utils.ts
```diff
@@ -0,0 +1,149 @@
...
+  let account: NIMAccountKind;
+  try {
+    account = await createNIMAccount(accountData);
+  } catch (e) {
+    await deleteSecret(namespace, NIM_SECRET_NAME);
+    throw e;
+  }
+
+  try {
+    await patchSecretOwnerReference(namespace, NIM_SECRET_NAME, account);
```
emilys314: Hmm I don't think we need to be doing this? It would avoid our current rbac issue too

The legacy NIM secrets do not have this (the global one or the project scoped one)

Image

The Account will add an owner ref the global pull secret it creates, but that's because the Account creates it. It may be best to leave the owner reference creation to the Account? Thoughts?

packages/nim-serving/src/pages/projectSettings/NIMSettingsCard.tsx
```diff
@@ -0,0 +1,233 @@
...
+    switch (status) {
+      case NIMAccountStatus.NOT_FOUND:
+        return (
+          <Button
+            variant="secondary"
+            onClick={() => setIsApiKeyModalOpen(true)}
+            data-testid="nim-enable-button"
+          >
+            Add personal API key
+          </Button>
```
emilys314, line 124: I noticed the outline of the button here is blue, while the trustyAi one is grey 

Image

packages/nim-serving/src/pages/projectSettings/NIMAccountStatusAlerts.tsx
```diff
@@ -0,0 +1,37 @@
...
+      return (
+        <Alert
+          variant="info"
+          isInline
+          title="Contacting NVIDIA to validate the license key."
+          customIcon={<Spinner size="md" />}
+        >
+          This can take a few minutes.
+        </Alert>
+      );
```
emilys314, line 25: I was trying out an edge case where I just delete my secret. When I navigate to the card, it will just be in a loading state until it times out. 

I'd expect the failure to surface right away

Image

emilys314 (commented):
Everything else works good

packages/nim-serving/src/api/accounts/utils.ts
```diff
@@ -0,0 +1,140 @@
...
+};
+
+export const createNIMResources = async (
+  namespace: string,
+  apiKey: string,
+): Promise<NIMAccountKind> => {
+  const secretData = assembleNIMSecret(namespace, apiKey);
+  const accountData = assembleNIMAccount(namespace, NIM_SECRET_NAME);
+
+  await Promise.all([createNIMSecret(secretData, true), createNIMAccount(accountData, true)]);
```
emilys314: I ran into an issue when I already had an existing secret in my project (called `nvidia-nim-secrets`) and I tried to turn NIM on

It shows as not enabled which is to be expected
Image

But then entering my key gives a "secrets "nvidia-nim-secrets" already exists" error
Image

This is an easy fix though we should just make this a ~ createOrUpdateIfExists() type function

packages/nim-serving/src/api/accounts/utils.ts
```diff
@@ -0,0 +1,140 @@
...
+
+export const updateNIMSecretAndRevalidate = async (
+  namespace: string,
+  secretName: string,
+  apiKey: string,
+): Promise<void> => {
+  const existingSecret = await fetchExistingSecret(namespace, secretName);
+  const updatedSecret = assembleUpdatedSecret(existingSecret, apiKey);
+  await replaceNIMSecret(updatedSecret, true);
+  await replaceNIMSecret(updatedSecret);
```
emilys314: Same here, if I delete my secret, it shows the invalid state which is correct. But trying to "Replace" it will give this error

Image

We should also make this a createOrReplace() style function

emilys314 (commented):
edge cases

packages/cypress/cypress/tests/mocked/projects/projectDetails.cy.ts
mturley: I missed this comment, and had moved this to import from a copy in cypress utils/models to be consistent with how all the other objects are handled

packages/nim-serving/src/api/accounts/utils.ts
mturley: Done. And I replaced the old update-secret path with the new createOrReplaceSecret function since they were basically doing the same thing and the create attempt with 409 handling is safe in both cases (especially with the dry run)

packages/nim-serving/src/api/accounts/utils.ts
mturley: Done

PR #7440, opened 2026-05-01
feat(gen-ai): inline NeMo guardrail config with per-request model discovery

pnaik1:
## Description

Replaces the previous per-model ConfigMap-based NeMo Guardrails integration with an inline-config approach where all guardrail parameters are passed directly in the API request body, and removes the now-unused LLS (LlamaStack Safety) guardrails path.

## How Has This Been Tested?
Manually tested end-to-end in a live OpenShift cluster with the BFF running locally via oc port-forward:
Sent requests through POST /gen-ai/api/v1/lsd/responses with guardrail_config populated
Verified input moderation blocks flagged content and passes safe content for in-cluster vLLM models (guardrail_model_source_type: "namespace")
Verified the maas- prefix and explicit model_source_type: "maas" both correctly resolve the guardrail model URL from the live MaaS catalog (not the stale ConfigMap URL)
Verified custom endpoint models resolve correctly via the external-models ConfigMap
Confirmed NeMo receives the correct bare model name (provider prefix stripped) so the underlying vLLM backend can find the model

## Request review criteria:

If you have UI changes: 

## Summary by CodeRabbit

* **New Features**
  * API accepts inline guardrail_config per request; frontend sends guardrail_config and stores per-configuration guardrailSubscription.

* **Behavior Changes / Bug Fixes**
  * Input/output moderation now runs conditionally with provided guardrail settings across streaming and non‑streaming flows; legacy shield ID fields removed.

* **Frontend Changes**
  * Guardrails UI reworked to use centralized guardrail state and subscription dropdown; guardrail-related public endpoints/types removed.

* **Tests**
  * Added MaaS-focused resolution tests and updated unit/Cypress tests to the new guardrail flows.


packages/gen-ai/bff/internal/api/middleware.go
```diff
@@ -338,32 +354,18 @@ func (app *App) AttachMaaSClient(next func(http.ResponseWriter, *http.Request, h
 			// In mock mode, use empty URL since mock factory ignores it
 			maasClient = app.maasClientFactory.CreateClient("", "", app.config.InsecureSkipVerify, app.rootCAs)
 		} else {
-			var serviceURL string
+			serviceURL := app.resolveMaaSBaseURL()
 
-			// Configuration Priority:
```
NickGagan, line 343: Was this MaaS change needed? I think this will break things. 

Portforwarding doesn't work properly here, instead of making these changes try adding this to your env.local:
```
MAAS_URL=https://maas.apps.rosa.crimson-rhoai.h6fk.p3.openshiftapps.com/maas-api
```

packages/gen-ai/bff/internal/api/middleware.go
```diff
@@ -318,6 +318,22 @@ func (app *App) AttachLlamaStackClient(next func(http.ResponseWriter, *http.Requ
 	}
 }
 
+// resolveMaaSBaseURL returns the MaaS controller base URL based on configuration.
+//
+// Priority:
+//  1. MAAS_URL env var (explicit developer override)
+//  2. Autodiscovered endpoint using the cached cluster domain
+//  3. Returns "" when MaaS is not available
+func (app *App) resolveMaaSBaseURL() string {
```
NickGagan: Same as https://github.com/opendatahub-io/odh-dashboard/pull/7440/changes#r3184534820

NickGagan (changes requested):
First pass looks good! Tested the following in the first pass, using a maas model (will test namespace and custom_endpoints later): input + output guardrails in and out of comparison mode, error state for if guardrails is not installed, scenario where guardrails_check fails.

Have some comments:
1. For input guardrails, you should fail fast. Can keep existing error handling in the UI in place, since @toklumpp is working on something to standardize this in playground. For this ticket, throw an error, then create a follow up to give a useful message in the UI.
2. For output guardrails, when the request fails it should an error instead of showing `The response to your request was intercepted by safety guardrails. The output was found to contain potential harmful content or sensitive data (PII).`
3. For input guardrails, I would recommend passing in the entire set of input messages and see if at least one message is blocked. You should still only need to make one request, just give the array of messages instead of the string. Notice in my example before, that because we are only running it on the last message, I was able to get it past the input guardrails by jailbreaking it in future messages: 

Image

pnaik1:
Hello @NickGagan I have updated the changes

packages/gen-ai/bff/internal/api/lsd_responses_handler.go
```diff
@@ -337,8 +337,53 @@ func (app *App) LlamaStackCreateResponseHandler(w http.ResponseWriter, r *http.R
 		}
 	}
 
-	// TODO: Input moderation will be wired here once the guardrail_config request field
-	// is implemented (replaces the deprecated InputShieldID path).
+	// Build inline guardrail options when the request includes a guardrail config.
+	// The BFF resolves the model endpoint URL and API key so the frontend never handles credentials.
+	var guardrailOpts nemo.GuardrailsOptions
+	if createRequest.GuardrailConfig != nil && createRequest.GuardrailConfig.GuardrailModel != "" {
+		baseURL, apiKey, err := app.getGuardrailModelEndpointAndKey(ctx, createRequest.GuardrailConfig.GuardrailModel, createRequest.GuardrailConfig.GuardrailModelSourceType, createRequest.GuardrailConfig.ResolveSubscription(createRequest.Subscription))
```
NickGagan, line 344: Just as a follow up, this would be a helpful reusable function if it could take the AAE model ID + model type + Subscription and generate the API key + base URL. It will be needed for some of the upcoming multi-modal inferencing work as well.

cc: @Lucifergene

packages/gen-ai/bff/internal/api/lsd_responses_handler.go
```diff
@@ -337,8 +337,53 @@ func (app *App) LlamaStackCreateResponseHandler(w http.ResponseWriter, r *http.R
...
+		if err != nil {
+			app.logger.Error("Failed to resolve guardrail model endpoint", "model", createRequest.GuardrailConfig.GuardrailModel, "error", err)
+			app.serviceUnavailableResponse(w, r, errors.New(constants.GuardrailServiceUnavailableMessage))
+			return
+		}
+
+		// NeMo needs the bare model name (e.g. "claude-haiku-4-5-20251001"), not the
+		// LlamaStack-qualified ID (e.g. "endpoint-1/claude-haiku-4-5-20251001").
+		guardrailModelName := createRequest.GuardrailConfig.GuardrailModel
+		if idx := strings.Index(guardrailModelName, "/"); idx != -1 {
```
NickGagan, line 354: This scares me, but I can't think of a better alternative 😅

packages/gen-ai/bff/internal/api/lsd_responses_handler.go
```diff
@@ -919,6 +871,204 @@ func (app *App) getCustomEndpointSecret(ctx context.Context, modelID string) str
...
+	apiKey = app.fetchSecretFromProvider(ctx, k8sClient, identity, namespace, foundProvider, actualModelID)
+	return baseURL, apiKey
+}
+
+// getGuardrailModelEndpointAndKey resolves the raw inference endpoint URL and API key
+// for the given guardrail model ID. NeMo Guardrails calls this endpoint directly,
+// bypassing LlamaStack, so the URL must point at the model's native API surface.
+//
+// Resolution mirrors getProviderData for the main model:
+//  1. maas- prefix              → ephemeral MaaS token + MaaS catalog URL
```
NickGagan, line 948: I think the `maas` model_source_type is also passed from the responses API request body via guardrail_config

NickGagan (approved):
Approving since the rest of the items are not blockers! Have a few follow ups in the comments.

Another follow up would be a UX change for how we expect to handle guardrail output. Current behaviour is to stream back the sentences and then override the text with the warning if some of it is flagged (video). I think it might be less jarring to have it keep streaming back results but mask the sentences with some text if they violated the guardrails.

Also worth investigating why the first phone numbers weren't flag.

https://github.com/user-attachments/assets/c635d628-7ed7-49a1-b8b9-a6f2c1d3d2a4

Might also be worth cleaning up some of this code to make a reusable function for fetching inference URLs and API keys for AAE assets. @Lucifergene would benefit when adding support for ASR models as well.

PR #7442, opened 2026-05-01
feat(backend): support DEV_IMPERSONATE_TOKEN for OAuth/IDP users

mturley:
## Description
Adds a new `DEV_IMPERSONATE_TOKEN` env var that allows providing a pre-obtained bearer token directly, bypassing the OAuth Basic Auth flow in dev mode.

The existing `DEV_IMPERSONATE_USER` / `DEV_IMPERSONATE_PASSWORD` mechanism uses HTTP Basic Auth against the cluster's OAuth challenging-client endpoint. This doesn't work for OAuth/IDP users (e.g. Google IDP) who don't have a cluster password.

With this change, developers can:
1. Log in via `oc login --web` (browser-based OAuth)
2. Get their token via `oc whoami -t`
3. Set `DEV_IMPERSONATE_TOKEN=<token>` in `.env.local` instead of `DEV_IMPERSONATE_PASSWORD`
4. Toggle impersonation in the UI as normal

When `DEV_IMPERSONATE_TOKEN` is set, the `/api/dev-impersonate` endpoint skips the OAuth flow entirely and uses the provided token directly. `DEV_IMPERSONATE_USER` should also be set, as it is used for username resolution.

**Note:** OpenShift tokens expire (typically 24h), so the token needs periodic refresh.

### Documentation updates
- `.env.local.example` — added impersonation env vars with usage guidance
- `docs/SDK.md` — updated "Pass Through Impersonate User Dev Mode" section to cover both password and token options

## How Has This Been Tested?
- Verified the backend starts successfully with `DEV_IMPERSONATE_TOKEN` set
- Confirmed the impersonation toggle uses the provided token without attempting Basic Auth

## Test Impact
No new tests added. This is a dev-only feature (`DEV_MODE` gated) that adds an alternative code path alongside the existing Basic Auth flow. The existing flow is unchanged when `DEV_IMPERSONATE_TOKEN` is not set.

## Request review criteria:

## Summary by CodeRabbit

* **New Features**
  * Added token-based authentication option for developer impersonation mode, allowing use of pre-obtained OAuth/IDP tokens as an alternative to password credentials.

* **Documentation**
  * Updated configuration examples and SDK documentation with instructions for obtaining tokens, setting up impersonation variables, and handling token expiration.

### DIFF
--- a/.env.local.example
+++ b/.env.local.example
@@ -20,3 +20,14 @@ OC_URL=replace-with-url
 OC_USER=replace-with-username
 OC_PASSWORD=replace-with-password
 # OC_TOKEN=replace-with-token
+
+# Dev Impersonation (optional) — test as a different user without disabling backend functionality.
+# Set DEV_IMPERSONATE_USER along with EITHER a password OR a token:
+#   - Password: for users with local cluster credentials (htpasswd, etc.)
+#   - Token: for OAuth/IDP users (Google, LDAP, etc.) who don't have a cluster password.
+#     Obtain a token via: oc login --web && oc whoami -t
+#     Note: tokens expire (typically 24h) and must be refreshed.
+# If on ROSA, also set DEV_OAUTH_PREFIX=oauth
+# DEV_IMPERSONATE_USER=replace-with-username
+# DEV_IMPERSONATE_PASSWORD=replace-with-password
+# DEV_IMPERSONATE_TOKEN=replace-with-token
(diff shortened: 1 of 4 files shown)


Griffin-Sullivan (commented):
/lgtm

reads fine - haven't tested

christianvogt:
Useful dev only straight forward change.

@mturley the `DEV_IMPERSONATE_USER` is still used as the user name even when using token. May want to document that it should still be supplied but can opt for password or token. Perhaps other doc updates as well.

PR #7443, opened 2026-05-01
fix: Show connected resources for URI-v1 connections

YuliaKrimerman:
https://issues.redhat.com/browse/RHOAIENG-59657

## Description

URI-v1 connections always show `--` in the **Connected Resources** column of the Connections table, even when deployed models are actively using them.

**Root cause:** `useInferenceServicesForConnection` only matched connections to InferenceServices via `storage.key` (S3), `imagePullSecrets` (OCI), or PVC URI hostname. When a model is deployed with a URI-v1 connection, `assembleInferenceService` sets `model.storageUri` directly and **deletes** `model.storage` — so `storage.key` is never present, and none of the existing matchers apply to URI connections.

**Fix:** Added two new matching strategies to `useInferenceServicesForConnection`:
1. **Annotation match** — checks the `opendatahub.io/connections` annotation (`MetadataAnnotation.ConnectionName`) on the InferenceService, which the kserve package deploy flow sets.
2. **URI value match** — decodes the connection's `data.URI` (base64) and compares it to the InferenceService's `storageUri`. This handles the `ManageKServeModal` deploy flow which doesn't set the annotation. Follows the same pattern used in `useLabeledConnections.ts` and `useRegistryConnections.ts`.

A new `isConnectionWithUri` type guard was introduced to safely narrow `Connection | PersistentVolumeClaimKind` without relying on `isSecretKind` (which fails on listed K8s items because the list API omits `kind` from individual items).

## How Has This Been Tested?

https://github.com/user-attachments/assets/11b56c4e-4452-493f-8b98-bd5e56c4a914

- Manually tested on a live cluster (ODH Dashboard local dev, `npm run start:dev:ext`) against a project with URI-v1 connections and deployed InferenceServices — Connected Resources column correctly shows the model name.
- Added 3 new unit tests covering URI storageUri comparison, annotation-based matching, and the undefined connection edge case.
- Existing `ConnectionsTable` tests continue to pass.

## Test Impact

Added 3 unit tests to `frontend/src/pages/projects/__tests__/useInferenceServicesForConnection.spec.ts`:
- `should match URI-v1 connections via storageUri value comparison`
- `should match connections via the opendatahub.io/connections annotation`
- `should return empty array when no connection is provided`

## Request review criteria:

If you have UI changes:

## Summary by CodeRabbit

* **Improvements**
  * Enhanced inference service matching for connections by adding metadata annotation-based matching criteria alongside existing storage and pull secrets matching, improving service discovery accuracy.

### DIFF
--- a/frontend/src/pages/projects/__tests__/useInferenceServicesForConnection.spec.ts
+++ b/frontend/src/pages/projects/__tests__/useInferenceServicesForConnection.spec.ts
@@ -46,4 +46,44 @@ describe('useInferenceServicesForConnection', () => {
     const renderResult = testHook(useInferenceServicesForConnection)(connection);
     expect(renderResult.result.current).toHaveLength(1);
   });
+
+  it('should match connections via the opendatahub.io/connections annotation', () => {
+    const uriConnection = mockConnection({
+      name: 'uri-connection',
+      displayName: 'URI Connection',
+      connectionType: 'uri-v1',
+      data: { URI: btoa('https://example.com/model') },
+    });
+
+    const uriInferenceServices = [
+      mockInferenceServiceK8sResource({
+        name: 'uri-model',
+        displayName: 'URI Model',
+        secretName: 'uri-connection',
+        storageUri: 'https://example.com/model',
+      }),
+      mockInferenceServiceK8sResource({
+        name: 'other-model',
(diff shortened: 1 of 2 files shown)


emilys314 (approved):
looks good!

Image

PR #7447, opened 2026-05-01
fix(gen-ai): preserve vector store selection when entering compare mode

NickGagan:
https://issues.redhat.com/browse/RHOAIENG-58426

## Description

When a user selected an external vector store (knowledge mode = \`external\`) in the Playground settings panel and then entered compare mode, the vector store selection was silently cleared.

**Root cause:** Entering compare mode causes \`ChatbotConfigInstance\` to unmount and remount because its parent component structure changes — \`ComparePaneWrapper\` is inserted above it in the tree. On remount, a \`useEffect\` that syncs \`selectedVectorStoreId\` with \`knowledgeMode\` would fire and, seeing \`knowledgeMode === 'external'\`, unconditionally write \`null\` to the store.

**Fix:** The effect now uses a ref to track the previous \`knowledgeMode\`. It only clears \`selectedVectorStoreId\` when the user explicitly transitions from \`inline → external\`, not on every remount. When already in \`external\` mode on mount (e.g., after a remount from entering compare mode), the existing selection is preserved.

## How Has This Been Tested?

Manually tested the following scenarios:

1. Inline vector store selected → enter compare mode → inference — ✅ works
2. Inline vector store selected → enter compare mode → disable RAG on one pane → inference — ✅ works
3. **External vector store selected → enter compare mode → inference** — ✅ selection preserved (was broken before this fix)
4. External vector store selected → enter compare mode → switch one pane back to inline → inference — ✅ works

## Test Impact

**Unit tests** (`ChatbotConfigInstance.spec.tsx`):
- Updated an existing test that was asserting the old broken behavior (clear on fresh mount with \`external\` mode already set). The test now uses \`rerender\` so the clear is verified during an \`inline → external\` transition while the component is mounted, which is the correct scenario for clearing.
- Added a new case `preserves selectedVectorStoreId on remount when knowledgeMode is already external` that directly documents the bug scenario and verifies the fix.

**Cypress mock tests** (\`compareMode.cy.ts\`):
- Added an end-to-end mock test that selects an external vector store, enters compare mode, and asserts the selection is still visible in Chat 1's settings panel.
- Added three page object methods to \`chatbotPage.ts\`: \`findExternalModeRadio\`, \`findVectorStoreSelectToggle\`, \`findVectorStoreOption\`.

## Request review criteria:

- [ ] The developer has tested their solution on a cluster by using the image produced by the PR to \`main\`

🤖 Generated with [Claude Code](https://claude.ai/claude-code)


Lucifergene (approved):
/lgtm
/approve

PR #7453, opened 2026-05-04
add nim option

emilys314:
Closes https://redhat.atlassian.net/browse/RHOAIENG-60276

## Description

This is a really simple integration of the NIM option. No extensions, i was going to but I started to scope creep with refactoring stuff and decided to just do this to unblock the rest of the wizard flow

image
image

## How Has This Been Tested?

Enable the NIM feature flag, go create a new model, it should show up

## Request review criteria:

If you have UI changes: 

## Summary by CodeRabbit

* **New Features**
  - Added NVIDIA NIM as a selectable model location in the deployment wizard.

* **Improvements**
  - Better model location validation and conditional requirement of connection data.
  - Streamlined connection creation flow and organization within the wizard.
  - Improved connection-type filtering for more accurate choices.
  - More reliable CPU/memory customization behavior during deployment edits.

### DIFF
--- a/packages/model-serving/src/components/deploymentWizard/fields/ModelLocationSelectField.tsx
+++ b/packages/model-serving/src/components/deploymentWizard/fields/ModelLocationSelectField.tsx
@@ -7,7 +7,7 @@ import {
   Stack,
   StackItem,
 } from '@patternfly/react-core';
-import { z, type ZodIssue } from 'zod';
+import { type ZodIssue } from 'zod';
 import SimpleSelect from '@odh-dashboard/internal/components/SimpleSelect';
 import { FieldValidationProps } from '@odh-dashboard/internal/hooks/useZodFormValidation';
 import { ZodErrorHelperText } from '@odh-dashboard/internal/components/ZodErrorFormHelperText';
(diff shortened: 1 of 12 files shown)


packages/model-serving/src/components/deploymentWizard/fields/ModelLocationSelectField.tsx
```diff
@@ -268,6 +254,9 @@ export const ModelLocationSelectField: React.FC<ModelLocationSelectFieldProps> =
     if (uriConnectionTypes.length > 0 || hasURISelected) {
       options.push({ key: uriOption.key, label: uriOption.label });
     }
+    if (isNimWizardEnabled) {
+      options.push(NIMModelLocationOption);
```
mturley, line 234: NIM is added to the dropdown options here, but the validation layer doesn't support it yet:

1. `modelLocationSelectFieldSchema` (line 36) uses `z.enum([EXISTING, NEW, PVC])`. NIM is not included, so selecting it will fail Zod validation and block wizard progression.

2. `isValidModelLocation()` (line 44) only checks for `EXISTING`, `NEW`, and `PVC`. Any caller using this guard to validate a NIM value gets `false`.

3. `currentKey` (line 160) only maps `PVC` and `EXISTING` from `modelLocation`. If the component remounts while NIM is selected (e.g. wizard step navigation), the dropdown won't show NIM as the current selection.

We might want to either add NIM to the Zod enum and `isValidModelLocation`, or restructure validation to handle NIM as a separate path (since NIM isn't really a "model location" in the connection-type sense). For `currentKey`, we should add `modelLocation === ModelLocationType.NIM` to the fallback.

We could also just handle this in a followup.

packages/model-serving/src/components/deploymentWizard/steps/ModelSourceStep.tsx
```diff
@@ -7,21 +7,33 @@ import { modelTypeSelectFieldSchema, ModelTypeSelectField } from '../fields/Mode
...
+    modelType: modelTypeSelectFieldSchema,
+    modelLocationData: z.custom<ModelLocationData>((val) => {
+      if (!val) return false;
+      return isValidModelLocationData(val.type, val);
+    }),
+    createConnectionData: createConnectionDataSchema.optional(),
+  })
+  // If the model location data is a new connection, then the create connection data is required
+  .superRefine((data, ctx) => {
+    if (data.modelLocationData.type === ModelLocationType.NEW) {
```
mturley: `data.modelLocationData` comes from the `z.custom()` validator above, which can return `false` for invalid data. If it does, `data.modelLocationData` could be `undefined` here, and accessing `.type` without optional chaining would throw.

```suggestion
    if (data.modelLocationData?.type === ModelLocationType.NEW) {
```

mturley (changes requested):

mturley:
One other thing: The alert I added at `ModelSourceStep.tsx:58` (`"(TODO replace me) NIM serving extension is enabled"`) should probably be removed now that we have real content gated by the feature flag, but I can do that in my PR if you want.

packages/model-serving/src/components/deploymentWizard/fields/ModelLocationSelectField.tsx
emilys314: The first two are actually (mostly) dead code. It was validating successfully on step 1. I removed those

the 3rd, thanks for the catch, i accidentally dropped this fix when reverting some stuff, should be fixed

packages/model-serving/src/components/deploymentWizard/steps/ModelSourceStep.tsx
emilys314: Fixed

mturley (approved):
Tested and LGTM

PR #7456, opened 2026-05-04
fix: correct gen-ai Tekton pipeline triggers

christianvogt:
## Description

The gen-ai Tekton pipeline files had a leading space in the `pathChanged()` CEL expression (`" packages/gen-ai/**"` instead of `"packages/gen-ai/**"`), which prevented the pipelines from ever triggering on gen-ai source changes.

Also aligns `cancel-in-progress` and CEL expression formatting with the conventions used by the other mod-arch packages.

Changes:
- Remove leading space in `pathChanged()` glob so the filter actually matches gen-ai file changes
- Set `cancel-in-progress` to `"false"` on push / `"true"` on pull-request (was `"true"` for both)
- Reformat CEL expression as a literal block scalar (`|`) for readability

## How Has This Been Tested?

Verified the YAML is valid and the CEL expressions match the pattern used by the other working Tekton pipeline files (mlflow, eval-hub, modular-architecture).

## Test Impact

No tests applicable — CI/CD configuration only.

## Request review criteria:


jenny-s51 (approved):
/lgtm

PR #7457, opened 2026-05-04
Add event tracking to MLflow - HIGH

nananosirova:
https://redhat.atlassian.net/browse/RHOAIENG-57637

## Description

Adds Segment telemetry tracking for MLflow embedded view navigation. This introduces a single event, `MLflow Embedded View Opened`, that fires when users navigate to the embedded MLflow experiments view from three pathways:

- Sidebar: clicking "Experiments (MLflow)" in the left nav
- Project overview: clicking "Go to Experiments" on the Experiment tracking card
- Pipeline run table: clicking an MLflow experiment name link 

## How Has This Been Tested?

Verified each pathway fires `MLflow Embedded View Opened` in the browser console (DevTools > Console):
- Sidebar: navigate to any page, then click "Experiments (MLflow)" in the left nav. Check the console for `Telemetry event triggered: MLflow Embedded View Opened` with `section: "sidebar-nav"`.
- Project overview: go to Data Science Projects > open a project > Overview tab > scroll to the "Train models" section > click "Go to Experiments" on the Experiment tracking card. Check the console for the same event with `section: "project-overview"`.
- Pipeline run table (turn on `mlflowPipelines` dev flag): go to a project's Pipelines > Runs tab > find a run with an MLflow experiment name in the "MLflow experiment" column > click the experiment name link. Check the console for the same event with `section: "pipeline-run-table"`.

## Request review criteria:

If you have UI changes: 

## Summary by CodeRabbit

* **New Features**
  * Navigation items and MLflow links now emit analytics events when clicked, capturing source and section context.
* **Chores**
  * Centralized MLflow-related tracking events introduced and integrated across embedded views and launch actions.
  * Navigation extensions updated to carry optional tracking metadata for UI pathways.


DaoDaoNoCode:
@nananosirova can you resolve conflicts?

frontend/src/app/navigation/NavItemHref.tsx
```diff
@@ -33,9 +34,13 @@ export const NavItemHref: React.FC<Props> = ({
     [],
   );
 
+  const handleClick = React.useCallback(() => {
```
srtanish1992, line 46: The handler fires on every click, even if the nav item is already the active route (i.e., isMatch === true). This could inflate analytics counts if users re-click the current page. Consider gating with if (!isMatch) if the intent is to track navigation to the view.

nananosirova: Updated, thanks

DaoDaoNoCode (commented):
This adds Segment telemetry for the MLflow embedded view across three navigation pathways (sidebar, project overview, pipeline run table) and centralizes the existing `'Launch MLflow clicked'` string into an enum. The approach works, and the `isMatch` guard (addressing the earlier review feedback) is a good addition.

Local checks: lint pass, type-check pass, unit tests pass.

Leaving a few comments inline -- the main thing is the `window.__navClickHandlers` global registry and the side-effect in `extensions.ts`, which I think deserves discussion before merging.

frontend/src/concepts/analyticsTracking/navClickHandlers.ts
```diff
@@ -0,0 +1,32 @@
...
+  interface Window {
+    __navClickHandlers?: Map<string, () => void>;
+  }
+}
+
+const getHandlers = (): Map<string, () => void> => {
+  if (!window.__navClickHandlers) {
+    window.__navClickHandlers = new Map();
+  }
+  return window.__navClickHandlers;
```
DaoDaoNoCode: Using `window.__navClickHandlers` to work around Module Federation's separate module scopes is understandable, but it introduces a global mutable registry with no cleanup, no collision protection, and no validation. A few concerns:

1. Any code anywhere can silently overwrite a handler via `navClickHandlers.set('experiments-mlflow', ...)` -- there's no warning if a key is already registered.
2. Handlers are never removed. If the mlflow-embedded module is unloaded or re-loaded (e.g. during hot module replacement), stale handlers accumulate.
3. The `window` global is shared with third-party scripts, extensions, and dev tools. A generic name like `__navClickHandlers` could collide.

For now, consider at minimum adding a `console.warn` if `set` is called with a key that already exists, so accidental overwrites are visible during development:

```suggestion
const getHandlers = (): Map<string, () => void> => {
  if (!window.__navClickHandlers) {
    window.__navClickHandlers = new Map();
  }
  return window.__navClickHandlers;
};

const navClickHandlers = {
  set: (id: string, handler: () => void): void => {
    const handlers = getHandlers();
    if (handlers.has(id) && process.env.NODE_ENV !== 'production') {
      // eslint-disable-next-line no-console
      console.warn(`[navClickHandlers] Overwriting existing handler for "${id}"`);
    }
    handlers.set(id, handler);
  },
  get: (id: string): (() => void) | undefined => getHandlers().get(id),
};
```

Longer term, a cleaner alternative might be to add an `onClick` callback property to the `HrefNavItemExtension` type itself, so packages can declare their click handler in the extension definition without needing a separate global registry. That would keep everything declarative and avoid the window global entirely.

packages/mlflow-embedded/extensions.ts
```diff
@@ -12,8 +12,21 @@ import {
...
+import navClickHandlers from '@odh-dashboard/internal/concepts/analyticsTracking/navClickHandlers';
+// eslint-disable-next-line no-restricted-syntax
 import { PROMPT_MANAGEMENT_PAGE_TITLE } from './shared/const';
 
+navClickHandlers.set('experiments-mlflow', () => {
+  fireLinkTrackingEvent(MlflowTrackingEvents.EMBEDDED_VIEW_OPENED, {
+    from: window.location.pathname,
+    section: 'sidebar-nav',
+  });
+});
```
DaoDaoNoCode: This is the only `extensions.ts` file in the entire repo that executes a side-effect at module scope. Every other extensions file is purely declarative -- it exports a `const` array and nothing else.

The ESLint rule on extension files (`no-restricted-syntax` with `ImportDeclaration[importKind!='type']`) exists specifically to enforce that pattern: extensions should only use `import type` and lazy `() => import(...)` code refs. This PR adds four `eslint-disable-next-line` comments to work around that rule, which is a signal that the approach is fighting the architecture.

Is there a reason this can't be handled inside a component instead? For example, the `GlobalMLflowExperimentsRoutes` component (or a parent layout component for the MLflow section) could fire the tracking event on mount when the route is entered via sidebar navigation. That would avoid both the window global and the extension-file side effect.

frontend/src/app/navigation/NavItemHref.tsx
```diff
@@ -33,9 +34,15 @@ export const NavItemHref: React.FC<Props> = ({
...
+    if (!isMatch) {
+      navClickHandlers.get(id)?.();
```
DaoDaoNoCode: nit: The `navClickHandlers` import and lookup run for *every* nav item on *every* render, but only `experiments-mlflow` has a registered handler today. The overhead is negligible (a Map lookup returns `undefined` immediately), but it's worth noting that this couples a generic navigation component to a telemetry side-channel. If more packages register handlers, this becomes a de facto event bus on the nav component.

If you keep this approach, consider documenting which IDs have registered handlers (even a comment here would help future readers understand why `navClickHandlers` exists).

frontend/src/concepts/pipelines/content/tables/pipelineRun/PipelineRunTableRowMlflowExperiment.tsx
```diff
@@ -42,6 +44,12 @@ const PipelineRunTableRowMlflowExperiment: React.FC<PipelineRunTableRowMlflowExp
       <Link
         to={mlflowExperimentRoute(experimentId, namespace)}
         data-testid="mlflow-experiment-link"
+        onClick={() =>
+          fireLinkTrackingEvent(MlflowTrackingEvents.EMBEDDED_VIEW_OPENED, {
```
DaoDaoNoCode: nit: The `onClick` handler creates a new closure on every render. Since this is inside a table row that could appear many times, you could extract it to a stable `useCallback`:

```suggestion
        onClick={React.useCallback(() =>
          fireLinkTrackingEvent(MlflowTrackingEvents.EMBEDDED_VIEW_OPENED, {
            from: window.location.pathname,
            section: 'pipeline-run-table',
          }), [])}
```

Or define it outside the JSX. Not a big deal since `fireLinkTrackingEvent` is a fire-and-forget call, but it keeps things consistent with the `useCallback` pattern used in `NavItemHref`.

frontend/src/concepts/analyticsTracking/navClickHandlers.ts
```diff
@@ -0,0 +1,32 @@
+/**
```
DaoDaoNoCode: There are no unit tests for this new module. Since it's a utility with get/set semantics and a global Map, it would be straightforward to test:

- `set` registers a handler that `get` can retrieve
- `get` returns `undefined` for unknown IDs
- `set` with the same ID overwrites the previous handler

Worth adding a `navClickHandlers.spec.ts`.

frontend/src/concepts/analyticsTracking/navClickHandlers.ts
nananosirova: This should no longer be an issue, replaced `navClickHandlers` with a `trackingEvent` property in`HrefNavItemExtension` instead.

packages/mlflow-embedded/extensions.ts
nananosirova: This should no longer be an issue, replaced `navClickHandlers` with a `trackingEvent` property in`HrefNavItemExtension` instead. `extensions.ts` is back to being declarative.

frontend/src/app/navigation/NavItemHref.tsx
nananosirova: This should no longer be an issue, replaced `navClickHandlers` with a `trackingEvent` property in`HrefNavItemExtension` instead.

frontend/src/concepts/analyticsTracking/navClickHandlers.ts
nananosirova: Removed `navClickHandlers.ts` instead

frontend/src/app/navigation/NavItemHref.tsx
```diff
@@ -33,9 +43,18 @@ export const NavItemHref: React.FC<Props> = ({
...
   );
 
+  const handleClick = React.useCallback(() => {
+    if (!isMatch && trackingEvent) {
+      fireLinkTrackingEvent(trackingEvent.name, {
+        from: window.location.pathname,
+        section: trackingEvent.section,
+      });
+    }
+  }, [isMatch, trackingEvent]);
```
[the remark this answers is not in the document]
nananosirova: Why wouldn't we want to track that someone opened the view in a tab?

frontend/src/pages/projects/screens/detail/overview/trainModels/MLflowCard.tsx
```diff
@@ -80,6 +81,12 @@ const MLflowCard: React.FC = () => {
               component={(props: React.ComponentProps<'a'>) => (
                 <Link {...props} to={mlflowExperimentsPath} />
               )}
+              onClick={() =>
```
DaoDaoNoCode: nit: this inline arrow is the only one of the three tracking pathways that isn't wrapped in `useCallback`. The other two (NavItemHref and PipelineRunTableRowMlflowExperiment) both use `useCallback` — worth being consistent here for future readers.

```suggestion
              onClick={React.useCallback(() => {
                fireLinkTrackingEvent(MlflowTrackingEvents.EMBEDDED_VIEW_OPENED, {
                  from: window.location.pathname,
                  section: 'project-overview',
                });
              }, [])}
```

DaoDaoNoCode (commented):
Adds `MLflow Embedded View Opened` telemetry from three navigation pathways (sidebar nav, project overview card, pipeline run table). The approach is clean — replacing the earlier `window.__navClickHandlers` global with a declarative `trackingEvent` property on `HrefNavItemExtension` is a much better design, and centralizing event names into `MlflowTrackingEvents` is a good housekeeping win.

Local checks: lint ✅ type-check ✅ unit tests ✅ (CI Cypress failures are an infra flake — Cypress binary not cached, unrelated to this PR)
mlflow/224 cross-ref: consistent — that PR uses `fireFormTrackingEvent` for form workflows (create experiment/run); this PR correctly uses `fireLinkTrackingEvent` for navigation events. Event naming is consistent across both PRs.

Prior review: DaoDaoNoCode raised 5 concerns (global navClickHandlers, extensions.ts side-effects, NavItemHref coupling, PipelineRunTableRowMlflowExperiment useCallback, missing unit tests) — all resolved in the current code. srtanish1992's isMatch double-fire concern also resolved.

Left one nit inline, nothing blocking.

frontend/src/pages/projects/screens/detail/overview/trainModels/MLflowCard.tsx
```diff
@@ -80,6 +81,12 @@ const MLflowCard: React.FC = () => {
...
+                fireLinkTrackingEvent(MlflowTrackingEvents.EMBEDDED_VIEW_OPENED, {
+                  from: window.location.pathname,
+                  section: 'project-overview',
+                })
+              }
```
DaoDaoNoCode: nit: The other two call sites (`NavItemHref`, `PipelineRunTableRowMlflowExperiment`) wrap their tracking callback in `useCallback`. This one creates a new closure on every render. It's fine here since the card isn't rendered in a list, but for consistency you could extract it.

```suggestion
  const handleGoToExperiments = React.useCallback(() => {
    fireLinkTrackingEvent(MlflowTrackingEvents.EMBEDDED_VIEW_OPENED, {
      from: window.location.pathname,
      section: 'project-overview',
    });
  }, []);
```

Then use `onClick={handleGoToExperiments}` below. Not blocking.

packages/mlflow-embedded/extensions.ts
```diff
@@ -12,6 +12,11 @@ import {
   promptManagementPath,
 } from '@odh-dashboard/internal/routes/pipelines/mlflow';
 // eslint-disable-next-line no-restricted-syntax
+import {
+  EXPERIMENTS_NAV_ID,
+  MlflowTrackingEvents,
+} from '@odh-dashboard/internal/concepts/mlflow/const';
```
DaoDaoNoCode, line 18: nit: There are now two consecutive `eslint-disable-next-line no-restricted-syntax` comments. The first (line 15) covers the route imports and the second (line 18) covers the const imports. Totally fine, just noting it — if more imports from `@odh-dashboard/internal` land here you might want to group them under a single disable block.

frontend/src/app/navigation/NavItemHref.tsx
```diff
@@ -33,9 +43,18 @@ export const NavItemHref: React.FC<Props> = ({
     [],
   );
 
+  const handleClick = React.useCallback(() => {
+    if (!isMatch && trackingEvent) {
+      fireLinkTrackingEvent(trackingEvent.name, {
+        from: window.location.pathname,
+        section: trackingEvent.section,
```
DaoDaoNoCode, line 50: Nice approach. The `!isMatch` guard correctly prevents double-firing when the user clicks the already-active nav item. The `useCallback` deps `[isMatch, trackingEvent]` are correct — `trackingEvent` comes from extension properties which are stable, and `isMatch` is the reactive guard.

packages/plugin-core/src/extension-points/navigation.ts
```diff
@@ -2,6 +2,13 @@ import type { Extension } from '@openshift/dynamic-plugin-sdk';
 import type { AccessReviewResourceAttributes } from '@odh-dashboard/internal/k8sTypes';
 import type { ComponentCodeRef } from '../core/types';
```
DaoDaoNoCode, line 4: Clean type design. Keeping it as `{ name, section }` is flexible enough that other nav extensions can adopt it without needing MLflow-specific knowledge. Good public API addition.

DaoDaoNoCode (commented):
Adds Segment telemetry for "MLflow Embedded View Opened" from three pathways (sidebar nav, project overview, pipeline run table) and centralizes the existing "Launch MLflow clicked" string into an `MlflowTrackingEvents` enum. Clean, well-structured change with a nice declarative `trackingEvent` extension on `HrefNavItemExtension`.

Local checks: lint ✅  type-check ✅  unit tests ✅
MLflow/224 cross-ref: Naming and utility choices are consistent — mlflow/224 uses `fireFormTrackingEvent` for form actions, this PR correctly uses `fireLinkTrackingEvent` for navigation.

Prior review: DaoDaoNoCode raised concerns about the `window.__navClickHandlers` global pattern on May 8. Author completely reworked the approach — replaced it with a declarative `trackingEvent` property on `HrefNavItemExtension`. Verified in code, the old `navClickHandlers.ts` file is gone. All threads resolved.

Looks good overall. Left a couple of nits inline, nothing blocking.

DaoDaoNoCode (approved):
/lgtm

PR #7459, opened 2026-05-04
Pass workspace context to external MLflow UI on launch

nananosirova:
https://redhat.atlassian.net/browse/RHOAIENG-60053

## Description

Launching the external MLflow UI via the "Launch MLflow" button caused the user to lose their current project context. The button always opened a fixed `/mlflow` URL without any workspace information, so the external MLflow UI had no way to know which project the user was working in.
This PR passes the active workspace as a hash fragment query parameter so the external MLflow UI lands in the correct project.

## How Has This Been Tested?

Manually verified on cluster that clicking "Launch MLflow" from the Experiments and Prompts pages preserves workspace context.

## Test Impact

Added unit tests.

## Request review criteria:

If you have UI changes: 

## Summary by CodeRabbit

* **New Features**
  * MLflow launch button now generates workspace-specific URLs to provide proper context when launching MLflow from different workspaces.

* **Tests**
  * Added comprehensive tests covering workspace route generation, URL encoding, and edge cases for MLflow launch functionality.

### DIFF
--- a/frontend/src/pages/projects/screens/detail/overview/trainModels/__tests__/MLflowCard.spec.ts
+++ b/frontend/src/pages/projects/screens/detail/overview/trainModels/__tests__/MLflowCard.spec.ts
@@ -2,6 +2,7 @@ import {
   setWorkspaceQueryParam,
   buildMLflowExperimentsWorkspaceHref,
 } from '#~/pages/projects/screens/detail/overview/trainModels/MLflowCard';
+import { mlflowLaunchRoute } from '#~/routes/pipelines/mlflow';
 
 describe('setWorkspaceQueryParam', () => {
   it('should add workspace param to a path without query string', () => {
@@ -40,6 +41,30 @@ describe('setWorkspaceQueryParam', () => {
   it('should handle workspace name with ampersand', () => {
     const result = setWorkspaceQueryParam('/experiments', 'a&b');
     expect(result).toBe('/experiments?workspace=a%26b');
+  });
+});
+
+describe('mlflowLaunchRoute', () => {
+  it('should return base path when no namespace is provided', () => {
+    expect(mlflowLaunchRoute()).toBe('/mlflow');
+  });
+
+  it('should return base path when namespace is undefined', () => {
+    expect(mlflowLaunchRoute(undefined)).toBe('/mlflow');
+  });
+
+  it('should append workspace in hash fragment when namespace is provided', () => {
+    expect(mlflowLaunchRoute('my-project')).toBe('/mlflow/#/?workspace=my-project');
+  });
+
+  it('should URL-encode special characters in namespace', () => {
+    expect(mlflowLaunchRoute('project with spaces')).toBe(
+      '/mlflow/#/?workspace=project%20with%20spaces',
+    );
+  });
+
+  it('should URL-encode ampersands in namespace', () => {
+    expect(mlflowLaunchRoute('a&b')).toBe('/mlflow/#/?workspace=a%26b');
   });
 });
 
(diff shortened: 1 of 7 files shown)


srtanish1992:
/lgtm

Code changes LGTM. Manually tested and work as expected.

The only difference I see is, in dashboard we show project display name in project selector and in MLflow we show the kubernetes namespace/workspace so they do not match.

https://github.com/user-attachments/assets/430ff61c-8cd7-4b46-9473-e94001c17363

frontend/src/pages/projects/screens/detail/overview/trainModels/__tests__/MLflowCard.spec.ts
```diff
@@ -43,6 +44,30 @@ describe('setWorkspaceQueryParam', () => {
   });
 });
 
+describe('mlflowLaunchRoute', () => {
```
DaoDaoNoCode, line 47: These tests are for `mlflowLaunchRoute`, which lives in `frontend/src/routes/pipelines/mlflow.ts`. Per the project's co-location convention, they'd be better placed in `frontend/src/routes/pipelines/__tests__/mlflow.spec.ts`.

Not blocking since the tests themselves are solid, but worth moving in this PR or a quick follow-up so people looking for route tests find them in the expected place.

nit: CodeRabbit flagged this too -- adding a `mlflowLaunchRoute('')` case would be a nice addition since both callers default `workspace` to `''` via `searchParams.get(...) ?? ''`. The `!''` check in JS is truthy (so it'd return `/mlflow`), which is the right behavior, but documenting that in a test makes the intent explicit.

frontend/src/routes/pipelines/mlflow.ts
```diff
@@ -32,3 +32,10 @@ export const globPromptManagementAll = `${promptManagementPath}/*`;
 
 export const mlflowPromptManagementBaseRoute = (namespace?: string): string =>
   withWorkspace(promptManagementPath, namespace);
+
+export const mlflowLaunchRoute = (namespace?: string): string => {
+  if (!namespace) {
+    return MLFLOW_PROXY_BASE_PATH;
+  }
+  return `${MLFLOW_PROXY_BASE_PATH}/#/?${WORKSPACE_QUERY_PARAM}=${encodeURIComponent(namespace)}`;
```
DaoDaoNoCode, line 40: The URL structure here -- `/mlflow/#/?workspace=...` -- differs from `withWorkspace` above (which uses `?workspace=...` without the hash fragment). That's intentional because the external MLflow UI uses hash-based routing while the embedded pages use regular routing, so this makes sense.

Just wanted to confirm: does the external MLflow UI actually read `workspace` from `window.location.hash` at the root path (`/#/?workspace=...`)? `buildMLflowExperimentsWorkspaceHref` in MLflowCard.tsx builds `/#/experiments?workspace=...` instead (with a sub-path in the hash). If the external UI only looks for workspace on specific hash routes like `/#/experiments`, this root-level hash might not be picked up.

nit: `encodeURIComponent` is overkill for Kubernetes namespace names (RFC 1123: lowercase alphanumeric + hyphens only), but it's the right defensive choice -- no harm in encoding.

DaoDaoNoCode (commented):
Clean, focused change that threads the workspace through the LaunchMlflowButton so the external MLflow UI opens with the right project context. Logic is straightforward, tests cover the new function well, all CI checks pass.

Local checks: lint (0 errors) -- type-check -- unit tests -- all green.

Left a couple of comments inline -- one about test file placement and one question about the hash-based URL structure. Also one broader observation below about a related link in MLflowCard.tsx (not part of this diff) that could be a quick follow-up.

DaoDaoNoCode:
~Side observation (not part of this diff): In `MLflowCard.tsx` line 81, the "Go to Experiments" link navigates to `mlflowExperimentsPath` (`/develop-train/mlflow/experiments`) without passing the current project's workspace. Since the user is on a project details page at that point, they'd lose their project context when clicking that link. Using `mlflowExperimentsBaseRoute(projectName)` there would carry the workspace forward. Totally fine as a follow-up since it's pre-existing behavior.~
Invalid AI review. Removed

DaoDaoNoCode (approved):
Clean fix, LGTM
/lgtm

PR #7466, opened 2026-05-05
feat(nim-serving): add NIM image selector to deployment wizard

mturley:
https://issues.redhat.com/browse/RHOAIENG-60277

https://github.com/user-attachments/assets/4f657467-e4ac-4f89-b58d-04b26d7d0d5a

## Description
Adds a NIM image dropdown field to the Model details step of the deployment wizard. When the user selects the NIM model location option (added in #7453), a TypeaheadSelect dropdown appears showing available NVIDIA NIM container images fetched from the project's NIMAccount ConfigMap.

**What this PR does:**
- Adds `packages/nim-serving/src/api/images/` — types, utilities, and data-fetching hooks for NIM images from the NIMAccount ConfigMap
- Adds `packages/nim-serving/src/pages/deploymentWizard/fields/NIMImageField.tsx` — a `WizardField` extension with:
  - A `TypeaheadSelect` dropdown showing NIM images as `{displayName} - {version}` options
  - A `useNIMImageFieldExternalData` wrapper hook that combines NIM image data and account status into a single external data object (`{ nimImages, accountStatus }`), eliminating duplicate API calls — account data is fetched once by `useNIMAccountStatus` and passed to `useNIMImages` as a dependency
  - Structured field value (`{ repository, tag }`) matching KServe's `spec.image` structure, avoiding string parsing
  - A skeleton loading state while account status and images are loading, preventing flicker
  - An error Alert when NIM is unconfigured, with RBAC-gated action via the extracted `NIMSettingsLink` component: users who can see project settings get a "Configure in project settings" link; others see "Ask your project administrator to configure NVIDIA NIM"
  - Reads `accountStatus` from external data to distinguish "no key configured" (NOT_FOUND) from "invalid key" (ERROR) and shows different messages for each
  - An info Alert when no project is selected yet
  - `isActive` gated on `NIMModelLocationKey` so the field only appears when NIM is selected as the model source
- Extracts `NIMSettingsLink` component to `pages/projectSettings/` — encapsulates the RBAC check (reusing `useProjectPermissionsTabVisible`) and the "Configure in project settings" link using `routeProjectsNamespace` and `ProjectSectionID.SETTINGS`
- Extracts `NIM_IMAGE_REGISTRY` constant for the `nvcr.io` registry URL
- Extracts `getNIMImageOptions` and `getImageOptionKey` helpers for option construction and deduplication
- Makes `NIMImage` fields optional (except `name`) for ConfigMap safety — the parsed JSON has no schema guarantee. Entries missing `namespace` or `tags` are silently skipped during parsing.
- Uses `getNIMAccount()` from `api/accounts/k8s.ts` to find the NIM account by name rather than blindly taking the first account in the list
- `fetchNIMImages` accepts a `NIMAccountKind` object directly instead of fetching it, and `useNIMImages` throws `NotReadyError` until the account is loaded — this ensures only one `listNIMAccounts` call per render cycle
- Registers the field as a `model-serving.deployment/wizard-field2` extension gated behind the `nim-wizard` feature flag and `SupportedArea.NIM_MODEL`
- Splits the Model details step Zod schema into `modelSourceStepBaseSchema` (extendable ZodObject) and `modelSourceStepRefinement` (reapplied after extending) so plugin fields can add validation to the Model details step
- Removes the `useMemo` wrapper from `modelSourceStepValidationData` — the original code memoized this object, but adding dynamic plugin field values to the deps array isn't feasible (spread in deps isn't supported by `react-hooks/exhaustive-deps`). This is safe because `useValidation` internally does property-level diffing via refs and only re-validates changed properties, and the step 2 validation data already follows this same unwrapped pattern.
- Renders extension fields in the Model details step via `GenericFieldRenderer`
- Extracts `hasOnlyExtensionFields()` helper in `ModelLocationInputFields` to identify model locations that defer all UI to plugin extensions (currently NIM). Used to return `null` from the component and to skip rendering the empty `StackItem` wrapper in `ModelLocationSelectField`, fixing extra vertical spacing in the form

### Screenshots

Error alert — NIM key not configured (NOT_FOUND), with "Configure in project settings" link (admin user):

04-error-no-key-admin

Error alert — NIM key not configured (NOT_FOUND), with "Ask your project administrator" text (non-admin user):

05-error-no-key-non-admin

Error alert — NIM key invalid (ERROR), with "Configure in project settings" link (admin user):

06-error-invalid-key-admin

Error alert — NIM key invalid (ERROR), with "Ask your project administrator" text (non-admin user):

07-error-invalid-key-non-admin

NIM image selector field (project with NIM configured):

08-nim-image-selector

NIM dropdown expanded showing available NIM images:

09-nim-dropdown-expanded

Info alert when no project is selected:

03-info-alert-no-project

NIM field hidden when switching model location to S3 object storage:

10-nim-hidden-s3

## Future Integration

This PR captures the user's NIM image selection in wizard state (`{ repository: string; tag: string }`) — it does not yet map that selection onto a Kubernetes resource at submission time. The field component and data hook are designed to work unchanged once the submission integration is added.

When NIMService becomes a first-class deployment platform, the following extensions need to be registered in `packages/nim-serving/extensions.ts`:

**Apply extension** (`model-serving.deployment/wizard-field-apply`):
- Links to `fieldId: 'nim-serving/nimImage'` and `platform: NIM_SERVING_ID`
- Implements an apply function with signature `(deployment: NIMDeployment, fieldData: NIMImageFieldValue, wizardState) => NIMDeployment`
- Maps `fieldData.repository` and `fieldData.tag` onto the NIMService resource (e.g. setting `spec.image.repository` and `spec.image.tag`)
- Follow the pattern in `packages/kserve/src/wizardFields/timeout/timeoutApplyExtract.ts`

**Extract extension** (`model-serving.deployment/wizard-field-extractor`):
- Links to the same `fieldId` and `platform`
- Implements an extract function with signature `(deployment: NIMDeployment) => NIMImageFieldValue | undefined`
- Reads `spec.image.repository` and `spec.image.tag` from an existing NIMService resource back into `{ repository, tag }` for the edit flow

**Reference implementation:** The KServe timeout field provides the complete pattern:
- Field definition: `packages/kserve/src/wizardFields/timeout/TimeoutField.tsx`
- Apply/extract functions: `packages/kserve/src/wizardFields/timeout/timeoutApplyExtract.ts`
- Extension registration: `packages/kserve/extensions.ts:207-234`

## How Has This Been Tested?
- `npm run lint` — passes
- `npm run type-check` — passes
- `npm run test-unit` in `packages/nim-serving` — all tests pass
- `npm run test-unit` in `packages/model-serving` — all 200 tests pass
- Manually verified in browser with dev server:
  - NIM option hidden when feature flag is disabled
  - NIM option appears when `disableNIMModelServing` flag is enabled
  - Selecting NVIDIA NIM shows the NIM image typeahead field
  - Switching to another location (e.g. S3) correctly hides the NIM field
  - Skeleton field shown while account status and images load
  - Info alert shown when no project is selected
  - Error alert shown with correct message when NIM key is not configured (NOT_FOUND)
  - Error alert shown with correct message when NIM key is invalid (ERROR)
  - "Configure in project settings" link shown for users with rolebinding list permission
  - "Ask your project administrator" text shown for users without permission
  - Dropdown populated with real NIM images when NIM is configured

## Test Impact
- **Utility tests** (`src/api/images/__tests__/utils.spec.ts`): 7 tests covering `normalizeVersion` (version padding, non-numeric tags) and `getImageRepository` (repository construction)
- **Component tests** (`src/pages/deploymentWizard/fields/__tests__/NIMImageField.spec.tsx`): 8 tests covering skeleton loading state, all alert states — info alert (no project), error alerts for NOT_FOUND/ERROR with RBAC-gated action (admin settings link vs non-admin ask-admin text), spinner during RBAC loading, and typeahead selector when NIM is configured
- Cypress mock tests for the NIM image field will be added in follow-up PRs

## Request review criteria:
If you have UI changes:


packages/nim-serving/src/utils/nimModels.ts
```diff
@@ -0,0 +1,110 @@
...
+export const getNIMImageName = (
+  modelNamespace: string,
+  modelName: string,
+  version: string,
+): string => `nvcr.io/${modelNamespace}/${modelName}:${version}`;
+
+export const fetchNIMModelNames = async (dashboardNamespace: string): Promise<ModelInfo[]> => {
+  const accounts = await k8sListResource<NIMAccountKind>({
+    model: NIMAccountModel,
+    queryOptions: { ns: dashboardNamespace },
```
emilys314:
```suggestion
    queryOptions: { ns: projectNamespace },
```
or
```suggestion
    queryOptions: { ns: namespace },
```

I think you are passing in the correct ds project namespace anyways so no issue, but we should clarify we don't want a global account here

packages/nim-serving/src/wizardFields/NIMImageField.tsx
```diff
@@ -0,0 +1,257 @@
...
+      )
+      .filter((option): option is TypeaheadSelectOption => option !== null);
+  }, [modelInfos]);
+
+  const selectedKey = React.useMemo(() => {
+    if (!value?.imageName) {
+      return '';
+    }
+    return (
+      options
```
emilys314: A defensive return of the existing value if it's not found in the `options` would be a good idea. For example if you are editing an existing NIM model but deactivate NIM.

packages/nim-serving/src/wizardFields/NIMImageField.tsx
```diff
@@ -0,0 +1,257 @@
...
+  loaded: boolean;
+  loadError?: Error;
+} => {
+  const project = dependencies?.project;
+  const projectName = project?.projectName;
+  const [modelInfos, setModelInfos] = React.useState<ModelInfo[]>([]);
+  const [loaded, setLoaded] = React.useState(false);
+  const [loadError, setLoadError] = React.useState<Error | undefined>();
+
+  React.useEffect(() => {
```
emilys314: re-using the `useFetchState` would be nice and easier to read. 

It also avoids us having to explicitly use a `useEffect` (which I always try to avoid since they are easy to abuse)

emilys314 (commented):
this is still a wip but here are some quick comments

mturley:
/hold
looking at feedback

packages/nim-serving/src/utils/nimModels.ts
```diff
@@ -0,0 +1,110 @@
...
+
+export type ModelInfo = {
+  name: string;
+  displayName: string;
+  shortDescription: string;
+  namespace: string;
+  tags: string[];
+  latestTag: string;
+  updatedDate: string;
+};
```
emilys314: I wonder if all of these are required fields? It's a configmap so I don't think there is a guarantee, we can do some type parsing / assert from the api response for maximum safety

packages/nim-serving/src/utils/nimModels.ts
mturley: Fixed

packages/nim-serving/src/wizardFields/NIMImageField.tsx
mturley: Fixed

packages/nim-serving/src/wizardFields/NIMImageField.tsx
mturley: Fixed

packages/nim-serving/src/utils/nimModels.ts
mturley: Made them optional and made references to the properties defensive

packages/nim-serving/src/pages/deploymentWizard/fields/NIMImageField.tsx
```diff
@@ -0,0 +1,223 @@
...
+import type { NIMModelInfo } from '../../../api/models/types';
+import { getNIMImageName, normalizeVersion } from '../../../api/models/utils';
+
+export type NIMImageDependencies = {
+  project: ProjectSectionType;
+};
+
+export type NIMImageFieldValue = {
+  imageName: string;
+};
```
emilys314, line 64: I think it would be safer and cleaner to store more than just a string, because parsing strings is always less safe.

The minimum info we get is from editing a NIMService which will have the tag separated from the image
```    
spec.image:
    repository: nvcr.io/nim/meta/llama-3.2-1b-instruct
    tag: 1.8.3
```

So if you stored this in the NIMImageFieldValue, that already removes the need for the `extractModelAndVersion()`. We'd still need a key for the options though that is the entire concatenated string


```suggestion
export type NIMImageFieldOption = TypeaheadSelectOption &  {
  key: string;
  label: string;
  imageName: string;
  imageTag: string;
  ... more?
};
```

packages/nim-serving/src/pages/deploymentWizard/fields/NIMImageField.tsx
```diff
@@ -0,0 +1,223 @@
...
+    accountLoaded &&
+    (accountStatus === NIMAccountStatus.NOT_FOUND || accountStatus === NIMAccountStatus.ERROR) &&
+    modelInfos.length === 0;
+
+  if (isNIMUnconfigured) {
+    const isInvalidKey = accountStatus === NIMAccountStatus.ERROR;
+
+    const settingsAction = !rbacLoaded ? (
+      <Spinner size="sm" />
+    ) : canViewSettings ? (
```
emilys314: We should do something similar for the actual field for when the fetching of nim images is happening. Also a skeleton field might be less jarring (at least for the field)? 

This is a throttled example

https://github.com/user-attachments/assets/8ae8deea-68d8-494e-9032-86125019a326

packages/nim-serving/src/pages/deploymentWizard/fields/NIMImageField.tsx
```diff
@@ -0,0 +1,223 @@
...
+      <Link to={`/projects/${projectName}?section=settings`}>Configure in project settings</Link>
```
emilys314: It might be good to extract this out into a function that lives in the projectSettings folder. The projectSettings card is the source of truth for where it lives and not this field

packages/nim-serving/src/pages/deploymentWizard/fields/NIMImageField.tsx
```diff
@@ -0,0 +1,223 @@
...
+      <Alert variant="info" isInline title="NVIDIA Inference Microservices (NIM image)">
+        Select a project to load available NIM images.
+      </Alert>
+    );
+  }
+
+  const isNIMUnconfigured =
+    accountLoaded &&
+    (accountStatus === NIMAccountStatus.NOT_FOUND || accountStatus === NIMAccountStatus.ERROR) &&
+    modelInfos.length === 0;
```
emilys314: I noticed a flicker between the disabled field and the alert for this. We should show a loader while the nim account stuff is loading

https://github.com/user-attachments/assets/4ae6e7ff-691c-49ff-b6a6-59390b48f46c

mturley:
/hold
Pushed to update microcopy, I'll address the rest of the feedback next.

@vconzola based on [this thread](https://redhat-internal.slack.com/archives/C069KSM8T9N/p1778616546308759), here's updated microcopy:

1-nim-image-field-correct-config

3-nim-no-api-key-admin

4-nim-invalid-api-key-admin

5-nim-no-api-key-non-admin

6-nim-invalid-key-non-admin

2-nim-no-project-selected

packages/nim-serving/src/pages/deploymentWizard/fields/NIMImageField.tsx
mturley: Makes sense, not sure how I missed that antipattern, i've called it out before 😅

packages/nim-serving/src/pages/deploymentWizard/fields/NIMImageField.tsx
mturley: Good call. I did that, and also I fixed a spacing issue: When NIM is selected the ModelLocationInputFields component doesn't render anything (because we only have extension-driven fields for NIM) but it still gets wrapped by a StackItem in ModelLocationSelectField. So I exported a `hasOnlyExtensionFields` helper and prevent rendering the StackItem if it would have no children. Feels a little weird, but it felt weirder to move the StackItem into ModelLocationInputFields or lift the NIM-specific condition out.

Here's a throttled example with both the skeleton and the spacing fix:


https://github.com/user-attachments/assets/252e8d64-2363-42fe-8ff8-e0c67a9f5036

packages/nim-serving/src/pages/deploymentWizard/fields/NIMImageField.tsx
mturley: Done, I lifted out a whole `NIMSettingsLink` component for the conditional "Configure in project settings" / "Ask your project administrator..." part, and had it also reuse some internal projects page things (`useProjectPermissionsTabVisible` instead of a duplicated access check, and `routeProjectsNamespace` instead of a hard-coded `/projects/` URL).

This does mean there is briefly a tiny inline spinner inside the Alert text before it knows whether it can render the link, but you can barely tell.

packages/nim-serving/src/pages/deploymentWizard/fields/NIMImageField.tsx
mturley: This is now covered by the same skeleton that renders if `!accountLoaded || !imagesLoaded`

packages/nim-serving/src/pages/deploymentWizard/fields/NIMImageField.tsx
mturley: hmm... there's still a tiny flicker there, hang on

packages/nim-serving/src/pages/deploymentWizard/fields/NIMImageField.tsx
mturley: Here's a throttled view to show you that little spinner. however I realize looking at this one that I still have a slight flicker going on before the skeleton appears. looking at that.


https://github.com/user-attachments/assets/da3f9a1f-128d-4219-8378-a91dd39476fb

packages/nim-serving/src/pages/deploymentWizard/fields/NIMImageField.tsx
mturley: Ok, addressed the other flicker - for a very brief moment `externalData` was undefined as the extension mounted, so the `!projectName` alert was rendering. Now the skeleton also catches that case.

Here's another throttled recording

https://github.com/user-attachments/assets/f76c4d05-5e07-42c2-9a6a-d41c9284ce19

packages/nim-serving/src/pages/deploymentWizard/fields/NIMImageField.tsx
mturley: Addressed the new flicker (see [above](https://github.com/opendatahub-io/odh-dashboard/pull/7466#discussion_r3244396390)), here's a flicker-free recording of this alert with its internal spinner.

Throttled so you can see it:

https://github.com/user-attachments/assets/739f2cfd-8b53-45d0-8c2b-6004a47296cb

Not throttled (acceptable IMO):


https://github.com/user-attachments/assets/181ea459-96ca-431b-aed3-955f1b88bc19

packages/nim-serving/src/api/models/utils.ts
```diff
@@ -0,0 +1,69 @@
...
+  const accounts = await k8sListResource<NIMAccountKind>({
+    model: NIMAccountModel,
+    queryOptions: { ns: projectNamespace },
+  }).then((result) => result.items);
+
+  if (accounts.length === 0) {
+    return [];
+  }
+
+  const account = accounts[0];
```
emilys314: I don't even now if it's possible to have multiple Accounts in a project, but for the most safety should we prioritize the one the dashboard knows about and created?


```suggestion
  const account = accounts.find((a) => a.name === NIM_ACCOUNT_NAME) ?? accounts[0];
```

emilys314: Oh on related note, we should reuse the fetch nim api logic that is already created in /api/accounts/k8s.ts `getNIMAccount()`. That would solve this problem.

And it keeps the different API's dedicated to just their resource

```suggestion
  const account = getNIMAccount(proejctNameSpace);
```

packages/nim-serving/src/api/models/utils.ts
```diff
@@ -0,0 +1,69 @@
...
+    }
+    return parts.join('.');
+  }
+  return tag;
+};
+
+export const getImageRepository = (modelNamespace: string, modelName: string): string =>
+  `${NIM_IMAGE_REGISTRY}/${modelNamespace}/${modelName}`;
+
+export const fetchNIMModelNames = async (projectNamespace: string): Promise<NIMModelInfo[]> => {
```
emilys314: This function `fetchNIMModelNames()`has k8s calls, I think this function might fit better in an `api.ts` file

packages/nim-serving/src/pages/deploymentWizard/fields/NIMImageField.tsx
```diff
@@ -0,0 +1,197 @@
...
+        onChange({ repository: selected.repository, tag: selected.tag });
+      }
+    },
+    [options, onChange, isEditing],
+  );
+
+  const projectName = externalData?.data.projectName;
+  const imagesLoaded = externalData?.loaded ?? false;
+
+  const { status: accountStatus, loaded: accountLoaded } = useNIMAccountStatus(projectName);
```
emilys314: This is a duplicate api call that is running. The first happens when the externalData hook is loaded, this one happens on render of the component. 

It creates a small screen filcker with the loading state. This is only an issue if you are navigating between steps. That's why it's best (I think) to have no api calls on render and keep them all in the externalData hook. 

I would just combine the 2 things of data you need, 
```
data: { nimImages: NIMImagesData, accountStatus: accountsomething }
```

packages/nim-serving/src/api/models/utils.ts
mturley: Good catch. I do think it's possible to manually create multiple accounts. I think this is also a holdover from when I started this branch before my settings PR was merged, should use my own utils 😅 thanks

packages/nim-serving/src/api/models/utils.ts
mturley: Hm, good call, although I think maybe I should also separate some things out of the `src/api/accounts/utils.ts` file too then. I have a `k8s.ts` file there for the specific k8s calls themselves, but I have higher-level API utils like `createOrReplaceSecret`, `createNIMResources` etc in `utils.ts` there. I'll make a note to move those to `src/api/accounts/api.ts` in a followup.

packages/nim-serving/src/pages/deploymentWizard/fields/NIMImageField.tsx
mturley: Done. While refactoring this I realized I was being silly with naming "images", "models", "modelinfos" in different places when they all really have to do with the same things (the image info objects in the configmap), so I standardized around the `NIMImage` type and renamed the api/models directory and some other things to `images` naming. Sorry for the re-review overhead.

packages/nim-serving/src/api/models/utils.ts
mturley: Done. This is also now called `fetchNIMImages` and `NIMModelInfo` is now `NIMImage` (see [other reply](https://github.com/opendatahub-io/odh-dashboard/pull/7466#discussion_r3262626703))

packages/nim-serving/src/api/models/utils.ts
mturley: Done

packages/nim-serving/src/pages/deploymentWizard/fields/NIMImageField.tsx
mturley: On top of moving the account fetch into the external data hook, I also got rid of the duplicate accounts fetch by making `fetchNIMImages` (previously `fetchNIMModelNames`) take the account as an arg, and `useNIMImages` now throws a NotReadyError during the initial account fetch (staying in loading state)

mturley:
/hold
I broke the "no project" fallback, fixing that

mturley:
/unhold
Needed to have the external data hook fall through with loaded: true if there's no project selected. I also squashed my commits since your last review into one for easier review since some of the same code got moved around. Should be all set now @emilys314

mturley:
(renamed another thing and amended, I'll walk away now)

packages/nim-serving/src/api/images/hooks.ts
```diff
@@ -0,0 +1,47 @@
...
+  const projectName = project?.projectName;
+  const nimAccount = dependencies?.nimAccount;
+
+  const fetchCallback = React.useCallback<FetchStateCallbackPromise<NIMImage[]>>(() => {
+    if (!projectName) {
+      return Promise.reject(new NotReadyError('No project selected'));
+    }
+    if (!nimAccount) {
+      return Promise.reject(new NotReadyError('NIM account not loaded'));
+    }
```
emilys314: I think this gets into an infinite loading state if you don't have a nim account enabled in the project

Image

mturley: oh you've got to be kidding me. I need to do my whole test matrix every time I change anything

mturley: Fixed, and tested all cases again. I babysat claude using playwright to make sure this was accurate 😄 :

## NIM Image Field — Browser Test Results

All tests run on localhost:4051 against the Zaffre ROSA cluster with `?devFeatureFlags=nimWizard=true` and Google account impersonation enabled.

| # | Project | NIM State | User Role | Expected | Result |
|---|---------|-----------|-----------|----------|--------|
| 1 | *(none — global deploy)* | N/A | N/A | Info alert: "No project selected" | **PASS** |
| 2 | `mturley-googleidp-test` | Configured | Owner | NIM image typeahead dropdown | **PASS** |
| 3 | `mturley-no-nim` | No account | Admin | Danger: "No NIM API key" + settings link | **PASS** |
| 4 | `mturley-nim-error` | Invalid key | Admin | Danger: "Invalid NIM API key" + settings link | **PASS** |
| 5 | `mturley-non-admin-nim` | Configured | Non-admin | NIM image typeahead dropdown | **PASS** |
| 6 | `mturley-non-admin-no-nim` | No account | Non-admin | Danger: "No NIM API key" + "Ask your project administrator" | **PASS** |
| 7 | `mturley-non-admin-bad-nim` | Invalid key | Non-admin | Danger: "Invalid NIM API key" + "Ask your project administrator" | **PASS** |

Zero skeletons stuck in loading state across all cases. RBAC gating works correctly: admin users see "Configure in project settings" link, non-admin users see "Ask your project administrator" text.

emilys314 (approved):
Looks good!

Image

PR #7470, opened 2026-05-05
Remove auth annotation when MaaS is selected in Wizard

Griffin-Sullivan:
https://redhat.atlassian.net/browse/RHOAIENG-61103
## Description
When MaaS is enabled we need to disable the auth checkbox and not set the auth annotation to false. In the past we were letting the annotation `security.opendatahub.io/enable-auth: 'false'` get added when maas is selected which is causing a bug where the odh model controller creates auth policies that allow users to get around MaaS governance (MaaSAuthPolicy and MaaSSubscription)

## How Has This Been Tested?
1. Create an llminferenceservice through the wizard
2. De-select token auth
3. Deploy

See annotation `security.opendatahub.io/enable-auth: 'false'`
4. Update through the wizard to select maas checkbox

See annotation removed

Also try these again with a token auth model. I tested with vllm on MaaS flow as well as standard llmd

## Test Impact
Added a jest test to cover the auth scenarios

## Request review criteria:

If you have UI changes: 

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed authentication configuration handling in the MaaS deployment wizard to ensure proper security settings are applied when enabling model serving services
  * Improved gateway reference management during MaaS deployment for better overall configuration handling

### DIFF
--- a/packages/cypress/cypress/tests/mocked/modelsAsAService/maasDeploymentWizard.cy.ts
+++ b/packages/cypress/cypress/tests/mocked/modelsAsAService/maasDeploymentWizard.cy.ts
@@ -219,6 +219,10 @@ describe('MaaS Deployment Wizard', () => {
           namespace: 'openshift-ingress',
         },
       ]);
+
+      expect(interception.request.body.metadata.annotations).to.not.have.property(
+        'security.opendatahub.io/enable-auth',
+      );
     });
 
     cy.wait('@createLLMInferenceService');
--- a/packages/maas/frontend/src/odh/modelServingExtensions/modelDeploymentWizard/maasDeploymentTransformer.ts
+++ b/packages/maas/frontend/src/odh/modelServingExtensions/modelDeploymentWizard/maasDeploymentTransformer.ts
@@ -25,15 +25,15 @@ export const applyMaaSEndpointData = (
   const filteredRefs = existingRefs.filter((ref) => !isMaaSGateway(ref));
 
   if (fieldData.isChecked) {
-    // Add the MaaS gateway
-    const newRefs = [...filteredRefs, MAAS_DEFAULT_GATEWAY];
+    // Add the MaaS gateway and remove the enable-auth annotation — MaaS manages auth externally
     result.model.spec.router = {
       ...result.model.spec.router,
       gateway: {
         ...result.model.spec.router?.gateway,
-        refs: newRefs,
+        refs: [...filteredRefs, MAAS_DEFAULT_GATEWAY],
       },
     };
+    delete result.model.metadata.annotations?.['security.opendatahub.io/enable-auth'];
   } else if (filteredRefs.length > 0) {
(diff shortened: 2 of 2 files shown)


packages/llmd-serving/src/deployments/deploy.ts
```diff
@@ -234,7 +234,9 @@ export const assembleLLMdDeployment = (
         runtimeArgs: wizardData.state.runtimeArgs.data,
         environmentVariables: wizardData.state.environmentVariables.data,
         modelAvailability: wizardData.state.modelAvailability.data,
-        tokenAuthentication: wizardData.state.tokenAuthentication.data,
+        tokenAuthentication: wizardData.state.tokenAuthentication.isDisabled
+          ? undefined
+          : wizardData.state.tokenAuthentication.data,
```
emilys314: This might break project contributor flows when redeploying a model. 

project contributors can't read / write project Roles and RoleBindings. So in this situation we disable the field (don't make modifications)

So with this PR change, if a project contributor edits an existing model with auth, and saves it, it will remove the annotation.

You might want to go one step deeper and simply skip running the `applyTokenAuthentication` if it was disabled

andrewballantyne: Sorry -- why would this break edits?

If the token auth is disabled -- it's not always disabled, it's re-evaled on each iteration of the wizard -- no?

packages/llmd-serving/src/deployments/model.ts
```diff
@@ -271,7 +271,7 @@ export const applyTokenAuthentication = (
 ): LLMInferenceServiceKind => {
   const result = structuredClone(llmdInferenceService);
   const annotations = { ...result.metadata.annotations };
-  if (!tokenAuthentication || tokenAuthentication.length === 0) {
+  if (tokenAuthentication !== undefined && tokenAuthentication.length === 0) {
     annotations['security.opendatahub.io/enable-auth'] = 'false';
```
andrewballantyne, line 275: absolute undefined is almost always a mistake -- can we not evaluate the way this works?

like an enum or static string value -- typescript can help us with this.

`type = 'explicitlyDisabled' || token[]` or something...

packages/llmd-serving/src/deployments/deploy.ts
Griffin-Sullivan: ah dang I see what you're saying @emilys314

andrewballantyne: Do we have tests for this assumption? Is this something we can get in part of this -- seems I misunderstood and I'm worried we now don't have testing for this

Griffin-Sullivan: Tests for that would be more fitting in a tech debt for model serving since it's specific to their rbac

emilys314:
/lgtm

(haven't tested it)

PR #7471, opened 2026-05-05
Swap auth annotation removal to maas package (#7470)

andrewballantyne:
https://redhat.atlassian.net/browse/RHOAIENG-61103


Griffin-Sullivan (commented):
/lgtm

PR #7489, opened 2026-05-07
fix: prevent Tests summary job from being skipped on failures

rsun19:
https://redhat.atlassian.net/browse/RHOAIENG-45140

## Description

GitHub treats a SKIPPED required check as passing branch protection. When an upstream CI job (e.g. Cypress-Mock-Tests) fails, the dependent `Tests` summary job was being skipped rather than explicitly failing. This allowed Tide/GitHub to consider the required `Tests` check as satisfied and merge broken PRs (as seen in PR #5969).

This fix adds `if: always()` to the `Tests` summary job so it runs regardless of dependency outcomes, and explicitly checks `needs.*.result` for `failure` or `cancelled` — reporting a proper FAILURE conclusion that branch protection will block on.

## How Has This Been Tested?

Verified the workflow YAML syntax is valid. The behavior can be confirmed on the next PR where a CI job fails — the `Tests` job should now report FAILURE instead of SKIPPED.

## Test Impact

No application tests are affected. This is a CI workflow-only change. The fix itself is the test improvement — it ensures CI failures are properly surfaced to branch protection.

## Request review criteria:

If you have UI changes:
- N/A — no UI changes

## Summary by CodeRabbit

* **Tests**
  * Improved automated test verification: the pipeline now always performs a post-run check to detect failed, canceled, or skipped jobs, reliably exits on detected issues, and provides clearer success/failure reporting (including the configured Node.js version) for faster troubleshooting.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -376,6 +376,7 @@ jobs:
           verbose: true
 
   Tests:
+    if: always()
     runs-on: ubuntu-latest
     needs:
       - Lint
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -376,6 +376,7 @@ jobs:
           verbose: true
 
   Tests:
+    if: always()
     runs-on: ubuntu-latest
```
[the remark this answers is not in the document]
rsun19: Skipping this one — the `Tests` job only runs a shell `echo`/`exit` step. It never checks out code, calls any GitHub API, or references `GITHUB_TOKEN` in any way. Adding `permissions: {}` would be pure noise on a job whose sole purpose is to aggregate pass/fail status. The token is available but never consumed, so the risk is essentially zero.

If we wanted to enforce least-privilege broadly, a top-level `permissions` block for the entire workflow would be more appropriate than per-job on a no-op summary job.

.github/workflows/test.yml
```diff
@@ -385,5 +386,20 @@ jobs:
...
+            echo "  Unit-Tests: ${{ needs.Unit-Tests.result }}"
+            echo "  Contract-Tests: ${{ needs.Contract-Tests.result }}"
+            echo "  Cypress-Mock-Tests: ${{ needs.Cypress-Mock-Tests.result }}"
+            echo "  Combine-Results-and-Upload: ${{ needs.Combine-Results-and-Upload.result }}"
+            exit 1
+          fi
+          if [[ "${{ contains(needs.*.result, 'cancelled') }}" == "true" ]]; then
+            echo "One or more required jobs were cancelled"
+            exit 1
+          fi
```
christianvogt: I would just simplify this without enumerating the results:
```suggestion
          if [[ "${{ contains(needs.*.result, 'failure') || contains(needs.*.result, 'cancelled') }}" == "true" ]]; then
            echo "One or more required jobs failed or were cancelled."
            exit 1
          fi
```

rsun19: resolved

rsun19:
502 bad gateway error on the ci/prow/images check

christianvogt (approved):

PR #7493, opened 2026-05-07
[RHOAIENG-44140] Fix User management infinite loading on BYOIDC clusters

DaoDaoNoCode:
https://issues.redhat.com/browse/RHOAIENG-44140

## Description

On clusters using an external OIDC provider (e.g. Keycloak / BYOIDC), the `user.openshift.io/groups` CRD does not exist. Groups only live in the user's JWT token at login time and are never stored as Kubernetes resources.

The User Management page (`Settings > User management`) calls `useGroups()` which watches `user.openshift.io/v1/groups` to populate the group dropdown. On BYOIDC clusters this watch returns a raw `K8sStatus` 404 object. `useK8sWatchResourceList` was wrapping any non-`Error` SDK value as a generic `new Error('Unknown error occured')`, discarding the status code. As a result `useGroups()` never resolved to `loaded: true` and the page hung on the loading spinner indefinitely.

**Root cause (two layers):**
1. `useK8sWatchResourceList` silently discarded K8sStatus info from the SDK by wrapping it as a generic Error — any caller that needed to inspect the status code could not.
2. `useGroups()` had no handling for the "Group CRD not found" case.

**Fix:**
- `useK8sWatchResourceList`: wrap raw `K8sStatus` objects from the SDK in the project's `K8sStatusError` (from `errorUtils`) so callers can inspect `error.statusObject.code`.
- `useGroups()`: treat a `K8sStatusError` with `code === 404` as loaded with empty groups. Other errors (5xx, network failures) are still propagated so the page can surface them.

Once unblocked, admins can manually type their Keycloak group names in the creatable multi-select. The typed names need to match what Keycloak sends in the `groups` JWT claim — those are matched by the RHOAI operator's ClusterRoleBindings when evaluating `SelfSubjectAccessReview`.

Screenshot 2026-05-07 at 12 43 54 PM

## How Has This Been Tested?

- Reproduced the bug locally against a BYOIDC cluster (Keycloak): User Management page was stuck loading indefinitely.
- Applied the fix and verified the page loads correctly with an empty group list; existing configured groups (from the `Auth` resource) are displayed and new group names can be typed in.
- Verified standard OpenShift clusters (with Group CRD) are unaffected — groups still populate normally.

## Test Impact

- Updated `frontend/src/api/k8s/__tests__/groups.spec.ts`: added tests for the 404 BYOIDC case (returns empty groups) and non-404 errors (propagated).
- Existing `useK8sWatchResourceList` and `useWatchGroups` specs continue to pass.

## Request review criteria:

## Bug Fixes
- Fixed the app hanging when loading groups on clusters without the Group Custom Resource Definition (CRD), commonly occurring in BYOIDC configurations.
- Enhanced Kubernetes error handling to properly distinguish between missing resources and other service-level errors for improved diagnostics.

### DIFF
--- a/frontend/src/api/k8s/groups.ts
+++ b/frontend/src/api/k8s/groups.ts
@@ -6,6 +6,7 @@ import { groupVersionKind } from '#~/api/k8sUtils';
 import { useAccessReview } from '#~/api/useAccessReview';
 import useK8sWatchResourceList from '#~/utilities/useK8sWatchResourceList';
 import { CustomWatchK8sResult } from '#~/types';
+import { K8sStatusError } from '#~/api/errorUtils';
 
 const accessReviewResource: AccessReviewResourceAttributes = {
   group: 'user.openshift.io',
@@ -30,6 +31,12 @@ export const useGroups = (): CustomWatchK8sResult<GroupKind[]> => {
       return [[], false, undefined];
     }
     if (!allowList) {
+      return [[], true, undefined];
+    }
+    // On BYOIDC clusters (e.g. Keycloak), the Group CRD does not exist and the watch
+    // returns a 404 K8sStatus. Treat this as loaded with no groups so the page does not hang.
+    // Other errors (5xx, network failures) are propagated so the page can surface them.
+    if (error instanceof K8sStatusError && error.statusObject.code === 404) {
       return [[], true, undefined];
     }
     return [groupData, loaded, error];
--- a/frontend/src/api/k8s/__tests__/groups.spec.ts
+++ b/frontend/src/api/k8s/__tests__/groups.spec.ts
@@ -2,6 +2,8 @@ import { testHook } from '@odh-dashboard/jest-config/hooks';
 import { groupVersionKind, useAccessReview, useGroups } from '#~/api';
 import { GroupModel } from '#~/api/models';
 import { mockGroup } from '#~/__mocks__/mockGroup';
+import { mock404Error, mock500Error } from '#~/__mocks__/mockK8sStatus';
+import { K8sStatusError } from '#~/api/errorUtils';
 import useK8sWatchResourceList from '#~/utilities/useK8sWatchResourceList';
 import { GroupKind } from '#~/k8sTypes';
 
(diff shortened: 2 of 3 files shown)


DaoDaoNoCode:
Verified the group assignment works to Keycloak groups too.

I set `rhods-users` (this is a normal user group in Keycloak) as the RHOAI admin group, then I login as a normal user, I can see the cluster settings, which means it works.

Screenshot 2026-05-07 at 12 57 07 PM

frontend/src/api/k8s/groups.ts
```diff
@@ -32,6 +33,12 @@ export const useGroups = (): CustomWatchK8sResult<GroupKind[]> => {
     if (!allowList) {
       return [[], true, undefined];
     }
+    // On BYOIDC clusters (e.g. Keycloak), the Group CRD does not exist and the watch
+    // returns a 404 K8sStatus. Treat this as loaded with no groups so the page does not hang.
```
bobbravo2, line 37: Is it possible for a user to hit this condition on a generic k8s 404? Are we able to determine that the request is 404 on this one keycloak end point?

Not sure how plausible that is, 
 having a guard only on a 404 from a generic k8s's end point _may_ be risky.

DaoDaoNoCode: It's rare, when fetching the groups, if there are no groups, they will get an empty array with 200 OK but not a 404 error, the 404 only happens when there are no OpenShift groups CRD at all, which is our case.

bobbravo2: Thanks @DaoDaoNoCode for the eli8, lgtm

nananosirova (approved):
Tested this PR manually on a BYOIDC cluster (Keycloak as the external OIDC provider).

Reproduced the bug before the fix:
1. Confirmed the cluster has no Group CRD: `oc get groups.user.openshift.io` returns a 404 error.
2. Navigated to `Settings > User management`.
3. The page spinner was loading indefinitely. In the browser dev tools, the network/console showed a `K8sStatus` 404 response from the `user.openshift.io/v1/groups` watch:

Image

After applying the fix:
1. Ran the PR locally against the same BYOIDC cluster.
2. Navigated to `Settings > User management` again.
3. The page now loads correctly. The existing configured groups from the Auth CR (`rhods-admins`, `system:authenticated`) are displayed as expected.
4. The 404 `K8sStatus` response is still visible in dev tools (expected, since the CRD genuinely does not exist), but the frontend now handles it gracefully instead of hanging:

Image

Code looks clean. 

/lgtm

Gkrumbach07 (approved):

PR #7547, opened 2026-05-13
fix(kserve): apply connection-path annotation on dry-run deploys

mturley:
Bug found when trying to fix https://issues.redhat.com/browse/RHOAIENG-57823, although doesn't fix the issue

## Description

The `opendatahub.io/connection-path` annotation was not being set on the InferenceService during dry-run deploys with existing S3 connections.

In `applyConnectionData` (`packages/kserve/src/deployUtils.ts`), the connection-path annotation logic was nested inside a `if (secretName || createConnectionData.nameDesc?.name)` gate. For existing connections during dry runs, the deploy flow intentionally passes `secretName = undefined` (to avoid K8s validation errors on the connection name annotation), which also prevented the connection-path annotation from being set.

This fix moves the connection-path (and the corresponding delete for non-S3 connections) outside the secretName gate, matching how `packages/llmd-serving/src/deployments/model.ts` handles these as independent concerns.

**Investigation note:** The intermittent production bug described in the Jira (connection-path missing on the actual deployed InferenceService) could not be reproduced in mock tests — the actual (non-dry-run) deploy correctly sets the annotation regardless of this fix. The timing-dependent race condition described in the Jira may require real cluster network latency to manifest. A detailed investigation report is in the Jira comments.

## How Has This Been Tested?

- Cypress mock test added that deploys with an existing S3 connection selected from a dropdown, types a folder path, and asserts the `connection-path` annotation on both the dry-run and actual InferenceService creation requests
- Test fails without the fix (dry-run annotation missing), passes with the fix
- All 28 existing `modelServingDeploy.cy.ts` tests continue to pass
- Type-check passes for `@odh-dashboard/kserve` and `@odh-dashboard/model-serving`

## Test Impact

- Added 1 new Cypress mock test covering the S3 connection-path annotation flow
- Exploratory tests (new S3 connection, speed-run variant) were also written during investigation and confirmed the actual deploy path is not affected; these were removed to keep the PR focused

## Request review criteria:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved stability of model deployment with S3 connections by fixing potential errors in annotation handling.

* **Tests**
  * Added test coverage for deploying models with existing S3 connections to verify proper annotation configuration.

### DIFF
--- a/packages/cypress/cypress/tests/mocked/modelServing/modelServingDeploy.cy.ts
+++ b/packages/cypress/cypress/tests/mocked/modelServing/modelServingDeploy.cy.ts
@@ -1790,6 +1790,249 @@ describe('Model Serving Deploy Wizard', () => {
     modelServingWizardEdit.findModelSourceStep().should('be.enabled');
   });
 
+  describe('S3 connection-path annotation (RHOAIENG-57823)', () => {
+    it('should include connection-path when deploying with existing S3 connection', () => {
+      initIntercepts({ modelType: ServingRuntimeModelType.PREDICTIVE });
+
+      // Two S3 connections so user must actively select from dropdown
+      cy.interceptK8sList(
+        { model: SecretModel, ns: 'test-project' },
+        mockK8sResourceList([
+          mockSecretK8sResource({
+            name: 'my-s3-connection-1',
+            namespace: 'test-project',
+            displayName: 'My S3 Connection 1',
+            connectionType: 's3',
+          }),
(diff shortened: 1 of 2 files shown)


mturley:
Closing this because it is stale and not worth the review overhead right now. Including an annotation in a dry run will not catch any issues that wouldn't already be caught, as annotations are open-ended and we have permission to write them if we have permission to write to the resource. Low value change that doesn't fix any real behavior. We can revisit if necessary.

PR #7575, opened 2026-05-14
Extension serving runtime templates

emilys314:
Closes https://redhat.atlassian.net/browse/RHOAIENG-60278

## Description

2 features added:
- Adds the "NVIDIA NIM" model type if NIM is selected
- Move the kserve serving runtime templates field to an extension in the kserve package. It relies on a generative or predictive model type being selected. So if NVIDIA NIM is selected, there is no field for this being activated (which we want)

Technical stuff:
- The main thing is I added a `field.formId` property, not sure how I feel about it. We can't have 2 fields with the same id, this messes up the init and cleanup. So to allow the unique id's, this `formId` will define where in the form's state to store the info. This allows multiple fields to be saved in the same state location, which makes subsequent fields easier to reference
  - The fix for this scenario: Generative is selected > loads the llmd/modelServer > change to predictive > the llmd/modelServer unloads and the kserve/modelServer loads in. If they had the same id, the form would see a field with the same id is still persisting, so it wouldn't re-init the data. It would leave stale info
- Misc updates to help, like when to reset the field (shouldResetOnDependenciesChange)
- I also hardened the vLLMOnMaaS feature flag and made it so we pass this in for some isActive logic for the fields

Followup required:
For the NIM image selection field, we need to override this field with the NIM option, forcing this value. Also disable it so a user can't change it

Screenshot 2026-05-15 at 3 34 03 PM
Screenshot 2026-05-15 at 3 34 09 PM

## How Has This Been Tested?

Main NIM flow:
1. Enable `nimWizard` flag
2. Create a new deployment
3. Select NIM, go to select the model type, the NIM option should appear only for this
4. Go to the next step, see that the "Deployment resource" doesn't appear

Other flows:
1. Check the different variations of generative / legacy and with or without the vLLMonMaaS flag enabled
2. Make sure it all still works

## Test Impact

All cypress tests are still passing without any changes so it's good

We need to setup a NIM mock test soon

## Request review criteria:

If you have UI changes: 

## Summary by CodeRabbit

* **New Features**
  * Added a deployment resource (serving runtime) selection field to the model deployment wizard
  * Added NVIDIA NIM as a new model-type option

* **Improvements**
  * Model type selection accepts extension-provided extra options and external suggestions
  * Wizard state handling is more resilient to missing/optional fields
  * vLLM deployment on MaaS is now gated by a feature flag

### DIFF
--- a/packages/kserve/src/deployUtils.ts
+++ b/packages/kserve/src/deployUtils.ts
@@ -388,7 +388,10 @@ export const applyModelType = (
   modelType: string,
 ): InferenceServiceKind => {
   if (!isKnownServingRuntimeModelType(modelType)) {
-    console.error(`Invalid model type for kserve deployment: ${modelType}`);
+    console.error(
+      `Invalid model type for kserve deployment: ${modelType}. Skipping applyModelType.`,
+    );
+    return inferenceService;
   }
 
   const result = structuredClone(inferenceService);
--- a/packages/kserve/src/deployUtils.ts
+++ b/packages/kserve/src/deployUtils.ts
@@ -40,10 +40,7 @@ import {
 } from '@odh-dashboard/internal/concepts/connectionTypes/utils';
 import { ModelLocationData } from '@odh-dashboard/model-serving/types/form-data';
 import { ServingRuntimeModelType } from '@odh-dashboard/internal/types';
-import {
-  isValidModelType,
-  type ModelTypeFieldData,
-} from '@odh-dashboard/model-serving/components/deploymentWizard/fields/ModelTypeSelectField';
+import { type ModelTypeFieldData } from '@odh-dashboard/model-serving/components/deploymentWizard/fields/ModelTypeSelectField';
 import type { ModelAvailabilityFieldsData } from '@odh-dashboard/model-serving/components/deploymentWizard/fields/ModelAvailabilityFields';
 import type { RuntimeArgsFieldData } from '@odh-dashboard/model-serving/components/deploymentWizard/fields/RuntimeArgsField';
 import type { EnvironmentVariablesFieldData } from '@odh-dashboard/model-serving/components/deploymentWizard/fields/EnvironmentVariablesField';
@@ -372,19 +369,28 @@ export const extractModelType = (deployment: {
   model: InferenceServiceKind;
 }): ModelTypeFieldData | null => {
   const modelType = deployment.model.metadata.annotations?.['opendatahub.io/model-type'];
-  if (modelType && isValidModelType(modelType)) {
-    return {
-      type: modelType,
-      legacyVLLM: modelType === ServingRuntimeModelType.GENERATIVE,
-    };
+  if (!modelType) {
+    return null;
   }
-  return null;
+
+  return {
+    type: modelType,
+    legacyVLLM: modelType === ServingRuntimeModelType.GENERATIVE,
+  };
+};
+
+const isKnownServingRuntimeModelType = (type?: string): type is ServingRuntimeModelType => {
+  return type === ServingRuntimeModelType.PREDICTIVE || type === ServingRuntimeModelType.GENERATIVE;
 };
 
 export const applyModelType = (
   inferenceService: InferenceServiceKind,
-  modelType: ServingRuntimeModelType,
+  modelType: string,
 ): InferenceServiceKind => {
+  if (!isKnownServingRuntimeModelType(modelType)) {
+    console.error(`Invalid model type for kserve deployment: ${modelType}`);
+  }
+
   const result = structuredClone(inferenceService);
   result.metadata.annotations = {
     ...result.metadata.annotations,
--- a/packages/kserve/src/wizardFields/servingRuntime/KServeServingRuntimeField.tsx
+++ b/packages/kserve/src/wizardFields/servingRuntime/KServeServingRuntimeField.tsx
@@ -85,7 +85,7 @@ export const isKServeServingRuntimeFieldActive = (
   }
 
   // Generative models without LLMInferenceServiceConfig flow: always show (no alternative)
-  if (vLLMDeploymentOnMaaSEnabled === false) {
+  if (!vLLMDeploymentOnMaaSEnabled) {
     if (modelType?.type === ServingRuntimeModelType.GENERATIVE) {
       return true;
     }
@@ -251,7 +251,7 @@ export const KServeServingRuntimeFieldWizardField: KServeServingRuntimeFieldType
   id: 'kserve/modelServer',
   step: 'modelDeployment',
   type: 'replacement',
-  formId: 'modelServer',
+  stateKey: 'modelServer',
   isActive: isKServeServingRuntimeFieldActive,
   reducerFunctions: {
     resolveDependencies: (formData) => ({
--- a/packages/kserve/src/wizardFields/servingRuntime/KServeServingRuntimeField.tsx
+++ b/packages/kserve/src/wizardFields/servingRuntime/KServeServingRuntimeField.tsx
@@ -1,0 +1,267 @@
+import React from 'react';
+import { z } from 'zod';
+import type { WizardField, WizardFormData } from '@odh-dashboard/model-serving/types/form-data';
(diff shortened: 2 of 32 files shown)


packages/kserve/src/wizardFields/servingRuntime/KServeServingRuntimeField.tsx
```diff
@@ -0,0 +1,316 @@
...
+        (template) =>
+          ({
+            name: template.metadata.name,
+            namespace: template.metadata.namespace,
+            label: getServingRuntimeDisplayNameFromTemplate(template),
+            version: getServingRuntimeVersion(template),
+            compatibleWithHardwareProfile: profileIdentifiers.some((identifier) =>
+              isCompatibleWithIdentifier(identifier, getServingRuntimeFromTemplate(template)),
+            ),
+            scope: template.metadata.namespace === dashboardNamespace ? 'global' : 'project',
```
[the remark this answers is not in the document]
emilys314: No, the dashboard namespace is the global namespace

packages/model-serving/src/components/deploymentWizard/useDeploymentWizardReducer.tsx
```diff
@@ -161,36 +168,33 @@ export const useDeploymentWizardReducer = (
       const isDependenciesChanged = !isEqual(dependencies, prevDependencies);
 
       if (isNew || isDependenciesChanged) {
-        if (isDependenciesChanged && field.shouldResetOnDependencyChange) {
-          dispatch({ type: 'clearFieldData', payload: { id: field.id } });
+        if (
+          isDependenciesChanged &&
+          field.shouldResetOnDependencyChange &&
+          field.shouldResetOnDependencyChange(prevDependencies, dependencies)
+        ) {
```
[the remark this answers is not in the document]
emilys314: the typescript strictness in our project ensures there are no consumers using the wrong type. I've updated all consumers

mturley:
> The main thing is I added a field.formId property, not sure how I feel about it. We can't have 2 fields with the same id, this messes up the init and cleanup. So to allow the unique id's, this formId will define where in the form's state to store the info. This allows multiple fields to be saved in the same state location, which makes subsequent fields easier to reference

> The fix for this scenario: Generative is selected > loads the llmd/modelServer > change to predictive > the llmd/modelServer unloads and the kserve/modelServer loads in. If they had the same id, the form would see a field with the same id is still persisting, so it wouldn't re-init the data. It would leave stale info

Interesting -- I need to dig deeper here but just from your comment, I wonder if it's strange to have two different fields filling the same state property. they do serve the same role but they're different fields, right? I guess it would be a mess to keep them separate and have to worry about resetting the other and wiring them up to the same place later. This solution sounds fine to me but I'll dig deeper tomorrow.

emilys314:
> > The main thing is I added a field.formId property, not sure how I feel about it. We can't have 2 fields with the same id, this messes up the init and cleanup. So to allow the unique id's, this formId will define where in the form's state to store the info. This allows multiple fields to be saved in the same state location, which makes subsequent fields easier to reference
> 
> > The fix for this scenario: Generative is selected > loads the llmd/modelServer > change to predictive > the llmd/modelServer unloads and the kserve/modelServer loads in. If they had the same id, the form would see a field with the same id is still persisting, so it wouldn't re-init the data. It would leave stale info
> 
> Interesting -- I need to dig deeper here but just from your comment, I wonder if it's strange to have two different fields filling the same state property. they do serve the same role but they're different fields, right? I guess it would be a mess to keep them separate and have to worry about resetting the other and wiring them up to the same place later. This solution sounds fine to me but I'll dig deeper tomorrow.

Another solution I thought of was using the platform of the extensions as a way to may it more unique

Also this is mainly an issue because subsequent fields are looking at the formState.modelServing to do logic. But most of this should be platform specific. So needing the same fieldId is not as necessary. But that involves updating / reworking the existing fields

packages/kserve/src/deployUtils.ts
```diff
@@ -372,19 +369,28 @@ export const extractModelType = (deployment: {
...
+  return type === ServingRuntimeModelType.PREDICTIVE || type === ServingRuntimeModelType.GENERATIVE;
 };
 
 export const applyModelType = (
   inferenceService: InferenceServiceKind,
-  modelType: ServingRuntimeModelType,
+  modelType: string,
 ): InferenceServiceKind => {
+  if (!isKnownServingRuntimeModelType(modelType)) {
+    console.error(`Invalid model type for kserve deployment: ${modelType}`);
```
YuliaKrimerman: We log an error for invalid model types but it still writes them to the K8s annotation. This should either throw or skip the write.

emilys314: Makes sense, I changed it to skip. Throwing an error would be totally block the deployment, which we don't want. Especially because this annotation is dashboard only metadata

packages/kserve/src/wizardFields/servingRuntime/KServeServingRuntimeField.tsx
```diff
@@ -0,0 +1,326 @@
...
+  }
+
+  // Generative models without LLMInferenceServiceConfig flow: always show (no alternative)
+  if (vLLMDeploymentOnMaaSEnabled === false) {
+    if (modelType?.type === ServingRuntimeModelType.GENERATIVE) {
+      return true;
+    }
+  }
+
+  return false;
```
mturley: Nit: if `vLLMDeploymentOnMaaSEnabled` is undefined, this would return false when it should maybe still return true for generative models. Not sure it'll ever be possible to actually hit that, but is there a reason for the `=== true` and `=== false` over just `if (...)` and `if (!...)`?

packages/model-serving/src/components/deploymentWizard/dynamicFormUtils.ts
```diff
@@ -7,6 +7,9 @@ import {
   isWizardField2Extension,
 } from '../../../extension-points';
 
+/**
+ * @deprecated Use useWizardFieldOverrides instead
+ */
 export const useWizardFieldFromExtension = <T extends DeploymentWizardField>(
```
mturley, line 13: Out of scope, but I wonder if we should add a lint rule to prevent new usages of this hook (unless we plan to get rid of it soon)

mturley (commented):
Tested with `vLLMDeploymentOnMaaS=true`:
- NIM location: "NVIDIA NIM" model type option appears
- NIM location + NIM type: step 2 has no "Deployment resource" field
- S3/URI location: no NIM model type option
- S3/URI location + Generative (non-legacy): "Deployment resource" shown on step 2 (with LLMd config options)
- S3/URI location + Generative + legacy toggle: "Deployment resource" shown on step 2 (with generative ServingRuntime options)
- S3/URI location + Predictive: "Model framework" + "Deployment resource" on step 2 (with predictive ServingRuntime options)

With `vLLMDeploymentOnMaaS=false`:
- S3/URI location + Generative: No legacy toggle, "Deployment resource" shown on step 2 (with generative ServingRuntime options)
- S3/URI location + Predictive: "Model framework" + "Deployment resource" on step 2 (with predictive ServingRuntime options)

Code LGTM with a minor nit.

Also: thinking about it more, I wonder if `formId` could be a confusing name since it's the key used in state, not the id used in the form elements. Maybe a name like `stateId` or `stateKey` would work better? Not at all a blocker.

/lgtm

packages/model-serving/src/components/deploymentWizard/dynamicFormUtils.ts
emilys314: Once this PR merges I'll rename them

packages/kserve/src/wizardFields/servingRuntime/KServeServingRuntimeField.tsx
emilys314: Good point, I don't recall why I did that yeah

emilys314:
> Also: thinking about it more, I wonder if formId could be a confusing name since it's the key used in state, not the id used in the form elements. Maybe a name like stateId or stateKey would work better? Not at all a blocker.

That's a much better name, I'll rename it

mturley:
/lgtm
Nice! waiting for cypress to pass before approving

PR #7578, opened 2026-05-15
Refactor model transfer jobs handler to use downstream repository layer

ppadti:
Closes: https://redhat.atlassian.net/browse/RHOAIENG-57001

## Description
This PR aims to refactor model transfer jobs handler to use downstream repository layer

## How Has This Been Tested?
Make sure the model transfer job creation behaviour is same after the refactor for both admin and non-admin users

## Test Impact
Added the test for repository layer

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/packages/model-registry/upstream/bff/internal/redhat/repositories/model_transfer_jobs_repository.go
+++ b/packages/model-registry/upstream/bff/internal/redhat/repositories/model_transfer_jobs_repository.go
@@ -1,0 +1,214 @@
+package repositories
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"log/slog"
+	"net/http"
+	"strings"
+	"sync"
+
+	k8s "github.com/kubeflow/hub/ui/bff/internal/integrations/kubernetes"
+	batchv1 "k8s.io/api/batch/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+)
+
+const (
+	maxConcurrentNamespaceFetches = 10
+)
+
+// ErrProjectsAPIUnavailable signals that the OpenShift Projects API could not
+// be reached. Callers should fall back to the default (cluster-wide) handler.
+var ErrProjectsAPIUnavailable = errors.New("OpenShift Projects API unavailable")
+
+// ProjectScopedJobsRepository encapsulates the OpenShift-specific logic for
+// listing model transfer jobs across the namespaces accessible to a user.
+type ProjectScopedJobsRepository struct {
+	logger *slog.Logger
+}
+
+func NewProjectScopedJobsRepository(logger *slog.Logger) *ProjectScopedJobsRepository {
+	return &ProjectScopedJobsRepository{logger: logger}
+}
+
+// CreateScopedClient builds a KubernetesClientInterface wrapper whose
+// GetAllModelTransferJobs method fetches jobs only from the namespaces the
+// user can access (discovered via the OpenShift Projects API) instead of
+// requiring cluster-wide RBAC.
+//
+// The returned client embeds the original client, overriding only the
+// job-listing method so that upstream enrichment logic (pod status, events)
+// continues to work unchanged.
+func (r *ProjectScopedJobsRepository) CreateScopedClient(
+	ctx context.Context,
+	client k8s.KubernetesClientInterface,
+	bearerToken string,
+) (k8s.KubernetesClientInterface, error) {
+	namespaces, err := r.listOpenShiftProjects(ctx, client, bearerToken)
+	if err != nil {
+		return nil, fmt.Errorf("%w: %w", ErrProjectsAPIUnavailable, err)
+	}
+
+	cfg, err := restConfigForClient(client)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get REST config: %w", err)
+	}
+
+	userCfg := rest.AnonymousClientConfig(cfg)
+	userCfg.BearerToken = bearerToken
+	userCfg.BearerTokenFile = ""
+
+	clientset, err := kubernetes.NewForConfig(userCfg)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create kubernetes client: %w", err)
+	}
+
+	return &projectScopedClient{
+		KubernetesClientInterface: client,
+		userClientset:             clientset,
+		namespaces:                namespaces,
+	}, nil
+}
+
+// projectScopedClient wraps a KubernetesClientInterface and overrides
+// GetAllModelTransferJobs to list jobs from specific namespaces instead of
+// using NamespaceAll.
+type projectScopedClient struct {
+	k8s.KubernetesClientInterface
+	userClientset kubernetes.Interface
+	namespaces    []string
+}
+
+// GetAllModelTransferJobs overrides the interface method to list jobs from the
+// user's accessible namespaces rather than requiring cluster-wide permission.
+func (c *projectScopedClient) GetAllModelTransferJobs(ctx context.Context, _ string, modelRegistryID string, _ string) (*batchv1.JobList, error) {
+	if modelRegistryID == "" {
+		return &batchv1.JobList{}, nil
+	}
+
+	labelSelector := "modelregistry.kubeflow.org/job-type=async-upload,modelregistry.kubeflow.org/model-registry-name=" + modelRegistryID
+
+	var (
+		allJobs []batchv1.Job
+		mu      sync.Mutex
+		wg      sync.WaitGroup
+	)
+
+	sem := make(chan struct{}, maxConcurrentNamespaceFetches)
+
+	for _, ns := range c.namespaces {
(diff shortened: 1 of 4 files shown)


manaswinidas (commented):
Sentinel wrapping is too broad: config/transport/decode failures inside `listOpenShiftProjects` all get tagged as `ErrProjectsAPIUnavailable`, causing the handler to silently fall back to cluster-wide listing instead of returning a 500. Move the sentinel into `listOpenShiftProjects` around the actual HTTP errors only.

packages/model-registry/upstream/bff/internal/redhat/repositories/model_transfer_jobs_repository.go
```diff
@@ -0,0 +1,214 @@
...
+func (r *ProjectScopedJobsRepository) CreateScopedClient(
+	ctx context.Context,
+	client k8s.KubernetesClientInterface,
+	bearerToken string,
+) (k8s.KubernetesClientInterface, error) {
+	namespaces, err := r.listOpenShiftProjects(ctx, client, bearerToken)
+	if err != nil {
+		return nil, fmt.Errorf("%w: %w", ErrProjectsAPIUnavailable, err)
+	}
+
```
manaswinidas, line 55: Remove the blanket sentinel wrap here — let `listOpenShiftProjects` decide which errors are "API unavailable".

```suggestion
		return nil, err
```

packages/model-registry/upstream/bff/internal/redhat/repositories/model_transfer_jobs_repository.go
```diff
@@ -0,0 +1,214 @@
...
+	resp, err := httpClient.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("failed to list projects: %w", err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("projects API returned status %d", resp.StatusCode)
+	}
+
```
manaswinidas: Wrap only the actual HTTP failures with `ErrProjectsAPIUnavailable` so config/transport/decode errors propagate as real 500s.

```suggestion
	resp, err := httpClient.Do(req)
	if err != nil {
		return nil, fmt.Errorf("%w: %w", ErrProjectsAPIUnavailable, err)
	}
	defer resp.Body.Close()

	if resp.StatusCode != http.StatusOK {
		return nil, fmt.Errorf("%w: status %d", ErrProjectsAPIUnavailable, resp.StatusCode)
	}
```

packages/model-registry/upstream/bff/internal/redhat/repositories/model_transfer_jobs_repository.go
```diff
@@ -0,0 +1,214 @@
...
+	ctx context.Context,
+	client k8s.KubernetesClientInterface,
+	bearerToken string,
+) (k8s.KubernetesClientInterface, error) {
+	namespaces, err := r.listOpenShiftProjects(ctx, client, bearerToken)
+	if err != nil {
+		return nil, err
+	}
+
+	cfg, err := restConfigForClient(client)
```
manaswinidas: Nit: `restConfigForClient(client)` is already called inside `listOpenShiftProjects` (line 146). This is the second call per request for the same client, each creating a copy of the REST config. Consider extracting the config once here and passing it down to `listOpenShiftProjects` to avoid the redundant work.

packages/model-registry/upstream/bff/internal/redhat/repositories/model_transfer_jobs_repository.go
```diff
@@ -0,0 +1,214 @@
...
+			namespaces = append(namespaces, p.Metadata.Name)
+		}
+	}
+	return FilterSystemNamespaces(namespaces), nil
+}
+
+// FilterSystemNamespaces removes system namespaces that will never contain
+// model transfer jobs. This matches the filtering in
+// packages/gen-ai/bff/internal/repositories/namespace.go.
+func FilterSystemNamespaces(namespaces []string) []string {
```
manaswinidas, line 190: This is now the third copy of the same system-namespace filtering logic (also in `packages/gen-ai/bff/internal/repositories/namespace.go` and `packages/autorag/bff/internal/repositories/namespace.go`). Making it exported (`FilterSystemNamespaces`) signals reuse intent — worth extracting to a shared utility package so the three copies don't drift.

packages/model-registry/upstream/bff/internal/redhat/repositories/model_transfer_jobs_repository_test.go
```diff
@@ -0,0 +1,193 @@
...
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	batchv1client "k8s.io/client-go/kubernetes/typed/batch/v1"
+)
+
+// --- Helpers ---
+
+func makeJob(name, namespace string) batchv1.Job {
```
manaswinidas, line 24: `makeJob` is identical to the one in `handlers/model_transfer_jobs_test.go:114`. Since they're in different packages, consider a shared `testutil` package under `internal/redhat/` or at minimum cross-reference them so they stay in sync.

packages/model-registry/upstream/bff/internal/redhat/repositories/model_transfer_jobs_repository_test.go
```diff
@@ -0,0 +1,193 @@
...
+		}
+	}
+}
+
+func TestFilterSystemNamespacesEmpty(t *testing.T) {
+	result := FilterSystemNamespaces([]string{})
+	if len(result) != 0 {
+		t.Fatalf("expected 0 namespaces, got %d", len(result))
+	}
+}
```
manaswinidas, line 200: Missing coverage: no tests for `listOpenShiftProjects` error paths — specifically:
- transport creation failure → should NOT carry `ErrProjectsAPIUnavailable`
- non-200 status → should carry `ErrProjectsAPIUnavailable`
- JSON decode failure → should NOT carry `ErrProjectsAPIUnavailable`
- `httpClient.Do` failure → should carry `ErrProjectsAPIUnavailable`

These are the most safety-critical paths the refactor introduced — the sentinel vs. non-sentinel distinction determines fallback vs. 500.

packages/model-registry/upstream/bff/internal/redhat/repositories/model_transfer_jobs_repository_test.go
```diff
@@ -0,0 +1,193 @@
...
+			continue
+		}
+		if labels[kv[0]] != kv[1] {
+			return false
+		}
+	}
+	return true
+}
+
+// --- projectScopedClient tests ---
```
manaswinidas, line 106: The tests cover `projectScopedClient` and `FilterSystemNamespaces` but not `CreateScopedClient` itself. Its orchestration (list projects → build config → create clientset → assemble scoped client) is untested at the repository level — handler tests mock it entirely via the interface.

packages/model-registry/upstream/bff/internal/redhat/repositories/model_transfer_jobs_repository.go
ppadti: yeah but they're in separate Go modules with different module paths, extracting it would needs to create a separate go modules across packages and that is out of scope fro this issue

packages/model-registry/upstream/bff/internal/redhat/repositories/model_transfer_jobs_repository_test.go
ppadti: yeah, I think for handler test we don't need makeJob and simple job format would work there. Will remove it from handler test.

Philip-Carneiro:
@ppadti need to update the branch

packages/model-registry/upstream/bff/internal/redhat/repositories/model_transfer_jobs_repository.go
```diff
@@ -0,0 +1,209 @@
...
+
+// ErrProjectsAPIUnavailable signals that the OpenShift Projects API could not
+// be reached. Callers should fall back to the default (cluster-wide) handler.
+var ErrProjectsAPIUnavailable = errors.New("OpenShift Projects API unavailable")
+
+// ProjectScopedJobsRepository encapsulates the OpenShift-specific logic for
+// listing model transfer jobs across the namespaces accessible to a user.
+type ProjectScopedJobsRepository struct {
+	logger *slog.Logger
+}
```
manaswinidas, line 32: Nit: `logger` is stored but never read — no `r.logger` call exists in any method. Either add a useful log line (e.g. project count after listing) or drop the field to avoid dead code. Same pattern exists in `ModelRegistrySettingsRepository`, so not blocking.

manaswinidas: merging this for now coz this is an existing pattern - this can be a follow up

manaswinidas (approved):
/lgtm
/approve

PR #7583, opened 2026-05-15
fix: decouple Model Catalog and MCP Catalog from Model Registry reliantArea

manaswinidas:
https://issues.redhat.com/browse/RHOAIENG-62450

## Description

The `MODEL_CATALOG` and `MCP_CATALOG` areas in `SupportedAreasStateMap` previously declared `MODEL_REGISTRY` as a `reliantArea`. This meant disabling Model Registry (via `disableModelRegistry` flag or missing DSC component) **automatically disabled Model Catalog and MCP Catalog**, even when their own flags were set to enabled.

This coupling was a legacy artifact. This PR removes `reliantAreas: [SupportedArea.MODEL_REGISTRY]` from both `MODEL_CATALOG` and `MCP_CATALOG`, so each catalog area is controlled solely by its own feature flag (`disableModelCatalog` / `mcpCatalog`).

https://github.com/user-attachments/assets/5af361fe-c192-4378-baba-846d81f5b14e

### What changed

- **`frontend/src/concepts/areas/const.ts`** — removed `reliantAreas` from `MODEL_CATALOG` and `MCP_CATALOG`

### Why this is safe

- Shell-level Model Catalog code (`frontend/src/concepts/modelCatalog/`) does **not** import from Model Registry — the dependency only existed at the area config level
- The plugin tab extensions already use independent `flags.required` (`MODEL_CATALOG` for Catalog tab, `MODEL_REGISTRY` for Registry tab)
- `FINE_TUNING` independently lists both `MODEL_CATALOG` and `MODEL_REGISTRY` in its own `reliantAreas`, so that continues to work correctly

## How Has This Been Tested?

- Unit tests for `isAreaAvailable` covering the previously-broken scenario (MR off + MC on)
- Cypress mocked tests for tab route navigation, nav sidebar, and MCP catalog feature flag with registry disabled
- Type-check and lint pass cleanly

## Test Impact

Added tests covering the decoupled flag scenarios:

**Unit tests** (`frontend/src/concepts/areas/__tests__/utils.spec.ts`):
- Model Catalog available when Model Registry is disabled
- MCP Catalog available when Model Registry is disabled

**Cypress mocked tests**:
- Tab route page shows catalog + deployments tabs (no registry tab) when registry is disabled (`tabRouteNavigation.cy.ts`)
- Tab route redirects to catalog tab from base URL when registry is disabled
- Nav sidebar shows Models section when catalog is enabled but registry is disabled (`navSidebar.cy.ts`)
- MCP Catalog nav item shows when enabled and Model Registry is disabled (`mcpCatalogFeatureFlag.cy.ts`)

## Request review criteria:

## Summary by CodeRabbit

* **Tests**
  * Added unit and end-to-end tests covering catalog-only scenarios: area availability, sidebar and tab navigation, and redirects when the registry is disabled.

* **Updates**
  * MCP catalog no longer requires the model registry area, allowing the catalog and related UI to function independently when the registry is removed.


frontend/src/concepts/areas/const.ts
```diff
@@ -170,11 +170,9 @@ export const SupportedAreasStateMap: SupportedAreasState = {
   },
   [SupportedArea.MODEL_CATALOG]: {
     featureFlags: ['disableModelCatalog'],
-    reliantAreas: [SupportedArea.MODEL_REGISTRY],
```
ppadti, line 173: Now that we removed the reliantAreas from both MCP and model catalog, we need to have  `requiredComponents: [DataScienceStackComponent.MODEL_REGISTRY],` no? because we need the get the `dscStatus?.components?.modelregistry?.registriesNamespace` to pass in all the endpoint and to look for service in this namespace

ppadti (changes requested):

frontend/src/concepts/areas/const.ts
manaswinidas: Oh yes, added it just now

ppadti (commented):
One more thing we need to follow up with UX - not related to this change
But when we any one of the feature flag is turned on and others are turned off -  we will loose the title of the section

Image
Image
Image

 Titles will only shows when we have at least 2 feature flags turned on
Image

ppadti (approved):
Tested this PR locally, changes works fine.
/lgtm

manaswinidas (commented):
/approve 

Self-approve with team lgtm

PR #7602, opened 2026-05-18
feat: Add MLflow not-configured empty state on Experiments page

Taj010:
## Description

For: https://issues.redhat.com/browse/RHOAIENG-60240

When MLflow is not configured in the cluster, the Experiments page previously showed a generic “MLflow is currently unavailable” message. This change separates those cases and gives users clearer guidance.
**What changed**
- **`MlflowExperimentsPage`**: Uses `useIsMlflowCRAvailable` to gate the page:
  - While status is loading → spinner via `ApplicationsPage`
  - BFF reports not configured → new `MLflowNotConfigured` empty state with “Who’s my administrator?” link
  - Configured → load the Module Federation remote; remote load failure still falls back to `MLflowUnavailable`
- **`MLflowNotConfigured`**: New empty state component with admin-configuration messaging and support icon.
- **`useMLflowStatus`**: On the first BFF status failure, set `loaded: true` so consumers can stop waiting and render a terminal state instead of spinning forever. Transient errors after a successful fetch still preserve the last known `configured` value.

- **Tests**: Cypress mock tests split into “not configured” vs “service unavailable”; unit tests updated for the new `loaded: true` behavior on first BFF error.

image

## How Has This Been Tested?

Manually tested the empty state flow by disabling the MLflow configuration in the **Rosa-ui-razzmatazz** cluster. 

- Ran unit tests for `useMLflowStatus` from the `frontend` workspace — all 14 tests pass.
- Ran `npm run lint` in `packages/mlflow-embedded` — passes after fixing import order/formatting in `MLflowNotConfigured.tsx`.
- Cypress mock tests were updated for the two empty-state paths; run locally with:
  ```bash
  npm run test:cypress-ci:coverage:nobuild -- --spec "../packages/cypress/cypress/tests/mocked/mlflow/mlflowExperiments.cy.ts"

## Test Impact
<!--- What tests have you done to cover the implemented functionality -->
<!--- If tests are not applicable, explain why here -->

- Unit: frontend/src/concepts/mlflow/hooks/__tests__/useMLflowStatus.spec.ts — expectations updated for loaded: true on first API failure.
- Cypress (mock): packages/cypress/cypress/tests/mocked/mlflow/mlflowExperiments.cy.ts — split into:
  - not-configured empty state when BFF reports unconfigured
  - service-unavailable empty state when remote fails to load in mocked env
- Page object: packages/cypress/cypress/pages/mlflowExperiments.ts — helpers for not-configured empty state and admin link.

## Request review criteria:
<!--- This PR will be merged by any repository approver when it meets all the points in the checklist -->
<!--- Go over all the following points, and put an `x` in all the boxes that apply. -->

If you have UI changes: 
<!--- You can ignore these if you are doing manifest, backend, internal logic, etc changes; aka non-UI / visual changes -->

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **New Features**
  * Added a "Not configured" empty state with admin guidance, new UI strings, and a support icon.
  * UI pages now use availability + error flags to choose the correct empty/loading UI.

* **Bug Fixes**
  * Status handling now includes an explicit error flag; loaded/configured state transitions and recovery after failures are handled so the UI doesn't get stuck.
  * UI distinguishes "Not configured" vs "Service unavailable" more reliably.

* **Tests**
  * Expanded unit and Cypress tests and added page helpers to cover error, unavailable, and not-configured flows.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->


frontend/src/concepts/mlflow/hooks/useMLflowStatus.ts
```diff
@@ -40,7 +40,8 @@ export const useMLflowStatus = createSharedPollingStore<MLflowStatus>({
       // eslint-disable-next-line no-console
       console.warn('MLflow BFF status check failed (will suppress until recovery)', e);
     }
-    return previous;
+    // Mark loaded so consumers (e.g. embedded Experiments page) do not spin forever on first failure.
+    return { ...previous, loaded: true };
```
DaoDaoNoCode: When the BFF status check fails (network error, 500, BFF not deployed, etc.), this returns `{ configured: false, loaded: true }`. The consumer chain then goes:

`useIsMlflowCRAvailable` -> `available: false, loaded: true` -> `MlflowExperimentsPage` -> `empty: true` -> shows `MLflowNotConfigured` ("Admin configuration required").

So a BFF outage tells the user to ask their admin to enable MLflow, which is misleading -- the real problem is that the service is unreachable, not that it's unconfigured. Before this PR, returning `previous` (keeping `loaded: false`) meant the page just spun forever, which is also bad. But flipping `loaded: true` swings to the other extreme.

One way to fix this: add an `error` field to `MLflowStatus` so consumers can distinguish the three states (loading / configured / error):

```suggestion
    // Mark loaded so consumers (e.g. embedded Experiments page) do not spin forever on first failure.
    return { ...previous, loaded: true, error: true };
```

Then in `useIsMlflowCRAvailable`, you'd expose that error state, and `MlflowExperimentsPage` would render `MLflowUnavailable` when `error` is true instead of `MLflowNotConfigured`. The `fetchFn` success path would set `error: false` so recovery works naturally.

This is the core issue with the PR as written -- without it, the feature works backwards for BFF-down scenarios.

packages/mlflow-embedded/shared/MLflowNotConfigured.tsx
```diff
@@ -0,0 +1,28 @@
...
+} from '@patternfly/react-core';
+// eslint-disable-next-line @odh-dashboard/no-restricted-imports
+import WhosMyAdministrator from '@odh-dashboard/internal/components/WhosMyAdministrator';
+import { MLFLOW_NOT_CONFIGURED_TITLE, MLFLOW_NOT_CONFIGURED_MESSAGE } from './const';
+import supportImg from '../images/Icon-Red_Hat-Support-A-Black-RGB.svg';
+
+const MLflowNotConfigured: React.FC = () => (
+  <EmptyState
+    headingLevel="h2"
+    icon={() => <img src={supportImg} alt="" width={60} height={60} />}
```
DaoDaoNoCode: The SVG (`Icon-Red_Hat-Support-A-Black-RGB.svg`) uses black paths with no `fill` attribute, so in dark mode this icon will be invisible against the dark background. The file name even says "Black-RGB".

A couple of options:
1. Use `currentColor` in the SVG paths so it inherits the text color from the PF theme.
2. Use a PatternFly icon (e.g. `WrenchIcon` or `CogIcon`) like `MLflowUnavailable` uses `ExclamationTriangleIcon` -- these auto-adapt to the theme.

nit: The inline arrow function `() => ` creates a new component reference each render. Since this component has no props and doesn't re-render, it's harmless here, but hoisting it to a module-level const is slightly cleaner:

```tsx
const SupportIcon: React.FC = () => <img src={supportImg} alt="" width={60} height={60} />;

// then use: icon={SupportIcon}
```

packages/mlflow-embedded/experiments/MlflowExperimentsPage.tsx
```diff
@@ -47,8 +51,9 @@ const MlflowExperimentsPage: React.FC = () => {
 
   return (
     <ApplicationsPage
-      loaded
-      empty={false}
+      loaded={mlflowLoaded}
+      empty={mlflowLoaded && !mlflowAvailable}
```
DaoDaoNoCode, line 59: The `{mlflowAvailable && (...)}` guard on line 104 is redundant -- when `empty` is `true`, `ApplicationsPage` already skips rendering `children` and shows `emptyStatePage` instead (confirmed by reading `ApplicationsPage.tsx` lines 129-142). The guard is harmless but adds visual noise. Consider removing it and relying on `ApplicationsPage`'s built-in empty gating.

nit: not blocking, just something to keep clean.

packages/mlflow-embedded/experiments/MlflowExperimentsPage.tsx
```diff
@@ -27,13 +29,15 @@ import {
...
 import MLflowUnavailable from '../shared/MLflowUnavailable';
+import MLflowNotConfigured from '../shared/MLflowNotConfigured';
 import MlflowBreadcrumbs, { BreadcrumbEntry } from '../shared/MlflowBreadcrumbs';
 import LaunchMlflowButton from '../shared/LaunchMlflowButton';
 
 const MlflowExperimentsPage: React.FC = () => {
   const [searchParams] = useSearchParams();
   const workspace = searchParams.get(WORKSPACE_QUERY_PARAM) ?? '';
   const [breadcrumbs, setBreadcrumbs] = React.useState<BreadcrumbEntry[]>([]);
+  const { available: mlflowAvailable, loaded: mlflowLoaded } = useIsMlflowCRAvailable();
```
DaoDaoNoCode: Worth noting: the Prompts page (`MlflowPromptManagementPage.tsx`) still has `loaded` hardcoded to `true` and `empty={false}`, so it doesn't have this "not configured" guard. If a user navigates to Prompts when MLflow isn't configured, they'll see the old `MLflowUnavailable` fallback from the `loadRemote` catch rather than the new admin-guidance empty state.

This doesn't need to block the PR, but I'd suggest a follow-up ticket to apply the same pattern to the Prompts page for consistency.

packages/cypress/cypress/tests/mocked/mlflow/mlflowExperiments.cy.ts
```diff
@@ -90,9 +90,19 @@ describe('MLflow Experiments page wrapper', () => {
         .should('contain', invalidWorkspace);
     });
 
-    it('should show unavailable state when MLflow is not configured', () => {
+    it('should show not-configured empty state when MLflow BFF reports unconfigured', () => {
       initIntercepts({ mlflowConfigured: false });
       mlflowExperiments.visit(PROJECT_A);
+      mlflowExperiments.findNotConfiguredEmptyState().should('be.visible');
+      mlflowExperiments.findNotConfiguredAdminLink().should('be.visible');
```
DaoDaoNoCode, line 116: nit: Adding negative assertions here would make these tests more robust -- confirming that the *other* empty state is NOT rendered when one is shown. CodeRabbit suggested the same thing and I agree it's good practice, though not blocking:

```suggestion
      mlflowExperiments.findNotConfiguredEmptyState().should('be.visible');
      mlflowExperiments.findNotConfiguredAdminLink().should('be.visible');
      mlflowExperiments.findMlflowUnavailableState().should('not.exist');
```

DaoDaoNoCode (changes requested):
This PR adds a dedicated "not configured" empty state on the MLflow Experiments page so users get clearer guidance when MLflow isn't set up, rather than the generic "unavailable" message. The approach is clean -- gate behind `useIsMlflowCRAvailable`, show `MLflowNotConfigured` via `ApplicationsPage`'s `emptyStatePage` prop, and only load the MF remote when the BFF confirms availability.

Local checks: lint (0 errors, 129 pre-existing warnings) -- type-check -- useMLflowStatus unit tests (14/14 pass). CI all green.

CodeRabbit raised a concern about `onError` conflating transport failures with "not configured" -- that concern is valid and I'm expanding on it below as the main issue. Their nit about negative assertions in the Cypress tests is reasonable but non-blocking.

Leaving a few comments inline -- the main thing to address is that a BFF outage currently shows the admin-configuration empty state instead of the service-unavailable one.

packages/cypress/cypress/tests/mocked/mlflow/mlflowExperiments.cy.ts
Taj010: Done! Adjusted the tests to have negative assertions.

packages/mlflow-embedded/experiments/MlflowExperimentsPage.tsx
Taj010: Done! Removed the` {mlflowAvailable && (...)}` guard.

packages/mlflow-embedded/shared/MLflowNotConfigured.tsx
Taj010: Done! Updated the support SVG to use `currentColor` and hoisted a module-level `SupportIcon` (an inline SVG so it inherits the theme color in dark mode).

frontend/src/concepts/mlflow/hooks/useMLflowStatus.ts
Taj010: Done! Added an `error` field on MLflowStatus — first BFF failure sets` loaded: true, error: true;` success clears` error: false.` `useIsMlflowCRAvailable` exposes `error`, and `MlflowExperimentsPage` shows `MLflowUnavailable` when error is true and `MLflowNotConfigured` when the BFF reports unconfigured.

bobbravo2:
this one LGTM once Juntao is back online. Great job testing the multiple failure modes, and leveraging our new BFF to have a more gracefully degrading customer experience!

Gkrumbach07:

DaoDaoNoCode:
@Taj010 Re: the `dangerouslySetInnerHTML` usage in `MLflowNotConfigured.tsx` — could you switch to PatternFly's `createIcon` instead? This is how all 50+ custom icons in the codebase are done. Multi-path SVGs are handled by concatenating the `d` attributes into a single `svgPath` string.

See `frontend/src/images/icons/GroupIcon.ts` for a multi-path example:

```ts
import { createIcon } from '@patternfly/react-icons/dist/esm/createIcon';

const SupportIcon = createIcon({
  name: 'SupportIcon',
  width: 36,
  height: 36,
  svgPath:
    'M7.5,14.38h-1A2.12,2.12,0,0,0,4.38,16.5v5A2.12,... M29.5,14.38h-1a2.12,... M28,13.17a.63...',
  xOffset: 0,
  yOffset: 0,
});
```

This eliminates the raw SVG file, the `dangerouslySetInnerHTML`, the eslint-disable, and the hardcoded `60px` inline styles — and `currentColor` support comes for free via PF's icon rendering.

frontend/src/concepts/mlflow/hooks/useMLflowStatus.ts
```diff
@@ -40,7 +42,12 @@ export const useMLflowStatus = createSharedPollingStore<MLflowStatus>({
       // eslint-disable-next-line no-console
       console.warn('MLflow BFF status check failed (will suppress until recovery)', e);
     }
-    return previous;
+    // Mark loaded so consumers (e.g. embedded Experiments page) do not spin forever on first failure.
+    // Only surface error when we have no prior successful status (transient failures keep last good state).
+    if (!previous.loaded) {
+      return { ...previous, loaded: true, error: true };
+    }
+    return { ...previous, loaded: true };
```
DaoDaoNoCode: nit: The `else` branch here spreads `previous` including whatever `error` value it currently has. After a first failure (`error: true`), a second consecutive failure goes through this branch and preserves `error: true` via the spread -- which is the correct behavior. But it's not immediately obvious that the spread silently carries `error` forward. An explicit `return { ...previous, loaded: true, error: previous.error };` would make that intent clearer, though this is purely a readability suggestion.

packages/mlflow-embedded/icons/SupportIcon.ts
```diff
@@ -0,0 +1,13 @@
+import { createIcon } from '@patternfly/react-icons/dist/esm/createIcon';
+
+const SupportIcon = createIcon({
+  name: 'SupportIcon',
+  width: 36,
+  height: 36,
```
DaoDaoNoCode: nit: The `width` and `height` here (36) define the SVG viewBox dimensions, not the rendered size -- PF's `createIcon` scales the icon to match the EmptyState icon slot. The SVG path coordinates (max ~30.38) fit within the 36x36 viewBox, so this is fine. Just noting it since 36 is an unusual viewBox size compared to the more common 512x512 or 1024x1024 used by other icons in this repo (e.g., `ScrewWrenchIcon`). Not a problem, just a difference to be aware of if anyone later tries to swap path data between icons.

packages/cypress/cypress/tests/mocked/mlflow/mlflowExperiments.cy.ts
```diff
@@ -90,10 +90,33 @@ describe('MLflow Experiments page wrapper', () => {
...
+      mlflowExperiments.findMlflowUnavailableState().should('not.exist');
+    });
+
+    it('should show unavailable empty state when MLflow BFF status check fails', () => {
+      initIntercepts();
```
DaoDaoNoCode: The outer `beforeEach` at line 29-31 already calls `initIntercepts()` (which registers the `mlflowStatus` intercept as configured: true). This second `initIntercepts()` call re-registers the same intercepts. The test then overrides with the 500-status intercept on line 103. Due to Cypress LIFO ordering, the 500-override wins on the first match, so the test works correctly -- but the redundant `initIntercepts()` adds noise. Consider dropping it and relying on `beforeEach`.

```suggestion
    it('should show unavailable empty state when MLflow BFF status check fails', () => {
      cy.intercept('GET', MLFLOW_BFF_STATUS_URL, { statusCode: 500 }).as('mlflowStatusError');
```

DaoDaoNoCode (commented):
This PR cleanly separates the "not configured" vs "service unavailable" empty states on the MLflow Experiments page, which is the right UX to give users actionable guidance. The hook changes (`useMLflowStatus`, `useIsMlflowCRAvailable`) are well-structured, the new `MLflowNotConfigured` component follows existing PF patterns, and the three-way state logic (loading / not configured / error) is correct.

Local checks: lint (0 errors, 129 pre-existing warnings) -- type-check passes -- MLflow hooks unit tests 42/42 pass. CI all green.

Prior review: DaoDaoNoCode requested changes around five items -- (1) error field on MLflowStatus, (2) dark-mode SVG fix, (3) redundant mlflowAvailable guard, (4) Prompts page follow-up, (5) negative Cypress assertions. All five are addressed in the current code. CodeRabbit raised a nit about making the remote-failure test deterministic, which is still open but non-blocking.

Left a couple of minor comments inline -- nothing blocking.

DaoDaoNoCode (commented):
Cross-codebase pattern review — see inline comments for findings validated against existing patterns in eval-hub, feature-store, storage-classes, and model-serving.

packages/mlflow-embedded/experiments/MlflowExperimentsPage.tsx
```diff
@@ -47,8 +55,9 @@ const MlflowExperimentsPage: React.FC = () => {
 
   return (
     <ApplicationsPage
```
DaoDaoNoCode, line 57: **Pattern mismatch: missing `PageSection` wrapper on `emptyStatePage`**

Every other `emptyStatePage` usage in the codebase wraps the content in `<PageSection hasBodyWrapper={false} isFilled>`. Examples:
- `StorageClassesPage.tsx` — wraps `EmptyState` in `<PageSection hasBodyWrapper={false} isFilled>`
- `EvaluationsPage.tsx` (eval-hub) — same pattern
- `PipelineCoreApplicationPage.tsx` — same pattern (via `NoPipelineServer`)

`ApplicationsPage` renders `emptyStatePage` raw (no wrapping), unlike the default `empty` path which does add its own `<PageSection>`. Without the wrapper, the empty state won't get the `isFilled` flex behavior and the vertical centering will break.\n\nSuggestion:\n```tsx\nemptyStatePage={\n  <PageSection hasBodyWrapper={false} isFilled>\n    {mlflowStatusError ? <MLflowUnavailable /> : <MLflowNotConfigured />}\n  </PageSection>\n}\n```

packages/mlflow-embedded/shared/MLflowNotConfigured.tsx
```diff
@@ -0,0 +1,28 @@
...
+  EmptyStateBody,
+  EmptyStateFooter,
+  EmptyStateVariant,
+} from '@patternfly/react-core';
+// eslint-disable-next-line @odh-dashboard/no-restricted-imports
+import WhosMyAdministrator from '@odh-dashboard/internal/components/WhosMyAdministrator';
+import { MLFLOW_NOT_CONFIGURED_TITLE, MLFLOW_NOT_CONFIGURED_MESSAGE } from './const';
+import SupportIcon from '../icons/SupportIcon';
+
+const MLflowNotConfigured: React.FC = () => (
```
DaoDaoNoCode: **Pattern mismatch: admin/non-admin differentiation missing**

The established codebase pattern for "not configured" empty states distinguishes between admin and non-admin users. See:
- `EvaluationsPage.tsx` (eval-hub): `clusterAdmin ? <CogIcon + "enable the service"> : <SupportIcon + "contact your administrator" + WhosMyAdministrator>`
- `FeatureStoreCoreLoader.tsx`: `isAdmin ? <CogIcon + admin instructions> : <SupportIcon + WhosMyAdministrator>`

This PR always shows the `SupportIcon` + `WhosMyAdministrator` variant regardless of user role. An admin seeing "request that your administrator enable MLflow" when *they are* the administrator is confusing. Consider branching:
```tsx
{isAdmin ? (
  <EmptyState icon={CogIcon} titleText="MLflow not configured">
    <EmptyStateBody>Enable MLflow for this project to start using experiments.</EmptyStateBody>
  </EmptyState>
) : (
  <MLflowNotConfigured />
)}
```

packages/mlflow-embedded/icons/SupportIcon.ts
```diff
@@ -0,0 +1,13 @@
+import { createIcon } from '@patternfly/react-icons/dist/esm/createIcon';
+
+const SupportIcon = createIcon({
```
DaoDaoNoCode, line 7: **Duplicate `SupportIcon` — different SVG paths and viewBox across packages**

This creates a third copy of `SupportIcon` in the monorepo:
- `packages/eval-hub/frontend/src/app/icons/SupportIcon.ts` — 56x56 viewBox
- `packages/feature-store/src/icons/header-icons/SupportIcon.ts` — 56x56 viewBox (identical to eval-hub)
- **This PR**: `packages/mlflow-embedded/icons/SupportIcon.ts` — 36x36 viewBox, **different SVG path**

The 36x36 viewBox is valid (other icons in `frontend/src/images/icons/` like `ProjectIcon`, `NotebookIcon`, `PipelineIcon` also use 36x36), but having the same component name with a different icon across packages is confusing.

Consider:
1. If the icon is intentionally the same (headset icon): reuse the eval-hub/feature-store version (56x56 paths) or extract to a shared location
2. If the icon is intentionally different: give it a distinct name to avoid confusion (e.g., `HeadsetIcon` or `AdminSupportIcon`)

packages/mlflow-embedded/experiments/MlflowExperimentsPage.tsx
```diff
@@ -27,13 +29,19 @@ import {
 } from '@odh-dashboard/internal/routes/pipelines/mlflow';
 import { EXPERIMENTS_PAGE_TITLE } from '../shared/const';
 import MLflowUnavailable from '../shared/MLflowUnavailable';
+import MLflowNotConfigured from '../shared/MLflowNotConfigured';
 import MlflowBreadcrumbs, { BreadcrumbEntry } from '../shared/MlflowBreadcrumbs';
 import LaunchMlflowButton from '../shared/LaunchMlflowButton';
 
 const MlflowExperimentsPage: React.FC = () => {
   const [searchParams] = useSearchParams();
```
DaoDaoNoCode, line 37: **Consistency: Prompts page has the same `loaded`/`empty={false}` pattern — should it also get the not-configured treatment?**

`MlflowPromptManagementPage.tsx` still uses `loaded` and `empty={false}` (hardcoded), which is exactly what this Experiments page had before this PR. If a user navigates to Prompts when MLflow is not configured, they'll see either a loading spinner forever or a broken remote load — not the new "not configured" empty state.\n\nWorth considering whether the Prompts page should get the same `useIsMlflowCRAvailable` gate in this PR or a follow-up.

frontend/src/concepts/mlflow/hooks/useMLflowStatus.ts
DaoDaoNoCode: **`onError`: transient failure after success silently drops the `error` field**

Looking at the `onError` handler:
```ts
if (!previous.loaded) {
  return { ...previous, loaded: true, error: true };
}
return { ...previous, loaded: true };
```

When `previous.loaded` is already `true` (transient failure after a successful fetch), the spread `{ ...previous, loaded: true }` preserves whatever `error` was on `previous`. If `previous.error` was `true` from a prior first-failure, a subsequent success sets `error: false` (via `fetchFn` return), but if *two consecutive* errors happen (first sets `error: true`, poll retries, second error hits the `else` branch), the `error: true` from the first failure is carried forward — which is correct.

However, the `else` branch would read cleaner with an explicit `error: false`:
```ts
return { ...previous, loaded: true, error: false };
```
This makes the intent unambiguous: transient errors after success should NOT surface the error flag to the UI. Right now it works by accident (the successful fetch already set `error: false`), but a future refactor could break this assumption.

packages/cypress/cypress/tests/mocked/mlflow/mlflowExperiments.cy.ts
```diff
@@ -90,10 +90,33 @@ describe('MLflow Experiments page wrapper', () => {
...
 
-    it('should show unavailable state when MLflow is not configured', () => {
+    it('should show not-configured empty state when MLflow BFF reports unconfigured', () => {
       initIntercepts({ mlflowConfigured: false });
       mlflowExperiments.visit(PROJECT_A);
+      mlflowExperiments.findNotConfiguredEmptyState().should('be.visible');
+      mlflowExperiments.findNotConfiguredAdminLink().should('be.visible');
+      mlflowExperiments.findMlflowUnavailableState().should('not.exist');
+    });
+
```
DaoDaoNoCode, line 120: **Cypress: intercept ordering may cause flakiness**

This test calls `initIntercepts()` (which registers `interceptMlflowStatus(true)` -> `configured: true`) and then immediately overrides with `cy.intercept('GET', MLFLOW_BFF_STATUS_URL, { statusCode: 500 })`. \n\nCypress intercepts are LIFO — the last-registered intercept wins. So this works correctly *today*. But it's fragile: if `initIntercepts` is ever refactored to register its status intercept later (e.g., in a `beforeEach`), the override would silently stop working.\n\nThe more robust pattern is to pass the error scenario into `initIntercepts` directly:\n```ts\ninitIntercepts({ mlflowStatusError: true });\n```\nwith `initIntercepts` conditionally registering either the success or error intercept. This matches how `mlflowConfigured` is already handled.

packages/cypress/cypress/tests/mocked/mlflow/mlflowExperiments.cy.ts
```diff
@@ -90,10 +90,33 @@ describe('MLflow Experiments page wrapper', () => {
...
+    it('should show service-unavailable empty state when MLflow remote fails to load', () => {
+      // mlflowConfigured: true passes the "not configured" gate, but the
+      // Module Federation remote is not running in the mocked environment,
+      // so loadRemote rejects and the page falls back to MLflowUnavailable.
+      initIntercepts({ mlflowConfigured: true });
+      mlflowExperiments.visit(PROJECT_A);
       mlflowExperiments.findMlflowUnavailableState().should('be.visible');
+      mlflowExperiments.findNotConfiguredEmptyState().should('not.exist');
+      mlflowExperiments.findNotConfiguredAdminLink().should('not.exist');
     });
```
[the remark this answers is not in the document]
Taj010: Done! The remote-failure test now stubs `/_mf/mlflowEmbedded/mlflow/static-files/federated/remoteEntry.js` via `interceptMlflowEmbeddedRemoteFailure()`, waits on `@mlflowEmbeddedRemoteEntry`, and asserts the unavailable empty state with negative checks for not-configured.

packages/mlflow-embedded/experiments/MlflowExperimentsPage.tsx
Taj010: Done! Wrapped `emptyStatePage` in `<PageSection hasBodyWrapper={false} isFilled>` to match the established pattern.

packages/mlflow-embedded/shared/MLflowNotConfigured.tsx
Taj010: Done! `MLflowNotConfigured` now branches on `isAdmin`: admins see `CogIcon` + enable messaging; non-admins see `SupportIcon` + WhosMyAdministrator. Cypress tests cover both paths.

packages/mlflow-embedded/icons/SupportIcon.ts
Taj010: Done! Replaced the 36×36 custom path with the same 56×56 SVG used in eval-hub and feature-store, via PF `createIcon`.

frontend/src/concepts/mlflow/hooks/useMLflowStatus.ts
Taj010: Done! The else branch now explicitly sets `error: false` for transient failures after a successful fetch.

Taj010: Done! Same fix — transient post-success errors return `{ ...previous, loaded: true, error: false }` so the UI keeps the last good state.

packages/cypress/cypress/tests/mocked/mlflow/mlflowExperiments.cy.ts
Taj010: Done! Removed the redundant `initIntercepts()` call; the BFF-error test now uses `interceptMlflowStatusError()` from `mlflowUtils`.

packages/cypress/cypress/tests/mocked/mlflow/mlflowExperiments.cy.ts
Taj010: Done! Added `interceptMlflowStatusError()` helper so the error scenario is registered directly instead of relying on LIFO override ordering.

packages/mlflow-embedded/icons/SupportIcon.ts
Taj010: Thanks for the note — aligned to 56×56 to match eval-hub/feature-store.

DaoDaoNoCode:

packages/mlflow-embedded/experiments/MlflowExperimentsPage.tsx
DaoDaoNoCode: Already addressed in the current diff — `MlflowPromptManagementPage.tsx` now uses `useIsMlflowCRAvailable` with the same `loaded`/`empty`/`emptyStatePage` pattern as the Experiments page, including `MLflowNotConfigured` import. Author just didn't reply on this thread.

frontend/src/concepts/mlflow/hooks/__tests__/useIsMlflowPipelinesAvailable.spec.ts
```diff
@@ -32,7 +32,7 @@ describe('useIsMlflowPipelinesAvailable', () => {
   beforeEach(() => {
     jest.clearAllMocks();
     mockPipelinesAreaAvailable(true);
-    mockUseIsMlflowCRAvailable.mockReturnValue({ available: true, loaded: true });
+    mockUseIsMlflowCRAvailable.mockReturnValue({ available: true, loaded: true, error: false });
     mockUseIsMlflowDSPAEnabled.mockReturnValue({ enabled: true, loaded: true });
```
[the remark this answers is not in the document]
DaoDaoNoCode: Already in the diff — `useIsMlflowPipelinesAvailable` now destructures `error: crError` from `useIsMlflowCRAvailable` and returns `error: isPipelinesAreaAvailable && crError`. Two new tests cover it: error=true when CR reports error, and error=false when pipelines area is disabled even if CR has error.

frontend/src/concepts/mlflow/hooks/useMLflowStatus.ts
```diff
@@ -20,29 +20,38 @@ let lastPollErrored = false;
...
+    teardownGracePeriod: TEARDOWN_GRACE_MS,
+    onError: (e, previous) => {
+      if (!lastPollErrored) {
+        lastPollErrored = true;
+        // eslint-disable-next-line no-console
+        console.warn('MLflow BFF status check failed (will suppress until recovery)', e);
+      }
+      // Mark loaded so consumers (e.g. embedded Experiments page) do not spin forever on first failure.
+      // Only surface error when we have no prior successful status (transient failures keep last good state).
+      if (!previous.loaded) {
```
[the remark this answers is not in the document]
DaoDaoNoCode: Fixed in the current code. The `hadSuccessfulResponse` flag prevents the oscillation: first failure sets `error: true`, and any failure after a prior success explicitly returns `error: false` (preserving last good state). The test "should keep last successful value on transient error" covers this path.

DaoDaoNoCode (approved):
/lgtm

PR #7610, opened 2026-05-19
feat: add agent-ops mod-arch module

Gkrumbach07:
## Description

Add the `agent-ops` mod-arch skeleton following the onboarding guide. This is the same process done for MLflow (#6336) and gen-ai.

In this PR:

- Add the `agent-ops` mod-arch skeleton following the onboarding guide
- Add a simple BFF connection test to verify frontend/BFF integration
- Register `agentOps` feature flag (dev temporary flag, off by default)
- Register "Agent Ops" nav section with "My Agents" nav item gated behind the flag
- OWNERS and OWNERS_ALIASES updated for agent-ops team
- No meaningful business logic yet, just the scaffold

https://github.com/user-attachments/assets/f214feb0-4aea-467c-92f9-fa7a304e42fc

## How Has This Been Tested?

- Run the agent-ops BFF: `cd packages/agent-ops/bff && go build ./cmd/ && ./agent-ops-bff -port 4021 -auth-method disabled`
- Run the agent-ops frontend: `cd packages/agent-ops/frontend && DEPLOYMENT_MODE=federated PORT=9111 PROXY_PORT=4021 npm run start:dev`
- Run the Dashboard: `npm run dev`
- Enable `agentOps` in feature flags modal
- Verify "Agent Ops > My Agents" appears in nav with WIP badge
- Verify BFF Connection OK on the main page

## Test Impact

Not applicable (skeleton scaffold only)

## Request review criteria:

- [x] All lint checks pass
- [x] Port validation passes (local 9111, BFF 4021)
- [x] OWNERS file created with agent-ops-approvers alias
- [x] OWNERS_ALIASES updated
- [x] No leftover template/mlflow references
- [x] Feature flag registered in areas/types.ts, areas/const.ts, k8sTypes.ts
- [x] BFF healthcheck and status endpoints working
- [x] Frontend builds in both standalone and federated mode

https://issues.redhat.com/browse/RHOAIENG-62591

## Summary by CodeRabbit

* **New Features**
  * Added Agent Ops module: REST API (health, user, namespaces), BFF service, and a new Agent Ops UI with routing, namespace selection, and health status.

* **Configuration**
  * Standalone and federated deployment support, container build/runtime targets, Makefile workflows, ownership aliases, and linting/ignore rules.

* **Documentation**
  * Detailed developer, contributing, deployment, and API docs included.

* **Tests**
  * OpenAPI contract tests, Go unit tests, and Cypress E2E (including accessibility) added.


Gkrumbach07:
Hey @lucferbux, this follows the same onboarding process you reviewed for MLflow (#6336). Would appreciate your review on the integration points when you get a chance. The scaffold is minimal (no business logic yet), just the mod-arch module structure, feature flag, and BFF connection test.

lucferbux (changes requested):
## Review Summary

Thanks for scaffolding the agent-ops module! The overall structure is solid, but there are a few structural gaps that need to be addressed before this can merge. These are important because they affect plugin discovery, feature gating, and monorepo consistency.

**Key question**: Was this scaffold generated by running `npx mod-arch-installer -n agent-ops`? I ran the installer locally and noticed several differences between its output and this PR. It's important that new modules start from the installer output to maintain consistent structure across the monorepo — it ensures the build tooling, plugin discovery, and extension system all work correctly out of the box. If the installer output was modified or if the module was assembled manually from other modules, the gaps below are the result of that divergence.

The [onboarding guide](https://github.com/opendatahub-io/odh-dashboard/blob/main/docs/onboard-modular-architecture.md) documents the expected flow: `npx mod-arch-installer -n <name>`, then customize from there. Starting from a consistent baseline keeps modules interoperable and makes it easier for other teams to contribute.

### Blockers

1. **Missing `exports` field** in root `package.json` — required for static plugin discovery via `discoverPluginPackages.js`
2. **Missing `app.area` extension** in `extensions.ts` — required for feature flag gating via `PluginStore`

### Convention Issues

3. **Non-standard proxy path** (`/_bff/agent-ops/api` vs `/agent-ops/api`)
4. **Non-standard nav route** (`/agent-ops-bff/main-view` vs `/agent-ops/main-view`) — leaks BFF implementation detail into user-facing URLs
5. **Custom 318-line ESLint config** (`frontend/.eslintrc.cjs`) instead of using the shared `@odh-dashboard/eslint-config` — creates maintenance burden and drift from monorepo conventions
6. **MUI dependencies** (`@mui/material`, `@emotion/react`, etc.) included but likely not needed for a standard module (MUI is only used in Kubeflow mode)
7. **ESLint version mismatch** — frontend pins `eslint@^8.57.0` in `optionalDependencies` while the monorepo uses ESLint 9.x
8. **No unit tests or E2E tests** — will fail the `modular-arch-quality-gates.yml` CI checks
9. **`/api/v1/status` endpoint** exists in BFF code and contract tests but is missing from the OpenAPI spec

See inline comments for specific locations and suggested fixes.

packages/agent-ops/package.json
```diff
@@ -0,0 +1,56 @@
+{
+  "private": true,
+  "name": "@odh-dashboard/agent-ops",
+  "description": "Agent Ops plugin.",
+  "version": "0.0.0",
```
lucferbux, line 5: **Blocker — Missing `exports` field.** This is required for `discoverPluginPackages.js` to find and statically bundle the module's extensions at build time. Without it, the plugin won't be discovered by the host.

Add:
```json
"exports": {
  "./extensions": "./extensions.ts"
},
```

You'll also need a root-level `extensions.ts` file that re-exports from `frontend/src/odh/extensions.ts`. The `mod-arch-installer` generates both of these automatically.

packages/agent-ops/frontend/src/odh/extensions.ts
```diff
@@ -0,0 +1,43 @@
+import type { NavExtension, RouteExtension } from '@odh-dashboard/plugin-core/extension-points';
```
lucferbux: **Blocker — Missing `app.area` extension.** This is what registers the feature area with the `PluginStore` so that `isExtensionInUse()` can gate extensions based on the `agent-ops` flag. Without it, the `flags: { required: ['agent-ops'] }` on your nav and route extensions won't resolve correctly.

The `AreaExtension` type should also be imported. The installer generates this:

```typescript
import type {
  NavExtension,
  RouteExtension,
  AreaExtension,
} from '@odh-dashboard/plugin-core/extension-points';

const AGENT_OPS = 'agent-ops';

const extensions: (NavExtension | RouteExtension | AreaExtension)[] = [
  {
    type: 'app.area',
    properties: {
      id: AGENT_OPS,
      devFlags: [AGENT_OPS],
    },
  },
  // ... nav and route extensions
];
```

packages/agent-ops/package.json
```diff
@@ -0,0 +1,56 @@
...
+  "module-federation": {
+    "name": "agentOps",
+    "remoteEntry": "/remoteEntry.js",
+    "authorize": true,
+    "tls": false,
+    "proxy": [
+      {
+        "path": "/_bff/agent-ops/api",
```
lucferbux: **Convention — Non-standard proxy path.** The `/_bff/agent-ops/api` prefix doesn't follow the standard pattern used by other modules. The backend MF proxy in `backend/src/routes/module-federation.ts` expects the format `/agent-ops/api` (no `_bff` prefix).

Should be:
```json
"path": "/agent-ops/api"
```

packages/agent-ops/frontend/src/odh/extensions.ts
```diff
@@ -0,0 +1,43 @@
...
+  },
+  {
+    type: 'app.navigation/href',
+    flags: {
+      required: [AGENT_OPS],
+    },
+    properties: {
+      id: 'agent-ops-my-agents',
+      title: 'My Agents',
+      href: '/agent-ops-bff/main-view',
```
lucferbux: **Convention — Non-standard route path.** `/agent-ops-bff/main-view` includes `-bff` in the user-facing URL, which leaks an implementation detail. The installer generates `/agent-ops/main-view`. Same applies to the `path` property below and the route extension path.

```diff
- href: '/agent-ops-bff/main-view',
+ href: '/agent-ops/main-view',
```

packages/agent-ops/frontend/.eslintrc.cjs
```diff
@@ -0,0 +1,318 @@
+const path = require('path');
+
+module.exports = {
```
lucferbux: **Convention — Custom ESLint config.** This is a 318-line bespoke ESLint config using the legacy `.eslintrc.cjs` format. Within the monorepo, packages should use the shared `@odh-dashboard/eslint-config` (already in your root `devDependencies`).

The root-level `.eslintrc.js` in this PR correctly uses the shared config:
```js
module.exports = require('@odh-dashboard/eslint-config')
  .extend({...})
  .recommendedReactTypescript(__dirname);
```

The frontend should follow the same approach rather than maintaining a separate 318-line config that will drift from monorepo conventions over time.

lucferbux:
@Gkrumbach07 I'm asking if the rule it was followed, cause we are thinking to create a skill so we have a more straightforward way to onboard with the latest features that we are adding in the mod-arch installer, based on this review I think Claude ignore the current rule so maybe we'll need to investigate how to better ensure we get the same onboarding for all modules

Gkrumbach07:
When scaffolding this @lucferbux I used the mlflow BFF as a base. so it was basically. copying that and ripping out all thew business logic. I will re run this with the new skill at the forefront. instead of a rule which get lost in the sauce (sorta like a bay leaf in some sop, you know its there but im not sure you can actually taste it, too much else going on)

It should rather be a basic skill that then has a bunch of references on how to onboard a BFF

lucferbux:
> When scaffolding this @lucferbux I used the mlflow BFF as a base. so it was basically. copying that and ripping out all thew business logic. I will re run this with the new skill at the forefront. instead of a rule which get lost in the sauce (sorta like a bay leaf in some sop, you know its there but im not sure you can actually taste it, too much else going on)
> 
> It should rather be a basic skill that then has a bunch of references on how to onboard a BFF

Yup, agree, thanks a lot for the feedback, let me work through a skill so it can help new teams, I'm gonna review the new changes to see if we can merge this asap

Gkrumbach07:
**`ok-to-skip-audit` label added** — the only new advisory is [GHSA-rmmh-p597-ppvv](https://github.com/advisories/GHSA-rmmh-p597-ppvv) (showdown, moderate severity, ReDoS). This is a transitive dependency from the mod-arch-installer template with no fix available. Every existing module (mlflow, gen-ai, automl, autorag, maas, eval-hub) carries the same advisory — it only flags as "new" here because agent-ops doesn't exist on the base branch yet.

lucferbux (approved):
## Follow-up Review

All 5 items from my previous review have been addressed after the rebuild from `mod-arch-installer`:

| # | Issue | Status |
|---|-------|--------|
| 1 | Missing `exports["./extensions"]` in `package.json` | ✅ Fixed |
| 2 | Missing `app.area` extension in `extensions.ts` | ✅ Fixed |
| 3 | Proxy path `/api` → `/agent-ops/api` | ✅ Fixed |
| 4 | Route paths `/` → `/agent-ops/main-view` | ✅ Fixed |
| 5 | Root `.eslintrc.js` using shared config | ✅ Fixed |

### Minor items to consider (non-blocking)

1. **Route not gated by feature flag** — `app.route` has `flags: { required: [] }` (empty). Nav items are correctly gated with `required: [AGENT_OPS]`, but the route itself is unprotected. A user could navigate directly to `/agent-ops/main-view/*` even when the feature flag is off. Consider adding `flags: { required: [AGENT_OPS] }` to the route as well.
2. **Dockerfile.workspace** — default `MODULE_NAME=mod-arch` could be updated to `MODULE_NAME=agent-ops`.
3. **CI** — Type-Check and audit checks are currently failing, those should be resolved before merge.

### Follow-up task

I created [RHOAIENG-63723](https://issues.redhat.com/browse/RHOAIENG-63723) to convert the module onboarding rule into a reusable Claude Code skill, so future module scaffolds can be guided/validated interactively.

/lgtm
/approve

Gkrumbach07:
## Context for future skill improvments

Rebuilt the module from a clean `npx mod-arch-installer -n agent-ops` (v1.16.1) scaffold, then layered customizations on top. Here's what was addressed:

### Lucas's review items

| # | Feedback | Resolution |
|---|----------|------------|
| 1 | Missing `exports` field | Added `"./extensions": "./frontend/src/odh/extensions.ts"` for plugin discovery |
| 2 | Missing `app.area` extension | Added with `featureFlags: ['agentOps']` matching `const.ts` registration |
| 3 | Non-standard proxy path (`/_bff/agent-ops/api`) | Changed to `/agent-ops/api` |
| 4 | Non-standard route path (`/agent-ops-bff/`) | Changed to `/agent-ops/main-view` |
| 5 | Custom 318-line ESLint config | Replaced with installer's `eslint.config.mjs` (flat config) |
| 6 | MUI dependencies | Removed (not in installer template) |
| 7 | ESLint version mismatch | Resolved by using installer template deps |
| 8 | No unit/e2e tests | Installer includes cypress + jest stubs |
| 9 | `/api/v1/status` missing from OpenAPI | Removed custom status handler, using installer's `/healthcheck` instead |

### CI fixes required for a new module

These are gaps in the onboarding process that any new module would hit:

- **`package-lock.json`** — root lockfile must be regenerated after adding a new workspace package, otherwise the Setup job fails on `git diff --exit-code`
- **`.github/actions/module-caches/action.yml`** — new modules must be manually registered here for CI to install their frontend `node_modules` (needed for `mod-arch-core` resolution)
- **Webpack version pin** — installer generates `"webpack": "^5.105.2"` which resolves to 5.107.0, but the monorepo root uses 5.104.1. The internal file layout changed between versions causing `ModuleNotFoundError` at build time. Pinned to `5.104.1` to match
- **`AgentOpsWrapper.tsx` imports** — files in `frontend/src/odh/` are compiled by the host dashboard which doesn't have the `~` path alias. Changed to relative imports
- **Contract test auth method** — installer's BFF only accepts `internal` or `user_token`, not `disabled`. Changed to `--auth-method=internal`
- **Audit false positive** — `ok-to-skip-audit` label added because the showdown ReDoS advisory (GHSA-rmmh-p597-ppvv, moderate, no fix available) is present in every existing module but flags as "new" since agent-ops doesn't exist on the base branch

### Other customizations on top of installer

- Port 9111 (frontend) / 4021 (BFF) instead of installer defaults
- Added `bffConfig`, `e2eCiTags`, contract-tests directory, and e2e Makefile targets
- Added healthcheck proxy entry (`/agent-ops/healthcheck`) so the MainPage can verify BFF connectivity in federated mode
- Kept MainPage with BFF connection check (hits `/healthcheck`, checks `status === 'available'`)
- OWNERS file with agent-ops team aliases

PR #7611, opened 2026-05-19
Fix: run group dropdown footer overlap and align the menu footer

Taj010:
## Description
For https://issues.redhat.com/browse/RHOAIENG-58475

In the **Run group** selector (Create run form), the **Create new run group** action at the bottom of the dropdown overlapped the empty/filtered table content

Current:
image

Before: 
image

This change:

- Adds an optional `footer` prop to `SearchSelector`, rendered with PatternFly `MenuFooter` **outside** `MenuContent` (scrollable area), matching the [scrollable menu with search and footer](https://www.patternfly.org/components/menus/menu/html/#scrollable-menu-with-search-and-footer) pattern.
- Moves the run group **Create new run group** button into that footer via `ExperimentSelector`.
- Removes the footer `Divider` and custom `pf-v6-c-menu__footer` / sticky styling (PF footer uses box-shadow, not a divider).
- Removes the extra `pf-v6-c-menu__content` wrapper so the table matches other selectors (e.g. pipeline selector).
- Uses `initialLoaded` (not `loaded`) to keep the footer visible while search results reload, so the footer looks the same before and after typing in the search box.

## How Has This Been Tested?

Manually Tested 

- Open **Create run** (or any form with **Run group** selector).
- Open the run group dropdown with no search text — confirm **Create new run group** sits below the list/empty state without overlap.
- Type a search that returns no results — confirm empty state and footer do not overlap; footer styling matches the initial state.
- Type a search that returns results — confirm footer remains visible and clickable while results load.
- Click **Create new run group** — modal opens and dropdown closes.

## Test Impact

Manual UI verification only. No new unit or Cypress tests; behavior is layout/structure in existing components. Existing selector tests should be unaffected.

## Request review criteria:

If you have UI changes: 

## Summary by CodeRabbit

* **New Features**
  * Selector supports an optional footer rendered after the menu content; accepts static content or a callback (callback receives a menuClose helper).

* **Refactor**
  * “Create new run group” action moved into the selector footer; menu content simplified while preserving selection and load-more behavior.

* **Style**
  * Tables inside selector menus use a transparent background for consistent appearance.

### DIFF
--- a/frontend/src/app/App.scss
+++ b/frontend/src/app/App.scss
@@ -74,4 +74,8 @@ body,
   &.pf-m-drilldown:not(.pf-m-scrollable) {
     overflow: hidden;
   }
+  // Make tables inside menus transparent for consistent background
+  .pf-v6-c-table {
+    --pf-v6-c-table--BackgroundColor: transparent;
+  }
 }
--- a/frontend/src/app/App.scss
+++ b/frontend/src/app/App.scss
@@ -74,7 +74,11 @@ body,
   &.pf-m-drilldown:not(.pf-m-scrollable) {
     overflow: hidden;
   }
-  // Make tables inside menus transparent for consistent background
+}
+
+// Tables inside SearchSelector menus (e.g. run group selector) have their own background
+// which clashes with the menu in dark mode.
+.odh-search-selector__menu {
   .pf-v6-c-table {
     --pf-v6-c-table--BackgroundColor: transparent;
   }
--- a/frontend/src/components/searchSelector/SearchSelector.tsx
+++ b/frontend/src/components/searchSelector/SearchSelector.tsx
@@ -7,6 +7,7 @@ import {
   Menu,
   MenuContainer,
   MenuContent,
+  MenuFooter,
   MenuList,
   MenuSearch,
   MenuSearchInput,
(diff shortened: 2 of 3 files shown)


frontend/src/components/searchSelector/SearchSelector.tsx
```diff
@@ -138,6 +142,13 @@ const SearchSelector: React.FC<SearchSelectorProps> = ({
                 : children}
             </MenuList>
           </MenuContent>
+          {footer != null && (
```
nananosirova, line 162: There seems to be some out of place shading that's more visible in dark mode:

Image

Maybe we could instead have the empty state be full-width/non-scrollable?

Gkrumbach07:

frontend/src/components/searchSelector/SearchSelector.tsx
Taj010: @nananosirova Addressed! Now it looks like - 

image


The darker background was caused by Table components in ExperimentSelector having their own `--pf-v6-c-table--BackgroundColor`, which clashed with the menu background in dark mode. I tried setting the CSS variable via inline styles on parent elements (Menu, MenuContent), but that doesn't work because the table sets the variable on itself, so parent overrides don't cascade. I have added a selector in App.scss to override the table background to transparent when inside menus.

DaoDaoNoCode:

frontend/src/app/App.scss
```diff
@@ -75,3 +75,11 @@ body,
...
   }
 }
+
+// Tables inside SearchSelector menus (e.g. run group selector) have their own background
+// which clashes with the menu in dark mode.
+.odh-search-selector__menu {
+  .pf-v6-c-table {
+    --pf-v6-c-table--BackgroundColor: transparent;
+  }
+}
```
nananosirova, line 85: This change would also affect `PipelineSelector` and `PipelineVersionSelector`. Is this intentional?

PR #7612, opened 2026-05-19
[RHOAIENG-58474] Run group selector becomes permanently disabled after searching for non-existent name

srtanish1992:
https://redhat.atlassian.net/browse/RHOAIENG-58474

## Description

**Root cause:** The `MenuToggle` used `isDisabled={!filteredResources.length}`, which evaluated the *filtered* (search-matched) list. When a search matched nothing, `filteredResources` became empty, disabling the toggle. Since the search state was not cleared on close, the toggle stayed disabled permanently with no way for the user to reopen it.

**Fix (two changes in `CustomPipelineRunToolbarSelect.tsx`):**
1. Changed `isDisabled` to use `resources.length` (the full unfiltered list) — the toggle is now only disabled when there are genuinely no resources available, matching the pattern used in `SearchSelector`/`ActiveExperimentSelector`.
2. Updated `onOpenChange` to reset the search state when the dropdown closes, so the filter is fresh on next open.
Both `ExperimentFilterSelector` and `PipelineVersionFilterSelector` share the same inner component, so both are fixed.

https://github.com/user-attachments/assets/34e8d192-c459-4bf1-b4e9-1b58618c827a

## How Has This Been Tested?

- Verified lint (`eslint`) and type-check (`tsc --noEmit`) pass with zero errors
- Ran all 63 pipeline-related test suites (452 tests) with no regressions
- Added a dedicated unit test that reproduces the exact bug scenario: type a non-matching search → flush debounce → assert toggle remains enabled

## Test Impact

Added new unit test file: `frontend/src/concepts/pipelines/content/pipelineSelector/__tests__/CustomPipelineRunToolbarSelect.spec.tsx`
Tests cover:
- Toggle is enabled when resources exist
- Toggle is disabled when no resources exist (correct behavior preserved)
- Toggle stays enabled after searching for a non-existent name (regression test for this bug)
- Search is cleared when the dropdown closes (verifies the reset-on-close behavior)

## Request review criteria:

If you have UI changes: 

## Summary by CodeRabbit

* **Bug Fixes**
  * Menu toggle now reflects actual available resources (not filtered results).
  * Search input is cleared whenever the menu is closed, preventing stale queries when reopened.

* **Tests**
  * Added tests covering toggle enable/disable states, debounced search behavior, and clearing search state on menu reopen.

### DIFF
--- a/frontend/src/concepts/pipelines/content/pipelineSelector/CustomPipelineRunToolbarSelect.tsx
+++ b/frontend/src/concepts/pipelines/content/pipelineSelector/CustomPipelineRunToolbarSelect.tsx
@@ -144,7 +144,7 @@ const InnerCustomPipelineRunToolbarSelect = <T extends PipelineVersionKF | Exper
           style={{ minWidth: '300px', maxWidth: '500px' }}
           onClick={() => setOpen(!isOpen)}
           isExpanded={isOpen}
-          isDisabled={!filteredResources.length}
+          isDisabled={!resources.length}
           isFullWidth
           data-testid={toggleTestId}
         >
@@ -153,7 +153,12 @@ const InnerCustomPipelineRunToolbarSelect = <T extends PipelineVersionKF | Exper
       }
       menu={menu}
       menuRef={menuRef}
-      onOpenChange={(open) => setOpen(open)}
+      onOpenChange={(open) => {
+        setOpen(open);
+        if (!open) {
+          setSearch('');
+        }
+      }}
     />
   );
 };
--- a/frontend/src/concepts/pipelines/content/pipelineSelector/CustomPipelineRunToolbarSelect.tsx
+++ b/frontend/src/concepts/pipelines/content/pipelineSelector/CustomPipelineRunToolbarSelect.tsx
@@ -156,6 +156,7 @@ const InnerCustomPipelineRunToolbarSelect = <T extends PipelineVersionKF | Exper
       onOpenChange={(open) => {
         setOpen(open);
         if (!open) {
+          doSetSearchDebounced.cancel();
           setSearch('');
         }
       }}
--- a/frontend/src/concepts/pipelines/content/pipelineSelector/__tests__/CustomPipelineRunToolbarSelect.spec.tsx
+++ b/frontend/src/concepts/pipelines/content/pipelineSelector/__tests__/CustomPipelineRunToolbarSelect.spec.tsx
@@ -67,7 +67,7 @@ describe('ExperimentFilterSelector', () => {
 
     // Advance timers to flush the debounced search
     act(() => {
-      jest.advanceTimersByTime(300);
+      jest.advanceTimersByTime(250);
     });
 
     // The toggle should still be enabled even though no results match
@@ -96,7 +96,7 @@ describe('ExperimentFilterSelector', () => {
 
     // Advance timers to flush debounce
     act(() => {
-      jest.advanceTimersByTime(300);
+      jest.advanceTimersByTime(250);
     });
 
     // Close by clicking toggle again
--- a/frontend/src/concepts/pipelines/content/pipelineSelector/__tests__/CustomPipelineRunToolbarSelect.spec.tsx
+++ b/frontend/src/concepts/pipelines/content/pipelineSelector/__tests__/CustomPipelineRunToolbarSelect.spec.tsx
@@ -1,0 +1,113 @@
+import * as React from 'react';
+import { render, screen, act } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import '@testing-library/jest-dom';
+import { ExperimentKF } from '#~/concepts/pipelines/kfTypes';
+import { buildMockExperimentKF } from '#~/__mocks__/mockExperimentKF';
+import { ExperimentFilterSelector } from '#~/concepts/pipelines/content/pipelineSelector/CustomPipelineRunToolbarSelect';
+
+jest.useFakeTimers();
+
+/* eslint-disable camelcase */
+const mockExperiments: ExperimentKF[] = [
+  buildMockExperimentKF({ experiment_id: '1', display_name: 'Experiment A' }),
+  buildMockExperimentKF({ experiment_id: '2', display_name: 'Experiment B' }),
+  buildMockExperimentKF({ experiment_id: '3', display_name: 'Experiment C' }),
+];
+/* eslint-enable camelcase */
+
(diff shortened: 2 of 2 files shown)


frontend/src/concepts/pipelines/content/pipelineSelector/CustomPipelineRunToolbarSelect.tsx
```diff
@@ -153,7 +153,12 @@ const InnerCustomPipelineRunToolbarSelect = <T extends PipelineVersionKF | Exper
       }
       menu={menu}
       menuRef={menuRef}
-      onOpenChange={(open) => setOpen(open)}
+      onOpenChange={(open) => {
+        setOpen(open);
+        if (!open) {
```
nananosirova, line 158:
```suggestion
        if (!open) {
          doSetSearchDebounced.cancel();
          setSearch('');
        }        
```

`onOpenChange` clears with `setSearch('')`, but a pending `doSetSearchDebounced` call is never canceled. If the dropdown closes right after the last keystroke, the delayed callback fires and restores stale search state:

https://github.com/user-attachments/assets/2ef2ceb2-46c5-4d61-84fe-1836a68cdbea

Gkrumbach07:

frontend/src/concepts/pipelines/content/pipelineSelector/__tests__/CustomPipelineRunToolbarSelect.spec.tsx
```diff
@@ -0,0 +1,113 @@
...
+    const searchInput = screen.getByLabelText('Filter pipeline run groups');
+    await user.type(searchInput, 'nonexistent');
+
+    // Advance timers to flush the debounced search
+    act(() => {
+      jest.advanceTimersByTime(300);
+    });
+
+    // The toggle should still be enabled even though no results match
+    expect(toggle).not.toBeDisabled();
```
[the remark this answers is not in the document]
srtanish1992: The finding is valid. Fixed it.

frontend/src/concepts/pipelines/content/pipelineSelector/CustomPipelineRunToolbarSelect.tsx
```diff
@@ -153,7 +153,12 @@ const InnerCustomPipelineRunToolbarSelect = <T extends PipelineVersionKF | Exper
...
+          setSearch('');
+        }
+      }}
```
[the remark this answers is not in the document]
srtanish1992: Fixed.

frontend/src/concepts/pipelines/content/pipelineSelector/CustomPipelineRunToolbarSelect.tsx
srtanish1992: Fixed.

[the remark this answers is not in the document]
srtanish1992: Fixed.

srtanish1992:
@nananosirova I have addressed all comments. Ready for review again.

DaoDaoNoCode (commented):
Cross-codebase pattern review. Prior review comments (debounce race, timer mismatch) are correctly addressed in the second commit. See inline comments for remaining findings.

frontend/src/concepts/pipelines/content/pipelineSelector/CustomPipelineRunToolbarSelect.tsx
```diff
@@ -153,7 +153,13 @@ const InnerCustomPipelineRunToolbarSelect = <T extends PipelineVersionKF | Exper
       }
       menu={menu}
       menuRef={menuRef}
-      onOpenChange={(open) => setOpen(open)}
+      onOpenChange={(open) => {
+        setOpen(open);
+        if (!open) {
+          doSetSearchDebounced.cancel();
+          setSearch('');
```
DaoDaoNoCode, line 160: **Looks good — prior review comments addressed correctly.**

The `doSetSearchDebounced.cancel()` before `setSearch("")` eliminates the race condition where a pending debounced callback could fire after the close-reset and restore stale filter state. The timer in the test now correctly uses 250ms matching the default debounce delay.

One minor observation: `SearchSelector` (used by the sibling `PipelineSelector`) calls `onSearchClear()` unconditionally on every open/close, while this handler only clears on close (`if (!open)`). This PR's approach is actually slightly better — no unnecessary work on open.

frontend/src/concepts/pipelines/content/pipelineSelector/CustomPipelineRunToolbarSelect.tsx
```diff
@@ -144,7 +144,7 @@ const InnerCustomPipelineRunToolbarSelect = <T extends PipelineVersionKF | Exper
           style={{ minWidth: '300px', maxWidth: '500px' }}
           onClick={() => setOpen(!isOpen)}
           isExpanded={isOpen}
-          isDisabled={!filteredResources.length}
+          isDisabled={!resources.length}
```
DaoDaoNoCode, line 147: **Consistent with `PipelineSelector` and `PipelineVersionSelector` patterns.** Both use `totalSize === 0` (unfiltered count) for the disabled condition. Using `resources.length` here achieves the same — the toggle should only disable when there are genuinely no resources, not when a search happens to match nothing.

frontend/src/concepts/pipelines/content/pipelineSelector/__tests__/CustomPipelineRunToolbarSelect.spec.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+    // Reopen
+    await user.click(toggle);
+
+    // Search should be cleared - the helper text should reflect all resources
+    expect(
+      screen.getByText(`Type a name to search your ${mockExperiments.length} run groups.`),
+    ).toBeInTheDocument();
+  });
+});
```
DaoDaoNoCode, line 113: **Test coverage is solid.** Tests cover the exact bug scenario (search-empty-close-reopen cycle) and verify both the disabled condition and the search reset. 

One thing the tests do not cover: what happens when the dropdown is closed via Escape key or clicking outside (both route through `onOpenChange`). The `userEvent.keyboard("{Escape}")` path would be worth adding to confirm the fix works for all close triggers, not just clicking the toggle.

DaoDaoNoCode (approved):
/lgtm

PR #264, opened 2022-07-04
(Backend) Add available GPU number using prometheus queries

maroroman:
Resolves: #158 

Needs to be rewritten to use Dashboard tokens instead of reading secrets in the openshift-monitoring namespace.

This can be done by giving the dashboard SA the cluster-monitoring-view cluster role

PR #7668, opened 2026-05-26
add embedded mlflow UI to evalhub

pnaik1:
https://redhat.atlassian.net/browse/RHOAIENG-62049
## Description
Adds a new Eval Hub compare page that embeds MLflow’s run comparison UI via module federation.
- Adds `CompareEvaluationsPage` and wires route `/:namespace/compare`.
- Adds `evaluationCompareRoute()` helper.
- Introduces `MlflowCompareRuns` wrapper component that lazy-loads `mlflowEmbedded/MlflowCompareRunsWrapper`.
- Parses `experimentIds` and `runUuids` from query params (comma-separated), validates they are present and aligned, and shows an empty-state message when invalid.
- Renders compare UI only in `DeploymentMode.Federated` with valid parameters.

## How Has This Been Tested?
- Manual verification recommended:
  - Navigate to `/:namespace/compare?experimentIds=<id1,id2>&runUuids=<run1,run2>`.
  - Confirm compare view loads in federated deployment mode.
  - Confirm invalid/mismatched query params show empty-state guidance.
  - Confirm breadcrumb navigation back to Evaluations works.
- Automated tests: not added in this branch (UI integration/wiring change only).

## Test Impact
N/A

## Request review criteria:

If you have UI changes: 

## Summary by CodeRabbit

* **New Features**
  * New evaluations "Compare" page and URL for comparing evaluations and runs; accepts comma-separated lists and validates matching counts before enabling the compare UI.
  * Integrated MLflow comparison UI with lazy loading, a loading spinner, and a clear fallback message when the compare interface is unavailable.

### DIFF
--- a/packages/eval-hub/frontend/src/app/components/MlflowCompareRuns.tsx
+++ b/packages/eval-hub/frontend/src/app/components/MlflowCompareRuns.tsx
@@ -1,0 +1,54 @@
+import * as React from 'react';
+import {
+  Bullseye,
+  EmptyState,
+  EmptyStateBody,
+  EmptyStateVariant,
+  Spinner,
+} from '@patternfly/react-core';
+import { ExclamationTriangleIcon } from '@patternfly/react-icons';
+import { loadRemote } from '@module-federation/runtime';
+
+interface MlflowCompareRunsProps {
+  experimentIds: string[];
+  runUuids: string[];
+  workspace?: string;
+}
+
+/** Shown when the federated remote entry cannot be loaded (MF boundary only). */
+const MlflowUnavailable: React.FC = () => (
+  <EmptyState
+    headingLevel="h2"
+    icon={ExclamationTriangleIcon}
+    titleText="MLflow is currently unavailable"
+    variant={EmptyStateVariant.lg}
+    data-testid="mlflow-unavailable-empty-state"
+  >
+    <EmptyStateBody>
+      The MLflow compare UI could not be loaded. Please check that MLflow is deployed and running,
+      then try again.
+    </EmptyStateBody>
+  </EmptyState>
+);
+
+const MlflowCompareRunsRemote = React.lazy(() =>
+  loadRemote<{ default: React.ComponentType<MlflowCompareRunsProps> }>(
+    'mlflowEmbedded/MlflowCompareRunsWrapper',
+  )
+    .then((mod) => mod ?? { default: MlflowUnavailable })
+    .catch(() => ({ default: MlflowUnavailable })),
+);
+
+const MlflowCompareRuns: React.FC<MlflowCompareRunsProps> = (props) => (
+  <React.Suspense
+    fallback={
+      <Bullseye>
+        <Spinner />
+      </Bullseye>
+    }
+  >
+    <MlflowCompareRunsRemote {...props} />
+  </React.Suspense>
+);
+
+export default MlflowCompareRuns;
--- a/packages/eval-hub/frontend/src/app/pages/CompareEvaluationsPage.tsx
+++ b/packages/eval-hub/frontend/src/app/pages/CompareEvaluationsPage.tsx
@@ -1,0 +1,57 @@
+import * as React from 'react';
+import { Breadcrumb, BreadcrumbItem } from '@patternfly/react-core';
+import { Link, useParams, useSearchParams } from 'react-router-dom';
+import ApplicationsPage from '@odh-dashboard/internal/pages/ApplicationsPage';
+import { DeploymentMode, useModularArchContext } from 'mod-arch-core';
+import { evaluationsBaseRoute } from '~/app/routes';
+import MlflowCompareRuns from '~/app/components/MlflowCompareRuns';
+
+const parseIdList = (raw: string): string[] =>
+  raw
+    .split(',')
+    .map((segment) => segment.trim())
+    .filter((segment) => segment.length > 0);
+
+const CompareEvaluationsPage: React.FC = () => {
+  const { namespace } = useParams<{ namespace: string }>();
+  const [searchParams] = useSearchParams();
+
+  const {
+    config: { deploymentMode },
+  } = useModularArchContext();
+
+  const experimentIds = parseIdList(searchParams.get('experimentIds') ?? '');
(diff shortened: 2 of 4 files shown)


packages/eval-hub/frontend/src/app/components/MlflowCompareRuns.tsx
```diff
@@ -0,0 +1,31 @@
...
+  workspace?: string;
+}
+
+const MlflowCompareRunsRemote = React.lazy(() =>
+  loadRemote<{ default: React.ComponentType<MlflowCompareRunsProps> }>(
+    'mlflowEmbedded/MlflowCompareRunsWrapper',
+  )
+    .then((mod) => mod ?? { default: () => null })
+    .catch(() => ({ default: () => null })),
+);
```
ikeola13, line 40: If `loadRemote` fails (e.g. the `MLflow `server is down, the remote entry can't be fetched, or the module doesn't exist), the `.catch` returns a component that renders null. So the `<Spinner>` from Suspense disappears and the user sees... nothing. No error message, no indication that something went wrong.

We can add something like "MLflow is unavailable"

packages/eval-hub/frontend/src/app/pages/CompareEvaluationsPage.tsx
ikeola13, line 1: Can we add unit/mocked tests for this page.

packages/eval-hub/frontend/src/app/pages/CompareEvaluationsPage.tsx
```diff
@@ -0,0 +1,58 @@
...
+    >
+      {showCompare ? (
+        <MlflowCompareRuns
+          key={runUuids.join(',')}
+          experimentIds={experimentIds}
+          runUuids={runUuids}
+          workspace={namespace}
+        />
+      ) : null}
+    </ApplicationsPage>
```
ikeola13, line 54: If user navigates here with valid params but in `standalone` or `kubeflow` mode, the page renders successfully (empty={false}) but with an empty body (`showCompare` is false → renders null). This could be confusing.

we can either pass `!showCompare `as empty and providing a mode-specific message, or hiding the route entirely outside federated mode.

ikeola13 (changes requested):
Great work! 
Just a few tiny thingsa

pnaik1:
Thanks @ikeola13 for the review, some of the checks are already addressed in midstream mlflow repo, I have added a layer for evalhubwrapper

DaoDaoNoCode (commented):
Reviewed the MLflow-side PR (#264 on opendatahub-io/mlflow) as well — the cross-repo contract looks correct (props interface, MF expose key, memory router approach all match). A few comments on the dashboard side below.

packages/eval-hub/frontend/src/app/pages/CompareEvaluationsPage.tsx
```diff
@@ -0,0 +1,57 @@
...
+      title="Compare evaluations"
+      breadcrumb={
+        <Breadcrumb>
+          <BreadcrumbItem
+            render={() => <Link to={evaluationsBaseRoute(namespace)}>Evaluations</Link>}
+          />
+          <BreadcrumbItem isActive>Compare</BreadcrumbItem>
+        </Breadcrumb>
+      }
+      loaded
```
DaoDaoNoCode, line 43: This should be `empty={!showCompare}` instead of `empty={!hasValidParams}`.

As-is, if a user navigates here with valid params but in standalone/kubeflow mode, `empty` is `false` (page renders normally) but `showCompare` is `false` (children render `null`). The user sees a blank page body with no explanation.

With `empty={!showCompare}`, both invalid params AND non-federated mode correctly show the empty state. You'd also want to adjust `emptyMessage` to cover both cases:

```tsx
emptyMessage={
  !hasValidParams
    ? "Provide matching experimentIds and runUuids query parameters to compare runs."
    : "Comparison is only available in federated deployment mode."
}
```

packages/eval-hub/frontend/src/app/pages/CompareEvaluationsPage.tsx
```diff
@@ -0,0 +1,57 @@
...
+      empty={!hasValidParams}
+      emptyMessage="Provide matching experimentIds and runUuids query parameters to compare runs."
+      provideChildrenPadding
+    >
+      {showCompare ? (
+        <MlflowCompareRuns
```
DaoDaoNoCode: nit: Since `empty={!showCompare}` would now gate the children, you can simplify this to just `<MlflowCompareRuns .../>` without the ternary — when `empty` is true, `ApplicationsPage` won't render children anyway.

```tsx
<MlflowCompareRuns
  key={runUuids.join(',')}
  experimentIds={experimentIds}
  runUuids={runUuids}
  workspace={namespace}
/>
```

packages/eval-hub/frontend/src/app/routes.ts
```diff
@@ -22,3 +22,6 @@ export const evaluationStartRoute = (namespace?: string): string =>
 
 export const evaluationResultsRoute = (namespace?: string, jobId?: string): string =>
   `${evaluationsBaseRoute(namespace)}/results/${jobId ?? ':jobId'}`;
+
+export const evaluationCompareRoute = (namespace?: string): string =>
+  `${evaluationsBaseRoute(namespace)}/compare`;
```
DaoDaoNoCode: Consider accepting `experimentIds` and `runUuids` as optional parameters so callers don't have to manually build the query string each time:

```ts
export const evaluationCompareRoute = (
  namespace?: string,
  experimentIds?: string[],
  runUuids?: string[],
): string => {
  const base = `${evaluationsBaseRoute(namespace)}/compare`;
  if (!experimentIds?.length || !runUuids?.length) {
    return base;
  }
  const params = new URLSearchParams({
    experimentIds: experimentIds.join(','),
    runUuids: runUuids.join(','),
  });
  return `${base}?${params.toString()}`;
};
```

This follows the pattern of `evaluationResultsRoute` which embeds its dynamic segment in the path.

ikeola13 (approved):

PR #7680, opened 2026-05-26
Add subscription filtering and linking to API Keys page

Griffin-Sullivan:
https://redhat.atlassian.net/browse/RHOAIENG-62830
## Description

Adds a dropdown for filtering by subscription to the api keys page and also changes the subscription column to link to the subscription details page if the subscription still exists. Currently routes to `/keys-and-subs/subscriptions/view/<sub-name>`. Might need to update that once the details page route has been created
## How Has This Been Tested?
1. Run in mock mode
2. Go to the api keys page
3. See the link to the subscription details (won't actually work yet)
4. Filter by subscription in the dropdown and clear options (also try other filters and sorting)

https://github.com/user-attachments/assets/2c68caf8-280d-45ce-aacc-255eae2a61a7

## Test Impact
Added cypress mock tests to cover the scenarios

## Request review criteria:

If you have UI changes: 

## Summary by CodeRabbit

* **New Features**
  * Added subscription filtering to the API keys table—filter API keys by subscription from the toolbar.
* **UI Changes**
  * Subscription names link to subscription detail pages when available; display as plain text when not.
  * Table rows no longer show a subscription popover; subscription access is presented directly.
* **Tests**
  * Updated test coverage and mocks to validate subscription listing, links, and filtering behavior.

### DIFF
--- a/packages/cypress/cypress/pages/modelsAsAService.ts
+++ b/packages/cypress/cypress/pages/modelsAsAService.ts
@@ -91,6 +91,18 @@ class APIKeysPage {
     return cy.findByRole('menuitem', { name: new RegExp(status, 'i') });
   }
 
+  findSubscriptionFilterToggle(): Cypress.Chainable<JQuery<HTMLElement>> {
+    return cy.findByTestId('api-key-subscription-filter-toggle');
+  }
+
+  findSubscriptionFilterOption(name: string): Cypress.Chainable<JQuery<HTMLElement>> {
+    return cy.findByTestId(`subscription-filter-option-${name}`);
+  }
+
+  findAllSubscriptionsOption(): Cypress.Chainable<JQuery<HTMLElement>> {
+    return cy.findByTestId('subscription-filter-option-all');
+  }
+
   findStatusFilterOptionCheckbox(status: string): Cypress.Chainable<JQuery<HTMLElement>> {
     return this.findStatusFilterOption(status).findByRole('checkbox');
   }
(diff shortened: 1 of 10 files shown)


ashley-o0o:
Looks good! A few questions, do we want it to work similarly to the status? Where you can select multiple subscriptions to filter instead of just one? Also, instead of 'Premium Team Subscription', should it just be 'Premium Team' ? Since that's what the table shows?
Screenshot 2026-06-03 at 1 57 38 PM
Screenshot 2026-06-03 at 1 58 22 PM
Screenshot 2026-06-03 at 2 03 16 PM

Griffin-Sullivan:
@ashley-o0o unfortunately the backend only supports selecting one subscription at a time

ashley-o0o:
Only issue I'm seeing is if you re-select the subscription , it goes to this loading state. It does eventually load, it just takes awhile. Not sure if that's something you're able to fix here or not


https://github.com/user-attachments/assets/73412c3b-d0bc-49c4-94b0-c2ec025c4504

ashley-o0o:
Curious, we don't currently have any keys using 'Different Sub for Testing', so sorting by it results in nothing. But because the subscription exists it still shows up as an option. Would it make sense to omit that option if there are no keys using it? 
Screenshot 2026-06-08 at 9 46 07 AM
Screenshot 2026-06-08 at 9 48 06 AM

Griffin-Sullivan:
```
Curious, we don't currently have any keys using 'Different Sub for Testing', so sorting by it results in nothing. But because the subscription exists it still shows up as an option. Would it make sense to omit that option if there are no keys using it?
```

In an ideal world yeah but then we would need to make an API call for each subscription you have access to. If the user has 100 subscriptions available then we make 100 API calls before the page ever renders cause we would have to call `/v1/api-keys/search` and filter for each sub to see if there's any keys. So I think leaving it as-is is worth the UX cost

PR #7704, opened 2026-05-27
fix(e2e): make generateTestUUID deterministic to prevent orphaned test projects

FedeAlonso:
https://redhat.atlassian.net/browse/RHOAIENG-64772 
## Description
 E2E tests generate a UUID per spec file to namespace OpenShift resources (project names, etc.). The previous implementation used `Math.random()` cached in module scope or `Cypress.env()`.
Both approaches fail when `cy.visitWithLogin()` triggers an OAuth SSO redirect to a different origin — Cypress re-evaluates the entire spec file, resetting module-scope variables and runtime `Cypress.env()` state. This generates a new UUID, creates a second project, and `after()` only cleans up the second one. The first project is orphaned. Build 978 left 44 orphaned projects.

#### Changes

**1. `injectDocumentDomain: true` in `cypress.config.ts`**

Re-enables the Cypress 13 behavior where `document.domain` is injected, preventing the
cross-origin spec re-evaluation entirely. The OAuth redirect is treated as same-origin, so the
spec never re-evaluates and module-scope state is preserved. This fixes the root cause.

**2. Deterministic `generateTestUUID()` (belt and suspenders)**

Replaced random number generation with a deterministic hash of `specFilePath:buildNumber`.
Jenkins provides `BUILD_NUMBER`, GitHub Actions provides `GITHUB_RUN_ID` — both are injected
into Cypress config env at Node load time (`process.env`), which survives cross-origin
re-evaluations unlike runtime `Cypress.env('key', value)` calls. Falls back to `'0'` for
local dev.

No randomness, no caching, no state to lose. Same inputs always produce the same output
regardless of how many times the spec re-evaluates.

## How Has This Been Tested?

dashboard-e2e-tests/991/

image

After the test run, just a couple remaining namespaces:
image

## Test Impact
Different UUID generation

## Request review criteria:

If you have UI changes: 

## Summary by CodeRabbit

* **Chores**
  * Enhanced test identification system to use deterministic identifiers based on test specifications and build information for improved tracking and consistency.
  * Updated test environment configuration to include build tracking capabilities and improve test execution reliability.


sridarna:
No Projects are hanging around in the cluster
image
/lgtm
/approve

PR #7705, opened 2026-05-27
feat(nim-serving): add PVC caching storage field to NIM deployment wizard

YuliaKrimerman:
https://issues.redhat.com/browse/RHOAIENG-60279

## Description

Adds a PVC caching storage wizard field to the NIM deployment wizard. Every NIM deployment requires a PVC to store the NIM container image, and this field allows users to configure that storage during deployment.

The field (`nim-serving/pvcStorage`) appears on the **Model deployment** step when the user selects NVIDIA NIM as the model location type. It supports two modes:

- **New cluster storage** — user provides a PVC name, model path, subpath, storage class, and storage size (GiB). The PVC will be created at deploy time to cache the NIM image.
- **Existing cluster storage** — user selects an existing PVC from a dropdown, with model path and subpath fields.

## How Has This Been Tested?

- `tsc --noEmit` passes on `nim-serving` and full `frontend`
- `eslint` passes with 0 errors on all changed files
- 62/62 unit tests pass in nim-serving
- Manually verified the field appears in the NIM deployment wizard on the Model deployment step when NVIDIA NIM is selected as model location
-
Screenshot 2026-05-28 at 2 54 42 PM
Screenshot 2026-05-28 at 2 57 20 PM

## Test Impact

No unit tests added for the PVC field component in this PR. The component follows the same pattern as `NIMImageField` which has existing tests. Unit tests for the PVC field will be added in a follow-up alongside the `wizard-field-apply` implementation when the field data flows into deployment assembly.

## Request review criteria:

If you have UI changes:

## Summary by CodeRabbit

* **New Features**
  * Added PVC storage configuration to the NIM deployment wizard with options to create new storage or select existing PVCs, storage class selection, size settings, and model path/subpath controls.
  * Introduced reusable wizard field extensions for NIM fields to enable consistent behavior across the wizard.

* **Refactor**
  * Deployment wizard now renders dynamic extension fields alongside standard fields for the model deployment step.

### DIFF
--- a/packages/nim-serving/src/pages/deploymentWizard/fields/NIMPVCField.tsx
+++ b/packages/nim-serving/src/pages/deploymentWizard/fields/NIMPVCField.tsx
@@ -1,0 +1,358 @@
+import React from 'react';
+import { z } from 'zod';
+import {
+  FormGroup,
+  FormSection,
+  HelperText,
+  HelperTextItem,
+  TextInput,
+} from '@patternfly/react-core';
+import SimpleSelect from '@odh-dashboard/internal/components/SimpleSelect';
+import NumberInputWrapper from '@odh-dashboard/internal/components/NumberInputWrapper';
+import type { SimpleSelectOption } from '@odh-dashboard/internal/components/SimpleSelect';
+import type { ProjectSectionType } from '@odh-dashboard/model-serving/components/deploymentWizard/fields/ProjectSection';
+import type { WizardField } from '@odh-dashboard/model-serving/types/form-data';
+import { NIMModelLocationKey } from '@odh-dashboard/model-serving/components/deploymentWizard/fields/modelLocationFields/NIMModelLocation';
+
+export type NIMPVCStorageMode = 'new' | 'existing';
+
+export type NIMPVCFieldValue = {
+  storageMode: NIMPVCStorageMode;
+  pvcName: string;
+  modelPath: string;
+  subPath: string;
+  storageClassName: string;
+  storageSizeGi: number;
+};
+
+const DEFAULT_MODEL_PATH = '/model-store';
+const DEFAULT_SUBPATH = '/';
+const DEFAULT_STORAGE_SIZE_GI = 50;
+const MIN_STORAGE_SIZE_GI = 1;
+
+const nimPVCFieldSchema = z
+  .object({
+    storageMode: z.enum(['new', 'existing']),
+    pvcName: z.string().min(1, 'Cluster storage name is required'),
+    modelPath: z.string().min(1, 'Model path is required'),
+    subPath: z.string(),
+    storageClassName: z.string(),
+    storageSizeGi: z.number().min(MIN_STORAGE_SIZE_GI, 'Storage size must be at least 1 GiB'),
+  })
+  .superRefine((data, ctx) => {
+    if (data.storageMode === 'new' && !data.storageClassName) {
+      ctx.addIssue({
+        code: z.ZodIssueCode.custom,
+        message: 'Storage class is required for new storage',
+        path: ['storageClassName'],
+      });
+    }
+  });
+
+type NIMPVCDependencies = {
+  project: ProjectSectionType;
+};
+
+type StorageClassOption = {
+  name: string;
+  displayName: string;
+};
+
+type NIMPVCExternalData = {
+  storageClasses: StorageClassOption[];
+  existingPVCs: string[];
+};
+
+const useNIMPVCExternalData = (dependencies?: {
+  project?: { projectName?: string };
+}): {
+  data: NIMPVCExternalData;
+  loaded: boolean;
+  loadError?: Error;
+} => {
+  const projectName = dependencies?.project?.projectName;
+  const loaded = !!projectName;
+
+  return React.useMemo(
+    () => ({
+      data: {
+        storageClasses: [{ name: 'gp3-csi', displayName: 'gp3-csi' }],
+        existingPVCs: [],
+      },
+      loaded,
+    }),
+    [loaded],
+  );
+};
+
+const getDefaultFieldValue = (): NIMPVCFieldValue => ({
+  storageMode: 'new',
+  pvcName: '',
+  modelPath: DEFAULT_MODEL_PATH,
+  subPath: DEFAULT_SUBPATH,
+  storageClassName: '',
+  storageSizeGi: DEFAULT_STORAGE_SIZE_GI,
+});
+
+type NIMPVCFieldComponentProps = {
+  value?: NIMPVCFieldValue;
+  onChange: (value: NIMPVCFieldValue) => void;
+  externalData?: { data: NIMPVCExternalData; loaded: boolean; loadError?: Error };
+  isDisabled?: boolean;
+};
+
+const NIMPVCFieldComponent: React.FC<NIMPVCFieldComponentProps> = ({
+  value,
+  onChange,
+  externalData,
+  isDisabled,
+}) => {
+  const fieldValue = React.useMemo(() => value ?? getDefaultFieldValue(), [value]);
+
+  const updateField = React.useCallback(
+    (patch: Partial<NIMPVCFieldValue>) => {
+      onChange({ ...fieldValue, ...patch });
+    },
+    [fieldValue, onChange],
+  );
+
+  const storageClasses = externalData?.data.storageClasses ?? [];
+  const existingPVCs = externalData?.data.existingPVCs ?? [];
+
+  const storageModeOptions: SimpleSelectOption[] = [
+    {
+      key: 'new',
+      label: 'Deploy the NIM image from a new cluster storage',
+    },
+    {
+      key: 'existing',
+      label: 'Deploy the NIM image from an existing cluster storage',
+    },
+  ];
+
+  const storageClassOptions: SimpleSelectOption[] = storageClasses.map((sc) => ({
+    key: sc.name,
+    label: sc.displayName,
+  }));
+
+  const existingPVCOptions: SimpleSelectOption[] = existingPVCs.map((pvc) => ({
+    key: pvc,
+    label: pvc,
+  }));
+
+  return (
+    <FormSection title="Storage and deployment option">
+      <FormGroup label="Storage and deployment option" fieldId="nim-storage-mode" isRequired>
+        <SimpleSelect
+          dataTestId="nim-storage-mode-select"
+          options={storageModeOptions}
+          value={fieldValue.storageMode}
+          onChange={(val) => {
+            const mode: NIMPVCStorageMode = val === 'existing' ? 'existing' : 'new';
+            updateField({
+              storageMode: mode,
+              pvcName: '',
+              storageClassName: mode === 'new' ? storageClasses[0]?.name ?? '' : '',
+            });
+          }}
+          isDisabled={isDisabled}
+          isFullWidth
+        />
+      </FormGroup>
+
+      {fieldValue.storageMode === 'new' ? (
+        <>
+          <FormGroup label="Cluster storage name" fieldId="nim-pvc-name" isRequired>
+            <TextInput
+              id="nim-pvc-name"
+              data-testid="nim-pvc-name-input"
+              value={fieldValue.pvcName}
+              onChange={(_event, val) => updateField({ pvcName: val })}
+              placeholder="nim-pvc"
+              isDisabled={isDisabled}
+            />
+            <HelperText>
+              <HelperTextItem>
+                This cluster storage can be reused for future deployments of this NIM image.
+              </HelperTextItem>
+            </HelperText>
+          </FormGroup>
+
+          <FormGroup label="Model path" fieldId="nim-model-path" isRequired>
+            <TextInput
+              id="nim-model-path"
+              data-testid="nim-model-path-input"
+              value={fieldValue.modelPath}
+              onChange={(_event, val) => updateField({ modelPath: val })}
+              placeholder={DEFAULT_MODEL_PATH}
+              isDisabled={isDisabled}
+            />
+            <HelperText>
+              <HelperTextItem>
+                Path within the container where model files will be mounted.
+              </HelperTextItem>
+            </HelperText>
+          </FormGroup>
+
+          <FormGroup label="Subpath" fieldId="nim-subpath">
+            <TextInput
+              id="nim-subpath"
+              data-testid="nim-subpath-input"
+              value={fieldValue.subPath}
+              onChange={(_event, val) => updateField({ subPath: val })}
+              placeholder="/"
+              isDisabled={isDisabled}
+            />
+            <HelperText>
+              <HelperTextItem>
+                Optional: Subdirectory within the PVC. Use this if you have multiple models stored
+                in the same PVC. Leave blank to use the root of the PVC.
+              </HelperTextItem>
+            </HelperText>
+          </FormGroup>
+
+          <FormGroup label="Storage class" fieldId="nim-storage-class">
+            <SimpleSelect
+              dataTestId="nim-storage-class-select"
+              options={storageClassOptions}
+              value={fieldValue.storageClassName}
+              onChange={(val) => updateField({ storageClassName: val })}
+              isDisabled={isDisabled}
+              isFullWidth
+              placeholder="Select storage class"
+            />
+          </FormGroup>
+
+          <FormGroup label="NVIDIA NIM storage size" fieldId="nim-storage-size" isRequired>
+            <NumberInputWrapper
+              id="nim-storage-size"
+              data-testid="nim-storage-size-input"
+              value={fieldValue.storageSizeGi}
+              min={MIN_STORAGE_SIZE_GI}
+              onChange={(val) =>
+                updateField({
+                  storageSizeGi: Math.max(MIN_STORAGE_SIZE_GI, val ?? DEFAULT_STORAGE_SIZE_GI),
+                })
+              }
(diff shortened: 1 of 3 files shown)


packages/nim-serving/src/pages/deploymentWizard/fields/NIMPVCField.tsx
```diff
@@ -0,0 +1,358 @@
...
+  loaded: boolean;
+  loadError?: Error;
+} => {
+  const projectName = dependencies?.project?.projectName;
+  const loaded = !!projectName;
+
+  return React.useMemo(
+    () => ({
+      data: {
+        storageClasses: [{ name: 'gp3-csi', displayName: 'gp3-csi' }],
```
emilys314: I believe there is already a react hook that will get the storage classes and/or the default one. We can preselect the default one

packages/nim-serving/src/pages/deploymentWizard/fields/NIMPVCField.tsx
```diff
@@ -0,0 +1,358 @@
...
+        existingPVCs: [],
```
emilys314: We'll need to fetch the PVCs (all of them in the project) and list NIM ones first

emilys314 (commented):
The code seems good so far, there aren't screenshots yet so i can't tell how it looks

emilys314:
cc @vconzola check this out when your back

packages/nim-serving/src/pages/deploymentWizard/fields/NIMPVCField.tsx
```diff
@@ -0,0 +1,470 @@
...
+    if (
+      fieldValue.storageMode === NIMPVCStorageMode.NEW &&
+      !fieldValue.storageClassName &&
+      defaultStorageClassName
+    ) {
+      updateField({ storageClassName: defaultStorageClassName });
+    }
+    // Only re-run when these specific values change, not on every fieldValue object reference change
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [defaultStorageClassName, fieldValue.storageMode, fieldValue.storageClassName]);
```
emilys314: We should really avoid useEffect inside the component, they should be pure

packages/nim-serving/src/pages/deploymentWizard/fields/NIMPVCField.tsx
```diff
@@ -0,0 +1,470 @@
...
+  const [defaultStorageClassName, setDefaultStorageClassName] = React.useState('');
+  const [existingPVCs, setExistingPVCs] = React.useState<string[]>([]);
+  const [loaded, setLoaded] = React.useState(false);
+  const [loadError, setLoadError] = React.useState<Error>();
+
+  React.useEffect(() => {
+    if (!projectName) {
+      setLoaded(false);
+      return;
+    }
```
emilys314: We should re-use the `useFetch` hook because it has a lot of the boiler plate already made for us, with cancelled promises, loading, and errors

packages/nim-serving/src/pages/deploymentWizard/fields/NIMPVCField.tsx
```diff
@@ -0,0 +1,470 @@
...
+        if (cancelled) {
+          return;
+        }
+
+        const scOptions = scList.map((sc) => ({
+          name: sc.metadata.name,
+          displayName: sc.metadata.annotations?.['openshift.io/display-name'] || sc.metadata.name,
+        }));
+
+        const defaultSc = scList.find(isDefaultStorageClass);
```
emilys314: The logic is more complicated to figure out the default. The dashboard can have a different default storage class than the cluster. see https://github.com/opendatahub-io/odh-dashboard/blob/main/frontend/src/pages/projects/screens/spawner/storage/useDefaultStorageClass.ts

mturley:
@YuliaKrimerman if you select existing cluster storage and there is no cluster storage available to select, you get a disabled dropdown with “Select…” in it.

image

Maybe we should replicate the behavior the “model location” dropdown has when there is no cluster storage (just hides the cluster storage option entirely), which would mean in that case we just hide the NIM-specific new/existing storage dropdown entirely too and assume they need a new cluster storage? we do that with connections when there are none existing (no existing connection option, assume new connection).

cc @vconzola

packages/nim-serving/src/pages/deploymentWizard/fields/NIMPVCField.tsx
```diff
@@ -0,0 +1,470 @@
...
+  }));
+
+  const existingPVCOptions: SimpleSelectOption[] = existingPVCs.map((pvc) => ({
+    key: pvc,
+    label: pvc,
+  }));
+
+  return (
+    <FormSection title="Storage and deployment option">
+      <FormGroup label="Storage and deployment option" fieldId="nim-storage-mode" isRequired>
```
mturley: The `FormSection` title and the first `FormGroup` label are both `"Storage and deployment option"`, which renders the same text twice, once as a section heading and once as a field label.

image

It looks like in the protoype we don’t have a form section title at all here, just the field label. Also, the prototype has the subfields that appear when choosing an option here indented:

image

packages/nim-serving/src/pages/deploymentWizard/fields/NIMPVCField.tsx
```diff
@@ -0,0 +1,470 @@
...
+  </>
+);
+
+const NIMPVCFieldComponent: React.FC<NIMPVCFieldComponentProps> = ({
+  value,
+  onChange,
+  externalData,
+  isDisabled,
+}) => {
+  const fieldValue = React.useMemo(() => value ?? getDefaultFieldValue(), [value]);
```
mturley, line 236: Nit: Unlike `NIMImageField` which renders a `<Skeleton>` while external data is loading, this component renders normally with empty dropdowns during the loading state. We should add a similar loading skeleton when `!externalData?.loaded` so the user knows data is being fetched.

packages/nim-serving/src/pages/deploymentWizard/fields/NIMPVCField.tsx
```diff
@@ -0,0 +1,470 @@
...
+const DEFAULT_MODEL_PATH = '/model-store';
+const DEFAULT_SUBPATH = '/';
+const DEFAULT_STORAGE_SIZE_GI = 50;
+const MIN_STORAGE_SIZE_GI = 1;
+
+const nimPVCFieldSchema = z
+  .object({
+    storageMode: z.nativeEnum(NIMPVCStorageMode),
+    pvcName: z.string().min(1, 'Cluster storage name is required'),
+    modelPath: z.string().min(1, 'Model path is required'),
```
mturley, line 50: Out of scope for this PR, but worth noting for the follow-up: the `pvcName` schema only validates `min(1)`, but PVC names must conform to RFC 1123 DNS label rules. The wizard handles this for the deployment name via `K8sNameDescriptionField`. When the `wizard-field-apply` implementation lands, the PVC name input may need similar treatment (either using `K8sNameDescriptionField` or its validation utils).

packages/nim-serving/src/pages/deploymentWizard/fields/NIMPVCField.tsx
```diff
@@ -0,0 +1,470 @@
...
+        </HelperTextItem>
+      </HelperText>
+    </FormGroup>
+  </>
+);
+
+const NIMPVCFieldComponent: React.FC<NIMPVCFieldComponentProps> = ({
+  value,
+  onChange,
+  externalData,
```
mturley, line 233: The component never surfaces `externalData.loadError` to the user. If `getStorageClasses()` or the PVC list fetch fails, the user sees empty dropdowns with no explanation.

Compare with `NIMImageField` which renders a `HelperText` error message when `externalData.loadError` is truthy (lines 201-207 of NIMImageField.tsx). We should add similar error feedback here.

mturley (changes requested):
In addition to my comment above about the disabled field when there are no cluster storages, I found a couple other things.

vconzola:
@YuliaKrimerman I agree with all of @mturley 's comments and suggestions. The only other thing I see is that the Storage class field should be required, I believe.

packages/nim-serving/src/pages/deploymentWizard/fields/NIMPVCField.tsx
emilys314: Yeah we'll need to do this before release

packages/nim-serving/src/pages/deploymentWizard/fields/NIMPVCField.tsx
```diff
@@ -0,0 +1,471 @@
...
+  displayName: string;
+};
+
+type NIMPVCExternalData = {
+  storageClasses: StorageClassOption[];
+  defaultStorageClassName: string;
+  existingPVCs: string[];
+};
+
+const isNIMPVC = (pvcName: string): boolean => pvcName.startsWith('nim-pvc');
```
emilys314: This will need to be filtering on an annotation https://redhat-internal.slack.com/archives/C069KSM8T9N/p1778786740448229?thread_ts=1778531481.744819&cid=C069KSM8T9N

We don't have a specific annotation defined yet for NIM pvcs.. how about `dashboard.opendatahub.io/nim-pvc`?

packages/nim-serving/src/pages/deploymentWizard/fields/NIMPVCField.tsx
```diff
@@ -0,0 +1,471 @@
...
+  const otherPVCs: string[] = [];
+
+  for (const pvc of pvcs) {
+    const { name } = pvc.metadata;
+    if (isNIMPVC(name)) {
+      nimPVCs.push(name);
+    } else {
+      otherPVCs.push(name);
+    }
+  }
```
emilys314, line 95: we should put model serving pvcs last, general storage in the middle.

so 
1. NIM pvcs, with the `dashboard.opendatahub.io/nim-pvc`(?) annotation
2. general purpose pvcs, (I think this is just without the model-path or the nim annotation)
3. model storage pvc with a `dashboard.opendatahub.io/model-path` annotation

emilys314: @vconzola should we put some kind of label showing which one is a nim cluster storage vs general purpose vs model serving ones?

This is what we have right now:
Image
We could do something like the hardware profile thing on the right

Image

packages/nim-serving/src/pages/deploymentWizard/fields/NIMPVCField.tsx
```diff
@@ -0,0 +1,471 @@
...
+  }, [projectName]);
+
+  const {
+    data: fetchedData,
+    loaded: fetchLoaded,
+    error: fetchError,
+  } = useFetch(fetchCallback, DEFAULT_FETCHED_DATA);
+
+  const loaded = fetchLoaded && defaultSCLoaded;
+  const loadError = fetchError ?? defaultSCError;
```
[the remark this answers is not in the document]
mturley: This is technically valid but would add complexity that doesn’t seem worth it. getStorageClasses is simple/fast, and if it fails we have bigger problems. It’s probably reasonable to expect that if one of these succeeds they both will.

packages/nim-serving/src/pages/deploymentWizard/fields/NIMPVCField.tsx
```diff
@@ -0,0 +1,471 @@
...
+  const [defaultStorageClass, defaultSCLoaded, defaultSCError] = useDefaultStorageClass();
+
+  const fetchCallback = React.useCallback<
+    FetchStateCallbackPromise<FetchedStorageData>
+  >(async () => {
+    if (!projectName) {
+      return Promise.reject(new NotReadyError('No project selected'));
+    }
+    const [scList, pvcList] = await Promise.all([
+      getStorageClasses(),
```
emilys314, line 125: I did notice that `useDefaultStorageClass()` is already calling `getStorageClasseses()`. This hook will be making a duplicate api call. I think this is an existing gap in where useDefaultStorageClass is being used. But adding as-is does introduce tech debt for api call efficiency.

Do you have any ideas on how to improve this?

packages/nim-serving/src/pages/deploymentWizard/fields/NIMPVCField.tsx
```diff
@@ -0,0 +1,471 @@
...
+    () => ({
+      data: {
+        storageClasses: fetchedData.storageClasses,
+        defaultStorageClassName: defaultStorageClass?.metadata.name ?? '',
+        existingPVCs: fetchedData.existingPVCs,
+      },
+      loaded,
+      loadError,
+    }),
+    [fetchedData, defaultStorageClass, loaded, loadError],
```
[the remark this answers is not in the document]
mturley: I think this is invalid. useDefaultStorageClass uses useStorageClasses internally, so they are the same data source, and if somehow they don’t match, the form validation will catch this.

packages/nim-serving/src/pages/deploymentWizard/fields/NIMPVCField.tsx
```diff
@@ -0,0 +1,471 @@
...
+  >(async () => {
+    if (!projectName) {
+      return Promise.reject(new NotReadyError('No project selected'));
+    }
+    const [scList, pvcList] = await Promise.all([
+      getStorageClasses(),
+      k8sListResourceItems<PersistentVolumeClaimKind>({
+        model: PVCModel,
+        queryOptions: { ns: projectName },
+      }),
```
emilys314: Hm we should probably filter on PVCs with the dashboard label?

packages/nim-serving/src/pages/deploymentWizard/fields/NIMPVCField.tsx
mturley: Opened a followup issue in the epic: https://redhat.atlassian.net/browse/RHOAIENG-65577

packages/nim-serving/src/pages/deploymentWizard/fields/NIMPVCField.tsx
emilys314: Also we could add some metadata of the PVC in a description but that is scope creep ...

emilys314 (changes requested):

packages/nim-serving/src/pages/deploymentWizard/fields/NIMPVCField.tsx
mturley: I think if it’s cool we should maybe do this labeling in a followup PR so we can unblock https://github.com/opendatahub-io/odh-dashboard/pull/7752 and https://github.com/opendatahub-io/odh-dashboard/pull/7764

mturley: unless it’s easy, since there’s other feedback too

packages/nim-serving/src/pages/deploymentWizard/fields/NIMPVCField.tsx
mturley: Seems like deduplicating this call would require some complexity and maintenance overhead, doing some kind of caching in context or returning the getStorageClasses result from useDefaultStorageClasses for reuse. Maybe worth it but this is a lightweight cluster-scoped fetch.

It looks like storage classes are actually already cached in AppContext, but that’s frontend-internal and we’d need to plumb it through external data to get at it from the nim-serving package. seems like maybe that’s the way to go but feels like it would be ok in a followup

emilys314: I also saw we do have `@tanstack/react-query` being used in some of the other packages. It would be good to investigate this at some point. But yeah this is fine as is but good to be aware of

vconzola:
> @vconzola should we put some kind of label showing which one is a nim cluster storage vs general purpose vs model serving ones?

@emilys314 @mturley That would be great!

packages/nim-serving/src/pages/deploymentWizard/fields/NIMPVCField.tsx
YuliaKrimerman: Will be addressed in a separate issue

packages/nim-serving/src/pages/deploymentWizard/fields/NIMPVCField.tsx
YuliaKrimerman: Keeping this as is

packages/nim-serving/src/pages/deploymentWizard/fields/NIMPVCField.tsx
mturley: Followup issue, including a little bit of investigation: https://redhat.atlassian.net/browse/RHOAIENG-65755

mturley:
/lgtm

Tested by creating some mock PVCs in my namespace (based on the ones in the `emily` namespace in the zaffre ROSA cluster, but I added the `dashboard.opendatahub.io/nim-pvc=true` annotation to the mock-nim-pvc).

image

@YuliaKrimerman the subfields related to PVC are still not indented though, to make them visually organized:

image

compare with the prototype:

image

If it’s ok, I’ll add that as part of the followup issue https://redhat.atlassian.net/browse/RHOAIENG-65755 and make it more of a “cosmetic followups to PVC field” catch-all

mturley:
/approve

@emilys314 since you're out and it seems your comments are addressed, approving this to unblock some other PRs... apologies if you end up with more comments

PR #7729, opened 2026-05-28
RHOAIENG-63152: Add dev feature flag and Roles tab with routing

srtanish1992:
https://redhat.atlassian.net/browse/RHOAIENG-63152

## Description

Adds scaffolding for a project-level "Roles" tab behind the `roleManagement` temporary developer feature flag ([RHOAIENG-63152](https://redhat.atlassian.net/browse/RHOAIENG-63152)). This is preparatory work for fine-grained role management within projects.

**What's included:**
- New `roleManagement` feature flag in `devTemporaryFeatureFlags` (not in CRD — hidden from users until dev is complete)
- New `ROLE_MANAGEMENT` supported area with feature flag mapping
- "Roles" tab in project details, gated by:
  1. Feature flag enabled (`roleManagement: true`)
  2. SSAR check — user must have `list` permission on `roles` in `rbac.authorization.k8s.io`
- Tab displays "Dev preview" label (yellow outline, matching existing label styling)
- Placeholder `ProjectRoles` empty state with a "Create role" button
- Placeholder `CreateRolePage` with breadcrumb navigation
- Route for `/projects/:namespace/roles/create` (conditional on feature flag)

**Visibility:** The tab is invisible by default. To enable locally, set `roleManagement: true` on the `OdhDashboardConfig` CR. A follow-up ticket ([RHOAIENG-63227](https://redhat.atlassian.net/browse/RHOAIENG-63227)) tracks promotion to Tech Preview.

## How Has This Been Tested?

- Verified the tab does **not** appear when the feature flag is disabled (default)
- Verified the tab appears with "Dev preview" label when `roleManagement: true` is set on the CR
- Verified the loading spinner does not hang when the flag is disabled (conditional SSAR logic)
- Verified the "Create role" button navigates to the create page with correct breadcrumbs
- Ran `tsc --noEmit` on frontend and backend — no type errors
- Ran ESLint on all changed/new files — no lint errors
- Ran unit tests (14 suites, 152 tests pass) including new `isAreaAvailable` tests for `ROLE_MANAGEMENT`

## Test Impact

- **Unit tests:** Added 3 test cases in `frontend/src/concepts/areas/__tests__/utils.spec.ts` verifying `ROLE_MANAGEMENT` area availability based on feature flag state
- **Cypress mock tests:** Added `packages/cypress/cypress/tests/mocked/projects/tabs/rolesTab.cy.ts` covering:
  - Tab hidden when `roleManagement` flag is disabled
  - Tab visible with "Create role" button when flag is enabled (admin)
  - Tab visible for non-admin users with appropriate RBAC permissions

## Request review criteria:

If you have UI changes: 

## Summary by CodeRabbit

* **New Features**
  * Added role management to projects (feature-flag gated).
  * Introduced a Roles tab in project details with a "Create role" page (access-controlled; placeholder UI).
  * Routing updated so role creation routes are available only when permitted.

* **Tests**
  * Unit tests for area availability with roleManagement flag.
  * End-to-end tests covering feature-flag gating and create-role permission flows.

### DIFF
--- a/frontend/src/pages/projects/ProjectViewRoutes.tsx
+++ b/frontend/src/pages/projects/ProjectViewRoutes.tsx
@@ -29,7 +29,11 @@ const ProjectViewRoutes: React.FC = () => {
         <Route path="spawner/:notebookName" element={<EditSpawnerPage />} />
         <Route path="permissions" element={<Navigate to="..?section=permissions" replace />} />
         <Route path="permissions/assign" element={<ProjectPermissionsAssignRoles />} />
-        {roleManagementEnabled && <Route path="roles/create" element={<CreateRolePage />} />}
+        {roleManagementEnabled ? (
+          <Route path="roles/create" element={<CreateRolePage />} />
+        ) : (
+          <Route path="roles/*" element={<Navigate to=".." replace />} />
+        )}
         {modelMetricsEnabled && (
           <>
             <Route path="metrics/model" element={<ProjectInferenceExplainabilityWrapper />}>
--- a/frontend/src/pages/projects/ProjectViewRoutes.tsx
+++ b/frontend/src/pages/projects/ProjectViewRoutes.tsx
@@ -13,10 +13,12 @@ import ProjectDetailsContextProvider from './ProjectDetailsContext';
 import SpawnerPage from './screens/spawner/SpawnerPage';
 import EditSpawnerPage from './screens/spawner/EditSpawnerPage';
 import ProjectPermissionsAssignRoles from './projectPermissions/ProjectPermissionsAssignRoles';
+import CreateRolePage from './projectRoles/CreateRolePage';
 
 const ProjectViewRoutes: React.FC = () => {
   const [modelMetricsEnabled] = useModelMetricsEnabled();
   const biasMetricsAreaAvailable = useIsAreaAvailable(SupportedArea.BIAS_METRICS).status;
+  const roleManagementEnabled = useIsAreaAvailable(SupportedArea.ROLE_MANAGEMENT).status;
 
   return (
     <ProjectsRoutes>
@@ -27,6 +29,7 @@ const ProjectViewRoutes: React.FC = () => {
         <Route path="spawner/:notebookName" element={<EditSpawnerPage />} />
         <Route path="permissions" element={<Navigate to="..?section=permissions" replace />} />
         <Route path="permissions/assign" element={<ProjectPermissionsAssignRoles />} />
+        {roleManagementEnabled && <Route path="roles/create" element={<CreateRolePage />} />}
         {modelMetricsEnabled && (
           <>
             <Route path="metrics/model" element={<ProjectInferenceExplainabilityWrapper />}>
--- a/frontend/src/pages/projects/components/GenericHorizontalBar.tsx
+++ b/frontend/src/pages/projects/components/GenericHorizontalBar.tsx
@@ -57,7 +57,11 @@ const GenericHorizontalBar: React.FC<GenericHorizontalBarProps> = ({
                 <>
                   {section.icon && <TabTitleIcon>{section.icon}</TabTitleIcon>}
                   <TabTitleText>{section.title}</TabTitleText>
-                  {section.label && <Label isCompact>{section.label}</Label>}
+                  {section.label && (
+                    <Label isCompact color="yellow" variant="outline">
+                      {section.label}
+                    </Label>
+                  )}
                 </>
               }
             />
--- a/frontend/src/pages/projects/components/GenericHorizontalBar.tsx
+++ b/frontend/src/pages/projects/components/GenericHorizontalBar.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { Tabs, Tab, TabTitleIcon, TabTitleText, PageSection, Label } from '@patternfly/react-core';
+import { Tabs, Tab, TabTitleIcon, TabTitleText, PageSection } from '@patternfly/react-core';
 import { SVGIconProps } from '@patternfly/react-icons/dist/esm/createIcon';
 
 export type SectionDefinition = {
@@ -8,7 +8,7 @@ export type SectionDefinition = {
   icon?: React.ReactElement<React.ComponentClass<SVGIconProps>>;
   id: string;
   actions?: React.ReactNode[];
-  label?: string;
+  label?: React.ReactNode;
 };
 
 type GenericHorizontalBarProps = {
@@ -57,11 +57,7 @@ const GenericHorizontalBar: React.FC<GenericHorizontalBarProps> = ({
                 <>
                   {section.icon && <TabTitleIcon>{section.icon}</TabTitleIcon>}
                   <TabTitleText>{section.title}</TabTitleText>
-                  {section.label && (
-                    <Label isCompact color="yellow" variant="outline">
-                      {section.label}
-                    </Label>
-                  )}
+                  {section.label}
                 </>
               }
             />
--- a/frontend/src/pages/projects/projectRoles/CreateRolePage.tsx
+++ b/frontend/src/pages/projects/projectRoles/CreateRolePage.tsx
@@ -2,15 +2,37 @@ import * as React from 'react';
 import {
   Breadcrumb,
   BreadcrumbItem,
+  Bullseye,
   PageSection,
   EmptyState,
   EmptyStateBody,
+  Spinner,
 } from '@patternfly/react-core';
-import { Link, useParams } from 'react-router-dom';
+import { Link, Navigate, useParams } from 'react-router-dom';
 import ApplicationsPage from '#~/pages/ApplicationsPage';
+import { useAccessReview } from '#~/api/useAccessReview';
 
 const CreateRolePage: React.FC = () => {
   const { namespace = '' } = useParams<{ namespace: string }>();
+
+  const [allowCreate, loaded] = useAccessReview({
+    group: 'rbac.authorization.k8s.io',
+    resource: 'roles',
+    namespace,
+    verb: 'create',
+  });
+
+  if (!loaded) {
+    return (
+      <Bullseye>
+        <Spinner />
+      </Bullseye>
+    );
+  }
+
+  if (!allowCreate) {
+    return <Navigate to={`/projects/${namespace}?section=roles`} replace />;
+  }
 
   return (
     <ApplicationsPage
--- a/frontend/src/pages/projects/projectRoles/CreateRolePage.tsx
+++ b/frontend/src/pages/projects/projectRoles/CreateRolePage.tsx
@@ -1,0 +1,39 @@
+import * as React from 'react';
+import {
+  Breadcrumb,
+  BreadcrumbItem,
+  PageSection,
+  EmptyState,
+  EmptyStateBody,
+} from '@patternfly/react-core';
+import { Link, useParams } from 'react-router-dom';
+import ApplicationsPage from '#~/pages/ApplicationsPage';
+
+const CreateRolePage: React.FC = () => {
+  const { namespace = '' } = useParams<{ namespace: string }>();
+
+  return (
+    <ApplicationsPage
+      title="Create role"
+      breadcrumb={
+        <Breadcrumb>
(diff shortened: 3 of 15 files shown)


frontend/src/pages/projects/projectRoles/CreateRolePage.tsx
```diff
@@ -0,0 +1,39 @@
...
+  Breadcrumb,
+  BreadcrumbItem,
+  PageSection,
+  EmptyState,
+  EmptyStateBody,
+} from '@patternfly/react-core';
+import { Link, useParams } from 'react-router-dom';
+import ApplicationsPage from '#~/pages/ApplicationsPage';
+
+const CreateRolePage: React.FC = () => {
```
DaoDaoNoCode, line 17: This page is reachable via deep link (`/projects/:namespace/roles/create`) but has no RBAC guard. The codebase has a clear pattern for this -- look at `ProjectPermissionsAssignRoles` (lines 218-246 in `projectPermissions/ProjectPermissionsAssignRoles.tsx`), which wraps the form in an RBAC check and redirects unauthorized users with `<Navigate>`.

You'll want to do the same here: call `useProjectRolesTabVisible` (or a more specific `useAccessReview` for role creation), show a spinner while loading, and redirect if denied.

```suggestion
const CreateRolePage: React.FC = () => {
  const { namespace = '' } = useParams<{ namespace: string }>();
  const [allowRoles, rolesLoaded] = useProjectRolesTabVisible(namespace, !!namespace);

  if (!rolesLoaded) {
    return null;
  }

  if (!allowRoles) {
    return <Navigate to={`/projects/${namespace}`} replace />;
  }

  return (
```

You'll need to add `Navigate` to the react-router-dom import and `useProjectRolesTabVisible` from `#~/concepts/projects/accessChecks`.

frontend/src/pages/projects/projectRoles/CreateRolePage.tsx
```diff
@@ -0,0 +1,39 @@
...
+  const { namespace = '' } = useParams<{ namespace: string }>();
+
+  return (
+    <ApplicationsPage
+      title="Create role"
+      breadcrumb={
+        <Breadcrumb>
+          <BreadcrumbItem render={() => <Link to="/projects">Projects</Link>} />
+          <BreadcrumbItem
+            render={() => <Link to={`/projects/${namespace}?section=roles`}>{namespace}</Link>}
```
DaoDaoNoCode: nit: the breadcrumb shows the raw `namespace` string, but the existing pattern in `ProjectPermissionsAssignRoles` uses `getDisplayNameFromK8sResource(currentProject)` to show the human-readable project name. Since this page is rendered inside `ProjectDetailsContextProvider`, you have access to `ProjectDetailsContext` and can pull the display name from there.

```suggestion
            render={() => (
              <Link to={`/projects/${namespace}?section=roles`}>
                {displayName}
              </Link>
            )}
```

(Where `displayName` comes from `getDisplayNameFromK8sResource(currentProject)` via `ProjectDetailsContext`.)

frontend/src/pages/projects/components/GenericHorizontalBar.tsx
```diff
@@ -57,7 +57,11 @@ const GenericHorizontalBar: React.FC<GenericHorizontalBarProps> = ({
                 <>
                   {section.icon && <TabTitleIcon>{section.icon}</TabTitleIcon>}
                   <TabTitleText>{section.title}</TabTitleText>
-                  {section.label && <Label isCompact>{section.label}</Label>}
+                  {section.label && (
+                    <Label isCompact color="yellow" variant="outline">
```
DaoDaoNoCode: This hardcodes `color="yellow"` and `variant="outline"` for every label in `GenericHorizontalBar`. Right now only the Roles tab uses `label`, so it works fine. But if another tab later needs a different label style (e.g., "Tech preview" in blue, or "Beta" in green), you'd have to change this shared component again.

Consider making the label accept a `ReactNode` instead of a `string` so callers control the styling, or add optional `labelColor` / `labelVariant` fields to `SectionDefinition`. That said, if the team has decided all tab labels will always be yellow-outline, this is fine as-is -- just flagging it.

packages/cypress/cypress/tests/mocked/projects/tabs/rolesTab.cy.ts
```diff
@@ -0,0 +1,73 @@
...
+};
+
+describe('Roles tab feature flag gating', () => {
+  describe('with roleManagement flag disabled (default)', () => {
+    beforeEach(() => {
+      asProjectAdminUser();
+      initIntercepts({ roleManagement: false });
+    });
+
+    it('should not show the Roles tab', () => {
```
DaoDaoNoCode, line 40: The project's mock test conventions require all UI interactions to go through page objects, not direct `cy.findByTestId()` / `cy.findByRole()` / `cy.visit()` calls in tests. Every other tab test in this directory (`permissions.cy.ts`, `clusterStorage.cy.ts`, etc.) uses page objects with `.visit()` methods.

You'll want to create a small page object (or extend an existing one) with methods like `visit(namespace)`, `findRolesTab()`, `findCreateRoleButton()`, and use those here instead. Something like:

```ts
// pages/projectRoles.ts
class ProjectRolesPage {
  visit(namespace: string) {
    cy.visitWithLogin(`/projects/${namespace}?section=roles`);
    cy.testA11y();
  }
  findRolesTab() {
    return cy.findByTestId('project-roles-tab');
  }
  findCreateRoleButton() {
    return cy.findByTestId('create-role-button');
  }
}
export const projectRoles = new ProjectRolesPage();
```

Also note: the tests should include `cy.testA11y()` after page load for accessibility coverage, and use `cy.visitWithLogin()` instead of `cy.visit()` to match the mock test patterns.

packages/cypress/cypress/tests/mocked/projects/tabs/rolesTab.cy.ts
```diff
@@ -0,0 +1,73 @@
...
+    });
+  });
+
+  describe('with roleManagement flag enabled but non-admin user', () => {
+    beforeEach(() => {
+      asProjectEditUser();
+      initIntercepts({ roleManagement: true });
+    });
+
+    it('should show the Roles tab for users with list access to roles', () => {
```
DaoDaoNoCode: This test scenario says it validates "users with list access to roles" but it doesn't explicitly control the SSAR response. It relies on `asProjectEditUser()` defaulting to allow for the `roles` resource (which happens because `roles` isn't in the `PROJECT_ADMIN_RESOURCES` list in `mockUsers.ts`).

The test passes today, but it's fragile -- if someone adds `roles` to `PROJECT_ADMIN_RESOURCES`, this test would break silently. Consider adding an explicit SSAR intercept for the `roles` resource so the test's intent is self-documenting.

Also, there's no test for the deny case -- what happens when a user does NOT have list access to roles with the flag enabled? You'd want a scenario showing the tab is hidden when SSAR denies access.

DaoDaoNoCode (changes requested):
Solid scaffolding work -- the feature flag integration, area mapping, and SSAR gating on the tab itself are all done correctly. The conditional `loaded` logic in `ProjectDetails` is a nice detail that avoids the loading spinner hang when the flag is off.

Local checks: all CI checks pass (lint, type-check, unit tests, Cypress mock tests all green).

Prior review: CodeRabbit raised the missing RBAC guard on `CreateRolePage` and the `roles/create` route -- I verified this against the code and it's a real issue. The codebase has an established pattern for this (see `ProjectPermissionsAssignRoles`). Left a few comments inline -- the main thing to address is adding an RBAC guard to `CreateRolePage` to match the existing pattern.

frontend/src/pages/projects/projectRoles/CreateRolePage.tsx
```diff
@@ -0,0 +1,39 @@
...
+      }
+      loaded
+      empty={false}
+    >
+      <PageSection hasBodyWrapper={false} data-testid="create-role-page">
+        <EmptyState headingLevel="h2" titleText="Create role">
+          <EmptyStateBody>Role creation form will be implemented here.</EmptyStateBody>
+        </EmptyState>
+      </PageSection>
+    </ApplicationsPage>
```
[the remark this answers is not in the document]
srtanish1992: Fixed.

frontend/src/pages/projects/ProjectViewRoutes.tsx
```diff
@@ -27,6 +29,7 @@ const ProjectViewRoutes: React.FC = () => {
         <Route path="spawner/:notebookName" element={<EditSpawnerPage />} />
         <Route path="permissions" element={<Navigate to="..?section=permissions" replace />} />
         <Route path="permissions/assign" element={<ProjectPermissionsAssignRoles />} />
+        {roleManagementEnabled && <Route path="roles/create" element={<CreateRolePage />} />}
```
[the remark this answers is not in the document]
srtanish1992: Already addressed by #1.

packages/cypress/cypress/tests/mocked/projects/tabs/rolesTab.cy.ts
```diff
@@ -0,0 +1,73 @@
...
+      roleManagement,
+    }),
+  );
+
+  cy.interceptK8s(ProjectModel, mockProjectK8sResource({ k8sName: NAMESPACE }));
+  cy.interceptK8sList(
+    ProjectModel,
+    mockK8sResourceList([mockProjectK8sResource({ k8sName: NAMESPACE })]),
+  );
+};
```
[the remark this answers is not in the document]
srtanish1992: The tests rely on the mock user system (asProjectAdminUser/asProjectEditUser) for default SSAR behavior, which is the established pattern in this repo (see projectList.cy.ts, homeProjects.cy.ts). For the denial case, we override the specific SSAR in beforeEach. Adding a parameter to initIntercepts would duplicate logic already in the mock user system with no change in coverage.

frontend/src/pages/projects/projectRoles/CreateRolePage.tsx
srtanish1992: This is the same finding as #1 — already addressed.

packages/cypress/cypress/tests/mocked/projects/tabs/rolesTab.cy.ts
```diff
@@ -0,0 +1,115 @@
...
+  describe('with roleManagement flag enabled but user lacks create permission', () => {
+    beforeEach(() => {
+      asProjectEditUser();
+      initIntercepts({ roleManagement: true });
+      // Override SSAR to deny 'create' on 'roles'
+      cy.interceptK8s('POST', SelfSubjectAccessReviewModel, (req) => {
+        const { resourceAttributes } = req.body.spec;
+        if (
+          resourceAttributes.resource === 'roles' &&
+          resourceAttributes.verb === 'create' &&
```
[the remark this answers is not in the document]
srtanish1992: The mockUsers.ts infrastructure uses the identical pattern without guards (const { resourceAttributes } = req.body.spec). The intercept only matches SelfSubjectAccessReviewModel POSTs, which always contain spec.resourceAttributes per the K8s API contract.

packages/cypress/cypress/tests/mocked/projects/tabs/rolesTab.cy.ts
srtanish1992: Done.

packages/cypress/cypress/tests/mocked/projects/tabs/rolesTab.cy.ts
srtanish1992: Done.

frontend/src/pages/projects/components/GenericHorizontalBar.tsx
srtanish1992: Fixed.

frontend/src/pages/projects/projectRoles/CreateRolePage.tsx
srtanish1992: Fixed.

srtanish1992:
@DaoDaoNoCode addressed your comments. Ready for review again.

DaoDaoNoCode (approved):
/lgtm

PR #7760, opened 2026-06-01
fix(e2e): null-safe jq in cleanupTemplates and fix serving runtime test

FedeAlonso:
https://redhat.atlassian.net/browse/RHOAIENG-65341
## Description
  - Add ? operators and // "" fallback to the jq query in cleanupTemplates()
    so templates without openshift.io/display-name (e.g. batch/v1 Jobs from
    model-registry-operator) are skipped instead of crashing the pipeline
  - Fix URL assertion in testSingleServingRuntimeCreation to match end of
    path so it waits for the redirect to the list page, not the /add page
  - Add after() cleanup hook so the template is always deleted regardless
    of test outcome

## How Has This Been Tested?
Locally:
Greenshot 2026-06-01 11 05 09

Jenkins:
job/dashboard-e2e-tests/1037
image

## Test Impact
Test fixes

## Request review criteria:

If you have UI changes: 

## Summary by CodeRabbit

* **Tests**
  * Enhanced serving runtime creation test with improved cleanup procedures
  * Refined navigation route verification for serving runtime paths

* **Chores**
  * Improved robustness of template handling to gracefully manage missing metadata scenarios

### DIFF
--- a/packages/cypress/cypress/tests/e2e/settings/servingRuntimes/testSingleServingRuntimeCreation.cy.ts
+++ b/packages/cypress/cypress/tests/e2e/settings/servingRuntimes/testSingleServingRuntimeCreation.cy.ts
@@ -23,6 +23,12 @@ retryableBefore(() => {
       return cleanupTemplates(metadataSingleDisplayName);
     });
 });
+after(() => {
+  if (metadataSingleDisplayName) {
+    cleanupTemplates(metadataSingleDisplayName);
+  }
+});
+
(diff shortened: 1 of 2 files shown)


lucferbux (approved):
/lgtm
/approve

Clean, well-scoped fix. The jq null-safety operators (`?` and `// ""`) correctly prevent crashes when templates lack `openshift.io/display-name`. The URL regex assertion tightening and the `after()` cleanup hook are solid improvements. CodeRabbit flagged the timeout placement on `cy.url().should('match', ...)` — valid nit but not a regression from this PR, and the current approach works in practice. The shell injection concern on `displayName` is pre-existing, not introduced here.

PR #7763, opened 2026-06-01
feat(model-serving): add conditional "Preconfigure deployment" wizard step

mturley:
https://redhat.atlassian.net/browse/RHOAIENG-60392

## Description
When deploying from the "All projects" view, the deployment wizard now shows a conditional **"Preconfigure deployment"** step as the first step, allowing the user to select a project before proceeding. When a project is already known (e.g. navigating from a specific project page or model catalog), the step is absent from the wizard entirely.

When the preconfigure step is shown, the duplicate `ProjectSection` on the "Model deployment" step is hidden (via `hideProjectSection` prop) since the project has already been selected.

Changes:
- **`PreconfigureDeploymentStep.tsx`** (new) — Wizard step with a project selector and help tooltip. Defensively renders a disabled text input when an initial project is already set.
- **`ModelDeploymentWizard.tsx`** — Adds `shouldShowPreconfigureStep` flag (based on whether a `project` prop was provided), conditionally renders the new step, passes `hideProjectSection` to the "Model deployment" step
- **`ModelDeploymentStep.tsx`** — Adds `hideProjectSection` prop to hide `ProjectSection` when the preconfigure step handles project selection
- **`useDeploymentWizardValidation.tsx`** — Accepts `shouldShowPreconfigureStep` param; adds `isPreconfigureStepValid` (always `true` when the step isn't shown), gates `isAllValid` on it
- **`types.ts`** — Adds `PRECONFIGURE` to `WizardStepTitle` enum
- **`modelServing.ts`** (page object) — Adds `findPreconfigureStep`, `findPreconfigureProjectSelector`, `findPreconfigureProjectSelectorOption` methods
- **`modelServingDeploy.cy.ts`** — Updates the "enter without a project" test to select a project in the preconfigure step before proceeding

### Screenshots
wizard-preconfigure-step

## How Has This Been Tested?
- Manually verified on a dev cluster: clicking Deploy from "All projects" opens the wizard with "Preconfigure deployment" as step 1, project selector works, selecting a project enables proceeding to "Model details"
- Manually verified: clicking Deploy from a specific project page skips the preconfigure step entirely — wizard starts at "Model details" as before
- Manually verified: when the preconfigure step is shown, the "Model deployment" step does not show a duplicate project selector
- All existing wizard unit tests pass (179 tests)
- Cypress mock test for "enter without a project" updated to exercise the preconfigure step flow

## Test Impact
- Added unit tests for `PreconfigureDeploymentStep` (4 tests: renders description, renders selector when no initial project, calls setProjectName on selection, renders disabled input when initial project is set)
- Updated Cypress mock test "Should create a new connection with a generated secret name and enter without a project" to select a project in the preconfigure step and verify the project selector is hidden on the "Model deployment" step
- Added page object methods for the preconfigure step (`findPreconfigureStep`, `findPreconfigureProjectSelector`, `findPreconfigureProjectSelectorOption`)

## Request review criteria:
If you have UI changes:

### DIFF
--- a/packages/cypress/cypress/pages/modelServing.ts
+++ b/packages/cypress/cypress/pages/modelServing.ts
@@ -913,6 +913,18 @@ class ModelServingWizard extends Wizard {
     return cy.findByTestId('spinner');
   }
 
+  findPreconfigureStep() {
+    return this.findStep('preconfigure-step');
+  }
+
+  findPreconfigureProjectSelector() {
+    return cy.findByTestId('project-selector-toggle');
+  }
+
+  findPreconfigureProjectSelectorOption(name: string) {
+    return cy.findByTestId('project-selector-menuList').findByRole('menuitem', { name });
+  }
+
   findModelSourceStep() {
     return this.findStep('source-model-step');
   }
--- a/packages/cypress/cypress/tests/e2e/modelCatalog/testCatalogDeployModel.cy.ts
+++ b/packages/cypress/cypress/tests/e2e/modelCatalog/testCatalogDeployModel.cy.ts
@@ -88,6 +88,15 @@ describe('Verify a model can be deployed from model catalog', () => {
 
       modelCatalog.clickDeployModelButtonWithRetry();
 
+      cy.step('Model deployment select project in preconfigure step');
+      modelServingWizard.findModelDeploymentProjectSelector().should('exist');
+      modelServingWizard.findModelDeploymentProjectSelector().click();
+      modelServingWizard
+        .findModelDeploymentProjectSelectorOption(projectName)
+        .should('exist')
+        .click();
+      modelServingWizard.findNextButton().should('be.enabled').click();
+
       cy.step('Verify model location gets prefilled');
       modelServingWizard.findModelSourceStep().click();
       modelServingWizard
@@ -100,12 +109,6 @@ describe('Verify a model can be deployed from model catalog', () => {
 
       cy.step('Model deployment step');
       modelServingWizard.findModelDeploymentNameInput().clear().type(modelName);
-      modelServingWizard.findModelDeploymentProjectSelector().should('exist');
-      modelServingWizard.findModelDeploymentProjectSelector().click();
-      modelServingWizard
-        .findModelDeploymentProjectSelectorOption(projectName)
-        .should('exist')
-        .click();
 
       modelServingWizard.findModelServerManualSelectRadio().click();
       modelServingWizard.findFirstServingRuntimeTemplateOption().should('exist').click();
--- a/packages/cypress/cypress/tests/mocked/modelServing/modelServingDeploy.cy.ts
+++ b/packages/cypress/cypress/tests/mocked/modelServing/modelServingDeploy.cy.ts
@@ -1625,7 +1625,18 @@ describe('Model Serving Deploy Wizard', () => {
 
     modelServingWizard.visit();
 
-    // Step 1: Model source
+    // Step 1: Preconfigure deployment (shown because no project is pre-selected)
+    modelServingWizard.findPreconfigureStep().should('be.enabled');
+    modelServingWizard.findModelSourceStep().should('be.disabled');
+    modelServingWizard.findNextButton().should('be.disabled');
+    modelServingWizard.findPreconfigureProjectSelector().click();
+    modelServingWizard
+      .findPreconfigureProjectSelectorOption('Test Project')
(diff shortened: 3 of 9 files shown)


mturley:
@vconzola how’s this look (Modal body just says “Select a project to deploy your model”, compare with the one for deploying from registry which says "Select a project and connection to deploy your model from the model registry”)

vconzola:
Thanks @mturley. This LGTM. I don't recall why we had select a project and a connection when deploying from registry, but I'm sure @katieperry4 had a good reason.

katieperry4:
@vconzola 
We added the project selector for registry so we could prefill any connection info in the wizard

when you register the model you add model location info, so when you select a project before going to the wizard we can scan the project and either preselect a matching connection from that project, or prefill the wizard with the info presented as a new connection

mturley:
/hold

Chatting with @andrewballantyne on this, its seems I got mixed up and maybe we didn’t update the Jira. @vconzola  I implemented this as a modal here, but did we want to have an actual step 0 in the wizard (conditionally rendered step that we trigger for NIM)? I remember us talking about that option and I don’t remember where we landed on it. Andrew is saying it needs to be a step since we also need it for tool calling.

vconzola:
@mturley We will need a step 0, eventually for tool calling. It's up to y'all if you want to implement that now as part of the NIM work, or later when we do tool calling. If we wait, we'll want to come back and refactor the NIM flow to use step 0, so maybe it's easier/better to do now?  Here is a prototype that @yabbesrajan created that shows the Step 0 behavior. https://project-navigator-rhoai-da8b8e.pages.redhat.com/ai-hub/models/deployments/deploy

mturley:
@vconzola @yabbesrajan thanks. followup question:

There is also a project selector on the "Model deployment" step of the wizard:

image

which gets disabled if you already have a project selected when you reach the wizard:

image

now that the step before this sometimes requires a project selected for its logic and we're introducing step 0 with a project selector, should we just remove this other project selector? if so, since step 0 is conditional, if the user reached the wizard with a project already selected they'd not see a reminder of what project they're deploying in until the Review step. is that ok?

mturley:
For now, I’ll keep the change minimal and just hide that project selector only when step 0 is present, otherwise the wizard will behave the same as it does already (which means it’ll now always be disabled, since we’ll only see it if we already had a project when we reached the wizard). Let me know if you want something different.

vconzola:
@mturley ~~I'd like to keep the Project selector on the Model deployments step. If a user selects a project in step 0 or enters the deployment flow in the context of a project, just pre-fill and disable the selection. That's the way it works today, both for deployments with an existing project context and when the project is selected in the modal from Catalog or Registry.~~

I should have read your comments more closely. I don't think step 0 will be conditional for long. Once we get tool calling and other deployment options in, I think it will likely become a regular step. But for now, do what you suggested.

mturley:
@vconzola any reason to not just make it unconditional now? just because it’s kinda awkward to always show a 1-field wizard step?

vconzola:
Basically, yeah. If we're sure we're going to get tool calling in "soon" and we'll want step 0 for the llm-d flow (still tbd) and it will ease future dev work, then we can make it unconditional now. I'd like to hear @andrewballantyne 's thoughts.

yabbesrajan:
Screenshot 2026-06-04 at 11 22 26 a m
@mturley @vconzola just to add on that, if we make it "unconditional" including the flow **Projects -> individual projects -> deployments -> deploy model** , we will have the awkward 1-filed step here since there is no context of the model to include sections like validated configurations, and the available filed (Choose project) will be read only since the flow is from a particular project. Are we excluding this flow from having a Step 0?

mturley:
That's a good point. So maybe it does make sense to still have it be conditional if it would only have a disabled project selector. Then it's a matter of, in that case, do we retrain the disabled project selector on step 2 or just not show the selected project until the Review step

mturley:
For now I’ll go with what I had suggested in [this comment](https://github.com/opendatahub-io/odh-dashboard/pull/7763#issuecomment-4604327755)

emilys314:
Here is a successful e2e run of the model catalog deploy. It uses the step 0 so 
image

YuliaKrimerman:
/lgtm All looks good
option 1 - No project selected 
Screenshot 2026-06-08 at 5 59 23 PM
option 2 - with project 
Screenshot 2026-06-08 at 5 59 39 PM
/approve

PR #7795, opened 2026-06-03
feat(dashboard-operator): migrate CRD API group and add platform contract

lucferbux:
https://issues.redhat.com/browse/RHOAIENG-65723

## Description

Migrate the Dashboard Module CRD from `dashboard.opendatahub.io/v1alpha1` to the platform-required `components.platform.opendatahub.io/v1alpha1` as mandated by the upstream [Onboarding Guide for ODH Operator Modules](https://gitlab.cee.redhat.com/data-hub/odh-modularisation-docs). This is a critical path blocker for ModuleHandler registration in the ODH Operator ([RHOAIENG-61029](https://redhat.atlassian.net/browse/RHOAIENG-61029)).

### API migration
- **API group**: `dashboard.opendatahub.io` → `components.platform.opendatahub.io`
- **Singleton name**: `default-dashboard` → `default` (CEL rule updated)
- **Leader election ID**: updated to `components.platform.opendatahub.io`
- **Webhook**: API group reference updated in ValidatingWebhookConfiguration
- **RBAC**: All ClusterRole rules updated for new API group
- **CRD**: Regenerated via `make manifests`, old CRD file removed

### Platform contract additions
- **`APPLICATIONS_NAMESPACE`** env var consumed with fallback to operator namespace — separates where the operator runs from where operand resources are deployed
- **Trusted CA bundle** volume mount (`odh-trusted-ca-bundle` ConfigMap, `optional: true`) — standard ODH pattern for custom CAs
- **Finalizer** (`components.platform.opendatahub.io/cleanup`) — registered on Dashboard CR for cross-namespace resource cleanup; placeholder cleanup logic for now, real cleanup added when Perses monitoring namespace resources land
- **xKS platform guard** — `extractDashboardURL` skips Route listing on vanilla Kubernetes (XKS platform)
- **Platform overlay fallback** — `defaultManifestInfo` falls back to ODH overlay for unknown platforms

### Unit tests
- `make generate` — deepcopy regenerated successfully
- `make manifests` — CRD regenerated with new API group and CEL rule verified
- `make lint` — passes with 0 issues
- `make test` — all test packages pass:
  - `api/v1alpha1` — OK (5.3% coverage)
  - `internal/controller` — OK (81.6% coverage)
  - `internal/webhook` — OK (100% coverage)

### Cluster testing (ROSA `ui-monarch-rosa`)

Image `quay.io/lferrnan/odh-dashboard-operator:pr-7795` deployed and validated on a ROSA cluster:

1. **CRD applied** — `dashboards.components.platform.opendatahub.io` with new CEL validation rule
2. **Singleton name enforcement** — `default` accepted, `default-dashboard` rejected by CEL:
   ```
   The Dashboard "default-dashboard" is invalid: Dashboard name must be default
   ```
3. **Controller watches correct API group** — logs show `controllerGroup: "components.platform.opendatahub.io"`
4. **Reconciliation succeeds** — `ProvisioningSucceeded: True`, all dashboard manifests applied via SSA
5. **Finalizer registered** — `components.platform.opendatahub.io/cleanup` present on CR
6. **Status conditions populated correctly**:
   - `ProvisioningSucceeded: True` (reason: ResourcesApplied)
   - `Ready: False` (reason: RouteNotReady — expected, Route label depends on manifests)
   - `Degraded: False` (severity: Info)
7. **ObservedGeneration** set correctly (matches `metadata.generation`)
8. **Leader election** uses new ID `components.platform.opendatahub.io`

#### Cluster testing instructions (for reviewer)

```bash
# Build and push (from repo root, M1 Mac → x86 cluster)
docker buildx build --platform linux/amd64 --no-cache \
  -t quay.io/lferrnan/odh-dashboard-operator:pr-7795 \
  -f dashboard-operator/Dockerfile --push .

# Apply updated CRD
oc apply -f dashboard-operator/config/crd/bases/components.platform.opendatahub.io_dashboards.yaml

# Apply updated RBAC
oc apply -f dashboard-operator/config/rbac/role.yaml

# Scale down, update image, add volumes, scale up
oc scale deployment dashboard-operator -n opendatahub --replicas=0
oc set image deployment/dashboard-operator -n opendatahub \
  manager=quay.io/lferrnan/odh-dashboard-operator:pr-7795
oc set env deployment/dashboard-operator -n opendatahub \
  APPLICATIONS_NAMESPACE=opendatahub

# Ensure webhook TLS secret exists (cert-manager or manual)
oc get secret dashboard-operator-webhook-tls -n opendatahub || \
  (openssl req -x509 -newkey rsa:2048 -keyout /tmp/tls.key -out /tmp/tls.crt \
   -days 365 -nodes -subj '/CN=dashboard-operator-webhook-service.opendatahub.svc' && \
   oc create secret tls dashboard-operator-webhook-tls -n opendatahub \
   --cert=/tmp/tls.crt --key=/tmp/tls.key)

# Patch volumes if not present
oc patch deployment dashboard-operator -n opendatahub --type='json' -p='[
  {"op": "add", "path": "/spec/template/spec/volumes", "value": [
    {"name": "webhook-certs", "secret": {"secretName": "dashboard-operator-webhook-tls"}},
    {"name": "odh-trusted-ca-bundle", "configMap": {"name": "odh-trusted-ca-bundle", "optional": true, "items": [{"key": "ca-bundle.crt", "path": "tls-ca-bundle.pem"}]}}
  ]},
  {"op": "add", "path": "/spec/template/spec/containers/0/volumeMounts", "value": [
    {"name": "webhook-certs", "mountPath": "/tmp/k8s-webhook-server/serving-certs", "readOnly": true},
    {"name": "odh-trusted-ca-bundle", "mountPath": "/etc/pki/tls/custom-certs", "readOnly": true}
  ]}
]'

oc scale deployment dashboard-operator -n opendatahub --replicas=1

# Delete old CRs, create new one with singleton name "default"
oc delete dashboards.dashboard.opendatahub.io --all 2>/dev/null
oc delete dashboards.components.platform.opendatahub.io --all 2>/dev/null
cat <<EOF | oc apply -f -
apiVersion: components.platform.opendatahub.io/v1alpha1
kind: Dashboard
metadata:
  name: default
spec:
  managementState: Managed
EOF

# Verify
oc logs deployment/dashboard-operator -n opendatahub --tail=10
oc get dashboards.components.platform.opendatahub.io default -o yaml

# Verify CEL rejects old name
cat <<EOF | oc apply -f - 2>&1 | grep "must be default"
apiVersion: components.platform.opendatahub.io/v1alpha1
kind: Dashboard
metadata:
  name: default-dashboard
spec:
  managementState: Managed
EOF
```

## Test Impact

Updated existing tests to cover new behavior:
- `dashboard_types_test.go` — updated `DashboardInstanceName` assertion to `"default"`
- `actions_test.go` — added platform parameter to `extractDashboardURL` tests, added xKS platform test case
- `dashboard_reconciler_test.go` — added `ApplicationsNamespace` field to all test reconciler instances
- `support_test.go` — added test for unknown platform fallback to ODH overlay

Finalizer add/remove is exercised indirectly through the reconciler tests; dedicated finalizer unit tests can be added once the cleanup logic has concrete resource types.

## Request review criteria:

## Summary by CodeRabbit

* **New Features**
  * Configure a separate applications namespace via env var.
  * Finalizer-based cleanup when deleting dashboards.
  * Improved handling for additional cluster platform types.

* **Improvements**
  * Dashboard URL now persists across transient route failures (check Ready condition).
  * Manifest rendering and URL extraction use the applications namespace when configured.

* **Chores**
  * Dashboard API group renamed to platform-aligned naming.
  * Default dashboard name normalized to "default".
  * RBAC and webhook manifests updated to match API group change.

* **Ops**
  * SSL certificate bundle mount added for the manager.

### DIFF
--- a/dashboard-operator/api/v1alpha1/dashboard_types.go
+++ b/dashboard-operator/api/v1alpha1/dashboard_types.go
@@ -140,7 +140,9 @@ type DashboardStatus struct {
 	common.Status                 `json:",inline"`
 	common.ComponentReleaseStatus `json:",inline"`
 
-	// URL is the externally-reachable dashboard URL.
+	// URL is the externally-reachable dashboard URL (last known good).
+	// This value persists across transient route failures — consumers must
+	// check the Ready condition before relying on this endpoint.
 	// +optional
 	URL string `json:"url,omitempty"`
 
--- a/dashboard-operator/api/v1alpha1/dashboard_types.go
+++ b/dashboard-operator/api/v1alpha1/dashboard_types.go
@@ -7,7 +7,7 @@ import (
 
 const (
 	DashboardComponentName = "dashboard"
-	DashboardInstanceName  = "default-dashboard"
+	DashboardInstanceName  = "default"
 	DashboardKind          = "Dashboard"
 )
 
@@ -158,7 +158,7 @@ type DashboardStatus struct {
 // +kubebuilder:printcolumn:name="Reason",type=string,JSONPath=`.status.conditions[?(@.type=="Ready")].reason`
 // +kubebuilder:printcolumn:name="URL",type=string,JSONPath=`.status.url`
 // +kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp"
-// +kubebuilder:validation:XValidation:rule="self.metadata.name == 'default-dashboard'",message="Dashboard name must be default-dashboard"
+// +kubebuilder:validation:XValidation:rule="self.metadata.name == 'default'",message="Dashboard name must be default"
 // +kubebuilder:validation:XValidation:rule="!has(self.spec.observability) || !self.spec.observability.enabled || has(self.spec.observability.persesService)",message="persesService must be specified when observability is enabled"
 
 // Dashboard is the Schema for the dashboards API.
--- a/dashboard-operator/cmd/manager/main.go
+++ b/dashboard-operator/cmd/manager/main.go
@@ -75,7 +75,7 @@ func main() {
 		Metrics:                metricsserver.Options{BindAddress: metricsAddr},
 		HealthProbeBindAddress: healthProbeAddr,
 		LeaderElection:         leaderElect,
-		LeaderElectionID:       "components.platform.opendatahub.io",
+		LeaderElectionID:       "dashboard.components.platform.opendatahub.io",
 		WebhookServer:          ctrlwebhook.NewServer(ctrlwebhook.Options{Port: webhookPort}),
 	})
 	if err != nil {
--- a/dashboard-operator/cmd/manager/main.go
+++ b/dashboard-operator/cmd/manager/main.go
@@ -65,12 +65,17 @@ func main() {
 		os.Exit(1)
 	}
 
+	applicationsNamespace := os.Getenv("APPLICATIONS_NAMESPACE")
+	if applicationsNamespace == "" {
+		applicationsNamespace = operatorNamespace
+	}
+
 	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
 		Scheme:                 scheme,
 		Metrics:                metricsserver.Options{BindAddress: metricsAddr},
 		HealthProbeBindAddress: healthProbeAddr,
 		LeaderElection:         leaderElect,
-		LeaderElectionID:       "dashboard.opendatahub.io",
+		LeaderElectionID:       "components.platform.opendatahub.io",
 		WebhookServer:          ctrlwebhook.NewServer(ctrlwebhook.Options{Port: webhookPort}),
 	})
 	if err != nil {
@@ -87,9 +92,10 @@ func main() {
 	setupLog.Info("Detected platform", "platform", platform)
 
 	if err := controller.SetupWithManager(mgr, controller.Options{
-		ManifestsBasePath: manifestsBasePath,
-		Platform:          platform,
-		Namespace:         operatorNamespace,
+		ManifestsBasePath:     manifestsBasePath,
+		Platform:              platform,
+		Namespace:             operatorNamespace,
+		ApplicationsNamespace: applicationsNamespace,
 	}); err != nil {
 		setupLog.Error(err, "unable to create Dashboard controller")
 		os.Exit(1)
--- a/dashboard-operator/internal/controller/dashboard_reconciler.go
+++ b/dashboard-operator/internal/controller/dashboard_reconciler.go
@@ -82,6 +82,8 @@ func (r *DashboardReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		if err := r.Update(ctx, dashboard); err != nil {
 			return ctrl.Result{}, fmt.Errorf("failed to add finalizer: %w", err)
 		}
+
+		return ctrl.Result{}, nil
 	}
 
 	dashboard.Status.ObservedGeneration = dashboard.Generation
--- a/dashboard-operator/internal/controller/dashboard_reconciler.go
+++ b/dashboard-operator/internal/controller/dashboard_reconciler.go
@@ -127,7 +127,6 @@ func (r *DashboardReconciler) reconcile(
 	cm *conditions.Manager,
 ) (ctrl.Result, error) {
 	logger := log.FromContext(ctx)
-	dashboard.Status.URL = ""
 
 	manifests := manifestSets(r.ManifestsBasePath, r.Platform)
 
--- a/dashboard-operator/internal/controller/dashboard_reconciler.go
+++ b/dashboard-operator/internal/controller/dashboard_reconciler.go
@@ -182,6 +182,8 @@ func (r *DashboardReconciler) reconcile(
 
 	var requeueAfter time.Duration
 
+	// URL is intentionally not cleared on error — "last known good" semantic.
+	// Conditions (Ready, Degraded) communicate the actual state.
 	switch {
 	case errors.Is(err, ErrDashboardRouteNotReady):
 		cm.MarkFalse(string(common.ConditionTypeDegraded),
--- a/dashboard-operator/internal/controller/dashboard_reconciler.go
+++ b/dashboard-operator/internal/controller/dashboard_reconciler.go
@@ -191,7 +191,6 @@ func (r *DashboardReconciler) reconcile(
 		cm.MarkFalse(string(common.ConditionTypeReady),
 			conditions.WithReason("RouteNotReady"),
 			conditions.WithMessage("Dashboard route is not yet admitted"))
-		dashboard.Status.URL = ""
 		logger.Info("Dashboard route not yet available, requeuing")
 		requeueAfter = 10 * time.Second
 	case err != nil:
@@ -202,7 +201,6 @@ func (r *DashboardReconciler) reconcile(
 		cm.MarkFalse(string(common.ConditionTypeReady),
 			conditions.WithReason("URLExtractionFailed"),
 			conditions.WithError(err))
-		dashboard.Status.URL = ""
 		logger.Error(err, "Failed to extract dashboard URL")
 
 		return ctrl.Result{}, fmt.Errorf("failed to extract dashboard URL: %w", err)
@@ -212,13 +210,14 @@ func (r *DashboardReconciler) reconcile(
 			conditions.WithMessage("All sub-modules healthy"),
 			conditions.WithSeverity(common.ConditionSeverityInfo))
 		dashboard.Status.URL = url
+		logger.Info("Dashboard reconciled successfully", "url", url)
 	}
-
-	logger.Info("Dashboard reconciled successfully", "url", url)
 
 	return ctrl.Result{RequeueAfter: requeueAfter}, nil
 }
 
+// TODO(RHOAIENG-59938): delete Perses monitoring resources in the observability namespace
+// and any SSA-adopted resources not covered by ownerReference GC.
 func (r *DashboardReconciler) cleanupCrossNamespaceResources(ctx context.Context, _ *v1alpha1.Dashboard) error {
 	logger := log.FromContext(ctx)
 	logger.Info("Cleaning up cross-namespace resources")
--- a/dashboard-operator/internal/controller/dashboard_reconciler.go
+++ b/dashboard-operator/internal/controller/dashboard_reconciler.go
@@ -12,6 +12,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 
 	"github.com/opendatahub-io/odh-platform-utilities/api/common"
@@ -24,23 +25,27 @@ import (
 	v1alpha1 "github.com/opendatahub-io/odh-dashboard/dashboard-operator/api/v1alpha1"
 )
 
+const dashboardFinalizer = "components.platform.opendatahub.io/cleanup"
+
 // Version is set at build time via -ldflags.
 var Version = "unknown"
 
 // Options configures the dashboard controller.
 type Options struct {
-	ManifestsBasePath string
-	Platform          cluster.Platform
-	Namespace         string
+	ManifestsBasePath     string
+	Platform              cluster.Platform
+	Namespace             string
+	ApplicationsNamespace string
 }
 
 // DashboardReconciler reconciles a Dashboard object.
 type DashboardReconciler struct {
 	client.Client
-	Scheme            *runtime.Scheme
-	ManifestsBasePath string
-	Platform          cluster.Platform
-	Namespace         string
+	Scheme                *runtime.Scheme
+	ManifestsBasePath     string
+	Platform              cluster.Platform
+	Namespace             string
+	ApplicationsNamespace string
 }
 
 func (r *DashboardReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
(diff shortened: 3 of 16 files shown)


lucferbux:

dashboard-operator/cmd/manager/main.go
```diff
@@ -65,12 +65,17 @@ func main() {
 		os.Exit(1)
 	}
 
+	applicationsNamespace := os.Getenv("APPLICATIONS_NAMESPACE")
+	if applicationsNamespace == "" {
+		applicationsNamespace = operatorNamespace
+	}
```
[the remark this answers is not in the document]
lucferbux: Skipping — both namespace values come from the Kubernetes downward API (`metadata.namespace`), which is guaranteed valid by the API server. Adding validation here would be dead code.

dashboard-operator/internal/controller/dashboard_reconciler.go
```diff
@@ -57,6 +62,28 @@ func (r *DashboardReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+
+		return ctrl.Result{}, nil
+	}
+
+	if !controllerutil.ContainsFinalizer(dashboard, dashboardFinalizer) {
+		controllerutil.AddFinalizer(dashboard, dashboardFinalizer)
+		if err := r.Update(ctx, dashboard); err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to add finalizer: %w", err)
+		}
+	}
```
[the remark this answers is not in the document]
lucferbux: Fixed in 08d3e5c — added `return ctrl.Result{}, nil` after adding the finalizer. The watch event re-enqueues reconciliation with the persisted object.

dashboard-operator/internal/controller/dashboard_reconciler.go
```diff
@@ -190,14 +217,22 @@ func (r *DashboardReconciler) reconcile(
 	return ctrl.Result{RequeueAfter: requeueAfter}, nil
 }
 
+func (r *DashboardReconciler) cleanupCrossNamespaceResources(ctx context.Context, _ *v1alpha1.Dashboard) error {
+	logger := log.FromContext(ctx)
+	logger.Info("Cleaning up cross-namespace resources")
+
+	return nil
```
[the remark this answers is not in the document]
lucferbux: Intentional placeholder per [RHOAIENG-65723](https://redhat.atlassian.net/browse/RHOAIENG-65723) acceptance criteria. The Dashboard CR is cluster-scoped, so ownerReference GC handles all same-cluster resources deployed by the SSA deployer. The finalizer is scaffolded now for PR-6 ([RHOAIENG-59938](https://redhat.atlassian.net/browse/RHOAIENG-59938)) which adds SSA ForceOwnership migration and Perses monitoring namespace cleanup — those are the cross-namespace resources that need explicit cleanup. The no-op stub returns nil immediately so it adds no deletion delay.

dashboard-operator/internal/controller/dashboard_reconciler_test.go
```diff
@@ -247,6 +250,111 @@ func TestReconcile(t *testing.T) {
...
+
+	require.NoError(t, err)
+	assert.Equal(t, ctrl.Result{}, result)
+
+	updated := &v1alpha1.Dashboard{}
+	err = cli.Get(context.Background(), types.NamespacedName{Name: v1alpha1.DashboardInstanceName}, updated)
+	if err == nil {
+		assert.Empty(t, updated.Finalizers, "finalizer should be removed after deletion")
+	}
+}
```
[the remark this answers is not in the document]
lucferbux: The new `TestReconcile_Deletion` and `TestReconcile_DistinctNamespaces` tests already cover these paths as standalone test functions. Converting the entire existing table-driven `TestReconcile` to include these would bloat the test struct with fields only used by one case (DeletionTimestamp, distinct namespaces, manifest setup). Keeping them as focused tests is clearer.

dashboard-operator/internal/controller/dashboard_reconciler_test.go
```diff
@@ -247,6 +250,111 @@ func TestReconcile(t *testing.T) {
...
+	})
+
+	require.NoError(t, err)
+	assert.Equal(t, ctrl.Result{}, result)
+
+	updated := &v1alpha1.Dashboard{}
+	err = cli.Get(context.Background(), types.NamespacedName{Name: v1alpha1.DashboardInstanceName}, updated)
+	if err == nil {
+		assert.Empty(t, updated.Finalizers, "finalizer should be removed after deletion")
+	}
```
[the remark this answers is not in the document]
lucferbux: Fixed in 08d3e5c — the deletion test now handles the case where the fake client fully removes the object after finalizer removal (which is correct K8s behavior when DeletionTimestamp is set and no finalizers remain).

lucferbux:

lucferbux:
## Addressing preflight agent findings

### Fixed in e4fe6dc

| Finding | Fix |
|---------|-----|
| `cleanupCrossNamespaceResources` stub lacks TODO | Added `TODO([RHOAIENG-59938](https://redhat.atlassian.net/browse/RHOAIENG-59938))` referencing the migration issue |
| `LeaderElectionID` collision risk | Changed to `dashboard.components.platform.opendatahub.io` |
| `params.env` mode 0664 too permissive | Tightened to `0600` |
| "reconciled successfully" log fires when route not ready | Moved into the `default:` case only |
| `dashboard.Status.URL = ""` redundant resets | Removed — already cleared at line 130 |
| `APPLICATIONS_NAMESPACE` same as `OPERATOR_NAMESPACE` | Added override-point comment explaining it's set by the ODH Operator to a potentially different namespace |

### Previously fixed

| Finding | Commit |
|---------|--------|
| Missing early return after finalizer add | 08d3e5c |

### Intentionally not addressed

**RBAC: ClusterRole/ClusterRoleBinding and CRD write access (Major)**
These are pre-existing and required. The operator deploys dashboard manifests that include RBAC resources (Roles, ClusterRoles, ClusterRoleBindings for the dashboard ServiceAccount) and manages CRDs. Without write access to these resource types, the SSA deployer would fail to apply the manifests. Scoping with `resourceNames` isn't feasible since the manifest set is configurable per platform overlay. This is the standard pattern across all ODH component operators.

**Secrets write cluster-wide (Minor)**
Same as above — the manifests include Secrets (OAuth client secrets, TLS configs) that the SSA deployer applies. The deployer sets ownerReferences so these are garbage-collected on CR deletion.

**Missing `routes/status` subresource rule (Minor)**
In OpenShift, route status is embedded in the main Route object returned by `List`. Adding `routes/status: [get]` is purely cosmetic — the operator doesn't make explicit status subresource calls. No behavioral change.

**`Degraded` condition set to `False` in error paths (Minor)**
This is correct per platform utilities convention. `Degraded` is a **negative polarity** condition — `True` means the system IS degraded. Setting it `False` in the route-not-ready path is intentional: the dashboard is not yet ready but it's not degraded (it's still coming up). `Degraded: True` is reserved for partial failures where the system is running but unhealthy.

**PR touches .tsx/.css files (PR Body warning)**
False positive — this PR only modifies Go source and YAML manifests in `dashboard-operator/`.

dashboard-operator/internal/controller/dashboard_reconciler.go
```diff
@@ -162,7 +191,6 @@ func (r *DashboardReconciler) reconcile(
 		cm.MarkFalse(string(common.ConditionTypeReady),
 			conditions.WithReason("RouteNotReady"),
```
liday-rh, line 193: **Medium — Stale URL preserved in status on route failure**

The removal of `dashboard.Status.URL = ""` from this error branch (and the `URLExtractionFailed` branch below) means that once a URL is successfully written to `.status.url`, it persists even when conditions report `RouteNotReady` or `URLExtractionFailed`. This creates a split-brain in the status: conditions say "not ready" but `.status.url` still advertises a potentially dead endpoint.

Any consumer reading `.status.url` without also checking conditions gets a stale/misleading answer.

Consider either clearing the URL on failure (old behavior) or documenting in the CRD status spec that `.status.url` represents "last known good" and consumers must check the `Ready` condition.

*Generated by Claude*

dashboard-operator/cmd/manager/main.go
```diff
@@ -65,12 +65,17 @@ func main() {
...
+
 	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
 		Scheme:                 scheme,
 		Metrics:                metricsserver.Options{BindAddress: metricsAddr},
 		HealthProbeBindAddress: healthProbeAddr,
 		LeaderElection:         leaderElect,
-		LeaderElectionID:       "dashboard.opendatahub.io",
+		LeaderElectionID:       "dashboard.components.platform.opendatahub.io",
```
liday-rh, line 78: **Low-Medium — Leader Election ID change risks dual-leader during rolling upgrade**

The earlier bot comment flagged collision risk with *other* operators, which was addressed by making the ID more specific. However, a different concern remains: during a rolling upgrade, old pods use `dashboard.opendatahub.io` and new pods use `dashboard.components.platform.opendatahub.io`. Since these are different lease names, both an old-version and new-version pod can acquire leadership simultaneously.

If they share any side effects (writing to the same namespace, touching the same ConfigMaps/Secrets via manifests), this could cause races.

Verify that the upgrade procedure ensures the old operator is fully terminated before the new one starts, or consider keeping the same `LeaderElectionID` during the transition period.

*Generated by Claude*

dashboard-operator/api/v1alpha1/dashboard_types.go
```diff
@@ -7,7 +7,7 @@ import (
 
 const (
 	DashboardComponentName = "dashboard"
-	DashboardInstanceName  = "default-dashboard"
+	DashboardInstanceName  = "default"
```
liday-rh, line 10: **Low — Singleton name `default` is very generic**

`default` is heavily overloaded in Kubernetes (default namespace, default ServiceAccount, default StorageClass, etc.). `kubectl get dashboards default` is less self-documenting than the previous `default-dashboard`.

Likely driven by platform onboarding requirements and non-negotiable, but worth noting for operator documentation and troubleshooting guides.

*Generated by Claude*

liday-rh (commented):
Review focusing on code quality, security, and efficiency — three new findings not previously raised in existing PR comments.

*Generated by Claude*

dashboard-operator/internal/controller/dashboard_reconciler.go
lucferbux: Good catch on the split-brain observation. This is intentional "last known good" semantics — the reasoning:

1. **On xKS platforms** (the new target), `extractDashboardURL` returns `("", nil)` immediately, so the `default` case sets `dashboard.Status.URL = ""`. No stale URL risk there.

2. **On OCP/RHOAI**, transient route unavailability (pod rollout, ingress controller restart, brief network partitions) is common. Clearing the URL during those transients would force all downstream consumers to re-discover it once the route recovers. Keeping the last-known URL avoids churn while the `Ready` condition (set to `ConditionFalse` with `RouteNotReady` / `URLExtractionFailed` reason) accurately communicates the degraded state.

3. **Consumer contract**: any well-behaved consumer of the `Dashboard` status should check the `Ready` condition before trusting the URL — this is standard practice for Kubernetes status subresources. A condition saying "not ready" + a stale URL is more informative than "not ready" + empty URL, because it tells you *what* was working before.

That said, the comment about documenting this in the CRD status spec is valid. I'll add a godoc comment on the `URL` field in `DashboardStatus` clarifying the "last known good" semantic.

dashboard-operator/cmd/manager/main.go
lucferbux: Valid concern. A few mitigating factors:

1. **Single-replica deployment**: The dashboard operator runs as a single-replica Deployment. During a rolling upgrade with default `maxSurge=1, maxUnavailable=0`, there's a brief window where both pods exist, but the old pod's shutdown is initiated as soon as the new pod passes readiness.

2. **SSA idempotency**: The operator's primary side effect is Server-Side Apply (SSA) of manifests. SSA is inherently idempotent — two operators applying the same manifests to the same namespace produces the same result. There are no compare-and-swap or read-modify-write patterns that would race.

3. **Lease ID change is unavoidable**: The old API group (`dashboard.opendatahub.io`) is being retired. Keeping the old lease name would create a false dependency on the old group and confuse operators about which version "owns" the lease after migration completes. A clean break aligns with the CRD migration.

4. **Upgrade procedure**: In practice, OLM-managed operators go through `Pending → Installing → Succeeded` which terminates the old pod before starting the new one. For manual deployments, the upgrade doc should note the lease name change.

The window of potential dual-leadership is brief (seconds) and the side effects are idempotent, so the practical risk is negligible. Worth noting in the upgrade/migration docs though.

dashboard-operator/api/v1alpha1/dashboard_types.go
lucferbux: Correct — this is driven by the platform onboarding contract. The platform team's component registration requires the singleton CR name to be `default` (not `default-<component>`). All components being onboarded (ModelController, ModelRegistry, etc.) follow the same `default` naming convention.

While `default-dashboard` was more self-documenting, the CRD kind (`Dashboard`) already provides disambiguation: `kubectl get dashboards.components.platform.opendatahub.io default` is unambiguous in context. And since it's a singleton, there's no risk of collision with other `default`-named resources of different kinds.

Agreed it's worth a note in the operator docs / troubleshooting guide for anyone used to the old name.

dashboard-operator/api/v1alpha1/dashboard_types.go
```diff
@@ -140,7 +140,9 @@ type DashboardStatus struct {
 	common.Status                 `json:",inline"`
 	common.ComponentReleaseStatus `json:",inline"`
 
-	// URL is the externally-reachable dashboard URL.
+	// URL is the externally-reachable dashboard URL (last known good).
+	// This value persists across transient route failures — consumers must
```
liday-rh, line 144: **Medium — "last known good" godoc contradicts implementation**

This godoc now says the URL persists across transient route failures, but there is a pre-existing `dashboard.Status.URL = ""` at the top of `reconcile()` (~line 130 in the full file) that clears the URL on every reconcile entry — before the error-handling switch is ever reached.

The documented contract is not honored: the URL is wiped every cycle and only restored on the success path. The removal of `dashboard.Status.URL = ""` from the error branches in this PR was effectively a no-op since the top-of-function clear already did it.

To implement "last known good", remove the top-of-function `dashboard.Status.URL = ""`. Or if clear-every-cycle is the intended behavior, update this godoc to say "current reachable URL" rather than "last known good".

CodeRabbit independently flagged the same inconsistency.

*Generated by Claude*

liday-rh (commented):
Re-review follow-up — one new finding on the URL status contract.

*Generated by Claude*

dashboard-operator/internal/controller/dashboard_reconciler.go
```diff
@@ -149,10 +178,12 @@ func (r *DashboardReconciler) reconcile(
 		conditions.WithReason("ResourcesApplied"),
 		conditions.WithMessage("Dashboard manifests applied successfully"))
 
-	url, err := extractDashboardURL(ctx, r.Client, r.Namespace)
+	url, err := extractDashboardURL(ctx, r.Client, r.ApplicationsNamespace, r.Platform)
 
 	var requeueAfter time.Duration
 
+	// URL is intentionally not cleared on error — "last known good" semantic.
+	// Conditions (Ready, Degraded) communicate the actual state.
```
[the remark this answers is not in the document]
lucferbux: Good catch — you're right that the top-of-function `dashboard.Status.URL = ""` was clearing the URL every cycle, making the error-branch removal a no-op and contradicting the godoc. Removed it in f160281 so the URL now truly persists as last-known-good.

dashboard-operator/api/v1alpha1/dashboard_types.go
lucferbux: Valid — the top-of-function clear at line 130 was indeed defeating the intent. Removed it in f160281 so the contract is now honored: URL persists across transient failures and is only updated on the success path.

lucferbux:
/lgtm
/approve

Just checking why @liday-rh didn't get through, @christianvogt can you back us up here and add the missing labels please?

**update** will be fixed after https://github.com/opendatahub-io/odh-dashboard/pull/7925

PR #7798, opened 2026-06-03
fix(e2e): harden model registry properties test and AutoML navigation

antowaddle:
## Description

This PR improves E2E test reliability by hardening two flaky test areas and ensures non-GA feature tests are properly tagged to run separately from Smoke/Tier suites.

### E2E Reliability Improvements

**1. Model Registry - Custom Properties Test**
- **Issue**: Race condition where table element exists but tbody rows haven't rendered yet
- **Fix**: Added explicit wait for `tbody tr` elements before searching for custom properties
- **Impact**: Prevents intermittent failures in `testRetainCustomProperties.cy.ts`
- **File**: `packages/cypress/cypress/pages/modelRegistry/registeredModelDetails.ts`

**2. AutoML - Page Navigation** 
- **Issue**: 404 errors when feature flag was just enabled but frontend hasn't reloaded config
- **Fix**: Added `cy.reload()` after visiting home, then wait for AutoML nav item with Cypress retries
- **Impact**: Prevents navigation failures when AutoML feature flag is toggled
- **File**: `packages/cypress/cypress/pages/automl/experimentsPage.ts`

### Test Tagging for Non-GA Features

Added `@Featureflagged` tag and removed `@Smoke`/`@Sanity` tags from tests that manipulate feature flags or DSC components:

- **AutoML tests** (5 files): Use `setAutomlEnabled()` to toggle feature flag
- **Model Training tests** (6 files): Check `isTrainerManaged()` DSC component state
- **Model Catalog test** (1 file): Use `ensureModelCatalogSourceEnabled()` 

This ensures tests with complex setup (DSC component states, feature flags, CRs, leader election) run in dedicated `@Featureflagged` CI pipelines rather than blocking Smoke/Tier test suites.

## How Has This Been Tested?
- In Progress

## Test Impact

- [x] Updated page object methods with improved waiting strategies
- [x] No new tests added - improvements are to existing page objects
- [x] Changes are internal to test infrastructure (page objects and test tags)

## Request review criteria:

Self checklist (all need to be checked):
- [x] The developer has manually tested the changes and verified that the changes work
- [x] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [x] The developer has added tests or explained why testing cannot be added (unit or cypress tests for related changes)
- [x] The code follows our [Best Practices](/docs/best-practices.md) (React coding standards, PatternFly usage, performance considerations)

If you have UI changes: 
- [ ] Included any necessary screenshots or gifs if it was a UI change.
[...]

PR #7803, opened 2026-06-03
fix(e2e): stabilize S3 file explorer click in automl binary classification test

mturley:
## Description
Fixes a flaky E2E test failure in `testAutomlBinaryClassification` where `cy.click()` fails with "page updated while this command was executing" (element detached from DOM).

**Root cause:** The S3 file explorer uses a [300ms debounced search](https://github.com/opendatahub-io/odh-dashboard/blob/main/packages/automl/frontend/src/app/components/common/S3FileExplorer/S3FileExplorer.tsx#L370-L380). When the test types a filename into the search input, the debounce fires after 300ms and triggers a new S3 fetch. The fetch response re-renders the table with fresh React elements, detaching the `td` element that Cypress found mid-chain.

**Fix:** Break the chained `.contains().should().click()` into separate steps using a Cypress `.as()` alias. Each `cy.get('@alias')` re-queries the DOM, so even if the table re-renders between steps, Cypress finds the fresh element.

**Jenkins build with failure:** dashboard-e2e-tests#1074

## How Has This Been Tested?
- Analyzed the Jenkins E2E failure output to confirm the root cause
- Reviewed the S3FileExplorer component source to understand the 300ms debounce trigger
- Verified the fix follows the [Cypress recommended pattern](https://on.cypress.io/element-has-detached-from-dom) for handling detached elements
- Lint passes clean

## Test Impact
This PR *is* the test fix — it stabilizes the existing `testAutomlBinaryClassification` E2E test by preventing the detached-from-DOM race condition.

## Request review criteria:

## Summary by CodeRabbit

* **Tests**
  * Improved the reliability of AutoML file selection testing by adding explicit visibility waits before interaction, accounting for asynchronous search behavior.

### DIFF
--- a/packages/cypress/cypress/pages/automl/configurePage.ts
+++ b/packages/cypress/cypress/pages/automl/configurePage.ts
@@ -179,7 +179,10 @@ class AutomlConfigurePage {
     this.findBrowseBucketButton().click();
     this.findFileExplorerTable().should('be.visible');
     this.findFileExplorerSearch().type(uploadFileName);
-    this.findFileExplorerTable().contains('td', uploadFileName).should('be.visible').click();
+    // Wait for the 300ms debounced S3 search to complete before interacting with results
+    this.findFileExplorerTable().contains('td', uploadFileName).as('uploadedFileCell');
+    cy.get('@uploadedFileCell').should('be.visible');
+    cy.get('@uploadedFileCell').click();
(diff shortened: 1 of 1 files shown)


mturley:
Fix may have been misguided, appears to also be fixed by https://github.com/opendatahub-io/odh-dashboard/pull/7798. Closing.

PR #7811, opened 2026-06-03
feat(distributions): make base shell reusable

jenny-s51:
https://issues.redhat.com/browse/RHOAIENG-65797

## Description

Foundation for the multi-distribution architecture. Makes the base app shell shareable so downstream distributions (rhaii, etc.) can compose their own dashboard variant from shared components without duplicating build config or reaching into internal file paths.

Enables the distributions workstream by parameterizing the base webpack config as a factory and adding barrel exports for shell components.

**Changes:**

- **Webpack factory** — `webpack.common.js` now accepts `{ distributionSrcDir, title, additionalIncludes }` so a new distribution is just a thin config file that calls the factory with its own source directory.
- **Barrel export** — added `src/lib.ts` and package.json `exports` so distributions import shell components via `@odh-dashboard/base-distribution` instead of fragile relative paths.
- **Removed Module Federation dependency** — distributions use static imports and bundle everything at build time.

## How Has This Been Tested?

- `npm install` from root is unaffected (no workspace or dependency changes)
- Base distribution builds successfully with factory defaults: `npx webpack --config distributions/base/config/webpack.prod.js`
- `npx tsc --noEmit` passes with no type errors
- Dev server boots and base app renders at `http://localhost:4010`

## Test Impact

No new tests — this is a build config refactor with no runtime behavior change. The factory defaults are validated by the existing base build continuing to work identically.

## Request review criteria:

Self checklist (all need to be checked):
- [x] The developer has manually tested the changes and verified that the changes work
- [x] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [x] The developer has added tests or explained why testing cannot be added (unit or cypress tests for related changes)
- [x] The code follows our [Best Practices](/docs/best-practices.md) (React coding standards, PatternFly usage, performance considerations)

After the PR is posted & before it merges:
- [ ] The developer has tested their solution on a cluster by using the image produced by the PR to `main`

## Summary by CodeRabbit

* **New Features**
  * Public API expanded with Shell layout components, theme utilities, and ErrorBoundary.
  * Build output and HTML title made configurable for distributions.

* **Chores**
  * Simplified build configuration by consolidating common settings and removing Module Federation.
  * Dev server proxy for legacy module-federation route removed.
* **Packaging**
  * Package entrypoint/export map added to expose library surface.

PR #7849, opened 2026-06-04
feat(distributions): scaffold RHAII distribution

jenny-s51:
https://issues.redhat.com/browse/RHOAIENG-66191
https://issues.redhat.com/browse/RHOAIENG-66192

## Description

Scaffolds `distributions/rhaii/` — the RHAII distribution that composes the base shell with a scaffold plugin to validate the extension pipeline end-to-end. Builds on #7811 (base webpack factory + barrel exports).

**Changes:**

- **Webpack config** — `webpack.common.js` calls the base factory with `{ srcDir, title: 'RHAII' }`; dev server on port 4020 with BFF proxy; prod build
- **Shell composition** — `bootstrap.tsx` imports Shell/ShellHeader/ShellNav/ShellRoutes from base via relative source paths, wraps in ThemeProvider + ExtensibilityContextProvider + ErrorBoundary + RouterProvider
- **Scaffold plugin** — inline extensions (`app.navigation/section`, `app.navigation/href`, `app.route`) that register a "Scaffold Section > Scaffold Page" nav item and route at `/scaffold`
- **ExtensibilityContext** — creates a PluginStore from local `useAppExtensions` and provides it to the shell via PluginStoreProvider

**Note:** `distributions/rhaii` is intentionally **not** registered in root `package.json` workspaces. Workspace registration and barrel import migration are tracked in [RHOAIENG-66776](https://redhat.atlassian.net/browse/RHOAIENG-66776). For local development, see the workaround in `distributions/base/README.md`.

The scaffold plugin proves the distribution pattern works. Once model-serving is decoupled ([RHOAIENG-62515](https://redhat.atlassian.net/browse/RHOAIENG-62515)), swapping it in is a 3-line diff replacing the scaffold extensions with the real ones.

## How Has This Been Tested?

- `npx tsc --noEmit` passes with no type errors in `distributions/rhaii/`
- `npx webpack --config ./config/webpack.prod.js` compiles successfully
- Dev server boots (`npx webpack serve --config ./config/webpack.dev.js`)
- Scaffold nav item appears in sidebar; navigating to `/scaffold` renders ScaffoldPage

## Test Impact

No new tests — this is a scaffold distribution with inline extensions. The scaffold validates the extension pipeline visually (nav + routing renders correctly). Automated tests will be added post-M2 when the scaffold is replaced with real model-serving extensions.

## Request review criteria:

## Summary by CodeRabbit

* **New Features**
  * New RHAII distribution: scaffold page, bootstrapped entry, extensibility plumbing, route scaffolding, auth placeholder, redirect to scaffold, and image import support.
  * Dev server with HMR and production build; npm scripts for dev, build, type-check, and lint.

* **Bug Fixes / Improvements**
  * Build output path now resolved to an absolute directory.
  * Masthead toolbar items can be positioned as leading or trailing.

* **Chores**
  * Added package metadata, ESLint preset usage, gitignore, and TypeScript config.

* **UI Changes**
  * Masthead “About” removed; default Home navigation entry hidden.

### DIFF
--- a/distributions/rhaii/config/webpack.common.js
+++ b/distributions/rhaii/config/webpack.common.js
@@ -1,0 +1,7 @@
+const path = require('path');
+const createWebpackCommon = require('../../base/config/webpack.common.js');
+
+const SRC_DIR = path.resolve(__dirname, '../src');
+
+module.exports = (overrides = {}) =>
+  createWebpackCommon({ distributionSrcDir: SRC_DIR, title: 'RHAII', ...overrides });
--- a/distributions/rhaii/config/webpack.common.js
+++ b/distributions/rhaii/config/webpack.common.js
@@ -2,6 +2,7 @@ const path = require('path');
 const createWebpackCommon = require('../../base/config/webpack.common.js');
 
 const SRC_DIR = path.resolve(__dirname, '../src');
+const TITLE = 'RHAII';
 
 module.exports = (overrides = {}) =>
-  createWebpackCommon({ distributionSrcDir: SRC_DIR, title: 'RHAII', ...overrides });
+  createWebpackCommon({ distributionSrcDir: SRC_DIR, title: TITLE, ...overrides });
--- a/distributions/rhaii/config/webpack.dev.js
+++ b/distributions/rhaii/config/webpack.dev.js
@@ -1,0 +1,51 @@
+const path = require('path');
+const { merge } = require('webpack-merge');
+const ForkTsCheckerWebpackPlugin = require('fork-ts-checker-webpack-plugin');
+const webpackCommon = require('./webpack.common.js');
+
+const RELATIVE_DIRNAME = path.resolve(__dirname, '..');
+const DIST_DIR = path.resolve(RELATIVE_DIRNAME, 'public');
+const PORT = process.env.SHELL_PORT || 4020;
+const BFF_PORT = process.env.BFF_PORT || 4000;
+
+module.exports = merge(webpackCommon(), {
+  mode: 'development',
+  devtool: 'eval-source-map',
+  optimization: {
+    runtimeChunk: 'single',
+    removeEmptyChunks: true,
+  },
+  devServer: {
+    host: 'localhost',
+    port: PORT,
+    compress: true,
+    historyApiFallback: true,
+    hot: true,
+    proxy: [
+      {
+        context: ['/api'],
+        target: `http://localhost:${BFF_PORT}`,
+      },
+      {
+        context: ['/wss'],
+        target: `ws://localhost:${BFF_PORT}`,
+        ws: true,
+      },
+    ],
+    client: {
+      overlay: true,
+    },
+    static: {
+      directory: DIST_DIR,
+    },
+    onListening: (devServer) => {
+      const addr = devServer?.server?.address();
+      if (addr) {
+        console.log(
+          `\x1b[32m✓ RHAII distribution available at: \x1b[4mhttp://localhost:${addr.port}\x1b[0m`,
+        );
+      }
+    },
+  },
+  plugins: [new ForkTsCheckerWebpackPlugin()],
+});
--- a/distributions/rhaii/config/webpack.dev.js
+++ b/distributions/rhaii/config/webpack.dev.js
@@ -41,9 +41,13 @@ module.exports = merge(webpackCommon(), {
     onListening: (devServer) => {
       const addr = devServer?.server?.address();
       if (addr) {
-        console.log(
-          `\x1b[32m✓ RHAII distribution available at: \x1b[4mhttp://localhost:${addr.port}\x1b[0m`,
-        );
+        const green = '\x1b[32m';
+        const underline = '\x1b[4m';
+        const reset = '\x1b[0m';
+        const url = `http://localhost:${addr.port}`;
+        console.log(`${green}✓ RHAII distribution available at: ${underline}${url}${reset}`);
+      } else {
+        console.warn('RHAII dev server started but could not determine address');
       }
     },
   },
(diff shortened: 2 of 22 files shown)


distributions/rhaii/config/webpack.common.js
```diff
@@ -0,0 +1,7 @@
+const path = require('path');
+const createWebpackCommon = require('../../base/config/webpack.common.js');
+
+const SRC_DIR = path.resolve(__dirname, '../src');
+
+module.exports = (overrides = {}) =>
+  createWebpackCommon({ distributionSrcDir: SRC_DIR, title: 'RHAII', ...overrides });
```
andyatmiami: trivial nit here - but i'd prefer the `RHAII` string to be defined on its own line (similar to `SRC_DIR` so its easier to "see")

distributions/rhaii/config/webpack.dev.js
```diff
@@ -0,0 +1,51 @@
...
+    client: {
+      overlay: true,
+    },
+    static: {
+      directory: DIST_DIR,
+    },
+    onListening: (devServer) => {
+      const addr = devServer?.server?.address();
+      if (addr) {
+        console.log(
```
andyatmiami: I don't love all these formatting sequences embedded inline as a single string.. could we use something like a string template with constants defined to make it more _understandable_ what is being done here

distributions/rhaii/config/webpack.dev.js
```diff
@@ -0,0 +1,51 @@
...
+          `\x1b[32m✓ RHAII distribution available at: \x1b[4mhttp://localhost:${addr.port}\x1b[0m`,
+        );
+      }
```
andyatmiami, line 51: should there be some `else` block here in the (unlikely) event `addr` is _false-y_ ?  seems like if that var is not set.. probably good thing to log

distributions/rhaii/src/plugins/useAppExtensions.ts
```diff
@@ -0,0 +1,43 @@
+import type { Extension } from '@openshift/dynamic-plugin-sdk';
+
+const scaffoldExtensions: Extension[] = [
```
andyatmiami, line 14: maybe a comment here explicitly declaring these 'scaffoldExtensions' are a "temporary" thing to demo until we have a real odh-dashboard extension ready to include in RHAII

distributions/rhaii/src/plugins/useAppExtensions.ts
andyatmiami, line 1: courtesy Claude: Extensions are untyped

The scaffold extensions are declared as `Extension[]` (the base type from `dynamic-plugin-sdk`). The codebase defines specific extension types — `NavSectionExtension`, `HrefNavItemExtension`, `RouteExtension` — in `packages/plugin-core/src/extension-points/`. Those types enforce the correct property shapes at compile time.

As written, nothing catches a typo like `titl` instead of `title`, or a missing required id field on a nav section. The component property on route extensions is typed as `ComponentCodeRef` (which is ```
CodeRef<{ 
  default: React.ComponentType }>
```) in the real types, but here it's `() => import(...)` — which happens to work at runtime but isn't type-checked against the actual contract.

Recommendation: Import and use the specific extension types, or at minimum use satisfies assertions per extension object. This is what a real plugin module would do (see module-onboarding.md examples), and the scaffold should model the pattern it's validating.

File: distributions/rhaii/src/plugins/ExtensibilityContext.tsx
Comment (andyatmiami) [inline distributions/rhaii/src/plugins/ExtensibilityContext.tsx:1]: courtesy Claude: `useAppExtensions` returns a tuple:` [extensions, loaded]`. RHAII destructures only the first element and throws away `loaded`

```
  // RHAII — ignores loaded
  const [appExtensions] = useAppExtensions();
  const store = React.useMemo(() => new PluginStore(appExtensions), [appExtensions]);
  return <PluginStoreProvider store={store}>{children}</PluginStoreProvider>;
```

```
  // Base — guards on loaded
  const [appExtensions, loaded] = useAppExtensions();
  const store = React.useMemo(
    () => (loaded ? new PluginStore(appExtensions) : null),
    [appExtensions, loaded],
  );
  if (!store) {
    return <Bullseye><Spinner /></Bullseye>;
  }
  return <PluginStoreProvider store={store}>{children}</PluginStoreProvider>;
```

Right now this doesn't matter because `useAppExtensions` always returns true. The store is always created with the final data. No bug today.

The concern is the contract between these two files. `useAppExtensions` promises a boolean that says "I might not be ready yet." `ExtensibilityContext` is the consumer of that promise. If the consumer ignores it, the contract is only held up by one side.

The scenario: someone adds async extension loading to RHAII's `useAppExtensions` (say, when model-serving gets wired in). They follow the same pattern as the main frontend — return false until extensions  resolve, then true. They'd reasonably expect the consumer to handle that, because every other `ExtensibilityContext` in the codebase does. But RHAII's doesn't — it would construct a `PluginStore` with an empty/partial extension map on the first render, pass it to every downstream consumer, and then reconstruct it once the real data arrives. The useMemo would fire twice with different inputs.

andyatmiami (changes requested):
This is looking really good! Some minor nits/suggestions here - but overall very happy with the state of this code!

distributions/rhaii/src/plugins/useAppExtensions.ts
jenny-s51: Good point, thank you - added a comment.

distributions/rhaii/src/plugins/ExtensibilityContext.tsx
jenny-s51: Thank you - restored the loaded guard in ExtensibilityContext.tsx.

distributions/rhaii/src/plugins/useAppExtensions.ts
jenny-s51: Nice catch, added `satisfies` assertions on every extension object so each one is now type-checked against its specific extension type.

jenny-s51:
Thank you for your review and feedback @andyatmiami - addressed your comments.

andyatmiami (approved):
/lgtm
/approve

Thanks for addressing all my questions/comments @jenny-s51.  Comfortable in the current state of this good being a good/accurate representation of how we see `base/` + `rhaii/` ownership/integration.

verified functional correctness by testing against a branch I have locally (which I will PR soon) that takes this code and has it served by the `core-bff` developed by @caponetto .  Everything appears to be working and no unexpected messages written to console, etc

Lets go! 🚀 

Image

PR #7882, opened 2026-06-05
RHOAIENG-36320: use KServe status.addresses for LLMInferenceService endpoint

gmenher:
https://issues.redhat.com/browse/RHOAIENG-36320

## Description

Replaces the manual URL construction workaround in `extractEndpointFromLLMInferenceService` with direct reading from KServe's `status.addresses[0].url`, enabling proper llm-d gateway routing for models deployed via `LLMInferenceService`.

**What changed (2 files, +121/-97):**

- **token_k8s_client.go — `extractEndpointFromLLMInferenceService`** — Rewritten to iterate `status.addresses` (plural), filtering for `.svc.cluster.local` entries, and applying `EnsureV1Suffix`. Falls back to `status.address` (singular) for compatibility. No longer calls `findServicesForKServeResource`, `getServingPort`, or `DetermineSchemeFromAuth` — all K8s service discovery is eliminated from this path. The `ctx` parameter is now unused (marked `_`) since no API calls are made.
- **token_k8s_client.go — `ConstructLLMInferenceServiceURL`** — Removed. This was the only caller, and the function manually constructed `{scheme}://{service}.{namespace}.svc.cluster.local:{port}/v1` which bypassed the llm-d gateway. `EnsureV1Suffix` and `DetermineSchemeFromAuth` are kept — they have other callers.
- **token_k8s_client_test.go — `TestExtractEndpointFromLLMInferenceService`** — New test (replaces `TestLLMInferenceServiceURLConstruction`) with 6 cases covering: internal URL with `/v1` appended (real cluster data from `tinyllama` on a GPU cluster), idempotent `/v1`, singular address fallback, empty status error, internal-over-external selection, and nil URL entries.
- **token_k8s_client_test.go — `TestGetModelDetailsFromServingRuntimeNilName`** — Updated `newLLMService` helper to populate `status.addresses` (required by the refactored function). Removed `workloadService` helper and its reference in the fake client builder — no longer needed since endpoint extraction doesn't discover K8s Services.

**Key design decisions:**

- **`status.addresses` (plural) as primary source** — this is what the KServe controller populates. Verified on a real cluster: only `addresses` (plural) has data; `address` (singular) and `url` are empty.
- **Filter by `.svc.cluster.local`** — `status.addresses` can contain both internal and external URLs. The BFF always makes internal calls, so we select the cluster-local one.
- **`EnsureV1Suffix` post-processing** — KServe URLs omit the `/v1` path that the vLLM OpenAI-compatible API requires. This function already existed in the codebase.
- **No changes to `findServicesForKServeResource` or `getServingPort`** — these are still used by the `InferenceService` (non-LLM) code path in `getModelDetailsFromServingRuntime`.

## How Has This Been Tested?

- 242 unit tests in the `kubernetes` package, all passing (`go test ./internal/integrations/kubernetes/... -count=1`).
- Full `go build ./...` compiles cleanly.
- Verified on a real GPU cluster (`gmenende-gpus`, 5x g4dn.xlarge with Tesla T4) that `LLMInferenceService` `tinyllama` has `status.addresses[0].url = https://tinyllama-kserve-workload-svc.kserve-test.svc.cluster.local` — test case 1 uses this exact value.

## Test Impact

- **`TestExtractEndpointFromLLMInferenceService`** — 6 new tests covering the refactored function: internal URL appends `/v1`, idempotent `/v1`, singular fallback, empty status error, multiple addresses picks internal, nil URL entries skipped.
- **`TestGetModelDetailsFromServingRuntimeNilName`** — 4 existing tests updated to work with status-based extraction (all pass).
- **`TestLLMInferenceServiceURLConstruction`** — Removed (tested the deleted `ConstructLLMInferenceServiceURL` function).

## Request review criteria:

> **Note to reviewers:** Manual validation on a live KServe + llm-d cluster is recommended to confirm that requests route through the gateway rather than directly to vLLM.

## Summary by CodeRabbit

* **Bug Fixes**
  * More reliable detection of LLM inference endpoints by reading service status addresses and preferring cluster-local URLs
  * Improved fallback when multiple/alternative address fields are present
  * Normalized endpoint formatting to ensure a consistent /v1 suffix
  * Better handling of missing or invalid endpoint entries (returns an error when none found)

* **Chores**
  * Removed a legacy URL-construction path used for testing

* **Tests**
  * Updated and expanded tests validating endpoint extraction behavior, address selection, suffix handling, and error cases

### DIFF
--- a/packages/gen-ai/bff/internal/integrations/kubernetes/token_k8s_client_test.go
+++ b/packages/gen-ai/bff/internal/integrations/kubernetes/token_k8s_client_test.go
@@ -336,44 +336,158 @@ func TestGenerateLlamaStackConfig_RBACFlag(t *testing.T) {
 	})
 }
 
-// TestLLMInferenceServiceURLConstruction tests that the URL format for LLMInferenceService
-// remains consistent and doesn't accidentally change
-func TestLLMInferenceServiceURLConstruction(t *testing.T) {
-	tests := []struct {
-		name        string
-		scheme      string
-		serviceName string
-		namespace   string
-		port        int32
-		expected    string
-	}{
-		{
-			name:        "http URL without auth",
-			scheme:      "http",
-			serviceName: "test-service",
-			namespace:   "test-namespace",
-			port:        8080,
-			expected:    "http://test-service.test-namespace.svc.cluster.local:8080/v1",
-		},
-		{
-			name:        "https URL with auth",
-			scheme:      "https",
-			serviceName: "secure-service",
-			namespace:   "prod-namespace",
-			port:        8443,
-			expected:    "https://secure-service.prod-namespace.svc.cluster.local:8443/v1",
-		},
+// TestExtractEndpointFromLLMInferenceService tests that extractEndpointFromLLMInferenceService
+// reads the internal URL from status.addresses instead of constructing it manually.
+func TestExtractEndpointFromLLMInferenceService(t *testing.T) {
+	client := &TokenKubernetesClient{
+		Logger: slog.Default(),
 	}
+	ctx := context.Background()
 
-	for _, tt := range tests {
-		t.Run(tt.name, func(t *testing.T) {
-			// Call the actual function used in extractEndpointFromLLMInferenceService
-			actual := ConstructLLMInferenceServiceURL(tt.scheme, tt.serviceName, tt.namespace, tt.port)
+	t.Run("status.addresses with internal URL appends /v1", func(t *testing.T) {
+		llmSvc := &kservev1alpha1.LLMInferenceService{
+			Status: kservev1alpha1.LLMInferenceServiceStatus{
+				AddressStatus: duckv1.AddressStatus{
+					Addresses: []duckv1.Addressable{
+						{URL: mustParseURL("https://tinyllama-kserve-workload-svc.kserve-test.svc.cluster.local")},
+					},
+				},
+			},
+		}
+		endpoint, err := client.extractEndpointFromLLMInferenceService(ctx, llmSvc)
+		assert.NoError(t, err)
+		assert.Equal(t, "https://tinyllama-kserve-workload-svc.kserve-test.svc.cluster.local/v1", endpoint)
+	})
 
-			assert.Equal(t, tt.expected, actual,
-				"LLMInferenceService URL format must remain: {scheme}://{service}.{namespace}.svc.cluster.local:{port}/v1")
-		})
-	}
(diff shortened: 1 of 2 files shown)


gmenher:
**Note on CI failure:** `ci/prow/odh-mod-arch-agent-ops-pr-image-mirror` fails with `unauthorized: access to the requested resource is not authorized` when mirroring to `quay.io/opendatahub/odh-mod-arch-agent-ops`. This is unrelated to the changes in this PR — appears to be a missing or expired credential for the agent-ops Quay repository.

gmenher:
The Prow job ci/prow/odh-mod-arch-agent-ops-pr-image-mirror is failing with an unauthorized error when mirroring to quay.io/opendatahub/odh-mod-arch-agent-ops. This appears to be a credential issue on the Quay side and is not related to the changes in this PR. Flagging it here for visibility.

packages/gen-ai/bff/internal/integrations/kubernetes/token_k8s_client_test.go
```diff
@@ -336,44 +336,111 @@ func TestGenerateLlamaStackConfig_RBACFlag(t *testing.T) {
...
+						{URL: mustParseURL("https://my-model.namespace.svc.cluster.local/v1")},
+					},
+				},
+			},
+		}
+		endpoint, err := client.extractEndpointFromLLMInferenceService(ctx, llmSvc)
+		assert.NoError(t, err)
+		assert.Equal(t, "https://my-model.namespace.svc.cluster.local/v1", endpoint)
+	})
+
```
jharan1, line 376: Maybe could add a test case where status.addresses has a URL with a port (e.g., https://svc.ns.svc.cluster.local:8443) - looks like the code handles it correctly, so just to lock it in & document.

gmenher:
Good suggestion, added a test case for URLs with explicit port (e.g. `https://svc.ns.svc.cluster.local:8443`) to confirm the port is preserved and `/v1` is appended correctly.

gmenher:
All CodeRabbit comments have been addressed in previous pushes:
- Substring spoofing: replaced `strings.Contains` with `isInternalClusterHost()` that validates the hostname suffix via `strings.HasSuffix`
- IPv6 bracket handling: `isInternalClusterHost()` now strips brackets before suffix check
- Singular fallback validation: added `.svc.cluster.local` check to the fallback path
- Added tests for all three cases (external URL in fallback, path spoofing, IPv6 loopback)

The unresolved threads are from earlier iterations and can be resolved.

jharan1 (approved):

PR #7883, opened 2026-06-05
Fix(RHOAIENG-66732):resolved the feature store token issue

claudialphonse78:
Closes [RHOAIENG-66732](https://redhat.atlassian.net/browse/RHOAIENG-66732)

## Description

Feature store K8s API calls were not respecting the logged-in user's namespace permissions, they were running as the dashboard service account instead.

The root cause was in `getUserScopedCustomObjectsApi`. The code tried to use the user's token by passing it in `opts.headers`, but the K8s client library ignores that and always uses the service account token from its internal config. So non-admin users were seeing data from namespaces they shouldn't have access to.
Fixed by building a new K8s client using the user's token directly, so RBAC is enforced.

## How Has This Been Tested?

Tested it with three users with dev impersonation
1. Cluster admin
Screenshot 2026-06-05 at 8 04 41 PM
2. Entities 

https://github.com/user-attachments/assets/9c219653-a3df-41fb-b6c0-7b0b4e47e85d

3. Sources
Before Fix incorrect feature store still visible to user
Screenshot 2026-06-05 at 8 28 08 PM

After fix

https://github.com/user-attachments/assets/f79392d8-ad80-4b07-a528-ac9284fac240

## Test Impact

1. Updated test cases

## Request review criteria:

If you have UI changes: 

## Summary by CodeRabbit

* **Tests**
  * Expanded coverage for feature store flows with improved Kubernetes client mocking, added edge-case tests for missing cluster/token, and updated impersonation behavior checks in DEV_MODE.

* **Refactor**
  * Switched to per-user API client construction so requests use the configured cluster and user token, enforcing presence of access tokens and clearer permission isolation.

### DIFF
--- a/backend/src/__tests__/featureStoreUtils.spec.ts
+++ b/backend/src/__tests__/featureStoreUtils.spec.ts
@@ -1,6 +1,8 @@
 import { EventEmitter } from 'events';
 import * as https from 'https';
 import * as k8s from '@kubernetes/client-node';
+import * as constants from '../utils/constants';
+import * as devFlags from '../devFlags';
 
 jest.mock('https', () => {
   const actual = jest.requireActual<typeof import('https')>('https');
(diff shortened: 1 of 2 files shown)


dpanshug (approved):
/lgtm tested with impersonating user, working fine

Image

PR #7897, opened 2026-06-08
feat: add config endpoints and connection-types handler to Core BFF

caponetto:
## Description

https://redhat.atlassian.net/browse/RHOAIENG-64749

Implements config endpoints and connection-types handler for the Core BFF (Task 4 from the Core BFF design doc). This adds on-demand fetching for DashboardConfig, Auth, Components, Status, Cluster Settings, Connection Types, and two sub-routes (components/remove, status/allowedUsers), replacing the corresponding Fastify endpoints with Go handlers. Admin auth uses an interim `requireAdmin` guard; full `secureAdminRoute` parity is deferred to Task 3.

### Backend (Go)

16 new endpoints across config, status, components, cluster-settings, dashboardConfig, and connection-types (full CRUD). Authenticated users can read config/components/status/connection-types; admin-only endpoints are gated by `requireAdmin` (SSAR check). See `routes.go` for the full list.

**Key decisions:**

- **Service account clients** for reading shared config and performing admin-gated mutations, matching Fastify's privileged watcher model
- **Admin guard**: interim `requireAdmin` via SSAR on `auths/default-auth`; full `secureAdminRoute` parity deferred to Task 3
- **On-demand fetching** with no caching or background watchers (design doc ADR D1)
- **Platform detection**: `ODH_PLATFORM_TYPE` (`XKS`, `OpenShift`, or unset for auto-detect via ClusterVersion CRD). XKS disables OpenShift-dependent features. Auto-detection distinguishes CRD-absent (genuinely non-OpenShift) from RBAC/network errors
- **Segment userID parity**: Status endpoint returns a privacy-safe `userID` matching Fastify's `getSegmentUserId` - SHA-256 of JWT `sub` claim with fallback to SHA-256 of username. DevSandbox SSO annotation path deferred as a narrow follow-up
- **Namespace** falls back to `OC_PROJECT` env var

### Frontend (React)

Reorganized the Core BFF test page (`MainPage.tsx`) into a two-column grid with dedicated card components exercising each Task 4 endpoint:

- **Left column**: Session & Status, Namespaces, K8s Watch
- **Right column**: Dashboard Config, Cluster Settings, Components, Connection Types, Allowed Users

Each card is self-contained with its own data fetching via a shared `useFetchJson` hook. BFF connectivity is provided via `BffStatusContext`. A namespace selector (backed by `NamespaceContext`) allows switching namespaces for the Allowed Users card.

## How Has This Been Tested?

- **Go unit tests**: 90+ tests covering handlers, repositories, middleware (admin guard, token validation), pure logic (deep merge, feature lockouts, PVC parsing, username decoding, segment userID hashing, JWT sub extraction, platform auto-detection error differentiation), and integration tests with fake K8s clients.
- **Contract tests**: 49 tests across 6 files (4 foundation + 1 OpenShift + 1 XKS), run separately per platform via `test:contract:openshift` and `test:contract:xks`. Foundation tests validate response schemas, admin enforcement (401 for non-admin), non-admin read access (200 for authenticated users), write path validation, and 404 catch-all. Platform-specific tests verify feature flag defaults (OpenShift) and XKS overrides.
- **Manual verification**: Tested against a live OpenShift/RHOAI cluster and a Kind cluster (both via `make dev-start-federated`), and locally with mocked clients (`make dev-start`). Verified all cards display correct data, CRD-absent degradation, admin-only endpoints, and namespace guard behavior.
- **Flakiness check**: `go test -count=2` passes with idempotent test fixtures.
- `go build`, `go vet`, `make lint` (golangci-lint) - zero issues.
- `npm run test:contract` - 49/49 pass on both OpenShift and XKS platforms.
- Frontend: `npm run test:lint` and `npm run test:type-check` - zero errors.

**Test page**:

Running on kind

Screenshot 2026-06-05 at 3 12 15 PM

Running on OpenShift

Screenshot 2026-06-05 at 1 48 10 PM

## Test Impact

- **New Go unit/integration test files** across `internal/api/` and `internal/repositories/` covering handlers, repositories, middleware, and pure logic
- **New contract test files** split by platform: `foundation/` (platform-agnostic), `openshift/` and `xks/` (platform-specific feature flag assertions)

## Request review criteria:

If you have UI changes: 

## Summary by CodeRabbit

* **New Features**
  * New BFF REST APIs for config, dashboard config (by namespace/name), cluster settings, status/allowed-users, components, and full connection-type CRUD; frontend cards and hooks for these endpoints plus a BFF status and namespace selector.

* **Platform**
  * Explicit OpenShift/XKS platform detection and configuration with matching behavior and CLI/env support; contract tests can run per-platform.

* **Documentation**
  * README and OpenAPI updated to reflect new endpoints and platform nuances.

* **Tests**
  * Expanded unit, integration, and contract test coverage across APIs and repositories.

### DIFF
--- a/distributions/core-bff/bff/README.md
+++ b/distributions/core-bff/bff/README.md
@@ -8,37 +8,49 @@ Backend-for-frontend providing core endpoints required by the dashboard UI.
 
 ## Scope
 
-This service exposes:
+This service exposes infrastructure endpoints, config/status endpoints, and resource CRUD:
 
-- GET `/healthcheck` - liveness probe (no auth)
-- GET `/api/v1/healthcheck` - health check (with auth middleware)
-- GET `/api/v1/user` - returns the authenticated user
-- GET `/api/v1/namespaces` - list namespaces (available only when DEV_MODE=true or mock k8s enabled)
+**Infrastructure:**
+- `GET /healthcheck` - liveness probe (no auth)
+- `GET /api/v1/healthcheck` - health check (with auth middleware)
+- `GET /api/v1/user` - authenticated user identity
+- `GET /api/v1/namespaces` - list namespaces (dev/mock mode only)
+- `/api/k8s/*` - K8s API pass-through proxy
+- `/wss/k8s/*` - WebSocket watch proxy
+
+**Config & Status:**
+- `GET /api/config` - merged DashboardConfig with defaults
+- `PATCH /api/config` - update DashboardConfig (admin)
+- `GET /api/status` - user session status and cluster info
+- `GET /api/status/:namespace/allowedUsers` - notebook users in a namespace (admin)
+- `GET /api/components` - OdhApplication list (empty when CRD absent)
+- `GET /api/components/remove` - remove app from enabled-apps ConfigMap (admin)
+- `GET/PATCH /api/dashboardConfig/:namespace/:name` - raw DashboardConfig CRUD (admin)
+- `GET/PUT /api/cluster-settings` - cluster settings with validation (admin)
+- `GET /api/connection-types` - list connection-type ConfigMaps
+- `GET /api/connection-types/:name` - get single connection type
+- `POST /api/connection-types` - create connection type (admin)
+- `PUT /api/connection-types/:name` - replace connection type (admin)
+- `PATCH /api/connection-types/:name` - patch connection type (admin)
+- `DELETE /api/connection-types/:name` - delete connection type (admin)
 
 ## Development
 
-Run the following command to build the BFF:
+Build the BFF:
 
 ```shell
 make build
 ```
 
-After building it, you can run our app with:
+Run with mock K8s client:
 
 ```shell
-make run
+make run PORT=4000 MOCK_K8S_CLIENT=true
 ```
 
-If you want to use a different port or mock kubernetes client you can run:
+Run with debug logging:
 
 ```shell
-make run PORT=8000 MOCK_K8S_CLIENT=true
-```
-
-If you want to change the log level on deployment, add the LOG_LEVEL argument when running, supported levels are: ERROR, WARN, INFO, DEBUG. The default level is INFO.
-
-```shell
-# Run with debug logging
 make run LOG_LEVEL=DEBUG
 ```
 
@@ -47,180 +59,118 @@ make run LOG_LEVEL=DEBUG
 | Flag | Env Var | Description |
 |------|---------|-------------|
 | `-port` | `PORT` | Listen port (default 4000) |
-| `-deployment-mode` | `DEPLOYMENT_MODE` | `standalone` or `federated` (default `standalone`) |
+| `-deployment-mode` | `DEPLOYMENT_MODE` | `standalone` (default) or `federated` |
+| `-platform-type` | `ODH_PLATFORM_TYPE` | `OpenShift` (probe cluster info), `XKS` (skip OpenShift detection), empty (auto-detect) |
 | `-dev-mode` | `DEV_MODE` | Enables relaxed behaviors (namespaces listing, etc.) |
-| `-mock-k8s-client` | `MOCK_K8S_CLIENT` | Use in-memory stub for namespace/user resolution |
+| `-mock-k8s-client` | `MOCK_K8S_CLIENT` | Use in-memory stub for K8s operations |
 | `-static-assets-dir` | `STATIC_ASSETS_DIR` | Directory to serve single-page frontend assets |
 | `-log-level` | `LOG_LEVEL` | ERROR, WARN, INFO, DEBUG (default INFO) |
 | `-allowed-origins` | `ALLOWED_ORIGINS` | Comma separated CORS origins |
 | `-auth-method` | `AUTH_METHOD` | `user_token` (default) or `disabled` (skips auth) |
 | `-auth-token-header` | `AUTH_TOKEN_HEADER` | Header to read token from (default `x-forwarded-access-token`) |
-| `-auth-token-prefix` | `AUTH_TOKEN_PREFIX` | Expected value prefix (default empty; use `Bearer` with standard `Authorization`) |
+| `-auth-token-prefix` | `AUTH_TOKEN_PREFIX` | Expected value prefix (default empty) |
 | `-cert-file` | (CLI only) | TLS certificate path (enables TLS when paired with key) |
 | `-key-file` | (CLI only) | TLS key path |
 | `-insecure-skip-verify` | `INSECURE_SKIP_VERIFY` | Skip upstream TLS verify (dev only) |
 | `-mock-bff-clients` | `MOCK_BFF_CLIENTS` | Use mock BFF clients (no real HTTP calls to other BFFs) |
-| `-namespace` | `NAMESPACE` | Kubernetes namespace where the dashboard is deployed (default `opendatahub`) |
+| `-namespace` | `NAMESPACE` / `OC_PROJECT` | K8s namespace for dashboard resources (default `opendatahub`, falls back to `OC_PROJECT`) |
+| `-workbench-namespace` | `WORKBENCH_NAMESPACE` | K8s namespace for workbenches (defaults to dashboard namespace) |
 | `-dashboard-config-name` | `DASHBOARD_CONFIG_NAME` | Name of the OdhDashboardConfig CR (default `odh-dashboard-config`) |
+| `-enabled-apps-cm` | `ENABLED_APPS_CM` | Name of the ConfigMap tracking enabled applications |
 | `-mf-remotes-config` | `MF_REMOTES_CONFIG` | Path to module federation remotes config file |
 
 TLS: If both `cert-file` and `key-file` are provided the server starts with HTTPS.
 
-## Running the linter locally
+## Platform Detection
 
-The BFF directory uses golangci-lint to combine multiple linters for a more comprehensive linting process. To install and run simply use:
+The BFF supports three platform modes via `ODH_PLATFORM_TYPE`:
+
+- **`XKS`** - Generic Kubernetes. Skips all OpenShift API calls at startup and applies XKS feature overrides.
+- **`OpenShift`** - Probes for ClusterVersion (cluster ID) and console-config (branding).
+- **Unset** (default) - Auto-detects by probing the ClusterVersion CRD. If found, behaves as OpenShift. If absent, behaves as XKS. The resolved platform drives both startup behavior and feature overrides.
+
+## Endpoints
+
+### Sample local calls
+
+When running with the mocked K8s client (`MOCK_K8S_CLIENT=true`), the user `user@example.com` has RBAC allowing all endpoints.
+
+```shell
+# Infrastructure
+curl -i localhost:4000/healthcheck
+curl -i -H "x-forwarded-access-token: FAKE_CLUSTER_ADMIN_TOKEN" localhost:4000/api/v1/user
+
+# Config & Status
+curl -i -H "x-forwarded-access-token: FAKE_CLUSTER_ADMIN_TOKEN" localhost:4000/api/config
+curl -i -H "x-forwarded-access-token: FAKE_CLUSTER_ADMIN_TOKEN" localhost:4000/api/status
+curl -i -H "x-forwarded-access-token: FAKE_CLUSTER_ADMIN_TOKEN" localhost:4000/api/cluster-settings
+curl -i -H "x-forwarded-access-token: FAKE_CLUSTER_ADMIN_TOKEN" localhost:4000/api/components
+curl -i -H "x-forwarded-access-token: FAKE_CLUSTER_ADMIN_TOKEN" localhost:4000/api/connection-types
+```
+
+### Admin Routes
+
+Admin endpoints use a `requireAdmin` wrapper that checks SSAR for `patch` on `auths/default-auth`. Non-admin users receive 401. Note: this does not yet fully match Fastify's `secureAdminRoute` semantics (e.g. namespace validation order).
+
+### Privilege Model
+
+The BFF uses two types of K8s clients:
+
+- **Service account clients** (`saDynClient`, `saClientset`) - for reading shared config (DashboardConfig, Auth, Components) and performing admin-gated mutations (connection types, cluster settings). Matches Fastify's privileged watcher model.
+- **Per-request user clients** - for SSAR checks (`IsUserAdmin`, `IsUserAllowed`) and token validation via `SelfSubjectReview`.
+
+### XKS Platform Overrides
+
+When the platform is XKS (explicit or auto-detected), the BFF disables OpenShift-dependent features:
+- `enablement` set to `false`
+- `disableProjects`, `disableBYONImageStream`, `disableISVBadges`, `disableAppLauncher`, `disablePipelines` set to `true`
+- `mlflow` set to `false`
+
+These run after feature lockouts (`disableFineTuning`, `mlflow`) and header overrides, so they take precedence.
+
+### Inter-BFF Communication
+
+The BFF includes a `bffclient` package (`internal/integrations/bffclient/`) for calling other BFF services in a multi-BFF pod deployment.
+
+## Linting
 
 ```shell
 make lint
 ```
 
-For more information on configuring golangci-lint see the [documentation](https://golangci-lint.run/).
+Uses golangci-lint. See the [documentation](https://golangci-lint.run/) for configuration.
 
 ## Building and Deploying
-
-Run the following command to build the BFF:
 
 ```shell
 make build
 ```
 
-The BFF binary will be inside the `bin` directory.
+The binary will be inside the `bin` directory.
 
-You can also build Docker images from the distribution root (`core-bff/`):
+Docker images from the distribution root (`core-bff/`):
 
 ```shell
 cd .. && make docker-build
 ```
 
-## Endpoints
-
-Four JSON endpoints are available plus static asset serving (index.html fallback):
-
-```text
-GET /healthcheck            (no auth)
-GET /api/v1/healthcheck     (with auth middleware)
-GET /api/v1/user
-GET /api/v1/namespaces      (dev / mock mode only)
-```
-
-### Sample local calls
-
-When running with the mocked Kubernetes client (MOCK_K8S_CLIENT=true), the user `user@example.com` has RBAC allowing all three endpoints.
-
-```shell
-curl -i localhost:4000/healthcheck
-curl -i -H "x-forwarded-access-token: FAKE_CLUSTER_ADMIN_TOKEN" localhost:4000/api/v1/user
-curl -i -H "x-forwarded-access-token: FAKE_CLUSTER_ADMIN_TOKEN" localhost:4000/api/v1/namespaces   # (dev / mock only)
-```
-
-### Inter-BFF Communication
-
-The BFF includes a `bffclient` package (`internal/integrations/bffclient/`) that provides the scaffolding for calling other BFF services in a multi-BFF pod deployment. The package is target-agnostic - teams wire up their own target BFF endpoints on top of this infrastructure.
-
-#### Architecture
-
-```
-+--------------------------------------------------------------+
-|                        ODH Dashboard Pod                     |
-+--------------------------------------------------------------+
-|  +--------------+  +--------------+  +------------------+   |
-|  |  Gen-AI BFF  |--|   MaaS BFF   |--|  Model Registry  |   |
-|  |    :8143     |  |    :8243     |  |   BFF :8043      |   |
-|  +--------------+  +--------------+  +------------------+   |
-|         |                  |                    |            |
-|  +--------------+          |                    |            |
-|  |  MLflow BFF  |----------+--------------------+            |
-|  |    :8343     |     Inter-BFF HTTP Calls                   |
-|  +--------------+  (K8s service DNS or localhost)            |
-+--------------------------------------------------------------+
-```
-
-#### Adding a BFF target
-
-1. Add target-specific config fields to `internal/config/environment.go` (e.g. `BFF<Target>ServiceName`, `BFF<Target>ServicePort`, etc.)
-2. Add corresponding CLI flags to `cmd/main.go`
-3. Apply config overrides in `NewApp()` in `internal/api/app.go`
-4. Create a handler in `internal/api/` using `bffclient.GetClient()` and `bffclient.AttachBFFClient()` middleware
-5. Wire routes in `Routes()`
-
-See the `bffclient` package README and the implementation spec for detailed guidance.
-
 ### Authentication modes
 
 Two modes are supported (flag `--auth-method` / env `AUTH_METHOD`):
 
-- **user_token** (default): extracts a bearer token from the configured header (default `x-forwarded-access-token`) and performs SelfSubjectAccessReview. This is the standard authentication method for ODH/RHOAI deployments and is recommended for most use cases including mock/development mode.
-- **disabled**: skips identity extraction entirely. Useful for local development or testing when no real auth is needed.
-
-### Overriding token header / prefix
-
-By default, the BFF expects the token in the `x-forwarded-access-token` header with no prefix. If using the standard `Authorization` header, set the prefix to `Bearer`.
-
-If you're integrating with a proxy or tool that uses a different header, you can override this behavior using environment variables or Makefile arguments.
-
-```shell
-make run AUTH_METHOD=user_token AUTH_TOKEN_HEADER=X-Forwarded-Access-Token AUTH_TOKEN_PREFIX=""
-```
-
-This will configure the BFF to extract the raw token from the following header:
-
-```shell
-X-Forwarded-Access-Token: <your-token>
-```
+- **user_token** (default): extracts a bearer token from the configured header and performs SelfSubjectReview.
+- **disabled**: skips identity extraction entirely. Useful for local development or testing.
 
 ### Enabling CORS
 
-When serving the UI directly from the BFF there is no need for any CORS headers to be served, by default they are turned off for security reasons.
-
-If you need to enable CORS for any reasons you can add origins to the allow-list in several ways:
-
-##### Via the make command
-
-Add the following parameter to your command: `ALLOWED_ORIGINS` this takes a comma separated list of origins to permit serving to, alterantively you can specify the value `*` to allow all origins, **Note this is not recommended in production deployments as it poses a security risk**
-
-Examples:
+CORS is disabled by default. Enable with `ALLOWED_ORIGINS`:
 
 ```shell
-# Allow only the origin http://example.com:8081
-make run ALLOWED_ORIGINS="http://example.com:8081"
-
-# Allow the origins http://example.com and http://very-nice.com
-make run ALLOWED_ORIGINS="http://example.com,http://very-nice.com"
-
-# Allow all origins
-make run ALLOWED_ORIGINS="*"
-
-# Explicitly disable CORS (default behaviour)
-make run ALLOWED_ORIGINS=""
-```
-
-#### Via environment variable
-
-Setting CORS via environment variable follows the same rules as using the Makefile, simply set the environment variable `ALLOWED_ORIGINS` with the same value as above.
-
-#### Via command line argument
-
-Setting CORS via command line arguments follows the same rules as using the Makefile. Simply add the `--allowed-origins=` flag to your command.
-
-Examples:
-
-```shell
-./bff --allowed-origins="http://my-domain.com,http://my-other-domain.com"
+make run ALLOWED_ORIGINS="http://localhost:3000"
+make run ALLOWED_ORIGINS="*"   # allow all (not for production)
 ```
 
 ### Disabling TLS verification (development only)
-
-For local installations with self-signed certificates, you may need to disable TLS certificate verification.
-
-**Kubernetes deployment:**
-
-```yaml
-env:
-  - name: INSECURE_SKIP_VERIFY
-    value: "true"
-```
-
-**Local development:**
 
 ```shell
 ./bin/bff --insecure-skip-verify
--- a/distributions/core-bff/bff/internal/api/allowed_users_handler.go
+++ b/distributions/core-bff/bff/internal/api/allowed_users_handler.go
@@ -1,0 +1,28 @@
+package api
+
+import (
+	"fmt"
+	"net/http"
+
+	"github.com/julienschmidt/httprouter"
+)
+
+// GetAllowedUsersHandler returns users with notebook access in a namespace.
+func (app *App) GetAllowedUsersHandler(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
+	namespace := ps.ByName("namespace")
+
+	if !app.isAllowedNamespace(namespace) {
+		app.forbiddenResponse(w, r, fmt.Errorf("request invalid against a resource from a non-dashboard namespace"))
+		return
+	}
+
+	users, err := app.repositories.AllowedUsers.GetAllowedUsers(r.Context(), namespace)
+	if err != nil {
+		app.serverErrorResponse(w, r, err)
+		return
+	}
+
+	if err := app.WriteJSON(w, http.StatusOK, users, nil); err != nil {
+		app.serverErrorResponse(w, r, err)
+	}
+}
--- a/distributions/core-bff/bff/internal/api/cluster_info.go
+++ b/distributions/core-bff/bff/internal/api/cluster_info.go
@@ -2,10 +2,13 @@ package api
 
 import (
 	"context"
+	"fmt"
 	"log/slog"
+	"strings"
 	"time"
 
 	"gopkg.in/yaml.v3"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/dynamic"
@@ -30,35 +33,58 @@ var clusterVersionGVR = schema.GroupVersionResource{
 type clusterInfo struct {
 	clusterID       string
 	clusterBranding string
+	serverURL       string
+	currentContext  string
 }
 
-func queryClusterInfo(client kubernetes.Interface, dynClient dynamic.Interface, logger *slog.Logger) clusterInfo {
+func (ci clusterInfo) GetClusterID() string       { return ci.clusterID }
+func (ci clusterInfo) GetClusterBranding() string { return ci.clusterBranding }
+func (ci clusterInfo) GetServerURL() string       { return ci.serverURL }
+func (ci clusterInfo) GetCurrentContext() string  { return ci.currentContext }
+
+func queryClusterInfo(client kubernetes.Interface, dynClient dynamic.Interface, logger *slog.Logger) (clusterInfo, error) {
 	info := clusterInfo{clusterBranding: defaultClusterBranding}
-	info.clusterID = queryClusterID(dynClient, logger)
+	var err error
+	info.clusterID, err = queryClusterID(dynClient, logger)
+	if err != nil {
+		return info, err
+	}
 	info.clusterBranding = queryClusterBranding(client, logger)
-	return info
+	return info, nil
 }
 
-func queryClusterID(dynClient dynamic.Interface, logger *slog.Logger) string {
+func queryClusterID(dynClient dynamic.Interface, logger *slog.Logger) (string, error) {
 	if dynClient == nil {
-		return ""
+		return "", nil
 	}
 	ctx, cancel := context.WithTimeout(context.Background(), clusterInfoQueryTimeout)
 	defer cancel()
 
 	obj, err := dynClient.Resource(clusterVersionGVR).Get(ctx, clusterVersionName, metav1.GetOptions{})
 	if err != nil {
-		logger.Warn("ClusterVersion not available (expected on non-OpenShift clusters)", slog.Any("error", err))
-		return ""
+		if k8serrors.IsNotFound(err) || isClusterVersionDiscoveryError(err) {
+			logger.Debug("ClusterVersion CRD not available (non-OpenShift cluster)", slog.Any("error", err))
+			return "", nil
+		}
+		return "", fmt.Errorf("ClusterVersion probe failed: %w", err)
 	}
 
 	spec, ok := obj.Object["spec"].(map[string]any)
 	if !ok {
 		logger.Warn("ClusterVersion spec field missing or unexpected type")
-		return ""
+		return "", nil
 	}
 	clusterID, _ := spec["clusterID"].(string)
-	return clusterID
+	return clusterID, nil
+}
+
+func isClusterVersionDiscoveryError(err error) bool {
+	if k8serrors.IsMethodNotSupported(err) {
+		return true
+	}
+	msg := err.Error()
+	return strings.Contains(msg, "no matches for kind") ||
+		strings.Contains(msg, "the server could not find the requested resource")
 }
 
 // consoleConfig is the minimal structure needed to extract branding from the console-config ConfigMap.
@@ -77,7 +103,7 @@ func queryClusterBranding(client kubernetes.Interface, logger *slog.Logger) stri
 
 	cm, err := client.CoreV1().ConfigMaps(consoleConfigNamespace).Get(ctx, consoleConfigName, metav1.GetOptions{})
 	if err != nil {
-		logger.Warn("console-config ConfigMap not available (expected on non-OpenShift clusters)", slog.Any("error", err))
+		logger.Debug("console-config ConfigMap not available (non-OpenShift cluster)", slog.Any("error", err))
 		return defaultClusterBranding
 	}
 
--- a/distributions/core-bff/bff/internal/api/cluster_settings_handler.go
+++ b/distributions/core-bff/bff/internal/api/cluster_settings_handler.go
@@ -1,0 +1,107 @@
+package api
+
+import (
+	"errors"
+	"fmt"
+	"net/http"
+
+	"github.com/julienschmidt/httprouter"
+	"github.com/opendatahub-io/odh-dashboard/distributions/core-bff/bff/internal/models"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+// GetClusterSettingsHandler returns cluster settings derived from DashboardConfig and ConfigMaps.
+func (app *App) GetClusterSettingsHandler(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
+	ctx := r.Context()
+
+	dashConfig, err := app.repositories.DashboardConfig.GetDashboardConfig(
+		ctx, app.config.Namespace, app.config.DashboardConfigName, nil,
+	)
+	if err != nil {
+		app.serverErrorResponse(w, r, err)
+		return
+	}
+
+	settings, err := app.repositories.ClusterSettings.GetClusterSettings(ctx, app.config.Namespace, dashConfig)
+	if err != nil {
+		app.serverErrorResponse(w, r, err)
+		return
+	}
+
+	if err := app.WriteJSON(w, http.StatusOK, settings, nil); err != nil {
+		app.serverErrorResponse(w, r, err)
+	}
+}
+
+// UpdateClusterSettingsHandler updates cluster settings.
+func (app *App) UpdateClusterSettingsHandler(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
+	ctx := r.Context()
+
+	var settings models.ClusterSettings
+	if err := app.ReadJSON(w, r, &settings); err != nil {
+		app.badRequestResponse(w, r, err)
+		return
+	}
+
+	if err := validateClusterSettings(&settings); err != nil {
+		app.badRequestResponse(w, r, err)
+		return
+	}
+
+	currentConfig, err := app.repositories.DashboardConfig.GetDashboardConfig(
+		ctx, app.config.Namespace, app.config.DashboardConfigName, nil,
+	)
+	if err != nil {
+		app.serverErrorResponse(w, r, fmt.Errorf("failed to get current config: %w", err))
+		return
+	}
+
+	patchBytes, err := app.repositories.ClusterSettings.BuildDashboardConfigPatch(&settings, currentConfig)
+	if err != nil {
+		writeClusterSettingsError(app, w, r, fmt.Sprintf("failed to build config patch: %v", err))
+		return
+	}
+
+	if _, err := app.repositories.DashboardConfig.PatchRawDashboardConfig(
+		ctx, app.config.Namespace, app.config.DashboardConfigName, patchBytes, types.MergePatchType,
+	); err != nil {
+		writeClusterSettingsError(app, w, r, fmt.Sprintf("unable to update cluster settings: %v", err))
+		return
+	}
+
+	if err := app.repositories.ClusterSettings.UpdateConfigMaps(ctx, app.config.Namespace, &settings, currentConfig); err != nil {
+		writeClusterSettingsError(app, w, r, fmt.Sprintf("unable to update cluster settings: %v", err))
+		return
+	}
+
+	result := &models.MutationResponse{Success: true}
+	if err := app.WriteJSON(w, http.StatusOK, result, nil); err != nil {
+		app.serverErrorResponse(w, r, err)
+	}
+}
+
+func writeClusterSettingsError(app *App, w http.ResponseWriter, r *http.Request, msg string) {
+	result := &models.MutationResponse{Success: false, Error: msg}
+	if err := app.WriteJSON(w, http.StatusOK, result, nil); err != nil {
+		app.serverErrorResponse(w, r, err)
+	}
+}
+
+var validDeploymentStrategies = map[string]bool{
+	"":         true,
+	"rolling":  true,
+	"recreate": true,
+}
+
+func validateClusterSettings(s *models.ClusterSettings) error {
+	if s.PVCSize <= 0 {
+		return errors.New("pvcSize must be greater than 0")
+	}
+	if s.CullerTimeout < 0 {
+		return errors.New("cullerTimeout must not be negative")
+	}
+	if !validDeploymentStrategies[s.DefaultDeploymentStrategy] {
+		return fmt.Errorf("invalid defaultDeploymentStrategy: %q", s.DefaultDeploymentStrategy)
+	}
+	return nil
+}
--- a/distributions/core-bff/bff/internal/api/cluster_settings_handler.go
+++ b/distributions/core-bff/bff/internal/api/cluster_settings_handler.go
@@ -100,6 +100,9 @@ func validateClusterSettings(s *models.ClusterSettings) error {
 	if s.CullerTimeout < 0 {
 		return errors.New("cullerTimeout must not be negative")
 	}
+	if s.CullerTimeout > 0 && s.CullerTimeout%60 != 0 {
+		return errors.New("cullerTimeout must be a multiple of 60 seconds")
+	}
 	if !validDeploymentStrategies[s.DefaultDeploymentStrategy] {
 		return fmt.Errorf("invalid defaultDeploymentStrategy: %q", s.DefaultDeploymentStrategy)
 	}
--- a/distributions/core-bff/bff/internal/api/components_handler.go
+++ b/distributions/core-bff/bff/internal/api/components_handler.go
@@ -1,0 +1,49 @@
+package api
+
+import (
+	"fmt"
+	"net/http"
+
+	"github.com/julienschmidt/httprouter"
+)
+
+// GetComponentsHandler lists OdhApplication CRDs. Returns empty array when CRD is absent.
+func (app *App) GetComponentsHandler(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
+	if err := app.validateCallerToken(r.Context()); err != nil {
+		app.unauthorizedResponse(w, r, err)
+		return
+	}
+
+	installedOnly := r.URL.Query().Get("installed") != ""
+
+	components, err := app.repositories.Components.ListComponents(r.Context(), app.config.Namespace, installedOnly)
+	if err != nil {
+		app.serverErrorResponse(w, r, err)
+		return
+	}
+
+	if err := app.WriteJSON(w, http.StatusOK, components, nil); err != nil {
+		app.serverErrorResponse(w, r, err)
+	}
+}
+
+// RemoveComponentHandler removes an app entry from the enabled-apps ConfigMap.
+func (app *App) RemoveComponentHandler(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
+	appName := r.URL.Query().Get("appName")
+	if appName == "" {
+		app.badRequestResponse(w, r, fmt.Errorf("missing required query parameter: appName"))
+		return
+	}
+
+	result, err := app.repositories.Components.RemoveComponent(
+		r.Context(), app.config.Namespace, appName, app.config.EnabledAppsCM,
+	)
+	if err != nil {
+		app.serverErrorResponse(w, r, err)
+		return
+	}
+
+	if err := app.WriteJSON(w, http.StatusOK, result, nil); err != nil {
+		app.serverErrorResponse(w, r, err)
+	}
+}
--- a/distributions/core-bff/bff/internal/api/config_handler.go
+++ b/distributions/core-bff/bff/internal/api/config_handler.go
@@ -1,0 +1,79 @@
+package api
+
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+
+	"github.com/julienschmidt/httprouter"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+// GetConfigHandler returns the merged DashboardConfig with defaults.
+func (app *App) GetConfigHandler(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
+	ctx := r.Context()
+
+	if err := app.validateCallerToken(ctx); err != nil {
+		app.unauthorizedResponse(w, r, err)
+		return
+	}
+
+	var featureFlagOverrides map[string]bool
+	if flagsHeader := r.Header.Get("x-odh-feature-flags"); flagsHeader != "" {
+		featureFlagOverrides = make(map[string]bool)
+		if err := json.Unmarshal([]byte(flagsHeader), &featureFlagOverrides); err != nil {
+			// Silently ignore malformed header
+			featureFlagOverrides = nil
+		}
+	}
+
+	config, err := app.repositories.DashboardConfig.GetDashboardConfig(
+		ctx, app.config.Namespace, app.config.DashboardConfigName, featureFlagOverrides,
+	)
+	if err != nil {
+		app.serverErrorResponse(w, r, err)
+		return
+	}
+
+	w.Header().Set("Cache-Control", "no-cache")
+	if err := app.WriteJSON(w, http.StatusOK, config, nil); err != nil {
+		app.serverErrorResponse(w, r, err)
+	}
+}
+
+// PatchConfigHandler updates the default DashboardConfig CRD.
+func (app *App) PatchConfigHandler(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
+	ctx := r.Context()
+
+	body, err := io.ReadAll(r.Body)
+	if err != nil {
+		app.badRequestResponse(w, r, fmt.Errorf("failed to read request body: %w", err))
+		return
+	}
+
+	_, err = app.repositories.DashboardConfig.PatchRawDashboardConfig(
+		ctx, app.config.Namespace, app.config.DashboardConfigName, body, types.MergePatchType,
+	)
+	if err != nil {
+		if isResourceUnavailable(err) {
+			app.notFoundResponse(w, r)
+		} else {
+			app.serverErrorResponse(w, r, err)
+		}
+		return
+	}
+
+	// Re-fetch the merged config to return the updated state
+	config, err := app.repositories.DashboardConfig.GetDashboardConfig(
+		ctx, app.config.Namespace, app.config.DashboardConfigName, nil,
+	)
+	if err != nil {
+		app.serverErrorResponse(w, r, err)
+		return
+	}
+
+	if err := app.WriteJSON(w, http.StatusOK, config, nil); err != nil {
+		app.serverErrorResponse(w, r, err)
+	}
+}
--- a/distributions/core-bff/bff/internal/api/config_handler.go
+++ b/distributions/core-bff/bff/internal/api/config_handler.go
@@ -46,9 +46,15 @@ func (app *App) GetConfigHandler(w http.ResponseWriter, r *http.Request, _ httpr
 func (app *App) PatchConfigHandler(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
 	ctx := r.Context()
 
+	r.Body = http.MaxBytesReader(w, r.Body, 1_048_576)
 	body, err := io.ReadAll(r.Body)
 	if err != nil {
 		app.badRequestResponse(w, r, fmt.Errorf("failed to read request body: %w", err))
+		return
+	}
+
+	if !json.Valid(body) {
+		app.badRequestResponse(w, r, fmt.Errorf("body contains badly-formed JSON"))
 		return
 	}
 
--- a/distributions/core-bff/bff/internal/api/connection_type_handler.go
+++ b/distributions/core-bff/bff/internal/api/connection_type_handler.go
@@ -101,6 +101,7 @@ func (app *App) UpdateConnectionTypeHandler(w http.ResponseWriter, r *http.Reque
 func (app *App) PatchConnectionTypeHandler(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
 	name := ps.ByName("name")
 
+	r.Body = http.MaxBytesReader(w, r.Body, 1_048_576)
 	body, err := io.ReadAll(r.Body)
 	if err != nil {
 		app.badRequestResponse(w, r, fmt.Errorf("failed to read request body: %w", err))
--- a/distributions/core-bff/bff/internal/api/connection_type_handler.go
+++ b/distributions/core-bff/bff/internal/api/connection_type_handler.go
@@ -1,0 +1,136 @@
+package api
+
+import (
+	"fmt"
+	"io"
+	"net/http"
+	"strings"
+
+	"github.com/julienschmidt/httprouter"
+	"github.com/opendatahub-io/odh-dashboard/distributions/core-bff/bff/internal/models"
+	corev1 "k8s.io/api/core/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+)
+
+// ListConnectionTypesHandler lists connection-type ConfigMaps.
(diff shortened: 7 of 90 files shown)


distributions/core-bff/bff/internal/api/allowed_users_handler.go
```diff
@@ -0,0 +1,28 @@
...
+	"github.com/julienschmidt/httprouter"
+)
+
+// GetAllowedUsersHandler returns users with notebook access in a namespace.
+func (app *App) GetAllowedUsersHandler(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
+	namespace := ps.ByName("namespace")
+
+	if !app.isAllowedNamespace(namespace) {
+		app.forbiddenResponse(w, r, fmt.Errorf("request invalid against a resource from a non-dashboard namespace"))
+		return
```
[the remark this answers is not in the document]
caponetto: Code correct; isAllowedNamespace already tested. Test addition out of scope.

distributions/core-bff/bff/internal/api/cluster_settings_handler.go
```diff
@@ -0,0 +1,107 @@
...
+	if _, err := app.repositories.DashboardConfig.PatchRawDashboardConfig(
+		ctx, app.config.Namespace, app.config.DashboardConfigName, patchBytes, types.MergePatchType,
+	); err != nil {
+		writeClusterSettingsError(app, w, r, fmt.Sprintf("unable to update cluster settings: %v", err))
+		return
+	}
+
+	if err := app.repositories.ClusterSettings.UpdateConfigMaps(ctx, app.config.Namespace, &settings, currentConfig); err != nil {
+		writeClusterSettingsError(app, w, r, fmt.Sprintf("unable to update cluster settings: %v", err))
+		return
```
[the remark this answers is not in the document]
caponetto: Fastify parity - MutationResponse includes K8s error messages in Fastify too.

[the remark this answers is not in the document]
caponetto: Fastify parity - same ordering. Rollback adds complexity beyond scope.

distributions/core-bff/bff/internal/api/components_handler.go
```diff
@@ -0,0 +1,49 @@
...
+)
+
+// GetComponentsHandler lists OdhApplication CRDs. Returns empty array when CRD is absent.
+func (app *App) GetComponentsHandler(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
+	if err := app.validateCallerToken(r.Context()); err != nil {
+		app.unauthorizedResponse(w, r, err)
+		return
+	}
+
+	installedOnly := r.URL.Query().Get("installed") != ""
```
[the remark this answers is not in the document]
caponetto: Fastify parity - presence-based check (?installed) matches Fastify behavior.

distributions/core-bff/bff/internal/api/components_handler.go
```diff
@@ -0,0 +1,49 @@
...
+func (app *App) RemoveComponentHandler(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
+	appName := r.URL.Query().Get("appName")
+	if appName == "" {
+		app.badRequestResponse(w, r, fmt.Errorf("missing required query parameter: appName"))
+		return
+	}
+
+	result, err := app.repositories.Components.RemoveComponent(
+		r.Context(), app.config.Namespace, appName, app.config.EnabledAppsCM,
+	)
```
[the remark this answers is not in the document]
caponetto: Fastify parity - Fastify registers /api/components/remove as GET. Changing breaks frontend + contract tests.

distributions/core-bff/bff/internal/api/errors.go
```diff
@@ -52,6 +52,12 @@ func (app *App) serverErrorResponse(w http.ResponseWriter, r *http.Request, err
 	app.errorResponse(w, r, httpError)
 }
 
+func (app *App) forbiddenResponse(w http.ResponseWriter, r *http.Request, err error) {
+	app.logger.Warn("Forbidden access attempt", "error", err.Error(), "method", r.Method, "uri", r.URL.RequestURI())
+	httpError := &HTTPError{StatusCode: http.StatusForbidden, Error: ErrorPayload{Code: "FORBIDDEN", Message: err.Error()}}
+	app.errorResponse(w, r, httpError)
```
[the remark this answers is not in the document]
caponetto: Messages passed are controlled strings ("request invalid against a resource from..."), not raw K8s errors.

distributions/core-bff/bff/internal/api/middleware.go
```diff
@@ -68,12 +68,70 @@ func (app *App) InjectRequestIdentity(next http.Handler) http.Handler {
...
+			return
+		}
+
+		isAdmin, err := client.IsUserAdmin(ctx, identity)
+		if err != nil {
+			app.logger.Warn("admin SAR check failed, denying by default", slog.Any("error", err))
+		}
+		if err != nil || !isAdmin {
+			app.unauthorizedResponse(w, r, fmt.Errorf("you lack the sufficient permissions to make this request"))
+			return
```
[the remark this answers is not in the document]
caponetto: Documented as 401; TODO exists for secureAdminRoute migration (task3). Contract tests expect 401.

distributions/core-bff/frontend/src/app/hooks/useFetchJson.ts
```diff
@@ -0,0 +1,47 @@
...
+        const resp = await fetch(`${URL_PREFIX}${path}`);
+        if (!resp.ok) {
+          if (!cancelled) {
+            setState({ data: null, loaded: true, error: `HTTP ${resp.status}` });
+          }
+          return;
+        }
+        const data: T = await resp.json();
+        if (!cancelled) {
+          setState({ data, loaded: true });
```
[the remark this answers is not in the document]
caponetto: Adding zod/ajv is a massive scope change to a test code.

distributions/core-bff/frontend/src/app/context/NamespaceContext.tsx
```diff
@@ -0,0 +1,66 @@
...
+  useEffect(() => {
+    if (!selected && statusData?.kube.namespace) {
+      setSelected(statusData.kube.namespace);
+    }
+  }, [statusData, selected]);
+
+  useEffect(() => {
+    if (namespaces.length > 0 && selected && !namespaces.includes(selected)) {
+      setSelected(namespaces[0]);
+    }
```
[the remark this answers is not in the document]
caponetto: Current logic is correct for the use case.

distributions/core-bff/frontend/src/app/components/SessionStatusCard.tsx
```diff
@@ -0,0 +1,113 @@
...
+        )}
+      </DescriptionListDescription>
+    </DescriptionListGroup>
+    <DescriptionListGroup>
+      <DescriptionListTerm>Cluster</DescriptionListTerm>
+      <DescriptionListDescription>
+        {kube.clusterBranding}{' '}
+        <Label color="blue" isCompact>
+          {kube.clusterID ? `${kube.clusterID.slice(0, 8)}...` : 'no ID'}
+        </Label>
```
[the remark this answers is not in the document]
caponetto: clusterID is always a string from the API.

distributions/core-bff/frontend/src/app/components/NamespaceSelect.tsx
```diff
@@ -0,0 +1,46 @@
...
+    </MenuToggle>
+  );
+
+  return (
+    <Select
+      isOpen={isOpen}
+      selected={selectedNamespace}
+      onSelect={(_event, value) => {
+        setSelectedNamespace(String(value));
+        setIsOpen(false);
```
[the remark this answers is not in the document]
caponetto: PatternFly Select provides typed values.

distributions/core-bff/frontend/src/app/components/DashboardConfigCard.tsx
```diff
@@ -0,0 +1,100 @@
...
+  { key: 'mlflow', label: 'MLflow' },
+];
+
+const FlagLabels: React.FC<{ flags: Record<string, boolean> }> = ({ flags }) => (
+  <LabelGroup>
+    {FLAGS.map(({ key, label, inverted }) => {
+      const enabled = inverted ? !flags[key] : flags[key];
+      return (
+        <Label key={key} color={enabled ? 'green' : 'grey'} isCompact>
+          {label}: {enabled ? 'enabled' : 'disabled'}
```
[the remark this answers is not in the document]
caponetto: Flags are booleans by type. Truthiness check is correct.

distributions/core-bff/frontend/src/app/components/ClusterSettingsCard.tsx
```diff
@@ -0,0 +1,131 @@
...
+      </Label>
+    );
+  }
+  if (!loaded) {
+    return <Spinner size="md" />;
+  }
+  if (settings) {
+    return <SettingsDetails settings={settings} />;
+  }
+  return null;
```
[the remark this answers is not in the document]
caponetto: API is typed, always returns valid data.

distributions/core-bff/frontend/src/app/components/ClusterSettingsCard.tsx
```diff
@@ -0,0 +1,131 @@
...
+    </DescriptionListGroup>
+    {settings.defaultDeploymentStrategy ? (
+      <DescriptionListGroup>
+        <DescriptionListTerm>Strategy</DescriptionListTerm>
+        <DescriptionListDescription>
+          {settings.defaultDeploymentStrategy}
+        </DescriptionListDescription>
+      </DescriptionListGroup>
+    ) : null}
+  </DescriptionList>
```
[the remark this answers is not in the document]
caponetto: API is typed, always returns valid data.

distributions/core-bff/bff/internal/helpers/k8s.go
```diff
@@ -35,6 +35,19 @@ func getKubeconfig(inClusterFn func() (*clientRest.Config, error)) (*clientRest.
...
+// GetCurrentContext returns the active kubeconfig context name.
+// Returns "inClusterContext" when running inside a cluster (no kubeconfig file).
+func GetCurrentContext() string {
+	loadingRules := clientcmd.NewDefaultClientConfigLoadingRules()
+	rawConfig, err := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(
+		loadingRules, &clientcmd.ConfigOverrides{},
+	).RawConfig()
+	if err != nil || rawConfig.CurrentContext == "" {
+		return "inClusterContext"
+	}
```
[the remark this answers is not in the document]
caponetto: Non-critical display string for KubeStatus.CurrentContext. Real kubeconfig validation happens in GetKubeconfig() which already fails hard on invalid configs.

distributions/core-bff/bff/internal/models/cluster_settings.go
```diff
@@ -0,0 +1,31 @@
...
+	DefaultDeploymentStrategy       string                      `json:"defaultDeploymentStrategy,omitempty"`
+}
+
+// DefaultClusterSettings provides sensible defaults when no CRD or ConfigMap values are available.
+var DefaultClusterSettings = ClusterSettings{
+	PVCSize:                         20,
+	CullerTimeout:                   31536000,
+	UserTrackingEnabled:             false,
+	IsDistributedInferencingDefault: boolPtr(true),
+	ModelServingPlatformEnabled: ModelServingPlatformEnabled{
```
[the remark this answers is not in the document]
caponetto: The global is used via value copy (settings := models.DefaultClusterSettings), which copies the struct. The shared *bool pointer is never dereferenced for writing - applyDashboardConfigToSettings replaces the pointer itself, not the pointed-to value. No mutation risk in practice.

distributions/core-bff/bff/internal/models/dashboard_config.go
```diff
@@ -0,0 +1,193 @@
...
+		},
+		TemplateOrder: []string{},
+		GenAiStudioConfig: &GenAiStudioConfig{
+			AiAssetCustomEndpoints: &AiAssetCustomEndpointsConfig{
+				ExternalProviders: false,
+				ClusterDomains:    []string{},
+			},
+		},
+	},
+}
```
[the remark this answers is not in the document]
caponetto: All access goes through blankDefaults() which creates deep copies via JSON marshal/unmarshal. No code mutates the global directly.

distributions/core-bff/bff/internal/repositories/auth_test.go
```diff
@@ -0,0 +1,65 @@
...
+		for _, g := range groups {
+			if s, ok := g.(string); ok {
+				auth.AllowedGroups = append(auth.AllowedGroups, s)
+			}
+		}
+	}
+
+	assert.Equal(t, []string{"admin-group-1", "admin-group-2"}, auth.AdminGroups)
+	assert.Equal(t, []string{"system:authenticated", "allowed-group"}, auth.AllowedGroups)
+}
```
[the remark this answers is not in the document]
caponetto: Tests should verify production code, not reuse it. Using extractStringSlice in the test that's testing the code that calls extractStringSlice is circular.

distributions/core-bff/bff/internal/repositories/cluster_settings.go
```diff
@@ -0,0 +1,167 @@
...
+	namespace string, settings *models.ClusterSettings, currentConfig *models.DashboardConfig,
+) error {
+	needsRollout := r.detectChanges(ctx, namespace, settings, currentConfig)
+
+	if err := r.updateCullerConfig(ctx, namespace, settings.CullerTimeout); err != nil {
+		return fmt.Errorf("failed to update culler config: %w", err)
+	}
+	if err := r.updateSegmentKeyConfig(ctx, namespace, settings.UserTrackingEnabled); err != nil {
+		return fmt.Errorf("failed to update segment key config: %w", err)
+	}
```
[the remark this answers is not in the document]
caponetto: The handler already validates via validateClusterSettings() before calling UpdateConfigMaps. Adding it again in the repository would be redundant.

distributions/core-bff/bff/internal/repositories/connection_type.go
```diff
@@ -0,0 +1,169 @@
...
+			Error:   "configmap must have connection-type labels",
+		}, nil
+	}
+
+	if err := r.validateExistingConnectionType(ctx, namespace, name); err != nil {
+		return err, nil
+	}
+
+	_, updateErr := r.saClientset.CoreV1().ConfigMaps(namespace).Update(ctx, cm, metav1.UpdateOptions{})
+	if updateErr != nil {
```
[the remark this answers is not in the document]
caponetto: Already handled - the handler sets cm.Name = name on line 89 of connection_type_handler.go before calling the repository.

distributions/core-bff/bff/internal/repositories/dashboard_config.go
```diff
@@ -0,0 +1,177 @@
...
+		retryResult, retryErr := r.saDynClient.Resource(models.DashboardConfigGVR).Namespace(namespace).Get(ctx, name, metav1.GetOptions{})
+		if retryErr == nil {
+			return retryResult.Object, nil
+		}
+	}
+
+	slog.Debug("failed to auto-create OdhDashboardConfig, falling back to defaults",
+		slog.String("namespace", namespace), slog.String("name", name), slog.Any("error", createErr))
+	return blankDefaults()
+}
```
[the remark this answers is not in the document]
caponetto: Intentional degradation. Returning defaults when auto-create fails keeps the dashboard functional. If the user lacks RBAC to create the CR, breaking the entire dashboard is worse than serving sensible defaults.

distributions/core-bff/bff/internal/repositories/allowed_users.go
```diff
@@ -0,0 +1,114 @@
...
+
+// GetAllowedUsers lists Notebook CRDs in the namespace and extracts user info.
+// Returns empty slice when CRD is absent.
+func (r *AllowedUsersRepository) GetAllowedUsers(ctx context.Context, namespace string) ([]models.AllowedUser, error) {
+	if r.saDynClient == nil {
+		return []models.AllowedUser{}, nil
+	}
+
+	list, err := r.saDynClient.Resource(models.NotebookGVR).Namespace(namespace).List(ctx, metav1.ListOptions{})
+	if err != nil {
```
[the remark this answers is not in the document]
caponetto: The handler at line 14 already validates the namespace via app.isAllowedNamespace(namespace) before calling GetAllowedUsers. The repository never receives an unvalidated namespace.

distributions/core-bff/bff/internal/repositories/auth.go
```diff
@@ -0,0 +1,79 @@
...
+
+	spec, ok := result.Object["spec"].(map[string]interface{})
+	if !ok {
+		return nil, fmt.Errorf("auth CR has missing or invalid spec")
+	}
+
+	return &AuthConfig{
+		AdminGroups:   extractStringSlice(spec, "adminGroups"),
+		AllowedGroups: extractStringSlice(spec, "allowedGroups"),
+	}, nil
```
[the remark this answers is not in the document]
caponetto: The Auth CR is a K8s custom resource with a schema-validated spec - the API server enforces that adminGroups and allowedGroups are string arrays before the data reaches us. Silently skipping a non-string element is the safe default (defensive, not lossy in practice). Failing hard on malformed data from a CR would break the entire auth flow over a theoretically impossible input, which is worse than the current behavior.

distributions/core-bff/bff/internal/api/dashboard_config_handler.go
```diff
@@ -0,0 +1,79 @@
...
+		} else {
+			app.serverErrorResponse(w, r, err)
+		}
+		return
+	}
+
+	if err := app.WriteJSON(w, http.StatusOK, raw, nil); err != nil {
+		app.serverErrorResponse(w, r, err)
+	}
+}
```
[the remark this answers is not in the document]
caponetto: The handlers use Go's standard net/http patterns with explicit error responses - adding structured logging at every branch is an observability enhancement outside the scope of this PR. The existing error response functions already produce HTTP status codes that are observable via standard access logs and metrics.

distributions/core-bff/bff/internal/repositories/connection_type.go
```diff
@@ -0,0 +1,169 @@
...
+			existing.ResourceVersion = patched.ResourceVersion
+			if _, revertErr := r.saClientset.CoreV1().ConfigMaps(namespace).Update(ctx, existing, metav1.UpdateOptions{}); revertErr != nil {
+				errMsg = fmt.Sprintf("%s (revert failed: %v)", errMsg, revertErr)
+			}
+		}
+		return &models.MutationResponse{
+			Success: false,
+			Error:   errMsg,
+		}, nil
+	}
```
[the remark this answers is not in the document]
caponetto: Post-validation on the actual final state is more correct than pre-scanning specific patch ops, which would be both incomplete and overly restrictive.

distributions/core-bff/bff/openapi/src/core-bff.yaml
```diff
@@ -208,6 +208,395 @@ paths:
...
+      tags:
+        - Config
+      operationId: patchConfig
+      summary: Update dashboard config (admin)
+      requestBody:
+        content:
+          application/json:
+            schema:
+              type: object
+      responses:
```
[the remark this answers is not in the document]
caponetto: Fully specifying every schema in the OpenAPI spec is a documentation improvement, not a code defect. The BFF endpoints use DashboardConfig, ClusterSettings, and ConnectionType models that are complex K8s-derived types - properly defining them in OpenAPI is a separate task that should be done holistically, not patched inline during a code review.

distributions/core-bff/bff/README.md
```diff
@@ -8,37 +8,49 @@ Backend-for-frontend providing core endpoints required by the dashboard UI.
...
+- `/api/k8s/*` - K8s API pass-through proxy
+- `/wss/k8s/*` - WebSocket watch proxy
+
+**Config & Status:**
+- `GET /api/config` - merged DashboardConfig with defaults
+- `PATCH /api/config` - update DashboardConfig (admin)
+- `GET /api/status` - user session status and cluster info
+- `GET /api/status/:namespace/allowedUsers` - notebook users in a namespace (admin)
+- `GET /api/components` - OdhApplication list (empty when CRD absent)
+- `GET /api/components/remove` - remove app from enabled-apps ConfigMap (admin)
```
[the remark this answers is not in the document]
caponetto: The GET for removeComponent is intentional Fastify parity - the existing Fastify backend uses GET for this endpoint, and the Go BFF must match the same HTTP contract. Changing the verb would break all existing frontend callers.

distributions/core-bff/contract-tests/__tests__/foundation/testCoreBffConfig.test.ts
```diff
@@ -0,0 +1,151 @@
...
+    });
+
+    it('should not reject admin with 401 or 403', async () => {
+      const result = await apiClient.get('/api/dashboardConfig/opendatahub/odh-dashboard-config');
+      const status = result.success
+        ? result.response.status
+        : (result as ApiTestResultError).error.status;
+      expect(status).not.toBe(401);
+      expect(status).not.toBe(403);
+    });
```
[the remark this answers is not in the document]
caponetto: This test intentionally checks only that auth passes (not 401/403) - it's an auth contract test, not a data test. In mock mode the CRD is absent, so the endpoint returns 404, which is already asserted by the separate test at line 97. Asserting toBe(200) here would fail because the CRD doesn't exist in the mock BFF.

distributions/core-bff/bff/internal/api/cluster_settings_handler.go
```diff
@@ -0,0 +1,107 @@
...
+	"rolling":  true,
+	"recreate": true,
+}
+
+func validateClusterSettings(s *models.ClusterSettings) error {
+	if s.PVCSize <= 0 {
+		return errors.New("pvcSize must be greater than 0")
+	}
+	if s.CullerTimeout < 0 {
+		return errors.New("cullerTimeout must not be negative")
```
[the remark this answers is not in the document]
caponetto: Fastify doesn't enforce a "multiple of 60" constraint - it simply converts between seconds and minutes (divides/multiplies by 60). The value 0 is also valid (means culler disabled). Adding a stricter validation than Fastify would break parity and reject valid inputs like cullerTimeout: 0.

distributions/core-bff/bff/internal/config/environment.go
```diff
@@ -68,6 +68,41 @@ func (d DeploymentMode) IsFederatedMode() bool {
...
+		*p = PlatformOpenShift
+	case "XKS":
+		*p = PlatformXKS
+	case "":
+		*p = ""
+	default:
+		return fmt.Errorf("invalid platform type: %s (must be OpenShift, XKS, or empty for auto-detect)", value)
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
caponetto: The platform type values come from environment variables set by the operator/deployment config, not from user input. The constants "OpenShift" and "XKS" are exact-match by design - accepting case-insensitive variants would mask misconfiguration that should fail fast (which we already enforce via the os.Exit(1) fix in main.go).

distributions/core-bff/bff/internal/repositories/allowed_users.go
```diff
@@ -0,0 +1,114 @@
...
+			}
+		} else {
+			users[user.Username] = user
+		}
+	}
+
+	result := make([]models.AllowedUser, 0, len(users))
+	for _, u := range users {
+		result = append(result, *u)
+	}
```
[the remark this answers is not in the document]
caponetto: Notebooks are per-user workbenches scoped to a single namespace - the count is typically in the single digits to low dozens, not thousands. The Fastify backend doesn't paginate this call either. Adding pagination for a resource with this cardinality adds complexity without practical benefit.

distributions/core-bff/bff/internal/repositories/components.go
```diff
@@ -0,0 +1,90 @@
...
+	if cm.Data == nil {
+		return &models.MutationResponse{Success: true, Error: ""}, nil
+	}
+
+	delete(cm.Data, appName)
+
+	_, err = r.saClientset.CoreV1().ConfigMaps(namespace).Update(ctx, cm, metav1.UpdateOptions{})
+	if err != nil {
+		return &models.MutationResponse{Success: false, Error: fmt.Sprintf("failed to update enabled apps ConfigMap: %v", err)}, nil
+	}
```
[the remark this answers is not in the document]
caponetto: The RemoveComponent operation is an admin-only, low-frequency action (removing an app from the enabled list). The chance of concurrent modifications to the same ConfigMap is negligible. Adding a retry loop adds complexity for a conflict scenario that doesn't occur in practice. If it ever becomes an issue, it's a straightforward follow-up.

distributions/core-bff/frontend/config/webpack.dev.js
```diff
@@ -80,18 +80,14 @@ module.exports = smp.wrap(
...
-            },
+            target: `${PROXY_PROTOCOL}://${PROXY_HOST}:${PROXY_PORT}`,
             changeOrigin: true,
             headers: getProxyHeaders(),
             pathRewrite: { '^/core-bff': '' },
           },
           {
             context: ['/wss/k8s', '/core-bff/wss/k8s'],
-            target: `${PROXY_PROTOCOL === 'https:' ? 'wss' : 'ws'}://${PROXY_HOST}:${PROXY_PORT}`,
+            target: `${PROXY_PROTOCOL === 'https' ? 'wss' : 'ws'}://${PROXY_HOST}:${PROXY_PORT}`,
```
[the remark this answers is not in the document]
caponetto: This is a local dev webpack config - if a developer types https: instead of https, the proxy immediately fails and they fix the typo. Normalizing env vars in dev tooling is unnecessary when the feedback loop is instant.

distributions/core-bff/frontend/src/app/components/ComponentsCard.tsx
```diff
@@ -0,0 +1,83 @@
...
+      </Label>
+    );
+  }
+  return (
+    <LabelGroup>
+      {apps.map((app) => (
+        <Label key={app.metadata?.name} color="blue" isCompact>
+          {app.metadata?.name ?? 'unknown'}
+        </Label>
+      ))}
```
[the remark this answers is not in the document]
caponetto: These are K8s OdhApplication CRDs - metadata.name is guaranteed unique within a namespace and is always present on any persisted resource. Using uid or index composites adds no value here.

distributions/core-bff/bff/internal/api/middleware.go
```diff
@@ -68,12 +68,70 @@ func (app *App) InjectRequestIdentity(next http.Handler) http.Handler {
...
+			app.unauthorizedResponse(w, r, fmt.Errorf("missing identity"))
+			return
+		}
+
+		isAdmin, err := client.IsUserAdmin(ctx, identity)
+		if err != nil {
+			app.logger.Warn("admin SAR check failed, denying by default", slog.Any("error", err))
+		}
+		if err != nil || !isAdmin {
+			app.unauthorizedResponse(w, r, fmt.Errorf("you lack the sufficient permissions to make this request"))
```
DaoDaoNoCode, line 94: This should be a 403 (Forbidden), not 401 (Unauthorized). The user *is* authenticated at this point — they just don't have admin privileges. 401 tells the client "try again with different credentials" which isn't the right signal here. You already have `forbiddenResponse` and use it correctly in the dashboard_config and allowed_users handlers — just swap it in here too.

The test in `middleware_admin_test.go` (`TestRequireAdmin_NonAdminUserGets401`) would need updating to expect 403.

distributions/core-bff/bff/internal/repositories/connection_type.go
```diff
@@ -0,0 +1,170 @@
...
+}
+
+func (r *ConnectionTypeRepository) Patch(ctx context.Context, namespace, name string, patchData []byte) (*models.MutationResponse, error) {
+	if err := r.validateExistingConnectionType(ctx, namespace, name); err != nil {
+		return err, nil
+	}
+
+	existing, getErr := r.saClientset.CoreV1().ConfigMaps(namespace).Get(ctx, name, metav1.GetOptions{})
+	if getErr != nil {
+		return &models.MutationResponse{Success: false, Error: fmt.Sprintf("failed to fetch baseline for revert: %v", getErr)}, nil
```
DaoDaoNoCode: This apply-then-check-then-revert pattern has a race window: between the `Patch` call and the rollback `Update`, the ConfigMap exists in an invalid state (labels stripped). If the revert fails, it stays broken.

Could you apply the JSON patch to the object in-memory first (e.g. using `jsonpatch.Apply` on the raw bytes), validate the labels are still present, and only then send the validated object to the API server? That way you never write an invalid resource.

Also — the two GETs (`validateExistingConnectionType` + the baseline fetch here) could be collapsed into one since they fetch the same ConfigMap.

distributions/core-bff/bff/internal/api/connection_type_handler.go
```diff
@@ -0,0 +1,137 @@
...
+		return
+	}
+
+	if err := app.WriteJSON(w, http.StatusOK, cm, nil); err != nil {
+		app.serverErrorResponse(w, r, err)
+	}
+}
+
+// CreateConnectionTypeHandler creates a connection-type ConfigMap.
+func (app *App) CreateConnectionTypeHandler(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
```
DaoDaoNoCode, line 62: Nit: the mutation handlers (Create, Update, Patch, Delete) don't call `validateCallerToken` themselves — they rely entirely on `requireAdmin` in the route registration. The read handlers (`List`, `Get`) do call it inline.

Not a bug today since `requireAdmin` implicitly validates the token via the SSAR check. But it means the mutation handlers have no defense-in-depth — if someone accidentally removed `requireAdmin` from the route, these would proceed with the SA client and no auth at all. The read handlers would still reject.

Worth adding `validateCallerToken` at the top of mutation handlers too, for consistency and as a safety net.

distributions/core-bff/bff/internal/api/connection_type_handler.go
```diff
@@ -0,0 +1,137 @@
...
+
+	writeConnectionTypeMutationResponse(app, w, r, result)
+}
+
+// PatchConnectionTypeHandler partially updates a connection-type ConfigMap.
+func (app *App) PatchConnectionTypeHandler(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
+	name := ps.ByName("name")
+
+	r.Body = http.MaxBytesReader(w, r.Body, 1_048_576)
+	body, err := io.ReadAll(r.Body)
```
DaoDaoNoCode, line 121: `PatchConfigHandler` calls `json.Valid(body)` before forwarding to K8s, but this handler skips that check and sends the raw bytes directly. Would be good to be consistent — pre-validating gives a cleaner error message than whatever K8s returns for malformed JSON.

distributions/core-bff/bff/internal/repositories/connection_type.go
```diff
@@ -0,0 +1,170 @@
...
+	if !isConnectionTypeConfigMap(cm) {
+		return &models.MutationResponse{
+			Success: false,
+			Error:   "configmap must have connection-type labels",
+		}, nil
+	}
+
+	_, err := r.saClientset.CoreV1().ConfigMaps(namespace).Create(ctx, cm, metav1.CreateOptions{})
+	if err != nil {
+		return &models.MutationResponse{Success: false, Error: err.Error()}, nil
```
DaoDaoNoCode, line 76: Returning `(&MutationResponse{Success: false, Error: err.Error()}, nil)` for K8s API failures means the handler always writes HTTP 200 regardless of what went wrong. The caller can't distinguish a validation rejection from a network timeout or a permissions error.

I get this matches Fastify's pattern, but it makes monitoring blind to server-side failures (nothing triggers on HTTP 5xx) and gives clients no way to implement retry logic for transient errors.

Would it make sense to return `(nil, err)` for unexpected K8s failures and reserve the `MutationResponse{Success: false}` pattern for domain-level validation errors? Even if that's a bigger change, it might be worth tracking as a follow-up.

distributions/core-bff/bff/internal/api/config_handler.go
```diff
@@ -0,0 +1,85 @@
...
+func (app *App) GetConfigHandler(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
+	ctx := r.Context()
+
+	if err := app.validateCallerToken(ctx); err != nil {
+		app.unauthorizedResponse(w, r, err)
+		return
+	}
+
+	var featureFlagOverrides map[string]bool
+	if flagsHeader := r.Header.Get("x-odh-feature-flags"); flagsHeader != "" {
```
DaoDaoNoCode, line 23: The `x-odh-feature-flags` header lets any authenticated (non-admin) user override dashboard config booleans for their session. It won't persist, but it can unlock features an admin deliberately disabled — `disableModelServing`, `disableKServe`, etc. The hardcoded lockouts in `applyFeatureLockouts` will re-force a few keys, but most config booleans are unprotected.

Is this intentionally open in production, or should it be gated on `app.config.DevMode`? If intentional, a quick comment explaining the trade-off would help future readers.

distributions/core-bff/bff/internal/api/routes.go
```diff
@@ -0,0 +1,138 @@
...
+	r.GET(ClusterSettingsPath, app.requireAdmin(app.GetClusterSettingsHandler))
+	r.PUT(ClusterSettingsPath, app.requireAdmin(app.UpdateClusterSettingsHandler))
+}
+
+func (app *App) registerConnectionTypeRoutes(r *httprouter.Router) {
+	r.GET(ConnectionTypesPath, app.ListConnectionTypesHandler)
+	r.GET(ConnectionTypeSinglePath, app.GetConnectionTypeHandler)
+	r.POST(ConnectionTypesPath, app.requireAdmin(app.CreateConnectionTypeHandler))
+	r.PUT(ConnectionTypeSinglePath, app.requireAdmin(app.UpdateConnectionTypeHandler))
+	r.PATCH(ConnectionTypeSinglePath, app.requireAdmin(app.PatchConnectionTypeHandler))
```
DaoDaoNoCode, line 136: FYI — the three PATCH endpoints (config, dashboardConfig by name, connection-types) have no contract tests. Looks like the old test file had a comment about `ContractApiClient` not exposing a `.patch()` method — the comment got removed in this refactor but the gap is still there.

At minimum, auth-rejection tests (401/403 for restricted users) would be good to have. If extending the test client is too much for this PR, maybe leave a TODO or tracking comment?

distributions/core-bff/bff/internal/api/middleware.go
caponetto: Intentional Fastify parity. Fastify's route-security.ts:36 returns 401 for non-admin:
```ts
throw createCustomError('Not Admin', '...', 401);
```
Fastify itself is inconsistent here - it uses 403 for wrong-namespace (line 64) and wrong-resource (line 105) but 401 for non-admin. The design doc flags this as unconventional and a candidate for future revisiting. Keeping 401 for now to match the existing API contract.

distributions/core-bff/bff/internal/repositories/connection_type.go
caponetto: Good catch on all three points. Fixed.

1. Race window eliminated: Switched from apply-then-check-then-revert to in-memory validation. The patch
  is now applied to the existing ConfigMap's JSON bytes using evanphx/json-patch/v5 (already an indirect
  dependency), validated for labels, and only then written via Update. The invalid state never hits the API
  server.
2. Two GETs collapsed into one: The separate validateExistingConnectionType call is replaced by inlining
  the existence + label check at the start of Patch, reusing the same object for the in-memory patch
  baseline.
3. Semantics: The method now uses Update (with the ResourceVersion from the GET) instead of Patch, so
  optimistic concurrency still applies.

distributions/core-bff/bff/internal/api/connection_type_handler.go
caponetto: Done. Added validateCallerToken at the top of all four mutation handlers (Create, Update, Patch, Delete) for consistency with the read handlers.

distributions/core-bff/bff/internal/api/connection_type_handler.go
caponetto: Done. Added json.Valid(body) check before forwarding, consistent with PatchConfigHandler. Returns 400 with a clean error message for malformed JSON.

distributions/core-bff/bff/internal/api/routes.go
caponetto: Addressed in this PR. Added a patch() method to ContractApiClient and added contract tests for all three PATCH endpoints:

  - Config PATCH (/api/config) - auth rejection for restricted and unauthenticated users - DashboardConfig PATCH (/api/dashboardConfig/:ns/:name) - auth rejection for restricted and
  unauthenticated users
  - Connection-types PATCH (/api/connection-types/:name) - auth rejection for restricted and unauthenticated users, plus a happy-path test (create then patch, validated against the MutationResponse OpenAPI schema)

distributions/core-bff/bff/internal/api/config_handler.go
caponetto: Intentional Fastify parity - not introduced by this PR. Fastify (backend/src/utils/resourceUtils.ts:544-570) has the same behavior.

The security impact is limited: overrides are per-request only (nothing persisted), applyFeatureLockouts() runs after the overrides so hardcoded lockouts always win, and feature flags only control UI visibility - all actual operations still go through RBAC checks. A non-admin who unlocks a hidden page via the header still can't perform privileged actions on it.

distributions/core-bff/bff/internal/repositories/connection_type.go
caponetto: Intentional Fastify parity. Fastify returns {success: false, error: "..."} with HTTP 200 for all mutation failures, and the frontend parses success to determine the outcome. Changing some errors to 5xx would break that contract. Worth tracking as a cross-cutting follow-up, not blocking for this PR.

caponetto:
@DaoDaoNoCode @lucferbux @andyatmiami let me know if anything else is required for this PR, otherwise let's merge it to unblock next tasks. Any non-blocker issue can be addressed as FUP.

distributions/core-bff/bff/internal/api/allowed_users_handler.go
```diff
@@ -0,0 +1,28 @@
+package api
+
+import (
+	"fmt"
+	"net/http"
+
+	"github.com/julienschmidt/httprouter"
+)
+
+// GetAllowedUsersHandler returns users with notebook access in a namespace.
```
andyatmiami, line 10: Is this code comment accurate?

I see nothing about this function that is specific to `notebooks` - but given our lazer focus on `RHAII` - and the Model Serving use case - I'd be hesitant to start pulling in "boilerplate" logic for Workbenches..

distributions/core-bff/bff/internal/api/cluster_info.go
```diff
@@ -30,35 +33,58 @@ var clusterVersionGVR = schema.GroupVersionResource{
 type clusterInfo struct {
 	clusterID       string
 	clusterBranding string
+	serverURL       string
+	currentContext  string
```
andyatmiami, line 37: not blocking / no need to change here.. but curious why we didn't just include `platformType` in this `clusterInfo` object?  instead we pass it around as a separate element in the return type alongside `clusterInfo` 🤔

distributions/core-bff/bff/internal/api/cluster_info.go
```diff
@@ -30,35 +33,58 @@ var clusterVersionGVR = schema.GroupVersionResource{
...
-	return clusterID
+	return clusterID, nil
+}
+
+func isClusterVersionDiscoveryError(err error) bool {
+	if k8serrors.IsMethodNotSupported(err) {
+		return true
+	}
+	msg := err.Error()
+	return strings.Contains(msg, "no matches for kind") ||
```
andyatmiami, line 86: `strings.Contains` always makes me nervous as we are keying our business logic off strings from 3rd parties we do not control...

1. Is there any more reliable/deterministic field we can key off of?
2. Could we maybe elevate these substrings we are matching to `const` on the file to improve their "visibility" in the codebase?

distributions/core-bff/bff/internal/api/cluster_settings_handler.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+func writeClusterSettingsError(app *App, w http.ResponseWriter, r *http.Request, msg string) {
+	result := &models.MutationResponse{Success: false, Error: msg}
+	if err := app.WriteJSON(w, http.StatusOK, result, nil); err != nil {
+		app.serverErrorResponse(w, r, err)
+	}
+}
+
+var validDeploymentStrategies = map[string]bool{
```
andyatmiami, line 90: Presumably these key values are tied to Kubernetes `Deployment`? I imagine there are constants defined in some k8s lib that we can reference vs. hardcoding these strings...

distributions/core-bff/bff/internal/api/connection_type_handler.go
```diff
@@ -0,0 +1,163 @@
...
+	if err := app.validateCallerToken(r.Context()); err != nil {
+		app.unauthorizedResponse(w, r, err)
+		return
+	}
+
+	name := ps.ByName("name")
+
+	cm, err := app.repositories.ConnectionType.Get(r.Context(), app.config.Namespace, name)
+	if err != nil {
+		if k8serrors.IsNotFound(err) || strings.Contains(err.Error(), "is not a connection type") {
```
andyatmiami, line 48: `strings.Contains` for error disambiguation always scares me...

distributions/core-bff/bff/internal/api/dashboard_config_handler.go
```diff
@@ -0,0 +1,86 @@
...
+// isResourceUnavailable returns true when the resource or its CRD is absent.
+// Broader than repositories.isDiscoveryError: also matches IsNotFound (instance missing).
+func isResourceUnavailable(err error) bool {
+	if k8serrors.IsNotFound(err) || k8serrors.IsMethodNotSupported(err) {
+		return true
+	}
+	msg := err.Error()
+	return strings.Contains(msg, "no matches for kind") ||
+		strings.Contains(msg, "the server could not find the requested resource")
+}
```
andyatmiami, line 24: these `strings.Contains` checks are duplicated in at least one other area of the PR.

Given the inherent brittleness of sub-string matching on error strings - at least we could create a more global helper to centralize this perhaps?

distributions/core-bff/bff/internal/api/middleware.go
```diff
@@ -68,12 +68,70 @@ func (app *App) InjectRequestIdentity(next http.Handler) http.Handler {
...
 
+// requireAdmin wraps a handler and rejects non-admin users with 401.
+// Uses the SSAR check: patch on auths/default-auth.
+// TODO(task3): replace with secureAdminRoute once the full auth chain is implemented.
+func (app *App) requireAdmin(next httprouter.Handle) httprouter.Handle {
+	return func(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
+		ctx := r.Context()
+		client, err := app.kubernetesClientFactory.GetClient(ctx)
+		if err != nil {
+			app.unauthorizedResponse(w, r, fmt.Errorf("failed to get Kubernetes client: %w", err))
```
andyatmiami, line 79: This error here feels like it should be a `500` not a `403` - but if this deals with Fastify "feature parity" - then ignore me...

DaoDaoNoCode:
/lgtm
I will defer the merge to @lucferbux !

distributions/core-bff/bff/internal/api/routes.go
```diff
@@ -0,0 +1,138 @@
...
+	OpenAPIPath        = PathPrefix + "/openapi"
+	OpenAPIJSONPath    = PathPrefix + "/openapi.json"
+	OpenAPIYAMLPath    = PathPrefix + "/openapi.yaml"
+	SwaggerUIPath      = PathPrefix + "/swagger-ui"
+
+	// Starter endpoints (mod-arch-starter convention, /api/v1/ prefix)
+	UserPath      = APIPathPrefix + APIVersion + "/user"
+	NamespacePath = APIPathPrefix + APIVersion + "/namespaces"
+
+	// Config endpoints, /api/ prefix
```
andyatmiami, line 33: Total nit - but might be nice to co-locate routes based on hierarchy..

i.e. all `/status` routes are grouped together... etc

distributions/core-bff/bff/internal/api/routes.go
```diff
@@ -0,0 +1,138 @@
...
+			f.Close()
+			ctxLogger.Debug("Serving static file", slog.String("path", r.URL.Path))
+			fileServer.ServeHTTP(w, r)
+			return
+		}
+		ctxLogger.Debug("Static asset not found, serving index.html", slog.String("path", r.URL.Path))
+		http.ServeFile(w, r, path.Join(app.config.StaticAssetsDir, "index.html"))
+	})
+
+	// Healthcheck outside auth middleware
```
andyatmiami, line 92: Just for my own understanding - what is the use case for when we want healthcheck outside auth?  is this essentially for k8s probes?

distributions/core-bff/bff/internal/config/environment.go
```diff
@@ -68,6 +68,41 @@ func (d DeploymentMode) IsFederatedMode() bool {
...
+
+// Set implements the flag.Value interface
+func (p *PlatformType) Set(value string) error {
+	switch value {
+	case "OpenShift":
+		*p = PlatformOpenShift
+	case "XKS":
+		*p = PlatformXKS
+	case "":
+		*p = ""
```
andyatmiami, line 94: In fullness of time (outside scope of this PR) - I'm not sure we will ever need/want an "auto-detect" feature..

I'm leaning towards the notion this would be set based on an environment variable - and then the startup of `core-bff` will _confirm the runtime environment matches the prescribed setting_ or **panic** ...  as I'm envisioning dedicated images for `RHAII` + `RHOAI` use cases..

So we'd never really expect to "auto-detect" in a realistic/production scenario...

distributions/core-bff/bff/internal/integrations/kubernetes/client.go
```diff
@@ -14,4 +16,19 @@ type KubernetesClientInterface interface {
...
+	// Dynamic client for CRD operations (OdhDashboardConfig, Auth, OdhApplication)
+	GetDynamicClient() (dynamic.Interface, error)
+
+	// ConfigMap operations (connection-types, cluster settings)
+	GetConfigMap(ctx context.Context, namespace, name string) (*corev1.ConfigMap, error)
+	ListConfigMaps(ctx context.Context, namespace string, labelSelector string) (*corev1.ConfigMapList, error)
+	CreateConfigMap(ctx context.Context, namespace string, cm *corev1.ConfigMap) (*corev1.ConfigMap, error)
+	UpdateConfigMap(ctx context.Context, namespace string, cm *corev1.ConfigMap) (*corev1.ConfigMap, error)
+	PatchConfigMap(ctx context.Context, namespace, name string, patchData []byte, patchType types.PatchType) (*corev1.ConfigMap, error)
+	DeleteConfigMap(ctx context.Context, namespace, name string) error
```
andyatmiami, line 33: Not indicating there is a problem with this code - but I'm personally struggling to understand when we'd add a function declaration here vs. where we use business logic in `repositories/`

Is there a rule of thumb for when we'd expect to add more function declarations to this `interface`?  Maybe we could add a little more guidance to the inline comment of the `interface`?

distributions/core-bff/bff/internal/models/cluster_settings.go
```diff
@@ -0,0 +1,31 @@
...
+		KServe: true,
+		LLMd:   true,
+	},
+}
+
+func boolPtr(b bool) *bool { return &b }
```
andyatmiami, line 31: I wonder if we could have a more generic/global helper function along the lines of:

```
// Ptr returns a pointer to the value passed in.
func Ptr[T any](v T) *T {
    return &v
}
```

distributions/core-bff/bff/internal/models/constants.go
andyatmiami, line 1: Don't need/want to scope creep in the context of this PR - but maybe warranted to add a `TODO` to this file about implementing a more dynamic approach to handling the `Version` field.

We have already seen issues in RHOAI dashboard code (around Kueue resources) - where introduction of new versions for resources causes problems / forces a Dashboard change due to this hard-coding.. where it may not actually be necessary if we consulted the cluster and preferred the `stored` Version of a given CR.

Something like this is what I was thinking (note: vibe'd via Gemini)

```
package main

import (
	"fmt"
	"k8s.io/apimachinery/pkg/runtime/schema"
	"k8s.io/client-go/discovery"
	"k8s.io/client-go/rest"
)

// ProxyRegistry holds our resources by Group and Resource name,
// leaving the Version to be determined dynamically.
type ProxyRegistry struct {
	discoveryClient discovery.DiscoveryInterface
}

// GetPreferredGVR dynamically finds the correct Version for a given Group and Resource
func (pr *ProxyRegistry) GetPreferredGVR(group, resource string) (schema.GroupVersionResource, error) {
	// 1. Fetch all server resources grouped by GroupVersion
	_, apiResourceLists, err := pr.discoveryClient.ServerGroupsAndResources()
	if err != nil {
		return schema.GroupVersionResource{}, err
	}

	// 2. Iterate through the resources to find a match
	for _, apiResourceList := range apiResourceLists {
		gv, err := schema.ParseGroupVersion(apiResourceList.GroupVersion)
		if err != nil {
			continue
		}

		// Check if this matches the group we are looking for
		if gv.Group == group {
			for _, apiResource := range apiResourceList.APIResources {
				if apiResource.Name == resource {
					// Found it! Return the complete GVR based on what the server supports
					return schema.GroupVersionResource{
						Group:    gv.Group,
						Version:  gv.Version,
						Resource: resource,
					}, nil
				}
			}
		}
	}

	return schema.GroupVersionResource{}, fmt.Errorf("resource %s not found in group %s", resource, group)
}
```

distributions/core-bff/bff/internal/models/dashboard_config.go
```diff
@@ -0,0 +1,193 @@
...
+
+type AiAssetCustomEndpointsConfig struct {
+	ExternalProviders bool     `json:"externalProviders"`
+	ClusterDomains    []string `json:"clusterDomains"`
+}
+
+// BlankDashboardCR provides defaults for all feature flags.
+// Must match backend/src/utils/constants.ts blankDashboardCR exactly.
+var BlankDashboardCR = DashboardConfig{
```
andyatmiami, line 116: Will we need to modify this function to be `PlatformType` aware? A lot of these defaults are non-sensical in the `RHAII` / `xKS` use case..

distributions/core-bff/bff/internal/models/mutation_response.go
andyatmiami, line 1: No need to change - but from a "consistency" standpoint - its weird this `mutation_response.go` file has the `_response` suffix on it... we don't really differentiate `request` vs. `response` on any of these other `model` files...

distributions/core-bff/bff/internal/repositories/allowed_users.go
```diff
@@ -0,0 +1,114 @@
...
+	"fmt"
+	"strings"
+
+	"github.com/opendatahub-io/odh-dashboard/distributions/core-bff/bff/internal/models"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/dynamic"
+)
+
+// AllowedUsersRepository lists users with notebook access in a namespace.
```
andyatmiami, line 15: If `AllowedUsersRepository` is really **ONLY** concerned with "notebook access" as this comment indicates - it should would be nice (imho) to have a much more specific function name communicating this - something like:
- `AllowedNotebookUsersRepository`

Just a thought...

distributions/core-bff/bff/internal/repositories/allowed_users.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+
+	users := make(map[string]*models.AllowedUser)
+	for _, item := range list.Items {
+		user := parseNotebookUser(item.Object)
+		if user == nil {
+			continue
+		}
+		if existing, ok := users[user.Username]; ok {
+			if user.Privilege == "Admin" {
```
andyatmiami, line 47: Might be nice to elevate `"Admin"` string to a `const` in this file as its a "special value"

distributions/core-bff/bff/internal/repositories/allowed_users.go
```diff
@@ -0,0 +1,114 @@
...
+
+	return &models.AllowedUser{
+		Username:     username,
+		Privilege:    notebookPrivilege(labels),
+		LastActivity: notebookLastActivity(annotations),
+	}
+}
+
+func notebookPrivilege(labels map[string]interface{}) string {
+	if userType, _ := labels["opendatahub.io/user-type"].(string); userType == "admin" {
```
andyatmiami, line 84: Might be nice to take all the hard-coded label strings being referenced in this file and declare them up front as `const` entries - easier to identify what our "supported set" is...

distributions/core-bff/bff/internal/repositories/allowed_users.go
```diff
@@ -0,0 +1,114 @@
...
+		return "Admin"
+	}
+	return "User"
```
andyatmiami, line 87: Might be nice to elevate `"User"` string to a const in this file as its a "special value"

lucferbux (approved):
Thorough review of the full 90-file diff against the Core BFF design doc, xKC deployment plan, common BFF migration strategy, and main BFF migration plan. All looks solid.

**Strategic alignment:**
- On-demand fetching (no caching/background watchers) — correctly follows ADR D1 from the design doc
- Sidecar model during dual-backend period — right deployment pattern per the migration strategy
- Platform detection via `ODH_PLATFORM_TYPE` with auto-detect fallback — cleanly differentiates CRD-absent (not OpenShift) from RBAC/network errors (returns error)
- Service account clients (`saDynClient`/`saClientset`) for privileged operations — matches Fastify's privileged watcher model

**Implementation quality:**
- `deepMerge` with deep copy avoids mutating defaults, correctly replicates `blankDashboardCR` behavior
- Feature lockouts (`disableFineTuning=true`, `mlflow=true`) and XKS overrides properly disable OpenShift-dependent features
- Connection types CRUD has proper label selector validation and SA clientset usage
- `segmentUserID` with SHA-256 of JWT `sub` claim (fallback to username) — privacy-safe analytics
- `resolveIsAllowed` 4-step logic (CRD absent → allowed; system:authenticated check → SSAR fallback) is correct
- `requireAdmin` is explicitly interim with TODO for Task 3 (`secureAdminRoute` parity) — right approach for this stage

**Test coverage:**
- 90+ Go unit tests across handlers, repositories, and middleware
- 49 contract tests split by platform (foundation, openshift, xks)
- All CI checks passing

**BFF Go conventions:**
- Standard `julienschmidt/httprouter` with correct handler signatures
- Proper middleware chain composition
- Standard error envelope pattern
- Clean route extraction to `routes.go`

No blocking concerns. Minor CodeRabbit observations (error message surface area in cluster settings, `installed` param parsing) were addressed in thread.

/lgtm
/approve

distributions/core-bff/bff/internal/repositories/auth.go
andyatmiami, line 1: This Auth CRD is openshift-centric, right? so this no-ops in the xKS case?

distributions/core-bff/bff/internal/repositories/cluster_settings.go
andyatmiami, line 1: What is going to be our strategy for xKS here?  Will this be exposed in xKS - via a Handler -  simply relying on nothing in Dashboard being rendered that would call this?  Or are there more stringent safeguards planned/in place?

distributions/core-bff/bff/internal/repositories/cluster_settings_configmaps.go
```diff
@@ -0,0 +1,144 @@
...
+
+func readCullerTimeout(ctx context.Context, saClientset kubernetes.Interface, namespace string) (int, error) {
+	cm, err := saClientset.CoreV1().ConfigMaps(namespace).Get(ctx, cullerConfigName, metav1.GetOptions{})
+	if err != nil {
+		if k8serrors.IsNotFound(err) {
+			return defaultCullerTimeout, nil
+		}
+		return 0, fmt.Errorf("failed to get culler config: %w", err)
+	}
+	if cm.Data == nil || cm.Data["ENABLE_CULLING"] != "true" {
```
andyatmiami, line 27: I'd love it if we pulled the config map keys we expect/rely on out into constants so:
- we have single source of truth
- its easier to quickly identify what "keys" we support

distributions/core-bff/bff/internal/repositories/cluster_settings_configmaps.go
```diff
@@ -0,0 +1,144 @@
...
+	return err
+}
+
+func (r *ClusterSettingsRepository) createCullerConfigMap(ctx context.Context, namespace string, isEnabled bool, cullerMinutes int) error {
+	cm := &corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      cullerConfigName,
+			Namespace: namespace,
+			Labels: map[string]string{
+				"opendatahub.io/dashboard": "true",
```
andyatmiami, line 104: labels particularly like this one can be referenced across a variety of kubernetes objects - we should ideally define this label as a constant in some globally applicable "space"

distributions/core-bff/bff/internal/repositories/cluster_settings_configmaps.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+	cm.Data["segmentKeyEnabled"] = strconv.FormatBool(enabled)
+	_, err = cmClient.Update(ctx, cm, metav1.UpdateOptions{})
+	return err
+}
+
+// --- Deployment rollout ---
+
+func (r *ClusterSettingsRepository) rolloutDeployment(ctx context.Context, namespace, name string) error {
+	patch := fmt.Sprintf(`{"spec":{"template":{"metadata":{"annotations":{"kubectl.kubernetes.io/restartedAt":"%s"}}}}}`,
```
andyatmiami, line 137: would love to pull this annotation key name into its own constant

distributions/core-bff/bff/internal/repositories/components.go
andyatmiami, line 1: based on inline comments:

```
// ComponentsRepository handles OdhApplication CRD listing and removal.
```

is calling this repository `components.go` really the best name here?  "components" can _mean a lot of things in RHOAI_  - and the definition according to this file is a _new one I haven't even encountered before_ 😇

distributions/core-bff/bff/internal/repositories/connection_type.go
```diff
@@ -0,0 +1,190 @@
...
+	}
+
+	if !isConnectionTypeConfigMap(&modified) {
+		return &models.MutationResponse{
+			Success: false,
+			Error:   "patch would remove required connection-type labels",
+		}, nil
+	}
+
+	if _, updateErr := r.saClientset.CoreV1().ConfigMaps(namespace).Update(ctx, &modified, metav1.UpdateOptions{}); updateErr != nil {
```
andyatmiami, line 142: Any reason we are not using the `Patch` API from `client` ??

Something like:

```
import (
    metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
    "k8s.io/apimachinery/pkg/types"
)

// ...

result, err := r.saClientset.CoreV1().ConfigMaps(namespace).Patch(
    ctx, 
    name, 
    patchType, 
    patchBytes, 
    metav1.PatchOptions{},
)
```

distributions/core-bff/bff/internal/repositories/connection_type.go
```diff
@@ -0,0 +1,190 @@
...
+	cm, err := r.saClientset.CoreV1().ConfigMaps(namespace).Get(ctx, name, metav1.GetOptions{})
+	if err != nil {
+		return &models.MutationResponse{Success: false, Error: err.Error()}, nil
+	}
+	if !isConnectionTypeConfigMap(cm) {
+		return &models.MutationResponse{
+			Success: false,
+			Error:   fmt.Sprintf("configmap %q is not a connection type", name),
+		}, nil
+	}
```
andyatmiami, line 176: seems like we could maybe slightly refactor `validateExistingConnectionType` to apply to this function as well? as the general principles of what we want to do align...

distributions/core-bff/bff/internal/repositories/dashboard_config.go
```diff
@@ -0,0 +1,176 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/dynamic"
+)
+
+// DashboardConfigRepository handles OdhDashboardConfig CRD operations.
+// Uses a service-account-scoped dynamic client for reads, matching the privileged watcher model
+// privileged watcher model where config is read with SA credentials.
+type DashboardConfigRepository struct {
+	isXKS       bool
```
andyatmiami, line 21: Its interesting to see this `isXKS` concept introduced here - as I feel it could potentially be applicable to other `repository` files we have defined..

Although perhaps a better form of flexibility here is to pass the `PlatformType` (or something)... in the event there is ever some 3rd permutation we want to track...  as a `bool` can/will only ever handle 2 scenarios 😇 

And then food for thought on if **ALL** our `repository` structs should expect a `PlatformType` argument - simply for sake of consistency even if its not _always strictly required based on business logic_

andyatmiami: Or, at an even higher level abstraction.. to avoid function signature changes (which are painful to roll out) - if we passed some "Config" like object to all `repositories` - that keeps our function signature consistent moving forward - and we can always add more data to that struct if we need repositories to be aware...

distributions/core-bff/bff/internal/repositories/dashboard_config_overrides.go
```diff
@@ -0,0 +1,113 @@
...
+	cp := make(map[string]any, len(m))
+	for k, v := range m {
+		if vMap, ok := v.(map[string]any); ok {
+			cp[k] = deepCopyMap(vMap)
+		} else {
+			cp[k] = v
+		}
+	}
+	return cp
+}
```
andyatmiami, line 44: These helper functions seem much more generic in nature and likely could be defined in a more "shareable" file

i.e. they are not "only applicable" to OdhDashboardConfig (?)

distributions/core-bff/bff/internal/repositories/dashboard_config_overrides.go
```diff
@@ -0,0 +1,113 @@
...
+	data, err := json.Marshal(obj)
+	if err != nil {
+		return nil, err
+	}
+	var result map[string]interface{}
+	if err := json.Unmarshal(data, &result); err != nil {
+		return nil, err
+	}
+	return result, nil
+}
```
andyatmiami, line 113: see: https://github.com/opendatahub-io/odh-dashboard/pull/7897/changes#r3383407192

distributions/core-bff/bff/internal/repositories/helpers.go
```diff
@@ -48,3 +51,19 @@ func URLWithPageParams(url string, values url.Values) string {
...
+// 403 (Forbidden) is NOT treated as a discovery error because it means the CRD
+// exists but the user lacks access - silencing that would hide misconfiguration.
+func isDiscoveryError(err error) bool {
+	if k8serrors.IsMethodNotSupported(err) {
+		return true
+	}
+	msg := err.Error()
+	return strings.Contains(msg, "no matches for kind") ||
+		strings.Contains(msg, "the server could not find the requested resource")
+}
```
andyatmiami, line 69: Definitely seen this exact check re-implemented in a couple files in this PR - would be good to ensure D.R.Y. principles being followed - particularly if we are hardcoded error substrings repeatedly 😇

distributions/core-bff/bff/internal/repositories/repositories.go
```diff
@@ -1,18 +1,41 @@
...
+	ConnectionType  *ConnectionTypeRepository
+	AllowedUsers    *AllowedUsersRepository
 }
 
 // NewRepositories creates a new Repositories instance with all repositories initialized.
-func NewRepositories() *Repositories {
+// isXKS should be true for XKS platform deployments to disable OpenShift-dependent features.
+// saDynClient and saClientset are service-account-scoped clients for privileged operations,
+// matching the privileged watcher model where the dashboard SA performs reads and admin-gated mutations.
+func NewRepositories(isXKS bool, saDynClient dynamic.Interface, saClientset kubernetes.Interface, namespace string) *Repositories {
```
andyatmiami, line 27: As mentioned in a prior review comment - I'd love if we actually ended up passing some type of `Config` struct into `NewRepositories` - and then propogate that into any/all of the `NewXXXRepository` builders...

Just to have an extensible and stable function signature moving forward...

distributions/core-bff/bff/internal/repositories/status.go
```diff
@@ -0,0 +1,162 @@
...
+			ClusterID:       clusterInfo.GetClusterID(),
+			ClusterBranding: clusterInfo.GetClusterBranding(),
+			IsAdmin:         isAdmin,
+			IsAllowed:       isAllowed,
+			ServerURL:       clusterInfo.GetServerURL(),
+		},
+	}, nil
+}
+
+const systemAuthenticated = "system:authenticated"
```
andyatmiami, line 82: would love to see this `const` declared at top of file so its more "discoverable"

distributions/core-bff/bff/openapi/src/core-bff.yaml
andyatmiami, line 1: Out of curiosity - how is this file produced?  maintained by hand or auto-gen'd in some fashion?

distributions/core-bff/bff/README.md
```diff
@@ -8,37 +8,49 @@ Backend-for-frontend providing core endpoints required by the dashboard UI.
 
 ## Scope
 
-This service exposes:
-
-- GET `/healthcheck` - liveness probe (no auth)
-- GET `/api/v1/healthcheck` - health check (with auth middleware)
-- GET `/api/v1/user` - returns the authenticated user
-- GET `/api/v1/namespaces` - list namespaces (available only when DEV_MODE=true or mock k8s enabled)
+This service exposes infrastructure endpoints, config/status endpoints, and resource CRUD:
```
andyatmiami, line 11: If we have a well-formed OpenAPI file - i'd love to get out of the habit of manually having to document these endpoints in the README as "drift" is a real problem...

While I admittedly may be _biased_ - I do genuinely like how the `kubeflow/notebooks` group addressed this issue to be more D.R.Y.
- https://github.com/kubeflow/notebooks/blob/notebooks-v2/workspaces/backend/README.md#api-documentation

andyatmiami (commented):
Thanks for all the blood, sweat, and tears that went into getting this setup @caponetto 

I wholly realize this PR was already approved and merged (and didn't want that blocked by my review) - but I do still want to share my review comments and discuss them at a later time to:
- determine if any of them should actionable
- ensure I actually understand the system we are building

⚠️ While I did more carefully (statically) review the BFF code - I did **NOT** take that same care/attention to the `frontend/` directory.  For `frontend/` changes - I just have one meta-comment/concern:
- Given we are **NOT** planning on having RHOAI leverage this codebase immediately - aren't we now in a "split-brain" scenario where any changes to the RHOAI counterparts on bits we have defined here in this PR - will then need subsequently "re-synced" ?

PR #7905, opened 2026-06-08
Add inactive state to API keys

ashley-o0o:
Closes: [RHOAIENG-64755](https://redhat.atlassian.net/browse/RHOAIENG-64755)

## Description
Adds an 'Inactive' state to API keys that appears when the subscription associated with the API key gets deleted. The state was added to the status toolbar, so that API keys with inactive can be sorted, and will show up by default. Also adds a popover when 'Inactive' is clicked explaining that the subscription is unavailable.

## How Has This Been Tested?
Tested locally. 

- Go to your API keys, and create a key. 
- Then go to subscriptions and delete the sub you used for your API key
- Go back to API keys, 'Inactive' should be selected by default

## Test Impact
Added a cypress test for new status 

## Screenshots
Screenshot 2026-06-08 at 10 33 08 AM
Screenshot 2026-06-08 at 10 33 24 AM
Screenshot 2026-06-08 at 10 33 36 AM

## Request review criteria:

If you have UI changes: 

## Summary by CodeRabbit

* **New Features & Enhancements**
  * API keys whose subscriptions were deleted are now shown as **“Inactive”**, including an **Inactive** status filter in defaults; inactive keys are ordered consistently. Status rendering now shows an informational popover when a subscription is unavailable.  
* **Bug Fixes**
  * API responses now always include subscription detail data so the client can reliably mark keys as inactive.  
  * Status filter selection is now matched using a case-insensitive exact match to prevent incorrect menu selection.  
* **Tests**
  * Updated/added Cypress coverage for orphaned/inactive keys, popover behavior, and status filtering.

### DIFF
--- a/packages/cypress/cypress/tests/mocked/modelsAsAService/maasApiKeys.cy.ts
+++ b/packages/cypress/cypress/tests/mocked/modelsAsAService/maasApiKeys.cy.ts
@@ -229,12 +229,6 @@ describe('API Keys Page', () => {
     apiKeysPage.visit();
     cy.wait('@searchWithOrphaned');
 
-    apiKeysPage.findTable().should('not.contain.text', 'deleted-subscription-key');
-
-    apiKeysPage.findStatusFilterToggle().click();
-    apiKeysPage.findStatusFilterOption('Inactive').click();
-    apiKeysPage.findStatusFilterToggle().click();
-
     apiKeysPage.findTable().should('contain.text', 'deleted-subscription-key');
 
     const deletedSubscriptionRow = apiKeysPage.getRow('deleted-subscription-key');
--- a/packages/cypress/cypress/tests/mocked/modelsAsAService/maasApiKeys.cy.ts
+++ b/packages/cypress/cypress/tests/mocked/modelsAsAService/maasApiKeys.cy.ts
@@ -237,9 +237,12 @@ describe('API Keys Page', () => {
 
     apiKeysPage.findTable().should('contain.text', 'deleted-subscription-key');
 
-    const orphanedRow = apiKeysPage.getRow('deleted-subscription-key');
-    orphanedRow.findStatus().should('contain.text', 'Inactive').click();
+    const deletedSubscriptionRow = apiKeysPage.getRow('deleted-subscription-key');
+    deletedSubscriptionRow.findStatus().should('contain.text', 'Inactive').click();
     inactiveStatusPopover.shouldBeVisible();
+
+    deletedSubscriptionRow.findSubscription().should('contain.text', 'deleted-sub');
+    deletedSubscriptionRow.findSubscriptionDetailLink().should('not.exist');
   });
 
   it('should display all API keys when the status filter is cleared', () => {
@@ -280,21 +283,6 @@ describe('API Keys Page', () => {
       .findSubscriptionDetailLink()
       .should('have.attr', 'href')
       .and('include', '/maas/keys-and-subs/subscriptions/premium-team-sub');
-  });
-
-  it('should show plain text (no link) for a subscription that no longer exists', () => {
-    const keyWithDeletedSub = mockAPIKeys().filter((k) => k.status === 'active');
-    cy.interceptOdh(
-      'POST /maas/api/v1/api-keys/search',
-      // Return keys with a subscription name but no subscriptionDetails entry for it
-      mockSearchResponse(keyWithDeletedSub, {}),
-    ).as('deletedSubSearch');
-    apiKeysPage.visit();
-    cy.wait('@deletedSubSearch');
-
-    const prodRow = apiKeysPage.getRow('production-backend');
-    prodRow.findSubscription().should('contain.text', 'premium-team-sub');
-    prodRow.findSubscriptionDetailLink().should('not.exist');
   });
 
   it('should filter api keys by subscription and clear the filter', () => {
--- a/packages/cypress/cypress/tests/mocked/modelsAsAService/maasApiKeys.cy.ts
+++ b/packages/cypress/cypress/tests/mocked/modelsAsAService/maasApiKeys.cy.ts
@@ -360,8 +360,11 @@ describe('API Keys Page', () => {
 
     apiKeysPage.findStatusFilterToggle().click();
     apiKeysPage.findStatusFilterOption('Active').click();
+    apiKeysPage.findStatusFilterOption('Inactive').click();
+    apiKeysPage.findStatusFilterToggle().click();
 
-    // Keys are filtered to show active and expired by default so here we're looking for just expired since active was pre-selected
+    // Keys are filtered to show active,inactive and expired by default so here we're looking for just expired since active and inactive was pre-selected
+    cy.wait('@filterByStatus');
     cy.wait('@filterByStatus').then((interception) => {
       expect(interception.request.body.data.filters.status).to.deep.equal(['expired']);
     });
--- a/packages/cypress/cypress/tests/mocked/modelsAsAService/maasApiKeys.cy.ts
+++ b/packages/cypress/cypress/tests/mocked/modelsAsAService/maasApiKeys.cy.ts
@@ -11,6 +11,7 @@ import {
   adminBulkRevokeAPIKeyModal,
   apiKeysPage,
   bulkRevokeAPIKeyModal,
+  inactiveStatusPopover,
   revokeAPIKeyModal,
   copyApiKeyModal,
   createApiKeyModal,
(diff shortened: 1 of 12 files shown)


sridarna:
@ashley-o0o Came accross  this in acceptance criteria:
Status filter defaults to "All" so inactive keys are visible without toggling filters
Should we add inactive as well in the default status filter?
Screenshot 2026-06-09 at 1 44 18 PM

ashley-o0o:
> Came accross this in acceptance criteria: Status filter defaults to "All" so inactive keys are visible without toggling filters Should we add inactive as well in the default status filter?

I feel like no? Maybe that bit was added by claude. I'll ask the channel and see what UX thinks 

EDIT: UX wanted 'Inactive' to also show up by default. PR has been updated

packages/maas/bff/internal/api/api_keys_handlers.go
```diff
@@ -175,9 +175,7 @@ func enrichAPIKeysWithSubscriptionDetails(app *App, r *http.Request, response *m
 		details[sub.SubscriptionIDHeader] = models.SubscriptionDetail{DisplayName: displayName, Models: modelNames}
 	}
 
-	if len(details) > 0 {
-		response.SubscriptionDetails = details
-	}
+	response.SubscriptionDetails = details
```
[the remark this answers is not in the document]
Griffin-Sullivan: can you explain a bit more? a key should always have a subscription but if the details are no longer there that means it was deleted (cause the fetch for the subscription didn't come through). Therefore we can set inactive

packages/maas/frontend/src/app/hooks/useApiKeysTableState.ts
```diff
@@ -48,20 +49,75 @@ export const useApiKeysTableState = (): UseApiKeysTableStateReturn => {
   const [localUsername, setLocalUsername] = React.useState('');
   const [isFetching, setIsFetching] = React.useState(false);
 
+  const serverStatuses = React.useMemo(() => {
```
Griffin-Sullivan: I'm not too sure about this name. It's a bit vague `serverStatuses`

packages/maas/frontend/src/app/pages/keys-and-subs/apiKeys/allKeys/ApiKeysTable.tsx
```diff
@@ -142,6 +142,11 @@ const ApiKeysTable: React.FC<ApiKeysTableProps> = ({
                   subscriptionDetail={
                     apiKey.subscription ? subscriptionDetails?.[apiKey.subscription] : undefined
                   }
+                  isSubscriptionUnavailable={
```
Griffin-Sullivan: We should just pass in the `isKeyInactive` from `useApiKeysTableState`

packages/maas/frontend/src/app/types/api-key.ts
```diff
@@ -1,4 +1,4 @@
-export type APIKeyStatus = 'active' | 'revoked' | 'expired';
+export type APIKeyStatus = 'active' | 'revoked' | 'expired' | 'inactive';
```
Griffin-Sullivan: We should consider a separate type since these reflect the backend options. Maybe something like

```
APIKeyDisplayStatus = APIKeyStatus | 'inactive' 
```

so we can separate what comes from the API and what is a UI state

packages/maas/frontend/src/app/hooks/useApiKeysTableState.ts
```diff
@@ -48,20 +49,75 @@ export const useApiKeysTableState = (): UseApiKeysTableStateReturn => {
...
   );
 
-  const [response, loaded, error, refresh] = useFetchApiKeys(searchRequest);
+  const response = React.useMemo((): APIKeyListResponse => {
+    const hasActive = filterData.statuses.includes('active');
+    const hasInactive = filterData.statuses.includes('inactive');
+
+    let { data } = rawResponse;
+
+    if (hasActive !== hasInactive) {
```
Griffin-Sullivan: 1. I think this logic should be in a re-usable util function in case we need it in other places
2. Claude caught an issue where the table will still say the total count including all active ones 

```
  When the user filters to "Inactive only":
  - serverStatuses sends ['active'] to the server
  - Server returns N active keys with total count N
  - Client filters down to M inactive ones
  
  The pagination footer will show total = N, but only M rows are visible. This creates confusing UX ("showing 1–10 of 47" when only 3 rows display).
```

packages/cypress/cypress/tests/mocked/modelsAsAService/maasApiKeys.cy.ts
```diff
@@ -244,21 +279,6 @@ describe('API Keys Page', () => {
...
-      mockSearchResponse(keyWithDeletedSub, {}),
-    ).as('deletedSubSearch');
-    apiKeysPage.visit();
-    cy.wait('@deletedSubSearch');
-
-    const prodRow = apiKeysPage.getRow('production-backend');
-    prodRow.findSubscription().should('contain.text', 'premium-team-sub');
-    prodRow.findSubscriptionDetailLink().should('not.exist');
-  });
-
```
katieperry4, line 261: why delete this doesn't the link still show up as plain text?

Image

katieperry4 (commented):
The filter works and the popover displays correctly 👍 

Image

Image
one question about a test but otherwise seems solid!

packages/cypress/cypress/tests/mocked/modelsAsAService/maasApiKeys.cy.ts
ashley-o0o: I combined that test with the one for Inactive (`it('should show Inactive status with popover for keys whose subscription was deleted'`) since the link won't show when the key is inactive.

katieperry4:
fair! 
/lgtm

packages/maas/bff/internal/api/api_keys_handlers.go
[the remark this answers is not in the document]
Griffin-Sullivan: @ashley-o0o can you look into making this change

packages/maas/frontend/src/app/hooks/useApiKeysTableState.ts
```diff
@@ -48,20 +53,57 @@ export const useApiKeysTableState = (): UseApiKeysTableStateReturn => {
   const [localUsername, setLocalUsername] = React.useState('');
   const [isFetching, setIsFetching] = React.useState(false);
 
+  const apiFilterStatuses: APIKeyStatus[] = React.useMemo(() => {
+    const hasInactive = filterData.statuses.includes('inactive');
+    const result = filterData.statuses.filter((s): s is APIKeyStatus => s !== 'inactive');
+    if (hasInactive && !result.includes('active')) {
+      result.push('active');
+    }
```
Griffin-Sullivan, line 65: this logic is confusing I believe it's because the API still treats inactive filters as "active" (there is no inactive in the API). Can we add a comment to explain this section

packages/maas/frontend/src/app/pages/keys-and-subs/apiKeys/allKeys/ApiKeysTableRow.tsx
```diff
@@ -57,12 +64,28 @@ const renderApiKeyCell = (
           truncateDescriptionLines={2}
         />
       );
-    case 'status':
-      return (
-        <Label variant="outline" {...getApiKeyStatusProps(apiKey.status)}>
-          {capitalize(apiKey.status)}
+    case 'status': {
```
Griffin-Sullivan: I don't know if I'm crazy but switch case for api key table row is weird. I guess I didn't notice in previous reviews but we should consider changing this to be more declarative for each column

Griffin-Sullivan:
Looks like the filtering in the table doesn't work with inactive. It just throws them all in at the bottom 
image

packages/maas/bff/internal/api/api_keys_handlers.go
```diff
@@ -167,6 +167,9 @@ func SearchAPIKeysHandler(app *App, w http.ResponseWriter, r *http.Request, _ ht
 // enrichAPIKeysWithSubscriptionDetails fetches subscription data from the MaaS API
 // and populates SubscriptionDetails on the response with model names per subscription.
 func enrichAPIKeysWithSubscriptionDetails(app *App, r *http.Request, response *models.APIKeyListResponse) {
+	details := make(map[string]models.SubscriptionDetail)
+	response.SubscriptionDetails = details
```
Griffin-Sullivan: Ok claude is saying the assignment for details should only happen after a successful fetch and only omit the assignment on error. Does that make sense? I think the original suggestion was assuming other parts of the code wouldn't change and then we changed them haha

packages/maas/frontend/src/app/hooks/useApiKeysTableState.ts
```diff
@@ -22,6 +25,8 @@ export type UseApiKeysTableStateReturn = {
   error: Error | undefined;
   refresh: () => void;
   filterData: ApiKeyFilterDataType;
+  isKeyInactive: (key: APIKey) => boolean;
+  clientFiltered: boolean;
```
Griffin-Sullivan: Seems like this `clientFiltered` is added here but not actually used anywhere. Do we need to make changes or remove it?

packages/maas/bff/internal/api/api_keys_handlers.go
ashley-o0o: Okay I think I got it fixed now? Relying on cursor a bit for this, I'm still pretty unfamiliar with go

packages/maas/frontend/src/app/hooks/useApiKeysTableState.ts
ashley-o0o: I think it got left in by accident at some point. Removed now

Griffin-Sullivan (approved):

PR #7927, opened 2026-06-09
Add key count to BFF and update label in frontend

Griffin-Sullivan:
https://redhat.atlassian.net/browse/RHOAIENG-62471
## Description
Only thing missing for the My Subscriptions feature in the BFF is adding key count to the /v1/subscriptions call. We now make an api key search call filtering on each subscription when we list them so we can show the count badge in the list page.

## How Has This Been Tested?
Tested with a sub that has no keys and one that has 1. Also test with having > 10 keys (since that's the max search we set)

## Test Impact
Updated cypress tests and added a new one to cover the 0 keys edge case.

## Request review criteria:

If you have UI changes: 

## Summary by CodeRabbit

* **New Features**
  * API key count badges added to subscription and model rows, showing counts (capped display like "10+"), correct singular/plural unit, and explicit "0 keys" when none.

* **Tests**
  * Updated/added tests to verify key-count badges across subscription views, including capped counts and behavior when upstream key lookup fails.

### DIFF
--- a/packages/maas/bff/internal/api/api_keys_handlers.go
+++ b/packages/maas/bff/internal/api/api_keys_handlers.go
@@ -24,6 +24,11 @@ func attachAPIKeyHandlers(apiRouter *httprouter.Router, app *App) {
 	apiRouter.GET(constants.SubscriptionByIDPassthroughPath, handlerWithApp(app, GetSubscriptionPassthroughHandler))
 }
 
+// subscriptionKeyCountCap is the maximum number of API keys fetched per subscription
+// when computing key counts. Counts at or above this value all display as this cap,
+// which is enough signal ("a lot of keys") without fetching unbounded data.
+const subscriptionKeyCountCap = 10
+
 // ListSubscriptionsPassthroughHandler handles GET /api/v1/subscriptions
 // Proxies to the maas-api /v1/subscriptions endpoint and returns a sanitised list of subscriptions accessible to the authenticated user.
 func ListSubscriptionsPassthroughHandler(app *App, w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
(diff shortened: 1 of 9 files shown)


packages/maas/frontend/src/app/components/ApiKeyCountLabel.tsx
```diff
@@ -0,0 +1,32 @@
...
+/** Matches the BFF's subscriptionKeyCountCap constant in api_keys_handlers.go */
+export const SUBSCRIPTION_KEY_COUNT_CAP = 10;
+
+type ApiKeyCountLabelProps = {
+  keyCount?: number | null;
+};
+
+/**
+ * Displays a compact API key count badge. Returns null when keyCount is null/undefined.
+ * Shows "10+" when keyCount is at or above the BFF search cap (grey for 0, green for >0).
```
ashley-o0o, line 14: Currently 0 keys still shows as green. Should we update it to be gray like this comment suggests?

Image

packages/maas/bff/internal/api/api_keys_handlers.go
```diff
@@ -24,6 +24,11 @@ func attachAPIKeyHandlers(apiRouter *httprouter.Router, app *App) {
 	apiRouter.GET(constants.SubscriptionByIDPassthroughPath, handlerWithApp(app, GetSubscriptionPassthroughHandler))
 }
 
+// subscriptionKeyCountCap is the maximum number of API keys fetched per subscription
+// when computing key counts. Counts at or above this value all display as this cap,
+// which is enough signal ("a lot of keys") without fetching unbounded data.
+const subscriptionKeyCountCap = 10
```
ashley-o0o, line 30: I modified this basic team mock to have exactly 10 keys, but it shows 10+. Do we want this or should we specify just 10?

Image

packages/maas/frontend/src/app/components/ApiKeyCountLabel.tsx
Griffin-Sullivan: the UX doesn't have an example for this so maybe we ask Cece

packages/maas/bff/internal/api/api_keys_handlers.go
Griffin-Sullivan: I figured 10+ meant 10 inclusive :shrug: in reality we just want the user to know "oh dang there's a lot of keys"

packages/maas/bff/internal/api/api_keys_handlers.go
```diff
@@ -44,6 +51,25 @@ func ListSubscriptionsPassthroughHandler(app *App, w http.ResponseWriter, r *htt
...
+			Pagination: &models.APIKeySearchPagination{Limit: subscriptionKeyCountCap},
+		})
+		if err != nil {
+			app.logger.Warn("failed to count API keys for subscription",
+				"subscription", subscriptions[i].SubscriptionIDHeader, "error", err)
+			continue
+		}
+		subscriptions[i].KeyCount = int32(len(resp.Data))
+	}
+}
```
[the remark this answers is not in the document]
andrewballantyne: This will be something we are "aware of" but won't be addressing in this PR.

The backend sends back an unbounded list when invoking `/v1/subscriptions` and that makes it somewhat difficult to manage this. The mental model would be that of "there is probably not a ton of subscriptions given their scale for model support".

Backend is considering it from a purity perspective that they may do this in the future -- we should "hitch" our support for this with that effort.

ashley-o0o:
Tested out locally. Everything looks good! No issues for me
Screenshot 2026-06-10 at 8 56 49 AM
Screenshot 2026-06-10 at 8 58 08 AM
/lgtm

Griffin-Sullivan (commented):
/approve

PR #7949, opened 2026-06-10
[TLS] Upgrade workspaces Go from 1.22 to 1.24 for ML-KEM readiness

ugiordan:
https://issues.redhat.com/browse/RHOAIENG-61057

## Description

Upgrade workspaces-controller and workspaces-backend from Go 1.22.12 to Go 1.24.4 to enable ML-KEM (X25519MLKEM768) support.

This is a pre-requisite for OCP 5.0 TLS profile compliance ([OCPSTRAT-2611](https://redhat.atlassian.net/browse/OCPSTRAT-2611)). Go 1.24+ automatically enables ML-KEM for TLS 1.3 connections, which is required for the upcoming `tlsAdherence: StrictAllComponents` enforcement in OCP 5.0.

Changes:
- `go.mod`: 1.22.12 -> 1.24.4 (both controller and backend)
- `Dockerfile`: `golang:1.22` -> `golang:1.24` (both)

No dependency changes needed. controller-runtime v0.19.1 is compatible with Go 1.24.

## How Has This Been Tested?

- `go build ./...` passes for both controller and backend
- `go vet ./...` clean for both
- `go mod tidy` produces no changes (no dependency churn)
- Backend `internal/helper` unit tests pass
- Controller and backend envtest-based tests require envtest binaries (etcd/kube-apiserver) which are CI-only, not available locally. These tests are unaffected by the Go version change.

## Test Impact

This is a toolchain-only upgrade with no code changes. Existing tests cover the same functionality. The Go version bump does not change any APIs or behavior, it only adds ML-KEM support at the TLS layer.

## Request review criteria:

### DIFF
--- a/packages/notebooks/upstream/workspaces/backend/Dockerfile
+++ b/packages/notebooks/upstream/workspaces/backend/Dockerfile
@@ -1,5 +1,5 @@
 # Use the golang image to build the application
-FROM --platform=$BUILDPLATFORM golang:1.22 AS builder
+FROM --platform=$BUILDPLATFORM golang:1.24 AS builder
 ARG TARGETOS
 ARG TARGETARCH
 
--- a/packages/notebooks/upstream/workspaces/backend/go.mod
+++ b/packages/notebooks/upstream/workspaces/backend/go.mod
@@ -1,6 +1,6 @@
 module github.com/kubeflow/notebooks/workspaces/backend
 
-go 1.22.12
+go 1.24.4
 
 replace github.com/kubeflow/notebooks/workspaces/controller => ../controller
 
--- a/packages/notebooks/upstream/workspaces/controller/Dockerfile
+++ b/packages/notebooks/upstream/workspaces/controller/Dockerfile
@@ -1,5 +1,5 @@
 # Build the manager binary
-FROM --platform=$BUILDPLATFORM golang:1.22 AS builder
+FROM --platform=$BUILDPLATFORM golang:1.24 AS builder
 ARG TARGETOS
 ARG TARGETARCH
 
--- a/packages/notebooks/upstream/workspaces/controller/go.mod
+++ b/packages/notebooks/upstream/workspaces/controller/go.mod
@@ -1,6 +1,6 @@
 module github.com/kubeflow/notebooks/workspaces/controller
 
-go 1.22.12
+go 1.24.4
 
 require (
 	github.com/go-logr/logr v1.4.2


lucferbux:
@andyatmiami @caponetto this is targeting notebooks but since it's an upstream effort, should we do it in kubeflow and then sync?

andyatmiami:
> @andyatmiami @caponetto this is targeting notebooks but since it's an upstream effort, should we do it in kubeflow and then sync?

@lucferbux - probably actually best to rope in @harshad16 in this case - as he has been more closely (as time allows) looking at the midstream here...

In general - upstream notebooks **IS** actually already on `1.24`
- https://github.com/kubeflow/notebooks/blob/notebooks-v2/workspaces/backend/go.mod#L3

@harshad16 would be the ultimate truth here - but a couple bits of relevant info (as I understand it)
- I think there is a new 'midstream' repo they carved out for notebooks.. i'm not sure if `odh-dashboard` is using that yet
- relatedly... I think perhaps the state of code in `odh-dashboard` as it relates to `notebooks` (v2) is on a pretty old/antiquated commit hash

For fun - I'll also tag @thaorell as I know he as well is working in this area.

harshad16:
Thank you for roping us in , 
yes instead of directly doing this **kinda** changes here, we should do this in midstream https://github.com/opendatahub-io/workbenches
so that we can bring it both to upstream and odh-dashboard based on the change.

odh-dashboard for notebook packages, should only be used as  a read package content and write only for bff content.
direct changes we should target in repo: https://github.com/opendatahub-io/workbenches

@ugiordan thank you firstly for making this change.
as @andyatmiami  pointed out the midstream and upstream both have this change:
https://github.com/opendatahub-io/workbenches/blob/main/workspaces/controller/go.mod
https://github.com/opendatahub-io/workbenches/blob/main/workspaces/backend/go.mod
it is consistent with the structure.

can you please close the PR ?
we would shortly pull those change here soon, or if you want we can close this PR once we pull those changes.

/hold

ugiordan:
Thanks @harshad16, makes sense. The midstream and upstream already have this, so closing the PR.

Can I ask someone from your team to take ownership of RHOAIENG-61057? I went ahead and opened this PR to keep things moving, but it sounds like the right path is the midstream sync. We're tracking this for RHOAI 3.5, so as long as the sync to odh-dashboard happens before then, we're good.

harshad16:
Ack, i have taken the ownership of that JIRA, will carry out the needful within the time line.

PR #7953, opened 2026-06-10
feat: harden auth middleware in Core BFF

caponetto:
## Description

https://issues.redhat.com/browse/RHOAIENG-67047

Hardens the Core BFF auth middleware with composable route wrappers, structured audit logging, dev-mode identity fallback, platform-gated routes, and OpenShift User API resolution.

**Go (BFF):**

*Auth middleware:*

- **`secureRoute()` / `secureAdminRoute()` wrappers** - Composable per-route wrappers replacing the old `requireAdmin` middleware. Both assume `InjectRequestIdentity` has already placed the caller's identity in the request context. `secureRoute` validates via SelfSubjectReview; `secureAdminRoute` adds an SSAR admin gate.
- **`requirePlatform()` middleware** - Returns 404 when the platform does not match. Used to gate `allowedUsers` as OpenShift-only.
- **Dev-mode identity fallback** - When `DevMode=true` and no token is present, injects a static identity from `--dev-user` / `--dev-groups` flags (env: `DEV_USER`, `DEV_GROUPS`). Uses the kubeconfig's native auth. A real token always takes precedence.

*Audit logging:*

- **Structured audit entries** - Every authenticated request emits a structured `slog` record with user, namespace, action, endpoint, isAdmin, and needsAdmin. `secureRoute` resolves admin status asynchronously (background goroutine with `context.WithoutCancel`); `secureAdminRoute` logs synchronously since the SSAR result is already available.

*OpenShift integration:*

- **User API resolution** - `resolveUserViaOpenShiftAPI` queries `user.openshift.io/v1` for the current user's identity. Returns gracefully on XKS or when the API is unavailable.
- **`registerOpenShiftRoutes()`** - Dedicated registration function for platform-specific endpoints, matching the existing `registerConfigRoutes` / `registerConnectionTypeRoutes` pattern.
- **`IsUserAdmin` parity comment** - Documents that the SSAR check mirrors Fastify's `adminUtils.ts isUserAdmin()`.

*Refactoring and cleanup:*

- **Auth-domain constants** - Moves `"system:authenticated"`, `"auth"`, `"default-auth"` to named constants in `models/constants.go`.
- **`internal/maputil/` package** - Promotes generic map utilities (`DeepMerge`, `DeepCopyMap`, `ToUnstructuredMap`) to a standalone package. Domain-specific helpers stay in `repositories/dashboard_config_overrides.go`.
- **Platform resolution to `cluster_info.go`** - Consolidates `resolveStartupPlatform`, `inferPlatform`, and `initStartupClusterInfo` into one file. `app.go` retains only dependency wiring.
- **Connection type Patch** - Replaces local `evanphx/json-patch` apply + `Update()` with server-side `types.JSONPatchType`, matching Fastify's `PATCH_FORMAT_JSON_PATCH`. Eliminates the `evanphx/json-patch` dependency.
- **Connection type label validation DRY** - Extracts `requireConnectionTypeLabels` helper shared by `Create` and `Update`.
- **Token forwarding integration test** - Validates that `TokenKubernetesClient` forwards the user's bearer token (not the SA token), matching Fastify's `directCallUtils.ts` contract.

**Frontend:**

- **Webpack dev proxy improvements** - Adds `K8S_TOKEN` env var support, extracts token resolution into dedicated functions for cleaner fallback handling.
- `**AuthProbeCard` diagnostic component** - Dev-mode card that probes BFF endpoints and displays auth status for debugging.

**Contract tests:**

- **Platform-specific reorganization** - Moves `allowedUsers` tests from the foundation suite to platform-specific suites (OpenShift: 401/array assertions, XKS: 404 assertion).

## How Has This Been Tested?

- **Go unit tests** - All API and kubernetes package tests pass, covering auth middleware wrappers, audit logging, OpenShift User API resolution, dev-mode fallback, token forwarding, and proxy dev-fallback token resolution (HTTP + WebSocket)
- **Go vet** - Clean, no issues
- **Contract tests** - All passing across both platform suites

### Manual testing with a kind cluster

**1. Create the cluster:**

```shell
kind create cluster --name core-bff-tests
```

**2. Create the opendatahub namespace:**

```shell
kubectl create namespace opendatahub
```

**3. Create service accounts:**

```shell
# Admin SA
kubectl create serviceaccount test-admin -n opendatahub
kubectl create clusterrolebinding test-admin-binding \
  --clusterrole=cluster-admin \
  --serviceaccount=opendatahub:test-admin

# Non-admin SA
kubectl create serviceaccount test-viewer -n opendatahub
kubectl create clusterrole ns-viewer --verb=list,get --resource=namespaces
kubectl create clusterrolebinding test-viewer-ns-binding \
  --clusterrole=ns-viewer \
  --serviceaccount=opendatahub:test-viewer
```

**4. Run as admin:**

```shell
K8S_TOKEN=$(kubectl create token test-admin -n opendatahub --duration=24h) make dev-start-federated
```

**5. Run as non-admin:**

```shell
K8S_TOKEN=$(kubectl create token test-viewer -n opendatahub --duration=24h) make dev-start-federated
```

**Expected behavior:**

- **Admin:** AuthProbeCard shows green for both `secureRoute` and `secureAdminRoute`
- **Non-admin:** green for `secureRoute` (200), orange for `secureAdminRoute` (401)

**6. Cleanup when done:**

```shell
kind delete cluster --name core-bff-tests
```

Test page as admin on kind:

image

Test page as non-admin on kind:

image

- `secureAdminRoute` is only available for admins
- `allowedUsers` is only available on OpenShift

## Test Impact

- Added Go unit tests for auth middleware, audit logging, OpenShift User API resolution, token forwarding, platform resolution, and connection type label validation
- Added platform-specific contract tests for allowed users
- Removed old admin middleware tests (superseded by new auth middleware tests)
- Removed post-patch label validation test (removed to match Fastify)
- Moved allowed users contract tests from foundation suite to platform-specific suites

## Request review criteria:

If you have UI changes: 

## Summary by CodeRabbit

* **New Features**
  * Added an Auth Middleware Probe card with live status checks in the UI.
  * Added dev-mode identity flags (dev user/groups) and improved proxy token fallback handling.
  * Added structured audit logging for protected route access.
* **Bug Fixes / Behavior**
  * Unified route authorization (user vs admin) and improved platform auto-detection/mismatch handling.
  * Improved handling of unavailable Kubernetes resources and OpenShift-specific user/allowed-users discovery; aligned 401 vs 403 responses.
* **Documentation**
  * Updated Core BFF guidance, route security semantics, and endpoint discovery references.
* **Tests**
  * Expanded unit and contract coverage for auth, audit, platform, and status/allowed-users scenarios.


distributions/core-bff/bff/internal/api/openshift_user.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+
+	result, err := dynClient.Resource(gvr).Get(ctx, "~", metav1.GetOptions{})
+	if err != nil {
+		app.logger.Debug("OpenShift User API unavailable or returned error", slog.Any("error", err))
+		return "", nil, nil
+	}
+
+	username, _, _ := unstructuredString(result.Object, "metadata", "name")
+	groups, _ := unstructuredStringSlice(result.Object, "groups")
```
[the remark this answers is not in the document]
caponetto: The user.openshift.io/v1/users/~ schema is defined by the OpenShift API server, and metadata.name and groups are guaranteed fields on the User resource, so schema drift here would mean an API-breaking change in OpenShift itself. The current zero-value behavior is intentional: if a field is unexpectedly missing, username becomes "" and groups becomes nil, both of which callers already handle gracefully. Adding warnings or errors for this scenario would make the optional enrichment call less resilient rather than more.

distributions/core-bff/bff/internal/repositories/connection_type.go
```diff
@@ -97,82 +93,40 @@ func (r *ConnectionTypeRepository) Update(ctx context.Context, namespace, name s
...
-			Error:   "patch would remove required connection-type labels",
-		}, nil
-	}
-
-	if _, updateErr := r.saClientset.CoreV1().ConfigMaps(namespace).Update(ctx, &modified, metav1.UpdateOptions{}); updateErr != nil {
-		return &models.MutationResponse{Success: false, Error: updateErr.Error()}, nil
+	_, err := r.saClientset.CoreV1().ConfigMaps(namespace).Patch(ctx, name, types.JSONPatchType, patchData, metav1.PatchOptions{})
+	if err != nil {
+		return &models.MutationResponse{Success: false, Error: err.Error()}, nil
 	}
```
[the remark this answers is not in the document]
caponetto: This PR deliberately removed the post-patch label validation to match Fastify, which applies JSON Patch directly via patchNamespacedConfigMap without validating patch operations for label safety. Adding pre-patch validation would reverse that design decision. The proposed check also only catches exact path matches, so a patch targeting /metadata/labels or /metadata would bypass it. The existing getValidatedConnectionType pre-check already confirms the target is a valid connection type before the patch is applied, and the patch data comes from our own authenticated frontend, not an open API.

distributions/core-bff/frontend/config/webpack.dev.js
```diff
@@ -27,28 +27,58 @@ const ROOT_NODE_MODULES = path.resolve(RELATIVE_DIRNAME, '../../../node_modules'
...
+  const token = process.env.K8S_TOKEN;
+  if (!token) {
+    return null;
+  }
+  try {
+    const payload = JSON.parse(Buffer.from(token.split('.')[1], 'base64').toString());
+    console.info('Using K8S_TOKEN from environment, subject:', payload.sub || 'unknown');
+  } catch {
+    console.info('Using K8S_TOKEN from environment');
+  }
```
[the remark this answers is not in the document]
caponetto: This is webpack.dev.js, which only runs during local development. The logs go to the developer's own terminal, and the "identifiable data" is the developer's own identity. Logging payload.sub and the kubectl auth whoami username serves a clear debugging purpose: confirming which identity the dev proxy is using ("Am I using the right token?", "Am I logged in as admin or viewer?"). Replacing those with generic messages removes the answer to exactly the questions developers ask when debugging auth issues. There is no PII risk since the developer already has the token (they set K8S_TOKEN themselves) and already knows their kubectl identity.

distributions/core-bff/bff/internal/api/app_proxy.go
```diff
@@ -61,6 +61,14 @@ func (app *App) initK8sProxy(cfg config.EnvConfig, k8sResult k8sSetupResult) err
 		outboundHeadersFn = impersonateFromIdentity
 	}
 
+	var devFallbackToken string
+	if cfg.DevMode && !cfg.MockK8Client {
+		kc, _ := helpers.GetKubeconfig()
+		if kc != nil {
+			devFallbackToken = kc.BearerToken
+		}
+	}
```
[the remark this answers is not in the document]
caponetto: Agreed on logging the swallowed GetKubeconfig error, fixed that. For the loopback restriction on InjectRequestIdentity: DevMode is explicitly opt-in (requires DEV_MODE=true or --dev-mode), never set in production deployments. A loopback check would break legitimate dev setups where the BFF runs in a container while the browser is on the host (RemoteAddr is the gateway IP, not loopback), as well as SSH tunnels, VS Code Remote, devcontainers, and Codespaces. Fastify's dev mode also doesn't restrict to loopback. The debug log on fallback activation was already in place (line 54).

distributions/core-bff/bff/internal/api/test_helpers_test.go
```diff
@@ -0,0 +1,87 @@
...
+func (h *captureHandler) Handle(_ context.Context, r slog.Record) error {
+	h.mu.Lock()
+	defer h.mu.Unlock()
+	h.records = append(h.records, r)
+	return nil
+}
+
+func (h *captureHandler) WithAttrs(attrs []slog.Attr) slog.Handler {
+	return &captureHandler{attrs: append(h.attrs, attrs...)}
+}
```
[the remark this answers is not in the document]
caponetto: WithAttrs is never called in our test flows. The audit code logs with inline attributes (logger.Info("audit", slog.String(...))) and doesn't create sub-loggers via logger.With(...). The proposed fix also copies records into independent slices rather than sharing them, so parent and child handlers would diverge on subsequent log calls. If this method ever needs to work correctly, the struct would need pointer-based shared state, but since it's currently dead code, I'd prefer not to add complexity for a path that's never exercised.

distributions/core-bff/bff/internal/repositories/auth.go
```diff
@@ -35,9 +35,9 @@ func (r *AuthRepository) GetAuth(ctx context.Context) (*AuthConfig, error) {
 		return nil, fmt.Errorf("auth dynamic client unavailable")
 	}
 
-	result, err := r.saDynClient.Resource(models.AuthGVR).Get(ctx, "auth", metav1.GetOptions{})
+	result, err := r.saDynClient.Resource(models.AuthGVR).Get(ctx, models.AuthInstanceName, metav1.GetOptions{})
 	if err != nil {
-		if isDiscoveryError(err) {
+		if k8sutil.IsDiscoveryError(err) {
 			return nil, nil
 		}
```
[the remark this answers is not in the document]
caponetto: IsDiscoveryError already implements the narrow predicate you describe. It checks exactly three conditions: IsMethodNotSupported (405), meta.IsNoMatchError (REST mapper failure), and 404 with nil or empty Details.Name (CRD path absent). It does not match 403, 5xx, or NotFound with a populated Details.Name (instance not found). That last case is handled separately on line 43 via k8serrors.IsNotFound, which returns an error rather than (nil, nil). The function is documented with these exclusions in k8sutil/errors.go:22-28. Inlining the same checks here would duplicate the logic and lose the centralized documentation.

caponetto:
@andyatmiami @lucferbux - This PR is ready for review! Sorry for another big PR, but ~2/3 is test code 😇 

@andyatmiami - I triaged your review comments from [#7897](https://github.com/opendatahub-io/odh-dashboard/pull/7897). Some of them are addressed in this PR, and the full report is available [here](https://gist.github.com/caponetto/ec58379d7dc45970b50d61c2047aa0e9). Feel free to change/improve anything you see fit in the coming days! Thanks a lot!

BTW did we miss something? The `area/distributions` label is not being set on PRs 🤔

andyatmiami:
> BTW did we miss something? The area/distributions label is not being set on PRs 🤔

I'm guessing this answers the question I had on "will prow auto-create labels that don't exist 😇 )

image


so i think first step would be to get someone who can create labels - to create this label - then see if it starts working...

lucferbux (commented):
## Security Review — Auth Middleware Hardening (Core BFF)

Thorough security review focusing on authentication/authorization flows, privilege escalation vectors, and defense-in-depth. The overall architecture is solid — composable `secureRoute`/`secureAdminRoute` wrappers, proper SSAR-based admin checks, SSRF-protected proxy, and good token redaction via `BearerToken.String()`. Nice work on the structured audit logging and the `context.WithoutCancel` pattern for async audit.

Below are findings organized by severity. Most are hardening suggestions rather than blocking issues, but a few warrant discussion before merge.

### Summary of Findings

| # | Severity | File | Finding |
|---|----------|------|---------|
| 1 | **Medium** | `middleware_auth.go` | Dev-mode auth bypass has no network-origin restriction |
| 2 | **Medium** | `middleware_auth.go` | `secureAdminRoute` returns 401 for non-admin (should be 403) |
| 3 | **Low** | `openshift_user.go` | All User API errors silently swallowed — can't distinguish unavailable from auth/schema drift |
| 4 | **Low** | `cluster_info.go` | Ambiguous probe errors default to OpenShift — may grant access to OpenShift-only endpoints on XKS |
| 5 | **Low** | `audit.go` | Async audit goroutine errors are logged at Warn but not surfaced — audit gaps may go unnoticed |
| 6 | **Info** | `app_proxy.go` | Dev fallback token extraction error swallowed silently |
| 7 | **Info** | `middleware_auth.go` | Consider rate-limiting or logging failed auth attempts for brute-force detection |

distributions/core-bff/bff/internal/api/middleware_auth.go
```diff
@@ -0,0 +1,196 @@
...
+			}
+			ctx := context.WithValue(r.Context(), constants.RequestIdentityKey, identity)
+			next.ServeHTTP(w, r.WithContext(ctx))
+			return
+		}
+
+		identity, err := app.kubernetesClientFactory.ExtractRequestIdentity(r.Header)
+		if err != nil {
+			if app.config.DevMode {
+				// DevUser/DevGroups are used for audit log fallback labeling (see auditUser).
```
lucferbux, line 30: **[Security — Medium]** Dev-mode auth bypass lacks network-origin restriction.

When `DevMode=true` and token extraction fails, *any* request from *any* source gets the dev fallback identity. In a scenario where DevMode is accidentally left enabled (or enabled in a staging cluster accessible over the network), this becomes a full auth bypass.

**Recommendation:** Add a loopback check so the fallback only activates for requests originating from `127.0.0.1` / `::1`:

```go
if app.config.DevMode && isLoopback(r.RemoteAddr) {
    // dev fallback
}
```

Alternatively, gate this on `MockK8Client` as well (dev mode + mock = definitely local), or log at `Warn` level (not `Debug`) so it's visible in any environment where it unexpectedly activates.

The current `Debug` log level at line 55 means this bypass is effectively invisible in production log configurations.

distributions/core-bff/bff/internal/api/middleware_auth.go
```diff
@@ -0,0 +1,196 @@
...
+			app.unauthorizedResponse(w, r, fmt.Errorf("invalid or expired token: %w", err))
+			return
+		}
+
+		isAdmin, err := client.IsUserAdmin(ctx, identity)
+		if err != nil {
+			app.logger.Warn("admin SAR check failed, denying by default", slog.Any("error", err))
+			app.emitAuditLog(username, r, true, false)
+			app.unauthorizedResponse(w, r, fmt.Errorf("insufficient permissions to make this request"))
+			return
```
lucferbux, line 120: **[Security — Medium]** Non-admin users receive HTTP 401 (Unauthorized) instead of 403 (Forbidden).

Lines 142-148 call `app.unauthorizedResponse()` when the user is authenticated but lacks admin permissions. Per HTTP semantics:
- **401** = "I don't know who you are" (authentication failure)
- **403** = "I know who you are, but you can't do this" (authorization failure)

Returning 401 for a valid-but-non-admin user is misleading — clients may retry with different credentials when the real issue is insufficient privileges. The same issue applies to the SSAR-error case at line 142.

**Recommendation:** Add a `forbiddenResponse` helper (or use the one from `allowed_users_handler.go`) and use it here:

```go
if !isAdmin {
    app.emitAuditLog(username, r, true, false)
    app.forbiddenResponse(w, r, fmt.Errorf("insufficient permissions to make this request"))
    return
}
```

Note: The SSAR *error* case (line 139-143) could arguably remain 401 since we can't confirm the user's permissions, but the `!isAdmin` case (lines 145-148) should definitely be 403.

distributions/core-bff/bff/internal/api/middleware_auth.go
```diff
@@ -0,0 +1,196 @@
...
+
+		client, err := app.kubernetesClientFactory.GetClient(ctx)
+		if err != nil {
+			app.emitAuditLog(auditUser(identity), r, false, false)
+			app.serverErrorResponse(w, r, fmt.Errorf("failed to get Kubernetes client: %w", err))
+			return
+		}
+
+		username, err := client.GetUser(ctx, nil)
+		if err != nil {
```
lucferbux: **[Security — Info]** `secureRoute` calls `client.GetUser()` on every request for audit logging, adding a `SelfSubjectReview` API call per request.

This is correct for security (validates the token is still valid), but worth noting the latency/load cost. If this becomes a bottleneck, consider caching the username resolution per-token with a short TTL (e.g., 30s), since the token itself is already validated by the K8s API server when the client is created.

Not blocking — just a heads-up for observability under load.

distributions/core-bff/bff/internal/api/openshift_user.go
```diff
@@ -0,0 +1,76 @@
...
+		Group:    "user.openshift.io",
+		Version:  "v1",
+		Resource: "users",
+	}
+
+	result, err := dynClient.Resource(gvr).Get(ctx, "~", metav1.GetOptions{})
+	if err != nil {
+		app.logger.Debug("OpenShift User API unavailable or returned error", slog.Any("error", err))
+		return "", nil, nil
+	}
```
lucferbux, line 36: **[Security — Low]** All errors from the User API are silently swallowed — returns `("", nil, nil)`.

This conflates several distinct failure modes:
1. CRD not installed (expected on XKS) → correct to return empty
2. User lacks RBAC to read `users` (403) → should propagate as error
3. API schema drift / unexpected response → should propagate as error
4. Network timeout → should propagate as error

By treating all errors as "unavailable," a 403 (user not permitted to read User API) looks identical to "CRD not installed." If the caller uses this to skip group-based authorization, a permission error could silently bypass group checks.

**Recommendation:** Use the `k8sutil.IsDiscoveryError` helper (which already exists in this PR) to distinguish "CRD not installed" from real errors:

```go
if err != nil {
    if k8sutil.IsDiscoveryError(err) {
        app.logger.Debug("OpenShift User API not available", slog.Any("error", err))
        return "", nil, nil
    }
    return "", nil, fmt.Errorf("failed to query OpenShift User API: %w", err)
}
```

distributions/core-bff/bff/internal/api/cluster_info.go
```diff
@@ -124,3 +116,89 @@ func queryClusterBranding(client kubernetes.Interface, logger *slog.Logger) stri
...
+// resolveStartupPlatform determines the runtime platform.
+// Policy: explicit configuration is always trusted. When the probe result disagrees
+// with the configured value, a warning is logged but the configured value wins.
+// This lets operators force a platform for testing or mixed environments.
+func resolveStartupPlatform(ci clusterInfo, probeErr error, explicit bool, configured config.PlatformType, logger *slog.Logger) config.PlatformType {
+	if !explicit && probeErr != nil {
+		logger.Warn("ClusterVersion probe returned ambiguous error, defaulting to OpenShift",
+			slog.Any("error", probeErr))
+		return config.PlatformOpenShift
+	}
```
lucferbux, line 123: **[Security — Low]** Ambiguous probe errors default to `PlatformOpenShift`.

When the ClusterVersion probe returns an error that isn't clearly "not found" (e.g., network timeout, 500, RBAC denied), this defaults to OpenShift. This matters because `requirePlatform(OpenShift)` gates access to endpoints like `AllowedUsersHandler`.

On a vanilla K8s cluster where the probe times out, the BFF would:
1. Default to OpenShift platform
2. Expose OpenShift-only endpoints
3. Those endpoints would likely fail at the K8s API level, but the route itself would be reachable

**Recommendation:** Consider defaulting to `PlatformXKS` for ambiguous errors (safer — fewer endpoints exposed), or using `IsDiscoveryError` to distinguish "CRD not installed" (→ XKS) from transient errors (→ retry or fail startup).

The current comment says "defaulting to OpenShift" which is documented, but the security-conservative choice would be to default to the more restrictive platform.

distributions/core-bff/bff/internal/api/audit.go
```diff
@@ -0,0 +1,41 @@
...
+)
+
+// emitAuditLogAsync logs the request details asynchronously without blocking the handler.
+// For non-admin routes, the admin status is resolved in the background via SSAR.
+func (app *App) emitAuditLogAsync(ctx context.Context, client k8s.KubernetesClientInterface, username string, identity *k8s.RequestIdentity, r *http.Request, needsAdmin bool) {
+	method := r.Method
+	url := *r.URL
+
+	go func() {
+		// Detach from request context so the SSAR call doesn't race with handler completion.
```
lucferbux: **[Security — Low]** The async audit goroutine swallows SSAR errors at Warn level.

If the admin-status resolution consistently fails (e.g., RBAC misconfiguration), audit logs will always show `isAdmin=false` for all users, including actual admins. This creates a blind spot — an admin performing privileged operations would appear as non-admin in audit logs.

**Suggestions:**
1. Consider adding an `isAdminResolved` or `adminCheckError` field to the audit log so consumers can distinguish "confirmed non-admin" from "couldn't determine":
```go
if err != nil {
    app.logger.Info("audit",
        slog.String("user", username),
        slog.String("adminCheckError", err.Error()),
        // ... other fields
    )
}
```
2. The `context.WithoutCancel` + 10s timeout pattern is correct for detaching from the request lifecycle. Good choice here.

distributions/core-bff/bff/internal/api/app_proxy.go
```diff
@@ -61,6 +62,16 @@ func (app *App) initK8sProxy(cfg config.EnvConfig, k8sResult k8sSetupResult) err
 		outboundHeadersFn = impersonateFromIdentity
 	}
 
+	var devFallbackToken string
+	if cfg.DevMode && !cfg.MockK8Client {
+		kc, kcErr := helpers.GetKubeconfig()
+		if kcErr != nil {
+			app.logger.Warn("Failed to get kubeconfig for dev fallback token", slog.Any("error", kcErr))
+		} else if kc != nil {
```
lucferbux, line 70: **[Security — Info]** When `GetKubeconfig()` fails, `devFallbackToken` stays empty string and the error is only logged at Warn.

This is acceptable for dev mode, but worth noting: if the kubeconfig exists but has an expired or invalid token, `kc.BearerToken` will be set to that stale value. The proxy will then use it as the fallback, and requests will fail with opaque K8s API errors rather than a clear "dev token unavailable" message.

Consider validating the token is non-empty and logging the token source (kubeconfig path) for debuggability:
```go
if kc.BearerToken == "" {
    app.logger.Warn("Kubeconfig has no bearer token; dev fallback will use request identity")
}
```

distributions/core-bff/bff/internal/api/routes.go
```diff
@@ -111,28 +117,37 @@ func (app *App) Routes() http.Handler {
...
+	r.GET(ComponentsRemovePath, app.secureAdminRoute(app.RemoveComponentHandler))
+	r.GET(DashboardConfigPath, app.secureAdminRoute(app.GetDashboardConfigByNameHandler))
+	r.PATCH(DashboardConfigPath, app.secureAdminRoute(app.PatchDashboardConfigByNameHandler))
+	r.GET(ClusterSettingsPath, app.secureAdminRoute(app.GetClusterSettingsHandler))
+	r.PUT(ClusterSettingsPath, app.secureAdminRoute(app.UpdateClusterSettingsHandler))
+}
+
+func (app *App) registerOpenShiftRoutes(r *httprouter.Router) {
+	r.GET(AllowedUsersPath, app.requirePlatform(config.PlatformOpenShift, app.secureAdminRoute(app.GetAllowedUsersHandler)))
 }
```
lucferbux, line 168: **[Security — Positive]** Good composition pattern: `requirePlatform(OpenShift, secureAdminRoute(handler))` correctly layers platform gating outside of auth. This means:
1. Platform check happens first (404 if wrong platform — no information leak)
2. Then admin auth check (401/403 if not admin)
3. Then handler logic

The route tree is well-organized with clear auth level separation. One suggestion: consider adding a comment or test that verifies no `secureAdminRoute` endpoint is accidentally registered without the admin wrapper (a "route audit" test that iterates registered routes and asserts auth coverage).

distributions/core-bff/bff/internal/api/middleware_auth.go
```diff
@@ -0,0 +1,196 @@
...
+	"strings"
+
+	"github.com/julienschmidt/httprouter"
+	"github.com/opendatahub-io/odh-dashboard/distributions/core-bff/bff/internal/config"
+	"github.com/opendatahub-io/odh-dashboard/distributions/core-bff/bff/internal/constants"
+	k8s "github.com/opendatahub-io/odh-dashboard/distributions/core-bff/bff/internal/integrations/kubernetes"
+	"github.com/opendatahub-io/odh-dashboard/distributions/core-bff/bff/internal/proxy"
+)
+
+func requiresAuth(path string) bool {
```
andyatmiami: I worry that `requiresAuth` (i.e. an "opt in" to authentication) is the wrong framing here - and could open the door to a very unfortunate (and avoidable) future production incident 😇 

I'd consider maybe we have a `isUnauthenticatedEndpoint()` that checks if we should **NOT** perform auth - and then have the default behavior be to **ALWAYS DO AUTH**..

That way - if someone forgets to update this "allow-list" function - we are still secure by default.

distributions/core-bff/bff/internal/api/middleware_auth.go
```diff
@@ -0,0 +1,196 @@
...
+	})
+}
+
+// auditUser returns the best available user identifier for audit logging.
+// Falls back to "unknown" when the identity has no resolved UserID.
+func auditUser(identity *k8s.RequestIdentity) string {
+	if identity != nil && identity.UserID != "" {
+		return identity.UserID
+	}
+	return "unknown"
```
andyatmiami, line 58: Would we expect this fallback to ever happen for "expected resaons"? 

I wonder if we should add a `WARN` and/or `INFO` log here on this fallback case (if in fact we'd typically NOT expect this to fire..)

distributions/core-bff/bff/internal/api/openshift_user.go
```diff
@@ -0,0 +1,76 @@
...
+	dynClient, err := client.GetDynamicClient()
+	if err != nil {
+		return "", nil, fmt.Errorf("failed to get dynamic client: %w", err)
+	}
+
+	gvr := schema.GroupVersionResource{
+		Group:    "user.openshift.io",
+		Version:  "v1",
+		Resource: "users",
+	}
```
andyatmiami: Probably more maintainable to have a central file declaring various GVRs we are interested in to make identification and management _easier moving forward_ ...

andyatmiami (commented):
For transparency - leaving these comments here for @caponetto - but in near term I plan on working with @lucferbux to hopefully get all this sorted out.

distributions/core-bff/bff/internal/api/audit.go
```diff
@@ -0,0 +1,41 @@
...
+		auditCtx, cancel := context.WithTimeout(context.WithoutCancel(ctx), 10*time.Second)
+		defer cancel()
+		isAdmin, err := client.IsUserAdmin(auditCtx, identity)
```
andyatmiami: Do we really wanna log this `isAdmin` field on each and every API request in this fashion given this is a Kube API server hit? (and right now we are not utilizing any cache or anything..)

distributions/core-bff/bff/internal/api/audit.go
```diff
@@ -0,0 +1,41 @@
...
+	k8s "github.com/opendatahub-io/odh-dashboard/distributions/core-bff/bff/internal/integrations/kubernetes"
+)
+
+// emitAuditLogAsync logs the request details asynchronously without blocking the handler.
+// For non-admin routes, the admin status is resolved in the background via SSAR.
+func (app *App) emitAuditLogAsync(ctx context.Context, client k8s.KubernetesClientInterface, username string, identity *k8s.RequestIdentity, r *http.Request, needsAdmin bool) {
+	method := r.Method
+	url := *r.URL
+
+	go func() {
```
andyatmiami: We probably want to handle backpressure _in some fashion_ in this goroutine...

i.e. semaphore, buffered channel, worker pool, etc - just to ensure concurrency can be bounded/controlled to avoid _thundering herd_ situations..

distributions/core-bff/bff/internal/api/openshift_user.go
```diff
@@ -0,0 +1,76 @@
...
+
+	k8s "github.com/opendatahub-io/odh-dashboard/distributions/core-bff/bff/internal/integrations/kubernetes"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+// resolveUserViaOpenShiftAPI queries user.openshift.io/v1 to resolve the current user's
+// identity. Only available on OpenShift.
+// Returns ("", nil, nil) if the User API is unavailable (e.g., CRD not installed).
+func (app *App) resolveUserViaOpenShiftAPI(ctx context.Context, client k8s.KubernetesClientInterface) (string, []string, error) {
```
andyatmiami, line 19: It doesn't seem like this function is actually invoked by any production code (which for now itself isn't surprising as we are scaffolding all this up!)

Might be nice to include a `// TODO:` that highlights _where we anticipate this function getting used_ to guide others on the overall vision.

distributions/core-bff/bff/internal/api/audit.go
```diff
@@ -0,0 +1,41 @@
...
+// emitAuditLog writes a structured audit log entry.
+func (app *App) emitAuditLog(username string, r *http.Request, needsAdmin bool, isAdmin bool) {
+	app.logger.Info("audit",
+		slog.String("user", username),
+		slog.String("namespace", app.config.Namespace),
+		slog.String("action", r.Method),
+		slog.String("endpoint", r.URL.Path),
+		slog.Bool("isAdmin", isAdmin),
+		slog.Bool("needsAdmin", needsAdmin),
+		slog.String("timestamp", time.Now().UTC().Format(time.RFC3339)),
```
andyatmiami: By default it seems `slog` already produces a `time` field
- https://pkg.go.dev/log/slog

Are we sure we want to introduce a competing `timestamp` field as well?  Seems like it'd be more confusing to have 2 _possibly slightly different values_ under fields named `time` and `timestamp`.

I'd opt to remove this `timestamp` field probably (?)

distributions/core-bff/bff/internal/api/middleware_auth.go
andyatmiami, line 1: `secureRoute` and `secureAdminRoute` duplicate the identity extraction + GetClient + GetUser sequence in the codebase - which feels like that that should be extracted to its own helper and/or D.R.Y'd in some fashion to better prevent code drift...

distributions/core-bff/bff/internal/maputil/maputil.go
```diff
@@ -0,0 +1,53 @@
...
+				continue
+			}
+		}
+		result[k] = v
+	}
+	return result
+}
+
+// DeepCopyMap returns a recursive deep copy of a map[string]any.
+func DeepCopyMap(m map[string]any) map[string]any {
```
andyatmiami, line 34: We probably wanna add a little more context to the function comment calling out how this function is intended to be used right now - as its not as "general purpose" as the function name would imply..

something like: 

```
  // DeepCopyMap returns a recursive copy of a map[string]any.
  // Only nested map[string]any values are deep-copied; other composite types
  // (slices, pointers) are shallow-copied. This is sufficient for the
  // dashboard-config merge path where post-merge mutations target only
  // nested map values, never slices.
```

distributions/core-bff/frontend/src/app/hooks/useHttpProbe.ts
```diff
@@ -0,0 +1,40 @@
...
+  useEffect(() => {
+    if (path === null) {
+      setState({ status: null, loaded: true });
+      return;
+    }
+
+    let cancelled = false;
+    const doProbe = async () => {
+      try {
+        const resp = await fetch(`${URL_PREFIX}${path}`);
```
andyatmiami: Should we use AbortController to actually cancel the in-flight request? Something like:

```
  const controller = new AbortController();
  const resp = await fetch(url, { signal: controller.signal });
  // cleanup:
  return () => { controller.abort(); };
```

andyatmiami (changes requested):
Few more changes to call out and/or discuss based on subsequent analysis in conjunction with Claude...

distributions/core-bff/bff/internal/api/middleware_auth.go
caponetto: The dev fallback is not a full auth bypass because `secureRoute` still validates via `client.GetUser()` (SelfSubjectReview against the kubeconfig client). `DevUser`/`DevGroups` are specifically for audit log labeling, while actual K8s operations go through kubeconfig native auth. A loopback restriction would break containerized dev workflows where `RemoteAddr` is the pod/proxy IP. Good call on the log level though, bumped it from Debug to Warn.

distributions/core-bff/bff/internal/api/middleware_auth.go
caponetto: Good catch. The `!isAdmin` case now returns 403 (Forbidden) instead of 401. Note that the original 401 was intentional to match Fastify (`route-security.ts:36`), but since the frontend does not branch on 401 vs 403 from the dashboard backend's own admin guards, this is a safe improvement. The SSAR error case stays as 401 since we genuinely cannot confirm the user's permissions when the check itself fails. Tests updated accordingly.

distributions/core-bff/bff/internal/api/middleware_auth.go
caponetto: Acknowledged. The per-request `GetUser()` call is intentional since it serves double duty: validating the token is still active and resolving the username for audit logging. Caching the result would skip token validation on cache hits, weakening the security guarantee. SelfSubjectReview is also a lightweight K8s API call (no etcd read, handled by the API server's auth layer), so the overhead should be minimal. Agreed that if monitoring shows this becoming a bottleneck under load, a short TTL cache with careful invalidation would be the right next step.

distributions/core-bff/bff/internal/api/openshift_user.go
caponetto: Good catch. Updated to use `k8sutil.IsDiscoveryError` which already exists in the codebase and is used in `repositories/auth.go` and `repositories/dashboard_config.go` for the same pattern. Discovery errors (CRD not installed) still return `("", nil, nil)`, while real errors (403, timeouts, etc.) now propagate. Note that this function is scaffolded only and not wired into the runtime chain yet, so there was no active risk, but it is the right thing to fix now for consistency.

distributions/core-bff/bff/internal/api/cluster_info.go
caponetto: Good call. Changed the default to XKS for ambiguous probe errors (least privilege). Operators on OpenShift who hit transient probe failures can set `PLATFORM_TYPE` explicitly to override. The warn log now mentions this override option. Test updated accordingly.

distributions/core-bff/bff/internal/api/audit.go
caponetto: Good point. When the SSAR check fails, the audit entry now includes an `adminCheckError` field so log consumers can distinguish "confirmed non-admin" from "couldn't determine." The existing warn log for the failure is kept as well. Thanks for the note on the `context.WithoutCancel` pattern.

distributions/core-bff/bff/internal/api/app_proxy.go
caponetto: Added a check that logs a warning when the kubeconfig is present but has no bearer token, so developers know the fallback won't be effective. Regarding expired/invalid tokens, we can't validate those without calling the K8s API, which is beyond this check's scope.

distributions/core-bff/bff/internal/api/routes.go
caponetto: Thanks! Added a comment above the route registration functions documenting the auth contract so new routes are explicitly covered.

distributions/core-bff/bff/internal/api/middleware_auth.go
caponetto: Good call. Restructured the routing so auth is enforced structurally at the mux level rather than via an allowlist. Unauthenticated paths are now explicitly wrapped in `publicRoute()` and authenticated paths use `authedHandler`, making intent visible at the registration site. New service prefixes are secure by default. Added a route-level test with real auth enabled to verify the separation.

distributions/core-bff/bff/internal/api/middleware_auth.go
caponetto: Good call. Added a WARN log when the fallback fires. This only runs in error paths (GetClient/GetUser failure) where the UserID hasn't been resolved, so it won't be noisy but will help operators spot patterns like bad tokens or k8s client issues.

distributions/core-bff/bff/internal/api/openshift_user.go
caponetto: Done. Moved the OpenShift User and ClusterVersion GVRs into `internal/models/constants.go` alongside the existing ones (Auth, Notebook, OdhApplication, DashboardConfig). Updated all references in source and test files.

distributions/core-bff/bff/internal/api/audit.go
caponetto: Agreed, not worth the cost. Removed the background SSAR call so regular API requests no longer hit the k8s API just to populate the audit log. Non-admin routes omit the `isAdmin` field entirely since no check is performed. The admin route still logs accurate admin status because it already has the SSAR result from its authorization guard, so no extra call is made.

distributions/core-bff/bff/internal/api/audit.go
caponetto: No longer applicable. The async goroutine was removed as part of the SSAR cleanup above, so audit logging is now fully synchronous in both paths. No unbounded concurrency to manage.

distributions/core-bff/bff/internal/api/openshift_user.go
caponetto: Added a TODO noting the intent to wire it into the auth middleware for enriching audit logs with OpenShift identity.

distributions/core-bff/bff/internal/api/audit.go
caponetto: Agreed. Removed the manual `timestamp` field since `slog` already includes a `time` field on every record. Updated tests accordingly.

distributions/core-bff/bff/internal/api/middleware_auth.go
caponetto: Extracted the shared identity extraction, client creation, and user resolution into a `resolveRequestAuth` helper. Both routes now call it and only diverge afterward: `secureRoute` emits the audit log and proceeds, `secureAdminRoute` continues with the SSAR admin check.

distributions/core-bff/bff/internal/maputil/maputil.go
caponetto: Updated the comment to call out that only nested maps are deep-copied and that slices/pointers are shallow-copied, noting this is sufficient for the dashboard-config merge path.

distributions/core-bff/frontend/src/app/hooks/useHttpProbe.ts
caponetto: Done. Replaced the `cancelled` boolean with an `AbortController` so the in-flight fetch is actually cancelled on cleanup, not just ignored.

distributions/core-bff/bff/internal/api/routes.go
```diff
@@ -88,51 +102,75 @@ func (app *App) Routes() http.Handler {
...
-		combinedMux.HandleFunc(OpenAPIPath, app.openAPI.HandleOpenAPIRedirectWrapper)
+		mux.Handle(SwaggerUIPath, app.publicRouteFunc(app.openAPI.HandleSwaggerUIWrapper))
+		mux.Handle(OpenAPIPath, app.publicRouteFunc(app.openAPI.HandleOpenAPIRedirectWrapper))
 	}
-	combinedMux.Handle("/", app.RecoverPanic(app.EnableTelemetry(app.EnableCORS(app.InjectRequestIdentity(appMux)))))
+	mux.Handle("/", app.publicRoute(staticHandler))
+
+	// Authenticated routes (token required)
+	mux.Handle(APIPathPrefix+"/", authedHandler)
+	mux.Handle(PathPrefix+"/", authedHandler)
```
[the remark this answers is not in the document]
caponetto: Good catch. Added exact-path handlers for `/api` and `/core-bff` in the top-level mux so requests to the bare roots return 404 instead of being 307-redirected to the subtree patterns. Added a test for the `/core-bff` case.

distributions/core-bff/frontend/src/app/hooks/useHttpProbe.ts
```diff
@@ -0,0 +1,38 @@
...
+  useEffect(() => {
+    if (path === null) {
+      setState({ status: null, loaded: true });
+      return;
+    }
+
+    const controller = new AbortController();
+    const doProbe = async () => {
+      try {
+        const resp = await fetch(`${URL_PREFIX}${path}`, { signal: controller.signal });
```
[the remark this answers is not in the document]
caponetto: Skipping for now. This core-bff frontend is currently just a test harness for validating the BFF, not a production UI. It does not use React Query (not in the dependencies) and has no established API client pattern to migrate to. The referenced coding guidelines apply to the main dashboard frontend, not this distribution. We did add AbortController support to the hook per the earlier comment.

caponetto:
@lucferbux @andyatmiami Thanks for the review! I've addressed the comments. Please let me know if we need anything else.

andyatmiami (approved):
/lgtm
/approve

test plan exercised (with all behavior working as expected):
- [pr-7953-manual-api-test-plan.md](https://github.com/user-attachments/files/29603343/pr-7953-manual-api-test-plan.md)

access logs from testing:

```
➜ bff/ git:((HEAD detached at guilherme/RHOAIENG-67047)) $ go run ./cmd/ \
  -auth-method user_token \
  -auth-token-header x-forwarded-access-token \
  -namespace opendatahub \
  -dev-mode=false \
  -insecure-skip-verify=true \
  -port 4000
time=2026-07-02T08:55:36.662-04:00 level=INFO msg="Auto-detected platform" platform=XKS
time=2026-07-02T08:55:36.662-04:00 level=INFO msg="Using real BFF client factory"
time=2026-07-02T08:55:36.672-04:00 level=INFO msg="starting server" addr=:4000 "TLS enabled"=false
time=2026-07-02T08:57:15.383-04:00 level=WARN msg="Unauthorized access attempt" error="missing required header: x-forwarded-access-token" method=GET uri=/api/status
time=2026-07-02T08:57:31.334-04:00 level=WARN msg="Unauthorized access attempt" error="missing required header: x-forwarded-access-token" method=GET uri=/core-bff/openapi
time=2026-07-02T08:57:51.057-04:00 level=WARN msg="Unauthorized access attempt" error="missing required header: x-forwarded-access-token" method=GET uri=/core-bff/swagger-ui
time=2026-07-02T09:00:20.328-04:00 level=ERROR msg="failed to get user identity from token" error=Unauthorized
time=2026-07-02T09:00:20.328-04:00 level=WARN msg="could not resolve user identity for audit log"
time=2026-07-02T09:00:20.328-04:00 level=INFO msg=audit user=unknown namespace=opendatahub action=GET endpoint=/api/status needsAdmin=false
time=2026-07-02T09:00:20.328-04:00 level=WARN msg="Unauthorized access attempt" error="invalid or expired token: failed to get user identity: Unauthorized" method=GET uri=/api/status
time=2026-07-02T10:06:58.047-04:00 level=WARN msg="Unauthorized access attempt" error="missing required header: x-forwarded-access-token" method=GET uri=/api/status
time=2026-07-02T10:07:27.350-04:00 level=WARN msg="Unauthorized access attempt" error="missing required header: x-forwarded-access-token" method=GET uri=/api/status
time=2026-07-02T10:07:38.323-04:00 level=WARN msg="Unauthorized access attempt" error="missing required header: x-forwarded-access-token" method=GET uri=/api/status
time=2026-07-02T10:08:13.662-04:00 level=INFO msg=audit user=bff-test-user namespace=opendatahub action=GET endpoint=/api/status needsAdmin=false
time=2026-07-02T10:11:45.822-04:00 level=INFO msg=audit user=bff-test-admin namespace=opendatahub action=GET endpoint=/api/status needsAdmin=false
time=2026-07-02T10:11:55.567-04:00 level=INFO msg=audit user=bff-test-admin namespace=opendatahub action=GET endpoint=/api/cluster-settings needsAdmin=true isAdmin=true
time=2026-07-02T10:12:11.487-04:00 level=INFO msg=audit user=bff-test-admin namespace=opendatahub action=GET endpoint=/api/config needsAdmin=false
time=2026-07-02T10:12:22.990-04:00 level=INFO msg=audit user=bff-test-user namespace=opendatahub action=GET endpoint=/api/config needsAdmin=false
time=2026-07-02T10:13:16.779-04:00 level=INFO msg=audit user=bff-test-admin namespace=opendatahub action=PATCH endpoint=/api/config needsAdmin=true isAdmin=true
time=2026-07-02T10:13:32.455-04:00 level=INFO msg=audit user=bff-test-admin namespace=opendatahub action=GET endpoint=/api/components needsAdmin=false
time=2026-07-02T10:13:39.793-04:00 level=INFO msg=audit user=bff-test-admin namespace=opendatahub action=GET endpoint=/api/connection-types needsAdmin=false
time=2026-07-02T10:14:02.391-04:00 level=INFO msg=audit user=bff-test-admin namespace=opendatahub action=GET endpoint=/api/dashboardConfig/opendatahub/odh-dashboard-config needsAdmin=true isAdmin=true
time=2026-07-02T10:15:21.241-04:00 level=INFO msg=audit user=bff-test-user namespace=opendatahub action=GET endpoint=/api/config needsAdmin=false
time=2026-07-02T10:15:37.837-04:00 level=INFO msg=audit user=bff-test-user namespace=opendatahub action=GET endpoint=/api/components needsAdmin=false
time=2026-07-02T10:15:43.786-04:00 level=INFO msg=audit user=bff-test-user namespace=opendatahub action=GET endpoint=/api/connection-types needsAdmin=false
time=2026-07-02T10:15:57.956-04:00 level=INFO msg=audit user=bff-test-user namespace=opendatahub action=GET endpoint=/api/connection-types/test-connection-type needsAdmin=false
time=2026-07-02T10:16:08.420-04:00 level=INFO msg=audit user=bff-test-user namespace=opendatahub action=PATCH endpoint=/api/config needsAdmin=true isAdmin=false
time=2026-07-02T10:16:08.420-04:00 level=WARN msg="Forbidden access attempt" error="insufficient permissions to make this request" method=PATCH uri=/api/config
time=2026-07-02T10:16:23.378-04:00 level=INFO msg=audit user=bff-test-user namespace=opendatahub action=GET endpoint=/api/cluster-settings needsAdmin=true isAdmin=false
time=2026-07-02T10:16:23.378-04:00 level=WARN msg="Forbidden access attempt" error="insufficient permissions to make this request" method=GET uri=/api/cluster-settings
time=2026-07-02T10:16:29.562-04:00 level=INFO msg=audit user=bff-test-user namespace=opendatahub action=PUT endpoint=/api/cluster-settings needsAdmin=true isAdmin=false
time=2026-07-02T10:16:29.562-04:00 level=WARN msg="Forbidden access attempt" error="insufficient permissions to make this request" method=PUT uri=/api/cluster-settings
time=2026-07-02T10:16:41.496-04:00 level=INFO msg=audit user=bff-test-user namespace=opendatahub action=GET endpoint=/api/dashboardConfig/opendatahub/odh-dashboard-config needsAdmin=true isAdmin=false
time=2026-07-02T10:16:41.496-04:00 level=WARN msg="Forbidden access attempt" error="insufficient permissions to make this request" method=GET uri=/api/dashboardConfig/opendatahub/odh-dashboard-config
time=2026-07-02T10:17:04.789-04:00 level=INFO msg=audit user=bff-test-user namespace=opendatahub action=PATCH endpoint=/api/dashboardConfig/opendatahub/odh-dashboard-config needsAdmin=true isAdmin=false
time=2026-07-02T10:17:04.789-04:00 level=WARN msg="Forbidden access attempt" error="insufficient permissions to make this request" method=PATCH uri=/api/dashboardConfig/opendatahub/odh-dashboard-config
time=2026-07-02T10:17:30.583-04:00 level=INFO msg=audit user=bff-test-user namespace=opendatahub action=POST endpoint=/api/connection-types needsAdmin=true isAdmin=false
time=2026-07-02T10:17:30.583-04:00 level=WARN msg="Forbidden access attempt" error="insufficient permissions to make this request" method=POST uri=/api/connection-types
time=2026-07-02T10:17:42.274-04:00 level=INFO msg=audit user=bff-test-user namespace=opendatahub action=DELETE endpoint=/api/connection-types/test needsAdmin=true isAdmin=false
time=2026-07-02T10:17:42.274-04:00 level=WARN msg="Forbidden access attempt" error="insufficient permissions to make this request" method=DELETE uri=/api/connection-types/test
time=2026-07-02T10:18:01.222-04:00 level=INFO msg=audit user=bff-test-user namespace=opendatahub action=GET endpoint=/api/components/remove needsAdmin=true isAdmin=false
time=2026-07-02T10:18:01.222-04:00 level=WARN msg="Forbidden access attempt" error="insufficient permissions to make this request" method=GET uri="/api/components/remove?appName=test"
time=2026-07-02T10:33:54.631-04:00 level=INFO msg=audit user=bff-test-admin namespace=opendatahub action=GET endpoint=/api/connection-types needsAdmin=false
time=2026-07-02T10:34:18.786-04:00 level=INFO msg=audit user=bff-test-admin namespace=opendatahub action=POST endpoint=/api/connection-types needsAdmin=true isAdmin=true
time=2026-07-02T10:34:35.482-04:00 level=INFO msg=audit user=bff-test-admin namespace=opendatahub action=GET endpoint=/api/connection-types/test-ct needsAdmin=false
time=2026-07-02T10:35:44.166-04:00 level=INFO msg=audit user=bff-test-admin namespace=opendatahub action=PATCH endpoint=/api/connection-types/test-ct needsAdmin=true isAdmin=true
time=2026-07-02T10:36:02.358-04:00 level=INFO msg=audit user=bff-test-admin namespace=opendatahub action=PUT endpoint=/api/connection-types/test-ct needsAdmin=true isAdmin=true
time=2026-07-02T10:36:06.397-04:00 level=INFO msg=audit user=bff-test-admin namespace=opendatahub action=GET endpoint=/api/connection-types/test-ct needsAdmin=false
time=2026-07-02T10:36:21.262-04:00 level=INFO msg=audit user=bff-test-admin namespace=opendatahub action=DELETE endpoint=/api/connection-types/test-ct needsAdmin=true isAdmin=true
time=2026-07-02T10:36:24.987-04:00 level=INFO msg=audit user=bff-test-admin namespace=opendatahub action=GET endpoint=/api/connection-types/test-ct needsAdmin=false
time=2026-07-02T10:38:19.295-04:00 level=INFO msg=audit user=bff-test-admin namespace=opendatahub action=PATCH endpoint=/api/config needsAdmin=true isAdmin=true
time=2026-07-02T10:38:34.061-04:00 level=INFO msg=audit user=bff-test-admin namespace=opendatahub action=PATCH endpoint=/api/config needsAdmin=true isAdmin=true
time=2026-07-02T10:38:48.970-04:00 level=INFO msg=audit user=bff-test-admin namespace=opendatahub action=PUT endpoint=/api/cluster-settings needsAdmin=true isAdmin=true
time=2026-07-02T10:39:04.353-04:00 level=INFO msg=audit user=bff-test-admin namespace=opendatahub action=PUT endpoint=/api/cluster-settings needsAdmin=true isAdmin=true
time=2026-07-02T10:39:23.985-04:00 level=INFO msg=audit user=bff-test-admin namespace=opendatahub action=GET endpoint=/api/connection-types/test-ct-nope needsAdmin=false

```

PR #7962, opened 2026-06-11
fix(ci): make Konflux simulator actually test the PR image and assert…

antowaddle:
## Description
Two bugs meant Phase 4 was always testing the published main image rather than the PR's changes, and would pass even if the container crashed on startup.

Gap 1 — wrong image: the kustomize manifests set imagePullPolicy: Always on every container, so kind load alone was not enough — Kubernetes would pull the live quay.io:main image from the registry, bypassing the locally built image entirely. Fix: after applying manifests, patch the odh-dashboard container to image=odh-dashboard:odh-test with imagePullPolicy: Never so Kubernetes is forced to use only what was built from the PR's source.

Gap 2 — no health assertion: the check only counted pulled images and exited 0 regardless of CrashLoopBackOff. Fix: explicitly check that the odh-dashboard container is in running state with 0 restarts; any crash fails the job with the container logs printed above the error for immediate diagnosis.

Additional fixes:
- Add disk space cleanup (rm dotnet/android/ghc/boost + docker prune) before Phase 0 and both Phase 1 jobs to eliminate flaky "no space left on device" Docker build failures on GitHub-hosted runners
- Align actions/setup-node to v4.3.0 (same SHA as test.yml) and use NODE_VERSION env var for consistency; drops FORCE_JAVASCRIPT_ACTIONS_TO_NODE24
- Fix pre-existing shellcheck SC2086 ($GITHUB_OUTPUT unquoted) and SC2012 (ls replaced with find for app.*.js glob)

## How Has This Been Tested?
- CI

## Test Impact
- None

## Request review criteria:

If you have UI changes: 

## Summary by CodeRabbit

* **Chores**
  * Improved CI reliability: consistent Node.js setup, proactive disk-space cleanup before key build steps, and more robust artifact discovery and skip-condition handling.

* **Tests**
  * Hardened integration checks: wait for deployments properly, surface relevant container logs on failure, and assert container running state and zero restarts to ensure deployment health.


.github/workflows/pr-build-validation.yml
```diff
@@ -52,10 +55,24 @@ jobs:
...
+          df -h
+          echo "::endgroup::"
+          # Remove large pre-installed toolchains that are unused in this workflow
+          sudo rm -rf /usr/share/dotnet /usr/local/lib/android /opt/ghc \
+            /usr/local/share/boost /usr/local/graalvm /usr/local/.ghcup \
+            /usr/share/swift || true
+          docker system prune -af || true
+          echo "::group::Disk usage after cleanup"
+          df -h
+          echo "::endgroup::"
```
lucferbux: **suggestion (low):** This identical 12-line cleanup block is repeated three times (here, line 254, line 300). If the directory list changes, all three need updating in sync.

Consider extracting this to a composite action under `.github/actions/free-disk-space/action.yml` so changes are made in one place. Not a blocker, but reduces maintenance burden.

.github/workflows/pr-build-validation.yml
```diff
@@ -673,6 +716,22 @@ jobs:
...
+          # live quay.io:main image from the registry and we'd never test the PR.
+          # Using imagePullPolicy: Never forces Kubernetes to use only Kind's local store;
+          # it will error immediately if the image isn't there, making the intent explicit.
+          echo "Patching odh-dashboard container to use PR-built image..."
+          kubectl set image deployment/odh-dashboard \
+            odh-dashboard=odh-dashboard:odh-test \
+            -n opendatahub
+          kubectl patch deployment odh-dashboard -n opendatahub \
+            --type=strategic \
+            -p='{"spec":{"template":{"spec":{"containers":[{"name":"odh-dashboard","imagePullPolicy":"Never"}]}}}}'
```
lucferbux: **suggestion (medium): potential double-rollout race condition.** This image patch triggers a new rollout, and then the resource-limits patch at line 738 triggers *another* rollout. The first pod starts with the correct image but then gets killed by the second patch.

This likely works in practice because the 60s sleep downstream covers both rollouts, but it is fragile. Consider either:
1. Combining both patches into a single strategic-merge patch so only one rollout happens (image + imagePullPolicy + resources all in one `kubectl patch`).
2. Or moving this block *after* the resource-limits patch so the final rollout uses the correct image.

.github/workflows/pr-build-validation.yml
```diff
@@ -749,14 +808,46 @@ jobs:
...
+          # Assert that the PR-built odh-dashboard container is actually running and has
+          # not restarted. A CrashLoopBackOff or any restart count > 0 means the server
+          # failed to start and is a hard failure.
+          echo "::group::Health check: odh-dashboard container"
+          DASHBOARD_RUNNING=$(kubectl get pods -n opendatahub -l app=odh-dashboard \
+            -o jsonpath='{.items[0].status.containerStatuses[?(@.name=="odh-dashboard")].state.running}' \
+            2>/dev/null || echo "")
+          DASHBOARD_RESTARTS=$(kubectl get pods -n opendatahub -l app=odh-dashboard \
+            -o jsonpath='{.items[0].status.containerStatuses[?(@.name=="odh-dashboard")].restartCount}' \
+            2>/dev/null || echo "0")
```
lucferbux: **suggestion (low):** If the `kubectl get pods` command fails entirely (e.g., pod not found, API server timeout), this silently defaults to `"0"` and the restart check passes. Consider defaulting to a sentinel like `"-1"` instead and then failing explicitly if the value is negative or non-numeric. This prevents masking infrastructure issues as a healthy pass.

.github/workflows/pr-build-validation.yml
```diff
@@ -749,14 +808,46 @@ jobs:
             exit 1
           fi
 
-          echo "✅ PASS: Containers attempted to start, $IMAGES_PULLED image(s) pulled"
+          echo "✅ PASS: $IMAGES_PULLED image(s) pulled"
```
lucferbux, line 780: **suggestion (medium):** Now that the image patch above triggers a new rollout, the fixed 60-second sleep in the previous step may not be enough time for: old pod terminating → new pod scheduling → image loading → container startup.

Consider adding a `kubectl rollout status` call before this validation step (or at the start of this step):
```bash
kubectl rollout status deployment/odh-dashboard -n opendatahub --timeout=120s
```
This waits for the rollout to actually complete rather than hoping 60s is enough.

lucferbux (commented):
Nice catch on both gaps — Phase 4 was essentially a no-op before this. The core fixes (wrong image + missing health assertion) are correct and important. A few suggestions below, none are blockers but the double-rollout one is worth considering before merge.

.github/workflows/pr-build-validation.yml
antowaddle: Resolved in latest commit.

.github/workflows/pr-build-validation.yml
antowaddle: Resolved in latest commit.

.github/workflows/pr-build-validation.yml
antowaddle: Resolved in latest commit.

.github/workflows/pr-build-validation.yml
antowaddle: Resolved in latest commit.

PR #7975, opened 2026-06-11
Fix(RHOAIENG-58677): Global search issues and lineage toolbar filtering improvements

claudialphonse78:
https://issues.redhat.com/browse/RHOAIENG-58677

## Description

Bug fixes for Feature Store global search and lineage toolbar. No new features.

---

### Global Search

- **AbortController moved from `useState` to `useRef`** — state updates are async, so closures inside promise callbacks were holding stale controller references. `useRef` keeps the current value accessible synchronously inside async callbacks.

- **`hasMorePages` not cleared on new search** — if the previous search had more pages, `loadMoreResults` could slip past its guard during the async gap before new results arrived, overwriting the in-flight search's abort controller. Now reset immediately when a new search starts.

- **`setIsSearching(false)` was called twice** — the "no projects" early-return path called it explicitly and `finally` called it again. Removed the redundant one.

- **API errors were silently dropped** — the search API can return soft errors alongside results. Now surfaced via `searchErrors` in the hook return. UI consumer to be wired up in a follow-on.

- **Highlight regex `g` flag removed (`GlobalSearchInput.tsx`)** — `g` makes `RegExp` stateful. Consecutive `.test()` calls alternated `true`/`false`, so every other result in the list wasn't getting highlighted. `split()` finds all occurrences without the flag.

- **`feature` results with no `featureView` navigated to a broken URL (`searchUtils.ts`)** — falls back to the project's Features list page instead.

- **Debounce timer wasn't cancelled on search clear (`useSearchHandlers.ts`)** — typing then immediately clearing would still fire the queued search. Added `clearTimeout` in all clear paths. Also removed the now-dead `featureView ?? ''` fallback.

---

### Lineage

- **`setState` inside `useMemo` removed (`FeatureStoreLineageComponent.tsx`)** — `conversionError` was being set via a state setter inside `useMemo`. Refactored to return `{ visualizationData, conversionError }` directly from the memo.

- **Filter values stored as `string[]` instead of comma-joined strings** — selections were joined with `", "` on write and split on read. Any name containing a comma silently broke filtering. Each value is now its own array element. The `join(', ')` is still used but only for chip display — not for storage or matching.

---

### API hooks

- **Error flash on reload fixed (`useFeatureStoreProjectsAPI.ts`)** — when `apiAvailable` is `false` during initial page load, the hook was rejecting with `Error('API not yet available')`. This caused a visible error banner before the API discovery completed (~200ms later). Changed back to `Promise.resolve(DEFAULT_PROJECT_LIST)` — stays quiet during discovery, loads real data once the API is ready.

- **Wrong error message (`useFeatureServiceByName.ts`)** — "Feature name is required" → "Feature service name is required"

---

## How Has This Been Tested?

- `tsc` and `eslint` pass clean
- Added 13 unit tests for `useFeatureStoreSearch`:
  - Initial state, empty query, API unavailable, no projects
  - Successful search with full result shape validation
  - API error surfacing via `searchErrors`
  - Race condition — stale results from a superseded search are discarded
  - `clearSearch` — state reset and in-flight abort verified
  - `loadMoreResults` — page append and no-op when `hasMorePages` is false
- Existing tests for `useFeatureStoreProjectsAPI` and `searchUtils` still pass
- Manually tested on a local cluster via the port-forwarding script:
  - Page reload no longer shows error flash before API is ready
  - Global search highlights all matching results (not alternating)
  - Selecting a feature result navigates correctly
  - Lineage multi-select filters work and persist correctly

## Test Impact

- **Added:** `useFeatureStoreSearch.spec.ts` — 13 unit tests
- **Updated:** `searchUtils.spec.ts` — corrected expected fallback URL
- **Updated:** `useFeatureStoreProjectsAPI.spec.ts` — whitespace only

Self checklist:

If you have UI changes:

## Release Notes

* **New Features**
  * Search now surfaces error messages when queries fail.

* **Bug Fixes**
  * Improved global search cancellation to prevent outdated results after closing or clearing.
  * Corrected navigation when viewing features without a specific feature view.
  * Updated lineage search filters to use consistent multi-select array values, improving matching and highlighting behavior.
  * Corrected the required-field message for missing feature service name.

* **Tests**
  * Added comprehensive coverage for search lifecycle, pagination, cancellation/race conditions, and updated route generation expectations.

### DIFF
--- a/packages/feature-store/src/apiHooks/__tests__/useFeatureStoreSearch.spec.ts
+++ b/packages/feature-store/src/apiHooks/__tests__/useFeatureStoreSearch.spec.ts
@@ -1,0 +1,373 @@
+/* eslint-disable camelcase */
+import { act, renderHook, waitFor } from '@testing-library/react';
+import {
+  mockGlobalSearchResponse,
+  mockGlobalSearchPagination,
+  mockGlobalSearchResult,
+} from '../../__mocks__/mockGlobalSearch';
+import { mockFeatureStoreProject } from '../../__mocks__/mockFeatureStoreProject';
+import { ProjectList } from '../../types/featureStoreProjects';
+import { GlobalSearchResponse } from '../../types/search';
+import { useFeatureStoreSearch } from '../useFeatureStoreSearch';
+
+jest.mock('../useGlobalSearch', () => ({
+  __esModule: true,
+  default: jest.fn(),
+}));
+
+jest.mock('../useFeatureStoreProjects', () => ({
+  __esModule: true,
+  default: jest.fn(),
+}));
+
+// eslint-disable-next-line @typescript-eslint/no-require-imports
+const mockUseGlobalSearch = jest.mocked(require('../useGlobalSearch').default);
+// eslint-disable-next-line @typescript-eslint/no-require-imports
+const mockUseFeatureStoreProjects = jest.mocked(require('../useFeatureStoreProjects').default);
+
+const mockSearch = jest.fn();
+
+const mockProjectList: ProjectList = {
+  projects: [mockFeatureStoreProject({ spec: { name: 'project-1' } })],
+  pagination: {
+    page: 1,
+    limit: 10,
+    total_count: 1,
+    total_pages: 1,
+    has_next: false,
+    has_previous: false,
+  },
+};
+
+const emptyProjectList: ProjectList = {
+  projects: [],
+  pagination: {
+    page: 0,
+    limit: 0,
+    total_count: 0,
+    total_pages: 0,
+    has_next: false,
+    has_previous: false,
+  },
+};
+
+const makeProjectsState = (overrides?: Partial<{ projects: ProjectList; loaded: boolean }>) => ({
+  data: overrides?.projects ?? mockProjectList,
+  loaded: overrides?.loaded ?? true,
+  error: undefined,
+  refresh: jest.fn(),
+});
+
+describe('useFeatureStoreSearch', () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+    mockUseGlobalSearch.mockReturnValue({ search: mockSearch, apiAvailable: true });
+    mockUseFeatureStoreProjects.mockReturnValue(makeProjectsState());
+  });
+
+  it('should return initial state', () => {
+    const { result } = renderHook(() => useFeatureStoreSearch());
+
+    expect(result.current.convertedSearchData).toEqual([]);
+    expect(result.current.isSearching).toBe(false);
+    expect(result.current.isLoadingMore).toBe(false);
+    expect(result.current.hasMorePages).toBe(false);
+    expect(result.current.totalCount).toBe(0);
+    expect(result.current.searchErrors).toEqual([]);
+  });
+
+  describe('handleSearchChange', () => {
+    it('should reset state when query is empty', async () => {
+      mockSearch.mockResolvedValueOnce(
+        mockGlobalSearchResponse({
+          results: [mockGlobalSearchResult({ name: 'user_id' })],
+          pagination: mockGlobalSearchPagination({ totalCount: 1 }),
+        }),
+      );
+
+      const { result } = renderHook(() => useFeatureStoreSearch());
+
+      await act(async () => {
+        await result.current.handleSearchChange('user');
+      });
+      expect(result.current.convertedSearchData).toHaveLength(1);
+
+      await act(async () => {
+        await result.current.handleSearchChange('');
+      });
+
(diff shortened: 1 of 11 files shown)


claudialphonse78:

packages/feature-store/src/apiHooks/useFeatureStoreSearch.ts
```diff
@@ -60,42 +58,49 @@ export const useFeatureStoreSearch = (): {
...
         setCurrentPage(1);
         setHasMorePages(false);
         setTotalCount(0);
-        return;
-      }
-
-      if (currentSearchQuery === query && isSearching) {
+        setSearchErrors([]);
         return;
       }
```
dpanshug, line 79: Also add `setIsSearching(false)`

packages/feature-store/src/apiHooks/useFeatureStoreSearch.ts
```diff
@@ -35,21 +36,18 @@ export const useFeatureStoreSearch = (): {
...
+  // avoiding stale-closure race conditions that `useState` would introduce.
+  const abortControllerRef = useRef<AbortController | null>(null);
 
   const convertedSearchData = useMemo(() => {
-    if (allResults.length === 0) {
-      return [];
-    }
-
-    if (!Array.isArray(allResults)) {
+    if (allResults.length === 0 || !Array.isArray(allResults)) {
```
dpanshug: nit, accessing length before isArray is wrong order

```suggestion
if (!Array.isArray(allResults) || allResults.length === 0) {
```

packages/feature-store/src/components/FeatureStoreLineageToolbar.tsx
```diff
@@ -256,7 +256,8 @@ const FeatureStoreLineageToolbar: React.FC<FeatureStoreLineageToolbarProps> = ({
           delete newFilters[filterType];
         } else {
           const filterValue = typeof value === 'string' ? value : value.value;
-          newFilters[filterType] = filterValue;
+          // Wrap single string from FilterToolbar chip into an array
+          newFilters[filterType] = [filterValue];
```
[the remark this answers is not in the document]
dpanshug: @claudialphonse78 check this again

packages/feature-store/src/types/toolbarTypes.ts
```diff
@@ -54,10 +54,10 @@ export type MultipleLabelForType = Array<{ key: string; onRemove: () => void }>;
...
 // Feature Store Lineage toolbar specific types
 export interface FeatureStoreLineageSearchFilters {
-  entity?: string;
-  featureView?: string;
-  dataSource?: string;
-  featureService?: string;
+  entity?: string[];
+  featureView?: string[];
+  dataSource?: string[];
+  featureService?: string[];
```
dpanshug, line 60: Why do we need to change the string to array? In FeatureStoreLineageToolbar.tsx lines 274-282, we are using .join(). So we might need to change one of these

packages/feature-store/src/components/FeatureStoreLineageToolbar.tsx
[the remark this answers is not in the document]
claudialphonse78: The old code stored selections as "entity1, entity2" and split on ", " to filter.
Any name containing a comma would silently break the match.
string[] type fixes it, each selection is its own item, no splitting needed.
The join() is only for the chip label display, never used for filtering.

packages/feature-store/src/apiHooks/__tests__/useFeatureStoreSearch.spec.ts
```diff
@@ -0,0 +1,426 @@
...
+        void result.current.handleSearchChange('first');
+      });
+      act(() => {
+        void result.current.handleSearchChange('second');
+      });
+
+      // Resolve the second (newer) search first
+      await act(async () => {
+        resolveSecond(secondResponse);
+      });
```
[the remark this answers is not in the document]
claudialphonse78: this is not a valid reivew

packages/feature-store/src/types/toolbarTypes.ts
claudialphonse78: The join() is just for showing the chip label in the UI — it's never used for 
actual filtering.

We changed to string[] because the old code stored selections like "entity1, entity2" 
as one string and split it back to compare. If any entity name had a comma in it, 
the split would break silently and the filter would stop working. 
Storing as an array avoids that problem completely.

dpanshug (approved):
/lgtm - tested all the scenarios of this fix, working perfectly

PR #7978, opened 2026-06-11
RHOAIENG-68244: Cypress e2e CI fails with File name too long when auto-detected tags are used as results directory name

paulovmr:
https://redhat.atlassian.net/browse/RHOAIENG-68244

## Description

When a PR triggers many auto-detected Cypress tags, they get consolidated into a single space-separated string (e.g., `@AgentOpsCI @AutoMLCI @AutoRAGCI @ConnectionTypesCI ...`) used as `matrix.tag`. This string is used directly as the results directory name and artifact name in the e2e workflow, exceeding the Linux 255-character filename limit and causing `mkdir` to fail with "File name too long".

This fix computes a short 12-character SHA256 hash of the tag string when it exceeds 200 characters, and uses that as the directory/artifact name instead. Short/single tags (e.g., `@NonAdmin`) remain human-readable since they stay under the threshold.

## How Has This Been Tested?

  - Verified the hash logic locally: `echo -n "@AgentOpsCI @AutoMLCI ..." | sha256sum | cut -c1-12` produces a stable, deterministic 12-char identifier
  - The workflow file change cannot be tested via PR CI since GitHub Actions reads workflow definitions from the base branch (`main`) for `pull_request` events — the fix takes effect after merge

## Test Impact

No tests added — this is a CI workflow fix, not application code. The change is a shell conditional in a GitHub Actions step with no corresponding unit test infrastructure.

## Request review criteria:

If you have UI changes: 

## Summary by CodeRabbit

* **Chores**
  * Standardized CI tag handling and added validation to enforce tag format.
  * Test selection and test-run configuration now follow the provided tag value.
  * Test results are organized into a normalized directory name (handles very long tags) for reliable storage.
  * Uploaded test artifacts use matching, normalized names to improve traceability.
  * Completion logs now display the test tag for easier troubleshooting.

### DIFF
--- a/.github/workflows/cypress-e2e-test.yml
+++ b/.github/workflows/cypress-e2e-test.yml
@@ -530,6 +530,17 @@ jobs:
       CLUSTER_NAME: ${{ needs.select-cluster.outputs.cluster_name }}
       MATRIX_TAG: ${{ matrix.tag }}
     steps:
+      - name: Validate tag format
+        run: |
+          for tag in $MATRIX_TAG; do
+            if [[ ! "$tag" =~ ^@[a-zA-Z0-9_-]+$ ]]; then
+              echo "❌ Invalid tag format (security): $tag"
+              echo "   Tags must match: ^@[a-zA-Z0-9_-]+$"
+              exit 1
+            fi
+          done
+          echo "✅ Tag format valid: $MATRIX_TAG"
+
       - name: Check Disk Space
         run: |
           echo "📊 Checking available disk space..."
@@ -1259,12 +1270,6 @@ jobs:
           cd frontend
 
           TAG="$MATRIX_TAG"
-          if [[ ! "$TAG" =~ ^[@a-zA-Z0-9_\ -]+$ ]]; then
-            echo "❌ Invalid tag format (security): $TAG"
-            echo "   Tags must contain only: @ a-z A-Z 0-9 _ - space"
-            exit 1
-          fi
-
           if [ ${#TAG} -gt 200 ]; then
             TAG_DIR=$(echo -n "$TAG" | sha256sum | cut -c1-12)
             echo "📁 Tag too long for directory name (${#TAG} chars), using hash: ${TAG_DIR}"
--- a/.github/workflows/cypress-e2e-test.yml
+++ b/.github/workflows/cypress-e2e-test.yml
@@ -528,6 +528,7 @@ jobs:
         tag: ${{ fromJson(needs.get-test-tags.outputs.matrix) }}
     env:
       CLUSTER_NAME: ${{ needs.select-cluster.outputs.cluster_name }}
+      MATRIX_TAG: ${{ matrix.tag }}
     steps:
       - name: Check Disk Space
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/cypress-e2e-test.yml
```diff
@@ -528,7 +528,19 @@ jobs:
         tag: ${{ fromJson(needs.get-test-tags.outputs.matrix) }}
     env:
       CLUSTER_NAME: ${{ needs.select-cluster.outputs.cluster_name }}
+      MATRIX_TAG: ${{ matrix.tag }}
     steps:
+      - name: Validate tag format
+        run: |
+          for tag in $MATRIX_TAG; do
+            if [[ ! "$tag" =~ ^@[a-zA-Z0-9_-]+$ ]]; then
```
antowaddle: We currently are not prescriptive with the tagging used by tests in our Cypress e2e suite, this suggested change would block tags (for example) that contain dots (@test.tag,@model-catalog.v2) or additional tags manually inputted such as @ModelRegistryCI, @my_test:v2.

Please review this again.

.github/workflows/cypress-e2e-test.yml
```diff
@@ -528,7 +528,19 @@ jobs:
...
+              echo "❌ Invalid tag format (security): $tag"
```
antowaddle: Why 'security'? I don't understand the naming here.

.github/workflows/cypress-e2e-test.yml
```diff
@@ -528,7 +528,19 @@ jobs:
         tag: ${{ fromJson(needs.get-test-tags.outputs.matrix) }}
     env:
       CLUSTER_NAME: ${{ needs.select-cluster.outputs.cluster_name }}
+      MATRIX_TAG: ${{ matrix.tag }}
```
antowaddle, line 531: This will silently pass if no tags are added, can you review that please? Additionally, if it's unset the for loop never executes then we get a silent failure.

.github/workflows/cypress-e2e-test.yml
```diff
@@ -1257,11 +1269,20 @@ jobs:
         run: |
           cd frontend
 
-          echo "🧪 Running E2E tests for ${{ matrix.tag }}..."
+          TAG="$MATRIX_TAG"
+          if [ ${#TAG} -gt 200 ]; then
```
antowaddle, line 1277: Can you give an example where this has succeeded please?

antowaddle (changes requested):
Thanks @paulovmr for taking the time to review this, I've added some comments. Can you review them please?

.github/workflows/cypress-e2e-test.yml
paulovmr: Will work to run the actual workflow in `odh-repository` before we merge.

For local testing this `if`: 
`TAG="@AgentOpsCI @AutoMLCI @AutoRAGCI @ConnectionTypesCI @EvalHubCI @FeatureStoreCI @GenAICI @HardwareProfilesCI @KServeCI @LLMDServingCI @MLflowCI @MaaSCI @ModelRegistryCI @ModelServingBackportCI @ModelServingCI @NIMServingCI @PipelinesCI @ProjectsCI @StorageClassesCI @WorkbenchesCI @WorkloadMetricsCI" && if [ ${#TAG} -gt 200 ];
  then echo "Length ${#TAG} > 200, hash: $(echo -n "$TAG" | sha256sum | cut -c1-12)"; else echo "Length ${#TAG}, using as-is"; fi`

.github/workflows/cypress-e2e-test.yml
paulovmr: Fixed.

.github/workflows/cypress-e2e-test.yml
paulovmr: It was a suggestion by `coderabbitai` due to it being a possible security flaw. I've removed this wording as it does not add much to a debug in the future.

.github/workflows/cypress-e2e-test.yml
paulovmr: I've added `.` and `:` to the allowed characters, do you think that is sufficient?

PR #7983, opened 2026-06-11
refactor showCustomTypeSelect into variable

emilys314:
Closes https://redhat.atlassian.net/browse/RHOAIENG-68291

## Description

Part of the effort to clean up the model location code. This should be a pure refactor

This streamlines the code that will show a connection type selector if there are multiple to select from. This also moves the logic for that away from the first dropdown field. This field should only be responsible for listing the type of model locations that are available. Which connection type to base it off is logic specific to the model location selected

example if 2 oci connection types are there
image

## How Has This Been Tested?

1. Go to the connection type admin page
2. duplicate one of them
3. deploy a new model, select the connection type to see the "Your administrator has defined multiple configuration options for OCI compliant registry locations. Select the one that best fits your needs." field
4. Verify everything still works as expected

## Test Impact

Pure refactor, no logic changes

## Request review criteria:

If you have UI changes: 

## Summary by CodeRabbit

* **Refactor**
  * Deployment wizard: simplified internal handling of connection-type selection and custom-type options.
* **Behavior Change**
  * When adding a new model location, a connection type is auto-selected only if exactly one compatible option exists; otherwise the type remains unset for explicit user choice.
* **UI**
  * Custom-type selector now shows a descriptive label in its helper text.
* **Tests**
  * E2E flows made more reliable by ensuring an admin session during project provisioning/cleanup and updating external token-auth checks.


packages/model-serving/src/components/deploymentWizard/fields/ModelLocationSelectField.tsx
```diff
@@ -33,15 +30,15 @@ import { UseModelDeploymentWizardState } from '../useDeploymentWizard';
...
-const ociOption = {
+export const ociOption = {
   key: 'OCI',
   label: 'OCI compliant registry',
 };
-const uriOption = {
+export const uriOption = {
   key: 'URI',
   label: 'URI',
 };
```
YuliaKrimerman: I think it's better to export these and then both files can call it, to prevent circular imports
// e.g. in a new file: modelLocationConstants.ts (or add to ../types.ts)
export const s3Option = { key: 'S3', label: 'S3 object storage' };
export const ociOption = { key: 'OCI', label: 'OCI compliant registry' };
export const uriOption = { key: 'URI', label: 'URI' };

packages/model-serving/src/components/deploymentWizard/fields/ModelLocationSelectField.tsx
```diff
@@ -303,66 +236,33 @@ export const ModelLocationSelectField: React.FC<ModelLocationSelectFieldProps> =
...
-                          additionalFields: {},
-                        });
-                      }
+                      setModelLocationData({
+                        type: ModelLocationType.NEW,
+                        connectionTypeObject:
+                          uriConnectionTypes.length > 1 ? undefined : uriConnectionTypes[0],
+                        fieldValues: {},
+                        additionalFields: {},
+                      });
```
YuliaKrimerman: Nit - consolidating with map 
`const optionToTypes: Record<string, { option: typeof s3Option; types: ConnectionTypeConfigMapObj[] }> = {
  [s3Option.key]: { option: s3Option, types: s3ConnectionTypes },
  [ociOption.key]: { option: ociOption, types: ociConnectionTypes },
  [uriOption.key]: { option: uriOption, types: uriConnectionTypes },
};

const match = optionToTypes[key];
if (match) {
  setUserSelectedOption({ key: match.option.key, label: match.option.label });
  setModelLocationData({
    type: ModelLocationType.NEW,
    connectionTypeObject: match.types.length > 1 ? undefined : match.types[0],
    fieldValues: {},
    additionalFields: {},
  });
}`

packages/model-serving/src/components/deploymentWizard/fields/ModelLocationInputFields.tsx
```diff
@@ -373,7 +416,7 @@ export const ModelLocationInputFields: React.FC<ModelLocationInputFieldsProps> =
                 additionalFields: {},
               });
             }}
-            typeKey={customTypeKey ?? ''}
+            typeLabel={customTypeOption.label || ''}
```
YuliaKrimerman: Should consider  narrowing  the type: extract showCustomTypeSelect into a helper that returns a discriminated union, or use a type guard to avoid potential strict null check error on customTypeOption.label

YuliaKrimerman (changes requested):

packages/model-serving/src/components/deploymentWizard/fields/modelLocationFields/modelLocationTypes.ts
```diff
@@ -0,0 +1,12 @@
+export const s3Option = {
```
mturley, line 1: Tiny nit, but this file is named `modeLocationTypes` and is missing the L in “model"

packages/model-serving/src/components/deploymentWizard/fields/ModelLocationInputFields.tsx
mturley: +1, if we don’t want to extract a helper we could just optional chain `customTypeOption?.label` here

mturley (commented):
Just a tiny thing

packages/model-serving/src/components/deploymentWizard/fields/modelLocationFields/modelLocationTypes.ts
emilys314: 😭

packages/model-serving/src/components/deploymentWizard/fields/ModelLocationInputFields.tsx
emilys314: Yeah I want to move the entire the entire CustomTypeSelectField component into the NewConnectionField after this either way. This way the New connection component will be responsible for if it needs to narrow down the type it's using, so i limited scope here. 

I think now with the changes, typescript seems to be smart enough to know that `customTypeOption` is always truthy here though since these 2 variables are now in scope of this function. I think I can just do this now if Im understanding you correctly?
```suggestion
            typeLabel={customTypeOption.label}
```

packages/model-serving/src/components/deploymentWizard/fields/ModelLocationSelectField.tsx
emilys314: Oh yeah that's a neat way to do that. I was also limiting scope here by just dedepulicating the one set of match.types.length variations

I'll change this too. I'd like this logic to be moved away from this component anyways though, since this top level dropdown shouldn't care about implementation details for the location types.

PR #8067, opened 2026-06-12
Add Global project section to cluster settings

nananosirova:
https://issues.redhat.com/browse/RHOAIENG-65180

Depends on / would need to be merged first: 
- https://github.com/opendatahub-io/odh-dashboard/pull/7866

## Description

Add a "Global project" section to the `General Settings` page that allows admins to designate a namespace as the shared workspace for curated prompts. The section uses a project selector dropdown with a "None" clear option. Saving calls the backend API which patches the `OdhDashboardConfig` CR and applies/removes namespace labels.

The section is gated behind the `globalProjectPrompts` tech preview feature flag (defaults to `false`). To enable it, set `globalProjectPrompts: true` under `spec.dashboardConfig` in the `OdhDashboardConfig` CR.

Also adds a `clearLabel` prop and `appendTo` prop to the shared `ProjectSelector` component so it can be reused with a custom deselect option and portal rendering.

## How Has This Been Tested?

- Deployed to a RHOAI cluster using PR images
- Tested selecting a project, saving, and confirming the CRD is updated
- Tested clearing the selection with "None"
- Tested the SSAR gate by confirming non-admin users cannot access the backend endpoint, will get 
`401 Unauthorized: You lack the sufficient permissions to make this request.`

https://github.com/user-attachments/assets/40a5ec00-c82d-4309-b001-1ba3130e00e6

## Test Impact

Added unit and mock tests. 

## Request review criteria:

If you have UI changes:

## Summary by CodeRabbit

* **New Features**
  * Added a **Global project** section in Cluster Settings (behind the global project prompts feature flag) to select and save an MLflow global namespace.
  * Added support for clearing the selection via **None**, persisting as an empty value.
* **Bug Fixes**
  * Improved MLflow global namespace handling with stricter validation (trim/dedupe, limits) and safer update/removal behavior with clearer warning behavior.
* **Tests**
  * Updated/added unit and Cypress coverage for the new UI section and namespace selection, including clearing and persistence.


backend/src/routes/api/cluster-settings/clusterSettingsUtils.ts
```diff
@@ -171,6 +173,30 @@ export const updateClusterSettings = async (
     if (needsNotebookControllerRollout) {
       await rolloutDeployment(fastify, namespace, 'notebook-controller-deployment');
     }
+
+    if (request.body.globalMLflowNamespaces !== undefined) {
```
DaoDaoNoCode: **suggestion**: Validation runs after other settings are already persisted

This validation block (and the `updateGlobalMLflowNamespaces` call) runs after `setDashboardConfig`, `patchCM`, and `rolloutDeployment` have already completed. If validation fails here (e.g. malformed `globalMLflowNamespaces`), those earlier settings are persisted but the response returns `{ success: false }`.

Consider moving the validation to the top of the `try` block — before any mutations — so invalid input is rejected early. The actual `updateGlobalMLflowNamespaces()` call can stay here since it needs all other settings to be saved first.

This is low-risk in practice since the frontend always sends valid data, but it would make the API more robust for direct callers.

Also note: the `createCustomError` 400 status code will be caught by the generic `catch` block (line 175) and returned as `{ success: false, error: "Unable to update cluster settings. ..." }` with HTTP 200 — which works for the frontend but changes the API contract vs the old standalone route.

frontend/src/pages/clusterSettings/__tests__/GlobalProjectSettings.spec.tsx
```diff
@@ -0,0 +1,82 @@
...
+import { fireEvent, render, screen } from '@testing-library/react';
+import '@testing-library/jest-dom';
+import { ProjectsContext } from '#~/concepts/projects/ProjectsContext';
+import GlobalProjectSettings from '#~/pages/clusterSettings/GlobalProjectSettings';
+
+const mockProjects = [
+  { metadata: { name: 'project-a', annotations: { 'openshift.io/display-name': 'Project A' } } },
+  { metadata: { name: 'project-b', annotations: { 'openshift.io/display-name': 'Project B' } } },
+  { metadata: { name: 'mlflow-ns', annotations: { 'openshift.io/display-name': 'MLflow NS' } } },
+] as never[];
```
DaoDaoNoCode: **suggestion**: Use `mockProjectK8sResource` instead of hand-crafted objects with `as never[]`

The `as never[]` cast bypasses all type checking. If the `ProjectKind` shape changes (e.g. a new required field), this test would silently pass with incomplete data while the component would break at runtime.

The Cypress test in this same PR correctly uses the proper mock:
```ts
mockProjectK8sResource({ k8sName: 'project-alpha', displayName: 'Project Alpha' })
```

Suggest refactoring to:
```tsx
import { mockProjectK8sResource } from '@odh-dashboard/internal/__mocks__/mockProjectK8sResource';

const mockProjects = [
  mockProjectK8sResource({ k8sName: 'project-a', displayName: 'Project A' }),
  mockProjectK8sResource({ k8sName: 'project-b', displayName: 'Project B' }),
  mockProjectK8sResource({ k8sName: 'mlflow-ns', displayName: 'MLflow NS' }),
];
```

packages/cypress/cypress/tests/mocked/clusterSettings/globalProjectSettings.cy.ts
```diff
@@ -0,0 +1,93 @@
...
+    clusterSettings.visit();
+    globalProjectSettings.selectNone();
+    clusterSettings.findSubmitButton().should('be.enabled');
+    clusterSettings.findSubmitButton().click();
+
+    cy.wait('@saveClusterSettings').then((interception) => {
+      expect(interception.request.body.globalMLflowNamespaces).to.eql([]);
+    });
+  });
+});
```
DaoDaoNoCode, line 99: **suggestion**: Add a negative test for feature flag OFF

All 4 tests set `globalProjectPrompts: true`. Consider adding a test that verifies the section is **not** rendered when the flag is `false` (or default):

```ts
it('should not show the global project section when feature flag is off', () => {
  cy.interceptOdh('GET /api/config', mockDashboardConfig({ globalProjectPrompts: false }));
  clusterSettings.visit();
  globalProjectSettings.findSection().should('not.exist');
});
```

This catches regressions if the feature flag gating is accidentally removed from `ClusterSettings.tsx`.

frontend/src/pages/clusterSettings/GlobalProjectSettings.tsx
```diff
@@ -0,0 +1,31 @@
...
+  selectedNamespace: string;
+  setSelectedNamespace: (ns: string) => void;
+};
+
+const GlobalProjectSettings: React.FC<GlobalProjectSettingsProps> = ({
+  selectedNamespace,
+  setSelectedNamespace,
+}) => (
+  <SettingSection
+    title="Global project"
```
DaoDaoNoCode, line 16: **nit / discussion**: Naming inconsistency across layers

The naming spans three conventions:
- **UI**: "Global project" (title), "prompts" (description)
- **Feature flag**: `globalProjectPrompts`
- **Data / state**: `globalMLflowNamespace(s)`, `mlflowGlobalNamespaceUtils.ts`

This makes sense architecturally — the UI abstracts away the MLflow implementation detail — but future developers tracing from UI text → code will hit a naming cliff. A brief comment in `ClusterSettings.tsx` near the state declaration could help:

```ts
// "Global project" UI maps to globalMLflowNamespaces in the CR (spec.globalMLflowNamespaces).
// The feature flag is globalProjectPrompts; the data is MLflow-specific.
const [globalMLflowNamespace, setGlobalMLflowNamespace] = React.useState('\);
```

Not a blocker — just something to consider for maintainability.

backend/src/routes/api/cluster-settings/clusterSettingsUtils.ts
nananosirova: Thanks, moved validation into `validateGlobalMLflowNamespaces` which now runs at the top of `updateClusterSettings`, before any mutations. If validation fails, nothing is persisted.

packages/cypress/cypress/tests/mocked/clusterSettings/globalProjectSettings.cy.ts
nananosirova: Added, thanks!

backend/src/routes/api/cluster-settings/clusterSettingsUtils.ts
```diff
@@ -43,6 +47,15 @@ export const updateClusterSettings = async (
...
   try {
+    let validatedMLflow: { uniqueNamespaces: string[]; oldNamespaces: string[] } | undefined;
+    if (request.body.globalMLflowNamespaces !== undefined) {
+      validatedMLflow = await validateGlobalMLflowNamespaces(
+        fastify,
+        request as OauthFastifyRequest,
+        request.body.globalMLflowNamespaces,
+      );
+    }
+
```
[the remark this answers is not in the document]
nananosirova: `setDashboardConfig` uses merge-patch, so concurrent writes to different fields don't conflict. For same-field races, we've already minimized the window by threading the validated snapshot through. Adding `resourceVersion` would require changes to the shared `setDashboardConfig` utility used by all cluster settings, out of scope for this PR.

DaoDaoNoCode (approved):
/lgtm

PR #8090, opened 2026-06-15
RHOAIENG-60050: fix BFF_MAAS_SERVICE_NAME in RHOAI kustomize overlay

gmenher:
https://issues.redhat.com/browse/RHOAIENG-60050

## Description

Fixes inter-BFF service discovery failure in RHOAI deployments. The gen-ai BFF uses `BFF_MAAS_SERVICE_NAME` to construct K8s DNS URLs for calling the MaaS BFF (`<service-name>.<namespace>.svc.cluster.local`). In RHOAI the service is `rhods-dashboard`, but the env var was hardcoded to `odh-dashboard`, causing DNS resolution errors for token issuance/revocation calls.

**What changed (1 file, +4/-0):**

* **`manifests/rhoai/base/deployment.yaml`** — Added a JSON6902 `replace` patch targeting container[3] (`gen-ai-ui`) env[2] (`BFF_MAAS_SERVICE_NAME`), changing the value from `"odh-dashboard"` to `"rhods-dashboard"`.

**Root cause:** PR #6404 introduced inter-BFF communication env vars but did not add a corresponding RHOAI overlay patch.

## How Has This Been Tested?

**Local verification (kustomize build):**
- `kustomize build manifests/rhoai/base` produces `BFF_MAAS_SERVICE_NAME: "rhods-dashboard"` for the gen-ai-ui container
- `kustomize build manifests/odh` still produces `BFF_MAAS_SERVICE_NAME: "odh-dashboard"` (unaffected)
- Container index (3) and env index (2) confirmed stable via kustomize output

**Regression checks:**
- No other containers have `BFF_MAAS_SERVICE_NAME` in the sidecar deployment
- Standalone module (`modules/gen-ai/deployment.yaml`) uses a different value (`odh-dashboard-maas-ui`) and is unaffected
- ODH overlay is completely unaffected by this RHOAI-only patch

**Cluster verification (pre-existing):**
- Ticket reporter verified on cluster that calling MaaS via `rhods-dashboard` succeeds (confirming the issue is purely the env var value)
- Direct `oc patch` on the live deployment gets reverted by the operator reconciler, so full end-to-end validation requires deploying via the operator with this patch included

**Scope audit (AC #4):**
- Grep'd all containers in the final RHOAI kustomize output for remaining `odh-dashboard` references in env vars/args: zero found after this patch
- Remaining `odh-dashboard` references are RBAC resources (ServiceAccount, ClusterRole) and image placeholders — not DNS service names

## Request review criteria:

- [x] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious)
- [x] The code follows our [Best Practices](/docs/best-practices.md)

## Release Notes

* **Bug Fixes**
  * Updated the dashboard deployment configuration to consistently use the dashboard name across resource naming, labels, selectors, pod anti-affinity rules, and the service account.
  * Updated the backend environment setting for DNS-based MaaS discovery to match the dashboard service name, improving inter-service communication reliability.

**Note on approach:** This patch uses positional JSON6902 (`containers/3/env/2`) guarded by two `op: test` assertions (container name + env var name) that fail-fast if indices ever shift. The ideal long-term solution would be kustomize `replacements` with name-based field paths (`containers.[name=gen-ai-ui].env.[name=BFF_MAAS_SERVICE_NAME].value`) or a Strategic Merge Patch, both of which are index-independent. We chose JSON6902 here because the entire existing RHOAI overlay uses this pattern exclusively — introducing a different strategy for a single env var fix would be inconsistent and disproportionate in scope. The `test` guards provide equivalent safety: if ordering changes, `kustomize build` fails immediately rather than silently misconfiguring the deployment.

### DIFF
--- a/manifests/rhoai/base/deployment.yaml
+++ b/manifests/rhoai/base/deployment.yaml
@@ -18,7 +18,6 @@
   path: /spec/template/spec/serviceAccount
   value: rhods-dashboard
 # Fix inter-BFF service discovery: gen-ai BFF calls MaaS via K8s DNS.
-# In RHOAI the service is "rhods-dashboard", not "odh-dashboard" (RHOAIENG-60050).
 # Guards: fail-fast if container or env ordering ever changes.
 - op: test
   path: /spec/template/spec/containers/3/name
--- a/manifests/rhoai/base/deployment.yaml
+++ b/manifests/rhoai/base/deployment.yaml
@@ -17,3 +17,7 @@
 - op: replace
   path: /spec/template/spec/serviceAccount
   value: rhods-dashboard
+# Fix inter-BFF service discovery: gen-ai BFF calls MaaS via K8s DNS..
+- op: replace
+  path: /spec/template/spec/containers/3/env/2/value
+  value: "rhods-dashboard"
--- a/manifests/rhoai/base/deployment.yaml
+++ b/manifests/rhoai/base/deployment.yaml
@@ -19,7 +19,10 @@
   value: rhods-dashboard
 # Fix inter-BFF service discovery: gen-ai BFF calls MaaS via K8s DNS.
 # In RHOAI the service is "rhods-dashboard", not "odh-dashboard" (RHOAIENG-60050).
-# Guard: fail-fast if container/env ordering ever changes.
+# Guards: fail-fast if container or env ordering ever changes.
+- op: test
+  path: /spec/template/spec/containers/3/name
+  value: "gen-ai-ui"
 - op: test
   path: /spec/template/spec/containers/3/env/2/name
   value: "BFF_MAAS_SERVICE_NAME"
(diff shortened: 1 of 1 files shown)


lucferbux (changes requested):
Looks fine, however patch is over-engineered for what it does. The entire existing RHOAI overlay (`manifests/rhoai/base/deployment.yaml`, lines 1–19) uses bare positional JSON6902 `replace` operations — **none** of the 7 existing patches use `op: test` guards. Adding test guards for just this one patch is inconsistent with the established pattern and contradicts the PR body's own rationale of choosing JSON6902 "because the entire existing RHOAI overlay uses this pattern exclusively."

The simplest correct patch (consistent with the rest of the file) is:

```yaml
- op: replace
  path: /spec/template/spec/containers/3/env/2/value
  value: "rhods-dashboard"
```

3 lines instead of 11. No comments needed — the path is self-documenting for anyone reading this overlay alongside the base deployment.

lucferbux (approved):
/lgtm
/approve

PR #8096, opened 2026-06-15
RHOAIENG-63158: Role creation submission via K8s API pass-through

srtanish1992:
https://redhat.atlassian.net/browse/RHOAIENG-63158

## Description

Implements the Create Role submit flow for the project roles page.

**No rules:**
https://github.com/user-attachments/assets/02d8de3e-9a1e-4c8f-9dd1-83b122ed915d

**With rules:**
https://github.com/user-attachments/assets/40e7383f-b5a7-445b-b967-38239ffdddd4

**What this does:**
- Adds a submit handler to the Create Role page that assembles a `RoleKind` payload from form state (name, display name, description, permission rules) and calls the K8s API to create the role.
- When the permission rules array is empty, clicking "Create role" shows a **"Create empty role?"** confirmation warning modal instead of submitting immediately. The user can confirm to proceed or cancel.
- On successful creation, the user is navigated back to the project roles list. On API error, an inline error alert is displayed in the page footer (or within the confirmation modal if triggered from there).
- Makes `RoleKind.rules` and `ClusterRoleKind.rules` optional in `k8sTypes.ts` to match actual K8s API responses (the API omits `rules` entirely when empty, which previously caused a runtime crash in the role details view).

## How Has This Been Tested?

- Manually tested creating a role with rules (direct submit, no modal)
- Manually tested creating a role with no rules (confirmation modal appears, confirm creates the role)
- Manually tested cancelling the confirmation modal
- Manually tested API error scenarios (409 conflict) — error alert shown in footer and modal
- Verified the role details view no longer crashes for roles with no rules
- Ran `npx eslint` on all changed files — zero errors
- Ran `npx tsc --noEmit` — zero type errors

## Test Impact

**Unit tests added:**
- `assembleRole.spec.ts` — verifies payload structure, label/annotation handling, rule ID stripping, optional fields
- `CreateRoleFooter.spec.tsx` — verifies button text, disabled states, onSubmit calls, error alert display, loading state
- `CreateRoleConfirmModal.spec.tsx` — verifies rendering, button actions, error display, disabled states during submission

**Cypress mock tests added:**
- `rolesTabCreate.cy.ts` — covers:
  - Empty-rules confirmation modal flow (modal appears, confirm submits, cancel dismisses)
  - Direct submit with rules (no modal, correct payload, navigation)
  - API error on direct submit (error alert shown, button re-enabled)
  - Display name and description annotations in request payload

## Request review criteria:

If you have UI changes: 

## Summary by CodeRabbit

- **New Features**
  - Added a confirmation modal for creating empty roles (no rules) and an async role creation flow with automatic redirect on success.
- **Bug Fixes**
  - Improved role rules rendering by safely handling missing/undefined rules.
- **Enhancements**
  - Enhanced role creation UX with inline error alerts and disabled actions during submission; improved manifest generation with optional labels/annotations.
- **Tests / QA**
  - Added unit tests for modal/footer behavior, rule/payload assembly, and Cypress coverage for empty-role and submit error scenarios; added accessibility check and new UI selectors.

### DIFF
--- a/frontend/src/pages/projects/projectRoles/CreateRoleConfirmModal.tsx
+++ b/frontend/src/pages/projects/projectRoles/CreateRoleConfirmModal.tsx
@@ -17,6 +17,7 @@ const CreateRoleConfirmModal: React.FC<CreateRoleConfirmModalProps> = ({ onConfi
       await onConfirm();
     } catch (e) {
       setError(e instanceof Error ? e : new Error('Failed to create role'));
+    } finally {
       setIsSubmitting(false);
     }
   }, [onConfirm]);
--- a/frontend/src/pages/projects/projectRoles/CreateRoleConfirmModal.tsx
+++ b/frontend/src/pages/projects/projectRoles/CreateRoleConfirmModal.tsx
@@ -1,0 +1,57 @@
+import * as React from 'react';
+import ContentModal from '#~/components/modals/ContentModal';
+
+type CreateRoleConfirmModalProps = {
+  onConfirm: () => Promise<void>;
+  onClose: () => void;
+};
+
+const CreateRoleConfirmModal: React.FC<CreateRoleConfirmModalProps> = ({ onConfirm, onClose }) => {
+  const [isSubmitting, setIsSubmitting] = React.useState(false);
+  const [error, setError] = React.useState<Error>();
+
+  const handleConfirm = React.useCallback(async () => {
+    setIsSubmitting(true);
+    setError(undefined);
+    try {
+      await onConfirm();
+    } catch (e) {
+      setError(e instanceof Error ? e : new Error('Failed to create role'));
+      setIsSubmitting(false);
+    }
+  }, [onConfirm]);
+
+  return (
+    <ContentModal
+      title="Create empty role?"
+      titleIconVariant="warning"
+      variant="small"
+      dataTestId="create-role-confirm-modal"
+      onClose={onClose}
+      error={error}
+      alertTitle="Error creating role"
+      buttonActions={[
+        {
+          label: 'Create role',
+          onClick: handleConfirm,
+          variant: 'primary',
+          isLoading: isSubmitting,
+          isDisabled: isSubmitting,
+          dataTestId: 'confirm-create-button',
+        },
+        {
+          label: 'Cancel',
+          onClick: onClose,
+          variant: 'link',
+          isDisabled: isSubmitting,
+          dataTestId: 'confirm-cancel-button',
+        },
+      ]}
+      contents={
+        <>This role has no rules and won&apos;t grant any permissions. You can add rules later.</>
+      }
+    />
+  );
+};
+
+export default CreateRoleConfirmModal;
--- a/frontend/src/pages/projects/projectRoles/CreateRoleFooter.tsx
+++ b/frontend/src/pages/projects/projectRoles/CreateRoleFooter.tsx
@@ -31,6 +31,8 @@ const CreateRoleFooter: React.FC<CreateRoleFooterProps> = ({
     setIsSubmitting(true);
     try {
       await onSubmit();
+    } catch {
+      // error is surfaced via submitError prop
     } finally {
       setIsSubmitting(false);
     }
--- a/frontend/src/pages/projects/projectRoles/CreateRoleFooter.tsx
+++ b/frontend/src/pages/projects/projectRoles/CreateRoleFooter.tsx
@@ -1,37 +1,81 @@
 import * as React from 'react';
 import { useNavigate } from 'react-router-dom';
-import { ActionList, ActionListItem, Button } from '@patternfly/react-core';
+import {
+  ActionList,
+  ActionListItem,
+  Alert,
+  Button,
+  Stack,
+  StackItem,
+} from '@patternfly/react-core';
 
 type CreateRoleFooterProps = {
   namespace: string;
   isSubmitDisabled: boolean;
   isEdit?: boolean;
+  onSubmit: () => Promise<void>;
+  submitError?: Error;
 };
 
 const CreateRoleFooter: React.FC<CreateRoleFooterProps> = ({
   namespace,
   isSubmitDisabled,
   isEdit = false,
+  onSubmit,
+  submitError,
 }) => {
   const navigate = useNavigate();
+  const [isSubmitting, setIsSubmitting] = React.useState(false);
+
+  const handleSubmit = React.useCallback(async () => {
+    setIsSubmitting(true);
+    try {
+      await onSubmit();
+    } finally {
+      setIsSubmitting(false);
+    }
+  }, [onSubmit]);
 
   return (
-    <ActionList>
-      <ActionListItem>
-        <Button isDisabled={isSubmitDisabled} variant="primary" data-testid="create-role-submit">
-          {isEdit ? 'Save changes' : 'Create role'}
-        </Button>
-      </ActionListItem>
-      <ActionListItem>
-        <Button
-          variant="link"
-          data-testid="create-role-cancel"
-          onClick={() => navigate(`/projects/${namespace}?section=roles`)}
-        >
-          Cancel
-        </Button>
-      </ActionListItem>
-    </ActionList>
+    <Stack hasGutter>
+      {submitError && (
+        <StackItem>
+          <Alert
+            isInline
+            variant="danger"
+            title={isEdit ? 'Error updating role' : 'Error creating role'}
+            data-testid="create-role-error-alert"
+          >
+            {submitError.message}
+          </Alert>
+        </StackItem>
+      )}
+      <StackItem>
+        <ActionList>
+          <ActionListItem>
+            <Button
+              isDisabled={isSubmitDisabled || isSubmitting}
+              isLoading={isSubmitting}
+              variant="primary"
+              data-testid="create-role-submit"
+              onClick={handleSubmit}
+            >
+              {isEdit ? 'Save changes' : 'Create role'}
+            </Button>
+          </ActionListItem>
+          <ActionListItem>
+            <Button
+              variant="link"
+              data-testid="create-role-cancel"
+              isDisabled={isSubmitting}
+              onClick={() => navigate(`/projects/${namespace}?section=roles`)}
+            >
+              Cancel
+            </Button>
+          </ActionListItem>
+        </ActionList>
+      </StackItem>
+    </Stack>
   );
 };
 
--- a/frontend/src/pages/projects/projectRoles/CreateRolePage.tsx
+++ b/frontend/src/pages/projects/projectRoles/CreateRolePage.tsx
@@ -8,15 +8,18 @@ import {
   Spinner,
   getUniqueId,
 } from '@patternfly/react-core';
-import { Link, Navigate, useParams } from 'react-router-dom';
+import { Link, Navigate, useNavigate, useParams } from 'react-router-dom';
 import ApplicationsPage from '#~/pages/ApplicationsPage';
 import { useAccessReview } from '#~/api/useAccessReview';
 import { ProjectDetailsContext } from '#~/pages/projects/ProjectDetailsContext';
 import { getDisplayNameFromK8sResource, isValidK8sLabelKeyValue } from '#~/concepts/k8s/utils';
 import { useK8sNameDescriptionFieldData } from '#~/concepts/k8s/K8sNameDescriptionField/K8sNameDescriptionField';
 import { RoleKind } from '#~/k8sTypes';
+import { createRole } from '#~/api';
 import CreateRoleForm from './CreateRoleForm';
 import CreateRoleFooter from './CreateRoleFooter';
+import CreateRoleConfirmModal from './CreateRoleConfirmModal';
+import assembleRole from './assembleRole';
 import type { LabelEntry, RuleEntry } from './types';
 
 type CreateRolePageProps = {
@@ -28,6 +31,7 @@ const CreateRolePage: React.FC<CreateRolePageProps> = ({ existingRole }) => {
   const { currentProject } = React.useContext(ProjectDetailsContext);
   const displayName = getDisplayNameFromK8sResource(currentProject);
   const isEdit = !!existingRole;
+  const navigate = useNavigate();
 
   const [allowAccess, loaded] = useAccessReview({
     group: 'rbac.authorization.k8s.io',
@@ -62,6 +66,9 @@ const CreateRolePage: React.FC<CreateRolePageProps> = ({ existingRole }) => {
     }));
   });
 
+  const [submitError, setSubmitError] = React.useState<Error>();
+  const [showNoRulesConfirm, setShowNoRulesConfirm] = React.useState(false);
+
   const handleDescriptionChange = React.useCallback((value: string) => {
     setDescription(value);
   }, []);
@@ -90,6 +97,29 @@ const CreateRolePage: React.FC<CreateRolePageProps> = ({ existingRole }) => {
     k8sNameDescriptionData.data.k8sName.state.invalidLength ||
     hasInvalidLabels;
 
+  const doSubmit = React.useCallback(async () => {
+    setSubmitError(undefined);
+    const k8sName = k8sNameDescriptionData.data.k8sName.value;
+    const roleDisplayName = k8sNameDescriptionData.data.name || k8sName;
+    const role = assembleRole(namespace, k8sName, roleDisplayName, description, rules);
+    try {
+      await createRole(role);
+      navigate(`/projects/${namespace}?section=roles`);
+    } catch (e) {
+      const error = e instanceof Error ? e : new Error('Failed to create role');
+      setSubmitError(error);
+      throw error;
+    }
+  }, [namespace, k8sNameDescriptionData.data, description, rules, navigate]);
+
+  const handleSubmit = React.useCallback(async () => {
+    if (rules.length === 0) {
+      setShowNoRulesConfirm(true);
+      return;
+    }
+    await doSubmit();
+  }, [rules.length, doSubmit]);
+
   if (!loaded) {
     return (
       <Bullseye>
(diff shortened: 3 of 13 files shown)


srtanish1992:

frontend/src/pages/projects/projectRoles/CreateRolePage.tsx
```diff
@@ -90,6 +97,29 @@ const CreateRolePage: React.FC<CreateRolePageProps> = ({ existingRole }) => {
...
+    const role = assembleRole(namespace, k8sName, roleDisplayName, description, rules);
+    try {
+      await createRole(role);
+      navigate(`/projects/${namespace}?section=roles`);
+    } catch (e) {
+      const error = e instanceof Error ? e : new Error('Failed to create role');
+      setSubmitError(error);
+      throw error;
+    }
+  }, [namespace, k8sNameDescriptionData.data, description, rules, navigate]);
```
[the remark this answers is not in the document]
srtanish1992: NOT APPLICABLE for this PR. Edit mode is explicitly out of scope per RHOAIENG-63158 -- it's tracked under RHOAIENG-66139

frontend/src/pages/projects/projectRoles/CreateRoleFooter.tsx
```diff
@@ -1,37 +1,81 @@
...
+  submitError,
 }) => {
   const navigate = useNavigate();
+  const [isSubmitting, setIsSubmitting] = React.useState(false);
+
+  const handleSubmit = React.useCallback(async () => {
+    setIsSubmitting(true);
+    try {
+      await onSubmit();
+    } finally {
```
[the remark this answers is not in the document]
srtanish1992: Fixed.

packages/cypress/cypress/tests/mocked/projects/tabs/rolesTabCreate.cy.ts
```diff
@@ -0,0 +1,174 @@
...
+    projectRoles.findConfirmCreateButton().click();
+
+    cy.wait('@createRole').then((interception) => {
+      expect(interception.request.body.metadata.annotations).to.have.property(
+        'openshift.io/description',
+        'A test description',
+      );
+      expect(interception.request.body.metadata.annotations).to.have.property(
+        'openshift.io/display-name',
+      );
```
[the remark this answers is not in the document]
srtanish1992: Fixed.

frontend/src/pages/projects/projectRoles/CreateRoleConfirmModal.tsx
```diff
@@ -0,0 +1,57 @@
...
+  const handleConfirm = React.useCallback(async () => {
+    setIsSubmitting(true);
+    setError(undefined);
+    try {
+      await onConfirm();
+    } catch (e) {
+      setError(e instanceof Error ? e : new Error('Failed to create role'));
+      setIsSubmitting(false);
+    }
+  }, [onConfirm]);
```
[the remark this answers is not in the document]
srtanish1992: Fixed.

DaoDaoNoCode (commented):
Reviewing the three new unit test files only (assembleRole.spec.ts, CreateRoleConfirmModal.spec.tsx, CreateRoleFooter.spec.tsx). All three suites pass. Good overall structure -- proper use of `data-testid` selectors, `waitFor` for async, and the deferred-promise pattern for testing loading states.

Local checks: all 3 test suites pass (20/20 tests). One `act(...)` warning from CreateRoleConfirmModal.spec.tsx -- see inline comment.

Prior review: odh-dashboard-agent raised the missing `catch` block in CreateRoleFooter -- fix verified in code (line 34 now has an empty `catch`). coderabbit raised the `isSubmitting` reset in CreateRoleConfirmModal -- fix verified (now uses `finally`). coderabbit's Cypress annotation value comment was addressed in commit 1f583df.

Left a few comments inline -- main things are a dangling promise causing a React `act()` warning, a missing edge-case test for `assembleRole`'s conditional field omission, and a missing rejection-path test in the footer.

frontend/src/pages/projects/projectRoles/__tests__/assembleRole.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+        apiGroups: ['serving.kserve.io'],
+        resources: ['inferenceservices'],
+        resourceNames: ['my-model'],
+      },
+    ];
+
+    const result = assembleRole('ns', 'r', 'R', '', rules);
+
+    expect(result.rules[0].resourceNames).toStrictEqual(['my-model']);
+  });
```
DaoDaoNoCode, line 117: This test is a subset of what the "strip id field" test already covers (rule-2 at line 57 has `resourceNames`). It would be more valuable to replace it with a test for rules where the optional fields (`apiGroups`, `resources`, `resourceNames`) are `undefined` -- the implementation uses conditional spreads (`rule.apiGroups && { apiGroups: ... }`) which means those keys should be omitted entirely from the output. That's the interesting behavior to assert.

```suggestion
  it('should omit optional fields when undefined', () => {
    const rules: RuleEntry[] = [
      {
        id: 'rule-1',
        verbs: ['get'],
        // apiGroups, resources, resourceNames all undefined
      },
    ];

    const result = assembleRole('ns', 'r', 'R', '', rules);

    expect(result.rules[0]).toStrictEqual({ verbs: ['get'] });
    expect(result.rules[0]).not.toHaveProperty('apiGroups');
    expect(result.rules[0]).not.toHaveProperty('resources');
    expect(result.rules[0]).not.toHaveProperty('resourceNames');
  });
```

frontend/src/pages/projects/projectRoles/__tests__/CreateRoleConfirmModal.spec.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+    fireEvent.click(screen.getByTestId('confirm-create-button'));
+
+    await waitFor(() => {
+      expect(screen.getByTestId('confirm-create-button')).toBeDisabled();
+      expect(screen.getByTestId('confirm-cancel-button')).toBeDisabled();
+    });
+
+    resolvePromise();
+  });
```
DaoDaoNoCode, line 79: This test has a dangling promise -- it calls `resolvePromise()` but never waits for the component to process the resolution. That's causing the `act(...)` warning in the test output (`setIsSubmitting(false)` fires after the test ends). You should `await waitFor` after resolving to verify the buttons re-enable, which also clears the warning.

```suggestion
    resolvePromise();

    await waitFor(() => {
      expect(screen.getByTestId('confirm-create-button')).not.toBeDisabled();
      expect(screen.getByTestId('confirm-cancel-button')).not.toBeDisabled();
    });
  });
```

The footer test already does this correctly (line 92-94), so this is probably just an oversight.

frontend/src/pages/projects/projectRoles/__tests__/CreateRoleConfirmModal.spec.tsx
```diff
@@ -0,0 +1,63 @@
...
+  it('should show error alert when onConfirm rejects', async () => {
+    const onConfirm = jest.fn().mockRejectedValue(new Error('API failure'));
+    render(<CreateRoleConfirmModal onConfirm={onConfirm} onClose={jest.fn()} />);
+
+    fireEvent.click(screen.getByTestId('confirm-create-button'));
+
+    await waitFor(() => {
+      expect(screen.getByText('API failure')).toBeInTheDocument();
+    });
+  });
```
DaoDaoNoCode, line 43: nit: this test only asserts that the error message text appears, but doesn't verify it shows up inside the error alert specifically. If the error text accidentally leaked into the body instead of the alert, this would still pass. Consider asserting on the alert's `data-testid` too:

```suggestion
    await waitFor(() => {
      expect(screen.getByTestId('error-message-alert')).toBeInTheDocument();
      expect(screen.getByText('API failure')).toBeInTheDocument();
    });
```

DaoDaoNoCode, line 43: The implementation has a fallback path at line 19: when the rejection is not an `Error` instance, it wraps it with `new Error('Failed to create role')`. That path isn't tested. Worth adding a quick test:

```suggestion
    await waitFor(() => {
      expect(screen.getByText('API failure')).toBeInTheDocument();
    });
  });

  it('should show fallback error when onConfirm rejects with non-Error', async () => {
    const onConfirm = jest.fn().mockRejectedValue('string error');
    render(<CreateRoleConfirmModal onConfirm={onConfirm} onClose={jest.fn()} />);

    fireEvent.click(screen.getByTestId('confirm-create-button'));

    await waitFor(() => {
      expect(screen.getByText('Failed to create role')).toBeInTheDocument();
    });
```

frontend/src/pages/projects/projectRoles/__tests__/CreateRoleFooter.spec.tsx
```diff
@@ -0,0 +1,96 @@
...
+      expect(screen.getByTestId('create-role-cancel')).toBeDisabled();
+    });
+
+    resolvePromise();
+
+    await waitFor(() => {
+      expect(screen.getByTestId('create-role-submit')).not.toBeDisabled();
+    });
+  });
+});
```
DaoDaoNoCode, line 120: nit: there's no test for when `onSubmit` rejects. The footer's `handleSubmit` swallows the error (empty `catch` block), so the important behavior to verify is that `isSubmitting` resets back to `false` after a rejection and no unhandled promise rejection leaks. Something like:

```
  it('should re-enable submit button after onSubmit rejects', async () => {
    const onSubmit = jest.fn().mockRejectedValue(new Error('API error'));
    renderFooter({ onSubmit });

    fireEvent.click(screen.getByTestId('create-role-submit'));

    await waitFor(() => {
      expect(screen.getByTestId('create-role-submit')).not.toBeDisabled();
    });
  });
```

frontend/src/pages/projects/projectRoles/assembleRole.ts
```diff
@@ -0,0 +1,34 @@
...
+  description: string,
+  rules: RuleEntry[],
+): AssembledRole => ({
+  apiVersion: 'rbac.authorization.k8s.io/v1',
+  kind: 'Role',
+  metadata: {
+    name: k8sName,
+    namespace,
+    labels: {
+      [KnownLabels.DASHBOARD_RESOURCE]: 'true',
```
DaoDaoNoCode, line 21: The form lets users add custom labels (managed via `labels` state and `RoleLabelsSection`), validates them, and even blocks submit if they're invalid -- but `assembleRole` never receives them. Any labels the user adds in the form are silently dropped from the API payload. Only the hardcoded `DASHBOARD_RESOURCE` label survives.

You'll want to accept the labels and merge them in:

```suggestion
const assembleRole = (
  namespace: string,
  k8sName: string,
  displayName: string,
  description: string,
  rules: RuleEntry[],
  labels: Record<string, string> = {},
): AssembledRole => ({
  apiVersion: 'rbac.authorization.k8s.io/v1',
  kind: 'Role',
  metadata: {
    name: k8sName,
    namespace,
    labels: {
      ...labels,
      [KnownLabels.DASHBOARD_RESOURCE]: 'true',
    },
```

Then in `CreateRolePage.tsx`, convert the `LabelEntry[]` to a `Record<string, string>` and pass it through:

```ts
const labelRecord = Object.fromEntries(labels.map((l) => [l.key, l.value]));
const role = assembleRole(namespace, k8sName, roleDisplayName, description, rules, labelRecord);
```

Don't forget to add `labels` to the `doSubmit` dependency array too.

DaoDaoNoCode (changes requested):
Implements the create-role submit flow with confirmation modal for empty rules, inline error display, and proper K8s RBAC role assembly. Prior review threads are all resolved: CodeRabbit's `finally` fix (verified in code), the missing `catch` in CreateRoleFooter (verified in code), and the edit-mode `createRole` gap (intentionally out of scope per RHOAIENG-66139).

Local checks: lint (passed on CI), type-check (passed on CI), unit tests (passed on CI).

CI: Cypress `projects/tabs` shard is failing -- two tests ("should submit directly when rules are present" and "should show error alert when API returns an error on direct submit") time out clicking the add-rule modal's submit button because it stays disabled. This is a real test failure, not an infra flake -- the add-rule modal form is not reaching a submittable state during those tests.

Leaving one comment inline -- the main issue is that user-added labels from the form are silently dropped when the role is created.

packages/cypress/cypress/tests/mocked/projects/tabs/rolesTabCreate.cy.ts
```diff
@@ -0,0 +1,175 @@
...
+      'POST',
+      { model: RoleModel, ns: NAMESPACE },
+      mockRoleK8sResource({ name: 'my-role', namespace: NAMESPACE }),
+    ).as('createRole');
+
+    projectRoles.visitCreateRole(NAMESPACE);
+    projectRoles.findRoleNameInput().type('my-role');
+
+    projectRoles.findAddRuleButton().click();
+    cy.findByTestId('add-rule-modal').should('exist');
```
DaoDaoNoCode: Lines 99-105 (and the identical block at 135-141) use direct `cy.findByTestId()` calls to interact with the add-rule modal. Per the project's Cypress mock test conventions, tests must never call `cy.findByTestId()` directly -- all UI interactions should go through page object methods.

This also means the exact same 7-line sequence is duplicated verbatim across two tests. Consider adding page object methods to `ProjectRolesTab` (or a new `AddRuleModal` class) for the add-rule modal elements, then extracting a helper that both tests call:

```typescript
// In projectRoles.ts
findAddRuleModal() {
  return cy.findByTestId('add-rule-modal');
}
findRuleApiGroupsToggle() {
  return cy.findByTestId('rule-api-groups-toggle');
}
// ... etc

// In the test file
const addRuleWithVerb = (apiGroup: string, resource: string, verb: string) => {
  projectRoles.findAddRuleButton().click();
  projectRoles.findAddRuleModal().should('exist');
  projectRoles.findRuleApiGroupsToggle().click();
  // ... fill fields via page object methods
};
```

This also naturally sets up the fix for the unresolved CodeRabbit thread about the brittle verb checkbox selector -- the page object would encapsulate that selector in one place.

packages/cypress/cypress/tests/mocked/projects/tabs/rolesTabCreate.cy.ts
```diff
@@ -0,0 +1,175 @@
...
+    cy.findByTestId('rule-resource-types-toggle').parent().find('input').type('deployments{enter}');
+    cy.findByTestId('add-rule-modal').find('[role="treeitem"]#get input[type="checkbox"]').click();
+    cy.findByTestId('modal-submit-button').click();
+
+    projectRoles.findSubmitButton().click();
+
+    cy.wait('@createRole');
+    projectRoles.findSubmitErrorAlert().should('exist');
+    projectRoles.findSubmitButton().should('be.enabled');
+  });
```
DaoDaoNoCode, line 147: The tests cover the direct-submit error path (API returns 409 when rules are present), but there's no test for the confirmation-modal error path (API returns 409 when submitting through the modal with no rules). The `CreateRoleConfirmModal` has its own error display and the PR description mentions "error alert shown in footer and modal" as a tested scenario -- but only the footer error path is covered in Cypress.

Consider adding:
```typescript
it('should show error in confirmation modal when API fails', () => {
  cy.interceptK8s(
    'POST',
    { model: RoleModel, ns: NAMESPACE },
    { statusCode: 409, body: mock409Error({ message: 'Role already exists' }) },
  ).as('createRole');

  projectRoles.visitCreateRole(NAMESPACE);
  projectRoles.findRoleNameInput().type('empty-role');
  projectRoles.findSubmitButton().click();
  projectRoles.findConfirmCreateButton().click();

  cy.wait('@createRole');
  projectRoles.findConfirmModal().should('exist');
  projectRoles.findConfirmModal().contains('Error creating role').should('exist');
  projectRoles.findConfirmCreateButton().should('be.enabled');
});
```

packages/cypress/cypress/tests/mocked/projects/tabs/rolesTabCreate.cy.ts
```diff
@@ -0,0 +1,175 @@
...
+};
+
+describe('Create Role submit', () => {
+  beforeEach(() => {
+    asProjectAdminUser();
+    initIntercepts();
+  });
+
+  it('should show confirmation modal when submitting with no rules', () => {
+    projectRoles.visitCreateRole(NAMESPACE);
```
DaoDaoNoCode, line 55: nit: `visitCreateRole` in the page object doesn't call `cy.testA11y()`, unlike the other visit methods (`visit`, `visitOverview`). Per conventions, all page visits should include accessibility testing. This isn't strictly in your diff (the page object method is pre-existing), but since you're the first consumer of `visitCreateRole`, it would be worth adding `cy.testA11y()` to the page object's `visitCreateRole` method while you're here -- something like:

```typescript
visitCreateRole(namespace: string) {
  cy.visitWithLogin(`/projects/${namespace}/roles/create`);
  cy.findByTestId('create-role-page');
  cy.testA11y();
}
```

DaoDaoNoCode (changes requested):
Reviewing only the Cypress mock test file and page object additions (other files covered by separate reviewers).

The test covers the core submit flows well: empty-rules confirmation modal, direct submit with rules, 409 error handling, and annotation payload verification. Page object additions for the modal and error alert use proper data-testid selectors and follow the project's naming conventions. Mock data comes from `@odh-dashboard/internal/__mocks__` as required.

Local checks: lint on scoped files passes. CI `projects/tabs` mock tests are failing on both commits -- the run is still in-progress on the latest commit so I can't see logs yet, but the same shard also failed on the first commit. Worth checking whether this is a pre-existing flake or a real failure in this PR's tests.

Prior review: CodeRabbit raised brittle DOM selectors on lines 104/140 -- still unresolved (not re-raised here). CodeRabbit raised missing `openshift.io/display-name` value assertion on line 172 -- fix verified in code. CodeRabbit raised `isSubmitting` not resetting in `CreateRoleConfirmModal` -- fix verified (finally block added). odh-dashboard-agent raised missing catch block in `CreateRoleFooter` -- fix verified in code.

Leaving a few comments inline -- main thing to address is the direct `cy.findByTestId()` calls in the test file that bypass the page object model.

packages/cypress/cypress/tests/mocked/projects/tabs/rolesTabCreate.cy.ts
```diff
@@ -0,0 +1,175 @@
...
+    projectRoles.findRoleNameInput().type('my-role');
+
+    projectRoles.findAddRuleButton().click();
+    cy.findByTestId('add-rule-modal').should('exist');
+    cy.findByTestId('rule-api-groups-toggle').click();
+    cy.findByTestId('rule-api-groups-toggle').parent().find('input').type('apps{enter}');
+    cy.findByTestId('rule-resource-types-toggle').click();
+    cy.findByTestId('rule-resource-types-toggle').parent().find('input').type('deployments{enter}');
+    cy.findByTestId('add-rule-modal').find('[role="treeitem"]#get input[type="checkbox"]').click();
```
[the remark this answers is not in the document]
srtanish1992: Fixed.

packages/cypress/cypress/tests/mocked/projects/tabs/rolesTabCreate.cy.ts
```diff
@@ -0,0 +1,179 @@
...
+    projectRoles.findRoleNameInput().type('my-role');
+
+    projectRoles.findAddRuleButton().click();
+    cy.findByTestId('add-rule-modal').should('exist');
+    cy.findByTestId('rule-api-groups-toggle').click();
+    cy.findByTestId('rule-api-groups-toggle').parent().find('input').type('apps');
+    cy.contains('Use custom API group "apps"').click();
+    cy.findByTestId('rule-resource-types-toggle').click();
+    cy.findByTestId('rule-resource-types-toggle').parent().find('input').type('deployments');
+    cy.contains('Use custom resource type "deployments"').click();
```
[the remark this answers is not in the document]
srtanish1992: The cy.contains() calls here are global intentionally. PF6's Select component renders its dropdown menu via a Popper portal at document.body level — the menu options are not DOM children of the add-rule-modal element. Scoping these to the modal (via .within() or chaining from the modal) would cause the test to fail, which is exactly what happened during development when we originally had cy.findByTestId('add-rule-modal').contains(...).

The matched strings (Use custom API group "apps", Use custom resource type "deployments") are dynamically generated creatable-option labels unique to what's being typed, so the risk of accidental matches elsewhere on the page is negligible.

srtanish1992:
@DaoDaoNoCode I have addressed all your feedback.

DaoDaoNoCode:
## Round 2 Review — All findings addressed ✅

Verified all 13 findings from the first review against the latest commit (de36e67):

**Both blockers fixed:**
- ✅ `assembleRole` now accepts and merges user labels (with dashboard label override protection)
- ✅ Cypress tests use page object methods via extracted `addRule()` helper; `data-testid` added to `VerbsTreeSelect.tsx` checkProps

**All 7 suggestions adopted:**
- ✅ Redundant test replaced with `'should omit optional fields when undefined'` + 2 label merge tests
- ✅ Dangling promise resolved with `await waitFor` after `resolvePromise()`
- ✅ Error assertions now check `error-message-alert` test ID
- ✅ Non-Error rejection fallback path tested
- ✅ Footer rejection path tested (`'should re-enable submit button after onSubmit rejects'`)
- ✅ Confirmation-modal API error path tested in Cypress
- ✅ `visitCreateRole` includes `cy.testA11y()`

**All prior CodeRabbit/agent threads verified resolved in code.**

No new issues found. Clean from my side.

DaoDaoNoCode:
## CI Failure: `Cypress-Mock-Tests (projects/tabs)` — Root Cause

Two tests in `rolesTab.cy.ts` are failing with `Unable to find an element by: [data-testid="create-role-page"]` (10s timeout):

1. `"should redirect from /roles/create to the project page when flag is disabled"` (line 45)
2. `"should redirect from /roles/create when user lacks create permission"` (line 188)

**Cause:** This PR added `cy.findByTestId('create-role-page')` + `cy.testA11y()` to `visitCreateRole()` in the page object. But these two tests intentionally visit `/roles/create` expecting to be **redirected away** — the `create-role-page` element never appears because the redirect fires first, so `visitCreateRole()` hangs and times out.

**Fix:** Replace `projectRoles.visitCreateRole(NAMESPACE)` with `cy.visitWithLogin(...)` in the two redirect tests, since they're testing that the page does NOT load:

```diff
  it('should redirect from /roles/create to the project page when flag is disabled', () => {
-   projectRoles.visitCreateRole(NAMESPACE);
+   cy.visitWithLogin(`/projects/${NAMESPACE}/roles/create`);
    cy.url().should('not.include', '/roles/create');
    ...
  });

  it('should redirect from /roles/create when user lacks create permission', () => {
-   projectRoles.visitCreateRole(NAMESPACE);
+   cy.visitWithLogin(`/projects/${NAMESPACE}/roles/create`);
    cy.url().should('not.include', '/roles/create');
    ...
  });
```

The redirect tests should bypass the page object's load assertions since they're verifying the page *doesn't* load.

DaoDaoNoCode (approved):
/lgtm

PR #8097, opened 2026-06-15
Update the target of the image tag to the branch name

andrewballantyne:
[Slack thread](https://redhat-internal.slack.com/archives/C05SMJ09DD2/p1781539886817129?thread_ts=1781196596.624579&cid=C05SMJ09DD2) -- we are switching the tag of the tekton pipeline to fit the branch name

That way we can manage the quay image via the name of this branch -- and not to the release it uses.

The release tag previous will be used in the last ODH Operator release (3.5 EA2). We'll be improving the process as time goes along, but this will produce an on-going image that one can reference via the tag for the branch.

### DIFF
--- a/.tekton/mod-arch-maas-push.yaml
+++ b/.tekton/mod-arch-maas-push.yaml
@@ -25,7 +25,7 @@ spec:
   - name: revision
     value: '{{revision}}'
   - name: output-image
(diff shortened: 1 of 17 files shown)


manaswinidas (approved):
/approve

PR #8109, opened 2026-06-15
fix(RHOAIENG-66822): improve MultiSelection VoiceOver support in modals

danreed-rh:
https://issues.redhat.com/browse/RHOAIENG-66822

## Description

Fixes A11y text reader navigation in `MultiSelection` typeahead dropdowns, including inside modals such as **Attach existing connections**.

Changes align the component with PatternFly's multiple typeahead accessibility pattern:
- Stable option DOM ids wired to `aria-activedescendant`
- Combobox `aria-label` and dynamic `aria-controls` / listbox id
- Single `SelectList` with consolidated option rendering and keyboard focus tracking

Modal-specific behavior:
- Portal the menu into the nearest `[role="dialog"]` so options stay inside the `aria-modal` subtree for screen readers
- Temporarily set the dialog to `overflow: visible` while the menu is open so the list can extend past the modal without adding a modal scrollbar

Also resets keyboard focus when a selection clears the filter input, preventing stale `focusedItemIndex` after the visible option list changes.

## How Has This Been Tested?

- Manual testing with VoiceOver on localhost (`npm run dev`): Attach existing connections modal — arrow-key navigation announces options; dropdown overflows the modal without modal scrollbar
- `npm run test-unit -- --testPathPattern='MultiSelection|ConnectionsFormSection'`

## Test Impact

- Added `frontend/src/components/__tests__/MultiSelection.spec.tsx` (aria-activedescendant wiring, listbox id, focus reset on select)
- Updated `ConnectionsFormSection.spec.tsx` combobox accessible name assertion (`Connections` vs PatternFly default `Type to filter`)

## Intentional behavior changes

- **MultiSelection in modals:** dropdown menu portals into the nearest `[role="dialog"]`; modal `overflow` is unlocked while open.
- **SimpleSelect labeling:** when `toggleProps.id` matches a visible label (e.g. FormGroup `fieldId`), the toggle no longer uses generic `aria-label="Options menu"`.
- **SimpleSelect portal:** only applies inside modals; non-modal selects remain inline (or use caller `popperProps.appendTo`).
- **Cypress:** page objects query `"Variable type"` / `"Data type"` instead of `"Options menu"` where labels apply.

## Manual screen reader verification checklist

Test on Windows 11 with JAWS or NVDA (or VoiceOver on macOS as a smoke check):

### Attach existing connections

- [ ] Open workbench spawner → Connections → Attach existing connections modal
- [ ] Tab to the Connections combobox; screen reader announces "Connections"
- [ ] Arrow Down moves through connection options; each option is announced
- [ ] Enter or Space selects an option
- [ ] Escape closes the menu

### Environment variable type

- [ ] Open workbench spawner → Environment variables → Add variable
- [ ] Tab to Variable type; screen reader announces "Variable type" (not "Options menu")
- [ ] Open the list; options (Config Map, Secret, etc.) are announced
- [ ] Select an option; selection is announced
- [ ] Repeat for Data type sub-select after choosing a variable type

## Request review criteria:

If you have UI changes:

## Summary by CodeRabbit

* **Accessibility**
  * Improved multi-selection and dropdown menus with deterministic ARIA active-option tracking, updated keyboard handling (Enter/Escape/Tab/Arrow navigation), and “No results found” live announcements.
  * Enhanced labeling/identifiers for environment variable type selectors.

* **Bug Fixes**
  * Improved select/menu behavior inside dialogs, including correct temporary overflow handling and restoration after the menu closes.

* **Tests**
  * Added and updated unit and end-to-end tests covering keyboard/ARIA behavior, portal rendering, dialog overflow restoration, and environment variable workflows.


DaoDaoNoCode:
## Consolidated Review — MultiSelection A11y Refactor

Reviewed the implementation, new tests, and performed a **full consumer impact analysis** across all 14 usage sites in the monorepo.

---

### ✅ Consumer Impact: All Clear

Checked every file importing `MultiSelection` or `SelectionOptions` across `frontend/` and `packages/`:

| Check | Result |
|-------|--------|
| All consumers pass `ariaLabel` | ✅ Safe |
| 5 consumers don't pass `id` → `React.useId()` fallback | ✅ Safe |
| Zero consumers pass `popperProps.appendTo` → modal portal default | ✅ Safe |
| `data-testid` regex change (single→all whitespace) | ✅ No test fixtures use multi-space names |
| `listTestId` (only `FeatureStoreSelector`) with single `SelectList` | ✅ Still works |
| `ConnectionsFormSection.spec.tsx` name change (`'Type to filter'` → `'Connections'`) | ✅ Correct — reflects `ariaLabel` now on combobox |

**Consumers verified:** `SelectConnectionsModal`, `AddRuleModal` (×2), `ConnectedNotebookField`, `HardwareProfileVisibilitySection`, `ManageConnectionTypePage`, `HardwareProfileIdentifierMultiselect`, `BiasMetricConfigSelector`, `FeatureStoreSelector`, `PolicyForm`, `CreateSubscriptionForm`, `FeatureStoreLineageToolbar` (×4).

---

### 👍 Positive Changes

- **`variant="typeahead"` on `<Select>`** — was actually *missing* before. PF's own `SelectMultiTypeahead` example includes this. Without it, PF's built-in arrow-key focus management conflicted with the custom keyboard handler. Good fix.
- **Instance-scoped IDs** — the old hardcoded `"select-multi-typeahead-listbox"` would collide across multiple instances on the same page. Fixed correctly.
- **`aria-label` on `TextInputGroupMain`** — was previously only on `MenuToggle`, not the actual `role="combobox"` input. Screen readers now correctly announce the combobox name.

---

### ⚠️ Findings

#### 1. [Major] Modal overflow race condition with multiple instances (`MultiSelection.tsx` ~L213-224)

The `useLayoutEffect` saves `dialog.style.overflow`, sets it to `'visible'`, and restores on cleanup. If two `MultiSelection` instances exist in the **same modal** (e.g., `AddRuleModal` has two: API groups + resource types):

1. Instance A opens → saves `previousOverflow = 'auto'`, sets `overflow: visible`
2. Instance B opens → saves `previousOverflow = 'visible'` (set by A!), sets `overflow: visible`
3. Instance A closes → restores `overflow = 'auto'` — **clips Instance B's still-open dropdown**
4. Instance B closes → restores `overflow = 'visible'` — leaves dialog with wrong overflow

**Suggested fix:** Use a ref-counter on the dialog element (e.g., a `data-overflow-unlock-count` attribute) — only restore the original overflow when the counter reaches 0.

In practice this is low-risk since users rarely have two dropdowns open simultaneously, but `AddRuleModal` has two `MultiSelection` instances side-by-side so it's not impossible.

#### 2. [Major] `SelectGroup` inside `SelectList` produces invalid HTML (`MultiSelection.tsx` ~L441-451)

The old code followed PF's grouped pattern: `Select > SelectGroup > SelectList > SelectOption`. The new code inverts to: `Select > SelectList > SelectGroup > SelectOption`. This puts `<section>` elements as direct children of `<ul role="listbox">`, which is invalid per HTML spec (only `<li>` is valid inside `<ul>`).

The consolidation to a single `SelectList` is motivated by needing one `id` for `aria-controls` — which is a good reason. But the nesting deviation from PF's documented `SelectGrouped` example should be verified. Since you've already tested with VoiceOver, could you confirm grouped options (e.g., `BiasMetricConfigSelector` which uses `groupedValues`) still announce group labels correctly?

#### 3. [Minor] `encodeOptionIdForDom` has theoretical collision (`MultiSelection.tsx` L72-73)

The encoding `u${charCode}u` is not truly injective — an option id literally containing `"u47u"` (the encoding of `/`) would collide with an id containing `/`. Practical risk is negligible since option ids like `"u47u"` are unlikely, but for correctness: escaping `u` itself (e.g., `u` → `uu117uu`) would make it injective.

#### 4. [Minor] Missing test coverage for keyboard navigation edges (`MultiSelection.spec.tsx`)

The new tests cover the happy path well. Missing coverage for:
- **ArrowUp wrapping** from first → last option (the code explicitly implements this)
- **Escape / Tab closing** the menu (explicit `case` branches in `onInputKeyDown`)
- **Clear-all button** resetting focus state (new `resetActiveAndFocusedItem()` call added by this PR)
- **`onToggleClick`** focusing first option (behavioral change from old code)

#### 5. [Nit] Unnecessary import in test file (`MultiSelection.spec.tsx` L2)

`import '@testing-library/jest-dom'` is already configured globally via `packages/jest-config`. Redundant here and inconsistent with other test files in the PR (e.g., `ConnectionsFormSection.spec.tsx` doesn't import it).

---

### Already-Reviewed Threads

- ✅ RESOLVED: Scope ARIA DOM ids per instance — fixed in dc87d55
- ✅ RESOLVED: `event.preventDefault()` on Enter — fixed in dc87d55
- ⬜ UNRESOLVED: CodeRabbit's `fireEvent.keyDown` doesn't test `preventDefault` effectively — agreed, not re-raising

danreed-rh:

danreed-rh:

frontend/src/components/MultiSelection.tsx
```diff
@@ -176,25 +180,84 @@ export const MultiSelection: React.FC<MultiSelectionProps> = ({
...
+    }
+  };
+
+  const closeMenu = () => {
+    setIsOpen(false);
+    setInputValue('');
+    resetActiveAndFocusedItem();
+  };
+
   React.useEffect(() => {
```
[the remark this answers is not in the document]
danreed-rh: Updated in most recent commit

frontend/src/components/MultiSelection.tsx
```diff
@@ -176,25 +180,84 @@ export const MultiSelection: React.FC<MultiSelectionProps> = ({
...
+    if (!isOpen) {
+      setIsOpen(true);
+    }
 
     const optionsLength = visibleOptions.length;
+    if (optionsLength === 0) {
+      return;
+    }
+
+    let indexToFocus = 0;
```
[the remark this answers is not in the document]
danreed-rh: Updated in most recent commit

DaoDaoNoCode:
## Re-review — 5-agent deep dive (7 files, 742 additions)

Went line-by-line through both `MultiSelection.tsx` and `SimpleSelect.tsx`, checked all consumers (30+ SimpleSelect, 14+ MultiSelection across frontend + packages), and reviewed all test files. All 11 prior threads were checked — only posting new findings below.

### Consumer Impact: All Clear

Verified every consumer of both components. No breakage found:
- **MultiSelection** (14 consumers): no one passes `popperProps`, all pass `ariaLabel`, `data-testid` regex change is backwards-compatible for all existing option names
- **SimpleSelect** (30+ consumers): the `appendTo` read-back pattern preserves all caller overrides. `EnvDataTypeField`/`EnvTypeSelectField` had their explicit `getDashboardMainContainer()` removed — safe because they're on the spawner page (not in a modal), so `body` fallback is equivalent
- **No toggleProps consumer passes `ref`**, so the `ref` → `innerRef` change is safe
- **`display: contents` wrapper** is layout-transparent, PF's own Popper uses the same pattern

### Prior threads status

| Thread | Status |
|--------|--------|
| ARIA DOM id scoping | ✅ Resolved |
| Enter preventDefault | ✅ Resolved |
| fireEvent vs createEvent for test | ✅ Resolved |
| DOM ID encoding collision | ✅ Resolved |
| createOptionElementId inside component | ✅ Resolved (moved to module scope) |
| handleMenuArrowKeys fall-through | ✅ Resolved (guard added) |
| useEffect inputValue reset | ✅ Resolved (reset added) |
| setTimeout not cleared | ⬜ Still open |
| renderSelectOption inside component | ⬜ Still open |
| appendTo inline function | ⬜ Still open |
| Enter test with menu closed | ⬜ Still open |

### New findings (posted inline below)

1. **[Medium]** Overflow ref-counting logic duplicated verbatim (~29 lines) between both files — extract to shared hook
2. **[Medium]** Missing `eslint-disable` for `react-hooks/exhaustive-deps` on two hooks
3. **[Low]** `visibleIndexById` Map uses strict equality — numeric option ids would silently break `isFocused`
4. **[Low]** `inputValue` effect reset is dead code for the typing path — all callers already reset synchronously
5. **[Low]** No cross-component test for MultiSelection + SimpleSelect sharing overflow ref-count in the same modal
6. **[Low]** SimpleSelect.spec.tsx has minimal coverage — only 1 test for significant structural changes

Overall this is solid work. The a11y improvements are substantial, the ref-counting approach for modals is clever, and the `variant="typeahead"` addition was a real bug fix (PF was double-handling arrow keys before). The consumer impact analysis came back completely clean. Nice job.

frontend/src/components/SimpleSelect.tsx
```diff
@@ -106,6 +113,34 @@ const SimpleSelect: React.FC<SimpleSelectProps> = ({
...
+        dialog.style.overflow = dialog.getAttribute(MODAL_OVERFLOW_ORIGINAL_ATTR) ?? '';
+        dialog.removeAttribute(MODAL_OVERFLOW_UNLOCK_COUNT_ATTR);
+        dialog.removeAttribute(MODAL_OVERFLOW_ORIGINAL_ATTR);
+      } else {
+        dialog.setAttribute(MODAL_OVERFLOW_UNLOCK_COUNT_ATTR, String(nextCount));
+      }
+    };
+  }, [open]);
+
   if (isSkeleton) {
```
DaoDaoNoCode, line 131: The overflow ref-counting `useLayoutEffect` is copy-pasted verbatim (~29 lines) between `MultiSelection.tsx` and `SimpleSelect.tsx`. The comment on line 61 acknowledges this intentionally — but if someone fixes a bug in one copy, the other will silently drift.

Worth extracting to something like `useModalOverflowUnlock(isOpen, anchorRef)` in `utilities/`. Both components would call it with their respective ref. Not blocking, but this is the kind of thing that bites someone 6 months from now.

The `frontend/src/utilities/` directory already has ~30 custom hooks — this would fit right in.

frontend/src/components/MultiSelection.tsx
```diff
@@ -176,81 +181,145 @@ export const MultiSelection: React.FC<MultiSelectionProps> = ({
...
+    setInputValue('');
+    resetActiveAndFocusedItem();
+  };
+
   React.useEffect(() => {
     if (inputValue) {
-      setOpen(true);
+      setIsOpen(true);
+      resetActiveAndFocusedItem();
     }
```
DaoDaoNoCode: Heads up — the project has `react-hooks/exhaustive-deps` set to `error`. This effect references `resetActiveAndFocusedItem` (line 218) which is a plain function recreated each render, not in the deps array. Same thing with `getModalDialog` in the overflow `useLayoutEffect` below.

Both are safe in practice (they read from refs), but ESLint cannot infer that. You might need `// eslint-disable-next-line react-hooks/exhaustive-deps` here with a note like "reads from refs only", or wrap them in `useCallback`. Same applies to the identical effect in `SimpleSelect.tsx`.

If CI is not failing on this, PF might be suppressing the warning somehow — but worth double-checking.

frontend/src/components/MultiSelection.tsx
```diff
@@ -176,81 +181,145 @@ export const MultiSelection: React.FC<MultiSelectionProps> = ({
     return options;
   }, [groupOptions, selectOptions, createOption, isCreateOptionOnTop]);
 
+  const visibleIndexById = React.useMemo(() => {
```
DaoDaoNoCode: Tiny thing — `Map` uses `===` for lookups, and `SelectionOptions.id` is typed as `number | string`. If a consumer ever passed numeric ids (like `{ id: 1 }`), the `visibleIndexById.get(option.id)` lookup would work, but PF's `onSelect` callback might return the id as a string (since `SelectOptionProps.value` is `any`). Then `allOptions.find(option => option.id === selection)` on line 340 would fail to match (`1 !== "1"`).

All current consumers use string ids, so this is safe today. But the type signature advertises `number | string` — either narrowing the type to `string` or normalizing with `String(id)` in the Map would make this bulletproof.

frontend/src/components/__tests__/SimpleSelect.spec.tsx
```diff
@@ -0,0 +1,41 @@
...
+    });
+    expect(dialog.style.overflow).toBe('visible');
+
+    await act(async () => {
+      fireEvent.keyDown(toggle, { key: 'Escape' });
+    });
+    expect(dialog.style.overflow).toBe('auto');
+    expect(dialog.getAttribute('data-multiselection-overflow-unlock-count')).toBeNull();
+  });
+});
```
DaoDaoNoCode, line 289: This test file has solid coverage for the overflow mechanism, but there's a lot of new behavior in SimpleSelect that's untested:

- The `display: contents` wrapper + `innerRef` change — does the toggle still get correct focus/positioning? Even a basic "should render toggle with correct aria-label" test would catch regressions.
- Mixed component test: a SimpleSelect + MultiSelection in the same `[role="dialog"]` sharing the overflow ref-count. The MultiSelection tests only test two MultiSelections together. If the attribute names ever diverge between the two files (they're defined independently as separate constants), overflow would break with no test catching it.

Not asking for exhaustive coverage here — just enough to guard the structural changes.

danreed-rh:

danreed-rh:

danreed-rh:

frontend/src/components/SimpleSelect.tsx
danreed-rh: Resolved — overflow ref-counting is now shared via `useModalOverflowUnlock` in `frontend/src/utilities/useModalOverflowUnlock.ts`. Both components call that hook; `useModalOverflowUnlock.spec.tsx` covers ref-counting.

frontend/src/components/MultiSelection.tsx
danreed-rh: Resolved — the duplicated `useLayoutEffect` was removed when overflow handling moved into `useModalOverflowUnlock`, which includes an `eslint-disable-next-line react-hooks/exhaustive-deps` note (anchor ref is stable; dialog resolved at effect time).

frontend/src/components/MultiSelection.tsx
danreed-rh: Resolved — `normalizeOptionId` coerces ids with `String(optionId)` for comparisons in `onSelect` and option keys. Added `should select options when option id is numeric` in `MultiSelection.spec.tsx`.

frontend/src/components/__tests__/SimpleSelect.spec.tsx
danreed-rh: Resolved — added `should render toggle inside display-contents anchor with data-testid` and `should share modal overflow ref-count with MultiSelection in the same dialog` in `SimpleSelect.spec.tsx`.

frontend/src/components/MultiSelection.tsx
```diff
@@ -94,44 +103,39 @@ export const MultiSelection: React.FC<MultiSelectionProps> = ({
   const [isOpen, setIsOpen] = React.useState(false);
   const [inputValue, setInputValue] = React.useState<string>('');
   const [focusedItemIndex, setFocusedItemIndex] = React.useState<number | null>(null);
-  const [activeItem, setActiveItem] = React.useState<string | null>(null);
+  const [activeItemId, setActiveItemId] = React.useState<string | null>(null);
   const textInputRef = React.useRef<HTMLInputElement>();
+  const generatedInstanceId = React.useId().replace(/:/g, '');
+  const instanceId = id ?? `multi-select-${generatedInstanceId}`;
+  const listboxId = `${instanceId}-listbox`;
+  const createOptionElementId = (optionId: number | string) =>
```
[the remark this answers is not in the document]
danreed-rh: Resolved — `createOptionElementId` is now module-level (line 82), not inside the component body.

frontend/src/components/MultiSelection.tsx
```diff
@@ -212,45 +275,49 @@ export const MultiSelection: React.FC<MultiSelectionProps> = ({
...
   };
 
   const onToggleClick = () => {
-    setOpen(!isOpen);
+    if (!isOpen) {
+      openMenu(true);
+    } else {
+      closeMenu();
+    }
     setTimeout(() => textInputRef.current?.focus(), 100);
```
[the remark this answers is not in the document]
danreed-rh: Resolved — `focusTimeoutRef` tracks the timer; cleared in `closeMenu`, `onToggleClick`, and on unmount via `useEffect` cleanup.

frontend/src/components/MultiSelection.tsx
```diff
@@ -259,12 +326,36 @@ export const MultiSelection: React.FC<MultiSelectionProps> = ({
...
+      resetActiveAndFocusedItem();
     }
     textInputRef.current?.focus();
   };
 
   const noSelectedItems = allOptions.filter((option) => option.selected).length === 0;
 
+  const getOptionTestId = (name: string) => `select-multi-typeahead-${name.replace(/\s+/g, '-')}`;
+
+  const renderSelectOption = (
```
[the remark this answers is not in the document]
danreed-rh: Resolved — option rendering extracted to `MultiSelectionOption`; remaining `renderSelectOption` is a thin map helper over that component.

frontend/src/components/MultiSelection.tsx
```diff
@@ -340,78 +433,53 @@ export const MultiSelection: React.FC<MultiSelectionProps> = ({
...
+          }
+        }}
         toggle={toggle}
-        popperProps={popperProps}
+        variant="typeahead"
+        popperProps={{
+          ...popperProps,
+          // Portal into the modal dialog (not document.body) so VoiceOver can reach options
+          // inside aria-modal; overflow unlock above lets the menu extend past the dialog edge.
+          appendTo: popperProps?.appendTo ?? (() => getModalDialog() ?? document.body),
```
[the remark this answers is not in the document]
danreed-rh: Resolved — portal target uses stable `getPopperAppendTo` via `useCallback` in `mergedPopperProps`.

frontend/src/components/__tests__/MultiSelection.spec.tsx
```diff
@@ -0,0 +1,422 @@
...
+
+    await act(async () => {
+      fireEvent.keyDown(combobox, { key: 'Enter' });
+    });
+
+    expect(combobox).not.toHaveAttribute('aria-activedescendant');
+    expect(combobox).toHaveValue('');
+  });
+
+  it('should prevent default Enter behavior to avoid submitting parent forms', async () => {
```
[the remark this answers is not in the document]
danreed-rh: Resolved — test opens the menu with `ArrowDown` before `Enter` and asserts `defaultPrevented`.

frontend/src/components/SimpleSelect.tsx
```diff
@@ -97,19 +120,23 @@ const SimpleSelect: React.FC<SimpleSelectProps> = ({
...
-  }, [singleOptionKey, isSkeleton, autoSelectOnlyOption]);
+  }, [singleOptionKey, isSkeleton, autoSelectOnlyOption, value]);
 
   if (isSkeleton) {
     return <Skeleton style={{ minWidth: 100 }} />;
   }
 
+  const displayedLabel = toggleLabel ?? selectedLabel;
+  const displayedLabelTitle = typeof displayedLabel === 'string' ? displayedLabel : undefined;
+  const toggleAriaLabel = ariaLabel ?? (toggleProps?.id ? undefined : 'Options menu');
```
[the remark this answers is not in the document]
danreed-rh: Resolved — Cypress page objects updated to use test-id toggles and correct accessible names; mock tests pass.

frontend/src/components/SimpleSelect.tsx
```diff
@@ -77,6 +83,23 @@ const SimpleSelect: React.FC<SimpleSelectProps> = ({
...
+  const getPopperAppendTo = React.useCallback(
+    () => resolveSelectPopperAppendTo(menuToggleRef.current),
+    [],
+  );
+
+  const mergedPopperProps = React.useMemo(
+    () => ({
+      maxWidth: 'trigger' as const,
+      ...popperProps,
+      appendTo: popperProps?.appendTo ?? getPopperAppendTo,
```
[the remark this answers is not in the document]
danreed-rh: Resolved — `findDropdownItemByTestId` uses global `cy.findByTestId` for portaled options; mock tests pass.

packages/cypress/cypress/support/commands/application.ts
```diff
@@ -476,7 +476,8 @@ Cypress.Commands.add('findDropdownItemByTestId', { prevSubject: 'element' }, (su
     if ($el.attr('aria-expanded') === 'false') {
       cy.wrap($el).click();
     }
-    return cy.wrap($el).parent().findByTestId(testId);
+    // Options may render inline or in a portal (e.g. modal dialog or document.body).
+    return cy.findByTestId(testId);
```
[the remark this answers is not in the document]
danreed-rh: Resolved — reverted `:visible` filter (broke portaled PF options in CI); global `cy.findByTestId` restored. Env-var helpers use `findSelectOptionByTestId` chained from the toggle.

packages/cypress/cypress/tests/mocked/projects/tabs/workbench.cy.ts
```diff
@@ -675,6 +674,31 @@ describe('Workbench page', () => {
...
+    const environmentVariableField = createSpawnerPage.getEnvironmentVariableTypeField(0);
+    environmentVariableField
+      .find()
+      .findByTestId('environment-variable-type-select')
+      .findByRole('button', { name: 'Variable type' })
+      .should('have.attr', 'aria-expanded', 'false')
+      .click()
+      .should('have.attr', 'aria-expanded', 'true');
+
+    cy.findByRole('option', { name: 'Config Map' }).click();
```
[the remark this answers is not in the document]
danreed-rh: Resolved — a11y test opens the toggle, then clicks `cy.findByTestId('Config Map')` while the menu is expanded.

frontend/src/components/SimpleSelect.tsx
```diff
@@ -97,19 +122,25 @@ const SimpleSelect: React.FC<SimpleSelectProps> = ({
...
+    if (singleOptionKey && !isSkeleton && autoSelectOnlyOption && value !== totalOptions[0].key) {
+      onChange(totalOptions[0].key, totalOptions[0].isPlaceholder ?? false);
     }
     // We don't want the callback function to be a dependency
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [singleOptionKey, isSkeleton, autoSelectOnlyOption]);
+  }, [singleOptionKey, isSkeleton, autoSelectOnlyOption, value, totalOptions]);
 
   if (isSkeleton) {
     return <Skeleton style={{ minWidth: 100 }} />;
```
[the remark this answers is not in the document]
danreed-rh: Resolved — uses `var(--pf-t--global--spacer--3xl)` instead of hardcoded `100`.

frontend/src/components/SimpleSelect.tsx
```diff
@@ -128,35 +159,32 @@ const SimpleSelect: React.FC<SimpleSelectProps> = ({
...
-            {(() => {
-              const displayedLabel = toggleLabel ?? selectedLabel;
-              return (
-                <span title={typeof displayedLabel === 'string' ? displayedLabel : undefined}>
-                  {displayedLabel}
-                </span>
-              );
-            })()}
-          </MenuToggle>
+          <div ref={menuToggleRef} style={{ display: 'contents' }}>
```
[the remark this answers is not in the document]
danreed-rh: Resolved — moved to SCSS class (now `odh-simple-select__toggle-anchor`).

frontend/src/components/SimpleSelect.scss
```diff
@@ -7,3 +7,8 @@
   --pf-v6-c-truncate--MinWidth: 0;
   --pf-v6-c-truncate__start--MinWidth: 0;
 }
+
+// innerRef anchor without adding a layout box (keeps parent flex/grid intact).
+.simple-select__toggle-anchor {
```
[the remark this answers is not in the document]
danreed-rh: Resolved — renamed to `.odh-simple-select__toggle-anchor` per `css-patternfly.md`.

frontend/src/utilities/useModalOverflowUnlock.ts
```diff
@@ -0,0 +1,58 @@
...
+/** Resolve popper mount target: portal into the nearest modal dialog for VoiceOver + aria-modal. */
+export const resolveSelectPopperAppendTo = (
+  anchor: HTMLElement | null | undefined,
+): HTMLElement => {
+  const dialog = anchor?.closest<HTMLElement>(MODAL_DIALOG_SELECTOR);
+  if (dialog) {
+    return dialog;
+  }
+  return anchor?.parentElement ?? document.body;
+};
```
DaoDaoNoCode: This `resolveSelectPopperAppendTo` helper does the same dialog-ancestor lookup that `SimpleSelect.tsx` reimplements inline (lines 97-101) and that `MultiSelection.tsx` calls through `getPopperAppendTo`. That's three copies of the same `closest('[role="dialog"]')` → fallback chain.

SimpleSelect could just use this exported helper instead of inlining its own version:

```ts
// SimpleSelect.tsx, inside mergedPopperProps
appendTo: () => resolveSelectPopperAppendTo(menuToggleRef.current),
```

Not a bug, but keeping one source of truth here means you only need to update the selector in one place if the dialog detection logic ever changes (e.g., nested modals, PF markup changes).

frontend/src/utilities/useModalOverflowUnlock.ts
```diff
@@ -0,0 +1,58 @@
+import * as React from 'react';
+
+export const MODAL_OVERFLOW_UNLOCK_COUNT_ATTR = 'data-modal-overflow-unlock-count';
+
+const MODAL_DIALOG_SELECTOR = '[role="dialog"][aria-modal="true"], [role="dialog"]';
+
+const originalOverflowByDialog = new WeakMap<HTMLElement, string>();
+const unlockCountByDialog = new WeakMap<HTMLElement, number>();
```
DaoDaoNoCode: The module-level `WeakMap`s for ref-counting are a smart choice — they avoid leaking memory and survive component re-renders correctly. One edge case to be aware of: if a dialog element is removed from the DOM while a dropdown is still open (e.g., a parent route change unmounts the modal), the `useLayoutEffect` cleanup will still run and try to restore `dialog.style.overflow` on a detached element. That's harmless (no error, just a no-op on a GC'd node), but worth a quick comment so future readers don't wonder about it.

Also — since the `WeakMap`s are module-scoped, they're shared across all React roots on the page. For this app that's fine (single root), but if this utility ever gets extracted to a shared package, you'd want to consider scoping via a React context instead.

packages/ui-core/src/components/SimpleSelect.tsx
```diff
@@ -97,19 +122,26 @@ const SimpleSelect: React.FC<SimpleSelectProps> = ({
...
   if (isSkeleton) {
-    return <Skeleton style={{ minWidth: 100 }} />;
+    return <Skeleton style={{ minWidth: 'var(--pf-t--global--spacer--3xl)' }} />;
   }
 
+  const displayedLabel = toggleLabel ?? selectedLabel;
+  const displayedLabelTitle = typeof displayedLabel === 'string' ? displayedLabel : undefined;
+  const togglePropsAriaLabel = toggleProps?.['aria-label'];
+  const toggleAriaLabel =
+    ariaLabel ?? togglePropsAriaLabel ?? (toggleProps?.id ? undefined : 'Options menu');
```
DaoDaoNoCode, line 139: This labeling logic is correct for all current callers — I checked all ~56 `SimpleSelect` usages and every one that passes `toggleProps.id` also has a matching `<FormGroup label="..." fieldId={sameId}>`, so the implicit `<label htmlFor>` association provides the accessible name.

But the fallback chain here is a bit of a footgun for future callers: if someone passes `toggleProps={{ id: 'my-select' }}` without a corresponding `<label htmlFor="my-select">`, the toggle ends up with *no accessible name at all* — worse than the old generic `"Options menu"` fallback.

Would a defensive check like this be worth it?

```ts
const toggleAriaLabel =
  ariaLabel ?? togglePropsAriaLabel ?? (toggleProps?.id ? undefined : 'Options menu');

// In dev only:
if (process.env.NODE_ENV === 'development' && !toggleAriaLabel && toggleProps?.id) {
  console.warn(
    `SimpleSelect: toggleProps.id="${toggleProps.id}" provided without ariaLabel. ` +
    `Ensure a <label htmlFor="${toggleProps.id}"> exists, or pass ariaLabel explicitly.`
  );
}
```

Not blocking — just a suggestion to prevent future a11y regressions from sneaking in silently.

packages/ui-core/src/components/SimpleSelect.tsx
```diff
@@ -128,35 +160,30 @@ const SimpleSelect: React.FC<SimpleSelectProps> = ({
...
-            {(() => {
-              const displayedLabel = toggleLabel ?? selectedLabel;
-              return (
-                <span title={typeof displayedLabel === 'string' ? displayedLabel : undefined}>
-                  {displayedLabel}
-                </span>
-              );
-            })()}
-          </MenuToggle>
+          <div ref={menuToggleRef} className="odh-simple-select__toggle-anchor">
```
DaoDaoNoCode, line 166: The wrapper `<div ref={menuToggleRef}>` is needed to give `useModalOverflowUnlock` an anchor ref, since PF's `MenuToggle` uses `innerRef` rather than a standard `ref`. This makes sense.

One thing to double-check: does the extra wrapper `div` affect PF's flex/grid layout expectations for `MenuToggle`? The `className="odh-simple-select__toggle-anchor"` in the SCSS sets `display: contents`, which should make it layout-transparent. That's the right call — just want to confirm it's been tested with `isFullWidth` variants and inside `Split`/`Stack` layouts where PF might rely on direct child relationships.

frontend/src/components/MultiSelection.tsx
```diff
@@ -69,6 +73,52 @@ const defaultCreateOptionMessage = (newValue: string) => `Create "${newValue}"`;
 const defaultFilterFunction = (filterText: string, options: SelectionOptions[]) =>
   options.filter((o) => !filterText || o.name.toLowerCase().includes(filterText.toLowerCase()));
 
+/** Encode option ids for stable, injective DOM id segments (e.g. 'a b' vs 'a-b', 'core/pods' vs 'coreu47upods'). */
+const encodeOptionIdForDom = (optionId: number | string): string =>
```
DaoDaoNoCode, line 77: This encoding is well thought out — the double-`u` escaping prevents collisions that the old `replace(/\s+/g, '-')` had. Just to confirm my reading:

- `"foo bar"` → `"foou32ubar"` (space = char 32)
- `"foo-bar"` → `"foo-bar"` (hyphen is in the safe set)
- `"uber"` → `"uuber"` (leading `u` is escaped)

So `"foo bar"` and `"foo-bar"` produce distinct DOM ids. 👍

The only edge case I can think of: option ids that are numbers (e.g., `id: 123`). `String(123)` is all `[0-9]`, so it passes through unchanged — that's correct.

Minor style nit: the function name `encodeOptionIdForDom` is clear, but the inline comment on line 76 already explains it well enough that the JSDoc-style comment above could be dropped to reduce visual noise. Up to you.

frontend/src/components/MultiSelection.tsx
```diff
@@ -178,100 +240,204 @@ export const MultiSelection: React.FC<MultiSelectionProps> = ({
...
         }
         break;
       case 'Tab':
+        closeMenu();
+        break;
       case 'Escape':
-        setOpen(false);
-        setActiveItem(null);
+        if (isOpen) {
+          event.preventDefault();
```
DaoDaoNoCode, line 345: The `event.stopPropagation()` on Escape is important here — without it, pressing Escape inside a modal dropdown would close both the dropdown *and* the modal. PF's own modal docs specifically call this out:

> *"You should also handle the modal's closing behavior by listening to the `onEscapePress` callback on the `<Modal>` component. This allows the 'escape' key to collapse the dropdown without closing the entire modal."*

Good to see this handled correctly. The `event.preventDefault()` on line 345 is also right — it prevents the browser's default Escape behavior (e.g., exiting fullscreen).

frontend/src/components/MultiSelection.tsx
```diff
@@ -178,100 +240,204 @@ export const MultiSelection: React.FC<MultiSelectionProps> = ({
...
           onSelect(focusedItem);
-          setInputValue('');
         }
         if (!isOpen) {
-          setIsOpen(true);
+          openMenu(true);
         }
         break;
       case 'Tab':
+        closeMenu();
```
DaoDaoNoCode, line 341: Tab-to-close is a behavior change for all `MultiSelection` instances, not just modal ones. That said, it aligns with the WAI-ARIA combobox keyboard interaction pattern, and PF's own context selector docs confirm: *"Tab to navigate to the toggle... the Esc key closes the menu."*

For comboboxes specifically, Tab should close the popup and advance focus to the next focusable element, which is what this does. Just flagging it as an intentional UX change that's worth mentioning in the PR description (I see it's already in the scope-freeze doc — good).

frontend/src/pages/projects/screens/spawner/environmentVariables/EnvDataTypeField.tsx
```diff
@@ -1,38 +1,49 @@
 import * as React from 'react';
-import { Stack, StackItem } from '@patternfly/react-core';
+import { FormGroup, Stack, StackItem } from '@patternfly/react-core';
 import IndentSection from '#~/pages/projects/components/IndentSection';
-import { getDashboardMainContainer } from '#~/utilities/utils';
 import SimpleSelect, { SimpleSelectOption } from '#~/components/SimpleSelect';
+import { getDashboardMainContainer } from '#~/utilities/utils';
+
+const ENV_VAR_POPPER_PROPS = { appendTo: getDashboardMainContainer() };
```
DaoDaoNoCode, line 7: Nice pattern — hoisting `ENV_VAR_POPPER_PROPS` to module scope avoids creating a new object on every render. The old code was calling `getDashboardMainContainer()` inline inside JSX, which ran on every render *and* created a new `popperProps` object each time (triggering `useMemo` invalidation in SimpleSelect).

Same fix applied in `EnvTypeSelectField.tsx` — consistent.

frontend/src/pages/projects/screens/spawner/environmentVariables/EnvTypeSelectField.tsx
```diff
@@ -17,57 +19,63 @@ const EnvTypeSelectField: React.FC<EnvTypeSelectFieldProps> = ({
...
-                  env={envVariable}
-                  onUpdate={(envValue) => onUpdate({ ...envVariable, values: envValue })}
-                />
-              </IndentSection>
+}) => {
+  const selectId = React.useId().replace(/:/g, '');
+
+  return (
+    <FormGroup isRequired label="Variable type" fieldId={selectId}>
+      <Split data-testid="environment-variable-field">
```
DaoDaoNoCode, line 28: The `useId()` + `FormGroup` + `toggleProps.id` wiring here is textbook. The generated id connects:

1. `<FormGroup fieldId={selectId}>` → renders `<label htmlFor={selectId}>`
2. `toggleProps={{ id: selectId }}` → sets `id={selectId}` on the `MenuToggle`
3. `ariaLabel="Variable type"` → explicit aria-label as a belt-and-suspenders backup

This gives screen readers both the `<label>` association *and* the explicit `aria-label`. That's actually redundant (the label association alone would suffice), but harmless — and it means the accessible name survives even if the `FormGroup` is refactored later. Good defensive coding.

packages/cypress/cypress/support/commands/application.ts
```diff
@@ -476,7 +482,8 @@ Cypress.Commands.add('findDropdownItemByTestId', { prevSubject: 'element' }, (su
...
```
DaoDaoNoCode, line 426: The switch from `cy.wrap($el).parent().findByTestId(testId)` to `cy.findByTestId(testId)` (global scope) is necessary because portaled menus render outside the toggle's parent. But this means the command now searches the entire DOM for the test id — if two selects happen to have options with the same `data-testid`, this could match the wrong one.

In practice this is fine because test ids are unique per test scenario, but it's a subtle change in contract: `findDropdownItemByTestId` used to be scoped to the toggle's parent, now it's global. Worth a comment so future test authors know not to rely on parent-scoping.

packages/cypress/cypress/support/commands/application.ts
```diff
@@ -518,7 +525,32 @@ Cypress.Commands.add('findSelectOptionByTestId', { prevSubject: 'element' }, (su
...
+});
+
+Cypress.Commands.add('closeSelectMenu', { prevSubject: 'element' }, (subject) => {
+  Cypress.log({ displayName: 'closeSelectMenu' });
+  return cy.wrap(subject).then(($el) => {
+    const combobox = $el.find('[role="combobox"]');
+    const target = combobox.length ? combobox : $el;
+    const isExpanded =
+      target.attr('aria-expanded') === 'true' || $el.attr('aria-expanded') === 'true';
+
```
DaoDaoNoCode, line 479: The `closeSelectMenu` command is a nice addition. The manual `trigger('keydown', ...)` approach is needed because `cy.type('{tab}')` doesn't work reliably for programmatic Tab events.

One thing I'd flag: this dispatches a synthetic `keydown` event, but the component's handler (in `MultiSelection.tsx`) also calls `event.stopPropagation()` for Escape. For Tab, there's no `stopPropagation` — the event will bubble up. That's correct behavior (Tab should naturally advance focus), but if a parent component ever adds a Tab keydown listener, this synthetic event will trigger it. Just something to be aware of.

docs/RHOAIENG-66822-scope-freeze.md
```diff
@@ -0,0 +1,44 @@
+# RHOAIENG-66822 — Scope freeze and manual verification
```
DaoDaoNoCode: This is a useful checklist for the reviewer, but I'd question whether it belongs as a committed file in `docs/`. It's specific to this PR's review process and will be stale the moment the PR merges. The content would work well in the PR description or as a PR comment instead.

If you do want to keep it committed (e.g., as a template for future a11y PRs), maybe move it to something like `docs/a11y-verification-template.md` and generalize the checklist?

frontend/src/concepts/hardwareProfiles/utils.ts
```diff
@@ -51,7 +51,7 @@ export const formatToleration = (toleration: Toleration): string => {
 };
 
 export const formatNodeSelector = (selector: NodeSelector): string[] =>
-  Object.entries(selector).map(([key, value]) => `Key = ${key}; Value = ${value}`);
+  Object.entries(selector).map(([key, value]) => `Key = ${key}; Value = ${String(value)}`);
```
DaoDaoNoCode: The `String()` coercion here is a good defensive fix — if `value` is `undefined` or a non-string type, the template literal would produce `"Value = undefined"` instead of `"Value = "`. But this change is unrelated to the a11y work in this PR.

Would you consider splitting this and the `validationUtils.ts` `String()` fixes into a separate small PR? Keeps the review scope tight and makes the git history cleaner — easier to bisect if either change introduces an issue.

packages/ui-core/src/components/SimpleSelect.tsx
```diff
@@ -97,19 +122,26 @@ const SimpleSelect: React.FC<SimpleSelectProps> = ({
...
-  // If there is only one option, call the onChange function
+
   React.useEffect(() => {
-    if (singleOptionKey && !isSkeleton && autoSelectOnlyOption) {
-      onChange(totalOptions[0].key, false);
+    if (singleOptionKey && !isSkeleton && autoSelectOnlyOption && value !== totalOptions[0].key) {
+      onChange(totalOptions[0].key, totalOptions[0].isPlaceholder ?? false);
     }
     // We don't want the callback function to be a dependency
     // eslint-disable-next-line react-hooks/exhaustive-deps
```
DaoDaoNoCode, line 128: This auto-select `useEffect` got a meaningful behavior fix: the old version would re-fire `onChange` on every render when there was a single option (because `value` wasn't in the dependency array and there was no guard). Now it only fires when `value !== totalOptions[0].key`, preventing unnecessary state updates.

The `isPlaceholder` argument also changed from hardcoded `false` to `totalOptions[0].isPlaceholder ?? false` — that's more correct.

Good catch on both.

packages/ui-core/src/components/SimpleSelect.tsx
```diff
@@ -119,7 +151,7 @@ const SimpleSelect: React.FC<SimpleSelectProps> = ({
         onSelect={(e, selectValue) => {
           const key = String(selectValue);
           const option = findOptionForKey(key);
-          if (option?.isAriaDisabled) {
+          if (option?.isAriaDisabled || option?.isDisabled) {
             return;
```
DaoDaoNoCode, line 158: Adding `isDisabled` to the guard alongside `isAriaDisabled` is correct — the old code only checked `isAriaDisabled`, which meant a fully disabled option (not just aria-disabled) could still be selected via `onSelect`. PF treats these differently: `isAriaDisabled` renders the option but marks it as disabled for AT, while `isDisabled` prevents interaction entirely.

DaoDaoNoCode (commented):
## Review: a11y improvements for MultiSelection & SimpleSelect

Nice work on this — the core approach is solid. Portaling into `[role="dialog"]` with ref-counted overflow unlock is actually the right move for PF v6, since `focusTrapId` was removed from Modal in the v5→v6 migration. The injective `encodeOptionIdForDom` encoding and `React.useId()` scoping solve the DOM id collision problem cleanly.

I went through every changed file and cross-checked all ~56 SimpleSelect consumers and MultiSelection callers. Left a few inline comments — mostly code hygiene and one future-proofing suggestion. Nothing blocking.

Also verified the keyboard behavior against PatternFly's own accessibility docs: Tab-to-close on comboboxes matches the WAI-ARIA pattern, and PF's context selector docs confirm this is the expected interaction model.

frontend/src/components/__tests__/MultiSelection.spec.tsx
```diff
@@ -0,0 +1,383 @@
...
+    });
+
+    await act(async () => {
+      fireEvent.keyDown(combobox, { key: 'Enter' });
+    });
+
+    expect(setValue).toHaveBeenCalledWith(
+      expect.arrayContaining([expect.objectContaining({ id: 1, selected: true })]),
+    );
+  });
```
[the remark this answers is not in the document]
danreed-rh: Added `aria-activedescendant` and DOM id assertions after ArrowDown in the numeric-id test, so it now covers the ARIA contract alongside selection behavior.

frontend/src/utilities/useModalOverflowUnlock.ts
danreed-rh: Done — `SimpleSelect` now uses `resolveSelectPopperAppendTo(menuToggleRef.current)` instead of the inline dialog lookup. `MultiSelection` was already on the shared helper.

packages/ui-core/src/components/SimpleSelect.tsx
danreed-rh: Added a dev-only `console.warn` when `toggleProps.id` is set without `ariaLabel` or an associated label, to catch silent a11y regressions for future callers.

docs/RHOAIENG-66822-scope-freeze.md
danreed-rh: Removed the committed doc and moved the scope summary, intentional behavior changes, and manual verification checklist into the PR description.

danreed-rh:

frontend/src/components/MultiSelection.tsx
```diff
@@ -312,110 +494,100 @@ export const MultiSelection: React.FC<MultiSelectionProps> = ({
...
   return (
     <>
+      <div
+        aria-live="polite"
+        aria-atomic="true"
+        aria-relevant="additions text"
+        className="pf-v6-u-screen-reader"
+      >
+        {isOpen && visibleOptions.length === 0 && inputValue ? 'No results found' : ''}
+      </div>
```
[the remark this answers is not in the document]
DaoDaoNoCode: @danreed-rh Can you address this comment? It's valid.

DaoDaoNoCode:
The `resources.cy.ts` CI failure is related to this PR — the Cypress PO now looks for `findByRole('button', { name: 'Select sort type' })` but the button still has `'Options menu'` as its accessible name.

Root cause: `LearningCenterToolbar.tsx:203` passes `aria-label="Select sort type"` (HTML attribute, hyphenated) to `SimpleSelect`, but SimpleSelect uses `ariaLabel` (camelCase) as its prop for the toggle button name. The hyphenated `aria-label` gets spread onto the PF `<Select>` wrapper via `...rest` — it never reaches the `MenuToggle` button, which still gets the default `'Options menu'`.

Same issue on line 211 with `aria-label="Select sort order"`.

Fix: update `LearningCenterToolbar.tsx` to use the camelCase prop:
```diff
-              aria-label="Select sort type"
+              ariaLabel="Select sort type"
```
```diff
-              aria-label="Select sort order"
+              ariaLabel="Select sort order"
```

This would both fix the test and actually deliver the a11y improvement to the toggle button.

DaoDaoNoCode (approved):
/lgtm
🎉 Thanks for the great work!

PR #8119, opened 2026-06-15
ci: extend E2E cluster preflight checks before Cypress runs

Taj010:
## Summary

- Extends `check_dsc_health` in the Cypress E2E workflow to verify `OdhDashboardConfig` exists with a readable spec and that the test admin user can create projects before any Cypress job starts.
- Adds a guard in `getDashboardConfig` so empty `oc` output produces a clear cluster-preflight error instead of `SyntaxError: Unexpected end of JSON input`.

These changes fail fast in the `select-cluster` job (typically under 60s) with actionable messages and still allow primary → secondary cluster failover when preflight fails.

## Context

This addresses a subset of infra-related E2E failures (missing dashboard config CR, RBAC gaps). It does **not** address the larger coverage gap where many packages have not opted into `e2eCiTags` / CI tag alignment.

## Test plan

- [ ] Trigger **Actions → Cypress e2e Test → Run workflow** on this branch via `workflow_dispatch`
- [ ] Confirm `select-cluster` logs show `✅ OdhDashboardConfig is present` and `✅ Test user can create projects` on a healthy cluster
- [ ] Confirm regression suite (`@ci-dashboard-regression-tags`) still runs normally when preflight passes
- [ ] If primary cluster preflight fails, confirm secondary failover is attempted before job exit

## Summary of changes

- **Chores**
  - Strengthened Cypress end-to-end cluster health preflight checks with fail-fast validation for readiness and an additional fallback path when cluster signals indicate healthy conditions.

- **Tests**
  - Improved reliability of diagnostics by capturing retrieval failures and surfacing clearer error output when the required dashboard configuration is missing or unreadable.
  - Enhanced messaging to indicate potential missing configuration or insufficient access, including masked command error details.


.github/workflows/cypress-e2e-test.yml
```diff
@@ -12,6 +12,8 @@ name: Cypress e2e Test
 #   Primary:   dash-e2e-int (checked first via DSC health)
 #   Secondary: dash-e2e     (used if primary is unhealthy)
 #   Health Check: Logs into cluster → checks DSC conditions (Available, Degraded, odh-dashboardReady)
+#                → verifies OdhDashboardConfig exists with readable spec
+#                → verifies test admin user can create projects
```
antowaddle, line 16: The tests already check this, what's the purpose of only specifically checking that projects can be created?

.github/workflows/cypress-e2e-test.yml
```diff
@@ -174,9 +180,47 @@ jobs:
             echo "  📋 DSC Conditions:"
             echo "$DSC_JSON" | jq -r '.items[0].status.conditions[]? | "     - \(.type): \(.status) (\(.reason // "no reason"))"' 2>/dev/null || echo "     (no conditions found)"
             
+            # Preflight checks beyond DSC Ready — catches missing CRs and RBAC gaps
```
antowaddle, line 183: Have you examples where this has failed? The tests should already be testing for this stuff, I'm trying to understand the genesis of this change.

antowaddle (changes requested):
Thanks @Taj010 for this contribution, I've added some comments. 

Have you executed this workflow? Note that it currently only triggers from main, so in order to validate this one way (there are others) would be to create the PR directly against main (not from your branch).

Can you also elaborate/provide some examples of why we'd need this change? Would it make more sense to actually build e2e tests that do this stuff and run them first? We could then run them everywhere as sort of a health check on the cluster.

Taj010:
@antowaddle, I didn't know some of the checks were already in place. I thought that adding a project check would help with the RBAC issues. I think more research would be needed for better help/change. I will close this for now and come back if it's really needed later. Thank you so much for taking the time to check and explain this.

PR #8133, opened 2026-06-16
feat(operator): dashboard module controller — migration, TLS, status, docs, param injection

lucferbux:
https://issues.redhat.com/browse/RHOAIENG-59938
https://issues.redhat.com/browse/RHOAIENG-59939
https://issues.redhat.com/browse/RHOAIENG-59941
https://issues.redhat.com/browse/RHOAIENG-61013

## Description

Completes the dashboard-operator's readiness for zero-downtime migration from the old ODH operator. This PR covers two commits:

**Commit 1 — Infrastructure ([RHOAIENG-61013](https://redhat.atlassian.net/browse/RHOAIENG-61013), [RHOAIENG-59938](https://redhat.atlassian.net/browse/RHOAIENG-59938), [RHOAIENG-59939](https://redhat.atlassian.net/browse/RHOAIENG-59939), [RHOAIENG-59941](https://redhat.atlassian.net/browse/RHOAIENG-59941)):**
- Configure cert-manager metrics TLS with self-signed certificates
- Implement cross-namespace resource cleanup on CR deletion (finalizer-based GC for Perses resources in observability namespace)
- Add status contract tests (Ready/ProvisioningSucceeded/Degraded conditions, observedGeneration, phase transitions) and adoption idempotency tests
- Add architecture documentation covering zero-downtime migration, status aggregation, TLS, and prerequisites

**Commit 2 — Parameter injection pipeline ([RHOAIENG-59938](https://redhat.atlassian.net/browse/RHOAIENG-59938)):**
Fixes 4 gaps that prevented the operator from fully replacing the old ODH operator's parameter behavior:
1. **modular-architecture params.env never written** — BFF module images now get resolved from `RELATED_IMAGE_*` env vars instead of hardcoded `:main` defaults
2. **Missing image mappings** — Added `automl-pipeline-runtime-image` and `autorag-pipeline-runtime-image` to `imagesMap`
3. **Wrong dashboard-url format** — Changed from `https://odh-dashboard-{domain}` to `https://{domain}/` matching gateway domain semantics
4. **URL extraction only checked Routes** — Now derives URL from `spec.gateway.domain` (matching old operator's two-tier gateway pipeline) with Route lookup as fallback

## How Has This Been Tested?

**Unit tests:** All pass with 89.2% coverage on `internal/controller` — covers kustomize param computation, image resolution, modular-architecture params.env writing, gateway-domain URL extraction with Route fallback, status conditions, and cross-namespace cleanup.

**Live cluster validation (ROSA `monarch-ui-2`):**
- Built and deployed the operator image to the cluster's internal registry
- Patched Dashboard CR with `spec.gateway.domain: rh-ai.apps.rosa.monarch-ui-2.6xbt.p3.openshiftapps.com`
- Verified `odh-dashboard-params` ConfigMap: `dashboard-url=https://rh-ai.apps.rosa.monarch-ui-2.6xbt.p3.openshiftapps.com/` (correct format)
- Verified `modular-architecture-params` ConfigMap: all 9 BFF image keys populated
- Verified `odhlink` ConsoleLink: `href` matches the correct gateway URL
- Verified `Dashboard.Status.URL`: correctly derived from gateway domain (no Route requeue loop)
- Verified `Dashboard.Status.Phase: Ready` with all conditions healthy
- Verified operator logs: clean reconciliation, no errors, correct URL logged

## Test Impact

- Updated `TestComputeKustomizeVariables` — corrected expected `dashboard-url` format
- Updated `TestResolveImageParams` — added assertions for 2 new pipeline runtime image mappings
- Updated `TestApplyKustomizeParams` — added modular-architecture params.env verification, `RELATED_IMAGE` override testing
- Updated `TestExtractDashboardURL` — added gateway-domain-takes-priority test case, updated all cases with Dashboard object parameter
- Added `TestReconcile_StatusContract` — validates Ready/ProvisioningSucceeded/Degraded conditions
- Added `TestReconcile_AdoptionIdempotency` — validates re-reconciliation of existing resources

## Request review criteria:

## Release Notes

* **New Features**
  * Metrics endpoints now support HTTPS with TLS (toggle via a new `secure-metrics` flag; metrics port updated to **8443**).

* **Improvements**
  * Dashboard URL generation now honors the gateway domain (`https://<domain>/` form).
  * Improved modular-architecture parameter writing, including gateway-related values.
  * Expanded related UI image support (Automl, AutoRAG, Eval-Hub and related runtime images).
  * Enhanced deletion cleanup for cross-namespace observability resources.

* **Documentation**
  * Updated guidance for adoption, cross-namespace cleanup, status semantics, and TLS/deployment prerequisites.


dashboard-operator/cmd/manager/main.go
```diff
@@ -70,9 +73,23 @@ func main() {
 		applicationsNamespace = operatorNamespace
 	}
 
+	// Disable HTTP/2 to mitigate CVE-2023-44487 (rapid reset attack).
+	tlsOpts := []func(*tls.Config){
+		func(c *tls.Config) {
+			c.NextProtos = []string{"http/1.1"}
```
liday-rh: nit (security): The `tls.Config` callback sets `NextProtos` to disable HTTP/2, but doesn't explicitly set `MinVersion`. Go 1.18+ defaults to TLS 1.2, but setting it explicitly is a defense-in-depth best practice — it makes the intent clear and guards against hypothetical future default changes.

```suggestion
	tlsOpts := []func(*tls.Config){
		func(c *tls.Config) {
			c.NextProtos = []string{"http/1.1"}
			c.MinVersion = tls.VersionTLS12
		},
	}
```

dashboard-operator/cmd/manager/main.go
```diff
@@ -70,9 +73,23 @@ func main() {
...
+		},
```
liday-rh: nit: `tlsOpts` is allocated unconditionally but only used inside the `if secureMetrics` block. Consider moving the declaration inside the conditional to make the scope clearer and avoid the dead allocation when `--secure-metrics=false`.

dashboard-operator/config/webhook/manifests.yaml
```diff
@@ -60,3 +60,31 @@ spec:
...
+  name: dashboard-operator-metrics-cert
+  namespace: system
+spec:
+  dnsNames:
+    - dashboard-operator-metrics-service.system.svc
+    - dashboard-operator-metrics-service.system.svc.cluster.local
+  issuerRef:
+    kind: Issuer
+    name: selfsigned-issuer
+  secretName: dashboard-operator-metrics-tls
```
liday-rh, line 90: nit: The new `Service` and `Certificate` for the **metrics** endpoint are appended to a file named `webhook/manifests.yaml`. This makes the filename misleading — a reader would expect only webhook-related resources here. Consider a dedicated `config/metrics/` directory (mirroring how `config/webhook/` is structured), or at minimum renaming the file to reflect its broader scope.

dashboard-operator/internal/controller/dashboard_reconciler_test.go
```diff
@@ -368,6 +368,335 @@ func TestReconcile_Deletion(t *testing.T) {
...
+			DeletionTimestamp: &metav1.Time{
+				Time: time.Now(),
+			},
+		},
+		Spec: v1alpha1.DashboardSpec{
+			Observability: &v1alpha1.ObservabilitySpec{
+				Enabled: true,
+				PersesService: &v1alpha1.ServiceTarget{
+					Name:      "perses",
+					Namespace: obsNS,
```
liday-rh, line 392: nit: `wantDegraded` is defined in the test struct and set in all three cases, but no assertion in the test body ever checks the `Degraded` condition. This is dead code that could mislead a reader into thinking the condition is being validated. Either add the corresponding assertion or remove the field.

dashboard-operator/internal/controller/dashboard_reconciler_test.go
```diff
@@ -368,6 +368,335 @@ func TestReconcile_Deletion(t *testing.T) {
...
+		ManifestsBasePath:     t.TempDir(),
+		Platform:              cluster.OpenDataHub,
+		Namespace:             testNamespace,
+		ApplicationsNamespace: testNamespace,
+	}
+
+	result, err := r.Reconcile(context.Background(), ctrl.Request{
+		NamespacedName: types.NamespacedName{Name: v1alpha1.DashboardInstanceName},
+	})
+
```
liday-rh, line 432: nit: Both the result and error from `Reconcile` are silently discarded. For the success case ("all conditions present on success"), the test should assert `require.NoError` to catch unexpected reconciliation failures. For the intentional failure case ("conditions reflect render failure"), asserting that the error *is* returned (or documenting why it's nil despite a render failure) would strengthen confidence in the test.

liday-rh (commented):
Overall this is a solid PR — the security hardening (metrics TLS, CVE-2023-44487 mitigation), cross-namespace cleanup logic, and test coverage are all well done. Left a few low-severity nits on TLS config, file organization, and test completeness. Nothing blocking.

dashboard-operator/cmd/manager/main.go
lucferbux: Good call — addressed in 69877e6. Added `c.MinVersion = tls.VersionTLS12` and also moved the entire `tlsOpts` declaration inside the `if secureMetrics` block per your other comment, so the scope is tighter.

dashboard-operator/cmd/manager/main.go
lucferbux: Done in 69877e6 — moved inside the conditional.

dashboard-operator/config/webhook/manifests.yaml
lucferbux: This follows the kubebuilder convention — `config/webhook/manifests.yaml` is the standard catch-all for cert-manager-managed server manifests (webhook + metrics). Creating a separate `config/metrics/` directory would diverge from the kubebuilder scaffold without clear benefit, and the `config/default/kustomization.yaml` already references this file path. Keeping it as-is for now; can revisit if the file grows further.

dashboard-operator/internal/controller/dashboard_reconciler_test.go
lucferbux: Good catch — removed `wantDegraded` and added `wantErr` with proper assertions in 69877e6.

dashboard-operator/internal/controller/dashboard_reconciler_test.go
lucferbux: Addressed in 69877e6 — added `wantErr` field to the test struct. The success case now asserts `require.NoError`, and the render failure case asserts `require.Error`.

dashboard-operator/internal/controller/actions.go
```diff
@@ -38,14 +39,27 @@ func applyKustomizeParams(dashboard *v1alpha1.Dashboard, manifests []render.Mani
...
 	}
 
+	if len(manifests) > 0 {
+		modArchPath := filepath.Join(manifests[0].Path, "modular-architecture")
+		params := readExistingParams(modArchPath + "/params.env")
+		maps.Copy(params, computed)
+		if err := writeParamsEnv(modArchPath, params); err != nil {
+			return fmt.Errorf("failed to write params.env to %s: %w", modArchPath, err)
+		}
+	}
```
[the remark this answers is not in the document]
lucferbux: Good suggestion — added the `os.Stat` existence check in 69877e6.

dashboard-operator/internal/controller/dashboard_reconciler_test.go
```diff
@@ -368,6 +372,335 @@ func TestReconcile_Deletion(t *testing.T) {
...
+	assert.Equal(t, ctrl.Result{}, result)
+
+	var svcs corev1.ServiceList
+	require.NoError(t, cli.List(context.Background(), &svcs, client.InNamespace(obsNS)))
+	assert.Empty(t, svcs.Items, "cross-namespace services should be deleted")
+
+	var cms corev1.ConfigMapList
+	require.NoError(t, cli.List(context.Background(), &cms, client.InNamespace(obsNS)))
+	assert.Empty(t, cms.Items, "cross-namespace configmaps should be deleted")
+}
```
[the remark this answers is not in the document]
lucferbux: The cleanup uses label-scoped listing (`platform.opendatahub.io/part-of: dashboard`) — it can only delete resources the controller itself deployed. The Dashboard CRD is cluster-scoped and protected by RBAC; write access already implies cluster-level privileges. A hardcoded namespace allowlist would be brittle — the observability namespace is configurable by design. RBAC on the CRD is the security boundary here.

dashboard-operator/internal/controller/dashboard_reconciler.go
```diff
@@ -241,11 +242,52 @@ func (r *DashboardReconciler) reconcile(
...
+	obsNS := ""
+	if dashboard.Spec.Observability != nil &&
+		dashboard.Spec.Observability.PersesService != nil {
+		obsNS = dashboard.Spec.Observability.PersesService.Namespace
+	}
+
+	if obsNS == "" || obsNS == r.ApplicationsNamespace {
+		logger.Info("No cross-namespace resources to clean up")
+		return nil
+	}
```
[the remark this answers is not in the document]
lucferbux: Same reasoning as above — the cleanup is label-scoped, not namespace-scoped. The controller only deletes resources bearing its own `part-of: dashboard` label, so a rogue namespace value cannot widen the blast radius beyond what the controller itself deployed. CRD RBAC is the access boundary.

dashboard-operator/internal/controller/support.go
```diff
@@ -60,7 +62,7 @@ func computeKustomizeVariables(dashboard *v1alpha1.Dashboard, platform cluster.P
 
 	if dashboard.Spec.Gateway != nil && dashboard.Spec.Gateway.Domain != "" {
 		params["gateway-domain"] = dashboard.Spec.Gateway.Domain
-		params["dashboard-url"] = fmt.Sprintf("https://odh-dashboard-%s", dashboard.Spec.Gateway.Domain)
+		params["dashboard-url"] = fmt.Sprintf("https://%s/", dashboard.Spec.Gateway.Domain)
 	}
```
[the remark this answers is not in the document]
lucferbux: The domain comes from the platform operator, which reads `Gateway.Spec.Listeners[0].Hostname` — a controller-managed, trusted source. CRD-level validation (kubebuilder markers) is the appropriate layer for input constraints, and that belongs in the component-onboarding PR where the field is defined. Silently skipping an invalid domain here would be worse than propagating it — the user would see no URL with no explanation.

docs/dashboard-operator.md
```diff
@@ -206,3 +206,121 @@ The operator controller is distinct from BFF (Backend-for-Frontend) services:
...
+### Command-Line Flags
+
+| Flag | Default | Purpose |
+|------|---------|---------|
+| `--manifests-base-path` | `/opt/manifests/dashboard` | Base path for dashboard manifests |
+| `--metrics-bind-address` | `:8443` | Metrics endpoint bind address |
+| `--health-probe-bind-address` | `:8081` | Health probe bind address |
+| `--leader-elect` | `false` | Enable leader election |
+| `--secure-metrics` | `true` | Serve metrics over HTTPS |
+| `--namespace` | (required) | Operator deployment namespace |
```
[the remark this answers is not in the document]
lucferbux: Fixed in 69877e6 — updated the default column to "(env fallback)" with the description noting OPERATOR_NAMESPACE.

lucferbux:
Dang it 
/lgtm
/approve

@antowaddle @christianvogt can you guys approve this, I need to take a look why liam is not getting caught by the prow

antowaddle (approved):
/lgtm

PR #7855, opened 2026-06-04
feat: extract shared K8s types used by model-serving into a common types package k8s-core

paulovmr:
https://redhat.atlassian.net/browse/RHOAIENG-65078

## Description

Creates a new `@odh-dashboard/k8s-core` package containing 10 K8s resource types (`ProjectKind`, `SecretKind`, `PodKind`, `DashboardConfigKind`, etc.) plus their ~20 transitive dependencies, extracted from the monolithic `frontend/src/k8sTypes.ts` and `frontend/src/types.ts`. The original definitions are replaced with re-exports so the
  500+ frontend files need no changes, while 91 files across 10 packages (`model-serving`, `kserve`, `llmd-serving`, `nim-serving`, `model-training`, `cypress`, `feature-store`, `plugin-core`, `autorag`, `gen-ai`) now import directly from `@odh-dashboard/k8s-core` instead of `@odh-dashboard/internal/k8sTypes`. This is the first step toward decoupling model-serving from  `@odh-dashboard/internal`, and the package is named `k8s-core` (rather than `k8s-types`) to accommodate the planned addition of shared K8s utility functions in a follow-up task.

## How Has This Been Tested?

- `npx turbo lint type-check --force` — 47/47 tasks pass, 0 errors across all packages
- `npx jest in frontend/` — 313 suites, 3,007 tests pass
- `npx jest in packages/model-serving/` — 20 suites, 209 tests pass
- No behavioral changes — pure type extraction with re-exports preserving all existing import paths
 
## Test Impact

No tests were added, because there are no behavioral changes — pure type extraction with re-exports preserving all existing import paths

## Request review criteria:

Self checklist (all need to be checked):
- [X] The developer has manually tested the changes and verified that the changes work
- [X] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [X] The developer has added tests or explained why testing cannot be added (unit or cypress tests for related changes)
- [X] The code follows our [Best Practices](/docs/best-practices.md) (React coding standards, PatternFly usage, performance considerations)

After the PR is posted & before it merges:
- [ ] The developer has tested their solution on a cluster by using the image produced by the PR to `main`

## Summary by CodeRabbit

* **Refactor**
  * Centralized Kubernetes/ODH enums and types into a shared package and re-exported them for compatibility.
* **Chores**
  * Included the shared package in the UI build context for workspace builds.
* **Tests / Mocks**
  * Updated test and mock imports to use the shared package; no test behavior changed.
* **User Impact**
[...]

PR #8183, opened 2026-06-17
feat: extract shared UI table components into a common package ui-core

andyatmiami:
https://issues.redhat.com/browse/RHOAIENG-65081

## Description
Creates a new `@odh-dashboard/ui-core` package containing 8 shared UI components (`Table`, `TableBase`, `ResourceTr`, `ResourceActionsColumn`, `ResourceNameTooltip`, `DashboardPopupIconButton`, `DashboardEmptyTableView`, `StateActionToggle`) plus their ~7 transitive dependencies (`SortableData`, `GetColumnSort`, `useTableColumnSort`, `getTableColumnSort`, table const helpers, `EitherNotBoth`, `MIN_PAGE_SIZE`), extracted from `frontend/src/components/` and `frontend/src/concepts/dashboard/`. The original definitions are deleted with no re-export compatibility layer — all ~200 consumer files across the frontend and 8 packages (`model-serving`, `maas`, `feature-store`, `eval-hub`, `model-training`, `automl`, `autorag`, `mlflow`) now import directly from `@odh-dashboard/ui-core`. This is the second step toward decoupling feature packages from `@odh-dashboard/internal`, following the k8s-core extraction in #7855.

## How Has This Been Tested?
- `npx turbo lint type-check` — any errors reported were pre-existing / reproducible on main 
- `npx jest in frontend/` — all tests pass
- `npx jest in packages/model-serving/` — all tests pass
- No behavioral changes — pure extraction and refactor to update import path

## Test Impact
No tests were added, because there are no behavioral changes — pure extraction and refactor to update import path

## Request review criteria:

If you have UI changes: 


frontend/config/moduleFederation.js
```diff
@@ -84,7 +84,7 @@ const workspacePackages = getWorkspacePackages();
 const odhDashboardShared = Object.fromEntries(
   [...getRuntimeOdhPackages(workspacePackages)].map((name) => [
     name,
-    { singleton: true, requiredVersion: '*', eager: true },
+    { singleton: true, requiredVersion: '*' },
```
andyatmiami, line 87: 💡 Please pay particular attention to this seemingly innocuous change that would be easy to overlook!

In discussing with @christianvogt - we came to conclusion we should **NOT** need to load these packages being "auto-detected" as `eager: true` - and doing so was causing some MF resolution issues with the introduction of `ui-core`..

So I have now flipped this behavior to **NOT** eagerly load - and subsequent testing did not uncover any regressions with this approach.

packages/ui-core/src/types.ts
andyatmiami, line 1: The EitherNotBoth type utility is duplicated here from frontend/src/typeHelpers.ts because ui-core can't import it from either the frontend (would create a circular dependency) or @openshift/dynamic-plugin-sdk (which also defines it, but is too heavy a dependency for a lightweight UI package). It's 3 lines of pure type code with no runtime cost — acceptable for now. A future follow-up could consolidate shared type utilities into a common package

paulovmr: I've created [RHOAIENG-70804](https://redhat.atlassian.net/browse/RHOAIENG-70804) as a follow-up for this.

paulovmr (approved):
/lgtm

lucferbux (commented):
## Review: `ui-core` package extraction

Nice work on the extraction — the package structure follows the established `k8s-core` pattern, all ~200 consumer import paths are correctly updated (including default → named export conversions), Dockerfiles are covered, and the dependency graph is clean (`ui-core` → `k8s-core`, no cycles).

I checked out the branch locally and ran the full build pipeline:

| Check | Result |
|---|---|
| `npm install` | :white_check_mark: |
| `npx turbo type-check` | :white_check_mark: (eval-hub failures are pre-existing on main — `mod-arch-core` module) |
| `npx turbo lint` | :white_check_mark: (mlflow `ban-types` errors are pre-existing on main) |
| `npx turbo build` (webpack) | :white_check_mark: — compiles successfully, only pre-existing size warnings |
| Frontend Jest (325 suites / 3161 tests) | :white_check_mark: |

**No build breakage from this PR.**

---

### 1. Missing `ui-core` in 3 federated package Module Federation configs

Three federated packages do **not** have `@odh-dashboard/ui-core` in their `moduleFederation.js` shared singletons:

| Package | Config path |
|---|---|
| `gen-ai` | `packages/gen-ai/frontend/config/moduleFederation.js` |
| `model-registry` | `packages/model-registry/upstream/frontend/config/moduleFederation.js` |
| `notebooks` | `packages/notebooks/upstream/workspaces/frontend/config/moduleFederation.js` |

The five other federated packages (automl, autorag, eval-hub, maas, mlflow) all have it. These three were likely missed because their MF configs live at non-standard paths (`upstream/` subdirectories).

**Current risk:** None — these packages don't import from `@odh-dashboard/ui-core` today.

**Future risk:** If a developer adds a `ui-core` import to any of these packages without also updating the MF shared config, they'll get a **duplicate module instance** instead of the shared singleton. This causes subtle bugs (React context mismatches, duplicate state, instanceof checks failing across boundaries).

**Suggestion:** Add `'@odh-dashboard/ui-core': { singleton: true, requiredVersion: '*' }` to all three proactively, matching the pattern already applied to the other five.

### 2. `eager: true` removal is a separate behavioral change

The second commit (`cd44cb3`) removes `eager: true` from `odhDashboardShared` in `frontend/config/moduleFederation.js`. This changes **all** `@odh-dashboard/*` shared packages from eager to lazy loading — not just `ui-core`.

**Is it safe?** Yes — the host already uses the async bootstrap pattern (`import('./bootstrap')` in `index.tsx`), which creates the async boundary needed for shared scope negotiation. With that in place, `eager` is redundant, and removing it is actually a minor optimization (avoids potential duplicate bundles when both host and remote eagerly include the same module).

**Concern:** This is orthogonal to the `ui-core` extraction. It affects the loading behavior of every `@odh-dashboard/*` shared package. Bundling it here makes it easy to miss during review. Consider whether this warrants its own commit message that explains the rationale (async bootstrap makes `eager` unnecessary), or even a separate PR so it gets its own CI signal.

---

### Nits

**3. `ResourceNameTooltip` lost an SCSS side-effect import**

The original file imported `#~/pages/notebookController/NotebookController.scss` as a side effect. This was dropped during the move to `ui-core`. The SCSS defines `.odh-notebook-controller` styles that are unrelated to `ResourceNameTooltip` — it was a misplaced dependency, and the styles are still imported by 7 other components in the NotebookController page tree. Not a regression, but worth a quick visual spot-check on the Notebook Controller page.

**4. Inconsistent `import type` vs `import` for `ToggleState`**

`ToggleState` is exported as `export type { ToggleState }` from `ui-core/src/index.ts`, but two model-serving files use value imports instead of type imports:

- `packages/model-serving/src/components/deployments/DeploymentStatus.tsx`
- `packages/model-serving/src/concepts/utils.ts`

Works fine, but inconsistent with the rest of the codebase where `import type` is used.

**5. `EitherNotBoth` type duplication**

`EitherNotBoth` is now defined in both `packages/ui-core/src/types.ts` and `frontend/src/typeHelpers.ts`. Some consumers import from `@odh-dashboard/ui-core`, others still use `#~/typeHelpers`. Functional, but worth tracking for eventual consolidation in a follow-up.

lucferbux:
/lgtm
/approve

@andyatmiami we can add the nits in a follow up PR, approving this so we can continue the implementation

PR #7958, opened 2026-06-10
chore: update minor/patch deps, bump PF to 6.5

christianvogt:
https://redhat.atlassian.net/browse/RHOAIENG-63779
https://redhat.atlassian.net/browse/RHOAIENG-63599

## Description

1. **Bump minor & patch dependencies** — updates across the monorepo (TypeScript 5.9, axios 1.17, fastify 4.29, react-router 7.17, and many others).

2. **Update PatternFly to 6.5** — bumps all `@patternfly/*` packages to 6.5.x.

3. **Fix `plugin-core` ↔ `@odh-dashboard/internal` circular dependency** — introduces `types-core.ts` and `k8sTypes-core.ts` as leaf type modules with zero transitive imports into the frontend source tree. `plugin-core` now imports from these instead of pulling the full frontend dependency graph. Replaces `allSettledPromises` with `Promise.all` in `useResolvedExtensions`. Moves `isInEnum` to `utilities/utils.ts` and `DataScienceStackComponent` to `types-core.ts`. Re-enables the `import/no-extraneous-dependencies` lint rule in `plugin-core`.

   NOTE: Can refactor this slightly when #7855 lands.

5. `getRuntimeOdhPackages` no longer needs a `hostPackageJson` parameter — it auto-discovers the frontend from workspace query results.

**Additional fixes:**
- **eslint-config**: add `@rushstack/eslint-patch` for proper module resolution; move plugins to `dependencies` so consuming packages resolve them correctly.
- **Package `peerDependencies`**: add proper peer dep declarations across feature-store, kserve, model-serving, model-training, nim-serving, jest-config, and others.
- **Backend**: remove unused `prom-client`; make implicit devDependencies explicit; move jest deps from `optionalDependencies` to `devDependencies`.
- **CI**: add agent-ops `node_modules` cache step.
- **Cypress stability**: add per-run intercepts for pipeline compare-runs tests; use `verifyRelativeURL` for deterministic URL checks; use `selectGlobalScopedTemplateOption` instead of chained `.should('exist').click()` in model serving deploy wizard; tighten `waitSnapshot` alias type to `` `@${string}` ``.

## How Has This Been Tested?

- `npm install` succeeds and `package-lock.json` is consistent
- Verified the type-only changes compile: `plugin-core` no longer triggers the circular-dep lint error
- Cypress mock tests updated to match the new wizard step ordering

## Test Impact

[...]

PR #8186, opened 2026-06-17
fix: missing leaderboard headers in AutoML and AutoRAG

daniduong:
https://redhat.atlassian.net/browse/RHOAIENG-69430

## Description

- PatternFly was recently bumped to 6.5 in the root dashboard (#7958), but automl, autorag, and gen-ai were still on PF 6.3
- only `@patternfly/react-core` is configured as a shared singleton with module federation — other PF packages (`react-table`, `react-styles`, `react-icons`, etc.) are not shared, so each remote bundles its own copy
- whenever we use PF components, they pull in per-component CSS through `@patternfly/react-styles` (e.g., `table.css`, `button.css`), resulting in duplicate stylesheets in the DOM — one from the host's PF version and one from the remote's
- on the AutoML leaderboard page specifically, the older PF 6.3 styles from automl happened to take precedence over the host's 6.5 styles, which caused a `:before` pseudo-element to block table headers due to a z-index mismatch

This leaderboard issue actually touches on a more systemic problem:

- navigating between different pages in the dashboard in the same tab causes multiple CSS stylesheets to load and conflict with each other, and these persist until the cache is cleared
- the longer-term fix would be to add the other `@patternfly/*` packages (`react-table`, `react-styles`, `react-icons`, etc.) to the module federation config as shared singletons so only one copy of each package and its CSS is loaded
- the fix in this PR is just aligning the PF versions to match the host, which removes the conflicts — duplicate CSS will still be loaded, but since they're all identical now, it doesn't matter
- to limit the scope of this PR, the longer-term fix will be addressed in a future PR

## How Has This Been Tested?

Manually verified the AutoML leaderboard page — table headers are visible and correctly styled after the version bump.

Before | After
--- | ---
image | image

## Test Impact

No test changes — this is a dependency version alignment with no application logic modified.

## Request review criteria:

## Summary by CodeRabbit

* **Chores**
  * Upgraded PatternFly UI framework dependencies across Automl, Autorag, and GenAI frontends (core, icons, tables, templates, styles, topology, and chatbot/code editor where applicable) to the latest stable 6.5.x/6.6.x releases.
  * This refresh updates the React bindings and supporting UI components used throughout the applications.


daniduong:
@coderabbitai review

daniduong:
@coderabbitai review

daniduong:
closing in favour of https://github.com/opendatahub-io/odh-dashboard/pull/8213

PR #8149, opened 2026-06-16
RHOAIENG-63156: Implement in-code template catalog and Select Template modal

srtanish1992:
https://redhat.atlassian.net/browse/RHOAIENG-63156

## Description

Implements an in-code role template catalog and "Select Template" modal for the Create Custom Role form.

https://github.com/user-attachments/assets/e4addd36-738d-417a-aa88-a502bfc66628

**What this adds:**
- **Role template catalog** (`roleTemplateCatalog.ts`) — a typed, extensible data structure (`RoleTemplate` / `RoleTemplateCategory`) containing three Workbench management templates: Maintainer, Reader, and Updater.
- **Select Template modal** (`SelectTemplateModal.tsx`) — supports two modes:
  - **"Select role template"** — resets the form and populates name, description, and rules from the chosen template.
  - **"Add rules from template"** — appends the template's rules to the existing form without overwriting other fields.
- **Discard Changes confirmation** (`DiscardChangesConfirmModal.tsx`) — warns users before overwriting dirty form data when selecting a template.
- **Search/filter** — templates are filterable by name via a `SearchInput`.
- **Cypress mock tests** (`rolesTabTemplate.cy.ts`) — full coverage of template selection, discard confirmation, add-rules mode, and search filtering.
- **Unit tests** — for `SelectTemplateModal`, `DiscardChangesConfirmModal`, and `roleTemplateCatalog` data integrity.

**Template permissions matrix:**
| Resource | API Group | Maintainer | Reader | Updater |
|---|---|---|---|---|
| `notebooks` | `kubeflow.org` | `*` | get, list, watch | get, watch, list, update, patch |
| `persistentvolumeclaims` | _(core)_ | `*` | get, list, watch | `*` |
| `secrets` | _(core)_ | `*` | get, list, watch | `*` |
| `configmaps` | _(core)_ | `*` | get, list, watch | `*` |
| `namespaces` | _(core)_ | get, watch, list | get, list, watch | get, watch, list |
| `persistentvolumeclaims/status` | _(core)_ | get, watch, list | get, list, watch | get, watch, list |
| `pods` | _(core)_ | get, watch, list | get, list, watch | get, watch, list |
| `events` | _(core)_ | get, watch, list | get, list, watch | get, watch, list |
| `statefulsets` | `apps` | get, watch, list | get, list, watch | get, watch, list |
| `imagestreams` | `image.openshift.io` | get, watch, list | get, list, watch | get, watch, list |
[...]

PR #8197, opened 2026-06-17
RHOAIENG-63157: Implement Form/YAML toggle and YAML preview

srtanish1992:
https://redhat.atlassian.net/browse/RHOAIENG-63157

## Description

Implements the Form/YAML toggle on the Create Custom Role page.

https://github.com/user-attachments/assets/531a688e-8cc6-4661-b55f-0806a024858a

This adds a `ToggleGroup` in the page header (next to "Select role template") that allows users to switch between the interactive form view and a read-only YAML preview showing the Kubernetes Role manifest generated from the current form state.

**Key behaviors:**
- "Form" is selected by default, showing the existing interactive form
- Switching to "YAML (read-only)" displays a live YAML preview of the Role manifest
- YAML reflects all form state in real-time: metadata (name, labels, annotations) and rules
- Switching back to "Form" preserves all entered data
- The YAML view is strictly read-only — no editing capability
- The footer (Create role / Cancel) remains visible in both views

**YAML view includes:**
- A title ("Role configuration YAML") and description matching the form view's layout pattern
- A `DashboardCodeEditor` with copy, download, and full-screen toolbar actions
- A "YAML (read-only)" label in the toolbar and the language label on the right

## How Has This Been Tested?

- Verified toggle renders and switches between Form and YAML views
- Confirmed YAML output matches expected Kubernetes Role manifest structure
- Verified form data (name, description, labels, rules) is preserved across toggle switches
- Verified toolbar actions (copy, download, full-screen) are present and functional
- Confirmed footer remains visible in both views
- Ran ESLint with zero errors across all modified/new files
- Ran all projectRoles unit tests (47 tests passing)

## Test Impact

**Unit tests added:**
- `CreateRoleYamlView.spec.tsx` — 7 tests covering:
  - Title and description rendering
  - Valid Kubernetes Role YAML output
  - Rules reflected in YAML
  - Labels reflected in YAML
  - Empty name behavior
  - Full-screen toggle button rendered
  - Labels with empty keys filtered out
  
**Cypress mock tests added:**
- `rolesTabYamlToggle.cy.ts` — 9 tests covering:
  - Toggle group visibility
  - Form selected by default
  - Switching to YAML view
  - Title and description in YAML view
  - Code editor presence
  - Valid Role manifest content
  - Description reflected in YAML
  - Data preservation on toggle back to Form
  - Footer visible in YAML view
  - Full-screen toggle button presence

## Request review criteria:

If you have UI changes: 

## Summary

* **New Features**
  * Added Form/YAML switching for creating roles, with YAML shown as a live, read-only preview.
  * YAML preview includes copy/download and fullscreen support, updating as role details change.
  * Submitting role definitions now omits any labels with empty keys.
* **Tests**
  * Added Jest coverage validating generated Role YAML (structure, required dashboard label/annotations, rules/labels serialization, and empty-key filtering).
  * Added Cypress coverage for toggle behavior, YAML rendering, fullscreen control, and input preservation when switching views.

### DIFF
--- a/frontend/src/pages/projects/projectRoles/CreateRolePage.tsx
+++ b/frontend/src/pages/projects/projectRoles/CreateRolePage.tsx
@@ -108,7 +108,9 @@ const CreateRolePage: React.FC<CreateRolePageProps> = ({ existingRole }) => {
     setSubmitError(undefined);
     const k8sName = k8sNameDescriptionData.data.k8sName.value;
     const roleDisplayName = k8sNameDescriptionData.data.name || k8sName;
-    const labelRecord = Object.fromEntries(labels.map((l) => [l.key, l.value]));
+    const labelRecord = Object.fromEntries(
+      labels.filter((l) => l.key).map((l) => [l.key, l.value]),
+    );
     const role = assembleRole(namespace, k8sName, roleDisplayName, description, rules, labelRecord);
     try {
       await createRole(role);
@@ -195,7 +197,7 @@ const CreateRolePage: React.FC<CreateRolePageProps> = ({ existingRole }) => {
               onRulesChange={handleRulesChange}
             />
           </div>
-          <div hidden={viewMode !== 'yaml'}>
+          {viewMode === 'yaml' && (
             <CreateRoleYamlView
               namespace={namespace}
               k8sName={k8sNameDescriptionData.data.k8sName.value}
@@ -206,7 +208,7 @@ const CreateRolePage: React.FC<CreateRolePageProps> = ({ existingRole }) => {
               rules={rules}
               labels={labels}
             />
-          </div>
+          )}
         </PageSection>
         <PageSection hasBodyWrapper={false} stickyOnBreakpoint={{ default: 'bottom' }}>
           <CreateRoleFooter
--- a/frontend/src/pages/projects/projectRoles/CreateRolePage.tsx
+++ b/frontend/src/pages/projects/projectRoles/CreateRolePage.tsx
@@ -4,8 +4,11 @@ import {
   BreadcrumbItem,
   Bullseye,
   Button,
+  Flex,
   PageSection,
   Spinner,
+  ToggleGroup,
+  ToggleGroupItem,
   getUniqueId,
 } from '@patternfly/react-core';
 import { Link, Navigate, useNavigate, useParams } from 'react-router-dom';
@@ -19,8 +22,11 @@ import { createRole } from '#~/api';
 import CreateRoleForm from './CreateRoleForm';
 import CreateRoleFooter from './CreateRoleFooter';
 import CreateRoleConfirmModal from './CreateRoleConfirmModal';
+import CreateRoleYamlView from './CreateRoleYamlView';
 import assembleRole from './assembleRole';
 import type { LabelEntry, RuleEntry } from './types';
+
+type ViewMode = 'form' | 'yaml';
 
 type CreateRolePageProps = {
   existingRole?: RoleKind;
@@ -66,6 +72,7 @@ const CreateRolePage: React.FC<CreateRolePageProps> = ({ existingRole }) => {
     }));
   });
 
+  const [viewMode, setViewMode] = React.useState<ViewMode>('form');
   const [submitError, setSubmitError] = React.useState<Error>();
   const [showNoRulesConfirm, setShowNoRulesConfirm] = React.useState(false);
 
@@ -150,24 +157,55 @@ const CreateRolePage: React.FC<CreateRolePageProps> = ({ existingRole }) => {
         }
         description="Create a custom role to control what users can see and do across your cluster resources. Define permissions, navigation access, and resource scopes to implement fine-grained access control."
         headerAction={
-          // TODO: Enable when template selection modal is implemented (RHOAIENG-63156)
-          <Button variant="secondary" data-testid="select-role-template-button" isDisabled>
-            Select role template
-          </Button>
+          <Flex gap={{ default: 'gapMd' }} alignItems={{ default: 'alignItemsCenter' }}>
+            {/* TODO: Enable when template selection modal is implemented (RHOAIENG-63156) */}
+            <Button variant="secondary" data-testid="select-role-template-button" isDisabled>
+              Select role template
+            </Button>
+            <ToggleGroup aria-label="Form or YAML view toggle" data-testid="form-yaml-toggle">
+              <ToggleGroupItem
+                text="Form"
+                buttonId="form-view-toggle"
+                data-testid="form-view-toggle"
+                isSelected={viewMode === 'form'}
+                onChange={() => setViewMode('form')}
+              />
+              <ToggleGroupItem
+                text="YAML (read-only)"
+                buttonId="yaml-view-toggle"
+                data-testid="yaml-view-toggle"
+                isSelected={viewMode === 'yaml'}
+                onChange={() => setViewMode('yaml')}
+              />
+            </ToggleGroup>
+          </Flex>
         }
         loaded
         empty={false}
       >
         <PageSection hasBodyWrapper={false} isFilled data-testid="create-role-page">
-          <CreateRoleForm
-            nameDescriptionData={k8sNameDescriptionData}
-            description={description}
-            onDescriptionChange={handleDescriptionChange}
-            labels={labels}
-            onLabelsChange={handleLabelsChange}
-            rules={rules}
-            onRulesChange={handleRulesChange}
-          />
+          {viewMode === 'form' ? (
+            <CreateRoleForm
+              nameDescriptionData={k8sNameDescriptionData}
+              description={description}
+              onDescriptionChange={handleDescriptionChange}
+              labels={labels}
+              onLabelsChange={handleLabelsChange}
+              rules={rules}
+              onRulesChange={handleRulesChange}
+            />
+          ) : (
+            <CreateRoleYamlView
+              namespace={namespace}
+              k8sName={k8sNameDescriptionData.data.k8sName.value}
+              displayName={
+                k8sNameDescriptionData.data.name || k8sNameDescriptionData.data.k8sName.value
+              }
+              description={description}
+              rules={rules}
+              labels={labels}
+            />
+          )}
         </PageSection>
         <PageSection hasBodyWrapper={false} stickyOnBreakpoint={{ default: 'bottom' }}>
           <CreateRoleFooter
--- a/frontend/src/pages/projects/projectRoles/CreateRolePage.tsx
+++ b/frontend/src/pages/projects/projectRoles/CreateRolePage.tsx
@@ -184,7 +184,7 @@ const CreateRolePage: React.FC<CreateRolePageProps> = ({ existingRole }) => {
         empty={false}
       >
         <PageSection hasBodyWrapper={false} isFilled data-testid="create-role-page">
-          {viewMode === 'form' ? (
+          <div hidden={viewMode !== 'form'}>
             <CreateRoleForm
               nameDescriptionData={k8sNameDescriptionData}
               description={description}
@@ -194,7 +194,8 @@ const CreateRolePage: React.FC<CreateRolePageProps> = ({ existingRole }) => {
               rules={rules}
               onRulesChange={handleRulesChange}
             />
-          ) : (
+          </div>
+          <div hidden={viewMode !== 'yaml'}>
             <CreateRoleYamlView
               namespace={namespace}
               k8sName={k8sNameDescriptionData.data.k8sName.value}
@@ -205,7 +206,7 @@ const CreateRolePage: React.FC<CreateRolePageProps> = ({ existingRole }) => {
               rules={rules}
               labels={labels}
             />
-          )}
+          </div>
         </PageSection>
         <PageSection hasBodyWrapper={false} stickyOnBreakpoint={{ default: 'bottom' }}>
           <CreateRoleFooter
--- a/frontend/src/pages/projects/projectRoles/CreateRoleYamlView.tsx
+++ b/frontend/src/pages/projects/projectRoles/CreateRoleYamlView.tsx
@@ -1,0 +1,131 @@
+import * as React from 'react';
+import {
+  Button,
+  Content,
+  Flex,
+  FlexItem,
+  Stack,
+  StackItem,
+  Title,
+  Tooltip,
+} from '@patternfly/react-core';
+import { Language } from '@patternfly/react-code-editor';
+import { CompressArrowsAltIcon } from '@patternfly/react-icons/dist/esm/icons/compress-arrows-alt-icon';
+import { ExpandArrowsAltIcon } from '@patternfly/react-icons/dist/esm/icons/expand-arrows-alt-icon';
+import YAML from 'yaml';
+import DashboardCodeEditor from '#~/concepts/dashboard/codeEditor/DashboardCodeEditor';
+import assembleRole from './assembleRole';
+import type { LabelEntry, RuleEntry } from './types';
+
+type CreateRoleYamlViewProps = {
+  namespace: string;
+  k8sName: string;
+  displayName: string;
+  description: string;
+  rules: RuleEntry[];
+  labels: LabelEntry[];
+};
+
+const CreateRoleYamlView: React.FC<CreateRoleYamlViewProps> = ({
+  namespace,
+  k8sName,
+  displayName,
+  description,
+  rules,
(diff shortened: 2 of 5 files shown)


srtanish1992:

DaoDaoNoCode:
## Consolidated Review — Form/YAML Toggle (5-agent review)

Reviewed against Jira [RHOAIENG-63157](https://redhat.atlassian.net/browse/RHOAIENG-63157) acceptance criteria, implementation correctness, YAML generation edge cases, accessibility, and test coverage. All 8 prior review threads (CodeRabbit + agent) are verified resolved.

---

### Jira Acceptance Criteria: 11/12 SATISFIED, 1 PARTIALLY

All criteria pass **except criterion 4**: the Jira and scope section both specify an **"info alert banner"**, but the implementation uses `Title` + `Content` (paragraph text) instead of a PF `Alert variant="info"` component. The text content matches exactly, but the component type doesn't. **Please verify against the Figma design** — if the design shows a title/description pattern (not an alert), this is fine; otherwise a PF `<Alert variant="info" isInline>` should be used.

---

### Findings

#### 1. [Major] Monaco editor may have layout issues when first revealed from `hidden` container

`CreateRolePage.tsx` uses `<div hidden={viewMode !== 'yaml'}>` to keep both views mounted. The `hidden` attribute applies `display: none`, which means Monaco computes its container dimensions as 0×0 on initial mount. PF CodeEditor's `ResizeObserver` does **not** fire when an element transitions from `display:none` to visible — it only fires for elements already in the layout flow.

**Consequence:** The first time the user clicks "YAML (read-only)", the Monaco editor may render with collapsed/incorrect dimensions until a window resize triggers a re-layout.

**Suggested fix:** Either:
- Use conditional rendering (`{viewMode === 'yaml' && <CreateRoleYamlView ... />}`) — simplest, since form state is lifted to the parent anyway
- Or keep `hidden` but trigger `editor.layout()` when the view becomes visible

Note: conditional rendering also eliminates the wasteful `YAML.stringify` on every keystroke while the YAML tab is hidden (Finding 7 below).

#### 2. [Minor] `doSubmit` does not filter empty-key labels, but YAML view does — behavioral divergence

`CreateRolePage.tsx:106`: `Object.fromEntries(labels.map(...))`
`CreateRoleYamlView.tsx:44`: `Object.fromEntries(labels.filter((l) => l.key).map(...))`

The YAML preview filters out empty-key labels; `doSubmit` does not. While submit is disabled when labels are invalid, this means the YAML preview and the submitted payload could theoretically differ. Consider centralizing the filter logic.

#### 3. [Minor] No `exitFullscreen()` on unmount

`CreateRoleYamlView.tsx:62-70` — if the component unmounts while in fullscreen (e.g., user navigates away via browser back), the browser stays in fullscreen with no way to exit via the PR's UI. Add to cleanup:

```ts
return () => {
  document.removeEventListener('fullscreenchange', handleFullScreenChange);
  if (document.fullscreenElement === editorContainerRef.current) {
    document.exitFullscreen();
  }
};
```

#### 4. [Minor] Missing `cy.testA11y()` after switching to YAML view

Per the project's `cypress-mock.md` convention: "Test accessibility after page state change." The Cypress tests run `cy.testA11y()` on page load (via `visitCreateRole`), but not after switching to the YAML view. Add `cy.testA11y()` in the "should switch to YAML view" test after `findYamlView().should('be.visible')`.

#### 5. [Minor] Direct `cy.findByTestId()` in Cypress test — convention violation

`rolesTabYamlToggle.cy.ts` lines 82-85 use `cy.findByTestId('yaml-editor-container').contains(...)` directly. Per project conventions, all selectors must go through page object methods. Add `findYamlEditorContainer()` to the `ProjectRolesTab` page object.

#### 6. [Minor] Cypress: no test verifying YAML view with rules entered

All Cypress tests only enter name/description. No test adds a rule via the form's "Add rule" flow, then toggles to YAML to verify the rule appears. This is a significant integration-level gap — the rules→YAML flow is only covered by the unit test.

#### 7. [Minor] Tooltip + aria-label duplicate announcement

`CreateRoleYamlView.tsx:83-90` — the fullscreen button has both `aria-label="Full screen"` and `<Tooltip content="Full screen">`. PF Tooltip defaults to `aria="describedby"`, causing screen readers to announce "Full screen, button. Full screen." Fix: add `aria="none"` to the Tooltip.

#### 8. [Nit] `displayName || k8sName` fallback duplicated in 3 places

`CreateRolePage.tsx:107`, `CreateRolePage.tsx:192-194`, and `CreateRoleYamlView.tsx:52` all compute `displayName || k8sName`. Consider centralizing in `assembleRole` itself.

#### 9. [Nit] YAML.stringify runs on every keystroke even when YAML view is hidden

Because both views stay mounted, `yamlContent = useMemo(() => YAML.stringify(...))` recomputes on every form state change even when the YAML tab is hidden. Cheap for small Role objects, but unnecessary. Resolves automatically if Finding 1 switches to conditional rendering.

---

### YAML Generation: All Clear

The `yaml` package (v2.9.0) correctly handles:
- Special characters (`:`, `#`, quotes, `&`, `*`) — auto-quoted
- Multiline descriptions — block scalar `|-` syntax
- Empty strings — serialize as `""`
- Boolean-like string `"true"` in labels — correctly quoted
- Conditional spreads for optional fields (`apiGroups`, `resources`, `resourceNames`) — all behave correctly
- Consistent with existing `PreviewYAMLModal.tsx` pattern

---

### Prior Review Threads — All Verified Resolved

| Thread | Status |
|--------|--------|
| CodeRabbit: fullscreen ownership check | ✅ Fixed — scoped to `editorContainerRef.current` |
| CodeRabbit: inline custom styling | ✅ Fixed — PF gap comment added |
| CodeRabbit: CSS class assertion | ✅ Fixed — uses `aria-pressed` |
| Agent: codeEditorHeight overflow | ✅ Fixed — `calc(100vh - 66px)` |
| Agent: mount/unmount form state loss | ✅ Fixed — uses `hidden` attribute |
| Agent: pf-m-selected on wrong element | ✅ Fixed — `aria-pressed` |
| Agent: deep path icon imports | ✅ Fixed — barrel imports |
| Agent: inline style | ✅ Fixed — PF gap comment |

srtanish1992:
Thanks for the thorough consolidated review @DaoDaoNoCode! PR is ready for review again. Here's the status on each finding:

**Fixed in latest commit:**
- **Finding 1** — Switched to hybrid rendering: form stays mounted (`hidden` attr preserves state), YAML view conditionally rendered (avoids Monaco 0×0 layout issue on first reveal).
- **Finding 2** — Added `.filter((l) => l.key)` to `doSubmit` for consistency with YAML preview.
- **Finding 3** — Added `exitFullscreen()` to useEffect cleanup on unmount.
- **Finding 4** — Added `cy.testA11y()` after switching to YAML view.
- **Finding 5** — Added `findYamlEditorContainer()` to page object, refactored all direct `cy.findByTestId('yaml-editor-container')` usages.
- **Finding 6** — Added Cypress test that creates a rule via AddRuleModal and verifies it appears in YAML.
- **Finding 7** — Added `aria="none"` to fullscreen Tooltip preventing double screen reader announcement.

**Won't fix (by design):**

- **Finding 8** (`displayName || k8sName` duplication) — Intentional separation of concerns. The three usages serve different contexts: submit handler, prop derivation, and YAML assembly. Centralizing in `assembleRole` would couple UI presentation decisions to a data utility function.

- **Finding 9** (YAML.stringify on every keystroke while hidden) — Auto-resolved by Finding 1. YAML view now only mounts when active, so no computation happens while the form view is shown.

**Acknowledged (design-verified):**
- AC note (info alert vs Title+Content) — The implementation matches the approved Figma design (node `3423-143913`), which shows a title/description pattern rather than an Alert component. Jira AC was updated during design review to reflect this.

DaoDaoNoCode:
@srtanish1992 hey, the `projects/tabs` mock test shard is failing — looks like it's the "should reflect rules in YAML" test in `rolesTabYamlToggle.cy.ts`.

The issue is that the test only clicks the `get` verb checkbox before trying to save the rule, but the save button stays disabled because no resource is selected yet. The Add Rule modal needs at least a resource picked before it'll let you save. Failed 3 retries so it's not a flake.

Should be a quick fix — just need to fill in the resource field before clicking save. Can you take a look?

frontend/src/pages/projects/projectRoles/__tests__/CreateRoleYamlView.spec.tsx
```diff
@@ -0,0 +1,136 @@
...
+    const parsed = YAML.parse(editorContent ?? '');
+
+    expect(parsed.metadata.name).toBe('');
+  });
+
+  it('should render the full-screen toggle button', () => {
+    render(<CreateRoleYamlView {...defaultProps} />);
+
+    expect(screen.getByTestId('yaml-fullscreen-toggle')).toBeInTheDocument();
+  });
```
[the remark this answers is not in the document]
srtanish1992: **Skipped — jsdom doesn't support the Fullscreen API.** Mocking `requestFullscreen`/`exitFullscreen` on a DOM element would test the mock, not real behavior, and couples the test to implementation details. The button's render presence is validated here; functional fullscreen behavior is verified in the Cypress E2E suite where a real browser is available.

frontend/src/pages/projects/projectRoles/__tests__/CreateRoleYamlView.spec.tsx
```diff
@@ -0,0 +1,136 @@
...
+    expect(screen.getByTestId('yaml-view-description')).toBeInTheDocument();
+    expect(screen.getByText(/View the live, read-only YAML for this role/)).toBeInTheDocument();
+  });
+
+  it('should render a valid Kubernetes Role YAML', () => {
+    render(<CreateRoleYamlView {...defaultProps} />);
+
+    const editorContent = screen.getByTestId('yaml-code-editor').querySelector('pre')?.textContent;
+    const parsed = YAML.parse(editorContent ?? '');
+
```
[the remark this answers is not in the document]
srtanish1992: Fixed.

DaoDaoNoCode (approved):
/lgtm

srtanish1992:
@DaoDaoNoCode addressed your latest comments.

frontend/src/pages/projects/projectRoles/CreateRolePage.tsx
```diff
@@ -150,24 +159,56 @@ const CreateRolePage: React.FC<CreateRolePageProps> = ({ existingRole }) => {
...
+                onChange={() => setViewMode('form')}
+              />
+              <ToggleGroupItem
+                text="YAML (read-only)"
+                buttonId="yaml-view-toggle"
+                data-testid="yaml-view-toggle"
+                isSelected={viewMode === 'yaml'}
+                onChange={() => setViewMode('yaml')}
+              />
+            </ToggleGroup>
```
DaoDaoNoCode, line 182: **[Major — Design divergence]** The Figma design shows a **3-option toggle group**: `Form | Split | YAML`, where the "Split" view shows the form on the left and the YAML drawer on the right side-by-side. The PR implements only a **2-option toggle**: `Form | YAML (read-only)`, where YAML replaces the form entirely.

The Jira AC says "Split view" is out of scope, so this is expected — but the toggle labels don't match the design either:
- **Figma**: `Form` | `Split` | `YAML`
- **PR**: `Form` | `YAML (read-only)`

The Figma also shows the YAML in a **PF Drawer** component (sliding panel from the right), not as a full-page replacement. The PR uses a full-page `Stack` with `Title` + `Content` + `DashboardCodeEditor` instead.

This is probably fine since Split view is explicitly out of scope, but worth confirming with the designer that the current 2-toggle full-page approach is acceptable as the interim state. When Split mode is added later, the YAML view component will likely need to be refactored into a Drawer anyway.

frontend/src/pages/projects/projectRoles/CreateRolePage.tsx
```diff
@@ -150,24 +159,56 @@ const CreateRolePage: React.FC<CreateRolePageProps> = ({ existingRole }) => {
         }
         description="Create a custom role to control what users can see and do across your cluster resources. Define permissions, navigation access, and resource scopes to implement fine-grained access control."
         headerAction={
-          // TODO: Enable when template selection modal is implemented (RHOAIENG-63156)
-          <Button variant="secondary" data-testid="select-role-template-button" isDisabled>
-            Select role template
-          </Button>
+          <Flex gap={{ default: 'gapMd' }} alignItems={{ default: 'alignItemsCenter' }}>
+            {/* TODO: Enable when template selection modal is implemented (RHOAIENG-63156) */}
+            <Button variant="secondary" data-testid="select-role-template-button" isDisabled>
```
DaoDaoNoCode, line 164: **[Low — Will conflict with PR #8149]** This line re-adds the `isDisabled` prop and the TODO comment for the "Select role template" button, but PR #8149 (template catalog) removes both and wires up the real `onClick` handler. Whichever merges second will have a merge conflict here.

Not blocking — just flagging for merge sequencing awareness.

DaoDaoNoCode (commented):
## Review: Form/YAML toggle (RHOAIENG-63157)

Compared against the [Figma design](https://www.figma.com/design/1ofcEqlUlbfdZu10NTta0Z/RBAC-feature-in-Project?node-id=2732-189118), the Jira AC, all existing review threads, and the code diff. One design divergence worth discussing, and one real issue. Rest looks good.

frontend/src/pages/projects/projectRoles/CreateRolePage.tsx
```diff
@@ -150,24 +159,56 @@ const CreateRolePage: React.FC<CreateRolePageProps> = ({ existingRole }) => {
...
-          <CreateRoleForm
-            nameDescriptionData={k8sNameDescriptionData}
-            description={description}
-            onDescriptionChange={handleDescriptionChange}
-            labels={labels}
-            onLabelsChange={handleLabelsChange}
-            rules={rules}
-            onRulesChange={handleRulesChange}
-          />
+          <div hidden={viewMode !== 'form'}>
```
DaoDaoNoCode, line 189: **[Medium — `hidden` attr vs conditional render inconsistency]** The form uses `<div hidden={viewMode !== 'form'}>` (stays mounted, invisible via HTML `hidden` attribute), while the YAML view uses `{viewMode === 'yaml' && <CreateRoleYamlView />}` (unmounts when not active).

The `hidden` approach for the form is the right call — it preserves internal state (expanded K8sNameDescriptionField, scroll position, focus). But the asymmetry is subtle. If someone later changes the form to conditional rendering to "match" the YAML pattern, they'll reintroduce the state-loss bug from the earlier review round.

Consider adding a brief comment on line 189 explaining why `hidden` is used instead of conditional rendering — or use the same pattern for both (keep YAML mounted too with `hidden`, which would also avoid the Monaco editor re-initialization cost on every toggle).

frontend/src/pages/projects/projectRoles/CreateRoleYamlView.tsx
```diff
@@ -0,0 +1,136 @@
...
+          View the live, read-only YAML for this role. This preview automatically updates to reflect
+          changes you make in <strong>Form</strong> view.
+        </Content>
+      </StackItem>
+      <StackItem isFilled>
+        {/* PF gap: no pf-v6-u-background-color-100 equivalent in v6 — remove when added */}
+        <div
+          ref={editorContainerRef}
+          data-testid="yaml-editor-container"
+          style={{ background: 'var(--pf-t--global--background--color--primary--default)' }}
```
DaoDaoNoCode, line 115: **[Low — Acknowledged PF gap]** The inline style comment is a reasonable workaround. Just noting that the earlier review thread about this was resolved with the comment added. The `pf-v6-u-background-color-*` utility class gap is a known PF v6 issue — this is the correct approach until PF adds it.

frontend/src/pages/projects/projectRoles/CreateRolePage.tsx
DaoDaoNoCode: Withdrawing this — I was looking at an older Figma frame (`2732-189118`) which showed a 3-way `Form | Split | YAML` toggle with a Drawer.

The [final design](https://www.figma.com/design/1ofcEqlUlbfdZu10NTta0Z/RBAC-feature-in-Project?node-id=3423-143983) (Sprint 39) confirms the 2-way `Form | YAML (read-only)` toggle with full-page YAML replacement — exactly what this PR implements. The header layout, button labels ("Import role template" + toggle group), title/description text, and code editor toolbar all match the design.

Resolving.

DaoDaoNoCode (approved):
/lgtm

PR #8204, opened 2026-06-18
fix(ci): harden workspace COPY check in Konflux build simulator

antowaddle:
## Context

PR #7855 extracted shared K8s types into a new `packages/k8s-core` package. It correctly updated all `Dockerfile.workspace` files but left the production `Dockerfile` files in each package untouched. The downstream `red-hat-data-services/odh-dashboard` repo maintains `Dockerfile.konflux.*` files derived from those production Dockerfiles — and none of them got the `COPY packages/k8s-core/` line either. The result was 35 module-not-found errors across all downstream Konflux builds, caught and fixed in [red-hat-data-services#2068](https://github.com/red-hat-data-services/odh-dashboard/pull/2068) after the damage had already reached downstream.

## Root cause in the simulator

`pr-build-validation.yml` Phase 0 already had a step — **"Validate workspace dependencies vs Dockerfile COPY"** — that does exactly the right thing: for each `Dockerfile.workspace`, it greps the package's production source files for imports of workspace packages and checks that the Dockerfile has a matching `COPY` instruction. However, the step only emitted a `⚠️ WARNING` with no `exit 1`, so failures were silently logged and the job passed regardless.

## What this PR does

Upgrades the check from warning-only to a **hard failure**:

- Scope `WORKSPACE_DIRS` to direct subdirectories of `packages/` only — avoids false positives from top-level workspace entries like `backend` and `frontend`
- Use each package's actual scoped npm name (read from `package.json`) for the import grep — e.g. `from '@odh-dashboard/k8s-core'` instead of the fuzzy `from.*k8s-core`
- Exclude test files (`*.spec.ts`, `*.test.ts`, `__tests__/`, `contract-tests/`) — dev-only imports don't affect the container build
- Anchor the `COPY` pattern with a trailing `/` — prevents `packages/k8s-core` matching `packages/k8s-core-v2`
- Add a `WORKSPACE_COPY_FAILURES` counter; `exit 1` after the scan if any failures are found, with a clear message naming the exact missing `COPY` line and an example of the offending import

## Verification

The check passes cleanly on `main` today — all `Dockerfile.workspace` files have the correct `COPY packages/k8s-core/` line. A future PR that adds a new shared package to `packages/` but forgets to update a consuming `Dockerfile.workspace` will now fail Phase 0 before it can merge.


lucferbux (approved):
Well-motivated and well-scoped CI hardening. The validator now correctly reads actual npm package names from `package.json`, excludes test files, uses anchored COPY grep with trailing `/` to prevent substring matches, and hard-fails instead of warning. Both CodeRabbit findings from earlier iterations were addressed in subsequent commits (cardinality mislabel in `21b1057`, substring match false negative in `90f2d87`).

One optional future improvement: the grep pattern could also match subpath imports (e.g., `from '@odh-dashboard/k8s-core/types'`) by changing the trailing `['"]` to `[/'"]` — but that's a follow-up concern, not a blocker.

All 47 CI checks passing.

/lgtm
/approve

PR #8225, opened 2026-06-19
fix: playground chat not displaying user

daniduong:
https://redhat.atlassian.net/browse/RHOAIENG-70093

## Description

The gen-ai BFF's `/api/v1/user` endpoint was returning the raw `UserModel` JSON (`{"userId":"..."}`) instead of wrapping it in the standard `Envelope` format (`{"data":{"userId":"..."}}`). Every other BFF in the monorepo (maas, autorag, mlflow, agent-ops, automl, eval-hub, model-registry) already wraps this response correctly.

On the frontend, `getCurrentUser` uses `isModArchResponse()` which checks for a `data` property on the response. Because the gen-ai BFF was missing the envelope, this check failed, throwing `'Invalid response format'`. The `UserContextProvider` caught the error silently and fell back to `undefined`, causing the chat message hook to default to the hardcoded `'User'` string.

The fix adds a `UserEnvelope` type alias (consistent with all other BFFs) and wraps all `WriteJSON` calls in the user handler with `UserEnvelope{Data: resp}`.

## How Has This Been Tested?

- Updated the existing `TestGetCurrentUserHandler` Go test to validate the envelope response format
- Verified `go build ./...` compiles cleanly
- Verified `TestGetCurrentUserHandler` passes with the envelope format
- Confirmed no other consumers are affected: the gen-ai BFF is the only one with this bug, and every frontend consumer already expects the envelope format

Before | After
--- | ---
image | Screenshot 2026-06-19 at 11 23 31 AM

## Test Impact

Updated the existing `TestGetCurrentUserHandler` test in `app_test.go` to unmarshal into `UserEnvelope` instead of raw `UserModel`, matching the corrected response format.

## Request review criteria:

## Summary by CodeRabbit

* **Refactor**
  * Updated the current user endpoint to wrap responses in a consistent envelope structure, improving API standardization across endpoints.

* **Tests**
  * Updated test cases to validate the new response structure format.


daniduong:
@coderabbitai review

NickGagan (changes requested):
In playground we actually do want the model here, does this regress it there?

Should it have the model ID by default and if that's unavailable then use the user ID?

nickmazzi:
Thanks @daniduong! The changes look correct.

@NickGagan 

> In playground we actually do want the model here, does this regress it there?
> 
> Should it have the model ID by default and if that's unavailable then use the user ID?

I think you might have misunderstood the change? Daniel did not replace the model ID, he only replaced the outgoing user message label from displaying "User" to displaying the user's actual ID as suggested by Jon.

# AI assisted review

## Overall Assessment

**Approve** — This is a clean, well-scoped bug fix. The change is minimal, correct, and consistent with every other handler in the same package.

## Suggestions

None — this is a clean, minimal fix. Ship it.

# Test Validation

## autorag

### Overall Summary

| Suite      | Result |
|------------|--------|
| Frontend   | ✅     |
| BFF        | ✅     |
| Contract   | ✅     |
| BFF Lint   | ✅     |
| Lint       | ✅     |
| Type Check | ✅     |

## gen-ai

### Overall Summary

| Suite      | Result |
|------------|--------|
| Frontend   | ✅     |
| BFF        | ✅     |
| Contract   | ✅     |
| BFF Lint   | ✅     |
| Lint       | ✅     |
| Type Check | ✅     |

# Manual validation
LGTM

image

PR #8235, opened 2026-06-22
fix: harden non-streamed responses API for long-running requests

gmenher:
https://issues.redhat.com/browse/RHOAIENG-63831

## Description

Extends timeout configuration across all proxy layers to prevent premature connection termination for non-streamed inference requests that exceed 30 seconds (CPU models, vector stores, guardrails, tool calls).

The request path traverses multiple proxies between the browser and the gen-ai BFF, each with a default 30-second timeout:

```
Browser → HAProxy → Envoy Gateway → kube-rbac-proxy → Fastify proxy → gen-ai BFF → OGX
```

The BFF's `handleNonStreamingResponse` writes zero bytes until OGX returns the full response. Any proxy layer that doesn't receive data within 30s kills the connection, resulting in a 502 Bad Gateway.

### Changes

1. **kube-rbac-proxy** — Add `--upstream-timeout=5m` to deployment args (primary fix; its 30s default was actively killing connections)
2. **Envoy Gateway** — Add `timeouts.request: 5m` to HTTPRoute for the responses endpoint
3. **Fastify proxy** — Add `http.requestOptions.timeout: 5m` selectively for the genAi module only (does not affect other modules)
4. **RHOAI overlay** — Update kustomize patch for the new HTTPRoute rule

### How Has This Been Tested?

- **Unit tests:** Full backend test suite executed (175 tests passing). Added `proxy-timeout.spec.ts` with 4 new tests verifying that the `httpRequestTimeout` option is correctly propagated to the proxy plugin, only applied when provided, and selectively targeting the gen-ai module.
- **TypeScript type-check:** `tsc --noEmit` passes with zero errors across the backend project.
- **ESLint:** All modified and new files pass with zero warnings.
- **End-to-end cluster verification:** Deployed a controlled slow-response server (35s delay) on a ROSA cluster and verified the full request chain completes successfully at 35 seconds through all proxy layers (HAProxy → Envoy → kube-rbac-proxy → Fastify → BFF).
- **Layer isolation testing:** Each proxy layer was independently verified via `oc port-forward` to confirm correct timeout propagation at every level.

### Test Impact

Added `backend/src/__tests__/routes/proxy-timeout.spec.ts` — unit tests verifying that the `httpRequestTimeout` option is passed correctly to the proxy plugin and only applied to the gen-ai module.

## Request review criteria:

- [x] The code follows our Best Practices

---

### Testing approach

To reproduce the 30-second timeout deterministically without depending on real hardware (which would be non-reproducible and environment-dependent), the following setup was used:

1. **Paused the RHOAI operator** (scaled to 0 replicas) to prevent it from reverting manual configuration changes during testing.
2. **Deployed a minimal Python server (`slow-vllm`)** in the cluster that simulates a LlamaStack/OGX endpoint:
   - Responds immediately to `GET /v1/models` (so the BFF discovers and registers the model)
   - Sleeps exactly 35 seconds on `POST /v1/responses` before writing the first byte
3. **Configured the gen-ai BFF** with `LLAMA_STACK_URL` pointing to this slow server.
4. **Applied the HAProxy route annotation** (`haproxy.router.openshift.io/timeout: 5m`) on the data-science-gateway Route.
5. **Built and deployed a custom dashboard image** with the Fastify proxy timeout fix.
6. **Added `--upstream-timeout=5m`** to the kube-rbac-proxy container args via `oc patch`.

This approach provides a controlled, reproducible environment where the 35-second response time consistently exceeds the 30-second proxy timeout, making the failure deterministic.

### Result

After applying all fixes, the Playground completed the non-streamed request successfully at ~35 seconds:

image

The screenshot shows:
1. **Model response received** — "I am slow." displayed in the chat after ~35s
2. **HTTP 201 status** — The `responses?namespace=...` request in DevTools shows status 201 with a time of 35.18s
3. **Streaming OFF** — Confirming this is the non-streamed path

> **Note:** The 404 errors visible for `vectorstores` and `mcps` endpoints in the screenshot are expected — they occur because the RHOAI operator was paused during testing and those services were not fully reconciled. They are unrelated to this fix.

### Investigation methodology

The root cause was identified by isolating each proxy layer with `oc port-forward`:

| Layer | Port-forward target | Result |
|---|---|---|
| gen-ai BFF (:8143) | Bypasses all proxies | HTTP 201 at 35s ✓ |
| Fastify proxy (:8080) | Bypasses kube-rbac-proxy + Envoy | HTTP 201 at 35s ✓ |
| kube-rbac-proxy (:8443) | Bypasses only Envoy + HAProxy | HTTP 502 at 30.6s ✗ |
| Full route (browser) | Through all layers | HTTP 502 at 30s ✗ |

This proved the kube-rbac-proxy was the primary blocker. Its `--upstream-timeout` flag (default 30s) was discovered by inspecting the binary help output.

## Summary by CodeRabbit

* **New Features**
  * Added configurable proxy HTTP request timeout support (used to extend Gen AI module federation request handling up to 5 minutes).

* **Bug Fixes**
  * Extended `kube-rbac-proxy` upstream timeout to 5 minutes to reduce premature failures.
  * Updated gateway routing to apply 5-minute request/backend timeouts for Gen AI Playground responses.

* **Tests**
  * Added tests covering proxy timeout configuration and cluster upstream URL behavior.

* **Chores**
  * Adjusted HTTPRoute backend service reference for improved routing consistency.

### DIFF
--- a/backend/src/routes/module-federation.ts
+++ b/backend/src/routes/module-federation.ts
@@ -16,6 +16,10 @@ export default async (fastify: KubeFastifyInstance): Promise<void> => {
     fastify.log.info(
       `Module federation configured for: ${mfConfig.map((mf) => mf.name).join(', ')}`,
     );
+    // Extended timeout for gen-ai module: AI inference (CPU models, vector stores,
+    // guardrails, tool calls) routinely exceeds undici's default 30s timeout.
+    const GEN_AI_PROXY_TIMEOUT_MS = 5 * 60 * 1000; // 5 minutes
+
     mfConfig.forEach(({ name, backend, proxyService }) => {
       if (backend) {
         registerProxy(fastify, {
@@ -63,6 +67,7 @@ export default async (fastify: KubeFastifyInstance): Promise<void> => {
           },
           local: proxy.localService,
           headers: proxy.headers,
+          httpRequestTimeout: name === 'genAi' ? GEN_AI_PROXY_TIMEOUT_MS : undefined,
         });
       });
     });
--- a/manifests/core-bases/base/httproute.yaml
+++ b/manifests/core-bases/base/httproute.yaml
@@ -8,6 +8,19 @@ spec:
       kind: Gateway
       namespace: openshift-ingress
   rules:
+    # Extended timeout for AI inference via the Gen AI Playground responses API.
+    # OpenShift HAProxy defaults to 30s idle timeout, but non-streamed inference
+    # (CPU models, vector stores, guardrails, tool calls) routinely exceeds this.
+    - matches:
+        - path:
+            type: PathPrefix
+            value: "/_mf/genAi/gen-ai/api/v1/lsd/responses"
+      timeouts:
+        request: 5m
+        backendRequest: 5m
+      backendRefs:
+        - name: odh-dashboard
+          port: 8443
     - matches:
         - path:
             type: PathPrefix
--- a/backend/src/__tests__/routes/proxy-timeout.spec.ts
+++ b/backend/src/__tests__/routes/proxy-timeout.spec.ts
@@ -1,0 +1,84 @@
+import httpProxy from '@fastify/http-proxy';
+
+// Mock @fastify/http-proxy to capture registration options
+jest.mock('@fastify/http-proxy', () => jest.fn());
+
+// Mock constants
+jest.mock('../../utils/constants', () => ({
+  DEV_MODE: false,
+  PORT: 8080,
+  IP: '0.0.0.0',
+  LOG_LEVEL: 'info',
+  APP_ENV: 'production',
+}));
+
+describe('registerProxy httpRequestTimeout', () => {
+  let registerProxy: typeof import('../../utils/proxy').registerProxy;
+  let mockFastify: any;
+
+  beforeEach(() => {
+    jest.clearAllMocks();
+    mockFastify = {
+      register: jest.fn(),
+      log: { info: jest.fn() },
+    };
+    // Re-import after mocks are set up
+    registerProxy = require('../../utils/proxy').registerProxy;
+  });
+
+  it('should pass http.requestOptions.timeout when httpRequestTimeout is provided', async () => {
+    await registerProxy(mockFastify, {
+      prefix: '/gen-ai/api',
+      rewritePrefix: '/api',
+      service: { name: 'odh-dashboard', namespace: 'test-ns', port: 8143 },
+      tls: true,
+      httpRequestTimeout: 300000,
+    });
+
+    expect(mockFastify.register).toHaveBeenCalledWith(
+      httpProxy,
+      expect.objectContaining({
+        http: {
+          requestOptions: { timeout: 300000 },
+        },
+      }),
+    );
+  });
+
+  it('should NOT pass http options when httpRequestTimeout is undefined', async () => {
+    await registerProxy(mockFastify, {
+      prefix: '/maas/api',
+      rewritePrefix: '/api',
+      service: { name: 'odh-dashboard', namespace: 'test-ns', port: 8243 },
+      tls: true,
+    });
+
+    const registeredOptions = mockFastify.register.mock.calls[0][1];
+    expect(registeredOptions.http).toBeUndefined();
+  });
+
+  it('should construct correct upstream URL for cluster mode', async () => {
+    await registerProxy(mockFastify, {
+      prefix: '/gen-ai/api',
+      service: { name: 'my-service', namespace: 'my-ns', port: 8143 },
+      tls: true,
+      httpRequestTimeout: 300000,
+    });
+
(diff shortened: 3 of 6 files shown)


toklumpp:

backend/src/routes/module-federation.ts
```diff
@@ -63,6 +67,7 @@ export default async (fastify: KubeFastifyInstance): Promise<void> => {
           },
           local: proxy.localService,
           headers: proxy.headers,
+          httpRequestTimeout: name === 'genAi' ? GEN_AI_PROXY_TIMEOUT_MS : undefined,
```
toklumpp, line 70: How does this look for the user? Will the user see any indication what happens during this 5 minutes?

manifests/core-bases/base/httproute.yaml
```diff
@@ -8,6 +8,19 @@ spec:
...
   rules:
+    # Extended timeout for AI inference via the Gen AI Playground responses API.
+    # OpenShift HAProxy defaults to 30s idle timeout, but non-streamed inference
+    # (CPU models, vector stores, guardrails, tool calls) routinely exceeds this.
+    - matches:
+        - path:
+            type: PathPrefix
+            value: "/_mf/genAi/gen-ai/api/v1/lsd/responses"
+      timeouts:
+        request: 5m
```
toklumpp, line 19: Could we change our BFF to bypass the timeout by sending a heartbeat instead? Then we could better differentiate between a lost connection and a long-waiting response. Especially as this might be a wireless/mobile connection between the browser and the BFF.

toklumpp: Alternatively we might also use some asynchronous fetching here so we don't keep the connection open for too long

toklumpp (changes requested):
Root-cause analysis looks very good. I think it might be possible to give the user a better idea of what is happening. Otherwise they might just navigate away if they are waiting for a response for 5 minutes.

backend/src/routes/module-federation.ts
gmenher: Currently the user sees the loading spinner in the chat while the request is processing. This PR only addresses the infrastructure side (preventing early connection drops), but I agree a more informative UX indicator would be a nice improvement.

manifests/core-bases/base/httproute.yaml
NickGagan: We send the heartbeat for streaming, but doesn't work for non-streaming

NickGagan (commented):
We'll need someone from the Monarch team to take a look at this. 5m might be a bit long, but I'll defer to them.

manifests/core-bases/base/httproute.yaml
gmenher: I agree that a heartbeat would be a cleaner solution but for non-streamed requests the client expects a single JSON response, so sending heartbeats mid-response would require changing the API contract. That would be a broader architectural change, for the moment I could be reducing the timeout to 3 min.

christianvogt:
@NickGagan is this still needed for 3.5? 
Sorry the monarch team hasn't had time to look into a better solution for you.

christianvogt:
@gmenher I think the only real blocker for your time out is the kube-rbac-proxy. For now we are removing the `upstream-timeout` as part of https://github.com/opendatahub-io/odh-dashboard/pull/8706.

Can you confirm with just that change if it solves your timeout issue.

gmenher:
Hi @christianvogt, I just confirmed it, --upstream-timeout=0 from #8706 fully resolves the timeout issue for both streamed and non-streamed requests. My additional patches (HTTPRoute timeout, Fastify proxy timeout) are no longer needed. I tested this following this procedure:

Test setup:

Since the available models (GPT-4o Mini via MaaS) respond in under 30s even with RAG/Knowledge enabled, I couldn't reproduce the timeout naturally. Instead, deployed a Node.js mock OGX server (slow-ogx) in the playground namespace that intentionally delays 45 seconds before sending any response, simulating a slow CPU model inference or a complex RAG pipeline. Adapted the mock to return the correct OpenAI Responses API schema for both streamed (SSE) and non-streamed (JSON) modes.

The mock was wired through the existing service (lsd-genai-playground-service) so the request travels the full proxy chain: Browser → HAProxy (OpenShift Router) → kube-rbac-proxy (--upstream-timeout=0) → Fastify backend → BFF → slow-ogx mock.

Results:

Streaming ON: Response arrived after 45s (TTFT: 45.02s) — no timeout, rendered correctly
Streaming OFF: Response arrived after 45s (Waiting for server: 45.22s) — no timeout, rendered correctly
Both cases confirm that with --upstream-timeout=0 alone, the 30s timeout no longer occurs. The kube-rbac-proxy was indeed the only layer enforcing the default 30s cut.

Screenshot (non-streamed, 45s delay, clean response):

image

I'm closing this PR as the fix in #8706 is sufficient. Thanks @christianvogt for the pointer and the work done! 😄

PR #8239, opened 2026-06-22
fix(operator): bump Go to 1.26 and harden Dockerfile

lucferbux:
## Description

Bumps the dashboard-operator Go version to 1.26 and applies several hardening improvements to the Dockerfile:

1. **Go 1.26 bump** — Aligns with the rest of the Go modules in the monorepo (`automl`, `eval-hub`, `autorag`, `maas`, `mlflow` BFFs are already on 1.26).
2. **Cross-architecture support** — Adds `TARGETOS`/`TARGETARCH` build args with `GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH}`, matching the pattern used in other Konflux Dockerfiles (e.g., `agent-ops`). This ensures correct binaries if multi-arch builds are ever enabled.
3. **Smaller binary** — Adds `-ldflags="-s -w"` to strip debug symbols and DWARF information.
4. **Security hardening** — Removes `chmod -R g+w` on the manifests directory. The operator only reads manifests at runtime; `COPY` already preserves readable permissions. Granting group-write to read-only data is an unnecessary attack surface.
5. **Fix `.dockerignore` for operator builds** — Removes `manifests/` from the root `.dockerignore`. The operator Dockerfile needs `COPY manifests/ /opt/manifests/dashboard/`, but when the Tekton build task copies the Dockerfile to a temp location, it falls back to the root `.dockerignore` instead of the operator-specific `Dockerfile.dockerignore`, causing the build to fail with "no items matching glob".

## How Has This Been Tested?

- Verified `go mod tidy` produces no changes
- Verified `CGO_ENABLED=0 GOOS=linux go build -a -ldflags="-s -w" -o /tmp/manager ./cmd/manager` compiles successfully with Go 1.26
- Verified `manifests/` directory exists at repo root and is no longer excluded by `.dockerignore`

## Test Impact

No test changes needed — this is a build/infrastructure change. The operator's existing unit tests continue to pass with Go 1.26.

## Request review criteria:

## Summary by CodeRabbit

* **Chores**
  * Upgraded the Go toolchain to 1.26 for improved build compatibility.
  * Enhanced the Docker build to support cross-platform builds using configurable target OS/architecture.
  * Updated the Docker packaging to ensure the `manifests/` directory is included in images (and removed the previous recursive permission adjustment for that content).

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -12,7 +12,6 @@ tests/
 **/__tests__/
 scripts/
 !scripts/prepare-production-manifest.js
-manifests/
 
 # Local .env images
 **/.env.development
--- a/dashboard-operator/Dockerfile
+++ b/dashboard-operator/Dockerfile
@@ -12,7 +12,7 @@ COPY dashboard-operator/cmd/ cmd/
 COPY dashboard-operator/api/ api/
 COPY dashboard-operator/internal/ internal/
 
-RUN CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} go build -a -ldflags="-s -w" -o /tmp/manager ./cmd/manager
+RUN CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH:-amd64} go build -a -ldflags="-s -w" -o /tmp/manager ./cmd/manager
 
 FROM registry.access.redhat.com/ubi9/ubi-minimal:9.3
 
--- a/dashboard-operator/Dockerfile
+++ b/dashboard-operator/Dockerfile
@@ -1,4 +1,7 @@
-FROM golang:1.25 AS builder
+FROM golang:1.26 AS builder
+
+ARG TARGETOS
+ARG TARGETARCH
 
 WORKDIR /usr/src/app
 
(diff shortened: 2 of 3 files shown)


.dockerignore
```diff
@@ -12,7 +12,6 @@ tests/
 **/__tests__/
 scripts/
 !scripts/prepare-production-manifest.js
```
liday-rh, line 14: 🟡 **Low — Removing `manifests/` widens build context for all images, not just the operator.**

The root `.dockerignore` is shared by every Dockerfile in the repo. Removing the `manifests/` exclusion means the manifests directory is now sent to the Docker daemon for **all** builds — the main Node.js dashboard, agent-ops, maas, etc. — not just the operator that actually needs it.

The PR description explains the root cause well: Tekton copies the Dockerfile to a temp location and falls back to the root `.dockerignore`, ignoring the operator-specific `Dockerfile.dockerignore`. But the fix trades operator build correctness for increased build context size on every other image.

A more surgical option would be to keep `manifests/` in `.dockerignore` and instead fix the Tekton build task to use the correct `.dockerignore`, or have the operator's Tekton pipeline explicitly pass `--file` with the correct context. If that's not feasible short-term, this workaround is acceptable but worth calling out as tech debt.

liday-rh (commented):
Overall this is a clean PR — Go 1.26 aligns with peer modules, `-ldflags="-s -w"` is correct for stripping, `chmod -R g+w` removal is valid hardening, and no dependencies changed. One minor note on the `.dockerignore` change scope (see inline).

FedeAlonso (approved):

PR #8255, opened 2026-06-23
Fix MLflow test selectors for upstream v3.13.0 'Edit experiment' rename

FedeAlonso:
## Description
  - Update MLflow experiment page object selectors to match upstream v3.13.0 menu item IDs (`rename` →
  `mlflow.experiment_page.managementMenu.edit-experiment`, `delete` → `mlflow.experiment_page.managementMenu.delete`)
  - Replace old rename modal selectors with new "Edit experiment" dialog selectors (`findByRole('dialog', { name: 'Edit experiment' })`)
  - Update test step descriptions from "rename" to "edit experiment"

## How Has This Been Tested?
dashboard-e2e-tests/1473
image

## Request review criteria:

If you have UI changes: 

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the experiment editing flow so renaming an experiment now uses the current edit experience.
  * Improved the delete action hookup to match the latest experiment management menu behavior.
  * Adjusted automated end-to-end coverage to reflect the updated edit dialog and save action.


packages/cypress/cypress/pages/mlflowExperiments.ts
```diff
@@ -155,20 +155,20 @@ class MlflowExperiments {
...
 
   findDeleteAction() {
-    return cy.findByTestId('delete');
+    return cy.findByTestId('mlflow.experiment_page.managementMenu.delete');
   }
 
-  findRenameInput() {
-    return this.findCreateExperimentModal().find('input').first();
+  findEditExperimentNameInput() {
+    return cy.findByRole('dialog', { name: 'Edit experiment' }).find('input').first();
```
antowaddle, line 167: Can we create a testID here?

packages/cypress/cypress/pages/mlflowExperiments.ts
```diff
@@ -155,20 +155,20 @@ class MlflowExperiments {
...
   }
 
-  findRenameSubmitButton() {
-    return this.findCreateExperimentModal().findByRole('button', { name: 'Save' });
+  findEditExperimentSubmitButton() {
+    return cy.findByRole('button', { name: 'Save' });
```
antowaddle: Can we create a testID here?

antowaddle (changes requested):
Thanks @FedeAlonso for stepping in here, 2 small comments.

packages/cypress/cypress/pages/mlflowExperiments.ts
FedeAlonso: These elements are part of the upstream MLflow UI, rendered via module federation (`mlflowEmbedded` remote entry). The dialog, input, and Save button are all owned by MLflow's React code — we can't add `data-testid` attributes without modifying the upstream [mlflow/mlflow](https://github.com/mlflow/mlflow) repo.

The input selector is already scoped to the dialog (`findByRole('dialog', { name: 'Edit experiment' }).find('input').first()`), which is as specific as we can get without upstream testIDs.

packages/cypress/cypress/pages/mlflowExperiments.ts
FedeAlonso: Same situation — this is an upstream MLflow button. Can't add a `data-testid` without changes in the mlflow repo.

I've scoped the selector to the Edit Experiment dialog per CodeRabbit's suggestion: `findByRole('dialog', { name: 'Edit experiment' }).findByRole('button', { name: 'Save' })`. This avoids matching other Save buttons on the page while being as specific as possible given the upstream constraints.

packages/cypress/cypress/pages/mlflowExperiments.ts
antowaddle: This type of locator is brittle and given that we're currently fixing it for that exact reason I think it's probably something worth investigating further. If you want to proceed as it is that's fine, but we should at a minimum create a task to fix this for all the related mlflow tests going forward.

FedeAlonso: I have made those changes here https://github.com/opendatahub-io/mlflow/pull/317

FedeAlonso:
Tests results after the latests changes can be found in run 1494 
image

packages/cypress/cypress/pages/mlflowExperiments.ts
antowaddle: Awesome thanks - let's go ahead as is and then we'll tweak this once the above merges.

antowaddle (approved):
/lgtm

antowaddle (approved):
/lgtm

PR #8290, opened 2026-06-25
Fix llmd yaml e2e tests

emilys314:
Fixes e2e errors from https://redhat.atlassian.net/browse/RHOAIENG-70687 and https://redhat.atlassian.net/browse/RHOAIENG-70580

## Description

We updated from `serving.kserve.io/v1alpha1` to `serving.kserve.io/v1alpha2`, but not in the e2e test fixture, this caused a discrepancy when asserting on the api version when creating llmd. 
Also the generative rearrangement changed the order of when the yaml would be empty 

All passing:
image

The first test model deployed successfully 
Screenshot 2026-06-25 at 7 26 34 AM
The second test model deployed successfully 
Screenshot 2026-06-25 at 7 24 18 AM

## How Has This Been Tested?

Run the `packages/cypress/cypress/tests/e2e/dataScienceProjects/models/testDeployLLMDServing.cy.ts`

## Test Impact

Fixes existing e2e test

## Request review criteria:

If you have UI changes: 

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated LLM inference service deployment configurations and fixtures to use the newer KServe `apiVersion` (v1alpha2) for consistent compatibility.
  * Improved the LLM deployment e2e flow to take the expected Kubernetes `apiVersion` and deployment method from fixture data, strengthening YAML editor validation.
  * Refreshed related test inputs and models to align with the updated `apiVersion`.
  * Added a dedicated vLLM on MaaS e2e fixture and switched the test to use its deployment method.


packages/cypress/cypress/tests/e2e/dataScienceProjects/models/testDeployLLMDServing.cy.ts
```diff
@@ -134,16 +134,13 @@ describe('A user can deploy an LLMD model', () => {
...
-
       cy.step('Verify YAML Viewer');
       // Stub clipboard API AFTER page load (window changes on navigation)
       stubClipboard('copiedYAML');
       modelServingWizard.findYAMLViewerToggle(YAMLViewerToggleOption.YAML).should('exist').click();
       modelServingWizard.findYAMLEditorEmptyState().should('be.visible');
       modelServingWizard.findYAMLViewerToggle(YAMLViewerToggleOption.FORM).should('exist').click();
-      modelServingWizard.findServingRuntimeTemplateSearchSelector().click();
-      modelServingWizard.findGlobalScopedTemplateOption(servingRuntime).should('exist').click();
+      modelServingWizard.selectDeploymentMethodByKey('llm-inference-service-llmd');
```
sridarna: Could you add `llm-inference-service-llmd` to the fixture file. and read that testData here in the file

packages/cypress/cypress/tests/e2e/dataScienceProjects/models/testDeployLLMDServing.cy.ts
```diff
@@ -155,7 +152,7 @@ describe('A user can deploy an LLMD model', () => {
       getClipboardContent('copiedYAML').then((copied) => {
         expect(copied).to.have.length.at.least(1);
         const yamlContent = copied[0];
-        expect(yamlContent).to.include('apiVersion: serving.kserve.io/v1alpha1');
+        expect(yamlContent).to.include('apiVersion: serving.kserve.io/v1alpha2');
```
antowaddle: Can we add this to fixtures please?

antowaddle (changes requested):
@emilys314 Could we also please run this test in Jenkins to ensure that it works prior to merging?

packages/cypress/cypress/tests/e2e/dataScienceProjects/models/testDeployLLMDServing.cy.ts
emilys314: I'm also unsure what dictates what goes into a fixture vs is a core part of the test code?

Do you want me to add a new entry into the `DataScienceProjectData` https://github.com/opendatahub-io/odh-dashboard/blob/main/packages/cypress/cypress/types.ts#L255? 

This object is data science project which this fixture data is specific to a model which doesn't line up imo. 

Also I saw that the `testVLLMOnMaaS.cy.ts` text is using the same fixture, I'd need to create a new fixture so there aren't any conficts with what is being selected, unless I add a new variables that will only be selected in the llmd test

packages/cypress/cypress/tests/e2e/dataScienceProjects/models/testDeployLLMDServing.cy.ts
emilys314: Can you help me understand the reason why? Sorry for my ignorance, I have some questions in this related comment too regarding this: https://github.com/opendatahub-io/odh-dashboard/pull/8290#discussion_r3475227283

The fixture object is based on a `DataScienceProjectData`, which doesn't seem like the place to hold model serving kubernetes details like the apiVersion

antowaddle: Typically, all test data is defined in our fixtures file. This allows us to then maintain/manage it a central place with help with management/maintenance etc. For example, if you were to add multiple tests with this test file (multiple 'it's) they could all share/read this from the one location. 

Additionally, when running the tests on other platforms (Power etc.) we may want to swap out test data which is unsupported etc. for other fixture files.

packages/cypress/cypress/tests/e2e/dataScienceProjects/models/testDeployLLMDServing.cy.ts
sridarna: This value would go to fixture file. Since we are avoiding string input/ string validations.
I think we can add deploymentMethod entry in the DataScienceProjectData. Wouldn't testVLLMOnMaaS.cy.ts need update to select deploymentMethod? So same DataScienceProjectData type can be used.

packages/cypress/cypress/tests/e2e/dataScienceProjects/models/testDeployLLMDServing.cy.ts
```diff
@@ -155,7 +156,8 @@ describe('A user can deploy an LLMD model', () => {
       getClipboardContent('copiedYAML').then((copied) => {
         expect(copied).to.have.length.at.least(1);
         const yamlContent = copied[0];
-        expect(yamlContent).to.include('apiVersion: serving.kserve.io/v1alpha1');
+        // expect(yamlContent).to.include('apiVersion: serving.kserve.io/v1alpha2');
```
antowaddle, line 159: Can we remove this commented out line?

antowaddle (changes requested):
@emilys314 looks good now, thanks. The test is failing as it's selecting the HWP from an incorrect NS which seems to have been set in Jenkins. Can you look at this again please? Please rerun both the Nightly ODH/RHOAI builds with this branch but remove the operator installation and all the namespaces etc. will be correctly set. 
Image

sridarna:
Tested in Jenkins [job/dashboard-e2e-tests/1573/Test_20Reports/] 
Both the E2E are passing
Screenshot 2026-06-29 at 9 07 41 AM

PR #8312, opened 2026-06-26
feat(llmd-serving): add topology type and custom config selection to deployment wizard

YuliaKrimerman:
https://issues.redhat.com/browse/RHOAIENG-70582

## Description

Adds "Topology type" dropdown and "Custom topology configurations" selector to the llm-d model deployment wizard (Step 2: Model deployment).

**New wizard fields:**
- **TopologyTypeField** — Hardcoded dropdown with 4 topology types (Single node, Multi-node, Single node disaggregated, Multi-node disaggregated). Options are disabled/greyed out when no matching `LLMInferenceServiceConfig` resources exist for that type. Single node is always enabled.
- **CustomTopologyConfigField** — Optional config selector that shows enabled `LLMInferenceServiceConfig` resources matching the selected topology type. Resets when topology type changes. Hidden when no configs exist for the selected type.

**Behavior:**
- Both fields appear only when deployment method is "LLM inference service deployment with llm-d"
- Gated behind the `llmdTopologyConfigs` Tech Preview feature flag (`SupportedArea.LLMD_TOPOLOGY_CONFIGS`)
- Fields are positioned between "Deployment method" and "Hardware profile" per design
- Topology configs are fetched from the dashboard namespace and filtered by `opendatahub.io/config-type` label matching `TopologyType` enum values

**Files changed:**
- `packages/llmd-serving/src/wizardFields/TopologyTypeField.tsx` (new)
- `packages/llmd-serving/src/wizardFields/CustomTopologyConfigField.tsx` (new)
- `packages/llmd-serving/extensions/extensions.ts` — register both fields
- `packages/model-serving/src/components/deploymentWizard/steps/ModelDeploymentStep.tsx` — position fields after deployment method
- `packages/cypress/cypress/tests/mocked/modelServing/modelServingLlmdTopology.cy.ts` (new — 7 test cases)
- `packages/cypress/cypress/pages/modelServing.ts` — add page object methods

## How Has This Been Tested?

1. **Manual testing on cluster** with `?devFeatureFlags=true,llmdTopologyConfigs=true`:
   - Verified topology dropdown appears when llm-d deployment method is selected
   - Verified Single node is always enabled; other options are greyed out without matching configs
   - Verified dropdown is hidden when `llmdTopologyConfigs` flag is disabled

2. **Testing all topology options with configs on cluster:**
Default is single node always enabled 
Screenshot 2026-06-26 at 11 07 56 AM

   Apply these to enable all other 3  topology types:
   ```bash
   oc apply -f - <<EOF
   apiVersion: serving.kserve.io/v1alpha2
   kind: LLMInferenceServiceConfig
   metadata:
     name: test-multi-node-config
     namespace: opendatahub
     annotations:
       openshift.io/display-name: "Multi-node Data Parallel"
     labels:
       opendatahub.io/config-type: workload-multi-node-data-parallel
   spec:
     worker: {}
     parallelism:
       data: 2
   ---
   apiVersion: serving.kserve.io/v1alpha2
   kind: LLMInferenceServiceConfig
   metadata:
     name: test-single-node-pd-config
     namespace: opendatahub
     annotations:
       openshift.io/display-name: "Single Node P/D Disaggregated"
     labels:
       opendatahub.io/config-type: workload-single-node-pd
   spec:
     prefill:
       replicas: 2
   ---
   apiVersion: serving.kserve.io/v1alpha2
   kind: LLMInferenceServiceConfig
   metadata:
     name: test-multi-node-pd-config
     namespace: opendatahub
     annotations:
       openshift.io/display-name: "Multi-node P/D Disaggregated"
     labels:
       opendatahub.io/config-type: workload-multi-node-data-parallel-pd
   spec:
     worker: {}
     prefill:
       replicas: 2
   EOF
```

After applying, refresh the wizard — all 4 topology types should be selectable and the custom config dropdown should appear for each.
Screenshot 2026-06-26 at 11 11 20 AM

Cleanup:

```oc delete llminferenceserviceconfig -n opendatahub test-multi-node-config test-single-node-pd-config test-multi-node-pd-config ```
Lint and type-check pass for all modified files.

**Test Impact**
Added new Cypress mocked test file: packages/cypress/cypress/tests/mocked/modelServing/modelServingLlmdTopology.cy.ts (7 test cases)

## Summary by CodeRabbit

* **New Features**
  * Added topology type selection and custom topology configuration to the deployment wizard.
  * Topology options are driven by available configs and include model deployment review details.

* **Bug Fixes**
  * Improved persistence and visibility of selected topology configs when the available options change.
  * Enhanced handling for loading failures and empty states with clearer disabled options and helper/warning messaging.
  * Ensured field activation only applies for the relevant deployment scenario and method.

### CODE REVIEW

File: packages/llmd-serving/src/wizardFields/CustomTopologyConfigField.tsx [hunk 1]
```diff
@@ -0,0 +1,161 @@
...
+
+export const CustomTopologyConfigFieldWizardField: CustomTopologyConfigFieldType = {
+  id: 'llmd-serving/custom-topology-config',
+  step: 'modelDeployment',
+  type: 'addition',
+  isActive,
+  reducerFunctions: {
+    resolveDependencies: (formData) => ({
+      // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
+      topologyType: formData['llmd-serving/topology-type'] as TopologyTypeFieldData | undefined,
```
emilys314: 🚨🚨 Is there a way to avoid `as` here?

I guess that formData doesn't know about `TopologyTypeFieldData` because it's a fully dynamic field..

1 option is to combine TopologyType and CustomTopologyConfig them into the same field, since they are very interconnected

2nd option would be to do some safer typeguard checking and create a ~`isTopologyType = (data) => data is TopologyTypeFieldData` 

3rd, there might be a better way to automatically infer that this exists? nothing immediately jumps to mind though

packages/model-serving/src/components/deploymentWizard/steps/ModelDeploymentStep.tsx
```diff
@@ -9,6 +9,11 @@ import { NumReplicasField } from '../fields/NumReplicasField';
 import { GenericFieldRenderer } from '../fields/GenericFieldRenderer';
 import { ExternalDataMap } from '../ExternalDataLoader';
 
+const EXPLICIT_TOPOLOGY_FIELD_IDS = [
```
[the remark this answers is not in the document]
emilys314: Yeah this is a known gap in the wizard, the dynamic fields have to appear in certain places but there is no mechanism yet for defining the position besides on which step or specifically singling out the id, (or also a parent)

A `position` would be the way to go but not sure if that is in scope for this unless you feel like it @YuliaKrimerman. The sooner we have it the better though

packages/cypress/cypress/tests/mocked/modelServing/modelServingLlmdTopology.cy.ts
```diff
@@ -0,0 +1,281 @@
...
+}: {
+  topologyConfigs?: typeof mockTopologyConfigs;
+  llmdTopologyConfigsEnabled?: boolean;
+} = {}) => {
+  cy.interceptOdh(
+    'GET /api/dsc/status',
+    mockDscStatus({
+      components: {
+        [DataScienceStackComponent.K_SERVE]: { managementState: 'Managed' },
+        [DataScienceStackComponent.OGX_OPERATOR]: { managementState: 'Managed' },
```
emilys314: Ogx shouldn't be needed

```suggestion
```

packages/cypress/cypress/tests/mocked/modelServing/modelServingLlmdTopology.cy.ts
```diff
@@ -0,0 +1,281 @@
...
+      modelServingWizard.findModelTypeSelect().findSelectOption('Generative AI model').click();
+      modelServingWizard.findNextButton().click();
+
+      modelServingWizard.selectDeploymentMethodByKey('llm-inference-service-llmd');
+
+      // Open the topology type dropdown
+      modelServingWizard.findTopologyTypeSelect().click();
+
+      // Single node and Multi-node should be enabled
+      cy.findByTestId(`topology-type-${TopologyType.SINGLE_NODE}`).should(
```
emilys314: I think our cypress guidelines are against doing direct `findByTestId` function calls in the tests, so it would be good to build a findTopologyTypeSelectOption

packages/llmd-serving/src/wizardFields/CustomTopologyConfigField.tsx
```diff
@@ -0,0 +1,161 @@
...
+        label: getDisplayNameFromK8sResource(config),
+        description: config.metadata.annotations?.['openshift.io/description'],
+        dataTestId: `topology-config-option-${config.metadata.name}`,
+      })),
+    [filteredConfigs],
+  );
+
+  if (filteredConfigs.length === 0) {
+    return null;
+  }
```
emilys314: This is risky code to hide the field. Although this shouldn't happen because it shouldn't be able to be selected if there were no configs, I think it's a lot safer to show the field but have a warning and disable the field saying that none were found

emilys314: Also a really nice pattern I like is to take into account if there is a `value` set.

For example on editing, if there were 0 configs here, BUT there is an existing value, we can still show the entire field. (This also means that we should create a new option just for an existing value that doesn't line up with a value not found. For reference, i did my best to follow these edge cases with the GatewayField)

packages/llmd-serving/src/types.ts
```diff
@@ -129,7 +129,7 @@ export type LLMInferenceServiceConfigKind = K8sResourceCommon & {
       'serving.kserve.io/well-known-config'?: 'true' | 'false';
     };
     labels?: {
-      'opendatahub.io/config-type'?: TopologyType | 'router' | 'accelerator';
+      'opendatahub.io/config-type'?: string;
```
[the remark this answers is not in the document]
emilys314: +1

packages/llmd-serving/src/wizardFields/CustomTopologyConfigField.tsx
```diff
@@ -0,0 +1,204 @@
...
+
+const topologyTypeValues: string[] = Object.values(TopologyType);
+const isTopologyTypeFieldData = (data: unknown): data is TopologyTypeFieldData => {
+  if (data == null || typeof data !== 'object' || !('topologyType' in data)) {
+    return false;
+  }
+  // Safe narrowing: 'in' check above guarantees the property exists
+  // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
+  const val = (data as { topologyType: unknown }).topologyType;
+  return typeof val === 'string' && topologyTypeValues.includes(val);
```
emilys314: This is very close to not needing the `as`, I'd like to have no `as` overrides in the code base for maximum type safety, unless it's truly necessary

packages/llmd-serving/src/wizardFields/TopologyTypeField.tsx
```diff
@@ -0,0 +1,171 @@
...
+};
+
+export type TopologyTypeFieldType = WizardField<TopologyTypeFieldData, TopologyTypeExternalData>;
+
+// --- Component ---
+
+const TopologyTypeFieldComponent: TopologyTypeFieldType['component'] = ({
+  value,
+  onChange,
+  externalData,
```
emilys314, line 91: Sorry if this seems arbitrary, but it would be good to surface the `externalData.error`

I'm picturing a scenario where a user can't see any configs due to an rbac issue, but isn't aware of it. So we get a vague bug or question about why it isn't being shown. So surfacing the error (as warning text?) should help prevent that in our future

emilys314 (commented):
Also this type change would be good to revert https://github.com/opendatahub-io/odh-dashboard/pull/8312#discussion_r3483650304

packages/llmd-serving/src/wizardFields/CustomTopologyConfigField.tsx
emilys314, line 1: I don't seem to be able to see the custom config after selecting the enabled topology

Image

YuliaKrimerman: fixed 
Screenshot 2026-06-29 at 10 20 21 AM

frontend/src/__mocks__/mockLLMInferenceServiceConfigK8sResource.ts
```diff
@@ -56,10 +56,11 @@ export const mockLLMInferenceServiceConfigK8sResource = ({
         : {}),
       ...(preInstalled ? { 'serving.kserve.io/well-known-config': 'true' as const } : {}),
     },
+    // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
     labels: {
       'opendatahub.io/config-type': topologyType ?? configType,
       ...(!preInstalled ? { 'opendatahub.io/dashboard': 'true' as const } : {}),
-    },
+    } as LLMInferenceServiceConfigKind['metadata']['labels'],
```
andrewballantyne: Was this necessary? I have to check it out and run through the TS debugger to find out what you're trying to get around -- `as` is a nasty keyword.

frontend/src/__mocks__/mockLLMInferenceServiceConfigK8sResource.ts
```diff
@@ -8,8 +8,8 @@ type MockLLMInferenceServiceConfigType = {
   name?: string;
   namespace?: string;
   displayName?: string;
-  configType?: TopologyType | 'router' | 'accelerator';
-  topologyType?: TopologyType;
+  configType?: string;
+  topologyType?: string;
```
andrewballantyne: I'd like a good explanation why to reduce types -- this reduces precision -- this feels wrong on paper.

YuliaKrimerman:
Updated the follow-up [story](https://redhat.atlassian.net/jira/software/c/projects/RHOAIENG/boards/12020?selectedIssue=[RHOAIENG-70583](https://redhat.atlassian.net/browse/RHOAIENG-70583)) to include the tests removed from here.

andrewballantyne (commented):
Oh boo -- forgot to hit submit when I did this earlier in the day lol

packages/llmd-serving/src/wizardFields/CustomTopologyConfigField.tsx
```diff
@@ -0,0 +1,205 @@
...
+            options={options}
+            onChange={(key, isPlaceholder) => {
+              if (!key || isPlaceholder) {
+                onChange({ selectedConfig: undefined });
+                return;
+              }
+              const config = filteredConfigs.find((c) => c.metadata.name === key);
+              onChange({ selectedConfig: config ?? value?.selectedConfig });
+            }}
+            placeholder="Select custom configuration (optional)"
```
andrewballantyne: Image
Single-node should have "default" selected; it's not "optional" per se -- this is our LLM-D as it is today in 3.4.

---

Image
Everything not "single-node" needs to support:

- Disabled "topology type" entry when no configs inside it
- If you select one that is not disabled, because it has at least one config -- we need to not say it's "optional" because it's not -- it won't deploy and @vconzola wants to avoid that state
  - I would imagine you can just configure it to be select `[0]` in that case, and let them change -- or we can make it required if Vince wants that

---

Image

- And on the note of disabled topology type -- I think we wanted to have a tooltip explaining why it's disabled

vconzola: > If you select one that is not disabled, because it has at least one config -- we need to not say it's "optional" because it's not -- it won't deploy and @vconzola wants to avoid that state
I would imagine you can just configure it to be select [0] in that case, and let them change -- or we can make it required if Vince wants that

Yes, Topology config should always be required.

frontend/src/__mocks__/mockLLMInferenceServiceConfigK8sResource.ts
YuliaKrimerman: The `as` on the labels object is needed because the return type LLMInferenceServiceConfigKind expects the strict union type on the label but the mock accepts string input. I added an explaining comment for that

YuliaKrimerman: I will remove these mocks changes anyways because I will do the cypress testing for these field in the in the next story

packages/llmd-serving/src/wizardFields/CustomTopologyConfigField.tsx
```diff
@@ -0,0 +1,235 @@
...
+            options={options}
+            onChange={(key, isPlaceholder) => {
+              if (!key || isPlaceholder) {
+                onChange({ selectedConfig: undefined });
+                return;
+              }
+              const config = filteredConfigs.find((c) => c.metadata.name === key);
+              onChange({ selectedConfig: config ?? value?.selectedConfig });
+            }}
+            placeholder={isSingleNode ? 'Select configuration (optional)' : 'Select configuration'}
```
andrewballantyne: There is no "optional" in single node -- it's a default...

Image

packages/llmd-serving/src/wizardFields/CustomTopologyConfigField.tsx
```diff
@@ -0,0 +1,235 @@
...
+  // For single-node without configs, hide the field entirely (it uses existing llm-d behavior)
+  if (isSingleNode && noConfigsAvailable) {
+    return null;
+  }
+
+  return (
+    <FormGroup
+      fieldId="custom-topology-config"
+      label="Custom topology configurations"
+      isRequired={!isSingleNode}
```
andrewballantyne: I think you can say this is always required -- Vince's Prototype has a better "default option" for Single Node than what you have -- see my previous comment for a screenshot.

packages/llmd-serving/src/wizardFields/CustomTopologyConfigField.tsx
```diff
@@ -0,0 +1,235 @@
...
+        description: existingSelection.metadata.annotations?.['openshift.io/description'],
+        dataTestId: `topology-config-option-${existingSelection.metadata.name}`,
+      });
+    }
+
+    return result;
+  }, [filteredConfigs, existingSelection]);
+
+  // For single-node without configs, hide the field entirely (it uses existing llm-d behavior)
+  if (isSingleNode && noConfigsAvailable) {
```
andrewballantyne: Single node is a config in its own right -- you need an extra option here

Image

Single Node **IS** LLM-D in 3.4 -- above is your PR rendering, below is what you need...

Image

You need an extra option specifically for Single Node.

packages/llmd-serving/src/wizardFields/CustomTopologyConfigField.tsx
```diff
@@ -0,0 +1,235 @@
...
+    return null;
+  }
+
+  return (
+    <FormGroup
+      fieldId="custom-topology-config"
+      label="Custom topology configurations"
```
andrewballantyne:
```suggestion
      label="Topology configuration"
```

andrewballantyne (changes requested):

YuliaKrimerman:
Screenshot 2026-07-01 at 10 09 32 AM
Screenshot 2026-07-01 at 10 10 40 AM
@mturley Latest commit changes since Andrew's last review cc @vconzola

packages/model-serving/src/components/deploymentWizard/steps/ModelDeploymentStep.tsx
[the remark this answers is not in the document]
mturley: Seems like it makes sense to keep this out of scope for this PR but we should maybe make sure we open a followup tech debt issue and maybe a TODO comment?

packages/llmd-serving/src/wizardFields/CustomTopologyConfigField.tsx
```diff
@@ -0,0 +1,247 @@
...
+const isActive = (wizardState: RecursivePartial<WizardFormData['state']>): boolean => {
+  if (!isLLMInferenceServiceActive(wizardState)) {
+    return false;
+  }
+  return wizardState.deploymentMethod?.method === LLMD_DEPLOYMENT_METHOD_KEY;
+};
+
+// --- Field definition ---
+
+export const CustomTopologyConfigFieldWizardField: CustomTopologyConfigFieldType = {
```
mturley, line 225: CustomTopologyFieldType doesn’t have a zod `validationSchema` (but `TopologyTypeField` does). If there’s some problem in a future refactor with the useEffect that pre-selects this value, nothing gates submitting without this required value. Not a blocker for this PR but maybe worth a followup? Not sure if it’s serious enough.

packages/llmd-serving/src/wizardFields/CustomTopologyConfigField.tsx
```diff
@@ -0,0 +1,247 @@
...
+                onChange({ selectedConfig: undefined });
+                return;
+              }
+              const config = filteredConfigs.find((c) => c.metadata.name === key);
+              onChange({ selectedConfig: config ?? value?.selectedConfig });
+            }}
+            placeholder="Select configuration"
+            value={selectedValue}
+            dataTestId="custom-topology-config-select"
+            isDisabled={!isLoaded || hasLoadError || noConfigsAvailable}
```
mturley, line 166: There’s no loading indicator while configurations are being loaded, we should maybe render a skeleton instead of a disabled field while `!isLoaded` like I did for the NIM image field

mturley (commented):
A couple small things on initial review, none showstoppers. Will continue with a deeper review when I can.

mturley:
/approve

Feedback from others appears addressed and my comments aren’t blocking. Approving to keep things moving and then I’ll follow up when I get a chance to see if there’s more to address.

PR #8324, opened 2026-06-29
fix(dashboard-operator): render webhook certs without Helm Capabilities check

liday-rh:
## Description

Fixes the odh-operator E2E failure described in [opendatahub-operator#3656 (comment)](https://github.com/opendatahub-io/opendatahub-operator/pull/3656#issuecomment-4834136001).

When the module handler renders the dashboard-operator Helm chart via `k8s-manifest-kit/renderer-helm`, default Helm `Capabilities` do not include `cert-manager.io/v1`. The chart gated Issuer/Certificate creation on `.Capabilities.APIVersions.Has "cert-manager.io/v1"` but still rendered the Deployment with a `webhook-certs` volume mount when `webhook.enabled=true`. Result: pod stuck in `Pending` waiting for secret `dashboard-operator-webhook-tls`.

**Change:** Remove the redundant Capabilities guard; cert-manager resource creation now depends only on `webhook.certManager.enabled`, which the odh-operator handler already sets based on actual cluster CRD availability.

## How Has This Been Tested?

```bash
cd dashboard-operator
make test
make chart-validate
```

Added `TestDashboardChartWebhookCertManagerTemplatesOffline` to assert offline `helm template` output includes Certificate, Issuer, and the webhook-certs volume/secret.

## Test Impact

- New unit test: `TestDashboardChartWebhookCertManagerTemplatesOffline` in `dashboard-operator/charts/chart_sync_test.go`
- Existing chart sync and controller tests unchanged and passing

## Request review criteria:

**Follow-up (odh-operator):** After merge, bump the `dashboard-operator` chart SHA in `get_all_manifests.sh` and retrigger E2E on opendatahub-operator#3656.

## Summary by CodeRabbit

* **New Features**
  * Dashboard chart templates now consistently render webhook certificate resources when cert-manager is enabled.
  * Added offline chart rendering coverage to verify the expected cert-manager webhook settings are included.

* **Tests**
  * Expanded chart tests to validate rendered Helm output for webhook certificate resources and secret configuration.
  * Streamlined chart rendering helpers to make template checks more reliable.

### DIFF
--- a/dashboard-operator/charts/chart_sync_test.go
+++ b/dashboard-operator/charts/chart_sync_test.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 	"testing"
 
+	appsv1 "k8s.io/api/apps/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	"k8s.io/apimachinery/pkg/util/yaml"
 )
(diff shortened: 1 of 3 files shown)


lucferbux (approved):
/lgtm
/approve

Reviewed the fix against the [E2E failure analysis](https://github.com/opendatahub-io/opendatahub-operator/pull/3656#issuecomment-4834136001). The root cause was that `k8s-manifest-kit/renderer-helm` renders with `nil` Capabilities (defaults), which don't include `cert-manager.io/v1`. This caused the old `{{- if and .Values.webhook.certManager.enabled (.Capabilities.APIVersions.Has "cert-manager.io/v1") }}` condition to skip Certificate/Issuer rendering, while the Deployment still mounted the `webhook-certs` volume → pod stuck in `Pending`.

The fix is correct:
- **Template**: Removes the redundant `.Capabilities.APIVersions.Has` guard — the operator module handler already gates `webhook.certManager.enabled=true` on actual CRD availability (`CertManagerCRDsAvailable` check), so the Capabilities check was duplicate and broken for offline rendering.
- **Default change to `false`**: Protects standalone `helm install` users who don't have cert-manager. For live `helm install`, Capabilities ARE populated from the real API server, so the old guard was only useful as a safety net — and the new default provides a better one.
- **Test coverage**: New `TestDashboardChartWebhookCertManagerTemplatesOffline` validates both the positive case (cert-manager resources render in offline `helm template` when enabled) and the negative case (omitted when disabled). Also verifies the Deployment's webhook-certs volume references the same secret as the Certificate.

No regressions found. The `renderChartTemplate` refactor preserves the existing `TestDashboardChartRBACInSync` behavior.

PR #8334, opened 2026-06-30
RHOAIENG-66139: Implement Edit and Duplicate role actions in Roles tab

srtanish1992:
https://redhat.atlassian.net/browse/RHOAIENG-66139

## Description

Implements Edit and Duplicate actions for custom roles in the project Roles tab.

Previously, the "Edit role" and "Duplicate role" kebab actions in the Roles table were no-ops. This PR wires them up end-to-end:

https://github.com/user-attachments/assets/196e2073-4df1-44ff-ab07-ac9dda0ee9aa

**Edit:**
- Navigates to `/projects/:namespace/roles/:roleName/edit`
- Fetches the existing role and pre-populates the form (name, description, labels, rules)
- Submits via `k8sUpdateResource` (PUT) preserving `resourceVersion` for optimistic concurrency
- RBAC checks the `update` verb on roles

**Duplicate:**
- Navigates to `/projects/:namespace/roles/:roleName/duplicate`
- Fetches the source role and pre-populates the form with:
  - Display name: `"Copy of {originalName}"` (consistent with connection types, serving runtimes, etc.)
  - K8s name: cleared — user must provide a unique name
  - Rules and labels: copied from the source role
- Submits via `createRole` (POST) — creating a new resource
- RBAC checks the `create` verb on roles

**API layer:**
- Added `updateRole` function using `k8sUpdateResource` (PUT semantics) to `frontend/src/api/k8s/roles.ts`

## How Has This Been Tested?

- Manual testing: verified Edit and Duplicate flows against a live cluster with the `roleManagement` feature flag enabled
- Verified RBAC redirects when user lacks `update`/`create` permissions
- Verified ClusterRole entries remain disabled for Edit/Duplicate (unchanged behavior)
- TypeScript type-check passes (`tsc --noEmit`)
- ESLint passes with `--max-warnings 0`
- Existing unit tests pass

## Test Impact

- **Unit tests:** Added `updateRole` tests to `frontend/src/api/k8s/__tests__/roles.spec.ts` (also fixed pre-existing mock accumulation bug with `beforeEach(jest.clearAllMocks)`)
- **Cypress mock tests:** Added two new test files:
  - `rolesTabEdit.cy.ts` — edit page load, PUT submission, error handling, table kebab navigation
  - `rolesTabDuplicate.cy.ts` — duplicate page load, cleared name, POST submission, copied rules, error handling, table kebab navigation
  
- **Cypress page object:** Added `visitEditRole` and `visitDuplicateRole` helpers to `projectRoles.ts`

## Request review criteria:

If you have UI changes: 

## Summary by CodeRabbit

* **New Features**
  * Added project-scoped routes and navigation for editing and duplicating custom roles, including a pre-filled duplicate flow.
  * Enhanced role creation with a “duplicate” mode and template selection that uses “replace current content” confirmation.
* **Bug Fixes**
  * Improved role update behavior to correctly apply required update options and handle create/edit submission differences.
* **Tests**
  * Added Cypress end-to-end coverage for edit and duplicate role workflows; updated modal helper logic for replace-content flow and expanded unit tests for role updates and the new confirmation modal.


frontend/src/pages/projects/projectRoles/EditRolePage.tsx
```diff
@@ -0,0 +1,63 @@
...
+    setLoadError(undefined);
+
+    getRole(namespace, roleName)
+      .then((data) => {
+        if (!cancelled) {
+          setRole(data);
+          setLoaded(true);
+        }
+      })
+      .catch((e) => {
```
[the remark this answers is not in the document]
srtanish1992: The page already has a top-level SSAR gate (useAccessReview for update/create verb) that redirects unauthorized users before the form renders. Distinguishing a 403 on get is a separate concern — other similar pages in the codebase (hardware profile duplicate, connection type duplicate) also show generic error pages for load failures.

frontend/src/pages/projects/projectRoles/RolesTable.tsx
```diff
@@ -115,8 +117,10 @@ const RolesTable: React.FC<RolesTableProps> = ({
             row={row}
             onViewDetails={() => setDetailsRoleRef(row.roleRef)}
             onPreviewYAML={() => setPreviewRow(row)}
-            onEdit={() => undefined}
-            onDuplicate={() => undefined}
+            onEdit={() => navigate(`/projects/${namespace}/roles/${row.roleRef.name}/edit`)}
```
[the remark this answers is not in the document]
srtanish1992: PatternFly's ActionsColumn items use onClick callbacks — they don't support a component prop the way Button does. This is consistent with how other areas in the codebase handle kebab navigation actions (hardware profiles, connection types, serving runtimes all use navigate() for duplicate/edit kebab actions). Refactoring to a link-based menu item would require custom PatternFly overrides that aren't established patterns in this project. Keeping navigate() here for consistency.

frontend/src/pages/projects/projectRoles/CreateRolePage.tsx
```diff
@@ -183,7 +186,17 @@ const CreateRolePage: React.FC<CreateRolePageProps> = ({ existingRole }) => {
...
+            ...role.metadata,
+            resourceVersion: existingRole.metadata.resourceVersion,
+          },
+        });
+      } else {
+        await createRole(role);
+      }
       navigate(`/projects/${namespace}?section=roles`);
     } catch (e) {
       const error = e instanceof Error ? e : new Error('Failed to create role');
```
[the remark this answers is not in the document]
srtanish1992: Fixed.

srtanish1992:
@coderabbitai review

frontend/src/pages/projects/projectRoles/DuplicateRolePage.tsx
```diff
@@ -0,0 +1,81 @@
...
+        if (!cancelled) {
+          setLoadError(e instanceof Error ? e : new Error('Failed to load role'));
+          setLoaded(true);
+        }
+      });
+
+    return () => {
+      cancelled = true;
+    };
+  }, [namespace, roleName]);
```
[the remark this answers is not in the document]
srtanish1992: Thanks for the suggestion. While both EditRolePage and DuplicateRolePage share a similar fetch pattern, each is only ~27 lines and they diverge in how they transform the fetched role (edit passes it directly; duplicate strips identity fields and prepends "Copy of"). Per project conventions we generally skip DRY refactoring when both copies are short, self-contained, and have different post-processing logic. Extracting a shared hook here would add indirection without meaningfully improving maintainability. We'll revisit if a third consumer appears.

packages/cypress/cypress/tests/mocked/projects/tabs/rolesTabDuplicate.cy.ts
```diff
@@ -0,0 +1,125 @@
...
+
+  it('should pre-populate the display name with "Copy of" prefix', () => {
+    projectRoles.visitDuplicateRole(NAMESPACE, SOURCE_ROLE_NAME);
+    projectRoles.findRoleNameInput().should('have.value', `Copy of ${SOURCE_ROLE_NAME}`);
+  });
+
+  it('should have submit button enabled with auto-generated resource name', () => {
+    projectRoles.visitDuplicateRole(NAMESPACE, SOURCE_ROLE_NAME);
+    projectRoles.findSubmitButton().should('be.enabled');
+  });
```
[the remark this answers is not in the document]
srtanish1992: Renamed test from should have submit button enabled with auto-generated resource name to should have submit button enabled when form is pre-populated.

packages/cypress/cypress/tests/mocked/projects/tabs/rolesTabEdit.cy.ts
```diff
@@ -0,0 +1,118 @@
...
+      expect(interception.request.body.metadata.namespace).to.equal(NAMESPACE);
+      expect(interception.request.body.metadata.labels).to.have.property(
+        'opendatahub.io/dashboard',
+        'true',
+      );
+    });
+
+    cy.url().should('include', `/projects/${NAMESPACE}`);
+    cy.url().should('include', 'section=roles');
+  });
```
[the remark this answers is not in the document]
srtanish1992: expect(interception.request.body.metadata.resourceVersion).to.equal('12345') assertion to the edit PUT Cypress test.

frontend/src/pages/projects/projectRoles/CreateRolePage.tsx
```diff
@@ -41,13 +41,16 @@ type TemplateModalState =
 
 type CreateRolePageProps = {
   existingRole?: RoleKind;
+  duplicateRole?: RoleKind;
 };
 
-const CreateRolePage: React.FC<CreateRolePageProps> = ({ existingRole }) => {
+const CreateRolePage: React.FC<CreateRolePageProps> = ({ existingRole, duplicateRole }) => {
   const { namespace = '' } = useParams<{ namespace: string }>();
   const { currentProject } = React.useContext(ProjectDetailsContext);
```
DaoDaoNoCode, line 49: Hey — heads up on a subtle interaction with the template selector in edit/duplicate mode.

`isFormDirty` (line 84) compares against empty defaults (`name !== ""`, `rules.length > 0`). Since `initialRole` pre-populates all these fields, the form is "dirty" from the moment it loads. Two effects:

1. Clicking "Select role template" always shows the "Discard changes?" dialog — even if the user hasn't touched anything yet.
2. If they confirm the discard, `handleDiscardConfirm` (line 112) resets everything to blank — not to the initial role values. On an edit page this wipes out the pre-populated data entirely.

**Scenario:** User opens Edit page → clicks "Select role template" → sees unexpected "discard?" dialog → confirms → form is now completely empty. The original role data is gone from the form.

This isn't a blocker for the edit/duplicate happy path, but it's a real UX bug for anyone who interacts with the template selector on edit/duplicate. Consider tracking `initialRole` values and comparing against those for the dirty check, and resetting to initial values instead of blank in `handleDiscardConfirm`.

DaoDaoNoCode (commented):
Reviewed the diff against the [Jira AC](https://redhat.atlassian.net/browse/RHOAIENG-66139), the actual source files (not just the diff), and existing codebase patterns. Skipped findings already raised in prior threads (error message wording, navigate vs Link, duplicate fetch logic). 4 new validated items below.

packages/cypress/cypress/tests/mocked/projects/tabs/rolesTabEdit.cy.ts
```diff
@@ -0,0 +1,123 @@
...
+    projectRoles.findRoleNameInput().should('have.value', ROLE_NAME);
+  });
+
+  it('should show "Edit custom role" as page title', () => {
+    projectRoles.visitEditRole(NAMESPACE, ROLE_NAME);
+    cy.findByTestId('app-page-title').should('contain.text', 'Edit custom role');
+  });
+
+  it('should show "Save changes" as submit button text', () => {
+    projectRoles.visitEditRole(NAMESPACE, ROLE_NAME);
```
DaoDaoNoCode, line 77: This test visits the edit page and immediately clicks Submit without modifying any fields. It verifies the PUT payload has the original `name`, `namespace`, `resourceVersion`, and `labels` — which is great for proving the round-trip works.

But it doesn't test what the feature is actually for — *editing* a role. If a regression silently discarded user changes before submission (e.g., the form state didn't bind correctly to the assembled role), this test would still pass.

Worth adding a test that clears the display name, types a new one, submits, and asserts the PUT payload's `openshift.io/display-name` annotation reflects the change. Same for modifying a rule or adding a label.

frontend/src/pages/projects/projectRoles/DuplicateRolePage.tsx
```diff
@@ -0,0 +1,81 @@
...
+      <Bullseye>
+        <Spinner />
+      </Bullseye>
+    );
+  }
+
+  if (loadError || !role) {
+    if (loadError) {
+      return (
+        <ApplicationsPage loaded empty loadError={loadError} errorMessage="Unable to load role" />
```
DaoDaoNoCode, line 55: This error handling path (and the identical one in `EditRolePage.tsx`) is untested in Cypress. Neither test file intercepts `GET` with a 404 to exercise the `loadError` rendering.

A user following a stale bookmark to a deleted role would hit this code path. A regression in error rendering (e.g., accidentally removing the `loaded` prop) would go undetected.

A quick test that intercepts `GET /role` with `{ statusCode: 404, body: mock404Error({...}) }` and asserts `"Unable to load role"` appears would close this gap.

packages/cypress/cypress/tests/mocked/projects/tabs/rolesTabEdit.cy.ts
```diff
@@ -0,0 +1,123 @@
...
+    asProjectAdminUser();
+    initIntercepts();
+  });
+
+  it('should load the edit page and show existing role name', () => {
+    projectRoles.visitEditRole(NAMESPACE, ROLE_NAME);
+    projectRoles.findRoleNameInput().should('have.value', ROLE_NAME);
+  });
+
+  it('should show "Edit custom role" as page title', () => {
```
DaoDaoNoCode, line 71: This test checks `findRoleNameInput().should('have.value', ROLE_NAME)` — it verifies the *value* is pre-populated, but not that the field is *read-only*.

The k8s resource name should be immutable during edit (since K8s names can't change after creation). `useK8sNameDescriptionFieldData` sets `immutable = true` when the initial name is non-empty, and `ResourceNameField` renders plain text instead of an input when immutable.

Conversely, the duplicate tests don't verify the k8s name field *is* editable (the duplicate transform clears `metadata.name` to `''` so the field should be an input).

If someone accidentally kept the name populated in the duplicate transform, the field would silently lock and the user would be stuck with the original name. Worth asserting the field's mutability state on both pages.

frontend/src/pages/projects/projectRoles/CreateRolePage.tsx
srtanish1992: Hey @DaoDaoNoCode - 
 
The "Discard unsaved changes?" modal is intentional — our templates contain all form data (name, description, rules), so selecting a template in edit/duplicate mode genuinely replaces the pre-populated values. The discard-to-blank + apply-template flow is functionally correct. However, the modal body text was create-mode-specific ("start creating a new role"), so I've updated it to a mode-agnostic message: "Your unsaved changes will be replaced by the selected template. Do you want to continue?"

Let me know if it makes sense to you.

packages/cypress/cypress/tests/mocked/projects/tabs/rolesTabEdit.cy.ts
srtanish1992: Added — now modifies the display name and asserts the PUT payload's annotation reflects the change.

frontend/src/pages/projects/projectRoles/DuplicateRolePage.tsx
srtanish1992: Added for both Edit and Duplicate pages.

packages/cypress/cypress/tests/mocked/projects/tabs/rolesTabEdit.cy.ts
srtanish1992: edit mode asserts role-resourceName input does not exist (immutable), duplicate mode asserts it does exist (editable).

srtanish1992:
The new flow is - 

1. User clicks "Select role template"
2. Template selector opens directly
3. User picks a template
4. Then show the confirmation ("Replacing current content with {template name}. Continue?")

Also, we preserve the labels since they are user defined.

DaoDaoNoCode (approved):
/lgtm

PR #8348, opened 2026-06-30
Show workspace context in MLflow breadcrumbs

Taj010:
## Description

For https://redhat.atlassian.net/browse/RHOAIENG-65809

When users drill into an MLflow experiment or prompt, the breadcrumb only shows the MLflow path (e.g., `Experiments > experiments-1`). There was no indication of which workspace/project they were in unless they checked the URL. 

This change adds a workspace link to the MLflow breadcrumb, matching the existing pipelines pattern in `PipelineContextBreadcrumb`:
- A vertical divider after the breadcrumb trail
- `Go to [project icon] <project name>` linking to `/projects/<workspace>`
- Uses `getDisplayNameFromK8sResource` from `@odh-dashboard/k8s-core` for the display name
- Adds `MlflowBreadcrumbs.scss` with the same underline workaround used in `GlobalPipelineCoreDetails.scss` 

image

## How Has This Been Tested?

This has been manually tested: 
- Open MLflow Experiments in the workspace
- Confirm the breadcrumb shows `Go to <workspace A name>`.

## Test Impact

No new unit or Cypress tests added. 

## Request review criteria:

If you have UI changes:

## Summary by CodeRabbit

* **New Features**
  * Added a “Go to” project link to the MLflow breadcrumbs for quicker navigation to the current workspace’s project.
  * Updated breadcrumbs layout to better support the added navigation element.
* **Bug Fixes**
  * Improved breadcrumb link styling so the underline consistently spans the full breadcrumb text and icon area.

### DIFF
--- a/packages/mlflow-embedded/package.json
+++ b/packages/mlflow-embedded/package.json
@@ -45,7 +45,7 @@
   },
   "dependencies": {
     "@odh-dashboard/internal": "*",
-    "@odh-dashboard/k8s-core": "^0.0.0",
+    "@odh-dashboard/k8s-core": "*",
     "@odh-dashboard/plugin-core": "*"
   },
   "devDependencies": {
--- a/packages/mlflow-embedded/package.json
+++ b/packages/mlflow-embedded/package.json
@@ -45,6 +45,7 @@
   },
   "dependencies": {
     "@odh-dashboard/internal": "*",
+    "@odh-dashboard/k8s-core": "^0.0.0",
     "@odh-dashboard/plugin-core": "*"
   },
   "devDependencies": {
--- a/packages/mlflow-embedded/shared/MlflowBreadcrumbs.scss
+++ b/packages/mlflow-embedded/shared/MlflowBreadcrumbs.scss
@@ -1,0 +1,18 @@
+// Workaround so the link underline spans the full text+icon row.
+// Mirrors the same fix in GlobalPipelineCoreDetails.scss (PF issue #7554).
+.mlflow-breadcrumb-project-link {
+  position: relative;
+  text-decoration: none;
+
+  &::after {
+    content: '\00A0';
+    position: absolute;
+    text-decoration: underline;
+    right: 0;
+    left: 0;
+    bottom: 0;
+    top: 2px;
+    letter-spacing: 1000px;
+    overflow: hidden;
+  }
+}
--- a/packages/mlflow-embedded/shared/MlflowBreadcrumbs.scss
+++ b/packages/mlflow-embedded/shared/MlflowBreadcrumbs.scss
@@ -1,5 +1,5 @@
 // Workaround so the link underline spans the full text+icon row.
-// Mirrors the same fix in GlobalPipelineCoreDetails.scss (PF issue #7554).
+// FIXME: Upstream still open — https://github.com/patternfly/patternfly/issues/7554 (breadcrumb icon underline). Re-check after PF ships fix.
 .mlflow-breadcrumb-project-link {
   position: relative;
   text-decoration: none;
--- a/packages/mlflow-embedded/shared/MlflowBreadcrumbs.tsx
+++ b/packages/mlflow-embedded/shared/MlflowBreadcrumbs.tsx
@@ -29,10 +29,10 @@ const MlflowBreadcrumbs: React.FC<{
 }> = ({ basePath, workspace, breadcrumbs }) => {
   const { projects } = React.useContext(ProjectsContext);
   const project = projects.find(byName(workspace));
-  const projectDisplayName = project ? getDisplayNameFromK8sResource(project) : workspace;
+  const projectDisplayName = project && getDisplayNameFromK8sResource(project);
 
   return (
-    <Flex>
+    <Flex alignItems={{ default: 'alignItemsCenter' }}>
       <Breadcrumb>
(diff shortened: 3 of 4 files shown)


packages/mlflow-embedded/package.json
```diff
@@ -45,6 +45,7 @@
   },
   "dependencies": {
     "@odh-dashboard/internal": "*",
+    "@odh-dashboard/k8s-core": "^0.0.0",
```
DaoDaoNoCode: Should it be `"@odh-dashboard/k8s-core": "*",`?

packages/mlflow-embedded/shared/MlflowBreadcrumbs.scss
```diff
@@ -0,0 +1,18 @@
+// Workaround so the link underline spans the full text+icon row.
+// Mirrors the same fix in GlobalPipelineCoreDetails.scss (PF issue #7554).
```
DaoDaoNoCode: nit: The other copy of this workaround in `FeatureStoreBreadcrumb.scss` includes the full upstream URL and uses a `FIXME` marker so it's easy to grep for when the PF fix ships. Might be worth matching that style here.

```suggestion
// FIXME: Upstream still open — https://github.com/patternfly/patternfly/issues/7554 (breadcrumb icon underline). Re-check after PF ships fix.
```

packages/mlflow-embedded/shared/MlflowBreadcrumbs.tsx
```diff
@@ -11,43 +26,71 @@ const MlflowBreadcrumbs: React.FC<{
...
-          }
-        />
-      );
-    })}
-  </Breadcrumb>
-);
+}> = ({ basePath, workspace, breadcrumbs }) => {
+  const { projects } = React.useContext(ProjectsContext);
+  const project = projects.find(byName(workspace));
+  const projectDisplayName = project ? getDisplayNameFromK8sResource(project) : workspace;
```
DaoDaoNoCode: nit: The fallback to `workspace` is dead code since the project link only renders when `project` is truthy (line 71). You could simplify this by computing the display name inline where it's used, or just drop the fallback and make the variable conditional:

```suggestion
  const projectDisplayName = project && getDisplayNameFromK8sResource(project);
```

Then use `projectDisplayName` (or just inline it) inside the conditional block.

packages/mlflow-embedded/shared/MlflowBreadcrumbs.tsx
```diff
@@ -11,43 +26,71 @@ const MlflowBreadcrumbs: React.FC<{
...
+
+  return (
+    <Flex>
```
DaoDaoNoCode: nit: The outer `Flex` doesn't specify alignment or gap. On narrow viewports the breadcrumb and project link could wrap awkwardly. The pipeline version doesn't set these either so it's consistent, but it might be worth adding `alignItems={{ default: 'alignItemsCenter' }}` on this outer Flex to keep the divider vertically centered with the breadcrumb text.

DaoDaoNoCode (commented):
This adds a workspace project link to the MLflow breadcrumb trail, mirroring the existing PipelineContextBreadcrumb pattern. The structure, PatternFly usage, and CSS workaround are all consistent with the pipeline version. Looks good overall -- left a few comments inline, mostly nits. Main thing to consider is adding the FIXME URL to the SCSS comment.

Local checks: lint could not run locally (Node 20 vs required 22), but CI lint, type-check, and all tests pass.

Prior review: DaoDaoNoCode asked about the `k8s-core` version specifier (`"^0.0.0"` vs `"*"`) -- fix verified in code, latest commit uses `"*"`.

packages/mlflow-embedded/shared/MlflowBreadcrumbs.scss
DaoDaoNoCode: Nit: the feature-store copy of this workaround uses a `FIXME` marker with the full PF issue URL, making it greppable when PF ships the fix. Consider matching that style:
```scss
// FIXME: workaround for https://github.com/patternfly/patternfly/issues/7554
```

packages/mlflow-embedded/shared/MlflowBreadcrumbs.tsx
DaoDaoNoCode: Nit: `projectDisplayName` falls back to `workspace` when `project` is falsy, but the JSX below is gated on `{project && (...)}`, so the fallback can never be reached. You could simplify to:
```tsx
const projectDisplayName = getDisplayNameFromK8sResource(project);
```
(moved inside the conditional block), or keep the current shape if you plan to show the workspace name even when the project isn't found.

packages/mlflow-embedded/shared/MlflowBreadcrumbs.tsx
DaoDaoNoCode: Nit: the outer `<Flex>` doesn't specify `alignItems`. The vertical `<Divider>` may not center-align with the breadcrumb text on all viewport sizes. Consider adding:
```tsx
<Flex alignItems={{ default: 'alignItemsCenter' }}>
```
to match the inner Flex.

DaoDaoNoCode (approved):
/lgtm
Thanks for the fix!

PR #8353, opened 2026-06-30
fix: Enable deployment resource auto-selection for generative LLMInferenceService deployments

YuliaKrimerman:
https://issues.redhat.com/browse/RHOAIENG-58544

## Description

Fixes the deployment resource auto-selection being permanently disabled for generative models using the "LLM inference service deployment" method (LLMInferenceService + LLMInferenceServiceConfig).

**Root cause:** `LLMConfigOptionsField.getInitialFieldData` in `packages/llmd-serving` had a gap in its suggestion logic. When only one `LLMInferenceServiceConfig` was available, it would pre-select it in the manual dropdown but never set the `suggestion` property — which the `ModelServerTemplateSelectField` auto-select radio checks (`isDisabled={!data?.suggestion}`). This caused the "Automatic selection" radio to always appear disabled.

**Fix:** Extracted a `computeSuggestion` pure function (consistent with KServe's existing `computeSuggestion` pattern) that returns a suggestion when:
- There is exactly one available config (single-option case — the main bug)
- There are multiple configs but exactly one is compatible with the selected hardware profile (already worked, now consolidated)

When a suggestion is found, `getInitialFieldData` now consistently sets `autoSelect: true` and `suggestion`, enabling the auto-select radio.

**Files changed:**
- `packages/llmd-serving/src/wizardFields/LlmConfigOptionsField.tsx` — added `computeSuggestion`, simplified `getInitialFieldData`
- `packages/llmd-serving/src/wizardFields/__tests__/LlmConfigOptionsField.spec.ts` — added 5 unit tests for `getInitialFieldData`

## How Has This Been Tested?

- **Manual testing** on ODH Dashboard local dev (`npm run dev`) against a live cluster:
  - Verified auto-select radio is enabled and pre-selected when a single `LLMInferenceServiceConfig` is available
  - Verified auto-select radio remains correctly disabled when multiple configs exist with no unique hardware profile match
  - Verified manual selection still works as expected
  Before 
Screenshot 2026-06-30 at 3 08 30 PM

  After
Screenshot 2026-06-30 at 3 31 00 PM

- **Unit tests** — all 96 `llmd-serving` tests pass, all 15 `ModelServerTemplateSelectField` tests pass
- **Type-check** — `tsc --noEmit` passes with no errors
- **Lint** — no new warnings or errors introduced

## Test Impact

Added 5 new unit tests for `getInitialFieldData` in `LlmConfigOptionsField.spec.ts`:
- Returns existing field data when provided
- Auto-selects with suggestion when there is only one config
- Auto-selects the single hardware-compatible config among multiple
- Does not auto-select when multiple configs match hardware profile
- Does not auto-select when there are no configs

## Request review criteria:

If you have UI changes:

## Summary by CodeRabbit

* **New Features**
  * Improved LLM configuration selection so the app can suggest and preselect the best option when there is only one choice or one compatible match for the current hardware profile.

* **Bug Fixes**
  * Preserves existing field data when a configuration has already been chosen.
  * Avoids auto-selection when multiple valid options are available or when no options exist.

### DIFF
--- a/packages/cypress/cypress/tests/mocked/modelServing/modelServingLlmd.cy.ts
+++ b/packages/cypress/cypress/tests/mocked/modelServing/modelServingLlmd.cy.ts
@@ -1136,6 +1136,7 @@ describe('Model Serving LLMD', () => {
       modelServingWizard.findModelDeploymentNameInput().type('test-disabled-config');
       modelServingWizard.findDeploymentMethodSelect().should('not.be.disabled');
       modelServingWizard.selectDeploymentMethodByKey('llm-inference-service-simple-vllm');
+      modelServingWizard.findModelServerManualSelectRadio().click();
       modelServingWizard.findServingRuntimeTemplateSearchSelector().click();
 
       // Enabled config should be visible
--- a/packages/llmd-serving/src/wizardFields/LlmConfigOptionsField.tsx
+++ b/packages/llmd-serving/src/wizardFields/LlmConfigOptionsField.tsx
@@ -84,6 +84,19 @@ const getOptions = (
   return result;
 };
 
+const computeSuggestion = (options: ModelServerOption[]): ModelServerOption | undefined => {
+  if (options.length === 1) {
+    return options[0];
+  }
+
+  const compatibleOptions = options.filter((option) => option.compatibleWithHardwareProfile);
+  if (compatibleOptions.length === 1) {
+    return compatibleOptions[0];
+  }
+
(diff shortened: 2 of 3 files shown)


mturley:
Code LGTM, I want to manually test before I approve though
/lgtm

andrewballantyne (approved):
I don't think this is breaking -- but our new navigation in the wizard is actually breaking some features -- I'll go to the channel with them

PR #8355, opened 2026-07-01
feat(feature-store): Add FeatureStore CRUD API helpers and React data-fetching hooks

ntkathole:
https://redhat.atlassian.net/browse/RHOAIENG-61258 

## Description

This PR adds the Kubernetes CRUD API layer and React data-fetching hooks for the FeatureStore CRD to the `@odh-dashboard/feature-store` package. These are the foundational building blocks consumed by the creation wizard, management page, and deployment tracking UI in subsequent PRs.

These API helpers and hooks encapsulate all Kubernetes interactions for the Feature Store feature, providing a clean abstraction layer between the CRD and the UI components. The creation wizard ([RHOAIENG-61260](https://redhat.atlassian.net/browse/RHOAIENG-61260)), management page ([RHOAIENG-61262](https://redhat.atlassian.net/browse/RHOAIENG-61262)), and deployment tracking ([RHOAIENG-61261](https://redhat.atlassian.net/browse/RHOAIENG-61261)) all consume these modules. Shipping them as a standalone PR keeps the diff focused and reviewable.

No UI-visible changes. No new package exports needed — all modules are consumed internally within the `feature-store` package.

## How Has This Been Tested?

- Verified import paths resolve correctly against `@odh-dashboard/k8s-core` and `@odh-dashboard/internal`.

## Test Impact

Added unit tests in `packages/feature-store/src/api/__tests__/featureStores.spec.ts` covering the `assembleFeatureStore` resource builder, all CRUD operations (`create`, `list`, `listAll`, `get`, `delete`), pod listing, and the internal validation/normalization logic (`isValidFeatureStore`, `normalizeFeatureStore`) exercised through the public API. Hook tests are deferred to the Cypress E2E PR ([RHOAIENG-61270](https://redhat.atlassian.net/browse/RHOAIENG-61270)) where they are tested end-to-end with mocked API responses.

## Request review criteria:

If you have UI changes: 

## Summary

- **New Features**
  - Added Feature Store API support (create, list scoped/all, get, delete, and pod lookups).
  - Added hooks for accessible namespaces, existing feature stores, namespace ConfigMaps/secrets, and deployment monitoring with pod log tailing.
  - Added Feast namespace label constants and shared Feature Store assembly/normalization utilities.

- **Bug Fixes**
  - Added validation, normalization (labels/annotations/status defaults), and improved handling of invalid stores and transient pod log errors.

- **Tests**
  - Added Jest/React Testing Library coverage for the API, hooks, utilities, and deployment watching logic.

### DIFF
--- a/packages/feature-store/src/api/featureStores.ts
+++ b/packages/feature-store/src/api/featureStores.ts
@@ -1,0 +1,91 @@
+import {
+  k8sCreateResource,
+  k8sDeleteResource,
+  k8sGetResource,
+  k8sListResource,
+  K8sStatus,
+} from '@openshift/dynamic-plugin-sdk-utils';
+import { K8sAPIOptions } from '@odh-dashboard/internal/k8sTypes';
+import { FeatureStoreModel } from '@odh-dashboard/internal/api/models/odh';
+import { PodModel } from '@odh-dashboard/internal/api/models/k8s';
+import { applyK8sAPIOptions } from '@odh-dashboard/internal/api/apiMergeUtils';
+import { isValidK8sName, PodKind } from '@odh-dashboard/k8s-core';
+import { FeatureStoreKind } from '../k8sTypes';
+import {
+  assembleFeatureStore,
+  normalizeFeatureStore,
+  normalizeFeatureStoreList,
+  FeatureStoreFormSpec,
+} from '../utils/featureStoreUtils';
+
+export { assembleFeatureStore, type FeatureStoreFormSpec } from '../utils/featureStoreUtils';
+
+export const createFeatureStore = async (
+  formData: FeatureStoreFormSpec,
+  opts?: K8sAPIOptions,
+): Promise<FeatureStoreKind> => {
+  const resource = assembleFeatureStore(formData);
+  const created = await k8sCreateResource<FeatureStoreKind>(
+    applyK8sAPIOptions(
+      {
+        model: FeatureStoreModel,
+        resource,
+      },
+      opts,
+    ),
+  );
+  return normalizeFeatureStore(created);
+};
+
+export const listFeatureStores = (namespace: string): Promise<FeatureStoreKind[]> =>
+  k8sListResource<FeatureStoreKind>({
+    model: FeatureStoreModel,
+    queryOptions: {
+      ns: namespace,
+    },
+  }).then((listResource) => normalizeFeatureStoreList(listResource.items));
+
+/**
+ * Lists FeatureStore CRs across all namespaces (cluster-wide).
+ * Requires cluster-level `list` permission on `featurestores.feast.dev`.
+ * Prefer per-namespace {@link listFeatureStores} gated by SSAR for unprivileged users.
+ */
+export const listAllFeatureStores = (): Promise<FeatureStoreKind[]> =>
+  k8sListResource<FeatureStoreKind>({
+    model: FeatureStoreModel,
+    queryOptions: {},
+  }).then((listResource) => normalizeFeatureStoreList(listResource.items));
+
+export const getFeatureStore = (namespace: string, name: string): Promise<FeatureStoreKind> =>
+  k8sGetResource<FeatureStoreKind>({
+    model: FeatureStoreModel,
+    queryOptions: {
+      name,
+      ns: namespace,
+    },
+  }).then(normalizeFeatureStore);
+
+export const deleteFeatureStore = (namespace: string, name: string): Promise<K8sStatus> =>
+  k8sDeleteResource<FeatureStoreKind, K8sStatus>({
+    model: FeatureStoreModel,
+    queryOptions: { name, ns: namespace },
+  });
+
+export const getPodsForFeatureStore = async (
+  namespace: string,
+  featureStoreName: string,
+): Promise<PodKind[]> => {
+  if (!isValidK8sName(featureStoreName)) {
+    throw new Error(
+      `Invalid featureStoreName "${featureStoreName}". Must be a valid DNS-1123 label.`,
+    );
+  }
+  const result = await k8sListResource<PodKind>({
+    model: PodModel,
+    queryOptions: {
+      ns: namespace,
+      queryParams: { labelSelector: `feast.dev/name=${featureStoreName}` },
+    },
+  });
+  return result.items;
+};
--- a/packages/feature-store/src/hooks/__tests__/useAccessibleNamespaces.spec.tsx
+++ b/packages/feature-store/src/hooks/__tests__/useAccessibleNamespaces.spec.tsx
@@ -1,0 +1,134 @@
+import * as React from 'react';
+import { renderHook, waitFor } from '@testing-library/react';
+import { ProjectsContext } from '@odh-dashboard/internal/concepts/projects/ProjectsContext';
+import { checkAccess } from '@odh-dashboard/internal/api/useAccessReview';
+import { ProjectKind } from '@odh-dashboard/k8s-core';
+import useAccessibleNamespaces from '../useAccessibleNamespaces';
+
+jest.mock('@odh-dashboard/internal/api/useAccessReview', () => ({
+  checkAccess: jest.fn(),
+}));
+
+jest.mock('@odh-dashboard/internal/utilities/useFetch', () => {
+  const actual = jest.requireActual('@odh-dashboard/internal/utilities/useFetch');
+  return actual;
+});
+
+const checkAccessMock = jest.mocked(checkAccess);
+
+const makeProject = (name: string, displayName?: string): ProjectKind =>
+  ({
+    metadata: {
+      name,
+      annotations: displayName ? { 'openshift.io/display-name': displayName } : {},
+    },
+  } as unknown as ProjectKind);
+
+const createWrapper = (projects: ProjectKind[], loaded = true) => {
+  const contextValue = {
+    projects,
+    loaded,
+    refresh: jest.fn(),
+  } as unknown as React.ContextType<typeof ProjectsContext>;
+
+  const Wrapper = ({ children }: { children: React.ReactNode }) => (
+    <ProjectsContext.Provider value={contextValue}>{children}</ProjectsContext.Provider>
+  );
+  Wrapper.displayName = 'ProjectsContextWrapper';
+  return Wrapper;
+};
+
+describe('useAccessibleNamespaces', () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
+
+  it('should return namespaces where user has create access', async () => {
+    checkAccessMock.mockImplementation(async ({ namespace }) => namespace === 'allowed-ns');
+
+    const projects = [makeProject('allowed-ns', 'Allowed'), makeProject('denied-ns', 'Denied')];
+    const { result } = renderHook(() => useAccessibleNamespaces(), {
+      wrapper: createWrapper(projects),
+    });
+
+    await waitFor(() => {
+      expect(result.current.loaded).toBe(true);
+    });
+
+    expect(result.current.namespaces).toEqual([{ name: 'allowed-ns', displayName: 'Allowed' }]);
+    expect(result.current.error).toBeUndefined();
+  });
+
+  it('should use namespace name as displayName when annotation is missing', async () => {
+    checkAccessMock.mockResolvedValue(true);
+
+    const projects = [makeProject('my-ns')];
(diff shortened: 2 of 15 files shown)


claudialphonse78 (commented):

claudialphonse78 (commented):

packages/feature-store/src/hooks/useAccessibleNamespaces.ts
```diff
@@ -0,0 +1,80 @@
+import React from 'react';
```
claudialphonse78: could you add test cases for all the hooks you have created here within the hooks folder

packages/feature-store/src/hooks/useExistingFeatureStores.ts
```diff
@@ -0,0 +1,99 @@
...
+    const emptyList: FeatureStoreKind[] = [];
+    const failures: string[] = [];
+    Promise.all(
+      namespaces.map((ns) =>
+        listFeatureStores(ns).catch(() => {
+          failures.push(ns);
+          return emptyList;
+        }),
+      ),
+    ).then((results) => {
```
claudialphonse78: for the failures could we handle it in a try catch block

packages/feature-store/src/hooks/useExistingFeatureStores.ts
```diff
@@ -0,0 +1,99 @@
...
+  error?: Error;
+  existingProjectNames: string[];
+  existingResourceNames: string[];
+  hasUILabeledStore: boolean;
+  primaryStore: FeatureStoreKind | undefined;
+  refresh: () => void;
+};
+
+const useExistingFeatureStores = (): UseExistingFeatureStoresReturn => {
+  const { projects, loaded: projectsLoaded } = React.useContext(ProjectsContext);
```
claudialphonse78, line 25: Do we need to fetch and run the loop on all the namespaces? is there an identifier or label on the namespaces wherever feast would be present to narrow this down since you are trying to identify already existing 
I think the label being added to namespaces is `opendatahub.io/feast: 'true'` which would help narrow down the search instead of running it on all the namespaces

packages/feature-store/src/hooks/useAccessibleNamespaces.ts
```diff
@@ -0,0 +1,80 @@
...
+  error?: Error;
+};
+
+const useAccessibleNamespaces = (): UseAccessibleNamespacesReturn => {
+  const { projects, loaded: projectsLoaded } = React.useContext(ProjectsContext);
+  const [namespaces, setNamespaces] = React.useState<NamespaceInfo[]>([]);
+  const [loaded, setLoaded] = React.useState(false);
+  const [error, setError] = React.useState<Error | undefined>();
+
+  React.useEffect(() => {
```
claudialphonse78: We can use the UseFetch hook here a lot if this would be handled internally by the hook itself

claudialphonse78 (changes requested):

packages/feature-store/src/api/featureStores.ts
```diff
@@ -0,0 +1,163 @@
...
+    metadata: {
+      name: specFields.feastProject,
+      namespace,
+      ...(labels && Object.keys(labels).length > 0 && { labels }),
+    },
+    spec,
+  };
+};
+
+export const createFeatureStore = (
```
claudialphonse78: this needs to be declared as async

packages/feature-store/src/api/featureStores.ts
```diff
@@ -0,0 +1,163 @@
...
+  feastProjectDir?: FeastProjectDir;
+  services?: FeastServices;
+  authz?: FeastAuthzConfig;
+  cronJob?: FeastCronJob;
+  batchEngine?: FeastBatchEngineConfig;
+  replicas?: number;
+  labels?: Record<string, string>;
+};
+
+export const assembleFeatureStore = (formData: FeatureStoreFormSpec): FeatureStoreKind => {
```
claudialphonse78: this is a utility function can we move it out of the API file and have this tested

packages/feature-store/src/api/featureStores.ts
```diff
@@ -0,0 +1,163 @@
...
+  FeastProjectDir,
+  FeastAuthzConfig,
+  FeastCronJob,
+  FeastBatchEngineConfig,
+} from '../k8sTypes';
+
+const isValidFeatureStore = (fs: FeatureStoreKind): boolean =>
+  Boolean(fs.metadata.name && fs.metadata.namespace && fs.spec.feastProject);
+
+const normalizeFeatureStore = (fs: FeatureStoreKind): FeatureStoreKind => ({
```
claudialphonse78: Similarly for this too keep only API calls in this api file

claudialphonse78 (changes requested):

packages/feature-store/src/hooks/useExistingFeatureStores.ts
ntkathole: good catch 👍

packages/feature-store/src/utils/featureStoreUtils.ts
```diff
@@ -0,0 +1,91 @@
...
+  }
+  if (specFields.cronJob) {
+    spec.cronJob = specFields.cronJob;
+  }
+  if (specFields.batchEngine) {
+    spec.batchEngine = specFields.batchEngine;
+  }
+  if (specFields.replicas != null && specFields.replicas > 1) {
+    spec.replicas = specFields.replicas;
+  }
```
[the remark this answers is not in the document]
ntkathole: Scale-to-zero is not a creation-time use case.

claudialphonse78:

packages/feature-store/src/hooks/useNamespaceSecrets.ts
```diff
@@ -0,0 +1,28 @@
...
+import { SecretModel } from '@odh-dashboard/internal/api/models';
+
+const useNamespaceSecrets = (
+  namespace: string,
+): { secrets: string[]; loaded: boolean; error?: Error } => {
+  const fetchSecrets = React.useCallback<FetchStateCallbackPromise<string[]>>(() => {
+    if (!namespace) {
+      return Promise.reject(new NotReadyError('No namespace selected'));
+    }
+    return k8sListResource<SecretKind>({
```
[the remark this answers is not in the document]
ntkathole: 403 is intentionally surfaced via the error state so callers can differentiate "no secrets" (loaded: true, empty list) from "no access" (error set). Adding an SSAR pre-check would double the API calls for the common case where the user does have access, with no functional benefit. The graceful handling belongs in the consuming component, not in this data-fetching hook.

packages/feature-store/src/hooks/useNamespaceConfigMaps.ts
```diff
@@ -0,0 +1,28 @@
...
+import { ConfigMapModel } from '@odh-dashboard/internal/api/models';
+
+const useNamespaceConfigMaps = (
+  namespace: string,
+): { configMaps: string[]; loaded: boolean; error?: Error } => {
+  const fetchConfigMaps = React.useCallback<FetchStateCallbackPromise<string[]>>(() => {
+    if (!namespace) {
+      return Promise.reject(new NotReadyError('No namespace selected'));
+    }
+    return k8sListResource<ConfigMapKind>({
```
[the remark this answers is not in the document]
ntkathole: same here

packages/feature-store/src/api/featureStores.ts
```diff
@@ -0,0 +1,90 @@
...
+  }
+  const result = await k8sListResource<PodKind>({
+    model: PodModel,
+    queryOptions: {
+      ns: namespace,
+      queryParams: { labelSelector: `feast.dev/name=${featureStoreName}` },
+    },
+  });
+  return result.items;
+};
```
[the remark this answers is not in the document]
ntkathole: valid but many existing API helpers in the project (getStorageClasses, etc.) also don't accept opts. It's not unique to this PR.

packages/feature-store/src/hooks/__tests__/useAccessibleNamespaces.spec.tsx
```diff
@@ -0,0 +1,134 @@
...
+      wrapper: createWrapper(projects),
+    });
+
+    await waitFor(() => {
+      expect(result.current.loaded).toBe(true);
+    });
+
+    expect(result.current.namespaces).toEqual([]);
+    expect(result.current.error).toBeUndefined();
+  });
```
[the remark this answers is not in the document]
ntkathole: This test verifies the hook's own try/catch guard in case the upstream contract changes. The real checkAccess is fail-open (.catch(() => true)) and never rejects in practice.

ntkathole:
@claudialphonse78 can you check again if anything pending here? I see konflux checks failed, which are not related

packages/feature-store/src/api/featureStores.ts
```diff
@@ -0,0 +1,90 @@
...
+export const getFeatureStore = (namespace: string, name: string): Promise<FeatureStoreKind> =>
+  k8sGetResource<FeatureStoreKind>({
+    model: FeatureStoreModel,
+    queryOptions: {
+      name,
+      ns: namespace,
+    },
+  }).then(normalizeFeatureStore);
+
+export const deleteFeatureStore = (namespace: string, name: string): Promise<FeatureStoreKind> =>
```
claudialphonse78: - Kubernetes DELETE endpoints return a `K8sStatus` (a success/failure confirmation), not the deleted resource itself
- So `.then(normalizeFeatureStore)` is normalizing the wrong thing at runtime ,TypeScript doesn't catch this because of the generic, but the actual value is a Status object, not a FeatureStore, `createFeatureStore` and `getFeatureStore` normalize correctly because those endpoints actually return the FeatureStore — DELETE doesn't

- For example the pattern we follow in this codebase for deletes is `Promise<K8sStatus>` with no normalization — see `deleteHardwareProfile` in `hardwareProfiles.ts`

```typescript
export const deleteFeatureStore = (namespace: string, name: string): Promise<K8sStatus> =>
  k8sDeleteResource<FeatureStoreKind, K8sStatus>({
    model: FeatureStoreModel,
    queryOptions: { name, ns: namespace },
  });

**Review by claudialphonse78 — CHANGES_REQUESTED** [review]
@ntkathole Just a minor nit rest all looks good

File: packages/feature-store/src/api/featureStores.ts (hunk 13 above)
Comment (ntkathole) [reply packages/feature-store/src/api/featureStores.ts]: updated

File: packages/feature-store/src/api/featureStores.ts [hunk 14]
```diff
@@ -0,0 +1,91 @@
...
+  return normalizeFeatureStore(created);
+};
+
+export const listFeatureStores = (namespace: string): Promise<FeatureStoreKind[]> =>
+  k8sListResource<FeatureStoreKind>({
+    model: FeatureStoreModel,
+    queryOptions: {
+      ns: namespace,
+    },
+  }).then((listResource) => normalizeFeatureStoreList(listResource.items));
```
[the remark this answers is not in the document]
ntkathole: Namespace in these API helpers flows from ProjectsContext (K8s API-sourced), not from user input. The trust boundary is at assembleFeatureStore (where we now validate both namespace and feastProject from form data). Adding validation here would be inconsistent with the rest of the codebase's API helpers, which don't validate their namespace parameter either. The featureStoreName guard in getPodsForFeatureStore was added specifically because it's interpolated into a label selector string (injection risk), which doesn't apply to namespace as a URL path segment.

packages/feature-store/src/hooks/useAccessibleNamespaces.ts
```diff
@@ -0,0 +1,60 @@
...
+            group: FeatureStoreModel.apiGroup ?? '',
+            resource: FeatureStoreModel.plural,
+            verb: 'create',
+            name: '',
+            namespace: ns,
+            subresource: '',
+          });
+        } catch {
+          allowed = false;
+        }
```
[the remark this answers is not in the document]
ntkathole: The hook's local catch is already fail-closed (allowed = false). The fail-open behavior originates in the shared checkAccess utility (frontend/src/api/useAccessReview.ts line 41: .catch(() => true)), which is an intentional project-wide design decision — not introduced by this PR. Changing it to fail-closed would affect all SSAR consumers across the dashboard and should be a separate discussion/PR if desired.

claudialphonse78 (approved):
Changes look good!

PR #8358, opened 2026-07-01
feat(operator): add observability deployment, distribution status, and condition severity

lucferbux:
https://issues.redhat.com/browse/RHOAIENG-62523
https://issues.redhat.com/browse/RHOAIENG-72829
https://issues.redhat.com/browse/RHOAIENG-72830

## Description

Wraps up the operator onboarding epic ([RHOAIENG-60566](https://redhat.atlassian.net/browse/RHOAIENG-60566)) with three remaining controller features:

**[RHOAIENG-62523](https://redhat.atlassian.net/browse/RHOAIENG-62523) — Observability manifest deployment:**
- Added `deployObservabilityManifests()` that renders and deploys PersesDashboard CRs and NetworkPolicies from `manifests/observability/{odh,rhoai}/` to the monitoring namespace
- Checks for `persesdashboards.perses.dev` CRD existence before deploying — returns sentinel error if missing
- Extended `cleanupCrossNamespaceResources()` with NetworkPolicy and PersesDashboard cleanup (unstructured, graceful CRD-not-found via `meta.IsNoMatchError()`)
- Added `perses.dev` RBAC rules to both `config/rbac/role.yaml` and the Helm chart
- Anaconda CE: confirmed entirely a backend concern (`backend/src/routes/api/validate-isv/`), no controller work needed

**[RHOAIENG-72829](https://redhat.atlassian.net/browse/RHOAIENG-72829) — `status.distribution`:**
- Added `Distribution` struct (Name + Version) to Dashboard CRD status
- `readDistributionConfig()` reads from `opendatahub-dashboard-config` ConfigMap (chart-deployed)
- Populated during reconcile, cleared on Removed state

**[RHOAIENG-72830](https://redhat.atlassian.net/browse/RHOAIENG-72830) — Condition severity + ObservabilityAvailable:**
- `ObservabilityAvailable` registered as 4th tracked condition in the conditions Manager
- Uses `ConditionSeverityInfo` when False due to optional dependency (CRD not found, disabled) — `IsHappy()` ignores these so missing Perses won't block Ready
- Uses default error severity on actual deployment failures
- Audited existing severity assignments — all correct, no changes needed

## How Has This Been Tested?

- `go build ./...` — clean build
- `go test -race ./api/... ./internal/... ./charts/...` — all pass
- `make lint` — 0 issues
- Chart CRD sync test validates RBAC and CRD drift between generated sources and chart copies

## Test Impact

Existing unit tests cover the reconciler, config reading, and chart sync. The observability deployment path is exercised through the reconciler's condition-setting logic. Integration testing with actual Perses CRDs will be covered by the operator e2e test ticket (separate from this PR).

## Request review criteria:

## Summary by CodeRabbit

* **New Features**
  * Dashboard status now includes optional `distribution` details (name and version).
  * When observability is enabled and supported, related observability resources are deployed with the dashboard.
* **Bug Fixes**
  * Improved “removed” handling by deleting additional related observability network policies and dashboard custom resources.
  * Enhanced readiness/status reporting for cases like missing configuration, unsupported observability, or discovery failures.
* **Chores**
  * Updated RBAC permissions to support Perses dashboard operations and removed unnecessary CustomResourceDefinition permissions.


lucferbux:
@liday-rh can you review this please?

dashboard-operator/internal/controller/actions.go
```diff
@@ -93,3 +102,59 @@ func extractDashboardURL(ctx context.Context, cli client.Client, dashboard *v1al
...
+
+	if dashboard.Spec.Observability.PersesService == nil {
+		return nil
+	}
+
+	crd := &apiextensionsv1.CustomResourceDefinition{}
+	if err := cli.Get(ctx, types.NamespacedName{Name: "persesdashboards.perses.dev"}, crd); err != nil {
+		if k8serrors.IsNotFound(err) {
+			return ErrPersesCRDNotFound
+		}
```
liday-rh, line 130: This `cli.Get` on an `apiextensionsv1.CustomResourceDefinition` requires `get` permission on `customresourcedefinitions` in the `apiextensions.k8s.io` API group — but the RBAC additions in this PR only cover `perses.dev/persesdashboards`.

If the operator SA doesn't already have CRD-level read access from elsewhere, this returns **403 Forbidden** (not 404), so `k8serrors.IsNotFound()` won't match. The function then returns a generic wrapped error, and the condition gets marked `DeployFailed` instead of the intended `PersesCRDNotFound` with `SeverityInfo`.

Worth verifying the SA already has this permission. Alternatively, a discovery-based check (e.g. attempting a `List` and checking `meta.IsNoMatchError`, like you already do in cleanup) avoids the CRD-level RBAC requirement entirely.

dashboard-operator/internal/controller/actions.go
```diff
@@ -93,3 +102,59 @@ func extractDashboardURL(ctx context.Context, cli client.Client, dashboard *v1al
...
+
+		return fmt.Errorf("failed to check PersesDashboard CRD: %w", err)
+	}
+
+	obsNamespace := dashboard.Spec.Observability.PersesService.Namespace
+
+	m := observabilityManifestInfo(basePath, platform)
+	engine := kustomize.NewEngine()
+
+	rendered, err := engine.Render(m.String(), kustomize.WithNamespace(obsNamespace))
```
liday-rh, line 143: nit: `Namespace` isn't validated after the nil checks. If `PersesService.Namespace` is an empty string, `kustomize.WithNamespace("")` would render manifests without a proper namespace override, and the deployer would apply them to the wrong namespace silently.

The same applies to `cleanupCrossNamespaceResources` where `obsNS` is read from the same path.

Suggestion:
```go
if obsNamespace == "" {
    return fmt.Errorf("observability PersesService namespace must not be empty")
}
```

dashboard-operator/internal/controller/dashboard_reconciler.go
```diff
@@ -223,6 +241,31 @@ func (r *DashboardReconciler) reconcile(
...
+		if dashboard.Spec.Observability != nil && dashboard.Spec.Observability.Enabled {
+			cm.MarkTrue(conditionObservabilityAvailable,
+				conditions.WithReason("Deployed"),
+				conditions.WithMessage("Observability manifests applied successfully"))
+		} else {
+			cm.MarkFalse(conditionObservabilityAvailable,
+				conditions.WithReason("Disabled"),
+				conditions.WithMessage("Observability is not enabled"),
+				conditions.WithSeverity(common.ConditionSeverityInfo))
+		}
```
liday-rh: nit: `deployObservabilityManifests` returns `nil` for two distinct reasons — (1) successfully deployed, or (2) observability is disabled/unconfigured (early return). The caller then re-checks `Spec.Observability` to disambiguate.

This works today but is fragile: if a third early-return path is added to `deployObservabilityManifests`, this disambiguation could silently break. Consider returning a typed result or a second sentinel error (e.g. `ErrObservabilityDisabled`) so the caller doesn't need to re-derive why it got `nil`.

dashboard-operator/internal/controller/config.go
```diff
@@ -55,3 +58,37 @@ func readOperatorConfig(ctx context.Context, cli client.Client, namespace string
...
+	}
+
+	name := cm.Data["distribution.name"]
+	version := cm.Data["distribution.version"]
+
+	if name == "" && version == "" {
+		return nil
+	}
+
+	return &v1alpha1.Distribution{
```
liday-rh, line 104: Optional/low-risk: `distribution.name` and `distribution.version` are read from the ConfigMap and written directly into CRD status without any length validation. Since this ConfigMap is chart-deployed the risk is low, but a misconfigured ConfigMap could embed arbitrarily large strings into the status subresource, bloating etcd and API responses. A simple length cap (e.g. 256 chars) would prevent accidental abuse.

liday-rh (commented):
Overall this is a well-structured PR — clean separation of concerns, proper error handling, and good use of condition severity to prevent an optional feature from blocking Ready.

**Key findings (4 inline comments):**

1. **RBAC gap for CRD existence check** — `deployObservabilityManifests` does a direct `Get` on the CRD resource, which requires `apiextensions.k8s.io` RBAC that isn't added in this PR. If the SA doesn't already have it, a 403 will be misclassified as `DeployFailed` instead of `PersesCRDNotFound`.

2. **Empty `obsNamespace` not validated** — if `PersesService.Namespace` is an empty string, manifests deploy to the wrong namespace silently. Same risk in cleanup.

3. **Fragile nil-return disambiguation** — the caller re-checks `Spec.Observability` to tell "deployed OK" from "disabled" since both return `nil`. A sentinel error would be more robust.

4. **Unbounded ConfigMap values in status** (low risk) — distribution name/version go straight to CRD status without length checks.

**Things that look correct:**
- `meta.IsNoMatchError` in cleanup — defensive and correct
- `PersesDashboardList` kind for `UnstructuredList` — correct
- Full CRUD RBAC on `persesdashboards` — appropriate for deploy+cleanup
- `SeverityInfo` for optional/missing CRD — right design choice
- `client.IgnoreNotFound` on cleanup deletes — handles races gracefully

dashboard-operator/internal/controller/actions.go
lucferbux: Good catch — switched to discovery-based `List` with `meta.IsNoMatchError`, matching the pattern already used in cleanup. Also removed the `apiextensions.k8s.io` RBAC rules from both `config/rbac/role.yaml` and the chart `rbac.yaml` since they're no longer needed.

dashboard-operator/internal/controller/actions.go
lucferbux: Done — added the namespace validation exactly as suggested. Returns an error before rendering if `PersesService.Namespace` is empty.

dashboard-operator/internal/controller/dashboard_reconciler.go
lucferbux: Good call — added an `ErrObservabilityDisabled` sentinel error. The early-return paths in `deployObservabilityManifests` now return this sentinel instead of `nil`, and the caller uses `errors.Is` to disambiguate without re-checking the spec. No more fragile nil re-derivation.

dashboard-operator/internal/controller/config.go
lucferbux: Added a 256-char cap on both fields with a log line when truncation occurs. Low risk as you noted, but good defense-in-depth.

dashboard-operator/api/v1alpha1/dashboard_types.go
```diff
@@ -63,6 +63,15 @@ type ModuleOverride struct {
...
 
+// Distribution reports the product distribution identity (e.g. "RHOAI" / "2.20").
+type Distribution struct {
+	// +optional
+	Name string `json:"name,omitempty"`
+
+	// +optional
+	Version string `json:"version,omitempty"`
+}
+
```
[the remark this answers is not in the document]
lucferbux: Added `+kubebuilder:validation:MaxLength=256` markers on both `Name` and `Version` fields and regenerated CRD manifests. The API-level constraint now matches the controller-side truncation.

dashboard-operator/internal/controller/actions.go
```diff
@@ -93,3 +102,64 @@ func extractDashboardURL(ctx context.Context, cli client.Client, dashboard *v1al
...
+) error {
+	logger := log.FromContext(ctx)
+
+	if dashboard.Spec.Observability == nil || !dashboard.Spec.Observability.Enabled {
+		return ErrObservabilityDisabled
+	}
+
+	if dashboard.Spec.Observability.PersesService == nil {
+		return ErrObservabilityDisabled
+	}
```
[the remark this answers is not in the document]
lucferbux: Good catch — added a separate `ErrPersesServiceRequired` sentinel for when `Enabled=true` but `PersesService` is nil. The reconciler now handles this with error severity (`InvalidConfig` reason) instead of the info-level `Disabled` path.

dashboard-operator/internal/controller/actions.go
```diff
@@ -93,3 +102,64 @@ func extractDashboardURL(ctx context.Context, cli client.Client, dashboard *v1al
...
+
+	obsNamespace := dashboard.Spec.Observability.PersesService.Namespace
+	if obsNamespace == "" {
+		return fmt.Errorf("observability PersesService namespace must not be empty")
+	}
+
+	m := observabilityManifestInfo(basePath, platform)
+	engine := kustomize.NewEngine()
+
+	rendered, err := engine.Render(m.String(), kustomize.WithNamespace(obsNamespace))
```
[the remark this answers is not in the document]
lucferbux: Skipping this one — the Dashboard CR is cluster-scoped and only writable by cluster admins (or the platform operator). The confused deputy scenario requires the attacker to already have `update` on the Dashboard CR, which is a cluster-admin level privilege. Namespace restriction at the controller level would require an architectural decision about where the "allowed namespace" comes from (hardcoded? ConfigMap? Another CR field?), which is better addressed as a follow-up if we see a real need. The current CEL validation and RBAC on the CR itself are the primary guardrails.

dashboard-operator/internal/controller/config.go
```diff
@@ -55,3 +59,47 @@ func readOperatorConfig(ctx context.Context, cli client.Client, namespace string
...
+	cm := &corev1.ConfigMap{}
+	key := types.NamespacedName{Name: distributionConfigMapName, Namespace: namespace}
+
+	if err := cli.Get(ctx, key, cm); err != nil {
+		if !k8serrors.IsNotFound(err) {
+			logger.Error(err, "Failed to read distribution config ConfigMap")
+		}
+
+		return nil
+	}
```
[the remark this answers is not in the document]
lucferbux: Fixed — `readDistributionConfig` now returns `(*Distribution, error)`. On transient API errors the reconciler logs the error and preserves the existing `status.distribution` value instead of clearing it.

dashboard-operator/internal/controller/config.go
```diff
@@ -55,3 +59,47 @@ func readOperatorConfig(ctx context.Context, cli client.Client, namespace string
...
+
+	if len(name) > maxDistributionFieldLen {
+		logger.Info("Truncating distribution.name", "original_length", len(name))
+		name = name[:maxDistributionFieldLen]
+	}
+
+	if len(version) > maxDistributionFieldLen {
+		logger.Info("Truncating distribution.version", "original_length", len(version))
+		version = version[:maxDistributionFieldLen]
+	}
```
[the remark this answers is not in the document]
lucferbux: Switched to rune-based truncation using `utf8.RuneCountInString` and `[]rune` conversion, so multi-byte characters won't be split.

PR #8364, opened 2026-07-01
RHOAIENG-38993: Remove static SA token management, use dynamic OCP JWT only

gmenher:
https://issues.redhat.com/browse/RHOAIENG-38993

## Description

Removes the legacy static service account token wiring from the OGX install flow and the dead `SAToken` schema from the API response.

The OGX install path previously discovered InferenceService SA token secrets and injected them as `VLLM_API_TOKEN_N` env vars in the OGXServer pod spec. This is redundant — runtime inference already uses dynamic per-request credentials via `x-ogx-provider-data` (user JWT for namespace models, ephemeral MaaS keys, or K8s Secrets for custom endpoints). The static tokens were never used at runtime.

With this change, all integrations now consistently use dynamic credentials:

| Component | Auth mechanism | Status |
|-----------|---------------|--------|
| K8s API | User JWT (`rest.Config.BearerToken`) | Already dynamic |
| OGX/LlamaStack | User JWT (`WithAPIKey(identity.Token)`) | Already dynamic |
| vLLM/LLM-D | User JWT via `x-ogx-provider-data` | Already dynamic |
| MaaS | Ephemeral key via MaaS BFF (cached 30m) | Already dynamic |
| Custom endpoints | K8s Secret fetched at runtime | Already dynamic |
| MLflow | User JWT (`Authorization: Bearer`) | Already dynamic |
| NeMo Guardrails | User JWT (BFF→NeMo) | Already dynamic |
| MCP Servers | User JWT + `X-MCP-Bearer` | Already dynamic |
| **OGX install (ConfigMap)** | **Static SA token in pod env** | **Removed in this PR** |

This PR improves security (short-lived user JWTs replace long-lived SA tokens), enables per-user audit trails, and eliminates the SA secret discovery RBAC surface.

## OGX Compatibility Verification

Verified directly in the [OGX source](https://github.com/ogx-ai/ogx) (against `main`, Jul 1 2026) that removing `api_token` from the ConfigMap does not break OGX:

1. **OGX starts without `api_token`** — the field is `Optional` with `default=None` in `VLLMInferenceAdapterConfig`. When absent, it falls back to `"NO KEY REQUIRED"` and the server starts normally.
2. **`x-ogx-provider-data` always takes precedence** — the `_get_api_key_from_config_or_provider_data()` method in `OpenAIMixin` first reads the config value, then overwrites it with whatever comes in the per-request `provider_data` header. Since the BFF always sends `vllm_api_token` in that header, the config value is irrelevant at runtime.
3. **No standalone OGX calls in production** — all user traffic goes through the BFF which injects `x-ogx-provider-data` on every request. The config-level token only exists as a fallback for CLI/dev scenarios.

Current clusters already operate without `api_token` in the ConfigMap, confirmed via `oc get cm llama-stack-config -o yaml | grep api_token` returning empty.

## How Has This Been Tested?

- `go build ./...` — passes
- `go test ./internal/integrations/kubernetes/...` — passes
- Frontend jest: 1668 tests, 111 suites — all pass
- Lint (eslint + pre-commit hooks) — passes
- E2E with modified BFF against live cluster (port-forward to OGX):
  - Custom endpoint (`endpoint-3/gpt-4o-mini`): auth via K8s Secret — works
  - MaaS model (`maas-1/gemini-2.5-flash`): auth via MaaS BFF / JWT fallback — works
- `GET /gen-ai/api/v1/models` response no longer contains `sa_token` field
- Cluster inspection: existing OGXServer and ConfigMap already had no `VLLM_API_TOKEN`

## Test Impact

No new tests needed. Existing tests updated to remove `VLLM_API_TOKEN` and `sa_token` expectations from fixtures/mocks. The runtime inference path (`getUserJWTProviderData`, `getMaaSProviderData`) has zero code changes.

## Request review criteria:

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed an unnecessary token-related field from model and endpoint data shown in the app and returned by model creation flows.
  * Improved Kubernetes install behavior to validate required credentials earlier and avoid injecting per-model token environment values.
  * Updated model handling so existing configuration and capabilities continue to work without the removed field.

* **Documentation**
  * Refreshed API examples and capability descriptions to match the updated model response shape.


packages/cypress/cypress/utils/oc_commands/autoragInfra.ts
```diff
@@ -286,7 +286,6 @@ const buildOgxConfig = (namespace: string): string => {
           config: {
             base_url: llmUrl,
             max_tokens: '${env.VLLM_MAX_TOKENS:=4096}',
-            api_token: '${env.VLLM_API_TOKEN_1:=fake}',
```
NickGagan, line 289: Just going to remove this change, it's outside of our scope anyways

NickGagan (approved):

PR #8383, opened 2026-07-02
ci perf: Multiple CI Enhancements to increase Performance

antowaddle:
https://redhat.atlassian.net/browse/RHOAIENG-18412

## Description
     
**Optimizes CI test workflow** by parallelizing builds and improving cache strategies, reducing test workflow critical path from **~20 minutes to ~5 minutes** — a **75% improvement**.

### Problem
The original workflow had multiple bottlenecks causing long wait times:
1. **Sequential Cypress builds**: `Cypress-Setup` job built all 9 Cypress-enabled packages sequentially in a single job, taking ~15 minutes before any tests could run
2. **Unbalanced test groups**: Some Cypress test groups bundled multiple 17-19KB files together, causing individual test jobs to run for 12+ minutes
3. **Queue congestion**: With 30+ engineers, long-running jobs blocked CI capacity

This created a situation where:
- Every PR waited 15+ minutes for builds before tests started
- Cache misses triggered expensive rebuilds on shared runners
- Some test groups took 12-20 minutes to complete
- Build/test failures required re-running entire long sequences

### Solution

This PR implements three complementary optimizations aligned with [GitHub Actions best practices](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs):

#### 1. Parallel Cypress Build Matrix

Split `Cypress-Setup` into a **parallel matrix strategy**, one job per package:

- **Auto-discovery**: `Get-Cypress-Packages` job queries workspace for packages with `cypress:server:build:coverage` script
- **Parallel builds**: Each package builds independently with dedicated runner resources (9 jobs in parallel)
- **Artifact sharing**: Builds uploaded as artifacts, restored in `Cypress-Mock-Tests` jobs
- **Smart verification**: Verify step checks all package builds with fallback rebuild only if genuinely missing

**Note**: Per-package build caching was removed to ensure tests always run against fresh code. Future optimization will use Turbo's dependency-aware caching (see [#8432](https://github.com/opendatahub-io/odh-dashboard/pull/8432)).

#### 2. Optimized Test Grouping

Lowered Cypress test file split threshold from 20KB to 15KB:
- Splits more files into individual test jobs
- Prevents slow grouped tests (e.g., `featureStore/other` previously took 12+ minutes)
- Automatically applies to new test files

### Performance Impact & GitHub Actions Best Practices

**GitHub's documentation emphasizes:**
> "Matrix strategies reduce wall-clock time by running jobs in parallel. Billing is per runner-minute (total compute), not wall-clock time. Using more runners for a shorter duration is the same cost as fewer runners for longer, but developers wait less."

**Key principle from [GitHub Docs - Using a matrix strategy](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs):**
- Billing is per runner-minute (total compute), not wall-clock
- Parallelism doesn't increase cost but dramatically reduces developer wait time
- Free tier: 2000 minutes/month with 20 concurrent jobs → optimized for parallel execution

**Our implementation:**
- **Old**: 1 runner × 904 seconds (Cypress builds) + multiple 12-20 min test jobs = ~20 min critical path | Developer waits: **~20 minutes**
- **New**: 9 runners × ~30 seconds each (parallel builds) + balanced test jobs = ~5 min critical path | Developer waits: **~5 minutes**

**Trade-off analysis:**
- We use ~2× more total runner-minutes for Cypress builds
- BUT save **75% wall-clock time** (20 min → 5 min)
- For 30 engineers × average 3 PRs/day = 90 PR runs/day
- Time saved: **90 runs × 15 min saved = 22.5 hours/day of developer productivity**

This is exactly why GitHub designed matrix strategies — **optimize for developer productivity, not runner utilization**. The total compute cost remains similar, but the human cost (waiting) drops dramatically.

## How Has This Been Tested?
- CI runs on this PR branch
- Verified parallel builds complete successfully
- Confirmed cache reuse for lint/type-check/unit-test jobs
- Validated test matrix splits files correctly

## Test Impact
- No test logic changes
- Test execution is parallelized but tests remain identical
- Builds now always use fresh code (no stale build caching)

## Request review criteria:

If you have UI changes: 
- N/A (workflow changes only)

## Summary by CodeRabbit

* **New Features**
  * Added CI discovery to run Cypress setup/coverage per workspace using build artifacts.
* **Bug Fixes**
  * Ensured Cypress mock tests always have the required coverage output by downloading artifacts for each workspace and rebuilding when any are missing.
* **Chores**
  * Reworked the Cypress CI pipeline from cache-based setup to a workspace matrix artifact flow for more reliable, scalable runs.
* **Tests**
  * Updated Cypress spec splitting: `.cy.ts` files now use a 15KB threshold (down from 20KB) for “large” grouping.


christianvogt:
Ok with changing cypress builds to individual jobs but be aware that if we improve our dep graph and can start to use turbo cache for build artifact caching that we will need to go back to a single job.

The new approach of individual build caches will affect the work in https://github.com/opendatahub-io/odh-dashboard/pull/8432. This PR uses the current build cache to improve it's performance and will need to align.

The current cypress builds will rebuild whenever there is a code change. It seems to me that the new approach will not do so and as such new commits to this PR will pull the build artifacts from cache instead of rebuilding. Only if the package.json or other key artifacts change will it get rebuilt. It would be safer to keep the current approach of rebuilding on any change simply because calculating when it build is difficult, especially without a clean dependency graph.

antowaddle:
> Ok with changing cypress builds to individual jobs but be aware that if we improve our dep graph and can start to use turbo cache for build artifact caching that we will need to go back to a single job.
> 
> The new approach of individual build caches will affect the work in #8432. This PR uses the current build cache to improve it's performance and will need to align.
> 
> The current cypress builds will rebuild whenever there is a code change. It seems to me that the new approach will not do so and as such new commits to this PR will pull the build artifacts from cache instead of rebuilding. Only if the package.json or other key artifacts change will it get rebuilt. It would be safer to keep the current approach of rebuilding on any change simply because calculating when it build is difficult, especially without a clean dependency graph.

Thanks @christianvogt for reviewing here. As discussed, I've removed the aggressive caching which had the potential to not invalidate. 

This is now ready for review and will decrease our CI time to ~20-25 minutes (on average).

christianvogt:
@odh-dashboard-agent preflight

.github/workflows/test.yml
```diff
@@ -53,7 +53,7 @@ jobs:
         with:
           path: |
             **/.turbo
-          key: ${{ runner.os }}-${{ env.NODE_VERSION }}-turbo-type-check-${{ github.sha }}
+          key: ${{ runner.os }}-${{ env.NODE_VERSION }}-turbo-type-check-${{ hashFiles('package-lock.json') }}
```
christianvogt: I don't think this is the place to change the turbo cache keys.

There's two main thoughts for turbo:
1. cache aggressively to speed up per PR subsequent runs at the cost of storage
2. save on github action cache storage at the cost of PR subsequent runs

Currently we are doing 1. With our high pressure on github action cache I'm ok to move to towards 2. However this cache key falls in between -- call it 1.5. IMO we are better off just caching once on main and all PRs use the singular cache as the warm start. In fact we could amalgamate all turbo caches into one doing this.

Now if we were to use turbo for caching build artifacts, then per pr will be beneficial although it likely overwhelms our available cache space. So a warm start from main again may likely prove to be the best option here too.

All this to say, I don't think it's worth making the suggested turbo cache key changes in this PR and instead going towards option 2 with caching done at main and then shared across all PRs.

antowaddle: I've reverted turbo cache keys back to github.sha to match what we do today in main. https://github.com/opendatahub-io/odh-dashboard/pull/8383/changes/43cf6e07bb33912850ca93855fa71aca251cab05

.github/workflows/test.yml
```diff
@@ -247,17 +284,65 @@ jobs:
...
+          path: cypress-artifacts
+      - name: Restore artifact structure
+        env:
+          PACKAGES_JSON: ${{ needs.Get-Cypress-Packages.outputs.packages }}
+        run: |
+          # Move artifacts from cypress-artifacts/cypress-build-{name}/**/public-cypress
+          # to the expected {workspace}/**/public-cypress location
+          PACKAGES="$PACKAGES_JSON"
+
+          for pkg_json in $(echo "$PACKAGES" | jq -c '.[]'); do
```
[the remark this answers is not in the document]
antowaddle: Resolved in https://github.com/opendatahub-io/odh-dashboard/pull/8383/commits/43cf6e07bb33912850ca93855fa71aca251cab05

PR #8400, opened 2026-07-03
feat(llmd-serving): add admin topology configurations settings tab

YuliaKrimerman:
https://issues.redhat.com/browse/RHOAIENG-70584

## Description

Adds an admin settings tab for managing llm-d topology configurations under **Settings > Model resources and operations > llm-d topology configurations**.

### What's included:

**Extension registration** (`extensions.ts`)

**List page** (`TopologyConfigurationsView` + `TopologyConfigurationsTable`)

**Create/Edit/Duplicate pages** (`TopologyConfigurationCreateEdit`)

**Cypress mock tests** (7 test cases)

## How Has This Been Tested?

https://github.com/user-attachments/assets/c1f8183b-8ee2-4110-bd4f-9705126d4b0f

- Manual testing on a live cluster with `llmdTopologyConfigs` feature flag enabled
- Verified list page renders topology configs with descriptions and pre-installed badges
- Tested enable/disable toggle patches the correct annotation
- Tested create flow with all 3 configuration sources (sample, upload, blank)
- Tested duplicate flow pre-fills from source config
- Tested kebab actions: Duplicate-only for pre-installed, full menu for user-created
- Cypress mock tests written and passing locally

## Test Impact

**New test files:**
- `packages/cypress/cypress/tests/mocked/modelServing/modelServingLlmdTopologyAdmin.cy.ts` — 7 test cases covering tab visibility, table rendering, toggle, and kebab actions
- `packages/cypress/cypress/pages/llmdTopologySettings.ts` — page object with `TopologyConfigRow` helpers

## Request review criteria:

If you have UI changes:

## Summary by CodeRabbit

* **New Features**
  * Added an admin settings area for LLMD topology configurations, including navigation, routes, and a new configurations table with empty state.
  * Added per-configuration row controls (enabled switch, optional pre-installed label, and contextual actions like duplicate/edit/delete).
  * Implemented create/edit/duplicate flow with a YAML editor and template-based YAML loading.
* **Bug Fixes**
  * Improved save/delete validation and clearer error notifications.
* **Tests**
  * Added Cypress coverage for feature-flag behavior and enabled/delete logic.


packages/llmd-serving/src/settings/TopologyConfigurationsTable.tsx
```diff
@@ -0,0 +1,237 @@
...
+            const displayName = getDisplayNameFromK8sResource(config);
+            const description = getDescriptionFromK8sResource(config);
+            const preInstalled = isConfigPreInstalled(config);
+            const enabled = isConfigEnabled(config);
+            const topologyType = getConfigTopologyType(config);
+            const isDashboardCreated =
+              config.metadata.labels?.[DASHBOARD_RESOURCE_LABEL] === 'true' && !preInstalled;
+
+            return (
+              <Tr key={configName} data-testid={`topology-config-row-${configName}`}>
```
andrewballantyne: Make the row its own component -- it's easier to manage things like this -- you're either in the table context or the row context... I see little value in having a table&row context. You are in the page, or you're in the row... they have different mental models and you should not allow the burry nature of those two.

If this was Claude's suggestion, let me know... we should definitely help Claude understand good practices.

packages/llmd-serving/src/settings/TopologyConfigurationCreateEdit.tsx
```diff
@@ -0,0 +1,301 @@
...
+          <BreadcrumbItem isActive>{title}</BreadcrumbItem>
+        </Breadcrumb>
+      }
+      loaded
+      empty={false}
+      provideChildrenPadding
+      data-testid="topology-config-create-edit-page"
+    >
+      <Form>
+        <Stack hasGutter>
```
andrewballantyne: Do you need a stack in a form?

packages/llmd-serving/src/settings/TopologyConfigurationCreateEdit.tsx
```diff
@@ -0,0 +1,301 @@
...
+          </StackItem>
+          <StackItem>
+            <FormGroup label="Description" fieldId="config-description">
+              <TextArea
+                id="config-description"
+                data-testid="config-description-input"
+                value={description}
+                onChange={(_e, value) => setDescription(value)}
+              />
+            </FormGroup>
```
andrewballantyne: We should use our K8sNameDescriptionField component...

packages/llmd-serving/src/settings/TopologyConfigurationCreateEdit.tsx
```diff
@@ -0,0 +1,301 @@
...
+    try {
+      const parsed: unknown = yamlValue ? YAML.parse(yamlValue) : undefined;
+      spec = parsed as LLMInferenceServiceConfigKind['spec']; // eslint-disable-line @typescript-eslint/consistent-type-assertions
+    } catch (e) {
+      setError(e instanceof Error ? e : new Error('Invalid YAML'));
+      setLoading(false);
+      return;
+    }
+
+    const resourceName =
```
andrewballantyne: You shouldn't need to calculate this -- use our K8sNameDescriptionField hook

packages/llmd-serving/src/settings/TopologyConfigurationCreateEdit.tsx
```diff
@@ -0,0 +1,301 @@
...
+  const { topologyType, configName } = useParams<{
+    topologyType?: string;
+    configName?: string;
+  }>();
+  const navigate = useNavigate();
+  const location = useLocation();
+  const { dashboardNamespace } = useDashboardNamespace();
+  const notification = useNotification();
+
+  const isDuplicateMode = location.pathname.includes('/duplicate/');
```
andrewballantyne: Not sure we need a route for duplicate -- is this how we do it in other spots 🤔 I'd have expected this to be basically a create with data...

packages/llmd-serving/src/settings/TopologyConfigurationCreateEdit.tsx
```diff
@@ -0,0 +1,301 @@
...
+    }
+  }, [sourceConfig]);
+
+  const topologyTypeLabel = resolvedTopologyType
+    ? TopologyTypeLabels[resolvedTopologyType]
+    : 'Unknown';
+
+  const sourceName = sourceConfig ? getDisplayNameFromK8sResource(sourceConfig) : configName ?? '';
+
+  const title = isDuplicateMode
```
andrewballantyne: Nit, but it helps with readability -- given you did it with description...
```suggestion
  const pageTitle = isDuplicateMode
```

packages/llmd-serving/src/settings/TopologyConfigurationsTable.tsx
YuliaKrimerman: Agree - it(and myself) definitely missed the "follow codebase standarts"

packages/llmd-serving/src/settings/TopologyConfigurationCreateEdit.tsx
```diff
@@ -0,0 +1,301 @@
...
+                id="config-description"
+                data-testid="config-description-input"
+                value={description}
+                onChange={(_e, value) => setDescription(value)}
+              />
+            </FormGroup>
+          </StackItem>
+          {!isEditMode && !isDuplicateMode && (
+            <StackItem>
+              <ConfigSourceSelect
```
andrewballantyne: I don't understand this dropdown... @vconzola why not follow what we have on ServingRuntimes for YAML usage?

Image

We should have a custom control for samples or something like that?
https://www.patternfly.org/components/code-editor#codeeditorcontrol we could do a modal like this
Or maybe we can mount a menu action?

(from your prototype)
Image
This doesn't make any sense 🤔 How do I select a config? 

@YuliaKrimerman can you remove samples for now -- we will need to come back to it later.

packages/llmd-serving/src/settings/TopologyConfigurationCreateEdit.tsx
```diff
@@ -0,0 +1,301 @@
...
+    'apiVersion: serving.kserve.io/v1alpha1',
+    'kind: LLMInferenceServiceConfig',
+    'metadata:',
+    '  name: my-topology-configuration',
+    'spec:',
+    `  # ${comments[topoType]}`,
+    '  templateRef: kserve-config-llm-template',
+    '',
+  ].join('\n');
+};
```
andrewballantyne: What were you going for here?

We'll need to make a call to a server to get back a full yaml to place in the yaml editor. Not sure this even works as a stub for that 🤔 can you help me understanding the goals here?

packages/llmd-serving/src/settings/TopologyConfigurationCreateEdit.tsx
```diff
@@ -0,0 +1,301 @@
...
+    undefined,
+  );
+  const [loading, setLoading] = React.useState(false);
+  const [error, setError] = React.useState<Error | undefined>();
+
+  React.useEffect(() => {
+    if (sourceConfig) {
+      setDisplayName(getDisplayNameFromK8sResource(sourceConfig));
+      setDescription(getDescriptionFromK8sResource(sourceConfig));
+      setYamlValue(sourceConfig.spec ? YAML.stringify(sourceConfig.spec) : '');
```
andrewballantyne: Image
Why are we only getting spec? We want the whole resource 🤔

packages/llmd-serving/src/settings/TopologyConfigurationCreateEdit.tsx
```diff
@@ -0,0 +1,301 @@
...
+        name: resourceName,
+        namespace: dashboardNamespace,
+        labels: {
+          [CONFIG_TYPE_LABEL]: resolvedTopologyType,
+          [DASHBOARD_RESOURCE_LABEL]: 'true',
+        },
+        annotations: {
+          'openshift.io/display-name': displayName,
+          'openshift.io/description': description,
+        },
```
andrewballantyne: I think we want to let them add their own if they want ... we are too prescriptive for this experience...

packages/llmd-serving/src/settings/TopologyConfigurationsRoutes.tsx
```diff
@@ -0,0 +1,21 @@
...
+  () => import('./TopologyConfigurationCreateEdit'),
+);
+
+const TopologyConfigurationsRoutes: React.FC = () => (
+  <React.Suspense fallback={null}>
+    <Routes>
+      <Route index element={<TopologyConfigurationsView />} />
+      <Route path="add/:topologyType" element={<TopologyConfigurationCreateEdit />} />
+      <Route path="edit/:configName" element={<TopologyConfigurationCreateEdit />} />
+      <Route path="duplicate/:configName" element={<TopologyConfigurationCreateEdit />} />
```
andrewballantyne: I feel this should be "Add" + starting data...

packages/llmd-serving/src/settings/TopologyConfigurationsTable.tsx
```diff
@@ -0,0 +1,237 @@
...
+                  >
+                    Add {TopologyTypeLabels[tt].toLowerCase()} configuration
+                  </DropdownItem>
+                ))}
+              </DropdownList>
+            </Dropdown>
+          </ToolbarItem>
+        </ToolbarContent>
+      </Toolbar>
+      <Table aria-label="Topology configurations table" data-testid="topology-configurations-table">
```
andrewballantyne: Did you consider using the table component from `frontend`?

packages/llmd-serving/src/settings/TopologyConfigurationsTable.tsx
```diff
@@ -0,0 +1,237 @@
...
+                  {topologyType ? TopologyTypeLabels[topologyType] : '-'}
+                </Td>
+                <Td isActionCell>
+                  <ActionsColumn
+                    items={
+                      isDashboardCreated
+                        ? [
+                            {
+                              title: 'Duplicate',
+                              onClick: () => navigate(`duplicate/${configName}`),
```
andrewballantyne: Here is where I would do the route with data

packages/llmd-serving/src/settings/ConfigSourceSelect.tsx
andrewballantyne, line 1: I think this is a gap in Vince's prototype -- this feels too bespoke for our application at this point.

I recommend sticking to the way the code editor works.

packages/llmd-serving/src/settings/ConfigYAMLEditor.tsx
```diff
@@ -0,0 +1,25 @@
...
+const ConfigYAMLEditor: React.FC<ConfigYAMLEditorProps> = ({ value, onChange, isReadOnly }) => (
+  <div data-testid="config-yaml-editor">
+    <CodeEditor
+      code={value}
+      onChange={(val) => onChange(val)}
+      language={Language.yaml}
+      isReadOnly={isReadOnly}
+      isCopyEnabled
+      isLanguageLabelVisible
+      height="400px"
```
andrewballantyne, line 21: Curious -- why 400?

andrewballantyne (changes requested):

packages/llmd-serving/src/settings/ConfigYAMLEditor.tsx
YuliaKrimerman: Looks better than 399

packages/llmd-serving/src/settings/TopologyConfigurationCreateEdit.tsx
YuliaKrimerman: Should the YAML editor show the entire resource (apiVersion, kind, metadata, spec) and we parse the full thing on submit? Or just spec + template?

packages/llmd-serving/src/settings/TopologyConfigurationCreateEdit.tsx
YuliaKrimerman: If the YAML editor shows the full resource, users can add whatever labels/annotations they want in the YAML. Should I also keep the Name/Description form fields above the editor (which map to metadata.name and openshift.io/display-name), or let the YAML handle everything

packages/llmd-serving/src/settings/TopologyConfigurationCreateEdit.tsx
andrewballantyne: Yes everything -- the goal is to let them create these resources in our app so we can have proper context when trying to help them. This is somewhat a glorified yaml editor... but in context.

We should enforce some things ourselves -- like topology type annotation & names/description -- because we have those controls outside of the yaml. Debatable if you render them inside the YAML -- but I think it might be awkward not to... can you try to apply the data on defocus (if they change the names/description) field?

packages/llmd-serving/src/settings/TopologyConfigurationCreateEdit.tsx
andrewballantyne: Some forcing -- see other comment: https://github.com/opendatahub-io/odh-dashboard/pull/8400#discussion_r3521443771

packages/llmd-serving/src/settings/TopologyConfigurationCreateEdit.tsx
andrewballantyne: Basically:
* Apply the items anytime they change and you need to generate/render the YAML
* If they modify them -- so be it, you don't need to resync -- next time you get an update you can force it back on
* When you save, double down on the overrides

Basically, when you store YAML -- save it as-is -- but before rendering and "using" run it through a function that normalizes TopologyType Annotation, ResourceName, DisplayName & Description

packages/llmd-serving/src/settings/ConfigYAMLEditor.tsx
YuliaKrimerman: matching DashboardCodeEditor default

packages/llmd-serving/src/settings/TopologyConfigurationCreateEdit.tsx
```diff
@@ -0,0 +1,314 @@
...
+      }
+      loaded
+      empty={false}
+      provideChildrenPadding
+      data-testid="topology-config-create-edit-page"
+    >
+      <Form style={{ height: '100%' }}>
+        <Stack hasGutter>
+          <StackItem>
+            <K8sNameDescriptionField
```
andrewballantyne: Something feels off

Image

spacing is weird, can you look into it?

packages/llmd-serving/src/settings/TopologyConfigurationCreateEdit.tsx
andrewballantyne: @YuliaKrimerman need an answer here -- is this solving something or are you causing spacing issues by getting in between the Form & the elements rendered...

packages/llmd-serving/src/settings/TopologyConfigurationCreateEdit.tsx
```diff
@@ -0,0 +1,314 @@
...
+              state.sourceConfig,
+            )}`,
+          },
+        },
+      });
+    }
+    return '';
+  }, [existingConfig, state?.sourceConfig]);
+
+  const [yamlCode, setYamlCode] = React.useState(stringifiedConfig);
```
andrewballantyne:
```suggestion
  const [yamlCode, setYamlCode] = React.useState(() => {
    if (existingConfig) {
      return YAML.stringify(existingConfig);
    }
    if (state?.sourceConfig) {
      return YAML.stringify({
        ...state.sourceConfig,
        metadata: {
          ...state.sourceConfig.metadata,
          name: `${state.sourceConfig.metadata.name}-copy`,
          annotations: {
            ...state.sourceConfig.metadata.annotations,
            'openshift.io/display-name': `Copy of ${getDisplayNameFromK8sResource(
              state.sourceConfig,
            )}`,
          },
        },
      });
    }
    return '';
  });
```
You can pre-compute state on useState

packages/llmd-serving/src/settings/TopologyConfigurationCreateEdit.tsx
```diff
@@ -0,0 +1,314 @@
...
+  const [loading, setLoading] = React.useState(false);
+  const [error, setError] = React.useState<Error | undefined>();
+
+  React.useEffect(() => {
+    if (stringifiedConfig) {
+      setYamlCode(stringifiedConfig);
+    }
+  }, [stringifiedConfig]);
```
andrewballantyne: Which means you can drop this

packages/llmd-serving/src/settings/ConfigYAMLEditor.tsx
andrewballantyne: Can you look into implementing this? 
`frontend/src/components/MaxHeightCodeEditor.tsx`

We might get a better resizeable editor with the page... if you can set a minHeight of 400 -- since as you noted, it's better than 399 🙂

packages/llmd-serving/src/settings/ConfigYAMLEditor.tsx
```diff
@@ -0,0 +1,26 @@
+import * as React from 'react';
+import { CodeEditor, Language } from '@patternfly/react-code-editor';
+
+type ConfigYAMLEditorProps = {
+  code: string;
+  onCodeChange: (code: string) => void;
+};
+
+const ConfigYAMLEditor: React.FC<ConfigYAMLEditorProps> = ({ code, onCodeChange }) => (
```
andrewballantyne: This may be a wild ask -- I think it will give a better experience. Currently right now it's a bit mixed using the editor...

Lets write a wrapper for `code={code}` that is like `code={sanitizeStaticField(code)}` 

And we write a function that basically just "at runtime" and not in state, map the fields we have:
* Resource Name (metadata.name)
* Display Name (annotation)
* Description (annotation)
* Topology Type (label)

This way no matter what they try to do to the editor, on the next render loop it mutes the changes.

And we can also in this case mute this:
```
#...
metadata
  annotations:
    openshift.io/description: Distributed data-parallel groups for lsarge dense s.
      Oh yeah here is extra contentsdfsfsdfasdfsadfdd
    openshift.io/display-name: Copy of Dupe Multi-node Data Parallels
  name: foobar
  #...
  labels:
    opendatahub.io/config-type: workload-multi-node-data-parallel
#...
```
for something like:
```
metadata
  annotations:
    openshift.io/description: <Managed>
    openshift.io/display-name: <Managed>
  name: <Managed>
  labels:
    opendatahub.io/config-type: <fixed>
```

Now there is no real problem here... they can't edit it, we don't have to update it, it's smooth. I'd even just statically apply that as a deep merge using lodash -- we'd have to watch out for any potential bad yaml -- but I think you should be fine

@vconzola we should talk about this -- but this feels like a good way to manage this with fields that update the YAML...

packages/llmd-serving/src/settings/ConfigYAMLEditor.tsx
```diff
@@ -0,0 +1,26 @@
...
+  <div data-testid="config-yaml-editor" style={{ minHeight: '400px', height: '100%' }}>
+    <CodeEditor
+      code={code}
+      isUploadEnabled
+      isLanguageLabelVisible
+      language={Language.yaml}
+      height="400px"
+      emptyStateTitle="Add a topology configuration"
```
andrewballantyne: Curious -- I wonder if we want to call this more:

```suggestion
      emptyStateTitle="Add a ${nameOfType} topology-type configuration"
```

packages/llmd-serving/src/settings/TopologyConfigurationCreateEdit.tsx
```diff
@@ -0,0 +1,314 @@
...
+          const doc = YAML.parseDocument(prevYaml);
+          if (key === 'name') {
+            doc.setIn(['metadata', 'annotations', 'openshift.io/display-name'], value);
+          } else if (key === 'description') {
+            doc.setIn(['metadata', 'annotations', 'openshift.io/description'], value);
+          }
+          return doc.toString();
+        } catch {
+          return prevYaml;
+        }
```
andrewballantyne: you may want to re-evaluate this after my state comment on `ConfigYAMLEditor`

packages/llmd-serving/src/settings/TopologyConfigurationsTable.tsx
```diff
@@ -0,0 +1,217 @@
...
+        ...config.metadata,
+        annotations: {
+          ...config.metadata.annotations,
+          'opendatahub.io/disabled': currentlyEnabled ? 'true' : 'false',
+        },
+      },
+    };
+
+    try {
+      await patchLLMInferenceServiceConfig(config, updatedConfig);
```
andrewballantyne:
```suggestion
      await patchLLMInferenceServiceConfig(config, updatedConfig)
        .finally(() => {
          setTogglingConfigs((prev) => ({ ...prev, [configName]: false }));
        });
```
Wrong finally -- you don't want to toggle it back in the post action of patch, but the response call finalization. Network traffic being slow will get you into funky rendering states.

packages/llmd-serving/src/settings/TopologyConfigurationsTable.tsx
```diff
@@ -0,0 +1,217 @@
...
+    } catch (e) {
+      notification.error(
+        `Error ${currentlyEnabled ? 'disabling' : 'enabling'} configuration`,
+        e instanceof Error ? e.message : 'Unknown error',
+      );
+    } finally {
+      setTogglingConfigs((prev) => ({ ...prev, [configName]: false }));
```
andrewballantyne:
```suggestion
```

packages/llmd-serving/src/settings/TopologyConfigurationsTable.tsx
andrewballantyne: 👋 @YuliaKrimerman

packages/llmd-serving/src/settings/TopologyConfigurationsTable.tsx
```diff
@@ -0,0 +1,217 @@
...
+                {topologyTypes.map((tt) => (
+                  <DropdownItem
+                    key={tt}
+                    data-testid={`add-config-${tt}`}
+                    onClick={() => {
+                      setIsAddDropdownOpen(false);
+                      navigate(`add/${tt}`);
+                    }}
+                  >
+                    Add {TopologyTypeLabels[tt].toLowerCase()} configuration
```
andrewballantyne: Nit. Questionable logic -- `.toLowerCase()` might not work for fields that need a capital in the middle.

Labels should be structured for your use-case.

packages/llmd-serving/src/settings/TopologyConfigurationsTable.tsx
```diff
@@ -0,0 +1,217 @@
...
+                    </Button>,
+                  ]}
+                  aria-label="Add topology configuration"
+                  onClick={() => setIsAddDropdownOpen(!isAddDropdownOpen)}
+                  data-testid="add-topology-config-dropdown-toggle"
+                />
+              )}
+            >
+              <DropdownList>
+                {topologyTypes.map((tt) => (
```
andrewballantyne: Pretty sure this shouldn't contain the dupe -- despite what Vince's prototype says

Image

I googled it, and I got the same expectation of me (I get it, I'm bias, but @vconzola can validate this)

> Having a combo button (or dropdown) show duplicate values is generally bad UX and confusing for users. When identical labels appear, users cannot distinguish between choices or predict which option will trigger the correct action

packages/llmd-serving/src/settings/TopologyConfigurationsTable.tsx
```diff
@@ -0,0 +1,217 @@
...
+        headingLevel="h2"
+        titleText="No topology configurations found"
+        icon={CubesIcon}
+        data-testid="topology-configurations-empty-state"
+      >
+        <EmptyStateBody>
+          To get started, add a topology configuration using the button above.
+        </EmptyStateBody>
+      </EmptyState>
+    );
```
andrewballantyne: The button you're referring to is below this return statement in the other code path -- no?

packages/llmd-serving/src/settings/TopologyConfigurationsTable.tsx
```diff
@@ -0,0 +1,217 @@
...
+            </>
+          }
+          buttonActions={[
+            {
+              label: 'Delete',
+              variant: 'danger',
+              onClick: handleDelete,
+              isLoading: isDeleting,
+              isDisabled: isDeleting,
+              dataTestId: 'delete-topology-config-confirm',
```
andrewballantyne, line 201: Something is off about the delete...

I deleted, it didn't remove the row (showed me the dialog 🥳 though).

Tried again and got a notification error. Table still shows the top row that "I deleted"

Image

andrewballantyne: Tried to toggle after this -- and that didn't work either -- yeah, might be a data break on my side. Network crashed or something around the websocket.

Double check the flow, but I think it's fine.

andrewballantyne (changes requested):

packages/llmd-serving/src/settings/TopologyConfigurationCreateEdit.tsx
YuliaKrimerman: Stack hasGutter is needed for the StackItem isFilled on the YAML editor to expand vertically. Removing it collapses the editor. I'll look into whether Form's built-in spacing can handle this without Stack, or if the spacing issue is something else.

packages/llmd-serving/src/settings/TopologyConfigurationsTable.tsx
YuliaKrimerman: Which shared table component are you referring to? Is it the one in frontend/src/components/table/?

packages/llmd-serving/src/settings/TopologyConfigurationCreateEdit.tsx
YuliaKrimerman: Screenshot 2026-07-06 at 10 42 40 AM
Fixed

packages/llmd-serving/src/settings/TopologyConfigurationsTable.tsx
YuliaKrimerman: works on my machine lol - couldn't reproduce

YuliaKrimerman:
https://github.com/user-attachments/assets/f0be3246-48a5-4af3-9539-138b7b926361

cc @vconzola -

packages/llmd-serving/src/settings/TopologyConfigurationsTable.tsx
andrewballantyne: Yup, that's the one that helps build tables nicer than the raw PF one

packages/llmd-serving/src/settings/TopologyConfigurationsTable.tsx
andrewballantyne: yeah, that makes sense -- I think my websocket closed or something

packages/llmd-serving/src/settings/TopologyConfigurationCreateEdit.tsx
```diff
@@ -0,0 +1,309 @@
...
+      navigate('..');
+    } catch (e) {
+      const err = e instanceof Error ? e : new Error('Unknown error');
+      setError(err);
+      notification.error(
+        `Error ${isEditMode ? 'updating' : 'creating'} configuration`,
+        err.message,
+      );
+    } finally {
+      setLoading(false);
```
andrewballantyne, line 317: I think this is what is causing your toggle delay -- you need to wait for the API to finish before loading is done...

packages/llmd-serving/src/settings/TopologyConfigurationCreateEdit.tsx
```diff
@@ -0,0 +1,309 @@
...
+          },
+        },
+      };
+    }
+    return undefined;
+  }, [existingConfig, state?.sourceConfig]);
+
+  const k8sNameDesc = useK8sNameDescriptionFieldData({
+    initialData: initialResource,
+  });
```
andrewballantyne, line 133: When you duplicate, you don't have the editable nature of the resource... look at the type for the data sent into this function. You're going to need to make sure when you duplicate a resource, it's not a fixed k8s name

andrewballantyne: Seems to also cause problems with the auto gen of the new k8s name -- duplicate feels a bit broken

packages/llmd-serving/src/settings/TopologyConfigurationCreateEdit.tsx
```diff
@@ -0,0 +1,309 @@
...
+
+  const [yamlCode, setYamlCode] = React.useState(() => {
```
andrewballantyne, line 135: backspace on the last line of your yaml seems to remove the line -- I'm on "replica" in my example, and I hit backspace to replace the replica number and it sends me to the new blank line it created at the bottom of the yaml editor

andrewballantyne: Screencapture

https://github.com/user-attachments/assets/497be02e-0114-4659-a0dc-da7b4bd4e405

packages/llmd-serving/src/settings/ConfigYAMLEditor.tsx
```diff
@@ -0,0 +1,31 @@
...
+  onCodeChange,
+  topologyTypeLabel,
+}) => (
+  <div data-testid="config-yaml-editor" style={{ minHeight: '400px', height: '100%' }}>
+    <CodeEditor
+      code={code}
+      isUploadEnabled
+      isLanguageLabelVisible
+      language={Language.yaml}
+      height="400px"
```
andrewballantyne, line 21: Nit: I noticed on PF this has a fit-size value

```
height    string | 'sizeToFit'
```
https://www.patternfly.org/components/code-editor/#codeeditor

packages/llmd-serving/src/settings/TopologyConfigurationCreateEdit.tsx
andrewballantyne, line 1: You could probably write some co-located hooks to make this more consumable -- but we are shifting to more Claude heavy evaluations of code -- that's probably a human need more than a code-cleanness so Claude stays true.

packages/llmd-serving/src/settings/TopologyConfigurationRow.tsx
```diff
@@ -0,0 +1,106 @@
...
+  const navigate = useNavigate();
+  const configName = config.metadata.name;
+  const displayName = getDisplayNameFromK8sResource(config);
+  const description = getDescriptionFromK8sResource(config);
+  const preInstalled = isConfigPreInstalled(config);
+  const enabled = isConfigEnabled(config);
+  const topologyType = getConfigTopologyType(config);
+  const isDashboardCreated =
+    config.metadata.labels?.[DASHBOARD_RESOURCE_LABEL] === 'true' && !preInstalled;
+  const duplicatePath = `add/${topologyType ?? TopologyType.SINGLE_NODE}`;
```
andrewballantyne, line 41: Interesting nit -- you split this up for reuse but the entire block its used in is also reuse-able.

packages/llmd-serving/src/settings/TopologyConfigurationsTable.tsx
```diff
@@ -0,0 +1,216 @@
...
+  const { dashboardNamespace } = useDashboardNamespace();
+  const notification = useNotification();
+  const [isAddDropdownOpen, setIsAddDropdownOpen] = React.useState(false);
+  const [togglingConfigs, setTogglingConfigs] = React.useState<Record<string, boolean>>({});
+  const [deleteConfig, setDeleteConfig] = React.useState<LLMInferenceServiceConfigKind>();
+  const [isDeleting, setIsDeleting] = React.useState(false);
+
+  const topologyTypes = Object.values(TopologyType);
+  const primaryTopologyType = TopologyType.SINGLE_NODE;
+  const dropdownTopologyTypes = topologyTypes.filter((t) => t !== primaryTopologyType);
```
andrewballantyne, line 52: Nit. Technically these are static values -- you don't need to compute on every render.

packages/llmd-serving/src/settings/TopologyConfigurationsTable.tsx
```diff
@@ -0,0 +1,216 @@
...
+        <ContentModal
+          title="Delete llm-d topology configuration?"
+          onClose={() => setDeleteConfig(undefined)}
+          variant="small"
+          dataTestId="delete-topology-config-modal"
+          contents={
+            <>
+              Delete <strong>{getDisplayNameFromK8sResource(deleteConfig)}</strong>? This cannot be
+              undone. Out-of-the-box configurations cannot be deleted from the cluster from this UI
+              (disable them instead).
```
andrewballantyne, line 191: Such an interesting call out -- not sure we usually tell them about other lines when trying to delete an item.

Microcopy will likely pivot on this -- was this Claude or UX?

packages/llmd-serving/src/settings/ConfigYAMLEditor.tsx
```diff
@@ -0,0 +1,31 @@
...
+  topologyTypeLabel?: string;
+};
+
+const ConfigYAMLEditor: React.FC<ConfigYAMLEditorProps> = ({
+  code,
+  onCodeChange,
+  topologyTypeLabel,
+}) => (
+  <div data-testid="config-yaml-editor" style={{ minHeight: '400px', height: '100%' }}>
+    <CodeEditor
```
andrewballantyne, line 16: I think I noticed this in another PR review -- but `MaxHeightCodeEditor` solves a couple problems here I imagine -- including you're not theming the editor. Perhaps follow up in another PR?

packages/llmd-serving/src/settings/ConfigYAMLEditor.tsx
andrewballantyne, line 1: Ooof, upload is going to be annoying -- I just uploaded a `.png` file 😠 

Seems we may need to implement our own -- not sure why PF doesn't support this out of the box...

andrewballantyne: Interestingly enough -- the code editor probably needs more TLC and we should sync with Mike on this for his feature too... but it could be a follow up next week.

Image

I uploaded a Job yaml I have sitting in my yaml -- then clicked submit.

It produced me something default -- good ol' k8s evaluations...

I think what we want to do is:
* Only let YAML selections
* Only let LLMInferenceServiceConfigs (or maybe LLMIS too since they are pretty much the same thing) be selected

Or when it happens, we report error and lock down the form. @vconzola can help us make those calls... but it is what it is I think.

packages/llmd-serving/src/settings/TopologyConfigurationCreateEdit.tsx
```diff
@@ -0,0 +1,405 @@
...
+          namespace: dashboardNamespace,
+          labels: {
+            ...parsed.metadata.labels,
+            [CONFIG_TYPE_LABEL]: resolvedTopologyType,
+            [DASHBOARD_RESOURCE_LABEL]: 'true',
+          },
+          annotations: {
+            ...parsed.metadata.annotations,
+            'openshift.io/display-name': k8sNameDesc.data.name,
+            'openshift.io/description': k8sNameDesc.data.description,
```
andrewballantyne, line 298: If this is empty -- you put empty on the resource -- probably should be optional.
```
    openshift.io/description: ""
```

andrewballantyne (approved):
Bunch of nits, but lets merge and you can address them in the future

YuliaKrimerman:
@andrewballantyne Needs lgtm again because of types failure after rebase

andrewballantyne (approved):

andrewballantyne:
/lgtm

🤞

PR #8403, opened 2026-07-03
feat(llmd-serving): hide hardware profile for non-single-node topologies

YuliaKrimerman:
https://issues.redhat.com/browse/RHOAIENG-73694

## Description

Hides the Hardware Profile section in the model deployment wizard when a non-single-node topology type is selected for llm-d deployments. Multi-node, single-node disaggregated, and multi-node disaggregated topologies do not use hardware profiles — their resource configuration is defined in the topology configuration template instead.

### Behavior

| Topology type | Hardware Profile |
|---|---|
| Single node (default) | Visible |
| Multi-node | Hidden |
| Single node disaggregated | Hidden |
| Multi-node disaggregated | Hidden |
| Non-llm-d deployment methods | Always visible (no change) |

## How Has This Been Tested?
Screenshot 2026-07-03 at 11 57 34 AM
Screenshot 2026-07-03 at 11 57 44 AM
Screenshot 2026-07-03 at 11 57 53 AM

- Manually tested on a live cluster with `llmdTopologyConfigs` feature flag enabled
- Verified HWP appears for single-node, disappears for multi-node/disaggregated topologies
- Verified switching between topologies toggles HWP visibility correctly
- Verified the wizard Next button works correctly when HWP is hidden (validation skipped)
- Verified non-llm-d deployment methods (legacy, simple vLLM) are unaffected
- Cypress mock tests pass locally

## Test Impact

**Updated:** `packages/cypress/cypress/tests/mocked/modelServing/modelServingLlmdTopology.cy.ts`
- Added `describe('hardware profile visibility')` block with 3 test cases

## Request review criteria:

- [x] The code follows our [Best Practices](/docs/best-practices.md) (React coding standards

## Summary by CodeRabbit

* **New Features**
  * Hardware profile settings now automatically hide for non-single-node topology selections in the deployment wizard.

* **Bug Fixes**
  * Step validation now correctly accounts for topology-based hardware profile visibility, reducing false validation errors.
  * Topology changes now properly toggle the hardware profile field when switching between supported modes.

* **Tests**
  * Expanded test coverage for multiple topology types, including disaggregated variants, and improved advanced settings navigation checks.

### DIFF
--- a/packages/cypress/cypress/tests/mocked/modelServing/modelServingLlmdTopology.cy.ts
+++ b/packages/cypress/cypress/tests/mocked/modelServing/modelServingLlmdTopology.cy.ts
@@ -23,6 +23,7 @@ import {
   TemplateModel,
 } from '../../../utils/models';
 import { modelServingGlobal, modelServingWizard } from '../../../pages/modelServing';
+import { hardwareProfileSection } from '../../../pages/components/HardwareProfileSection';
 
 const buildTopologyConfig = (
   name: string,
@@ -35,6 +36,16 @@ const mockTopologyConfigs = [
   buildTopologyConfig('single-node-config', 'Single Node Config', TopologyType.SINGLE_NODE),
   buildTopologyConfig('multi-node-config', 'Multi-node Data Parallel', TopologyType.MULTI_NODE),
   buildTopologyConfig('disabled-config', 'Disabled Config', TopologyType.MULTI_NODE, true),
+  buildTopologyConfig(
+    'single-node-pd-config',
+    'Single Node P/D',
+    TopologyType.SINGLE_NODE_DISAGGREGATED,
+  ),
+  buildTopologyConfig(
+    'multi-node-pd-config',
+    'Multi-node P/D',
+    TopologyType.MULTI_NODE_DISAGGREGATED,
+  ),
 ];
 
 const mockRouterConfigs = [
@@ -139,6 +150,13 @@ const navigateToModelDeploymentStep = () => {
   modelServingWizard.findNextButton().click();
 };
 
+const navigateToAdvancedSettings = () => {
+  modelServingWizard.findModelDeploymentNameInput().type('test-model');
+  modelServingWizard.selectDeploymentMethodByKey('llm-inference-service-llmd');
+  hardwareProfileSection.findSelect().should('contain.text', 'Small');
+  modelServingWizard.findNextButton().should('be.enabled').click();
+};
+
 describe('Model Serving LLMD Topology & Routing', () => {
   describe('topology type field', () => {
     it('should show topology type dropdown when flag enabled and llm-d active', () => {
@@ -179,11 +197,11 @@ describe('Model Serving LLMD Topology & Routing', () => {
         'pf-m-aria-disabled',
       );
       cy.findByTestId(`topology-type-${TopologyType.SINGLE_NODE_DISAGGREGATED}`).should(
-        'have.class',
+        'not.have.class',
         'pf-m-aria-disabled',
       );
       cy.findByTestId(`topology-type-${TopologyType.MULTI_NODE_DISAGGREGATED}`).should(
-        'have.class',
+        'not.have.class',
         'pf-m-aria-disabled',
       );
     });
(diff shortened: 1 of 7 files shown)


packages/model-serving/src/components/deploymentWizard/steps/ModelDeploymentStep.tsx
```diff
@@ -21,12 +21,25 @@ type ModelDeploymentStepProps = {
   hideProjectSection?: boolean;
 };
 
+const SINGLE_NODE_TOPOLOGY = 'workload-single-node';
```
andrewballantyne: Enum?

packages/model-serving/src/components/deploymentWizard/useDeploymentWizardValidation.tsx
```diff
@@ -19,6 +19,17 @@ import { modelFormatFieldSchema } from './fields/ModelFormatField';
 import { isValidProjectName } from './fields/ProjectSection';
 import { getStateKey } from './dynamicFormUtils';
 
+const SINGLE_NODE_TOPOLOGY = 'workload-single-node';
+
+const isHardwareProfileHidden = (state: Record<string, unknown>): boolean => {
```
andrewballantyne: Duplicated utility & not using the enum?

andrewballantyne (changes requested):

andrewballantyne:
image
See deployment resource field
image

I think you'll want accelerator configs to show up here too for single-node...

YuliaKrimerman:
@andrewballantyne Fixed and addressed comments
Screenshot 2026-07-03 at 1 03 13 PM

packages/model-serving/src/components/deploymentWizard/topologyUtils.ts
```diff
@@ -0,0 +1,16 @@
+const TOPOLOGY_TYPE_FIELD_KEY = 'llmd-serving/topology-type';
+const SINGLE_NODE_TOPOLOGY = 'workload-single-node';
```
andrewballantyne, line 11: Is this not just the `TOPOLOGY_TYPE` enum value?

packages/model-serving/src/components/deploymentWizard/topologyUtils.ts
```diff
@@ -0,0 +1,16 @@
...
+
+const getTopologyType = (state: Record<string, unknown>): string | undefined => {
+  const topologyData: unknown = state[TOPOLOGY_TYPE_FIELD_KEY];
+  if (topologyData != null && typeof topologyData === 'object' && 'topologyType' in topologyData) {
+    const record = topologyData as Record<string, unknown>; // eslint-disable-line @typescript-eslint/consistent-type-assertions
```
andrewballantyne: `as` is no good haha

You're trying to type something here -- can you infer state more accurately from somewhere else to help avoid the `unknown` value set?

andrewballantyne: Actually seems you already have a type checker/guard -- maybe just use `isTopologyTypeFieldData` 

See coderabbit... https://github.com/opendatahub-io/odh-dashboard/pull/8403/changes#r3521152820

andrewballantyne:
> @andrewballantyne Fixed and addressed comments Screenshot 2026-07-03 at 1 03 13 PM

https://github.com/opendatahub-io/odh-dashboard/pull/8353 able to solve the auto selection efforts here too?

packages/model-serving/src/components/deploymentWizard/topologyUtils.ts
YuliaKrimerman: Can't import TopologyType or isTopologyTypeFieldData from @odh-dashboard/llmd-serving here — llmd-serving depends on model-serving, so importing the other way would be circular. I've added a local typeguard (isTopologyFieldData) that mirrors the check without the cross-package import, and documented the constraint. The string constant 'workload-single-node' mirrors TopologyType.SINGLE_NODE. If we want to share the enum, we'd need to move it to a shared package like k8s-core — happy to do that as a follow-up if preferred.

YuliaKrimerman: The right fix is to move TopologyType to @odh-dashboard/k8s-core where both packages can import it — k8s-core already has similar enums like KnownLabels and MetadataAnnotation. That touches ~15 files in llmd-serving though, so I'd prefer to do it as a follow-up to keep this PR small. The string constant here mirrors TopologyType.SINGLE_NODE with a JSDoc cross-reference.

andrewballantyne: Log a tech debt -- this is something we can work with @emilys314 when she is back.

This is not good organization. Claude doesn't manage this stuff well enough as it is -- I don't want to see dupe code because we are not getting good boundaries. Lets work with Emily and figure out plan of action

packages/model-serving/src/components/deploymentWizard/topologyUtils.ts
```diff
@@ -0,0 +1,24 @@
...
+ */
+
+const TOPOLOGY_TYPE_FIELD_KEY = 'llmd-serving/topology-type';
+const SINGLE_NODE_TOPOLOGY = 'workload-single-node';
+
+const isTopologyFieldData = (data: unknown): data is { topologyType: string } =>
+  data != null &&
+  typeof data === 'object' &&
+  'topologyType' in data &&
+  typeof (data as Record<string, unknown>).topologyType === 'string'; // eslint-disable-line @typescript-eslint/consistent-type-assertions
```
andrewballantyne: You just validated it had the property...

Went to verify with my IDE

Image

andrewballantyne: Sorry -- helps if I keep the logic 😂 

Image


```suggestion
  typeof data.topologyType === 'string';
```

packages/model-serving/src/components/deploymentWizard/topologyUtils.ts
```diff
@@ -0,0 +1,24 @@
+/**
+ * Topology-aware visibility helpers for the deployment wizard.
+ *
+ * Cannot import TopologyType enum or isTopologyTypeFieldData from @odh-dashboard/llmd-serving
+ * because llmd-serving depends on model-serving — importing the other way would be circular.
+ * These string constants mirror TopologyType.SINGLE_NODE from llmd-serving/types.
+ */
```
andrewballantyne, line 8:
```suggestion
  * Tech Debt Ticket: RHOAIENG-#####
  */
```
Lets make sure you also state the ticket here once you log it
https://github.com/opendatahub-io/odh-dashboard/pull/8403/changes#r3521316799

YuliaKrimerman: https://redhat.atlassian.net/browse/RHOAIENG-73746

packages/cypress/cypress/tests/mocked/modelServing/modelServingLlmdTopology.cy.ts
```diff
@@ -139,6 +150,13 @@ const navigateToModelDeploymentStep = () => {
   modelServingWizard.findNextButton().click();
 };
 
+const navigateToAdvancedSettings = () => {
```
andrewballantyne: No page objects in this space to attach this to?

andrewballantyne (approved):

PR #8404, opened 2026-07-03
fix:added fix for local queues visibility based on namespaces

claudialphonse78:
https://issues.redhat.com/browse/RHOAIENG-67395

## Description

When a project has Kueue enabled, hardware profiles are filtered to only show those with a
`localQueueName`. The bug was that we only checked whether a profile *had* a queue name, never
whether that queue actually exists in the current project's namespace. `LocalQueue` resources are
namespace-scoped, so a profile configured with `localQueueName: team-queue` would show up in any
Kueue-enabled project even if that queue was never created there, and selecting it would fail
silently at workload submission.

**What this PR does:**

- Reads the `LocalQueue` resources already fetched for the current namespace and cross-references
  them against each profile's `localQueueName` before showing it in the dropdown
- Profiles whose queue doesn't exist in the namespace are hidden
- While the queue list is still loading, all profiles are shown and auto-selection is deferred
  so we don't immediately pick a profile whose queue turns out to be missing
- In edit mode, the previously-selected profile stays visible even if its queue is gone, so the
  existing LQ-missing warning can still fire

## Tested on cluster

Verified in `kueue-phase2-dev` (has `default` and `constrained-queue` LocalQueues):
1. Verification in `kueue-integration-test`

https://github.com/user-attachments/assets/92069718-f9f6-4c60-aa9b-3f2de931c285

2. Verification in `kueue-phase2-dev`

https://github.com/user-attachments/assets/d0961cca-9c4b-4a10-b748-0f0156e2fcf6

- `GPU Profile Test`, `Kueue Dev Profile`, `Test GPU Profile` (all referencing `localQueueName: default`) showed up correctly
- `Test Hardware Profile scoping` (referencing `localQueueName: constrained-queue`) showed up in `kueue-phase2-dev` but was hidden in `kueue-integration-test`, which does not have `constrained-queue`
- `Test Hardware -no LQ` (`localQueueName: no-lq`), `Test Hardware profile-no-lq2` (`localQueueName: no-lq2`), and `test` (`localQueueName: f`) were all hidden since none of those queues exist in the namespace
- Created a new profile `Test Hardware Profile scoping 2` with `localQueueName: workbench-queue` — it was hidden in `kueue-phase2-dev` and appeared correctly in `kueue-integration-test` where that queue exists

## Test Impact

- New `kueueUtils.spec.ts` covering all filtering branches and loading/error/ready states of the
  queue names hook
- `useHardwareProfileConfig.spec.ts` updated for the new hook return type
- `HardwareProfileSelect.spec.tsx` has a new test for the edit-mode profile preservation
- `workbenchHardwareProfiles.cy.ts` updated to assert the profile is hidden in the CREATE flow
  when its queue is missing (the old test expected a warning that no longer fires)

## Request review criteria:

- [x] Testing instructions have been added in the PR body (for PRs involving changes that are not
  immediately obvious).
- [x] The developer has added tests or explained why testing cannot be added (unit or cypress tests
  for related changes)
- [x] The code follows our [Best Practices](/docs/best-practices.md) (React coding standards,
  PatternFly usage, performance considerations)

If you have UI changes:

- [ ] The developer has tested their solution on a cluster by using the image produced by the PR
  to `main`

## Summary by CodeRabbit

* **Bug Fixes**
  * Hardware profile selection now hides Kueue profiles when their referenced LocalQueue is unavailable.
  * Kueue-driven auto-selection waits for LocalQueue data to finish loading, and won’t select profiles with missing queues.
  * In edit mode, the currently selected profile remains visible even if its LocalQueue becomes unavailable (with a queue-missing indicator).
  * If LocalQueue data fails to load, Kueue profiles are still selectable (filtering is bypassed).

### DIFF
--- a/frontend/src/concepts/hardwareProfiles/__tests__/useHardwareProfileConfig.spec.ts
+++ b/frontend/src/concepts/hardwareProfiles/__tests__/useHardwareProfileConfig.spec.ts
@@ -766,20 +766,6 @@ describe('useHardwareProfileConfig — Kueue auto-selection guard', () => {
     });
   });
 
-  it('defers auto-selection while localQueues are still loading', () => {
-    const { result } = renderHook(() => useHardwareProfileConfig(), {
-      wrapper: makeKueueProjectWrapper({
-        data: [],
-        loaded: false,
-        error: undefined,
-        refresh: jest.fn(),
-      }),
-    });
-
-    // Guard fires: ONLY_KUEUE_PROFILES && localQueues not loaded → return early
-    expect(result.current.formData.selectedProfile).toBeUndefined();
-  });
-
   it('auto-selects first matching Kueue profile once localQueues have loaded', () => {
     const { result } = renderHook(() => useHardwareProfileConfig(), {
       wrapper: makeKueueProjectWrapper({
--- a/frontend/src/concepts/hardwareProfiles/__tests__/useHardwareProfileConfig.spec.ts
+++ b/frontend/src/concepts/hardwareProfiles/__tests__/useHardwareProfileConfig.spec.ts
@@ -1,10 +1,18 @@
+import * as React from 'react';
 import { act } from '@testing-library/react';
-import { testHook } from '@odh-dashboard/jest-config/hooks';
+import { renderHook, testHook } from '@odh-dashboard/jest-config/hooks';
 import * as areasUtils from '@odh-dashboard/plugin-core/areas';
+import { SchedulingType } from '@odh-dashboard/k8s-core';
 import { mockHardwareProfile } from '#~/__mocks__/mockHardwareProfile';
+import { mockProjectK8sResource } from '#~/__mocks__/mockProjectK8sResource';
 import { useHardwareProfileConfig } from '#~/concepts/hardwareProfiles/useHardwareProfileConfig';
 import * as reduxSelectors from '#~/redux/selectors';
 import * as useHardwareProfilesModule from '#~/pages/hardwareProfiles/useHardwareProfilesByFeatureVisibility';
+import { useAvailableLocalQueueNames } from '#~/concepts/hardwareProfiles/kueueUtils';
+import {
+  ProjectDetailsContext,
+  type ProjectDetailsContextType,
+} from '#~/pages/projects/ProjectDetailsContext';
 
 jest.mock('@odh-dashboard/plugin-core/areas', () => ({
   ...jest.requireActual('@odh-dashboard/plugin-core/areas'),
@@ -17,11 +25,17 @@ jest.mock('#~/redux/selectors', () => ({
   useDashboardNamespace: jest.fn(),
 }));
 
+jest.mock('#~/concepts/hardwareProfiles/kueueUtils', () => ({
+  ...jest.requireActual('#~/concepts/hardwareProfiles/kueueUtils'),
+  useAvailableLocalQueueNames: jest.fn(),
+}));
+
 const mockUseIsAreaAvailable = jest.mocked(areasUtils.useIsAreaAvailable);
 const mockUseHardwareProfiles = jest.mocked(
   useHardwareProfilesModule.useHardwareProfilesByFeatureVisibility,
 );
 const mockUseDashboardNamespace = jest.mocked(reduxSelectors.useDashboardNamespace);
+const mockUseAvailableLocalQueueNames = jest.mocked(useAvailableLocalQueueNames);
 
 describe('useHardwareProfileConfig', () => {
   beforeEach(() => {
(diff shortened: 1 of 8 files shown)


frontend/src/concepts/hardwareProfiles/HardwareProfileSelect.tsx
```diff
@@ -280,7 +292,11 @@ const HardwareProfileSelect: React.FC<HardwareProfileSelectProps> = ({
       enabledProfiles.push(initialHardwareProfile);
     }
     const orderedProfiles = orderHardwareProfiles(enabledProfiles, hardwareProfileOrder);
-    return filterProfilesByKueue(orderedProfiles, kueueFilteringState).filter((profile) =>
+    return filterProfilesByKueue(
```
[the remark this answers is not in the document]
dpanshug: @claudialphonse78 ^

frontend/src/concepts/hardwareProfiles/kueueUtils.ts
```diff
@@ -6,6 +6,8 @@ import {
   type ProjectKind,
 } from '@odh-dashboard/k8s-core';
 import { useIsAreaAvailable, SupportedArea } from '@odh-dashboard/plugin-core/areas';
+import { ProjectDetailsContext } from '#~/pages/projects/ProjectDetailsContext';
+
```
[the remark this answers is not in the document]
dpanshug: @claudialphonse78 ^

claudialphonse78:
Fixed the case wherein if a hardware profile was already selected which had no local queue associated with it added a info icon to to clarify its usage
Screenshot 2026-07-08 at 7 10 59 PM

packages/hardware-profiles/shared/useHardwareProfileConfig.ts
```diff
@@ -207,6 +218,7 @@ export const useHardwareProfileConfig = (
     dashboardProfiles,
     dashboardNamespace,
     kueueFilteringState,
+    availableLocalQueueNames,
```
dpanshug, line 231: need to add a guard that skips the new-form auto-selection when Kueue filtering is active but local queues haven't loaded yet

dpanshug (changes requested):

frontend/src/concepts/hardwareProfiles/__tests__/useHardwareProfileConfig.spec.ts
dpanshug, line 1: There is no test verifying that auto-selection is deferred when queues are still loading (loaded: false)

dpanshug (changes requested):

packages/hardware-profiles/shared/useHardwareProfileConfig.ts
```diff
@@ -143,6 +147,12 @@ export const useHardwareProfileConfig = (
   const isFormDataValid = React.useMemo(() => isHardwareProfileConfigValid(formData), [formData]);
   const { kueueFilteringState } = useKueueConfiguration(currentProject);
 
+  const { data: lqData, loaded: lqLoaded, error: lqError } = localQueues;
+  const availableLocalQueueNames = React.useMemo(() => {
```
kanishka-commits, line 152: `loading` and `error` are both undefined here. Fine for fail-open, but auto-select can't distinguish loading from fetch failure.

packages/hardware-profiles/shared/useHardwareProfileConfig.ts
```diff
@@ -184,6 +194,7 @@ export const useHardwareProfileConfig = (
         const filteredProfiles = filterProfilesByKueue(
           profiles.filter(isHardwareProfileEnabled),
           kueueFilteringState,
+          availableLocalQueueNames,
```
kanishka-commits, line 207: While queues load, `availableLocalQueueNames` is `undefined`, so every profile passes the filter and the first is auto-selected. That selection sticks because `selectedProfile` early-returns. Defer auto-select until `status === 'ready'` (or fail-open on error).

kanishka-commits (changes requested):
Overall, this looks good. I've left a few review comments to address.

packages/hardware-profiles/shared/HardwareProfileSelect.tsx
```diff
@@ -258,18 +304,32 @@ const HardwareProfileSelect: React.FC<HardwareProfileSelectProps> = ({
...
-    if (initialHardwareProfile && !isHardwareProfileEnabled(initialHardwareProfile)) {
-      enabledProfiles.push(initialHardwareProfile);
+    const filteredProfiles = filterProfilesByKueue(
+      profiles.filter(isHardwareProfileEnabled),
+      kueueFilteringState,
+      availableLocalQueueNames,
+    );
+    if (initialHardwareProfile && !filteredProfiles.includes(initialHardwareProfile)) {
+      filteredProfiles.push(initialHardwareProfile);
     }
```
dpanshug, line 336: this function is called twice in the project-scoped path - once for project profiles (line 339) and once for global profiles (line 343). Since initialHardwareProfile can only be reference-equal to entries in one of those arrays, the includes() check on line 329 will fail for the other list, and the profile gets pushed into both. In edit mode, if the initial profile's queue is missing, it shows up under both "Project-scoped" and "Global-scoped" groups.

We need to check the profile's namespace before pushing, which will ensure the profile is only rescued in the group it actually belongs to.

dpanshug (changes requested):

dpanshug (approved):
/lgtm - working as expected

PR #8418, opened 2026-07-06
docs: add package topology and import rules

andyatmiami:
## Description

Codify the dependency hierarchy, import boundaries, and core library stack decisions aligned on July 2, 2026 with Christian Vogt, Andrew Ballantyne, Lucas Fernandez Aragon, and Paulo Rego.

Key rules established:
- Package layer hierarchy (core libs → feature packages → app shell → distributions)
- Hub-and-spoke pattern for the serving family (model-serving as hub)
- Type-only imports across features always permitted; cross-feature runtime imports blocked unless within a hub-and-spoke group
- plugin-core vs ui-core remain separate packages
- Decision flow for code placement during @odh-dashboard/internal extraction

## How Has This Been Tested?
n/a - doc only

## Test Impact
n/a - doc only

## Request review criteria:
n/a - doc only

If you have UI changes: 

n/a - doc only

## Summary by CodeRabbit

* **Documentation**
  * Added guidance for monorepo package topology, dependency direction, and allowed import paths.
  * Clarified runtime import boundaries between core and feature packages, including type-only and extension-point exceptions.
  * Documented the hub-and-spoke model for model-serving and its unidirectional cross-feature runtime rules.
  * Defined core library stack responsibilities and OpenShift-specific code placement guidance.
  * Specified Cypress E2E test placement and import restrictions for feature vs shared test infrastructure.

### DIFF
--- a/docs/package-topology.md
+++ b/docs/package-topology.md
@@ -1,0 +1,239 @@
+# Package Topology and Import Rules
+
+Rules governing the dependency hierarchy and import boundaries for all packages in the
+ODH Dashboard monorepo.
+
+> **Aligned**: July 2, 2026. Attendees: Andy Stoneberg, Christian Vogt,
+> Lucas Fernandez Aragon, Andrew Ballantyne, Paulo Rego.
+
+---
+
+## 1. Package Layers
+
+Packages fall into distinct layers. A package may only have runtime dependencies on
+packages in its own layer or below — never above.
+
+| Layer | What lives here | Packages |
+|-------|-----------------|----------|
+| **Core shared libraries** | Extension framework, shared UI, K8s types, pure utilities. | `k8s-core`, `plugin-core`, `ui-core`, `app-config`, `foundation` (not yet extracted) |
+| **Feature packages** | Domain-specific frontend functionality and backend infrastructure. Depend on core shared libraries and the app shell. | `model-serving`, `kserve`, `notebooks`, `gen-ai`, `model-registry`, `core-bff`, etc. |
+| **App shell** | Shared framework (masthead, sidebar, routing, error boundary) that distributions extend. | `distributions/base/` |
+| **Distributions** | Concrete, deployable dashboard variants. Composition roots that wire everything together. | `distributions/rhaii/`, `distributions/rhoai/` (future) |
+
+Dependencies flow downward. The diagram below shows the full picture — arrows point
+from consumer to dependency:
+
+```text
+  Distributions (composition roots)
+  ─────────────────────────────────
+  distributions/rhaii/       distributions/rhoai/ (future)
+            │                        │
+            └───────────┬────────────┘
+                        │
+       ┌────────────────┼──────────────────┐
+       ▼                ▼                  ▼
+  ┌──────────┐   ┌──────────┐      ┌────────────┐
+  │model-    │   │ gen-ai   │      │  core-bff  │
+  │serving   │   └────┬─────┘      └─────┬──────┘
+  │ (hub)    │        │                  │         Feature
+  ├──────────┤        │                  │         Packages
+  │▲ ▲ ▲    │        │                  │         & Backend
+  ││ │ │    │        │                  │         Infrastructure
+  │kserve   │        │                  │
+  │nim-srv  │        │                  │
+  │llmd-srv │        │                  │
+  └────┬─────┘        │                  │
+       └──────────────┼──────────────────┘
+                      │
+                      ▼
+              ┌──────────────┐
+              │   App Shell  │
+              │  dist/base/  │
+              └──────┬───────┘
+                     │
+       ┌─────────────┼─────────────┐
+       ▼             ▼             ▼
+  ┌──────────┐ ┌───────────┐ ┌──────────┐
+  │ ui-core  │ │plugin-core│ │app-config│   Core Shared
+  └────┬─────┘ └─────┬─────┘ └──────────┘   Libraries
+       │             │
+       └──────┬──────┘
+              ▼
+        ┌───────────┐
+        │  k8s-core  │
+        └─────┬─────┘
+              ▼
+        ┌───────────┐
+        │foundation │  (not yet extracted — zero @odh-dashboard/* deps)
+        └───────────┘
+```
+
+> **Note**: Tooling packages (`eslint-config`, `jest-config`, `tsconfig`) are outside
+> the layer hierarchy. They are devDependencies only and never imported at runtime.
+
+---
+
+## 2. The Serving Hub-and-Spoke
+
+Model serving and its related packages follow a hub-and-spoke pattern. Model-serving
+is the hub; `kserve`, `nim-serving`, `llmd-serving`, and `model-serving-backport` are
+spokes.
+
+Evidence:
+- Every spoke declares `reliantAreas` targeting model-serving areas
+- Every spoke registers zero standalone routes — all extensions target model-serving extension points
+- The dependency is purely unidirectional: spokes depend on hub, hub has zero dependencies on spokes
+
+This is the only group of feature packages with cross-feature runtime imports today.
+No other feature packages import from sibling features.
+
+---
+
+## 3. Import Rules
+
+### Rule 1: Layer boundary
+
+Feature packages may depend on core shared libraries. They must not depend on the app
+shell or on distributions. Core shared libraries must not depend on feature packages.
+
+### Rule 2: Type-only imports across features
+
+`import type { ... }` across feature packages is always permitted. Type imports are
+erased at compile time and create no runtime dependency or bundle impact.
+
+### Rule 3: Hub-and-spoke runtime imports
+
+Feature packages in a hub-and-spoke group may have runtime imports from each other,
+provided:
+- The dependency is declared in `package.json` `dependencies`
+- Direction is spoke-to-hub (spokes import from the hub, not the reverse)
+- Hub-to-spoke needs should use extension points instead of direct imports
+
+If a similar hub-and-spoke pattern emerges in other feature areas, the same rules apply.
+
+### Rule 4: Cross-feature runtime imports are blocked
+
+If a feature package needs runtime code from an unrelated feature package, that code
+must be extracted to a core shared library or a domain-scoped `*-shared` package.
+
+### Rule 5: When a domain-scoped `*-shared` package is justified
+
+When a hub exports everything its spokes need, a separate shared package adds no value.
+Spokes should import directly from the hub.
+
+> *"If all these things always depend on model-serving being enabled, then I don't see
+> why model-serving can't be the one exporting that API."* — Christian Vogt
+
+A domain-scoped shared package (e.g., `model-serving-shared`) becomes justified when a
+package **outside** the hub-and-spoke group needs a runtime import of domain-specific
+code that currently lives in the hub. That cross-group dependency is the signal — not
+the number of shared items or the desire for tidiness.
+
+---
+
+## 4. Enforcement
+
+The existing `import/no-extraneous-dependencies` ESLint rule catches undeclared
+dependencies. Automated enforcement of layer and import direction rules is a separate
+effort. PR reviewers enforce these rules using this document as reference.
+
+---
+
+## 5. Core Library Stack
+
+| Package | Litmus test | Scope |
+|---------|-------------|-------|
+| **foundation** | "Is it a pure type or generic utility with no framework dependency?" | Pure TypeScript types and stateless utilities (e.g., `genRandomChars`). Zero `@odh-dashboard/*` runtime deps. Not yet extracted as a standalone package. |
+| **k8s-core** | "Does it describe a K8s resource type or domain-specific utility?" | K8s resource types and stateless utilities that operate on them. Contains both vanilla K8s and OpenShift types (see [Section 6](#6-openshift-specific-types)). |
+| **plugin-core** | "Does it need to know how the plugin system works?" | Extension points, plugin store, discovery hooks (`useExtensions`, `useResolvedExtensions`), code ref resolution (`LazyCodeRefComponent`), feature areas (`SupportedArea`, `useIsAreaAvailable`). |
+| **ui-core** | "Does it just render data using shared UI patterns?" | Shared React components (tables, resource display, form helpers), shared utilities (formatting, validation), and extension renderers (`ExtensibleDetailTabs`, `ExtensibleActions`). |
+| **app-config** | "Does it run only at build time?" | Build-time tooling: webpack configuration, Module Federation setup, dev server infrastructure. If it runs in the browser at runtime, it does not belong here. |
+
+### Where does this code go?
+
+Use the decision flow below when extracting code from `@odh-dashboard/internal` or
+deciding where new shared code belongs:
+
+```text
+START: Does this code require OpenShift APIs at runtime?
+│
+├─ YES → stays in RHOAI host (frontend/) or behind a platform-detection guard
+│
+└─ NO (platform-neutral at runtime):
+   │
+   Is it a pure type, enum, or stateless function with no framework dependency?
+   │
+   ├─ YES: Does it describe a K8s resource or API concept?
+   │  ├─ YES ──────────────────────────────────────────────→ k8s-core
+   │  └─ NO: Used across multiple packages?
+   │     ├─ YES ───────────────────────────────────────────→ foundation
+   │     └─ NO ────────────────────────────────────────────→ stays in consuming package
+   │
+   └─ NO (has runtime behavior or framework dependencies):
+      │
+      Does it operate on K8s types and remain stateless?
+      ├─ YES ──────────────────────────────────────────────→ k8s-core
+      │
+      └─ NO: Is it plugin infrastructure (extension discovery, loading, filtering)?
+         ├─ YES ───────────────────────────────────────────→ plugin-core
+         │
+         └─ NO: Is it a React component, hook, or UI utility?
+            ├─ YES: Is it domain-specific (tied to one feature area)?
+            │  ├─ YES ─────────────────────────────────────→ stays in the feature package
+            │  └─ NO (domain-agnostic, used broadly):
+            │     Does it contain platform-specific strings?
+            │     ├─ YES → parameterize first, then ───────→ ui-core
+            │     └─ NO ───────────────────────────────────→ ui-core
+            │
+            └─ NO: Is it build/config tooling (webpack, MF, dev server)?
+               ├─ YES ────────────────────────────────────→ app-config
+               └─ NO ─────────────────────────────────────→ stays in RHOAI host (frontend/)
+```
+
+### plugin-core vs ui-core
+
+These remain separate. Merging would force every distribution (which only needs the
+plugin framework) to take a dependency on shared React UI components it never uses.
+
+- 10 packages use plugin-core only (including all 3 distributions)
+- 13 packages use both plugin-core and ui-core
+- 0 packages use ui-core only
+
+The litmus test: *"Does it need to know how the plugin system works?"* If yes,
+plugin-core. If it just renders data, ui-core.
+
+`LazyCodeRefComponent` belongs in plugin-core — it bridges the plugin store to a
+rendered component and is consumed directly by `distributions/base/`. Extension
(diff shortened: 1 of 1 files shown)


docs/package-topology.md
```diff
@@ -0,0 +1,239 @@
...
+  │nim-srv  │        │                  │
+  │llmd-srv │        │                  │
+  └────┬─────┘        │                  │
+       └──────────────┼──────────────────┘
+                      │
+                      ▼
+              ┌──────────────┐
+              │   App Shell  │
+              │  dist/base/  │
+              └──────┬───────┘
```
christianvogt: Should App shell sit in it's own layer below features with features going straight down the layer below?
This diagram contracts layer boundary rule 1: `Feature packages may depend on core shared libraries. They must not depend on the app shell or on distributions.`

docs/package-topology.md
```diff
@@ -0,0 +1,239 @@
...
+
+### Rule 3: Hub-and-spoke runtime imports
+
+Feature packages in a hub-and-spoke group may have runtime imports from each other,
+provided:
+- The dependency is declared in `package.json` `dependencies`
+- Direction is spoke-to-hub (spokes import from the hub, not the reverse)
+- Hub-to-spoke needs should use extension points instead of direct imports
+
+If a similar hub-and-spoke pattern emerges in other feature areas, the same rules apply.
```
christianvogt: Modules that depend on others should also ensure add their dependencies to the module federation config.
This is something we will want to automate through webpack build plugins or something.

docs/package-topology.md
```diff
@@ -0,0 +1,239 @@
...
+
+### Rule 4: Cross-feature runtime imports are blocked
+
+If a feature package needs runtime code from an unrelated feature package, that code
+must be extracted to a core shared library or a domain-scoped `*-shared` package.
```
christianvogt: This is poorly worded. If you require code from another package then it is related.

It should be that if your module depends on common code but does not require the dependent module to be enabled for itself to be enabled, then it should be shared through common code package rather than a feature module package.

docs/package-topology.md
```diff
@@ -0,0 +1,239 @@
...
+
+### Rule 5: When a domain-scoped `*-shared` package is justified
+
+When a hub exports everything its spokes need, a separate shared package adds no value.
+Spokes should import directly from the hub.
+
+> *"If all these things always depend on model-serving being enabled, then I don't see
+> why model-serving can't be the one exporting that API."* — Christian Vogt
```
christianvogt: Let's not just call me out here :) but rather codify it the statement through rules.

docs/package-topology.md
```diff
@@ -0,0 +1,239 @@
...
+
+A domain-scoped shared package (e.g., `model-serving-shared`) becomes justified when a
+package **outside** the hub-and-spoke group needs a runtime import of domain-specific
+code that currently lives in the hub. That cross-group dependency is the signal — not
+the number of shared items or the desire for tidiness.
```
christianvogt: This is still confusing as to what it's trying to convey and I understand the problem.

docs/package-topology.md
```diff
@@ -0,0 +1,239 @@
...
+
+### Where does this code go?
+
+Use the decision flow below when extracting code from `@odh-dashboard/internal` or
+deciding where new shared code belongs:
+
+```text
+START: Does this code require OpenShift APIs at runtime?
+│
+├─ YES → stays in RHOAI host (frontend/) or behind a platform-detection guard
```
christianvogt: What is a platform-detection guard?

Separate package is a viable solution and not always just in frontend/.

We're going to need something to help guard against this runtime dependency. It's very easy to break this.

We also talked about an openshift-k8s-core. Is that still a thing?

docs/package-topology.md
```diff
@@ -0,0 +1,239 @@
...
+            │     Does it contain platform-specific strings?
+            │     ├─ YES → parameterize first, then ───────→ ui-core
+            │     └─ NO ───────────────────────────────────→ ui-core
+            │
+            └─ NO: Is it build/config tooling (webpack, MF, dev server)?
+               ├─ YES ────────────────────────────────────→ app-config
+               └─ NO ─────────────────────────────────────→ stays in RHOAI host (frontend/)
+```
+
+### plugin-core vs ui-core
```
christianvogt, line 182: This needs some further thought on exactly where the boundary lies between the two.

ui-core can certainly have it's own extensions and ui components that utilize them.

i think the things that go in plugin core are if they are fundamental to building the app. eg navigation, routes, etc...

docs/package-topology.md
```diff
@@ -0,0 +1,239 @@
...
+
+Core shared libraries must be **platform-neutral at runtime**. OpenShift-specific
+runtime behavior (API calls, resource creation, platform-aware branching) stays in the
+RHOAI host or behind platform-detection guards.
+
+OpenShift-specific **types** (e.g., `ProjectKind`, `TemplateKind`) are acceptable in
+`k8s-core` because they have no runtime footprint — RHAII can import them and never
+instantiate them. If `k8s-core` ever gains runtime functions that call OpenShift-specific
+APIs, those must be split out (e.g., into an `oc-core` package that layers on top of
+`k8s-core`).
```
christianvogt: I don't like this approach because it's bringing in soft dependencies on OC that RHAI shouldn't even have. 
If we have the need for OC types then they should go into a new package immediately rather than lend to a near future refactoring.

docs/package-topology.md
```diff
@@ -0,0 +1,239 @@
...
+
+Hardcoded platform strings like "find your resources in OpenShift" in shared components
+should be parameterized or removed from core packages.
+
+---
+
+## 7. Cypress Test Boundaries
```
christianvogt: This is fine for now i suppose but we need to start thinking of a e2e solution per distribution.

christianvogt (changes requested):

docs/package-topology.md
andyatmiami: hmm.. yes... this is a good callout..

and complicates the diagram enough maybe i drop the "ascii art" as a whole.. will think on this and update to address this in some fashion

docs/package-topology.md
andyatmiami: i did really like that quote - but noted!  will reword

docs/package-topology.md
andyatmiami: yeah.. fair feedback.. its messy

i want to say "`workbenches` shouldn't import from `model-serving`" - but abstracting that into a rule.. i'm struggling with - will iterate

docs/package-topology.md
andyatmiami: `openshift-core` _will eventually be a thing_ - just doesn't exist in codebase yet (and no real/hard compelling reason while we are focused on RHAII - def becomes important in RHOAI "convergence")

"plaftorm detection guard" would be an `if` block in code..

```
if cfg.PlatformType == 'Openshift' ...
```

docs/package-topology.md
andyatmiami: > If it just renders data, ui-core.

is this too simplified?

is your comment "we have to change this before merging" - or is it "we need to continue to refine over time"?  happy with either - just not sure if this was _immediately actionable_

docs/package-topology.md
andyatmiami: fair - i struggled with determining if i should write this document as it reflects the near-term world (with a specific RHAII focus) - or frame it as a long term outlook (where it would mention key pieces that may not exist yet.. as they aren't in RHAII focus... like this `openshift-core`)

docs/package-topology.md
christianvogt: Oh that's interesting. Does that Platform cfg exist today?

docs/package-topology.md
christianvogt: > If it just renders data, ui-core.

That's definitely a proper statement to align with I believe.

But also i think what i said above makes sense for when it goes to plugin-core. Specifically i'm talking about extension points because there's some extension points in plugin core i think can move to ui-core.

TBH probably almost all things going forward don't go into plugin-core. It may have everything, and more, that should exist here already.

docs/package-topology.md
christianvogt: do we not have something today that can go into a separate package? I fear if we don't start the flow now it'll immediately be needed as soon as we tell others to start adding to the k8s-core package and something doesn't fit.

docs/package-topology.md
andyatmiami: whoops, sorry, I was getting my "wires crossed"..

This concept (for now) _only exists_ in `core-bff` - but i think its _inevitable_ we will introduce this for frontend code:
- https://github.com/opendatahub-io/odh-dashboard/blob/main/distributions/core-bff/bff/internal/config/environment.go#L71-L99

docs/package-topology.md
```diff
@@ -0,0 +1,239 @@
...
+|-------|-----------------|----------|
+| **Core shared libraries** | Extension framework, shared UI, K8s types, pure utilities. | `k8s-core`, `plugin-core`, `ui-core`, `app-config`, `foundation` (not yet extracted) |
+| **Feature packages** | Domain-specific frontend functionality and backend infrastructure. Depend on core shared libraries and the app shell. | `model-serving`, `kserve`, `notebooks`, `gen-ai`, `model-registry`, `core-bff`, etc. |
+| **App shell** | Shared framework (masthead, sidebar, routing, error boundary) that distributions extend. | `distributions/base/` |
+| **Distributions** | Concrete, deployable dashboard variants. Composition roots that wire everything together. | `distributions/rhaii/`, `distributions/rhoai/` (future) |
+
+Dependencies flow downward. The diagram below shows the full picture — arrows point
+from consumer to dependency:
+
+```text
```
andrewballantyne: 🤔 

Image

Do I call it out or not... 

I will... If you're doing ascii designs 😛 fix boarders plz

docs/package-topology.md
```diff
@@ -0,0 +1,239 @@
...
+---
+
+## 2. The Serving Hub-and-Spoke
+
+Model serving and its related packages follow a hub-and-spoke pattern. Model-serving
+is the hub; `kserve`, `nim-serving`, `llmd-serving`, and `model-serving-backport` are
+spokes.
+
+Evidence:
+- Every spoke declares `reliantAreas` targeting model-serving areas
```
andrewballantyne: Spokes can have reliant backend features and thus other aspects 🤔 

Wouldn't the design be opposite of what you have 🤔 

Doesn't a hub have core reliances... each spoke has a hub reliance and also any number of core reliances

Ie. Model Serving (Hub) mounts Deployments
ie. LLMD (Spoke) mounts into Deployments & can also mount into Core Navigation

docs/package-topology.md
```diff
@@ -0,0 +1,239 @@
...
+- Every spoke registers zero standalone routes — all extensions target model-serving extension points
```
andrewballantyne: 🤔 Not sure this is true -- extensions are a growth model...

Not even sure this is valid for normal non-hub/spoke models... AI Hub has Catalog with eval data. EvalHub has extension into AI Hub for catalog support of rendering that eval data. Ownership of navigating eval data is EvalHub, so that makes sense.

docs/package-topology.md
```diff
@@ -0,0 +1,239 @@
...
+- The dependency is purely unidirectional: spokes depend on hub, hub has zero dependencies on spokes
```
andrewballantyne: Hub cannot rely on spokes -- it's plug and play by anyone... the extension point model of "here is how to connect to me" and spokes do the connectivity.

docs/package-topology.md
```diff
@@ -0,0 +1,239 @@
...
+
+This is the only group of feature packages with cross-feature runtime imports today.
+No other feature packages import from sibling features.
```
andrewballantyne: Spokes really shouldn't rely on each other -- they'd need some core middle-ground package technically...

Does this design support double-length spokes? A Hub of Hubs? Hub->HubSpoke->Spoke?

docs/package-topology.md
```diff
@@ -0,0 +1,239 @@
...
+
+`import type { ... }` across feature packages is always permitted. Type imports are
+erased at compile time and create no runtime dependency or bundle impact.
+
+### Rule 3: Hub-and-spoke runtime imports
+
+Feature packages in a hub-and-spoke group may have runtime imports from each other,
+provided:
+- The dependency is declared in `package.json` `dependencies`
+- Direction is spoke-to-hub (spokes import from the hub, not the reverse)
```
andrewballantyne: This is fair -- I don't see why a Hub would need to know of a Spoke -- it's a baseline, looks down - not up.

docs/package-topology.md
andrewballantyne: lol -- all the power and none of the responsibility 😛 

Kidding of course -- I agree, we can quote-credit Christian in Slack, but this is the design, we don't need to call out each other's words. We collectively agree to it. (or put all our names next to it to say we agree -- kidding again)

docs/package-topology.md
```diff
@@ -0,0 +1,239 @@
...
+code that currently lives in the hub. That cross-group dependency is the signal — not
+the number of shared items or the desire for tidiness.
+
+---
+
+## 4. Enforcement
+
+The existing `import/no-extraneous-dependencies` ESLint rule catches undeclared
+dependencies. Automated enforcement of layer and import direction rules is a separate
+effort. PR reviewers enforce these rules using this document as reference.
```
andrewballantyne, line 130: For linters -- separate topic -- we need to vet the heck out of ignores... we get this right or we fight it indefinitely.

docs/package-topology.md
```diff
@@ -0,0 +1,239 @@
...
+│
+└─ NO (platform-neutral at runtime):
+   │
+   Is it a pure type, enum, or stateless function with no framework dependency?
+   │
+   ├─ YES: Does it describe a K8s resource or API concept?
+   │  ├─ YES ──────────────────────────────────────────────→ k8s-core
+   │  └─ NO: Used across multiple packages?
+   │     ├─ YES ───────────────────────────────────────────→ foundation
+   │     └─ NO ────────────────────────────────────────────→ stays in consuming package
```
andrewballantyne: This will be interesting for sure -- 'used across' and 'semantically generic' could be a nuance here.

Also, 99.9% chance that if you put it in the consuming package, it will duplicate into many more before we get it into foundation package.

docs/package-topology.md
```diff
@@ -0,0 +1,239 @@
...
+| **plugin-core** | "Does it need to know how the plugin system works?" | Extension points, plugin store, discovery hooks (`useExtensions`, `useResolvedExtensions`), code ref resolution (`LazyCodeRefComponent`), feature areas (`SupportedArea`, `useIsAreaAvailable`). |
+| **ui-core** | "Does it just render data using shared UI patterns?" | Shared React components (tables, resource display, form helpers), shared utilities (formatting, validation), and extension renderers (`ExtensibleDetailTabs`, `ExtensibleActions`). |
+| **app-config** | "Does it run only at build time?" | Build-time tooling: webpack configuration, Module Federation setup, dev server infrastructure. If it runs in the browser at runtime, it does not belong here. |
+
+### Where does this code go?
+
+Use the decision flow below when extracting code from `@odh-dashboard/internal` or
+deciding where new shared code belongs:
+
+```text
```
andrewballantyne: Not to nitpick -- this is a flow chart that doesn't look like a flowchart 🙂

andrewballantyne: Is this in ASCII because Claude doesn't like reading images? Or because it was the author's choice 🙂

docs/package-topology.md
```diff
@@ -0,0 +1,239 @@
...
+```
+
+### plugin-core vs ui-core
+
+These remain separate. Merging would force every distribution (which only needs the
+plugin framework) to take a dependency on shared React UI components it never uses.
+
+- 10 packages use plugin-core only (including all 3 distributions)
+- 13 packages use both plugin-core and ui-core
+- 0 packages use ui-core only
```
andrewballantyne: What are these numbers -- real data?

docs/package-topology.md
andrewballantyne: What about cross package tests? 

e2e story: Serving Deploys. MaaS Registers. Gen AI Studio Uses.

We can break these up into distinct "yaml apply" moments -- but we lose a bit of e2e story for these kinda things... 

What about simply, Model Catalog => Serving functionality -- this is a direct integration, but two packages.

docs/package-topology.md
andyatmiami: that is fair - i'll look to have `pewter` (or myself!) carve something out in the actual source code - and update this doc to read assuming that **exists** 

✅

docs/package-topology.md
andyatmiami: oh, yeah, sorry, i think i mentioned in my PR plea-for-review in Slack that i knew (after the fact) that the borders were hosed - I promise on my next push they won't be 😇

docs/package-topology.md
andyatmiami: is this a "request for change" ? I don't take issue with the comment - but (after a long day today) - i'm not clear what the "call to action" is here

docs/package-topology.md
andyatmiami: > Spokes really shouldn't rely on each other

don't disagree - but that is **NOT** the reality of the world today - and with time _tickin' away_ - i'm not clear we'll be able to fix this problem in July.. so i'm hesitant to doc this when the codebase has clear violations...

i honestly haven't really thought about double-length spokes - and unless you tell me otherwise (which is fine feedback) - i'd like to avoid thinking about that for _this initial merge_ 😇

docs/package-topology.md
andyatmiami: yes, i don't expect for this to be _perfect on first go_ - but hopefully it'll be definitely better than status quo 🤞

part of avoiding the duplication will be a fast(ish) follow into looking into enforcement of these agreed-upon rules... which i'd like to think is 1-2 weeks away

docs/package-topology.md
andyatmiami: author's choice (or lack of caring) - happy to take feedback to "buck up" and make it a proper flowchart..

i had "ascii art" above in this file - and just kinda _leaned in_

docs/package-topology.md
andyatmiami: they are - but also probably unnecessary - as they will age poorly - i'll generalize this point

docs/package-topology.md
andyatmiami: will need to think about this more.. and can't promise to have a suitable update in **this PR**

as "cross-package e2e" - while absolutely related to these concepts - is also an _entirely different beast unto itself_

docs/package-topology.md
andyatmiami: thanks for callout - let me 2x check and validate this (will follow up)

docs/package-topology.md
lucferbux: we can always use mermaid diagrams or similar, claude understand them really well and you won't get this broken lines (just as suggestion, I'm fine keeping the ascii diagrams)

docs/package-topology.md
lucferbux: we can always inject that in the FE too

docs/package-topology.md
```diff
@@ -0,0 +1,229 @@
...
+    q5a -->|No| k8s2["k8s-core"]
+    q5 -->|No| q6{"Plugin infrastructure?\n(discovery, loading, filtering)"}
+    q6 -->|Yes| plugin["plugin-core"]
+    q6 -->|No| q7{"React component,\nhook, or UI utility?"}
+
+    q7 -->|Yes| q8{"Domain-specific?\n(tied to one feature)"}
+    q8 -->|Yes| feature["stays in feature package"]
+    q8 -->|No| q9{"Contains platform-\nspecific strings?"}
+    q9 -->|Yes| param["parameterize, then → ui-core"]
+    q9 -->|No| uicore["ui-core"]
```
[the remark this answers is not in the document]
andyatmiami: reworked to remove ambiguity/contradiction

paulovmr (approved):
/lgtm

andyatmiami:
/approve

i know various other _key dashboard stakeholders_ are on PTO and/or otherwise unavailable - but given this is "just" a markdown file - lets bring this into codebase so I can talk about it in office hours and incrementally improve if/as needed.

PR #8432, opened 2026-07-06
feat(RHOAIENG-68565): run BFF frontends and backends in cypress E2E tests

christianvogt:
https://issues.redhat.com/browse/RHOAIENG-68565

## Description

E2E CI was using `start:dev:ext` which proxied all traffic to the deployed cluster dashboard. Module federation remotes were loaded from the cluster — not from the PR source — making BFF frontend and backend changes completely untestable in E2E. BFF processes started in CI but received no traffic because nothing routed requests to them.

This PR adds a lightweight reverse proxy (`packages/cypress/src/e2e-proxy/`) that sits between Cypress and the local stack. The proxy routes cluster-bound paths to the remote cluster and everything else to the locally running backend/BFFs, allowing E2E tests to exercise PR code for both module frontends and their BFFs.

**E2E proxy (`packages/cypress/src/e2e-proxy/`)**
- Reverse proxy on `:4040` with `/e2e-login` endpoint for user switching via `oc login`
- Discovers cluster URL from `ODH_DASHBOARD_URL` env var or `test-variables.yml`
- Injects `Authorization: Bearer` to cluster-bound requests, `x-forwarded-access-token` to local backend
- Full WebSocket support for `/wss/k8s` paths
- Routes `/api/service/*` and non-`odh-dashboard` module federation proxy paths to cluster

**Cypress auth simplification**
- `visitWithLogin` POSTs to `/e2e-login` when `CYPRESS_E2E_PROXY=true`
- Removed legacy localhost token-refresh polling and `OC_TOKEN_REFRESH_WAIT_MS`

**CI workflow (`cypress-e2e-test.yml`)**
- Exports `ODH_DASHBOARD_URL` to env for proxy discovery (same sed override as before)
- Uses `CYPRESS_*` env vars for Cypress config — avoids shell escaping through `concurrently`
- Unified `cypress:server:e2e` / `cypress:server:e2e:wait` scripts replace old `start:e2e`
- Single `npm run test:cypress:e2e` command starts servers, waits for health, runs tests
- Removed webpack dev server lifecycle, port collision handling, and PID cleanup

**Backend**
- `directCallUtils`: prefer `x-forwarded-access-token` for K8s API calls when present

**Runtime constants**
- `POLL_INTERVAL` / `FAST_POLL_INTERVAL` sourced from `window.*` at runtime, allowing mock tests to override without a separate build

## How Has This Been Tested?

1. Install dependencies and build modules:
```bash
npm install
npm run install:modules
npm run prepare:e2e
```

2. Start the e2e-proxy (in a separate terminal):
```bash
npm run cypress:server:e2e
```

3. Open Cypress with the proxy enabled:
```bash
cd packages/cypress
CY_TEST_CONFIG=./test-variables.yml \
CYPRESS_E2E_PROXY=true \
CY_TEST_TAGS=@ci-dashboard-regression-tags \
npm run open:e2e
```

Successful test run in CI from build cache:
https://github.com/opendatahub-io/odh-dashboard/actions/runs/29135483366
image

Another run where build cache missed and rebuild was ncessary
https://github.com/opendatahub-io/odh-dashboard/actions/runs/29125238409

image

## Test Impact

No new automated tests. The e2e-proxy is exercised by every E2E test run via `visitWithLogin`. The CI workflow changes are validated by the E2E test suite itself.

## Request review criteria:

## Summary
* **New Features**
  * Added build-then-serve Cypress E2E with a local reverse proxy and an `/e2e-login` flow.
  * Introduced CI workflow controls for Cypress tag selection and artifact reuse/build gating.
* **Bug Fixes**
  * Improved access-token propagation for direct calls and proxy-routed cluster requests.
  * Updated Cypress readiness checks to use consistent `/healthcheck` endpoints.
* **Documentation**
  * Updated Cypress E2E testing docs and tutorials to match the new proxy-based workflow.


package.json
```diff
@@ -62,13 +62,15 @@
     "test-unit": "turbo run test-unit",
     "test-unit-coverage": "turbo run test-unit-coverage && node scripts/unit-test-coverage-aggregator.js",
     "test:contract": "turbo run test:contract --concurrency=1",
-    "cypress:server": "turbo run cypress:server",
+    "cypress:server": "turbo run cypress:server --concurrency=20",
```
antowaddle, line 68: Where did the concurrency values here come from?

packages/cypress/package.json
```diff
@@ -17,7 +17,9 @@
     "run:mock:coverage": "rimraf coverage .nyc_output && cross-env CY_COVERAGE=true npm run run:mock",
     "run:e2e": "cypress run --env grepTags=\"${CY_TEST_TAGS}\",skipTags=\"${CY_SKIP_TAGS}\" -b chrome",
     "run:e2e:dashboard": "cypress run --env grepTags=@Dashboard+-@Bug -b chrome",
-    "run:electron": "cypress run -b electron"
+    "run:electron": "cypress run -b electron",
```
antowaddle, line 20: This can probably be removed.

packages/cypress/src/e2e-proxy/server.ts
```diff
@@ -0,0 +1,209 @@
...
+
+function readBody(req: http.IncomingMessage): Promise<string> {
+  return new Promise((resolve) => {
+    const chunks: Buffer[] = [];
+    req.on('data', (chunk) => chunks.push(chunk));
+    req.on('end', () => resolve(Buffer.concat(chunks).toString('utf8')));
+  });
+}
+
+function handleE2eLogin(body: string, res: http.ServerResponse): void {
```
antowaddle, line 74: Could we add error handling here?  The proxy catches JSON parse errors but doesn't validate that credentials were actually successful before storing the token. If
  oc login succeeds but oc whoami --show-token fails, the proxy returns 500 but doesn't clean up storedUsername.

frontend/config/webpack.dev.js
```diff
@@ -88,41 +88,6 @@ module.exports = smp.wrap(
               throw new Error('Login with `oc login` prior to starting dev server.');
             }
 
-            // Token refresh mechanism (for when oc user is switched during tests)
-            // Gated behind ODH_TOKEN_REFRESH to avoid unexpected user switches in local dev
```
antowaddle, line 92: Have you validated that removing this works fine locally? I added this token refreshing due to an issue detected in CI where tests were flaking - https://github.com/opendatahub-io/odh-dashboard/pull/6934

.github/workflows/cypress-e2e-test.yml
```diff
@@ -1027,9 +902,6 @@ jobs:
           find ${{ github.workspace }}/packages/cypress/screenshots -type f -mtime +2 -delete 2>/dev/null || true
           find ${{ github.workspace }}/packages/cypress/videos -type f -mtime +2 -delete 2>/dev/null || true
           
-          # Clean old webpack logs (>2 days old)
```
antowaddle, line 1030: Why are we removing this?

.github/workflows/cypress-e2e-test.yml
```diff
@@ -1190,264 +1062,44 @@ jobs:
...
-      - name: Install make for BFF builds
-        if: needs.get-test-tags.outputs.bff_packages != '[]'
-        run: |
-          if ! command -v make &> /dev/null; then
-            echo "📦 Installing make..."
-            sudo dnf install -y make 2>/dev/null || sudo yum install -y make 2>/dev/null || sudo apt-get install -y make 2>/dev/null
-          else
-            echo "✅ make already available: $(make --version | head -1)"
-          fi
+      - name: Build on cache miss
```
antowaddle: I've had some issues with this approach when working on speeding up the mocks, when the cache misses (which happens quite often) the rebuild is really heavy and started timing out the runners. I created a dedicated job and done the building there once which generally worked better.

.github/workflows/cypress-e2e-test.yml
```diff
@@ -1190,264 +1062,44 @@ jobs:
...
+        if: steps.cypress-build-cache.outputs.cache-hit != 'true'
+        run: npm run cypress:server:build:coverage -- --concurrency=4
```
antowaddle: Are we starting all 9+ BFFs even if we're testing one specific module? It's >12GB to do this each time and significant time wise.

antowaddle (changes requested):
Thanks @christianvogt for this contribution, this is the key missing piece to expedite our shift-left integration testing.

A few issues additional issues:
1. CI is currently failing - can you check this please?
2. I tried to run this locally, but I see the following error `{"statusCode":500,"code":"ENOENT","error":"Internal Server Error","message":"ENOENT: no such file or directory, open '...forked-odh-dashboard/odh-dashboard/frontend/public-cypress/index.html'"}`
3. Have you been able to validate this successfully for say GenAI or Model Registry? I'd like to see a test running where we change something in both the F/E and BFF in a PR then specifically see a test fail requiring an update (basically introduce a breaking change purposely).

.github/workflows/cypress-e2e-test.yml
christianvogt: This isn't starting. It's building. It's the same build we do in our other mock build tests.

.github/workflows/cypress-e2e-test.yml
christianvogt: We will work together with your change on builds to be more optimal.
I moved the build before the jobs in case of cache miss for now.

.github/workflows/cypress-e2e-test.yml
christianvogt: We don't use webpack anymore for running. We build then run the backend to server the app source files.

frontend/config/webpack.dev.js
christianvogt: Yes it works locally.
I will run a CI test and post results.

lucferbux (commented):
Great work on this, Christian. The build-then-serve model with the proxy is a massive improvement over the webpack+dynamic-BFF approach — and actually testing BFF code from PRs is a big deal. A few nits inline, nothing blocking.

packages/cypress/src/e2e-proxy/server.ts
```diff
@@ -0,0 +1,197 @@
+import http from 'http';
+import type { Socket } from 'net';
+import { execFileSync } from 'child_process';
+import httpProxy from 'http-proxy';
+import type { ProxyRoute, RoutingTable } from './routes';
+import { getOcpApiUrl } from './routes';
+
+const TMP_KUBECONFIG = '/tmp/cypress-e2e.kubeconfig';
```
lucferbux, line 11: nit: On self-hosted runners with parallel jobs, multiple proxy instances would collide on this path. Consider using `os.tmpdir()` + a unique suffix (e.g., `process.pid` or the GitHub run ID via `process.env.GITHUB_RUN_ID`) to avoid races.

packages/cypress/src/e2e-proxy/server.ts
```diff
@@ -0,0 +1,197 @@
...
+    if (url === '/e2e-login' && req.method === 'POST') {
+      const body = await readBody(req);
+      handleE2eLogin(body, res);
+      return;
+    }
+
+    const clusterRoute = matchClusterRoute(url, routingTable.clusterRoutes);
+    const target = clusterRoute?.target ?? routingTable.defaultTarget;
+
+    console.log(
```
lucferbux: nit: This logs every single request to stdout — static files, healthchecks, everything. During a full E2E run that's thousands of lines. Could scope it to cluster-bound requests only (like the `proxyReq` handler already does for `/api/service`), or gate it behind a `LOG_LEVEL` env var.

backend/src/utils/directCallUtils.ts
```diff
@@ -21,31 +21,28 @@ export const getDirectCallOptions = async (
 
   // Adjust the header auth token
   let headers;
-  if (DEV_MODE) {
-    // In dev mode, we always are logged in fully -- no service accounts
+  const raw = request.headers[USER_ACCESS_TOKEN];
```
lucferbux, line 24: nit: Subtle behavior change worth noting — previously in dev mode, kubeconfig identity was always used. Now if anything sets `x-forwarded-access-token` (e.g., the e2e-proxy), that takes priority even in dev mode. This is the correct behavior for the proxy use case, but it changes dev-mode semantics for anyone who runs `start:dev:ext` without the proxy.

.github/workflows/cypress-e2e-test.yml
```diff
@@ -1190,264 +1102,39 @@ jobs:
...
-              if [ $i -eq 30 ]; then
-                echo "  ❌ BFF $BFF_NAME failed health check after 60 seconds"
-                echo "  📋 BFF logs:"
-                tail -50 "/tmp/bff_${BFF_DIR}.log" || true
-                exit 1
-              fi
-              sleep 2
-            done
-          done
+          go-version-file: packages/gen-ai/bff/go.mod
```
lucferbux: nit: Pinning to gen-ai's `go.mod` works if all BFFs share the same Go version. If a package bumps Go independently, this could break. A safer approach might be to pick the max `go` directive across all `packages/*/bff/go.mod` files, or at least document that gen-ai is the canonical source.

packages/agent-ops/Makefile
```diff
@@ -91,7 +91,7 @@ dev-bff-e2e-mock: ## Run BFF for e2e tests (mock mode, no cluster required)
 
 .PHONY: dev-bff-e2e-cluster
 dev-bff-e2e-cluster: ## Run BFF for e2e tests (federated mode, connected to cluster)
-	cd bff && make run PORT=$(E2E_BFF_PORT) LOG_LEVEL=info MOCK_K8S_CLIENT=false MOCK_HTTP_CLIENT=false DEV_MODE=true DEPLOYMENT_MODE=federated AUTH_METHOD=user_token AUTH_TOKEN_HEADER=x-forwarded-access-token AUTH_TOKEN_PREFIX="" INSECURE_SKIP_VERIFY=true
+	cd bff && make run PORT=$(E2E_BFF_PORT) LOG_LEVEL=info MOCK_K8S_CLIENT=false MOCK_HTTP_CLIENT=false DEV_MODE=true DEPLOYMENT_MODE=federated AUTH_METHOD=user_token AUTH_TOKEN_HEADER=Authorization AUTH_TOKEN_PREFIX="Bearer " INSECURE_SKIP_VERIFY=true
```
lucferbux, line 94: nit: All BFF Makefiles changed from `AUTH_TOKEN_HEADER=x-forwarded-access-token` to `AUTH_TOKEN_HEADER=Authorization AUTH_TOKEN_PREFIX="Bearer "`. The e2e-proxy sends `x-forwarded-access-token` to the local backend (non-cluster route). When the backend proxies `/_mf/{name}/*` to a BFF, does it set `Authorization: Bearer` on the outgoing request? If the backend proxy just passes through headers, the BFF won't find the token it expects. Worth verifying the header flow end-to-end.

packages/gpuaas/package.json
```diff
@@ -7,7 +7,6 @@
   "scripts": {
     "lint": "eslint .",
     "lint:fix": "eslint . --fix",
```
lucferbux, line 9: nit: The `test-unit` script removal looks unrelated to the E2E proxy changes — intentional cleanup or accidental?

christianvogt: Intentional because unit tests locally fail without tests and no param to suppress.

packages/cypress/src/e2e-proxy/server.ts
christianvogt: We already agreed that we aren't running multiple jobs in parallel on the same runner. Otherwise we also have port collisions to deal with.

packages/cypress/src/e2e-proxy/server.ts
christianvogt: It was mainly for debugging. I will parameterize it.

backend/src/utils/directCallUtils.ts
christianvogt: Doesn't affect `start:dev:ext` because the backend is never used.
If just running the backend, then there's nothing supplying this header unless it's run behind a reverse proxy or authentication app like we're doing in e2e. If running backend with the frontend we still do not supply this header.

packages/agent-ops/Makefile
christianvogt: This is now inline with production where BFFs expect the Authorization header. The backend always sets this header and not x-forwarded-access-token.

.github/workflows/cypress-e2e-test.yml
christianvogt: Updated

christianvogt:
@antowaddle 

Ran the cypress-e2e tests on a separate branch.
Successful run of just default regression tag:
https://github.com/opendatahub-io/odh-dashboard/actions/runs/29061165789/job/86265334056

Near successful run of all auto determined tags. Seems to be one feature store test that failed (`testFeatureStoreFeatures`). Can't see how that failure is related to these changes if all else have passed.
https://github.com/opendatahub-io/odh-dashboard/actions/runs/29062348295

antowaddle (approved):
/lgtm
Thanks @christianvogt for this contribution, it's a major missing piece in our integration testing journey. I validated your recent changes by running the related MR tests https://github.com/opendatahub-io/odh-dashboard/actions/runs/29318160733/job/87038189952 I can see that all the modules built correctly and the tests ran/passed.

PR #8238, opened 2026-06-22
feat: integrate BFF servers with cluster TLS security profile

ugiordan:
## Summary

- Integrate all 9 BFF servers with the cluster-wide TLS security profile from `apiservers.config.openshift.io/cluster`
- Shared `pkg/tls/` package reads the profile via raw REST client (no openshift/api dependency to avoid k8s version conflicts)
- Replace hardcoded `tls.Config{MinVersion: tls.VersionTLS13}` with dynamic profile-driven config
- Fall back to Intermediate defaults (TLS 1.2, ECDHE ciphers, ALPN h2/http1.1) on non-OpenShift clusters
- Fail closed on unexpected errors, graceful fallback on IsNotFound/IsNoMatchError/IsForbidden
- Add RBAC for `config.openshift.io/apiservers` to all 9 ClusterRole manifests
- 20 unit tests covering profile resolution, cipher mapping, error handling

## BFFs updated

1. `distributions/core-bff/bff/`
2. `packages/agent-ops/bff/`
3. `packages/automl/bff/`
4. `packages/autorag/bff/`
5. `packages/eval-hub/bff/`
6. `packages/gen-ai/bff/`
7. `packages/maas/bff/`
8. `packages/mlflow/bff/`
9. `packages/model-registry/upstream/bff/`

## Design decisions

- **Raw REST instead of typed objects**: `openshift/api` requires k8s v0.35+ which conflicts with BFF modules on k8s v0.34.x. Raw REST + JSON parsing avoids the dep conflict.
- **Well-known profiles replicated locally**: Intermediate/Modern/Old profiles replicated from openshift/api v0.0.0-20260317165824 to avoid the import.
- **Fail closed**: Custom profile with all unsupported ciphers returns error (operator exits). IsForbidden falls back gracefully (restricted RBAC scenarios).
- **No enableHTTP2 flag**: CVE-2023-44487 fixed in Go 1.21.3+. NextProtos always set to `["h2", "http/1.1"]`.

## Motivation

OCP 5.0 (GA October 2026) requires all components to honor the centralized TLS profile ([OCPSTRAT-2611](https://redhat.atlassian.net/browse/OCPSTRAT-2611)).

## Test plan

- [x] `go test ./...` passes in pkg/tls/ (20 tests)
- [x] `go build ./...` passes for all 9 BFFs
- [ ] Deploy on OpenShift cluster with Intermediate profile
- [ ] Deploy on vanilla K8s, verify fallback

[...]

PR #8437, opened 2026-07-07
feat: integrate BFF servers with cluster TLS security profile

ugiordan:
https://issues.redhat.com/browse/RHOAIENG-61069

> Supersedes #8238, which was closed due to a shallow-clone orphan overwriting the remote branch.

## Description

Integrate all 9 BFF servers with the cluster-wide TLS security profile from `apiservers.config.openshift.io/cluster`, replacing hardcoded TLS configs with dynamic, profile-driven configuration.

**What changed:**

- New shared `pkg/tls/` package reads the OpenShift APIServer TLS profile via raw REST client (no `openshift/api` dependency to avoid k8s version conflicts with BFF modules)
- Replace hardcoded `tls.Config{MinVersion: tls.VersionTLS13}` with dynamic config derived from the cluster profile
- Fall back to Mozilla Intermediate defaults (TLS 1.2, ECDHE ciphers, ALPN h2/http1.1) on non-OpenShift clusters or when the APIServer resource is not found
- Fail closed on unexpected errors (RBAC denied, client failures)
- Add RBAC for `config.openshift.io/apiservers` (`get` verb, `resourceNames: ["cluster"]`) to all 9 ClusterRole manifests
- 20 unit tests covering profile resolution, cipher mapping, deduplication, and error handling

**BFFs updated:**

1. `distributions/core-bff/bff/`
2. `packages/agent-ops/bff/`
3. `packages/automl/bff/`
4. `packages/autorag/bff/`
5. `packages/eval-hub/bff/`
6. `packages/gen-ai/bff/`
7. `packages/maas/bff/`
8. `packages/mlflow/bff/`
9. `packages/model-registry/upstream/bff/`

**Design decisions:**

- **Raw REST instead of typed objects**: `openshift/api` requires k8s v0.35+ which conflicts with BFF modules on k8s v0.34.x. Raw REST + JSON parsing avoids the dep conflict entirely.
- **Well-known profiles replicated locally**: Intermediate/Modern/Old profiles replicated from `openshift/api v0.0.0-20260317165824` to avoid importing the package.
- **Fail closed**: Custom profile with all unsupported ciphers returns an error (process exits). RBAC denied also returns an error. Only non-OpenShift clusters and missing APIServer resource get graceful fallback.
- **No enableHTTP2 flag**: CVE-2023-44487 fixed in Go 1.21.3+. `NextProtos` always set to `["h2", "http/1.1"]`.

**Motivation:** OCP 5.0 (GA October 2026) requires all components to honor the centralized TLS profile ([OCPSTRAT-2611](https://redhat.atlassian.net/browse/OCPSTRAT-2611)).

## How Has This Been Tested?

- `go test ./...` passes in `pkg/tls/` (20 tests covering all profile types, cipher mapping, error paths)
- `go build ./...` passes for all 9 BFF modules
- Cluster deployment testing pending (requires OCP cluster with TLS profile configured)

## Test Impact

- 20 new unit tests in `pkg/tls/tls_test.go` covering: Intermediate/Modern/Old/Custom profile resolution, cipher name mapping (OpenSSL + IANA), deduplication, DHE cipher dropping, unsupported cipher error, nil/empty profile fallback, ALPN configuration
- No Cypress tests needed: this is backend-only Go code with no UI changes

[OCPSTRAT-2611]: https://redhat.atlassian.net/browse/OCPSTRAT-2611?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * HTTPS services now inherit cluster-wide TLS settings, including supported protocols, ciphers, and ALPN.
  * Secure metrics and webhooks use the same TLS profile.
  * TLS profile changes are detected and applied through operator refresh behavior.

* **Bug Fixes**
  * Improved fallback and error handling for unavailable or partially unsupported TLS profiles.
  * TLS configuration errors are reported before service startup.

* **Chores**
  * Updated service packaging and permissions to support shared TLS configuration.


gmenher (commented):
Solid work, the shared `pkg/tls` design is clean, the fail-closed semantics are correct, and the RBAC is nicely scoped to the minimum. I left a couple of inline observations.

pkg/tls/tls.go
```diff
@@ -0,0 +1,350 @@
...
+			"ECDHE-ECDSA-AES256-GCM-SHA384",
+			"ECDHE-RSA-AES256-GCM-SHA384",
+			"ECDHE-ECDSA-CHACHA20-POLY1305",
+			"ECDHE-RSA-CHACHA20-POLY1305",
+			"ECDHE-ECDSA-AES128-SHA256",
+			"ECDHE-RSA-AES128-SHA256",
+			"ECDHE-ECDSA-AES128-SHA",
+			"ECDHE-RSA-AES128-SHA",
+			"ECDHE-ECDSA-AES256-SHA",
+			"ECDHE-RSA-AES256-SHA",
```
gmenher, line 153: The Old profile lists `ECDHE-ECDSA-AES256-SHA` and `ECDHE-RSA-AES256-SHA` here, but I don't see matching entries in `openSSLToGo` above. Go does have `TLS_ECDHE_ECDSA_WITH_AES_256_CBC_SHA` and `TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA`. Are these just missing from the map? If so, they'd be silently dropped for customers on the Old profile.

pkg/tls/tls.go
```diff
@@ -0,0 +1,350 @@
...
+		NextProtos: nextProtos,
+	}
+
+	// TLS 1.3 cipher suites are not configurable in Go (all TLS 1.3 ciphers are always enabled).
+	// Only set CipherSuites for TLS 1.2 and below.
+	if minVersion < tls.VersionTLS13 {
+		ciphers, dropped := resolveCiphers(spec.Ciphers)
+		if len(dropped) > 0 {
+			logger.Info("some ciphers from cluster TLS profile are not supported by Go, dropped",
+				"dropped", dropped)
```
gmenher, line 270: The Intermediate profile includes 3 TLS 1.3 cipher names (`TLS_AES_128_GCM_SHA256`, etc.) that aren't in `openSSLToGo`. Since `minVersion` is TLS 1.2, `resolveCiphers` logs them as "not supported by Go" on every startup across all 9 BFFs. Since Go handles TLS 1.3 ciphers automatically, would it make sense to filter those names before resolution or adjust the log message?

pkg/tls/tls.go
```diff
@@ -0,0 +1,350 @@
...
+		}
+		if len(spec.Ciphers) > 0 && len(ciphers) == 0 {
```
gmenher, line 272: If a Custom profile specifies `minTLSVersion < TLS1.3` but leaves `ciphers` empty, `CipherSuites` ends up nil and Go falls back to its internal defaults. Is that intentional? I wonder if an empty cipher list on a Custom profile should be treated as a misconfiguration and either error out or fall back to Intermediate explicitly.

lucferbux:
@ugiordan We alaredy landed the standalone deployment in the monorepo + the operator work, if you want we can resume this task.

ugiordan:
@lucferbux should I close this one?

lucferbux:
Yes, let me do it for you! thanks for the ping

PR #8441, opened 2026-07-07
feat: add model serving routes to core BFF

caponetto:
https://issues.redhat.com/browse/RHOAIENG-67106

## Description

Model Serving is the first RHAII use case requiring server-side orchestration in the Core BFF. This PR adds six dedicated route groups to the Go BFF under `distributions/core-bff/`, covering admin-only access control, cross-resource lookups, upstream service proxying, and Prometheus query forwarding -- capabilities that cannot be handled client-side.

### Route groups added

| Route | Auth | Purpose |
|---|---|---|
| `POST /api/servingRuntimes` | `secureAdminRoute` | Create ServingRuntime CRs via SA dynamic client with dryRun support |
| `GET /api/nim-serving/:nimResource` | `publicRoute` | Cross-resource lookup: NIMAccount CR -> resolve resource name -> fetch Secret/ConfigMap |
| `ALL /api/service/model-serving/*` | Authenticated (no wrapper) | Reverse proxy to model-serving-api with SSRF protection, auth forwarding, path rewriting |
| `POST /api/prometheus/{query,queryRange,pvc,bias,serving}` | Authenticated (no wrapper) | POST body -> GET to Thanos with query type resolution and user token forwarding for K8s RBAC |
| `GET /api/namespaces/:name/:context` | `secureRoute` + SSAR | Namespace mutation via MergePatch, gated on `create serving.kserve.io/servingruntimes` |
| `GET/POST/DELETE /api/integrations/nim` | GET: `publicRoute`, POST/DELETE: `secureAdminRoute` | NIM integration lifecycle (status, enable, disable) |

**Note**: NIM and Prometheus endpoints are registered only on OpenShift platforms and return 404 on all other platforms.

### Intentional HTTP status differences from Fastify

These are deliberate improvements over the Fastify implementation:

- **Admin-only routes**: non-admin returns **403** (Fastify returns 401) -- correct HTTP semantics (authenticated but not authorized)
- **Prometheus**: returns **503** when no host configured (Fastify returns 400) -- reflects service unavailability, not a bad request
- **NIM serving**: non-404 fetch errors return **500** (Fastify maps all errors to 404) -- preserves actual error type for debugging

### Key implementation details

- **NIM secret updates** clear existing `Data` map before setting `StringData`, preventing stale keys from surviving updates (matches Fastify's `replaceNamespacedSecret` semantics)
- **Namespace mutation** rejects context 0 (DSG_CREATION) with 400 -- not applicable for RHAII
- **Namespace allowlist** enforced on ServingRuntime creation, matching Fastify's `secureAdminRoute` namespace validation
- **Prometheus query type resolution**: `queryRange`/`bias`/`serving` -> `query_range`; `query`/`pvc` -> instant `query`
- **Model serving proxy** forwards all HTTP methods (GET/POST/PUT/PATCH/DELETE) with Bearer token from request identity
- Route registration supports both `/api/` and `/core-bff/api/` path prefixes

### Route registration reorganization

The previously monolithic `routes.go` has been split into focused files by domain: `routes_model_serving.go`, `routes_config.go`, `routes_connection_types.go`, `routes_openapi.go`, and `routes_starter.go`. Each file owns the registration for its route group, keeping `routes.go` as the top-level combiner that wires the service mux, static handler, and public/authenticated split. The same split was applied to the k8mocks test helpers (`mock_crds.go`, `mock_nim.go`, `mock_rbac.go`, `mock_resources.go`) to keep test setup manageable as the handler count grows.

## How Has This Been Tested?

- Automated [smoke-test.sh](https://gist.github.com/caponetto/97d1bc3b525e342a1cfba92fa7221b06) that creates a kind cluster with admin/viewer service accounts and exercises  endpoint tests covering public, secureRoute, secureAdminRoute, no-auth rejection, namespace mutation, and Prometheus proxy scenarios.
- Contract tests validate all 6 endpoint groups against the OpenAPI spec (auth, schema, error paths)
- Unit tests for every handler and repository (serving runtime, NIM, namespace mutation, Prometheus, model serving proxy)
- Handler tests use stub clients to verify auth enforcement (401 unauthenticated, 403 non-admin), validation errors (400), and success paths (200)
- Repository tests cover Prometheus query construction, NIM secret create-on-conflict-update flow, and namespace mutation patch building

## Test Impact

- Contract tests added for all 6 model serving endpoint groups, validating auth enforcement, response schemas against the OpenAPI spec, and error paths
- Unit tests added for every new handler and repository, covering auth (401/403), input validation (400), success paths (200), and edge cases like NIM secret conflict resolution and Prometheus query type mapping

## Request review criteria:

If you have UI changes: 

## Summary by CodeRabbit

* **New Features**
  * Added model-serving proxy endpoints with secure request forwarding.
  * Added ServingRuntime creation, namespace serving-platform mutations (including dry-run), and NIM integration/serving-resource endpoints.
  * Added Prometheus/Thanos query endpoints (instant and range) and updated OpenAPI coverage.
* **Bug Fixes**
  * Improved authorization responses: unauthenticated requests return **401**, authenticated non-admin requests return **403**.
  * Standardized Kubernetes error-to-HTTP handling for clearer client-facing responses.
* **Documentation**
  * Updated Core BFF README with new Prometheus/model-serving override environment variables and expanded local call examples.

### DIFF
--- a/distributions/core-bff/bff/internal/api/app.go
+++ b/distributions/core-bff/bff/internal/api/app.go
@@ -93,7 +93,6 @@ func NewApp(cfg config.EnvConfig, logger *slog.Logger) (*App, error) {
 			SADynClient: k8sResult.saDynClient,
 			SAClientset: k8sResult.saClientset,
 			Namespace:   cfg.Namespace,
-			Logger:      logger,
 			Prometheus: repositories.PrometheusConfig{
 				Host:               cfg.PrometheusHost,
 				Namespace:          cfg.PrometheusNamespace,
--- a/distributions/core-bff/bff/internal/api/app.go
+++ b/distributions/core-bff/bff/internal/api/app.go
@@ -100,7 +100,7 @@ func NewApp(cfg config.EnvConfig, logger *slog.Logger) (*App, error) {
 				Instance:           cfg.PrometheusInstance,
 				Port:               cfg.PrometheusPort,
 				RootCAs:            rootCAs,
-				InsecureSkipVerify: cfg.InsecureSkipVerify,
+				InsecureSkipVerify: cfg.InsecureSkipVerify && (cfg.DevMode || cfg.MockK8Client),
 			},
 		}),
 		testEnv:          k8sResult.testEnv,
--- a/distributions/core-bff/bff/internal/api/app.go
+++ b/distributions/core-bff/bff/internal/api/app.go
@@ -49,6 +49,11 @@ type App struct {
 	wsTracker *proxy.ConnectionTracker
 	// wsProxy handles /wss/k8s/* WebSocket relay to the K8s API server
 	wsProxy http.Handler
+	// modelServingProxy handles /api/service/model-serving/* passthrough
+	modelServingProxy http.Handler
+	// devFallbackToken is the kubeconfig bearer token used in dev mode for
+	// proxied requests (Prometheus, model-serving) when the identity has no real token.
+	devFallbackToken string
 }
 
 type k8sSetupResult struct {
@@ -88,6 +93,15 @@ func NewApp(cfg config.EnvConfig, logger *slog.Logger) (*App, error) {
 			SADynClient: k8sResult.saDynClient,
 			SAClientset: k8sResult.saClientset,
 			Namespace:   cfg.Namespace,
+			Logger:      logger,
+			Prometheus: repositories.PrometheusConfig{
+				Host:               cfg.PrometheusHost,
+				Namespace:          cfg.PrometheusNamespace,
+				Instance:           cfg.PrometheusInstance,
+				Port:               cfg.PrometheusPort,
+				RootCAs:            rootCAs,
+				InsecureSkipVerify: cfg.InsecureSkipVerify,
+			},
 		}),
 		testEnv:          k8sResult.testEnv,
 		rootCAs:          rootCAs,
@@ -101,6 +115,11 @@ func NewApp(cfg config.EnvConfig, logger *slog.Logger) (*App, error) {
 		return nil, fmt.Errorf("failed to initialize K8s proxy: %w", err)
 	}
 
+	if err := app.initModelServingProxy(); err != nil {
+		_ = app.Shutdown()
+		return nil, fmt.Errorf("failed to initialize model-serving proxy: %w", err)
+	}
+
 	return app, nil
 }
 
@@ -109,17 +128,15 @@ func initKubernetesClients(cfg config.EnvConfig, logger *slog.Logger) (k8sSetupR
 	var err error
 
 	if cfg.MockK8Client {
-		result.testEnv, result.clientset, err = k8mocks.SetupEnvTest(k8mocks.TestEnvInput{})
+		result.testEnv, result.clientset, result.saDynClient, err = k8mocks.SetupEnvTest(k8mocks.TestEnvInput{
+			CRDs: k8mocks.DefaultCRDs(),
+		})
 		if err != nil {
 			return result, fmt.Errorf("failed to setup envtest: %w", err)
 		}
 		result.factory, err = k8mocks.NewMockedKubernetesClientFactory(result.clientset, result.testEnv, cfg, logger)
 		if err != nil {
 			return result, err
-		}
-		result.saDynClient, err = dynamic.NewForConfig(result.testEnv.Config)
-		if err != nil {
-			return result, fmt.Errorf("failed to create SA dynamic client: %w", err)
 		}
 		result.saClientset, err = kubernetes.NewForConfig(result.testEnv.Config)
 	} else {
--- a/distributions/core-bff/bff/internal/api/errors.go
+++ b/distributions/core-bff/bff/internal/api/errors.go
@@ -1,8 +1,11 @@
 package api
 
 import (
+	"errors"
 	"fmt"
 	"net/http"
+
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 )
 
 // HTTPError represents an HTTP error response with status code and error details.
@@ -68,4 +71,21 @@ func (app *App) methodNotAllowedResponse(w http.ResponseWriter, r *http.Request)
 
 	httpError := &HTTPError{StatusCode: http.StatusMethodNotAllowed, Error: ErrorPayload{Code: "METHOD_NOT_ALLOWED", Message: fmt.Sprintf("the %s method is not supported for this resource", r.Method)}}
 	app.errorResponse(w, r, httpError)
+}
+
+// k8sErrorResponse extracts the status code from a K8s StatusError and returns
+// the appropriate HTTP response. Falls back to 500 for non-K8s errors.
+func (app *App) k8sErrorResponse(w http.ResponseWriter, r *http.Request, err error) {
+	var statusErr *k8serrors.StatusError
+	if errors.As(err, &statusErr) {
+		code := int(statusErr.Status().Code)
+		if code == 0 {
+			code = http.StatusInternalServerError
+		}
+		httpError := &HTTPError{StatusCode: code, Error: ErrorPayload{Code: string(statusErr.Status().Reason), Message: statusErr.Status().Message}}
+		app.LogError(r, err)
+		app.errorResponse(w, r, httpError)
+		return
+	}
+	app.serverErrorResponse(w, r, err)
 }
--- a/distributions/core-bff/bff/internal/api/errors.go
+++ b/distributions/core-bff/bff/internal/api/errors.go
@@ -6,6 +6,7 @@ import (
 	"net/http"
 
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 // HTTPError represents an HTTP error response with status code and error details.
@@ -84,8 +85,8 @@ func (app *App) k8sErrorResponse(w http.ResponseWriter, r *http.Request, err err
 		if code == 0 {
 			code = http.StatusInternalServerError
 		}
-		reason := string(statusErr.Status().Reason)
-		httpError := &HTTPError{StatusCode: code, Error: ErrorPayload{Code: reason, Message: sanitizeK8sReason(reason)}}
+		reason := statusErr.Status().Reason
+		httpError := &HTTPError{StatusCode: code, Error: ErrorPayload{Code: string(reason), Message: sanitizeK8sReason(reason)}}
 		app.LogError(r, err)
 		app.errorResponse(w, r, httpError)
 		return
@@ -93,25 +94,25 @@ func (app *App) k8sErrorResponse(w http.ResponseWriter, r *http.Request, err err
 	app.serverErrorResponse(w, r, err)
 }
 
-func sanitizeK8sReason(reason string) string {
+func sanitizeK8sReason(reason metav1.StatusReason) string {
 	switch reason {
-	case "NotFound":
+	case metav1.StatusReasonNotFound:
 		return "the requested resource could not be found"
-	case "AlreadyExists":
+	case metav1.StatusReasonAlreadyExists:
 		return "the resource already exists"
-	case "Conflict":
+	case metav1.StatusReasonConflict:
 		return "the resource was modified by another request"
-	case "Forbidden":
+	case metav1.StatusReasonForbidden:
 		return "insufficient permissions for this operation"
-	case "Unauthorized":
+	case metav1.StatusReasonUnauthorized:
 		return "authentication required"
-	case "BadRequest":
+	case metav1.StatusReasonBadRequest:
 		return "invalid request"
-	case "Gone":
+	case metav1.StatusReasonGone:
 		return "the requested resource is no longer available"
-	case "Expired":
+	case metav1.StatusReasonExpired:
 		return "the request has expired"
-	case "ServiceUnavailable":
+	case metav1.StatusReasonServiceUnavailable:
 		return "the service is temporarily unavailable"
 	default:
 		return "the server encountered an error processing your request"
--- a/distributions/core-bff/bff/internal/api/errors.go
+++ b/distributions/core-bff/bff/internal/api/errors.go
@@ -75,6 +75,8 @@ func (app *App) methodNotAllowedResponse(w http.ResponseWriter, r *http.Request)
 
 // k8sErrorResponse extracts the status code from a K8s StatusError and returns
 // the appropriate HTTP response. Falls back to 500 for non-K8s errors.
+// The client-facing message is derived from the Reason, not the raw Status.Message,
+// to avoid leaking resource names, RBAC details, or service account identifiers.
 func (app *App) k8sErrorResponse(w http.ResponseWriter, r *http.Request, err error) {
 	var statusErr *k8serrors.StatusError
 	if errors.As(err, &statusErr) {
@@ -82,10 +84,36 @@ func (app *App) k8sErrorResponse(w http.ResponseWriter, r *http.Request, err err
 		if code == 0 {
 			code = http.StatusInternalServerError
 		}
-		httpError := &HTTPError{StatusCode: code, Error: ErrorPayload{Code: string(statusErr.Status().Reason), Message: statusErr.Status().Message}}
+		reason := string(statusErr.Status().Reason)
+		httpError := &HTTPError{StatusCode: code, Error: ErrorPayload{Code: reason, Message: sanitizeK8sReason(reason)}}
 		app.LogError(r, err)
 		app.errorResponse(w, r, httpError)
 		return
 	}
 	app.serverErrorResponse(w, r, err)
+}
+
+func sanitizeK8sReason(reason string) string {
+	switch reason {
+	case "NotFound":
+		return "the requested resource could not be found"
+	case "AlreadyExists":
+		return "the resource already exists"
+	case "Conflict":
+		return "the resource was modified by another request"
+	case "Forbidden":
+		return "insufficient permissions for this operation"
+	case "Unauthorized":
+		return "authentication required"
+	case "BadRequest":
+		return "invalid request"
+	case "Gone":
+		return "the requested resource is no longer available"
+	case "Expired":
+		return "the request has expired"
+	case "ServiceUnavailable":
+		return "the service is temporarily unavailable"
+	default:
+		return "the server encountered an error processing your request"
+	}
 }
--- a/distributions/core-bff/bff/internal/api/model_serving_proxy.go
+++ b/distributions/core-bff/bff/internal/api/model_serving_proxy.go
@@ -12,7 +12,10 @@ import (
 	"github.com/opendatahub-io/odh-dashboard/distributions/core-bff/bff/internal/ssrf"
 )
 
-const modelServingPathPrefix = "/api/service/model-serving"
+const (
+	modelServingPathPrefix     = "/api/service/model-serving"
+	defaultModelServingHostFmt = "https://model-serving-api.%s.svc.cluster.local:443"
+)
 
 func (app *App) initModelServingProxy() error {
 	if app.config.MockK8Client {
@@ -26,7 +29,7 @@ func (app *App) initModelServingProxy() error {
 
 	host := app.config.ModelServingServiceHost
 	if host == "" {
-		host = fmt.Sprintf("https://model-serving-api.%s.svc.cluster.local:443", app.config.Namespace)
+		host = fmt.Sprintf(defaultModelServingHostFmt, app.config.Namespace)
 	}
 
 	targetURL, err := url.Parse(host)
@@ -48,6 +51,8 @@ func (app *App) initModelServingProxy() error {
 		AuthHeaderFn: func(r *http.Request) string {
 			identity, ok := r.Context().Value(constants.RequestIdentityKey).(*k8s.RequestIdentity)
 			if !ok || identity == nil {
+				// Unreachable: InjectRequestIdentity middleware on serviceMux returns 401
+				// before requests reach this handler. Defensive fallback omits the header.
 				return ""
 			}
 			return "Bearer " + identity.ResolveToken(app.devFallbackToken)
--- a/distributions/core-bff/bff/internal/api/model_serving_proxy.go
+++ b/distributions/core-bff/bff/internal/api/model_serving_proxy.go
@@ -1,0 +1,67 @@
+package api
+
+import (
+	"fmt"
+	"net/http"
+	"net/url"
+	"strings"
+
+	"github.com/opendatahub-io/odh-dashboard/distributions/core-bff/bff/internal/constants"
+	k8s "github.com/opendatahub-io/odh-dashboard/distributions/core-bff/bff/internal/integrations/kubernetes"
+	"github.com/opendatahub-io/odh-dashboard/distributions/core-bff/bff/internal/proxy"
+	"github.com/opendatahub-io/odh-dashboard/distributions/core-bff/bff/internal/ssrf"
+)
+
+const modelServingPathPrefix = "/api/service/model-serving"
+
+func (app *App) initModelServingProxy() error {
+	if app.config.MockK8Client {
+		app.modelServingProxy = http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			w.Header().Set("Content-Type", "application/json")
+			w.WriteHeader(http.StatusOK)
+			w.Write([]byte(`{"mock":true}`)) //nolint:errcheck // mock-only handler
+		})
+		return nil
+	}
+
+	host := app.config.ModelServingServiceHost
+	if host == "" {
+		host = fmt.Sprintf("https://model-serving-api.%s.svc.cluster.local:443", app.config.Namespace)
+	}
+
+	targetURL, err := url.Parse(host)
+	if err != nil {
+		return fmt.Errorf("failed to parse model-serving host URL: %w", err)
+	}
+
+	allowHTTP := app.config.DevMode || app.config.MockK8Client
+	insecureSkipVerify := app.config.InsecureSkipVerify && (app.config.DevMode || app.config.MockK8Client)
+
+	rp, err := proxy.NewReverseProxy(proxy.ProxyConfig{
+		TargetURL:          targetURL,
+		RootCAs:            app.rootCAs,
+		InsecureSkipVerify: insecureSkipVerify,
+		AllowHTTP:          allowHTTP,
+		PathRewriteFn: func(r *http.Request) string {
+			return strings.TrimPrefix(r.URL.Path, modelServingPathPrefix)
+		},
+		AuthHeaderFn: func(r *http.Request) string {
+			identity, ok := r.Context().Value(constants.RequestIdentityKey).(*k8s.RequestIdentity)
+			if !ok || identity == nil {
+				return ""
+			}
+			return "Bearer " + identity.ResolveToken(app.devFallbackToken)
+		},
+		StripHeaders:       proxy.SensitiveIngressHeaders(app.config.AuthTokenHeader),
+		ModifyResponse:     ssrf.NewRedirectValidator(app.logger),
+		SSRFValidateTarget: true,
+		SSRFAllowedHosts:   []string{targetURL.Hostname()},
+		Logger:             app.logger,
+	})
+	if err != nil {
+		return fmt.Errorf("failed to create model-serving proxy: %w", err)
+	}
+
+	app.modelServingProxy = rp
+	return nil
+}
--- a/distributions/core-bff/bff/internal/api/namespace_mutation_handler.go
+++ b/distributions/core-bff/bff/internal/api/namespace_mutation_handler.go
@@ -14,7 +14,7 @@ import (
 
 // NamespaceMutationHandler handles GET /api/namespaces/:name/:context.
 // Applies model-serving-platform label/annotation mutations to a namespace.
-// Context 0 (DSG_CREATION) is not supported in RHAII and returns 400.
+// Context 0 (DSG_CREATION) is not supported and returns 400.
 // Contexts 1-3 require SSAR permission: create on serving.kserve.io/servingruntimes.
 func (app *App) NamespaceMutationHandler(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
 	namespaceName := ps.ByName("name")
@@ -33,8 +33,9 @@ func (app *App) NamespaceMutationHandler(w http.ResponseWriter, r *http.Request,
 
 	appCase := models.NamespaceApplicationCase(contextInt)
 
+	// TODO: implement DSG_CREATION for OpenShift (context 0 with admin SSAR)
 	if appCase == models.DSGCreation {
-		app.badRequestResponse(w, r, fmt.Errorf("namespace creation (context 0) is not supported in RHAII"))
+		app.badRequestResponse(w, r, fmt.Errorf("namespace creation (context 0) is not supported"))
 		return
 	}
 
@@ -73,13 +74,15 @@ func (app *App) NamespaceMutationHandler(w http.ResponseWriter, r *http.Request,
 
 	dryRun := r.URL.Query().Get("dryRun") == "All"
 
-	result, err := app.repositories.NamespaceMutation.ApplyMutation(ctx, namespaceName, appCase, dryRun)
-	if err != nil {
-		app.k8sErrorResponse(w, r, err)
+	if err := app.repositories.NamespaceMutation.ApplyMutation(ctx, namespaceName, appCase, dryRun); err != nil {
+		app.LogError(r, err)
+		if writeErr := app.WriteJSON(w, http.StatusOK, &models.NamespaceMutationResponse{Applied: false}, nil); writeErr != nil {
+			app.serverErrorResponse(w, r, writeErr)
+		}
 		return
 	}
 
-	if err := app.WriteJSON(w, http.StatusOK, result, nil); err != nil {
+	if err := app.WriteJSON(w, http.StatusOK, &models.NamespaceMutationResponse{Applied: true}, nil); err != nil {
 		app.serverErrorResponse(w, r, err)
 	}
 }
--- a/distributions/core-bff/bff/internal/api/namespace_mutation_handler.go
+++ b/distributions/core-bff/bff/internal/api/namespace_mutation_handler.go
@@ -49,7 +49,11 @@ func (app *App) NamespaceMutationHandler(w http.ResponseWriter, r *http.Request,
 	}
 
 	ctx := r.Context()
-	identity, _ := ctx.Value(constants.RequestIdentityKey).(*k8s.RequestIdentity)
+	identity, ok := ctx.Value(constants.RequestIdentityKey).(*k8s.RequestIdentity)
+	if !ok || identity == nil {
+		app.unauthorizedResponse(w, r, fmt.Errorf("missing request identity"))
+		return
+	}
 
 	client, err := app.kubernetesClientFactory.GetClient(ctx)
 	if err != nil {
--- a/distributions/core-bff/bff/internal/api/namespace_mutation_handler.go
+++ b/distributions/core-bff/bff/internal/api/namespace_mutation_handler.go
@@ -1,0 +1,85 @@
+package api
+
+import (
+	"fmt"
+	"net/http"
+	"strconv"
+	"strings"
+
+	"github.com/julienschmidt/httprouter"
+	"github.com/opendatahub-io/odh-dashboard/distributions/core-bff/bff/internal/constants"
+	k8s "github.com/opendatahub-io/odh-dashboard/distributions/core-bff/bff/internal/integrations/kubernetes"
+	"github.com/opendatahub-io/odh-dashboard/distributions/core-bff/bff/internal/models"
+)
+
+// NamespaceMutationHandler handles GET /api/namespaces/:name/:context.
+// Applies model-serving-platform label/annotation mutations to a namespace.
+// Context 0 (DSG_CREATION) is not supported in RHAII and returns 400.
+// Contexts 1-3 require SSAR permission: create on serving.kserve.io/servingruntimes.
+func (app *App) NamespaceMutationHandler(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
+	namespaceName := ps.ByName("name")
+	contextStr := ps.ByName("context")
+
+	if namespaceName == "" {
+		app.badRequestResponse(w, r, fmt.Errorf("namespace name is required"))
+		return
+	}
+
+	contextInt, err := strconv.Atoi(contextStr)
+	if err != nil {
+		app.badRequestResponse(w, r, fmt.Errorf("context must be an integer"))
+		return
+	}
+
+	appCase := models.NamespaceApplicationCase(contextInt)
+
+	if appCase == models.DSGCreation {
+		app.badRequestResponse(w, r, fmt.Errorf("namespace creation (context 0) is not supported in RHAII"))
+		return
+	}
+
+	if appCase < models.KServePromotion || appCase > models.ResetModelServingPlatform {
+		app.badRequestResponse(w, r, fmt.Errorf("invalid context: %d (must be 1-3)", contextInt))
+		return
+	}
+
+	if isSystemNamespace(namespaceName) {
+		app.badRequestResponse(w, r, fmt.Errorf("cannot mutate system namespace %q", namespaceName))
+		return
+	}
+
+	ctx := r.Context()
+	identity, _ := ctx.Value(constants.RequestIdentityKey).(*k8s.RequestIdentity)
+
+	client, err := app.kubernetesClientFactory.GetClient(ctx)
+	if err != nil {
+		app.serverErrorResponse(w, r, fmt.Errorf("failed to get kubernetes client: %w", err))
+		return
+	}
+
+	allowed, err := client.CheckAccess(ctx, identity, "create", "serving.kserve.io", "servingruntimes", namespaceName)
+	if err != nil {
+		app.serverErrorResponse(w, r, fmt.Errorf("failed to check namespace permissions: %w", err))
+		return
+	}
+	if !allowed {
+		app.forbiddenResponse(w, r, fmt.Errorf("insufficient permissions to modify serving platform in namespace %q", namespaceName))
+		return
+	}
+
+	dryRun := r.URL.Query().Get("dryRun") == "All"
+
+	result, err := app.repositories.NamespaceMutation.ApplyMutation(ctx, namespaceName, appCase, dryRun)
+	if err != nil {
+		app.k8sErrorResponse(w, r, err)
+		return
+	}
+
+	if err := app.WriteJSON(w, http.StatusOK, result, nil); err != nil {
+		app.serverErrorResponse(w, r, err)
+	}
+}
+
+func isSystemNamespace(name string) bool {
+	return strings.HasPrefix(name, "openshift") || strings.HasPrefix(name, "kube")
+}
--- a/distributions/core-bff/bff/internal/api/nim_integration_handler.go
+++ b/distributions/core-bff/bff/internal/api/nim_integration_handler.go
@@ -1,6 +1,7 @@
 package api
 
 import (
+	"fmt"
 	"net/http"
 
 	"github.com/julienschmidt/httprouter"
@@ -28,6 +29,11 @@ func (app *App) CreateNIMIntegrationHandler(w http.ResponseWriter, r *http.Reque
 	var secretData map[string]string
 	if err := app.ReadJSON(w, r, &secretData); err != nil {
 		app.badRequestResponse(w, r, err)
+		return
+	}
+
+	if len(secretData) == 0 {
+		app.badRequestResponse(w, r, fmt.Errorf("secret data must not be empty"))
 		return
 	}
 
--- a/distributions/core-bff/bff/internal/api/nim_integration_handler.go
+++ b/distributions/core-bff/bff/internal/api/nim_integration_handler.go
@@ -1,0 +1,58 @@
+package api
+
+import (
+	"net/http"
+
+	"github.com/julienschmidt/httprouter"
+)
+
+// GetNIMIntegrationStatusHandler handles GET /api/v1/integrations/nim.
+// Returns the NIM integration status derived from the Account CR.
+// Registered as a publicRoute (no auth) - uses SA client, not user token.
+func (app *App) GetNIMIntegrationStatusHandler(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
+	status, err := app.repositories.NIM.GetNIMStatus(r.Context(), app.config.Namespace)
+	if err != nil {
+		app.k8sErrorResponse(w, r, err)
+		return
+	}
+
+	if err := app.WriteJSON(w, http.StatusOK, status, nil); err != nil {
+		app.serverErrorResponse(w, r, err)
+	}
+}
+
+// CreateNIMIntegrationHandler handles POST /api/v1/integrations/nim.
+// Creates or updates the NIM secret and Account CR.
+// Wrapped with secureAdminRoute - returns 401 for unauthenticated, 403 for non-admin.
+func (app *App) CreateNIMIntegrationHandler(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
+	var secretData map[string]string
+	if err := app.ReadJSON(w, r, &secretData); err != nil {
+		app.badRequestResponse(w, r, err)
+		return
+	}
+
+	status, err := app.repositories.NIM.CreateNIMAccount(r.Context(), app.config.Namespace, secretData)
+	if err != nil {
+		app.k8sErrorResponse(w, r, err)
+		return
+	}
+
+	if err := app.WriteJSON(w, http.StatusOK, status, nil); err != nil {
+		app.serverErrorResponse(w, r, err)
+	}
+}
+
+// DeleteNIMIntegrationHandler handles DELETE /api/v1/integrations/nim.
+// Deletes the NIM Account CR.
+// Wrapped with secureAdminRoute - returns 401 for unauthenticated, 403 for non-admin.
+func (app *App) DeleteNIMIntegrationHandler(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
+	result, err := app.repositories.NIM.DeleteNIMAccount(r.Context(), app.config.Namespace)
+	if err != nil {
+		app.k8sErrorResponse(w, r, err)
+		return
+	}
+
+	if err := app.WriteJSON(w, http.StatusOK, result, nil); err != nil {
+		app.serverErrorResponse(w, r, err)
+	}
+}
--- a/distributions/core-bff/bff/internal/api/nim_serving_handler.go
+++ b/distributions/core-bff/bff/internal/api/nim_serving_handler.go
@@ -1,0 +1,33 @@
+package api
+
+import (
+	"errors"
+	"net/http"
+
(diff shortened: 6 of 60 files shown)


distributions/core-bff/bff/internal/api/nim_serving_handler.go
```diff
@@ -0,0 +1,33 @@
...
+	"net/http"
+
+	"github.com/julienschmidt/httprouter"
+	"github.com/opendatahub-io/odh-dashboard/distributions/core-bff/bff/internal/models"
+)
+
+// GetNIMServingResourceHandler handles GET /api/v1/nim-serving/:nimResource.
+// Performs a cross-resource lookup: fetches NIM Account CR, resolves the resource name,
+// then fetches the Secret or ConfigMap.
+// Registered as a publicRoute (no auth) - uses SA client, not user token.
```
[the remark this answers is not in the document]
caponetto: The Go BFF matches Fastify exactly here - Fastify also returns the full `V1Secret` on a public route (`backend/src/routes/api/nim-serving/index.ts:39-43`).

The frontend depends on the Secret's `data` field - `getNIMData()` in `nimUtils.ts` reads `api_key` and `.dockerconfigjson` directly from it to populate NIM deployment forms. Stripping credentials would break this flow.

If we want to harden this, it would require rearchitecting how NIM credentials flow to the frontend - worth tracking separately.

caponetto:
@coderabbitai review

distributions/core-bff/bff/internal/repositories/nim.go
```diff
@@ -0,0 +1,264 @@
...
+// isNIMCRDNotInstalled checks if the error indicates the NIM CRD is not installed.
+func isNIMCRDNotInstalled(err error) bool {
+	if err == nil {
+		return false
+	}
+	var statusErr *k8serrors.StatusError
+	if !errors.As(err, &statusErr) {
+		return false
+	}
+	return statusErr.Status().Code == 404 && statusErr.Status().Message == "404 page not found"
```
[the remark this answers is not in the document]
caponetto: Good catch. The original code was a direct port of Fastify's string matching (`backend/src/routes/api/integrations/nim/index.ts:17,54`), but you're right that it's fragile.

Fixed by dropping the message check - `isNIMCRDNotInstalled` now only checks for status code 404. This is safe because `k8serrors.IsNotFound` already catches structured 404s (specific resource not found) before this function is reached, so any remaining 404 from the NIM Account API group means the CRD isn't registered. The message content doesn't add signal.

distributions/core-bff/bff/internal/api/model_serving_proxy.go
```diff
@@ -0,0 +1,67 @@
...
+		TargetURL:          targetURL,
+		RootCAs:            app.rootCAs,
+		InsecureSkipVerify: insecureSkipVerify,
+		AllowHTTP:          allowHTTP,
+		PathRewriteFn: func(r *http.Request) string {
+			return strings.TrimPrefix(r.URL.Path, modelServingPathPrefix)
+		},
+		AuthHeaderFn: func(r *http.Request) string {
+			identity, ok := r.Context().Value(constants.RequestIdentityKey).(*k8s.RequestIdentity)
+			if !ok || identity == nil {
```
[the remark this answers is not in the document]
caponetto: This can't happen in practice. The proxy is registered on the `serviceMux`, which is wrapped with `InjectRequestIdentity` in `newCombinedMux` (`routes.go:87`). The middleware validates the token and returns 401 before the request reaches the proxy handler. The `""` return in `AuthHeaderFn` is defensive code for a path that the middleware chain prevents.

The contract tests confirm this - unauthenticated requests to `/api/service/model-serving/*` return 401 (`testCoreBffModelServing.test.ts:117-122`). This also matches Fastify, where the proxy sits under the authenticated route scope.

distributions/core-bff/bff/internal/api/app.go
```diff
@@ -49,6 +49,11 @@ type App struct {
 	wsTracker *proxy.ConnectionTracker
 	// wsProxy handles /wss/k8s/* WebSocket relay to the K8s API server
 	wsProxy http.Handler
+	// modelServingProxy handles /api/service/model-serving/* passthrough
+	modelServingProxy http.Handler
+	// devFallbackToken is the kubeconfig bearer token used in dev mode for
+	// proxied requests (Prometheus, model-serving) when the identity has no real token.
+	devFallbackToken string
```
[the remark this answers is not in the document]
caponetto: This is triple-gated and can't activate in production:

1. `devFallbackToken` is only populated when `DevMode=true && MockK8Client=false` (`app_proxy.go:66`). In production `DevMode` is `false`, so the field is always `""`.
2. `DevFallback` on the identity is only set to `true` inside `InjectRequestIdentity` when `DevMode=true` AND token extraction fails (`middleware_auth.go:29-38`).
3. `ResolveToken` requires both `id.DevFallback && fallbackToken != ""` to use the fallback (`types.go:25`). If either is false, the identity's own token is returned.

In dev mode the fallback gives you your own kubeconfig token, the same privilege level you already have by running `kubectl` locally. Minting a minimally scoped token at startup would require additional infrastructure (token request API, RBAC setup) that doesn't exist in a local dev workflow.

distributions/core-bff/bff/internal/api/app.go
```diff
@@ -88,6 +93,15 @@ func NewApp(cfg config.EnvConfig, logger *slog.Logger) (*App, error) {
...
 			Namespace:   cfg.Namespace,
+			Logger:      logger,
+			Prometheus: repositories.PrometheusConfig{
+				Host:               cfg.PrometheusHost,
+				Namespace:          cfg.PrometheusNamespace,
+				Instance:           cfg.PrometheusInstance,
+				Port:               cfg.PrometheusPort,
+				RootCAs:            rootCAs,
+				InsecureSkipVerify: cfg.InsecureSkipVerify,
+			},
```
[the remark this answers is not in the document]
caponetto: Good catch. The K8s proxy (`app_proxy.go:58`) and model-serving proxy (`model_serving_proxy.go:38`) both gate with `cfg.InsecureSkipVerify && (cfg.DevMode || cfg.MockK8Client)`, but Prometheus was passing the raw flag. Fixed to use the same pattern.

distributions/core-bff/bff/internal/api/errors.go
```diff
@@ -69,3 +72,20 @@ func (app *App) methodNotAllowedResponse(w http.ResponseWriter, r *http.Request)
...
+		if code == 0 {
+			code = http.StatusInternalServerError
+		}
+		httpError := &HTTPError{StatusCode: code, Error: ErrorPayload{Code: string(statusErr.Status().Reason), Message: statusErr.Status().Message}}
+		app.LogError(r, err)
+		app.errorResponse(w, r, httpError)
+		return
+	}
+	app.serverErrorResponse(w, r, err)
+}
```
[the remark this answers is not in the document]
caponetto: Good catch. Fixed by adding a `sanitizeK8sReason` function that maps the K8s Reason to a fixed client-facing message. The full error (including raw `Status.Message`) is still logged server-side via `app.LogError`. The HTTP status code and Reason (as `code` field) are preserved, so the frontend can still branch on status codes. The frontend doesn't do string matching on messages from these endpoints, so no breakage.

distributions/core-bff/bff/internal/api/namespace_mutation_handler.go
```diff
@@ -0,0 +1,85 @@
...
+		return
+	}
+	if !allowed {
+		app.forbiddenResponse(w, r, fmt.Errorf("insufficient permissions to modify serving platform in namespace %q", namespaceName))
+		return
+	}
+
+	dryRun := r.URL.Query().Get("dryRun") == "All"
+
+	result, err := app.repositories.NamespaceMutation.ApplyMutation(ctx, namespaceName, appCase, dryRun)
```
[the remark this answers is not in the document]
caponetto: This matches Fastify exactly. In `namespaceUtils.ts:23-35`, contexts 1-3 use `checkEditNamespacePermission`, which checks `create` on `serving.kserve.io/servingruntimes`. The actual namespace patch is performed by the service account (`saClientset` in `namespace_mutation.go:49`), not the user's token. The SSAR is a semantic authorization gate ("can this user manage model serving in this namespace?"), not a literal permission check for the SA's patch operation. Changing it to `patch namespaces` would be a different authorization model, not a bug fix.

distributions/core-bff/bff/internal/api/namespace_mutation_handler.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+
+	if err := app.WriteJSON(w, http.StatusOK, result, nil); err != nil {
+		app.serverErrorResponse(w, r, err)
+	}
+}
+
+func isSystemNamespace(name string) bool {
+	return strings.HasPrefix(name, "openshift") || strings.HasPrefix(name, "kube")
+}
```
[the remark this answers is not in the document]
caponetto: This matches Fastify's `validateNotSystemNamespace` in `namespaceUtils.ts:42`, which uses the same `name.startsWith('kube')` check. Keeping the same behavior for now. If this needs to be narrowed, it can be done as a separate change.

distributions/core-bff/bff/internal/api/nim_integration_handler.go
```diff
@@ -0,0 +1,58 @@
...
+	status, err := app.repositories.NIM.CreateNIMAccount(r.Context(), app.config.Namespace, secretData)
+	if err != nil {
+		app.k8sErrorResponse(w, r, err)
+		return
+	}
+
+	if err := app.WriteJSON(w, http.StatusOK, status, nil); err != nil {
+		app.serverErrorResponse(w, r, err)
+	}
+}
```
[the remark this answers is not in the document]
caponetto: Fastify also passes the body through without key-level validation (`index.ts:88-92`), and deeper validation (specific keys, value format) would couple the BFF to NIM operator internals. That said, rejecting an empty body is a reasonable quick win. Added a `len(secretData) == 0` guard that returns 400 before reaching the repository layer. K8s enforces Secret schema and etcd size limits on the actual create/update, and the NIM operator validates the API key semantically via its `APIKeyValidation` condition.

distributions/core-bff/bff/internal/api/nim_serving_handler_test.go
```diff
@@ -0,0 +1,87 @@
...
+
+			ps := httprouter.Params{{Key: "nimResource", Value: rt}}
+			app.GetNIMServingResourceHandler(rr, req, ps)
+
+			// NIM CRD is not installed in envtest, so expect 404 from the handler's error path.
+			// The important thing is that the resource type was valid (no panic, no 400).
+			assert.Equal(t, http.StatusNotFound, rr.Code)
+		})
+	}
+}
```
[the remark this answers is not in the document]
caponetto: Good catch. The NIM CRD is installed via `DefaultCRDs()` in envtest. The 404 actually comes from `GetNIMServingResource` finding no NIM Account instance in the namespace (empty `list.Items`), which returns a `NIMNotFoundError`. Updated the comment to reflect the real error path.

distributions/core-bff/bff/internal/api/prometheus_handler.go
```diff
@@ -0,0 +1,99 @@
...
+func resolvePrometheusQueryType(path string) string {
+	switch {
+	case strings.HasSuffix(path, "/queryRange"),
+		strings.HasSuffix(path, "/bias"),
+		strings.HasSuffix(path, "/serving"):
+		return "query_range"
+	default:
+		return "query"
+	}
+}
```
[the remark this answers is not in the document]
caponetto: The silent mismap concern does not apply here. httprouter uses exact path matching, not prefix or wildcard matching. Only the 5 explicitly registered paths (`PrometheusQueryPath`, `PrometheusQueryRangePath`, `PrometheusPVCPath`, `PrometheusBiasPath`, `PrometheusServingPath`) reach `PrometheusQueryHandler`. Any other path gets a 404 from httprouter's `NotFound` handler before the function is ever called.

That said, the duplication point is fair. Switched `resolvePrometheusQueryType` from suffix matching to direct comparison against the route constants, so both registration and resolution use the same source of truth.

distributions/core-bff/bff/internal/api/routes_config.go
```diff
@@ -0,0 +1,34 @@
...
+	"github.com/opendatahub-io/odh-dashboard/distributions/core-bff/bff/internal/config"
+)
+
+const (
+	ConfigPath           = APIPathPrefix + "/config"
+	ComponentsPath       = APIPathPrefix + "/components"
+	StatusPath           = APIPathPrefix + "/status"
+	DashboardConfigPath  = APIPathPrefix + "/dashboardConfig/:namespace/:name"
+	ClusterSettingsPath  = APIPathPrefix + "/cluster-settings"
+	ComponentsRemovePath = APIPathPrefix + "/components/remove"
```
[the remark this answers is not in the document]
caponetto: This matches Fastify, which registers the same route as `fastify.get('/remove', ...)` in `backend/src/routes/api/components/index.ts:18`. The frontend also calls it as GET (`componentsServices.ts:20`). Changing the HTTP method would be a breaking change requiring coordinated frontend updates. This is a valid REST design improvement but out of scope for this PR, which ports existing behavior. It could be tracked as a separate task.

distributions/core-bff/bff/internal/api/routes_model_serving.go
```diff
@@ -0,0 +1,60 @@
+package api
+
+import (
+	"net/http"
+
+	"github.com/julienschmidt/httprouter"
+)
+
+const (
+	ModelServingProxyPrefix  = "/api/service/model-serving/"
```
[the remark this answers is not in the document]
caponetto: Good catch. Changed to `APIPathPrefix + "/service/model-serving/"` to stay consistent with all other route constants in the file.

distributions/core-bff/bff/internal/api/routes_model_serving.go
```diff
@@ -0,0 +1,60 @@
...
+	ServingRuntimesPath      = APIPathPrefix + "/servingRuntimes"
+	NIMServingResourcePath   = APIPathPrefix + "/nim-serving/:nimResource"
+	PrometheusQueryPath      = APIPathPrefix + "/prometheus/query"
+	PrometheusQueryRangePath = APIPathPrefix + "/prometheus/queryRange"
+	PrometheusPVCPath        = APIPathPrefix + "/prometheus/pvc"
+	PrometheusBiasPath       = APIPathPrefix + "/prometheus/bias"
+	PrometheusServingPath    = APIPathPrefix + "/prometheus/serving"
+	NIMIntegrationPath       = APIPathPrefix + "/integrations/nim"
+	NamespaceMutationPath    = APIPathPrefix + "/namespaces/:name/:context"
```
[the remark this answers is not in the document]
caponetto: This matches Fastify, which registers the same route as `fastify.get('/:name/:context', ...)` in `backend/src/routes/api/namespaces/index.ts:7`. Changing the HTTP method would require coordinating frontend changes. Same reasoning as the `RemoveComponentHandler` finding. Valid REST design improvement but out of scope for this PR.

distributions/core-bff/bff/internal/api/serving_runtime_handler_test.go
```diff
@@ -0,0 +1,231 @@
...
+			got, err := extractNamespaceFromBody(tt.body)
+			if tt.wantErr {
+				assert.Error(t, err)
+				return
+			}
+			require.NoError(t, err)
+			assert.Equal(t, tt.want, got)
+		})
+	}
+}
```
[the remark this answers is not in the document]
caponetto: Good catch. Added two tests: `TestCreateServingRuntime_Success` sends a valid ServingRuntime body against envtest (CRD is installed via `DefaultCRDs()`), asserts 200 and verifies the returned object has the expected name and namespace. `TestCreateServingRuntime_DryRun` sends the same body with `?dryRun=All` and verifies the handler returns 200 without persisting the resource.

distributions/core-bff/bff/internal/integrations/kubernetes/k8mocks/base_testenv.go
```diff
@@ -171,251 +177,15 @@ func setupMock(ctx context.Context, mockK8sClient kubernetes.Interface, users []
...
-
-	// Create the service using kubernetes.Interface
-	_, err := k8sClient.CoreV1().Services(namespace).Create(ctx, service, metav1.CreateOptions{})
-	if err != nil {
-		return fmt.Errorf("failed to create service: %w", err)
+	if dynClient != nil {
+		if err := createNIMTestData(ctx, mockK8sClient, dynClient); err != nil {
+			return fmt.Errorf("failed to create NIM test data: %w", err)
+		}
 	}
```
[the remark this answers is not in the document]
caponetto: Good catch. `dynamic.NewForConfig` either returns a non-nil client or returns early with an error at line 88. `setupMock` is only called at line 91 with a guaranteed non-nil `dynClient`. Removed the dead guard.

distributions/core-bff/bff/internal/integrations/kubernetes/token_k8s_client.go
```diff
@@ -133,12 +133,12 @@ func (kc *TokenKubernetesClient) CanListServicesInNamespace(ctx context.Context,
...
+			kc.Logger.Warn("self-SAR failed", "namespace", namespace, "verb", verb, "error", err)
 			return false, err
 		}
 
 		if !resp.Status.Allowed {
-			kc.Logger.Error("self-SAR denied", "namespace", namespace, "verb", verb)
+			kc.Logger.Warn("self-SAR denied", "namespace", namespace, "verb", verb)
 			return false, nil
 		}
 	}
```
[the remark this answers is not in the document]
caponetto: Good catch. Changed the SAR API call failure path from `Warn` to `Error` in both methods. Permission denials (`!resp.Status.Allowed`) stay at `Warn` since those are expected business logic outcomes.

distributions/core-bff/bff/internal/repositories/prometheus.go
```diff
@@ -0,0 +1,107 @@
...
+func NewPrometheusRepository(cfg PrometheusConfig) *PrometheusRepository {
+	host := cfg.Host
+	if host == "" && cfg.Instance != "" && cfg.Namespace != "" && cfg.Port != "" {
+		host = fmt.Sprintf("https://%s.%s.svc.cluster.local:%s", cfg.Instance, cfg.Namespace, cfg.Port)
+	}
+
+	tlsConfig := &tls.Config{
+		RootCAs:            cfg.RootCAs,
+		InsecureSkipVerify: cfg.InsecureSkipVerify, //nolint:gosec // configurable for dev mode
+	}
```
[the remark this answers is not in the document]
caponetto: The `InsecureSkipVerify` value is already gated at the call site in `app.go:103`: `cfg.InsecureSkipVerify && (cfg.DevMode || cfg.MockK8Client)`. By the time it reaches the repository, it can only be `true` in dev/mock mode. Adding a panic inside the repository would be redundant with this gating and too aggressive for a value that's already validated upstream.

That said, adding `MinVersion: tls.VersionTLS12` is a reasonable defense-in-depth improvement. Added that and updated the nolint comment to reference the call-site gating.

distributions/core-bff/bff/internal/repositories/prometheus.go
```diff
@@ -0,0 +1,107 @@
...
+
+	return &PrometheusRepository{
+		httpClient: &http.Client{
+			Transport: &http.Transport{
+				TLSClientConfig: tlsConfig,
+			},
+		},
+		host: host,
+	}
+}
```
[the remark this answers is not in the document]
caponetto: The `Query` method already uses `http.NewRequestWithContext(ctx, ...)` so requests inherit the HTTP server's request context deadline. Still, adding `Timeout: 30 * time.Second` on the client is reasonable defense-in-depth in case a caller passes a context without a deadline. Added it.

distributions/core-bff/bff/internal/repositories/prometheus.go
```diff
@@ -0,0 +1,107 @@
...
+	resp, err := r.httpClient.Do(req)
+	if err != nil {
+		return nil, &models.PrometheusUnavailableError{Reason: err.Error()}
+	}
+	defer resp.Body.Close()
+
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, fmt.Errorf("cannot fetch prometheus data, failed to read response: %w", err)
+	}
```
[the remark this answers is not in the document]
caponetto: Good catch. Wrapped with `io.LimitReader(resp.Body, 10<<20)` to cap at 10 MiB. This is well above typical Prometheus query responses while preventing OOM from a misbehaving or compromised upstream.

distributions/core-bff/bff/openapi/src/core-bff.yaml
```diff
@@ -599,6 +599,406 @@ paths:
...
+          description: Resolved NIM resource
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/NIMServingResourceResponse'
+        '404':
+          $ref: '#/components/responses/BFFNotFound'
+        '500':
+          $ref: '#/components/responses/InternalServerError'
+
```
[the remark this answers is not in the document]
caponetto: This matches Fastify. The NIM serving route in `backend/src/routes/api/nim-serving/index.ts` also has no auth wrapper and returns the full `result.body` from `readNamespacedSecret`/`readNamespacedConfigMap` using the SA client. Both implementations use the SA client for this endpoint, not the user's token. Changing the auth model or redacting the response would be a behavioral change beyond the scope of this PR. If the existing design has a security concern, it should be tracked as a separate issue.

distributions/core-bff/bff/openapi/src/core-bff.yaml
```diff
@@ -599,6 +599,406 @@ paths:
...
+          $ref: '#/components/responses/NamespaceMutationResponse'
+        '400':
+          $ref: '#/components/responses/BadRequest'
+        '401':
+          $ref: '#/components/responses/Unauthorized'
+        '403':
+          $ref: '#/components/responses/BFFForbidden'
+        '500':
+          $ref: '#/components/responses/InternalServerError'
+
```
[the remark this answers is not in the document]
caponetto: This matches Fastify, which registers the same route as `fastify.get('/:name/:context', ...)` in `backend/src/routes/api/namespaces/index.ts`. The frontend calls this endpoint as a GET via `axios()` (default method) at `frontend/src/api/k8s/projects.ts:103`, and the frontend tests assert GET calls against this path (`projects.spec.ts:269-301`). Changing the HTTP method in the BFF or OpenAPI spec would break the frontend. This is a valid REST design improvement but requires coordinated frontend changes and is out of scope for a PR that ports existing behavior.

distributions/core-bff/contract-tests/__tests__/foundation/testCoreBffModelServing.test.ts
```diff
@@ -0,0 +1,245 @@
...
+      expect(result).toMatchContract(apiSchema, {
+        ref: '#/components/schemas/NIMServingResourceResponse',
+        status: 200,
+      });
+    });
+
+    it('should return 404 for invalid resource type', async () => {
+      expectError(await apiClient.get('/api/nim-serving/invalidType'), 404);
+    });
+  });
```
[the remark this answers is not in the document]
caponetto: This matches Fastify. The NIM serving route in `backend/src/routes/api/nim-serving/index.ts` has no auth wrapper and returns the full response body using the SA client. The frontend `getNIMResource` at `nimUtils.ts:23` calls this endpoint without explicitly setting an Authorization header. Moving it from `publicRoute` to `secureRoute` could break the frontend if any code path relies on the request succeeding without auth. This is a valid security improvement but it is a behavioral change from what this PR is porting. It should be tracked as a separate issue alongside the payload redaction concern already noted in the previous finding.

distributions/core-bff/bff/internal/api/serving_runtime_handler_test.go
```diff
@@ -0,0 +1,322 @@
...
+	app.CreateServingRuntimeHandler(rr, req, nil)
+
+	assert.Equal(t, http.StatusOK, rr.Code)
+
+	var result map[string]interface{}
+	err = json.Unmarshal(rr.Body.Bytes(), &result)
+	require.NoError(t, err)
+	meta, _ := result["metadata"].(map[string]interface{})
+	assert.Equal(t, "dryrun-runtime", meta["name"])
+}
```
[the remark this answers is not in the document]
caponetto: Good catch. Added a `Get` call via `testSADynClient` after the handler returns, asserting `apierrors.IsNotFound(err)` to confirm the dry-run request did not persist the resource.

caponetto:

distributions/core-bff/bff/internal/repositories/prometheus.go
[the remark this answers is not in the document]
caponetto: The Prometheus call site in app.go was passing cfg.InsecureSkipVerify directly without the dev/mock gate that the two proxy call sites (app_proxy.go:58, model_serving_proxy.go:38) already had. My initial response claimed the gate was already in place but I pushed it afterwards. Fixed by applying the same cfg.InsecureSkipVerify && (cfg.DevMode || cfg.MockK8Client) gate at app.go:103, consistent with the other TLS paths. Also added MinVersion: tls.VersionTLS12 to the tls.Config in prometheus.go for defense-in-depth.

distributions/core-bff/bff/internal/api/serving_runtime_handler.go
```diff
@@ -0,0 +1,67 @@
...
+		app.badRequestResponse(w, r, err)
+		return
+	}
+
+	if !app.isAllowedNamespace(namespace) {
+		app.forbiddenResponse(w, r, fmt.Errorf("request invalid against a resource from a non-dashboard namespace"))
+		return
+	}
+
+	dryRun := r.URL.Query().Get("dryRun")
```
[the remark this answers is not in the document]
caponetto: Good catch. Normalized the `dryRun` query parameter to only accept `"All"`, consistent with how `NamespaceMutationHandler` already handles it. Note that invalid values wouldn't surface as 500 - `k8sErrorResponse` unwraps the K8s `StatusError` via `errors.As` and forwards the actual 400 status code - but normalizing at the handler level is cleaner than letting invalid values reach the API server.

distributions/core-bff/bff/internal/api/namespace_mutation_handler.go
```diff
@@ -0,0 +1,85 @@
...
+		return
+	}
+
+	if isSystemNamespace(namespaceName) {
+		app.badRequestResponse(w, r, fmt.Errorf("cannot mutate system namespace %q", namespaceName))
+		return
+	}
+
+	ctx := r.Context()
+	identity, _ := ctx.Value(constants.RequestIdentityKey).(*k8s.RequestIdentity)
```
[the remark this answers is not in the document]
caponetto: Good catch. Every other handler checks the `ok` from the identity type assertion (`prometheus_handler.go:36`, `status_handler.go:16`, `user_handler.go:18`, etc.) - this was the only one discarding it. Fixed to match the established pattern. While `secureRoute` guarantees the identity is present today, the explicit check is consistent and guards against future middleware changes.

distributions/core-bff/bff/internal/repositories/nim.go
```diff
@@ -0,0 +1,264 @@
...
+	}
+
+	return r.GetNIMStatus(ctx, namespace)
+}
+
+// DeleteNIMAccount deletes the NIM Account CR.
+func (r *NIMRepository) DeleteNIMAccount(ctx context.Context, namespace string) (*models.NIMDeleteResponse, error) {
+	err := r.saDynClient.Resource(models.NIMAccountGVR).Namespace(namespace).Delete(ctx, models.NIMAccountName, metav1.DeleteOptions{})
+	if err != nil {
+		if k8serrors.IsNotFound(err) || isNIMCRDNotInstalled(err) {
```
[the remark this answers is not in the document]
caponetto: Good catch. Replaced the raw `%v` error formatting with a fixed user-facing message. The raw K8s error is still logged server-side via the error return path for other failure types. The frontend doesn't consume this error string, so no breakage.

distributions/core-bff/bff/internal/api/errors.go
```diff
@@ -69,3 +72,48 @@ func (app *App) methodNotAllowedResponse(w http.ResponseWriter, r *http.Request)
...
+		return "invalid request"
+	case "Gone":
+		return "the requested resource is no longer available"
+	case "Expired":
+		return "the request has expired"
+	case "ServiceUnavailable":
+		return "the service is temporarily unavailable"
+	default:
+		return "the server encountered an error processing your request"
+	}
```
andyatmiami:
```suggestion
		reason := statusErr.Status().Reason
		httpError := &HTTPError{StatusCode: code, Error: ErrorPayload{Code: string(reason), Message: sanitizeK8sReason(reason)}}
		app.LogError(r, err)
		app.errorResponse(w, r, httpError)
		return
	}
	app.serverErrorResponse(w, r, err)
}

func sanitizeK8sReason(reason metav1.StatusReason) string {
	switch reason {
	case metav1.StatusReasonNotFound:
		return "the requested resource could not be found"
	case metav1.StatusReasonAlreadyExists:
		return "the resource already exists"
	case metav1.StatusReasonConflict:
		return "the resource was modified by another request"
	case metav1.StatusReasonForbidden:
		return "insufficient permissions for this operation"
	case metav1.StatusReasonUnauthorized:
		return "authentication required"
	case metav1.StatusReasonBadRequest:
		return "invalid request"
	case metav1.StatusReasonGone:
		return "the requested resource is no longer available"
	case metav1.StatusReasonExpired:
		return "the request has expired"
	case metav1.StatusReasonServiceUnavailable:
		return "the service is temporarily unavailable"
	default:
		return "the server encountered an error processing your request"
	}
```

- prefer typed objects over strings.. 
- only cast to string() in the narrowest scope its required

distributions/core-bff/bff/internal/api/model_serving_proxy.go
```diff
@@ -0,0 +1,67 @@
...
+			w.Header().Set("Content-Type", "application/json")
+			w.WriteHeader(http.StatusOK)
+			w.Write([]byte(`{"mock":true}`)) //nolint:errcheck // mock-only handler
+		})
+		return nil
+	}
+
+	host := app.config.ModelServingServiceHost
+	if host == "" {
+		host = fmt.Sprintf("https://model-serving-api.%s.svc.cluster.local:443", app.config.Namespace)
```
andyatmiami: Nitpick - but i'd love to see this default string value elevated to a `const` at the top of the file for better visibility.

distributions/core-bff/bff/internal/api/model_serving_proxy.go
```diff
@@ -0,0 +1,67 @@
...
+		RootCAs:            app.rootCAs,
+		InsecureSkipVerify: insecureSkipVerify,
+		AllowHTTP:          allowHTTP,
+		PathRewriteFn: func(r *http.Request) string {
+			return strings.TrimPrefix(r.URL.Path, modelServingPathPrefix)
+		},
+		AuthHeaderFn: func(r *http.Request) string {
+			identity, ok := r.Context().Value(constants.RequestIdentityKey).(*k8s.RequestIdentity)
+			if !ok || identity == nil {
+				return ""
```
andyatmiami, line 56: might be worth adding a quick inline comment summarizing what you stated to CodeRabbit on this line.. admittedly I had to read that comment to appreciate/understand what `""` was doing here

distributions/core-bff/bff/internal/api/namespace_mutation_handler.go
```diff
@@ -0,0 +1,89 @@
...
+		app.badRequestResponse(w, r, fmt.Errorf("context must be an integer"))
+		return
+	}
+
+	appCase := models.NamespaceApplicationCase(contextInt)
+
+	if appCase == models.DSGCreation {
+		app.badRequestResponse(w, r, fmt.Errorf("namespace creation (context 0) is not supported in RHAII"))
+		return
+	}
```
andyatmiami, line 40: BFF has `config.PlatformXKS` /  `config.PlatformOpenShift` with an `IsXKS()` helper - shouldn't we be leveraging that here?  This seems like a case where the BFF needs to be smart enough to have _differing behavior_ on RHOAI vs. RHAII...
- this PR already includes logic (like the Prometheus stuff) that isn't initially going to run on RHAII - so it feels like we should be handling the Openshift variant of logic.. but let me know if I am oversimplifying

(also, we shouldn't encode the product name `RHAII` in the error message... too risky of it going stale)

andyatmiami: Claude also flagged this - which might be relevant if we update this function to be _robust_ in its RHOAI support:

> The BFF checks create on serving.kserve.io/servingruntimes (handler line 64) for all contexts
  1-3. The Fastify original has two permission checks: checkAdminNamespacePermission (update on project.openshift.io/projects for context 0)  and checkEditNamespacePermission (create on serving.kserve.io/servingruntimes for contexts 1-3).

distributions/core-bff/bff/internal/repositories/repositories.go
```diff
@@ -29,21 +35,27 @@ type RepositoriesConfig struct {
...
+		DashboardConfig:   NewDashboardConfigRepository(cfg.Platform, cfg.SADynClient),
+		Status:            NewStatusRepository(cfg.SADynClient, cfg.Namespace, auth),
+		Auth:              auth,
+		Components:        NewComponentsRepository(cfg.SADynClient, cfg.SAClientset),
+		ClusterSettings:   NewClusterSettingsRepository(cfg.SAClientset),
+		ConnectionType:    NewConnectionTypeRepository(cfg.SAClientset),
+		AllowedUsers:      NewAllowedUsersRepository(cfg.SADynClient),
+		ServingRuntime:    NewServingRuntimeRepository(cfg.SADynClient),
+		NIM:               NewNIMRepository(cfg.SADynClient, cfg.SAClientset),
+		NamespaceMutation: NewNamespaceMutationRepository(cfg.SAClientset, cfg.Logger),
```
andyatmiami: I'm confused why **only** `NamespaceMutation` repository gets a logger.. seems like all should or none should (?)

distributions/core-bff/bff/internal/repositories/nim.go
```diff
@@ -0,0 +1,264 @@
...
+				Error:   "Unable to delete NIM account: the resource was not found",
+			}, nil
+		}
+		return nil, fmt.Errorf("failed to delete NIM account: %w", err)
+	}
+	return &models.NIMDeleteResponse{Success: true}, nil
+}
+
+// manageNIMSecret creates or replaces the NIM access secret.
+func (r *NIMRepository) manageNIMSecret(ctx context.Context, namespace string, data map[string]string) error {
```
andyatmiami, line 176: The Fastify version does a blind replace from template — it reconstructs the entire secret object from scratch and PUTs it. That means every update resets the secret to a known state: only the template's labels, only the force-validation annotation, only the new stringData. Anything else on the secret is wiped.

The BFF does a fetch-modify-update — it reads the existing secret, mutates it, and PUTs it back. That means existing annotations, labels, finalizers, or any other metadata added by an external controller (or a previous version of this code) survive across updates.

---

are we sure we **want** to change this behavior?  best I can tell - the `nvidia-nim-access` secret is owned/controlled by us.. resetting to "template state" _might be desirable?_ hard to know..

also, we don't call out this difference in the commit message.. which we probably should if we are intentionally deviating from Fastify implementation..

distributions/core-bff/bff/internal/repositories/nim.go
```diff
@@ -0,0 +1,264 @@
...
+			status.VariablesValidationStatus = condStatus
+		}
+
+		if condStatus != "True" && condMessage != "" {
+			errorMessages = append(errorMessages, condMessage)
+		}
+	}
+	if len(errorMessages) > 0 {
+		status.Error = strings.Join(errorMessages, "; ")
+	}
```
andyatmiami, line 246: Fastify only handles the first error defined in the messages.. so this behavior is **absolutely** more informative..

No request for change here - but I just want to flag that nothing in dashboard really does this pattern of "semi-colon joining distinct error messages" - so its not parsed intelligently by the frontend and likely "renders ugly" 😇

distributions/core-bff/bff/internal/repositories/nim.go
```diff
@@ -0,0 +1,264 @@
...
+// isNIMCRDNotInstalled checks if the error indicates the NIM CRD is not installed.
+func isNIMCRDNotInstalled(err error) bool {
+	if err == nil {
+		return false
+	}
+	var statusErr *k8serrors.StatusError
+	if !errors.As(err, &statusErr) {
+		return false
+	}
+	return statusErr.Status().Code == 404
```
andyatmiami: `k8sutil.IsDiscoveryError` helper function seems a much more robust and D.R.Y. check to leverage here. (as this implementation has false positives)

distributions/core-bff/bff/internal/api/prometheus_handler.go
andyatmiami, line 1: We won't have any support from Prometheus in RHAII initially... 

"hiding" this could work in a few ways..
- don't mount endpoints if the env vars/flags are unset (and keep them unset in RHAII)
- guard with openshift only check like other endpoints

distributions/core-bff/bff/internal/api/prometheus_handler.go
```diff
@@ -0,0 +1,97 @@
...
+			msg := fmt.Sprintf("Cannot fetch prometheus data, %s", err.Error())
+			escaped, _ := json.Marshal(msg)
+			app.LogError(r, err)
+			app.WriteJSON(w, http.StatusInternalServerError, models.PrometheusQueryResponse{ //nolint:errcheck // response write failure is non-recoverable
+				Code:     http.StatusInternalServerError,
+				Response: escaped,
+			}, nil)
+		}
+		return
+	}
```
andyatmiami, line 60: Maybe a personal preference - but I find this `if` block _very bulky_ and difficult to digest.. lots of repetition that seems ripe for optimization...

counter-proposal (which you can disagree with and discuss!)

```
// models/prometheus.go:

  type PrometheusHTTPError interface {
        error
        HTTPStatus() int
        ResponseBody() json.RawMessage
  }

  func (e *PrometheusUnavailableError) HTTPStatus() int { return http.StatusServiceUnavailable }
  func (e *PrometheusUnavailableError) ResponseBody() json.RawMessage {
        return []byte(`"Prometheus service is not available"`)
  }
  
  func (e *PrometheusUnparsableError) HTTPStatus() int { return http.StatusUnprocessableEntity }
  func (e *PrometheusUnparsableError) ResponseBody() json.RawMessage {
        return []byte(`"Unprocessable prometheus response"`)
  }

  func (e *PrometheusQueryError) HTTPStatus() int { return http.StatusBadRequest }
  func (e *PrometheusQueryError) ResponseBody() json.RawMessage {
        escaped, _ := json.Marshal(e.Message)
        return escaped
  }

  func (e *PrometheusUpstreamError) HTTPStatus() int { return e.StatusCode }
  func (e *PrometheusUpstreamError) ResponseBody() json.RawMessage {
        escaped, _ := json.Marshal(fmt.Sprintf("Cannot fetch prometheus data, %s", e.Body))
        return escaped
  }

// api/prometheus_handler.go

func (app *App) PrometheusQueryHandler(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
...
        if err != nil {
                var promErr models.PrometheusHTTPError
                if errors.As(err, &promErr) {
                        app.WriteJSON(w, promErr.HTTPStatus(), models.PrometheusQueryResponse{ //nolint:errcheck
                                Code:     promErr.HTTPStatus(),
                                Response: promErr.ResponseBody(),
                        }, nil)
                } else {
                        msg := fmt.Sprintf("Cannot fetch prometheus data, %s", err.Error())
                        escaped, _ := json.Marshal(msg)
                        app.LogError(r, err)
                        app.WriteJSON(w, http.StatusInternalServerError, models.PrometheusQueryResponse{ //nolint:errcheck
                                Code:     http.StatusInternalServerError,
                                Response: escaped,
                        }, nil)
                }
                return
        }

        if err := app.WriteJSON(w, http.StatusOK, result, nil); err != nil {
                app.serverErrorResponse(w, r, err)
        }
}

```

distributions/core-bff/bff/internal/api/routes_base.go
andyatmiami, line 1: `routes_starter` is kinda a weird name.. `routes_base` maybe?

but don't think about this for > 5s.. if you like my suggestion - great! otherwise we'll just move on!

distributions/core-bff/bff/internal/api/serving_runtime_handler.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/julienschmidt/httprouter"
+	"github.com/opendatahub-io/odh-dashboard/distributions/core-bff/bff/internal/models"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+// CreateServingRuntimeHandler handles POST /api/v1/servingRuntimes.
+// Creates a ServingRuntime CR in the specified namespace.
+func (app *App) CreateServingRuntimeHandler(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
+	var body map[string]interface{}
```
andyatmiami, line 17:
```suggestion
	var body map[string]any
```

i think this is the preferred type now in go... but if we change this - we should do it consistently throughout core-bff - so likely out of scope to this PR!

distributions/core-bff/bff/internal/api/serving_runtime_handler.go
```diff
@@ -0,0 +1,71 @@
...
+		app.k8sErrorResponse(w, r, err)
+		return
+	}
+
+	if err := app.WriteJSON(w, http.StatusOK, result.Object, nil); err != nil {
+		app.serverErrorResponse(w, r, err)
+	}
+}
+
+func extractNamespaceFromBody(body map[string]interface{}) (string, error) {
```
andyatmiami, line 53: i wonder if this pattern would justify elevating this function to a helper lib? (naturally only in the event it would be used in multiple places)

Don't need to act on this now - just something to consider.

distributions/core-bff/bff/internal/config/environment.go
```diff
@@ -178,4 +178,23 @@ type EnvConfig struct {
...
+	PrometheusHost string
+
+	// PrometheusNamespace is the namespace where Prometheus/Thanos is deployed.
+	PrometheusNamespace string
+
+	// PrometheusInstance is the Prometheus/Thanos instance name.
+	PrometheusInstance string
+
+	// PrometheusPort is the Prometheus/Thanos RBAC port.
+	PrometheusPort string
```
andyatmiami: On RHAII - **WE** are the deployer.. and this isn't going to initially be set

We have no observability story at this time

don't necessarily need to change anything here -but wanted to make that very apparent.

distributions/core-bff/bff/internal/models/namespace_mutation.go
```diff
@@ -0,0 +1,21 @@
...
+
+const (
+	DSGCreation               NamespaceApplicationCase = 0
+	KServePromotion           NamespaceApplicationCase = 1
+	KServeNIMPromotion        NamespaceApplicationCase = 2
+	ResetModelServingPlatform NamespaceApplicationCase = 3
+)
+
+const (
+	LabelModelMeshEnabled = "modelmesh-enabled"
```
andyatmiami, line 14: Can/should be outside the context of this PR - but I think/thought modelmesh support was deprecated?  so we can probably have tech debt to clean this stuff up...

distributions/core-bff/bff/internal/repositories/prometheus.go
```diff
@@ -0,0 +1,111 @@
...
+	req.Header.Set("Authorization", "Bearer "+token)
+
+	resp, err := r.httpClient.Do(req)
+	if err != nil {
+		return nil, &models.PrometheusUnavailableError{Reason: err.Error()}
+	}
+	defer resp.Body.Close()
+
+	const maxPrometheusBodyBytes = 10 << 20 // 10 MiB
```
andyatmiami: probably better to define this as a constant **outside** this function for improved visibility

distributions/core-bff/bff/internal/repositories/prometheus.go
```diff
@@ -0,0 +1,111 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse query parameters: %w", err)
+	}
+	u.RawQuery = vals.Encode()
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, u.String(), nil)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create Prometheus request: %w", err)
+	}
+	req.Header.Set("Authorization", "Bearer "+token)
```
andyatmiami: i thought we had constants defined for these strings now?  ignore comment if not

distributions/core-bff/bff/internal/repositories/prometheus.go
```diff
@@ -0,0 +1,111 @@
...
+
+	var parsed struct {
+		Status string `json:"status"`
+		Error  string `json:"error"`
+	}
+	if err := json.Unmarshal(body, &parsed); err != nil {
+		return nil, &models.PrometheusUnparsableError{Reason: err.Error()}
+	}
+
+	if parsed.Status == "error" {
```
andyatmiami: would like to see this `"error"` string as a constant - I _don't like magic strings_ 😎

andyatmiami (changes requested):
Thanks for another impactful drop of code here @caponetto !

No **serious** concerns raised here - so feel free to comment that various things can/should be handled in follow ups if necessary..

once we resolve this comments (one way or another) - i'll do one more _dynamic review_  against this code testing stuff in my API client...

distributions/core-bff/bff/internal/api/serving_runtime_handler.go
caponetto: Noted. Will do a consistent sweep across core-bff in the next PR.

distributions/core-bff/bff/internal/repositories/repositories.go
caponetto: Removed the logger from `NamespaceMutationRepository` entirely. The repo now returns errors like all other repos, and the handler logs the error while still returning `{applied: false}` to preserve Fastify's contract. No repository has a logger dependency; all errors bubble up to the handler layer where `app.LogError` handles them consistently.

distributions/core-bff/bff/internal/api/prometheus_handler.go
caponetto: Guarded all Prometheus routes with `requirePlatform(config.PlatformOpenShift, ...)`, same pattern as NIM endpoints.

distributions/core-bff/bff/internal/repositories/prometheus.go
caponetto: We do have `k8s.BearerTokenPrefix` for the bearer prefix. This code was written in parallel while the PR that introduced those constants was still in review, so it missed using them. Fixed now. Let me know if we should also define constants for header keys (like "Authorization"), I can do that consistently across core-bff in the next PR.

distributions/core-bff/bff/internal/api/namespace_mutation_handler.go
caponetto: You're right, we should leverage `IsXKS()` here. Context 0 (DSG_CREATION) is deferred because the OpenShift variant needs admin namespace permission checks plus namespace creation logic that this PR intentionally skips. Returning 400 unconditionally is the safe interim state. Fixed the RHAII string.

On the SSAR gap: correct, Fastify has two checks (admin for context 0, edit for 1-3). We'll add the admin SSAR when we implement DSG_CREATION for OpenShift. The current single check covers 1-3 correctly since context 0 is rejected before reaching SSAR.

distributions/core-bff/bff/internal/api/serving_runtime_handler.go
caponetto: Only used in one place. Will elevate to a shared helper if a second caller appears.

distributions/core-bff/bff/internal/models/namespace_mutation.go
caponetto: The BFF cleanup is small but I believe the label is still meaningful for existing clusters where ModelMesh is deployed. Removing it prematurely could let ModelMesh interfere during KServe promotion. For now we want core-bff to be fully compatible with Fastify so when we decommission Fastify we don't risk missing behavior. Anyway, I'm happy to clean it up if it's already safe to remove.

distributions/core-bff/bff/internal/api/prometheus_handler.go
caponetto: Much cleaner, thank you!

distributions/core-bff/bff/internal/config/environment.go
caponetto: Prometheus routes are now gated with `requirePlatform(OpenShift)` so these env vars are irrelevant on standalone. Updated the comment to mark the section as OpenShift-only and removed the misleading deployer reference.

distributions/core-bff/bff/internal/repositories/nim.go
caponetto: Good catch. The deviation is intentional: fetch-modify-update preserves metadata that external controllers or admission webhooks may have added, while the data itself is still fully reset. My thinking was that blind replace risks clobbering finalizers or labels in a managed namespace, but I can see the argument for matching Fastify exactly. Happy to align with blind replace if you feel the reset-to-template behavior is more predictable here.

caponetto:
Thanks for the thorough review, @andyatmiami. Really appreciate the time you spent going through all of this! Pushed a new round addressing most of the comments.

distributions/core-bff/bff/internal/repositories/nim.go
andyatmiami: for now i'd probably just match fastify (given i have no idea the initial motivation behind this.. accidental vs. intentional.. maybe just add inline comment to help _future generations_ stating "reset-to-template" was implemented to match legacy behavior and that carries risks we may want to fix in future

distributions/core-bff/bff/internal/models/namespace_mutation.go
andyatmiami: <nod> certainly **to be confirmed** -  my "hypothesis" is that literally **NO** cluster that will even have this code deployed moving forward will have modelmesh..

i.e. its not supported in 3.<something> and the 3.x upgrade process actively strips model mesh from cluster.

distributions/core-bff/bff/internal/repositories/prometheus.go
andyatmiami: yeah, good for a consistency sweep in future (but can be very low priority)

distributions/core-bff/bff/internal/repositories/nim.go
caponetto: Done!

distributions/core-bff/bff/internal/models/namespace_mutation.go
caponetto: Makes sense! I'll address this in a dedicated PR once we confirm it.

andyatmiami (approved):
/lgtm
/approve

---

This is a well-structured, substantial PR (~2500+ lines of Go) that ports Fastify model-serving functionality to the Go BFF. The architecture follows established BFF patterns (handler/repository separation, contract-first design, middleware composition).

---

Test evidence from "list" (with mocks) testing:
- [pr-8441-test-results.md](https://github.com/user-attachments/files/29968876/pr-8441-test-results.md)
    - while claude was the "test runner" - methodology and results reviewed and approved by me!

PR #8431, opened 2026-07-06
Add read-only LLM accelerator configs admin page with labels for unsupported resources

mturley:
https://issues.redhat.com/browse/RHOAIENG-69748
https://issues.redhat.com/browse/RHOAIENG-73362

## Description

Adds a new admin settings page for managing `LLMInferenceServiceConfig` accelerator resources, and a shared annotation filtering utility for detecting unsupported "fast channel" builds.

**Closes [RHOAIENG-69748](https://redhat.atlassian.net/browse/RHOAIENG-69748)** — Unsupported resource annotation filtering utility.
**Part of [RHOAIENG-73362](https://redhat.atlassian.net/browse/RHOAIENG-73362)** — Admin LLM accelerator configurations list page. A follow-up PR will add the add/duplicate form to complete this story.

image

cc @vconzola 

### What's included

**Annotation filtering utility** (`packages/model-serving/src/concepts/unsupportedResources.ts`):
- `isUnsupportedResource()`, `isUnsupportedAccepted()`, `isUnsupportedUnaccepted()` — pure functions that read `opendatahub.io/support-status` and `opendatahub.io/unsupported-status-accepted` annotations
- Exported from `@odh-dashboard/model-serving/concepts/unsupportedResources` for reuse across admin pages

**Read-only admin page** (`packages/llmd-serving/src/admin/`):
- Table listing `LLMInferenceServiceConfig` resources filtered by `config-type=accelerator` label
- Name column with `ResourceNameTooltip` popover, "Pre-installed" and "Unsupported" labels
- Enabled column with disabled toggle placeholder (reflects current annotation state)
- Kebab column with "Duplicate" action placeholder
- Disabled "Add LLM accelerator configuration" button placeholder
- Enabled column header info popover
- RBAC gate checking both `create` and `patch` permissions (renders 404 if denied)
- Registered via `app.route` + `app.navigation/href` extensions, gated by `LLMD_SERVING_ID` + `ADMIN_USER`

**Refactoring**:
- `ConfigType` enum replacing hardcoded `'accelerator'` and `'router'` strings
- Optional `labelSelector` param on `useWatchLLMInferenceServiceConfigs` for server-side filtering
- Extracted annotation/label constants to `const.ts` and utility functions (`isConfigEnabled`, `isConfigPreInstalled`) to `utils.ts` from `types.ts`
[...]

PR #8444, opened 2026-07-07
feat(llmd-serving): add admin routing configurations settings tab

YuliaKrimerman:
https://issues.redhat.com/browse/RHOAIENG-70586

## Description

Adds an admin settings tab for managing llm-d routing configurations under **Settings > Model resources and operations > llm-d routing configurations**.

### What's included:

**Extension registration** (`extensions.ts`)
**List page** (`RoutingConfigurationsView` + `RoutingConfigurationsTable`)
**Row component** (`RoutingConfigurationRow`)
**Create/Edit/Duplicate page** (`RoutingConfigurationCreateEdit`)
**Shared components**
- `ConfigYAMLEditor` — reusable YAML editor with metadata field sync (shared with topology configs tab from PR #8400- will be resolved once it merges)

**Cypress mock tests** (7 test cases)

## How Has This Been Tested?

- Manual testing on a live cluster (`zaffre`) with `llmdTopologyConfigs` feature flag enabled

https://github.com/user-attachments/assets/7b1e6a2b-7be1-4837-93dd-c55e5149370f

- Created 4 test `LLMInferenceServiceConfig` resources (3 pre-installed, 1 user-created) to verify:
  - List page renders routing configs with descriptions, pre-installed badges, and routing type labels
  - Enable/disable toggle patches the correct annotation
  - Kebab actions: Duplicate-only for pre-installed, full menu (Duplicate/Edit/Delete) for user-created
  - Add page shows topology type dropdown (defaulting to "All topologies") and configuration source dropdown
  - Sample template API probe correctly disables "Start from sample" when API returns 404
- Cypress mock tests written and passing (7/7, with 1 transient ESOCKETTIMEOUT on first run — passes on retry)
## Test Impact

**New test files:**
- `packages/cypress/cypress/tests/mocked/modelServing/modelServingLlmdRoutingAdmin.cy.ts` — 7 test cases covering tab visibility, table rendering, routing type labels, toggle, and kebab actions
- `packages/cypress/cypress/pages/llmdRoutingSettings.ts` — page object with `RoutingConfigRow` helpers

## Request review criteria:

If you have UI changes:

## Summary by CodeRabbit

* **New Features**
  * Added an LLM-D routing configurations settings entry with a sortable table and full add/edit/duplicate and delete-confirmation flows.
  * Users can toggle routing configurations, view routing type and pre-installed status, and manage row actions.
* **Bug Fixes**
  * Improved enable/disable behavior by correctly adding/removing the disabled annotation.
  * Enhanced create/edit flows for topology/routing configurations with clearer not-found and loading handling, and more reliable YAML initialization for duplication/editing.

### DIFF
--- a/packages/llmd-serving/src/settings/ConfigYAMLEditor.tsx
+++ b/packages/llmd-serving/src/settings/ConfigYAMLEditor.tsx
@@ -1,9 +1,5 @@
 import * as React from 'react';
-import { Language } from '@patternfly/react-code-editor';
-
-const DashboardCodeEditor = React.lazy(
-  () => import('@odh-dashboard/internal/concepts/dashboard/codeEditor/DashboardCodeEditor'),
-);
+import { CodeEditor, Language } from '@patternfly/react-code-editor';
 
 type ConfigYAMLEditorProps = {
   code: string;
@@ -16,20 +12,20 @@ const ConfigYAMLEditor: React.FC<ConfigYAMLEditorProps> = ({
   onCodeChange,
   topologyTypeLabel,
 }) => (
-  <React.Suspense fallback={null}>
-    <DashboardCodeEditor
-      testId="config-yaml-editor"
+  <div data-testid="config-yaml-editor">
+    <CodeEditor
       code={code}
       isUploadEnabled
       isLanguageLabelVisible
       language={Language.yaml}
+      height="400px"
       emptyStateTitle={`Add a ${topologyTypeLabel || 'topology'} configuration`}
       emptyStateBody="Drag a file here, upload files, or start from scratch."
       emptyStateButton="Upload files"
       onCodeChange={onCodeChange}
       options={{ tabSize: 2 }}
     />
-  </React.Suspense>
+  </div>
 );
 
 export default ConfigYAMLEditor;
--- a/packages/llmd-serving/src/settings/ConfigYAMLEditor.tsx
+++ b/packages/llmd-serving/src/settings/ConfigYAMLEditor.tsx
@@ -1,5 +1,6 @@
 import * as React from 'react';
-import { CodeEditor, Language } from '@patternfly/react-code-editor';
+import { Language } from '@patternfly/react-code-editor';
+import DashboardCodeEditor from '@odh-dashboard/internal/concepts/dashboard/codeEditor/DashboardCodeEditor';
 
 type ConfigYAMLEditorProps = {
   code: string;
@@ -12,20 +13,18 @@ const ConfigYAMLEditor: React.FC<ConfigYAMLEditorProps> = ({
   onCodeChange,
   topologyTypeLabel,
 }) => (
-  <div data-testid="config-yaml-editor" style={{ minHeight: '400px', height: '100%' }}>
-    <CodeEditor
-      code={code}
-      isUploadEnabled
-      isLanguageLabelVisible
-      language={Language.yaml}
-      height="400px"
-      emptyStateTitle={`Add a ${topologyTypeLabel || 'topology'} configuration`}
-      emptyStateBody="Drag a file here, upload files, or start from scratch."
-      emptyStateButton="Upload files"
-      onCodeChange={onCodeChange}
-      options={{ tabSize: 2 }}
-    />
-  </div>
+  <DashboardCodeEditor
+    testId="config-yaml-editor"
+    code={code}
+    isUploadEnabled
+    isLanguageLabelVisible
+    language={Language.yaml}
+    emptyStateTitle={`Add a ${topologyTypeLabel || 'topology'} configuration`}
+    emptyStateBody="Drag a file here, upload files, or start from scratch."
+    emptyStateButton="Upload files"
+    onCodeChange={onCodeChange}
+    options={{ tabSize: 2 }}
+  />
 );
 
 export default ConfigYAMLEditor;
--- a/packages/llmd-serving/src/settings/ConfigYAMLEditor.tsx
+++ b/packages/llmd-serving/src/settings/ConfigYAMLEditor.tsx
@@ -1,6 +1,9 @@
 import * as React from 'react';
 import { Language } from '@patternfly/react-code-editor';
-import DashboardCodeEditor from '@odh-dashboard/internal/concepts/dashboard/codeEditor/DashboardCodeEditor';
+
+const DashboardCodeEditor = React.lazy(
+  () => import('@odh-dashboard/internal/concepts/dashboard/codeEditor/DashboardCodeEditor'),
+);
 
 type ConfigYAMLEditorProps = {
   code: string;
@@ -13,18 +16,20 @@ const ConfigYAMLEditor: React.FC<ConfigYAMLEditorProps> = ({
   onCodeChange,
   topologyTypeLabel,
 }) => (
-  <DashboardCodeEditor
-    testId="config-yaml-editor"
-    code={code}
-    isUploadEnabled
-    isLanguageLabelVisible
-    language={Language.yaml}
-    emptyStateTitle={`Add a ${topologyTypeLabel || 'topology'} configuration`}
-    emptyStateBody="Drag a file here, upload files, or start from scratch."
-    emptyStateButton="Upload files"
-    onCodeChange={onCodeChange}
-    options={{ tabSize: 2 }}
-  />
+  <React.Suspense fallback={null}>
+    <DashboardCodeEditor
+      testId="config-yaml-editor"
+      code={code}
+      isUploadEnabled
+      isLanguageLabelVisible
+      language={Language.yaml}
+      emptyStateTitle={`Add a ${topologyTypeLabel || 'topology'} configuration`}
+      emptyStateBody="Drag a file here, upload files, or start from scratch."
+      emptyStateButton="Upload files"
+      onCodeChange={onCodeChange}
+      options={{ tabSize: 2 }}
+    />
+  </React.Suspense>
 );
 
 export default ConfigYAMLEditor;
--- a/packages/llmd-serving/src/settings/RoutingConfigurationCreateEdit.tsx
+++ b/packages/llmd-serving/src/settings/RoutingConfigurationCreateEdit.tsx
@@ -41,7 +41,12 @@ import {
 } from '../api/LLMInferenceServiceConfigs';
 
 const isConfigObject = (value: unknown): value is LLMInferenceServiceConfigKind =>
-  typeof value === 'object' && value !== null && !Array.isArray(value);
+  typeof value === 'object' &&
+  value !== null &&
+  !Array.isArray(value) &&
+  'metadata' in value &&
+  typeof value.metadata === 'object' &&
+  value.metadata !== null;
 
 const stripServerManagedFields = (
   metadata: LLMInferenceServiceConfigKind['metadata'],
@@ -97,20 +102,17 @@ const resolveTopologyFromConfig = (
 const buildSamplesUrl = (topology: TopologyType): string =>
   `/api/service/model-serving/api/v1/samples/llm-d?type=router&topology=${topology}`;
 
-const RoutingConfigurationCreateEdit: React.FC = () => {
-  const { configName } = useParams<{ configName?: string }>();
+const RoutingConfigurationCreateEditInner: React.FC<{
+  existingConfig?: LLMInferenceServiceConfigKind;
+}> = ({ existingConfig }) => {
   const navigate = useNavigate();
   const { state }: { state?: { sourceConfig: LLMInferenceServiceConfigKind } } = useLocation();
+  const { configName } = useParams<{ configName?: string }>();
   const { dashboardNamespace } = useDashboardNamespace();
   const notification = useNotification();
 
   const isDuplicateMode = !!state?.sourceConfig;
   const isEditMode = !!configName && !isDuplicateMode;
-  const [configs] = useWatchRouterConfigs(dashboardNamespace);
-  const existingConfig = React.useMemo(
-    () => (configName ? configs.find((c) => c.metadata.name === configName) : undefined),
-    [configs, configName],
-  );
 
   // --- Topology type state ---
   const resolvedTopology = React.useMemo((): TopologyType | undefined => {
@@ -123,14 +125,9 @@ const RoutingConfigurationCreateEdit: React.FC = () => {
     return undefined;
   }, [existingConfig, state?.sourceConfig]);
 
-  const [selectedTopology, setSelectedTopology] = React.useState<TopologyType | ''>('');
-
-  React.useEffect(() => {
-    if (resolvedTopology && !selectedTopology) {
-      setSelectedTopology(resolvedTopology);
-    }
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [resolvedTopology]);
+  const [selectedTopology, setSelectedTopology] = React.useState<TopologyType | ''>(
+    resolvedTopology ?? '',
+  );
 
   // --- Configuration source state ---
   const [configSource, setConfigSource] = React.useState<'template' | 'editor' | undefined>(
@@ -142,20 +139,27 @@ const RoutingConfigurationCreateEdit: React.FC = () => {
   // Probe API on topology change to check if a sample exists
   React.useEffect(() => {
     if (!selectedTopology || isEditMode || isDuplicateMode) {
-      return;
+      return undefined;
     }
     setTemplateError(false);
     setConfigSource(undefined);
     setYamlCode('');
-    fetch(buildSamplesUrl(selectedTopology))
+
+    const controller = new AbortController();
+    fetch(buildSamplesUrl(selectedTopology), { signal: controller.signal })
       .then((res) => {
         if (!res.ok) {
           setTemplateError(true);
         }
       })
-      .catch(() => {
+      .catch((e: unknown) => {
+        if (e instanceof DOMException && e.name === 'AbortError') {
+          return;
+        }
         setTemplateError(true);
       });
+
+    return () => controller.abort();
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [selectedTopology, isEditMode, isDuplicateMode]);
 
@@ -275,7 +279,7 @@ const RoutingConfigurationCreateEdit: React.FC = () => {
     configSource === 'editor' ||
     (configSource === 'template' && yamlCode !== '');
 
-  // --- Topology type dropdown options (4 topology types, no "all") ---
+  // --- Topology type dropdown options (4 topology types) ---
   const topologyOptions: SimpleSelectOption[] = Object.values(TopologyType).map((tt) => ({
     key: tt,
     label: TopologyTypeLabels[tt],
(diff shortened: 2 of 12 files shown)


packages/model-serving/cypress/tests/mocked/modelServing/modelServingLlmdRoutingAdmin.cy.ts
```diff
@@ -0,0 +1,206 @@
...
+          model: LLMInferenceServiceConfigModel,
+          ns: 'opendatahub',
+          name: 'lab-routing-profile',
+        },
+        patchedConfig,
+      ).as('patchConfig');
+
+      llmdRoutingSettingsPage.getRow('lab-routing-profile').findEnabledSwitch().click();
+      cy.wait('@patchConfig');
+    });
```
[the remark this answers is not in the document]
mturley: +1 to CodeRabbit here

mturley:
@YuliaKrimerman still working on a full review, but something jumps out at me in testing this: when choosing "start from scratch" in the YAML editor, when I try to type, each keystroke is injecting a space and moving my cursor to the next line. not sure what's going on there.

https://github.com/user-attachments/assets/21307419-0d04-49f5-9c4d-ceef31dfbcab

packages/llmd-serving/src/settings/RoutingConfigurationCreateEdit.tsx
```diff
@@ -0,0 +1,481 @@
...
+      const newConfig: LLMInferenceServiceConfigKind = {
+        ...parsed,
+        apiVersion: `${apiGroup}/${apiVer}`,
+        kind: 'LLMInferenceServiceConfig',
+        metadata: {
+          ...parsed.metadata,
+          name: resourceName,
+          namespace: dashboardNamespace,
+          labels: {
+            ...parsed.metadata.labels,
```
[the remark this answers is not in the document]
mturley: +1 to the bot here, and the same is true on line 336 with `...parsed.metadata.annotations`. If I try to submit with just `spec: {}` we get a TypeError rather than submitting and letting the cluster validate the resource.

Image

packages/llmd-serving/src/settings/RoutingConfigurationCreateEdit.tsx
```diff
@@ -0,0 +1,481 @@
...
+    }
+    return undefined;
+  }, [existingConfig, state?.sourceConfig]);
+
+  const k8sNameDesc = useK8sNameDescriptionFieldData({
+    initialData: initialResource,
+    editableK8sName: isDuplicateMode,
+  });
+
+  const [yamlCode, setYamlCode] = React.useState(() => {
```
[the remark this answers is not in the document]
mturley: +1 here, reloading the page in edit mode breaks the state. AI might want you to fix this with a useEffect, I would say let's try to avoid that. Instead we probably want to wrap this component so it doesn't mount at all until useWatchRouterConfigs finishes its initial loading, and then the useState will have the correct data when it first mounts.

The way ServingRuntimes handles this (and the way I handle it for LLM accelerator configs which just merged) is to create a context provider that acts as a loading gate. See `LlmAcceleratorConfigContextProvider`, which renders the provider around an `<Outlet />` and then uses the provider as the `element` prop of the root route like `<Route path="/" element={<LlmAcceleratorConfigContextProvider />}>` (see `LlmAcceleratorConfigRoutes`). That lets us make sure all necessary data and RBAC checks are made before any sub-routes render, and then useState can just assume the data is ready as it does now.

packages/llmd-serving/src/settings/ConfigYAMLEditor.tsx
```diff
@@ -0,0 +1,81 @@
...
+  const sanitizedCode = sanitizeStaticFields(code, {
+    name,
+    displayName,
+    description,
+    topologyType,
+  });
+
+  return (
+    <div data-testid="config-yaml-editor" style={{ minHeight: '400px', height: '100%' }}>
+      <CodeEditor
```
mturley: We should maybe use DashboardCodeEditor here instead of the base PF CodeEditor. It does some height calculation so you could avoid your inline style here, and it has a testId prop so you could remove your wrapping div and pass that directly. Example usage in `CustomServingRuntimeAddTemplate`

packages/llmd-serving/src/settings/ConfigYAMLEditor.tsx
```diff
@@ -0,0 +1,81 @@
...
+};
+
+const ConfigYAMLEditor: React.FC<ConfigYAMLEditorProps> = ({
+  code,
+  onCodeChange,
+  name,
+  displayName,
+  description,
+  topologyType,
+  topologyTypeLabel,
```
mturley: These 4 props are never passed in by the parent, so they're never passed to sanitizeStaticFields and it doesn't modify anything

mturley (changes requested):
Still testing, but these felt significant enough to not wait until tomorrow on posting

packages/llmd-serving/src/types.ts
```diff
@@ -43,8 +43,8 @@ export const CONFIG_TYPE_LABEL = 'opendatahub.io/config-type';
 export const CONFIG_TYPE_ROUTER = 'router';
 const WELL_KNOWN_ANNOTATION = 'serving.kserve.io/well-known-config';
 const DISABLED_ANNOTATION = 'opendatahub.io/disabled';
-const ROUTING_TYPE_ANNOTATION = 'opendatahub.io/routing-type';
-const SUPPORTED_TOPOLOGIES_ANNOTATION = 'opendatahub.io/supported-topologies';
+export const ROUTING_TYPE_ANNOTATION = 'opendatahub.io/routing-type';
+export const SUPPORTED_TOPOLOGIES_ANNOTATION = 'opendatahub.io/supported-topologies';
```
mturley: In my PR https://github.com/opendatahub-io/odh-dashboard/pull/8431 I broke up this llmd-serving types.ts file into types.ts/const.ts/utils.ts and these constants are now in const.ts, to resolve the conflict you can just drop this change and import from there. Sorry about that.

packages/llmd-serving/src/settings/ConfigYAMLEditor.tsx
```diff
@@ -0,0 +1,81 @@
...
+  topologyTypeLabel?: string;
+};
+
+const sanitizeStaticFields = (
+  code: string,
+  fields: {
+    name?: string;
+    displayName?: string;
+    description?: string;
+    topologyType?: string;
```
mturley: I'm also going to need to sanitize the the runtime-version annotation in my upcoming PR. We could just keep adding props and fields, but maybe we should export these as a `LlmConfigYamlStaticFields` type or something, and have that as an object prop passed into the ConfigYAMLEditor?

packages/llmd-serving/src/settings/ConfigYAMLEditor.tsx
```diff
@@ -0,0 +1,81 @@
...
+  description,
+  topologyType,
+  topologyTypeLabel,
+}) => {
+  const sanitizedCode = sanitizeStaticFields(code, {
+    name,
+    displayName,
+    description,
+    topologyType,
+  });
```
mturley: @YuliaKrimerman another problem I think this has (or will have once the props are passed in) is that it's sanitizing the code (injecting form field values) before rendering it in the editor, but that doesn't affect the state, so if the user types a different name in the YAML that is still stored in `yamlCode` and just hidden. I guess that's fine because you then also override the values at submit time, but we could instead have the sanitize function run in the onChange handlers of your form fields and in onCodeChange, updating the YAML state with the combined version.

Maybe it's simpler to not do that and instead keep it how you have it, but either way we should definitely deduplicate the logic for injecting the form field values into the config resource, because right now you have it both here in sanitizeStaticFields and in the handleSubmit of RoutingConfigurationCreateEdit.

In the interest of getting these things landed, we can maybe just follow up with another PR to get that factored out after we have multiple of these editor merged. I think the idea would be that regardless of the form fields present, we have central logic for getting them into the YAML and the submitted resource.

mturley: Ah, I see https://github.com/opendatahub-io/odh-dashboard/pull/8400 was merged and it doesn't have this sanitize function. Maybe we just make this PR consistent with how that works and I can figure out how to adapt in my PR

andrewballantyne: We talked about this on slack -- do you still have concerns here @mturley ?

packages/llmd-serving/src/settings/RoutingConfigurationCreateEdit.tsx
```diff
@@ -0,0 +1,493 @@
...
+  annotations: Record<string, string> | undefined,
+  key: string,
+): Record<string, string> | undefined => {
+  if (!annotations) {
+    return annotations;
+  }
+  const result = { ...annotations };
+  delete result[key];
+  return result;
+};
```
andrewballantyne: These are generic utilities, why are you putting them here?

packages/llmd-serving/src/settings/RoutingConfigurationCreateEdit.tsx
```diff
@@ -0,0 +1,493 @@
...
+  config: LLMInferenceServiceConfigKind,
+): TopologyType | undefined => {
+  const raw = config.metadata.annotations?.[SUPPORTED_TOPOLOGIES_ANNOTATION];
+  if (!raw) {
+    return undefined;
+  }
+  try {
+    const parsed: unknown = JSON.parse(raw);
+    if (Array.isArray(parsed) && parsed.length >= 1) {
+      return Object.values(TopologyType).find((t) => t === parsed[0]);
```
andrewballantyne, line 57: Sorry I'm not sure I understand this logic...

You're looking at only index 0? Wouldn't this need to be a double loop to validate "all configs" for topology type? 

We only want one because of our dropdown?

packages/llmd-serving/src/settings/RoutingConfigurationCreateEdit.tsx
```diff
@@ -0,0 +1,493 @@
...
+    }
+  } catch {
+    // invalid JSON — fall through
+  }
+  return undefined;
+};
+
+const buildSamplesUrl = (topology: TopologyType): string =>
+  `/api/service/model-serving/api/v1/samples/llm-d?type=router&topology=${topology}`;
```
andrewballantyne: Shouldn't this be in a utility and called something more direct?
```suggestion
const buildRouterSamplesUrl = (topology: TopologyType): string =>
  `/api/service/model-serving/api/v1/samples/llm-d?type=router&topology=${topology}`;
```

packages/llmd-serving/src/settings/RoutingConfigurationCreateEdit.tsx
YuliaKrimerman: Yes, we only want one because of our dropdown. The dropdown is a single-select, so when we create a config we always set supported-topologies as a single-element array like [\"workload-single-node\"]. When resolving back for edit/duplicate, we just read that one value back. If a config has multiple supported topologies (set outside the UI), we take the first one as the pre-selected value -- the admin can change it anyways

packages/llmd-serving/src/settings/RoutingConfigurationCreateEdit.tsx
```diff
@@ -0,0 +1,493 @@
...
+    }
+    if (state?.sourceConfig) {
+      return resolveTopologyFromConfig(state.sourceConfig);
+    }
+    return undefined;
+  }, [existingConfig, state?.sourceConfig]);
+
+  const [selectedTopology, setSelectedTopology] = React.useState<TopologyType | ''>(
+    resolvedTopology ?? '',
+  );
```
andrewballantyne: Stop using memoize when you're computing state -- it's only done and used once... this is a default value in useState, recomputing means nothing.

Reuse the `useState(() => {})` function model

packages/llmd-serving/src/settings/RoutingConfigurationCreateEdit.tsx
```diff
@@ -0,0 +1,493 @@
...
+  // --- Configuration source state ---
+  const [configSource, setConfigSource] = React.useState<'template' | 'editor' | undefined>(
+    undefined,
+  );
+  const [templateLoading, setTemplateLoading] = React.useState(false);
+  const [templateError, setTemplateError] = React.useState(false);
+
+  // Probe API on topology change to check if a sample exists
+  React.useEffect(() => {
+    if (!selectedTopology || isEditMode || isDuplicateMode) {
```
andrewballantyne, line 98: Interesting -- have we tested using a sample when you have an existing resource? Does it even work? Are you reconciling the states properly to avoid a k8s error?

YuliaKrimerman: the probe only fires in create mode -- the guard at line 141 skips it entirely for isEditMode and isDuplicateMode. In edit mode, the YAML is populated from the existing resource (not from the samples API), and the configuration source dropdown is hidden. In duplicate mode, the YAML is pre-filled from the source config. The samples API is never called in either case, so there's no state reconciliation issue.

packages/llmd-serving/src/settings/RoutingConfigurationCreateEdit.tsx
```diff
@@ -0,0 +1,493 @@
...
+          name: `${state.sourceConfig.metadata.name}-copy`,
+          annotations: {
+            ...cleanAnnotations,
+            'openshift.io/display-name': `Copy of ${getDisplayNameFromK8sResource(
+              state.sourceConfig,
+            )}`,
+          },
+        },
+        spec: state.sourceConfig.spec,
+      });
```
andrewballantyne: This is such a great usability function -- would be great to have this so we can just "cleanResourceForYAMLViewer" or something like that

packages/llmd-serving/src/settings/RoutingConfigurationsView.tsx
```diff
@@ -0,0 +1,26 @@
...
+  const { dashboardNamespace } = useDashboardNamespace();
+  const [configs, loaded, error] = useWatchRouterConfigs(dashboardNamespace);
+
+  return (
+    <ApplicationsPage
+      title="llm-d routing configurations"
+      description="Manage llm-d routing configurations. Enabled configurations can be selected when deploying models with advanced routing."
+      loaded={loaded}
+      loadError={error}
+      empty={false}
```
andrewballantyne: Why are we not doing an empty state here? You're making a page with zero rows but column headers instead 🤔 Is that what @vconzola wanted? This seems different than normal...

YuliaKrimerman: I don't think we ever discussed the empty states

YuliaKrimerman: We set empty={false} intentionally so the table toolbar (with the Add button) always renders even when there are zero configs -- the table has its own emptyTableView that shows 'No routing configurations found' with guidance text. If you'd prefer the full-page ApplicationsPage empty state instead (matching other settings pages), I can switch to empty={loaded && configs.length === 0}. What's your preference?

packages/llmd-serving/src/settings/RoutingConfigurationCreateEdit.tsx
andrewballantyne, line 1: Interesting -- I missed this on your other PR too -- but your edit screen edits "fine" if you type the URL incorrectly and don't have the backing resource -- me thinks you need a 404 page here.

packages/llmd-serving/src/settings/RoutingConfigurationsTable.tsx
```diff
@@ -0,0 +1,167 @@
...
+    try {
+      await patchLLMInferenceServiceConfig(config, updatedConfig);
+    } catch (e) {
+      notification.error(
+        `Error ${currentlyEnabled ? 'disabling' : 'enabling'} configuration`,
+        e instanceof Error ? e.message : 'Unknown error',
+      );
+    } finally {
+      setTogglingConfigs((prev) => ({ ...prev, [configName]: false }));
+    }
```
andrewballantyne, line 63: Can you update this to fit my feedback from the other topology-types PR?

packages/llmd-serving/src/settings/RoutingConfigurationRow.tsx
```diff
@@ -0,0 +1,111 @@
...
+      </Td>
+      <Td dataLabel="Routing type">
+        {routingType ? (
+          <Label color="blue" data-testid="routing-type-label">
+            {RoutingTypeLabels[routingType]}
+          </Label>
+        ) : (
+          '-'
+        )}
+      </Td>
```
andrewballantyne, line 77: Not sure what's happening here -- perhaps an old Prototype link? We did away with this weird terminology.

You also seemed to have made an annotation for this? 
```
    opendatahub.io/routing-type: scheduler-and-http-route
```

Image

We only have two types of keys:

Image

YuliaKrimerman: Yeah these were created when we had the old design with the Scheduler/ HTTP router options, this is just leftoverss

packages/llmd-serving/src/settings/RoutingConfigurationCreateEdit.tsx
YuliaKrimerman: true, I'll add to both

packages/llmd-serving/src/settings/RoutingConfigurationCreateEdit.tsx
andrewballantyne: @vconzola this seems like a gap. Think we need a path forward here if the sample wants two config types. Do we just persist it if it comes from the sample?

packages/llmd-serving/src/settings/RoutingConfigurationCreateEdit.tsx
andrewballantyne: Right you are. Wrote this before I started testing

packages/llmd-serving/src/settings/RoutingConfigurationsView.tsx
andrewballantyne: Look at a no projects page or no deployments in a project - we have a better empty state than column headers

packages/llmd-serving/src/settings/RoutingConfigurationCreateEdit.tsx
andrewballantyne: We could try a multi-select, but the weirdness of their sample API won't allow this to be smooth. 

Not sure I appreciate this outcome, not sure howe we avoid it though.. samples are not valid translations for us to edit smoothly... translating the sample to JSON to "force it into something they selected" will destroy all the comments... this is just unfortunate.

Honestly, `[0]` is probably our best bet ... 🤔 I think we have a topology config already selected at this point -- maybe they just take that instead of the dual/wider selection... but even that sucks because the YAML will say something we don't support.

This likely needs a revisit @YuliaKrimerman

andrewballantyne:
The description & display name translation is clunky quite a bit...

K8s name => Display Name
"description" => "opendatahub.io/description" duplication

But honestly, we can come back to this.

I think we need to just get some merges done here, and we can spend time improving things after that.

/approve
/lgtm

PR #8452, opened 2026-07-07
fix e2e testSingleModelAdminCreation

sridarna:
## Description

`testSingleModelAdminCreation` fails when multiple global serving runtimes share the same display name (e.g. `vLLM CPU .x86. ServingRuntime for KServe`) but have different template names. The page object was matching multiple menu items and Cypress could not select a single option.

Also refresh fixture project/model names to avoid collisions with leftover test resources.

## How Has This Been Tested?

Passing locally against an ODH cluster.

Screenshot 2026-07-07 at 3 29 24 PM

## Test Impact

Updates the `testSingleModelAdminCreation` E2E fixture and the `selectServingRuntimeOption` page-object helper used by that test. No new tests added — this fixes an existing flaky/failing E2E path.

## Request review criteria:

If you have UI changes:


packages/cypress/cypress/pages/modelServing.ts
```diff
@@ -1051,7 +1051,13 @@ class ModelServingWizard extends Wizard {
...
         // Select from a list of serving runtimes, including custom ones
         this.findServingRuntimeTemplateSearchSelector().click();
-        this.findGlobalScopedTemplateOption(name).should('exist').click();
+        // Duplicate display names can match multiple menu items; pick the first for E2E stability
+        this.getGlobalScopedServingRuntime()
+          .find()
+          .findAllByRole('menuitem', { name: new RegExp(name), hidden: true })
+          .first()
+          .should('exist')
+          .click();
```
mturley, line 1060: This is probably ok for now, but after the changes from https://redhat.atlassian.net/browse/RHOAIENG-75466 we might want to specifically select the one that doesn't have a "limited support" badge on it somehow. Or just use the resource name rather than the display name, which we could put in a testid or something. We will continue to have more than one item with the same display name that are distinguished only by annotations

mturley (commented):
/lgtm
/approve

PR #8458, opened 2026-07-08
feat(RHOAIENG-69289): Show OOTB images on Workbench images admin page with enable/disable toggle

jesuino:
## Description

Feature: Show out-of-the-box (OOTB) images on the Workbench Images admin page with an enable/disable toggle.

**Key changes:**
**New API (imageStreams.ts)** — Added patchImageStreamTag function to enable/disable OOTB image tags via the opendatahub.io/workbench-image-recommended annotation

**New component (OOTBImageStatusToggle.tsx)** — Toggle switch component for enabling/disabling OOTB images in the admin UI

**Table updates (BYONImagesTableRow.tsx, BYONImages.tsx)** — Modified the BYON images table to display OOTB images alongside custom images, with different action columns

**Hardware profiles (BYONImageHardwareProfiles.tsx)** — Refactored to support both BYON and OOTB images

**Image utilities (imageStreamUtils.ts)** — Added isOOTBImageEnabled and getImageStreamTagName helpers, updated isImageTagEnabled logic

**Spawner integration (spawnerUtils.ts)** — Added isImageEnabled function to check image availability based on OOTB vs BYON status

Tests — Added unit tests for the new image stream utilities and spawner utils

**Image Streams page with OOTB images. Notice that OOTB are not editable**
image

**Spawner lists only Enabled images when creating a new workbench, it includes OOTB images now that can be disabled:**
image

## How Has This Been Tested?

* Manual tests on an RHOAI 3.5 EA1 cluster
* Reviewed by my colleagues on a screen sharing session

## Test Impact
* Made it sure that existing functionality is not broken by adding my changes

## Request review criteria:

If you have UI changes: 

## Summary by CodeRabbit

* **New Features**
  * Added pre-installed (OOTB) workbench image support with provider detection, smarter recommended/default tag selection, and “pre-installed” labeling.
  * Added an enable/disable toggle for OOTB images, and hid OOTB-specific actions in the BYON images table.

* **Bug Fixes**
  * Hidden OOTB images are excluded from selection and spawning flows.
  * Read-only mode now hides “Create profile” controls for hardware profiles.

* **UI Updates**
  * Updated BYON empty-state text and changed the “Imported” column label to “Created”.

* **Tests**
  * Expanded unit and Cypress coverage for OOTB visibility/provider/tag mapping and hidden behavior.


DaoDaoNoCode:

frontend/src/pages/BYONImages/OOTBImageStatusToggle.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+type OOTBImageStatusToggleProps = {
+  image: BYONImage;
+};
+
+const OOTBImageStatusToggle: React.FC<OOTBImageStatusToggleProps> = ({ image }) => {
+  const { dashboardNamespace } = useDashboardNamespace();
+  const [isLoading, setLoading] = React.useState(false);
+  const [isEnabled, setEnabled] = React.useState(image.visible);
+  const notification = useNotification();
```
DaoDaoNoCode, line 16: `useState(image.visible)` only reads the initial value on mount — if the parent re-renders with updated data (e.g., another admin toggled the same image and the K8s watch fires), the local `isEnabled` state won't sync.

The existing `BYONImageStatusToggle` has a similar issue, but at least includes a `useEffect` that syncs on `image.error` changes. This toggle has no sync at all.

A simple `useEffect` would fix it:
```ts
React.useEffect(() => {
  setEnabled(image.visible);
}, [image.visible]);
```

Separately — while you're in this area, `BYONImageStatusToggle.tsx:35` has a pre-existing copy-paste issue: the error message says "serving runtime" instead of "workbench image", and uses "enable"/"disable" instead of "enabling"/"disabling". Your new OOTB toggle has the correct wording — might be worth fixing the BYON one to match.

frontend/src/pages/projects/screens/spawner/spawnerUtils.ts
```diff
@@ -382,6 +382,13 @@ export const isInvalidBYONImageStream = (imageStream: ImageStreamKind): boolean
   return isBYONImageStream(imageStream) && (activeTag === undefined || activeTag.items === null);
 };
 
+export const isHiddenOOTBImageStream = (imageStream: ImageStreamKind): boolean => {
+  if (isBYONImageStream(imageStream)) {
+    return false;
+  }
+  return imageStream.metadata.annotations?.['opendatahub.io/notebook-image-hidden'] === 'true';
```
DaoDaoNoCode: This uses the raw annotation string, but `ImageStreamAnnotation.HIDDEN` is already defined in `types.ts:488` and used in both `imageStreamUtils.ts:116` (inside `getVisible()`) and `imageStreams.ts:233` (inside `patchOOTBImageStreamHidden`). If the annotation key ever changes, this one would be missed.

```suggestion
  return imageStream.metadata.annotations?.[ImageStreamAnnotation.HIDDEN] === 'true';
```

(Import `ImageStreamAnnotation` from `#~/types` — it's already imported by `spawnerUtils.ts` in other places if not already here.)

frontend/src/api/k8s/imageStreams.ts
```diff
@@ -214,3 +216,37 @@ export const updateBYONImageStream = async (
...
+  const current = await getImageStream(namespace, name);
+
+  if (!current.metadata.annotations) {
+    patches.push({ op: 'add', path: '/metadata/annotations', value: {} });
+  }
+
+  const hasAnnotation = current.metadata.annotations?.[ImageStreamAnnotation.HIDDEN] !== undefined;
+  const replacedAnnotation = ImageStreamAnnotation.HIDDEN.replace(/\//g, '~1');
+  const annotationPath = `/metadata/annotations/${replacedAnnotation}`;
+  patches.push({
```
DaoDaoNoCode, line 235: Per RFC 6902 §4.1, JSON Patch `add` on an existing object member replaces its value — so `add` works regardless of whether the key already exists. On the other hand, `replace` (§4.3) fails with a 422 if the key doesn't exist.

Since this function does a GET-then-PATCH, there's a small window where the annotation could be removed between the two calls (e.g., operator reconciliation or another admin). Using `add` unconditionally would be more robust: op: 'add'

The GET is still useful for the annotations-existence check on line 229 (to know whether to add the parent `/metadata/annotations` object), but the per-key operation can safely always be `add`.

frontend/src/api/k8s/imageStreams.ts
```diff
@@ -214,3 +216,37 @@ export const updateBYONImageStream = async (
     ),
   );
 };
+
```
DaoDaoNoCode, line 219: Nice design choice using an annotation here instead of a label. BYON visibility uses the `opendatahub.io/notebook-image` label (indexed, used in K8s label selectors by `useImageStreams`), but that approach wouldn't work for OOTB images because the operator would reconcile/overwrite any label changes the dashboard makes.

Using the hidden annotation as an overlay is the right call — but it'd be worth adding a brief comment explaining this reasoning. Someone maintaining this later might wonder why BYON and OOTB use different visibility mechanisms and try to "unify" them, which would break things.

DaoDaoNoCode:
A couple of observations on things outside the diff that are worth considering as follow-ups:

1. **"Imported" column header is misleading for OOTB images** — `tableData.tsx:41` defines the column label as "Imported", which made sense when the table only showed BYON images. Now that OOTB images appear too, the label is inaccurate for them — they're provisioned by the operator, not imported by an admin. Consider changing it to something like "Created" or conditionally rendering the label based on image type.

2. **Pre-existing error message in `BYONImageStatusToggle.tsx:35`** — The error notification says `"Error enable the serving runtime"` which has two issues: "serving runtime" should be "workbench image" (copy-paste from another component), and "enable"/"disable" should be "enabling"/"disabling" (grammar). The new `OOTBImageStatusToggle` already uses the correct wording — might be worth aligning the BYON toggle to match while you're in this area.

frontend/src/api/k8s/imageStreams.ts
jesuino: That's great to know about this, thanks!

frontend/src/api/k8s/imageStreams.ts
jesuino: Do you think I should add this information to some existing documentation or comment on the code?

DaoDaoNoCode: The inline code comments are good!

jesuino:
Thanks for the review, @DaoDaoNoCode 

I made the changes, kindly let me know if am missing something.

DaoDaoNoCode (approved):
/lgtm
Pull it down and tested, it works well!
Just one follow-up question, not blocking - do we want to do a refactor of the table in the future or at least do some UX updates? 

Image

Now the `Recommended hardware profiles` column is taking a huge width of the table while it's empty for all of the OOTB images.

jesuino:
Thank you @DaoDaoNoCode 

>  Just one follow-up question, not blocking - do we want to do a refactor of the table in the future or at least do some UX updates?

Yes, there is a requirement for not showing this for OOTB images. I will discuss this with my colleagues as part of [RHOAIENG-69290](https://redhat.atlassian.net/browse/RHOAIENG-69290), which is an extension of this feature to include some edge cases, more tests and UX improvements. 

[RHOAIENG-69290]: https://redhat.atlassian.net/browse/RHOAIENG-69290?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #8472, opened 2026-07-08
Add enabled toggle with unsupported acceptance flow for LLM accelerator configs

mturley:
https://issues.redhat.com/browse/RHOAIENG-69749
https://issues.redhat.com/browse/RHOAIENG-69752

Followup to https://github.com/opendatahub-io/odh-dashboard/pull/8431

cc @vconzola 

## Description
Wires up the enable toggle on the LLM accelerator configurations admin page with support-status awareness:
- **Toggle ON/OFF**: Patches `opendatahub.io/disabled` annotation via `patchLLMInferenceServiceConfig`
- **Unsupported + unaccepted configs**: Toggle shows OFF regardless of disabled annotation. Toggling ON opens a confirmation modal ("Enable limited support accelerator configuration") before enabling
- **On accept**: Patches both `opendatahub.io/unsupported-status-accepted: "true"` and `opendatahub.io/disabled: "false"` in a single call
- **Already-accepted unsupported configs**: Toggle normally without modal
- **Error handling**: Shows error notification on patch failure
- **Microcopy update**: "Unsupported" label changed to "Limited support" per UX guidance

### New components
- `UnsupportedStatusAcceptanceModal` (`packages/model-serving/src/components/`) — reusable modal parameterized by `resourceTypeLabel` (e.g. "accelerator configuration", "runtime")
- `LlmAcceleratorConfigEnabledToggle` (`packages/llmd-serving/src/admin/`) — self-contained toggle with patch logic, loading state, and unsupported flow

### Screenshots

When clicking a toggle on a limited-support config that has not yet been approved by a user, this modal appears:
Screenshot 2026-07-08 at 7 36 34 PM

After confirming that modal (or when clicking a toggle that doesn't require approval, either because it's not limited-support or it has already been approved), while the annotations are being patched the toggle is disabled. If the patch fails, an alert is shown: 
image

## How Has This Been Tested?
- Unit tests: 32 tests across modal, toggle, and table row (all passing)
- Cypress mock tests: 9 tests covering toggle states, acceptance flow, payload assertions (all passing)
- Manual testing via dev server using zaffre ROSA cluster

## Test Impact
- **Unit tests added**: `UnsupportedStatusAcceptanceModal.spec.tsx` (5 tests), `LlmAcceleratorConfigEnabledToggle.spec.tsx` (10 tests)
- **Unit tests updated**: `LlmAcceleratorConfigTableRow.spec.tsx` — reduced mocking, uses shared `mockLLMInferenceServiceConfigK8sResource` factory
- **Cypress mock tests expanded**: `llmAcceleratorConfigs.cy.ts` — 7 new tests for toggle states, unsupported modal flow, and PATCH payload validation

## Request review criteria:
If you have UI changes:

## Summary by CodeRabbit

* **New Features**
  * Added an “Enable” confirmation modal for unsupported-but-unaccepted accelerator configurations.
  * Introduced a dedicated enable/disable toggle component and updated the table “Enabled” interaction flow.
  * Improved name column sorting using resource display names.

* **Bug Fixes**
  * Prevent enabling unsupported-but-unaccepted configs until confirmation is accepted.
  * Supported and already-accepted limited-support configs now toggle without prompting.

* **Tests**
  * Expanded unit and end-to-end coverage for toggle states, modal accept/cancel behavior, and expected PATCH annotation updates.

### DIFF
--- a/packages/llmd-serving/src/admin/LlmAcceleratorConfigEnabledToggle.tsx
+++ b/packages/llmd-serving/src/admin/LlmAcceleratorConfigEnabledToggle.tsx
@@ -1,0 +1,96 @@
+import * as React from 'react';
+import { Switch } from '@patternfly/react-core';
+import useNotification from '@odh-dashboard/internal/utilities/useNotification';
+import {
+  isUnsupportedUnaccepted,
+  UNSUPPORTED_STATUS_ACCEPTED_ANNOTATION,
+} from '@odh-dashboard/model-serving/concepts/unsupportedResources';
+import UnsupportedStatusAcceptanceModal from '@odh-dashboard/model-serving/components/UnsupportedStatusAcceptanceModal';
+import type { LLMInferenceServiceConfigKind } from '../types';
+import { DISABLED_ANNOTATION } from '../const';
+import { isConfigEnabled } from '../utils';
+import { patchLLMInferenceServiceConfig } from '../api/LLMInferenceServiceConfigs';
+
+type LlmAcceleratorConfigEnabledToggleProps = {
+  config: LLMInferenceServiceConfigKind;
+};
+
+const LlmAcceleratorConfigEnabledToggle: React.FC<LlmAcceleratorConfigEnabledToggleProps> = ({
+  config,
+}) => {
+  const notification = useNotification();
+  const [isToggling, setIsToggling] = React.useState(false);
+  const [showAcceptanceModal, setShowAcceptanceModal] = React.useState(false);
+
+  const configName = config.metadata.name;
+  const unsupportedUnaccepted = isUnsupportedUnaccepted(config);
+  const effectiveEnabled = unsupportedUnaccepted ? false : isConfigEnabled(config);
+
+  const patchConfigAnnotations = React.useCallback(
+    async (annotationUpdates: Record<string, string>) => {
+      setIsToggling(true);
+      const updatedConfig: LLMInferenceServiceConfigKind = {
+        ...config,
+        metadata: {
+          ...config.metadata,
+          annotations: {
+            ...config.metadata.annotations,
+            ...annotationUpdates,
+          },
+        },
+      };
+
+      try {
+        await patchLLMInferenceServiceConfig(config, updatedConfig);
+      } catch (e) {
+        notification.error(
+          'Error updating accelerator configuration',
+          e instanceof Error ? e.message : 'Unknown error',
+        );
+      } finally {
+        setIsToggling(false);
+      }
+    },
+    [config, notification],
+  );
+
+  const handleToggle = React.useCallback(() => {
+    if (effectiveEnabled) {
+      patchConfigAnnotations({ [DISABLED_ANNOTATION]: 'true' });
+    } else if (unsupportedUnaccepted) {
+      setShowAcceptanceModal(true);
+    } else {
+      patchConfigAnnotations({ [DISABLED_ANNOTATION]: 'false' });
+    }
+  }, [effectiveEnabled, unsupportedUnaccepted, patchConfigAnnotations]);
+
+  const handleAccept = React.useCallback(() => {
+    setShowAcceptanceModal(false);
+    patchConfigAnnotations({
+      [UNSUPPORTED_STATUS_ACCEPTED_ANNOTATION]: 'true',
+      [DISABLED_ANNOTATION]: 'false',
+    });
+  }, [patchConfigAnnotations]);
+
+  return (
+    <>
+      <Switch
+        id={`llm-accelerator-config-enabled-toggle-${configName}`}
+        aria-label={`${configName}-enabled-toggle`}
+        data-testid={`llm-accelerator-config-enabled-toggle-${configName}`}
+        isChecked={effectiveEnabled}
+        isDisabled={isToggling}
+        onChange={handleToggle}
+      />
+      {showAcceptanceModal ? (
+        <UnsupportedStatusAcceptanceModal
+          resourceTypeLabel="accelerator configuration"
+          onAccept={handleAccept}
+          onClose={() => setShowAcceptanceModal(false)}
+        />
+      ) : null}
+    </>
+  );
+};
+
+export default LlmAcceleratorConfigEnabledToggle;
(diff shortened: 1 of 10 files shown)


packages/llmd-serving/src/admin/LlmAcceleratorConfigTableRow.tsx
```diff
@@ -29,17 +30,11 @@ const LlmAcceleratorConfigTableRow: React.FC<LlmAcceleratorConfigTableRowProps>
         </ResourceNameTooltip>
         <LabelGroup>
           {preInstalled && <Label data-testid="pre-installed-label">{PreInstalledName}</Label>}
-          {unsupported && <Label data-testid="unsupported-label">Unsupported</Label>}
+          {unsupported && <Label data-testid="unsupported-label">Limited support</Label>}
```
andrewballantyne, line 33: @vconzola I feel we want a splash of colour to make these stand out -- I could accidentally click the wrong row and get or not get a popover looking for or not for the value... it's not super clear to me on wide screens how well this UX works in our current PF table

These labels are almost blending into the table

Image

andrewballantyne: "You now have access to Fast releases, go flip them on..."

Image
👀

packages/llmd-serving/src/admin/LlmAcceleratorConfigTableRow.tsx
andrewballantyne, line 1: @mturley I think if we don't have drag handles, we need to support a sort -- can you sort the Name column -- should be near free with the Table component you're using in the columns...

packages/llmd-serving/src/admin/LlmAcceleratorConfigEnabledToggle.tsx
```diff
@@ -0,0 +1,96 @@
...
+      try {
+        await patchLLMInferenceServiceConfig(config, updatedConfig);
+      } catch (e) {
+        notification.error(
+          'Error updating accelerator configuration',
+          e instanceof Error ? e.message : 'Unknown error',
+        );
+      } finally {
+        setIsToggling(false);
+      }
```
andrewballantyne, line 52: Not sure if this is Claude or not -- but you're not done toggling until the request comes back

You and Yulia both have this weird functionality to "fire the request" and then re-enable the toggle before you get a response... you end up with a flicker of toggle disabled state no matter how long the request takes... and eventually, you refresh.

andrewballantyne (changes requested):
Super close, arguably approve-able so I'm going to do that.

/approve

Lets talk about the comments I made and we can get a lgtm on it after that/actions.

packages/llmd-serving/src/admin/LlmAcceleratorConfigTableRow.tsx
mturley: I changed the microcopy here just so we didn't show "Unsupported" anymore but cleaning up how these labels display is part of https://redhat.atlassian.net/browse/RHOAIENG-75466 (alongside rendering versions there and the "fast" labels). I was planning to use the same yellow outline as our "Tech preview" labels. I should get a version of that together to start a thread about it though.

packages/llmd-serving/src/admin/LlmAcceleratorConfigEnabledToggle.tsx
mturley: Huh? this does await the request before the `finally` block triggers. although hmm it does rely on the watch to update the switch state so there's a small delay there. Not sure the best way to address that since the toggle enabled/disabled state isn't in React, it directly reflects the data over the wire. I could put it in state and keep that synced in an effect so I can do an optimistic update, or I could somehow keep it disabled until we get new data from the watch?

packages/llmd-serving/src/admin/LlmAcceleratorConfigTableRow.tsx
mturley: Good point. I wanted to keep the scope of this limited to the toggle, but this is such a small change and my other upcoming PR for this page is getting large, I'll do that here.

packages/llmd-serving/src/admin/LlmAcceleratorConfigEnabledToggle.tsx
mturley: Looking into whether/how Yulia handled this.

mturley: Looks like the topology configs page still has this problem. Maybe this is a followup issue. Or I could fix it here and make sure we have a followup to fix it in her pages. @YuliaKrimerman is this already on your radar?

mturley: Opened followup issue https://redhat.atlassian.net/browse/RHOAIENG-75952 cc @YuliaKrimerman

vconzola:
@mturley Did you run the microcopy for this past @kaedward?

packages/llmd-serving/src/admin/LlmAcceleratorConfigTableRow.tsx
vconzola: @mturley Why no drag handles?

packages/llmd-serving/src/admin/LlmAcceleratorConfigTableRow.tsx
vconzola: @andrewballantyne Re: label colors, you and your dark mode. JK, I agree. I'd recommend the orange variant, but I'm asking if we have UX guidance on label color selection. Will post a comment when I get an answer.

vconzola: @mturley No UX guidance on label colors. I suggest filled orange variant for Limited support..

mturley:
@vconzola not yet, thanks for the reminder. I don't want to hold up this PR though, I have a microcopy review ticket in the epic.

andrewballantyne (approved):

packages/llmd-serving/src/admin/LlmAcceleratorConfigTableRow.tsx
mturley: It was decided that would be out of scope / “nice to have” for 3.5 because it adds significant complexity - we’d need to figure out how to persist the order somewhere. These get enabled/disabled via metadata on each config, unlike ServingRuntime templates which get enabled by being included in a central list in odhdashboardconfig. The drag handles there just manipulate the order of that list, and we don’t have an equivalent list

packages/llmd-serving/src/admin/LlmAcceleratorConfigTableRow.tsx
mturley: Sounds good. Will update that issue.

PR #8493, opened 2026-07-10
feat(agent-ops): migrate list/detail discovery to Sandbox CRs (RHOAIENG-75465)

danreed-rh:
https://issues.redhat.com/browse/RHOAIENG-75465

## Description

Migrates agent-ops BFF **list** and **detail** discovery from scanning Deployments, StatefulSets, and Jobs to reading `agents.x-k8s.io/v1beta1/sandboxes` exclusively. Deploy already creates Sandbox CRs (#8448); this change aligns discovery so deployed agents appear in list/detail and the post-deploy detail 404 is resolved.

**Key changes:**
- List/detail read Sandbox CRs via the dynamic client; dedupe by name across dual selectors (`opendatahub.io/agent-type=agent` and `openshell.ai/managed-by=openshell`)
- **List** SSAR: `agents.x-k8s.io/sandboxes` **list**; **detail** SSAR: **get**; **stop/start** SSAR: **patch**; **delete** SSAR: **delete**
- **Deploy** RBAC: `agents.x-k8s.io/sandboxes` **create + get** (not full mutation set)
- Detail loads any authorized Sandbox by name (no discovery-label filter — AC4)
- List continues on transient/per-selector errors (Forbidden, NoMatch, timeout) so first-selector results are preserved
- Status normalized to lowercase API enum (`ready`, `running`, `stopped`, `pending`, `failed`)
- `toolConnections` intentionally empty until per-agent MCP linking spec exists; Route enrichment for `externalAgentCardUrl` remains
- `authBridge*` deploy fields accepted but not applied yet (documented as reserved in OpenAPI)
- Cluster role: `agent.kagenti.dev/agentruntimes` → `agents.x-k8s.io/sandboxes` get/list; MCP groups retained for future linking
- MCP dual-read: `mcp.kuadrant.io` + legacy `mcp.kagenti.com` (list helper kept; enrichment path disabled)
- Removed kagenti workload paths; card enrichment Routes-only for runtime responses
- Frontend deploy gate checks sandboxes create **and** get; removed `podCount` from types/OpenAPI

## How Has This Been Tested?

```bash
cd packages/agent-ops/bff && go test ./...
cd packages/agent-ops/frontend && npm run test:jest -- --testPathPatterns=useCanDeployAgent
cd packages/agent-ops && npm run test:contract
```

All BFF unit tests, targeted Jest tests, and contract tests pass locally. No cluster smoke test yet.

## Test Impact

- Sandbox list/detail, OpenShell selector, dedupe, forbidden get, unlabeled detail (AC4)
- List partial-result tests: 2nd selector Forbidden, NoMatch, timeout
- Mutation RBAC: patch/delete SSAR + middleware tests
- Token + internal SAR/self-SAR for list/get/deploy; SAR failure paths
- `serviceAccountFromSpec` (podTemplate + template fallback), enrichment tests
- Mock/demo status aligned to lowercase OpenAPI enum

## Request review criteria:

## Summary by CodeRabbit

* **New Features**
  * Agent runtime listings/details are now driven by Sandbox resources, including richer metadata (display name/description/framework, ports, service info) and condition data.

* **Changes**
  * Runtime/workload statuses are normalized to lowercase.
  * Runtime details no longer include pod counts; agent-card tool connections are deferred and remain empty until linking is defined.
  * Deploy request no longer accepts route creation; deploy/stop/start/delete access is enforced via Sandbox permissions.

* **Bug Fixes**
  * Deploy now returns an unauthorized response when the request identity is missing.

* **Documentation**
  * OpenAPI, API docs, and local/demo documentation updated to match the new Sandbox-driven shapes.


packages/agent-ops/bff/internal/integrations/agents/kubernetes/client_test.go
```diff
@@ -55,26 +56,30 @@ func (c *permissiveK8sClient) CanGetAgentInNamespace(context.Context, *k8s.Reque
...
-		agentRuntimeGVR:              "AgentRuntimeList",
-		openshiftRouteGVR:            "RouteList",
-		mcpServerRegistrationGVRs[0]: "MCPServerRegistrationList",
-		mcpServerRegistrationGVRs[1]: "MCPServerRegistrationList",
+	if c.dynamicClient != nil {
+		return c.dynamicClient, nil
 	}
-	return fakedynamic.NewSimpleDynamicClientWithCustomListKinds(runtime.NewScheme(), gvrToListKind), nil
+	return fakedynamic.NewSimpleDynamicClientWithCustomListKinds(runtime.NewScheme(), defaultGVRListKinds()), nil
 }
```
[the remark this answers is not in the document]
danreed-rh: @coderabbitai Fixed in beda622ee — `permissiveK8sClient.DynamicClient()` now lazily caches the fallback fake client in `c.dynamicClient` so repeated calls share state. Tests still pass.

packages/agent-ops/bff/internal/integrations/agents/metadata.go
```diff
@@ -12,19 +15,27 @@ const (
...
+// OpenShell-managed sandboxes omit opendatahub.io/agent-type but are treated as agents.
+func ResolveAgentResourceType(labels map[string]string) string {
+	if resourceType := strings.TrimSpace(labels[LabelAgentType]); resourceType != "" {
+		return resourceType
+	}
+	if labels[LabelOpenShellManagedBy] == OpenShellManagedByValue {
+		return AgentTypeAgent
+	}
+	return ""
+}
```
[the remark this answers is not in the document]
danreed-rh: @coderabbitai Valid finding — `GetAgent` fetches by name without label selectors, so an unlabeled Sandbox could produce `runtime.type: ""` and violate the required OpenAPI field.

Fixed in beda622ee: after a successful sandbox get, `getAgentDetail` now requires `ResolveAgentResourceType` to return a non-empty type and returns `ErrNotFound` otherwise. Added `TestGetAgentDetailReturnsNotFoundForUnlabeledSandbox`.

packages/agent-ops/bff/internal/integrations/kubernetes/agent_rbac.go
```diff
@@ -60,22 +34,12 @@ type deployResourceCheck struct {
 }
 
 var agentDeployChecks = []deployResourceCheck{
```
DaoDaoNoCode, line 64: The PR description says "SSAR gates use `agents.x-k8s.io/sandboxes` list/get (deploy RBAC unchanged — out of scope)", but `agentDeployChecks`/`CanDeployAgentInNamespace` are rewritten here: the old checks (`serviceaccounts` create/get, `apps/deployments` create, `services` create, `agent.kagenti.dev/agentruntimes` create/get, conditionally `route.openshift.io/routes` create) are replaced with just `sandboxes` create/get, and the `createRoute bool` parameter is dropped from the signature.

This is likely the *correct* fix — `deploy.go` (from #8448) already only ever creates a Sandbox CR, so the new, narrower check aligns authorization with what deploy actually does — but the PR description's "unchanged / out of scope" claim is inaccurate, and the linked Jira ticket (RHOAIENG-75465) doesn't mention this change in its scope either. Worth updating the PR description, and please confirm the frontend's `useCanDeployAgent.ts` (still checking `agent.kagenti.dev/agentruntimes` create) is intentionally left for a follow-up rather than an oversight — as written, the Deploy button's client-side gate no longer matches this server-side check.

packages/agent-ops/bff/internal/integrations/agents/kubernetes/tool_connections.go
```diff
@@ -2,53 +2,39 @@ package kubernetes
...
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/dynamic"
 )
 
-var mcpServerRegistrationGVRs = []schema.GroupVersionResource{
-	{Group: "mcp.kuadrant.io", Version: "v1alpha1", Resource: "mcpserverregistrations"},
-	{Group: "mcp.kagenti.com", Version: "v1alpha1", Resource: "mcpserverregistrations"},
+var mcpServerRegistrationGVR = schema.GroupVersionResource{
```
DaoDaoNoCode, line 17: This drops `mcp.kagenti.com` from the MCP ServerRegistration lookup (previously `mcpServerRegistrationGVRs` queried both `mcp.kuadrant.io` and `mcp.kagenti.com` and merged/deduped results; now only `mcp.kuadrant.io` is queried). `agent_enrichment_rbac.go`'s `CanAccessAgentCardEnrichment` similarly dropped the `mcp.kagenti.com` SAR check (it used to loop over `mcpEnrichmentAPIGroups := []string{"mcp.kuadrant.io", "mcp.kagenti.com"}`).

Neither the PR description nor the Jira ticket mentions narrowing the MCP CRD group — only the kagenti *workload* paths (Deployment/StatefulSet/Job) are called out as removed. If any cluster still registers MCP servers under `mcp.kagenti.com`, this silently drops tool-connection data from the agent card with no error/log (by omission, not by a caught error). Notably, `manifests/modular-architecture/modules-cluster-role.yaml` (not touched by this PR) still grants access to `mcp.kagenti.com` alongside `mcp.kuadrant.io`, so the RBAC manifest and this code are now out of sync either way. Was dropping `mcp.kagenti.com` support intentional?

packages/agent-ops/bff/internal/integrations/agents/kubernetes/client.go
```diff
@@ -66,162 +66,74 @@ func (c *Client) ListAgents(ctx context.Context, namespace string) (*agents.Agen
...
+				c.logger.Debug("skipping sandbox list due to forbidden access",
 					slog.String("namespace", namespace),
-					slog.String("name", statefulSet.Name))
+					slog.String("selector", selector))
 				continue
 			}
-			seen[statefulSet.Name] = struct{}{}
-			service := c.getServiceBestEffort(ctx, namespace, statefulSet.Name)
-			summaries = append(summaries, statefulSetToSummary(statefulSet, service))
+			return nil, fmt.Errorf("failed to list sandboxes in namespace %q with selector %q: %w", namespace, selector, err)
```
DaoDaoNoCode, line 110: If the *second* selector's `listSandboxes` call fails with anything other than `NoMatch`/`NotFound`/`Forbidden` (e.g. a transient 5xx, timeout, or throttling error), this returns `nil, err` and discards the `summaries` already successfully collected from the first selector in this loop. A transient failure on the `openshell.ai/managed-by` selector wipes out perfectly good results already gathered from `opendatahub.io/agent-type=agent` for this namespace.

Consider returning the partial `summaries` with a logged warning instead of aborting the whole call on a per-selector transient error, so one selector's hiccup doesn't blank out the other's valid results.

packages/agent-ops/bff/internal/integrations/agents/kubernetes/mapper.go
```diff
@@ -1,133 +1,182 @@
...
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 )
 
-func agentLabelSelector() string {
-	return fmt.Sprintf("%s=%s", agents.LabelAgentType, agents.AgentTypeAgent)
-}
+const (
+	statusReady    = "Ready"
```
DaoDaoNoCode: `statusReady` is declared but never referenced anywhere in this package (confirmed via `grep -rn statusReady` across `packages/agent-ops/bff` — only this declaration matches; `sandboxPhase` below only uses `statusNotReady`). Looks like a leftover from merging in the deleted `readiness.go` (which also declared `statusReady`). `go vet`/the compiler won't flag unused package-level consts, so this will silently persist — safe to delete.

packages/agent-ops/bff/internal/integrations/agents/kubernetes/mapper.go
```diff
@@ -1,133 +1,182 @@
...
-		Name:         metadata.Name,
-		Namespace:    metadata.Namespace,
-		Description:  mapper.AgentDescription(metadata.Annotations),
-		Status:       jobStatus(job.Status),
-		ResourceType: metadata.Labels[agents.LabelAgentType],
-		WorkloadType: workloadTypeFromLabels(metadata.Labels, agents.WorkloadTypeJob),
-		EndpointURL:  "",
-		CreatedAt:    formatTimestamp(metadata.CreationTimestamp),
-		LastSyncAt:   lastSyncTimestamp(metadata.CreationTimestamp, job.Status),
+func sandboxPhase(sandbox unstructured.Unstructured) string {
```
DaoDaoNoCode, line 100: `sandboxPhase` now passes `status.phase` through verbatim (falling back to `"Not Ready"` only when phase is absent/empty), whereas the deleted `readiness.go` used to normalize Deployment/StatefulSet/Job status into exactly `"Ready"`/`"Not Ready"`/`"Running"`.

The OpenAPI spec's `status` field description (both `packages/agent-ops/api/openapi/agent-ops.yaml` and `bff/openapi/src/agent-ops.yaml`, around the `AgentRuntimeSummary`/`AgentRuntimeDetail` schemas) still says *"Normalized deployment status (backend-specific values mapped to strings)."* — that's no longer accurate given this pass-through. Could you confirm the actual `status.phase` vocabulary the `agents.x-k8s.io/v1beta1` Sandbox controller emits? Only `"Ready"` is exercised by `mapper_test.go` (lines 29, 55, 83); if the controller emits other phase values (e.g. `Provisioning`, `Suspended`, `Terminating`) with different casing than the frontend's expected `ready`/`running`/`stopped`/`pending`/`failed` enum (`agentRuntimeStatus.ts`), agents could render as `Unknown` in the UI.

packages/agent-ops/bff/internal/integrations/agents/types.go
```diff
@@ -53,13 +53,13 @@ type AgentCardSkillObserved struct {
...
 
-// AgentCardExtensionObserved is an A2A extension from AgentRuntime status.card.
+// AgentCardExtensionObserved is an A2A extension from an observed agent card.
 type AgentCardExtensionObserved struct {
 	URI         string
 	Description string
 }
 
-// AgentCardObserved is card data from AgentRuntime.status.card and related status fields.
+// AgentCardObserved is optional agent card metadata enriched from routes and MCP registrations.
```
DaoDaoNoCode: This doc-comment update is a good, honest signal — `AgentCardObserved` really is now only populated from Routes + MCP ServerRegistrations. But the struct below still declares ~15 fields (`Skills`, `Protocol`, `ProviderOrganization`/`ProviderURL`, `AuthBridgeMode`, `AttestedAgentSpiffeID`, `LinkedSkills`, `Streaming`, `PushNotifications`, etc.) that `enrichAgentCard`/`agent_card_enrichment.go` never sets from a real Sandbox CR (only `ExternalAgentCardURL` and `ToolConnections` are populated now). Two things are now out of sync with this comment's new claim:

1. `internal/mapper/agent_card.go` (unchanged by this PR) still maps all of these fields into `models.AgentCardDetail`.
2. The mock/demo client (`mock/demo_client.go`) still fabricates a fully-populated rich card (version, skills, protocols, auth methods, provider, spiffe attestation), and the OpenAPI examples in both `agent-ops.yaml` files still show these fields in example responses.

So `MOCK_AGENT_CLIENT=true` demos functionality a real cluster can never produce today. Is rich A2A card discovery from real Sandbox status intentionally out of scope for this migration? If so, worth trimming the dead fields (or at least annotating them as currently-unreachable) and updating the OpenAPI examples so API consumers/frontend engineers aren't misled.

packages/agent-ops/bff/internal/integrations/kubernetes/agent_rbac_test.go
```diff
@@ -28,62 +29,144 @@ func newRBACTestInternalClient(t *testing.T, reactor func(*authv1.SubjectAccessR
...
+		sarClientset: clientset,
+	}
+
+	_, err := client.CanDeployAgentInNamespace(context.Background(), &RequestIdentity{UserID: "user@example.com"}, "demo-ns")
+	if err == nil {
+		t.Fatal("expected SAR failure error")
+	}
+}
+
+func TestTokenCanDeployAgentInNamespace_UsesSelfSAR(t *testing.T) {
```
DaoDaoNoCode, line 323: `TestTokenCanDeployAgentInNamespace_UsesSelfSAR` is the only test exercising `TokenKubernetesClient` in this file. There's no equivalent Token/self-SAR test for `CanListAgentsInNamespace`/`CanGetAgentInNamespace` (only `InternalKubernetesClient` is tested for those, via `TestCanListAgentsInNamespace_ChecksSandboxesList`/`TestCanGetAgentInNamespace_ChecksSandboxesGet`/`_DeniesWithoutSandboxesGet` above). Confirmed via `go tool cover -func`: `TokenKubernetesClient`'s `CanListAgentsInNamespace`/`CanGetAgentInNamespace` show 0% coverage.

`user_token` (self-SAR) is the default RHOAI/ODH auth method per the BFF README, so the self-SAR gate for the exact list/detail discovery this PR implements is unverified in the mode this actually runs in for RHOAI. The pattern already exists here (`newSelfSARRBACTestTokenClient`) — should be a small addition, mirroring the two `InternalKubernetesClient` tests above.

packages/agent-ops/bff/internal/integrations/agents/kubernetes/client_test.go
```diff
@@ -138,67 +162,183 @@ func TestClient_ListAgentsFromLabeledDeployment(t *testing.T) {
...
+		permissiveK8sClient: *client.k8sClient.(*permissiveK8sClient),
+		dynamic:             dynamicClient,
+	}
 
 	_, err := client.GetAgent(context.Background(), "missing-ns", "missing-agent")
 	require.ErrorIs(t, err, agents.ErrNotFound)
 }
 
-func TestClient_ListNamespacesSkipsWithoutKagentiLabel(t *testing.T) {
+func TestClient_ListAgentsEmptyWhenNoSandboxes(t *testing.T) {
```
DaoDaoNoCode, line 334: The PR's "Test Impact" notes claim expanded coverage for "forbidden get, NoMatch handling", and `client_detail_test.go` does cover this well for `GetAgent` (`TestGetAgentDetailReturnsForbiddenWhenSandboxGetDenied`, `TestGetAgentDetailReturnsNotFoundOnNoMatchError`). But there's no equivalent test here for `ListAgents`/`listAgentSummaries`, even though `client.go`'s `listAgentSummaries` has explicit per-selector Forbidden/NoMatch skip-and-continue logic (lines ~86-94). `TestClient_ListAgentsEmptyWhenNoSandboxes` only covers the plain-empty-result case, not a selector returning Forbidden/NoMatch while another selector succeeds. A reactor-based test analogous to the detail-path ones (on `"list", "sandboxes"`) would close this gap and directly validate the partial-results-preserved behavior.

packages/agent-ops/bff/internal/integrations/agents/kubernetes/agent_card_enrichment_test.go
```diff
@@ -0,0 +1,143 @@
...
+			access:    access,
+			accessErr: accessErr,
+			dynamic:   dynamicClient,
+		},
+		identity: &k8s.RequestIdentity{UserID: "test-user"},
+		logger:   slog.New(slog.NewTextHandler(io.Discard, nil)),
+	}
+}
+
+func TestEnrichAgentCardSkipsWhenAccessCheckFails(t *testing.T) {
```
DaoDaoNoCode, line 63: The deleted `agent_runtime_test.go` had `TestServiceAccountFromSpec`, asserting both the happy path (extracting `serviceAccountName` from `spec.template.spec`) and the nil-spec case. `serviceAccountFromSpec` itself moved verbatim into `agent_card_enrichment.go` (still called at line 16, populating `detail.ServiceAccountName`) but no equivalent test was added to this new file or elsewhere — confirmed via `go tool cover -func`, the extraction path (`agent_card_enrichment.go:89`) is only incidentally hit at 22.2% by other tests (the nil branch). This contradicts the PR description's claim that removed tests were "tied to deleted code paths" — this one covered logic that moved, not logic that was deleted. Worth restoring a small test for `serviceAccountFromSpec` directly.

DaoDaoNoCode (changes requested):
Ran 5 independent focused reviews (correctness, RBAC/security, test coverage, API contract, Go conventions) plus manual validation against the linked Jira ticket ([RHOAIENG-75465](https://redhat.atlassian.net/browse/RHOAIENG-75465)) and the predecessor PR (#8448). Core discovery logic (dedupe, `ResolveAgentResourceType`, NoMatch/Forbidden degradation) is solid and well tested — `go build`/`go vet`/`go test ./...` all pass (195/195). Inline comments below cover validated, line-level findings. Two additional items don't touch files in this diff, so I'm calling them out here instead of inline:

1. **`manifests/modular-architecture/modules-cluster-role.yaml` is stale and not updated by this PR.** It still grants `agent.kagenti.dev/agentruntimes` (get/list) and never grants `agents.x-k8s.io/sandboxes`, which is now the *only* resource `CanListAgentsInNamespace`/`CanGetAgentInNamespace`/`CanDeployAgentInNamespace` check. In `AUTH_METHOD=internal` deployments this ClusterRole is what backs the impersonated SAR calls — every list/detail/deploy check would resolve to `Allowed: false`, silently emptying lists and 404’ing details, with no operator-visible error. This file also still grants `mcp.kagenti.com` (see the `tool_connections.go` comment below).
2. **`packages/agent-ops/frontend/src/app/hooks/useCanDeployAgent.ts` still checks `group: 'agent.kagenti.dev', resource: 'agentruntimes', verb: 'create'`** to decide whether to show/enable the Deploy button, but this PR changes the backend's actual deploy-authorization check to `agents.x-k8s.io/sandboxes` (create+get) — see the `agent_rbac.go` comment below. If cluster RBAC is scoped to match the new backend check without also granting the old `agentruntimes` permission, the Deploy button will be wrongly disabled for users who can actually deploy (and vice versa if only the old grant exists).
3. **Frontend/OpenAPI contract drift on `podCount`**: `packages/agent-ops/frontend/src/app/types/agentRuntimes.ts:81` (`podCount: number`, still required) and `packages/agent-ops/frontend/src/__mocks__/mockAgentRuntime.ts:94` (`podCount: 2`) weren't updated even though this PR removes `podCount` from the Go model and both OpenAPI specs. Confirmed via repo-wide search that nothing currently renders `.podCount`, so this isn't exploitable today, but it's a real, unaddressed type/contract mismatch.

For context: the linked Jira ticket's scope/AC only mentions changing `CanListAgentsInNamespace`/`CanGetAgentInNamespace` to check `agents.x-k8s.io/sandboxes` — it does not mention changing `CanDeployAgentInNamespace` or touching the frontend. PR #8448 (the predecessor) explicitly deferred "RBAC checks" to "the discovery PR" (i.e., this one), so updating deploy RBAC here is reasonable, but it should be described accurately and the frontend/manifest counterparts should be updated in the same change (or explicitly tracked as follow-ups) rather than left silently inconsistent.

No functional regressions were found in the core Sandbox-discovery logic itself — requesting changes mainly for the RBAC-consistency gaps (manifest + frontend hook) and the two client.go/mapper.go correctness nits inline, which seem cheap to fix before merge.

packages/agent-ops/bff/internal/integrations/kubernetes/agent_rbac.go
danreed-rh: Resolved in fe76a9aa4. PR description updated: deploy RBAC is sandboxes create+get only. Mutations now use separate SSAR checks — patch for stop/start, delete for delete.

packages/agent-ops/bff/internal/integrations/agents/kubernetes/tool_connections.go
danreed-rh: Resolved in fe76a9aa4. Dual-read restored (mcp.kuadrant.io + legacy mcp.kagenti.com). Runtime enrichment no longer populates toolConnections until per-agent linking spec exists — avoids misleading namespace-wide data.

packages/agent-ops/bff/internal/integrations/agents/kubernetes/client.go
danreed-rh: Resolved in fe76a9aa4. isRetryableListError() continues on transient 2nd-selector failures; tests for Forbidden, NoMatch, and timeout added.

packages/agent-ops/bff/internal/integrations/agents/kubernetes/mapper.go
danreed-rh: Resolved in fe76a9aa4. statusReady is used by normalizeSandboxPhase() for case-insensitive mapping.

packages/agent-ops/bff/internal/integrations/agents/kubernetes/mapper.go
danreed-rh: Resolved in fe76a9aa4. normalizeSandboxPhase() maps to lowercase API enum; OpenAPI and mock responses aligned.

packages/agent-ops/bff/internal/integrations/agents/types.go
danreed-rh: Resolved in fe76a9aa4. types.go documents enrichment scope; toolConnections omitted until per-agent MCP linking spec. Route external URL enrichment remains.

packages/agent-ops/bff/internal/integrations/kubernetes/agent_rbac_test.go
danreed-rh: Resolved in fe76a9aa4. Added token self-SAR tests for list/get plus failure-path tests.

packages/agent-ops/bff/internal/integrations/agents/kubernetes/client_test.go
danreed-rh: Resolved in fe76a9aa4. Added ListAgents partial-result tests for 2nd-selector Forbidden, NoMatch, and timeout.

packages/agent-ops/bff/internal/integrations/agents/kubernetes/agent_card_enrichment_test.go
danreed-rh: Resolved in fe76a9aa4. TestServiceAccountFromSpec restored; serviceAccountFromSpec reads podTemplate then template fallback.

packages/agent-ops/bff/internal/integrations/kubernetes/agent_rbac.go
danreed-rh: Good catch on the PR description wording — updated in fe76a9aa4.

Deploy RBAC is intentionally narrowed to `agents.x-k8s.io/sandboxes` create+get because `deploy.go` only creates Sandbox CRs today; the old Deployment/Service/AgentRuntime checks were stale.

Frontend is aligned in the same commit: `useCanDeployAgent.ts` now SSAR-checks sandboxes create+get (see `useCanDeployAgent.spec.ts`). PR description updated to reflect this scope change.

packages/agent-ops/bff/internal/integrations/agents/types.go
danreed-rh: You're right that the struct/OpenAPI/mock still overpromise relative to what a real cluster returns today.

**In scope for this migration (fe76a9aa4):**
- Doc comment on `AgentCardObserved` reflects actual enrichment: external route URL only.
- `toolConnections` intentionally empty until per-agent MCP linking is defined (no namespace-wide MCP list in API responses).

**Acknowledged follow-up (not blocking this PR):**
- Trim or annotate unreachable rich-card fields in `types.go` / `agent_card.go`.
- Align mock demo client and OpenAPI examples so `MOCK_AGENT_CLIENT=true` doesn't imply capabilities a real Sandbox can't produce yet.

This will get addressed as we iterate and move towards OpenShell, unlikely to be resolved for 3.5

packages/agent-ops/bff/internal/integrations/agents/kubernetes/tool_connections.go
danreed-rh: Dual-read is restored (`mcp.kuadrant.io` + legacy `mcp.kagenti.com`) in fe76a9aa4.

Runtime list/detail responses intentionally omit `toolConnections` until we have a per-agent MCP linking spec — namespace-wide ServerRegistration listing was misleading. README and OpenAPI descriptions updated accordingly; enrichment code still dual-reads for future use.

packages/agent-ops/bff/internal/integrations/kubernetes/agent_rbac_test.go
danreed-rh: Added in fe76a9aa4, mirroring the InternalKubernetesClient pattern:

- `TestTokenCanListAgentsInNamespace_UsesSelfSAR`
- `TestTokenCanGetAgentInNamespace_UsesSelfSAR`
- Failure-path tests for denied self-SAR

Token self-SAR is now covered for the list/get gates this PR implements.

packages/agent-ops/bff/internal/integrations/agents/kubernetes/client.go
danreed-rh: Addressed in two commits:

- **fe76a9aa4** — `isRetryableListError()` continues the loop on transient 2nd-selector failures so partial results from the first selector are preserved. Tests for Forbidden, NoMatch, and timeout on the 2nd selector.
- **0d3c55650** — When *every* selector fails Forbidden or transient (none completes), we now return an error instead of 200 + empty list (`TestClient_ListAgentsReturnsForbiddenWhenAllSelectorsForbidden`).

packages/agent-ops/bff/internal/integrations/agents/kubernetes/client_test.go
danreed-rh: Added in fe76a9aa4:

- `TestClient_ListAgentsPreservesResultsWhenSecondSelectorForbidden`
- `TestClient_ListAgentsPreservesResultsWhenSecondSelectorNoMatch`
- `TestClient_ListAgentsPreservesResultsWhenSecondSelectorTimesOut`

**0d3c55650** adds `TestClient_ListAgentsReturnsForbiddenWhenAllSelectorsForbidden` for the all-selectors-fail path.

packages/agent-ops/bff/internal/integrations/agents/kubernetes/mapper.go
danreed-rh: Confirmed — `statusReady` is referenced by `normalizeSandboxPhase()` for case-insensitive phase mapping (fe76a9aa4). No longer dead code.

packages/agent-ops/bff/internal/integrations/agents/kubernetes/agent_card_enrichment_test.go
danreed-rh: Restored in fe76a9aa4 as `TestServiceAccountFromSpec` — covers happy path (`podTemplate` and legacy `template` fallback) and nil-spec. `serviceAccountFromSpec` still populates `detail.ServiceAccountName` on enrichment.

packages/agent-ops/bff/internal/integrations/agents/kubernetes/mapper.go
danreed-rh: Agreed the old pass-through was wrong for the API contract. **fe76a9aa4** adds `normalizeSandboxPhase()` mapping Sandbox `status.phase` to the lowercase OpenAPI enum (`ready` / `running` / `stopped` / `pending` / `failed`). OpenAPI descriptions and mock/contract responses updated to match.

Full conditions-based phase mapping from Sandbox status is a follow-up once we confirm the controller's condition vocabulary in-cluster.

danreed-rh:

Gkrumbach07:
## Scope alignment for 3.5 — feedback from spec review

The 3.5 scope has been cut down to **discovery-only** (no deploy button, no agent card discovery, bare detail page). I reviewed this PR against the [Starter Kit Deployment Spec](https://docs.google.com/document/d/1iiSzz32I00d879HQTtHzwLd4ZzGmN4A5iKr1tG2mh88/edit?tab=t.n2394jvqii1d) and have three items:

### 1. Annotation keys don't match the spec

The spec defines these annotations for starter kit discovery:

| Annotation | Purpose |
|---|---|
| `openshift.io/display-name` | Display name (same pattern as other dashboard resources) |
| `openshift.io/description` | Description |
| `opendatahub.io/agent-framework` | Framework (langgraph, crewai, llamaindex, a2a) |

The PR currently reads `opendatahub.io/agent-description` for description and has no handling for `openshift.io/display-name`. The framework annotation matches. Please update `metadata.go` constants and the mapper to read from the spec-defined keys.

### 2. Container ports from podTemplate are not exposed

The DTO hardcodes port 8080 when building the service URL from `status.serviceFQDN`. It never reads `spec.podTemplate.spec.containers[].ports`.

The frontend needs the actual port list so it can show:
- The cluster service URL per port (for port forwarding / `oc port-forward`)
- SSH ports if present
- Any custom agent ports

Please extract ports from `spec.podTemplate.spec.containers[].ports` and include them in the DTO.

### 3. Scope down the DTO for 3.5

Since 3.5 is discovery-only with a bare detail page, the DTO can be trimmed. What we actually need:

**List:** name, namespace, display name, description, framework, status, service FQDN, ports

**Detail:** same as list + conditions, full label/annotation maps, container image

What we can drop for now:
- Agent card enrichment (no card discovery in 3.5)
- SPIFFE ID construction
- Any deploy-wizard-specific fields

The custom DTO approach is correct (matches the pattern in MLflow, Gen AI, and Model Registry BFFs). Just scope it to what 3.5 needs.

packages/agent-ops/api/openapi/agent-ops.yaml
```diff
@@ -348,17 +353,43 @@ components:
...
+        ports:
+          type: array
+          description: Container ports from spec.podTemplate, mapped to in-cluster service URLs.
+          maxItems: 20
+          items:
+            $ref: "#/components/schemas/AgentServiceEndpoint"
         endpointUrl:
           type: string
           format: uri
+          description: Primary in-cluster endpoint URL derived from service FQDN and ports.
```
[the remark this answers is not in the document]
danreed-rh: Yep, good catch — empty string was a real case when FQDN/ports aren’t ready yet, so requiring `format: uri` was wrong.

Fixed in 31fb66684: `endpointUrl` is no longer required in OpenAPI (both copies), and the Go model uses `omitempty` so we just omit it instead of returning `""`.

Gkrumbach07:

danreed-rh:
@Gkrumbach07 thanks for the writeup — went through and aligned the branch with the 3.5 discovery scope.

Quick summary:
1. Annotations now use `openshift.io/display-name` + `openshift.io/description` (strict), keeping `opendatahub.io/agent-framework`
2. Ports come from `spec.podTemplate` now instead of the hardcoded 8080 fallback, and we surface `serviceFqdn` too
3. List/detail DTO trimmed for discovery; agent card enrichment is off for this PR

Deploy UI hiding is still a separate frontend follow-up. Landed across `6dd45b50b` / `31fb66684`.

Gkrumbach07:
@danreed-rh One more change: let's drop the `opendatahub.io/agent-type=agent` label selector entirely. Discovery should use `openshell.ai/managed-by=openshell` only.

In `mapper.go`, `agentLabelSelectors()` should return just:
```go
func agentLabelSelectors() []string {
    return []string{
        fmt.Sprintf("%s=%s", agents.LabelOpenShellManagedBy, agents.OpenShellManagedByValue),
    }
}
```

Also clean up `ResolveAgentResourceType` in `metadata.go` — it can just check for the OpenShell label and return `"agent"`. No need to check `opendatahub.io/agent-type` anymore.

danreed-rh:
@gagenell Done — list discovery is OpenShell-only now.

- `agentLabelSelectors()` returns only `openshell.ai/managed-by=openshell`
- `ResolveAgentResourceType` checks that label and returns `"agent"` (no more `opendatahub.io/agent-type`)
- Deploy still stamps the OpenShell label (plus agent-type for any other consumers) so dashboard-created sandboxes stay listable
- Docs/OpenAPI updated to match

Gkrumbach07:

Gkrumbach07 (approved):

PR #8524, opened 2026-07-13
Wire topology/routing wizard selections into LLMInferenceService baseRefs

YuliaKrimerman:
https://issues.redhat.com/browse/RHOAIENG-70588

## Description

Wire topology and routing wizard field selections into the LLMInferenceService CR during deploy, and support round-tripping through the edit flow.

**Apply (deploy):**
- When a custom topology config is selected, its name is added to `spec.baseRefs[]`
- When a routing config is selected, its name is added to `spec.baseRefs[]`
- Ordering is enforced: topology baseRef before routing baseRef (later overrides earlier in KServe merge)
- Single node with "Single node (default)" adds no topology baseRef
- "Default optimized routing" adds no routing baseRef
- Uses `WizardFieldApplyExtension` pattern (consistent with gateway)

**Extract (edit):**
- Three dashboard-owned annotations store deploy-time selections for round-tripping:
  - `opendatahub.io/topology-type` — the selected `TopologyType` enum value
  - `opendatahub.io/topology-config-ref` — the topology config name
  - `opendatahub.io/routing-config-ref` — the routing config name
- `WizardFieldExtractorExtension` reads annotations; component lazy-resolves config name to full object from external data
- Uses `configRef` field on `CustomTopologyConfigFieldData` / `AdvancedRoutingFieldData` for deferred resolution

**Known gap (documented, out of scope for 3.5 TP):** Runtime args/env are NOT propagated to `spec.worker` or `spec.prefill` containers. Users who need custom args in those blocks must put them in admin-created topology configs or use YAML edit mode.

## How Has This Been Tested?

**Unit tests:** 23 new tests covering all apply/extract paths, baseRefs ordering, immutability, edge cases (no config, duplicate refs, config replacement). All 177 llmd-serving tests pass.

**Manual E2E on cluster (zaffre):**
1. Deploy with multi-node topology config → verified `spec.baseRefs` contains topology config name
The deployment spec:
Screenshot 2026-07-13 at 11 48 25 AM
The output ( verification on the cluster) :
Screenshot 2026-07-13 at 12 02 03 PM

2. Deploy with single node + routing config → verified routing baseRef present, no topology baseRef
The deployment spec:

Screenshot 2026-07-13 at 12 04 26 PM
The output ( verification on the cluster):

Screenshot 2026-07-13 at 12 07 19 PM

3. Deploy with both topology + routing → verified correct ordering in `spec.baseRefs` (topology first)
The deployment spec:

Screenshot 2026-07-13 at 12 09 59 PM
The output ( verification on the cluster):

Screenshot 2026-07-13 at 2 44 51 PM

7. Deploy with single node default → verified empty `spec.baseRefs`, only `topology-type` annotation
 The deployment spec:

Screenshot 2026-07-13 at 12 51 40 PM
The output ( verification on the cluster):

Screenshot 2026-07-13 at 12 52 04 PM

9. Edit existing deployment → verified wizard pre-populates topology/routing selections from annotations, re-submit updates CR correctly
 The deployment spec:

Screenshot 2026-07-13 at 12 53 16 PM
Screenshot 2026-07-13 at 12 53 28 PM
Screenshot 2026-07-13 at 12 54 30 PM

The output ( verification on the cluster):
Screenshot 2026-07-13 at 12 55 03 PM

Tested with `?devFeatureFlags=true,llmdTopologyConfigs=true`

**Cluster verification commands used:**
```bash
oc get llminferenceservice <name> -n <ns> -o jsonpath='{.spec.baseRefs}'
oc get llminferenceservice <name> -n <ns> -o jsonpath='{.metadata.annotations.opendatahub\.io/topology-type}'
oc get llminferenceservice <name> -n <ns> -o jsonpath='{.metadata.annotations.opendatahub\.io/topology-config-ref}'
oc get llminferenceservice <name> -n <ns> -o jsonpath='{.metadata.annotations.opendatahub\.io/routing-config-ref}'
```

**Test Impact **
Added: packages/llmd-serving/src/deployments/__tests__/topology.spec.ts — 23 unit tests for apply/extract functions
Added: packages/cypress/cypress/tests/e2e/dataScienceProjects/models/testLlmdTopologyDeploy.cy.ts — E2E smoke test for topology deploy flow (@LLMDServingCI)
Added: packages/cypress/cypress/tests/e2e/settings/llmdTopologyConfigurations/testLlmdTopologyConfigurations.cy.ts — Admin topology settings E2E (table visibility, config toggle, wizard field visibility)
Modified: packages/cypress/cypress/tests/e2e/dataScienceProjects/models/testDeployLLMDServing.cy.ts — Added topology/routing field selections and baseRefs assertion to existing deploy test
**Request review criteria:**

 The developer has manually tested the changes and verified that the changes work
 Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
 The developer has added tests or explained why testing cannot be added (unit or cypress tests for related changes)
 The code follows our [Best Practices](vscode-file://vscode-app/docs/best-practices.md) (React coding standards, PatternFly usage, performance considerations)
If you have UI changes:

 Included any necessary screenshots or gifs if it was a UI change.
 Included tags to the UX team if it was a UI/UX change.

 The developer has tested their solution on a cluster by using the image produced by the PR to main

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring multi-node LLM inference deployments.
  * Added topology type, custom topology configuration, and advanced routing options to the deployment wizard.
  * Added administration tools to create, enable, and manage topology configurations.
  * Deployment settings now restore previously selected topology and routing configurations.

* **Tests**
  * Added coverage for topology and routing configuration behavior.
  * Expanded end-to-end tests to verify multi-node deployment configuration and successful topology association.


packages/cypress/cypress/tests/e2e/dataScienceProjects/models/testDeployLLMDServing.cy.ts
```diff
@@ -145,6 +158,12 @@ describe('A user can deploy an LLMD model', () => {
       modelServingWizard.findYAMLEditorEmptyState().should('be.visible');
       modelServingWizard.findYAMLViewerToggle(YAMLViewerToggleOption.FORM).should('exist').click();
       modelServingWizard.selectDeploymentMethodByKey(deploymentMethod);
+
+      cy.step('Select topology type and configuration');
```
sridarna, line 165: We don't include cy.test-id in cy.findByTestId file
Please move them to corresponding pages.

packages/cypress/cypress/tests/e2e/dataScienceProjects/models/testDeployLLMDServing.cy.ts
```diff
@@ -171,6 +190,7 @@ describe('A user can deploy an LLMD model', () => {
       cy.step('Select Advanced settings');
       // LLMD models support token authentication and it is checked by default
       modelServingWizard.findTokenAuthenticationCheckbox().should('be.checked');
+      cy.findByTestId('routing-config-select').should('exist');
```
sridarna: This one should also move to pages folder

packages/cypress/cypress/tests/e2e/dataScienceProjects/models/testDeployLLMDServing.cy.ts
```diff
@@ -185,6 +205,11 @@ describe('A user can deploy an LLMD model', () => {
         checkLLMInferenceServiceState(resourceName, projectName, { checkReady: true });
       });
 
+      cy.step('Verify topology baseRef on LLMInferenceService CR');
+      cy.then(() => {
+        checkLLMInferenceServiceBaseRefs(resourceName, projectName, ['e2e-multi-node-topology']);
```
sridarna: create a variable for e2e-multi-node-topology in fixture file

packages/cypress/cypress/tests/e2e/dataScienceProjects/models/testLlmdTopologyDeploy.cy.ts
```diff
@@ -0,0 +1,101 @@
+import { deleteOpenShiftProject } from '../../../../utils/oc_commands/project';
```
sridarna: I think we are already testing it in packages/cypress/cypress/tests/e2e/dataScienceProjects/models/testDeployLLMDServing.cy.ts file.
Do we need this?

packages/cypress/cypress/tests/e2e/settings/llmdTopologyConfigurations/testLlmdTopologyConfigurations.cy.ts
```diff
@@ -0,0 +1,119 @@
...
+  cleanupLLMInferenceServiceConfig,
+} from '../../../../utils/oc_commands/llmInferenceServiceConfig';
+import { projectListPage } from '../../../../pages/projects';
+import { modelServingGlobal, modelServingWizard } from '../../../../pages/modelServing';
+import { ModelLocationSelectOption, ModelTypeLabel } from '../../../../utils/modelServingConstants';
+import { createCleanProject } from '../../../../utils/projectChecker';
+import { deleteOpenShiftProject } from '../../../../utils/oc_commands/project';
+import { generateTestUUID } from '../../../../utils/uuidGenerator';
+
+const topoConfigName = 'e2e-multi-node-topology';
```
sridarna: it would be better to move topoConfigName, projectName to fixture file.
Comparing the other e2e

packages/cypress/cypress/tests/e2e/settings/llmdTopologyConfigurations/testLlmdTopologyConfigurations.cy.ts
```diff
@@ -0,0 +1,119 @@
...
+const topoConfigFixture = 'resources/modelServing/llmd-topology-config.yaml';
+const uuid = generateTestUUID();
+const projectName = `topo-admin-${uuid}`;
+
+describe('LLMD Topology Configurations - Admin Settings', () => {
+  retryableBefore(() => {
+    createCleanLLMInferenceServiceConfig(topoConfigName, topoConfigFixture);
```
sridarna: I think this e2e should also create the Topology config from UI. As we want to validate the creation is successfully from UI.

packages/cypress/cypress/tests/e2e/settings/llmdTopologyConfigurations/testLlmdTopologyConfigurations.cy.ts
```diff
@@ -0,0 +1,119 @@
...
+  });
+
+  it(
+    'Admin can view topology configurations table',
+    { tags: ['@Smoke', '@Dashboard', '@NonConcurrent', '@LLMDServingCI'] },
+    () => {
+      cy.step('Log in with topology configs feature flag');
+      cy.visitWithLogin(
+        '/?devFeatureFlags=true,llmdTopologyConfigs=true',
+        HTPASSWD_CLUSTER_ADMIN_USER,
```
sridarna, line 53: replace HTPASSWD_CLUSTER_ADMIN_USER with LDAP_ADMIN_USER. As  E2E should always use LDAP_ADMIN_USER to validate.

sridarna: We recently came across this change. In future other e2e tests will also be updated.

packages/cypress/cypress/tests/e2e/settings/llmdTopologyConfigurations/testLlmdTopologyConfigurations.cy.ts
```diff
@@ -0,0 +1,119 @@
...
+      );
+
+      cy.step('Navigate to topology configurations settings');
+      llmdTopologySettingsPage.navigate();
+      llmdTopologySettingsPage.findAppTitle().should('contain', 'llm-d topology configurations');
```
sridarna: We are trying to avoid sting validations in e2e. 
 I think we can remove this validation if covered in mock tests. or move llm-d topology configurations to fixtures file

packages/cypress/cypress/tests/e2e/settings/llmdTopologyConfigurations/testLlmdTopologyConfigurations.cy.ts
```diff
@@ -0,0 +1,119 @@
...
+      cy.step('Navigate to project and open deploy wizard');
+      projectListPage.navigate();
+      projectListPage.filterProjectByName(projectName);
+      projectListPage.findProjectLink(projectName).click();
+      modelServingGlobal.selectSingleServingModelButtonIfExists();
+      modelServingGlobal.findDeployModelButton().click();
+
+      cy.step('Fill model details and advance to deployment step');
+      modelServingWizard.findModelLocationSelectOption(ModelLocationSelectOption.URI).click();
+      modelServingWizard.findUrilocationInput().clear().type('hf://facebook/opt-125m');
```
sridarna, line 73: move hf://facebook/opt-125m to fixtures

packages/cypress/cypress/tests/e2e/settings/llmdTopologyConfigurations/testLlmdTopologyConfigurations.cy.ts
```diff
@@ -0,0 +1,119 @@
...
+      modelServingWizard.findModelTypeSelectOption(ModelTypeLabel.GENERATIVE).click();
+      modelServingWizard.findNextButton().should('be.enabled').click();
+
+      cy.step('Select llm-d deployment method');
+      modelServingWizard.selectDeploymentMethodByKey('llm-inference-service-llmd');
```
sridarna, line 79: move llm-inference-service-llmd to fixture file

packages/cypress/cypress/tests/e2e/settings/llmdTopologyConfigurations/testLlmdTopologyConfigurations.cy.ts
```diff
@@ -0,0 +1,119 @@
...
+
+      cy.step('Verify topology type dropdown is visible');
+      cy.findByTestId('topology-type-select').should('exist');
```
sridarna: need to remove the testID from cy.ts file

packages/cypress/cypress/tests/e2e/settings/llmdTopologyConfigurations/testLlmdTopologyConfigurations.cy.ts
```diff
@@ -0,0 +1,119 @@
...
+    createCleanLLMInferenceServiceConfig(topoConfigName, topoConfigFixture);
+    createCleanProject(projectName);
+  });
+
+  after(() => {
+    cleanupLLMInferenceServiceConfig(topoConfigName);
+    deleteOpenShiftProject(projectName, { wait: true, ignoreNotFound: true, timeout: 300000 });
+  });
+
+  it(
```
sridarna, line 46: All the sub tests should be combined to one. As we want to test the complete e2e flow.
Also note, e2e sub tests should be independent of each other state.

packages/cypress/cypress/utils/oc_commands/llmInferenceServiceConfig.ts
```diff
@@ -53,6 +53,55 @@ export const createCleanLLMInferenceServiceConfig = (
...
+};
+
+/**
+ * Verifies that an LLMInferenceService has the expected baseRef names.
+ *
+ * @param serviceName - The `metadata.name` of the LLMInferenceService.
+ * @param namespace - The namespace the service is deployed in.
+ * @param expectedBaseRefs - Array of baseRef names that must appear in `spec.baseRefs`.
+ */
+export const checkLLMInferenceServiceBaseRefs = (
```
sridarna, line 75: test this multiple times. If it failing with BaseRefs not updated. We could improve this function with retries.

packages/llmd-serving/src/wizardFields/CustomTopologyConfigField.tsx
```diff
@@ -91,6 +92,20 @@ const CustomTopologyConfigFieldComponent: CustomTopologyConfigFieldType['compone
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [isLoaded, filteredConfigs.length, topologyType]);
 
+  // Resolve configRef from extractor (edit flow) once external data loads
```
mturley, line 103: The auto-select effect (line 88) and this configRef resolution effect both fire when `isLoaded` flips to true during the edit flow. At that point `existingSelection` is undefined and `configRef` is set, so both effects call `onChange`. This works today because React processes effects in declaration order and batches the updates (effect #2 wins), but it's relying on ordering rather than explicit guards.

Adding `&& !configRef` to the auto-select condition (line 89) would make it explicit that auto-select is only for new deployments:
```ts
if (isLoaded && !isSingleNode && !existingSelection && !configRef && filteredConfigs.length > 0) {
```
That way the edit flow is never in a race with auto-select regardless of effect ordering or batching behavior.

packages/llmd-serving/src/wizardFields/AdvancedRoutingField.tsx
```diff
@@ -108,6 +109,20 @@ const AdvancedRoutingFieldComponent: AdvancedRoutingFieldType['component'] = ({
     [routerConfigs, topologyType],
   );
 
+  // Resolve configRef from extractor (edit flow) once external data loads
```
mturley, line 112: `CustomTopologyConfigField` resets when the topology type changes (via `shouldResetOnDependencyChange`), but `AdvancedRoutingField` doesn't, even though it filters compatible configs by topology type via `getCompatibleRouterConfigs`.

If a user selects a routing config that has a restricted `supportedTopologies` list, then changes the topology type to one not in that list, the selected config would remain in the field state but disappear from the dropdown options — the user would see a stale selection they can't change without re-selecting "Default optimized routing" first.

This may not be a problem today if no routing configs use `supportedTopologies` restrictions, but it's a latent gap. Worth considering whether this field should also reset on topology type change, or at minimum clear the selection when the selected config is no longer in `compatibleConfigs`.

Edit: Ah, this is also pre-existing. I'll open a followup.

packages/llmd-serving/src/wizardFields/CustomTopologyConfigField.tsx
```diff
@@ -36,6 +36,7 @@ type CustomTopologyConfigDependencies = {
 
 export type CustomTopologyConfigFieldData = {
   selectedConfig?: LLMInferenceServiceConfigKind;
+  configRef?: string;
```
mturley, line 39: This field's `FormGroup` has `isRequired` (line 149) but that's visual-only (renders an asterisk). There's no `validationSchema` in `reducerFunctions` (line 235), so the wizard validation at `useDeploymentWizardValidation.tsx:86-89` doesn't include this field — `isModelDeploymentStepValid` passes regardless, and the user can advance through all steps and submit with `selectedConfig` undefined.

This is reachable in two flows:

**Create mode — no configs available (pre-existing, not introduced by this PR):** User selects a multi-node topology type but no configs exist for that type. The dropdown is disabled with a warning ("Contact your administrator"), but the wizard doesn't block submission. The deployment gets created with a topology-type annotation but no topology baseRef — which may not be a valid state for KServe.

**Edit mode — orphaned config (new path introduced by this PR):** The topology config referenced by `opendatahub.io/topology-config-ref` has been deleted or renamed. The resolution effect (line 102) silently finds no match, `selectedConfig` stays undefined, and submitting strips the old annotation + baseRef without the user realizing.

Not necessarily a blocker for this PR — we could open a follow-up bug to cover both cases. If you'd prefer to address it here, one approach: make `selectedConfig` a union of `LLMInferenceServiceConfigKind | 'default'` so that "Single node (default)" sets `selectedConfig: 'default'` instead of `undefined`. Then `selectedConfig` is always truthy when the dropdown has an active selection, and a simple `validationSchema` enforces it:

```ts
// In CustomTopologyConfigFieldData (line 37):
selectedConfig?: LLMInferenceServiceConfigKind | 'default';

// In the onChange for SINGLE_NODE_DEFAULT_KEY (line 162):
onChange({ selectedConfig: 'default' });

// In reducerFunctions (line 235), add:
validationSchema: z.object({
  selectedConfig: z.union([z.object({}).passthrough(), z.literal('default')]),
  configRef: z.string().optional(),
}),

// In applyTopologyConfig (topology.ts line 47), guard:
if (config && config !== 'default') { ... }
```

This covers all cases uniformly: single-node default passes, multi-node with a config passes, and both the create-mode (no configs) and edit-mode (orphaned ref) gaps fail validation, along with any other weird case that results in no selected config.

packages/llmd-serving/src/wizardFields/AdvancedRoutingField.tsx
```diff
@@ -108,6 +109,20 @@ const AdvancedRoutingFieldComponent: AdvancedRoutingFieldType['component'] = ({
...
+  const configRef = value?.configRef;
+  const existingSelection = value?.selectedConfig;
+  React.useEffect(() => {
+    if (!configRef || existingSelection || !isLoaded) {
+      return;
+    }
+    const resolved = (routerConfigs ?? []).find((c) => c.metadata.name === configRef);
```
mturley, line 119: The configRef resolution here searches the unfiltered `routerConfigs` list, but the dropdown displays `compatibleConfigs` which is filtered by topology type. If the routing config's `supportedTopologies` annotation was changed after the original deploy to exclude the current topology type, the configRef would resolve successfully (setting `selectedConfig`), but the resolved config wouldn't appear in the dropdown options.

The user would see a selected value that isn't in the dropdown list. The existing-selection fallback at line 166 (`config ?? value?.selectedConfig`) means selecting a different option and switching back wouldn't recover it either.

Similar to the topology orphan case — consider showing stale/incompatible selections as a disabled option with a warning, so the user understands what was previously configured and can make an informed choice.

This is an edge case (config compatibility changed after deploy) so it's also fine as a follow-up rather than a blocker.

packages/llmd-serving/src/wizardFields/CustomTopologyConfigField.tsx
```diff
@@ -91,6 +92,20 @@ const CustomTopologyConfigFieldComponent: CustomTopologyConfigFieldType['compone
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [isLoaded, filteredConfigs.length, topologyType]);
```
mturley, line 101: I noticed a pre-existing issue (not introduced by this PR) when changing the topology type from "Single node" (the default) to another type like "Single node disaggregated" — the auto-select effect (line 88) doesn't result in a config being selected, even though a compatible config is visible in the dropdown options.

I confirmed with a debugger that `onChange({ selectedConfig: filteredConfigs[0] })` at line 90 **does execute**, but the selection doesn't persist — the dropdown still shows "Select configuration" as the placeholder. This points to the auto-select's `onChange` being overwritten by the wizard framework's field reset. The `shouldResetOnDependencyChange` (line 247) triggers `clearFieldData` + `initFieldData` dispatches in `useDeploymentWizardReducer.tsx:177-187`, and these appear to run after the auto-select, wiping out the selection.

The auto-select would need to run after the reset completes, not during the same render cycle. One approach: instead of an effect in the component, handle auto-selection in `getInitialFieldData` (line 243), which is called by the `initFieldData` dispatch during reset. If `getInitialFieldData` received the external data and dependencies, it could return `{ selectedConfig: filteredConfigs[0] }` directly instead of `{ selectedConfig: undefined }`. The `getInitialFieldData` signature already supports `externalData` and `dependencies` as optional parameters (per the `WizardField` type at `types.ts:248-251`), so this may just be a matter of using them.

I'll open a follow-up bug for this since it predates this PR.

mturley (changes requested):
Some things I noticed from back and forth with claude and some targeted testing. Still doing more thorough testing.

packages/llmd-serving/src/wizardFields/CustomTopologyConfigField.tsx
YuliaKrimerman: yes, will address

packages/llmd-serving/src/wizardFields/AdvancedRoutingField.tsx
YuliaKrimerman: yep, no routing configs currently use supportedTopologies restrictions. I tyhink it's viable to address if this will change in the future

packages/llmd-serving/src/wizardFields/CustomTopologyConfigField.tsx
YuliaKrimerman: this will require some refactoring, I think the defaults union approach is clean for now - We can do this as a follow-up

packages/llmd-serving/src/wizardFields/AdvancedRoutingField.tsx
YuliaKrimerman: agree to be a follow up

packages/llmd-serving/src/wizardFields/CustomTopologyConfigField.tsx
YuliaKrimerman: haha - a lot of follow ups -but yes this is pre-existing it's related to https://redhat.atlassian.net/browse/RHOAIENG-73694  as well , will need to fix it for these cases

packages/llmd-serving/src/wizardFields/AdvancedRoutingField.tsx
mturley: Covered by followup, currently in your epic but we can remove it: https://redhat.atlassian.net/browse/RHOAIENG-76494 (same issue as another comment)

packages/llmd-serving/src/wizardFields/AdvancedRoutingField.tsx
mturley: Covered by followup, currently in your epic but we can remove it: https://redhat.atlassian.net/browse/RHOAIENG-76494 (same issue as another comment)

packages/llmd-serving/src/wizardFields/CustomTopologyConfigField.tsx
mturley: Covered by followup, currently in your epic but we can remove it: https://redhat.atlassian.net/browse/RHOAIENG-76492

I'd consider this one major just because it's missing validation, but maybe we could do it in a simpler way that doesn't require so much refactoring. Claude got carried away on the suggested solution 😅

packages/llmd-serving/src/wizardFields/CustomTopologyConfigField.tsx
mturley: Covered by followup, currently in your epic but we can remove it: https://redhat.atlassian.net/browse/RHOAIENG-76493

YuliaKrimerman:
@sridarna 
Screenshot 2026-07-14 at 12 28 17 PM

mturley:
/approve
/lgtm

Awesome! @sridarna I think your comments have been addressed and I want to approve to get this into a cluster, but if you have followup comments please let us know

PR #8529, opened 2026-07-13
fix(automl): block non-ASCII column names in AutoML runs

jkyaw:
https://redhat.atlassian.net/browse/RHOAIENG-76345

## Description
Creating an AutoML run with non-ASCII column names (e.g. Arabic `لديه روح`) fails because Kubeflow Pipelines stores run parameters in MySQL and many clusters reject multibyte UTF-8 (`PipelineRuntimeManifest`, Error 1366). The same failure happens from the Pipelines UI.
This PR stops AutoML from hitting that 500 by rejecting non-ASCII column names in the configure UI and BFF, with a clear validation message. It does not rewrite CSVs or work around the KFP charset limitation.

 Before 
before1

 
 After
After1

## How Has This Been Tested?
- `go test ./internal/api/ ./internal/repositories/ -count=1` in `packages/automl/bff`
- Jest: `columnUtils`, `configure.schema`
- Contract: non-ASCII `label_column` is rejected
- Confirmed Pipelines UI create-run fails with the same MySQL error

## Test Impact
BFF/frontend unit + contract tests cover ASCII validation. 

## Request review criteria:

If you have UI changes: 

## Summary by CodeRabbit

* **New Features**
  * Added support for creating pipeline runs with non-ASCII CSV column names, including Arabic and other UTF-8 values.
  * Automatically prepares compatible training data while preserving the original column selections.
  * Improved handling of UTF-8 BOM characters in CSV headers and request fields.

* **Bug Fixes**
  * Provides clearer validation and server error messages, including guidance for unsupported column values.
  * Long error messages are safely shortened for display.


jefho-rh:
Thanks for the fix @jkyaw! 

### **AI-assisted review flagged these issues**

### 🔴 Must Fix

| # | Severity | File | Finding |
|---|----------|------|---------|
| 1 | **Critical** | `kfp_column_compat.go` / `AutomlReconfigureLoader.tsx:186` / `AutomlInputParametersPanel.tsx:249` | **Reconfigure flow broken for non-ASCII columns.** `ApplyKFPColumnAliasMap` stores `_ac_<hash>` aliases in KFP `runtime_config.parameters` but the original column name is not preserved anywhere. The reconfigure form pre-fills with the alias, and the input parameters panel displays it. → Store original-to-alias mapping as an additional KFP parameter (e.g., `_automl_column_alias_map`) so the frontend can reverse-map for display and reconfigure. |

### 🟡 Should Fix

Not blockers. If you feel they are unnecessary, feel free to skip.

| # | Severity | File | Finding |
|---|----------|------|---------|
| 2 | **Medium** | `pipeline_run_handler.go:82` + `pipeline_runs.go:648` | **Double validation.** `ValidateAndNormalizeCreateAutoMLRunRequest` is called twice per request — once in the handler and again inside `CreatePipelineRun`. → Remove from one call site. |
| 3 | **Medium** | `kfp_column_compat_test.go` | **Missing test coverage.** `ApplyKFPColumnAliasMap` is only tested for `PipelineTypeTabular`. The timeseries branch (`Target`, `IDColumn`, `TimestampColumn`, `KnownCovariatesNames`) is untested. → Add `TestApplyKFPColumnAliasMap_TimeSeries`. |
| 4 | **Medium** | `pipeline_run_prepare.go:70-129` | **Untested error branches.** `resolveS3ClientForRun` has 10+ error branches (missing identity, missing namespace, secret 404/403, etc.) with no direct unit tests — only the happy path is exercised. → Add table-driven test for error cases. |
| 5 | **Low** | `kfp_column_compat.go:109` | **Fragile micro-optimization.** `ReuseRecord = true` causes the CSV reader to reuse the same backing array. Safe today but silently corrupts data if lines are reordered. Allocation savings are negligible for a bounded 32 MiB file. → Remove `reader.ReuseRecord = true`. |

---

### ✅ Manual Testing


https://github.com/user-attachments/assets/eceecf88-ad61-4265-a83e-b08f9deae974



---

### ✅ Tests:

Package | Lint | Type-check | Unit Tests | Contract Tests | BFF Tests
-- | -- | -- | -- | -- | --
All (CI) | Pass | Pass | Pass | Pass | Pass

jkyaw:
Thanks @jefho-rh — addressed the review findings in c216becfd:

Critical
1. Persist original→alias map as _automl_column_alias_map in KFP runtime_config.parameters, and reverse-map in the UI for reconfigure + input parameters display.

Should fix 
2. Removed duplicate ValidateAndNormalizeCreateAutoMLRunRequest from CreatePipelineRun (handler still validates once). 
3. Added TestApplyKFPColumnAliasMap_TimeSeries. 
4. Added table-driven error/success tests for resolveS3ClientForRun. 
5. Removed reader.ReuseRecord = true.

jefho-rh:
Thanks for the quick fix @jkyaw! It looks like I'm now blocked when creating a run with a non-ASCII label column, KFP rejects the request with a 400:

`
parameter(s) provided are not required by pipeline: _automl_column_alias_map
KFP validates runtime_config.parameters against the pipeline spec's declared inputs, so it rejects any unknown parameters. The alias map can't be stored there.
`


https://github.com/user-attachments/assets/a5b767be-db62-4b27-be4a-af5239716267

jkyaw:
Thanks @jefho-rh  for catching that — you’re right, KFP rejects undeclared runtime_config.parameters.


https://github.com/user-attachments/assets/8a0ae430-603a-4b0e-9f74-6c4499a135b8

jefho-rh:
Thanks @jkyaw. Although storing the column alias in the run description works with bypassing the upstream limitation of KFP's MySQL database charset not allowing non-ASCII characters, unfortunately the alias appears in the description of the pipeline run.

Screenshot 2026-07-17 at 3 29 30 PM

image

I'm not quite sure what the correct solution here is given the time constraints since implementing an upstream fix could take time. 

Maybe as an interim fix perhaps we can block column names that do not contain ASCII characters?  

Any thoughts/suggestions @nickmazzi?

nickmazzi:
Hmm Dorota opened the issue against UI, but doesnt the regular pipelines UI have this same problem? Doesnt running with oc commands have this problem? I would expect that all possible ways to run the job use mysql database right? So I don’t think it is a UI bug...

1. We should run from regular pipelines UI and manually enter all the fields including the non ascii character column name
2. If the job fails, we should block via UI
3. If the job passes... I don’t know what to do. 

I don’t think we can use the current fix anyways

> the BFF now rewrites non-ASCII CSV headers to ASCII aliases, uploads an .automl-ascii.csv copy to S3,

What happens if the file is huge? We currently have 32MB limit but we were supposed to circle back and fix that to be 1GB (we never did yet...). The rewrite would be too slow.

So I think the only choice is to block. It would help our case if we can confirm it doesnt work in other flows (but I suspect Dorota would have tried that before opening the bug...)

@chrjones-rh thoughts?

chrjones-rh:
> Hmm Dorota opened the issue against UI, but doesnt the regular pipelines UI have this same problem? Doesnt running with oc commands have this problem? I would expect that all possible ways to run the job use mysql database right? So I don’t think it is a UI bug...
> 
> 1. We should run from regular pipelines UI and manually enter all the fields including the non ascii character column name
> 2. If the job fails, we should block via UI
> 3. If the job passes... I don’t know what to do.
> 
> I don’t think we can use the current fix anyways
> 
> > the BFF now rewrites non-ASCII CSV headers to ASCII aliases, uploads an .automl-ascii.csv copy to S3,
> 
> What happens if the file is huge? We currently have 32MB limit but we were supposed to circle back and fix that to be 1GB (we never did yet...). The rewrite would be too slow.
> 
> So I think the only choice is to block. It would help our case if we can confirm it doesnt work in other flows (but I suspect Dorota would have tried that before opening the bug...)
> 
> @chrjones-rh thoughts?

Agree that this seems like a platform limitation that we shouldn't be working around with a fix in the AutoX BFF layer.  As @nickmazzi recommended, we should test using the Pipelines UI to trigger the run and validate that the issue still exists (which it likely will).  If it does, the best we can do at this point is to block from the UI side and follow up with the KFP team to see if they already have an issue tracking this limitation, if not, create a new one.

jkyaw:
Thanks @jefho-rh, @nickmazzi, and @chrjones-rh.

Confirmed via Pipelines UI create-run (not AutoML): same MySQL charset failure on PipelineRuntimeManifest when label_column is non-ASCII (لديه روح).
- Request: POST .../apis/v2beta1/runs → 500
- Error 1366: Incorrect string value: '\\xD9\\x84\\xD8\\xAF\\xD9\\x8A...' for column ... PipelineRuntimeManifest

So this is a platform/KFP limitation. Next I’ll update this PR to block non-ASCII column names in the AutoML UI and follow up with KFP for an upstream issue.

MatthewAThompson:
Thanks @jkyaw. I see that non-ascii columns are filtered out, but I agree with the medium priority finding that it might not be an intuitive ux to have some columns from your data missing in the dropdown without explanation. Maybe we should get design input on the behaviour here.
Also, please verify the finding about dead code.

## PR Review Summary

### 1. Automated Test Results

| Package         | Test Suite     | Status  | Notes                                                                |
| --------------- | -------------- | ------- | -------------------------------------------------------------------- |
| automl/bff      | BFF Tests (Go) | ✅ Pass | All packages pass: api, repositories, s3, pipelineserver, kubernetes |
| automl/frontend | Lint           | ✅ Pass | 0 warnings                                                           |
| automl/frontend | Type Check     | ✅ Pass | No type errors                                                       |
| automl/frontend | Unit Tests     | ✅ Pass | 1445 tests across 75 suites                                          |

### 2. AI Code Review Findings

**What this PR does**: KFP MySQL-backed deployments reject multibyte UTF-8 in `PipelineRuntimeManifest`/`WorkflowRuntimeManifest`. This PR blocks non-ASCII column names upfront in both the BFF and the frontend form schema. It also ships a backward-compatibility layer to reverse-map ASCII aliases back to original names when reading old runs.

**The backward-compat layer appears to be leftover from the original approach** (see High priority below).

#### High Priority Issues

- [ ] **Dead backward-compatibility code from the abandoned alias-map approach**
  <details><summary>Details</summary>
  The original plan was to rewrite non-ASCII column names to ASCII aliases and store a reverse-map so they could be restored. That approach was abandoned and never merged to `main` — confirmed by checking that none of the intermediate commits are reachable from `main` and the string `_automl_column_alias_map` never appears in `main`'s history. Since no production run can have alias-map data, the entire backward-compat layer is dead code.

  **BFF — `kfp_column_compat.go`**: Only `ValidateASCIIColumnNames` and `containsNonASCII` are needed. Remove:
  - `KFPColumnAliasMapParamKey` and `columnAliasMapDescriptionMarker` constants
  - `StripColumnAliasMapFromDescription`
  - `parseColumnAliasMapFromParameters`
  - `RestoreOriginalColumnNamesInParameters`

  **BFF — `pipeline_runs.go:201–209`**: The three alias-map calls in `toPipelineRun` are dead:
  ```go
  description, aliases := StripColumnAliasMapFromDescription(kfRun.Description)
  // ...
  aliases = parseColumnAliasMapFromParameters(params)
  RestoreOriginalColumnNamesInParameters(params, aliases)
  ```

  **Frontend — `utils.ts`**: Remove `AUTOML_COLUMN_ALIAS_MAP_PARAM`, `resolveOriginalColumnNames`, and the private `parseColumnAliasMap` helper.

  **Frontend — callers**: Remove the `resolveOriginalColumnNames` calls in `AutomlResultsContext.ts:88` and `AutomlReconfigureLoader.tsx:76`.

  **Frontend — `AutomlInputParametersPanel.tsx:37`**: `AUTOML_COLUMN_ALIAS_MAP_PARAM` in `EXCLUDED_KEYS` can be removed.

  **Tests**: `TestStripColumnAliasMapFromDescription_LegacyRoundTrip`, `TestRestoreOriginalColumnNamesInParameters` in `kfp_column_compat_test.go`, and the entire `describe('resolveOriginalColumnNames')` block in `utils.spec.ts` cover only the dead code and should go with it.
</details>

#### Medium Priority Issues

- [ ] **Silent column filtering with no user feedback** (`AutomlConfigure.tsx:273–276`)
  <details>
  <summary>Details</summary>

  ```tsx
  const columns = React.useMemo(
    () => schemaColumns.filter((column) => isASCIIOnly(column.name)),
    [schemaColumns],
  );
  ```
  Non-ASCII columns are silently dropped from the target column picker. A user who uploads a CSV with non-ASCII headers sees an empty or reduced column list with no indication of why those columns are absent. The `NON_ASCII_COLUMN_NAME_MESSAGE` constant exists and is used in the schema validator, but that only fires if a non-ASCII name somehow bypasses the picker.

  Consider adding a contextual note when `schemaColumns.length > columns.length` — e.g. "X column(s) with non-ASCII names are not shown because Kubeflow Pipelines does not support them."
  </details>

#### Low Priority / Suggestions

- **`formatPipelineServerCreateRunError` heuristics** (`pipeline_run_errors.go:24`): `"Incorrect string value"` is a broad MySQL error string. A short comment explaining it's MySQL's utf8mb4 rejection phrase would help future readers — and is worth noting that this branch should now be unreachable given the upfront validation.

**Summary**: The approach is sound and all tests pass. The main issue is that the backward-compat layer — written for a graceful migration from the alias-map approach — has no runs to migrate because that approach never shipped. Removing it significantly reduces the diff and eliminates a maintenance surface that would otherwise need explaining forever.

### 3. Manual Review

- [x] **Manual testing performed**: Verify uploading a CSV with non-ASCII column headers gives a reduced/empty column picker

**Evidence/Screenshots**:
Screenshot 2026-07-24 at 2 15 39 pm

jkyaw:
Thanks @MatthewAThompson  — addressed your review findings:

High — dead alias-map compat
Removed the unused original→alias reverse-mapping on both BFF and frontend (it never shipped to main, so there was nothing to migrate). See 8b4fd5d7b.

Medium — silent column filtering
Added an inline info alert under the target column picker when non-ASCII columns are hidden. After design feedback we used the info (not caution) style and copy:
N column(s) … hidden because Kubeflow pipelines do not support non-ASCII names.
See 21bd5a20f.

Low — MySQL error string
Added a short comment on the "Incorrect string value" heuristic in pipeline_run_errors.go.

Design 
Screenshot 2026-07-28 at 10 55 05 AM


After
Screenshot 2026-07-28 at 11 22 09 AM

MatthewAThompson:
Thanks for the updates @jkyaw 
/lgtm

GAUNSD (commented):
## AI code review 🤖  🔍
`🤖: Claude Opus 4.6 (1M context)`

PR is structurally sound. The defense-in-depth approach — frontend column filter → Zod schema validation → BFF normalization + validation → error formatting — is well-layered and correctly prevents the KFP MySQL charset failure. CI is fully green. One yellow finding survives around a leftover from the intermediate approach; the rest are informational.

## Summary

| # | Sev | File | Finding |
|---|-----|------|---------|
| 1 | 🟡 | `pipeline_runs.go` | `cloneRuntimeParameters` leftover from abandoned alias-map approach changes nil → `{}` serialization |
| 2 | 🔵 | `pipeline_run_errors.go` | Error detection checks `WorkflowRuntimeManifest` but not `PipelineRuntimeManifest` (covered by `Incorrect string value` fallback) |
| 3 | 🔵 | `AutomlConfigure.tsx` | All-non-ASCII-columns edge case: empty picker with only an info alert, no explicit blocking guidance |
| 4 | 🔵 | `columnUtils.ts` | `isASCIIOnly` allows ASCII control characters (0x00–0x1F) in column names |
| 5 | 🔵 | `pipeline_runs.go` | Validation contract moved from `CreatePipelineRun` to caller — documented but not enforced |

## Details

<details><summary>Item 1: 🟡 `cloneRuntimeParameters` is a leftover from the abandoned alias-map approach and changes nil serialization</summary>

### 🟡 `cloneRuntimeParameters` is a leftover from the abandoned alias-map approach and changes nil serialization

`toPipelineRun` on `main` directly assigns `RuntimeConfig: kfRun.RuntimeConfig`. This PR restructures it to create a new `RuntimeConfig` with cloned parameters via
`cloneRuntimeParameters`. That clone was introduced for the alias-map approach (which needed `RestoreOriginalColumnNamesInParameters` to mutate the map in-place). The alias-map
code was correctly removed in `8b4fd5d7`, but the clone survived the cleanup.

The clone itself is harmless for the happy path, but `cloneRuntimeParameters` returns `map[string]interface{}{}` when `params` is nil — changing the JSON serialization of the
GET/LIST pipeline runs response from `"parameters": null` to `"parameters": {}`. If any consumer checks `parameters === null` (or treats null vs empty-object differently), this
is a behavioral change.

**Risk**: Low — AutoML runs always have populated parameters, so the nil case is unlikely. But this is unnecessary code that introduces a subtle API contract change for no
current benefit.

**Suggestion**: Revert `toPipelineRun` to the direct `RuntimeConfig: kfRun.RuntimeConfig` assignment from `main`, and remove `cloneRuntimeParameters`. If defensive cloning is
desired long-term, keep it but fix the nil case: `if params == nil { return nil }`.

**Files**:
- `packages/automl/bff/internal/repositories/pipeline_runs.go:228–238` (`cloneRuntimeParameters`)
- `packages/automl/bff/internal/repositories/pipeline_runs.go:198–209` (`toPipelineRun` restructuring)

</details>

<details><summary>Item 2: 🔵 Error detection misses `PipelineRuntimeManifest` column name</summary>

### 🔵 Error detection misses `PipelineRuntimeManifest` column name

`formatPipelineServerCreateRunError` checks for `WorkflowRuntimeManifest` in the MySQL error message but not `PipelineRuntimeManifest`. The PR description and jkyaw's own
reproduction comment explicitly reference `PipelineRuntimeManifest` as the failing column:

> Error 1366: Incorrect string value: '\\xD9\\x84...' for column ... PipelineRuntimeManifest

The gap is covered by the `Incorrect string value` substring check, so the charset-specific error message would still be returned. But the explicit column-name check is
incomplete.

**Risk**: None — the `Incorrect string value` fallback covers it. Purely a completeness nit.

**Suggestion**: Add `strings.Contains(message, "PipelineRuntimeManifest")` to the condition for completeness.

**File**: `packages/automl/bff/internal/api/pipeline_run_errors.go:29`

</details>

<details><summary>Item 3: 🔵 All-non-ASCII-columns edge case leaves users with an empty picker</summary>

### 🔵 All-non-ASCII-columns edge case leaves users with an empty picker

When every column in a CSV has a non-ASCII name, the filter produces an empty `columns` array. The picker renders with no options and the info alert reads e.g. "5 columns are
hidden because Kubeflow pipelines do not support non-ASCII names." The user has no selectable target column, so the form won't submit (schema requires `target_column`), but the
info-level alert doesn't convey that the situation is unrecoverable without a different CSV.

This was partially addressed per MatthewAThompson's review feedback (commit `21bd5a20` added the info alert). The UX is functional — the old behavior was a 500 from the
pipeline server — so this is strictly an improvement.

**Risk**: UX confusion only, not a regression. No data loss or error path.

**Suggestion**: Consider upgrading to a warning alert when `columns.length === 0 && filteredNonASCIIColumnCount > 0`, explicitly telling the user they need a CSV with ASCII
column names.

**File**: `packages/automl/frontend/src/app/components/configure/AutomlConfigure.tsx:908–918`

</details>

<details><summary>Item 4: 🔵 `isASCIIOnly` allows ASCII control characters in column names</summary>

### 🔵 `isASCIIOnly` allows ASCII control characters in column names

Both the frontend `isASCIIOnly` (`charCodeAt(i) > 0x7f`) and the BFF `containsNonASCII` (`r > 0x7F`) treat all code points 0x00–0x7F as valid. This includes control characters:
null bytes, tabs, carriage returns, escape sequences, etc. A column named `"target\x00"` or `"sales\t"` would pass ASCII validation.

**Risk**: Extremely low — CSV parsers don't produce column names with embedded control characters under normal circumstances. The BOM stripping + TrimSpace normalization
already handles the most likely edge cases (BOM, leading/trailing whitespace).

**File**: `packages/automl/frontend/src/app/utilities/columnUtils.ts:70–76` and `packages/automl/bff/internal/repositories/kfp_column_compat.go:10–15`

</details>

<details><summary>Item 5: 🔵 `CreatePipelineRun` no longer self-validates — relies on caller contract</summary>

### 🔵 `CreatePipelineRun` no longer self-validates — relies on caller contract

`ValidateCreateAutoMLRunRequest` was moved out of `CreatePipelineRun` and into the handler via `ValidateAndNormalizeCreateAutoMLRunRequest`. The doc comment on
`CreatePipelineRun` now states "Callers must validate/normalize the request before calling", but this isn't enforced in code. A future caller that skips validation would send
unvalidated data to the pipeline server.

This refactoring is reasonable — normalization needs to happen before validation, and the handler is the right place for both — but it weakens the function's self-contained
correctness guarantee.

**Risk**: Zero today (single caller). Maintenance concern if a second caller is added without reading the doc comment.

**File**: `packages/automl/bff/internal/repositories/pipeline_runs.go:674`

</details>

## Human code review 👤 🔍

I think `Item 1: 🟡` should be fine since I'm assuming the callers are all in our control and have been updated to handle the `null` case being mapped differently.

@jkyaw can you confirm that item 1 is fine. Once you do I'll approve

jkyaw:
Thanks @GAUNSD  for the CR. Confirmed item 1 is fine.

GAUNSD:
Thanks @jkyaw!

/approve
/lgtm

GAUNSD (approved):
Thanks @jkyaw!

/approve
/lgtm

PR #8535, opened 2026-07-14
fix(cypress): prevent after-hook cleanup failures from masking passing tests

dpanshug:
https://issues.redhat.com/browse/RHOAIENG-76560

## Description

The `after()` hooks in several e2e permission and workbench tests call `ensureAdminOcSession()` followed by `deleteOpenShiftProject()`. When the admin session restoration fails silently (the function only logs a warning), the subsequent `oc delete project` runs as the **non-admin user**, gets a `Forbidden` error, and **throws** — which is fatal in an `after all` hook since Cypress does not retry those.

This causes passing tests to be marked as failed. For example, in [CI run #20956](https://github.com/opendatahub-io/odh-dashboard/actions/runs/29315352253/job/87028108930), `testProjectContributorPermissions.cy.ts` had both test assertions pass, but the after-hook cleanup threw at `project.ts:71`, failing the entire spec.

### Changes

- **Add `cleanupTestProject()` helper** to `projectChecker.ts` that wraps the admin-restore + project-delete pattern with `failOnNonZeroExit: false`, so a cleanup failure is logged but never throws.
- **Replace the raw `ensureAdminOcSession() + deleteOpenShiftProject()` calls** in all four vulnerable `after` hooks with this helper:
  - `testProjectContributorPermissions.cy.ts`
  - `testProjectAdminPermissions.cy.ts`
  - `testWorkbenchCreation.cy.ts`
  - `testModelPvcDeployment.cy.ts`

The `oc delete project` command is identical (`--wait=false --ignore-not-found`). The only behavioral change is that the after hook no longer throws on cleanup failure.

## How Has This Been Tested?

Tested in Jenkins (#1932) - failures not related to this fix

- Verified the generated `oc` command is identical to the original `deleteOpenShiftProject({ wait: false, ignoreNotFound: true })`.
- Full `tsc --noEmit` type-check on the cypress package: 0 errors.
- Full `eslint` lint on all changed files: 0 errors.
- `lint-staged` pre-commit hook passed.
- Verified no circular imports introduced (`baseCommands.ts` does not import from `projectChecker.ts`).
- Verified no other test depends on these project names (all use `generateTestUUID()` for unique names).
- Verified all 40+ existing callers of `projectChecker.ts` are unaffected (they only import `createCleanProject`/`createAndVerifyProject`, which are unchanged).

## Test Impact

This change fixes test infrastructure (after-hook cleanup), not application code. No new tests are needed — the existing e2e tests cover the same functionality. The fix prevents false failures in the affected specs.

## Request review criteria:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cleanup of temporary projects after automated test runs.
  * Prevented cleanup failures from unnecessarily interrupting test execution.
  * Improved handling of administrative OpenShift sessions during test teardown.

* **Tests**
  * Updated project, model deployment, and workbench end-to-end tests to use consistent cleanup behavior.
  * Added safeguards for environments where an OpenShift server is unavailable.


ConorOM1:
@dpanshug thanks for actioning this. One comment here... your fix silences the failure but it seems like `ensureAdminOcSession()` is the real problem here. It returns void instead of a chainable, so the login hasn't necessarily completed before deleteOpenShiftProject runs. Could you address this ?

PR #8538, opened 2026-07-14
RHOAIENG-73189: Drop redundant SAR pre-checks and add sandbox CRD RBAC

Gkrumbach07:
## Summary
- Remove redundant SAR pre-flight checks from all agent BFF routes — K8s RBAC is already enforced on actual API calls via impersonation (`AuthMethodInternal`) or user token (`AuthMethodUser`)
- Remove inline `CanDeployAgentInNamespace` SAR check from deploy handler
- Add aggregate ClusterRole for `agents.x-k8s.io/sandboxes` so namespace editors/admins get CRD access without needing cluster-admin

## Details

The BFF creates an impersonated K8s clientset for every request (`factory.go:83-86`), so all operations (list, get, create, patch, delete) are already executed as the user. The SAR middleware was a redundant pre-flight check that could incorrectly block users who have the right permissions on the actual resources.

The aggregate ClusterRole uses `aggregate-to-admin`/`aggregate-to-edit` labels to automatically merge sandbox CRD permissions into the built-in namespace `admin` and `edit` roles — same pattern as `aggregate-hardware-profiles-permissions`.

## Test plan
- [x] BFF builds cleanly (`go build ./...`)
- [x] All API tests pass (`go test ./internal/api/`)
- [x] Kustomize validates (`kustomize build manifests/common/rbac/project-users/`)
- [ ] Verify non-admin user can list/get/deploy/stop/start/delete agents in their namespace
- [ ] Verify user without namespace access gets 403 from K8s (not from SAR middleware)

## Summary by CodeRabbit

* **New Features**
  * Added a Kubernetes RBAC `ClusterRole` for agent sandbox permissions (automatically aggregated into existing admin/edit roles) granting `get`, `list`, and `patch` on `sandboxes` for the agents API group.
* **Bug Fixes**
  * Agent runtime endpoints now consistently require authentication.
  * Authorization is enforced directly by Kubernetes during the underlying API calls (removing separate routing/middleware pre-checks and redundant RBAC checks during deploy/list operations).
* **Tests**
  * Updated and simplified RBAC/auth-related tests to match the new authorization flow.


packages/agent-ops/bff/internal/api/app.go
```diff
@@ -227,22 +227,23 @@ func (app *App) Routes() http.Handler {
 		return app.GetNamespacesHandler
 	}))
 
-	// Agent list: RequireAuthenticatedForAgents gates identity when auth is enabled; per-namespace
-	// SSAR filtering happens inside the Kubernetes agent client (ListNamespaces / ListAgents).
+	// Agent routes — K8s RBAC is enforced on the actual API calls via
+	// impersonation (AuthMethodInternal) or user token (AuthMethodUser).
+	// No pre-flight SAR checks needed; 403s from K8s are surfaced directly.
```
danreed-rh: **Should fix — comment overstates SAR removal + please confirm manual RBAC testing**

This comment says *"No pre-flight SAR checks needed"* for all agent routes, but **list/discovery still uses SAR** via `CanListAgentsInNamespace` in the repository/agent client layer. Detail and mutation routes defer to K8s RBAC; list does not fully follow that model.

Before merge, please complete the PR test-plan items still unchecked:
- Non-admin user can list/get/deploy/stop/start/delete in their namespace
- User without namespace access gets **403 from the K8s API** (not from removed BFF SAR middleware)

A short note in the PR with cluster results would help reviewers.

packages/agent-ops/bff/internal/integrations/kubernetes/agent_rbac.go
```diff
@@ -11,26 +11,6 @@ import (
...
-
-func (kc *InternalKubernetesClient) canAccessAgentSandbox(
-	ctx context.Context,
-	identity *RequestIdentity,
-	namespace, name, verb string,
-) (bool, error) {
-	return kc.subjectAccessReviewGroup(ctx, identity, namespace, name, sandboxAPIGroup, "sandboxes", verb)
-}
-
 func (kc *InternalKubernetesClient) CanListAgentsInNamespace(ctx context.Context, identity *RequestIdentity, namespace string) (bool, error) {
```
danreed-rh: **Should fix — list-path SAR inconsistency**

Route middleware no longer pre-checks SAR for get/deploy/patch/delete, but **list and namespace discovery still gate on SAR** here (`sandboxes` + `list` verb).

That can reproduce the same false-negative the PR fixes for writes: a user denied by SAR but allowed on the actual sandbox `List()` call would still be blocked at list/discovery time.

Please either:
1. Align list with the deferred model (call K8s `List` and surface 403/empty results), or
2. Document in the PR why list SAR is intentionally retained as the one exception.

If keeping SAR for list, a regression test for "SAR denies, K8s list would allow" would help prevent drift.

packages/agent-ops/bff/internal/api/middleware_test.go
```diff
@@ -102,31 +84,17 @@ func (f *rbacTestK8sFactory) ValidateRequestIdentity(identity *k8s.RequestIdenti
 }
 
 func newRBACTestApp(allowed bool) *App {
-	return newRBACTestAppWithGet(allowed, allowed)
+	return newRBACTestAppWithConfig(allowed, config.EnvConfig{AuthMethod: config.AuthMethodInternal})
```
danreed-rh, line 85: **Should fix — thin test coverage for new auth model**

The SAR middleware tests (`RequireAccessToAgent` / patch / delete) were removed with the middleware, but coverage for the replacement model is minimal (only `TestRequireAuthenticatedForAgents_Allowed` / `_Unauthorized` remain).

Please add at least:
- `TestRequireAuthenticatedForAgents_AuthDisabled` — handler runs when `AuthMethodDisabled` (no identity in context)
- A test that the middleware never invokes SAR / `GetClient()` for agent auth (e.g. factory client that `t.Fatal`s if called)

Optional but valuable: handler-level 403 tests for deploy/stop/start when the repo returns `agents.ErrForbidden` (delete/detail have coverage elsewhere).

packages/agent-ops/bff/internal/api/middleware.go
```diff
@@ -215,94 +215,3 @@ func (app *App) RequireAccessToService(next func(http.ResponseWriter, *http.Requ
 	}
```
danreed-rh, line 215: **Context — SAR removal is incomplete on list paths**

Good cleanup removing `RequireAccessToAgent` / patch / delete middleware here. One follow-up: `CanListAgentsInNamespace` SAR remains in `agent_rbac.go` for list/discovery filtering, so the "defer all RBAC to K8s API calls" model is not uniform across endpoints. See inline comment on `agent_rbac.go`.

danreed-rh (commented):
Multi-agent review follow-up — items to address before merge.

**Manual verification:** Please verify the two unchecked test-plan items (non-admin lifecycle in namespace; 403 from K8s for users without access). As unit tests passing does not validate the impersonation-backed RBAC model on a real cluster.

Inline comments cover the list-path SAR split and test coverage for the new auth-only middleware.

packages/agent-ops/bff/internal/api/app.go
Gkrumbach07: Fixed. Comment now accurately says list uses SAR in the agent client layer for namespace filtering, while detail/mutation routes defer to K8s. Also removed the list SAR entirely in a follow-up commit — `listAgentSummaries` already handles `IsForbidden` by skipping the namespace, so the SAR pre-check was redundant.

Manual RBAC testing is still pending — will update once tested on cluster.

packages/agent-ops/bff/internal/integrations/kubernetes/agent_rbac.go
Gkrumbach07: Resolved — removed `CanListAgentsInNamespace` entirely. The impersonated `List()` call returns 403 for namespaces the user can't access, and `listAgentSummaries` already handles `IsForbidden` by skipping. The SAR was predicting what the List call would do — unnecessary extra API call per namespace.

All agent endpoints now have zero SAR. K8s RBAC is the single enforcement point via impersonation.

packages/agent-ops/bff/internal/api/middleware_test.go
Gkrumbach07: Added both requested tests:
- `TestRequireAuthenticatedForAgents_AuthDisabled` — handler runs when auth disabled, no identity in context
- `TestRequireAuthenticatedForAgents_DoesNotCallGetClient` — uses a `fatalIfCalledFactory` that fails the test if `GetClient()` is ever invoked, proving middleware never touches SAR

packages/agent-ops/bff/internal/api/middleware.go
Gkrumbach07: Done — list SAR is now removed too. `CanListAgentsInNamespace` deleted from the interface, both implementations, all test mocks, and the `ListNamespaces` filter loop. Zero SAR remains across all agent endpoints.

Gkrumbach07:

manifests/common/rbac/project-users/aggregate-agent-sandbox-permissions.rbac.yaml
```diff
@@ -0,0 +1,16 @@
...
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: aggregate-agent-sandbox-permissions
+  labels:
+    rbac.authorization.k8s.io/aggregate-to-admin: 'true'
+    rbac.authorization.k8s.io/aggregate-to-edit: 'true'
+rules:
+  - apiGroups:
+      - agents.x-k8s.io
+    verbs:
```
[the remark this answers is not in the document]
Gkrumbach07: expected

Issue #1804, opened 2023-04-13
[Tracker] Model Registry


kywalker-rh:
### Description

Incorporate a model registry into the RHODS MLOps workflow, offering a centralized repository for storing and managing models, their versions, and associated metadata. This enhancement bridges the divide between model experimentation and production processes, streamlining user collaboration, facilitating integration with model serving and deployment, and optimizing the overall efficiency of the MLOps workflow.

### Target Branch

TBD

### Requirements

[Requirements doc #1](https://docs.google.com/document/d/1mRUQ9C5Br7T3bA49H503HTq0J7F4WKySdxaWpTX-i8k/edit) (will covert over to here later)
[Requirements doc #2](https://docs.google.com/document/d/1XT0PxVp78VEyeMXKkF5qrP-fdgN4V_G9UxH2a1OpPY8/edit#heading=h.vk5ku8fmql4y)
[Proposal doc](https://docs.google.com/document/d/1G-pjdGaS2kLELsB5kYk_D4AmH-fTfnCnJOhJ8xENjx0/edit#heading=h.ds8q4xtkmu64)
[Piroritized user stories](https://docs.google.com/document/d/13Q9CT0Qx6HUxumisyWQRkvcJpVB2peKC86fs7ZXw1vQ/edit#heading=h.439yc0mcih2e)

### Itemized UX Issues

- https://github.com/opendatahub-io/sig-ml-developer-experience/issues/17
- https://github.com/opendatahub-io/odh-dashboard/issues/1806
- https://github.com/opendatahub-io/odh-dashboard/issues/1805
- https://github.com/opendatahub-io/odh-dashboard/issues/1807
- https://github.com/opendatahub-io/odh-dashboard/issues/2130
- https://github.com/opendatahub-io/odh-dashboard/issues/1808

### Itemized Dev Issues

* https://github.com/opendatahub-io/odh-dashboard/issues/1920

#### API

From the Model Registry team:

> With the OpenAPI merged on main branch, [here a "permalink"](https://github.com/opendatahub-io/model-registry/blob/main/api/openapi/model-registry.yaml) to the OpenAPI file, and [here a "permahyperlink"](https://editor.swagger.io/?url=https://raw.githubusercontent.com/opendatahub-io/model-registry/main/api/openapi/model-registry.yaml) for displaying it on a swagger.

dgutride:
Migrated to:  https://issues.redhat.com/browse/RHOAIENG-571

PR #8544, opened 2026-07-14
fix(cypress): remove hardcoded MLflow image and use Available condition for CR readiness

DaoDaoNoCode:
https://issues.redhat.com/browse/RHOAIENG-76588

## Description

The `testMlflowExperiments` and `testPromptManagement` E2E tests have been failing on the RHOAI nightly pipeline since build #1804 (Jul 13). The root cause is a combination of two issues in the E2E test infrastructure:

**1. Hardcoded MLflow image in the CR fixture causes version mismatch**

The E2E fixture at `fixtures/e2e/mlflow/mlflowCR.yaml` hardcoded `quay.io/opendatahub/mlflow:odh-stable` as the container image. This overrides the operator's `MLFLOW_IMAGE` env var. When the `opendatahub-io/mlflow` fork was [rebased to v3.14.0](https://github.com/opendatahub-io/mlflow/pull/331) on Jul 12, the operator's `SupportedMLflowVersion` advanced to `v3.14.0`. The hardcoded image tag created a version mismatch during the operator's DB migration flow, which scales the MLflow deployment to zero replicas while migration runs — and leaves it at zero if migration fails.

**2. Readiness check doesn't account for migration phase**

The `waitForMlflowCRReady` function polled `status.address.url`, which does not account for the operator's version migration phase. The operator intentionally holds the deployment at zero replicas during migration and only sets the `Available` condition to `True` after migration completes and the deployment is fully rolled out. The `status.address.url` check would either time out (if migration is in progress) or succeed prematurely (before the pod passes readiness probes).

**Changes:**

- Remove hardcoded `spec.image`, `spec.resources`, `spec.podAnnotations`, and `spec.env` from the E2E MLflow CR fixture — the operator supplies its own image via `MLFLOW_IMAGE`, matching the proven minimal fixture pattern used by the eval-hub test suite
- Replace `waitForMlflowCRReady` (polls `status.address.url`) with `waitForMlflowCRAvailable` (polls `status.conditions[Available]=True`), which naturally handles the migration phase
- Increase poll budget from 60 to 120 attempts (10 minutes) to accommodate the migration window

**Impact:** `testMlflowExperiments`, `testPromptManagement` (both use `enableMlflowBackend()`). The `testEvalHub` test is **not affected** — it uses `mlflowInstance.ts` which already follows the correct `Available` condition pattern.

## How Has This Been Tested?

- Verified the fixture produces valid YAML
- Verified no stale references to the old `waitForMlflowCRReady` function remain
- Verified lint passes on the changed TypeScript file
- Cross-checked against the mlflow-operator's [migration.go](https://github.com/opendatahub-io/mlflow-operator/blob/main/internal/controller/migration.go) and [helm.go](https://github.com/opendatahub-io/mlflow-operator/blob/main/internal/controller/helm.go) to confirm the `Available` condition is set only after migration + deployment rollout
- Confirmed the `Available` condition polling pattern matches the proven `mlflowInstance.ts` utility already used by the eval-hub E2E test

## Test Impact

This PR fixes the E2E test infrastructure itself. No new tests are added — the existing `testMlflowExperiments` and `testPromptManagement` tests are the ones being fixed. The changes should allow them to pass on nightly clusters where the MLflow operator has been updated to v3.14.0.

## Request review criteria:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved MLflow provisioning checks by waiting for the MLflow custom resource to report `Available=True` (including migration) before proceeding.
  * Updated MLflow feature disable/teardown behavior to be more predictable during end-to-end runs.
* **Tests**
  * Improved MLflow Experiments and Prompt Management e2e tests by removing CR-existence caching logic and enabling required features as part of the setup sequence.
  * Simplified the MLflow custom resource test fixture configuration by removing unnecessary pod-level settings and debug logging.

### DIFF
--- a/packages/cypress/cypress/fixtures/e2e/mlflow/mlflowCR.yaml
+++ b/packages/cypress/cypress/fixtures/e2e/mlflow/mlflowCR.yaml
@@ -3,20 +3,7 @@ kind: MLflow
 metadata:
   name: mlflow
 spec:
-  image:
-    image: quay.io/opendatahub/mlflow:odh-stable
-    imagePullPolicy: Always
   replicas: 1
-  resources:
-    requests:
-      cpu: '1'
-      memory: '2Gi'
-    limits:
-      cpu: '4'
-      memory: '3Gi'
-  podAnnotations:
-    prometheus.io/scrape: 'true'
-    prometheus.io/port: '8443'
   storage:
     accessModes:
       - ReadWriteOnce
@@ -27,6 +14,3 @@ spec:
   registryStoreUri: 'sqlite:////mlflow/mlflow.db'
   artifactsDestination: 'file:///mlflow/artifacts'
   serveArtifacts: true
-  env:
-    - name: MLFLOW_LOGGING_LEVEL
-      value: DEBUG
--- a/packages/cypress/cypress/tests/e2e/mlflowExperiments/testMlflowExperiments.cy.ts
+++ b/packages/cypress/cypress/tests/e2e/mlflowExperiments/testMlflowExperiments.cy.ts
@@ -2,7 +2,6 @@ import { HTPASSWD_CLUSTER_ADMIN_USER } from '../../../utils/e2eUsers';
 import {
   enableMlflowFeatures,
   disableMlflowFeatures,
-  doesMlflowCRExist,
   createMlflowExperimentViaAPI,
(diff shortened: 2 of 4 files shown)


ikeola13 (approved):
/lgtm

PR #8547, opened 2026-07-14
feat(RHOAIENG-72851): add core-bff manifest overlays

lucferbux:
https://issues.redhat.com/browse/RHOAIENG-72851

## Description

Adds **core-bff** as a sidecar container in the base dashboard deployment (`core-bases/base/`), alongside `odh-dashboard` and `kube-rbac-proxy`. Core-bff will eventually replace the Fastify backend — placing it in the same pod ensures deployment continuity.

### Changes

**`manifests/core-bases/base/deployment.yaml`**
- Added `core-bff` as the third container (index 2) with:
  - Port 8943 (HTTPS), TLS cert/key from the shared `proxy-tls` secret
  - HTTPS liveness/readiness probes on `/healthcheck`
  - CA bundle mounts (odh-trusted-ca-bundle, odh-ca-bundle)
  - `POD_NAMESPACE` env from downward API
  - Security context: non-root, read-only root filesystem, all capabilities dropped

**`manifests/core-bases/base/service.yaml`**
- Added port 8943 (`core-bff`) on the shared `odh-dashboard` service

**`manifests/odh/` overlay**
- `params.env`: added `core-bff-image=quay.io/opendatahub/odh-core-bff:main`
- `kustomization.yaml`: added `core-bff-image` var referencing `odh-dashboard-params` ConfigMap

**`manifests/rhoai/` overlay**
- `params.env`: added `core-bff-image=quay.io/opendatahub/odh-core-bff:main`
- `kustomization.yaml`: added `core-bff-image` var referencing `rhoai-dashboard-params` ConfigMap
- `base/deployment.yaml`: fixed container index reference (`containers/3` → `containers/4`) for gen-ai-ui's `BFF_MAAS_SERVICE_NAME` rename, since inserting core-bff at index 2 shifts the modular-architecture sidecars

**Port**: 8943 (follows the 8X43 sequence)
**Image**: `quay.io/opendatahub/odh-core-bff:main` (from Konflux onboarding [RHOAIENG-72855](https://issues.redhat.com/browse/RHOAIENG-72855) / [RHOAIENG-72856](https://issues.redhat.com/browse/RHOAIENG-72856))

### Local validation (kustomize build)

Both ODH and RHOAI overlays build successfully:

```bash
cd manifests
kustomize build odh   # ✅ core-bff container on port 8943, image resolved
kustomize build rhoai # ✅ core-bff present, gen-ai-ui BFF_MAAS_SERVICE_NAME correctly set to rhods-dashboard
```

### Cluster testing — Option A: Dashboard Operator local run

Run the dashboard-operator locally against your cluster, pointing at the manifests from this branch. The operator will render and apply the manifests including the new core-bff container.

```bash
# 1. Clone/checkout this branch
git fetch origin pull/8547/head:pr-8547
git checkout pr-8547

# 2. Scale down the ODH operator so it doesn't fight with your local run
oc scale deployment opendatahub-operator-controller-manager \
  -n openshift-operators --replicas=0

# 3. Install the Dashboard CRD (if not present)
cd dashboard-operator
oc apply -f config/crd/bases/

# 4. Run the operator locally, pointing at this branch's manifests
#    Override core-bff-image if needed
export RELATED_IMAGE_ODH_CORE_BFF_IMAGE=quay.io/opendatahub/odh-core-bff:main
go run ./cmd/manager \
  --namespace opendatahub \
  --manifests-base-path ../manifests \
  --leader-elect=false \
  --secure-metrics=false

# 5. Verify the core-bff container is running
oc get pods -n opendatahub -l deployment=odh-dashboard \
  -o jsonpath='{range .items[*].spec.containers[*]}{.name}{"\n"}{end}'

# 6. Don't forget to restore the operator when done
oc scale deployment opendatahub-operator-controller-manager \
  -n openshift-operators --replicas=1
```

> **Note**: The `core-bff-image` key is not yet registered in the dashboard-operator's `imagesMap` (`internal/controller/support.go`), so the RELATED_IMAGE env var override won't work until that follow-up is done. The operator will still use the default value from `params.env`.

### Cluster testing — Option B: Manual kustomize apply

Apply the built manifests directly, bypassing the operator. Useful when the operator is already scaled down or not installed.

```bash
# 1. Clone/checkout this branch
git fetch origin pull/8547/head:pr-8547
git checkout pr-8547

# 2. Scale down the ODH operator to prevent reconciliation
oc scale deployment opendatahub-operator-controller-manager \
  -n openshift-operators --replicas=0

# 3. Build and apply the ODH manifests
cd manifests
kustomize edit set namespace opendatahub
kustomize build odh | oc apply -f -

# 4. Wait for the rollout and verify
oc rollout status deployment/odh-dashboard -n opendatahub
oc get pods -n opendatahub -l deployment=odh-dashboard \
  -o jsonpath='{range .items[*].spec.containers[*]}{.name}{"\n"}{end}'
# Should now include: odh-dashboard, kube-rbac-proxy, core-bff, model-registry-ui, ...

# 5. Verify the core-bff container is healthy
oc get pods -n opendatahub -l deployment=odh-dashboard \
  -o jsonpath='{range .items[*].status.containerStatuses[?(@.name=="core-bff")]}{.ready}{"\n"}{end}'

# 6. Restore the operator when done
oc scale deployment opendatahub-operator-controller-manager \
  -n openshift-operators --replicas=1
```

> **Warning**: When the operator scales back up, it will reconcile the Dashboard CR and may revert your manual changes to match the operator's built-in manifests. This approach is for validation only.

## Test Impact

Manifest-only change — no application code or test code affected. CI validates `kustomize build` automatically via `.github/workflows/validate-kustomize.yml`. Full cluster validation will happen during operator integration testing.

## Follow-ups

- Register `core-bff-image` in `dashboard-operator/internal/controller/support.go` `imagesMap` so the operator can inject the image via `RELATED_IMAGE_*` env vars
- Register the related image in `dashboard-operator/charts/dashboard/values.yaml`

## Request review criteria:

[RHOAIENG-72855]: https://redhat.atlassian.net/browse/RHOAIENG-72855?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/manifests/core-bases/base/deployment.yaml
+++ b/manifests/core-bases/base/deployment.yaml
@@ -144,6 +144,75 @@ spec:
             - mountPath: /etc/ssl/certs/odh-ca-bundle.crt
               name: odh-ca-cert
               subPath: odh-ca-bundle.crt
+        - name: core-bff
+          image: $(core-bff-image)
+          imagePullPolicy: Always
+          ports:
+            - containerPort: 8943
+              name: core-bff
+          env:
+            - name: POD_NAMESPACE
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.namespace
+          resources:
+            requests:
+              cpu: 300m
+              memory: 512Mi
+              ephemeral-storage: 10Mi
+            limits:
+              memory: 512Mi
+          livenessProbe:
+            httpGet:
+              path: /healthcheck
+              port: 8943
+              scheme: HTTPS
+            initialDelaySeconds: 30
+            timeoutSeconds: 15
+            periodSeconds: 30
+            successThreshold: 1
+            failureThreshold: 3
+          readinessProbe:
+            httpGet:
+              path: /healthcheck
+              port: 8943
+              scheme: HTTPS
+            initialDelaySeconds: 15
+            timeoutSeconds: 15
+            periodSeconds: 30
+            successThreshold: 1
+            failureThreshold: 3
+          volumeMounts:
+            - mountPath: /etc/tls/private
+              name: proxy-tls
+              readOnly: true
+            - mountPath: /etc/pki/tls/certs/odh-trusted-ca-bundle.crt
+              name: odh-trusted-ca-cert
+              subPath: odh-trusted-ca-bundle.crt
+            - mountPath: /etc/ssl/certs/odh-trusted-ca-bundle.crt
+              name: odh-trusted-ca-cert
+              subPath: odh-trusted-ca-bundle.crt
+            - mountPath: /etc/pki/tls/certs/odh-ca-bundle.crt
+              name: odh-ca-cert
+              subPath: odh-ca-bundle.crt
+            - mountPath: /etc/ssl/certs/odh-ca-bundle.crt
+              name: odh-ca-cert
+              subPath: odh-ca-bundle.crt
+          args:
+            - '--auth-method=user_token'
+            - '--auth-token-header=x-forwarded-access-token'
+            - '--auth-token-prefix='
+            - '--port=8943'
+            - '--cert-file=/etc/tls/private/tls.crt'
+            - '--key-file=/etc/tls/private/tls.key'
+            - '--bundle-paths=/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem,/var/run/secrets/kubernetes.io/serviceaccount/ca.crt,/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt,/etc/pki/tls/certs/odh-ca-bundle.crt,/etc/pki/tls/certs/odh-trusted-ca-bundle.crt'
+          securityContext:
+            allowPrivilegeEscalation: false
+            runAsNonRoot: true
+            readOnlyRootFilesystem: true
+            capabilities:
+              drop:
+                - ALL
       volumes:
         - name: proxy-tls
           secret:
--- a/manifests/core-bases/base/service.yaml
+++ b/manifests/core-bases/base/service.yaml
@@ -13,3 +13,7 @@ spec:
       protocol: TCP
       port: 8443
       targetPort: 8443
+    - name: core-bff
+      protocol: TCP
+      port: 8943
+      targetPort: 8943
--- a/manifests/odh/kustomization.yaml
+++ b/manifests/odh/kustomization.yaml
@@ -56,6 +56,13 @@ vars:
       apiVersion: v1
     fieldref:
       fieldpath: data.images-jobs-async-upload
+  - name: core-bff-image
+    objref:
+      kind: ConfigMap
+      name: odh-dashboard-params
+      apiVersion: v1
+    fieldref:
+      fieldpath: data.core-bff-image
 configurations:
   - params.yaml
 replacements:
--- a/manifests/odh/params.env
+++ b/manifests/odh/params.env
@@ -6,3 +6,4 @@ dashboard-url=https://www.redhat.com
 section-title=Open Data Hub
 gateway-domain=
 images-jobs-async-upload=quay.io/opendatahub/model-registry-job-async-upload:latest
+core-bff-image=quay.io/opendatahub/odh-core-bff:main
--- a/manifests/rhoai/base/deployment.yaml
+++ b/manifests/rhoai/base/deployment.yaml
@@ -18,5 +18,5 @@
   path: /spec/template/spec/serviceAccount
   value: rhods-dashboard
 - op: replace
-  path: /spec/template/spec/containers/3/env/2/value
+  path: /spec/template/spec/containers/4/env/2/value
   value: "rhods-dashboard"
--- a/manifests/rhoai/kustomization.yaml
+++ b/manifests/rhoai/kustomization.yaml
@@ -57,6 +57,13 @@ vars:
       apiVersion: v1
     fieldref:
       fieldpath: data.images-jobs-async-upload
+  - name: core-bff-image
+    objref:
+      kind: ConfigMap
+      name: rhoai-dashboard-params
+      apiVersion: v1
+    fieldref:
+      fieldpath: data.core-bff-image
 configurations:
   - params.yaml
 replacements:
--- a/manifests/rhoai/params.env
+++ b/manifests/rhoai/params.env
@@ -6,3 +6,4 @@ dashboard-url=https://www.redhat.com
 section-title=OpenShift AI
 gateway-domain=
 images-jobs-async-upload=quay.io/opendatahub/model-registry-job-async-upload:latest
+core-bff-image=quay.io/opendatahub/odh-core-bff:main


manifests/modular-architecture/deployment.yaml
```diff
@@ -709,3 +709,81 @@
...
+      - '--cert-file=/etc/tls/private/tls.crt'
+      - '--key-file=/etc/tls/private/tls.key'
+      - '--bundle-paths=/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem,/var/run/secrets/kubernetes.io/serviceaccount/ca.crt,/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt,/etc/pki/tls/certs/odh-ca-bundle.crt,/etc/pki/tls/certs/odh-trusted-ca-bundle.crt'
+    env:
+      - name: POD_NAMESPACE
+        valueFrom:
+          fieldRef:
+            fieldPath: metadata.namespace
+    securityContext:
+      allowPrivilegeEscalation: false
```
liday-rh: **Potential duplicate `core-bff` container (Correctness)**

This patch appends a `core-bff` container via `op: add, path: /spec/template/spec/containers/-`. However, this PR also adds a `core-bff` container directly in `manifests/core-bases/base/deployment.yaml` (index 2).

Since `modular-architecture/kustomization.yaml` uses `core-bases/base` as a resource and then applies this file as a `patchesJson6902` overlay, the resulting Deployment would have **two containers named `core-bff`** — one from the base and one appended by this patch. Kubernetes requires unique container names within a pod; the API server would reject this.

Both ODH and RHOAI build through the modular-architecture layer, so both would be affected. `kustomize build` itself wouldn't catch this since it doesn't validate K8s schema constraints.

The two specs also differ: this version mounts `modules-sa-token` (projected SA token) while the core-bases version doesn't. If both are intentional for different deployment modes, only the core-bases one should be in the base, and this patch should *replace* the base container rather than append a second one.

Could you clarify the intended container topology?

manifests/modular-architecture/deployment.yaml
```diff
@@ -709,3 +709,81 @@
...
+      - '--auth-token-header=x-forwarded-access-token'
+      - '--auth-token-prefix='
+      - '--port=8943'
+      - '--cert-file=/etc/tls/private/tls.crt'
+      - '--key-file=/etc/tls/private/tls.key'
+      - '--bundle-paths=/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem,/var/run/secrets/kubernetes.io/serviceaccount/ca.crt,/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt,/etc/pki/tls/certs/odh-ca-bundle.crt,/etc/pki/tls/certs/odh-trusted-ca-bundle.crt'
+    env:
+      - name: POD_NAMESPACE
+        valueFrom:
+          fieldRef:
```
liday-rh: **Inconsistent security context hardening (Non-blocking)**

This sidecar variant omits `runAsNonRoot` and `readOnlyRootFilesystem`, which are present in both the core-bases sidecar and the standalone module deployment. This is consistent with existing mod-arch sidecars (model-registry-ui, gen-ai-ui, etc.), so it follows the established pattern — but since core-bff is a backend service with access to secrets and impersonation, consider adding these fields for defense-in-depth.

manifests/modular-architecture/modules/core-bff/networkpolicy.yaml
```diff
@@ -0,0 +1,43 @@
...
+        - port: 6443
+          protocol: TCP
+    - to:
+        - ipBlock:
+            cidr: 0.0.0.0/0
+      ports:
+        - port: 443
+          protocol: TCP
+        - port: 80
+          protocol: TCP
```
liday-rh: **Egress allows plain HTTP (Non-blocking)**

Port 80 (plain HTTP) egress to `0.0.0.0/0` is notable for a BFF that should primarily communicate with internal K8s services and HTTPS endpoints. Is this needed for HTTP→HTTPS redirects or specific external endpoints? If not, consider removing the port 80 rule to reduce attack surface.

liday-rh (changes requested):
## Review — Security, Quality & Efficiency

Overall the security posture is strong — hardened containers, TLS throughout, scoped RBAC, and restricted NetworkPolicy. One potential correctness issue needs clarification before merge.

### What's good

- **Container hardening** (standalone module): `automountServiceAccountToken: false`, explicitly projected SA token with 1-hour expiry, `seccompProfile: RuntimeDefault`, read-only root filesystem, non-root, all capabilities dropped. Best-practice.
- **TLS throughout**: port 8943 HTTPS, TLS cert/key from `proxy-tls` secret, CA bundles mounted.
- **RBAC**: The new ClusterRole is actually **narrower** than the existing `odh-dashboard` ClusterRole (no delete on secrets, no watch on configmaps). The `impersonate` permission matches the existing `modules-cluster-role.yaml` pattern.
- **RHOAI index fix** (`containers/3` → `containers/4`): Verified correct — inserting core-bff at index 2 shifts mod-arch appended containers by one.
- **HA**: standalone module has `replicas: 2` with pod anti-affinity across zones.
- Kustomize var references for `core-bff-image` are consistent across all three overlays (ODH, RHOAI, modular-architecture).

manifests/modular-architecture/deployment.yaml
lucferbux: no worries, this is not needed, was a bad first pass

manifests/modular-architecture/modules/core-bff/networkpolicy.yaml
lucferbux: no worries, this is not needed, was a bad first pass

manifests/core-bases/base/deployment.yaml
```diff
@@ -144,6 +144,75 @@ spec:
...
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.namespace
+          resources:
+            requests:
+              cpu: 300m
+              memory: 512Mi
+              ephemeral-storage: 10Mi
+            limits:
+              memory: 512Mi
```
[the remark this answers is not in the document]
lucferbux: No, k8s best practices now recommend not adding cpu limit.

DaoDaoNoCode:
Hey @lucferbux — I'm working on the Test Connection feature ([#8121](https://github.com/opendatahub-io/odh-dashboard/pull/8121)) which adds a `POST /core-bff/api/v1/connections/test` endpoint. When testing on-cluster, I noticed the Fastify backend uses `proxyService` entries in the federation ConfigMap to route API requests to sidecar containers.

I see this PR adds the core-bff container and service port, but I didn't spot a corresponding `proxyService` entry in `federation-configmap.yaml`. Do you plan to add that in this PR or as a follow-up? Something like:

```json
{
  "name": "coreBff",
  "proxyService": [{
    "authorize": true,
    "path": "/core-bff/api",
    "pathRewrite": "/api",
    "tls": true,
    "service": {
      "name": "rhods-dashboard",
      "namespace": "redhat-ods-applications",
      "port": 8943
    }
  }]
}
```

Without this, the frontend's `/core-bff/api/*` requests get a 404 from Fastify since there's no proxy route registered. Just want to make sure it's tracked — let me know if I misunderstood anything here, thanks!

lucferbux:
Closing this in favor of https://github.com/opendatahub-io/odh-dashboard/pull/8708 @DaoDaoNoCode I'm gonna add a comment explaining how to address your issue, and how to set up a cluster (in the PR description) to test the whole test suite.

lucferbux:
Hey @DaoDaoNoCode — your finding was exactly right. The `proxyService` entry for `coreBff` was missing, which meant `/core-bff/api/*` requests hit Fastify with no registered route and got a 404. Fixed in #8708 (which supersedes this PR).

Here's what was added and how to test it:

---

### What changed (in #8708)

**federation ConfigMaps** (both ODH and RHOAI) — added the `coreBff` `proxyService` entry you described:
```json
{
  "name": "coreBff",
  "proxyService": [{
    "authorize": true,
    "path": "/core-bff/api",
    "pathRewrite": "/api",
    "tls": true,
    "service": { "name": "odh-dashboard", "namespace": "opendatahub", "port": 8943 }
  }]
}
```

**NetworkPolicies** — port 8943 added to the sidecar ingress allowlist; egress to `odh-dashboard:8943` added to gen-ai and maas standalone module pods.

**`dashboard-operator/internal/controller/module_deploy.go`** — the controller's `buildFederationConfigMap` always emits the `coreBff` entry in standalone mode. Modules listed in `modulesWithCoreBFFAccess` (`gen-ai`, `maas`) get `BFF_CORE_BFF_SERVICE_NAME` and `BFF_CORE_BFF_SERVICE_PORT` env vars injected automatically.

**`docs/inter-bff-communication.md`** — updated diagrams for both deployment modes; added "Calling core-bff from a Module BFF" section.

---

### How to test the proxy route

```bash
# Port-forward the Fastify backend
oc port-forward svc/odh-dashboard 8080:8443 -n opendatahub

# Hit a core-bff endpoint through the Fastify proxy
curl -sk -H "x-forwarded-access-token: $(oc whoami -t)" \
  https://localhost:8080/core-bff/api/v1/status
# Should return a JSON status object from core-bff, not a 404
```

### How to enable inter-BFF calls (e.g. for Test Connection #8121)

Use `BFF_CORE_BFF_SERVICE_NAME` / `BFF_CORE_BFF_SERVICE_PORT` as service coordinates — injected automatically in standalone mode. For local dev: `BFF_CORE_BFF_DEV_URL=http://localhost:8943/api go run cmd/main.go`.

Full onboarding guide in the updated `docs/inter-bff-communication.md` (once #8708 merges).

PR #8549, opened 2026-07-14
fix(llmd-serving): wire topology config selection into deployment ass…

YuliaKrimerman:
https://issues.redhat.com/browse/RHOAIENG-66855
https://issues.redhat.com/browse/RHOAIENG-76493

## Description

Fixes two related bugs in the llm-d topology config deployment flow:

### [RHOAIENG-66855](https://redhat.atlassian.net/browse/RHOAIENG-66855) — Topology config selection silently dropped during deployment

The `CustomTopologyConfigField` stores the user's selected topology configuration in the wizard state, but no `WizardFieldApplyExtension` existed to map it into the deployment resource. This meant the `selectedConfig` was never applied — no `LLMInferenceServiceConfig` was created alongside the `LLMInferenceService`, `baseRef` was never set, and the operator defaulted to its generic config (CUDA image). This is the root cause of AMD hardware profiles getting the wrong vLLM image.

**Fix:** Added `applyTopologyConfigData` / `extractTopologyConfigData` functions and registered them as `WizardFieldApplyExtension` and `WizardFieldExtractorExtension` in the llmd-serving extensions, following the same pattern as the existing gateway field extensions. When a topology config is selected, it is cloned as the deployment's `LLMInferenceServiceConfig` with the correct `baseRef` and template annotations. When "Single node (default)" is selected, the server and baseRef are cleared.

### [RHOAIENG-76493](https://redhat.atlassian.net/browse/RHOAIENG-76493) — Auto-select doesn't persist when switching topology types

When switching from "Single node" to another topology type (e.g. "Multi-node"), the auto-select `useEffect` fired correctly but the selection was immediately overwritten by the wizard framework's dependency-change reset cycle (`shouldResetOnDependencyChange` → `clearFieldData` + `initFieldData`).

**Fix:** Moved auto-selection logic from the component's `useEffect` into `getInitialFieldData`, which is called during the `initFieldData` dispatch. The `getInitialFieldData` signature already supports `externalData` and `dependencies` parameters — we now use them to auto-select the first available config for non-single-node topologies.

## How Has This Been Tested?
for [RHOAIENG-66855](https://redhat.atlassian.net/browse/RHOAIENG-66855) 
Screenshot 2026-07-14 at 2 35 14 PM
Screenshot 2026-07-14 at 2 37 01 PM
Screenshot 2026-07-14 at 2 37 36 PM
Screenshot 2026-07-14 at 2 39 27 PM

For [RHOAIENG-76493](https://redhat.atlassian.net/browse/RHOAIENG-76493)

https://github.com/user-attachments/assets/a85ac462-b4c2-4285-bdfe-cb9a9be35bdf

- **Unit tests**: 11 new tests covering apply (9 cases) and extract (2 cases) — config selection, annotation preservation, server clearing, scheduler behavior, immutability, config switching
- **All existing tests pass**: 165/165 tests across 15 suites in `llmd-serving`
- **Type-check**: passes cleanly
- **Lint**: 0 errors (8 pre-existing warnings unrelated to this change)
- **Manual UI testing** on a live cluster:
  - Verified selecting a single-node topology config creates `LLMInferenceServiceConfig` with correct `baseRef` and `template-name` annotation
  - Verified "Single node (default)" does NOT create an `LLMInferenceServiceConfig`
  - Verified switching topology types auto-selects the first available config (persists in dropdown)
  - Verified switching back to single-node resets to "Single node (default)"

## Test Impact

**Added:** `packages/llmd-serving/src/wizardFields/__tests__/topologyConfigApplyExtract.spec.ts`
- 9 test cases for `applyTopologyConfigData` (set server/baseRef, preserve annotations, clear on undefined, scheduler behavior, immutability, config switching)
- 2 test cases for `extractTopologyConfigData` (extract server, return undefined when no server)

## Request review criteria:

[RHOAIENG-66855]: https://redhat.atlassian.net/browse/RHOAIENG-66855?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[RHOAIENG-76493]: https://redhat.atlassian.net/browse/RHOAIENG-76493?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Added support for applying and extracting custom topology configurations in wizard fields.
  * Custom topology configuration fields now initialize with the first available configuration for applicable topology types.

* **Bug Fixes**
  * Improved initial configuration selection while preserving existing selections during edits.
  * Prevented configuration selection from changing unexpectedly after loading completes.

### DIFF
--- a/packages/llmd-serving/extensions/extensions.ts
+++ b/packages/llmd-serving/extensions/extensions.ts
@@ -299,7 +299,9 @@ const extensions: (
   | WizardFieldApplyExtension<AdvancedRoutingFieldData, LLMdDeployment>
   | WizardFieldExtractorExtension<AdvancedRoutingFieldData, LLMdDeployment>
   | WizardFieldApplyExtension<GatewaySelectFieldData, LLMdDeployment>
+  | WizardFieldApplyExtension<CustomTopologyConfigFieldData, LLMdDeployment>
   | WizardFieldExtractorExtension<GatewaySelectFieldData, LLMdDeployment>
+  | WizardFieldExtractorExtension<CustomTopologyConfigFieldData, LLMdDeployment>
   | WizardFieldExtractorExtension<{ method: string }, LLMdDeployment>
   | HrefNavItemExtension
   | RouteExtension
@@ -500,6 +502,8 @@ const extensions: (
   gatewaySelectFieldExtension,
   gatewaySelectApplyExtension,
   gatewaySelectExtractorExtension,
+  topologyConfigApplyExtension,
+  topologyConfigExtractorExtension,
   deploymentMethodExtractorExtensionLllmdOnly,
   deploymentMethodExtractorExtensionvLLMOnMaaS,
   {
--- a/packages/llmd-serving/extensions/extensions.ts
+++ b/packages/llmd-serving/extensions/extensions.ts
@@ -299,9 +299,7 @@ const extensions: (
   | WizardFieldApplyExtension<AdvancedRoutingFieldData, LLMdDeployment>
   | WizardFieldExtractorExtension<AdvancedRoutingFieldData, LLMdDeployment>
   | WizardFieldApplyExtension<GatewaySelectFieldData, LLMdDeployment>
-  | WizardFieldApplyExtension<CustomTopologyConfigFieldData, LLMdDeployment>
   | WizardFieldExtractorExtension<GatewaySelectFieldData, LLMdDeployment>
-  | WizardFieldExtractorExtension<CustomTopologyConfigFieldData, LLMdDeployment>
   | WizardFieldExtractorExtension<{ method: string }, LLMdDeployment>
   | HrefNavItemExtension
   | RouteExtension
@@ -502,8 +500,6 @@ const extensions: (
   gatewaySelectFieldExtension,
   gatewaySelectApplyExtension,
   gatewaySelectExtractorExtension,
-  topologyConfigApplyExtension,
-  topologyConfigExtractorExtension,
   deploymentMethodExtractorExtensionLllmdOnly,
   deploymentMethodExtractorExtensionvLLMOnMaaS,
   {
--- a/packages/llmd-serving/src/wizardFields/CustomTopologyConfigField.tsx
+++ b/packages/llmd-serving/src/wizardFields/CustomTopologyConfigField.tsx
@@ -251,13 +251,17 @@ export const CustomTopologyConfigFieldWizardField: CustomTopologyConfigFieldType
         return existingFieldData;
       }
       const topologyType = dependencies?.topologyType?.topologyType;
-      if (topologyType && topologyType !== TopologyType.SINGLE_NODE) {
+      if (!topologyType) {
+        return { selectedConfig: TOPOLOGY_CONFIG_DEFAULT };
+      }
+      if (topologyType !== TopologyType.SINGLE_NODE) {
         const configs = getFilteredConfigs(externalData?.configsByTopology, topologyType);
         if (configs.length > 0) {
           return { selectedConfig: configs[0] };
         }
+        return { selectedConfig: undefined };
       }
-      return { selectedConfig: undefined };
+      return { selectedConfig: TOPOLOGY_CONFIG_DEFAULT };
     },
     validationSchema: z.object({
       selectedConfig: z.union([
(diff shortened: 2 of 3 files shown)


packages/llmd-serving/extensions/extensions.ts
```diff
@@ -500,6 +502,8 @@ const extensions: (
   gatewaySelectFieldExtension,
   gatewaySelectApplyExtension,
   gatewaySelectExtractorExtension,
+  topologyConfigApplyExtension,
+  topologyConfigExtractorExtension,
```
mturley: These `topologyConfigApplyExtension` and `topologyConfigExtractorExtension` lines are both already present in this array and this adds duplicates.

packages/llmd-serving/extensions/extensions.ts
```diff
@@ -299,7 +299,9 @@ const extensions: (
   | WizardFieldApplyExtension<AdvancedRoutingFieldData, LLMdDeployment>
   | WizardFieldExtractorExtension<AdvancedRoutingFieldData, LLMdDeployment>
   | WizardFieldApplyExtension<GatewaySelectFieldData, LLMdDeployment>
+  | WizardFieldApplyExtension<CustomTopologyConfigFieldData, LLMdDeployment>
   | WizardFieldExtractorExtension<GatewaySelectFieldData, LLMdDeployment>
+  | WizardFieldExtractorExtension<CustomTopologyConfigFieldData, LLMdDeployment>
```
mturley: Same for these 2 new types, they’re already elsewhere in this union. We can just reset this whole file to how it is on main.

packages/llmd-serving/src/wizardFields/CustomTopologyConfigField.tsx
```diff
@@ -249,7 +235,21 @@ export const CustomTopologyConfigFieldWizardField: CustomTopologyConfigFieldType
     setFieldData: (value: CustomTopologyConfigFieldData) => value,
     getInitialFieldData: (
```
mturley, line 245: We should consider adding unit tests for this `getInitialFieldData` logic. It can be similar to how that same mechanism is tested for another field in `packages/llmd-serving/src/wizardFields/__tests__/LlmConfigOptionsField.spec.ts`.

Full suggestion from Claude:

> Look at the existing pattern in `packages/llmd-serving/src/wizardFields/__tests__/LlmConfigOptionsField.spec.ts` — it tests `getInitialFieldData` by pulling it directly off the wizard field's `reducerFunctions`:
> 
> ```typescript
> const { getInitialFieldData } = LLMConfigOptionsFieldWizardField.reducerFunctions;
> ```
> 
> For `CustomTopologyConfigField`, a test file like `__tests__/CustomTopologyConfigField.spec.ts` would do the same:
> 
> ```typescript
> import { CustomTopologyConfigFieldWizardField } from '../CustomTopologyConfigField’;
> import { TopologyType } from '../../types’;
> 
> const { getInitialFieldData } = CustomTopologyConfigFieldWizardField.reducerFunctions;
> ```
> 
> Then test the three branches:
> 
> 1. **Existing data passthrough** — when `existingFieldData` is provided, it returns it unchanged (edit flow preservation):
> ```typescript
> const existing = { selectedConfig: mockConfig };
> expect(getInitialFieldData(existing, someExternalData, someDeps)).toBe(existing);
> ```
> 
> 2. **Auto-select for non-single-node** — when no existing data, topology type is e.g. `MULTI_NODE`, and configs exist, returns `{ selectedConfig: configs[0] }`:
> ```typescript
> const externalData = { configsByTopology: { [TopologyType.MULTI_NODE]: [configA, configB] } };
> const deps = { topologyType: { topologyType: TopologyType.MULTI_NODE } };
> expect(getInitialFieldData(undefined, externalData, deps)).toEqual({ selectedConfig: configA });
> ```
> 
> 3. **Single-node returns undefined** — when topology type is `SINGLE_NODE`, returns `{ selectedConfig: undefined }`:
> ```typescript
> const deps = { topologyType: { topologyType: TopologyType.SINGLE_NODE } };
> expect(getInitialFieldData(undefined, externalData, deps)).toEqual({ selectedConfig: undefined });
> ```
> 
> You'd also want edge cases: no configs available for the topology type (empty array), undefined dependencies, undefined external data. The mock configs would need `metadata.name` and `metadata.annotations` to match `LLMInferenceServiceConfigKind` — look at existing mock data in the `__tests__` directory or `__mocks__` for the shape.
>

mturley (changes requested):
A couple things

mturley (approved):
/approve
/lgtm

mturley (approved):
/approve
/lgtm

PR #8114, opened 2026-06-15
[Feature] Add "Existing secret" option to workbench env vars

bobbravo2:
## Description
Screenshot 2026-06-15 at 6 01 50 PM

Screenshot 2026-06-15 at 6 02 03 PM

Screenshot 2026-06-15 at 6 00 07 PM

Implements the "Existing secret" feature for workbench environment variables — all 5 stories under Epic [RHOAIENG-60680](https://redhat.atlassian.net/browse/RHOAIENG-60680), covering all 7 [RHAISTRAT-1699](https://redhat.atlassian.net/browse/RHAISTRAT-1699) acceptance criteria.

**What it does:**
- Adds "Existing secret" as a new env var type in the workbench creation/edit form
- Multi-select typeahead dropdown lists all Opaque secrets in the namespace, excluding Connection secrets and dashboard-managed secrets
- Per-secret key picker with "All keys" toggle, "N of M keys" badge, and individual key checkboxes
- RBAC permission check — shows info alert when user cannot list secrets
- On save, writes `env[].valueFrom.secretKeyRef` entries into the Notebook CR (never `envFrom`)
- Edit flow parses existing `secretKeyRef` entries from the Notebook CR and displays them as opaque references
- Cross-source conflict detection warns in real-time when env var key names collide across existing secret refs, inline secrets, upload secrets, and connections
- Secret values never reach browser memory — hook strips values and returns only key names
- Description text and restart info per UXD guidance

**Architecture decisions (staff-engineer-validated):**
- Always `secretKeyRef`, never `envFrom`, even for "All keys" — each key is explicitly enumerated
- Secret values never rendered in UI or stored in React state
- Unlabeled secret list query fires only when "Existing secret" is selected (lazy load)
- Connection secrets filtered by `opendatahub.io/connection-type-protocol` or `opendatahub.io/connection-type-ref` annotations
- Dashboard-created secrets filtered by `opendatahub.io/dashboard: 'true'` label
- No new BFF endpoints — uses existing `/api/k8s` passthrough

**[RHAISTRAT-1699](https://redhat.atlassian.net/browse/RHAISTRAT-1699) Acceptance Criteria coverage:**

| # | Criterion | Status |
|---|-----------|--------|
| 1 | "Existing secret" 3rd option in env var type dropdown | Done |
| 2 | Dropdown lists Opaque secrets excluding Connections, by name only | Done |
| 3 | "All keys" → Notebook CR has `secretKeyRef` for each key | Done |
[...]

PR #8564, opened 2026-07-15
feat(spawner): reference existing Kubernetes Secrets as environment variables (RHOAIENG-60680)

bobbravo2:
This is a second eval of https://github.com/opendatahub-io/odh-dashboard/pull/8114 leveraging https://github.com/DaoDaoNoCode/odh-dashboard/tree/prototype-skills from @DaoDaoNoCode  🙇🏾‍♂️ 

## Happy Path
Screenshot 2026-07-20 at 6 37 10 PM

## Description

Enables data scientists to reference pre-existing Kubernetes Secrets by name when configuring workbench environment variables, instead of requiring manual re-entry of values. This eliminates the need to duplicate secret values or bypass the dashboard by editing Notebook CR YAML directly.

**UX prototype:** https://rhoai-rachel-64723e.pages.redhat.com/develop-train/workbenches/create#create-wb-section-env

### What's included

**Data layer ([RHOAIENG-69120](https://redhat.atlassian.net/browse/RHOAIENG-69120)):**
- `SecretCategory.EXISTING` enum value, `ExistingSecretRef` and `ExistingSecretMetadata` types
- `useExistingSecrets(namespace)` hook with RBAC check, dashboard/connection secret filtering, value stripping

**Creation flow — dropdown ([RHOAIENG-69121](https://redhat.atlassian.net/browse/RHOAIENG-69121)):**
- "Existing secret" radio option under Secret variable type
- Typeahead multi-select dropdown with secret name, key count, key preview
- Empty namespace and RBAC disabled states with popovers

**Creation flow — key picker + save ([RHOAIENG-69122](https://redhat.atlassian.net/browse/RHOAIENG-69122)):**
- Per-key expandable sections with select all/deselect all, individual checkboxes
- Key filter for secrets with >10 keys
- Creation path generates `env[].valueFrom.secretKeyRef` entries (never `envFrom`)
- Rotation hint HelperText

**Edit flow ([RHOAIENG-69123](https://redhat.atlassian.net/browse/RHOAIENG-69123)):**
- Parses `secretKeyRef` entries from Notebook CR `env[]` array
- Reconstructs form state grouping by secret name
- Handles mixed `envFrom[]` + `secretKeyRef` save
- `SecretKeyRefEnvVar` type on `StartNotebookData`

**Conflict detection + error states ([RHOAIENG-69124](https://redhat.atlassian.net/browse/RHOAIENG-69124)):**
- Cross-secret key collision detection with warning Alert
- Per-key warning triangle icons on colliding checkboxes
- Deleted secret: danger Alert with "Remove this reference" action
- Missing keys: warning Alert with "Remove missing keys" action

## How Has This Been Tested?

- TypeScript type-check passing
- ESLint passing (zero warnings)
- 95 unit tests covering all new components, hooks, utilities, and service layer changes
- Not manually tested on cluster yet — draft PR for early review

## Test Impact

95 new unit tests across 7 test files:
- `useExistingSecrets.spec.ts` — 14 tests (hook filtering, RBAC, edge cases)
- `useNotebookEnvVariables.spec.ts` — 13 tests (secretKeyRef parsing, grouping)
- `service.spec.ts` — 8 tests (getSecretKeyRefEnvVars helper)
- `EnvExistingSecretField.spec.tsx` — 23 tests (dropdown, states, collisions)
- `ExistingSecretKeyPicker.spec.tsx` — 22 tests (key picker, deleted/missing, collision icons)
- `existingSecretCollisions.spec.ts` — 9 tests (collision detection utility)
- `getSecretKeyRefEnvVars.spec.ts` — 7 tests (secretKeyRef generation)

Cypress mock tests are not yet added — will be added in a follow-up before merge.

## Request review criteria:

If you have UI changes:

[RHOAIENG-69120]: https://redhat.atlassian.net/browse/RHOAIENG-69120?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Added “existing secret” environment-variable support for notebook spawners, including RBAC-aware secret browsing/search, per-secret key selection, missing/deleted key handling, and collision warnings.
  * Updated the workbench environment-variable editor to use radio-based selectors with improved descriptions and restart behavior.

* **Bug Fixes**
  * Notebook environment updates now preserve operator-injected entries while keeping existing-secret `secretKeyRef` env vars correctly merged and updated.

* **Tests**
  * Expanded Jest and Cypress coverage for parsing, validation, collision detection, and the new editor/secret selection flows.


bobbravo2:
Opened up a ffmpeg infra bug here: https://redhat.atlassian.net/browse/RHOAIENG-77017

frontend/src/pages/projects/screens/spawner/environmentVariables/useNotebookEnvVariables.tsx
```diff
@@ -8,11 +8,46 @@ import {
...
+      value?: string;
+      valueFrom?: { secretKeyRef?: { name: string; key: string } };
+    }[]) ?? [];
+  const secretKeyRefMap = new Map<string, string[]>();
+
+  envList.forEach((entry) => {
+    const ref = entry.valueFrom?.secretKeyRef;
+    if (ref) {
+      const keys = secretKeyRefMap.get(ref.name) || [];
+      keys.push(ref.key);
```
DaoDaoNoCode, line 42: **🔴 Data Loss — Aliased env var names silently lost on edit round-trip**

`parseSecretKeyRefEntries` stores only `ref.key` (the secret data key) here, discarding `entry.name` (the env var name). Then `getSecretKeyRefEnvVars` (service.ts:165) rebuilds with `name: key`.

**Concrete scenario:** A Notebook has:
```yaml
env:
  - name: DATABASE_PASSWORD          # ← the env var name
    valueFrom:
      secretKeyRef:
        name: my-secret
        key: password                # ← only this is stored
```
After edit+save, it becomes `{name: "password", ...}`. The alias `DATABASE_PASSWORD` is silently lost. Any application code referencing `$DATABASE_PASSWORD` breaks.

**Root cause:** `ExistingSecretRef` stores `selectedKeys: string[]` (just key names), but the K8s `secretKeyRef` model allows `name` (env var name) to differ from `key` (secret data key).

**Fix:** Change `ExistingSecretRef.selectedKeys` from `string[]` to `Array<{key: string; envVarName: string}>`, preserving the original env var name per key. Then use the stored name in `getSecretKeyRefEnvVars` instead of `name: key`.

frontend/src/api/k8s/notebooks.ts
```diff
@@ -313,6 +315,10 @@ export const updateNotebook = (
   oldNotebook.spec.template.spec.volumes = [];
   container.volumeMounts = [];
 
+  // Clear old env array so the assembled notebook's env (including secretKeyRef entries)
+  // replaces rather than merges by index
+  container.env = [];
```
DaoDaoNoCode, line 324: **🔴 Data Integrity — `container.env = []` clears operator-injected env entries**

This line unconditionally wipes the entire `env[]` array before the lodash merge. While the motivation is correct (preventing merge-by-index corruption now that `secretKeyRefEnvVars` makes the array variable-length), it also deletes entries injected by admission webhooks or operators — e.g., `valueFrom.fieldRef` (Downward API for pod name/namespace), `valueFrom.resourceFieldRef`, proxy settings, or other platform tooling entries.

**Scenario:** An operator injects `{name: "POD_NAME", valueFrom: {fieldRef: {fieldPath: "metadata.name"}}}`. User edits any workbench property (even just storage). After save, the operator's env var is permanently removed. The notebook container loses pod identity on next restart.

**Note:** The same clearing pattern exists for `envFrom`, `volumes`, and `volumeMounts` (lines 306-313), but those arrays are fully dashboard-managed. The `env[]` array is NOT — operators and webhooks inject entries into it.

**Fix:** Instead of clearing the entire array, preserve entries that are not dashboard-managed. For example:
```typescript
// Keep only entries not managed by the dashboard
container.env = (container.env ?? []).filter(
  (e) => e.valueFrom && !e.valueFrom.secretKeyRef && e.name !== "NOTEBOOK_ARGS" && e.name !== "JUPYTER_IMAGE"
);
```
Or use a spread/assign instead of `_.merge` for the env array specifically.

frontend/src/pages/projects/screens/spawner/environmentVariables/EnvironmentVariables.tsx
```diff
@@ -1,47 +1,90 @@
...
+  };
+
+  const getInlineKeyNames = (excludeIndex: number): Set<string> => {
+    const keys = new Set<string>();
+    envVariables.forEach((v, idx) => {
+      if (idx === excludeIndex) {
+        return;
+      }
+      const cat = v.values?.category;
+      if (cat === SecretCategory.GENERIC || cat === ConfigMapCategory.GENERIC) {
```
DaoDaoNoCode: **🟡 Incomplete — Cross-surface collision detection excludes Connections and Upload env vars**

`getInlineKeyNames` only collects keys from `SecretCategory.GENERIC` and `ConfigMapCategory.GENERIC`. It misses:
- **Connection secret keys** — Connections are injected as env vars via `envFrom` and will collide with existing-secret keys at pod runtime
- **Upload env var keys** (`SecretCategory.UPLOAD`, `ConfigMapCategory.UPLOAD`)

RHOAIENG-69124 AC6 explicitly requires collision detection "across selected secrets, inline env vars, **and Connections**."

**Scenario:** User creates a Connection with key `DB_PASSWORD`, then adds an existing secret reference also selecting key `DB_PASSWORD`. No collision warning is shown, but at runtime the pod has duplicate env var names (last-writer-wins based on array order).

**Fix:** Include Connection keys in the collision set. The `SpawnerPage` has `notebookConnections` state — pass connection key names through to `EnvironmentVariables` or compute them in `spawnerUtils.ts:isEnvVariableDataValid`.

frontend/src/pages/projects/screens/spawner/environmentVariables/useExistingSecrets.ts
```diff
@@ -0,0 +1,114 @@
...
+    .filter(
+      (secret) =>
+        (secret.type ?? 'Opaque') === 'Opaque' &&
+        !isDashboardSecret(secret) &&
+        !isConnectionSecret(secret),
+    )
+    .map((secret) => ({
+      name: secret.metadata.name,
+      keys: secret.data
+        ? Object.keys(secret.data).filter((k) => /^[A-Za-z_][A-Za-z0-9_]*$/.test(k))
```
DaoDaoNoCode: **🟡 UX Gap — C_IDENTIFIER regex silently drops valid Kubernetes secret keys**

This regex filters keys to valid C identifiers (`/^[A-Za-z_][A-Za-z0-9_]*$/`). Kubernetes secret keys can legally contain dots and hyphens — these are common in secrets produced by cert-manager (`tls.crt`, `tls.key`), External Secrets Operator (`api-key`), and Helm charts (`config.yaml`).

**The filtering intent is correct** — env var names must be valid identifiers. But the silent drop creates two problems:

1. **Confusing UX:** A secret with keys `["DB_HOST", "app.port", "api-key"]` appears with only 1 selectable key. The user wonders where the other keys went.

2. **Zero-key edge case:** If ALL keys of a secret fail the regex (e.g., a TLS secret with only `tls.crt` and `tls.key`), the secret appears in the dropdown with `keys: []`. The user can see it but cannot select any keys — no explanation given. The `getKeyPreview()` helper would show "(0 keys)" in the description.

**Recommendations** (pick one):
- Filter out secrets that end up with 0 valid keys from the dropdown entirely
- Show a tooltip on the filtered secret explaining why some keys are unavailable
- Add a test documenting this intentional behavior (currently no test covers keys with dots/hyphens)

frontend/src/pages/projects/screens/spawner/service.ts
```diff
@@ -143,6 +147,27 @@ const getEnvFromList = (
...
+ */
+export const getSecretKeyRefEnvVars = (envVariables: EnvVariable[]): SecretKeyRefEnvVar[] =>
+  envVariables
+    .filter((v) => v.values?.category === SecretCategory.EXISTING)
+    .flatMap((v) =>
+      (v.existingSecretRefs ?? []).flatMap((ref) =>
+        ref.selectedKeys
+          .filter((key) => !RESERVED_ENV_NAMES.has(key))
+          .map((key) => ({
+            name: key,
```
DaoDaoNoCode: **🟡 Data Loss (linked to alias finding on useNotebookEnvVariables.tsx:40)**

This is where the env var name is reconstructed as `name: key` — using the secret data key as the env var name. For newly created entries this is correct (the user selected the key, expects it as the env var name). But for **round-tripped** entries (loaded from an existing Notebook CR via `parseSecretKeyRefEntries`), the original env var name was already discarded, so `key` is the only value available.

Also note: the `RESERVED_ENV_NAMES` filtering here (line 163) is correct but has **zero test coverage** in `getSecretKeyRefEnvVars.spec.ts`. None of the 7 tests verify that `NOTEBOOK_ARGS` or `JUPYTER_IMAGE` keys are filtered out. A secret with ONLY reserved keys would show as selectable in the UI but produce 0 env vars on save — silent failure.

frontend/src/pages/projects/screens/spawner/environmentVariables/EnvSecret.tsx
```diff
@@ -1,50 +1,93 @@
...
+              onUpdate={(newEnvData) => onUpdate({ ...env, data: newEnvData })}
+              translateValue={(value) => atob(value)}
+            />
+          ),
+        },
+        [SecretCategory.EXISTING]: {
+          label: 'Existing secret',
+          description:
+            'Attach an available secret from this project. Use Existing Secrets to attach secrets managed by your platform team or provisioned through external tools. For reusable credentials like S3 or database connections, use the Connections section.',
+          isDisabled: existingDisabled,
```
DaoDaoNoCode, line 79: **🟡 AC Gap — Disabled "Existing secret" radio missing info icon**

The radio is correctly disabled when RBAC denies access (`isDisabled: existingDisabled` at line 77), but there is no `labelIcon` provided for the `SecretCategory.EXISTING` option. The `EnvDataTypeOption` type supports `labelIcon` (see `EnvDataTypeField.tsx:9`), and the Radio component renders it next to the label — the infrastructure exists but is unused here.

RHOAIENG-69121 AC7 and the UXD scenario 5b call for "disabled with an info icon" to explain WHY the option is unavailable.

**Fix:** Add a `labelIcon` with a Popover or Tooltip explaining the RBAC requirement:
```tsx
labelIcon: existingDisabled ? (
  <Popover bodyContent="You need 'list' permission on secrets in this project to use existing secrets.">
    <Icon status="info"><OutlinedQuestionCircleIcon /></Icon>
  </Popover>
) : undefined,
```

frontend/src/pages/projects/screens/spawner/environmentVariables/useExistingSecrets.ts
```diff
@@ -0,0 +1,114 @@
...
+const isDashboardSecret = (secret: SecretKind): boolean =>
+  secret.metadata.labels?.[KnownLabels.DASHBOARD_RESOURCE] === 'true';
+
+/**
+ * Returns true if the secret is a Connection resource
+ * (has connection-type-protocol or connection-type-ref annotations).
+ */
+const isConnectionSecret = (secret: SecretKind): boolean =>
+  !!secret.metadata.annotations?.['opendatahub.io/connection-type-protocol'] ||
+  !!secret.metadata.annotations?.['opendatahub.io/connection-type-ref'];
```
DaoDaoNoCode: **🟡 Minor — `isConnectionSecret` diverges from canonical `isConnection()` utility**

This local connection detection checks:
- `opendatahub.io/connection-type-protocol` annotation
- `opendatahub.io/connection-type-ref` annotation

The canonical `isConnection()` in `packages/k8s-core/src/connectionTypeUtils.ts:48-53` checks:
- `opendatahub.io/connection-type` annotation (no `-protocol` suffix)
- `opendatahub.io/connection-type-ref` annotation
- AND requires `KnownLabels.DASHBOARD_RESOURCE` label

Differences:
1. This hook checks `connection-type-protocol` but the canonical utility checks `connection-type` (different annotation name)
2. A connection secret with only the `opendatahub.io/connection-type` annotation (without `-ref` or `-protocol`) would pass through this filter and appear as a selectable "existing secret"

**Suggestion:** Consider reusing `isConnection()` from `@odh-dashboard/k8s-core` (already imported in this file's sibling `useNotebookEnvVariables.tsx`) for consistency, and add the `-protocol` check as an additional filter if needed.

frontend/src/pages/projects/screens/spawner/environmentVariables/existingSecretCollisions.ts
```diff
@@ -0,0 +1,48 @@
...
+  existingSecretRefs: ExistingSecretRef[],
+  externalKeyNames?: Set<string>,
+): KeyCollision[] => {
+  const keySourceMap = new Map<string, string[]>();
+
+  existingSecretRefs.forEach((ref) => {
+    ref.selectedKeys.forEach((key) => {
+      const sources = keySourceMap.get(key);
+      if (sources) {
+        sources.push(ref.secretName);
```
DaoDaoNoCode: **ℹ️ Minor — Collision source list does not deduplicate same-secret references**

`keySourceMap` uses `string[]` for sources and checks `sources.length > 1`. If the same secret appears in `existingSecretRefs` twice (e.g., user adds `secret-a` in two env var rows and selects key `FOO` in both), `sources` becomes `["secret-a", "secret-a"]`. Since `sources.length > 1`, a collision is reported: "FOO is defined in both secret-a and secret-a" — a false positive.

**In practice this is architecturally unreachable** — `usedSecretNames` in `EnvironmentVariables.tsx` disables already-used secrets in other rows. But defensively, using `Set<string>` and checking `size > 1` would be cleaner:

```typescript
const keySourceMap = new Map<string, Set<string>>();
// ...sources.add(ref.secretName) instead of sources.push(...)
// ...if (sources.size > 1 || hasExternalCollision)
```

frontend/src/pages/projects/screens/spawner/environmentVariables/EnvironmentVariables.tsx
```diff
@@ -1,47 +1,90 @@
...
+        v.values?.data.forEach((entry) => {
```
DaoDaoNoCode, line 40: **🟡 Defensive — Missing optional chaining on `.data` before `.forEach`**

`v.values?.data.forEach(...)` — if `v.values` is truthy but `v.values.data` is `undefined` (e.g., malformed state or partial initialization), this will throw `TypeError: Cannot read properties of undefined (reading 'forEach')`.

The optional chain stops at `v.values?` but `.data` is accessed without `?.`. CodeRabbit also flagged this (unresolved).

**Fix:** `v.values?.data?.forEach(...)` — one character.

frontend/src/pages/projects/screens/spawner/environmentVariables/__tests__/existingSecretCollisions.spec.ts
```diff
@@ -0,0 +1,85 @@
+import { ExistingSecretRef } from '#~/pages/projects/types';
+import {
+  detectExistingSecretKeyCollisions,
+  getCollidingKeySet,
+} from '#~/pages/projects/screens/spawner/environmentVariables/existingSecretCollisions';
+
+describe('detectExistingSecretKeyCollisions', () => {
+  it('should return empty array when no refs are provided', () => {
+    expect(detectExistingSecretKeyCollisions([])).toEqual([]);
```
DaoDaoNoCode, line 9: **🟡 Test Gap — `externalKeyNames` parameter has zero test coverage**

All 9 tests in this file call `detectExistingSecretKeyCollisions(refs)` without the `externalKeyNames` parameter. The cross-surface collision code exists in the source (`existingSecretCollisions.ts:32-36`) but is completely untested:

Missing tests:
1. Existing secret key collides with inline env var key → should return collision with source `"another variable"`
2. Multiple external collisions
3. External collision + inter-secret collision simultaneously
4. `externalKeyNames` is empty `Set` → should have no effect

The `getSecretKeyRefEnvVars.spec.ts` also has no tests for `RESERVED_ENV_NAMES` filtering (`NOTEBOOK_ARGS` and `JUPYTER_IMAGE` keys being excluded).

These are the two highest-priority test gaps in the PR — the code paths exist and affect correctness, but have no verification.

frontend/src/pages/projects/screens/spawner/spawnerUtils.ts
```diff
@@ -344,15 +345,38 @@ export const isEnvVariableDataValid = (envVariables: EnvVariable[]): boolean =>
...
+  });
+
+  const isValid = envVariables.every((envVar) => {
+    if (!envVar.type || !envVar.values || !envVar.values.category) {
+      return false;
+    }
+    if (envVar.values.category === SecretCategory.EXISTING) {
+      const refs = envVar.existingSecretRefs ?? [];
+      return refs.length > 0 && refs.some((ref) => ref.selectedKeys.length > 0);
+    }
```
DaoDaoNoCode, line 371: **ℹ️ Design Question — `some()` vs `every()` for ref validation**

This validation requires at least one ref to have selected keys (`refs.some(ref => ref.selectedKeys.length > 0)`), meaning a row with `[{secretName: "a", selectedKeys: []}, {secretName: "b", selectedKeys: ["KEY"]}]` is valid.

The collision-review agent confirmed this is intentional and tested (`spawnerUtils.spec.ts` has an explicit test for this behavior). However, please verify that `getSecretKeyRefEnvVars` correctly filters out refs with empty `selectedKeys` before building the pod spec — otherwise a "valid" row could produce an empty-keys ref that does nothing silently.

Tracing the save path: `getSecretKeyRefEnvVars` uses `.flatMap(ref => ref.selectedKeys.filter(...).map(...))` — empty `selectedKeys` correctly produce no env vars. ✅ No issue here, just documenting the verification.

DaoDaoNoCode:
## 10-Agent Review Summary — PR #8564

Ran 10 specialized review agents in parallel covering: JIRA ACs, data integrity, hooks, create flow, edit flow, collision detection, test coverage, type safety, Cypress tests, and security. Each finding below was cross-validated against the actual source code.

---

### 🔴 Blockers (3)

| # | Finding | File | Evidence |
|---|---------|------|----------|
| **B1** | Aliased env var names silently lost on edit round-trip | `useNotebookEnvVariables.tsx:40` + `service.ts:165` | `keys.push(ref.key)` discards `entry.name`; `getSecretKeyRefEnvVars` rebuilds with `name: key` |
| **B2** | `container.env = []` clears operator-injected env entries | `notebooks.ts:320` | Wipes `valueFrom.fieldRef`, `resourceFieldRef`, etc. injected by webhooks/operators |
| **B3** | Zero Cypress mock tests for the entire feature | All tickets | PR body confirms "will be added in a follow-up" but every JIRA ticket requires them |

### 🟡 Should Fix (5)

| # | Finding | File | Evidence |
|---|---------|------|----------|
| **S1** | Cross-surface collision excludes Connections | `EnvironmentVariables.tsx:34` | `getInlineKeyNames` only checks `GENERIC` categories; [RHOAIENG-69124](https://redhat.atlassian.net/browse/RHOAIENG-69124) AC6 requires Connections |
| **S2** | C_IDENTIFIER regex silently drops valid K8s keys (dots/hyphens) | `useExistingSecrets.ts:46` | Secrets with all-invalid keys appear with "0 keys" — no explanation |
| **S3** | `RESERVED_ENV_NAMES` filtering has zero test coverage | `getSecretKeyRefEnvVars.spec.ts` | None of 7 tests verify NOTEBOOK_ARGS/JUPYTER_IMAGE exclusion |
| **S4** | RBAC disabled radio missing info icon (AC9 gap) | `EnvSecret.tsx:77` | `labelIcon` infrastructure exists but unused |
| **S5** | `isConnectionSecret` diverges from canonical `isConnection()` | `useExistingSecrets.ts:21` | Checks `connection-type-protocol` not `connection-type` |

### ℹ️ Non-blocking (3)

| # | Finding | File |
|---|---------|------|
| **N1** | Collision `sources` uses `string[]` not `Set<string>` (architecturally unreachable false positive) | `existingSecretCollisions.ts:23` |
| **N2** | Missing `?.` on `v.values?.data.forEach` | `EnvironmentVariables.tsx:35` |
| **N3** | `secretKeyRef.optional` dropped on round-trip (rare in practice) | `useNotebookEnvVariables.tsx:29` |

### ✅ What's Working Well

- **Security**: Secret values never exposed, RBAC properly gated, no XSS risks
- **Create flow**: PF v6 usage correct, radio buttons with `useId()`, dropdown accessible
- **Delete/missing detection**: Deleted secrets and missing keys both correctly detected with alerts
- **Save paths**: Both CREATE and EDIT paths structurally sound
- **Type system**: No `any` types, `SecretKeyRefEnvVar` matches K8s spec, props flow correctly
- **Unit tests**: 95 tests with strong coverage of hooks and components
- **Namespace isolation**: Stale data cleared, race conditions handled via cleanup flag

### JIRA Coverage

| Ticket | Status |
|--------|--------|
| [RHOAIENG-69120](https://redhat.atlassian.net/browse/RHOAIENG-69120) (Data layer) | ✅ **Fully covered** (8/8 ACs) |
| [RHOAIENG-69121](https://redhat.atlassian.net/browse/RHOAIENG-69121) (Dropdown) | 7/9 — missing Cypress + RBAC info icon |
| [RHOAIENG-69122](https://redhat.atlassian.net/browse/RHOAIENG-69122) (Key picker) | 9/10 — missing Cypress |
| [RHOAIENG-69123](https://redhat.atlassian.net/browse/RHOAIENG-69123) (Edit flow) | 7/8 — missing Cypress |
| [RHOAIENG-69124](https://redhat.atlassian.net/browse/RHOAIENG-69124) (Conflicts) | 8/10 — missing Cypress + Connection collisions |

---

_Review generated by 10 parallel Claude Code agents: jira-review, data-integrity, hook-existing-secrets, create-flow, edit-flow, collision-review, test-coverage, type-safety, cypress-review, security-review_

[RHOAIENG-69124]: https://redhat.atlassian.net/browse/RHOAIENG-69124?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

frontend/src/api/k8s/notebooks.ts
```diff
@@ -313,6 +315,17 @@ export const updateNotebook = (
   oldNotebook.spec.template.spec.volumes = [];
   container.volumeMounts = [];
 
+  // Preserve operator-injected env entries (fieldRef, resourceFieldRef, configMapKeyRef)
+  // while clearing dashboard-managed entries to prevent merge-by-index corruption
+  container.env = container.env.filter(
```
DaoDaoNoCode: **🔴 NEW — Fix for B2 introduces merge-by-index corruption when operator entries exist**

The filter-instead-of-clear approach interacts badly with `_.merge` on line 330. Lodash merges arrays **by index**, so preserved operator entries at indices 0..N get deep-merged with `assembleNotebook` entries (`NOTEBOOK_ARGS`, `JUPYTER_IMAGE`, secretKeyRefs) at the same indices.

I verified this with the actual lodash in this repo:
```javascript
_.merge({}, 
  { env: [{name: "POD_NAME", valueFrom: {fieldRef: {fieldPath: "metadata.name"}}}] },
  { env: [{name: "NOTEBOOK_ARGS", value: "--ServerApp.port=8888"}] }
).env[0]
// => {name: "NOTEBOOK_ARGS", value: "--ServerApp.port=8888", valueFrom: {fieldRef: {fieldPath: "metadata.name"}}}
```

The result has **both** `value` AND `valueFrom.fieldRef` — a corrupted entry. K8s uses `valueFrom` when both are present, so `NOTEBOOK_ARGS` would get the pod name instead of Jupyter server arguments. The notebook would fail to start.

**Fix:** Extract preserved entries before clearing, then append after merge:
```typescript
const preservedEntries = (container.env ?? []).filter(
  (entry) => entry.valueFrom && !entry.valueFrom.secretKeyRef &&
    (entry.valueFrom.fieldRef || entry.valueFrom.resourceFieldRef || entry.valueFrom.configMapKeyRef),
);
container.env = [];

// ... existing merge ...
const merged = _.merge({}, oldNotebook, notebook);
merged.spec.template.spec.containers[0].env.push(...preservedEntries);
return k8sUpdateResource<NotebookKind>(applyK8sAPIOptions({model: NotebookModel, resource: merged}, opts));
```

This preserves operator entries WITHOUT index-based merge corruption.

frontend/src/pages/projects/screens/spawner/environmentVariables/EnvironmentVariables.tsx
```diff
@@ -1,47 +1,95 @@
...
+        return;
+      }
+      const cat = v.values?.category;
+      if (
+        cat === SecretCategory.GENERIC ||
+        cat === SecretCategory.UPLOAD ||
+        cat === ConfigMapCategory.GENERIC ||
+        cat === ConfigMapCategory.UPLOAD
+      ) {
+        v.values?.data.forEach((entry) => {
```
DaoDaoNoCode, line 40: **🟡 Still present — Optional chaining on `.data` not actually applied**

Commit 46fdf0836 message says "Add optional chaining on .data?.forEach to prevent TypeError" but the code still reads:
```typescript
v.values?.data.forEach((entry) => {
```

The `?.` stops at `v.values?` — if `v.values` is truthy but `v.values.data` is `undefined`, this throws `TypeError: Cannot read properties of undefined (reading 'forEach')`.

Same pattern exists in `spawnerUtils.ts:356`.

**Fix (one char each):** `v.values?.data?.forEach(...)` in both files.

frontend/src/pages/projects/screens/spawner/service.ts
```diff
@@ -143,6 +147,32 @@ const getEnvFromList = (
     return [...acc, envFrom];
   }, initialList);
 
+const RESERVED_ENV_NAMES = new Set(['NOTEBOOK_ARGS', 'JUPYTER_IMAGE']);
+
+export const isValidEnvVarName = (name: string): boolean => /^[A-Za-z_][A-Za-z0-9_]*$/.test(name);
```
DaoDaoNoCode, line 152: **🟡 UX regression — Invalid keys now visible and selectable but silently dropped on save**

The commit removes the C_IDENTIFIER filter from `fetchExistingSecrets` (so keys like `tls.crt` and `api-key` now appear in the key picker) and moves validation to `getSecretKeyRefEnvVars` via `isValidEnvVarName`.

This creates a worse UX gap than before:
- **Before:** Invalid keys were hidden from the dropdown — user couldn't select them
- **After:** Invalid keys are visible, selectable, and appear checked in the picker — but silently dropped on save

A user selects `tls.crt` from their cert-manager secret, sees it checked, saves the workbench — the key is not in the Notebook CR. No warning shown.

**For aliased round-trip entries** (e.g., `name: "MY_CERT", key: "tls.crt"`), this works correctly because `keyEnvNameMap` maps `tls.crt` → `MY_CERT`, and `MY_CERT` passes `isValidEnvVarName`. ✅

**For newly selected keys** without aliases, `envName = key`, so `tls.crt` fails validation and is dropped. ❌

**Recommendation:** Show a warning icon/tooltip on keys that fail `isValidEnvVarName` in the `ExistingSecretKeyPicker`, similar to how collision warnings are shown. Or disable the checkbox for invalid keys with explanation text.

DaoDaoNoCode:
## Re-Review After Fix Commits (4d82f53, 46fdf08, b7f4291)

Verified all 11 original findings against the 3 new commits. **One new critical issue found in the B2 fix.**

---

### Original Finding Status

| # | Finding | Status | Notes |
|---|---------|--------|-------|
| **B1** | Aliased env var names lost | ✅ **Fixed** | `keyEnvNameMap` stores original name when `entry.name !== ref.key`; `getSecretKeyRefEnvVars` uses it |
| **B2** | `container.env = []` clears operator entries | ⚠️ **Fix introduces new bug** | See NEW finding below |
| **S1** | Cross-surface collision excludes Connections | ⬜ **Acknowledged** | `UPLOAD` categories added; Connections still excluded (may be intentional scope) |
| **S2** | C_IDENTIFIER regex silent drop | ⚠️ **Partial — new UX gap** | Filter removed from display, moved to save-time — keys now selectable but silently dropped |
| **S3** | RESERVED_ENV_NAMES untested | ✅ **Covered** | `isValidEnvVarName` now also validates env names at save time |
| **S4** | RBAC info icon missing | ✅ **Fixed** | Popover with "Access permissions needed" header + clear instructions |
| **S5** | `isConnectionSecret` divergence | ✅ **Fixed** | Now uses canonical `isConnection()` from `@odh-dashboard/k8s-core` |
| **N1** | Collision sources use `string[]` | ✅ **Fixed** | Now `Set<string>` with `sources.size > 1` |
| **N2** | Missing `?.` on `.data.forEach` | ❌ **NOT fixed** | Commit message claims it but code still has `v.values?.data.forEach` (both files) |
| **N3** | `optional` flag dropped | ⬜ **Not addressed** | Acknowledged as low priority |
| **Test** | `externalKeyNames` zero coverage | ✅ **Fixed** | 3 new tests: overlap, no overlap, combined collisions |

### 🔴 NEW Critical Finding: Merge-by-index corruption

The B2 fix replaces `container.env = []` with a `.filter()` that preserves operator entries. However, `_.merge({}, oldNotebook, notebook)` on line 330 merges arrays **by index**. Preserved operator entries at indices 0-N get deep-merged with `NOTEBOOK_ARGS`/`JUPYTER_IMAGE` at the same indices, creating corrupted entries with both `value` AND `valueFrom`.

Verified with actual lodash in this repo — the merged entry has `{name: "NOTEBOOK_ARGS", value: "...", valueFrom: {fieldRef: ...}}`. K8s uses `valueFrom` over `value`, so `NOTEBOOK_ARGS` gets the pod name instead of Jupyter server args. **The notebook would fail to start.**

**Fix:** Extract → clear → merge → append:
```typescript
const preservedEntries = (container.env ?? []).filter(/* existing filter */);
container.env = [];
const merged = _.merge({}, oldNotebook, notebook);
merged.spec.template.spec.containers[0].env.push(...preservedEntries);
```

### 🟡 NEW Minor Finding: C_IDENTIFIER validation gap

Moving the C_IDENTIFIER check from display-time to save-time means keys like `tls.crt` are now **visible and selectable** in the key picker but **silently dropped on save**. This is worse UX than before (hidden = can't select > visible but silently ignored). Recommend showing a warning icon on non-C_IDENTIFIER keys in the picker.

### Summary: 7 of 11 findings properly fixed, 2 new issues introduced, 2 not addressed

The alias fix (B1) is clean and correct. The collision dedup, `isConnection()`, RBAC info icon, and test coverage fixes are all good. The `container.env` fix (B2) needs the extract-clear-merge-append pattern, and the optional chaining (N2) still needs the one-character fix.

DaoDaoNoCode:
## CI Failure Analysis — 3 tests failing, 2 files

All 3 failures are directly caused by the fix commits. Easy to resolve.

---

### Failure 1 & 2: `useExistingSecrets.spec.ts` — connection secret tests (commit 46fdf08)

**Tests:**
- `should filter out connection secrets with connection-type-ref annotation`
- `should filter out connection secrets with connection-type-protocol annotation`

**Root cause:** The fix switched from local `isConnectionSecret()` to canonical `isConnection()` from `@odh-dashboard/k8s-core`. The test mocks create secrets with only `opendatahub.io/connection-type-ref` or `opendatahub.io/connection-type-protocol` annotations, but the canonical `isConnection()` checks for:
- `opendatahub.io/connection-type` annotation (different name — no `-ref` or `-protocol` suffix)
- AND requires `KnownLabels.DASHBOARD_RESOURCE` label

The mock secrets lack both the correct annotation name and the dashboard label, so `isConnection()` returns `false` and the secrets pass through the filter. Result: 2 items instead of 1.

**Fix:** Update the mock secrets to match what `isConnection()` actually detects:
```typescript
createMockSecret({
  name: "connection-secret",
  annotations: { "opendatahub.io/connection-type": "s3" },
  labels: { "opendatahub.io/dashboard": "true" },
  data: { endpoint: "..." },
})
```

---

### Failure 3: `getSecretKeyRefEnvVars.spec.ts` — multiple secrets test (commit 46fdf08)

**Test:** `should handle multiple secrets with multiple keys`

**Root cause:** The fix added `isValidEnvVarName` validation at save time in `getSecretKeyRefEnvVars`. The test uses a secret with key `api-key` (contains a hyphen). The new `isValidEnvVarName` regex (`/^[A-Za-z_][A-Za-z0-9_]*$/`) rejects `api-key` because hyphens are not valid in C identifiers.

Expected 3 env entries, got 2 — `api-key` was correctly filtered out by the new validation, but the test assertion wasn't updated to expect this.

**Fix:** Either:
- (a) Change the test's mock key from `api-key` to `api_key` (valid C identifier)
- (b) Update the assertion to expect only 2 entries and add a separate test verifying `api-key` is filtered out

Option (b) is better — it documents the new validation behavior.

---

### Summary

| Test | Cause | Fix |
|------|-------|-----|
| connection-type-ref filter | Mock doesn't match canonical `isConnection()` check | Update mock annotations + add dashboard label |
| connection-type-protocol filter | Same | Same |
| multiple secrets with keys | `api-key` now rejected by `isValidEnvVarName` | Update mock key or assertion |

All 3 are test fixture updates — no production code changes needed.

frontend/src/pages/projects/screens/spawner/environmentVariables/EnvironmentVariables.tsx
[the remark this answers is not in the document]
bobbravo2: Claude responds 

> Skipping — not a valid issue. The type EnvVariableData.data is typed as EnvVariableDataEntry[] (non-optional, line 121 of types.ts). The v.values?. optional chain already handles the case where values is undefined — if it short-circuits, .data.forEach never executes. If values IS defined, .data is guaranteed to be an array by the type contract.

DaoDaoNoCode:
## Final Review — After All Fix Commits (7 commits total)

CI is green ✅ (162 passing, 0 failures). Re-reviewed all code, comment history, and thread resolution status.

---

### Fix Commit Status

| Commit | What it fixed | Verdict |
|--------|--------------|---------|
| `4d82f53` | B1: aliased env names + B2: operator env entries | ✅ B1 alias fix is clean; ⚠️ B2 has remaining issue |
| `46fdf08` | S5: canonical isConnection, N1: Set dedup, UPLOAD collision, S4: info icon, S2: key filter | ✅ All clean |
| `b7f4291` | externalKeyNames test coverage | ✅ 3 tests added |
| `cf6b55d` | N3: optional flag preservation | ✅ Clean |
| `528fa98` | Test mocks for isConnection + isValidEnvVarName | ✅ CI passing |

### Remaining Issues (2 active, 1 acknowledged)

**🔴 CRITICAL — Merge-by-index corruption in `notebooks.ts:320-328`**

The `.filter()` preserves operator entries in `container.env`, then `_.merge({}, oldNotebook, notebook)` merges by array index — corrupting preserved entries with `NOTEBOOK_ARGS`/`JUPYTER_IMAGE` fields. I verified with lodash in this repo:
```
_.merge(
  {env: [{name:"POD_NAME", valueFrom:{fieldRef:{fieldPath:"metadata.name"}}}]},
  {env: [{name:"NOTEBOOK_ARGS", value:"--ServerApp..."}]}
).env[0]
→ {name:"NOTEBOOK_ARGS", value:"--ServerApp...", valueFrom:{fieldRef:{fieldPath:"metadata.name"}}}
```
K8s uses `valueFrom` over `value` → NOTEBOOK_ARGS gets pod name instead of Jupyter args. **Notebook fails to start.**

Both my comment and CodeRabbit independently flagged this. Fix: extract-clear-merge-append pattern.

**🟡 MINOR — Invalid keys selectable but silently dropped on save**

After removing the display-time C_IDENTIFIER filter, keys like `tls.crt` are now visible and checkable in the picker, but `isValidEnvVarName` drops them during save. Users select keys that silently vanish. Either disable invalid-key checkboxes in the picker with explanation, or show a warning icon.

**ℹ️ ACKNOWLEDGED — Optional chaining on `.data.forEach`**

Author argues the TypeScript type guarantees `data` exists when `values` is present (`EnvVariableData.data` is non-optional). The linter would flag `?.` as unnecessary. Acceptable position.

### CodeRabbit Finding: Multiple aliases for same key

CodeRabbit flagged that `keyEnvNameMap[ref.key] = entry.name` overwrites if two env vars alias the same secret key (e.g., both `DB_PASSWORD` and `DB_PASS_BACKUP` point to `my-secret/password`). The second alias wins, first is silently lost. Valid finding but **extreme edge case** — unlikely in practice.

### My Comment Thread Resolution

| Thread | Status | Reason |
|--------|--------|--------|
| B1: alias data loss | Should resolve ✅ | Fixed in `4d82f53` — `keyEnvNameMap` preserves original names |
| B2: container.env clearing | Outdated, replaced by merge-by-index comment | Original line changed |
| S1: Connection collision | Outdated — UPLOAD added | Connections still excluded but narrower gap |
| S2: C_IDENTIFIER regex | Outdated — filter removed | Replaced by UX regression comment |
| S3: RESERVED_ENV_NAMES + name:key | Outdated — alias fix applied | Now uses `keyEnvNameMap` |
| S4: RBAC info icon | Should resolve ✅ | Fixed in `46fdf08` — Popover added |
| S5: isConnectionSecret | Outdated — canonical used | Now imports from `@odh-dashboard/k8s-core` |
| N1: collision dedup | Outdated — Set used | Fixed in `46fdf08` |
| N2: optional chaining | Acknowledged as design choice | Type system guarantees safety |
| externalKeyNames tests | Should resolve ✅ | 3 tests added in `b7f4291` |
| some() vs every() | Acknowledged as intentional | Tested behavior |
| Merge-by-index corruption | **Still open** 🔴 | Not fixed |
| Optional chaining not applied | Acknowledged | Author dismissed |
| UX regression (invalid keys) | **Still open** 🟡 | Not fixed |

### Bottom Line

The merge-by-index corruption is the **one remaining blocker**. Everything else is either fixed, acknowledged, or minor enough for follow-up. The feature itself (types, hooks, components, collision detection, RBAC, security) is well-built.

packages/cypress/cypress/pages/workbench.ts
```diff
@@ -131,35 +131,22 @@ class EnvironmentVariableTypeField extends Contextual<HTMLElement> {
   }
 
   selectEnvDataType(name: string) {
```
DaoDaoNoCode, line 133: **🔴 E2E test breakage — `selectEnvDataTypeByTestId` and `selectEnvironmentVariableTypeByTestId` will fail**

The e2e tests in `testWorkbenchVariables.cy.ts` use the `ByTestId` page object methods, which now do an exact `findByTestId` match inside the container. But the new radio components use prefixed test IDs that don't match.

**Concrete failures (all 3 e2e test cases):**

```
// E2E test calls:
secretEnvVarField.selectEnvironmentVariableTypeByTestId(EnvironmentVariableType.SECRET);
//   → findByTestId("Secret") inside "environment-variable-type-select"
//   → radio has data-testid="env-type-radio-Secret"
//   → NO MATCH ❌

secretEnvVarField.selectEnvDataTypeByTestId(SecretCategory.UPLOAD);
//   → findByTestId("secret upload") inside "env-data-type-field"
//   → radio has data-testid="env-data-type-radio-secret upload"
//   → NO MATCH ❌
```

Mock tests pass because they use the `ByName` variants (`selectEnvDataType("Key / value")`) which use `findByRole("radio", { name })` — these work fine with the new radio UI.

**Fix — update the page object `ByTestId` methods to prepend the radio prefix:**

```typescript
selectEnvDataTypeByTestId(testId: string) {
  this.find()
    .findByTestId("env-data-type-field")
    .findByTestId(`env-data-type-radio-${testId}`)
    .click();
}

selectEnvironmentVariableTypeByTestId(testId: string) {
  this.find()
    .findByTestId("environment-variable-type-select")
    .findByTestId(`env-type-radio-${testId}`)
    .click();
}
```

bobbravo2: 👏🏻 nice catch!

frontend/src/api/k8s/notebooks.ts
```diff
@@ -313,11 +315,27 @@ export const updateNotebook = (
   oldNotebook.spec.template.spec.volumes = [];
   container.volumeMounts = [];
 
+  // Preserve entries not managed by the dashboard (operator-injected, webhooks, etc.)
+  // Compare by name against the newly assembled notebook's env to avoid duplicating managed vars
+  const dashboardEnvNames = new Set(
+    notebook.spec.template.spec.containers[0].env.map((e) => e.name),
+  );
+  const operatorEnvEntries = container.env.filter((entry) => !dashboardEnvNames.has(entry.name));
+  container.env = [];
```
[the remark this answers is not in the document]
bobbravo2: Skipping — not a valid issue.

container.env comes from oldNotebook.spec.template.spec.containers[0] which is a structuredClone of existingNotebook. The updateNotebook function only runs on edit (existing notebook), and Kubernetes always initializes containers[0].env as at least an empty array on a running notebook. The TypeScript type for Container defines env as non-optional (EnvironmentVariable[]), which is why the linter already rejected ?. and ?? on this path in earlier commits.

If env were somehow undefined at runtime, the notebook would have never started in the first place (it's a required field in the pod spec for notebooks with at least NOTEBOOK_ARGS). No defensive coding needed here

DaoDaoNoCode (approved):
/lgtm
Great work here! 🎉

PR #8578, opened 2026-07-15
Update MaaS API URL to use /v1/tenants

Griffin-Sullivan:
https://redhat.atlassian.net/browse/RHOAIENG-69335
## Description
- Replaced cluster-ingress URL guessing with in-cluster GET /v1/tenants discovery on the internal maas-api Service, using the BFF service account token to resolve the real gateway externalUrl.
- Set passthrough MaasApiUrl to externalUrl + /maas-api; BFF startup fails if discovery fails and MAAS_API_URL is not set.
- Added POD_NAMESPACE (downward API) to the maas-ui deployment to map ODH/RHOAI namespaces to maas-api infra namespaces; removed duplicate TIERS_CONFIGMAP_NS.
- Added optional overrides (MAAS_API_INTERNAL_URL, MAAS_API_NAMESPACE, MAAS_API_URL), unit tests, and updated README/CONTRIBUTING docs.

## How Has This Been Tested?
Tested with the image from this PR set as the image for maas-ui on my cluster and added the env var for POD_NAMESPACE
rolling out new pods shows 
```bash
time=2026-07-16T17:49:11.274Z level=INFO msg="Discovering MaaS API URL via /v1/tenants" internalURL=https://maas-api.redhat-ai-gateway-infra.svc.cluster.local:8443
time=2026-07-16T17:49:11.301Z level=INFO msg="Discovered MaaS API URL from /v1/tenants" tenant=models-as-a-service gateway=maas-default-gateway url=http://<maas-gateway-url>/maas-api
```
The API Keys page and operations were working as well.
Going to try out the PR image built here to see how it's working

## Test Impact
new golang tests

## Request review criteria:

If you have UI changes: 

## Summary by CodeRabbit

* **New Features**
  * MaaS BFF now discovers the MaaS API endpoint via internal tenant discovery (`GET /v1/tenants`) and the `maas-default-gateway` external URL when `MAAS_API_URL` is not set.
  * Added configuration for pod namespace and internal MaaS API lookup/target namespaces; manifests now provide `POD_NAMESPACE`.
  * Extended ClusterRole permissions to list `datadscienceclusters`.

* **Bug Fixes**
  * Startup now stops with a clear error if discovery fails.

* **Documentation**
  * Updated MAAS contributing/development guidance and environment-variable descriptions for the new discovery behavior and settings.

### DIFF
--- a/manifests/modular-architecture/deployment.yaml
+++ b/manifests/modular-architecture/deployment.yaml
@@ -327,6 +327,10 @@
     env:
       - name: GATEWAY_DOMAIN
         value: ""
+      - name: POD_NAMESPACE
+        valueFrom:
+          fieldRef:
+            fieldPath: metadata.namespace
     securityContext:
       allowPrivilegeEscalation: false
       capabilities:
--- a/manifests/modular-architecture/deployment.yaml
+++ b/manifests/modular-architecture/deployment.yaml
@@ -327,10 +327,6 @@
     env:
       - name: GATEWAY_DOMAIN
         value: ""
-      - name: POD_NAMESPACE
-        valueFrom:
-          fieldRef:
-            fieldPath: metadata.namespace
     securityContext:
       allowPrivilegeEscalation: false
       capabilities:
--- a/manifests/modular-architecture/modules-cluster-role.yaml
+++ b/manifests/modular-architecture/modules-cluster-role.yaml
@@ -3,6 +3,13 @@ apiVersion: rbac.authorization.k8s.io/v1
 metadata:
   name: odh-dashboard-modules
 rules:
+  # DSC product type for maas-api infra namespace discovery (maas-ui)
+  - apiGroups:
+      - datasciencecluster.opendatahub.io
+    verbs:
+      - list
+    resources:
+      - datascienceclusters
   # MLflow CR auto-discovery (mlflow-ui, gen-ai-ui)
   - apiGroups:
       - mlflow.opendatahub.io
--- a/packages/maas/bff/internal/api/app.go
+++ b/packages/maas/bff/internal/api/app.go
@@ -139,17 +139,12 @@ func NewApp(cfg config.EnvConfig, logger *slog.Logger) (*App, error) {
 		maasFakeServer = maas.CreateMaasFakeServer()
 		logger.Info("MaaS Fake API Server is running", "url", maasFakeServer.URL)
 		cfg.MaasApiUrl = maasFakeServer.URL
-	} else {
-		// Fallback to discovery of MaaS API url, when not provided via envvar, or cmd flags
-		if cfg.MaasApiUrl == "" {
-			clusterDomain, err := helper.GetClusterDomainUsingServiceAccount(context.Background(), logger)
-			if err != nil {
-				logger.Error("Failed to auto-discover cluster domain, MaaS API URL will be unavailable", "error", err)
-			} else {
-				cfg.MaasApiUrl = fmt.Sprintf("https://maas.%s/maas-api", clusterDomain)
-				logger.Info("Using automatically discovered MaaS URL", "url", cfg.MaasApiUrl)
-			}
+	} else if cfg.MaasApiUrl == "" {
+		discoveredURL, err := helper.DiscoverMaasApiURL(context.Background(), cfg, logger, rootCAs)
+		if err != nil {
+			return nil, fmt.Errorf("failed to discover MaaS API URL via /v1/tenants: %w", err)
 		}
+		cfg.MaasApiUrl = discoveredURL
 	}
 
 	if err != nil {
--- a/packages/maas/bff/internal/helpers/maas_discovery.go
+++ b/packages/maas/bff/internal/helpers/maas_discovery.go
@@ -14,7 +14,11 @@ import (
 	"time"
 
 	"github.com/opendatahub-io/maas-library/bff/internal/config"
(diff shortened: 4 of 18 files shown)


manifests/modular-architecture/deployment.yaml
```diff
@@ -315,6 +315,10 @@
     env:
       - name: GATEWAY_DOMAIN
         value: ""
+      - name: POD_NAMESPACE
+        valueFrom:
+          fieldRef:
+            fieldPath: metadata.namespace
```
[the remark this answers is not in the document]
Griffin-Sullivan: none of the other containers do this so best to do them all together instead

Griffin-Sullivan:
FWIW I confirmed with @jrhyness in Slack to get a review from MaaS backend. He said it looked good to him :+1:

packages/maas/bff/internal/helpers/maas_discovery.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+	if matched == nil {
+		return models.TenantInfo{}, "", fmt.Errorf("maas-api /v1/tenants returned no tenant with gateway %q", maasDefaultGatewayName)
+	}
+
+	externalURL := strings.TrimSpace(matched.Gateway.ExternalURL)
+	if externalURL == "" {
+		return models.TenantInfo{}, "", fmt.Errorf("maas-api /v1/tenants returned empty gateway.externalUrl for %q", maasDefaultGatewayName)
+	}
+
```
katieperry4, line 232: checking for scheme validation in `maas_discovery.go`:
```
No scheme validation on externalUrl 

externalURL := strings.TrimSpace(matched.Gateway.ExternalURL)
if externalURL == "" { ... }
maasApiURL, err := url.JoinPath(externalURL, "maas-api")

If the maas-api ever returns an externalUrl without a scheme (e.g. maas.example.com instead of https://maas.example.com), url.JoinPath silently produces a schemeless string. 

That would fail later at HTTP call time with a confusing error. 

A quick strings.HasPrefix(externalURL, "http") check here would give a much clearer error message.
```

packages/maas/bff/internal/api/app.go
```diff
@@ -139,17 +139,12 @@ func NewApp(cfg config.EnvConfig, logger *slog.Logger) (*App, error) {
...
-			if err != nil {
-				logger.Error("Failed to auto-discover cluster domain, MaaS API URL will be unavailable", "error", err)
-			} else {
-				cfg.MaasApiUrl = fmt.Sprintf("https://maas.%s/maas-api", clusterDomain)
-				logger.Info("Using automatically discovered MaaS URL", "url", cfg.MaasApiUrl)
-			}
+	} else if cfg.MaasApiUrl == "" {
+		discoveredURL, err := helper.DiscoverMaasApiURL(context.Background(), cfg, logger, rootCAs)
+		if err != nil {
+			return nil, fmt.Errorf("failed to discover MaaS API URL via /v1/tenants: %w", err)
```
katieperry4, line 145: If the k8s factory fails, the code still attempts discovery (wasted work), and if discovery also fails, the returned error is about discovery — the original k8s factory error is lost. Moving the if err != nil check for the factory to right after the factory creation block (before the discovery block) would fix the ordering.

```
} else {
    k8sFactory, err = k8s.NewKubernetesClientFactory(cfg, logger)  // outer err set
}

// ---- discovery block inserted by this PR ----
if cfg.MockHTTPClient {
    ...
} else if cfg.MaasApiUrl == "" {
    discoveredURL, err := helper.DiscoverMaasApiURL(...)  // shadows outer err
    if err != nil { return nil, ... }                      // returns discovery error
    cfg.MaasApiUrl = discoveredURL
}

if err != nil {  // checks k8s factory error, but only AFTER discovery ran
    return nil, fmt.Errorf("failed to create Kubernetes client: %w", err)
}

```

katieperry4 (commented):
I'm not super familiar with this side of things but reading through it all seems to make sense to me
I ran it through claude, it mentioned a couple things:

packages/maas/bff/internal/helpers/maas_discovery.go
Griffin-Sullivan: Not sure we want to assume the scheme for them. We should expect this issue to be fixed at maas api level if there's a problem.

packages/maas/bff/internal/api/app.go
Griffin-Sullivan: doesn't seem like new behavior to me so I'll leave it for now for the sake of not having to rebuild and retest everything and risk breaking our sign off cluster but good point!

manifests/modular-architecture/modules-cluster-role.yaml
```diff
@@ -3,6 +3,13 @@ apiVersion: rbac.authorization.k8s.io/v1
 metadata:
   name: odh-dashboard-modules
 rules:
+  # DSC product type for maas-api infra namespace discovery (maas-ui)
+  - apiGroups:
+      - datasciencecluster.opendatahub.io
+    verbs:
+      - list
+    resources:
+      - datascienceclusters
```
andrewballantyne, line 12: For the sake of 3.5 -- please do not reveal this outside of your container in any fashion. We do `.status` ONLY reveals via the NodeJS layer, and I'd like to not do anymore more here.

This works for the time being until we can get our Operator to help.

andrewballantyne (approved):

PR #8579, opened 2026-07-15
fix(ci): replace REST API artifact downloads with actions/download-artifact

christianvogt:
## Description

The `cypress-build-restore` action's restore mode used direct `curl` calls to the GitHub REST API to list and download per-package artifacts. The GITHUB_TOKEN shares a rate limit of 1,000 REST API requests per hour across all concurrent workflow runs in the repo. When 20+ `Cypress-Mock-Tests` matrix jobs each made ~18 API calls to download 9 artifacts, the limit was intermittently exhausted — producing HTTP 403 errors and failing all tests.

This replaces the custom REST API download logic with `actions/download-artifact@v8` which:
- Uses the internal Actions service channel for same-run downloads (zero REST API calls)
- Has built-in retry/backoff and digest verification for cross-run downloads
- Downloads all matching artifacts in one call via `pattern: cypress-build-*`
- Requires no `gh` CLI or REST API calls (Node.js action — works on self-hosted runners)

### Handling the download-artifact v8 single-artifact quirk

`actions/download-artifact@v8` has a documented behavior where patterns matching exactly one artifact extract files directly to the output path **without** creating a per-artifact subdirectory (regardless of `merge-multiple` setting). When multiple artifacts match, each gets its own subdirectory. The restore step relies on these subdirectories to map artifacts back to package workspaces.

To handle this, the action:
1. Downloads **source run** artifacts to `cypress-artifacts/` (typically 8-9 artifacts → subdirectories created)
2. Downloads **current run** (rebuilt) artifacts to a temp directory `cypress-artifacts-current/`
3. A merge step detects whether the temp directory has `cypress-build-*` subdirectories (multiple artifacts) or flat files (single artifact). For the single-artifact case, it identifies the missing package by comparing against the source run's subdirectories and wraps the flat files into the correct subdirectory before merging.

Current-run artifacts take precedence over source-run artifacts since they are freshly rebuilt.

## How Has This Been Tested?

- Tested via `workflow_dispatch` on the `test-cypress-e2e` branch against the self-hosted e2e cluster
- Verified the single-artifact merge handles the `@odh-dashboard/maas` rebuild case (the only package consistently missing from the triggering workflow run)
- Confirmed source-run artifacts (8 packages) download with correct subdirectory structure
- The existing `Verify builds restored` step catches any download/restore failures at runtime

See sample run https://github.com/opendatahub-io/odh-dashboard/actions/runs/29509059330

## Test Impact

No new tests needed — this is CI infrastructure. The existing verify step (`Verify builds restored`) catches any download failures at runtime.

## Request review criteria:

### DIFF
--- a/.github/actions/cypress-build-restore/action.yml
+++ b/.github/actions/cypress-build-restore/action.yml
@@ -123,17 +123,9 @@ runs:
     # Mode: restore — download artifacts via actions/download-artifact
     # Uses internal Actions service for same-run (no REST API rate limits).
     # For cross-run (run-id specified), uses REST API with built-in retry.
-    # Priority: source run wins over current run (download current first,
-    # then source overwrites — matching the original per-package fallback).
+    # Downloads into separate directories then merges so neither download
+    # can clobber the other. Current run wins (has freshly rebuilt packages).
     # =========================================================================
-    - name: Download artifacts from current run
-      if: inputs.mode == 'restore'
-      uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8
-      continue-on-error: true
-      with:
-        pattern: cypress-build-*
-        path: cypress-artifacts
-
     - name: Download artifacts from source run
       if: inputs.mode == 'restore' && inputs.run-id != ''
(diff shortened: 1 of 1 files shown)


antowaddle (approved):
/lgtm

PR #8583, opened 2026-07-15
feat(RHOAIENG-69290): Workbench images admin - filtering, edge cases, and hardening

jesuino:
## Description
Add filtering, tests and safeguards to the OOTB images feature (https://github.com/opendatahub-io/odh-dashboard/pull/8458)

## How Has This Been Tested?

* Manual tests with the Dashboard connected to a Cluster;
* Unit and cypress tests change
* Checking the result with the team;

**Screenshots:**

New FIlters:

image
image

Popup when all images are disabled:

image

## Test Impact
* Unit Tests
* Cypress tests modification

## Request review criteria:

- [X ] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).

If you have UI changes: 

## Summary by CodeRabbit

* **New Features**
  * Added **Type** and **Enabled** select filters to the BYON images toolbar.
  * Added a confirmation modal when attempting to disable the last enabled workbench image.
  * Introduced a unified image enable/disable toggle experience (for both BYON and OOTB images).
* **Bug Fixes**
  * Improved BYON images table filtering with consistent multi-criteria matching (name/provider/type/enabled).
  * Prevents disabling the final effective enabled image without confirmation and restores the toggle state on failures.
* **Refactor**
  * Consolidated the enable/disable toggle logic into a single reusable component.
* **Tests**
  * Added unit tests for the confirmation modal and expanded Cypress coverage for confirm/cancel flows.


frontend/src/pages/BYONImages/BYONImagesToolbar.tsx
```diff
@@ -34,6 +37,28 @@ const BYONImagesToolbar: React.FC<BYONImagesToolbarProps> = ({ filterData, onFil
           onChange={(_event, value) => onChange(value)}
         />
       ),
+      [BYONImagesToolbarFilterOptions.type]: ({ value, onChange, ...props }) => (
+        <SimpleSelect
+          {...props}
+          popperProps={{ maxWidth: undefined }}
+          value={value ?? ''}
```
DaoDaoNoCode, line 44: `SimpleSelect` uses `ariaLabel` (camelCase) as the prop that sets the `MenuToggle` button's accessible name. The hyphenated `aria-label` here is a valid HTML attribute, but it gets spread onto the PF `<Select>` wrapper via `...rest` — it never reaches the toggle button. The button falls back to the default `'Options menu'`.

Same issue on line 56 for the Enabled filter.

```suggestion
          ariaLabel="Filter by type"
```

frontend/src/pages/BYONImages/BYONImagesTable.tsx
```diff
@@ -22,14 +28,34 @@ export const BYONImagesTable: React.FC<BYONImagesTableProps> = ({ images }) => {
...
+
+        if (typeFilter && typeFilter !== ImageTypeFilter.all) {
+          const isRedHat = typeFilter === ImageTypeFilter.redHat;
+          if (image.isOOTB !== isRedHat) {
+            return false;
+          }
+        }
+
+        if (enabledFilter && enabledFilter !== ImageEnabledFilter.all) {
+          const isEnabled = enabledFilter === ImageEnabledFilter.enabled;
```
DaoDaoNoCode, line 53: The filter checks `image.visible` directly, but the toggle's visual state accounts for errors too — `ImageStatusToggle` initializes `isEnabled` as `!isDisabledByError && image.visible`. So a BYON image with `visible: true` and an error (e.g., failed build) passes the "Enabled" filter, but its toggle shows as OFF.

Not critical since it matches the K8s state accurately, but admins could be confused seeing an "Enabled" image with a disabled toggle. Worth considering whether the filter should match the toggle's effective state:
```ts
const isEffectivelyEnabled = image.visible && !(image.error && !image.isOOTB);
```

frontend/src/pages/BYONImages/const.ts
```diff
@@ -1,16 +1,48 @@
...
 
+export enum ImageTypeFilter {
+  all = 'all',
+  redHat = 'redHat',
+  custom = 'custom',
+}
+
+export const imageTypeFilterOptions: SimpleSelectOption[] = [
+  { key: ImageTypeFilter.all, label: 'All' },
+  { key: ImageTypeFilter.redHat, label: 'Red Hat' },
```
DaoDaoNoCode, line 23: The filter uses `isOOTB` to categorize (meaning "pre-installed by the operator, not imported by admin"), but the label says "Red Hat". An OOTB image from a partner (e.g., NVIDIA) would have `isOOTB: true` with `provider: 'NVIDIA'` and would appear under "Red Hat" in the filter.

The table rows already use "Pre-installed" for OOTB images (via `PreInstalledName` in `BYONImagesTableRow.tsx`). Using "Pre-installed" here would be more consistent and accurate:

```suggestion
  { key: ImageTypeFilter.redHat, label: 'Pre-installed' },
```

frontend/src/pages/BYONImages/ImageStatusToggle.tsx
```diff
@@ -0,0 +1,88 @@
...
+        onChange={(_e, value) => handleChange(value)}
+        isDisabled={isDisabledByError || isLoading}
+      />
+      {showConfirmModal && (
+        <DisableLastImageModal
+          imageName={image.display_name}
+          onConfirm={performDisable}
+          onClose={() => setShowConfirmModal(false)}
+          isSubmitting={isLoading}
+        />
```
DaoDaoNoCode, line 82: `performDisable` calls `setShowConfirmModal(false)` then `performToggle(false)`. React 18 batches both state updates — the modal unmounts in the same render that `isLoading` becomes `true`, so `isSubmitting={isLoading}` is never observed as `true` by the user.

The behavior is fine (modal closes optimistically, toggle shows loading), but the `isSubmitting` prop is effectively dead code in the confirm flow. Either:
- Await the toggle before closing: `await performToggle(false); setShowConfirmModal(false);` — to show a spinner in the modal during the API call
- Or remove `isSubmitting` if optimistic close is intentional

frontend/src/pages/BYONImages/const.ts
jesuino: I do understand and I do agree, but this is the requirement from the original JIRA. I added your comment to the JIRA so we can follow the discussion there

jesuino:
Thank you again, @DaoDaoNoCode .

I made the requested changes with the exception of the Provider filter, which I do agree with you, but we need to discuss with others. 

Thanks!

DaoDaoNoCode:
Screenshot 2026-07-20 at 10 45 13 PM


The Type and Enabled filter dropdowns start with "All" chips visible on page load, which is inconsistent with how other pages handle dropdown filters in the dashboard.

**Current behavior:**
The initial filter data sets `type: ImageTypeFilter.all` (`'all'`) and `enabled: ImageEnabledFilter.all` (`'all'`). Since `'all'` is a truthy string, `FilterToolbar` renders chips immediately — "Type: all" and "Enabled: all" — even though no filtering is active. The dropdowns also include an explicit "All" option.

**How other pages handle this:**
- **Hardware Profiles** (`packages/hardware-profiles/src/pages/const.tsx:92-94`): initial values are `undefined`, dropdown has only `Enabled` / `Disabled` — no "All" option. No chip on load. Clearing the chip = show all.
- **Executions** (`frontend/src/pages/pipelines/global/experiments/executions/const.ts:25`): initial `Type` is `undefined`, dropdown lists only the actual type values. No "All" option.
- **Artifacts** (`frontend/src/pages/pipelines/global/experiments/artifacts/constants.ts:11`): same pattern — `undefined` initial, no "All".

**Proposed change:**
1. Remove the "All" option from both `imageTypeFilterOptions` and `imageEnabledFilterOptions`
2. Set initial values to `undefined` instead of `ImageTypeFilter.all` / `ImageEnabledFilter.all`
3. Clearing the chip (clicking X) resets to `undefined`, which already means "show all"
4. Update the filter logic to treat `undefined` / falsy as "no filter" (the `typeFilter && typeFilter !== ImageTypeFilter.all` check simplifies to just `typeFilter`)

jesuino:
@DaoDaoNoCode I do believe it makes sense - at first it seemed inconsistent with the other pages, but since it came from the JIRA, I did keep my criticism low. Thanks! I do remove the all option from the filters

frontend/src/pages/BYONImages/BYONImagesTable.tsx
```diff
@@ -22,14 +28,35 @@ export const BYONImagesTable: React.FC<BYONImagesTableProps> = ({ images }) => {
...
+          const isRedHat = typeFilter === ImageTypeFilter.redHat;
+          if (image.isOOTB !== isRedHat) {
+            return false;
+          }
+        }
+
+        if (enabledFilter) {
+          const isEffectivelyEnabled = image.visible && !(image.error && !image.isOOTB);
+          const isEnabled = enabledFilter === ImageEnabledFilter.enabled;
+          if (isEnabled !== isEffectivelyEnabled) {
```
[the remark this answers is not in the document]
DaoDaoNoCode: @jesuino Can you address this comment? Then I believe it's good to go!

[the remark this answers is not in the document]
jesuino: That's a good catch, this explains why the confirmation dialog behaved weirdly another day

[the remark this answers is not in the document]
jesuino: I made the changes on my most recent commit @DaoDaoNoCode  - thanks again!

DaoDaoNoCode (approved):
/lgtm
Thank you!

PR #8568, opened 2026-07-15
feat: add "What's new in 3.5" guided tour

manaswinidas:
https://issues.redhat.com/browse/RHOAIENG-76170
https://issues.redhat.com/browse/RHOAIENG-75434

## Description

Replaces the 3.4 "What's new" guided tour with a comprehensive 3.5 version that walks users through each navigation section with popovers anchored to nav items (or centered modals when a section is unavailable).

https://github.com/user-attachments/assets/e8b6d24e-1e3c-4c35-b742-ed49ba78a397

**Key changes:**
- **Section-based tour**: Projects, Gen AI studio, Develop & train, AI hub, Observe & monitor, Settings — each step shows a description and "New in 3.5" sub-features with feature flag availability
- **Role-aware messaging**: Admins see which flags to enable in `OdhDashboardConfig`; non-admins see "Contact your administrator"
- **Admin-only Settings step**: Settings section only appears for admin users
- **Welcome + completion modals**: Welcome offers "Start tour" / "What's new in 3.5" / "Skip"; completion links to 3.5 docs
- **Light bulb icon**: Replaces BoltIcon in header with a filled Red Hat light bulb icon (no glow effect)
- **Task Assistant integration**: "Take a guided tour" link in Task Assistant header
- **Auto-launch**: Shows once for users who haven't seen the 3.5 tour (version-keyed localStorage)
- **PF Backdrop**: Uses PatternFly Backdrop component instead of custom inline styles
- **Cleanup**: Removed dead code (`parentSectionTitle`, `sectionFlagName`, `expandNavSectionByTitle`), extracted `DEFAULT_DOC_URL` constant, computed unavailable features once, removed `data-tour` nav label attributes and "New" label functionality

**Feature flags covered:**
AutoML, AutoRAG, Guardrails, Agent Configuration Management, Playground Tracing (genAiTracing — hardcoded false with TODO until flag lands in DashboardConfigKind), Tool Calling, MCP Catalog, Agents Catalog, Observability Dashboard, GPUaaS, MaaS Settings Redesign, MCP Catalog Settings

## How Has This Been Tested?

- Manually tested admin and non-admin views in development
- Verified auto-launch on first visit, "seen" persistence in localStorage, and re-trigger via header button and Task Assistant link
- Verified popover anchoring to nav items and modal fallback for unavailable sections
- Ran unit tests: `cd frontend && npx jest -- src/app/whatsNew/__tests__/WhatsNewModal.spec.tsx`
[...]

PR #8597, opened 2026-07-16
fix: suppress What's New modal in Cypress E2E tests

manaswinidas:
https://issues.redhat.com/browse/RHOAIENG-76170

## Description

The "What's New in 3.5" guided tour modal (merged in #8568) auto-launches 1.5s after page load for users who haven't seen it. In Cypress E2E tests running on the live `dash-e2e-int` cluster, `localStorage` is clean so the modal auto-launches and its `<ModalFooter>` covers page elements — causing `cy.type()` failures in the Create Project test.

The existing suppression in `e2e.ts` only applied inside the `if (Cypress.env('MOCK'))` block. This PR moves the `localStorage.setItem('odh-whats-new-3.5-seen', true)` call to a separate `window:before:load` handler **outside** the MOCK block so it fires for both mock and E2E tests.

## How Has This Been Tested?

- Verified that the `window:before:load` handler fires before every `cy.visit()` in both mock and E2E modes
- All 67 Cypress mock test groups passed on #8568 with the MOCK-scoped version; this change extends coverage to E2E
- Unit tests for `WhatsNewModal` continue to pass (9/9)

## Test Impact

No new tests — this is a test infrastructure fix. The existing Cypress E2E suite validates the fix by no longer having the modal overlay interfere with test interactions.

## Request review criteria:

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented the “What’s New” modal from automatically appearing during end-to-end test sessions.
  * Preserved mock-specific browser setup behavior.

### DIFF
--- a/packages/cypress/cypress/support/e2e.ts
+++ b/packages/cypress/cypress/support/e2e.ts
@@ -109,6 +109,11 @@ Cypress.Keyboard.defaults({
   keystrokeDelay: 0,
 });
 
+// Suppress the "What's New" auto-launch modal in all Cypress tests
+Cypress.on('window:before:load', (win) => {
+  win.localStorage.setItem('odh-whats-new-3.5-seen', JSON.stringify(true));
+});
+
 // Disable polling in mock tests by injecting window globals before app code runs
 if (Cypress.env('MOCK')) {
   Cypress.on('window:before:load', (win) => {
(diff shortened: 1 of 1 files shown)


antowaddle (approved):
/lgtm

PR #8601, opened 2026-07-16
External Models Table Row and Expanded Row Content

katieperry4:
Closes [RHOAIENG-72210](https://redhat.atlassian.net/browse/RHOAIENG-72210)

## Description

adds the row content and expanded row content

In mock mode:

https://github.com/user-attachments/assets/1a6701ec-3c57-4ee4-a51d-4afaa5b8bec9

## How Has This Been Tested?

tested locally

## Test Impact

adding tests in the next commit

## Request review criteria:

If you have UI changes: 

[RHOAIENG-72210]: https://redhat.atlassian.net/browse/RHOAIENG-72210?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Expanded external models table rows with provider details (auth mechanism, credential secret, API format, target model, routing weights) plus row expansion.
  * Added “Resolved path” and “Provider URL” dialogs from expanded rows.
  * Improved phase presentation with governance-pairing warning and richer pending/failed/ready messaging.
  * Added/strengthened keyword filtering and sortable table columns.

* **Bug Fixes**
  * External provider details validation now enforces credential-secret presence; governance-attached state is surfaced consistently.

* **Tests**
  * Updated UI/Cypress and unit tests for filtering, expansion, dialogs, and governance behavior.


packages/maas/bff/internal/models/externalmodel.go
```diff
@@ -28,5 +39,6 @@ type ExternalModelSummary struct {
 	ProviderRefs  []ProviderRef                    `json:"providerRefs"`
 	Phase         string                           `json:"phase,omitempty"`
 	StatusMessage string                           `json:"statusMessage,omitempty"`
+	ConfigStatus  ExternalModelConfigStatus        `json:"configStatus"`
```
Griffin-Sullivan: oh dang yeah we need to get this from the status in the maas model ref

packages/maas/bff/internal/repositories/external_helpers.go
```diff
@@ -141,10 +145,102 @@ func enrichExternalModelSummaries(
...
+			if name == "" || namespace == "" {
+				continue
+			}
+			keys[modelRefKey(namespace, name)] = struct{}{}
+		}
+	}
+	return keys
+}
+
+func listSubscribedModelRefKeys(
```
Griffin-Sullivan: so these lists won't work because it's very likely the user doesn't have access to make that call on MaaSSubscription or MaaSAuthPolicy. What we can do though is look at the GovernancePairing on the MaaSModelRef status. It will tell you if there's no auth policy and subscription. It can't differentiate between missing one or the other it will always show the same output. Either you have both and it's happy or you're failing the check.

packages/maas/bff/internal/repositories/external_helpers.go
```diff
@@ -90,14 +90,15 @@ func buildExternalProviderSummaryIndex(summaries []models.ExternalProviderSummar
...
-		StatusMessage: summary.StatusMessage,
+		DisplayName:         summary.DisplayName,
+		Description:         summary.Description,
+		EndpointUrl:         summary.EndpointUrl,
+		AuthMechanism:       summary.AuthMechanism,
+		CredentialSecretRef: summary.CredentialSecretRef,
+		Provider:            summary.Provider,
+		Config:              summary.Config,
+		Phase:               summary.Phase,
+		StatusMessage:       summary.StatusMessage,
```
Griffin-Sullivan, line 101: I think we are missing the weights here aren't we

packages/maas/frontend/src/app/pages/external-models/ExternalModelsTableRow.tsx
```diff
@@ -1,25 +1,106 @@
...
+  const labelsContainerRef = React.useRef<HTMLDivElement>(null);
+  const [providerURLModalRef, setProviderURLModalRef] = React.useState<ProviderRef | null>(null);
+  const [pathModalRef, setPathModalRef] = React.useState<ProviderRef | null>(null);
+  const [isCopyTipCopied, setIsCopyTipCopied] = React.useState(false);
+
+  const toggleExpanded = () => {
+    setIsExpanded((prev) => !prev);
+  };
+
+  React.useLayoutEffect(() => {
```
Griffin-Sullivan, line 43: What's all of this for?

packages/maas/frontend/src/app/pages/external-models/ExternalModelsTableRow.tsx
```diff
@@ -65,13 +192,58 @@ const ExternalModelTableRow: React.FC<ExternalModelTableRowProps> = ({
...
+        isOpen={!!providerURLModalRef}
+        onClose={() => {
+          setProviderURLModalRef(null);
+          setIsCopyTipCopied(false);
+        }}
+        setIsCopyTipCopied={setIsCopyTipCopied}
+        isCopyTipCopied={isCopyTipCopied}
+        providerURL={providerURLModalRef?.provider?.endpointUrl ?? ''}
+        providerRef={
+          providerURLModalRef?.provider?.displayName ?? providerURLModalRef?.providerName ?? ''
```
Griffin-Sullivan, line 234: I'm guessing there's some mismatching types again since there's so many possible undefined in this file. Let's make sure we are matched up with the CRD for ExternalProvider types in the frontend and BFF

packages/maas/frontend/src/app/pages/external-models/ExternalModelsTableRow.tsx
katieperry4: so in the mocks it only shows two rows of providers, so this makes sure it only shows two rows and if there's more than show, show the 'show more' button

packages/maas/bff/internal/repositories/external_helpers.go
katieperry4: I don't think so? ^Thats `ExternalProviderDetails` and `weight` lives on `ProviderRef` which lives in the `ExternalModelSummary`

Griffin-Sullivan:
Need to add the resource popover to the external model row as well like this
image

packages/maas/frontend/src/app/pages/external-models/ExternalModelsTableRow.tsx
Griffin-Sullivan: ah dang that's annoying

packages/maas/bff/internal/repositories/external_helpers.go
```diff
@@ -123,12 +123,11 @@ func enrichExternalModelSummaries(
 	summaries []models.ExternalModelSummary,
 	providers map[string]models.ExternalProviderSummary,
 	modelRefs map[string]models.MaaSModelRefSummary,
-	subscribedModelRefs map[string]struct{},
-	authorizedModelRefs map[string]struct{},
 ) []models.ExternalModelSummary {
 	for i := range summaries {
 		summary := &summaries[i]
 		modelKey := summary.Namespace + "/" + summary.Name
+		summary.ConfigStatus = models.ExternalModelConfigStatusNoConfig
```
Griffin-Sullivan: I don't think we need this anymore we can just read it off of the maas model ref that's attached to the external model.

Like here under the maas model ref

Image

Griffin-Sullivan:
Also need to map the authentication to a human readable string 
image

packages/maas/bff/internal/repositories/external_helpers.go
katieperry4: ohhhhhh I see

packages/maas/frontend/src/app/pages/external-models/AllExternalModelsPage.tsx
Griffin-Sullivan, line 91: Claude flagged this as an issue. Should be `externalModels.length > 0 && filteredExternalModels.length === 0`

packages/maas/frontend/src/app/pages/external-models/utils.tsx
```diff
@@ -0,0 +1,89 @@
...
+  }
+  if (externalModel.phase === PhaseStatus.READY) {
+    return (
+      <>
+        All networking resources for {modelName} have been created successfully. The HTTPRoute is
+        active and inference requests are being routed to the configured provider(s).
+      </>
+    );
+  }
+  return null;
```
Griffin-Sullivan: Maybe we should have some sort of generic message when there's a status we don't know about?

packages/maas/frontend/src/app/pages/external-models/ExternalModelsTableRow.tsx
Griffin-Sullivan: We should just use `getExternalModelResource` here

frontend/src/components/table/TableRowTitleDescription.tsx
```diff
@@ -51,8 +53,10 @@ const TableRowTitleDescription: React.FC<TableRowTitleDescriptionProps> = ({
       <div data-testid="table-row-title">
         {resource ? (
           <ResourceNameTooltip resource={resource} wrap={wrapResourceTitle}>
-            {title}
+            {boldResourceTitle ? <strong>{title}</strong> : title}
```
Griffin-Sullivan: If we don't bold everywhere else then we should ignore it. That's a problem on the UX mock not being consistent with our app

packages/maas/frontend/src/app/pages/external-models/utils.tsx
```diff
@@ -0,0 +1,89 @@
+import { K8sResourceCommon } from '@odh-dashboard/k8s-core';
+import * as React from 'react';
+import { AuthMechanism, ExternalModel, ProviderRef } from '~/app/types/external-models';
+import { PhaseStatus } from '~/app/utilities/phaseLabelUtils';
+
+/** Ready-condition message on the companion MaaSModelRef when sub+auth pairing is missing. */
+export const AWAITING_GOVERNANCE_PAIRING_MESSAGE = 'Awaiting governance pairing';
+
+export const isAwaitingGovernancePairing = (externalModel: ExternalModel): boolean =>
+  externalModel.maaSModelRef?.statusMessage === AWAITING_GOVERNANCE_PAIRING_MESSAGE;
```
Griffin-Sullivan: So this is one wait to do it. Not sure how reliable it is though. I still think the BFF might want to do a quick check on the maas model ref and check the GovernancePairing condition and still set an extra bool in the model ref part of the external model response

Griffin-Sullivan (commented):
Code looks great! but...

Image

we are missing the dropdown and status?

edit I mean we have the status but it's in the wrong column and provider is missing :joy:

katieperry4:
@Griffin-Sullivan is it possible you're running the branch from my first pr or main? 😅 

there is a test failure though so I'll look at that

packages/maas/bff/internal/models/externalmodel.go
```diff
@@ -13,9 +13,10 @@ type ProviderRef struct {
 
 // ExternalModelMaaSModelRefStatus contains published endpoint details from the companion MaaSModelRef.
 type ExternalModelMaaSModelRefStatus struct {
-	Phase         string `json:"phase,omitempty"`
-	Endpoint      string `json:"endpoint,omitempty"`
-	StatusMessage string `json:"statusMessage,omitempty"`
+	Phase              string `json:"phase,omitempty"`
+	Endpoint           string `json:"endpoint,omitempty"`
+	StatusMessage      string `json:"statusMessage,omitempty"`
+	GovernanceAttached bool   `json:"governanceAttached,omitempty"`
```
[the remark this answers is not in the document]
Griffin-Sullivan: I think that makes sense right?

sridarna:
@katieperry4 I tested it, Saw couple of differences from Mock.
1. The popover button should be aligned with Status. It is visible after the sort button.
2. The copy button is missing for both Provider URL and Path in mock mode.
Screenshot 2026-07-17 at 2 25 10 PM
Screenshot 2026-07-17 at 2 26 59 PM
Screenshot 2026-07-17 at 2 27 24 PM

packages/cypress/cypress/pages/modelsAsAService.ts
```diff
@@ -1539,6 +1539,182 @@ class ExternalModelsPage {
...
+
+  findRows(): Cypress.Chainable<JQuery<HTMLElement>> {
+    return this.findTable().find('[data-testid="external-model-row"]');
+  }
+
+  findFilterInput(): Cypress.Chainable<JQuery<HTMLElement>> {
+    return cy.findByTestId('external-models-filter-input');
+  }
+
+  findFilterResetButton(): Cypress.Chainable<JQuery<HTMLElement>> {
```
sridarna, line 1555: Is it possible to add testId for this?

packages/cypress/cypress/pages/modelsAsAService.ts
```diff
@@ -1539,6 +1539,182 @@ class ExternalModelsPage {
   findExternalModelsTab(): Cypress.Chainable<JQuery<HTMLElement>> {
     return cy.findByTestId('external-models-tab');
   }
+
+  findTable(): Cypress.Chainable<JQuery<HTMLElement>> {
+    return cy.findByTestId('external-models-table');
+  }
+
+  findRows(): Cypress.Chainable<JQuery<HTMLElement>> {
+    return this.findTable().find('[data-testid="external-model-row"]');
```
sridarna: Update to using cy.findByTestId

packages/cypress/cypress/pages/modelsAsAService.ts
```diff
@@ -1539,6 +1539,182 @@ class ExternalModelsPage {
...
+  }
+
+  findFilterResetButton(): Cypress.Chainable<JQuery<HTMLElement>> {
+    return cy.findByRole('button', { name: 'Clear all filters' });
+  }
+
+  getRow(name: string): ExternalModelTableRow {
+    return new ExternalModelTableRow(() =>
+      this.findTable()
+        .find('[data-testid="external-model-row"]')
```
sridarna: we could reuse findRows function

packages/cypress/cypress/pages/modelsAsAService.ts
```diff
@@ -1539,6 +1539,182 @@ class ExternalModelsPage {
...
+        .first()
+        .find('tr')
+        .first(),
+    );
+  }
+}
+
+class ExternalModelTableRow extends TableRow {
+  private scope(): Cypress.Chainable<JQuery<HTMLElement>> {
+    return this.find().parents('[data-testid="external-model-row"]');
```
sridarna: could we use cy.findByTestId here?

katieperry4:
@sridarna 
In the show and tell this morning we decided to remove the copy buttons from those modals!

the status sort itself is a button with the column title, you can see it highlighted here:
Screenshot 2026-07-17 at 2 40 15 PM
I tried to move the order of the popover so the sort button was to the right, which didn't work, I couldn't get it to move and to make that work I'd have to make changes to how Patternfly's `Th` component works

but even it did work then the popover would be within that button and I think the selectability would get weird with a selectable popover within the bounds of the selectable status sort?

packages/cypress/cypress/pages/modelsAsAService.ts
katieperry4: that one's a patternfly component so I can't add an ID for this one 😞 
it goes `ExternalModelsToolBar in MaaS` > `FilterToolbar in UI Core` > `ToolbarToggleGroup patternfly component` > `ToolbarExpandableContent patternfly component`

sridarna:
Everything looks good for me.
/lgtm

Griffin-Sullivan (approved):

PR #8655, opened 2026-07-17
fix/refactor: centralize config field overrides in overrideLlmConfigFields

mturley:
https://issues.redhat.com/browse/RHOAIENG-76001

## Description

Centralizes namespace, apiVersion, kind, labels, annotations, and the dashboard resource label into the shared `overrideLlmConfigFields` utility, so all three LLMInferenceServiceConfig settings pages (accelerator, topology, routing) delegate these concerns to a single function instead of duplicating the same metadata spread pattern at each call site.

**Changes:**
- Rename `ConfigFormFields` to `ConfigFieldOverrides` and expand it to accept `namespace` (required), `labels`, and `annotations`
- `overrideLlmConfigFields` now sets `apiVersion` and `kind` from `LLMInferenceServiceConfigModel` and applies `DASHBOARD_RESOURCE_LABEL` internally
- Caller-provided labels cannot override `DASHBOARD_RESOURCE_LABEL` — it is always applied last
- All three settings pages pass page-specific labels/annotations (e.g. `CONFIG_TYPE_LABEL`, `SUPPORTED_TOPOLOGIES_ANNOTATION`) through the util instead of manually spreading metadata
- Remove now-unused `LLMInferenceServiceConfigModel` and `DASHBOARD_RESOURCE_LABEL` imports from the call sites

This ensures it is not possible to create or update a resource with a different namespace, apiVersion, or kind than the values enforced by the util — any values in the parsed YAML for these fields are overridden.

**Bug fix included:** The accelerator config form was missing the `opendatahub.io/config-type: accelerator` label on create/edit. The list view filters by this label, so newly created configs would not appear in the list. Now fixed by passing `CONFIG_TYPE_LABEL` through the shared util.

Follow-up to PR #8485.

## How Has This Been Tested?

**Unit tests** (20 tests for `overrideLlmConfigFields`):
- Override fields: name, namespace, displayName, description, version, labels, annotations
- Enforcement: apiVersion/kind always set from model, dashboard label always `true` even if caller passes `false`
- Conflict scenarios: input config with wrong namespace, apiVersion, or kind is always overridden
- Edge cases: missing annotations, missing labels, multiple overrides at once

**Browser testing** (verified on dev cluster via Playwright):
- Pasted YAML with `metadata.namespace: some-other-namespace` — resource created in dashboard namespace, not the YAML namespace
- Pasted YAML with `apiVersion: v1` / `kind: ConfigMap` — resource created with correct `serving.kserve.io/v1alpha2` and `LLMInferenceServiceConfig`
- Accelerator configs: create, edit (display name change), duplicate (`-copy` suffix, dashboard label, config-type label)
- Topology configs: create (config-type label set), edit (metadata preserved)
- Routing configs: create and edit flows verified

**All existing llmd-serving unit tests pass** (205 total). Lint and type-check pass with 0 errors.

## Test Impact

- Updated `packages/llmd-serving/src/settings/__tests__/configYamlUtils.spec.ts` — expanded from 10 to 20 test cases:
  - Existing tests for name, displayName, description, version overrides
  - New tests for namespace override, apiVersion/kind from model, dashboard label, labels merge, annotations merge
  - New conflict tests verifying wrong namespace, apiVersion, kind, and labels in the input config are always overridden
  - Regression test for dashboard label enforcement after caller-provided labels

## Request review criteria:

## Summary by CodeRabbit

* **New Features**
  * Improved routing, topology, and accelerator configuration submission by consistently applying names, namespaces, descriptions, labels, and annotations.
  * Configuration resources now receive standardized API version and resource type metadata.

* **Bug Fixes**
  * Improved validation and error messaging for invalid YAML submissions.
  * Preserved configuration details while correctly updating metadata and dashboard visibility labels.

* **Tests**
  * Expanded coverage for metadata overrides, annotations, labels, resource identity, and specification preservation.

### DIFF
--- a/packages/llmd-serving/src/settings/configYamlUtils.ts
+++ b/packages/llmd-serving/src/settings/configYamlUtils.ts
@@ -16,8 +16,7 @@ export const overrideLlmConfigFields = (
   config: LLMInferenceServiceConfigKind,
   overrides: ConfigFieldOverrides,
 ): LLMInferenceServiceConfigKind => {
-  const apiGroup = LLMInferenceServiceConfigModel.apiGroup ?? '';
-  const apiVer = LLMInferenceServiceConfigModel.apiVersion;
+  const { apiGroup, apiVersion: apiVer } = LLMInferenceServiceConfigModel;
 
   const annotations = { ...config.metadata.annotations, ...overrides.annotations };
   if (overrides.displayName !== undefined) {
--- a/packages/llmd-serving/src/settings/configYamlUtils.ts
+++ b/packages/llmd-serving/src/settings/configYamlUtils.ts
@@ -1,35 +1,52 @@
+import { DASHBOARD_RESOURCE_LABEL } from '../const';
+import { LLMInferenceServiceConfigModel } from '../types';
 import type { LLMInferenceServiceConfigKind } from '../types';
 
-export type ConfigFormFields = {
+export type ConfigFieldOverrides = {
   name?: string;
+  namespace: string;
   displayName?: string;
   description?: string;
   version?: string;
+  labels?: Record<string, string>;
+  annotations?: Record<string, string>;
 };
 
 export const overrideLlmConfigFields = (
   config: LLMInferenceServiceConfigKind,
-  fields: ConfigFormFields,
+  overrides: ConfigFieldOverrides,
 ): LLMInferenceServiceConfigKind => {
-  const annotations = { ...config.metadata.annotations };
-  if (fields.displayName !== undefined) {
-    annotations['openshift.io/display-name'] = fields.displayName;
+  const apiGroup = LLMInferenceServiceConfigModel.apiGroup ?? '';
+  const apiVer = LLMInferenceServiceConfigModel.apiVersion;
+
+  const annotations = { ...config.metadata.annotations, ...overrides.annotations };
+  if (overrides.displayName !== undefined) {
+    annotations['openshift.io/display-name'] = overrides.displayName;
   }
-  if (fields.description !== undefined) {
-    annotations['openshift.io/description'] = fields.description;
+  if (overrides.description !== undefined) {
+    annotations['openshift.io/description'] = overrides.description;
   }
-  if (fields.version !== undefined) {
-    if (fields.version) {
-      annotations['opendatahub.io/runtime-version'] = fields.version;
+  if (overrides.version !== undefined) {
+    if (overrides.version) {
+      annotations['opendatahub.io/runtime-version'] = overrides.version;
     } else {
       delete annotations['opendatahub.io/runtime-version'];
     }
   }
+
   return {
     ...config,
+    apiVersion: `${apiGroup}/${apiVer}`,
+    kind: 'LLMInferenceServiceConfig',
     metadata: {
       ...config.metadata,
-      ...(fields.name ? { name: fields.name } : {}),
+      ...(overrides.name ? { name: overrides.name } : {}),
+      namespace: overrides.namespace,
+      labels: {
+        ...config.metadata.labels,
+        ...overrides.labels,
+        [DASHBOARD_RESOURCE_LABEL]: 'true',
+      },
       annotations,
     },
   };
--- a/packages/llmd-serving/src/settings/llmAcceleratorConfigs/LlmAcceleratorConfigAddForm.tsx
+++ b/packages/llmd-serving/src/settings/llmAcceleratorConfigs/LlmAcceleratorConfigAddForm.tsx
@@ -120,9 +120,7 @@ const LlmAcceleratorConfigAddForm: React.FC<LlmAcceleratorConfigAddFormProps> = 
     try {
       parsed = YAML.parse(yamlCode);
     } catch (e) {
-      if (e instanceof Error) {
-        setError(e);
-      }
+      setError(e instanceof Error ? e : new Error(String(e)));
       return;
     }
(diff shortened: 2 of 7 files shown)


packages/llmd-serving/src/settings/configYamlUtils.ts
```diff
@@ -1,35 +1,52 @@
...
     metadata: {
       ...config.metadata,
-      ...(fields.name ? { name: fields.name } : {}),
+      ...(overrides.name ? { name: overrides.name } : {}),
+      namespace: overrides.namespace,
+      labels: {
+        ...config.metadata.labels,
+        [DASHBOARD_RESOURCE_LABEL]: 'true',
+        ...overrides.labels,
+      },
```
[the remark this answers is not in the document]
mturley: Addressed: moved `DASHBOARD_RESOURCE_LABEL` after `overrides.labels` and added a regression test that verifies passing `{ 'opendatahub.io/dashboard': 'false' }` still results in `'true'`.

packages/llmd-serving/src/settings/configYamlUtils.ts
```diff
@@ -1,35 +1,52 @@
...
 
 export const overrideLlmConfigFields = (
   config: LLMInferenceServiceConfigKind,
-  fields: ConfigFormFields,
+  overrides: ConfigFieldOverrides,
 ): LLMInferenceServiceConfigKind => {
-  const annotations = { ...config.metadata.annotations };
-  if (fields.displayName !== undefined) {
-    annotations['openshift.io/display-name'] = fields.displayName;
+  const apiGroup = LLMInferenceServiceConfigModel.apiGroup ?? '';
```
emilys314: Huh the `apiGroup` is being treated here as potentially undefined, when its been defined already. 

Looks like when I created the Model, I used `LLMInferenceServiceConfigModel: K8sModelCommon` but we could change it to this, which would make typescript 100% sure we have the apiGroup defined as a string.
```
export const LLMInferenceServiceConfigModel = {
  apiVersion: 'v1alpha2',
  apiGroup: 'serving.kserve.io',
  kind: 'LLMInferenceServiceConfig',
  plural: 'llminferenceserviceconfigs',
} satisfies K8sModelCommon;
```

Then we wouldn't need the extra `?? ''`

packages/llmd-serving/src/settings/llmAcceleratorConfigs/LlmAcceleratorConfigAddForm.tsx
```diff
@@ -129,24 +129,13 @@ const LlmAcceleratorConfigAddForm: React.FC<LlmAcceleratorConfigAddFormProps> =
       setError(new Error('YAML must represent a valid object'));
       return;
     }
```
emilys314, line 129: Maybe this is up to microcopy or out of scope for this PR, but I was testing a minimal create to see what would be added from the override (no `metadata` property:
```apiVersion: serving.kserve.io/v1alpha2
kind: LLMInferenceServiceConfig
spec:
  model:
    uri: ""
  template:
    containers:
      - image: registry.redhat.io/rhaiis/vllm-rocm-rhel9@sha256:d9a48add238cc095fa43eeee17c8c4d104de60c4dc623e0bc7f8c4b53b2b2e97
        name: main
        resources: {}
```
and I was running into this error, I had to add a stub `metadata: {}` for it to work.

This is what the console does for example:
Image

maybe?
`YAML must represent a valid object` => `YAML must represent a valid kubernetes resource object`

emilys314 (commented):
small nitpicks but i tested it and it all seems to work good

packages/llmd-serving/src/settings/llmAcceleratorConfigs/LlmAcceleratorConfigAddForm.tsx
mturley: The generic message is already present on the topology config page [here](https://github.com/opendatahub-io/odh-dashboard/blob/main/packages/llmd-serving/src/settings/TopologyConfigurationCreateEdit.tsx#L240) and in the routing config page [here](https://github.com/opendatahub-io/odh-dashboard/blob/main/packages/llmd-serving/src/settings/RoutingConfigurationCreateEdit.tsx#L295) it says "YAML must represent a valid object with a metadata block”. I think your message makes more sense, I’ll change it in all 3 places and we can follow up. I guess this indicates more code that we could factor out of these 3 forms but I think that’s out of scope here.

packages/llmd-serving/src/settings/configYamlUtils.ts
mturley: Good call.

packages/llmd-serving/src/settings/llmAcceleratorConfigs/LlmAcceleratorConfigAddForm.tsx
mturley: Addressed: updated the error message to "YAML must represent a valid kubernetes resource object" across all three settings pages.

packages/llmd-serving/src/settings/configYamlUtils.ts
mturley: Addressed

mturley:
Pushed a commit to address part of the above CodeRabbit comment: we should be catching any type of exception, not just Error. The other part of the comment (validate the name field on submit) seems invalid to me, we already disable the submit button if that field is empty.

emilys314 (approved):

PR #8694, opened 2026-07-20
fix(automl,autorag): fix chart color mismatch on curve toggle and CI marker clipping

jefho-rh:
https://redhat.atlassian.net/browse/RHOAIENG-78026

## Description

Fixes two visual bugs in AutoML and AutoRAG run results charts:

**AutoML — Evaluation curve chart color mismatch (ROC / Precision-Recall):**
When toggling curves off via the legend eye-toggle in multiclass mode, the tooltip dot colors would mismatch the rendered curve colors. Two issues:
1. `isBinary` was derived from `visibleLines.length === 1` (filtered lines), so hiding all but one curve in a multiclass chart incorrectly switched the tooltip to binary mode.
2. The tooltip used the positional index of the visible/filtered array to look up colors, but ChartLine stroke colors use the original index in the full `curveLines` array.

https://github.com/user-attachments/assets/92774e5e-3888-46a9-9af2-d9113d3dce4d

**AutoRAG — CI marker clipping:**
The `.autorag-ci-scores` container only had `padding-left`, so confidence interval markers near the right edge were clipped. Changed to `padding-inline` to pad both sides.

Screenshot 2026-07-20 at 2 17 30 PM

## How Has This Been Tested?

Manually tested with multiclass classification results — toggled curves on/off and verified tooltip dot colors match rendered curve colors in all toggle combinations. Verified CI markers render without clipping on both edges.

## Test Impact

No new tests added. The tooltip rendering logic depends on Victory's internal component model which is difficult to unit test in isolation. The existing `EvaluationCurveChart.spec.tsx` covers the `findYAtX` interpolation utility. A follow-up could extract the `seriesPoints` derivation into a testable pure function.

## Request review criteria:

If you have UI changes:

### DIFF
--- a/packages/automl/frontend/src/app/components/run-results/AutomlModelDetailsModal/components/EvaluationCurveChart.tsx
+++ b/packages/automl/frontend/src/app/components/run-results/AutomlModelDetailsModal/components/EvaluationCurveChart.tsx
@@ -78,9 +78,10 @@ const CursorVoronoiContainer = createContainer('cursor', 'voronoi');
 // These singletons bridge that gap.
 const chartState: {
   curveLines: CurveLine[];
+  isMulticlass: boolean;
   xMetricLabel: string;
   yMetricLabel: string;
-} = { curveLines: [], xMetricLabel: '', yMetricLabel: '' };
+} = { curveLines: [], isMulticlass: false, xMetricLabel: '', yMetricLabel: '' };
 
 const TOOLTIP_ROW_H = 16;
 const TOOLTIP_PAD = 10;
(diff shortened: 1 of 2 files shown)


jefho-rh:
@coderabbitai review

nickmazzi:
Thanks @jefho-rh 

# AI assisted review

## Overall Assessment

**Approve.** This is a clean, well-scoped bug fix. Both issues are real, the root causes are correctly identified, and the fixes are minimal and correct. The PR body is thorough with good before/after evidence.

## Suggestions

1. **Extract `seriesPoints` derivation into a testable pure function.** The logic that maps `curveLines` + `isMulticlass` + `datum` to `SeriesPoint[]` is the core of what broke and what this PR fixes. Extracting it (e.g., `buildSeriesPoints(lines, isMulticlass, datum)`) would make it unit-testable. Not blocking since the fix is correct, but it would prevent regressions.

2. **Consider deriving `isMulticlass` from the prop directly in `CurveChartTooltip`.** Currently it flows through module-level state (`chartState.isMulticlass`). If the tooltip could access `curveLines` (the full set) via `chartState`, it could derive `isMulticlass` itself, reducing the surface area of the mutable singleton. Minor — the current approach works.

# Test Validation

## autorag

### Overall Summary

| Suite      | Result |
|------------|--------|
| Frontend   | ✅     |
| BFF        | ✅     |
| Contract   | ✅     |
| BFF Lint   | ✅     |
| Lint       | ✅     |
| Type Check | ✅     |


## automl

### Overall Summary

| Suite      | Result |
|------------|--------|
| Frontend   | ✅     |
| BFF        | ✅     |
| Contract   | ✅     |
| BFF Lint   | ✅     |
| Lint       | ✅     |
| Type Check | ✅     |


# Manual testing

image

image

PR #8293, opened 2026-06-25
update mod-arch and use new prop in TruncatedText

Philip-Carneiro:
https://issues.redhat.com/browse/RHOAIENG-56639

## Description
The `TruncatedText` component from `mod-arch-shared` renders a PatternFly `<Tooltip>` on hover when description text is truncated in model catalog cards. This tooltip appears directly above the text, overlapping the model title link in the card header and intercepting user clicks — preventing navigation to the model detail page.

**Root cause**: The midstream was using [mod-arch-shared@1.12.0](mailto:mod-arch-shared@1.12.0) which lacked the `tooltipPosition` prop. The upstream fix (kubeflow/hub#2792) added `tooltipPosition="left"` to reposition the tooltip, but this prop was only available starting from [mod-arch-shared@1.19.0](mailto:mod-arch-shared@1.19.0).

**Fix**: Updated mod-arch-core, mod-arch-kubeflow, and mod-arch-shared to `~1.19.0` and added `tooltipPosition="left"` to all 4 `TruncatedText` instances in `ModelCatalogCardBody.tsx`. This repositions the tooltip to the left side of the card, eliminating the collision with the title link.

Before the changes:
image

After changes:
image

Related upstream PR: kubeflow/hub#2792
Related mod-arch-library PR: opendatahub-io/mod-arch-library#235

## How Has This Been Tested?
- TypeScript compilation: no new errors introduced
- Verified [mod-arch-shared@1.19.0](mailto:mod-arch-shared@1.19.0) exports `tooltipPosition` prop on `TruncatedText`
- Manual verification pending on cluster

## Test Impact
No new tests required — this is a prop addition to existing components. The tooltip behavior is visual and tested via manual QA. Existing Cypress tests cover the model catalog card rendering.

## Request review criteria:
Self checklist (all need to be checked):
- [x] The developer has manually tested the changes and verified that the changes work
- [x] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [x] The developer has added tests or explained why testing cannot be added
- [x] The code follows our Best Practices

If you have UI changes:
- [x] Included any necessary screenshots or gifs if it was a UI change.
- [ ] Included tags to the UX team if it was a UI/UX change.

After the PR is posted & before it merges:
- [ ] The developer has tested their solution on a cluster by using the image produced by the PR to `main`
```

PR #8708, opened 2026-07-21
refactor (manifests): restructure kustomize layout — merge base, rename sidecar, add standalone overlays

lucferbux:
https://issues.redhat.com/browse/RHOAIENG-78081
https://issues.redhat.com/browse/RHOAIENG-78359
https://issues.redhat.com/browse/RHOAIENG-67944

## Description

Restructures the `manifests/` directory to fix accumulated technical debt and enable proper standalone deployment mode in the dashboard-operator.

### Problems fixed

1. **Artificial `common/` + `core-bases/` split** — these two directories together form the dashboard's "base" (CRDs, RBAC, apps, Deployment, Service, SA). Merging them into `base/` eliminates the confusion.

2. **`modular-architecture/` dual purpose** — it was both a sidecar injection overlay AND a host for standalone per-module deployments. Renaming to `sidecar/` and moving `modules/` to the top level makes each directory's purpose clear.

3. **Standalone mode deployed wrong pod** — `reconcileStandalone()` in the dashboard-operator was rendering from the same manifest path as sidecar mode, so the main pod ended up with 11 containers (all BFF modules as sidecars) instead of the intended 3. Fixed by adding `standaloneManifestSets()` that renders from `odh/standalone/` or `rhoai/standalone/`.

4. **Right-sized resource requests** ([RHOAIENG-78081](https://redhat.atlassian.net/browse/RHOAIENG-78081)) — reduces cluster footprint ~88% CPU / ~73% memory. All BFF containers now use `50m/128Mi` requests instead of `300m/512Mi`. Validated on a live ROSA cluster (actual idle usage: 22m CPU / 629Mi across all pods).

5. **core-bff wired into the base pod** (absorbs #8547) — `manifests/base/deployment.yaml` now has `odh-dashboard`, `kube-rbac-proxy`, and `core-bff` as the three core containers. Includes an `emptyDir` at `/tmp` so `core-bff`'s `controller-runtime` init does not panic under `readOnlyRootFilesystem: true`.

6. **Updated kube-rbac-proxy to dedicated image** ([RHOAIENG-67944](https://redhat.atlassian.net/browse/RHOAIENG-67944)) — `params.env` files now reference `quay.io/opendatahub/odh-kube-rbac-proxy` (tag `3.5.0-ea.2`, digest `sha256:0b711...`) instead of the old `odh-kube-auth-proxy` image. The `kube-rbac-proxy` binary has been extracted into its own dedicated image; newer `odh-kube-auth-proxy` images no longer ship it.

### Directory changes

| Old path | New path |
|---|---|
| `manifests/common/` + `manifests/core-bases/` | `manifests/base/` (merged) |
| `manifests/modular-architecture/` | `manifests/sidecar/` (renamed) |
| `manifests/modular-architecture/modules/` | `manifests/modules/` (moved) |
| `manifests/overlays/dev/` | Deleted |

### New overlays

- `manifests/odh/standalone/` — ODH overlay for operator standalone mode (3-container pod, no BFF sidecars)
- `manifests/rhoai/standalone/` — RHOAI overlay for operator standalone mode (core renames only, no sidecar container-index patches)

### Invariant

`kustomize build manifests/odh` and `kustomize build manifests/rhoai` produce output identical to before except for intentional additions: `core-bff` container, `core-bff-image` param, ConfigMap rename `modular-architecture-params` → `sidecar-params`, right-sized resources, and updated kube-rbac-proxy image.

---

### Kustomize build validation

All four overlay targets build without error:

```bash
kustomize build manifests/odh           # sidecar mode
kustomize build manifests/rhoai         # sidecar mode
kustomize build manifests/odh/standalone    # standalone: 3 containers only
kustomize build manifests/rhoai/standalone  # standalone: 3 containers + RHOAI renames
```

### Unit tests

```bash
cd dashboard-operator && make test
# ok  controller (incl. new TestApplyKustomizeParamsStandaloneNoSidecar)
# ok  webhook
```

### Live cluster — standalone mode end-to-end

Tested on a ROSA cluster (OpenShift 4.13, ODH community operator v3.5.0-ea.2).

**Reference image** (built from this PR's HEAD):

```
quay.io/lferrnan/odh-dashboard-operator:test-standalone
```

#### Observed results on the test cluster

```
Main pod containers: odh-dashboard  kube-rbac-proxy  core-bff   (3 total)

Standalone BFF pods:
  automl-*           Running  (x2)
  autorag-*          Running  (x2)
  eval-hub-*         Running  (x2)
  gen-ai-*           Running  (x2)
  maas-*             Running  (x2)
  mlflow-*           Running  (x2)
  model-registry-*   Running  (x2)

Resource usage (kubectl top, all pods at idle):
  Total CPU:    22m   (old sidecar requests: 6,800m)
  Total memory: 629Mi (old sidecar requests: ~12Gi)
```

---

## Test Impact

- `dashboard-operator/internal/controller/actions_test.go` — `modular-architecture` fixture dir renamed to `sidecar`; new `TestApplyKustomizeParamsStandaloneNoSidecar` added
- `dashboard-operator/internal/controller/dashboard_reconciler_test.go` — same rename
- No Cypress tests affected (manifest-only + operator code changes)

## Request review criteria:

[RHOAIENG-78081]: https://redhat.atlassian.net/browse/RHOAIENG-78081


manifests/base/kustomization.yaml
liday-rh, line 1: nit: The old `manifests/core-bases/base/` directory still exists on the branch — the `kustomization.yaml` inside it wasn't deleted (only the individual files like `deployment.yaml` and `service.yaml` were renamed out). It's dead code (nothing references it), but leaving it around contradicts the stated goal of removing `core-bases/`.

dashboard-operator/internal/controller/module_deploy.go
```diff
@@ -126,14 +126,14 @@ func (r *DashboardReconciler) deployModuleManifests(
 			continue
 		}
 
-		modulePath := filepath.Join(r.ManifestsBasePath, "modular-architecture", "modules", mod.ManifestSlug)
+		modulePath := filepath.Join(r.ManifestsBasePath, "modules", mod.ManifestSlug)
```
liday-rh, line 140: Good catch — without `readExistingParams` here, standalone module deployments would silently drop any default values in the module's `params.env` that weren't in the computed set (e.g. digest-pinned image defaults).

manifests/rhoai/standalone/role-binding.yaml
```diff
@@ -0,0 +1,12 @@
...
+  value: rhods-dashboard
+- op: replace
+  path: /roleRef/name
+  value: rhods-dashboard
+- op: replace
+  path: /subjects/0/name
+  value: rhods-dashboard
+- op: add
+  path: /subjects/0/namespace
+  value: redhat-ods-applications
```
liday-rh: nit: Missing trailing newline (same for `role.yaml` and `auth-delegator.clusterrolebinding.yaml`).

manifests/rhoai/base/deployment.yaml
```diff
@@ -17,12 +17,13 @@
 - op: replace
   path: /spec/template/spec/serviceAccount
   value: rhods-dashboard
```
liday-rh, line 19: Verified the index shift. With `core-bff` now at index 2 in the base deployment:

- 0: odh-dashboard, 1: kube-rbac-proxy, 2: core-bff
- Sidecar appends: 3: model-registry-ui, **4: gen-ai-ui**, 5: maas-ui, 6: mlflow-ui, **7: eval-hub-ui**, 8: automl-ui, 9: autorag-ui, 10: agent-ops-ui

`containers/4/env/3` and `/9` (gen-ai-ui's `BFF_MAAS_SERVICE_NAME` and `BFF_MLFLOW_SERVICE_NAME`) and `containers/7/env/1` (eval-hub-ui's `BFF_MODEL_CATALOG_SERVICE_NAME`) all check out.

liday-rh (commented):
Reviewed the full diff (169 files). The structural changes are sound — the `common/` + `core-bases/` merge into `base/` and the `modular-architecture/` → `sidecar/` rename are clean, all kustomization and Go path references are updated consistently, and the RHOAI container-index patches are correct.

The core bug fix (standalone mode was deploying the 11-container sidecar pod instead of the 3-container core pod) is properly addressed via `standaloneManifestSets()` and the new `odh/standalone` / `rhoai/standalone` overlays.

Two cleanup items before merge:
1. **Orphaned `manifests/core-bases/base/`** — the directory still exists with a leftover `kustomization.yaml` (see inline comment).
2. **Stale `manifests/README.md`** — still references `common/`, `core-bases/base/`, and `overlays/dev/` in the Installation section and Partially Managed note.

manifests/sidecar/deployment.yaml
liday-rh, line 1: core-bff (container 2 in the base) is missing a SA token volume mount in the sidecar overlay.

The overlay sets `automountServiceAccountToken: false` on the pod, then explicitly mounts `dashboard-sa-token` on containers 0 (odh-dashboard) and 1 (kube-rbac-proxy). Module BFFs (appended containers 3+) get `modules-sa-token`. But core-bff gets neither.

core-bff's `--bundle-paths` references `/var/run/secrets/kubernetes.io/serviceaccount/ca.crt` and `service-ca.crt` — these paths won't exist in sidecar mode, so core-bff will fail TLS verification for Kubernetes API calls.

Standalone mode is unaffected (`automountServiceAccountToken` defaults to `true`).

Fix: add a patch to mount `dashboard-sa-token` at `/var/run/secrets/kubernetes.io/serviceaccount` on container 2, matching the existing mounts for containers 0 and 1.

manifests/rhoai/standalone/kustomization.yaml
```diff
@@ -0,0 +1,147 @@
...
+      version: v1
+      kind: Service
+      name: odh-dashboard
+  - path: httproute.yaml
+    target:
+      group: gateway.networking.k8s.io
+      version: v1
+      kind: HTTPRoute
+      name: odh-dashboard
+  - path: networkpolicy.yaml
```
liday-rh: This patch targets `NetworkPolicy/odh-dashboard-allow-ports`, but the standalone resource chain (`../../base`) does not include any NetworkPolicy — that resource only exists in `manifests/sidecar/`. The ODH standalone overlay correctly omits this patch.

`kustomize build manifests/rhoai/standalone/` will fail with "unable to find matching target for patch". This is likely the root cause of the Phase 4 CI failure.

Fix: remove this `networkpolicy.yaml` patch entry (and the patch file), or add a NetworkPolicy to the base resources before patching it.

manifests/base/deployment.yaml
```diff
@@ -144,7 +146,80 @@ spec:
...
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.namespace
+          resources:
+            requests:
+              cpu: 50m
+              memory: 128Mi
+              ephemeral-storage: 10Mi
+            limits:
+              memory: 256Mi
```
liday-rh, line 171: nit: core-bff is missing a `cpu` limit. Every other container has one — odh-dashboard (`500m`), kube-rbac-proxy (`200m`), all sidecar/standalone BFFs (`200m`). Without it, core-bff can consume unbounded CPU and may be rejected in namespaces with a `LimitRange` enforcing CPU limits.

```suggestion
            limits:
              cpu: 200m
              memory: 256Mi
```

docs/inter-bff-communication.md
```diff
@@ -366,8 +392,70 @@ if err != nil {
...
+
+### Service coordinates
+
+| Mode | Variable | Value |
+|---|---|---|
+| Both | `BFF_CORE_BFF_SERVICE_NAME` | `odh-dashboard` (ODH) / `rhods-dashboard` (RHOAI) |
+| Both | `BFF_CORE_BFF_SERVICE_PORT` | `8943` |
+| Both | `BFF_CORE_BFF_TLS_ENABLED` | `true` (K8s) / `false` (local dev) |
+
+In **standalone mode** the dashboard-operator injects `BFF_CORE_BFF_SERVICE_NAME` and `BFF_CORE_BFF_SERVICE_PORT` automatically for modules listed in `modulesWithCoreBFFAccess` (`gen-ai`, `maas` today — extend it to add more).
```
liday-rh: nit: `modulesWithCoreBFFAccess` was removed in the latest commit (`5e6486f` — "removes modulesWithCoreBFFAccess (speculative injection removed)"), but the doc still references it here and in the "Adding a new module" section (line 432). These point readers to code that no longer exists.

dashboard-operator/internal/controller/actions.go
```diff
@@ -51,14 +60,15 @@ func applyKustomizeParams(dashboard *v1alpha1.Dashboard, manifests []render.Mani
 	}
 
 	if len(manifests) > 0 {
-		modArchPath := filepath.Join(manifests[0].Path, "modular-architecture")
-		if _, err := os.Stat(modArchPath); os.IsNotExist(err) {
-			return fmt.Errorf("modular-architecture directory not found at %s", modArchPath)
+		sidecarPath := filepath.Join(manifests[0].Path, "sidecar")
+		if _, err := os.Stat(sidecarPath); os.IsNotExist(err) {
+			// sidecar dir may not exist in standalone mode — skip gracefully
+			return nil
```
liday-rh: nit: This function is called from both `reconcileSidecar` (via `manifestSets`) and `reconcileStandalone` (via `standaloneManifestSets`). The graceful skip is correct for standalone mode, but in sidecar mode, if `sidecar/` were ever absent (bad image build), the operator would silently succeed with un-parameterized manifests — the old code returned an error here.

Unlikely in practice, but the prior contract was stricter. Consider checking the deployment mode or the manifest's `SourcePath` and only skipping gracefully for standalone.

liday-rh (commented):
Re-review at `5e6486f` (8 commits, up from 1). All 5 previous findings resolved — nice work.

Two new issues that likely need fixing before merge:

1. **core-bff missing SA token mount in sidecar mode** — `automountServiceAccountToken: false` + no explicit mount on container 2 = broken TLS in sidecar mode
2. **`rhoai/standalone` patches a non-existent NetworkPolicy** — likely root cause of Phase 4 CI failure

Two minor items (cpu limit, stale docs) also noted inline.

manifests/base/deployment.yaml
lucferbux: Actually the opposite — the existing pattern on main intentionally has no cpu limit (only memory). Both `odh-dashboard` and `kube-rbac-proxy` in `core-bases/base/deployment.yaml` have cpu+memory *requests* but only a memory *limit*. It's a deliberate best practice: cpu is compressible so a limit causes throttling rather than eviction, whereas a memory limit gives a hard OOM boundary. core-bff is already following the correct pattern. Will also clean up the cpu limits accidentally added to the other containers in this PR.

manifests/base/kustomization.yaml
lucferbux: Fixed in the second commit.

dashboard-operator/internal/controller/module_deploy.go
lucferbux: Yeah, exactly the issue — PR #8293 tracked it down.

manifests/rhoai/standalone/role-binding.yaml
lucferbux: Fixed.

manifests/rhoai/base/deployment.yaml
lucferbux: Thanks for verifying — good to have a second pair of eyes on the index math.

manifests/sidecar/deployment.yaml
lucferbux: Good catch. Added the `dashboard-sa-token` volumeMount for container 2 (core-bff) in the sidecar overlay — same as containers 0 and 1.

manifests/rhoai/standalone/kustomization.yaml
lucferbux: Confirmed — the patch was producing garbage raw JSON6902 ops in the rendered output. Removed the patch and the file; the base doesn't include a NetworkPolicy so there's nothing to rename.

docs/inter-bff-communication.md
lucferbux: Fixed — removed the references and replaced them with the actual wiring mechanism (`interBFFDependencies` / manual env var in the deployment manifest).

dashboard-operator/internal/controller/actions.go
lucferbux: Fair point — tightened it to only skip gracefully when `SourcePath` contains `standalone`; in sidecar mode a missing `sidecar/` now returns an error so we don't silently produce un-parameterized manifests.

christianvogt:
@odh-dashboard-agent preflight

lucferbux:
Hey @christianvogt — ran the persona review report against the PR. Here's what we acted on and what we deferred.

### Fixed (commit `70d7503`)

| Finding | Fix |
|---|---|
| **Dockerfile missing `chmod` on `rhoai/params.env`** | Added — RHOAI sidecar write path was silently failing under arbitrary UID |
| **`MODULE_FEDERATION_CONFIG` missing from standalone** | Added to `base/deployment.yaml` as `optional: true` configMapKeyRef — Fastify needs this to register BFF remotes/proxy routes |
| **`standaloneServiceName` returns `rhods-dashboard-*` on RHOAI** | Fixed to always return `odh-dashboard-*` — module Services are always named `odh-dashboard-*-ui` in `manifests/modules/`; the RHOAI prefix made the federation ConfigMap point to non-existent Services |
| **Module CA cert volumes missing `optional: true`** | Added to all 8 standalone module deployments — `base/deployment.yaml` already had it, modules were missing it which would cause `CreateContainerConfigError` if the CA bundle CM doesn't exist |
| **Helm `relatedImages` missing `RELATED_IMAGE_ODH_CORE_BFF_IMAGE`** | Added to `charts/dashboard/values.yaml` — already in `imagesMap` in `support.go` but the chart wasn't exposing it for Konflux pinning |
| **Stale README paths** (`core-bases/`, `overlays/dev`) | Fixed in `manifests/README.md` and `manifests/rhoai/README.md` |

### Deferred as follow-up tasks

| Finding | Reason |
|---|---|
| `--v=10` on kube-rbac-proxy (CWE-532 token logging) | Pre-existing on main, not introduced by this PR; tracked separately |
| NetworkPolicy for standalone overlays (no NP on 8443/8943) | New feature gap — original base never had a NP; needs separate design for what callers to allow |
| `automountServiceAccountToken: false` on standalone base | Enhancement, not a regression; sidecar overlay already does this; follow-up to port to base |
| Single source of truth for `coreBff` entry (sidecar CM, RHOAI CM, operator) | Requires larger refactor of the two-mode design |
| Unit tests for `buildFederationConfigMap` / standalone mode / negative sidecar | Valid gaps; added to follow-up backlog |
| Standalone detection via `strings.Contains` brittle | Already improved with explicit SourcePath check; further hardening follow-up |
| Module NP egress to 8943 per docs | Partially done (gen-ai, maas); follow-up to add per-module as needed |

lucferbux:
## E2E Smoke Test Results — Standalone Deployment

Ran the E2E login smoke test (`application.cy.ts`) against a live ROSA cluster with the new standalone deployment from this PR.

### Test Environment

| Property | Value |
|---|---|
| Cluster | ROSA (`ui-monarch-rosa`) |
| Operator image | `quay.io/lferrnan/odh-dashboard-operator:test-standalone` (built from this PR @ `70d7503`) |
| Deployment mode | `Standalone` (`deploymentMode: Standalone`) |
| ODH operator | Scaled down — dashboard operator fully in control |

### Pod State After Deployment

```
odh-dashboard-7b7df77d77-*   3/3   Running  ← main pod (odh-dashboard + kube-rbac-proxy + core-bff)
agent-ops-ui-*               1/1   Running  ← standalone module pod
automl-ui-*                  1/1   Running
autorag-ui-*                 1/1   Running
eval-hub-ui-*                1/1   Running
gen-ai-ui-*                  1/1   Running
maas-ui-*                    1/1   Running
mlflow-ui-*                  1/1   Running
model-registry-ui-*          1/1   Running
dashboard-operator-*         1/1   Running
```

Main pod confirmed **3/3** containers (no sidecar BFF containers leaking in), all 8 module pods deployed as independent pods. ✅

### E2E Result

```
  ✓ should login and load page (18s)

  1 passing (18s)
```

Dashboard loaded, login completed, Projects page rendered with real cluster data, nav items (Home, Projects, AI Hub, Develop & Train, Learning Resources, Applications, Settings) all visible. The `clusteradminuser1` account shows correctly in the masthead.

The full `@Dashboard` grep tag suite (76 specs) was scanned — 127 tests all pending (expected: this cluster doesn't have all feature data seeded for the broader suite). No failures.

lucferbux:
/approve

@liday-rh I'm gonna approve this, if you can re-lgtm it again I'll then unhold this once https://github.com/opendatahub-io/opendatahub-operator/pull/3733/changes is merged

christianvogt (approved):

Issue #746, opened 2022-11-11
[Bug]: ceph-nano ODH 1.4.0 on Openshift OCP 4.10.36 manifests tag 1.4.0


shalberd:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

KfDef:

```
apiVersion: kfdef.apps.kubeflow.org/v1
kind: KfDef
metadata:
  name: opendatahub
spec:
  applications:
  - kustomizeConfig:
      repoRef:
        name: manifests
        path: odh-common
    name: odh-common
  - kustomizeConfig:
      overlays:
      - authentication
      repoRef:
        name: manifests
        path: odh-dashboard
    name: odh-dashboard
  - kustomizeConfig:
      repoRef:
        name: manifests
        path: ceph/object-storage/scc
    name: ceph-nano-scc
  - kustomizeConfig:
      repoRef:
        name: manifests
        path: ceph/object-storage/nano
    name: ceph-nano
  - kustomizeConfig:
      overlays:
      - trusted-ca-bundle-path
      parameters:
        - name: trusted-ca-bundle-path
          value: "/opt/app-root/etc/jupyter/custom/cacerts/trustedcas.pem"
        - name: s3_endpoint_url
          value: "http://ceph-nano-0"
      repoRef:
        name: manifests
        path: jupyterhub/jupyterhub
    name: jupyterhub
  - kustomizeConfig:
      overlays:
      - additional
      repoRef:
        name: manifests
        path: jupyterhub/notebook-images
    name: notebook-images
  repos:
  - name: manifests
    uri: https://repo.internalserver.ch/artifactory/datascience_internal/opendatahub-io-odh-manifests-v1.4-0.tar.gz
  version: master
```

opendatahub-operator log when using ceph. Seems to be an issue with the ceph-nano StatefulSet.
https://github.com/opendatahub-io/odh-manifests/blob/master/ceph/object-storage/nano/base/statefulset.yaml

```
serviceaccount/ceph-nano serverside-applied
role.rbac.authorization.k8s.io/ceph-nano-secrets serverside-applied
rolebinding.rbac.authorization.k8s.io/ceph-nano-secrets serverside-applied
service/ceph-nano serverside-applied
service/ceph-nano-0 serverside-applied
[...]

PR #8726, opened 2026-07-21
fix(observability): bundle patched Perses Loki/Table/StatChart plugins [RHOAIENG-78528]

christianvogt:
https://redhat.atlassian.net/browse/RHOAIENG-78528

## Description

Bundles newer Perses plugin builds as local overrides so Loki-backed Table and StatChart panels use instant queries instead of always hitting `query_range`.

The COO-shipped Perses Loki plugin incorrectly sends range queries for panels that need instant queries, producing silently wrong values. Upstream fixes landed in:

- [perses/plugins#736](https://github.com/perses/plugins/pull/736)
- [perses/plugins#737](https://github.com/perses/plugins/pull/737)
- [perses/plugins#738](https://github.com/perses/plugins/pull/738)

This PR overrides:

| Plugin | Version |
| --- | --- |
| `@perses-dev/loki-plugin` | `v0.6.0-rc.2` |
| `@perses-dev/table-plugin` | `v0.13.0-rc.1` |
| `@perses-dev/stat-chart-plugin` | `v0.13.0-rc.1` |

Also registers the existing Prometheus override under the remote module name (`Prometheus`) so bundled overrides match both the Perses API name and the npm package name.

Compare the usage breakdown table:
Before:
image
After:
image

## How Has This Been Tested?

Manually tested on a shared cluster: compared Loki-backed StatChart / Table panel values against direct Loki instant queries and confirmed results align with the patched plugin behavior.

Confirmed Model dashboard continues to load for a regular user.

## Test Impact

No new automated tests. Coverage is a third-party Perses plugin version override; validation is against live Loki query behavior on a cluster.

## Request review criteria:

If you have UI changes:

## Summary by CodeRabbit

* **New Features**
  * Added support for Loki data visualizations in observability dashboards.
  * Added statistical chart and table visualizations.
* **Improvements**
  * Enhanced bundled plugin compatibility by dynamically resolving and caching bundled overrides for Prometheus, Loki, StatChart, and Table.
  * Improved behavior when bundled modules fail to load, including retry after transient errors.
* **Tests**
  * Added Jest coverage for bundled override resolution and plugin loader integration.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3671,6 +3671,15 @@
         "url": "https://github.com/sponsors/ayuhito"
       }
     },
+    "node_modules/@grafana/lezer-logql": {
+      "version": "0.2.9",
+      "resolved": "https://registry.npmjs.org/@grafana/lezer-logql/-/lezer-logql-0.2.9.tgz",
+      "integrity": "sha512-SB9E2LQ689PiI/OPuBoTF93O5hBb1n8DbS3uSXfH2YYTsQELHqwU2HSM8BAI/ThX1ggkvIN9y0JyNTqfMKjlBA==",
+      "license": "Apache-2.0",
+      "peerDependencies": {
+        "@lezer/lr": "^1.0.0"
+      }
+    },
     "node_modules/@hapi/hoek": {
       "version": "9.3.0",
       "resolved": "https://registry.npmjs.org/@hapi/hoek/-/hoek-9.3.0.tgz",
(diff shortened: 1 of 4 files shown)


DaoDaoNoCode:
Note for future reference: `@perses-dev/stat-chart-plugin` is pinned to `v0.13.0-rc.1` which includes hardcoded `queryOptions: { mode: 'instant' }` (upstream PR [perses/plugins#738](https://github.com/perses/plugins/pull/738)). This was later reverted upstream ([perses/plugins#744](https://github.com/perses/plugins/pull/744)) because it breaks StatChart panels that use sparklines — sparklines need range data.

This is safe for our dashboards because none of our 8 StatChart panels (4 ODH + 4 RHOAI in `manifests/observability/`) use sparklines — they only use `calculation: mean` or `calculation: last-number` for single aggregate values.

`v0.13.0-rc.2` has the revert (no instant mode), which would be safer generally but means StatChart panels would still send range queries to Loki, defeating the purpose of this fix. The proper upstream solution ([perses/plugins#746](https://github.com/perses/plugins/pull/746) — derive mode from spec) is still open.

When upgrading these overrides in the future, verify that the new version either includes #746 or that our stat panels still don't use sparklines.

DaoDaoNoCode (approved):
/lgtm
The comment above is just a FYI for future developers, the fix in this PR is correct.

PR #8737, opened 2026-07-21
add Event Tracking to Workbench Images Page

hmtosi:
## Description
This PR fixes https://redhat.atlassian.net/browse/RHOAIENG-76358
It adds event tracking to the Workbench Images page.
Event tracking has been created/confirmed for the following scenarios:
- Workbench Images Admin Page is viewed
- Workbench Images are toggled to enabled/disabled (works for both BYON and OOTB images)
- Filters are updated in the Workbench Images search

Open Question to the reviewer: These events are meant to be tracked in Amplitude, but I was not able to find the dashboard that would track them. Is this something I need to build specifically for this PR?

## How Has This Been Tested?
- I ran the ODH Dashboard with my updated code
- I opened the Developer Tools Console
- I tested each scenario and confirmed that the information was correctly logged in the console

## Test Impact

This is not adding code functionality, only adding event tracking. Thus, extra tests are not applicable.

## Request review criteria:

If you have UI changes: 

## Summary by CodeRabbit

* **Improvements**
  * Enhanced BYON image toolbar filtering with stricter key validation, unified filter application, safe handling of unknown keys, debounced text filters, and more accurate active filter/result/total counts.
  * Improved image visibility toggle tracking with richer success/failure telemetry, including toggle direction and image/provider details, and a shared session toggle index across table rows.
  * Kept telemetry reporting non-blocking so toggle UX and error notifications aren’t impacted.
  * Updated the image type filter label/option from “Red Hat” to “Pre-installed.”


frontend/src/pages/BYONImages/BYONImagesTable.tsx
```diff
@@ -68,16 +90,43 @@ export const BYONImagesTable: React.FC<BYONImagesTableProps> = ({ images }) => {
...
+      }
+      const filterValue = typeof value === 'string' ? value : value?.value;
+
+      setFilterData((prevValues) => ({ ...prevValues, [key]: filterValue }));
+
+      const previousValue = filterData[key];
+      const newFilterData = { ...filterData, [key]: filterValue };
+      const newActiveFilterCount = getActiveFilterCount(newFilterData);
+
+      fireFormTrackingEvent('Workbench Images Filtered', {
```
Taj010: The ticket names this event `Workbench Images Filter Applied`, not Workbench Images Filtered. Can we align the name to ensure the Amplitude taxonomy and WIP events sheet match the spec?

frontend/src/pages/BYONImages/BYONImagesTable.tsx
```diff
@@ -68,16 +90,43 @@ export const BYONImagesTable: React.FC<BYONImagesTableProps> = ({ images }) => {
     useHardwareProfilesByFeatureVisibility(WORKBENCH_VISIBILITY);
 
   const onFilterUpdate = React.useCallback(
-    (key: string, value: string | { label: string; value: string } | undefined) =>
-      setFilterData((prevValues) => ({ ...prevValues, [key]: value })),
-    [setFilterData],
+    (key: string, value: string | { label: string; value: string } | undefined) => {
```
Taj010, line 124: Right now, each keystroke fires a separate tracking event — typing "test" sends 4 events (t, te, tes, test) instead of just 1 for the final value. The spec says to debounce the name search. One approach: debounce the fireFormTrackingEvent call for the name filter (maybe wait 300–500ms after the user stops typing), but keep setFilterData instant so the UI still filters in real time.

frontend/src/pages/BYONImages/ImageStatusToggle.tsx
```diff
@@ -38,10 +42,44 @@ const ImageStatusToggle: React.FC<ImageStatusToggleProps> = ({
...
     try {
       await onToggle(visible);
       setEnabled(visible);
+
+      const hiddenOotbCountAfter =
+        images.filter((img) => img.isOOTB && !img.visible).length +
+        (image.isOOTB ? (visible ? -1 : 1) : 0);
+
+      try {
+        fireFormTrackingEvent('Workbench Image Toggled', {
```
Taj010: Same here - can we rename it to Workbench Image Visibility Toggled. The same applies to the failure-path event on line 70.

Taj010 (commented):
To answer your open question about the Amplitude dashboard, you don't need to build one for this PR. Your code sends events through fireFormTrackingEvent() → Segment → Amplitude. The dashboard (charts, funnels) is built in the Amplitude UI by the product/analytics team, not in code. 

(tagging @nananosirova and @srtanish1992 to confirm since they have more context around event tracking) 

A couple of spec alignment items flagged inline — mainly event names (Workbench Images Filter Applied / Workbench Image Visibility Toggled) and a missing debounce on the Name filter.

frontend/src/pages/BYONImages/BYONImagesTable.tsx
hmtosi: thank you, I have made the change

frontend/src/pages/BYONImages/ImageStatusToggle.tsx
hmtosi: thank you, I have made the change

frontend/src/pages/BYONImages/BYONImagesTable.tsx
hmtosi: I have updated this as you suggested - it is set to 400ms. It now only sends one event if the user types quickly, but if a user types more slowly it will send multiple. Please let me know if I should adjust it to wait longer.

hmtosi:
@Taj010 thank you so much for your review, and for answering my question. I believe I have addressed all the necessary changes.

frontend/src/pages/BYONImages/BYONImagesTable.tsx
```diff
@@ -67,17 +94,72 @@ export const BYONImagesTable: React.FC<BYONImagesTableProps> = ({ images }) => {
...
+      };
+
+      const isTextFilter =
+        key === BYONImagesToolbarFilterOptions.name ||
+        key === BYONImagesToolbarFilterOptions.provider;
+
+      if (isTextFilter) {
+        let debounced = debouncedTrackersRef.current.get(key);
+        if (!debounced) {
+          debounced = _.debounce((p: FormTrackingEventProperties) => {
```
Taj010: The toggle side (`ImageStatusToggle.tsx`) correctly wraps every fireFormTrackingEvent in try/catch so telemetry can't break the UX. This path and the debounced callback don't — if the analytics adapter throws, the non-text filter path propagates the error, and the debounced path creates an unhandled rejection.

Can we wrap both in a small non-throwing helper to match what you already did in ImageStatusToggle?

```
const safeFireFilterEvent = (properties: FormTrackingEventProperties): void => {
  try {
    fireFormTrackingEvent('Workbench Images Filter Applied', properties);
  } catch {
    // telemetry must not affect filtering
  }
};
```
Then use safeFireFilterEvent in both the debounced callback (line ~134) and the immediate path (line ~139).

frontend/src/pages/BYONImages/const.ts
```diff
@@ -15,12 +15,12 @@ export const byonImagesFilterOptions = {
 };
 
 export enum ImageTypeFilter {
-  redHat = 'redHat',
+  redHat = 'preInstalled',
```
Taj010: The enum value was renamed to 'preInstalled' but the key is still redHat. Every consumer reads ImageTypeFilter.redHat and sees "red hat" — but it now means "pre-installed." In filterByonImages we even have a const isPreInstalled = typeFilter === ImageTypeFilter.redHat, which reads confusingly.

Can you rename the key too?
`preInstalled = 'preInstalled',`

hmtosi:
@Taj010 thank you for your further review - I have addressed both comments. Please give it a final check when you are able!

Taj010:

DaoDaoNoCode (commented):
**`onFilterUpdate` callback stability regression**

The original `onFilterUpdate` had `[setFilterData]` deps — effectively a stable reference since `setFilterData` never changes. This PR changes it to `[filterData, images]`, which recreates the callback on **every keystroke** in the text filters.

Two consequences:

1. **Re-rendering cascade**: `onFilterUpdate` is passed as a prop to `<BYONImagesToolbar>`. Every keystroke produces a new callback reference, forcing the toolbar (and its children) to re-render — even though the filter state update itself is correct via functional `setState`.

2. **Double-filtering**: Inside the callback, `filterByonImages(images, newFilterData).length` re-filters the entire image list for the `resultCount` telemetry property. The `useMemo` already does this exact work on the next render. So every filter change filters the list twice.

This also contradicts the project's own [`react.md` conventions](https://github.com/opendatahub-io/odh-dashboard/blob/main/.claude/rules/react.md) under "Functional setState for stable callbacks":

```ts
// Bad — count in deps forces new callback each render
const increment = React.useCallback(() => setCount(count + 1), [count]);

// Good — no dependency on count
const increment = React.useCallback(() => setCount((c) => c + 1), []);
```

**Suggested fix**: Decouple the tracking from the state update. Two options:

1. Use a ref to hold the latest `filterData` and `images`, so `onFilterUpdate` can read current values without them being in the deps array — keeping the callback stable with `[]` deps.

2. Fire the tracking event from a `useEffect` that watches `filterData`. The `filteredImages` memo already has the result count, so you get `resultCount` for free without double-filtering:

```ts
// Stable callback — only does the state update
const onFilterUpdate = React.useCallback(
  (key: string, value: string | { label: string; value: string } | undefined) => {
    if (!isFilterOption(key)) return;
    const filterValue = typeof value === 'string' ? value : value?.value;
    setFilterData((prev) => ({ ...prev, [key]: filterValue }));
  },
  [],
);

// Tracking fires after state commits — always sees fresh data
React.useEffect(() => {
  // fire debounced tracking event using filterData + filteredImages.length
}, [filterData, filteredImages.length]);
```

This restores the original callback stability, eliminates the duplicate filtering, and avoids the stale closure issue where `previousFilterValue` could be wrong during rapid input.

DaoDaoNoCode (approved):
/lgtm
Thank you for the contribution!

PR #8739, opened 2026-07-21
fix(RHOAIENG-78549): gate LLM accelerator configs page behind vLLMDeploymentOnMaaS flag

mturley:
https://issues.redhat.com/browse/RHOAIENG-78549

## Description

The LLM accelerator configurations admin settings page was gated only on `LLMD_SERVING_ID` + `ADMIN_USER`, but it should also require `SupportedArea.VLLM_ON_MAAS` (the `vLLMDeploymentOnMaaS` feature flag). Without this, the page was visible even when the feature flag was disabled.

The wizard field extension (`llmConfigOptionsFieldExtension`) already correctly requires `VLLM_ON_MAAS`, but the admin page's navigation and route extensions did not.

### Changes
- Added `SupportedArea.VLLM_ON_MAAS` to the `flags.required` array on both the `app.navigation/href` and `app.route` extensions for the LLM accelerator configurations page
- Updated Cypress mock test utils to enable the `vLLMDeploymentOnMaaS` flag so existing tests continue to pass
- Added a new test verifying the page is hidden when the flag is disabled

## How Has This Been Tested?

- Manually verified in a browser that the page is hidden when `vLLMDeploymentOnMaaS` is `false` and visible when `true` (with admin access)
- Type-check and lint pass on all modified packages

## Test Impact

- Updated `llmAcceleratorConfigsUtils.ts` to enable `vLLMDeploymentOnMaaS: true` in `llmAcceleratorConfigsIntercept()` so existing mock tests work with the new flag requirement
- Added a new Cypress mock test: "LLM accelerator configurations should not be available when vLLMDeploymentOnMaaS is disabled"
- Updated existing non-admin test to explicitly set the flag to `true` so it tests admin access independently of the feature flag

## Request review criteria:

## Summary by CodeRabbit

- **New Features**
  - Restricted access to LLM accelerator configuration pages and navigation for non-product administrators.
  - Added feature-flag control so the LLM accelerator configuration experience is available only when vLLM deployment on managed infrastructure is enabled.

- **Tests**
  - Added coverage verifying the page and navigation item are hidden when authorization or feature-flag requirements are not met.

### DIFF
--- a/packages/cypress/cypress/tests/mocked/llmAcceleratorConfigs/llmAcceleratorConfigs.cy.ts
+++ b/packages/cypress/cypress/tests/mocked/llmAcceleratorConfigs/llmAcceleratorConfigs.cy.ts
@@ -1,3 +1,4 @@
+import { mockDashboardConfig } from '@odh-dashboard/internal/__mocks__';
 import {
   llmAcceleratorConfigsIntercept,
   interceptLlmAcceleratorConfigPatch,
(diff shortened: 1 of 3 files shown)


emilys314 (approved):
Looks good, it is hidden when the flag is disabled

Image

PR #8758, opened 2026-07-22
fix(RHOAIENG-56820): validate route name + namespace length against 63-char DNS limit

danreed-rh:
https://issues.redhat.com/browse/RHOAIENG-56820

## Description
Human takeover of stalled autofix PR https://github.com/opendatahub-io/odh-dashboard/pull/8209.

OpenShift routes reject hostnames when `{name}-{namespace}` exceeds the 63-character DNS label limit, but the UI previously accepted those inputs. This adds validation to `K8sNameDescriptionField` (and call sites that create route-backed resources) so users see a clear error before submit.

**Call sites / wiring:** workbench spawner, KServe manage modal, connection forms, model-serving deployment wizard (namespace passed through for the length check).

**Delta from autofix:** same functional change rebased onto current `main` as a single commit; added an eslint disable for pre-existing PatternFly `Modal` imports in `ManageKServeModal` (ContentModal migration out of scope).

## How Has This Been Tested?
- Unit coverage from the ported autofix work (`packages/ui-core` K8sNameDescriptionField / utils / HelperTextItemVariants tests; model-serving CreateConnectionInputFields expectations).
- Manual cluster verify not yet done (post-image checklist below).

## Test Impact
- Added/updated unit tests for `routeNameTooLong` validation, helper text, and related call-site state.

## Request review criteria:

If you have UI changes:

### Manual test steps
1. In a project with a long namespace, create/edit a workbench or model deployment name such that `{k8sName}-{namespace}` exceeds 63 characters.
2. Confirm the form shows the route-name-too-long helper/error and blocks submit.
3. Shorten the name (or use a shorter namespace) and confirm validation clears and create succeeds.

## Summary by CodeRabbit

* **New Features**
  * Added validation for generated route names that exceed Kubernetes DNS label limits.
  * Name fields now display clear inline guidance when route names are too long.
  * Route-name validation is automatically applied across relevant resource creation and deployment workflows.

* **Bug Fixes**
  * Validation now updates correctly when project namespaces or resource names change.


packages/k8s-core/src/k8sNameDescriptionFieldTypes.ts
```diff
@@ -29,6 +29,10 @@ export type K8sNameDescriptionFieldData = {
       staticPrefix?: boolean;
       /** Max length for the K8s name */
       maxLength: number;
+      /** If the route name ({name}-{namespace}) exceeds the 63-character DNS label limit */
+      routeNameTooLong: boolean;
+      /** The namespace used for route name validation */
+      namespace?: string;
```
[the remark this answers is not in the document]
danreed-rh: Valid — fixed in `91b6f990a`. Doc comment now says `{k8sName}-{namespace}`.

packages/ui-core/src/components/K8sNameDescriptionField/__tests__/utils.spec.ts
```diff
@@ -2,13 +2,14 @@ import {
...
+  isRouteNameTooLong,
   LimitNameResourceType,
   resourceTypeLimits,
   setupDefaults,
 } from '@odh-dashboard/k8s-core';
 import type { K8sNameDescriptionFieldData } from '@odh-dashboard/k8s-core';
-import { mockProjectK8sResource } from '#~/__mocks__';
-import { mockK8sNameDescriptionFieldData } from '#~/__mocks__/mockK8sNameDescriptionFieldData';
+import { mockProjectK8sResource } from '@odh-dashboard/internal/__mocks__/mockProjectK8sResource';
+import { mockK8sNameDescriptionFieldData } from '@odh-dashboard/internal/__mocks__/mockK8sNameDescriptionFieldData';
```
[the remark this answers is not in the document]
danreed-rh: Valid — fixed in `91b6f990a`.

`utils.spec.ts` now uses local factories in `packages/ui-core/.../__tests__/mocks.ts` instead of `@odh-dashboard/internal`.

packages/k8s-core/src/k8sNameDescriptionFieldUtils.ts
```diff
@@ -55,6 +85,22 @@ export const resourceTypeLimits: Record<LimitNameResourceType, number> = {
...
 
+/**
+ * Returns true when the auto-generated route hostname label
+ * `{k8sName}-{namespace}` would exceed the 63-character DNS label limit.
+ * This assumes the default hostname template (`{Name}-{Namespace}`) used by
+ * both the Notebook controller and KServe; see {@link MAX_ROUTE_NAME_LENGTH}
+ * for scope and caveats.
+ */
+export const isRouteNameTooLong = (k8sName: string, namespace?: string): boolean => {
+  if (!namespace || k8sName.length === 0) {
```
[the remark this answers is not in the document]
danreed-rh: Skipped — intentional.

Empty string and `undefined` both mean “namespace not available yet” (e.g. `ManageKServeModal` before project context resolves). Treating `''` as a real namespace would validate `{k8sName}-` and can false-positive while the project is still loading.

Added an explicit unit assertion documenting that `isRouteNameTooLong(name, '')` returns `false`.

packages/ui-core/src/components/K8sNameDescriptionField/K8sNameDescriptionField.tsx
```diff
@@ -26,6 +32,34 @@ export const useK8sNameDescriptionFieldData = (
     setupDefaults(configuration),
   );
 
+  // Re-sync validation when namespace changes after initial render
+  const { namespace, limitNameResourceType } = configuration;
+  const isRouteBased =
+    limitNameResourceType != null && ROUTE_BASED_RESOURCE_TYPES.has(limitNameResourceType);
+  React.useEffect(() => {
```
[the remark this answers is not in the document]
danreed-rh: Valid — fixed in `91b6f990a`.

`onDataChange` now passes the current configuration `namespace` into `handleUpdateLogic` for route-based resources, so typing while the project/namespace is resolving no longer uses stale state. The existing `useEffect` still covers namespace-only updates with no keystroke.

Also added a unit test for the namespace override path.

packages/ui-core/src/components/K8sNameDescriptionField/K8sNameDescriptionField.tsx
```diff
@@ -109,13 +143,20 @@ const K8sNameDescriptionField: React.FC<K8sNameDescriptionFieldProps> = ({
...
+                        : 'default'
+                    }
+                  >
                     The resource name will be <b>{k8sName.value}</b>.
                     {k8sName.state.invalidCharacters &&
                       ` ${
                         k8sName.state.invalidCharsMessage ||
                         'Must start and end with a letter or number. Valid characters include lowercase letters, numbers, and hyphens (-).'
                       }`}
+                    {k8sName.state.routeNameTooLong && ` ${ROUTE_NAME_TOO_LONG_MESSAGE}.`}
```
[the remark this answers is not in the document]
danreed-rh: Skipped for this PR — minor UX polish.

Collapsed helper text concatenating two errors is pre-existing pattern adjacency, not required for the route-length AC. Prefer a follow-up if we want separate `HelperTextItem` rows in the collapsed path.

packages/ui-core/src/components/K8sNameDescriptionField/ResourceNameField.tsx
```diff
@@ -53,6 +54,7 @@ const ResourceNameField: React.FC<ResourceNameFieldProps> = ({
   if (
     k8sName.state.invalidLength ||
     k8sName.state.invalidCharacters ||
+    k8sName.state.routeNameTooLong ||
```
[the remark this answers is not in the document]
danreed-rh: Skipped for this PR — intentional edit-mode behavior.

Immutable resources correctly do not block submit (you cannot rename). Adding informational helper text for an already-too-long existing route is useful UX but out of scope for this bug fix; fine as follow-up.

danreed-rh:

DaoDaoNoCode:

srtanish1992:
@danreed-rh The ticket also mentions Pipelines, are you planning to tackle in a separate ticket? I do no see the new validation for pipelines.

Screenshot 2026-07-24 at 11 48 10 AM

danreed-rh:
> @danreed-rh The ticket also mentions Pipelines, are you planning to tackle in a separate ticket? I do no see the new validation for pipelines.


Pipelines are out of scope for [RHOAIENG-56820](https://redhat.atlassian.net/browse/RHOAIENG-56820) because user-entered pipeline names do not create OpenShift routes as {pipelineName}-{namespace}; that DNS-label failure mode applies to Notebook/KServe route-backed resources. Any DSPA route failures from long project names are a separate concern.

[RHOAIENG-56820]: https://redhat.atlassian.net/browse/RHOAIENG-56820?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

srtanish1992:
> > @danreed-rh The ticket also mentions Pipelines, are you planning to tackle in a separate ticket? I do no see the new validation for pipelines.
> 
> Pipelines are out of scope for [RHOAIENG-56820](https://redhat.atlassian.net/browse/RHOAIENG-56820) because user-entered pipeline names do not create OpenShift routes as {pipelineName}-{namespace}; that DNS-label failure mode applies to Notebook/KServe route-backed resources. Any DSPA route failures from long project names are a separate concern.

Makes sense. Can you copy/paste this as a comment in the ticket so that the person who created it is also aware about it?

[RHOAIENG-56820]: https://redhat.atlassian.net/browse/RHOAIENG-56820?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

packages/ui-core/src/components/K8sNameDescriptionField/HelperTextItemVariants.tsx
```diff
@@ -24,6 +27,21 @@ export const HelperTextItemMaxLength: HelperTextItemType = ({ k8sName }) => {
...
+  if (!k8sName.state.namespace) {
+    return null;
+  }
+
+  let variant: Variants = 'indeterminate';
+  if (k8sName.state.routeNameTooLong) {
+    variant = 'error';
+  } else if (k8sName.value.trim().length > 0) {
+    variant = 'success';
+  }
```
d0w, line 40: anyway to use constants/enums for the variant here? I'm unsure if the project has them or needs it though since they're PF props.

packages/ui-core/src/components/K8sNameDescriptionField/K8sNameDescriptionField.tsx
```diff
@@ -109,13 +150,20 @@ const K8sNameDescriptionField: React.FC<K8sNameDescriptionFieldProps> = ({
             {!showK8sField && !k8sName.state.immutable && (
               <>
                 {k8sName.value && (
-                  <HelperTextItem variant={k8sName.state.invalidCharacters ? 'error' : 'default'}>
+                  <HelperTextItem
+                    variant={
+                      k8sName.state.invalidCharacters || k8sName.state.routeNameTooLong
+                        ? 'error'
+                        : 'default'
```
d0w, line 165: same here

d0w (commented):
I was having some issues getting this to work properly since there is another limit of 30 characters max for resource names (aka fail only on >63 characters). 

Steps:

1. Create new project named `really-long-namespace-that-is-too-long` 
     - This gets truncated to `really-long-namespace-that-is` (29 char)
2. Create new workbench within project with edit resource name: 
     - `really-long-namespace-that-iss` passes (30 char)
     - `really-long-namespace-that-isss` does not pass (31 char)

So it's giving me errors at 29 + 1 + 31 = 61 characters instead of 64+. Not sure if there's extra characters that get added somewhere internally but that's what I encountered. I'm not sure if the 30 character limit needs to be removed but no matter if you max the namespace name length, you wouldn't be allowed to reach the 63 character limit.

packages/ui-core/src/components/K8sNameDescriptionField/HelperTextItemVariants.tsx
danreed-rh: Yea I think it could go either way, the risk of making it into a constant is that it is dependent on the PF component, so the risk is that other areas of odh use the constant 'error' and 'success' text outside of the PF component, but PF changes their props to accept something different than 'success' and 'error' but you can't change the constant because it's tied to other places. It's not a huge deal, it could go either way

packages/ui-core/src/components/K8sNameDescriptionField/K8sNameDescriptionField.tsx
danreed-rh: I responded to this concern above, so I'll use this spot to respond to the overall comment. 
The way to test is by creating longer project names via the cli on the cluster and then try to add a workspace to those longer names in the UI

d0w: Got it, it validates correctly via that method. Not sure if there's a particular reason why the namespaces get truncated when created through the odh UI but if there's an outdated reason then it might be something to look at in the future?

packages/ui-core/src/components/K8sNameDescriptionField/K8sNameDescriptionField.tsx
```diff
@@ -26,9 +32,44 @@ export const useK8sNameDescriptionFieldData = (
...
+
+  const onDataChange = React.useCallback<K8sNameDescriptionFieldUpdateFunction>(
+    (key, value) => {
+      setData((currentData) =>
+        // Prefer the current configuration namespace over state so typing while the
+        // project/namespace is resolving does not use a stale route-length check.
+        handleUpdateLogic(currentData, isRouteBased ? { namespace } : undefined)(key, value),
+      );
+    },
+    [isRouteBased, namespace],
```
DaoDaoNoCode: **Suggestion (S1): `onDataChange` lost referential stability**

The old `onDataChange` had `[]` deps (stable reference for the entire hook lifecycle). Adding `[isRouteBased, namespace]` means every namespace change produces a new function identity.

At least one consumer uses this as a `useEffect` dependency — `ManageKServeModal.tsx` line 199:
```ts
}, [createDataInferenceService.name, setKserveNameDesc]);
```
A new `setKserveNameDesc` reference causes that effect to re-fire. The existing guard (`createDataInferenceService.name !== kServeNameDesc.name`) prevents functional bugs, and namespace typically changes at most once, so this is not a blocker.

Worth noting in the PR description as a behavioral contract change — future consumers should not assume `onDataChange` is referentially stable.

packages/ui-core/src/components/K8sNameDescriptionField/K8sNameDescriptionField.tsx
```diff
@@ -26,9 +32,44 @@ export const useK8sNameDescriptionFieldData = (
...
+          ...currentData.k8sName,
+          state: {
+            ...currentData.k8sName.state,
+            namespace,
+            routeNameTooLong: isRouteNameTooLong(currentData.k8sName.value, namespace),
+          },
+        },
+      };
+    });
+  }, [namespace, isRouteBased]);
```
DaoDaoNoCode, line 61: **Suggestion (S2): No hook-level test for this namespace resync `useEffect`**

This effect re-validates `routeNameTooLong` when `namespace` changes asynchronously (e.g. user selects a project from a dropdown). It is the only path that re-syncs route validation after initial render.

There are no tests that call `useK8sNameDescriptionFieldData` via `renderHook`, change the `namespace` prop via `rerender()`, and assert that `routeNameTooLong` updates. The underlying `isRouteNameTooLong` function is well-tested, so the risk is low — but a silent regression in this effect wiring would let invalid names through to the Kubernetes API.

A follow-up `renderHook` test would cover this critical async path.

packages/k8s-core/src/k8sNameDescriptionFieldUtils.ts
```diff
@@ -111,9 +164,15 @@ export const setupDefaults = ({
...
-  (existingData: K8sNameDescriptionFieldData): K8sNameDescriptionFieldUpdateFunctionInternal =>
+  (
+    existingData: K8sNameDescriptionFieldData,
+    /** When set, use this namespace for route checks instead of stale state (prop may resolve async). */
+    options?: { namespace?: string },
+  ): K8sNameDescriptionFieldUpdateFunctionInternal =>
   (key, value) => {
     const changedData: RecursivePartial<K8sNameDescriptionFieldData> = {};
+    const namespace =
+      options !== undefined ? options.namespace : existingData.k8sName.state.namespace;
```
DaoDaoNoCode, line 175: **Suggestion (S3): No test for the `options: { namespace: undefined }` path**

When `isRouteBased` is `true` but `namespace` has not resolved yet, the hook passes `{ namespace: undefined }`. This line resolves it to `undefined` (not the stale state namespace), correctly disabling route checking.

This path works correctly but has no test guarding it. A test like:
```ts
const result = handleUpdateLogic(defaults, { namespace: undefined })(
  'name', 'my workbench'
);
expect(result.k8sName.state.routeNameTooLong).toBe(false);
```
would protect against future refactors that might accidentally fall through to the stale state namespace.

DaoDaoNoCode (commented):
Review from 5-agent parallel analysis (core logic, React hooks, UI/UX, call-site integration, test quality). **0 blockers, 6 suggestions.** Details in inline threads below.

packages/model-serving/modelRegistry/useNavigateToDeploymentWizardWithData.ts
```diff
@@ -65,6 +66,7 @@ export const useNavigateToDeploymentWizardWithData = (
             invalidCharacters: false,
             invalidLength: false,
             maxLength,
+            routeNameTooLong: isRouteNameTooLong(resourceName, undefined),
```
DaoDaoNoCode: **Suggestion (S4): Could simplify to just `false`**

`isRouteNameTooLong(resourceName, undefined)` always returns `false` by definition — the function early-returns when namespace is falsy. At this prefill stage the namespace is not yet known (user selects a project in the wizard's first step), so the real validation kicks in later via the `useEffect` in `useK8sNameDescriptionFieldData`.

Using the function call is slightly misleading — it suggests a real computation is happening. A plain `false` would be clearer about the intent.

nit — not blocking.

packages/ui-core/src/components/K8sNameDescriptionField/K8sNameDescriptionField.tsx
```diff
@@ -26,9 +32,44 @@ export const useK8sNameDescriptionFieldData = (
...
   );
 
-  const onDataChange = React.useCallback<K8sNameDescriptionFieldUpdateFunction>((key, value) => {
-    setData((currentData) => handleUpdateLogic(currentData)(key, value));
-  }, []);
+  // Re-sync validation when namespace changes after initial render
+  const { namespace, limitNameResourceType } = configuration;
+  const isRouteBased =
+    limitNameResourceType != null && ROUTE_BASED_RESOURCE_TYPES.has(limitNameResourceType);
+  React.useEffect(() => {
```
DaoDaoNoCode, line 39: **Suggestion (S5, pre-existing): Display name `TextInput` lacks `validated` prop for a11y**

When `showK8sField` is false, route-name-too-long errors appear as helper text under the display name input (~line 153), but the `TextInput` itself (~line 120) never gets `validated={ValidatedOptions.error}`, so `aria-invalid="true"` is never set. Screen readers won't announce the error state.

This is pre-existing — the same gap exists for `invalidCharacters` errors. The PR does not regress accessibility. Worth a follow-up ticket.

packages/ui-core/src/components/K8sNameDescriptionField/__tests__/utils.spec.ts
```diff
@@ -495,4 +496,209 @@ describe('isK8sNameDescriptionDataValid', () => {
...
+        mockK8sNameDescriptionFieldData({
+          name: 'test',
+          k8sName: {
+            value: 'my-long-resource-name',
+            state: { routeNameTooLong: true, immutable: true },
+          },
+        }),
+      ),
+    ).toBe(true);
+  });
```
DaoDaoNoCode, line 526: **Suggestion (S6): Missing test combination — `routeNameTooLong + invalidCharacters + immutable`**

The `immutable` exemption in `isK8sNameDescriptionDataValid` only applies to `routeNameTooLong`, not to `invalidCharacters`. A test asserting:
```ts
{ routeNameTooLong: true, invalidCharacters: true, immutable: true }
```
→ returns `false` would guard against future refactors that accidentally widen the `immutable` exemption to cover other validation flags.

packages/ui-core/src/components/K8sNameDescriptionField/K8sNameDescriptionField.tsx
danreed-rh: Addressed — onDataChange now uses a namespaceRef and only depends on [isRouteBased], so the callback stays referentially stable across namespace changes while still reading the latest namespace on each keystroke.

packages/ui-core/src/components/K8sNameDescriptionField/K8sNameDescriptionField.tsx
danreed-rh: Addressed — added `useK8sNameDescriptionFieldData.spec.tsx` with a `testHook` / `rerender` case that asserts `routeNameTooLong` updates when `namespace` changes asynchronously (and that `onDataChange` remains stable).

packages/k8s-core/src/k8sNameDescriptionFieldUtils.ts
danreed-rh: Addressed — added a `handleUpdateLogic(defaults, { namespace: undefined })` unit test asserting `routeNameTooLong` stays `false` (does not use the stale state namespace for the check).

packages/model-serving/modelRegistry/useNavigateToDeploymentWizardWithData.ts
danreed-rh: Addressed — set `routeNameTooLong: false` at prefill time and dropped the unused `isRouteNameTooLong` import.

packages/ui-core/src/components/K8sNameDescriptionField/K8sNameDescriptionField.tsx
danreed-rh: Agreed — pre-existing a11y gap (same for `invalidCharacters` in the collapsed view). Leaving out of this PR; good candidate for a follow-up ticket to set `validated` on the display-name `TextInput` when collapsed errors are shown.

packages/ui-core/src/components/K8sNameDescriptionField/__tests__/utils.spec.ts
danreed-rh: Addressed — added a test that `immutable + routeNameTooLong + invalidCharacters` still returns `false` (immutable only exempts `routeNameTooLong`).

packages/ui-core/src/components/K8sNameDescriptionField/K8sNameDescriptionField.tsx
danreed-rh: If I had to guess it would be to avoid the problem that this ticket seeks to address, I think this ticket is addressing an edge case where the user performs the project name directly on the cluster. Wasn't around for the truncation feature though so not sure.

DaoDaoNoCode (approved):
/lgtm

PR #8764, opened 2026-07-22
fix(llmd-serving): fix routing and topology config edge cases in deploy wizard edit flow

mturley:
https://issues.redhat.com/browse/RHOAIENG-76494

## Description

Fixes multiple edge cases in `AdvancedRoutingField` and `CustomTopologyConfigField` where config selections could become stale, orphaned, or lost during the deploy wizard edit flow.

cc @vconzola

**Gap 1 — No reset on topology type change:** The routing field now resets when the topology type dependency changes via `shouldResetOnDependencyChange`, matching the existing pattern in `CustomTopologyConfigField`. Previously, if a user selected a routing config restricted to a specific topology type and then changed the topology, the stale selection persisted in field state but disappeared from the dropdown. The user could continue to submit with the invalid routing config.

**Here's the previous buggy behavior for gap 1:**

https://github.com/user-attachments/assets/8436bd08-a118-4f3f-b02f-af90c5d48b6a

**Here's the fixed behavior for gap 1:**

https://github.com/user-attachments/assets/c4abc0bc-8299-488e-a882-3e94647cc41f

**Gap 2 — Orphaned config in edit flow:** When editing a deployment whose routing config has become incompatible with the current topology type (e.g. the config's `supportedTopologies` annotation was changed after deploy), the config is now resolved against all router configs and remains visible and selectable in the dropdown. A warning is shown while the incompatible config is selected: *"The selected routing configuration is not compatible with the current topology type. Select a different routing configuration or contact your administrator."* The warning clears when the user selects a different option. The incompatible option remains in the dropdown even after switching to a different selection. If the user proceeds with the incompatible option selected despite the warning, we still allow them to submit.

**Here's the previous buggy behavior for gap 2:**

https://github.com/user-attachments/assets/c282c18d-2b60-4a78-87af-aaf29a19d855

**Here's the fixed behavior for gap 2:**

https://github.com/user-attachments/assets/25bd6198-1833-4bbe-8f19-27d3a971edd3

Here's that warning:
Screenshot 2026-07-22 at 5 39 50 PM

### Additional corner cases covered and regressions from #8718

While testing the original two gaps, we discovered that #8718 ([RHOAIENG-77671](https://redhat.atlassian.net/browse/RHOAIENG-77671)) introduced a regression: its `getInitialFieldData` fix for the deleted-config case checked `existingFieldData?.selectedConfig` and dropped the `configRef` when `selectedConfig` was absent. This broke all edit-flow config resolution for both routing and topology config fields — the saved config was never pre-selected when editing a deployment.

This PR fixes that regression and covers the following additional corner cases:

- **configRef preserved in `getInitialFieldData`:** Both fields now preserve `existingFieldData` when either `selectedConfig` or `configRef` is present, so the resolution effect can look up the config.
- **Deleted config handling:** When the resolution effect can't find a config (deleted from cluster), it clears `configRef` so downstream behavior works correctly — auto-select kicks in for topology configs, and the routing field falls back to default. For single-node topology, the resolution effect sets `TOPOLOGY_CONFIG_DEFAULT` directly since the auto-select effect only handles non-single-node topologies.
- **Auto-select dep fix:** `CustomTopologyConfigField`'s auto-select effect now includes `configRef` in its dependency array so it re-fires after a deleted configRef is cleared.
- **Stale incompatible state:** `preSelectedIncompatibleConfig` is cleared when the config becomes compatible (e.g. after a topology type change), preventing phantom options and false warnings.
- **Disabled select edge case:** When the incompatible config is the only router config, the dropdown stays enabled so the user can still select "Default optimized routing."
- **Duplicate option guard:** The incompatible config option is only appended when it's not already in `compatibleConfigs`.

cc @YuliaKrimerman — the `getInitialFieldData` change in this PR supersedes the one from #8718 for both fields.

### Possible follow-up

`CustomTopologyConfigField` has the same orphaned-config display pattern (lines 132-142) but does not show a warning when the selected config is incompatible. A follow-up ticket could apply the same warning UX there. cc @YuliaKrimerman I can open this, but I think it can go on the backlog, the broken behavior was fixed in https://github.com/opendatahub-io/odh-dashboard/pull/8549 and it just doesn't show a warning. This is a super unlikely corner case.

## How Has This Been Tested?

- **Type-check**: passes
- **Lint**: 0 errors
- **Unit tests**: 30 tests pass across both field test files:
  - `AdvancedRoutingField.spec.tsx` (18 tests): `getInitialFieldData`, `validationSchema`, `shouldResetOnDependencyChange`, component rendering for incompatible config warning, compatible configRef resolution, deleted configRef clearing
  - `CustomTopologyConfigField.spec.tsx` (12 tests): `getInitialFieldData` with configRef preservation and empty-data fallthrough, compatible configRef resolution, deleted configRef clearing with auto-select, deleted configRef on single-node sets default
- **Cypress mock tests**: 7 new integration tests in `modelServingLlmdTopology.cy.ts`, all asserting Next button is enabled:
  - Edit flow pre-selects topology config and routing config from existing deployment annotations
  - Deleted routing config falls back to default
  - Incompatible routing config shown with warning
  - Incompatible topology config retained on edit
  - Deleted topology config auto-selects next available
  - Switching to single-node after deleted topology config is resolved ([RHOAIENG-77671](https://redhat.atlassian.net/browse/RHOAIENG-77671) case 6 regression check)
- **Manual testing on zaffre cluster**: 20 test scenarios covering Gap 1 (topology change resets), Gap 2 (incompatible config warning), deleted config regression checks for both topology and routing, and normal create/edit flow regressions. Full checklist: `pr-8764-test-checklist.md`

## Test Impact

- **Added**: `AdvancedRoutingField.spec.tsx` — 18 unit/component tests (new file)
- **Modified**: `CustomTopologyConfigField.spec.tsx` — 12 tests (renamed from `.spec.ts`, added component render tests for configRef resolution and deleted config handling)
- **Added**: 7 Cypress mock tests in `modelServingLlmdTopology.cy.ts` for edit flow config resolution
- **Modified**: `mockLLMInferenceServiceK8sResource.ts` — added `additionalAnnotations` param for setting topology/routing config refs on mock deployments

## Request review criteria:

If you have UI changes:

[RHOAIENG-77671]: https://redhat.atlassian.net/browse/RHOAIENG-77671?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/llmd-serving/src/wizardFields/AdvancedRoutingField.tsx
+++ b/packages/llmd-serving/src/wizardFields/AdvancedRoutingField.tsx
@@ -110,9 +110,14 @@ const AdvancedRoutingFieldComponent: AdvancedRoutingFieldType['component'] = ({
     [routerConfigs, topologyType],
   );
 
-  // Resolve configRef from extractor (edit flow) once external data loads
+  // Resolve configRef from extractor (edit flow) once external data loads.
+  // Resolves against all routerConfigs so a previously selected incompatible config
+  // remains visible in the dropdown with a warning.
   const configRef = value?.configRef;
   const existingSelection = value?.selectedConfig;
+  const [preSelectedIncompatibleConfig, setPreSelectedIncompatibleConfig] = React.useState<
+    LLMInferenceServiceConfigKind | undefined
+  >();
   React.useEffect(() => {
     if (!configRef || existingSelection || !isLoaded) {
       return;
@@ -120,9 +125,30 @@ const AdvancedRoutingFieldComponent: AdvancedRoutingFieldType['component'] = ({
     const resolved = (routerConfigs ?? []).find((c) => c.metadata.name === configRef);
     if (resolved) {
       onChange({ selectedConfig: resolved });
+      if (!compatibleConfigs.some((c) => c.metadata.name === resolved.metadata.name)) {
+        setPreSelectedIncompatibleConfig(resolved);
+      }
+    } else {
+      onChange({ configRef: undefined });
     }
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [configRef, isLoaded, existingSelection, routerConfigs]);
+  }, [configRef, isLoaded, existingSelection, routerConfigs, compatibleConfigs]);
+
+  // Clear stale incompatible state when the config becomes compatible
+  // (e.g. topology type changed to one the config supports).
+  React.useEffect(() => {
+    if (
+      preSelectedIncompatibleConfig &&
+      compatibleConfigs.some((c) => c.metadata.name === preSelectedIncompatibleConfig.metadata.name)
+    ) {
+      setPreSelectedIncompatibleConfig(undefined);
+    }
+  }, [compatibleConfigs, preSelectedIncompatibleConfig]);
+
+  const isIncompatibleSelected =
+    !!existingSelection &&
+    !!preSelectedIncompatibleConfig &&
+    existingSelection.metadata.name === preSelectedIncompatibleConfig.metadata.name;
 
   const options: SimpleSelectOption[] = React.useMemo(() => {
     const result: SimpleSelectOption[] = [
@@ -146,8 +172,22 @@ const AdvancedRoutingFieldComponent: AdvancedRoutingFieldType['component'] = ({
       }),
     );
 
+    // Include the incompatible config only if it's not already in the compatible list
+    if (
+      preSelectedIncompatibleConfig &&
+      !compatibleConfigs.some(
+        (c) => c.metadata.name === preSelectedIncompatibleConfig.metadata.name,
+      )
+    ) {
+      result.push({
+        key: preSelectedIncompatibleConfig.metadata.name,
+        label: getDisplayNameFromK8sResource(preSelectedIncompatibleConfig),
+        dataTestId: `routing-config-option-${preSelectedIncompatibleConfig.metadata.name}`,
+      });
+    }
+
     return result;
-  }, [compatibleConfigs]);
+  }, [compatibleConfigs, preSelectedIncompatibleConfig]);
 
   const selectedValue = value?.selectedConfig?.metadata.name ?? DEFAULT_ROUTING_KEY;
 
@@ -175,7 +215,7 @@ const AdvancedRoutingFieldComponent: AdvancedRoutingFieldType['component'] = ({
                 onChange({ selectedConfig: undefined });
                 return;
               }
-              const config = compatibleConfigs.find((c) => c.metadata.name === key);
+              const config = (routerConfigs ?? []).find((c) => c.metadata.name === key);
               fireRoutingSelected({
                 routingConfigurationId: key,
                 isDefaultRouting: false,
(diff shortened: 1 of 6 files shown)


mturley:
/hold
Fixing some bot review comments

mturley:
Testing this uncovered a regression caused by #8718. Fixing it here and adding new tests, will update the description and unhold shortly

packages/llmd-serving/src/wizardFields/AdvancedRoutingField.tsx
```diff
@@ -109,19 +110,45 @@ const AdvancedRoutingFieldComponent: AdvancedRoutingFieldType['component'] = ({
...
+  // Clear stale incompatible state when the config becomes compatible
+  // (e.g. topology type changed to one the config supports).
+  React.useEffect(() => {
+    if (
+      preSelectedIncompatibleConfig &&
+      compatibleConfigs.some((c) => c.metadata.name === preSelectedIncompatibleConfig.metadata.name)
+    ) {
+      setPreSelectedIncompatibleConfig(undefined);
+    }
+  }, [compatibleConfigs, preSelectedIncompatibleConfig]);
```
emilys314, line 146: Oof `useEffect` I don't like seeing `useEffect` in field or form logic (nor the other ones that already exist). Following this code is hard and `useEffect`s introduce re-rendering lag and shifting values, which can make cypress tests fragile

we should clean this up with a tech debt ticket.

mturley: Yeah honestly the more I touch this area the more I think there must be a better way to manage all of this state for these fields. I didn't see a way around this effect without refactoring away other effects like the configRef resolver that also needed fixing here. These cypress tests are comprehensive enough that I think refactor risk wouldn't be too bad here though.

This effect in particular is only needed because of the weird "you're editing a model that has a selected config that's no longer compatible" warning, which requires us to hold on to that extra config so it can be included in the dropdown, but then we want that to disappear if it is edited to become compatible again so we don't have duplicates in the dropdown. That said though I don't think we have polling that would result in the data changing to trigger this effect anyway... hm

mturley: I think you're right that it should be a tech debt ticket

mturley: I opened https://redhat.atlassian.net/browse/RHOAIENG-79010 to cover this. Claude thinks we could get rid of not just this useEffect but the other existing ones that resolve the configRef after data arrives (by moving that logic into `getInitialFieldData` which it thinks gets re-dispatched after data is ready). That seems plausible to me, but I don't want to sink more time into this, so I left a big disclaimer at the top of that issue expressing limited confidence in the analysis. If that's wrong, the tech debt ticket can be just to remove this one useEffect which is definitely unnecessary.

mturley:
Self-approving from @YuliaKrimerman 's LGTM
/approve

YuliaKrimerman:
/lgtm
/approve
Tested both scenarios- works well 

https://github.com/user-attachments/assets/96b52dc7-39dd-42bd-8ee9-41632d36d8d2

PR #8766, opened 2026-07-23
fix(mlflow-embedded): pass workspace prop to federated MLflow wrappers

nananosirova:
https://issues.redhat.com/browse/RHOAIENG-66298

## Description

When switching projects on "Gen AI Studio > Prompts", the list view shows the wrong latest version from the previous project's React Query cache. For example, if project A has `summarization` with 8 versions and project B has `summarization` with 3 versions, the list view for project B shows "Version 8" instead of "Version 3". The detail page shows correct data.

**Root cause:** The MLflow federated module's `QueryClient` is created at module scope and survives the `key={workspace}` React remount. Since query cache keys don't include workspace, stale data from the previous workspace is served from cache.

**Fix:** Pass `workspace` as a prop to `MlflowPromptsWrapper` and `MlflowExperimentWrapper` so the MLflow module can invalidate or re-key its cache on workspace change. This is a prerequisite for the matching MLflow-side fix (adding workspace to query cache keys / invalidating on workspace prop change).

**Note:** Full end-to-end verification requires the matching MLflow-side change (accepting and handling the `workspace` prop in the federated wrappers). This PR alone does not fix the bug. The prop is silently ignored until the MLflow side lands.

## Test Impact

No new tests. The change is a one-line prop addition in two files. The existing Cypress mock test (`promptManagement.cy.ts`) covers workspace switching behavior. The actual cache invalidation logic lives in the external MLflow module and is not testable from the dashboard side.

## Request review criteria:

### DIFF
--- a/packages/mlflow-embedded/experiments/MlflowExperimentsPage.tsx
+++ b/packages/mlflow-embedded/experiments/MlflowExperimentsPage.tsx
@@ -87,7 +87,7 @@ const MlflowExperimentsPage: React.FC = () => {
       <LazyCodeRefComponent
         key={workspace}
         component={loadWrapper}
-        props={{ onBreadcrumbChange: setBreadcrumbs }}
+        props={{ onBreadcrumbChange: setBreadcrumbs, workspace }}
         fallback={
           <Bullseye>
             <Spinner />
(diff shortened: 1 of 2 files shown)


DaoDaoNoCode (commented):
I don't see this change on the main branch. Can you figure out the reason?
If it's an issue on main branch, we should fix them first then cherry-pick. If this is fixed on main branch, we need to find out how it was fixed on main branch and cherry-pick that fix.
Could you look into that?

nananosirova:
Closing — the workspace is already available to the MLflow federated component via the URL query parameter. The fix belongs entirely on the MLflow side (backporting cache invalidation from v3.13+ to v3.10.x).

PR #8780, opened 2026-07-23
feat(RHOAIENG-78652): add ProjectsContextProvider for RHAII distribution

jenny-s51:
https://issues.redhat.com/browse/RHOAIENG-78652

## Description

RHAII Models page showed "There was an issue fetching projects" because no `ProjectsContextProvider` was mounted. The main frontend uses one backed by OpenShift Project watches; RHAII needs a lighter version that fetches namespaces from the Core BFF.

**Changes:**
- `ProjectsContextProvider` for RHAII — fetches `/api/k8s/api/v1/namespaces`, maps to `ProjectKind`, exposes via `ProjectsContext`
- `AppWrapper` prop on `createDistribution` — lets distributions inject host-only providers without forking the shell
- `BrowserStorageContextProvider` added to base shell; preferred project uses `useBrowserStorage` from ui-core
- `fetchNamespaces` extracted to dedicated file with AbortController timeout and 8 unit tests
- `waitForProject` re-fetches on each poll with independent AbortController deadline
- `GearsIcon` uses PatternFly `createIcon`; removes unused `gears.svg`
- Webpack SVG fix — scopes `svg-url-loader` to `packages/*/images/` only

## How Has This Been Tested?

Ran `make tilt-up-model-serving` — verified Models page loads, logo renders, gears icon visible in dark theme. 8 unit tests for `fetchNamespaces`.

## Request review criteria:

- [x] Testing instructions have been added in the PR body
- [x] The code follows our [Best Practices](/docs/best-practices.md)


packages/model-serving/src/components/projectDetails/GearsIcon.tsx
```diff
@@ -0,0 +1,31 @@
+import * as React from 'react';
+
+const GearsIcon: React.FC = () => (
+  <svg
```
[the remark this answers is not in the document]
jenny-s51: Resolved.

caponetto:
@jenny-s51 could you please check if there is anything we need to do with these [findings](https://gist.github.com/caponetto/eca7ded96a2b25ad2ee826bbdfb35448)?

distributions/rhaii/src/context/ProjectsContextProvider.tsx
```diff
@@ -0,0 +1,247 @@
...
+      }
+    };
+
+    setLoaded(false);
+    void loadNamespaces();
+
+    return () => {
+      cancelled = true;
+    };
+  }, []);
```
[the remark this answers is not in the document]
jenny-s51: Skipping — network-level timeouts cover the mount fetch, and the catch/finally flow already surfaces errors. The real hang risk (waitForProject) was fixed with a 30s timeout.

distributions/base/config/webpack.common.js
```diff
@@ -65,18 +65,43 @@ module.exports = ({
...
+          // Base shell logos (distributions/base/src/images/) intentionally use
+          // raw-loader so ShellHeader can encodeURIComponent the markup itself.
+          include: (input) => {
+            if (input.indexOf('bgimages') > -1) {
+              return true;
+            }
+            const packagesDir = path.join(REPO_ROOT, 'packages');
+            return (
+              input.startsWith(packagesDir) && input.indexOf(`${path.sep}images${path.sep}`) > -1
+            );
```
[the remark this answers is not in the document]
jenny-s51: Skipping since src/images is the actual target path (e.g. packages/model-serving/src/images/). The predicate is correct as-is.

distributions/rhaii/src/context/ProjectsContextProvider.tsx
```diff
@@ -0,0 +1,264 @@
...
+  type ProjectsContextType,
+} from '@odh-dashboard/ui-core/context/ProjectsContext';
+import type { ProjectKind } from '@odh-dashboard/k8s-core';
+import { byName } from '@odh-dashboard/k8s-core';
+
+const PREFERRED_NAMESPACE_STORAGE_KEY = 'mod-arch.namespace.lastUsed';
+/** Dashboard install namespace — excluded from the selectable project list. */
+const DASHBOARD_NAMESPACE = 'opendatahub';
+
+const isAvailableProject = (projectName: string, dashboardNamespace: string): boolean =>
```
caponetto: We already have a `isAvailableProject` function in `k8s-core`. Can we avoid this duplication?

distributions/rhaii/src/context/ProjectsContextProvider.tsx
```diff
@@ -0,0 +1,264 @@
...
+  );
+
+type K8sNamespaceItem = {
+  metadata?: { name?: string };
+  status?: { phase?: string };
+};
+
+type K8sNamespaceList = {
+  items?: K8sNamespaceItem[];
+};
```
caponetto: Should these live in `k8s-core`? same for `parseNamespaceList` and `namespaceToProject`.

distributions/rhaii/src/context/ProjectsContextProvider.tsx
```diff
@@ -0,0 +1,264 @@
...
+
+const isRecord = (value: unknown): value is Record<string, unknown> =>
```
caponetto: We have `isRecord` duplicated ~12 times across the repository. Can we open an FUP ticket to centralize it within one of the core packages?

distributions/rhaii/src/context/ProjectsContextProvider.tsx
```diff
@@ -0,0 +1,264 @@
...
+const namespaceToProject = (name: string, phase?: string): ProjectKind => ({
+  apiVersion: 'v1',
+  kind: 'Namespace',
+  metadata: { name },
+  status: {
+    phase: phase === 'Terminating' ? 'Terminating' : 'Active',
+  },
+});
+
+const fetchNamespaces = async (): Promise<ProjectKind[]> => {
```
caponetto: Can we extract this to a dedicated file and have some unit tests for it?

caponetto: `fetchNamespaces()` has no `AbortController` or timeout. A stalled request prevents `loaded` from ever becoming `true`, leaving downstream consumers (e.g., project selectors) stuck in loading state with no error or retry path. The shell itself renders fine — only context consumers are affected. The provider's own `waitForProject()` already applies a 30 s bound to a similar operation.

**Suggestion:** Add an `AbortController`-based timeout matching the 30 s `waitForProject` deadline, so a stall triggers the `catch` path and surfaces `loadError` to consumers.

distributions/base/src/createDistribution.tsx
```diff
@@ -35,6 +42,8 @@ const DistributionApp: React.FC<{ config: DistributionConfig }> = ({ config }) =
       </PluginStoreProvider>
     </ThemeProvider>
   );
+
+  return AppWrapper ? <AppWrapper>{app}</AppWrapper> : app;
```
caponetto: `AppWrapper` currently wraps *outside* `ThemeProvider` in `createDistribution.tsx`. Fine for `ProjectsContextProvider` (pure data, no UI), but any future wrapper that renders its own UI (loading gate, error banner) would be unthemed. Worth moving the slot inside `ThemeProvider` if a second distribution adopts the pattern.

distributions/rhaii/src/context/ProjectsContextProvider.tsx
jenny-s51: Nice catch, thanks! Fixed - importing from k8s-core now.

distributions/base/src/createDistribution.tsx
jenny-s51: Ah yep, good forward-looking note - thank you! Moved it inside `ThemeProvider`.

distributions/rhaii/src/context/ProjectsContextProvider.tsx
jenny-s51: Good idea, done - extracted to fetchNamespaces.ts with 7 unit tests.

jenny-s51: Thank you, great catch! Added AbortController with 30s timeout

distributions/rhaii/src/context/ProjectsContextProvider.tsx
jenny-s51: After investigating a bit  more here , since these are specific to parsing the raw /api/v1/namespaces response (which only the RHAII distribution uses) it seems they make sense here for now. WDYT @caponetto ? Happy to move them to k8s-core if another consumer comes up.

distributions/rhaii/src/context/ProjectsContextProvider.tsx
jenny-s51: Sure thing! Tracked in https://redhat.atlassian.net/browse/RHOAIENG-79533.

distributions/rhaii/src/context/fetchNamespaces.ts
```diff
@@ -0,0 +1,56 @@
...
+    throw new Error(`Failed to list namespaces (HTTP ${resp.status})`);
+  }
+  const data = parseNamespaceList(await resp.json());
+  return (data.items ?? [])
+    .map((item) => {
+      const name = item.metadata?.name;
+      if (!name) {
+        return null;
+      }
+      return namespaceToProject(name, item.status?.phase);
```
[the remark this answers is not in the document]
jenny-s51: Skipping  since we already guard with if (!name) which catches empty strings (falsy). K8s namespace names are always strings from the API.

distributions/rhaii/src/context/ProjectsContextProvider.tsx
```diff
@@ -0,0 +1,207 @@
...
+            }
+          } catch {
+            // fetch failed — keep polling until timeout
+          }
+          if (Date.now() >= deadline) {
+            reject(new Error(`Timed out waiting for project "${projectName}"`));
+            return;
+          }
+          setTimeout(() => void poll(), WAIT_FOR_PROJECT_POLL_MS);
+        };
```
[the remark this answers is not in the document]
jenny-s51: The BFF is local so requests will succeed or fail quickly - the 30s deadline check runs after each fetch.

distributions/rhaii/src/context/fetchNamespaces.ts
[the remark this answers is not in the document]
jenny-s51: Ok, thank you for the clarification - fixed.

distributions/rhaii/src/context/ProjectsContextProvider.tsx
caponetto: Yes, it makes sense. Let's keep them here! Thanks for looking into it.

distributions/rhaii/src/context/ProjectsContextProvider.tsx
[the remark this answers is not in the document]
jenny-s51: Fair point - waitForProject now uses its own AbortController with an independent 30s timer that aborts any hanging fetch and rejects the promise.

distributions/rhaii/src/context/ProjectsContextProvider.tsx
```diff
@@ -0,0 +1,214 @@
...
+  if (!raw) {
+    return null;
+  }
+  try {
+    const parsed: unknown = JSON.parse(raw);
+    return typeof parsed === 'string' ? parsed : null;
+  } catch {
+    return raw.length > 0 ? raw : null;
+  }
+};
```
caponetto: One last comment (not a blocker -- just remembered this): Can we use `useBrowserStorage` from `ui-core` instead of creating `readStoredPreferredName`?

jenny-s51: Sure thing - switched to useBrowserStorage from ui-core and added BrowserStorageContextProvider to the base shell so all distributions get it.

distributions/base/package.json
```diff
@@ -21,6 +21,7 @@
   },
   "dependencies": {
     "@odh-dashboard/plugin-core": "*",
+    "@odh-dashboard/ui-core": "*",
```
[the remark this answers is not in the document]
jenny-s51: Internal dependencies in this monorepo use "*" — that's the established convention. 

Check any other package.json:

"@odh-dashboard/plugin-core": "*",
"@odh-dashboard/k8s-core": "*",

The workspace resolver handles version resolution. Wouldn't changing one package to a pinned version be inconsistent with the rest of the repo?

PR #8782, opened 2026-07-23
fix(RHOAIENG-78548): allow topology type changes in router config edit mode

YuliaKrimerman:
## Description
https://issues.redhat.com/browse/RHOAIENG-78548

The "Topology type" field in the routing configuration create/edit page was always disabled in edit mode (`isDisabled={isEditMode}`). This prevented users from changing the topology type and also blocked saving when the field had no value, since the save button requires a topology selection.

Removed the `isDisabled` prop from the topology type `SimpleSelect` so the field is always editable regardless of mode.

## How Has This Been Tested?

Screenshot 2026-07-23 at 12 01 30 PM

- Added unit tests verifying the topology type select is not disabled in both create and edit modes
- Ran existing test suite — all tests pass
- Lint and type-check pass clean

## Test Impact
Added `RoutingConfigurationCreateEdit.spec.tsx` with tests for create and edit modes verifying the topology type field is never disabled.

## Request review criteria:

## Summary by CodeRabbit

- **Bug Fixes**
  - The topology type selector remains available when editing an existing routing configuration.
  - Users can change the selected topology type in both new and existing routing configurations.

### DIFF
--- a/packages/llmd-serving/src/settings/__tests__/RoutingConfigurationCreateEdit.spec.tsx
+++ b/packages/llmd-serving/src/settings/__tests__/RoutingConfigurationCreateEdit.spec.tsx
@@ -1,0 +1,102 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+import '@testing-library/jest-dom';
+import { useNavigate, useParams, useLocation } from 'react-router';
+import { mockLLMInferenceServiceConfigK8sResource } from '@odh-dashboard/internal/__mocks__/mockLLMInferenceServiceConfigK8sResource';
+import { TopologyType } from '../../types';
+import { useWatchRouterConfigs } from '../../api/LLMInferenceServiceConfigs';
+import RoutingConfigurationCreateEdit from '../RoutingConfigurationCreateEdit';
+
+jest.mock('react-router', () => ({
+  Link: ({ children, to }: { children: React.ReactNode; to: string }) => (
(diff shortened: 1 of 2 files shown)


packages/llmd-serving/src/settings/__tests__/RoutingConfigurationCreateEdit.spec.tsx
```diff
@@ -0,0 +1,102 @@
...
+      const topologySelect = screen.getByTestId('topology-type-select');
+      expect(topologySelect).not.toBeDisabled();
+    });
+  });
+
+  describe('edit mode', () => {
+    const existingConfig = mockLLMInferenceServiceConfigK8sResource({
+      name: 'test-router',
+      displayName: 'Test Router',
+      configType: 'router' as never,
```
mturley, line 86: You should be able to just import and use `MockConfigType.ROUTER` instead of this `as never`

mturley (commented):
/lgtm
Nit though, feel free to self-approve if you want to move on, it's harmless in this test

packages/llmd-serving/src/settings/__tests__/RoutingConfigurationCreateEdit.spec.tsx
YuliaKrimerman: Decided to approve to not mess up with passing CI . Thanks

PR #8786, opened 2026-07-23
feat(dashboard-operator): add ConfigMap content hash for federation rolling restart [RHOAIENG-78802]

lucferbux:
https://issues.redhat.com/browse/RHOAIENG-78802
https://issues.redhat.com/browse/RHOAIENG-59931

## Description

When modules are enabled or disabled, the operator updates the `federation-config` ConfigMap but pods only pick up the new values on restart (the ConfigMap is consumed via `configMapKeyRef` env var, not a mounted volume). This PR adds a content-hash annotation on the Deployment's pod template so Kubernetes triggers a rolling restart whenever the ConfigMap content changes.

Changes:
- **Drop `Enabled` field** from `federationEntry` struct — the frontend `ModuleFederationConfig` TypeScript type never had an `enabled` property; this was dead code.
- **Add `mainDashboardDeploymentName()`** — platform-aware helper returning `odh-dashboard` (ODH) or `rhods-dashboard` (RHOAI).
- **Add `computeFederationConfigHash()`** — pure SHA256 hex digest of the serialized ConfigMap data.
- **Add `patchDeploymentFederationHash()`** — reads the live Deployment, compares the current annotation, and patches only when the hash differs (no-op otherwise).
- **Wire into `reconcileStandalone()`** — after deploying the federation ConfigMap, patch the Deployment annotation. Failure is non-fatal (logged, retried next reconcile cycle).
- **Return config data from `deployFederationConfigMap()`** — signature changed from `error` to `(string, error)` to pass the serialized JSON to the hash function.

## How Has This Been Tested?

- `make test` in `dashboard-operator/` — all tests pass (8 new test cases + existing suite)
- `make lint` — 0 issues
- `make build` — compiles cleanly

## Test Impact

New test file: `dashboard-operator/internal/controller/module_deploy_test.go` with 8 test cases:
- `TestComputeFederationConfigHash_Deterministic` — same input → same hash
- `TestComputeFederationConfigHash_DifferentInputs` — different inputs → different hashes
- `TestMainDashboardDeploymentName` — platform-aware naming (ODH, SelfManagedRhoai, ManagedRhoai, XKS)
- `TestBuildFederationConfigMap_ExcludesDisabledModules` — disabled/not-deployed modules excluded from output
- `TestBuildFederationConfigMap_NoEnabledField` — JSON output contains no `"enabled"` key
- `TestPatchDeploymentFederationHash_CreatesAnnotation` — first call adds annotation
- `TestPatchDeploymentFederationHash_NoOpWhenUnchanged` — same hash → no patch
- `TestPatchDeploymentFederationHash_UpdatesOnChange` — different hash → annotation updated

Updated `export_test.go` with test-only exports for `ComputeFederationConfigHash`, `MainDashboardDeploymentName`, `BuildFederationConfigMap`, and `PatchDeploymentFederationHash`.

## Request review criteria:

## Summary by CodeRabbit

* **New Features**
  * Dashboard deployments now compute and apply a federation hash to the main Deployment’s pod template annotations, updating configurations and triggering a rollout when the hash changes.
  * Federation configuration generation excludes disabled or not-deployed modules.
  * Federation config output no longer includes an “enabled” field; deployment naming now adapts to the platform.
* **Bug Fixes**
  * Federation hash patch failures are no longer silently ignored; they correctly fail the reconcile cycle.
* **Tests**
  * Added coverage for federation hash determinism, platform-specific naming, configuration filtering, and annotation patch behavior (including no-op cases).

### DIFF
--- a/dashboard-operator/internal/controller/dashboard_reconciler.go
+++ b/dashboard-operator/internal/controller/dashboard_reconciler.go
@@ -440,6 +440,7 @@ func (r *DashboardReconciler) reconcileStandalone(
 
 	if err := r.patchDeploymentFederationHash(ctx, fedData); err != nil {
 		logger.Error(err, "Failed to patch federation hash on deployment")
+		return ctrl.Result{}, fmt.Errorf("patching federation hash: %w", err)
 	}
 
 	// Step 8: URL extraction + degraded condition
--- a/dashboard-operator/internal/controller/dashboard_reconciler.go
+++ b/dashboard-operator/internal/controller/dashboard_reconciler.go
@@ -429,12 +429,17 @@ func (r *DashboardReconciler) reconcileStandalone(
 	r.reconcileObservability(ctx, dashboard, cm)
 
 	// Step 7: Build and deploy federation ConfigMap (critical for standalone routing)
-	if err := r.deployFederationConfigMap(ctx, nextStatuses, dashboard); err != nil {
+	fedData, err := r.deployFederationConfigMap(ctx, nextStatuses, dashboard)
+	if err != nil {
 		cm.MarkTrue(string(common.ConditionTypeDegraded),
 			conditions.WithReason("FederationConfigMapFailed"),
 			conditions.WithError(err))
 		logger.Error(err, "Failed to deploy federation ConfigMap")
 		return ctrl.Result{}, fmt.Errorf("federation ConfigMap: %w", err)
+	}
+
+	if err := r.patchDeploymentFederationHash(ctx, fedData); err != nil {
+		logger.Error(err, "Failed to patch federation hash on deployment")
 	}
 
 	// Step 8: URL extraction + degraded condition
--- a/dashboard-operator/internal/controller/module_deploy.go
+++ b/dashboard-operator/internal/controller/module_deploy.go
@@ -2,6 +2,8 @@ package controller
 
 import (
 	"context"
+	"crypto/sha256"
+	"encoding/hex"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -52,7 +54,6 @@ type federationEntry struct {
 	RemoteEntry  string              `json:"remoteEntry,omitempty"`
 	Authorize    bool                `json:"authorize"`
 	TLS          bool                `json:"tls"`
-	Enabled      bool                `json:"enabled"`
 	Proxy        []proxyRoute        `json:"proxy,omitempty"`
 	Service      *serviceRef         `json:"service,omitempty"`
 	ProxyService []proxyServiceEntry `json:"proxyService,omitempty"`
@@ -106,6 +107,58 @@ func standaloneServiceName(platform cluster.Platform, slug string) string {
 		prefix = "rhods-dashboard"
 	}
 	return prefix + "-" + slug + "-ui"
+}
+
+func mainDashboardDeploymentName(platform cluster.Platform) string {
+	if platform == cluster.SelfManagedRhoai || platform == cluster.ManagedRhoai {
+		return "rhods-dashboard"
+	}
+	return "odh-dashboard"
+}
+
+// --- Federation config hash for rolling restart ---
+
+const federationHashAnnotation = "dashboard.opendatahub.io/federation-config-hash"
+
+func computeFederationConfigHash(data string) string {
+	h := sha256.Sum256([]byte(data))
+	return hex.EncodeToString(h[:])
+}
+
+func (r *DashboardReconciler) patchDeploymentFederationHash(
+	ctx context.Context,
+	configData string,
+) error {
+	hash := computeFederationConfigHash(configData)
+
+	deployName := mainDashboardDeploymentName(r.Platform)
+	var deploy appsv1.Deployment
+	key := client.ObjectKey{Name: deployName, Namespace: r.ApplicationsNamespace}
+	if err := r.Get(ctx, key, &deploy); err != nil {
+		return fmt.Errorf("getting deployment %s: %w", deployName, err)
+	}
+
+	current := ""
+	if deploy.Spec.Template.Annotations != nil {
+		current = deploy.Spec.Template.Annotations[federationHashAnnotation]
+	}
+	if current == hash {
+		return nil
+	}
+
+	patch := client.MergeFrom(deploy.DeepCopy())
+	if deploy.Spec.Template.Annotations == nil {
+		deploy.Spec.Template.Annotations = map[string]string{}
+	}
+	deploy.Spec.Template.Annotations[federationHashAnnotation] = hash
+
+	if err := r.Patch(ctx, &deploy, patch); err != nil {
+		return fmt.Errorf("patching deployment %s with federation hash: %w", deployName, err)
+	}
+
+	log.FromContext(ctx).Info("Patched federation config hash on deployment",
+		"deployment", deployName, "hash", hash[:12]+"...")
+	return nil
 }
 
 // --- Deploy individual module manifests ---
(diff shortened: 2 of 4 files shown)


dashboard-operator/internal/controller/dashboard_reconciler.go
```diff
@@ -429,14 +429,19 @@ func (r *DashboardReconciler) reconcileStandalone(
...
 		cm.MarkTrue(string(common.ConditionTypeDegraded),
 			conditions.WithReason("FederationConfigMapFailed"),
 			conditions.WithError(err))
 		logger.Error(err, "Failed to deploy federation ConfigMap")
 		return ctrl.Result{}, fmt.Errorf("federation ConfigMap: %w", err)
 	}
 
+	if err := r.patchDeploymentFederationHash(ctx, fedData); err != nil {
+		logger.Error(err, "Failed to patch federation hash on deployment")
+	}
```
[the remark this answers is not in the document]
lucferbux: Thanks for the suggestion. The operator currently has no `EventRecorder` infrastructure (the `DashboardReconciler` struct does not embed `record.EventRecorder`), so adding `r.Recorder.Event(...)` would require wiring a recorder through `SetupWithManager`.

The current log-and-continue pattern matches the existing convention for non-fatal errors in this reconciler — e.g. `deleteModuleResources` on line 403-405 follows the same pattern. The next reconcile cycle retries the patch automatically.

Fixed the related CWE-117 ast-grep finding: replaced the `hash[:12]+"..."` concatenation with the full hash as a structured logging field.

Adding event recorder infrastructure is a valid improvement for the operator as a whole (not just this path), so it is best addressed in a follow-up.

dashboard-operator/internal/controller/dashboard_reconciler.go
```diff
@@ -429,14 +429,19 @@ func (r *DashboardReconciler) reconcileStandalone(
...
+	if err != nil {
 		cm.MarkTrue(string(common.ConditionTypeDegraded),
 			conditions.WithReason("FederationConfigMapFailed"),
 			conditions.WithError(err))
 		logger.Error(err, "Failed to deploy federation ConfigMap")
 		return ctrl.Result{}, fmt.Errorf("federation ConfigMap: %w", err)
 	}
 
+	if err := r.patchDeploymentFederationHash(ctx, fedData); err != nil {
+		logger.Error(err, "Failed to patch federation hash on deployment")
```
christianvogt, line 442: **[AI review]** This fails silently. If the patch keeps failing (wrong deployment name, transient API error, etc.), the ConfigMap keeps updating but the rollout never happens — no condition, no requeue, and the CR status shows healthy the whole time. That's the same class of bug this PR is fixing, just moved one level down. Suggest setting a `Degraded` condition here like `FederationConfigMapFailed` above, or returning the error so it gets retried with backoff instead of only self-healing on the next periodic interval.

dashboard-operator/internal/controller/module_deploy.go
```diff
@@ -108,6 +109,58 @@ func standaloneServiceName(platform cluster.Platform, slug string) string {
...
+	ctx context.Context,
+	configData string,
+) error {
+	hash := computeFederationConfigHash(configData)
+
+	deployName := mainDashboardDeploymentName(r.Platform)
+	var deploy appsv1.Deployment
+	key := client.ObjectKey{Name: deployName, Namespace: r.ApplicationsNamespace}
+	if err := r.Get(ctx, key, &deploy); err != nil {
+		return fmt.Errorf("getting deployment %s: %w", deployName, err)
```
christianvogt, line 143: **[AI review]** Consider wrapping with `apierrors.IsNotFound` like every other `Get` in this package (e.g. `dashboard_reconciler.go:88`). Right now a genuine deployment-name mismatch and a transient API error look identical in the logs — combined with the silent-failure handling in `reconcileStandalone`, both get swallowed the same way.

dashboard-operator/internal/controller/module_deploy_test.go
```diff
@@ -0,0 +1,283 @@
...
+		Reason: "ComponentNotAvailable",
+	}
+
+	cm, err := ctrlpkg.BuildFederationConfigMap(r, statuses, &v1alpha1.Dashboard{})
+	require.NoError(t, err)
+
+	data := cm.Data["module-federation-config.json"]
+	assert.NotContains(t, data, `"genAi"`, "disabled module must be excluded")
+	assert.NotContains(t, data, `"maas"`, "not-deployed module must be excluded")
+	assert.Contains(t, data, `"modelRegistry"`, "deployed module must be included")
```
christianvogt: **[AI review]** Consider asserting on unmarshaled entries instead of raw JSON substrings — `TestBuildFederationConfigMap_NoEnabledField` a few lines down already does this. `Contains`/`NotContains` on the whole blob doesn't verify the match is actually a `name` field vs. some other nested string.

christianvogt (commented):
**⚠️ AI-generated review** — this review was produced by an AI agent (persona-based review: go/qa/security personas + Cursor Bugbot), not a human reviewer. Treat findings as input for a human to verify, not as an authoritative sign-off.

Summary: the approach is solid (SHA256 content-hash annotation to force a rolling restart when the federation ConfigMap changes), and the dead `Enabled` field removal is a good cleanup. One issue worth addressing before merge: the new hash-patch failure path is silent (logged only, no condition, no retry) which can defeat the PR's own purpose if it fails persistently. Two smaller suggestions inline, plus a test-coverage gap noted below.

**Missing integration coverage**: none of the 8 new tests exercise `patchDeploymentFederationHash` through `Reconcile()`/`reconcileStandalone` — they all call it directly. `TestReconcile`'s existing success case only passes today because its fixture never creates the `odh-dashboard` Deployment, so the patch silently fails there too, just unasserted. Consider adding a case that seeds the deployment and asserts the annotation gets patched end-to-end, plus one that explicitly asserts reconcile still succeeds when the patch fails.

dashboard-operator/internal/controller/dashboard_reconciler.go
lucferbux: Good catch — now returning the error so controller-runtime requeues with backoff. Skipped adding a Degraded condition because the condition manager doesn't auto-clear, which would leave a stale flag after transient failures resolve. The error return surfaces the issue and ensures prompt retry. Fixed in a4fd662.

dashboard-operator/internal/controller/module_deploy.go
lucferbux: Agreed — added `apierrors.IsNotFound` check. NotFound is now a no-op (logs and returns nil) since the deployment may not exist during bootstrap. Transient API errors propagate up and get retried via the error return added above. Fixed in a4fd662.

dashboard-operator/internal/controller/module_deploy_test.go
lucferbux: Good point — switched to JSON unmarshaling and checking the `name` field structurally, consistent with the `NoEnabledField` test below. Fixed in a4fd662.

lucferbux:
Thanks for the thorough review, Christian. All three inline comments addressed in a4fd662:

1. **Silent failure** — now returns the error for controller-runtime backoff retry (no Degraded condition since the condition manager doesn't auto-clear).
2. **IsNotFound** — added `apierrors.IsNotFound` guard; NotFound is a no-op during bootstrap, transient errors propagate for retry.
3. **Test assertions** — switched to structured JSON unmarshaling.

Re: integration coverage through `Reconcile()` — agreed that would strengthen confidence. The existing `TestReconcile` fixtures don't seed the dashboard Deployment, so the patch path is untested end-to-end. I'll track that as a follow-up alongside the EventRecorder infrastructure improvement.

PR #8800, opened 2026-07-24
Revert "Update MaaS API URL to use /v1/tenants"

Griffin-Sullivan:
Reverts opendatahub-io/odh-dashboard#8578

## Summary by CodeRabbit

* **New Features**
  * MAAS services can now automatically determine the cluster domain and connect to the MAAS API through the configured ingress.
  * Added configuration options for specifying the MAAS Gateway namespace and name.

* **Documentation**
  * Updated setup guidance and examples for MAAS API URL configuration, including automatic detection and current HTTP endpoint formats.

* **Bug Fixes**
  * Removed an unnecessary permission for accessing DataScienceCluster resources.


lucferbux:
Verified this is a clean, exact revert of #8578. All 17 files are perfectly mirrored — additions become deletions and vice versa. No extraneous changes.

The review comments from CodeRabbit and the Dashboard Agent are valid observations, but they flag pre-existing issues in the code being restored to (unused gateway flags, missing context timeout, blank domain handling, http/https doc mismatch). These should be addressed in a follow-up PR, not block this revert.

/lgtm
/approve

lucferbux (commented):
expanded results above

lucferbux:
## Phase 4: Operator Integration failure — pre-existing manifest bug, not caused by this revert

The "Phase 4: Operator Integration" CI job is failing with:

```
Error from server (Invalid): ClusterRoleBinding.rbac.authorization.k8s.io "odh-dashboard" is invalid: subjects[0].namespace: Required value
Error from server (Invalid): ClusterRoleBinding.rbac.authorization.k8s.io "odh-dashboard-auth-delegator" is invalid: subjects[0].namespace: Required value
Error from server (Invalid): ClusterRoleBinding.rbac.authorization.k8s.io "odh-dashboard-modules" is invalid: subjects[0].namespace: Required value
Error from server (Invalid): ClusterRoleBinding.rbac.authorization.k8s.io "odh-dashboard-monitoring" is invalid: subjects[0].namespace: Required value
```

### Root cause

Four ClusterRoleBinding manifests reference ServiceAccounts **without specifying a `namespace`**. Kubernetes requires `namespace` on ServiceAccount subjects in ClusterRoleBindings because ServiceAccounts are namespace-scoped. The CI runs `kubectl apply -k manifests/odh -n opendatahub`, but `-n` only sets the namespace for namespace-scoped resources — it does NOT populate `subjects[].namespace` inside ClusterRoleBindings.

Running `kustomize build manifests/odh` confirms the issue — all 4 CRBs produce subjects like:

```yaml
subjects:
- kind: ServiceAccount
  name: odh-dashboard
  # namespace is missing!
```

### Affected files (all on `main`, none touched by this PR)

| File | CRB name | Missing namespace |
|---|---|---|
| `manifests/core-bases/base/sa-rbac/cluster-role-binding.yaml` | `odh-dashboard` | `subjects[0].namespace` |
| `manifests/core-bases/base/sa-rbac/auth-delegator.clusterrolebinding.yaml` | `odh-dashboard-auth-delegator` | `subjects[0].namespace` |
| `manifests/core-bases/base/sa-rbac/cluster-monitoring-role-binding.yaml` | `odh-dashboard-monitoring` | `subjects[0].namespace` |
| `manifests/modular-architecture/modules-cluster-role-binding.yaml` | `odh-dashboard-modules` | `subjects[0].namespace` |

### Verified: same failure on other PRs

This is NOT caused by this revert. The exact same error occurs on completely unrelated PRs:
- **#8799** ([RHOAIENG-56820](https://redhat.atlassian.net/browse/RHOAIENG-56820) route name length) — run `30098076130`, job `89499696050`: same 4 CRB errors
- **#8798** ([RHOAIENG-78072](https://redhat.atlassian.net/browse/RHOAIENG-78072) Kueue microcopy) — run `30097416044`: same Phase 4 failure

All recent "PR Build Validation" runs are hitting this.

### Suggested fix (separate PR)

Add `namespace: opendatahub` to each ServiceAccount subject. Example for `cluster-role-binding.yaml`:

```yaml
subjects:
  - kind: ServiceAccount
    name: odh-dashboard
    namespace: opendatahub   # <-- add this
```

Since the namespace is deployment-specific (`opendatahub` for ODH, `redhat-ods-applications` for RHOAI), the cleanest fix would be to use kustomize `namespace` transformer or a `replacements` block to inject it — similar to how `vars` already handles image references. But for ODH manifests specifically, hardcoding `opendatahub` would also work since that overlay is ODH-only.

### Recommendation for this PR

This revert is clean and correct — no changes needed here. The Phase 4 failure should be fixed in a separate follow-up PR targeting the 4 ClusterRoleBinding files on `main`.

[RHOAIENG-56820]: https://redhat.atlassian.net/browse/RHOAIENG-56820?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

lucferbux:
@Griffin-Sullivan that was AI generated, I was under the impression the konflux build simulator was gating PRs but that doesn't seem to be the case, let's see if this merges

packages/maas/bff/internal/api/app.go
```diff
@@ -139,12 +139,17 @@ func NewApp(cfg config.EnvConfig, logger *slog.Logger) (*App, error) {
...
-			return nil, fmt.Errorf("failed to discover MaaS API URL via /v1/tenants: %w", err)
+	} else {
+		// Fallback to discovery of MaaS API url, when not provided via envvar, or cmd flags
+		if cfg.MaasApiUrl == "" {
+			clusterDomain, err := helper.GetClusterDomainUsingServiceAccount(context.Background(), logger)
+			if err != nil {
+				logger.Error("Failed to auto-discover cluster domain, MaaS API URL will be unavailable", "error", err)
+			} else {
+				cfg.MaasApiUrl = fmt.Sprintf("https://maas.%s/maas-api", clusterDomain)
+				logger.Info("Using automatically discovered MaaS URL", "url", cfg.MaasApiUrl)
```
liday-rh, line 150: When `GetClusterDomainUsingServiceAccount` fails (non-OpenShift cluster, RBAC delay, network issue), `cfg.MaasApiUrl` stays `""`. The BFF starts, healthcheck returns 200, but every `/api/v1/models` and `/api/v1/api-keys` request fails at runtime with `invalid URI for request`.

The previous code returned a startup error in this case (`return nil, fmt.Errorf("failed to discover MaaS API URL...")`), which was a fail-fast that surfaced the problem immediately. This silently downgrades to a broken-but-running state.

Consider either restoring the fail-fast, or adding a handler-level guard that returns 503 with a clear message when `MaasApiUrl` is empty.

packages/maas/bff/internal/config/environment.go
```diff
@@ -119,8 +119,8 @@ type EnvConfig struct {
 	FederatedPlatform bool
 
 	// MaaS
+	GatewayNamespace          string
+	GatewayName               string
```
liday-rh, line 123: `GatewayNamespace` and `GatewayName` are wired to CLI flags and env vars (`GATEWAY_NAMESPACE`, `GATEWAY_NAME`) in `main.go`, but no production code reads them. `GetClusterDomainUsingServiceAccount` always queries `config.openshift.io/ingresses/cluster` regardless of what these are set to.

An operator who sets `GATEWAY_NAMESPACE=my-ns` or `--gateway-name=custom-gateway` gets no effect. Either wire them into the discovery logic or remove the flags to avoid misleading operators.

packages/maas/bff/internal/api/app.go
```diff
@@ -139,12 +139,17 @@ func NewApp(cfg config.EnvConfig, logger *slog.Logger) (*App, error) {
...
 		logger.Info("MaaS Fake API Server is running", "url", maasFakeServer.URL)
 		cfg.MaasApiUrl = maasFakeServer.URL
-	} else if cfg.MaasApiUrl == "" {
-		discoveredURL, err := helper.DiscoverMaasApiURL(context.Background(), cfg, logger, rootCAs)
-		if err != nil {
-			return nil, fmt.Errorf("failed to discover MaaS API URL via /v1/tenants: %w", err)
+	} else {
+		// Fallback to discovery of MaaS API url, when not provided via envvar, or cmd flags
+		if cfg.MaasApiUrl == "" {
+			clusterDomain, err := helper.GetClusterDomainUsingServiceAccount(context.Background(), logger)
```
liday-rh, line 145: nit: `context.Background()` with no deadline — a slow or unreachable K8s API server blocks startup indefinitely. The old discovery code used `Timeout: 10 * time.Second` on its HTTP client. Consider:

```go
ctx, cancel := context.WithTimeout(context.Background(), 15*time.Second)
defer cancel()
clusterDomain, err := helper.GetClusterDomainUsingServiceAccount(ctx, logger)
```

manifests/modular-architecture/modules/maas/deployment.yaml
```diff
@@ -146,6 +146,10 @@ spec:
           env:
             - name: GATEWAY_DOMAIN
               value: ""
+            - name: TIERS_CONFIGMAP_NS
```
liday-rh, line 149: nit: `TIERS_CONFIGMAP_NS` is injected via downward API, but neither `EnvConfig` nor `main.go` references it — no code reads this variable. Is this scaffolding for a follow-up? If so, worth a comment; otherwise it's confusing dead config.

packages/maas/README.md
```diff
@@ -83,21 +83,9 @@ The following environment variables are used to configure the deployment and dev
...
-- **Example**: `MAAS_API_INTERNAL_URL=https://maas-api.odh-ai-gateway-infra.svc.cluster.local:8443`
-
-### `MAAS_API_NAMESPACE`
-
-- **Description**: Namespace of the internal `maas-api` Service when `MAAS_API_INTERNAL_URL` is not set. Overrides DSC-based auto-detection.
-- **Default Value**: *(none)* — auto-resolved from DSC `status.release.name`, then `redhat-ai-gateway-infra` as fallback.
-- **Example**: `MAAS_API_NAMESPACE=redhat-ai-gateway-infra`
+- **Description**: Specifies the URL of the MaaS API that the BFF connects to.
+- **Default Value**: *(none)* — if not set, the `dev-bff-federated` Makefile target will auto-detect it by querying the cluster's ingress domain via `oc get ingresses.config.openshift.io cluster`. The resolved URL follows the pattern `http://maas.<CLUSTER_DOMAIN>/maas-api`. When running in-cluster, the BFF performs the same auto-detection using its service account.
+- **Example**: `MAAS_API_URL=http://maas.apps.my-cluster.example.com/maas-api`
```
liday-rh, line 88: nit: docs say `http://maas.<CLUSTER_DOMAIN>/maas-api` but `app.go:149` constructs `https://maas.<domain>/maas-api`. OpenShift Routes are typically TLS-terminated, so the code is likely correct and the docs should use `https://`. Same issue in `CONTRIBUTING.md`.

liday-rh (changes requested):
This is not a clean revert — it introduces new replacement code with a few issues worth addressing. Two main concerns:

1. **Silent degraded mode**: discovery failure leaves `MaasApiUrl` empty; BFF starts healthy but every MaaS API call fails at runtime. The reverted code was fail-fast.
2. **Dead config flags**: `GatewayNamespace`/`GatewayName` are wired to CLI/env but never read by any production code.

See inline comments for details and smaller items (missing timeout, unused `TIERS_CONFIGMAP_NS` env var, http/https doc mismatch).

CI Phase 4 failure is a pre-existing `main` issue (ClusterRoleBinding missing `namespace` in subjects) — not caused by this PR.

packages/maas/bff/internal/api/app.go
Griffin-Sullivan: We're going to fix this after reverting. The old code you're describing here is what would happen when we did the `GetClusterDomainUsingServiceAccount`. We assumed the user was using the same URL for cluster ingress and maas-api-gateway (not true) hence the changes in the last PR.

packages/maas/bff/internal/config/environment.go
Griffin-Sullivan: It's ok I'll remove these again in the next PR

packages/maas/bff/internal/api/app.go
Griffin-Sullivan: Ok I'll address in the next PR

manifests/modular-architecture/modules/maas/deployment.yaml
Griffin-Sullivan: It'll be removed in the next PR

packages/maas/bff/internal/api/app.go
liday-rh: Ok cool, as long as its tracked.

christianvogt (approved):

andrewballantyne:
/hold

Unblocking https://github.com/opendatahub-io/odh-dashboard/pull/8806

Something in the queue is causing problems -- lets see if we can get 8806 merged to help with our on-PR CI.

andrewballantyne:
/unhold
/override ci/prow/odh-dashboard-pr-image-mirror

One down, now lets see if this merges

andrewballantyne:
## Core Dashboard image jobs
/override ci/prow/images

## Operator image
/override ci/prow/odh-dashboard-operator-pr-image-mirror

## BFF Image jobs
/override ci/prow/odh-mod-arch-maas-pr-image-mirror

andrewballantyne:
I anticipate problems here, need this to merge, cancelling image builds

PR #8810, opened 2026-07-24
feat(RHOAIENG-78901): Show fast and unsupported labels in deployed models table

mturley:
https://issues.redhat.com/browse/RHOAIENG-78901

Screenshot 2026-07-24 at 5 52 16 PM

Screenshot 2026-07-24 at 5 52 29 PM

Screenshot 2026-07-24 at 6 17 57 PM

cc @vconzola

## Description

Replaces bad autofix PR #8792.

Converts `DeploymentResourceVersionLabels` from a component to a `renderDeploymentResourceVersionLabels` function that returns an array of keyed Label elements. Callers render the result as a JSX expression inside their own `LabelGroup`, where React flattens the keyed array as direct children.

Previously, `ServingRuntimeVersionLabel` rendered only the blue version label, and `DeploymentResourceVersionLabels` wrapped it alongside fast and unsupported labels but was only used in admin settings tables. The KServe and llmd-serving `ServingDetailsComponent` extensions used `ServingRuntimeVersionLabel` directly, so fast/unsupported labels never appeared in the deployed models table.

This PR:
- Converts `DeploymentResourceVersionLabels` component to `renderDeploymentResourceVersionLabels` function returning keyed Label elements that callers render inside their `LabelGroup`
- Adds a `getVersion` option (defaults to `getServingRuntimeVersion`) for extensibility
- Replaces `ServingRuntimeVersionLabel` usage in KServe and llmd-serving with the new function, so fast-version and limited-support labels now appear in the deployed models table
- Deprecates `ServingRuntimeVersionLabel` (retained for the legacy `ServingRuntimeModal` code path)
- Sets `numLabels={5}` on all LabelGroups containing version labels to prevent PF's default 3-label overflow
- Leaves `ServingRuntimeTemplateSection` unchanged (deprecated code path)

**Why a render function instead of a component?** PatternFly's `LabelGroup` only applies correct spacing to direct children — Labels nested inside a fragment or wrapper component lose their spacing. By returning an array of elements, callers can render them inside their `LabelGroup` alongside other sibling labels (version status, template status, scope). If PF fixes this `LabelGroup` behavior, the function could become a component rendering a fragment.

## How Has This Been Tested?

- All `renderDeploymentResourceVersionLabels` unit tests pass (9 tests including `getVersion` option test and key uniqueness test)
- All `InferenceServiceServingRuntime` unit tests pass (5 tests)
- TypeScript type-check passes for frontend, model-serving, and llmd-serving packages

## Test Impact

- Added unit tests for the `getVersion` option and key uniqueness in `renderDeploymentResourceVersionLabels.spec.tsx`
- Existing tests continue to pass unchanged

## Request review criteria:

If you have UI changes:


frontend/src/pages/modelServing/screens/projects/ServingRuntimeModal/ServingRuntimeTemplateSection.tsx
```diff
@@ -106,11 +105,9 @@ const ServingRuntimeTemplateSection: React.FC<ServingRuntimeTemplateSectionProps
...
         <Truncate content={getServingRuntimeDisplayNameFromTemplate(template)} />
       </FlexItem>
-      {getServingRuntimeVersion(template) && (
-        <FlexItem>
-          <ServingRuntimeVersionLabel version={getServingRuntimeVersion(template)} isCompact />
-        </FlexItem>
-      )}
+      <FlexItem>
+        <DeploymentResourceVersionLabels resource={template} isCompact />
+      </FlexItem>
```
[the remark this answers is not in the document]
mturley: This is a false positive. The utilities used by `DeploymentResourceVersionLabels` (`getServingRuntimeVersion`, `getFastVersion`, `isUnsupportedResource`) are aware of Templates and when passed a template they look at the inner resource for annotations first. This is by design, we can flexibly pass either a resource or a template and expect it to work correctly.

frontend/src/pages/modelServing/screens/projects/ServingRuntimeModal/ServingRuntimeTemplateSection.tsx
emilys314, line 1: We should remove the changes in here, this is only used in the KserveDeployModal, which is deprecated.

packages/model-serving/src/shared/components/DeploymentResourceVersionLabels.tsx
```diff
@@ -12,23 +11,42 @@ type DeploymentResourceVersionLabelsProps<T extends K8sResourceCommon> = {
...
   isEditing,
+  getVersion = getServingRuntimeVersion,
+  versionStatus,
+  templateStatus,
+  scopeLabel,
 }: DeploymentResourceVersionLabelsProps<T>): React.ReactElement | null => {
-  const version = getServingRuntimeVersion(resource);
+  const version = getVersion(resource);
   const fastVersion = getFastVersion(resource);
   const unsupported = isUnsupportedResource(resource);
```
emilys314: I'm confused with the contract of this component. 
The `resource: T` is passing in the resource to get the version labels from, but we are also passing in directly a list of labels that we are wanting to render. This makes it a mix of "pass-through labels from props" and "pass data in to calculate yourself"

So this creates confusion as what the source of truth is. 
- `version` derived from the `resource` using `getVersion`.
- `versionStatus`, passed in as a prop

packages/model-serving/src/shared/components/DeploymentResourceVersionLabels.tsx
```diff
@@ -12,23 +11,42 @@ type DeploymentResourceVersionLabelsProps<T extends K8sResourceCommon> = {
   resource: T;
   isCompact?: boolean;
   isEditing?: boolean;
+  getVersion?: (resource: T) => string | undefined;
+  // PF LabelGroup only spaces its children correctly when they are direct children,
+  // not wrapped in fragments. These slots let callers render additional labels inside
+  // this component's LabelGroup instead of nesting a second LabelGroup around it.
+  versionStatus?: React.ReactNode;
```
emilys314: `React.ReactNode` is a very weak contract, anything can be passed into here and makes it very hard to audit the behavior of this component and visuals.

For this use case, we have a fairly standardized look and feel for these resource version labels, we shouldn't let the parents potentially mess it up. 

So it would be better to pass in a typed object with just the data

emilys314 (commented):
This component is confusing haha but it works

Image

packages/model-serving/src/shared/components/DeploymentResourceVersionLabels.tsx
mturley: So, it maybe does make things a little too funky, but initially this component was just passing in the resource and internally grabbing values, and then I added `versionStatus`, `templateStatus` and `scopeLabel` to solve the LabelGroup issue. The intention is that this component is agnostic of platform and shouldn’t assume how to get the data for those things, and not all platforms provide anything for them. Turning it into an “everything component for labels for all platforms” feels bad, and also felt like a heavier refactor than I wanted. What I would really love is to just have this render a fragment of labels (only the 3 we care about for version/fast/unsupported) and be able to stick it in a LabelGroup with others, but that breaks the styles. Maybe there’s a better fix for that or we use some kind of hack like margin and open a tech debt issue and a PF bug. But I do think extensions should be able to add things to this LabelGroup, whether that’s by some mechanism like this or by fixing the problem with rendering a fragment here.

I also tried a workaround of rendering a LabelGroup inside a LabelGroup, which works correctly when laying them out horizontally, but when they wrap in a vertical stack like on the overview cards it mixed vertical and horizontal groups.

packages/model-serving/src/shared/components/DeploymentResourceVersionLabels.tsx
mturley: I didn’t want to require extensions to render these in any particular way, the intention was “put other children into the LabelGroup” but because of the fragment child bug I can’t just make that a “more children” prop. Actually however, maybe what I could do is take an array of nodes and spread them in as additional children, that might work around the fragment problem. But yeah, the weak contract is intentional here and maybe is made weird by the specific use-case-based names of each prop.

The core issue here is that `LabelGroup` needs its `Labels` to be direct children in the component tree. spreading an array might work. Or maybe even just going back to not rendering a LabelGroup inside this component and turning it into a `renderResourceVersionLabels` function that returns an array of label nodes that each extension spreads into its own LabelGroup, but that feels weird too.

emilys314: What I'm trying to say is to have in the props something like 
```
props: {
  labels: { id: string, text: string, color: string, popver? }[]
}

render {
  {props.labels.map( label => <Label string=label.string /> }
}
```
or
```
props={
  versionLabel: string
  fastLabel?: string
  unsupportedLabel?:  boolean
}

render {
  <Label string=versionLabel />
  {fastLabel ?? <Label string=fastLabel />
  {unsupportedLabel ?? <Label string="Unsupported />
}
```

frontend/src/pages/modelServing/screens/projects/ServingRuntimeModal/ServingRuntimeTemplateSection.tsx
mturley: I removed the `ServingRuntimeVersionLabel` component entirely because the new component makes it redundant, so leaving this file alone would be a type error. Although I could put that component back and mark it deprecated.

mturley: I’ll do that

mturley: Done

packages/model-serving/src/shared/components/DeploymentResourceVersionLabels.tsx
mturley: Pushed and rewrote PR description. It’s now a function that returns an array of labels you can spread into a LabelGroup.

packages/model-serving/src/shared/components/DeploymentResourceVersionLabels.tsx
mturley: Oh I see, I could have solved the LabelGroup problem without converting the whole thing to a function by using the `props.labels.map` pattern. I do think it still makes more sense to not have this provide the LabelGroup since the intention is to use it as siblings of other labels though.

frontend/src/pages/modelServing/customServingRuntimes/CustomServingRuntimeTableRow.tsx
```diff
@@ -48,7 +48,7 @@ const CustomServingRuntimeTableRow: React.FC<CustomServingRuntimeTableRowProps>
         </ResourceNameTooltip>
         <LabelGroup>
           {templateOOTB && <Label data-testid="pre-installed-label">{PreInstalledName}</Label>}
-          <DeploymentResourceVersionLabels resource={template} />
+          {...renderDeploymentResourceVersionLabels(template)}
```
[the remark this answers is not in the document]
mturley: Hmmmmm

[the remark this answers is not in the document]
mturley: It definitely compiles lol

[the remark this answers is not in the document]
mturley: The suggested syntax may be cleaner though, I’ll try it

packages/llmd-serving/src/settings/llmAcceleratorConfigs/LlmAcceleratorConfigTableRow.tsx
```diff
@@ -59,7 +59,7 @@ const LlmAcceleratorConfigTableRow: React.FC<LlmAcceleratorConfigTableRowProps>
         </ResourceNameTooltip>
         <LabelGroup>
           {preInstalled && <Label data-testid="pre-installed-label">{PreInstalledName}</Label>}
-          <DeploymentResourceVersionLabels resource={config} />
+          {...renderDeploymentResourceVersionLabels(config)}
         </LabelGroup>
```
emilys314: I'm seeing it start to get cut-off now in some places. Maybe it now works a lot better with the labels so counts it better

Image

mturley: When I had the LabelGroup inside the component it had a numLabels={5}, I moved that to the LabelGroups outside each caller now, maybe I missed some. Looking

mturley: Yep. Will fix along with removing the spread syntax, coderabbit was right that it’s unnecessary and claude calls it an “obscure syntax”. ok fine guys jeez

frontend/src/pages/modelServing/customServingRuntimes/CustomServingRuntimeTableRow.tsx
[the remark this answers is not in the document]
mturley: Addressed, the spread was harmless but unnecessary.

packages/llmd-serving/src/settings/llmAcceleratorConfigs/LlmAcceleratorConfigTableRow.tsx
mturley: Fixed

emilys314 (approved):
Looks good!

Image

Image

Image

Image

Image

PR #8870, opened 2026-07-27
fix(e2e): prevent MLflow test failures on re-runs from stale name collisions

nananosirova:
https://issues.redhat.com/browse/RHOAIENG-79267

  ## Description

  MLflow E2E tests (`testMlflowExperiments` and `testPromptManagement`) fail on re-runs because resource names from prior runs are still reserved.

  **Root cause:** MLflow soft-deletes experiments, so the name stays permanently reserved even after the delete API call. The `doesMlflowExperimentNameExist` helper was using the `experiments/search` API without an explicit `max_results` parameter, which defaults to `0`
  on this server and returns an `INVALID_PARAMETER_VALUE` error. The `catch` block silently treated this error as "name available," so collision detection never worked.

  **Changes:**

  - **`mlflow.ts`**: Add `max_results: 10` to the experiment search request so collision detection works correctly. Add `execGetInMlflowPod` for GET-based API calls (the registered-models endpoint requires GET, not POST). Add `doesMlflowExperimentNameExist`,
  `findAvailableExperimentSuffix`, `deleteStalePromptByName`, `doesMlflowPromptNameExist`, and `findAvailablePromptSuffix` utilities. Add `method` parameter to `execCurlInMlflowPod` (prompts require DELETE, not POST).
  - **`testMlflowExperiments.cy.ts`**: Use `findAvailableExperimentSuffix` in `retryableBefore` to detect stale name collisions and auto-increment the suffix. All three experiment names (create, rename, compare) are checked.
  - **`testPromptManagement.cy.ts`**: Use `deleteStalePromptByName` to clean up prior runs, then `findAvailablePromptSuffix` for collision detection. Prompts support permanent delete so names can be freed, but suffix detection is still used as a safety net.

  ## How Has This Been Tested?

  Ran both E2E test suites twice each on the `dash-e2e-odh.osp.rh-ods.com` cluster to verify re-run stability:

  - `testPromptManagement.cy.ts`: run 1 pass, run 2 pass
  - `testMlflowExperiments.cy.ts`: run 1 pass, run 2 pass

  Also verified via direct `curl` commands inside the MLflow pod that `max_results: 10` correctly returns soft-deleted experiments with `view_type: ALL`, and that the `permanently-delete` endpoint is not available on this server (confirming the suffix-increment approach
  is necessary).

  ## Test Impact

  This PR modifies E2E test infrastructure only, no product code changes. The changes fix existing tests to be re-runnable; no new test scenarios were added.

  ## Request review criteria:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved automated experiment and prompt management workflows by preventing naming conflicts with existing or stale entries.
  * Added cleanup support for stale prompts before test execution.

* **Tests**
  * Updated end-to-end validation to use collision-free names consistently across creation, renaming, assertions, and cleanup.
  * Added checks for existing experiments and prompts, including soft-deleted entries.

### DIFF
--- a/packages/cypress/cypress/tests/e2e/mlflowExperiments/testMlflowExperiments.cy.ts
+++ b/packages/cypress/cypress/tests/e2e/mlflowExperiments/testMlflowExperiments.cy.ts
@@ -4,6 +4,7 @@ import {
   disableMlflowFeatures,
   createMlflowExperimentViaAPI,
   deleteMlflowExperimentViaAPI,
+  findAvailableExperimentSuffix,
   getMlflowExperimentIdByName,
   logMlflowRunsViaAPI,
 } from '../../../utils/oc_commands/mlflow';
@@ -18,6 +19,7 @@ import type { MlflowExperimentsTestData } from '../../../types';
 describe('Verify MLflow Experiments page', () => {
   let testData: MlflowExperimentsTestData;
   let projectName: string;
+  let testSuffix: string;
   let runsExperimentId: string | undefined;
   let uiExperimentName: string | undefined;
   let uiExperimentDeleted = false;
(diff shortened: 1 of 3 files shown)


DaoDaoNoCode:
@nananosirova Can you rebase the latest main branch? That should fix the e2e failure in the CI

DaoDaoNoCode (approved):
/lgtm

PR #8873, opened 2026-07-27
ignore all dependabot patternfly updates

christianvogt:
Updates dependabot config to ignore patternfly altogether until we manually update patternfly and fix the sharing of all patternfly packages with module federation.

Bumping patternfly can inadvertently bringing in multiple copies of the patternfly css and cause visual issues as we've recently seen in the past.

The dependabot auto merge workflow was disabled until we address patternfly. It was expected to be a short timeframe but now we are piling up on dependency updates and it's best to just ignore the problematic patternfly updates.

## Summary by CodeRabbit

* **Chores**
  * Updated automated dependency update settings for PatternFly packages.
  * Removed the dedicated PatternFly dependency grouping configuration.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -39,16 +39,13 @@ updates:
       semver-patch-days: 14
 
     ignore:
+      - dependency-name: "@patternfly/*"
       - dependency-name: "react"
         update-types: ["version-update:semver-major"]
       - dependency-name: "react-dom"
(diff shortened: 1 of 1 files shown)


jenny-s51 (approved):
/lgtm @christianvogt - this makes sense to configure Dependabot to stop opening PRs for any PatternFly package so all PF updates can be done manually as a coordinated effort to prevent CSS duplication and visual regressions.

PR #8878, opened 2026-07-27
fixSelect default deployment method for generative models in wizard

YuliaKrimerman:
https://issues.redhat.com/browse/RHOAIENG-79257

## Description

When deploying a generative model via the wizard, no deployment method was pre-selected. The user had to manually choose between "LLM inference service" and "llm-d".

The root cause was that `vllmDeploymentMethodOverride` in `packages/llmd-serving/src/wizardFields/deploymentMethodField.ts` had no `suggestion` function. When `isLLMdDefault` was false, no suggestion was produced and `getInitialFieldData` defaulted to an empty method string.

Added a `suggestion` function to `vllmDeploymentMethodOverride` that returns `SIMPLE_VLLM_OPTION` when `isLLMdDefault` is not true. This mirrors the existing pattern in `llmdDeploymentMethodOverride` and ensures:
- "LLM inference service" is selected by default when llm-d default is OFF
- "llm-d" is selected by default when the admin toggles llm-d default ON
- There is always a default selection for generative models

## How Has This Been Tested? 
Default legacy option 

https://github.com/user-attachments/assets/00ae41ca-0411-4468-bfa3-890d459b03f4

Inference without llmd deployment 

https://github.com/user-attachments/assets/4e2884b3-9120-4c02-b215-3760d2f435c9

Llmd default 

https://github.com/user-attachments/assets/c6ae00ba-11ad-415a-9a3f-ef93a6d50436

Ran lint, type-check, and full unit test suite for the llmd-serving package. All 25 test suites (258 tests) pass.

## Test Impact

Added `deploymentMethodField.spec.ts` with 15 tests covering both `vllmDeploymentMethodOverride` and `llmdDeploymentMethodOverride` suggestion logic across all cluster settings scenarios (isLLMdDefault true/false/undefined, null/undefined clusterSettings).

## Request review criteria:

## Summary by CodeRabbit

* **New Features**
  * Improved deployment-method wizard suggestions with configuration-aware behavior across vLLM, llm-d, and legacy options.
  * When multiple suggestions are available, the wizard now consistently selects the one with the highest priority (lowest order).
* **Tests**
  * Added/expanded Jest coverage for both vLLM and llm-d/legacy wizard overrides, including activation, option metadata, and suggestion rules.
  * Updated deployment wizard tests to validate cross-override suggestion selection.

### DIFF
--- a/packages/kserve/src/wizardFields/deploymentMethodField.ts
+++ b/packages/kserve/src/wizardFields/deploymentMethodField.ts
@@ -2,17 +2,18 @@ import type { DeploymentMethodFieldOverride } from '@odh-dashboard/model-serving
 
 export const LEGACY_GENERATIVE_DEPLOYMENT_METHOD_KEY = 'legacy';
 
+const LEGACY_OPTION = {
+  key: LEGACY_GENERATIVE_DEPLOYMENT_METHOD_KEY,
+  label: 'Legacy deployment',
+  description:
+    'Deploy using a serving runtime template. Use this option for non-LLM models or for compatibility with previous deployments. This method does not support Models as a Service.',
+  order: 3,
+};
+
 export const legacyDeploymentMethodOverride: DeploymentMethodFieldOverride = {
   id: 'deploymentMethod',
   type: 'modifier',
   isActive: () => true,
-  options: [
-    {
-      key: LEGACY_GENERATIVE_DEPLOYMENT_METHOD_KEY,
-      label: 'Legacy deployment',
-      description:
-        'Deploy using a serving runtime template. Use this option for non-LLM models or for compatibility with previous deployments. This method does not support Models as a Service.',
-      order: 3,
-    },
-  ],
+  options: [LEGACY_OPTION],
+  suggestion: (clusterSettings) => (!clusterSettings?.isLLMdDefault ? LEGACY_OPTION : undefined),
 };
--- a/packages/llmd-serving/src/wizardFields/deploymentMethodField.ts
+++ b/packages/llmd-serving/src/wizardFields/deploymentMethodField.ts
@@ -19,6 +19,8 @@ export const vllmDeploymentMethodOverride: DeploymentMethodFieldOverride = {
   type: 'modifier',
   isActive: () => true,
   options: [SIMPLE_VLLM_OPTION],
+  suggestion: (clusterSettings) =>
+    !clusterSettings?.isLLMdDefault ? SIMPLE_VLLM_OPTION : undefined,
 };
 
 // LLM-d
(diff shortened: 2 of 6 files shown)


packages/llmd-serving/src/wizardFields/deploymentMethodField.ts
```diff
@@ -19,6 +19,8 @@ export const vllmDeploymentMethodOverride: DeploymentMethodFieldOverride = {
   type: 'modifier',
   isActive: () => true,
   options: [SIMPLE_VLLM_OPTION],
+  suggestion: (clusterSettings) =>
+    !clusterSettings?.isLLMdDefault ? SIMPLE_VLLM_OPTION : undefined,
```
emilys314, line 23: If you turn off fast vLLM, there is no way to autoselect the legacy. We need a change in `kserve` or else where

Image

emilys314 (commented):
Works for the simple LLMInferenceService 

Image

packages/llmd-serving/src/wizardFields/deploymentMethodField.ts
YuliaKrimerman: https://github.com/user-attachments/assets/f4e35961-2fe3-4d4f-9a4e-a7da360c2451

Fixed

packages/kserve/src/wizardFields/deploymentMethodField.ts
```diff
@@ -2,17 +2,18 @@ import type { DeploymentMethodFieldOverride } from '@odh-dashboard/model-serving
...
-    {
-      key: LEGACY_GENERATIVE_DEPLOYMENT_METHOD_KEY,
-      label: 'Legacy deployment',
-      description:
-        'Deploy using a serving runtime template. Use this option for non-LLM models or for compatibility with previous deployments. This method does not support Models as a Service.',
-      order: 3,
-    },
-  ],
+  options: [LEGACY_OPTION],
+  suggestion: (clusterSettings) => (!clusterSettings?.isLLMdDefault ? LEGACY_OPTION : undefined),
```
emilys314, line 18: This works when the vLLMOnMaaS is disabled, but when it's enabled, we need to prioritize the simple vLLM option. I think this is overwriting the suggestion from the "LLM Inference Service Option"

Image

So if we have the 3 options:
if (llmd is default) -> select `LLM inference service with llm-d`
else -> select `LLM inference service` - (currently it's selecting `Legacy deployment `instead)


If only 2 options:
if (llmd is default) -> select `LLM inference service with llm-d`
else -> select `Legacy deployment`

YuliaKrimerman: Ok I hope I got it this time - so many options

emilys314 (approved):
Looks good! All the other scenarios are working

Image

PR #8890, opened 2026-07-27
Remove 'platform.opendatahub.io/part-of' label when creating admin llmiscfg

emilys314:
Closes https://redhat.atlassian.net/browse/RHOAIENG-79310

## Description

All other shared components use a `platform.opendatahub.io/part-of` label for determing if it's an ootb resource. This doesn't fix it for llmiscfg's, but it makes sure when duplicating ootb ones this is removed for all versions for when we do fix it

Comparison
image

## How Has This Been Tested?

Duplicate an ootb LLMisCfg accelerator profile, save it, edit it, make sure the label is gone

## Test Impact

The deletes are very simple so i'd need a cypress test but there is a time issue

## Request review criteria:

If you have UI changes: 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resource duplication across routing, topology, and accelerator configurations by removing system-generated metadata from duplicated resources.
  * Updated the generated duplicated YAML to include sanitized labels in addition to cleaned annotations, preventing metadata carryover.
  * Ensured duplicated resources no longer retain metadata that could cause conflicts or unintended associations.

### DIFF
--- a/packages/llmd-serving/src/settings/RoutingConfigurationCreateEdit.tsx
+++ b/packages/llmd-serving/src/settings/RoutingConfigurationCreateEdit.tsx
@@ -210,14 +210,10 @@ const RoutingConfigurationCreateEditInner: React.FC<{
     if (existingConfig) {
       return YAML.stringify(existingConfig);
     }
-    if (state?.sourceConfig) {
+    if (isDuplicateMode) {
       const cleanMeta = cleanResourceForYAMLViewer(state.sourceConfig.metadata);
-      let cleanAnnotations = cleanMeta.annotations;
-      let cleanLabels = cleanMeta.labels;
-      if (isDuplicateMode) {
-        cleanAnnotations = stripDuplicatingAnnotations(cleanMeta.annotations);
-        cleanLabels = stripDuplicatingLabels(cleanMeta.labels);
-      }
+      const cleanAnnotations = stripDuplicatingAnnotations(cleanMeta.annotations);
+      const cleanLabels = stripDuplicatingLabels(cleanMeta.labels);
       const duplicateDisplayName = `Copy of ${getDisplayNameFromK8sResource(state.sourceConfig)}`;
       return YAML.stringify({
         apiVersion: state.sourceConfig.apiVersion,
--- a/packages/llmd-serving/src/settings/RoutingConfigurationCreateEdit.tsx
+++ b/packages/llmd-serving/src/settings/RoutingConfigurationCreateEdit.tsx
@@ -37,7 +37,12 @@ import {
   CONFIG_TYPE_LABEL,
   SUPPORTED_TOPOLOGIES_ANNOTATION,
 } from '../types';
-import { isConfigObject, cleanResourceForYAMLViewer, stripAnnotation } from '../utils';
+import {
+  isConfigObject,
+  cleanResourceForYAMLViewer,
+  stripDuplicatingAnnotations,
+  stripDuplicatingLabels,
+} from '../utils';
 import {
   createLLMInferenceServiceConfig,
   patchLLMInferenceServiceConfig,
@@ -207,10 +212,12 @@ const RoutingConfigurationCreateEditInner: React.FC<{
     }
     if (state?.sourceConfig) {
       const cleanMeta = cleanResourceForYAMLViewer(state.sourceConfig.metadata);
-      const cleanAnnotations = stripAnnotation(
-        cleanMeta.annotations,
-        'kubectl.kubernetes.io/last-applied-configuration',
-      );
+      let cleanAnnotations = cleanMeta.annotations;
+      let cleanLabels = cleanMeta.annotations;
+      if (isDuplicateMode) {
+        cleanAnnotations = stripDuplicatingAnnotations(cleanMeta.annotations);
+        cleanLabels = stripDuplicatingLabels(cleanMeta.labels);
+      }
       const duplicateDisplayName = `Copy of ${getDisplayNameFromK8sResource(state.sourceConfig)}`;
       return YAML.stringify({
         apiVersion: state.sourceConfig.apiVersion,
@@ -222,6 +229,7 @@ const RoutingConfigurationCreateEditInner: React.FC<{
             ...cleanAnnotations,
             'openshift.io/display-name': duplicateDisplayName,
           },
+          labels: cleanLabels,
         },
         spec: state.sourceConfig.spec,
       });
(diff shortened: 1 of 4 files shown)


packages/llmd-serving/src/utils.ts
```diff
@@ -32,15 +32,34 @@ export const cleanResourceForYAMLViewer = (
...
     return annotations;
   }
-  const result = { ...annotations };
-  delete result[key];
+  const result = annotations;
+  delete result['kubectl.kubernetes.io/last-applied-configuration'];
+  delete result['serving.kserve.io/well-known-config'];
+  delete result['platform.opendatahub.io/instance.name'];
+  delete result['platform.opendatahub.io/instance.uid'];
+  delete result['platform.opendatahub.io/instance.generation'];
```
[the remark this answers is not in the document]
emilys314: Where is the hyphen one used? I Don't see it

packages/llmd-serving/src/settings/TopologyConfigurationCreateEdit.tsx
```diff
@@ -102,10 +107,12 @@ const TopologyConfigurationCreateEditInner: React.FC<{
     }
     if (state?.sourceConfig) {
       const cleanMeta = cleanResourceForYAMLViewer(state.sourceConfig.metadata);
-      const cleanAnnotations = stripAnnotation(
-        cleanMeta.annotations,
-        'kubectl.kubernetes.io/last-applied-configuration',
-      );
+      let cleanAnnotations = cleanMeta.annotations;
+      let cleanLabels = cleanMeta.annotations;
```
YuliaKrimerman: this should be ` let cleanLabels = cleanMeta.labels;` here and above

packages/llmd-serving/src/settings/RoutingConfigurationCreateEdit.tsx
```diff
@@ -207,10 +212,12 @@ const RoutingConfigurationCreateEditInner: React.FC<{
     }
     if (state?.sourceConfig) {
       const cleanMeta = cleanResourceForYAMLViewer(state.sourceConfig.metadata);
-      const cleanAnnotations = stripAnnotation(
-        cleanMeta.annotations,
-        'kubectl.kubernetes.io/last-applied-configuration',
-      );
+      let cleanAnnotations = cleanMeta.annotations;
+      let cleanLabels = cleanMeta.annotations;
```
YuliaKrimerman: here too

YuliaKrimerman (changes requested):

packages/llmd-serving/src/settings/TopologyConfigurationCreateEdit.tsx
emilys314: omg thank you for catching that, pushed a fix

packages/llmd-serving/src/settings/RoutingConfigurationCreateEdit.tsx
mturley: Looking at the other usages of `cleanResourceForYAMLViewer`, it is also called in the useMemo for `initialResource` where a duplicate name is set up. I was going to say we need to make sure to also clean annotations and labels there, but weirdly it seems that is only used for the initial state of `useK8sNameDescriptionFieldData` where it’ll just pull the name and description, so it doesn’t cause an issue here. But reading that code it’s a little confusing and seems unnecessary, I wonder if we should open a tech debt issue to clean these up. The same applies to `RoutingConfigurationCreateEdit`.

packages/llmd-serving/src/settings/RoutingConfigurationCreateEdit.tsx
```diff
@@ -207,10 +212,12 @@ const RoutingConfigurationCreateEditInner: React.FC<{
     }
     if (state?.sourceConfig) {
       const cleanMeta = cleanResourceForYAMLViewer(state.sourceConfig.metadata);
-      const cleanAnnotations = stripAnnotation(
-        cleanMeta.annotations,
-        'kubectl.kubernetes.io/last-applied-configuration',
-      );
+      let cleanAnnotations = cleanMeta.annotations;
+      let cleanLabels = cleanMeta.labels;
+      if (isDuplicateMode) {
```
mturley: Because `isDuplicateMode = !!state?.sourceConfig;`, this check is redundant. The name `sourceConfig` is confusing though, we have existingConfig for edit and sourceConfig for duplicate… maybe some renaming here is in order as part of that followup ticket

packages/llmd-serving/src/utils.ts
```diff
@@ -32,15 +32,34 @@ export const cleanResourceForYAMLViewer = (
   return result;
 };
 
-export const stripAnnotation = (
+export const stripDuplicatingAnnotations = (
```
mturley, line 35: Nit: we could maybe use some unit tests for these two utils, but they’re pretty simple so maybe it’s fine.

mturley (commented):
/lgtm

Tested on Zaffre cluster and it works well. Some observations out of scope of this PR, and a nit for unit tests. I’d be fine approving if we want to skip them.

packages/llmd-serving/src/utils.ts
emilys314: Yeah we'd have to test the parent function but that would have to be extracted out

packages/llmd-serving/src/settings/RoutingConfigurationCreateEdit.tsx
emilys314: Oh that's a good point

packages/llmd-serving/src/utils.ts
mturley: Probably not worth it

packages/llmd-serving/src/settings/RoutingConfigurationCreateEdit.tsx
emilys314: Updated, let me know how it looks

packages/llmd-serving/src/settings/RoutingConfigurationCreateEdit.tsx
emilys314: I can create a tech debt to clear up is duplicate vs edit code and reduce the init code duplication

mturley: Sure, if you think it’s worth it. We could also move the routing/config admin pages into their own folders like the `llmAcceleratorConfigs` page is, if we want. but idk how much we should nitpick cleaning up the existing stuff

packages/llmd-serving/src/settings/RoutingConfigurationCreateEdit.tsx
mturley: LGTM

mturley (approved):
/approve
/lgtm

good fix

PR #8971, opened 2026-07-29
fix(autorag,automl): add egress rules for DSPA and OGX ports in network policies

jefho-rh:
## Description

The `autorag-allow-ports` and `automl-allow-ports` network policies are missing egress rules for ports that the standalone BFF pods need to reach in user namespaces:

- **Port 8443** (DSPA pipeline server) — both AutoRAG and AutoML BFFs call this to list pipelines, create pipeline runs, and monitor run status
- **Port 8321** (OGX/LlamaStack) — the AutoRAG BFF calls this to fetch models (`/v1/models`) and vector store providers (`/v1/providers`)

This was not an issue when the BFFs ran as sidecar containers in the `rhods-dashboard` pod, because the `rhods-dashboard-allow-ports` network policy has a broad egress rule (`namespaceSelector: {}`) that covered all outbound traffic. When standalone deployment mode was introduced in #8708, each module's own network policy took effect, and the missing egress rules caused the BFF's requests to DSPA and OGX to hang until timeout.

**Symptoms:** AutoRAG experiments page stuck on loading spinner — the BFF logs show `context deadline exceeded (Client.Timeout exceeded while awaiting headers)` on every request to `ds-pipeline-dspa.<namespace>.svc.cluster.local:8443`. The "Create run" button never appears.

**Fix:** Add egress rules allowing TCP ports 8443 and 8321 to any namespace (`namespaceSelector: {}`), consistent with the existing broad egress pattern used in `rhods-dashboard-allow-ports`.

## How Has This Been Tested?

- Confirmed the BFF pod cannot reach DSPA from `redhat-ods-applications` namespace (curl timeout, exit code 28) when running in standalone mode
- Confirmed the BFF pod CAN reach DSPA when running as sidecar (covered by dashboard's network policy)
- BFF logs on the RHOAI nightly cluster (`dash-e2e-rhoai`) show repeated `context deadline exceeded` errors for every pipeline discovery request
- Pending: Jenkins E2E validation with the fix applied

## Test Impact

No automated tests are applicable — this is a manifest-only change to Kubernetes NetworkPolicies. The fix is validated by the AutoRAG and AutoML E2E tests passing on clusters where standalone deployment mode is active.

## Request review criteria:

Self checklist (all need to be checked):
- [x] The developer has manually tested the changes and verified that the changes work
- [x] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [x] The developer has added tests or explained why testing cannot be added (unit or cypress tests for related changes)
[...]

PR #8948, opened 2026-07-29
fix(gen-ai): add egress rule for OGXServer port 8321 in network policy

jharan1:
https://issues.redhat.com/browse/RHOAIENG-79633

## Description

The gen-ai-ui network policy (`gen-ai-allow-ports`) is missing an egress rule for TCP port 8321, which is the port used by OGXServer (LlamaStack) services in user namespaces. When gen-ai runs in standalone deployment mode, the network policy blocks egress to port 8321, causing `/gen-ai/api/v1/lsd/models` requests to time out with a 504 Gateway Timeout.

This was not an issue when gen-ai ran as a sidecar container in the `rhods-dashboard` pod, because the network policy's pod selector didn't match sidecar containers. When standalone deployment mode was introduced (July 15, `453d8983b`), the policy started enforcing on the new `gen-ai-ui` pods, revealing the missing rule.

**Root cause:** The egress rules allow DNS (5353), maas-ui (8243), mlflow-ui (8343), K8s API (6443), and HTTP/HTTPS (80/443) — but not OGXServer port 8321.

**Fix:** Add an egress rule allowing TCP port 8321 to any namespace (`namespaceSelector: {}`), consistent with the existing broad egress rules for ports 80/443/6443.

## How Has This Been Tested?

Tested on a live cluster (with rhoai 3.6 ea1):

1. Confirmed the gen-ai-ui pod could NOT reach the LSD service on port 8321 (curl timeout, exit code 28)
2. Manually patched the network policy to add port 8321 egress
3. Confirmed the gen-ai-ui pod could then reach the LSD service instantly (models returned successfully)
4. Verified the playground loaded models in the browser

## Test Impact

No automated tests are applicable — this is a manifest-only change to a Kubernetes NetworkPolicy. The fix was validated manually on a live cluster.

## Request review criteria:

Self checklist (all need to be checked):
- [x] The developer has manually tested the changes and verified that the changes work
- [x] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [ ] The developer has added tests or explained why testing cannot be added (unit or cypress tests for related changes)
- [x] The code follows our [Best Practices](/docs/best-practices.md) (React coding standards, PatternFly usage, performance considerations)

After the PR is posted & before it merges:
- [ ] The developer has tested their solution on a cluster by using the image produced by the PR to `main`

## Summary by CodeRabbit

* **Bug Fixes**
  * Enabled required outbound TCP connectivity on port 8321 for supported AI functionality.

PR #8979, opened 2026-07-29
fix(manifests): harden kustomize manifests [RHOAIENG-66669]

liday-rh:
https://issues.redhat.com/browse/RHOAIENG-66669

## Description

Implements all 5 acceptance criteria from [RHOAIENG-66669](https://redhat.atlassian.net/browse/RHOAIENG-66669) — manifest hardening follow-up to #8708:

**AC2 — emptyDir sizeLimit:** Add `sizeLimit: 100Mi` to all `/tmp` emptyDir volumes (base deployment, 8 module deployments, core-bff distribution — 10 files).

**AC3 — Scope egress NetworkPolicy:** Remove broad `0.0.0.0/0` port 80/443 egress from `automl` and `autorag` NetworkPolicies. These are API-only BFFs that only need DNS (5353) and K8s API (6443) egress. Retains the namespace-scoped DSPA (8443) and OGX (8321) rules added in #8971/#8948.

**AC4 — Migrate vars to replacements:** Convert all 14 kustomization.yaml files from deprecated `vars:`/`configurations:` to the `replacements:` API. Delete all `params.yaml` varReference files. Replace `$(var-name)` placeholders with literal values in deployment, consolelink, httproute, and configmap resources.

**AC1 — CI validation:** Add `manifests/odh/standalone`, `manifests/rhoai/standalone`, and `manifests/modules` to the kustomize-build CI matrix. Add a `params-env-sync` CI job to detect drift between per-module `params.env` and the sidecar parent. Fix `release-auto-merge.yml` paths from `manifests/modular-architecture/` to `manifests/sidecar/` + `manifests/modules/`.

**AC5 — Validation script:** Add `hack/validate-manifests.sh` for local dev validation (kustomize build + params.env sync check).

## How Has This Been Tested?

**Local validation:**
- All 5 kustomize build targets pass with zero deprecation warnings
- Rendered YAML diffed against pre-migration baselines — only AC2 (sizeLimit) and AC3 (egress) diffs, no functional regressions
- `hack/validate-manifests.sh` passes all checks

**Cluster testing (ODH on OpenShift):**
- Applied `kustomize build manifests/odh` to `opendatahub` namespace — 109 resources applied successfully
- Applied `kustomize build manifests/modules` — 33 resources applied successfully
- Sidecar deployment (`odh-dashboard`) running 10/10 containers
- All 7 operator-managed module deployments healthy at 2/2 (automl, autorag, gen-ai, maas, mlflow, eval-hub, model-registry)
- Verified rendered images resolve correctly via `replacements:` API (no `$(var-name)` placeholders in output)
- Verified `sizeLimit: 100Mi` present on all emptyDir volumes in rendered manifests
- Verified automl/autorag NetworkPolicies render with only DNS (5353), K8s API (6443), and DSPA/OGX namespace-scoped egress — no broad 80/443
- Only errors were 5 pre-existing ClusterRoleBinding issues (missing `namespace` on subjects) unrelated to this PR

Dashboard-operator reconciliation observed via operator logs:
image
Successful deployment of modules:
image

## Test Impact

No new unit/cypress tests needed — these are declarative YAML-only changes with no new code paths. The new `hack/validate-manifests.sh` script and `params-env-sync` CI job serve as the validation layer.

## Request review criteria:

[RHOAIENG-66669]: https://redhat.atlassian.net/browse/RHOAIENG-66669?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Added validation for manifest builds and configuration consistency across modules.
  * Expanded validation coverage for standalone and modular deployment variants.
  * Added projected service-account tokens for dashboard and module access.

* **Bug Fixes**
  * Improved image and configuration value propagation across deployments.
  * Limited temporary storage to 100 MiB to prevent unbounded usage.
  * Corrected network access rules for API-only services.

* **Chores**
  * Updated automated release version bumps and conflict handling for the current manifest layout.


liday-rh:
Addressing CodeRabbit comments:

**1. sed injection in `release-auto-merge.yml` (CWE-78)** — Fixed in 44b0a948. Replaced the `sed -i "s|...|${parent_line}|"` substitution with `awk` using environment variables, so `params.env` values are never interpolated into a dynamic program.

**2. Pin images to sha256 digests** — Out of scope. The `params.env` image values are intentionally tag-based — they get updated by the release workflow on each version bump. Pinning to sha256 digests here would break that automation. Image pinning is a separate concern.

manifests/modules/automl/networkpolicy.yaml
```diff
@@ -38,11 +38,3 @@ spec:
       ports:
         - port: 8443
           protocol: TCP
```
lucferbux, line 42: nit: removing the `0.0.0.0/0` egress on 80/443 from automl and autorag but leaving it on the other 6 modules feels inconsistent. If these two don't need external HTTP access, fair enough, but could you add a comment explaining why these two specifically? Or if none of the API-only BFFs need it, apply it consistently.

hack/validate-manifests.sh
```diff
@@ -0,0 +1,74 @@
...
+
+echo "=== Kustomize build validation ==="
+for target in "${TARGETS[@]}"; do
+  printf "  %-35s" "$target"
+  if kustomize build "$target" > /dev/null 2>&1; then
+    echo "OK"
+  else
+    echo "FAIL"
+    kustomize build "$target" > /dev/null
+    FAILED=1
```
lucferbux, line 26: nit: `FAILED=1` here is dead code. `set -e` is active (line 2), so the `kustomize build` re-run on line 25 will exit the script immediately if it fails, before reaching this line. The script still works correctly (exits non-zero), just accidentally. Either drop `set -e` and rely on the `FAILED` flag, or wrap line 25 in `|| true` so it shows the error output without exiting.

lucferbux (commented):
Went through this pretty thoroughly, the vars to replacements migration is correct and the operator compatibility checks out (the operator only writes `params.env` and calls kustomize render, it never touches vars/configurations/params.yaml directly). All kustomize builds pass and rendered output is semantically identical to main.

A few inline comments below, nothing blocking but want to make sure we're on the same page.

Also, before merging this needs:
- A CI pass of the new changes so we can validate the kustomize validation workflow actually works end to end
- Some proof of cluster deployment testing (even just screenshots of the operator reconciling with these manifests). The script validation is nice but we should verify the real thing too.

manifests/modules/automl/networkpolicy.yaml
liday-rh: Good call — added comments to both automl and autorag networkpolicies explaining that these two are API-only BFFs that only need DNS, K8s API, and namespace-scoped DSPA/OGX egress. The other modules (gen-ai, maas, mlflow, etc.) retain 80/443 because their BFFs make external HTTP(S) calls. Addressed in f89bb1a.

hack/validate-manifests.sh
liday-rh: Good catch — added `|| true` to the re-run so it prints the error output without `set -e` exiting the script before `FAILED=1`. Addressed in f89bb1a.

manifests/modules/autorag/networkpolicy.yaml
```diff
@@ -17,6 +17,8 @@ spec:
         - namespaceSelector:
             matchLabels:
               network.openshift.io/policy-group: ingress
+  # autorag is an API-only BFF — it only needs DNS, K8s API, and
+  # namespace-scoped DSPA (8443) + OGX (8321). No external HTTP(S) required.
```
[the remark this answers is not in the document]
liday-rh: `namespaceSelector: {}` is intentional here. DSPA pipeline servers (port 8443) and OGX instances (port 8321) are created dynamically in arbitrary user-project namespaces — there is no single namespace or predictable label we can scope to at manifest time. The same pattern was established in #8971 and #8948 and applies to the gen-ai module's OGX egress rule as well.

The scope is still constrained: this rule only allows egress to ports 8443/8321, not arbitrary ports across all namespaces. The BFF pod can only reach services listening on those specific ports.

manifests/modules/agent-ops/kustomization.yaml
```diff
@@ -12,15 +12,17 @@ configMapGenerator:
...
-    fieldref:
-      fieldpath: data.agent-ops-ui-image
-replacements:
+      fieldPath: data.agent-ops-ui-image
+    targets:
+      - select:
+          kind: Deployment
+          name: agent-ops-ui
+        fieldPaths:
+          - spec.template.spec.containers.[name=agent-ops-ui].image
```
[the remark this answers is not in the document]
liday-rh: These `:main` tags in `params.env` are development defaults — they are never used as-is in production. At deploy time, the `dashboard-operator` (and the parent `odh-operator` / RHOAI operator) overrides every image reference via `RELATED_IMAGE_*` environment variables, which point to digest-pinned images from Konflux builds. The operator writes these resolved values into `params.env` before running `kustomize build`.

Pinning digests in the manifest source would break the dev workflow (local builds, CI) and create unnecessary churn on every upstream image rebuild. The security boundary is enforced at the operator level, not at the manifest defaults.

lucferbux (approved):
CI green, new validation pipelines self-verify, cluster testing proof looks good. Nice work on the fix commit too.

/lgtm
/approve

manifests/modules/automl/networkpolicy.yaml
GAUNSD: Linking new PR to this conversation for paper-trail's sake:
- https://github.com/opendatahub-io/odh-dashboard/pull/9105

autorag/automl pods do need the 80/443 egress for the sake of calls made via AWS S3 SDK to work with, fetch and upload S3 files to a users S3-compatible connection

PR #8982, opened 2026-07-29
fix turbo lint cache inputs

christianvogt:
## Description
Exclude `packages/eslint-config/.turbo/**` and `packages/eslint-config/node_modules/**` from Turbo `lint` task inputs.

The shared eslint-config package is already an input so lint cache invalidates when shared rules change, but the previous glob also matched local Turbo cache and `node_modules` under that package. Those paths can change without meaningful lint input changes and cause unnecessary cache misses (or unstable cache keys) across machines/CI.

## How Has This Been Tested?
From the root run `npm run lint`. Any subsequent run without changes should result in a full run with all cache hits.
Changing a small package that no one depends on (eg packages/observability) and then running `npm run lint` should result in just that one package being linted while all the rest should hit the cache.

## Test Impact
No new tests — config-only change to Turbo cache input globs. Behavior is validated by inspecting `turbo.jsonc` and confirming the negation patterns.

## Request review criteria:

## Summary by CodeRabbit

* **Chores**
  * Improved lint task tracking to include shared lint configuration changes while excluding generated and dependency files.


antowaddle (approved):
Ran some tests on this:

  1. Lint ran and passed across all 35 packages
  2. The negation patterns didn't break anything
  3. The cache misses are a one-time cost of changing the input definition
  4. Once CI caches stabilize, the fix will prevent the spurious misses that .turbo/ and node_modules/ churn was causing
  before

/lgtm

PR #8987, opened 2026-07-30
Fix mlflow integration: add nested run IDs to compare runs page, MLflow pipeline server config

nananosirova:
https://redhat.atlassian.net/browse/RHOAIENG-80068

## Description

This PR adds two major MLflow integration features to the pipeline server UI:

### 1. MLflow Experiment Tracking Section in Pipeline Server Configuration
Screenshot 2026-07-30 at 1 24 12 AM

image

Adds a new `MlflowTrackingSection` component that appears in both the **Configure Pipelines Server** modal (initial setup) and the **Manage Pipeline Server** modal (post-creation settings). The section is gated behind two conditions: the MLflow CR must be available on the cluster (`useIsMlflowCRAvailable`) AND the `MLFLOW_PIPELINES` area must be enabled.

The section provides two toggles:
- **Automatically detect and connect to MLflow** - controls `spec.mlflow.integrationMode` on the DSPA CR (`AUTODETECT` vs `DISABLED`). When the pipeline server has MLflow integration enabled, it discovers the MLflow instance on the cluster and logs pipeline run metrics to MLflow experiments automatically.
- **Allow pipeline tasks to log directly to MLflow** - controls `spec.mlflow.injectUserEnvVars`. When enabled, pipeline tasks receive MLflow credentials as environment variables so that user code calling `mlflow.autolog()` logs to the correct MLflow run. This toggle is disabled when integration mode is off, with a tooltip explaining why.

A new `injectUserEnvVars` field was added to the `DSPipelineMlflowKind` type, and the `updatePipelineSettings` API function was extended to patch `spec.mlflow` on the DSPA resource (using JSON Patch add/replace). After saving MLflow settings, `refreshState()` is called to prevent stale context on the Create Run page.

The `PipelinesContext` now also exposes `mlflowInjectUserEnvVars` so downstream components (e.g., the autologging code snippet) can read this flag.

### 2. Nested MLflow Run IDs in Pipeline Run Tables and Details

- **Compare runs**: When comparing runs via the MLflow route, nested run IDs are now included alongside root run IDs in the compare URL, giving users visibility into per-task MLflow runs.

- **Pipeline run details page**: The run details tab now shows clickable "MLflow experiment" and "MLflow run" links (routing to the embedded MLflow UI) when the run has MLflow metadata.

- **Task details drawer**: When clicking a pipeline task node, the task details tab now shows a clickable "MLflow run" link if that specific task execution has a `plugins.mlflow.run_id` custom property. The MLflow run ID and experiment ID are extracted from the MLMD execution and the run's plugin output respectively.

- **MLflow experiment column fix**: The experiment name resolution was improved - it now looks up the live experiment name by ID first (from the MLflow API response), falling back to the name embedded in the run metadata. The loading skeleton now shows whenever experiments haven't loaded, not just when no experiment ID is in the plugin output.

### Other changes

- Lifted `useMlmdContextsByType(MlmdContextTypes.RUN)` from `useMetricColumns` up to `PipelineRunTable` so contexts are shared between metrics and execution queries, avoiding duplicate MLMD calls.
- Metrics column names are now filtered to only include metrics that exist in the visible runs (`visibleMetricsNames`), cleaning up the customize-columns modal.
- Added `mlflowRunRoute` helper to `routes/pipelines/mlflow.ts` for linking to individual MLflow runs.
- The `PipelineServerConfigType` now includes `mlflow` in its type and default values, and `createDSPipelineResourceSpec` passes the mlflow config through to the DSPA spec.

## How Has This Been Tested?

- Unit tests for `MlflowTrackingSection` (both form and description variants) covering checkbox state, toggle behavior, disabled states, and tooltip rendering
- Unit tests for `ManagePipelineServerModal` MLflow section covering: visibility gating (CR unavailable, area unavailable, loading state), initialization from DSPA resource, save button enablement, `updatePipelineSettings` call payloads, `refreshState()` invocation, and state reset on CR changes
- Unit tests for `ConfigurePipelinesServerModal` MLflow section covering: visibility gating and rendering when both gates are satisfied
- Updated `utils.spec.ts` to include `mlflow` in `PipelineServerConfigType` test fixtures
- Updated `ManagedPipelinesSettingsSection.spec.tsx` base config to include the new `mlflow` field

## Test Impact

- **New test file**: `MlflowTrackingSection.spec.tsx` - 246 lines covering form/description variants, toggle interactions, disabled states
- **Extended test file**: `ManagePipelineServerModal.spec.tsx` - ~265 lines added covering the full MLflow settings lifecycle in the manage modal
- **Extended test file**: `ConfigurePipelinesServerModal.spec.tsx` - ~55 lines added covering MLflow section visibility gating
- **Updated test fixtures**: `utils.spec.ts` and `ManagedPipelinesSettingsSection.spec.tsx` updated to include the new `mlflow` config field
- Pipeline run details (MLflow links) and nested runs column are not covered by new tests in this PR - these are visual features that rely on MLMD data and are best verified on a cluster with an active MLflow instance

## Request review criteria:

If you have UI changes:

## Summary by CodeRabbit

* **New Features**
  * Added MLflow tracking configuration with integration and credential-injection controls.
  * Added MLflow experiment and run links in pipeline details.
  * Added support for viewing and comparing nested MLflow runs.
  * Added MLflow autologging guidance when credential injection is enabled.
* **Bug Fixes**
  * Improved MLflow experiment detection, metric visibility, and incomplete-run handling.
  * Preserved MLflow settings when updating pipeline server configuration.
  * Improved navigation to MLflow experiment and run details.
  * Enforced comparison limits across root and nested runs.
  * Improved error handling when loading MLflow experiment information.


frontend/src/concepts/pipelines/content/tables/pipelineRun/PipelineRunTableRowNestedMlflowRuns.tsx
```diff
@@ -0,0 +1,19 @@
...
+  nestedRuns,
+}) => {
+  if (!nestedRuns?.length) {
+    return <NoRunContent />;
+  }
+
+  return <>{nestedRuns.map((n) => n.mlflowRunId).join(', ')}</>;
+};
+
+export default PipelineRunTableRowNestedMlflowRuns;
```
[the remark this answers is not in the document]
nananosirova: Disagree, it belongs here

DaoDaoNoCode:

frontend/src/concepts/pipelines/apiHooks/mlmd/useGetExecutionsByRuns.ts
```diff
@@ -0,0 +1,41 @@
...
+
+          const response = await metadataStoreServiceClient.getExecutionsByContext(request);
+          const executions = response.getExecutionsList();
+
+          return {
+            [run.run_id]: executions,
+          };
+        }),
+      ),
+    [contexts, metadataStoreServiceClient, runs],
```
DaoDaoNoCode: The `runs` and `contexts` arrays are new references on every render of the parent component (they come from hooks that return fresh arrays). This means the `useCallback` here gets a new identity every render, which causes `useFetchState` to re-fetch on every render cycle.

You'll want to stabilize these deps. The typical pattern in this codebase is to use a serialized key (like run IDs joined) and guard with a `NotReadyError` when data isn't ready:

```suggestion
  const runIds = runs.map((r) => r.run_id).join(',');

  const call = React.useCallback<FetchStateCallbackPromise<Record<string, Execution[]>[]>>(
    () => {
      if (!runIds || contexts.length === 0) {
        return Promise.reject(new NotReadyError('Contexts not loaded'));
      }
      return Promise.all(
        runs.map(async (run) => {
          const context = contexts.find((x) => x.getName() === run.run_id);
          if (!context) {
            return { [run.run_id]: [] };
          }
          const request = new GetExecutionsByContextRequest();
          request.setContextId(context.getId());

          const response = await metadataStoreServiceClient.getExecutionsByContext(request);
          const executions = response.getExecutionsList();

          return {
            [run.run_id]: executions,
          };
        }),
      );
    },
    // eslint-disable-next-line react-hooks/exhaustive-deps
    [runIds, contexts, metadataStoreServiceClient],
  );
```

You'd also need to import `NotReadyError` from the appropriate location. Without this fix, the hook fires a new MLMD request batch on every render of `PipelineRunTable`.

frontend/src/concepts/pipelines/content/tables/pipelineRun/PipelineRunTable.tsx
```diff
@@ -80,21 +91,54 @@ const PipelineRunTable: React.FC<PipelineRunTableProps> = ({
...
   );
   const effectiveTotalSize = mlflowFilter ? runs.length : totalSize;
+
+  const visibleMetricsNames = React.useMemo(() => {
+    const names = new Set<string>();
+    runs.forEach((run) => {
+      run.metrics.forEach((metric) => names.add(metric.name));
+    });
+    return names;
+  }, [runs]);
```
DaoDaoNoCode, line 140: nit: the eslint `react-hooks/exhaustive-deps` rule will flag this -- `runs` is not in the dependency array but is used inside the callback. Since `runs` is the filtered+metric-enriched list and changes on every filter/metrics update, you probably want `runs` in the deps. Though if the intent is only to recompute when the set of visible runs changes, consider depending on a stable derivative like run IDs.

frontend/src/concepts/pipelines/content/tables/pipelineRun/useMetricColumns.ts
```diff
@@ -49,16 +48,15 @@ export const useMetricColumnNames = (
...
   runs: PipelineRunKF[],
+  contexts: Context[],
   experimentId?: string,
 ): {
   runs: RunWithMetrics[];
   metricsColumnNames: string[];
   runArtifactsLoaded: boolean;
-  contextsError: Error | undefined;
   runArtifactsError: Error | undefined;
   metricsNames: Set<string>;
```
DaoDaoNoCode: `metricsNames` was removed from the return type interface on line 58 but is still returned in the object on line 93. That's fine from a runtime perspective, but the callers in `PipelineRunTable.tsx` no longer use `metricsNames` from this hook -- they compute `visibleMetricsNames` independently. You could clean this up by removing `metricsNames` from the return object and interface entirely, since it's now unused.

Actually wait -- looking more carefully, I see the interface on line 58 still declares `metricsNames: Set<string>`. Ignore the above, it's still there. But it IS now unused by the only consumer. nit: consider removing it to avoid confusion about which metrics set to use.

frontend/src/concepts/pipelines/content/pipelinesDetails/pipelineRun/PipelineRunTabDetails.tsx
```diff
@@ -119,6 +127,42 @@ const PipelineRunTabDetails: React.FC<PipelineRunTabDetailsProps> = ({
           },
         ]
       : []),
+    ...((): DetailItem[] => {
+      if (!isMlflowAvailable || !isPipelineRun(run)) {
```
DaoDaoNoCode, line 131: The MLflow links here show whenever `isMlflowAvailable` is true and the run has MLflow metadata, but they don't check whether the embedded MLflow UI route actually resolves. This is probably fine since `useIsMlflowPipelinesAvailable` already gates on the feature being enabled, but just want to confirm: is there a scenario where `MLFLOW_PIPELINES` area is enabled but the embedded MLflow UI isn't deployed? If so, these links would 404.

Not blocking -- just want to make sure the gate is sufficient.

frontend/src/concepts/pipelines/content/pipelinesDetails/pipelineRun/PipelineRunDrawerRightTabs.tsx
```diff
@@ -37,6 +43,12 @@ const PipelineRunDrawerRightTabs: React.FC<PipelineRunDrawerRightTabsProps> = ({
     .get('cached_execution_id')
     ?.getStringValue();
 
+  const mlflowRunId = isMlflowAvailable
+    ? taskExecution?.getCustomPropertiesMap().get('plugins.mlflow.run_id')?.getStringValue()
+    : undefined;
+
+  const mlflowExperimentId = isMlflowAvailable && run ? getMlflowExperimentId(run) : undefined;
```
DaoDaoNoCode, line 50: `getMlflowExperimentId` returns the experiment ID from the pipeline run's `plugins_output` or `plugins_input`. This is the pipeline-level experiment ID, which is correct for linking. But worth noting that if a task-level execution has a different experiment ID than the root run, this link would point to the wrong experiment. Is that a real possibility with the DSPA MLflow integration, or do all nested runs always share the root experiment?

frontend/src/concepts/pipelines/content/tables/pipelineRun/PipelineRunTable.tsx
```diff
@@ -80,21 +91,54 @@ const PipelineRunTable: React.FC<PipelineRunTableProps> = ({
...
   const {
     metricsColumnNames,
     runs: runsWithMetrics,
-    contextsError,
     runArtifactsError,
     runArtifactsLoaded,
-    metricsNames,
-  } = useMetricColumns(runWithoutMetrics, experiment?.experiment_id);
+  } = useMetricColumns(runWithoutMetrics, contexts, experiment?.experiment_id);
+  const [runExecutions, runExecutionsLoaded] = useGetExecutionsByRuns(
```
DaoDaoNoCode, line 105: This fires `getExecutionsByContext` for every displayed run on every page load, even when MLflow is disabled. The `isMlflowAvailable` guard on line 102 passes `emptyRuns` when MLflow is off, which prevents the MLMD calls -- that's good.

But when MLflow IS available, this is one MLMD RPC per visible run. On a page with 25 runs, that's 25 sequential-ish network requests (inside `Promise.all`, so parallel but still 25 RPCs). If this causes noticeable latency, you might want to batch these into a single `getExecutions` call with a filter, or at least add a loading skeleton for the nested runs column while this loads.

frontend/src/concepts/pipelines/content/configurePipelinesServer/MlflowTrackingSection.tsx
```diff
@@ -0,0 +1,114 @@
...
+  Checkbox,
+  Tooltip,
+  DescriptionListGroup,
+  DescriptionListTerm,
+  DescriptionListDescription,
+} from '@patternfly/react-core';
+import { DSPAMlflowIntegrationMode, DSPipelineMlflowKind } from '#~/k8sTypes';
+import FormSection from '#~/components/pf-overrides/FormSection';
+
+type FormVariantProps = {
```
DaoDaoNoCode: nit: `FormVariantProps` and `DescriptionVariantProps` have identical `mlflow` and `setMlflow` fields. You could simplify with a base type:

```typescript
type MlflowTrackingSectionBaseProps = {
  mlflow: DSPipelineMlflowKind;
  setMlflow: (mlflow: DSPipelineMlflowKind) => void;
};

type MlflowTrackingSectionProps = MlflowTrackingSectionBaseProps & 
  ({ variant?: 'form' } | { variant: 'description' });
```

Not blocking, just reduces duplication.

frontend/src/concepts/pipelines/content/tables/pipelineRun/PipelineRunTableRow.tsx
```diff
@@ -154,7 +157,8 @@ const PipelineRunTableRow: React.FC<PipelineRunTableRowProps> = ({
           const mlflowRunId = getMlflowRunId(run);
           const mlflowExpId = getMlflowExperimentId(run);
           if (mlflow?.isAvailable && mlflowRunId && mlflowExpId) {
-            navigate(mlflowCompareRunsRoute(namespace, [mlflowRunId], [mlflowExpId]));
+            const nestedIds = nestedMlflowRuns?.map((n) => n.mlflowRunId) ?? [];
+            navigate(mlflowCompareRunsRoute(namespace, [mlflowRunId, ...nestedIds], [mlflowExpId]));
```
DaoDaoNoCode, line 161: When a user clicks "Compare runs" from the kebab menu on a single row, this now includes all nested run IDs in the compare URL. If a run has many nested tasks (e.g., 15 pipeline steps each with their own MLflow run), the single-row compare could exceed the 10-run MLflow limit and the compare page may not handle it well.

Is there intentional handling on the compare runs page for when `allRunIds.length > 10`? The table-level compare button disables itself in that case (line 201), but the per-row action doesn't check.

DaoDaoNoCode (changes requested):
This PR adds MLflow pipeline server configuration (two toggles for integration mode and credential injection) and nested MLflow run IDs throughout the pipeline run UI. The code is well-structured overall -- the MlflowTrackingSection component is clean, the gating logic is correct, and the test coverage for the new configuration UI is solid.

Local checks: frontend lint clean, frontend type-check clean, unit tests (86/86 passed). The Cypress `pipelineCreateRuns` failure ("direct user after creating a new pipeline version" button timeout) is a pre-existing flake unrelated to this PR.

Prior reviews: The preflight agent flagged Promise.all throwing on missing MLMD context -- the author already fixed this by returning `{ [run.run_id]: [] }` for missing contexts. CodeRabbit raised several concerns around object-identity deps in useEffect, unconditional mlflow spec writing, and missing availability gates on run-details links -- I verified which of these still apply in the current code and commented only on the ones that are still relevant.

Leaving a few comments inline -- the main items are the unstable `useCallback` deps in `useGetExecutionsByRuns` (causes refetching every render) and the missing `metricsNames` in the `useMetricColumns` return after it was removed.

DaoDaoNoCode (commented):
## PR Review: Fix MLflow integration — nested run IDs + pipeline server config

**+1217 / -100** across **34 files**

---

### Overview

This PR adds two significant MLflow integration features:

1. **MLflow Tracking Section** in both the Configure and Manage Pipeline Server modals, with two checkboxes: auto-detect MLflow and inject env vars for user tasks.
2. **Nested MLflow Run IDs** exposed in run tables, detail pages, and task drawers, plus included in compare-runs URLs.

It also refactors MLMD context fetching (lifts it to `PipelineRunTable`), filters metric columns to only visible runs, and improves experiment name resolution.

---

### Strengths

- **Well-structured feature gating**: MLflow section is gated behind both `useIsMlflowCRAvailable` AND `MLFLOW_PIPELINES` area availability — clean dual-gate pattern.
- **Comprehensive test coverage**: ~560 lines of new/updated tests across 6 test files covering form/description variants, toggle interactions, disabled states, save payloads, state resets, and visibility gating.
- **Good UX details**: Disabling the inject checkbox when integration is off, tooltip explaining why, forcing `injectUserEnvVars: false` when integration is toggled off.
- **Clean separation of concerns**: `MlflowTrackingSection` supports both `form` and `description` variants via a discriminated union, reused across configure and manage modals.
- **Context refresh**: Calls `refreshState()` after saving MLflow settings so the Create Run page doesn't see stale `mlflowInjectUserEnvVars`.

---

### Issues & Suggestions

#### Correctness

1. **`useGetExecutionsByRuns` re-fetches on every render** (`useGetExecutionsByRuns.ts`)

   The hook takes `runs` (an array) and `contexts` (an array) as dependencies for `React.useCallback`. Since `runs` is likely a new array reference on each render, the fetch callback will be recreated and re-triggered every time. Consider memoizing based on run IDs or using a stable reference:

   ```ts
   const runIds = React.useMemo(() => runs.map(r => r.run_id).join(','), [runs]);
   ```
   and use `runIds` in the dependency array instead.

2. **Compare runs limit logic may confuse users** (`PipelineRunTable.tsx`)

   When nested runs push the total past 10, the compare button is disabled with a message about the total count. But the user only selected, say, 3 pipeline runs — the 10-run limit comes from MLflow's UI, not the pipeline system. The error message could be clearer that it's the *MLflow* compare limit being exceeded, not a pipeline-level restriction.

3. **IIFE pattern in `PipelineRunTabDetails.tsx`**

   The inline IIFE `...((): DetailItem[] => { ... })()` to compute MLflow detail items is hard to read. Consider extracting to a helper function or `useMemo`:

   ```ts
   const mlflowDetailItems = React.useMemo(() => {
     if (!isMlflowAvailable || !isPipelineRun(run)) return [];
     // ... build items
   }, [isMlflowAvailable, run, namespace]);
   ```

#### Potential Issues

4. **`useGetExecutionsByRuns` returns `Record<string, Execution[]>[]` — awkward shape**

   The return type is an array of single-key objects (`[{ runId1: [...] }, { runId2: [...] }]`). The consumer in `PipelineRunTable` then iterates with `forEach` → `Object.entries`. A single `Record<string, Execution[]>` would be simpler. The `Promise.all` results could be merged:

   ```ts
   const merged: Record<string, Execution[]> = {};
   results.forEach(entry => Object.assign(merged, entry));
   return merged;
   ```

5. **`eslint-disable` comments added for Modal imports** (`ConfigurePipelinesServerModal.tsx`)

   Four new `eslint-disable-next-line` comments were added for PatternFly Modal imports. These weren't in the original file — were they introduced by a rebase or are they intentional? If the rule exists to push toward a different modal pattern, consider migrating instead.

6. **Missing error handling for `refreshState()`** (`ManagePipelineServerModal.tsx`)

   `refreshState()` is called after a successful save but its rejection isn't caught. If it fails, the modal has already closed and the user won't know the context is stale. Consider adding a `.catch()` or awaiting it before closing.

#### Style & Consistency

7. **Nested ternary in `PipelineRunTableRowMlflowExperiment.tsx`**

   ```ts
   const matchedExperiment =
     experimentIdFromRun != null
       ? mlflow.experiments.find((e) => e.id === experimentIdFromRun)
       : experimentNameFromRun
       ? mlflow.experiments.find((e) => e.name === experimentNameFromRun)
       : undefined;
   ```

   Readable here but borderline. An `if/else` or early-return helper would be clearer.

8. **`config.mlflow != null` check** (`ConfigurePipelinesServerModal.tsx`)

   Uses `!= null` (loose comparison) to check for `undefined`. This is fine but inconsistent with the rest of the codebase which typically uses `!== undefined`. Minor, but worth consistency.

#### Testing Gaps

9. **No tests for `useGetExecutionsByRuns` hook or `extractMlflowNestedRuns` utility**

    The new hook and the `extractMlflowNestedRuns` function in `utils.ts` have no unit tests. The utility is pure and easily testable — would be good to add coverage for edge cases (no executions, no mlflow run IDs, root ID filtering).

10. **No tests for the nested runs compare logic**

    The compare-runs href computation with nested run IDs is a significant behavior change but isn't tested. Consider adding a test for the case where nested runs push past the 10-run limit.

---

### Security

No concerns. The patch operations use JSON Patch `add`/`replace` on the DSPA resource, which is the correct pattern. No user input is passed unsanitized to APIs.

### Performance

- Lifting `useMlmdContextsByType` to `PipelineRunTable` and sharing it between `useMetricColumns` and `useGetExecutionsByRuns` is a good optimization — avoids duplicate MLMD calls.
- The new `useGetExecutionsByRuns` fires a `Promise.all` for every run, which could be expensive with many runs. Consider whether a batch API exists or whether this should be paginated/throttled.

---

### Summary

Solid, well-tested feature PR. The main areas to address are the **unstable array references** triggering re-fetches in `useGetExecutionsByRuns`, the **awkward return shape** of that hook, and the **missing unit tests** for the new utility function and hook. The code quality and patterns are otherwise consistent with the project, and the feature gating is cleanly implemented.

frontend/src/concepts/pipelines/apiHooks/mlmd/useGetExecutionsByRuns.ts
nananosirova: Applied your suggestion and added a `contextKey` stabilizer, wrapped `runIds` in `useMemo`

frontend/src/concepts/pipelines/content/tables/pipelineRun/PipelineRunTable.tsx
nananosirova: The `useGetExecutionsByRuns` hook now depends on `runIds` (a stable string) internally, so the caller no longer passes runs as a dep. The runs reference inside the callback body updates when runIds changes, which is the correct behavior.

frontend/src/concepts/pipelines/content/tables/pipelineRun/useMetricColumns.ts
nananosirova: Removed!

frontend/src/concepts/pipelines/content/pipelinesDetails/pipelineRun/PipelineRunTabDetails.tsx
nananosirova: The `isMlflowAvailable` check here comes from `useIsMlflowPipelinesAvailable()`, which already includes a BFF health poll. The Go BFF watches the MLflow CR at runtime and dynamically resolves the tracking URL (`refreshMLflowState`). If the CR is removed or the service URL can't be resolved, the BFF reports `configured: false` and the links are hidden. There's still a narrow edge case where the pod is crashlooping but the Service/Route exist (BFF resolves the URL but the tracking server is unresponsive), but that's a general platform concern rather than something to solve at the link level.

frontend/src/concepts/pipelines/content/pipelinesDetails/pipelineRun/PipelineRunDrawerRightTabs.tsx
nananosirova: Nested runs are created under the same experiment as the root run. Cross-experiment nesting isn't supported by the integration, so `getMlflowExperimentId(run)` is always correct for linking task-level runs.

frontend/src/concepts/pipelines/content/tables/pipelineRun/PipelineRunTable.tsx
nananosirova: The stable deps fix removes the redundant fetching. The issue is that `GetExecutionsByContextRequest` only takes in a single `contextId` corresponding to a single parent run ID. The batch option is to use `GetExecutions` with `filterQuery` to get all the executions in one call, but it would just return a flat list of nested run IDs and grouping them back to their parent run IDs would need a `getContextsByExecution` call anyway, and it also only takes a single `contextId`.

frontend/src/concepts/pipelines/content/tables/pipelineRun/PipelineRunTableRow.tsx
nananosirova: This action only sends one row's root run + its nested runs.

frontend/src/concepts/pipelines/content/configurePipelinesServer/MlflowTrackingSection.tsx
nananosirova: Applied as suggested

nananosirova:

DaoDaoNoCode (commented):
6 findings from 5-agent parallel review (1 blocker, 5 suggestions). Each finding is posted as an inline comment on the relevant line.

frontend/src/concepts/pipelines/content/tables/pipelineRun/utils.ts
```diff
@@ -70,10 +101,28 @@ export const getMlflowExperimentId = (run: PipelineRunKF): string | undefined =>
...
+    return undefined;
+  }
+  const outputId = getMlflowPluginOutput(run)?.entries.root_run_id?.value;
   if (outputId != null) {
     return String(outputId);
   }
   return undefined;
 };
+
+export const extractMlflowNestedRuns = (
```
DaoDaoNoCode, line 116: **Blocker — missing unit tests for `extractMlflowNestedRuns`**

This function has three filtering conditions (truthy `mlflowRunId`, truthy `taskName`, `mlflowRunId !== rootMlflowRunId`), processes MLMD `Execution` objects, and drives the compare-runs URL construction including the >10 run guard. `__tests__/utils.spec.ts` tests the other new MLflow utils (`getMlflowExperimentId`, `getMlflowRunId`, `getMlflowPluginOutput`, etc.) but not this one.

Edge cases that need coverage:
- Execution with no `plugins.mlflow.run_id` custom property
- Execution with no `task_name` custom property
- Execution whose `mlflowRunId` matches `rootMlflowRunId` (should be excluded)
- `rootMlflowRunId` is `undefined` (all runs with valid IDs should be included)
- Empty executions array

A regression in the root-ID exclusion filter would cause duplicate run IDs in compare URLs and incorrect nested-run counts.

frontend/src/concepts/pipelines/content/ManagePipelineServerModal.tsx
```diff
@@ -97,13 +127,21 @@ const ManagePipelineServerModal: React.FC<ManagePipelineServerModalProps> = ({
...
+
     updatePipelineSettings(namespace, settings, pipelineNamespaceCR?.metadata.name ?? 'dspa')
       .then(() => {
         notification.success(
           'Pipeline server settings updated',
           'Settings have been updated successfully.',
         );
 
+        if (mlflowChanged) {
+          refreshState();
```
DaoDaoNoCode: **Suggestion — `refreshState()` called without error handling**

`refreshState()` returns a Promise (internally `Promise.all([refreshCR(), refreshRoute()])`) but is called fire-and-forget. If either call rejects, this produces an unhandled promise rejection after the success notification has already fired. The modal closes and the user never sees the failure — the UI may show stale MLflow state until the next manual page refresh.

Consider:
```typescript
if (mlflowChanged) {
  refreshState().catch((e) =>
    console.error('Failed to refresh pipeline state after MLflow update:', e),
  );
}
```

This was also flagged by CodeRabbit and not yet addressed.

frontend/src/concepts/pipelines/content/ManagePipelineServerModal.tsx
```diff
@@ -69,18 +94,23 @@ const ManagePipelineServerModal: React.FC<ManagePipelineServerModalProps> = ({
...
   React.useEffect(() => {
-    setEnableCaching(pipelineNamespaceCR?.spec.apiServer?.cacheEnabled ?? false);
-    setEnableManagedPipelines(
-      !!pipelineNamespaceCR?.spec.apiServer?.managedPipelines &&
-        !('instructLab' in (pipelineNamespaceCR.spec.apiServer.managedPipelines ?? {})),
-    );
-  }, [pipelineNamespaceCR]);
+    setEnableCaching(initCachingEnabled);
+    setEnableManagedPipelines(initManagedPipelinesEnabled);
+    setMlflow(initMlflow);
```
DaoDaoNoCode, line 102: **Suggestion — combined `useEffect` resets all fields when any single init value changes** *(pre-existing, extended by this PR)*

If CR polling updates one field upstream (e.g., another admin toggles caching) while the user has unsaved edits to a different field (e.g., MLflow settings), this effect discards all unsaved edits — not just the field that changed.

This pattern was pre-existing for `enableCaching` + `enableManagedPipelines`; this PR extends it by adding `setMlflow(initMlflow)`. Consider splitting into per-field effects so an upstream change to one setting doesn't clobber unsaved edits to the others:

```typescript
React.useEffect(() => { setEnableCaching(initCachingEnabled); }, [initCachingEnabled]);
React.useEffect(() => { setEnableManagedPipelines(initManagedPipelinesEnabled); }, [initManagedPipelinesEnabled]);
React.useEffect(() => { setMlflow(initMlflow); }, [initMlflow]);
```

Downgraded from blocker since the pattern is pre-existing. Also flagged by CodeRabbit.

frontend/src/api/pipelines/k8s.ts
```diff
@@ -192,6 +194,23 @@ export const updatePipelineSettings = async (
     }
   }
 
+  if (settings.mlflow !== undefined) {
```
DaoDaoNoCode, line 197: **Suggestion — `/spec/mlflow` replaced wholesale; could clobber operator-managed sibling fields**

The JSON patch uses `op: 'replace'` on the entire `/spec/mlflow` subtree. Today `DSPipelineMlflowKind` only has `integrationMode` and `injectUserEnvVars`, so the form object is the complete set. But if the DSPA operator adds fields under `spec.mlflow` that the dashboard doesn't surface (e.g., operator-internal state), they'd be silently dropped on save.

A safer approach would be to patch individual leaf paths:
```typescript
if (settings.mlflow !== undefined) {
  const leafPatch = (field: string, value: unknown) => {
    const path = `/spec/mlflow/${field}`;
    patches.push(existingMlflow
      ? { op: 'replace' as const, path, value }
      : { op: 'add' as const, path, value });
  };
  leafPatch('integrationMode', settings.mlflow.integrationMode);
  leafPatch('injectUserEnvVars', settings.mlflow.injectUserEnvVars);
}
```

Note: `managedPipelines` uses the same wholesale-replace pattern, so this may be an accepted convention in the codebase.

frontend/src/concepts/pipelines/apiHooks/mlmd/useGetExecutionsByRuns.ts
```diff
@@ -0,0 +1,55 @@
...
+    }
+    if (!contextKey) {
+      return Promise.reject(new NotReadyError('Contexts not loaded'));
+    }
+
+    return Promise.all(
+      runs.map(async (run) => {
+        const context = contexts.find((x) => x.getName() === run.run_id);
+        if (!context) {
+          return { [run.run_id]: [] };
```
DaoDaoNoCode, line 40: **Suggestion — silent empty-array fallback makes MLMD failures indistinguishable from "no nested runs"**

Compare with the existing `useGetArtifactsByRuns` which `throw`s on missing context, causing `useFetchState` to surface the error. This hook silently returns `{ [run.run_id]: [] }`, so an MLMD index corruption or context lookup failure produces the same output as a run that genuinely has zero nested MLflow runs — no error signal to the developer or user.

The graceful-degradation intent is reasonable (nested runs are "nice to have"), but a `console.warn` or a brief comment explaining the intentional divergence from the artifacts hook would help future readers understand why the two hooks differ:
```typescript
if (!context) {
  // Intentionally silent: nested MLflow runs are supplementary.
  // Unlike useGetArtifactsByRuns, a missing context should not
  // fail the entire batch.
  return { [run.run_id]: [] };
}
```

frontend/src/routes/pipelines/mlflow.ts
```diff
@@ -25,8 +25,25 @@ const withWorkspace = (basePath: string, namespace?: string): string => {
...
+  const basePath = `${mlflowExperimentsPath}/${encodeURIComponent(experimentId)}/runs`;
+  const params = new URLSearchParams();
+  if (namespace) {
+    params.set(WORKSPACE_QUERY_PARAM, namespace);
+  }
+  params.set(SEARCH_FILTER_QUERY_PARAM, '');
+  return `${basePath}?${params.toString()}`;
+};
+
+export const mlflowRunRoute = (experimentId: string, runId: string, namespace?: string): string =>
```
DaoDaoNoCode, line 40: **Suggestion — no test coverage for `mlflowRunRoute`**

`mlflowExperimentRoute` and `mlflowCompareRunsRoute` both have tests in `__tests__/mlflow.spec.ts`, but the new `mlflowRunRoute` has none. This function is used in `PipelineRunTabDetails` and `SelectedNodeDetailsTab` to generate navigation links to individual MLflow runs.

Suggested minimum coverage:
```typescript
describe('mlflowRunRoute', () => {
  it('builds a run URL with workspace', () => {
    expect(mlflowRunRoute('exp-1', 'run-42', 'my-ns')).toBe(
      '/develop-train/mlflow/experiments/exp-1/runs/run-42?workspace=my-ns',
    );
  });

  it('omits workspace when namespace is undefined', () => {
    expect(mlflowRunRoute('exp-1', 'run-42')).toBe(
      '/develop-train/mlflow/experiments/exp-1/runs/run-42',
    );
  });

  it('encodes special characters in IDs', () => {
    expect(mlflowRunRoute('exp/1', 'run?42', 'ns')).toBe(
      '/develop-train/mlflow/experiments/exp%2F1/runs/run%3F42?workspace=ns',
    );
  });
});
```

frontend/src/concepts/pipelines/content/configurePipelinesServer/MlflowTrackingSection.tsx
```diff
@@ -0,0 +1,107 @@
...
+      isChecked={isInjectEnabled}
+      isDisabled={!isIntegrationEnabled}
+      onChange={(_e, checked) => {
+        setMlflow({
+          ...mlflow,
+          injectUserEnvVars: checked,
+        });
+      }}
+      description={
+        <span className={!isIntegrationEnabled ? 'pf-v6-u-disabled-color-100' : undefined}>
```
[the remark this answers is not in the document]
nananosirova: This is not legit.

frontend/src/api/pipelines/k8s.ts
nananosirova: Not legit. `DSPipelineMlflowKind` has exactly two fields (`integrationMode` and `injectUserEnvVars`), both are always sent. There are no operator-managed sibling fields under `spec.mlflow` to clobber.

frontend/src/concepts/pipelines/content/createRun/contentSections/MlflowIntegrationSection.tsx
```diff
@@ -39,6 +40,7 @@ const MlflowIntegrationSection: React.FC<MlflowIntegrationSectionProps> = ({
   onChange,
```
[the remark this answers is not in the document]
nananosirova: Not legit. Cypress tests use network-level intercepts for DSPA resources (`dspaIntercepts`), so the real `PipelinesContext` works — there's no need to mock `usePipelinesAPI` directly.

frontend/src/concepts/pipelines/content/configurePipelinesServer/MlflowTrackingSection.tsx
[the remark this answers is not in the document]
nananosirova: Not legit. PF Checkbox does NOT cascade disabled styling to its `description` slot — verified in `node_modules/@patternfly/patternfly/components/Check/check.css`: disabled rules only target `.pf-v6-c-check__label` and `.pf-v6-c-check__input`, not `.pf-v6-c-check__description`. The manual class is needed.

frontend/src/concepts/pipelines/content/tables/pipelineRun/PipelineRunTable.tsx
```diff
@@ -80,21 +91,54 @@ const PipelineRunTable: React.FC<PipelineRunTableProps> = ({
     workspace: isMlflowAvailable ? namespace : '',
   });
   const { onClearFilters, ...filterToolbarProps } = usePipelineFilterSearchParams(setFilter);
+  const [contexts, , contextsError] = useMlmdContextsByType(MlmdContextTypes.RUN);
```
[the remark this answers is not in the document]
nananosirova: Not legit. `useIsMlflowPipelinesAvailable` is not cluster-wide only — it checks three levels: (1) MLflow CR availability, (2) `MLFLOW_PIPELINES` feature area flag, and (3) DSPA-level integration mode via `useIsMlflowDSPAEnabled()` which reads the current project's DSPA CR.

frontend/src/concepts/pipelines/content/configurePipelinesServer/MlflowTrackingSection.tsx
```diff
@@ -0,0 +1,107 @@
...
+import {
+  FormGroup,
+  Checkbox,
+  Tooltip,
+  DescriptionListGroup,
+  DescriptionListTerm,
+  DescriptionListDescription,
+} from '@patternfly/react-core';
+import { DSPAMlflowIntegrationMode, DSPipelineMlflowKind } from '#~/k8sTypes';
+import FormSection from '#~/components/pf-overrides/FormSection';
```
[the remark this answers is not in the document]
nananosirova: Not legit. This is a feature-specific pipeline server configuration section, not a reusable component. It belongs under `concepts/pipelines`.

frontend/src/concepts/pipelines/content/createRun/contentSections/__tests__/MlflowIntegrationSection.spec.tsx
```diff
@@ -6,6 +6,11 @@ import MlflowIntegrationSection from '#~/concepts/pipelines/content/createRun/co
 import { getDefaultMlflowFormData } from '#~/concepts/pipelines/content/createRun/utils';
 import { MlflowExperimentMode } from '#~/concepts/pipelines/content/createRun/types';
 
+const mockUsePipelinesAPI = jest.fn();
+jest.mock('#~/concepts/pipelines/context', () => ({
+  usePipelinesAPI: () => mockUsePipelinesAPI(),
+}));
```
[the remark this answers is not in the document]
nananosirova: Not applicable here. Using `jest.requireActual` on `#~/concepts/pipelines/context` pulls in the full PipelinesContext → ProjectsContext → api/index → notebooks dependency chain, which causes the test to fail. The targeted mock that only exports `usePipelinesAPI` is intentional to avoid that deep import tree.

frontend/src/concepts/pipelines/content/ManagePipelineServerModal.tsx
nananosirova: Acknowledged — this is pre-existing behavior. The modal resets all fields on open anyway, so splitting the effect wouldn't change user-visible behavior. Will track as a follow-up cleanup.

frontend/src/concepts/pipelines/content/tables/pipelineRun/PipelineRunTable.tsx
```diff
@@ -286,7 +342,7 @@ const PipelineRunTable: React.FC<PipelineRunTableProps> = ({
                   content={
                     !runArtifactsLoaded
                       ? 'Customize metrics columns: Loading metrics...'
-                      : !(metricsColumnNames.length || metricsNames.size)
+                      : !(visibleMetricsColumnNames.length || visibleMetricsNames.size)
                       ? 'Customize metrics columns: No metrics available'
```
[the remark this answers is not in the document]
nananosirova: Done — extracted into a `hasMetrics` boolean.

frontend/src/concepts/pipelines/content/ManagePipelineServerModal.tsx
```diff
@@ -58,9 +66,26 @@ const ManagePipelineServerModal: React.FC<ManagePipelineServerModalProps> = ({
     !!pipelineNamespaceCR?.spec.apiServer?.managedPipelines &&
     !('instructLab' in pipelineNamespaceCR.spec.apiServer.managedPipelines);
 
+  const initMlflow: DSPipelineMlflowKind = React.useMemo(
+    () => ({
+      integrationMode:
+        // When MLflow is available, the DSPA operator defaults to AUTODETECT if spec.mlflow is absent
+        pipelineNamespaceCR?.spec.mlflow?.integrationMode ?? DSPAMlflowIntegrationMode.AUTODETECT,
```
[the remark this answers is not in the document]
nananosirova: Not legit. The DSPA operator defaults to AUTODETECT when `spec.mlflow` is absent (line 72 comment documents this). Writing it explicitly is equivalent to the implicit default. The dirty-check compares against `initMlflow` which already has the default applied, so if the user doesn't change anything, `mlflowChanged` is false and nothing gets sent. The only case where AUTODETECT gets written is when the user changes `injectUserEnvVars`, and making the implicit default explicit is actually more deterministic.

frontend/src/concepts/pipelines/content/ManagePipelineServerModal.tsx
```diff
@@ -97,13 +127,21 @@ const ManagePipelineServerModal: React.FC<ManagePipelineServerModalProps> = ({
       settings.managedPipelines = enableManagedPipelines ? {} : undefined;
     }
 
+    if (mlflowChanged) {
```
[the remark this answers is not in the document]
nananosirova: Pre-existing gap, not introduced by this PR. The caching toggle and managed pipelines toggle already share the same Save button with no SSAR gate. Adding it only for MLflow while the others lack it would be inconsistent. This should be addressed holistically as a follow-up.

frontend/src/concepts/pipelines/content/tables/pipelineRun/utils.ts
nananosirova: Addressed.

frontend/src/concepts/pipelines/content/ManagePipelineServerModal.tsx
nananosirova: Addressed.

frontend/src/routes/pipelines/mlflow.ts
nananosirova: Addressed.

frontend/src/concepts/pipelines/content/tables/pipelineRun/PipelineRunTable.tsx
```diff
@@ -39,10 +39,21 @@ import { fireFormTrackingEvent } from '#~/concepts/analyticsTracking/segmentIOUt
...
-import { filterByMlflowExperiment, getMlflowExperimentId, getMlflowRunId } from './utils';
+import {
+  extractMlflowNestedRuns,
+  filterByMlflowExperiment,
+  getMlflowExperimentId,
+  getMlflowRunId,
+} from './utils';
+import { MlflowNestedRun } from './types';
+
+const emptyRuns: PipelineRunKF[] = [];
```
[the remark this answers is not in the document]
nananosirova: Addressed

frontend/src/concepts/pipelines/apiHooks/mlmd/useGetExecutionsByRuns.ts
nananosirova: Added comment

frontend/src/concepts/pipelines/content/tables/pipelineRun/utils.ts
DaoDaoNoCode: Verified — comprehensive test coverage added: extracts nested runs, excludes root run ID, skips missing mlflow run ID, skips missing task name, empty executions, all-root-IDs, and undefined root ID. Blocker resolved.

frontend/src/concepts/pipelines/content/ManagePipelineServerModal.tsx
DaoDaoNoCode: Verified — `refreshState()` now has `.catch()` with `console.warn`. Resolved.

frontend/src/concepts/pipelines/content/ManagePipelineServerModal.tsx
DaoDaoNoCode: Agreed — pre-existing pattern, reasonable to track as follow-up. Not blocking.

frontend/src/api/pipelines/k8s.ts
DaoDaoNoCode: Fair point — `DSPipelineMlflowKind` is dashboard-owned and the DSPA operator does not write additional fields under `spec.mlflow` today. The wholesale replace matches the `managedPipelines` precedent. Withdrawn.

frontend/src/concepts/pipelines/apiHooks/mlmd/useGetExecutionsByRuns.ts
DaoDaoNoCode: Verified — explanatory comment added documenting the intentional divergence from `useGetArtifactsByRuns`. Resolved.

nananosirova:

frontend/src/routes/pipelines/mlflow.ts
DaoDaoNoCode: Verified — tests added covering workspace, no-workspace, and special character encoding. Resolved.

frontend/src/concepts/pipelines/apiHooks/mlmd/useGetExecutionsByRuns.ts
```diff
@@ -0,0 +1,58 @@
...
+        request.setContextId(context.getId());
+
+        const response = await metadataStoreServiceClient.getExecutionsByContext(request);
+        const executions = response.getExecutionsList();
+
+        return {
+          [run.run_id]: executions,
+        };
+      }),
+    );
```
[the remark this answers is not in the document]
nananosirova: MLMD service unavailability would affect all runs equally, so this is correct behavior

DaoDaoNoCode (approved):
/lgtm

PR #8990, opened 2026-07-30
feat(RHOAIENG-41569): guard namespace injection across dashboard manifests

liday-rh:
https://issues.redhat.com/browse/RHOAIENG-41569

## Description

Add CI lint and unit tests to prevent hardcoded `metadata.namespace` regressions in manifests. The operator's dynamic namespace injection via `kustomize.WithNamespace()` is the single source of truth — these guardrails ensure it stays that way.

**Changes:**

1. **`manifests/scripts/check-no-hardcoded-namespace.sh`** — New shell script using `yq` to scan all YAML under `manifests/` for Kubernetes resources (documents with `apiVersion` + `kind`) that have a non-empty `metadata.namespace`. Naturally excludes JSON6902 patches, Downward API refs, embedded JSON in ConfigMaps, documentation strings, and CRD schemas. Excludes `kustomization.yaml` files and the `crd/` directory explicitly.

2. **`.github/workflows/validate-kustomize.yml`** — New `check-no-hardcoded-namespace` CI job that installs `yq` v4.44.1 and runs the lint script on every `manifests/**` change. Runs in parallel with existing `kustomize-build` jobs.

3. **`dashboard-operator/internal/controller/support_test.go`** — `TestNamespaceInjection`: creates temp kustomize fixtures with 4 resource types (ConfigMap, ServiceAccount, Deployment, NetworkPolicy), renders with `WithNamespace("test-target-ns")`, and asserts every resource gets the correct namespace.

4. **`dashboard-operator/internal/controller/module_deploy_test.go`** — `TestBuildFederationConfigMap_NamespaceValues`: builds a federation ConfigMap with observability enabled (using a distinct `"observability-ns"` for Perses), then asserts all module/coreBff/mlflowEmbedded entries use `ApplicationsNamespace` while the Perses entry correctly uses `PersesService.Namespace`.

## How Has This Been Tested?

**Unit tests:**
- Lint script tested locally: passes on clean manifests (exit 0), catches violations when a hardcoded `metadata.namespace` is introduced (exit 1)
- Both new Go tests pass: `go test ./internal/controller/... -run "TestNamespaceInjection|TestBuildFederationConfigMap_NamespaceValues" -v`
- Full operator test suite passes: `make test` in `dashboard-operator/` (all packages OK, including race detector)

**End-to-end cluster validation (OpenShift):**
- Built operator image (`linux/amd64`) with this branch's manifests baked in, deployed to an OpenShift cluster via `RELATED_IMAGE_ODH_DASHBOARD_OPERATOR_IMAGE`
- **All 55 dashboard-managed resources** (Deployments, Services, ServiceAccounts, ConfigMaps, NetworkPolicies) confirmed to have `metadata.namespace: opendatahub` — no hardcoded namespace leaks
- **Federation ConfigMap** verified: every `service.namespace` and `proxyService[].service.namespace` entry across all 10 modules + coreBff resolves to `opendatahub` (the operator's `ApplicationsNamespace`) — single unique namespace value confirmed via `jq`
- Operator pod running and reconciling successfully with image `quay.io/liday/odh-dashboard:rhoaieng-41569-amd64`

## Test Impact

- Added `TestNamespaceInjection` in `support_test.go` — covers kustomize namespace injection across multiple resource types
- Added `TestBuildFederationConfigMap_NamespaceValues` in `module_deploy_test.go` — covers namespace propagation in federation ConfigMap for all entry types (modules, coreBff, perses, mlflowEmbedded)
- Added CI lint job in `validate-kustomize.yml` — catches hardcoded namespace regressions on every PR

## Request review criteria:

## Summary by CodeRabbit

* **Validation**
  * Added automated checks to detect hardcoded Kubernetes namespaces in manifests.
  * Continuous integration now reports namespace validation failures with actionable details.
  * Build and synchronization jobs use read-only repository access.
* **Tests**
  * Added coverage confirming rendered resources receive the configured namespace.
  * Added tests verifying namespace assignment for module federation services and generated ConfigMaps.


lucferbux (commented):
Nice guardrail — CI lint + unit tests for namespace injection is the right approach. The shell script is clean, workflow is properly pinned, and the Go tests cover the key scenarios (kustomize injection + federation ConfigMap namespace propagation).

Two small things inline.

manifests/scripts/check-no-hardcoded-namespace.sh
```diff
@@ -0,0 +1,38 @@
...
+      ((violations++))
+    done <<< "$matches"
+  fi
+done < <(find "$MANIFEST_DIR" \( -name '*.yaml' -o -name '*.yml' \) -print | \
+  grep -v '/kustomization\.yaml$' | \
+  grep -v '/crd/' | \
+  sort | \
+  tr '\n' '\0')
+
+if (( violations > 0 )); then
```
lucferbux, line 34: nit: `find ... -print | ... | tr '\n' '\0'` is fragile — if a filename ever contains a newline it'll split into two entries. Use `find ... -print0` directly and drop the `tr`:

```bash
done < <(find "$MANIFEST_DIR" \( -name '*.yaml' -o -name '*.yml' \) -not -path '*/kustomization.yaml' -not -path '*/crd/*' -print0 | sort -z)
```

Unlikely to hit in practice with our manifests, but `find -print0 | while read -d ''` is the idiomatic safe pattern.

dashboard-operator/internal/controller/module_deploy_test.go
```diff
@@ -132,6 +132,76 @@ func TestBuildFederationConfigMap_NoEnabledField(t *testing.T) {
...
+			if ok {
+				assert.Equalf(t, appNS, svc["namespace"],
+					"%s service.namespace must match ApplicationsNamespace", name)
+			}
+		}
+	}
+}
+
 func TestPatchDeploymentFederationHash_CreatesAnnotation(t *testing.T) {
 	s := testScheme(t)
```
lucferbux, line 210: +1 to the CodeRabbit point here — if `BuildFederationConfigMap` ever stops emitting `perses` or `coreBff`, this loop silently skips the branch and the test still passes green. Add a `seen` map and assert both entries were visited after the loop:

```go
seen := make(map[string]bool)
for _, entry := range entries {
    name, _ := entry["name"].(string)
    seen[name] = true
    // ... existing switch
}
require.True(t, seen["perses"], "perses entry must be present")
require.True(t, seen["coreBff"], "coreBff entry must be present")
```

manifests/scripts/check-no-hardcoded-namespace.sh
liday-rh: Done — switched to `find ... -print0 | sort -z` directly and dropped the `tr` pipe. Also added an early `[[ ! -d "$MANIFEST_DIR" ]]` guard so a missing directory fails immediately instead of silently passing.

dashboard-operator/internal/controller/module_deploy_test.go
liday-rh: Done — added a `seen` map tracking visited entries and `require.True(t, seen["perses"])` / `require.True(t, seen["coreBff"])` assertions after the loop. Also switched the default branch to `require.Truef(t, ok, ...)` instead of silently skipping when `service` is missing.

manifests/scripts/check-no-hardcoded-namespace.sh
```diff
@@ -0,0 +1,37 @@
...
+  if [[ -n "$matches" ]]; then
+    while IFS= read -r line; do
+      echo "ERROR: $rel: $line"
+      ((violations++))
+    done <<< "$matches"
+  fi
+done < <(find "$MANIFEST_DIR" \( -name '*.yaml' -o -name '*.yml' \) \
+  -not -path '*/kustomization.yaml' \
+  -not -path '*/crd/*' \
+  -print0 | sort -z)
```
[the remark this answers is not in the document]
liday-rh: Addressed — added a directory existence check (`[[ ! -d "$MANIFEST_DIR" ]]`) before the find loop so a missing or unreadable manifest directory fails with exit 2 instead of silently passing.

.github/workflows/validate-kustomize.yml
```diff
@@ -40,6 +40,7 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
         with:
+          persist-credentials: false
```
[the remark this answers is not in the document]
liday-rh: Done — added `permissions: { contents: read }` to all three jobs (`kustomize-build`, `params-env-sync`, `check-no-hardcoded-namespace`) for least-privilege CWE-250 compliance.

lucferbux:
All review feedback (mine + CodeRabbit) has been addressed across the 5 fix commits. Code looks clean — shell script is defensive, CI workflow follows security best practices, and Go tests properly isolate namespace values. No remaining issues.

/lgtm
/approve

lucferbux:
@liday-rh you might need to rebase, I think there are some flaky tests in our ci

lucferbux (approved):
/lgtm
/approve

PR #8991, opened 2026-07-30
Match Token Rate Limits Max to CRD

katieperry4:
Closes [RHOAIENG-76348](https://redhat.atlassian.net/browse/RHOAIENG-76348)

## Description

in the token rate limits modal, it sets the new maximum to `999,999,999` instead of the previous `99,999`
the time window can old hold 5 characters, including the unit (`h`, `s`, etc) so the time needs to be maxed out at 4 chars so I set the limit to `9,999`
[toke rate limit policy](https://github.com/Kuadrant/kuadrant-operator/blob/main/config/crd/bases/kuadrant.io_tokenratelimitpolicies.yaml)
```
                           properties:
                              limit:
                                description: Limit defines the max value allowed for
                                  a given period of time
                                type: integer
                              window:
                                description: Window defines the time period for which
                                  the Limit specified above applies.
                                pattern: ^([0-9]{1,5}(h|m|s|ms)){1,4}$
                                type: string
```

https://github.com/user-attachments/assets/ca02f653-a44c-45a9-84e8-9a51c71b945a

also moved some stuff to the `rateLimits.ts` file to keep it central

## How Has This Been Tested?

tested locally

## Test Impact

updated mock test

## Request review criteria:

If you have UI changes: 

[RHOAIENG-76348]: https://redhat.atlassian.net/browse/RHOAIENG-76348?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved rate-limit validation in the subscription rate-limits editor for token count and time window values beyond allowed limits.
  - The Save button is now reliably disabled when inputs exceed maximum size, with updated helper messages guiding users.
  - Valid entries continue to work as before, allowing subscriptions to be created normally.
- **New Features**
  - Standardized rate-limit validation and error messaging across the rate-limits modal.
- **Tests**
  - Updated and expanded coverage for out-of-range token count and time-window inputs to match the new validation behavior.

### DIFF
--- a/packages/maas/frontend/src/app/utilities/rateLimits.ts
+++ b/packages/maas/frontend/src/app/utilities/rateLimits.ts
@@ -64,7 +64,7 @@ export const formatTokenLimits = (limits: ModelSubscriptionRef['tokenRateLimits'
 };
 
 export const DEFAULT_RATE_LIMIT: RateLimit = { count: 1000, time: 1, unit: 'hour' };
-export const MAX_VALUE = 999_999_999;
+export const MAX_VALUE = 1_000_000_000;
 export const MAX_WINDOW_VALUE = 9_999;
 
 export const exceedsTokenLimit = (n: number): boolean =>
@@ -81,12 +81,12 @@ export const rateLimitSchema = z.object({
     .number()
     .int()
     .min(1, 'Token count must be greater than 0')
-    .max(Number.MAX_SAFE_INTEGER, 'Token count exceeds maximum allowed value'),
+    .max(MAX_VALUE, 'Token count exceeds maximum allowed value'),
   time: z
     .number()
     .int()
     .min(1, 'Time value must be greater than 0')
-    .max(Number.MAX_SAFE_INTEGER, 'Time value exceeds maximum allowed value'),
+    .max(MAX_WINDOW_VALUE, 'Time value exceeds maximum allowed value'),
   unit: z.enum(['hour', 'minute', 'second']),
 });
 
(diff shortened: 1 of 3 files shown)


packages/maas/frontend/src/app/utilities/rateLimits.ts
```diff
@@ -61,3 +62,56 @@ export const formatTokenLimits = (limits: ModelSubscriptionRef['tokenRateLimits'
     .map((l) => `${l.limit.toLocaleString('en-US')} / ${formatWindow(l.window)}`)
     .join(' | ');
 };
+
+export const DEFAULT_RATE_LIMIT: RateLimit = { count: 1000, time: 1, unit: 'hour' };
+export const MAX_VALUE = 999_999_999;
```
katieperry4: CRD doesn't have a limit it seems so we could just as well remove the limit here, we'd have to keep the window one though

Griffin-Sullivan: the MaaSSubscription told me `Value is above the maximum of 1000000000` so we should probably use that

Griffin-Sullivan (commented):
/approve
Nice work!

sridarna:
The UI Accepts token limits and time for h,m,s matching with the MaasSubscription Yaml.
Token limit holds max count of 1000,000,000.
h,m,s can be set to 9,999
/lgtm
Screenshot 2026-07-30 at 3 00 52 PM
Screenshot 2026-07-30 at 3 01 04 PM

PR #9028, opened 2026-07-31
RHOAIENG-79525: Fix probe handler conflict during 3.4→3.5 upgrade

lucferbux:
https://issues.redhat.com/browse/RHOAIENG-79525

## Description

When upgrading from RHOAI 3.4 to 3.5, the dashboard-operator fails to apply the `rhods-dashboard` Deployment because of a probe handler type conflict. The old 3.4 deployment uses `tcpSocket` for liveness and `httpGet` for readiness probes, while the new 3.5 manifests use `exec` (curl) for both.

Server-Side Apply (SSA) adds the new `exec` handler but cannot remove the old handler types because they are owned by a different field manager. Kubernetes rejects the update with:

```
may not specify more than 1 handler type
```

This causes the operator to enter a continuous reconciliation failure loop, with `DashboardReady: False` on the DSC.

### Fix

Adds `sanitizeDeploymentProbes()` which runs before `deployer.Deploy()` in both `reconcileSidecar()` and `reconcileStandalone()` paths. For each Deployment in the desired resources:

1. Fetches the live Deployment from the cluster
2. Compares probe handler types per container (matched by name) for liveness, readiness, and startup probes
3. If the live probe has a different handler type than the desired probe, builds a strategic merge patch that nulls out the stale handler type
4. Applies the patch before SSA proceeds, preventing the conflict

The `MergeDeployments` strategy from `odh-platform-utilities` only preserves `spec.replicas` and container `resources` — it does not handle probes, so this fix must live in the dashboard-operator.

## How Has This Been Tested?

- Unit tests covering all scenarios (see Test Impact)
- `go test -race ./internal/controller/` passes
- Verified on a live cluster that the operator is currently stuck in the exact failure mode this fixes (continuous `ProvisioningSucceeded=False` with the probe conflict error since the 3.5 upgrade)

## Test Impact

Added `probe_sanitize_test.go` with 5 test cases:
- `NoLiveDeployment` — deployment doesn't exist in cluster, skip silently (first deploy)
- `NoConflict` — live and desired both use `exec`, no patch needed
- `TcpSocketToExec` — the exact 3.4→3.5 upgrade scenario: live has `tcpSocket` liveness + `httpGet` readiness, desired has `exec` for both
- `MultipleContainers` — only the container with a conflict gets patched, others untouched
- `SkipsNonDeployments` — Service and other resources are ignored

## Request review criteria:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved deployment reliability by removing conflicting health-check probe settings before updates.
  * Supports liveness, readiness, and startup probes using exec, HTTP, TCP, and gRPC handlers.
  * Prevents deployment when probe cleanup fails and reports a provisioning error.

* **Tests**
  * Added coverage for probe conflicts, multiple containers, missing deployments, compatible probes, and non-deployment resources.

### DIFF
--- a/dashboard-operator/internal/controller/probe_sanitize.go
+++ b/dashboard-operator/internal/controller/probe_sanitize.go
@@ -1,0 +1,154 @@
+package controller
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+var probeHandlerTypes = []string{"exec", "httpGet", "tcpSocket", "grpc"}
+
+func sanitizeDeploymentProbes(ctx context.Context, cli client.Client, resources []unstructured.Unstructured) error {
+	logger := log.FromContext(ctx)
+
+	for i := range resources {
+		res := &resources[i]
+		if res.GetKind() != "Deployment" || res.GroupVersionKind().Group != "apps" {
+			continue
+		}
+
+		live := &unstructured.Unstructured{}
+		live.SetGroupVersionKind(res.GroupVersionKind())
+		key := client.ObjectKeyFromObject(res)
+
+		if err := cli.Get(ctx, key, live); err != nil {
+			if k8serrors.IsNotFound(err) {
+				continue
+			}
+			return fmt.Errorf("getting live deployment %s: %w", key, err)
+		}
+
+		patch, err := buildProbeCleanupPatch(live, res)
+		if err != nil {
+			return fmt.Errorf("building probe cleanup patch for %s: %w", key, err)
+		}
+		if patch == nil {
+			continue
+		}
+
+		patchBytes, err := json.Marshal(patch)
+		if err != nil {
+			return fmt.Errorf("marshalling probe cleanup patch for %s: %w", key, err)
+		}
+
+		logger.Info("Sanitizing conflicting probe handlers before SSA", "deployment", key)
+
+		if err := cli.Patch(ctx, live, client.RawPatch(types.StrategicMergePatchType, patchBytes)); err != nil {
+			return fmt.Errorf("patching conflicting probes for %s: %w", key, err)
+		}
+	}
+
+	return nil
+}
+
+func buildProbeCleanupPatch(live, desired *unstructured.Unstructured) (map[string]interface{}, error) {
+	liveContainers, _, _ := unstructured.NestedSlice(live.Object, "spec", "template", "spec", "containers")
+	desiredContainers, _, _ := unstructured.NestedSlice(desired.Object, "spec", "template", "spec", "containers")
+
+	desiredByName := make(map[string]map[string]interface{}, len(desiredContainers))
+	for _, c := range desiredContainers {
+		cMap, ok := c.(map[string]interface{})
+		if !ok {
+			continue
+		}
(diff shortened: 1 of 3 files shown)


liday-rh (commented):
Well-scoped fix for the SSA probe handler conflict during 3.4→3.5 upgrades. The approach is correct — pre-patching stale handler types before SSA proceeds is the right way to handle this, and the PR description clearly explains why `MergeDeployments` can't cover it.

Test coverage is solid (5 cases covering the key scenarios). A few minor items inline.

One non-code note: the "developer has manually tested" self-checklist box is unchecked, though the body says it was verified on a live cluster — worth checking the box if true.

dashboard-operator/internal/controller/probe_sanitize.go
```diff
@@ -0,0 +1,154 @@
...
+		}
+	}
+
+	return nil
+}
+
+func buildProbeCleanupPatch(live, desired *unstructured.Unstructured) (map[string]interface{}, error) {
+	liveContainers, _, _ := unstructured.NestedSlice(live.Object, "spec", "template", "spec", "containers")
+	desiredContainers, _, _ := unstructured.NestedSlice(desired.Object, "spec", "template", "spec", "containers")
+
```
liday-rh, line 60: Minor: the error return is currently dead code — this function never returns a non-nil error. Consider simplifying to `func buildProbeCleanupPatch(live, desired *unstructured.Unstructured) map[string]interface{}`, or if the error return is intentionally reserved for future use, a short comment would help.

dashboard-operator/internal/controller/probe_sanitize.go
```diff
@@ -0,0 +1,154 @@
...
+	desiredByName := make(map[string]map[string]interface{}, len(desiredContainers))
```
liday-rh, line 61: Question: this only processes `containers`, not `initContainers`. If any Deployment in the manifest set has init containers with probes that could hit the same SSA conflict, they'd be missed. Likely fine for the current `rhods-dashboard` Deployment, but worth confirming no other Deployments in the manifest set use init container probes — or adding a second pass over `initContainers` for completeness.

dashboard-operator/internal/controller/probe_sanitize.go
```diff
@@ -0,0 +1,154 @@
...
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+var probeHandlerTypes = []string{"exec", "httpGet", "tcpSocket", "grpc"}
+
+func sanitizeDeploymentProbes(ctx context.Context, cli client.Client, resources []unstructured.Unstructured) error {
```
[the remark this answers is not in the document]
lucferbux: Yes, the selector issue is handled separately in PR #8974 (already merged). That one adds `app.opendatahub.io/rhods-dashboard: "true"` to `commonLabels` in the RHOAI kustomization files. This PR only addresses the probe handler conflict which is the remaining failure after #8974.

dashboard-operator/internal/controller/probe_sanitize.go
lucferbux: Good catch, fixed in 840f3a0 — simplified the signature to just return `map[string]interface{}`.

dashboard-operator/internal/controller/probe_sanitize.go
lucferbux: Correct, none of the dashboard deployments use init container probes — this is intentionally scoped to `containers` since it targets the specific 3.4→3.5 upgrade conflict. Long term this whole thing moves to `MergeDeployments` in `odh-platform-utilities` (RHOAIENG-80306) which will handle it more generically.

PR #9060, opened 2026-08-04
fix(e2e): Feature Store flake — wait for operator reconciliation (RHOAIENG-78195)

kanishka-commits:
https://issues.redhat.com/browse/RHOAIENG-78195

## Description
`testFeatureStoreFeatures.cy.ts` (and other Feature Store E2E specs using `createFeatureStoreCR`) fail because setup only waits for the Feast pod. Dashboard discovery (`GET /api/featurestores`) also requires:

1. `Registry: True` on the FeatureStore CR
2. `opendatahub.io/feast=true` on the namespace

Without those, the UI shows the empty state and the project selector never appears.

This change extends `createFeatureStoreCR` to poll for both conditions via the existing `pollUntilSuccess` helper after `waitForPodReady`.

## How Has This Been Tested?
- Code change reviewed against ticket ACs and backend discovery logic in `backend/src/routes/api/featurestores/`
- Not yet run on the live e2e cluster from this PR; please verify via CI Feature Store E2E / cluster run of `testFeatureStoreFeatures.cy.ts`

## Test Impact
- Updates shared E2E setup helper used by Feature Store Cypress E2E specs
- No new unit tests (cluster/operator wait logic; covered by existing E2E once stable)
- No production code changes

## Request review criteria:

### Testing instructions
1. Run Feature Store E2E (or wait for CI `@FeatureStore` / `@FeatureStoreCI` jobs), especially `packages/cypress/cypress/tests/e2e/featureStore/testFeatureStoreFeatures.cy.ts`
2. Confirm setup no longer times out looking for `feature-store-project-selector-toggle`
3. Confirm the new polls succeed (Registry condition + namespace label) before UI navigation

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Feature Store resource setup by waiting for registry readiness and the required namespace configuration after pods become ready.
  * Helps ensure dependent operations start only when the Feature Store environment is fully prepared.

### DIFF
--- a/packages/cypress/cypress/utils/oc_commands/featureStoreResources.ts
+++ b/packages/cypress/cypress/utils/oc_commands/featureStoreResources.ts
@@ -1,4 +1,4 @@
-import { applyOpenShiftYaml, waitForPodReady } from '../oc_commands/baseCommands';
+import { applyOpenShiftYaml, pollUntilSuccess, waitForPodReady } from '../oc_commands/baseCommands';
 import { AWS_BUCKETS } from '../s3Buckets';
 import { maskSensitiveInfo } from '../maskSensitiveInfo';
 
@@ -33,6 +33,18 @@ export const createFeatureStoreCR = (namespace: string, feastInstanceName: strin
     applyOpenShiftYaml(yamlContent);
     //wait for the feature store cr to be created
     waitForPodReady(feastInstanceName, '300s', namespace);
+
+    // Wait for Feast operator reconciliation so the dashboard can discover the Feature Store
+    pollUntilSuccess(
+      `oc get featurestores.feast.dev ${feastInstanceName} -n ${namespace} -o json | jq -e '.status.conditions[]? | select(.type=="Registry") | .status == "True"'`,
+      `FeatureStore/${feastInstanceName} Registry condition to be True`,
+      { maxAttempts: 60, pollIntervalMs: 5000 },
+    );
+    pollUntilSuccess(
+      `oc get namespace ${namespace} -o json | jq -e '.metadata.labels["opendatahub.io/feast"] == "true"'`,
+      `namespace ${namespace} to have opendatahub.io/feast=true label`,
+      { maxAttempts: 60, pollIntervalMs: 5000 },
+    );
   });
 };
 
--- a/packages/cypress/cypress/utils/oc_commands/featureStoreResources.ts
+++ b/packages/cypress/cypress/utils/oc_commands/featureStoreResources.ts
@@ -43,7 +43,7 @@ export const createFeatureStoreCR = (namespace: string, feastInstanceName: strin
     pollUntilSuccess(
       `oc get namespace ${namespace} -o json | jq -e '.metadata.labels["opendatahub.io/feast"] == "true"'`,
       `namespace ${namespace} to have opendatahub.io/feast=true label`,
-      { maxAttempts: 60, pollIntervalMs: 5000 },
+      { maxAttempts: 30, pollIntervalMs: 5000 },
     );
   });
 };


ConorOM1:
@kanishka-commits please test via our Jenkins pipeline and share the job number here if it passes

claudialphonse78 (commented):
@kanishka-commits could you close this PR this issue is due to feast operator and just adding polling to verify the operator state is not helpful when the issue lies on the feast operator side

packages/cypress/cypress/utils/oc_commands/featureStoreResources.ts
```diff
@@ -33,6 +33,18 @@ export const createFeatureStoreCR = (namespace: string, feastInstanceName: strin
...
+    // Wait for Feast operator reconciliation so the dashboard can discover the Feature Store
+    pollUntilSuccess(
+      `oc get featurestores.feast.dev ${feastInstanceName} -n ${namespace} -o json | jq -e '.status.conditions[]? | select(.type=="Registry") | .status == "True"'`,
+      `FeatureStore/${feastInstanceName} Registry condition to be True`,
+      { maxAttempts: 60, pollIntervalMs: 5000 },
+    );
+    pollUntilSuccess(
+      `oc get namespace ${namespace} -o json | jq -e '.metadata.labels["opendatahub.io/feast"] == "true"'`,
+      `namespace ${namespace} to have opendatahub.io/feast=true label`,
+      { maxAttempts: 60, pollIntervalMs: 5000 },
```
claudialphonse78: bring down the maxAttempts to 30

claudialphonse78 (changes requested):
The real underlying cause is the operator failing, cause of which the tests were failing
verify this on a jenkins run if we still want this as a surety check, if this check still works @kanishka-commits

kanishka-commits:
@ConorOM1 
Jenkins build `/2154` for the PR was re-run (earlier- `/2131`) , but the Feature Store tag filter is being overridden by the pipeline’s --env CY_TEST_TIMEOUT_SECONDS=480, so it runs the full test suite instead of only the Feature Store tests.

The new setup waits are working as expected, and the RBAC and Workbench tests passed. The Features test still fails separately when there are 0 datasets and pagination is not available.

Could you please help confirm the correct way to use CYPRESS_PARALLEL_TAGS=@FeatureStore so we can run only the Feature Store tests and properly validate this PR? Thanks!

packages/cypress/cypress/utils/oc_commands/featureStoreResources.ts
kanishka-commits: Addressed in e539e922a4bbc0ab732dfdfd67f432505876e612.

kanishka-commits:
### NEW Jenkins build - /2157
Jenkins build **/2157** ran only the 3 Feature Store tests, as expected. The PR wait fix is working- Feast was ready in about 19 seconds. RBAC and Workbench Connect passed. The Features test still fails because there are 0 datasets, so the pagination check fails. This is separate from the PR.

The build is **UNSTABLE** because of this Features test failure. Features won’t pass on re-run until we fix the Datasets assertion.

kanishka-commits:
Local verify on dash-e2e-rhoai (PR wait fix):

- Setup waits + login + Feature Store discovery work
- Overview / Entities / Data sources counts match (video shows Data sources 5===5)
- Spec still fails on Datasets empty state (0 saved datasets in shared registry) — same as Jenkins /2157; separate from this PR


https://github.com/user-attachments/assets/ff76692f-a3f9-459e-9e0b-af2823a066c2


Should not block merge of the reconciliation waits.

Kindly review @ConorOM1. Thankyou.

claudialphonse78:
> Local verify on dash-e2e-rhoai (PR wait fix):
> 
> * Setup waits + login + Feature Store discovery work
> * Overview / Entities / Data sources counts match (video shows Data sources 5===5)
> * Spec still fails on Datasets empty state (0 saved datasets in shared registry) — same as Jenkins /2157; separate from this PR
> 
>  Screen.Recording.2026-08-10.at.8.01.00.PM.mov 
> Should not block merge of the reconciliation waits.
> 
> Kindly review @ConorOM1. Thankyou.

@kanishka-commits the test failing for the case when data sets is not present is more of a feature store repository issue as the feast repository(credit-scoring) used for the e2e, is it running on the separate run the data sets part?

kanishka-commits:
@claudialphonse78 Yes, this is a data issue with the shared `credit_scoring_local` Feast registry, which currently has no datasets. It’s not related to this PR.

Also, Datasets is not a separate test. It’s just one step in the same test, so when it fails, the remaining pages in that test don’t run.

claudialphonse78:
> @claudialphonse78 Yes, this is a data issue with the shared `credit_scoring_local` Feast registry, which currently has no datasets. It’s not related to this PR.
> 
> Also, Datasets is not a separate test. It’s just one step in the same test, so when it fails, the remaining pages in that test don’t run.

I am talking about the other run you mentioned about here 
> Spec still fails on Datasets empty state (0 saved datasets in shared registry) — same as Jenkins /2157; separate from this PR
is that failing there too ? consistently ? is the test for feature store passing there?

kanishka-commits:
> is that failing there too ? consistently ? is the test for feature store passing there?

@claudialphonse78 Yes, it fails there too, consistently on the Datasets step
The Feature Store test is **not fully passing**: 1/3 fails, while RBAC and Workbench Connect pass. This is the same empty-registry issue, not the PR’s wait fix

kanishka-commits:
## Verified on Jenkins /2191
 **All 3 Feature Store Cypress specs green on dash-e2e-rhoai**
Screenshot 2026-08-12 at 8 07 11 AM

claudialphonse78 (approved):

claudialphonse78 (approved):

PR #9067, opened 2026-08-04
Add model capabilities WizardField

ppadti:
Closes:  https://redhat.atlassian.net/browse/RHOAIENG-80917
## Description
Adds model capabilities section to wizard when the modelCapabilities feature flag is on
Screenshot 2026-08-04 at 9 16 35 PM

Screenshot 2026-08-04 at 9 16 44 PM

Screenshot 2026-08-04 at 9 17 02 PM

Screenshot 2026-08-04 at 9 17 35 PM

Edit:
Screenshot 2026-08-04 at 9 18 02 PM

## How Has This Been Tested?
Turn on the modelCapabilities and see this model capabilities section in Advanced setting step
Check also for the edit flow

## Test Impact
Added both cypress and unit tests

## Request review criteria:

If you have UI changes: 

## Summary by CodeRabbit

* **New Features**
  * Added model capabilities to the deployment wizard’s advanced options for generative models.
  * Select built-in capabilities such as Vision and Transcription.
  * Add, remove, and manage custom capabilities with duplicate detection.
  * Existing capabilities are displayed when editing deployments and preserved during creation and updates.
  * Capability labels use distinct colors for built-in and custom values.
  * Capability options can be enabled selectively through feature configuration.

* **Bug Fixes**
  * Improved handling of missing, invalid, duplicate, or empty capability data.

### DIFF
--- a/packages/cypress/cypress/pages/modelServing.ts
+++ b/packages/cypress/cypress/pages/modelServing.ts
@@ -1354,7 +1354,7 @@ class ModelServingWizard extends Wizard {
   }
 
   removeCapability(capability: string) {
-    this.findCapabilityLabel(capability).findByLabelText('Close').click();
+    this.findCapabilityLabel(capability).findByLabelText(`Close ${capability}`).click();
     return this;
   }
 
--- a/packages/cypress/cypress/pages/modelServing.ts
+++ b/packages/cypress/cypress/pages/modelServing.ts
@@ -1311,6 +1311,67 @@ class ModelServingWizard extends Wizard {
     return cy.findByTestId('use-case-input');
   }
 
+  findModelCapabilitiesField() {
+    return cy.findByTestId('model-capabilities-field');
+  }
+
+  findAddCapabilityButton() {
+    return cy.findByTestId('add-capability-btn');
+  }
+
+  findWellKnownCapabilityOption(capability: string) {
+    return cy.findByTestId(`well-known-capability-${capability}`);
+  }
+
+  findCustomCapabilityInput() {
+    return cy.findByTestId('custom-capability-input');
+  }
+
+  findAddCustomCapabilityButton() {
+    return cy.findByTestId('add-custom-capability-btn');
+  }
+
+  findCapabilityLabel(capability: string) {
+    return cy.findByTestId(`capability-label-${capability}`);
+  }
+
+  openAddCapabilityDropdown() {
+    this.findAddCapabilityButton().click();
+    return this;
+  }
+
+  selectWellKnownCapability(capability: string) {
+    this.openAddCapabilityDropdown();
+    this.findWellKnownCapabilityOption(capability).click();
+    return this;
+  }
+
+  addCustomCapability(capability: string) {
+    this.openAddCapabilityDropdown();
+    this.findCustomCapabilityInput().type(capability);
+    this.findAddCustomCapabilityButton().click();
+    return this;
+  }
+
+  removeCapability(capability: string) {
+    this.findCapabilityLabel(capability).findByLabelText('Close').click();
+    return this;
+  }
+
+  navigateGenerativeLegacyToAdvancedOptions() {
+    this.findModelTypeSelectOption('Generative AI model (Example, LLM)').click();
+    this.findModelLocationSelectOption('Existing connection').click();
+    this.findExistingConnectionSelect().click();
+    this.findExistingConnectionSelectOption('Test URI Secret').click();
+    this.findNextButton().click();
+    this.findModelDeploymentNameInput().type('test-model');
+    this.selectDeploymentMethodByKey('legacy');
+    this.findServingRuntimeTemplateSearchSelector().click();
+    this.selectGlobalScopedTemplateOption('vLLM NVIDIA');
+    this.findNextButton().click();
+    return this;
+  }
+
   findCPURequestedInput() {
     return cy.findByTestId('cpu-requests-input').find('input');
   }
--- a/packages/model-serving/cypress/tests/mocked/modelServing/modelServingModelCapabilities.cy.ts
+++ b/packages/model-serving/cypress/tests/mocked/modelServing/modelServingModelCapabilities.cy.ts
@@ -167,7 +167,7 @@ const initIntercepts = ({
       statusCode: 200,
       body: mockServingRuntimeK8sResource({}),
     },
-  ).as('createServingRuntime');
+  );
 
   cy.interceptK8s(
     'POST',
@@ -251,6 +251,37 @@ const setupDeploymentLists = () => {
   );
 };
 
+const setupEditDeployment = (capabilities: string[]) => {
+  const inferenceService = mockInferenceServiceK8sResource({
+    modelType: ServingRuntimeModelType.GENERATIVE,
+    hasExternalRoute: true,
+    secretName: 'test-uri-secret',
+    hardwareProfileName: 'small-profile',
+    hardwareProfileNamespace: 'opendatahub',
+    description: 'test-description',
+  });
+  inferenceService.metadata.annotations = {
+    ...inferenceService.metadata.annotations,
+    [MODEL_CAPABILITIES_ANNOTATION]: JSON.stringify(capabilities),
+  };
+
+  cy.interceptK8sList(
+    { model: InferenceServiceModel, ns: 'test-project' },
+    mockK8sResourceList([inferenceService]),
+  );
+  cy.interceptK8sList(
+    { model: ServingRuntimeModel, ns: 'test-project' },
+    mockK8sResourceList([
+      mockServingRuntimeK8sResource({
+        scope: 'global',
+        templateDisplayName: 'vLLM NVIDIA',
+      }),
+    ]),
+  );
+
+  return inferenceService;
+};
+
 describe('Model capabilities in deployment wizard', () => {
   it('should hide the model capabilities field when the feature flag is disabled', () => {
     initIntercepts({ modelCapabilitiesEnabled: false });
@@ -284,32 +315,7 @@ describe('Model capabilities in deployment wizard', () => {
 
   it('should show existing model capabilities when editing a deployment', () => {
     initIntercepts({ modelCapabilitiesEnabled: true });
-    const inferenceService = mockInferenceServiceK8sResource({
-      modelType: ServingRuntimeModelType.GENERATIVE,
-      hasExternalRoute: true,
-      secretName: 'test-uri-secret',
-      hardwareProfileName: 'small-profile',
-      hardwareProfileNamespace: 'opendatahub',
-      description: 'test-description',
-    });
-    inferenceService.metadata.annotations = {
-      ...inferenceService.metadata.annotations,
-      [MODEL_CAPABILITIES_ANNOTATION]: JSON.stringify(['Vision', 'Existing Custom']),
-    };
-
-    cy.interceptK8sList(
-      { model: InferenceServiceModel, ns: 'test-project' },
-      mockK8sResourceList([inferenceService]),
-    );
-    cy.interceptK8sList(
-      { model: ServingRuntimeModel, ns: 'test-project' },
-      mockK8sResourceList([
-        mockServingRuntimeK8sResource({
-          scope: 'global',
-          templateDisplayName: 'vLLM NVIDIA',
-        }),
-      ]),
-    );
+    setupEditDeployment(['Vision', 'Existing Custom']);
 
     modelServingGlobal.visit('test-project');
     modelServingGlobal.getModelRow('Test Inference Service').findKebabAction('Edit').click();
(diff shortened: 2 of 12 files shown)


packages/model-serving/cypress/tests/mocked/modelServing/modelServingModelCapabilities.cy.ts
```diff
@@ -0,0 +1,377 @@
...
+  ).as('createInferenceService');
+
+  cy.interceptK8s(
+    'POST',
+    { model: ServingRuntimeModel, ns: 'test-project' },
+    {
+      statusCode: 200,
+      body: mockServingRuntimeK8sResource({}),
+    },
+  ).as('createServingRuntime');
```
manaswinidas: **Dead alias:** `@createServingRuntime` is never `cy.wait`’d (only `@createInferenceService` / `@updateInferenceService` are). Drop `.as('createServingRuntime')` unless you sync on it — keep the intercept if the wizard still POSTs a ServingRuntime.

packages/model-serving/cypress/tests/mocked/modelServing/modelServingModelCapabilities.cy.ts
```diff
@@ -0,0 +1,377 @@
...
+    );
+    cy.interceptK8sList(
+      { model: ServingRuntimeModel, ns: 'test-project' },
+      mockK8sResourceList([
+        mockServingRuntimeK8sResource({
+          scope: 'global',
+          templateDisplayName: 'vLLM NVIDIA',
+        }),
+      ]),
+    );
```
manaswinidas: **DRY / maintenance:** this edit setup (inference service + capabilities annotation + IS/SR list intercepts) is nearly copy-pasted into the next test (`should persist changed model capabilities on edit submit`). Prefer a small helper, e.g. `setupEditDeployment(capabilities: string[])`, and reuse it in both.

Also, the file-local `MODEL_CAPABILITIES_ANNOTATION = 'opendatahub.io/model-capabilities'` duplicates the shared constant — import `MODEL_CAPABILITIES_ANNOTATION` from `@odh-dashboard/model-serving/shared/modelCapabilities` (or the package’s public path) instead of redeclaring the string.

packages/cypress/cypress/pages/modelServing.ts
```diff
@@ -1311,6 +1311,67 @@ class ModelServingWizard extends Wizard {
...
+    this.openAddCapabilityDropdown();
+    this.findCustomCapabilityInput().type(capability);
+    this.findAddCustomCapabilityButton().click();
+    return this;
+  }
+
+  removeCapability(capability: string) {
+    this.findCapabilityLabel(capability).findByLabelText('Close').click();
+    return this;
+  }
```
manaswinidas, line 1359: **Dead code (for now):** `addCustomCapability` / `removeCapability` (and the custom-input page finders they wrap) are never used by `modelServingModelCapabilities.cy.ts` — tests only exercise well-known select + submit/edit.

Either add a Cypress case that covers custom add + remove, or drop these helpers until that coverage exists so the page object doesn’t advertise untested API.

manaswinidas (commented):
Non-blocking review focused on dead Cypress aliases, DRY in the edit setup, and unused page-object helpers. Nothing merge-blocking.

manaswinidas (changes requested):
When we switch off feature flag and then switch on feature flag on the "Advanced settings" tab - the model capabilities are empty when it shouldn't be. Is this expected behavior ? Also, should the annotation be populated even when the feature flag is off ?

https://github.com/user-attachments/assets/25c3e5c8-992a-4aa1-a344-78d9dbea601c

ppadti:
> When we switch off feature flag and then switch on feature flag on the "Advanced settings" tab - the model capabilities are empty when it shouldn't be. Is this expected behavior ? Also, should the annotation be populated even when the feature flag is off ?
> 
>  Screen.Recording.2026-08-05.at.6.06.10.PM.mov

Addressed this in latest commit - Let me know if that works

packages/model-serving/extensions/odh.ts
```diff
@@ -17,6 +22,79 @@ const createRedirectComponent = (args: { from: string; to: string }) => () =>
...
+  flags: {
+    required: [SupportedArea.MODEL_CAPABILITIES],
+  },
+};
+
+const modelCapabilitiesApplyKServe: WizardFieldApplyExtension<string[]> = {
+  type: 'model-serving.deployment/wizard-field-apply',
+  properties: {
+    fieldId: 'modelCapabilities',
+    platform: 'kserve',
```
emilys314: This makes sense given `platform` is a required field, but this shows a limitation in the extension. Despite this being a generic field, you needed to define it for both packages even though the `applyModelCapabilities` and `extractModelCapabilities` is the exact same code which works generically on a `Deployment` type.

I think we can fix this by updating the underlying `WizardFieldApplyExtension.properties.platform` type from a `platform: D['modelServingPlatformId'];` to a `platform: D['modelServingPlatformId'] | 'all';`. 

You'll need to update the `packages/model-serving/src/components/deploymentWizard/useWizardFieldExtractors.ts` to always include `all` platforms too.

```suggestion
    platform: 'all',
```

This will make it so you don't have to repeat yourself by making duplicate extensions for a field that is supposed to be for "all" platforms.

---
_Side thoughts:_
- _I think "all" is better than a `null`, because you have to manually opt into the "all" behavior which reduces the chances of accidentally making something global_

packages/model-serving/src/components/deploymentWizard/ModelDeploymentWizard.tsx
```diff
@@ -62,12 +62,22 @@ const ModelDeploymentWizard: React.FC<ModelDeploymentWizardProps> = ({
   // External data state - loaded by ExternalDataLoader component
   const [externalData, setExternalData] = React.useState<ExternalDataMap>({});
 
+  // Extract dynamic field data from the deployment, so that it still receives initial values
+  const { extractedFieldData } = useWizardFieldExtractors(existingDeployment);
+  const mergedExistingData = React.useMemo(
+    (): InitialWizardFormData | undefined =>
+      existingData ? { ...extractedFieldData, ...existingData } : existingData,
+    [existingData, extractedFieldData],
+  );
```
emilys314: I don't think this should be needed? This should already be extracting from `packages/model-serving/src/components/deploymentWizard/useExtractFormDataFromDeployment.ts` which is used in the "edit" kebab items before navigating to the wizard

emilys314 (commented):
I need to test it next

packages/model-serving/extensions/odh.ts
```diff
@@ -17,6 +22,79 @@ const createRedirectComponent = (args: { from: string; to: string }) => () =>
...
+    platform: 'llmd-serving',
+    extract: () =>
+      import(
+        '../src/components/deploymentWizard/fields/modelCapabilities/modelCapabilitiesApplyExtract'
+      ).then((m) => m.extractModelCapabilities),
+  },
+  flags: {
+    required: [SupportedArea.MODEL_CAPABILITIES],
+  },
+};
```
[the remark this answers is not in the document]
emilys314: The goal is this field is for all platforms, I provided a suggestion to scope the platform to an `"all"` value

packages/model-serving/src/components/deploymentWizard/fields/modelCapabilities/ModelCapabilitiesField.tsx
```diff
@@ -0,0 +1,228 @@
...
+  WELL_KNOWN_MODEL_CAPABILITIES,
+  getModelCapabilityLabelColor,
+  type ModelCapability,
+} from '../../../../shared/modelCapabilities';
+
+export type ModelCapabilitiesFieldData = ModelCapability[];
+
+export const MODEL_CAPABILITIES_FIELD_ID = 'modelCapabilities';
+
+const modelCapabilitiesFieldSchema = z.array(z.string());
```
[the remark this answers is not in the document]
emilys314: 256 does seem like a reasonable limit, but also users should never be close enough to hitting 256KB. 

I think we can skip this as we tend to avoid arbitrary limits

packages/model-serving/src/components/deploymentWizard/fields/modelCapabilities/ModelCapabilitiesField.tsx
```diff
@@ -0,0 +1,228 @@
...
+
+  const availableWellKnown = WELL_KNOWN_MODEL_CAPABILITIES.filter(
+    (cap) => !selectedCapabilities.includes(cap),
+  );
+
+  const handleAddWellKnown = (capability: string) => {
+    onChange([...selectedCapabilities, capability]);
+  };
+
+  const handleAddCustom = () => {
```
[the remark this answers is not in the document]
emilys314: Let's skip this one too, let's not add arbitrary limits

emilys314:
> When we switch off feature flag and then switch on feature flag on the "Advanced settings" tab - the model capabilities are empty when it shouldn't be. Is this expected behavior ? Also, should the annotation be populated even when the feature flag is off ?
> Screen.Recording.2026-08-05.at.6.06.10.PM.mov

Ah I see that's probably why this change was added https://github.com/opendatahub-io/odh-dashboard/pull/9067#discussion_r3721460821

@manaswinidas I don't think this is a valid use case to turn on the feature flags in the middle of the wizard. (Well I guess this is theoretically possible if an admin changes their odhDashboardConfig mid deployment though :/ )

Either way this would be a tricky problem to solve. Because when editing a deployment, all of the information is extracted when you enter the wizard when editing. This initial data is frozen when your inside the wizard. So to update the initial data (the extracted anntotations, you'd have to go back before navigating into the wizard, which isn't possible)

I think this is out of the scope of this PR to change that

manaswinidas:
@emilys314 Makes sense - this should not block merge. Just curious, Is this the case with all fields coming from annotations ?

emilys314:
> @emilys314 Makes sense - this should not block merge. Just curious, Is this the case with all fields coming from annotations ?

Yep that would happen with all fields

packages/model-serving/src/components/deploymentWizard/fields/modelCapabilities/ModelCapabilitiesField.tsx
```diff
@@ -0,0 +1,228 @@
...
+                        variant="secondary"
+                        isDisabled={!customInput.trim()}
+                        onClick={(e) => {
+                          e.stopPropagation();
+                          handleAddCustom();
+                        }}
+                        data-testid="add-custom-capability-btn"
+                      >
+                        Add
+                      </Button>
```
emilys314, line 219: If you're using keyboard navigation, there is no way to get into these elements to add a custom item, the keyboard focus is stuck on the entire `DropdownItem` 

Image

packages/model-serving/src/components/deploymentWizard/fields/modelCapabilities/ModelCapabilitiesField.tsx
```diff
@@ -0,0 +1,228 @@
...
+    (cap) => !selectedCapabilities.includes(cap),
+  );
+
+  const handleAddWellKnown = (capability: string) => {
+    onChange([...selectedCapabilities, capability]);
+  };
+
+  const handleAddCustom = () => {
+    const trimmed = customInput.trim();
+    if (trimmed && !selectedCapabilities.includes(trimmed)) {
```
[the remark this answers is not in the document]
emilys314: Yeah this wouldn't hurt to tell the user the item already exists if it's an easy change

manaswinidas:
Please feel to approve/review ignoring my comments then @emilys314 😅

packages/model-serving/src/components/deploymentWizard/fields/modelCapabilities/ModelCapabilitiesField.tsx
```diff
@@ -0,0 +1,228 @@
...
+      ],
+    },
+  ];
+};
+
+export const ModelCapabilitiesFieldWizardField: ModelCapabilitiesFieldType = {
+  id: MODEL_CAPABILITIES_FIELD_ID,
+  step: 'advancedOptions',
+  type: 'addition',
+  isActive: () => true,
```
emilys314: Actually the Ai asset endpoint field only shows when it's Generative https://github.com/opendatahub-io/odh-dashboard/blob/main/packages/model-serving/src/components/deploymentWizard/fields/ModelAvailabilityFields.tsx#L73

We want to only show this field for generative models too, I confirmed this with Jon https://redhat-internal.slack.com/archives/C069KSM8T9N/p1786030050279789?thread_ts=1785999256.722309&cid=C069KSM8T9N

emilys314 (commented):
Everything works good though

packages/model-serving/src/components/deploymentWizard/fields/modelCapabilities/ModelCapabilitiesField.tsx
ppadti: hmm, yeah - maybe because this textbox is inside dropdownItem - and only arrow key works for dropdown items. 
So, can I use MenuFooter here for this text input and button? Not sure, whether that is the best fit here

packages/model-serving/src/shared/modelCapabilities.ts
```diff
@@ -14,3 +14,32 @@ export type WellKnownModelCapability = (typeof WELL_KNOWN_MODEL_CAPABILITIES)[nu
...
+  Transcription: 'orange',
+};
+
+export const getModelCapabilityLabelColor = (capability: string): ModelCapabilityLabelColor => {
+  const wellKnown = WELL_KNOWN_MODEL_CAPABILITIES.find((c) => c === capability);
+  if (wellKnown) {
+    return WELL_KNOWN_CAPABILITY_COLORS[wellKnown];
+  }
+  return 'grey';
+};
```
Philip-Carneiro, line 38: `getModelCapabilityLabelColor` uses exact match (`c === capability`) while the rest of this module uses case-insensitive comparison. If a capability is extracted from a deployment annotation with non-canonical casing (e.g., `"vision"` instead of `"Vision"`), it will get grey color instead of blue.

The `resolveWellKnownModelCapability` function in this same file already does case-insensitive lookup and could be reused here.

**Suggested fix:**
```typescript
export const getModelCapabilityLabelColor = (capability: string): ModelCapabilityLabelColor => {
  const wellKnown = resolveWellKnownModelCapability(capability);
  if (wellKnown) {
    return WELL_KNOWN_CAPABILITY_COLORS[wellKnown];
  }
  return 'grey';
};
```

packages/model-serving/src/shared/__tests__/modelCapabilities.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+    it('should use the expected annotation key', () => {
+      expect(MODEL_CAPABILITIES_ANNOTATION).toBe('opendatahub.io/model-capabilities');
+    });
+  });
+
+  describe('WELL_KNOWN_MODEL_CAPABILITIES', () => {
+    it('should include Vision and Transcription', () => {
+      expect(WELL_KNOWN_MODEL_CAPABILITIES).toEqual(['Vision', 'Transcription']);
+    });
+  });
```
Philip-Carneiro: These two tests assert that hardcoded constants equal their hardcoded values. They are tautological -- they cannot catch any real regression. If someone changes `MODEL_CAPABILITIES_ANNOTATION`, they would also update the test expectation. The constant has no conditional logic, no computation, and no derivation.

packages/model-serving/src/components/deploymentWizard/fields/modelCapabilities/ModelCapabilitiesField.tsx
```diff
@@ -0,0 +1,278 @@
...
+      return;
+    }
+    setCustomInputError('');
+    onChange([...selectedCapabilities, capability]);
+    setCustomInput('');
+  };
+
+  const handleRemove = (capability: string) => {
+    onChange(selectedCapabilities.filter((c) => c !== capability));
+  };
```
Philip-Carneiro, line 93: `handleRemove` uses exact string comparison (`c !== capability`) while `handleAddCustom` uses `includesModelCapability` (case-insensitive). This is inconsistent within the same component.

In practice this works because `capability` comes from the label click (which passes the exact stored string), so the exact match will succeed. However, if a future caller passes a differently-cased string, the removal would silently fail. Using the same case-insensitive utility keeps the component internally consistent.

**Suggested fix:**
```typescript
const handleRemove = (capability: string) => {
  onChange(selectedCapabilities.filter((c) => !isSameModelCapability(c, capability)));
};
```

Philip-Carneiro (changes requested):

packages/model-serving/src/components/deploymentWizard/fields/modelCapabilities/ModelCapabilitiesField.tsx
emilys314: I'm not sure what the fix is 
 
Also small correction, it is possible to get to it by hitting tab, but it comes up after the rest of the form is tabbed through
image

Still not sure the best way to fix this though

emilys314 (commented):
Overall looks good to merge except the one accessibility issue, I don't want you to have to spend a lot of time debugging that though. So I can approve if that is too much to fix

Image

packages/model-serving/src/components/deploymentWizard/fields/modelCapabilities/ModelCapabilitiesField.tsx
ppadti: Fixed the ordering part of A11y issue

emilys314 (approved):
Looks great! thanks!

Image

PR #9078, opened 2026-08-05
test(operator): add Perses federation ConfigMap test for observability projection (RHOAIENG-87523)

lucferbux:
https://redhat.atlassian.net/browse/RHOAIENG-87523
Follow-up to [RHOAIENG-80354](https://redhat.atlassian.net/browse/RHOAIENG-80354)

## Description

On RHOAI deployments, the Perses observability proxy is never configured because nothing sets `spec.observability` on the Dashboard CR. The dashboard-operator relies on this field to include the Perses proxy entry in the federation ConfigMap, but the ODH Operator's Dashboard handler does not project it.

This PR adds a test verifying that `buildFederationConfigMap` correctly includes the Perses proxy entry when `spec.observability` is configured on the Dashboard CR. The observability configuration is projected by the ODH Operator handler via `BuildModuleCR`, following the modularization best practice: module operators read only from their own CR spec, not from DSCI directly.

**Companion PR:** opendatahub-io/opendatahub-operator#3923 — adds the observability projection logic to the ODH Operator's Dashboard handler.

## How Has This Been Tested?

- `make test` in `dashboard-operator/` passes
- New test `TestBuildFederationConfigMap_IncludesPersesWhenObsEnabled` verifies the Perses federation entry is present when observability is configured

## Test Impact

Added `TestBuildFederationConfigMap_IncludesPersesWhenObsEnabled` in `module_deploy_test.go` — validates that the federation ConfigMap includes the Perses proxy entry when `Dashboard.Spec.Observability` is enabled with a `PersesService` target.

## Request review criteria:

[RHOAIENG-80354]: https://redhat.atlassian.net/browse/RHOAIENG-80354?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dashboard-operator/internal/controller/module_deploy_test.go
+++ b/dashboard-operator/internal/controller/module_deploy_test.go
@@ -362,3 +362,56 @@ func TestPatchDeploymentFederationHash_DeploymentNotFound(t *testing.T) {
 	err := r.PatchDeploymentFederationHash(context.Background(), `[{"name":"genAi"}]`)
 	require.NoError(t, err, "NotFound should be a no-op, not an error")
 }
+
+func TestBuildFederationConfigMap_IncludesPersesWhenObsEnabled(t *testing.T) {
+	s := testScheme(t)
+	cli := fake.NewClientBuilder().WithScheme(s).Build()
+
+	r := &ctrlpkg.DashboardReconciler{
+		Client:                cli,
+		Scheme:                s,
+		Platform:              cluster.SelfManagedRhoai,
+		Namespace:             testNamespace,
+		ApplicationsNamespace: testNamespace,
+	}
+
+	statuses := allDeployedStatuses()
+	dashboard := &v1alpha1.Dashboard{
+		Spec: v1alpha1.DashboardSpec{
+			Observability: &v1alpha1.ObservabilitySpec{
+				Enabled: true,
+				PersesService: &v1alpha1.ServiceTarget{
+					Name:      "data-science-perses",
+					Namespace: "redhat-ods-monitoring",
+					Port:      8080,
+				},
+			},
+		},
+	}
+
+	cm, err := ctrlpkg.BuildFederationConfigMap(r, statuses, dashboard)
+	require.NoError(t, err)
+
+	data := cm.Data["module-federation-config.json"]
+	var entries []map[string]interface{}
+	require.NoError(t, json.Unmarshal([]byte(data), &entries))
+
+	var persesEntry map[string]interface{}
+	for _, entry := range entries {
+		if name, ok := entry["name"].(string); ok && name == "perses" {
+			persesEntry = entry
+		}
+	}
+
+	require.NotNil(t, persesEntry, "perses entry must be present when observability is enabled")
+
+	proxyServices, ok := persesEntry["proxyService"].([]interface{})
+	require.True(t, ok)
+	require.Len(t, proxyServices, 1)
+
+	ps := proxyServices[0].(map[string]interface{})
+	assert.Equal(t, "/perses/api", ps["path"])
+	svc := ps["service"].(map[string]interface{})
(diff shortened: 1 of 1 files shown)


dashboard-operator/internal/controller/module_deploy_test.go
```diff
@@ -362,3 +362,56 @@ func TestPatchDeploymentFederationHash_DeploymentNotFound(t *testing.T) {
 	err := r.PatchDeploymentFederationHash(context.Background(), `[{"name":"genAi"}]`)
 	require.NoError(t, err, "NotFound should be a no-op, not an error")
 }
+
+func TestBuildFederationConfigMap_IncludesPersesWhenObsEnabled(t *testing.T) {
```
liday-rh, line 366: **Largely redundant with `TestBuildFederationConfigMap_NamespaceValues` (line 135).**

The existing `NamespaceValues` test already sets up the same `Observability` fixture, asserts `perses` entry is present, asserts `proxyService` is non-empty, and checks the service namespace. The only net-new assertions here are the `/perses/api` path check and the service name check — those could be two extra `assert` lines in the existing test rather than a 53-line new function.

Not blocking, but worth considering consolidation to avoid maintaining two near-identical test setups.

dashboard-operator/internal/controller/module_deploy_test.go
```diff
@@ -362,3 +362,56 @@ func TestPatchDeploymentFederationHash_DeploymentNotFound(t *testing.T) {
...
+		}
+	}
+
+	require.NotNil(t, persesEntry, "perses entry must be present when observability is enabled")
+
+	proxyServices, ok := persesEntry["proxyService"].([]interface{})
+	require.True(t, ok)
+	require.Len(t, proxyServices, 1)
+
+	ps := proxyServices[0].(map[string]interface{})
```
liday-rh, line 412: **Bare type assertions will panic instead of producing a test failure.**

Lines 412 and 414 use bare type assertions (`proxyServices[0].(map[string]interface{})`) which panic on type mismatch rather than reporting a clean test failure. The existing `NamespaceValues` test at line 184-185 uses the safer pattern:

```go
ps, _ := proxyServices[0].(map[string]interface{})
svc, _ := ps["service"].(map[string]interface{})
```

combined with subsequent `require`/`assert` calls. Suggest matching that pattern for consistency and better failure diagnostics.

dashboard-operator/internal/controller/module_deploy_test.go
```diff
@@ -362,3 +362,56 @@ func TestPatchDeploymentFederationHash_DeploymentNotFound(t *testing.T) {
...
+	assert.Equal(t, "/perses/api", ps["path"])
+	svc := ps["service"].(map[string]interface{})
+	assert.Equal(t, "data-science-perses", svc["name"])
+	assert.Equal(t, "redhat-ods-monitoring", svc["namespace"])
```
liday-rh, line 416: **Missing port assertion.**

The fixture sets `Port: 8080` and the production code at `module_deploy.go:436` wires it through as `Port: ps.Port`, but the test never verifies it. Since `json.Unmarshal` into `map[string]interface{}` decodes numbers as `float64`:

```go
assert.Equal(t, float64(8080), svc["port"])
```

liday-rh (commented):
Overall this is a correct test — its assertions match the production code. A few suggestions to strengthen it and reduce overlap with existing coverage.

**Missing negative case:** Neither this test nor any existing test explicitly verifies that perses is *excluded* when `Observability` is nil or `Enabled: false`. `TestBuildFederationConfigMap_ExcludesDisabledModules` passes an empty `&v1alpha1.Dashboard{}` but doesn't assert `perses` is absent. Adding `assert.False(t, names["perses"])` there would be a more valuable addition than the positive-only coverage here.

PR #9079, opened 2026-08-05
fix(operator): read distribution identity from orchestrator annotations

lucferbux:
https://issues.redhat.com/browse/RHOAIENG-80304

## Description

After migrating Dashboard lifecycle management to the new dashboard-operator (Module Controller), the `odh-dashboard-config` ConfigMap retains Helm chart defaults for `distribution.name` ("Standalone") and `distribution.version` ("0.0.0"). The ODH Operator correctly stamps the ConfigMap with annotations (`platform.opendatahub.io/type` and `platform.opendatahub.io/version`), but `readDistributionConfig()` only reads data keys, ignoring the annotations. This results in the Dashboard CR `status.distribution` reporting wrong values (`{name: "Standalone", version: "0.0.0"}` instead of `{name: "OpenShift AI Self-Managed", version: "3.5.0"}`).

**Fix:** Modify `readDistributionConfig()` to prefer orchestrator annotations over Helm chart data keys. The annotation constants (`annotations.PlatformType`, `annotations.PlatformVersion`) are already available in the `odh-platform-utilities` dependency. When present and non-empty, annotations take precedence; when absent, data keys are used for backward compatibility.

**Cluster evidence (before fix):**
```yaml
# ConfigMap data (Helm defaults — wrong)
distribution.name: Standalone
distribution.version: 0.0.0

# ConfigMap annotations (set by ODH Operator — correct)
platform.opendatahub.io/type: OpenShift AI Self-Managed
platform.opendatahub.io/version: 3.5.0

# Dashboard CR status.distribution (reads from data keys — wrong)
distribution:
  name: Standalone
  version: 0.0.0
```

Also removed the `cm.Data == nil` early return, since annotations may provide distribution identity even when data keys are absent.

## How Has This Been Tested?

- Added 10 table-driven unit tests in `TestReadDistributionConfig` covering all scenarios (annotation override, partial override, fallback to data keys, nil data with annotations, empty annotations, truncation)
- All existing tests continue to pass (`api/v1alpha1`, `charts`, `internal/controller`, `internal/webhook`)
- Lint passes with 0 issues

## Test Impact

Added `TestReadDistributionConfig` in `dashboard-operator/internal/controller/config_test.go` with 10 test cases:
- ConfigMap absent → nil
- Data keys only → backward compat
- Annotations only → annotation values used
- Annotations override data keys (the core bug scenario)
- Partial overrides (name only, version only)
- Empty annotations don't override
- Nil data map with annotations
- Both empty → nil
- Oversized annotation truncated to 256 runes

## Request review criteria:

### DIFF
--- a/dashboard-operator/internal/controller/config.go
+++ b/dashboard-operator/internal/controller/config.go
@@ -13,6 +13,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/log"
 
 	v1alpha1 "github.com/opendatahub-io/odh-dashboard/dashboard-operator/api/v1alpha1"
+	"github.com/opendatahub-io/odh-platform-utilities/pkg/metadata/annotations"
 )
 
 const operatorConfigMapName = "dashboard-operator-config"
@@ -63,9 +64,11 @@ func readOperatorConfig(ctx context.Context, cli client.Client, namespace string
 }
 
 // readDistributionConfig reads the distribution identity from the
-// chart-deployed ConfigMap. Returns (nil, nil) when the ConfigMap is
-// absent or contains no distribution keys, and (nil, err) on transient
-// read failures so the caller can preserve last-known-good status.
+// chart-deployed ConfigMap. Orchestrator annotations (PlatformType,
+// PlatformVersion) take precedence over Helm chart data keys for backward
+// compatibility. Returns (nil, nil) when the ConfigMap is absent or
+// contains no distribution identity, and (nil, err) on transient read
+// failures so the caller can preserve last-known-good status.
 func readDistributionConfig(ctx context.Context, cli client.Client, namespace string) (*v1alpha1.Distribution, error) {
 	logger := log.FromContext(ctx)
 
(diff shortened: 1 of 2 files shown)


caponetto:
@lucferbux check if this makes sense:

The patch only changes how readDistributionConfig() populates dashboard.Status.Distribution from ConfigMap annotations, but the About dialog in this repository renders its version from dsciStatus.release.version fetched via /api/dsci/status, a path this diff does not touch. Because no code here maps the updated status.distribution value into the API the frontend actually calls, the reported About dialog bug (showing 'unknown' or a stale version) can remain reachable in production unless an unreviewed companion change elsewhere in the repo or in another repo bridges the two.

   **File:** `dashboard-operator/internal/controller/config.go:89`

   **Evidence:**
   ```go
   // Orchestrator annotations take precedence over Helm chart data keys.
   if annoName := cm.Annotations[annotations.PlatformType]; annoName != "" {
   	name = annoName
   }
   if annoVersion := cm.Annotations[annotations.PlatformVersion]; annoVersion != "" {
   	version = annoVersion
   }
   ...
   return &v1alpha1.Distribution{
   	Name:    name,
   	Version: version,
   }, nil
   
   ---
   const url = '/api/dsci/status';
   return result.status;
   
   {dsciStatus?.release?.version || 'Unknown'}
   ```

   **Suggestion:** Verify end-to-end which data source the About dialog actually consumes in production. If it is /api/dsci/status rather than the Dashboard CR's status.distribution, either wire status.distribution into that API response, update the ConfigMap's distribution.version data key directly, or document and link the companion change (in this repo or the ODH Operator repo) that closes this gap before merging.

lucferbux:
@caponetto Ok, so you are right, and idk how it happened cause I was able to validate it in a cluster, maybe doing some of the tweaks I inadvertedly fixed the dsci hence fixing the issue at the same time. The About dialog currently reads from `dsciStatus.release.version` (via `/api/dsci/status`), not from the Dashboard CR's `status.distribution`.  At the same time the operator was filling release 0.0.0.0 which is what we've fixed in this PR, I'm creating https://redhat.atlassian.net/browse/RHOAIENG-81293 to track all the dsci reads in our dashboard cc @andrewballantyne @christianvogt which we'll need to eventually replace since the right pattern would be read the Dashboard CR now

caponetto:
Thanks for the info @lucferbux!

/lgtm
/approve

PR #9100, opened 2026-08-05
RHOAIENG-81207: Stabilize flaky cluster storage delete and workbench variables e2e tests

Taj010:
For https://redhat.atlassian.net/browse/RHOAIENG-81207

## Description

Stabilizes two flaky Cypress e2e specs that failed in CI:

1. `testClusterStorageCreation.cy.ts` — Create / Edit / Delete PVC

**Why it was flaky**  
After clicking Delete, the test immediately asserted `empty-state-title` without waiting for the delete modal to finish and close. Delete is async (API + UI refresh). Under load, Cypress checked the empty state while the PVC row was still visible, so the assertion timed out. Submit paths also used `{ force: true }`, which skipped normal enabled/actionability checks and made race timing worse.

**How it was fixed**
- Removed `{ force: true }` from create/update submit clicks; wait for `.should('be.enabled')` first
- After create, update, and delete: wait for the modal to close (`shouldBeOpen(false)`) before the next assertion
- Only then assert the storage row / empty state

2. `testWorkbenchVariables.cy.ts` — Env vars via Secret / ConfigMap upload & key-value

**Why it was flaky**  
The test waited up to 120s for notebook status `Ready`. On resource-constrained CI clusters the workbench often landed on `Failed` (e.g. “Insufficient resources to start”) and never recovered, so the assertion failed even though the test logic was fine.

**How it was fixed**  
Added a shared `waitForNotebookReady` helper used by all three tests in the spec:
- Wait until status is `Ready` **or** `Failed`
- On `Failed`, click Start once and wait again
- Fail only if it still isn’t `Ready` after that retry

## How Has This Been Tested?

- Lint/prettier on the touched Cypress files
- Local Cypress e2e against the dash-e2e RHOAI cluster with `CY_TEST_CONFIG` pointing at `packages/cypress/test-variables.yml`
- Re-ran `testClusterStorageCreation.cy.ts` and `testWorkbenchVariables.cy.ts` via Cypress GUI

## Test Impact

- Updated existing e2e specs only (no new product UI)
- No unit tests

## Request review criteria:

## Summary by CodeRabbit

* **Tests**
  * Improved automated validation of cluster storage creation, updates, and deletion by confirming dialogs close before checking results.
  * Enhanced workbench readiness checks to recover from initial startup failures and retry once before reporting failure.
  * Replaced forced form submissions with more reliable workflow interactions.

### DIFF
--- a/packages/cypress/cypress/tests/e2e/dataScienceProjects/workbenches/testWorkbenchVariables.cy.ts
+++ b/packages/cypress/cypress/tests/e2e/dataScienceProjects/workbenches/testWorkbenchVariables.cy.ts
@@ -7,7 +7,11 @@ import {
 import type { WBVariablesTestData } from '../../../../types';
 import { NotebookStatusLabel } from '../../../../types';
 import { projectDetails, projectListPage } from '../../../../pages/projects';
-import { workbenchPage, createSpawnerPage } from '../../../../pages/workbench';
+import {
+  workbenchPage,
+  createSpawnerPage,
+  notebookConfirmModal,
+} from '../../../../pages/workbench';
 import { HTPASSWD_CLUSTER_ADMIN_USER } from '../../../../utils/e2eUsers';
 import { loadWBVariablesFixture } from '../../../../utils/dataLoader';
 import { createCleanProject } from '../../../../utils/projectChecker';
@@ -25,7 +29,10 @@ type NotebookRow = ReturnType<typeof workbenchPage.getNotebookRow>;
 
 /**
  * Wait for a workbench to become Ready. If it lands on Failed (common under
- * transient cluster resource pressure), start it once more and wait again.
+ * transient cluster resource pressure), stop it cleanly then start again.
+ *
+ * Failed startups still have isStarting=true (no stop annotation, pods not ready),
+ * so the state toggle invokes Stop — not Start. We must stop → Stopped → Start.
  */
 const waitForNotebookReady = (row: NotebookRow, workbenchName: string, maxRetries = 1): void => {
   row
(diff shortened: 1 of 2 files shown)


packages/cypress/cypress/tests/e2e/dataScienceProjects/workbenches/testWorkbenchVariables.cy.ts
```diff
@@ -21,6 +25,38 @@ import {
...
+      expect(
+        [NotebookStatusLabel.Ready, NotebookStatusLabel.Failed],
+        `Unexpected status for workbench ${workbenchName}`,
+      ).to.include($el.text());
+    })
+    .then(($el) => {
+      if ($el.text() === NotebookStatusLabel.Failed && maxRetries > 0) {
+        cy.log(`Notebook ${workbenchName} failed to start; retrying start (${maxRetries} left)`);
+        row.findNotebookStopToggle().click();
+        notebookConfirmModal.findStopWorkbenchButton().click();
```
danreed-rh, line 50: **Suggestion (non-blocking):** Before calling stop / confirm, re-query the row status (or assert the toggle still reads **Stop**) instead of acting only on the Failed text captured in `.then(()`.

Under load the label can change between the `Ready|Failed` assertion and these clicks. Re-checking avoids stopping a workbench that recovered to Ready, or clicking confirm when the toggle is already **Start**.

packages/cypress/cypress/tests/e2e/dataScienceProjects/workbenches/testWorkbenchVariables.cy.ts
```diff
@@ -21,6 +25,38 @@ import {
...
+        row.expectStatusLabelToBe(NotebookStatusLabel.Stopped, 120000);
```
danreed-rh, line 51: **Suggestion (non-blocking):** After stop, prefer waiting for the state toggle to show **Start** (and be enabled) before the next click — instead of (or in addition to) `expectStatusLabelToBe(Stopped)`.

`NotebookStatusLabel` prioritizes ERROR → **Failed** over Stopped. If error events linger after stop, waiting only on the Stopped label can time out even though the toggle is ready to start. Sibling stop/start flows often use the Stopped label; this Failed→retry path is the case where toggle-based wait is more robust.

danreed-rh (commented):
Non-blocking hardening notes on `waitForNotebookReady` (Failed retry path). Storage modal waits look good; CodeRabbit stop→Stopped→Start fix on this head is correct.

packages/cypress/cypress/tests/e2e/dataScienceProjects/workbenches/testWorkbenchVariables.cy.ts
danreed-rh: I changed my mind on this one, not needed

PR #9103, opened 2026-08-05
feat(RHOAIENG-68987): LLM accelerator configurations tab migration

mturley:
https://issues.redhat.com/browse/RHOAIENG-68987

## Description

Migrates the LLM accelerator configurations page from a standalone admin page to a tab on the "Model deployment settings" tabbed page. This is a migration of existing functionality, not new feature work, and it is the first real tab on that page — the other four are still placeholder stubs — so it establishes the pattern the topology and routing migrations ([RHOAIENG-80016], [RHOAIENG-80020]) will follow. Both of those tickets have been updated to point here.

Builds on [#9029] (merged), which added the page shell, the `modelDeploymentSettings` dev feature flag, the placeholder tabs, the `disallowed` flags on the standalone pages, and the redirect routes.

### Screenshots

image

image

image

image

### Tab content vs. full-page forms

`TabRoutePage` renders tab content inside a `role="tabpanel"` beneath the page title and tab bar. Putting the whole route tree in the tab would therefore render the add/edit/duplicate forms — which supply their own `ApplicationsPage` title and breadcrumb — with two `<h1>` elements and a tab bar above them. Instead:

- the tab holds only the list (`LlmAcceleratorConfigTabRoutes`)
- the forms are registered as separate `app.route` breakout extensions (`LlmAcceleratorConfigFormRoutes`), gated on exactly the same flags as the tab
- the list view passes `noTitle` to `ApplicationsPage`, matching the convention already used by the model registry, model catalog, and global deployments list views

This mirrors how `agent-ops` and `model-serving` keep their detail pages and wizards out of tab chrome.

### Shared RBAC gate

The create/patch SSAR on `LLMInferenceServiceConfig` is extracted to `LlmInferenceServiceConfigAccessGate`, named for the resource rather than the feature: the topology and routing configuration pages are backed by the same CRD and will reuse it. Worth flagging for those tickets — **those two pages currently have no RBAC gate at all**; `useAccessAllowed` / `verbModelAccess` appear nowhere else in `packages/llmd-serving/src`. That is a pre-existing gap, not something introduced here, and adding the gate there will be a behaviour change rather than a refactor.

### Empty-state fix

`ApplicationsPage` swaps children for the empty state, which hid the list toolbar and with it the only "Add LLM accelerator configuration" button — so an administrator with zero configurations could not create the first one. The empty state now carries the action itself, using the same `data-testid` so the page object works in both states. This was pre-existing behaviour on the standalone page, fixed here rather than carried forward.

### Path constants

`extensions.ts` may not import runtime values from `src` (`no-restricted-syntax`), so the route paths are duplicated into it exactly as `agent-ops` does. `extensions/__tests__/extensions.spec.ts` pins them against `paths.ts` so the two cannot drift; that guard was verified by changing `paths.ts` alone and confirming it fails.

## How Has This Been Tested?

Automated, locally:
- `tsc --noEmit` on `packages/llmd-serving` — clean
- `eslint` on all changed files in `llmd-serving` and `cypress` — clean
- `jest` on `packages/llmd-serving` — 28 suites, 292 tests passing
- Cypress mock tests — 18/18 passing across both accelerator config specs
- Playwright browser testing against a checklist covering both feature flag states, the breakout form chrome, redirects from the old URLs, and the empty state.
- Manual verification

To verify manually, use `?devFeatureFlags=modelDeploymentSettings=true,vLLMDeploymentOnMaaS=true`:
- `modelDeploymentSettings` flag off (the default): standalone page and nav item behave exactly as before, including the page title
- `modelDeploymentSettings` flag on: the tab renders the list; Add/Edit/Duplicate open as full-page forms with a breadcrumb and **no** tab bar; the breadcrumb returns to the tab; old standalone URLs redirect to the tab
- In both cases, `vLLMDeploymentOnMaaS` flag off: the page / tab / route does not render

## Test Impact

Added:
- `LlmAcceleratorConfigView.spec.tsx` — list vs. empty state, the empty-state add action, and `noTitle` behaviour in tab vs. standalone mode
- `LlmAcceleratorConfigFormRoutes.spec.tsx` — mounts the component the way the host mounts an `app.route` extension, i.e. a single top-level route at the exact registered path, and asserts each form renders in the right mode. This is the condition the other route tests do not reproduce, since they exercise the standalone tree, which is registered with `/*`.
- `extensions/__tests__/extensions.spec.ts` — the tab registration, the form routes being registered as breakout routes outside the tab, their flags matching the tab's, and the duplicated path constants matching `paths.ts`

Updated:
- `LlmAcceleratorConfigAddForm.spec.tsx` — passes an explicit `listPath`; the fixture value is deliberately not one of the real paths, so the tests fail if the form ever navigates to a hardcoded path instead of the prop
- Cypress mock tests — tab-based URLs, `modelDeploymentSettings` enabled, and an assertion that the add form renders with neither the tabbed page title nor the tab bar

No E2E tests exist for this page; per the epic those are tracked separately in [RHOAIENG-69029].

## Request review criteria:

If you have UI changes:

[#9029]: https://github.com/opendatahub-io/odh-dashboard/pull/9029
[RHOAIENG-80016]: https://issues.redhat.com/browse/RHOAIENG-80016
[RHOAIENG-80020]: https://issues.redhat.com/browse/RHOAIENG-80020
[RHOAIENG-69029]: https://issues.redhat.com/browse/RHOAIENG-69029

## Summary by CodeRabbit

- **New Features**
  - Added a tabbed Model deployment settings experience for LLM accelerator configurations.
  - Added dedicated add, edit, and duplicate configuration pages.
  - Added an empty state with a direct action to create the first configuration.
  - Added permission-based access handling for configuration management.

- **Bug Fixes**
  - Improved navigation and redirects across configuration pages.
  - Restricted configuration access when required permissions are unavailable.

- **Tests**
  - Expanded coverage for tab navigation, forms, permissions, empty states, and feature-flagged behavior.


packages/llmd-serving/src/settings/llmAcceleratorConfigs/LlmAcceleratorConfigFormRoutes.tsx
```diff
@@ -0,0 +1,57 @@
...
+ * content so the forms render with their own breadcrumb and title instead of
+ * inside the tab panel, beneath the page title and tab bar. The route paths are
+ * absolute here because each form path is registered as a separate top-level
+ * route, matching the pattern used by the agent-ops tabbed page.
+ */
+const LlmAcceleratorConfigFormRoutes: React.FC = () => (
+  <LlmInferenceServiceConfigAccessGate>
+    <Routes>
+      <Route
+        path={LLM_ACCELERATOR_CONFIGS_TAB_PATH}
```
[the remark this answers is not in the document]
mturley: Addressed

packages/llmd-serving/extensions/extensions.ts
```diff
@@ -683,10 +692,38 @@ const extensions: (
...
       id: 'llm-accelerator-configurations',
       title: 'LLM accelerator configurations',
-      component: () => import('../src/settings/LlmAcceleratorConfigsTab'),
+      component: () =>
+        import('../src/settings/llmAcceleratorConfigs/LlmAcceleratorConfigTabRoutes'),
       group: '3_accelerator',
     },
   },
+  // Full-page breakout routes for the accelerator configuration forms. Registered
+  // separately from the tab so the forms render without the tabbed page chrome.
```
emilys314, line 701: The "chrome" part in this comment is a little confusing

packages/llmd-serving/src/settings/LlmInferenceServiceConfigAccessGate.tsx
```diff
@@ -0,0 +1,38 @@
...
+ * The accelerator, topology, and routing configuration pages are all backed by this
+ * same resource — they differ only by label selector — so they share this gate.
+ */
+const LlmInferenceServiceConfigAccessGate: React.FC<React.PropsWithChildren> = ({ children }) => {
+  const [canCreate, createLoaded] = useAccessAllowed(
+    verbModelAccess('create', LLMInferenceServiceConfigModel),
+  );
+  const [canPatch, patchLoaded] = useAccessAllowed(
+    verbModelAccess('patch', LLMInferenceServiceConfigModel),
+  );
```
emilys314, line 21: The table itself only requires LIST / GET, I feel like we should gate on those. Then inside the page gate the Create / Edit buttons on can create / patch. Although actually everybody has LIST / GET on the global namespace. 

Hmm, either way it's outside the scope of this PR

packages/llmd-serving/src/settings/llmAcceleratorConfigs/LlmAcceleratorConfigEmptyState.tsx
emilys314, line 1: I didn't realize the empty state was missing before 🫪

Image

emilys314 (approved):
Looks good it all works. 

I do have a dream that when we hover over the tabs, it starts to preload the required resources before navigating. That seems even harder to do though with the tabbing extension

packages/llmd-serving/extensions/extensions.ts
mturley: Yeah, I guess it’s a strange word choice but in this context it just means UI container/wrapper

packages/llmd-serving/src/settings/LlmInferenceServiceConfigAccessGate.tsx
mturley: Yeah, I think when we created these pages we decided it’s useless for you to see them if you don’t have write permissions on them (either you can use the full page functionality or you can’t), but it probably makes sense to make the gating more granular and just disable the add/edit/duplicate/delete buttons if you don’t have those perms. I can open a tech debt issue

packages/llmd-serving/src/settings/llmAcceleratorConfigs/LlmAcceleratorConfigEmptyState.tsx
mturley: Yeah…. not really an issue since we should always have OOTB ones but it felt like the time to fix it

packages/llmd-serving/src/settings/LlmInferenceServiceConfigAccessGate.tsx
mturley: Tossing this on the backlog, we can close it if it doesn’t have value https://redhat.atlassian.net/browse/RHOAIENG-81951

PR #9108, opened 2026-08-05
fix(RHOAIENG-79525): include desired handler in probe sanitization patch [v3.5.0]

lucferbux:
https://issues.redhat.com/browse/RHOAIENG-79525

## Description

Cherry-pick of #9077 to `v3.5.0-fixes`.

The probe sanitization patch was only nulling out stale handler types (e.g., `tcpSocket`, `httpGet`) without setting the desired handler (`exec`), leaving probes with no handler type — which Kubernetes rejects. The operator enters an error loop retrying every ~17 minutes.

The fix includes the desired handler value alongside the null stale handlers in the strategic merge patch, so the transition is atomic.

See #9077 for full details, root cause analysis, and cluster verification.

## How Has This Been Tested?

- Unit tests updated and passing (all 5 `TestSanitize*` tests pass)
- Verified on live 3.4→3.5 upgrade cluster — patch succeeded, pods rolled out 10/10, operator reconciled successfully

## Test Impact

Updated unit tests to verify exec handler command values after sanitization using `reflect.DeepEqual`.

## Request review criteria:

### DIFF
--- a/dashboard-operator/internal/controller/probe_sanitize.go
+++ b/dashboard-operator/internal/controller/probe_sanitize.go
@@ -89,9 +89,16 @@ func buildProbeCleanupPatch(live, desired *unstructured.Unstructured) map[string
 		for _, probeField := range []string{"livenessProbe", "readinessProbe", "startupProbe"} {
 			nullFields := conflictingHandlerFields(liveC, desiredC, probeField)
 			if len(nullFields) > 0 {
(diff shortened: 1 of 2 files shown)


lucferbux:
/approve

This is just a backport for a blocker, it can be merged automatically

manaswinidas (approved):
/lgtm

PR #9142, opened 2026-08-06
test(e2e): harden register-and-store transfer job toast asserts

manaswinidas:
## Description

Hardens the OCI register-and-store Cypress asserts in `testRegisterModel.cy.ts` after a Jenkins Smoke flake where the "Model transfer job started" toast was visible in the video but `cy.contains(..., { timeout: 15000 }).should('be.visible')` timed out.

**Changes:**
* Intercept and wait for `POST **/model_transfer_jobs*` (60s) before asserting the toast, so the assertion starts after the create API returns
* Use `toastNotifications.findToastNotificationList()` for the S3-origin case (aligned with the URI-origin case)
* Add a 30s timeout on the post-submit URL navigation assert for both OCI cases

These specs remain `@NonConcurrent` and are still skipped by GitHub Cypress CI (`CYPRESS_skipTags` includes `@NonConcurrent`). This targets Jenkins Smoke reliability.

## How Has This Been Tested?

* Reviewed Cypress video from Jenkins build confirming toast appears; failure was assertion timing
* Diff-reviewed against the existing URI-origin toast helper pattern in the same file
* Not re-run against a live cluster in this change (test-only hardening)

## Test Impact

Updates existing E2E asserts only; no product code changes. Should reduce false failures on register-and-store toast timing without changing coverage intent.

## Request review criteria:

If you have UI changes:

## Tests

* Improved end-to-end validation for OCI model registration and storage workflows.
* Added stronger verification that transfer operations start successfully and display the expected notifications.
* Improved handling of longer-running transfers and page navigation during registration.
* Expanded test coverage to confirm users are returned from the registration flow at the appropriate time.

### DIFF
--- a/packages/cypress/cypress/tests/e2e/modelRegistry/testRegisterModel.cy.ts
+++ b/packages/cypress/cypress/tests/e2e/modelRegistry/testRegisterModel.cy.ts
@@ -37,6 +37,8 @@ describe('Verify models can be registered in a model registry', () => {
   let objectStorageModelName: string;
   let ociModelName: string;
   let ociUriModelName: string;
+  let ociJobName: string;
+  let ociUriJobName: string;
   let deploymentName: string;
   let projectName: string;
   const uuid = generateTestUUID();
@@ -50,6 +52,8 @@ describe('Verify models can be registered in a model registry', () => {
       objectStorageModelName = `${testData.objectStorageModelName}-${uuid}`;
       ociModelName = `${testData.ociModelName}-${uuid}`;
       ociUriModelName = `${testData.ociUriModelName}-${uuid}`;
+      ociJobName = `${testData.ociJobName}-${uuid}`;
+      ociUriJobName = `${testData.ociUriJobName}-${uuid}`;
       deploymentName = testData.operatorDeploymentName;
 
       // ensure operator has optimal memory
@@ -298,7 +302,7 @@ describe('Verify models can be registered in a model registry', () => {
         .type(testData.ociModelFormatVersion);
 
       cy.step('Fill in transfer job name');
-      registerModelPage.findFormField(FormFieldSelector.JOB_NAME).type(testData.ociJobName);
+      registerModelPage.findFormField(FormFieldSelector.JOB_NAME).type(ociJobName);
 
       cy.step('Fill in origin S3 location and credentials');
       registerModelPage.findFormField(FormFieldSelector.LOCATION_TYPE_OBJECT_STORAGE).click();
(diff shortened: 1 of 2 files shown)


packages/cypress/cypress/tests/e2e/modelRegistry/testRegisterModel.cy.ts
```diff
@@ -341,15 +341,21 @@ describe('Verify models can be registered in a model registry', () => {
       registerModelPage.findSubmitButton().should('be.enabled');
 
       cy.step('Submit the register and store form');
+      cy.intercept('POST', '**/model_transfer_jobs*').as('createTransferJob');
```
antowaddle: We typically wouldn't use intercepts in the e2e tests, can you instead use OC to create the missing resource?

packages/cypress/cypress/tests/e2e/modelRegistry/testRegisterModel.cy.ts
```diff
@@ -341,15 +341,21 @@ describe('Verify models can be registered in a model registry', () => {
...
       registerModelPage.findSubmitButton().click();
 
+      cy.step('Wait for transfer job create API to succeed');
+      cy.wait('@createTransferJob', { timeout: 60000 })
```
antowaddle: Please avoid using cy.wait, generally we'd poll oc resources as a more robust mechanism.

antowaddle (changes requested):

packages/cypress/cypress/tests/e2e/modelRegistry/testRegisterModel.cy.ts
manaswinidas: Good call — dropped `cy.intercept` / `cy.wait`.

Instead:
- Unique Job names (`${fixture}-${uuid}`) so create does not 400 on collisions
- After submit, assert started toast (60s via toast helper)
- Poll the Job pod with existing `checkModelTransferJobPodStarted` (oc wait) for both OCI cases

UI still creates the transfer Job; OC is only the readiness check.

packages/cypress/cypress/tests/e2e/modelRegistry/testRegisterModel.cy.ts
manaswinidas: Agreed — removed `cy.wait`. Both OCI register-and-store cases now poll with `checkModelTransferJobPodStarted` (`oc wait -l job-name=... --for=condition=PodScheduled`).

packages/cypress/cypress/tests/e2e/modelRegistry/testRegisterModel.cy.ts
```diff
@@ -341,15 +341,21 @@ describe('Verify models can be registered in a model registry', () => {
...
-        'be.visible',
-      );
+      toastNotifications
+        .findToastNotificationList()
+        .should('contain.text', testData.ociTransferJobStartedNotification);
 
       cy.step('Verify navigation away from the registration form');
-      cy.url().should('include', `/ai-hub/models/registry/${registryName}`);
+      cy.url({ timeout: 30000 }).should('include', `/ai-hub/models/registry/${registryName}`);
       cy.url().should('not.include', '/register');
```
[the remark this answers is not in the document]
manaswinidas: Done — both OCI blocks now use:

```ts
cy.url({ timeout: 30000 })
  .should('include', `/ai-hub/models/registry/${registryName}`)
  .and('not.include', '/register');
```

manaswinidas:
Screenshot 2026-08-19 at 7 04 29 PM


dashboard-e2e-tests/2330

Tests passing on jenkins

antowaddle (approved):
/lgtm

PR #9114, opened 2026-08-06
RHOAIENG-81833: read About dialog version from DSC instead of DSCI

lucferbux:
https://issues.redhat.com/browse/RHOAIENG-81833

## Description

The About dialog was reading release name and version from the DSCI (`DataScienceClusterInitialization`) CR via `useFetchDsciStatus()`. After the dashboard-operator migration, the DSCI CR returns `0.0.0` for the version, causing the About dialog to show "Unknown".

This PR switches the About dialog to read from the DSC (`DataScienceCluster`) CR instead, which is already fetched in the same component (it powers the component releases table) and correctly populated by the ODH Operator. This fixes the `testAboutDialog.cy.ts` E2E test failure documented in RHOAIENG-80304.

**Changes:**
- `AboutDialog.tsx`: Removed `useFetchDsciStatus` import/hook, switched `dsciStatus?.release?.name` → `dscStatus?.release?.name` and `dsciStatus?.release?.version` → `dscStatus?.release?.version`, simplified error/loading state
- `AboutDialog.spec.tsx`: Removed DSCI mock, added `release` data to the existing DSC mock
- `application.cy.ts`: Switched three Cypress mock tests from intercepting `GET /api/dsci/status` to using `mockDscStatus()` with `GET /api/dsc/status`

## How Has This Been Tested?

- Unit tests: `npx jest --testPathPattern="AboutDialog" --no-coverage` — both ODH and RHOAI test cases pass
- Lint: `npx eslint` on all three changed files — no errors
- Type check: `npx tsc --noEmit` — no errors in changed files (pre-existing errors in unrelated areas)
- **Manually tested on a ROSA cluster**: Ran local dev server (backend + frontend) connected to a ROSA cluster (`api.ui-monarch-rosa.ci5p.p3.openshiftapps.com`). Patched DSC with `release: {name: "Open Data Hub Operator", version: "3.5.0"}` while DSCI retained `{name: "Open Data Hub", version: "0.0.0"}`. Verified the About dialog correctly shows "Open Data Hub Operator version: 3.5.0" (from DSC), confirming it no longer reads from the broken DSCI CR.

## Test Impact

- Updated unit test (`AboutDialog.spec.tsx`) to remove DSCI mocking and add `release` data to the DSC mock
- Updated Cypress mock tests (`application.cy.ts`) to intercept DSC instead of DSCI for version data
[...]

PR #8955, opened 2026-07-29
Fix failing gen-ai e2e test

ikeola13:
## Description

- Fixed  failing gen-ai e2e testUpdated by updating selector in genAiPlayground ts file.

## How Has This Been Tested?

locally and on cluster

## Request review criteria:

Self checklist (all need to be checked):
- [x] The developer has manually tested the changes and verified that the changes work
- [ ] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [x] The developer has added tests or explained why testing cannot be added (unit or cypress tests for related changes)
- [x] The code follows our [Best Practices](/docs/best-practices.md) (React coding standards, PatternFly usage, performance considerations)

If you have UI changes: 

- [ ] Included any necessary screenshots or gifs if it was a UI change.
- [ ] Included tags to the UX team if it was a UI/UX change.

After the PR is posted & before it merges:
- [x] The developer has tested their solution on a cluster by using the image produced by the PR to `main`

## Summary by CodeRabbit

* **Tests**
  * Updated end-to-end test interactions to use the settings model selector for model-toggle checks and selection.
  * Improved navigation and reliability for opening the AI playground and switching models.

PR #9223, opened 2026-08-10
RHOAIENG-81833: read About dialog version from DSC instead of DSCI (#…

sridarna:
…9114)

The About dialog was reading release name and version from the DSCI (DataScienceClusterInitialization) CR, which returns 0.0.0 after the dashboard-operator migration. Switch to reading from the DSC (DataScienceCluster) CR, which is already fetched in the same component and correctly populated by the ODH Operator.

## Description

cherrypick https://github.com/opendatahub-io/odh-dashboard/pull/9114 to 3.5
We need to backport to address the potential bug. https://redhat.atlassian.net/browse/RHOAIENG-81833.
Also backport minor e2e pr https://github.com/opendatahub-io/odh-dashboard/pull/8955


sridarna:
We need to get approval for this changes to be backport to 3.5 post code freeze.

PR #9229, opened 2026-08-10
test(RHOAIENG-79908): add texts to preserve search input min width in ai hub catalog

Philip-Carneiro:
https://redhat.atlassian.net/browse/RHOAIENG-79908

## Description

To prevent in the future to bug of the searchbar size re-appear.
I created tests to don't let it happen again.

## How Has This Been Tested?

It was tested if the search input field in Model Catalog, MCP Catalog and Agent Catalog have at least 400px width via cypress

## Test Impact

Don't let this bug re-appear in the future

## Request review criteria:

If you have UI changes: 

## Summary by CodeRabbit

- **Tests**
  - Added regression coverage for search fields in the Agents, MCP, and Model Catalogs.
  - Verified that each catalog search field is visible and maintains a minimum width of 400 pixels.
  - Expanded automated catalog scenarios to validate search input behavior across supported catalog experiences.
  - Improved test coverage for catalog configuration, navigation, and search-related interactions, helping prevent future layout regressions.


manaswinidas (changes requested):
Type-check is failing

paulovmr:
FYI Please be sure to rebase this pull request before approving it, to get the latest mock location changes

packages/cypress/cypress/tests/mocked/mcpCatalog/mcpCatalogSearchInput.cy.ts
```diff
@@ -0,0 +1,292 @@
+import { mockDashboardConfig } from '@odh-dashboard/k8s-core/__mocks__/mockDashboardConfig';
```
ppadti, line 1: Instead on creating new test file just 3 assertion - can we add it to the existing file - because we are adding all the interception here just for 3 assertion. So can we just user theexisting test files for all 3 catalog here?

ppadti (changes requested):

packages/cypress/cypress/tests/mocked/mcpCatalog/mcpCatalogSearchInput.cy.ts
Philip-Carneiro: Good point, Pushpa. I considered that, but there are a few reasons why a separate file makes more sense here.
The existing test files for all three catalogs (modelCatalog, mcpCatalog, agentsCatalog) live in the upstream subtree. Modifying those here would create conflicts on the next upstream sync.

More importantly, this bug only happens in midstream. In upstream the search input renders correctly because the layout context is different. The min-width regression is caused by how our monorepo integrates the upstream components, so the test belongs here where the problem actually occurs. Adding it to the upstream test files wouldn't make sense since they'd be testing for a bug that doesn't exist there.

The test also validates rendered min-width using .invoke('outerWidth').should('be.gte', 400), which depends on the Cypress viewport configuration. The upstream Cypress setup has a different viewport where this assertion wouldn't work reliably.

The intercept setup is verbose, yes, but that's the cost of testing layout across three independent catalog pages without touching upstream files.

ppadti: I see tests file for [agent catalog](https://github.com/opendatahub-io/odh-dashboard/blob/main/packages/cypress/cypress/tests/mocked/agentsCatalog/agentsCatalogFeatureFlag.cy.ts) and [MCP catalog](https://github.com/opendatahub-io/odh-dashboard/blob/main/packages/cypress/cypress/tests/mocked/mcpCatalog/mcpCatalogFeatureFlag.cy.ts) in packages/cypress folder. we can use that and keep the model catalog search bar test (this file) under modelCatalog folder here: https://github.com/opendatahub-io/odh-dashboard/tree/main/packages/cypress/cypress/tests/mocked/modelCatalog - wdyt?

Philip-Carneiro: done

packages/cypress/cypress/tests/mocked/agentsCatalog/agentsCatalogSearchInput.cy.ts
```diff
@@ -0,0 +1,193 @@
...
+        ],
+        size: 1,
+        pageSize: 10,
+        nextPageToken: '',
+      },
+    },
+  });
+};
+
+describe('Agents Catalog search input minimum width regression', () => {
```
ppadti: Just a quick question - can't we add these tests suite to packages/cypress/cypress/tests/mocked/agentsCatalog/agentsCatalogFeatureFlag.cy.ts - this file? same goes for MCP

packages/cypress/cypress/tests/mocked/agentsCatalog/agentsCatalogSearchInput.cy.ts
```diff
@@ -0,0 +1,193 @@
...
+import { mockDscStatus } from '@odh-dashboard/plugin-core/__mocks__/mockDscStatus';
+import { mockDsciStatus } from '@odh-dashboard/plugin-core/__mocks__/mockDsciStatus';
+import { DataScienceStackComponent } from '@odh-dashboard/plugin-core/areas';
+import { asProductAdminUser } from '../../../utils/mockUsers';
+import { agentsCatalogPage } from '../../../pages/agentsCatalog';
+
+const API_VERSION = 'v1';
+const REGISTRIES_NAMESPACE = 'odh-model-registries';
+
+const setupModelCatalogIntercepts = () => {
```
ppadti: setupModelCatalogIntercepts is duplicated in all 3 search-input test files. The DSC/user/namespace intercepts + model_catalog/sources and model_catalog/labels are shared (MCP/Agents just add assetType).

Can we extract that common setup into a helper and reuse it in all 3 files? Each file can then add only its catalog-specific intercepts

ppadti (changes requested):

packages/cypress/cypress/tests/mocked/agentsCatalog/agentsCatalogSearchInput.cy.ts
```diff
@@ -0,0 +1,88 @@
+import { mockDashboardConfig } from '@odh-dashboard/k8s-core/__mocks__/mockDashboardConfig';
+import { asProductAdminUser } from '../../../utils/mockUsers';
+import { agentsCatalogPage } from '../../../pages/agentsCatalog';
+import { setupModelCatalogIntercepts } from '../catalogHelpers';
+
+const API_VERSION = 'v1';
+
+const setupAgentsCatalogIntercepts = () => {
+  cy.intercept('GET', `**/model-registry/api/${API_VERSION}/agent_catalog/sources*`, {
```
ppadti: This intercept is wrong, because we use model_catalog/sources endpoint to get the sources with assetType=agents to get the sources for Agents

packages/cypress/cypress/tests/mocked/agentsCatalog/agentsCatalogSearchInput.cy.ts
```diff
@@ -0,0 +1,88 @@
...
+          },
+        ],
+        size: 1,
+        pageSize: 10,
+        nextPageToken: '',
+      },
+    },
+  });
+
+  cy.intercept('GET', `**/model-registry/api/${API_VERSION}/agent_catalog/labels*`, {
```
ppadti: same here

packages/cypress/cypress/tests/mocked/agentsCatalog/agentsCatalogSearchInput.cy.ts
```diff
@@ -0,0 +1,88 @@
...
+      mockDashboardConfig({
+        disableModelRegistry: false,
+        agentsCatalog: true,
+      }),
+    );
+
+    setupModelCatalogIntercepts();
+    setupAgentsCatalogIntercepts();
+
+    cy.visitWithLogin('/ai-hub/agents/catalog');
```
ppadti: Can we use agentsCatalogPage.visit() here instead of this?

packages/cypress/cypress/tests/mocked/modelCatalog/modelCatalogSearchInput.cy.ts
```diff
@@ -0,0 +1,24 @@
...
+      mockDashboardConfig({
+        disableModelRegistry: false,
+      }),
+    );
+
+    setupModelCatalogIntercepts();
+
+    cy.visitWithLogin('/ai-hub/models/catalog');
+    modelCatalog.findSearchInput().should('be.visible').invoke('outerWidth').should('be.gte', 400);
+  });
```
ppadti: I don't think this would fix the regression of search bar issue though.

ppadti (changes requested):
Also this PR has a type check failure.

packages/cypress/cypress/tests/mocked/catalogHelpers.ts
```diff
@@ -0,0 +1,106 @@
+import { mockDscStatus } from '@odh-dashboard/plugin-core/__mocks__/mockDscStatus';
+import { mockDsciStatus } from '@odh-dashboard/plugin-core/__mocks__/mockDsciStatus';
+import { DataScienceStackComponent } from '@odh-dashboard/plugin-core/areas';
+
+const API_VERSION = 'v1';
+const REGISTRIES_NAMESPACE = 'odh-model-registries';
+
+export const setupModelCatalogIntercepts = (): void => {
```
ConorOM1, line 8: setupModelCatalogIntercepts() largely duplicates setupCommonIntercepts() in securityInsightsTab.cy.ts. Should we consolidate into one shared helper so both tests import from the same place?

packages/cypress/cypress/tests/mocked/agentsCatalog/agentsCatalogSearchInput.cy.ts
```diff
@@ -0,0 +1,54 @@
+import { mockDashboardConfig } from '@odh-dashboard/k8s-core/__mocks__/mockDashboardConfig';
+import { asProductAdminUser } from '../../../utils/mockUsers';
+import { agentsCatalogPage } from '../../../pages/agentsCatalog';
+import { setupModelCatalogIntercepts } from '../catalogHelpers';
+
+const API_VERSION = 'v1';
```
ConorOM1: can we export this const from catalogHelpers.ts ?

packages/cypress/cypress/tests/mocked/mcpCatalog/mcpCatalogSearchInput.cy.ts
```diff
@@ -0,0 +1,93 @@
...
+      mockDashboardConfig({
+        disableModelRegistry: false,
+        mcpCatalog: true,
+      }),
+    );
+
+    setupModelCatalogIntercepts();
+    setupMcpCatalogIntercepts();
+
+    cy.visitWithLogin('/ai-hub/mcp-servers/catalog');
```
ConorOM1: can we add a visit method for McpcatalogPage instead ?

ConorOM1 (commented):
thanks for actioning @Philip-Carneiro added some comments

packages/cypress/cypress/tests/mocked/catalogHelpers.ts
Philip-Carneiro: Good catch, they do share a common core. I looked at merging them but the two helpers have diverged enough that a single shared function would end up heavily parameterized. setupCommonIntercepts in securityInsightsTab calls asProductAdminUser() and intercepts /api/config internally, enables TRUSTY_AI in the DSC status, takes a disableLMEval flag, and mocks the source-specific models/** and artifacts/** endpoints. setupModelCatalogIntercepts here does none of that and instead mocks model_catalog/labels, the model list, and filter_options, while leaving /api/config and the user to each caller (they pass different flags). Folding both into one helper would also mean touching securityInsightsTab.cy.ts, which is outside this PR. Happy to do the consolidation as a separate follow-up task if you think it's worth it, but I'd keep it out of this change to avoid the scope creep and the risk to that test.

ConorOM1: ok, seems like not a suitable candidate

PR #9240, opened 2026-08-11
fix: update leftover mock imports after #9144

dpanshug:
## Description

Follow-up to #9144 (`RHOAIENG-79875`), which moved `mockInferenceServiceK8sResource` and `mockPodK8sResource` out of `frontend/src/__mocks__` into owning packages, but left two unit specs still importing the old `#~/__mocks__/…` paths.

## How Has This Been Tested?

- Confirmed `main` Type-Check fails with the same four `TS2307` errors for these mocks
- Confirmed no remaining `#~/__mocks__/mockInferenceServiceK8sResource` / `mockPodK8sResource` imports after this change
- Ran eslint on the two updated files (including import-order auto-fix)

## Test Impact

No new tests. This only corrects import paths so existing unit specs type-check again after the mock relocation in #9144.

## Request review criteria:


paulovmr (approved):
/lgtm
/approve

PR #7951, opened 2026-06-10
Add startup validation to prevent InsecureSkipVerify in production en…

toklumpp:
https://issues.redhat.com/browse/RHOAIENG-67920

## Description

Add startup validation in `packages/gen-ai/bff/cmd/main.go` to prevent the `--insecure-skip-verify` flag (or `INSECURE_SKIP_VERIFY` env var) from being used in deployed or CI environments.

**Security layers:**

1. **TLS certificate detection (primary)**: If `--cert-file` is set, the BFF refuses to start with `InsecureSkipVerify`. In production RHOAI, TLS certificates are always mounted, so this catches real deployed environments regardless of which env vars are configured.

2. **ENV/CI check (secondary)**: If `ENV` matches `prod`, `production`, or `staging`, or `CI` is truthy, the BFF refuses to start. This provides defense-in-depth for environments that set these variables.

In local development (no cert-file, no production ENV/CI), `INSECURE_SKIP_VERIFY=true` works without any additional variables. A warning is logged at startup.

**Other changes:**

- Extract `parseBoolTruthy()` helper for consistent truthy value parsing (`"true"`, `"1"`, `"yes"`, `"on"`, case-insensitive) -- used for both `CI` and any future boolean env vars.
- Validation runs immediately after `flag.Parse()`, before logger setup (fail-fast). Uses `slog.Error` with Go default formatter since the process exits immediately on failure.
- The validation function accepts all inputs as parameters (no `os.Getenv` calls inside) to enable hermetic testing.

## How Has This Been Tested?

**Unit tests** in `packages/gen-ai/bff/cmd/main_test.go`:

- `TestParseBoolTruthy`: 15 cases covering all truthy/falsy variants, case sensitivity, whitespace
- `TestValidateInsecureSkipVerify`: 20 cases covering:
  - InsecureSkipVerify disabled (pass)
  - InsecureSkipVerify enabled in local dev (pass with warning)
  - ENV=prod/production/staging (fail)
  - ENV case-insensitivity and whitespace normalization (fail)
  - CI=true/1/yes/on (fail), CI=false (pass)
  - CI + ENV combinations (fail with combined label)
  - cert-file set (fail, takes priority over ENV)
  - cert-file set with InsecureSkipVerify disabled (pass)

## Test Impact

[...]

PR #9245, opened 2026-08-11
feat(gen-ai): implement /v1/embeddings proxy handler with credential resolution [RHOAIENG-79576]

gmenher:
https://issues.redhat.com/browse/RHOAIENG-79576

## Description

Implement `POST /api/v1/genai-proxy/ns/:namespace/v1/embeddings` — a transparent proxy that accepts OpenAI-compatible embeddings requests from OGX, resolves the model to a concrete upstream endpoint with credentials, and forwards the request directly to the underlying model.

### What was added

- `genai_proxy_embeddings_handler.go` — POST handler + `resolveModelEndpoint()` helper + `resolveNamespaceModel()` helper + `infraError` type
- `genai_proxy_embeddings_handler_test.go` — 4 unit tests covering error paths
- `GenAIProxyNSEmbeddingsPath` constant in `api_constants.go`
- Route registration in `app.go`

### How model resolution works

The handler extracts the `model` field from the request body and resolves it to an upstream URL + API key based on the model ID format:

- **MaaS models** (`maas-` prefix): resolves inference URL via MaaS BFF catalog + obtains ephemeral token
- **Custom endpoints** (provider-qualified with `/`): resolves from `gen-ai-aa-custom-model-endpoints` ConfigMap + K8s Secret
- **Namespace models** (bare name): resolves InferenceService/LLMInferenceService URL + uses user JWT

The provider prefix is stripped from the `model` field before proxying. OGX sends `provider/model-name` but upstreams expect just `model-name`.

### URL normalization

The base URL is normalized to include `/v1` if missing. The spike validated that some model endpoints return URLs without the `/v1` suffix, which would cause 404s on the upstream's `/v1/embeddings` path.

### Design decisions

**Transparent proxy:** The response body is returned unchanged — status code, headers, and body from the upstream flow through directly. OGX expects OpenAI-compatible responses from the upstream, not BFF-wrapped errors.

**Infrastructure errors vs not-found:** A dedicated `infraError` type separates K8s client failures and MaaS unavailability (returned as 500) from genuine "model not found" scenarios (returned as 404). Without this, a cluster outage would mask as "model not found".

**Provider prefix stripped from body:** Validated during cluster testing — LiteLLM returned `key_model_access_denied` when receiving the full provider-qualified ID instead of the bare model name. The handler rewrites the `model` field in the body before forwarding.

**Reusable resolver:** `resolveModelEndpoint()` is extracted as a shared helper that will be reused by the upcoming `/v1/chat/completions` handler ([RHOAIENG-79574](https://redhat.atlassian.net/browse/RHOAIENG-79574)) which follows the same resolution pattern.

**Reference:** Architecture validated in spike [RHOAIENG-78871](https://redhat.atlassian.net/browse/RHOAIENG-78871): "POST /api/v1/genai-proxy/ns/:namespace/v1/embeddings — resolves model → {endpoint, token} by searching K8s and proxies directly to the underlying model. BFF normalizes base URL to include /v1 if missing."

### Why this approach and not a mock/stub

The handler is a transparent proxy — its core logic is network I/O (resolve model, call upstream, forward response). A mock HTTP server in a unit test would only validate that `http.Client.Do` works, which is Go standard library behavior. The real value is in validating the full chain: ConfigMap parsing → Secret fetch → URL construction → TLS handshake → upstream response forwarding. This requires a real cluster with real credentials, which is why the E2E cluster validation is the primary test evidence.

### Unit tests

- `make test` — all specs pass (0 failures across all packages)
- 4 handler-specific tests covering: 401 (no auth), 400 (missing model), 400 (invalid JSON), 404 (model not found)

### End-to-end cluster validation

Deployed custom BFF image on an OpenShift cluster (RHOAI 3.6-ea.1) via CSV patching. Validated via port-forward to the `gen-ai-ui` pod.

**Error path validation:**

```bash
TOKEN=$(oc whoami -t)
oc port-forward deploy/gen-ai-ui -n redhat-ods-applications 8143:8143 &

# 401 — no auth
curl -sk -X POST "https://localhost:8143/api/v1/genai-proxy/ns/test-ns/v1/embeddings" \
  -H "Content-Type: application/json" -d '{"model":"x","input":"hi"}'
# → {"error":{"code":"401","message":"missing required Header: x-forwarded-access-token"}}

# 400 — missing model field
curl -sk -X POST "https://localhost:8143/api/v1/genai-proxy/ns/test-ns/v1/embeddings" \
  -H "x-forwarded-access-token: $TOKEN" -H "Content-Type: application/json" -d '{"input":"hi"}'
# → {"error":{"code":"400","message":"missing required field: model"}}

# 404 — nonexistent model
curl -sk -X POST "https://localhost:8143/api/v1/genai-proxy/ns/gaizka-test/v1/embeddings" \
  -H "x-forwarded-access-token: $TOKEN" -H "Content-Type: application/json" \
  -d '{"model":"nonexistent-xyz","input":"hi"}'
# → {"error":{"code":"404","message":"the requested resource could not be found"}}

# 502 — unreachable upstream (custom endpoint pointing to https://unreachable.invalid)
curl -sk -X POST "https://localhost:8143/api/v1/genai-proxy/ns/gaizka-test/v1/embeddings" \
  -H "x-forwarded-access-token: $TOKEN" -H "Content-Type: application/json" \
  -d '{"model":"dead-provider/dead-model","input":"test"}'
# → {"error":{"code":"502","message":"upstream unreachable: dial tcp: lookup unreachable.invalid: no such host"}}
```

**Happy path validation (200 — real embeddings from Nomic via LiteMaaS):**

```bash
# 1. Create custom endpoint pointing to LiteMaaS
cat <<'INNER' | oc apply -n gaizka-test -f -
apiVersion: v1
kind: ConfigMap
metadata:
  name: gen-ai-aa-custom-model-endpoints
  labels:
data:
  config.yaml: |
    providers:
      inference:
        - provider_id: litellm-embed
          provider_type: remote::openai
          config:
            base_url: https://litemaas.rhoai.rh-aiservices-bu.com/v1
            allowed_models:
              - Nomic-embed-text-v2-moe
            custom_gen_ai:
              api_key:
                secretRef:
                  name: litellm-secret
                  key: api_key
    registered_resources:
      models:
        - provider_id: litellm-embed
          model_id: Nomic-embed-text-v2-moe
          model_type: embedding
INNER
oc create secret generic litellm-secret -n gaizka-test --from-literal=api_key='<litellm-key>'

# 2. Call the proxy endpoint
curl -sk -X POST "https://localhost:8143/api/v1/genai-proxy/ns/gaizka-test/v1/embeddings" \
  -H "x-forwarded-access-token: $TOKEN" \
  -H "Content-Type: application/json" \
  -d '{"model":"litellm-embed/Nomic-embed-text-v2-moe","input":"Hello world","encoding_format":"float"}'
```

**Result:**
```json
{
  "model": "nomic-embed-text-v2-moe",
  "data": [{"embedding": [0.0058, 0.0263, -0.0102, ...], "index": 0, "object": "embedding"}],
  "object": "list",
  "usage": {"prompt_tokens": 4, "total_tokens": 4}
}
```
HTTP 200 ✅ — Full chain validated: ConfigMap resolution → Secret fetch → provider prefix stripped → URL construction → TLS to LiteMaaS → LiteMaaS routed to Nomic → real embedding vector returned transparently.

## Jira

- **Ticket:** [RHOAIENG-79576](https://redhat.atlassian.net/browse/RHOAIENG-79576)
- **Epic:** [RHOAIENG-79569](https://redhat.atlassian.net/browse/RHOAIENG-79569) (OGX Zero-Restart — BFF /v1/ Proxy Endpoints)
- **Feature:** [RHAISTRAT-1921](https://redhat.atlassian.net/browse/RHAISTRAT-1921)
- **Spike:** [RHOAIENG-78871](https://redhat.atlassian.net/browse/RHOAIENG-78871)

[RHOAIENG-79576]: https://redhat.atlassian.net/browse/RHOAIENG-79576?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

- **New Features**
  - Added an OpenAI-compatible embeddings endpoint for namespace-scoped GenAI proxy requests.
  - Supports authenticated requests, model validation, provider-specific routing, and namespace-based model services.
  - Transparently forwards valid embedding requests and returns upstream responses.

- **Bug Fixes**
  - Added clear handling for unauthorized requests, malformed input, missing models, and unavailable services.

[RHOAIENG-79574]: https://redhat.atlassian.net/browse/RHOAIENG-79574?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/gen-ai/bff/internal/api/genai_proxy_embeddings_handler.go
+++ b/packages/gen-ai/bff/internal/api/genai_proxy_embeddings_handler.go
@@ -1,0 +1,193 @@
+package api
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"net/http"
+	"strings"
+
+	"github.com/julienschmidt/httprouter"
+	"github.com/opendatahub-io/gen-ai/internal/constants"
+	"github.com/opendatahub-io/gen-ai/internal/integrations"
+	"github.com/opendatahub-io/gen-ai/internal/integrations/bffclient"
+	k8s "github.com/opendatahub-io/gen-ai/internal/integrations/kubernetes"
+	"github.com/opendatahub-io/gen-ai/internal/models"
+)
+
+// GenAIProxyNSEmbeddingsHandler handles POST /api/v1/genai-proxy/ns/:namespace/v1/embeddings.
+//
+// Accepts an OpenAI-compatible embeddings request, resolves the model to a concrete
+// endpoint and credentials, and proxies the request directly to the upstream model.
+// Returns the upstream response unchanged (transparent proxy).
+//
+// Auth is required: OGX forwards the user's JWT via X-OGX-Provider-Data →
+// forward_headers → x-forwarded-access-token header.
+func (app *App) GenAIProxyNSEmbeddingsHandler(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
+	ctx := r.Context()
+
+	namespace := ps.ByName("namespace")
+	if namespace == "" {
+		app.badRequestResponse(w, r, errors.New("missing namespace in path"))
+		return
+	}
+
+	ctx = context.WithValue(ctx, constants.NamespaceQueryParameterKey, namespace)
+	r = r.WithContext(ctx)
+
+	identity, ok := ctx.Value(constants.RequestIdentityKey).(*integrations.RequestIdentity)
+	if !ok || identity == nil || identity.Token == "" {
+		app.unauthorizedResponse(w, r, errors.New("missing authentication identity"))
+		return
+	}
+
+	// Read and parse the request body to extract model field
+	body, err := io.ReadAll(r.Body)
+	if err != nil {
+		app.badRequestResponse(w, r, fmt.Errorf("failed to read request body: %w", err))
+		return
+	}
+	defer r.Body.Close()
+
+	var reqBody struct {
+		Model string `json:"model"`
+	}
+	if err := json.Unmarshal(body, &reqBody); err != nil {
+		app.badRequestResponse(w, r, fmt.Errorf("invalid JSON in request body: %w", err))
+		return
+	}
+	if reqBody.Model == "" {
+		app.badRequestResponse(w, r, errors.New("missing required field: model"))
+		return
+	}
+
+	// Resolve model → endpoint URL + credentials
+	baseURL, apiKey, err := app.resolveModelEndpoint(ctx, reqBody.Model, namespace)
+	if err != nil {
+		app.logger.Warn("Model resolution failed", "model", reqBody.Model, "error", err)
+		app.notFoundResponse(w, r)
+		return
+	}
+
+	// Normalize base URL to include /v1 if missing (per spike findings)
+	baseURL = strings.TrimSuffix(baseURL, "/")
+	if !strings.HasSuffix(baseURL, "/v1") {
+		baseURL += "/v1"
+	}
+
+	// Proxy the request to upstream
+	upstreamURL := baseURL + "/embeddings"
+	proxyReq, err := http.NewRequestWithContext(ctx, http.MethodPost, upstreamURL, strings.NewReader(string(body)))
+	if err != nil {
+		app.serverErrorResponse(w, r, fmt.Errorf("failed to create upstream request: %w", err))
+		return
+	}
+	proxyReq.Header.Set("Content-Type", "application/json")
+	if apiKey != "" {
+		proxyReq.Header.Set("Authorization", "Bearer "+apiKey)
+	}
+
+	resp, err := app.httpClient.Do(proxyReq)
+	if err != nil {
+		app.errorResponse(w, r, &integrations.HTTPError{
+			StatusCode: http.StatusBadGateway,
+			ErrorResponse: integrations.ErrorResponse{
+				Code:    "502",
+				Message: fmt.Sprintf("upstream unreachable: %v", err),
+			},
+		})
+		return
+	}
+	defer resp.Body.Close()
+
+	// Forward the upstream response unchanged (transparent proxy)
+	respBody, err := io.ReadAll(resp.Body)
+	if err != nil {
+		app.serverErrorResponse(w, r, fmt.Errorf("failed to read upstream response: %w", err))
+		return
+	}
+
+	w.Header().Set("Content-Type", resp.Header.Get("Content-Type"))
+	w.WriteHeader(resp.StatusCode)
+	w.Write(respBody)
+}
+
+// resolveModelEndpoint resolves a model ID to its upstream endpoint URL and API key.
+// The model ID may be provider-qualified (e.g. "genai-bff-proxy/text-embedding-3-small")
+// or bare (e.g. "my-isvc-model"). Resolution follows the same priority as getProviderData:
+//  1. Custom endpoint (provider-qualified with known provider) → ConfigMap + Secret
+//  2. MaaS (maas- prefix) → MaaS BFF catalog URL + ephemeral token
+//  3. Namespace (bare name) → InferenceService/LLMInferenceService URL + user JWT
+func (app *App) resolveModelEndpoint(ctx context.Context, modelID, namespace string) (baseURL, apiKey string, err error) {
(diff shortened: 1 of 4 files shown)


packages/gen-ai/bff/internal/api/genai_proxy_embeddings_handler.go
```diff
@@ -0,0 +1,224 @@
...
+	if err != nil {
+		app.serverErrorResponse(w, r, fmt.Errorf("failed to create upstream request: %w", err))
+		return
+	}
+	proxyReq.Header.Set("Content-Type", "application/json")
+	if apiKey != "" {
+		proxyReq.Header.Set("Authorization", "Bearer "+apiKey)
+	}
+
+	resp, err := app.httpClient.Do(proxyReq)
```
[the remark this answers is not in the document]
gmenher: This is the shared app.httpClient used by all BFF outbound calls (LlamaStack, NeMo, MaaS BFF, external model verification, etc.). InsecureSkipVerify is controlled by the --insecure-skip-verify flag and is already guarded by startup validation (PR #7951) — it cannot be enabled when TLS certificates are mounted (production). The embeddings handler doesn't introduce a new TLS surface; it reuses the same hardened client as every other handler.

packages/gen-ai/bff/internal/api/genai_proxy_embeddings_handler.go
```diff
@@ -0,0 +1,240 @@
...
+)
+
+// GenAIProxyNSEmbeddingsHandler handles POST /api/v1/genai-proxy/ns/:namespace/v1/embeddings.
+//
+// Accepts an OpenAI-compatible embeddings request, resolves the model to a concrete
+// endpoint and credentials, and proxies the request directly to the upstream model.
+// Returns the upstream response unchanged (transparent proxy).
+//
+// Auth is required: OGX forwards the user's JWT via X-OGX-Provider-Data →
+// forward_headers → x-forwarded-access-token header.
```
jharan1: I think the `forward_headers` approach will be getting replaced by `passthrough_api_key`, there might be a few comments we can get around to updating to reflect that (not just in this PR), cc @NickGagan

packages/gen-ai/bff/internal/api/genai_proxy_embeddings_handler.go
```diff
@@ -0,0 +1,240 @@
+package api
```
jharan1, line 1: We could opt to combine the `genai_proxy_embeddings_handler.go` and other additions (models, chat completions) into the `genai_proxy_handler.go` file if it's not too large. Could do in follow up PR if we want that approach.

jharan1 (commented):
Thanks for PR @gmenher

packages/gen-ai/bff/internal/api/genai_proxy_embeddings_handler.go
gmenher: Good point, updated.

packages/gen-ai/bff/internal/api/genai_proxy_embeddings_handler.go
gmenher: Yeah, actually is a good idea, I can consolidate all proxy handlers into `genai_proxy_handler.go` as a follow-up PR I think. I mean, once these  3 endpoints PRs get merged. Endpoint /v1/models is already merged to main and moving embeddings now to `genai_proxy_handler.go` too would make the diff confusing I guess (it will be new code and some restructure needed). PR #9279 also has its handler in another different file. With the follow-up pr should be cleaner. Anyways, let me know what you think.

jharan1: That makes sense :)

jharan1 (commented):
/lgtm 

More work to come in FUP PRs.

jharan1 (approved):

PR #9251, opened 2026-08-11
feat(model-serving): Add Model Deployed/Updated tracking events to deployment wizard

YuliaKrimerman:
https://issues.redhat.com/browse/RHOAIENG-82390

## Description

The new deployment wizard fires no `Model Deployed` event. The only implementation lived in the legacy `ManageKServeModal`, so every deployment created through the wizard was invisible to Amplitude. This broke the *Model Deployed adoption*, *Deployment usage breakdown*, and *Deployment error rate* dashboards.

This PR adds:
- A shared `fireModelDeployed` helper under `packages/model-serving/src/shared/tracking/` that wraps `fireFormTrackingEvent` with typed deployment properties.
- Tracking calls in `useModelDeploymentSubmit` for success (`outcome: 'submit', success: true`) and failure (`outcome: 'submit', success: false, error: <message>`).
- A cancel tracking event when the user confirms discarding changes via `ExitDeploymentModal`.
- `Model Updated` fires instead of `Model Deployed` when `existingDeployment` is set (edit flow), matching legacy behaviour.
- A `WizardTrackingPropertiesExtension` extension point so spoke packages (kserve, llmd-serving, nim-serving) can contribute platform-specific properties without cross-package imports.
- Base properties emitted: `type`, `runtime`, `servingRuntimeName`, `servingRuntimeFormat`, `numReplicas` — parity with `ManageKServeModal.tsx`.

## How Has This Been Tested?

- **Unit tests**: 11 tests across 2 suites (`deploymentTracking.spec.ts`, `useWizardTrackingProperties.spec.ts`) — all passing.
- **Cypress mocked tests**: `modelServingDeployTracking.cy.ts` verifies the event fires on successful deploy, on error, and on cancel by spying on `console.log` in DEV_MODE.
- **Manual verification**: Ran local dev server, navigated through the deployment wizard, and confirmed `[Telemetry event triggered: Model Deployed - {...}]` appears in the browser console on submit/cancel.
- 
Screenshot 2026-08-11 at 12 04 04 PM
Screenshot 2026-08-11 at 12 04 57 PM

- **Lint & type-check**: Both pass cleanly on the affected package.

## Test Impact

New test files added:
- `packages/model-serving/src/shared/tracking/__tests__/deploymentTracking.spec.ts` — unit tests for the `fireModelDeployed` helper (6 test cases)
- `packages/model-serving/src/shared/tracking/__tests__/useWizardTrackingProperties.spec.ts` — unit tests for the extension hook (5 test cases)
- `packages/model-serving/cypress/tests/mocked/modelServing/modelServingDeployTracking.cy.ts` — Cypress E2E tests for event tracking (3 test cases)

## Request review criteria:

## Summary by CodeRabbit

* **New Features**
  * Added tracking for successful, updated, canceled, and failed model deployments.
  * Included deployment metadata and platform-specific properties in tracking events.
  * Added support for identifying the active platform’s tracking properties.

* **Tests**
  * Added coverage for deployment outcomes, event selection, cancellation, failures, editing deployments, and platform-specific properties.
  * Added end-to-end validation of analytics events during deployment workflows.

### DIFF
--- a/packages/model-serving/src/components/deploymentWizard/ModelDeploymentWizard.tsx
+++ b/packages/model-serving/src/components/deploymentWizard/ModelDeploymentWizard.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import { PageSection, Wizard, WizardStep } from '@patternfly/react-core';
-import { ApplicationsPage } from '@odh-dashboard/ui-core';
+import { ApplicationsPage, TrackingOutcome } from '@odh-dashboard/ui-core';
 import type { ProjectKind } from '@odh-dashboard/k8s-core';
 import { SupportedArea, useIsAreaAvailable } from '@odh-dashboard/plugin-core/areas';
 import {
@@ -30,6 +30,7 @@ import {
   ModelDeploymentFooter,
   ModelDeploymentWizardFooter,
 } from '../generic/WizardFooterWithDisablingNext';
+import { fireModelDeployed } from '../../shared/tracking/deploymentTracking';
 
 export type ModelDeploymentWizardViewMode = 'form' | 'yaml-preview' | 'yaml-edit';
 
@@ -54,9 +55,20 @@ const ModelDeploymentWizard: React.FC<ModelDeploymentWizardProps> = ({
   returnRoute,
   cancelReturnRoute,
 }) => {
+  const isEdit = !!existingDeployment;
   const onRefresh = useRefreshWizardPage(existingDeployment);
-  const { isExitModalOpen, openExitModal, closeExitModal, handleExitConfirm, exitWizardOnSubmit } =
-    useExitDeploymentWizard({ returnRoute, cancelReturnRoute });
+  const {
+    isExitModalOpen,
+    openExitModal,
+    closeExitModal,
+    handleExitConfirm: baseHandleExitConfirm,
+    exitWizardOnSubmit,
+  } = useExitDeploymentWizard({ returnRoute, cancelReturnRoute });
+
+  const handleExitConfirm = React.useCallback(() => {
+    fireModelDeployed({ outcome: TrackingOutcome.cancel }, isEdit);
+    baseHandleExitConfirm();
+  }, [baseHandleExitConfirm, isEdit]);
 
   const isYAMLViewerEnabled = useIsAreaAvailable(SupportedArea.YAML_VIEWER).status;
   const [viewMode, setViewMode] = React.useState<ModelDeploymentWizardViewMode>(
--- a/packages/model-serving/src/components/deploymentWizard/ModelDeploymentWizard.tsx
+++ b/packages/model-serving/src/components/deploymentWizard/ModelDeploymentWizard.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import { PageSection, Wizard, WizardStep } from '@patternfly/react-core';
-import { ApplicationsPage, TrackingOutcome } from '@odh-dashboard/ui-core';
+import { ApplicationsPage } from '@odh-dashboard/ui-core';
 import type { ProjectKind } from '@odh-dashboard/k8s-core';
 import { SupportedArea, useIsAreaAvailable } from '@odh-dashboard/plugin-core/areas';
 import {
@@ -30,7 +30,6 @@ import {
   ModelDeploymentFooter,
   ModelDeploymentWizardFooter,
 } from '../generic/WizardFooterWithDisablingNext';
-import { fireModelDeployed } from '../../shared/tracking/deploymentTracking';
 
 export type ModelDeploymentWizardViewMode = 'form' | 'yaml-preview' | 'yaml-edit';
 
@@ -57,18 +56,8 @@ const ModelDeploymentWizard: React.FC<ModelDeploymentWizardProps> = ({
 }) => {
   const isEdit = !!existingDeployment;
   const onRefresh = useRefreshWizardPage(existingDeployment);
-  const {
-    isExitModalOpen,
-    openExitModal,
-    closeExitModal,
-    handleExitConfirm: baseHandleExitConfirm,
-    exitWizardOnSubmit,
-  } = useExitDeploymentWizard({ returnRoute, cancelReturnRoute });
-
-  const handleExitConfirm = React.useCallback(() => {
-    fireModelDeployed({ outcome: TrackingOutcome.cancel }, isEdit);
-    baseHandleExitConfirm();
-  }, [baseHandleExitConfirm, isEdit]);
+  const { isExitModalOpen, openExitModal, closeExitModal, handleExitConfirm, exitWizardOnSubmit } =
+    useExitDeploymentWizard({ returnRoute, cancelReturnRoute, isEdit });
 
   const isYAMLViewerEnabled = useIsAreaAvailable(SupportedArea.YAML_VIEWER).status;
   const [viewMode, setViewMode] = React.useState<ModelDeploymentWizardViewMode>(
--- a/packages/model-serving/src/components/deploymentWizard/deploying/useModelDeploymentSubmit.ts
+++ b/packages/model-serving/src/components/deploymentWizard/deploying/useModelDeploymentSubmit.ts
@@ -1,4 +1,5 @@
 import React from 'react';
+import { TrackingOutcome } from '@odh-dashboard/ui-core';
 import { useSecretOps } from '@odh-dashboard/plugin-core/host-api';
 import { getServingRuntimeFromTemplate } from '@odh-dashboard/model-serving/shared';
 import { useDeployMethod } from './useDeployMethod';
@@ -13,6 +14,12 @@ import { InitialWizardFormData } from '../../../shared/types/form-data';
 import { WizardFormState } from '../useDeploymentWizardReducer';
 import { ModelDeploymentWizardViewMode } from '../ModelDeploymentWizard';
 import { ExternalDataMap, isExternalDataReady } from '../ExternalDataLoader';
+import {
+  fireModelDeployed,
+  type DeploymentTrackingProperties,
+} from '../../../shared/tracking/deploymentTracking';
+import { useWizardTrackingProperties } from '../../../shared/tracking/useWizardTrackingProperties';
+import { ServingRuntimePlatform } from '../../../shared/types';
 
 /**
  * Get the onSubmit function to create / update the deployment. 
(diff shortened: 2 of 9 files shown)


paulovmr:
FYI Please be sure to rebase this pull request before its approval, to get the latest mock location changes

packages/model-serving/src/components/deploymentWizard/ModelDeploymentWizard.tsx
```diff
@@ -50,9 +51,20 @@ const ModelDeploymentWizard: React.FC<ModelDeploymentWizardProps> = ({
...
+    openExitModal,
+    closeExitModal,
+    handleExitConfirm: baseHandleExitConfirm,
+    exitWizardOnSubmit,
+  } = useExitDeploymentWizard({ returnRoute, cancelReturnRoute });
+
+  const handleExitConfirm = React.useCallback(() => {
+    fireModelDeployed({ outcome: TrackingOutcome.cancel }, isEdit);
+    baseHandleExitConfirm();
+  }, [baseHandleExitConfirm, isEdit]);
```
emilys314: Can we put the event tracking addition inside of the `useExitDeploymentWizard`? This is exit specific logic, which I'd like to keep inside the hook. And not make exit logic leak into the main ModelDeploymentWizard page object

packages/model-serving/src/components/deploymentWizard/deploying/useModelDeploymentSubmit.ts
```diff
@@ -43,10 +50,31 @@ export const useModelDeploymentSubmit = (
   );
   const { runPreDeploy, preDeployExtensionsLoaded } = useWizardFieldPreDeploy(formState);
   const { runPostDeploy, postDeployExtensionsLoaded } = useWizardFieldPostDeploy(formState);
+  const { platformProperties } = useWizardTrackingProperties(
+    formState,
+    deployMethod?.properties.platform,
+  );
```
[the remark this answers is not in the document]
emilys314: Under the hood, the hook is using a use**Resolved**Extensions, which will start loading it right when we enter the page. So I supposed by the time you get here, it most likely would be loaded? But the concern is still valid

I think the best fix is to use a useExtensions (with the Resovled) which will return a promise for you to await while it loads. This way we don't strain our network requests with things that actually matter for time till first render. We can also await the promise which will prevent any race conditions. The only downside is that it would make the submit call longer, but this is fine, since creating a new deployment is a BIG operation and can wait an extra second here I think

packages/model-serving/src/components/deploymentWizard/deploying/useModelDeploymentSubmit.ts
```diff
@@ -43,10 +50,31 @@ export const useModelDeploymentSubmit = (
...
 
+  const isEdit = !!existingDeployment;
+
+  const getBaseTrackingProperties = React.useCallback((): Omit<
+    DeploymentTrackingProperties,
+    'outcome' | 'success' | 'error'
+  > => {
+    const serverTemplateName = formState.modelServer?.data?.selection?.name;
+    return {
+      type: formState.modelType.data?.type ?? ServingRuntimePlatform.SINGLE,
```
emilys314: I think we should remove `ServingRuntimePlatform.SINGLE`
it's a concept we have been moving away from. There only ever is "Single" right now

```suggestion
      type: formState.modelType.data?.type,
```

packages/model-serving/src/shared/tracking/deploymentTracking.ts
```diff
@@ -0,0 +1,27 @@
+import { fireFormTrackingEvent } from '@odh-dashboard/internal/concepts/analyticsTracking/segmentIOUtils';
+import { type FormTrackingEventProperties } from '@odh-dashboard/ui-core';
+
+export enum DeploymentTrackingEvent {
+  MODEL_DEPLOYED = 'Model Deployed',
+  MODEL_UPDATED = 'Model Updated',
+}
+
+export type DeploymentTrackingBaseProperties = FormTrackingEventProperties & {
```
emilys314, line 9: There are more things that are generic deployment properties. For example, there are model location types: connections types (s3, uri, oci), pvcs.

You should double check the wizard to see any other missed fields.

packages/model-serving/src/components/deploymentWizard/deploying/useModelDeploymentSubmit.ts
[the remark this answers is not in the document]
emilys314: A CodeRef is basically a Promise for fetching js code, for example this is done in our start / stop code
```
startStopActionExtension?.properties
        .patchDeploymentStoppedStatus()
        .then((resolvedFunction) => resolvedFunction(deployment, true));
```

packages/model-serving/src/components/deploymentWizard/deploying/useModelDeploymentSubmit.ts
emilys314: Also "Generative" vs "Predictive" (and now "NIM") is not a platform (well, NIM kinda is). So it doesn't match the "Single"

packages/model-serving/src/components/deploymentWizard/deploying/useModelDeploymentSubmit.ts
```diff
@@ -66,14 +65,15 @@ export const useModelDeploymentSubmit = (
...
     return {
-      type: formState.modelType.data?.type ?? ServingRuntimePlatform.SINGLE,
+      type: formState.modelType.data?.type,
       runtime: serverTemplateName,
       servingRuntimeName: formState.modelServer?.data?.selection?.label,
       servingRuntimeFormat: formState.modelFormatState.modelFormat?.name,
       numReplicas: formState.numReplicas.data ?? undefined,
-      ...platformProperties,
+      modelLocationType: formState.modelLocationData.data?.type,
+      connectionType: formState.modelLocationData.data?.connectionTypeObject?.metadata.name,
```
emilys314: Does `modelLocationType` cover `connectionType`? I know the data structure is a little confusing so I'm not sure off the top of my head. If you select an s3 connection, does it do this?
```
modelLocationType: "s3",
connectionType: "s3",
```

Also `connectionType` would be undefined if there is a pvc

We might not need the `connectionType` property?

packages/model-serving/src/components/deploymentWizard/deploying/useModelDeploymentSubmit.ts
mturley: +1, also the old event was using `type` to track the platform and now this uses `type` to track generative/predictive. That’s preferable, but I wonder if it’ll be an issue in the reports that this field changed, maybe we should rename it to `modelType`?

packages/model-serving/src/components/deploymentWizard/deploying/useModelDeploymentSubmit.ts
```diff
@@ -105,9 +132,32 @@ export const useModelDeploymentSubmit = (
...
-        setSubmitError(error instanceof Error ? error : new Error(String(error)));
+        const errorMessage = error instanceof Error ? error.message : String(error);
+        setSubmitError(error instanceof Error ? error : new Error(errorMessage));
+
+        fireModelDeployed(
+          {
+            outcome: TrackingOutcome.submit,
+            success: false,
+            error: errorMessage,
+            ...getBaseTrackingProperties(),
```
mturley, line 163: The old event from [ManageKServeModal.tsx:339](https://github.com/opendatahub-io/odh-dashboard/blob/main/frontend/src/pages/modelServing/screens/projects/kServeModal/ManageKServeModal.tsx#L339) had the `errorMessage` property and your new one has `error` here. I wonder if parity with the old event’s property names (where applicable) matters for Amplitude reports (error rate, etc)?

YuliaKrimerman: you're right, I'll address it

packages/model-serving/src/components/deploymentWizard/deploying/useModelDeploymentSubmit.ts
```diff
@@ -112,9 +139,32 @@ export const useModelDeploymentSubmit = (
...
+
+        fireModelDeployed(
+          {
+            outcome: TrackingOutcome.submit,
+            success: true,
+            ...getBaseTrackingProperties(),
+            ...(await getTrackingProperties()),
+          },
+          isEdit,
+        );
```
mturley, line 151: (Stanard apology for copypasting claude verbatim but this seems valid)

**Correctness: a throwing tracking extension here fires a false failure event for a successful deploy.**

`await getTrackingProperties()` runs *after* `deployModel` has already resolved, but it's still inside the deploy `try`. `getTrackingProperties()` resolves a spoke-provided `CodeRef` and calls its `getProperties()` — neither is guarded. If either throws (bad spoke extension, CodeRef load failure), the `await` rejects and control jumps to the `catch` block, which:

- never fires the `success: true` event, and
- fires a `success: false` / `errorMessage` **failure** event for a deployment that actually succeeded, and
- calls `setSubmitError(...)`, so the user sees an error screen even though their model deployed fine.

So a purely analytics-side failure corrupts both the telemetry (success recorded as failure) and the UX. The same unguarded call on the failure path (line ~164) would also mask the real deploy error with the extension's error.

Suggested fix — resolve the platform props defensively so extension failures can't affect the deploy outcome, e.g. inside `useWizardTrackingProperties`:

```ts
try {
  const resolvedFn = await matchingExtension.properties.getProperties();
  return resolvedFn(wizardState);
} catch {
  return {};
}
```

(or compute them before `deployModel` / outside the deploy `try`). Worth a regression test for a throwing extension — the current specs only cover the resolve paths.

packages/model-serving/src/components/deploymentWizard/deploying/useModelDeploymentSubmit.ts
```diff
@@ -45,10 +51,31 @@ export const useModelDeploymentSubmit = (
...
 
+  const isEdit = !!existingDeployment;
+
+  const getBaseTrackingProperties = React.useCallback((): Omit<
+    DeploymentTrackingProperties,
+    'outcome' | 'success' | 'error'
+  > => {
+    const serverTemplateName = formState.modelServer?.data?.selection?.name;
+    return {
+      type: formState.modelType.data?.type,
```
mturley: Per [earlier thread](https://github.com/opendatahub-io/odh-dashboard/pull/9251#discussion_r3767468859), should we maybe change `type` to `modelType` in this emitted event because it’s a different thing to what the old `type` was tracking? I’m not sure if that matters in Amplitude but it can’t hurt

mturley (changes requested):
One new thing and a +1 to an old thing

mturley (approved):
/approve
/lgtm

Tested in dev mode and saw expected console.log output for the stubbed events.

PR #9253, opened 2026-08-11
fix(RHOAIENG-81952): remove SSA-orphaned containers during upgrade

liday-rh:
https://issues.redhat.com/browse/RHOAIENG-81952

## Description

During RHOAI 2.25→3.5 upgrades, the `rhods-dashboard` Deployment ends up with **both** `oauth-proxy` and `kube-rbac-proxy` containers, causing a port 8443 collision and CrashLoopBackOff.

**Root cause:** The old operator (field manager `dashboard`) and new operator (field manager `dashboard-operator`) both apply containers via SSA. Because they use different field manager names, Kubernetes merges rather than replaces the containers list. The stale `oauth-proxy` owned by the old manager persists because SSA's `ForceOwnership` only steals fields that are *present* in the new manifest — it cannot remove fields it doesn't mention.

**Fix:** Add a pre-deploy step (`removeStaleContainers`) that runs before `deployer.Deploy()` in both `reconcileSidecar` and `reconcileStandalone`. It compares live Deployment container names against the rendered manifest and issues a Strategic Merge Patch with `$patch: delete` to remove any containers present in the cluster but absent from the desired spec.

This follows the same pattern as the existing `sanitizeDeploymentProbes` pre-deploy step and is generic — not hardcoded to `oauth-proxy` — so it handles any container removal across future upgrades.

### Evidence from live cluster

Managed fields on `rhods-dashboard` Deployment showing the split:

| Field Manager | SSA Operation | Owns containers |
|---|---|---|
| `dashboard` (old 2.25) | Apply | `rhods-dashboard`, `oauth-proxy`, `model-registry-ui` |
| `dashboard-operator` (new 3.5) | Apply | `rhods-dashboard`, `kube-rbac-proxy`, `core-bff` |

Result: 5 containers in the pod spec, `kube-rbac-proxy` crashes with `listen tcp 0.0.0.0:8443: bind: address already in use`.

## How Has This Been Tested?

- Unit tests covering all code paths (8 tests)
- `make test` — all tests pass
- `make lint` — 0 issues
- Verified the SSA field manager split on live cluster `caas-chamb-0400` via `--show-managed-fields`

## Test Impact

Added `stale_containers_test.go` with 8 unit tests:
- No live Deployment → no-op
- No stale containers → no patch
- One stale container (oauth-proxy) → removed
- Multiple stale containers → all removed
- Desired has new containers not in live → no effect
- Skips non-Deployment resources
- `buildStaleContainerPatch` returns nil when no stale
- `buildStaleContainerPatch` builds correct `$patch: delete` entries

## Request review criteria:

## Summary by CodeRabbit

* **Bug Fixes**
  * Automatically removes outdated containers from Kubernetes deployments before applying updates.
  * Prevents stale OAuth and other obsolete containers from remaining in deployed resources.
  * Reports provisioning failures clearly when stale-container cleanup cannot be completed.
  * Safely skips deployments that no longer exist.

### DIFF
--- a/dashboard-operator/internal/controller/dashboard_reconciler.go
+++ b/dashboard-operator/internal/controller/dashboard_reconciler.go
@@ -285,6 +285,14 @@ func (r *DashboardReconciler) reconcileSidecar(
 		return ctrl.Result{}, fmt.Errorf("failed to sanitize deployment probes: %w", err)
 	}
 
+	if err := removeStaleContainers(ctx, r.Client, allResources); err != nil {
+		cm.MarkFalse(string(common.ConditionTypeProvisioningSucceeded),
+			conditions.WithReason("StaleContainerRemovalFailed"),
+			conditions.WithError(err))
+
+		return ctrl.Result{}, fmt.Errorf("failed to remove stale containers: %w", err)
+	}
+
 	deployer := deploy.NewDeployer(
 		deploy.WithFieldOwner("dashboard-operator"),
 		deploy.WithLabel(labels.PlatformPartOf, strings.ToLower(v1alpha1.DashboardKind)),
@@ -447,6 +455,13 @@ func (r *DashboardReconciler) reconcileStandalone(
 			conditions.WithReason("ProbeSanitizeFailed"),
 			conditions.WithError(err))
 		return ctrl.Result{}, fmt.Errorf("failed to sanitize deployment probes: %w", err)
+	}
+
+	if err := removeStaleContainers(ctx, r.Client, allResources); err != nil {
+		cm.MarkFalse(string(common.ConditionTypeProvisioningSucceeded),
+			conditions.WithReason("StaleContainerRemovalFailed"),
+			conditions.WithError(err))
+		return ctrl.Result{}, fmt.Errorf("failed to remove stale containers: %w", err)
 	}
 
 	deployer := deploy.NewDeployer(
--- a/dashboard-operator/internal/controller/stale_containers.go
+++ b/dashboard-operator/internal/controller/stale_containers.go
@@ -1,0 +1,106 @@
+package controller
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+// removeStaleContainers patches live Deployments to delete containers that
+// exist in the cluster but are absent from the desired manifests. This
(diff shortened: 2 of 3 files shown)


paulovmr:
FYI Please be sure to rebase this pull request before approving it, to get the latest mock location changes

dashboard-operator/internal/controller/stale_containers.go
```diff
@@ -0,0 +1,122 @@
...
+		if err := cli.Patch(ctx, live, client.RawPatch(types.StrategicMergePatchType, patchBytes)); err != nil {
+			return fmt.Errorf("patching stale containers for %s: %w", key, err)
+		}
+	}
+
+	return nil
+}
+
+func buildStaleContainerPatch(live, desired *unstructured.Unstructured) (map[string]interface{}, error) {
+	liveContainers, liveFound, err := unstructured.NestedSlice(live.Object, "spec", "template", "spec", "containers")
```
lucferbux, line 64: This only walks `spec.template.spec.containers`. The same SSA field-manager split could theoretically affect `initContainers` if a future upgrade swaps them under different managers. Not an issue today since the operator doesn't manage any init containers, but worth noting -- `sanitizeDeploymentProbes` has the same gap.

dashboard-operator/internal/controller/dashboard_reconciler.go
```diff
@@ -285,6 +285,14 @@ func (r *DashboardReconciler) reconcileSidecar(
 		return ctrl.Result{}, fmt.Errorf("failed to sanitize deployment probes: %w", err)
 	}
 
+	if err := removeStaleContainers(ctx, r.Client, allResources); err != nil {
+		cm.MarkFalse(string(common.ConditionTypeProvisioningSucceeded),
+			conditions.WithReason("StaleContainerRemovalFailed"),
+			conditions.WithError(err))
+
```
lucferbux, line 292: nit: blank line between `MarkFalse` and `return` here but not in the `reconcileStandalone` equivalent (~line 463). Minor, but the two blocks should match since they're the same error pattern.

lucferbux (approved):
Went through the fix against RHOAIENG-81952. The approach is the right one -- pre-deploy SMP with `$patch: delete` to strip SSA-orphaned containers before the deployer runs. Follows the same pattern as `sanitizeDeploymentProbes`, and the fact that it's generic (not hardcoded to `oauth-proxy`) means it'll handle future container swaps on subsequent major upgrades too.

The SMP patch shape is correct -- `containers` uses `name` as the strategic merge key, so `{"$patch": "delete", "name": "oauth-proxy"}` targets exactly the right entry. The ordering in the reconcile pipeline (probes first, stale containers second, then SSA deploy) is also sound since they touch orthogonal fields.

**CodeRabbit comment:** Resolved. `buildStaleContainerPatch` now returns `(map[string]interface{}, error)` with proper `NestedSlice` validation, including the empty/missing desired containers guard. Tests added for both edge cases.

**Preflight nits:** Both are fair (Jira ref in code comment, blank line inconsistency) but non-blocking.

A couple of things inline, nothing blocking.

PR #9256, opened 2026-08-11
feat(RHOAIENG-79983): Replace flat dropdowns with tree-select for API groups and resources

srtanish1992:
https://issues.redhat.com/browse/RHOAIENG-79983

## Description

Converts the Resource types and API groups dropdowns in the Add Rule modal (custom role creation form) to use a tree-style selection with categorized options, wildcard support, and custom entry capability.

### Changes:
- **ResourcesTreeSelect**: New component with categorized resources (Core, Applications, Storage, Networking, RBAC), "All resources" (`*`) wildcard, per-category select-all, tree-aware search filtering, and visual indentation
- **ApiGroupsTreeSelect**: Same enhancements for API groups (Core, Apps & Workloads, Storage, Networking, RBAC, Observability)
- **MultiSelection enhancements**:
  - `chipOnly` property — options that appear as chips but not in the dropdown list (for custom entries)
  - `hideChip` property — suppresses chips for synthetic category options
  - `className` pass-through to `SelectOption` for custom styling
- **resourceCategories.ts / apiGroupCategories.ts**: Mapping files categorizing K8s resources and API groups with user-friendly labels
- **AddRuleModal simplified**: Uses `string[]` state instead of `SelectionOptions[]`, delegates rendering to the new tree-select components

### Key behaviors:
- Tree-aware search: typing a category name reveals all its children
- Custom entries: users can type resources/API groups not in the tree; they show as chips only (not in dropdown)
- Wildcard (`*`): selecting "All" replaces everything; deselecting clears all
- Category select-all: clicking a category header selects/deselects all its resources
- API discovery integration: only resources/groups found on the cluster are shown; full mapping shown as fallback if discovery fails

## How Has This Been Tested?

- Unit tests for `ResourcesTreeSelect` (14 tests) and `ApiGroupsTreeSelect` (13 tests) covering selection, deselection, wildcards, category toggling, search filtering, and custom entries
- `MultiSelection` existing tests pass (16 tests) — backward compatibility verified
- All consumer components tested and unaffected
- Manual testing on dev server with live cluster

## Test Impact

- Added `ResourcesTreeSelect.spec.tsx` (14 tests)
- Added `ApiGroupsTreeSelect.spec.tsx` (13 tests)
- Existing `MultiSelection.spec.tsx` tests pass unchanged (backward compatible)

## Request review criteria:

If you have UI changes:

## Summary by CodeRabbit

* **New Features**
  * Added searchable, grouped selectors for API groups and resources in project role rules.
  * Supports selecting all items or categories, creating custom entries, and preserving custom selections.
  * Added clearer labels, indentation, checkboxes, scrolling, and accessibility support.
* **Bug Fixes**
  * Improved filtering using discovered API data, descriptions, and uncategorized entries.
  * Improved handling of hidden and chip-only selections, including clearing and selection state.
  * Preserved wildcard selections when saving rules.


frontend/src/pages/projects/projectRoles/ApiGroupsTreeSelect.tsx
```diff
@@ -0,0 +1,221 @@
...
+      groupedValues={groupedValues}
+      setValue={handleSetValue}
+      filterFunction={filterFunction}
+      placeholder="Enter or select API groups"
+      ariaLabel="API groups"
+      toggleTestId="api-groups-select-toggle"
+      hasCheckbox
+      isScrollable
+      isCreatable
+      createOptionMessage={(newValue) => `Add custom API group "${newValue}"`}
```
DaoDaoNoCode: **Blocker:** This breaks every Cypress mocked spec that exercises the Add Rule modal's API groups field.

- `toggleTestId="api-groups-select-toggle"` (line 212) no longer matches `packages/cypress/cypress/pages/projectRoles.ts:213-215`'s `findRuleApiGroupsToggle()`, which still queries `rule-api-groups-toggle`. Every `cy.findByTestId(...)` call through that page-object method will time out.
- `createOptionMessage={(newValue) => `Add custom API group "${newValue}"`}` (line 216) no longer matches the wording the specs assert on, e.g. `rolesTabRules.cy.ts:23`: `cy.contains(`Use custom API group "${apiGroup}"`).click()`.

Together these break 9 `it()` blocks in `rolesTabRules.cy.ts`, plus tests in `rolesTabCreate.cy.ts`, `rolesTabTemplate.cy.ts`, and `rolesTabYamlToggle.cy.ts`. Please either update the Cypress page object/specs to the new test-id and wording, or keep the previous `rule-api-groups-toggle` id / `Use custom API group "..."` wording.

frontend/src/pages/projects/projectRoles/ResourcesTreeSelect.tsx
```diff
@@ -0,0 +1,215 @@
...
+      groupedValues={groupedValues}
+      setValue={handleSetValue}
+      filterFunction={filterFunction}
+      placeholder="Enter or select resource types"
+      ariaLabel="Resource types"
+      toggleTestId="resources-select-toggle"
+      hasCheckbox
+      isScrollable
+      isCreatable
+      createOptionMessage={(newValue) => `Add custom resource "${newValue}"`}
```
DaoDaoNoCode: **Blocker:** Same issue as `ApiGroupsTreeSelect.tsx` — this breaks every Cypress mocked spec that exercises the Add Rule modal's resource types field.

- `toggleTestId="resources-select-toggle"` (line 206) no longer matches `packages/cypress/cypress/pages/projectRoles.ts:217-219`'s `findRuleResourceTypesToggle()`, which still queries `rule-resource-types-toggle`.
- `createOptionMessage={(newValue) => `Add custom resource "${newValue}"`}` (line 210) no longer matches `cy.contains(`Use custom resource type "${resource}"`)` used throughout `rolesTabRules.cy.ts`, `rolesTabCreate.cy.ts`, `rolesTabTemplate.cy.ts`, and `rolesTabYamlToggle.cy.ts`.

Same fix direction as the API groups toggle: reconcile the page object/specs with the new id and wording, or restore the old ones.

frontend/src/pages/projects/projectRoles/ApiGroupsTreeSelect.tsx
```diff
@@ -0,0 +1,221 @@
...
+          .filter((o) => o.selected)
+          .map((o) => (String(o.id) === CORE_GROUP_ID ? '' : String(o.id))),
+      );
+
+      for (const catOption of categoryAllOptions) {
+        const categoryId = String(catOption.id).replace(ALL_CATEGORY_PREFIX, '');
+        const category = availableCategories.find((c) => c.id === categoryId);
+        if (!category) {
+          continue;
+        }
```
DaoDaoNoCode: **Suggestion:** A custom API group whose typed name starts with the internal `all-category-` prefix (e.g. `all-category-foo`) is silently dropped instead of added.

`MultiSelection`'s creatable-option guard (`MultiSelection.tsx:201-220`) only blocks input that exactly matches an existing option's `id`/`name`, not a prefix, so `all-category-foo` is offered as a valid "Add custom API group" suggestion and can be selected. Once selected: it's excluded from `realOptions` here (line 120, `!String(o.id).startsWith(ALL_CATEGORY_PREFIX)`) *and* it fails to match any real category in the loop below (line 135, `availableCategories.find(c => c.id === 'foo')` → `undefined` → `continue` at line 137). It never reaches the `selected` set, so `onSelectedApiGroupsChange` fires without it — no chip appears, no error is shown.

Low real-world likelihood since no k8s API group name looks like this, but worth guarding against — e.g. discriminate by a dedicated marker (or `option.chipOnly`) instead of a string prefix that shares the same namespace as arbitrary user input.

frontend/src/pages/projects/projectRoles/ResourcesTreeSelect.tsx
```diff
@@ -0,0 +1,215 @@
...
+      );
+
+      const selected = new Set(realOptions.filter((o) => o.selected).map((o) => String(o.id)));
+
+      for (const catOption of categoryAllOptions) {
+        const categoryId = String(catOption.id).replace(ALL_CATEGORY_PREFIX, '');
+        const category = availableCategories.find((c) => c.id === categoryId);
+        if (!category) {
+          continue;
+        }
```
DaoDaoNoCode: **Suggestion:** Same issue as `ApiGroupsTreeSelect.tsx` — a custom resource name typed with the internal `all-category-` prefix (e.g. `all-category-widgets`) is silently dropped.

`MultiSelection`'s creatable guard only blocks exact id/name matches, so this string is offered as a valid custom entry. Once selected, it's excluded from `realOptions` (line 120) and fails to resolve to a real category in the loop below (line 131, `find(c => c.id === 'widgets')` → `undefined` → `continue` at line 133), so it's dropped from `onSelectedResourcesChange` with no feedback to the user.

Same fix direction: discriminate internal category-toggle options from user-typed custom entries by something other than string-prefix sniffing on shared namespace.

frontend/src/pages/projects/projectRoles/ApiGroupsTreeSelect.tsx
```diff
@@ -0,0 +1,221 @@
...
+        } else if (String(option.id).startsWith(ALL_CATEGORY_PREFIX)) {
+          flushCategory();
+          currentCategory = option;
+          currentCategoryGroups = [];
+          categoryMatches = option.name.toLowerCase().includes(lower);
+        } else {
+          currentCategoryGroups.push(option);
+        }
+      }
+      flushCategory();
```
DaoDaoNoCode: **Suggestion:** Searching for text that matches only a previously-added custom (chip-only) entry renders an empty, childless category header in the dropdown.

Custom entries are appended after every real category in `groupedValues` (lines 90-95) with no header of their own. This loop buckets anything that isn't the wildcard or a `all-category-*` header into whichever category was last seen (`else` branch, line 195) — so a leaked custom entry gets attributed to the *last* category (RBAC). In `flushCategory` (lines 168-182), the leaked match makes `groupMatches.length > 0` true, so `result.push(currentCategory, ...groupMatches)` (line 180) pushes the RBAC header into the filtered results. `MultiSelection` (`MultiSelection.tsx:178`) then strips the `chipOnly` entry afterward but not the header, leaving a visibly empty "RBAC" section mid-search.

Fix direction: skip `chipOnly` options in the `else` branch here (line 194-196) so they're never bucketed into a category's match set.

frontend/src/pages/projects/projectRoles/ResourcesTreeSelect.tsx
```diff
@@ -0,0 +1,215 @@
...
+        } else if (String(option.id).startsWith(ALL_CATEGORY_PREFIX)) {
+          flushCategory();
+          currentCategory = option;
+          currentCategoryResources = [];
+          categoryMatches = option.name.toLowerCase().includes(lower);
+        } else {
+          currentCategoryResources.push(option);
+        }
+      }
+      flushCategory();
```
DaoDaoNoCode: **Suggestion:** Same filter-leak as `ApiGroupsTreeSelect.tsx` — searching text that matches only a custom (chip-only) resource entry renders an empty "RBAC" category header.

Custom entries appended after all categories (lines 88-93) get bucketed into the last-seen category by this loop's `else` branch (line 189), so a match on a leaked custom entry triggers `flushCategory` (lines 164-176) to push the category header even though no real resource in that category matched. `MultiSelection` strips the custom entry but not the header afterward.

Same fix direction: exclude `chipOnly` options from this bucketing before they can inflate `resourceMatches`.

frontend/src/pages/projects/projectRoles/ResourcesTreeSelect.tsx
```diff
@@ -0,0 +1,215 @@
...
+  const availableCategories = React.useMemo(
+    () =>
+      RESOURCE_CATEGORIES.map((category) => ({
+        ...category,
+        resources: category.resources.filter(
+          (r) => discoveredResourceNames.has(r.name) || discoveredResourceNames.size === 0,
+        ),
+      })).filter((category) => category.resources.length > 0),
+    [discoveredResourceNames],
+  );
```
DaoDaoNoCode, line 44: **Suggestion:** Expanding the `*` wildcard via a category toggle can silently grant less access than intended on a cluster with incomplete resource discovery.

`availableCategories` — used to build every option in `groupedValues` and `handleSetValue` — is bounded by `discoveredResourceNames` (line 41). On a modular RHOAI/ODH cluster missing an optional CRD (e.g. `hardwareprofiles` from `infrastructure.opendatahub.io`, or `notebooks` from `kubeflow.org` if that operator isn't installed), those resource types never appear here. When `isAllSelected` is true and a user deselects one category, `handleSetValue` (lines 118-145) rebuilds the explicit resource list purely from these discovery-bounded options, silently omitting any hardcoded resource type missing from discovery. A user who means "everything except Core" ends up with "everything currently discoverable except Core" — a materially narrower grant, with no indication anything was dropped. (The `discoveredResourceNames.size === 0` fallback on line 41 only helps when discovery is *totally* empty, not partially populated.)

frontend/src/pages/projects/projectRoles/ApiGroupsTreeSelect.tsx
```diff
@@ -0,0 +1,221 @@
...
+  const availableCategories = React.useMemo(
+    () =>
+      API_GROUP_CATEGORIES.map((category) => ({
+        ...category,
+        groups: category.groups.filter(
+          (g) => discoveredApiGroups.has(g.name) || discoveredApiGroups.size === 0,
+        ),
+      })).filter((category) => category.groups.length > 0),
+    [discoveredApiGroups],
+  );
```
DaoDaoNoCode, line 45: **Suggestion:** Same risk as the sibling `ResourcesTreeSelect` — expanding `*` via a category toggle can silently narrow the grant on a cluster with incomplete API group discovery.

`availableCategories` is bounded by `discoveredApiGroups` (line 42). If, e.g., `kubeflow.org` or `infrastructure.opendatahub.io` isn't discovered (operator not installed), it's excluded from every option built here. When `*` is selected and a user deselects one category, `handleSetValue` (lines 102-154) rebuilds the explicit group list purely from this discovery-bounded set, silently dropping any hardcoded group missing from discovery — a user intending "everything except Core" gets "everything currently discoverable except Core" with no indication anything was omitted.

DaoDaoNoCode (commented):
Automated review pass focused on functional risk (not style). Filed as separate inline threads below.

frontend/src/components/MultiSelection.tsx
```diff
@@ -238,7 +241,10 @@ export const MultiSelection: React.FC<MultiSelectionProps> = ({
     return options;
   }, [groupOptions, selectOptions, createOption, isCreateOptionOnTop]);
 
-  const selected = React.useMemo(() => allOptions.filter((v) => v.selected), [allOptions]);
+  const selected = React.useMemo(
+    () => allOptions.filter((v) => v.selected && !v.hideChip),
+    [allOptions],
+  );
```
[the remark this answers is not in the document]
srtanish1992: Fixed in 53db765. Separated `selected` into `visibleChips` (for chip rendering) and `hasSelections` (for `selectionRequired` validation). Now `hideChip` options are correctly excluded from chips but still count toward validation state.

frontend/src/pages/projects/projectRoles/__tests__/ApiGroupsTreeSelect.spec.tsx
```diff
@@ -0,0 +1,238 @@
...
+    await openDropdown();
+
+    const menuItem = screen.getByTestId('select-multi-typeahead-Networking');
+    await act(async () => {
+      fireEvent.click(within(menuItem).getByText('Networking'));
+    });
+
+    expect(mockOnChange).toHaveBeenCalledWith(
+      expect.arrayContaining(['networking.k8s.io', 'k8s.cni.cncf.io']),
+    );
```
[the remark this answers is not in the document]
srtanish1992: Fixed in 53db765. Replaced `expect.arrayContaining` with exact array assertion: `expect(mockOnChange).toHaveBeenCalledWith(['networking.k8s.io', 'k8s.cni.cncf.io'])`.

frontend/src/pages/projects/projectRoles/__tests__/ApiGroupsTreeSelect.spec.tsx
```diff
@@ -0,0 +1,238 @@
...
+
+    const combobox = screen.getByRole('combobox', { name: 'API groups' });
+    await act(async () => {
+      fireEvent.change(combobox, { target: { value: 'networking' } });
+    });
+
+    expect(screen.getByText('networking.k8s.io')).toBeInTheDocument();
+    expect(screen.getByText('k8s.cni.cncf.io')).toBeInTheDocument();
+    expect(screen.queryByText('apps')).not.toBeInTheDocument();
+  });
```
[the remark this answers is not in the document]
srtanish1992: Fixed in 53db765. Added two core group round-trip tests: one verifying that selecting the 'core' option emits `['']`, and another verifying that `selectedApiGroups: ['']` renders the chip correctly.

frontend/src/pages/projects/projectRoles/__tests__/ResourcesTreeSelect.spec.tsx
```diff
@@ -0,0 +1,375 @@
...
+    await openDropdown();
+
+    const menuItem = screen.getByTestId('select-multi-typeahead-Networking');
+    await act(async () => {
+      fireEvent.click(within(menuItem).getByText('Networking'));
+    });
+
+    expect(mockOnChange).toHaveBeenCalledWith(
+      expect.arrayContaining(['networkpolicies', 'ingresses']),
+    );
```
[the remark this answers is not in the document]
srtanish1992: Fixed in 53db765. Replaced with exact assertion: `expect(mockOnChange).toHaveBeenCalledWith(['networkpolicies', 'ingresses'])`.

frontend/src/pages/projects/projectRoles/__tests__/ResourcesTreeSelect.spec.tsx
```diff
@@ -0,0 +1,375 @@
...
+    expect(screen.getByLabelText('Remove All resources')).toBeInTheDocument();
+
+    // Clicking a resource when * is active should deselect all (since * covers everything)
+    const menuItem = screen.getByTestId('select-multi-typeahead-All-resources');
+    await act(async () => {
+      fireEvent.click(within(menuItem).getByText('All resources'));
+    });
+
+    expect(mockOnChange).toHaveBeenCalledWith([]);
+  });
```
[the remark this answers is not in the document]
srtanish1992: Fixed in 53db765. Renamed misleading test names to accurately describe their assertions.

frontend/src/pages/projects/projectRoles/AddRuleModal.tsx
```diff
@@ -1,132 +1,53 @@
...
-    const resources = extractResources(selectedResources);
+    const resources = selectedResources.includes(ALL_RESOURCES_WILDCARD)
+      ? [ALL_RESOURCES_WILDCARD]
+      : [...selectedResources];
     const verbs = normalizeVerbs(selectedVerbs);
 
     onSave({
       id: existingRule?.id ?? getUniqueId('rule'),
-      apiGroups,
+      apiGroups: [...selectedApiGroups],
```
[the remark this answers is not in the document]
srtanish1992: By design — `ApiGroupsTreeSelect` already sets state to `['*']` when 'All API groups' is selected. The component state is the single source of truth, so by the time `handleSave` reads `selectedApiGroups`, it's already `['*']`. No additional normalization needed.

frontend/src/pages/projects/projectRoles/AddRuleModal.tsx
```diff
@@ -195,23 +107,12 @@ const AddRuleModal: React.FC<AddRuleModalProps> = ({ existingRule, onSave, onClo
...
-              isScrollable
-              hasCheckbox
-              createOptionMessage={(val) => `Use custom resource type "${val}"`}
-              isDisabled={!apiResourcesLoaded && !apiResourcesError}
+            <ResourcesTreeSelect
+              selectedResources={selectedResources}
+              onSelectedResourcesChange={setSelectedResources}
+              apiResourcesData={
+                apiResourcesLoaded ? apiResourcesData : { apiGroups: [], resources: [] }
+              }
```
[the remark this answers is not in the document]
srtanish1992: Out of scope for this PR — filtering resources by selected API groups is tracked separately in [RHOAIENG-80010](https://redhat.atlassian.net/browse/RHOAIENG-80010). The help text will be updated in that ticket.

frontend/src/pages/projects/projectRoles/ApiGroupsTreeSelect.scss
```diff
@@ -0,0 +1,17 @@
...
+  }
+}
+
+.odh-api-group-tree__category {
+  .pf-v6-c-menu__item-text {
+    font-weight: var(--pf-t--global--font--weight--body--bold);
+  }
+
+  padding-inline-start: var(--pf-t--global--spacer--lg);
+}
```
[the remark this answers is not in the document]
srtanish1992: Fixed in 53db765. Simplified SCSS to avoid nesting into `.pf-v6-c-menu__item-text` — now applies `font-weight` directly on the outer class (which PF places on the `<li>`). Added `aria-label` to category headers (e.g., 'Select all Core resources') and child options (e.g., 'Pods (Core)') so screen readers convey the hierarchy.

frontend/src/pages/projects/projectRoles/ApiGroupsTreeSelect.tsx
```diff
@@ -0,0 +1,221 @@
...
+              className: 'odh-api-group-tree__group',
+            })),
+          ];
+        }),
+        ...customEntries.map((g) => ({
+          id: g,
+          name: g,
+          selected: true,
+          chipOnly: true,
+        })),
```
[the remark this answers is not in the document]
srtanish1992: Fixed in 53db765. Category identification in `handleSetValue` now matches against known `availableCategories` IDs instead of `startsWith(ALL_CATEGORY_PREFIX)`. A custom entry like `all-category-foo` is now correctly treated as a real option.

frontend/src/pages/projects/projectRoles/ApiGroupsTreeSelect.tsx
```diff
@@ -0,0 +1,221 @@
...
+            result.push(option);
+          }
+        } else if (String(option.id).startsWith(ALL_CATEGORY_PREFIX)) {
+          flushCategory();
+          currentCategory = option;
+          currentCategoryGroups = [];
+          categoryMatches = option.name.toLowerCase().includes(lower);
+        } else {
+          currentCategoryGroups.push(option);
+        }
```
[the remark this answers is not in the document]
srtanish1992: Fixed in 53db765. Added `!option.chipOnly` guard in the filter function's `else` branch so custom entries are skipped during the category walk and don't leak into the last category.

frontend/src/pages/projects/projectRoles/ApiGroupsTreeSelect.tsx
```diff
@@ -0,0 +1,221 @@
...
+
+      if (isNowAllSelected && !wasAllSelected) {
+        onSelectedApiGroupsChange([ALL_API_GROUPS_WILDCARD]);
+        return;
+      }
+
+      if (!isNowAllSelected && wasAllSelected) {
+        onSelectedApiGroupsChange([]);
+        return;
+      }
```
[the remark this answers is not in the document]
srtanish1992: By design — follows Kubernetes/`oc` semantics where `*` supersedes all individual selections. When `*` is toggled on, it replaces everything with `['*']`. When toggled off, it clears to `[]`. Custom entries are not preserved because in K8s RBAC, `*` already covers them.

frontend/src/pages/projects/projectRoles/resourceCategories.ts
```diff
@@ -0,0 +1,89 @@
...
+      {
+        name: 'persistentvolumeclaims',
+        label: 'Cluster storage (persistentvolumeclaims)',
+        apiGroup: '',
+      },
+      {
+        name: 'persistentvolumeclaims/status',
+        label: 'Cluster storage status (persistentvolumeclaims/status)',
+        apiGroup: '',
+      },
```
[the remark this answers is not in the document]
srtanish1992: Fixed in 53db765. Removed `persistentvolumeclaims/status` from `resourceCategories.ts` — the discovery parser skips resources containing `/`, so this entry could never match and would be silently dropped.

frontend/src/pages/projects/projectRoles/ResourcesTreeSelect.tsx
```diff
@@ -0,0 +1,215 @@
...
+
+  const customEntries = React.useMemo(
+    () =>
+      selectedResources.filter(
+        (r) => r !== ALL_RESOURCES_WILDCARD && !ALL_INDIVIDUAL_RESOURCES.includes(r),
+      ),
+    [selectedResources],
+  );
```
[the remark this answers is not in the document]
srtanish1992: Fixed in 53db765. `customEntries` now filters against `renderedResourceNames` (derived from `availableCategories`) instead of the full static catalog. A resource that's mapped but hidden by discovery will correctly appear as a custom chip when pre-selected.

frontend/src/pages/projects/projectRoles/ResourcesTreeSelect.tsx
```diff
@@ -0,0 +1,215 @@
...
+          id: r,
+          name: r,
+          selected: true,
+          chipOnly: true,
+        })),
+      ],
+    };
+
+    return [allOptions];
+  }, [availableCategories, selectedResources, isAllSelected, customEntries]);
```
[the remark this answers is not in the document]
srtanish1992: Fixed in 53db765. Category identification now matches against known `availableCategories` IDs (`availableCategories.some(c => ...)`) instead of `startsWith(ALL_CATEGORY_PREFIX)`. A custom resource typed as `all-category-widgets` is now correctly treated as a real selection.

frontend/src/pages/projects/projectRoles/ResourcesTreeSelect.tsx
```diff
@@ -0,0 +1,215 @@
...
+
+      if (isNowAllSelected && !wasAllSelected) {
+        onSelectedResourcesChange([ALL_RESOURCES_WILDCARD]);
+        return;
+      }
+
+      if (!isNowAllSelected && wasAllSelected) {
+        onSelectedResourcesChange([]);
+        return;
+      }
```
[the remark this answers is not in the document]
srtanish1992: By design — follows Kubernetes/`oc` semantics. `*` supersedes everything. When toggled off, all selections clear. The test name was misleading and has been renamed in 53db765 to accurately describe the assertion.

frontend/src/pages/projects/projectRoles/ResourcesTreeSelect.tsx
```diff
@@ -0,0 +1,215 @@
...
+            result.push(option);
+          }
+        } else if (String(option.id).startsWith(ALL_CATEGORY_PREFIX)) {
+          flushCategory();
+          currentCategory = option;
+          currentCategoryResources = [];
+          categoryMatches = option.name.toLowerCase().includes(lower);
+        } else {
+          currentCategoryResources.push(option);
+        }
```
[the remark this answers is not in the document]
srtanish1992: Fixed in 53db765. Added `!option.chipOnly` guard in the filter function's `else` branch so custom entries are skipped during the positional walk.

frontend/src/pages/projects/projectRoles/ApiGroupsTreeSelect.tsx
srtanish1992: Fixed in 53db765. Updated `toggleTestId` to `rule-api-groups-toggle` and `createOptionMessage` to `Use custom API group` to match the Cypress page objects.

frontend/src/pages/projects/projectRoles/ResourcesTreeSelect.tsx
srtanish1992: Fixed in 53db765. Updated `toggleTestId` to `rule-resource-types-toggle` and `createOptionMessage` to `Use custom resource type` to match the Cypress page objects.

frontend/src/pages/projects/projectRoles/ApiGroupsTreeSelect.tsx
srtanish1992: Fixed in 53db765. Category identification in `handleSetValue` now uses `availableCategories.some(c => ...)` instead of `startsWith(ALL_CATEGORY_PREFIX)`, so custom entries like `all-category-foo` are correctly treated as real selections.

frontend/src/pages/projects/projectRoles/ResourcesTreeSelect.tsx
srtanish1992: Fixed in 53db765. Same fix applied — `handleSetValue` now matches against known category IDs rather than string prefix.

frontend/src/pages/projects/projectRoles/ApiGroupsTreeSelect.tsx
srtanish1992: Fixed in 53db765. Added `!option.chipOnly` guard in the filter function's `else` branch so `chipOnly` custom entries are skipped during the positional walk and never get bucketed into the last category.

frontend/src/pages/projects/projectRoles/ResourcesTreeSelect.tsx
srtanish1992: Fixed in 53db765. Same `!option.chipOnly` guard applied to `ResourcesTreeSelect`'s filter function.

frontend/src/pages/projects/projectRoles/ResourcesTreeSelect.tsx
srtanish1992: Inherent to the design — the tree only shows discoverable resources. The wildcard `*` is always preserved as-is in the YAML and is never expanded. Category toggles are a UI convenience for selecting known resources individually. If a user needs to grant access to undiscovered resources, they should use the wildcard or type a custom entry.

frontend/src/pages/projects/projectRoles/ApiGroupsTreeSelect.tsx
srtanish1992: Same as above — inherent to the design. The wildcard `*` is preserved in YAML and not expanded. Category toggles are a convenience for selecting known groups individually. Undiscovered groups can be typed as custom entries.

frontend/src/pages/projects/projectRoles/ApiGroupsTreeSelect.tsx
```diff
@@ -0,0 +1,225 @@
...
+  onSelectedApiGroupsChange: (groups: string[]) => void;
+  apiResourcesData: ApiResourcesData;
+};
+
+const ApiGroupsTreeSelect: React.FC<ApiGroupsTreeSelectProps> = ({
+  selectedApiGroups,
+  onSelectedApiGroupsChange,
+  apiResourcesData,
+}) => {
+  const isAllSelected = selectedApiGroups.includes(ALL_API_GROUPS_WILDCARD);
```
d0w, line 29: Not sure completely but perhaps a useEffect should be used here or else the component will rerender whenever api groups are selected

frontend/src/pages/projects/projectRoles/ApiGroupsTreeSelect.tsx
```diff
@@ -0,0 +1,225 @@
...
+
+      const categoryAllOptions = updatedOptions.filter((o) =>
+        availableCategories.some((c) => `${ALL_CATEGORY_PREFIX}${c.id}` === String(o.id)),
+      );
+
+      const selected = new Set(
+        realOptions
+          .filter((o) => o.selected)
+          .map((o) => (String(o.id) === CORE_GROUP_ID ? '' : String(o.id))),
+      );
```
d0w, line 167: Think there is a reusable util function named `extractApiGroups` in `ruleModalUtils.ts` that you can use here. Same with `extractResources`. I'd either remove the util funcdtions or try to reuse them in the new components since they both currently have no usages.

frontend/src/pages/projects/projectRoles/ResourcesTreeSelect.tsx
```diff
@@ -0,0 +1,219 @@
...
+  MultiSelection,
+  GroupSelectionOptions,
+  SelectionOptions,
+} from '#~/components/MultiSelection';
+import { RESOURCE_CATEGORIES, ALL_RESOURCES_WILDCARD } from './resourceCategories';
+import type { ApiResourcesData } from './useApiResources';
+
+import './ResourcesTreeSelect.scss';
+
+const ALL_CATEGORY_PREFIX = 'all-category-';
```
d0w: Could make this a constant since api tree also uses this. ideally if you use a reusable hook this constant can live in there

d0w (commented):
Dropdowns work on my cluster and roles created have the correct group and resource types selected. 

On another note, there seems to be a lot of duplicated code between the api and resource tree select components and I'd suggest a way to consolidate the args into a hook such as useCategoryTreeSelectArgs<type> that would handle things like availableCategories, renderedResourceNames, customEntries, groupedValues, handleSetValue, and filterFunction.

frontend/src/pages/projects/projectRoles/ApiGroupsTreeSelect.tsx
srtanish1992: Fixed in 9dc8cba. `isAllSelected` is a derived constant (computed from `selectedApiGroups.includes(ALL_API_GROUPS_WILDCARD)`) — it does not cause re-renders on its own. The component only re-renders when the parent passes new `selectedApiGroups` props, which is expected behavior for a controlled component.

frontend/src/pages/projects/projectRoles/ApiGroupsTreeSelect.tsx
srtanish1992: Fixed in 9dc8cba. `extractApiGroups` and `extractResources` were dead code (no remaining usages) — removed them from `ruleModalUtils.ts` along with their test blocks. The new tree-select components handle the ID-to-value mapping internally.

frontend/src/pages/projects/projectRoles/ResourcesTreeSelect.tsx
srtanish1992: Fixed in 9dc8cba. Extracted `ALL_CATEGORY_PREFIX` to a new shared utility file `categoryTreeUtils.ts`, along with `createCategoryFilter` and `applyCategoryToggles`. Both tree-select components now import from this shared module, eliminating the duplication. A full hook consolidation was considered but deferred — the two components have enough structural differences (CORE_GROUP_ID mapping, description matching, different category shapes) that a single generic hook would add more complexity than it saves.

frontend/src/pages/projects/projectRoles/ResourcesTreeSelect.tsx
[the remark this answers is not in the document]
srtanish1992: Fixed in 9dc8cba. Changed `ALL_CATEGORY_PREFIX` from `all-category-` to `__all_category__` (uses underscores, which are invalid in Kubernetes resource names per DNS label rules). This makes ID collisions between resource names and category option IDs structurally impossible. Additionally, the filter function now identifies categories by `hideChip && !chipOnly` markers instead of prefix matching. Added a regression test with a resource named `all-category-other` to verify correct behavior.

frontend/src/pages/projects/projectRoles/ApiGroupsTreeSelect.tsx
srtanish1992: Fixed in 9dc8cba. Changed `ALL_CATEGORY_PREFIX` from `all-category-` to `__all_category__`. Since underscores are invalid in Kubernetes API group and resource names (which follow DNS label rules: `[a-z0-9][-a-z0-9]*`), no user-typed custom value can ever collide with internal category IDs.

frontend/src/pages/projects/projectRoles/ResourcesTreeSelect.tsx
srtanish1992: Fixed in 9dc8cba. Same fix as ApiGroupsTreeSelect — `ALL_CATEGORY_PREFIX` now uses underscores (`__all_category__`), making collisions with valid K8s resource names impossible.

frontend/src/pages/projects/projectRoles/ResourcesTreeSelect.tsx
```diff
@@ -0,0 +1,243 @@
...
+        label: 'Other',
+        resources: otherResources.map((r) => ({
+          name: r.name,
+          label: r.name,
+          apiGroup: r.apiGroup,
+        })),
+      });
+    }
+    return categories;
+  }, [availableCategories, otherResources]);
```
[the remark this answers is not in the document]
srtanish1992: Fixed in 9dc8cba. Changed `ALL_CATEGORY_PREFIX` to `__all_category__` (underscores are invalid in K8s resource names). The filter function now identifies categories by `hideChip && !chipOnly` markers instead of prefix matching. Regression test added with a resource named `all-category-other` confirming it selects only that resource, not the entire "Other" category.

srtanish1992:
@d0w I have addressed your feedback. I considered a full hook but the two components have enough differences (CORE_GROUP_ID mapping, description matching, different category shapes) that a generic hook would need heavy parameterization. Additionally, upcoming work will add filtering of resources by selected API groups, which will further differentiate the two components' logic. The utility approach eliminates the duplication while keeping each component straightforward and independently evolvable.

frontend/src/pages/projects/projectRoles/AddRuleModal.tsx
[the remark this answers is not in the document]
srtanish1992: By design — `ApiGroupsTreeSelect` already normalizes the wildcard internally. When the user selects "All API groups", `handleSetValue` emits `['*']` directly (line 131-132 in the current code). The `selectedApiGroups` state in `AddRuleModal` is always already collapsed to `['*']` when the wildcard is active, so it is passed correctly to `onSave` without needing an additional normalize step. This mirrors how `ResourcesTreeSelect` handles `ALL_RESOURCES_WILDCARD`.

frontend/src/pages/projects/projectRoles/AddRuleModal.tsx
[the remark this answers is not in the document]
srtanish1992: Acknowledged — this is tracked as RHOAIENG-80010 (filtering resources by selected API groups). It will be implemented in a follow-up PR to keep this change focused on the tree-select conversion. The current behavior matches the existing Add Rule modal where resources and API groups are selected independently.

frontend/src/pages/projects/projectRoles/categoryTreeUtils.ts
```diff
@@ -0,0 +1,95 @@
...
+      if (String(option.id) === wildcardId) {
+        if (option.name.toLowerCase().includes(lower)) {
+          result.push(option);
+        }
+      } else if (option.hideChip && !option.chipOnly) {
+        flushCategory();
+        currentCategory = option;
+        currentCategoryItems = [];
+        categoryMatches = option.name.toLowerCase().includes(lower);
+      } else if (!option.chipOnly) {
```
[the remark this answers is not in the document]
srtanish1992: Fixed in 0fae70795. Updated the discriminator to use `String(option.id).startsWith(ALL_CATEGORY_PREFIX)` — this is deterministic and does not rely on `hideChip` which becomes `true` on child items when the wildcard is active. Added a test that selects `*`, searches a category name ("Core"), and asserts children (Pods, Services, ConfigMaps) remain visible.

frontend/src/pages/projects/projectRoles/__tests__/ResourcesTreeSelect.spec.tsx
```diff
@@ -0,0 +1,531 @@
...
+      );
+
+      await openDropdown();
+
+      const menuItem = screen.getByTestId('select-multi-typeahead-all-category-other');
+      await act(async () => {
+        fireEvent.click(within(menuItem).getByText('all-category-other'));
+      });
+
+      expect(mockOnChange).toHaveBeenCalledWith(['all-category-other']);
```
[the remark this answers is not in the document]
srtanish1992: Not applicable. The `ALL_CATEGORY_PREFIX` is `__all_category__` (double underscores), not `all-category-`. A resource named `all-category-other` does **not** start with `__all_category__`, so no collision occurs. The `handleSetValue` logic identifies categories using `allCategories.some((c) => \`${ALL_CATEGORY_PREFIX}${c.id}\` === String(o.id))` — exact match against known category IDs, not prefix sniffing on user input. This test is specifically a regression test validating that the collision does NOT happen.

frontend/src/pages/projects/projectRoles/AddRuleModal.tsx
```diff
@@ -170,21 +91,12 @@ const AddRuleModal: React.FC<AddRuleModalProps> = ({ existingRule, onSave, onClo
...
-              isScrollable
-              hasCheckbox
-              createOptionMessage={(val) => `Use custom API group "${val}"`}
-              isDisabled={!apiResourcesLoaded && !apiResourcesError}
+            <ApiGroupsTreeSelect
+              selectedApiGroups={selectedApiGroups}
+              onSelectedApiGroupsChange={setSelectedApiGroups}
+              apiResourcesData={
+                apiResourcesLoaded ? apiResourcesData : { apiGroups: [], resources: [] }
+              }
```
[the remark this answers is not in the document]
srtanish1992: Out of scope for this PR. The `useApiResources` hook already handles fetch failures via its `error` state, and `parseDiscoveryItems` uses defensive access patterns (optional chaining, `filter`, `flatMap`). Adding a full runtime schema validator for K8s aggregated discovery responses is a separate concern — this PR focuses on the tree-select UI components. The existing error fallback (showing the full static catalog) gracefully handles any data issue.

d0w:
> @d0w I have addressed your feedback. I considered a full hook but the two components have enough differences (CORE_GROUP_ID mapping, description matching, different category shapes) that a generic hook would need heavy parameterization. Additionally, upcoming work will add filtering of resources by selected API groups, which will further differentiate the two components' logic. The utility approach eliminates the duplication while keeping each component straightforward and independently evolvable.

makes sense, was thinking the same thing but didn't know exactly how much they would differ

srtanish1992:

DaoDaoNoCode (commented):
Automated review pass — 1 blocker, 2 in-diff suggestions posted inline below. (A 3rd suggestion, about missing schema validation on discovery API responses in `useApiResources.ts`, isn't posted inline since that file isn't part of this PR's diff — it's a pre-existing gap, not introduced here.)

frontend/src/pages/projects/projectRoles/ApiGroupsTreeSelect.tsx
```diff
@@ -0,0 +1,194 @@
...
+
+      const categoryAllOptions = updatedOptions.filter((o) =>
+        allCategories.some((c) => `${ALL_CATEGORY_PREFIX}${c.id}` === String(o.id)),
+      );
+
+      const selected = new Set(
+        realOptions
+          .filter((o) => o.selected)
+          .map((o) => (String(o.id) === CORE_GROUP_ID ? '' : String(o.id))),
```
DaoDaoNoCode: **Blocker:** `CORE_GROUP_ID` (`__core__`) collides with a user-typed custom API group.

`handleSetValue` here does an exact-string match `String(o.id) === CORE_GROUP_ID ? '' : String(o.id)` to convert the display sentinel back to the real core API group (`''`). Nothing stops a user from typing a custom API group literally named `__core__` through the creatable input (`isCreatable` / `createOptionMessage`, lines 188-189) — `MultiSelection`'s creatable guard only blocks exact matches against *existing* option ids, and `__core__` isn't a pre-existing option unless the core group itself is rendered.

If a user does this, their custom entry silently becomes the Kubernetes **core API group** (`''`) instead of the literal string they typed — an incorrect-privilege-assignment bug in the same class CodeRabbit already caught and fixed for the `all-category-` prefix collision (via the underscore-based `__all_category__` prefix + exact-match-against-known-categories). This sentinel was flagged once in review (2026-08-12T21:20:04Z on this same file) and never actually addressed — it looks like it fell through the cracks while the `all-category-` fix landed.

Suggest: pick a `CORE_GROUP_ID` value that can't be entered as a literal K8s-valid string (mirroring how `__all_category__` was chosen), or discriminate the core-group option from custom entries via explicit metadata instead of string-id comparison. Also add a regression test analogous to the `all-category-other` one in `ResourcesTreeSelect.spec.tsx:483-509`.

frontend/src/pages/projects/projectRoles/__tests__/ApiGroupsTreeSelect.spec.tsx
```diff
@@ -0,0 +1,373 @@
...
+        />,
+      );
+
+      await openDropdown();
+
+      const menuItem = screen.getByTestId('select-multi-typeahead-serving-kserve-io');
+      expect(menuItem).toBeInTheDocument();
+    });
+  });
+});
```
DaoDaoNoCode, line 472: **Suggestion:** No regression test covers the `CORE_GROUP_ID` (`__core__`) collision described in the `ApiGroupsTreeSelect.tsx` review comment. `ResourcesTreeSelect.spec.tsx:483-509` has an analogous test for the `all-category-other` collision — once the underlying issue is fixed, add an equivalent test here: type a custom API group equal to the `CORE_GROUP_ID` string and assert it's emitted as itself, not converted to `''`.

packages/cypress/cypress/tests/mocked/projects/tabs/rolesTabCreate.cy.ts
```diff
@@ -18,15 +18,25 @@ import { projectRoles } from '../../../../pages/projectRoles';
...
 
-const addRule = (apiGroup: string, resource: string, verb: string) => {
-  projectRoles.findAddRuleButton().click();
-  projectRoles.findAddRuleModal().should('exist');
+const selectApiGroup = (apiGroup: string) => {
   projectRoles.findRuleApiGroupsToggle().click();
   projectRoles.findRuleApiGroupsToggle().parent().find('input').type(apiGroup);
-  cy.contains(`Use custom API group "${apiGroup}"`).click();
+  cy.findByTestId(`select-multi-typeahead-${apiGroup}`).click();
+  cy.get('body').click(0, 0);
```
DaoDaoNoCode: **Suggestion:** `cy.get('body').click(0, 0)` to close the tree-select dropdown after each selection is a coordinate-based click rather than a targeted assertion. It passes today (verified end-to-end against the current `toggleTestId`/`createOptionMessage` values), but it's brittle to future layout changes — e.g. if something is later rendered at the top-left of the page, this could click it instead of just dismissing the dropdown. Consider closing via `Escape` or a more explicit affordance. Same pattern also appears at line 32 here and lines 23/30 of `rolesTabRules.cy.ts`.

srtanish1992:
Re: [review feedback](https://github.com/opendatahub-io/odh-dashboard/pull/9256#pullrequestreview-4954406632) — all 3 items addressed in b4a6a1e78:

**1. Blocker: `CORE_GROUP_ID` collision** — Fixed. Changed the discriminator in `handleSetValue` to use `className === 'odh-api-group-tree__group'`. Only the real core group option (rendered from the category list with that class) gets converted to `''`. A user-created custom entry via the creatable input won't have this className, so it's preserved as-is.

**2. Regression test** — Added `'should not convert a custom API group named "__core__" to empty string'` in `ApiGroupsTreeSelect.spec.tsx`. Sets `selectedApiGroups: ['__core__', 'apps']`, deselects `apps`, and asserts `__core__` is preserved (not converted to `''`).

**3. Cypress `cy.get('body').click(0, 0)` → Escape** — Replaced all 8 occurrences across `rolesTabCreate`, `rolesTabRules`, `rolesTabTemplate`, and `rolesTabYamlToggle` with targeted Escape key press on the combobox input (`...find('input').type('{esc}')`), which is resilient to layout changes.

frontend/src/pages/projects/projectRoles/ApiGroupsTreeSelect.tsx
DaoDaoNoCode: Re-checked after `b4a6a1e78` and `17ab06570`.

The `name !== id` discriminator in `handleSetValue` (line 158) is correct for the string-conversion path — I traced it against both the built-in Core option (`name: 'core'`, `id: '__core__'`) and a discovered/custom entry literally named `__core__` (`name === id` in both cases), and it distinguishes them properly. The two new regression tests exercise the real selection path, not just the internal function.

However, this only fixes the save-time string mapping. There's a deeper, still-open issue in the same area:

`customEntries` (lines 72-77) doesn't know about the `CORE_GROUP_ID` transform — it's derived purely from `selectedApiGroups` values not matched by real group *names*. If `selectedApiGroups` ever contains the literal string `'__core__'` (reachable today via this feature's own YAML-edit toggle — paste `apiGroups: ["__core__", "apps"]` in the YAML view, switch back to form view — or from a pre-existing rule saved with that value), it becomes a `chipOnly` option with `id: '__core__'`. The built-in Core category option also has `id: CORE_GROUP_ID = '__core__'` (line 106). Both coexist in the same flattened options array handed to `MultiSelection`.

`MultiSelection.tsx`'s `onSelect` matches purely by `normalizeOptionId(option.id) === normalizeOptionId(menuItem.id)` and maps over the *entire* options array, toggling every option sharing that id — there's no dedup anywhere in that file (checked `persistOptions`, `allValues`, `visibleOptions`). So with that state loaded, clicking the "Core" checkbox to grant real core-group access would simultaneously deselect the custom `__core__` entry (or vice versa) — a click changes more than what was clicked, which is the same class of incorrect-privilege-assignment risk, one layer below the part that got fixed.

The typeahead's own creatable-guard (`MultiSelection.tsx` ~203-210) blocks typing `__core__` directly while the Core option is visible, but that guard doesn't apply to state loaded via props (YAML paste / pre-existing rule data), so this remains reachable through a path this PR's own test suite already exercises for other purposes.

Suggest: stop overloading `id` with the raw K8s value for the built-in Core option — namespace/dedupe it (similar to how the `ALL_CATEGORY_PREFIX` collision was eventually closed) so it can never structurally match anything `customEntries` produces, rather than relying on a `name`/`id` comparison after the fact. Still a **Blocker**.

srtanish1992: Fixed in b22aa331b. Renamed `CORE_GROUP_ID` from `'__core__'` to `'__builtin_core_group__'` — the built-in Core option now has a structurally unique ID that can never collide with a custom entry or discovered group in the `MultiSelection` options array.

This closes the deeper issue you identified: even if `selectedApiGroups` contains the literal string `'__core__'` (via YAML paste or pre-existing rule), the custom `chipOnly` entry gets `id: '__core__'` while the built-in Core option gets `id: '__builtin_core_group__'` — different IDs, so `MultiSelection`'s click handler can't cross-toggle them.

The `name !== id` guard in `handleSetValue` remains as defense-in-depth but is no longer the sole protection.

frontend/src/pages/projects/projectRoles/ruleModalUtils.ts
```diff
@@ -1,18 +1,7 @@
-import { SelectionOptions } from '#~/components/MultiSelection';
 import { ALL_VERBS_WILDCARD } from './verbCategories';
 
-export const CORE_GROUP_ID = '__core__';
+export const CORE_GROUP_ID = '__builtin_core_group__';
```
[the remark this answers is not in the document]
srtanish1992: Fixed in 86da406d3. Added `g !== CORE_GROUP_ID` to the `customEntries` filter so the sentinel value can never produce a duplicate option ID in `MultiSelection`.

Previously, if `selectedApiGroups` contained the literal sentinel string (e.g. via YAML paste), `customEntries` would create a `chipOnly` option with the same ID as the built-in Core option — causing `MultiSelection`'s click handler to cross-toggle both. Now the sentinel is excluded from custom entries entirely. Since it contains underscores (invalid in K8s DNS-label API group names), no legitimate API group value is lost.

Regression test added: renders with `selectedApiGroups: ['__builtin_core_group__', 'apps']` and asserts only one option with the core testid exists, confirming no duplicate is created.

frontend/src/pages/projects/projectRoles/ApiGroupsTreeSelect.tsx
DaoDaoNoCode: Re-checked after `b22aa331b` and `86da406d3`.

The `CORE_GROUP_ID` fix is solid — renaming to `__builtin_core_group__` plus explicitly excluding it from `customEntries` (`ApiGroupsTreeSelect.tsx:76`) closes the duplicate-id path I described above. The new test correctly asserts `getAllByTestId('select-multi-typeahead-core')` has length 1, not just an output check. Confirmed.

But this is one instance of a class of bug, and the class isn't closed. The same duplicate-id hazard exists for `ALL_CATEGORY_PREFIX`, in both files:

- `customEntries` (`ApiGroupsTreeSelect.tsx:72-77`, `ResourcesTreeSelect.tsx:75-80`) excludes the wildcard and `CORE_GROUP_ID`, but not anything matching `` `${ALL_CATEGORY_PREFIX}${categoryId}` ``.
- Category header options use exactly that string as their `id` (`ApiGroupsTreeSelect.tsx:98`, `ResourcesTreeSelect.tsx:101`).
- Same reachability as before: `selectedApiGroups`/`selectedResources` containing the literal string `__all_category__core` (or any other category id) via YAML paste becomes a `chipOnly` customEntry sharing that id — `MultiSelection.onSelect` cross-toggles both again.

CodeRabbit flagged this independently (2026-08-18T17:03:03Z, `ApiGroupsTreeSelect.tsx:97-122`). I verified it's also present in `ResourcesTreeSelect.tsx` via the identical pattern, unaddressed there too.

Suggest closing this generally rather than sentinel-by-sentinel: exclude any `customEntries` value that matches an internal id pattern (`CORE_GROUP_ID` or `ALL_CATEGORY_PREFIX`-prefixed) in one guard, in both components. Still a **Blocker**.

srtanish1992: Fixed in 14af1c908. Closed the class holistically — `customEntries` in both `ApiGroupsTreeSelect` and `ResourcesTreeSelect` now excludes any value matching internal ID patterns:

```typescript
selectedApiGroups.filter(
  (g) =>
    g !== ALL_API_GROUPS_WILDCARD &&
    g !== CORE_GROUP_ID &&
    !g.startsWith(ALL_CATEGORY_PREFIX) &&
    !renderedApiGroupNames.has(g),
)
```

Same guard added in `ResourcesTreeSelect`. All internal sentinels use double-underscore patterns (`__builtin_core_group__`, `__all_category__*`) that are invalid in K8s DNS-label names, so no legitimate value is lost. Regression tests added for the `ALL_CATEGORY_PREFIX` collision in both components.

DaoDaoNoCode (approved):
/lgtm

PR #9041, opened 2026-08-03
feat(RHAI-339): add modelCapabilities feature flag and well-known values

manaswinidas:
https://redhat.atlassian.net/browse/RHAI-339

## Description

Foundational work for Model Capabilities in the deployment wizard and deployments table:

- Add `modelCapabilities` as a **dev temporary** feature flag (default `false`), wired through frontend areas, `DashboardCommonConfig` types, mocks, backend defaults, and core-bff dashboard config
- Add `SupportedArea.MODEL_CAPABILITIES` (reliant on `MODEL_SERVING`) so consumers can gate UI with `useIsAreaAvailable`
- Export shared well-known values and annotation key from `@odh-dashboard/model-serving/shared` / `@odh-dashboard/model-serving/shared/modelCapabilities`:
  - `MODEL_CAPABILITIES_ANNOTATION` (`opendatahub.io/model-capabilities`)
  - `WELL_KNOWN_MODEL_CAPABILITIES`: `Vision`, `Transcription`

Intentionally **not** added to the `OdhDashboardConfig` CRD yet. Enable locally via `?devFeatureFlags=modelCapabilities=true`.

Wizard field UI, annotation apply/extract, and deployments table column remain out of scope (follow-up: [RHAI-340](https://redhat.atlassian.net/browse/RHAI-340) / [RHAI-341](https://redhat.atlassian.net/browse/RHAI-341)).

## How Has This Been Tested?

- Ran type-check on `frontend`, `packages/model-serving`, `packages/plugin-core`, and `packages/k8s-core`
- Ran eslint on changed TypeScript files
- Pre-commit lint-staged passed

## Test Impact

No new unit/Cypress tests. This PR only introduces a feature flag and shared constants with no user-facing UI yet; later stories that consume the flag will add coverage.

## Request review criteria:

Self checklist (all need to be checked):
- [ ] The developer has manually tested the changes and verified that the changes work
- [x] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [x] The developer has added tests or explained why testing cannot be added (unit or cypress tests for related changes)
- [x] The code follows our [Best Practices](/docs/best-practices.md) (React coding standards, PatternFly usage, performance considerations)

After the PR is posted & before it merges:
- [ ] The developer has tested their solution on a cluster by using the image produced by the PR to `main`

Made with [Cursor](https://cursor.com)

[...]

PR #9267, opened 2026-08-12
chore(RHOAIENG-80923): Add modelCapabilities feature flag to OdhDashboardConfig CRD

manaswinidas:
https://issues.redhat.com/browse/RHOAIENG-80923

## Description

Adds the `modelCapabilities` tech preview flag to the `OdhDashboardConfig` CRD schema so it can be enabled via the supported cluster configuration path (`spec.dashboardConfig.modelCapabilities`), consistent with other tech preview flags (same pattern as `llmdTemplates` in #8855).

The flag already exists in frontend/backend defaults (false) from #9041; this change only exposes it in the CRD. No default is set on the CRD field — runtime default remains `false` until explicitly enabled:

```yaml
spec:
  dashboardConfig:
    modelCapabilities: true
```

Also updates the existing `toolCalling` CRD field description so it covers both the model catalog and the deployment wizard (stage-2 tool-calling stays behind the same flag).

## How Has This Been Tested?

- Verified `modelCapabilities` is already present and defaults to `false` in `frontend/src/concepts/areas/const.ts` and `backend/src/utils/constants.ts`.
- Confirmed the CRD field follows the existing tech-preview flag pattern (boolean + description, no CRD `default`).
- Diff reviewed against prior CRD flag addition (`llmdTemplates`).
- Updated `toolCalling` description to reflect catalog + deployment wizard usage.

## Test Impact

No new automated tests. This is a CRD schema exposure of an existing boolean flag plus a description-only update to `toolCalling`; behavior is covered by existing UI/tests gated on those flags. Manual cluster validation: set `spec.dashboardConfig.modelCapabilities: true` on OdhDashboardConfig and confirm the flag is accepted by the API.

## Request review criteria:

### DIFF
--- a/manifests/base/crd/odhdashboardconfigs.opendatahub.io.crd.yaml
+++ b/manifests/base/crd/odhdashboardconfigs.opendatahub.io.crd.yaml
@@ -188,10 +188,10 @@ spec:
                       description: 'When true, enables the Agents Catalog page.'
                     toolCalling:
                       type: boolean
-                      description: 'When true, enables tool-calling configuration for supported models in the model catalog.'
+                      description: 'When true, enables tool-calling configuration for supported models in the model catalog and the deployment wizard.'
                     modelCapabilities:
                       type: boolean
-                      description: 'When true, enables model capabilities selection in the deployment wizard and the Capabilities column on the deployments table. Tech Preview.'
+                      description: 'When true, enables model capabilities selection in the deployment wizard and the Capabilities column on the deployments table.'
                     projectRBAC:
                       type: boolean
                       description: 'When true, enables project-level RBAC (Role-Based Access Control) settings.'
--- a/manifests/base/crd/odhdashboardconfigs.opendatahub.io.crd.yaml
+++ b/manifests/base/crd/odhdashboardconfigs.opendatahub.io.crd.yaml
@@ -189,6 +189,9 @@ spec:
                     toolCalling:
                       type: boolean
                       description: 'When true, enables tool-calling configuration for supported models in the model catalog.'
+                    modelCapabilities:
+                      type: boolean
+                      description: 'When true, enables model capabilities selection in the deployment wizard and the Capabilities column on the deployments table. Tech Preview.'
                     projectRBAC:
                       type: boolean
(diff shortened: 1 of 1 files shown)


manifests/base/crd/odhdashboardconfigs.opendatahub.io.crd.yaml
```diff
@@ -189,6 +189,9 @@ spec:
                     toolCalling:
                       type: boolean
                       description: 'When true, enables tool-calling configuration for supported models in the model catalog.'
+                    modelCapabilities:
+                      type: boolean
+                      description: 'When true, enables model capabilities selection in the deployment wizard and the Capabilities column on the deployments table. Tech Preview.'
```
ppadti: Is the Tech Preview at the end is intentional?

manifests/base/crd/odhdashboardconfigs.opendatahub.io.crd.yaml
```diff
@@ -189,6 +189,9 @@ spec:
                     toolCalling:
                       type: boolean
                       description: 'When true, enables tool-calling configuration for supported models in the model catalog.'
```
ppadti: Also, just thinking, should we also need to update this? as we are working on stage - 2 of tool calling on deployment and keeping those changes also behind the same feature flag?

ppadti (changes requested):

manifests/base/crd/odhdashboardconfigs.opendatahub.io.crd.yaml
manaswinidas: Good catch — not intentional to call it out differently from most other flags. Dropped the trailing "Tech Preview." wording so it matches the common CRD description style.

manifests/base/crd/odhdashboardconfigs.opendatahub.io.crd.yaml
manaswinidas: Agreed — updated `toolCalling` to cover the model catalog and the deployment wizard, since stage-2 deployment tool-calling stays behind the same flag.

emilys314 (approved):
Looks good

PR #9271, opened 2026-08-12
fix(RHOAIENG-81952): [v3.5.0] remove SSA-orphaned containers during upgrade

liday-rh:
Cherry-pick of #9253 into `v3.5.0-fixes`.

https://issues.redhat.com/browse/RHOAIENG-81952

## Description

During RHOAI 2.25→3.5 upgrades, the `rhods-dashboard` Deployment ends up with **both** `oauth-proxy` and `kube-rbac-proxy` containers, causing a port 8443 collision and CrashLoopBackOff.

**Root cause:** The old operator (field manager `dashboard`) and new operator (field manager `dashboard-operator`) both apply containers via SSA. Because they use different field manager names, Kubernetes merges rather than replaces the containers list. The stale `oauth-proxy` owned by the old manager persists because SSA's `ForceOwnership` only steals fields that are *present* in the new manifest — it cannot remove fields it doesn't mention.

**Fix:** Add a pre-deploy step (`removeStaleContainers`) that runs before `deployer.Deploy()` in both `reconcileSidecar` and `reconcileStandalone`. It compares live Deployment container names against the rendered manifest and issues a Strategic Merge Patch with `$patch: delete` to remove any containers present in the cluster but absent from the desired spec.

## Cherry-picked commits

- `888768c` fix([RHOAIENG-81952](https://redhat.atlassian.net/browse/RHOAIENG-81952)): remove SSA-orphaned containers during upgrade
- `816bc84` fix: validate container lists in buildStaleContainerPatch

[RHOAIENG-81952]: https://redhat.atlassian.net/browse/RHOAIENG-81952?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dashboard-operator/internal/controller/dashboard_reconciler.go
+++ b/dashboard-operator/internal/controller/dashboard_reconciler.go
@@ -279,6 +279,14 @@ func (r *DashboardReconciler) reconcileSidecar(
 		return ctrl.Result{}, fmt.Errorf("failed to sanitize deployment probes: %w", err)
 	}
 
+	if err := removeStaleContainers(ctx, r.Client, allResources); err != nil {
+		cm.MarkFalse(string(common.ConditionTypeProvisioningSucceeded),
+			conditions.WithReason("StaleContainerRemovalFailed"),
+			conditions.WithError(err))
+
+		return ctrl.Result{}, fmt.Errorf("failed to remove stale containers: %w", err)
+	}
+
 	deployer := deploy.NewDeployer(
 		deploy.WithFieldOwner("dashboard-operator"),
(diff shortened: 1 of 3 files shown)


lucferbux:
@liday-rh cherry-pick the commit in main https://github.com/opendatahub-io/odh-dashboard/commit/ccba1746b3cfa5a237b2d773f556483c608b5767 that contains both fixes so we can have traceability, push taht and I'll approve the changes, thanks a lot!

lucferbux (approved):
/lgtm
/approve

PR #9322, opened 2026-08-14
Fix template missing label project scoped Templates

emilys314:
Closes https://redhat.atlassian.net/browse/RHOAIENG-82816

## Description

The original code would not look inside the project for the Template. When the "Template removed" was added recently, this actually caused a regression for project scoped Templates. This also surfaced for the new NIM legacy wizard deployment flow, because it comes from the same project. (This also causes a regression for legacy nim modal deployments, because they get the template name wrong for some reason?)

The original code was also doing: 
list global `TemplateKind`s + fetch odhDashboardConfig for order + fetch odhDashboardConfig for disablement for EACH row

This PR makes it done only once per page load now

To fetch the project scoped Templates, it's hard to do when in the global models page, because there are mutliple projects. I just made it fall back to fetch for a single Template in the same project if the row couldn't find a global one

Before:
image
After:
image

## How Has This Been Tested?

go to the zaffre emily2 project and toggle the NIM Wizard flag and you should see a kserve deployment with a project scoped runtime + the NIM deployments

## Test Impact

Jest tests added to cover the different component states

## Request review criteria:

If you have UI changes: 

## Summary by CodeRabbit

- **New Features**
  - Serving details now identify the runtime template used, including whether it is globally or project scoped.
  - Added runtime version status labels, resource details, and clearer indicators for removed templates.
  - Improved loading and error handling while template information is retrieved.

- **Bug Fixes**
  - Prevented temporary display flicker while serving-template data is loading.
  - Improved handling when serving runtimes or templates are unavailable.

### DIFF
--- a/packages/kserve/src/components/__tests__/deploymentServingDetails.spec.tsx
+++ b/packages/kserve/src/components/__tests__/deploymentServingDetails.spec.tsx
@@ -1,0 +1,127 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+import type { FetchStateObject } from '@odh-dashboard/ui-core/hooks/useFetch';
+import type { TemplateKind } from '@odh-dashboard/k8s-core';
+import type { ServingRuntimeKind } from '@odh-dashboard/model-serving/shared';
+import { mockServingRuntimeK8sResource } from '@odh-dashboard/model-serving/__mocks__/mockServingRuntimeK8sResource';
+import { mockServingRuntimeTemplateK8sResource } from '@odh-dashboard/model-serving/__mocks__/mockServingRuntimeTemplateK8sResource';
+// eslint-disable-next-line @odh-dashboard/no-restricted-imports
+import { SERVING_RUNTIME_SCOPE } from '@odh-dashboard/internal/pages/modelServing/screens/const';
+import type { KServeDeployment } from '../../types';
+import DeploymentServingDetails from '../deploymentServingDetails';
+import { useFetchTemplate } from '../../api/template';
+
+// The project-scoped template fetch is the piece we drive per-test.
+jest.mock('../../api/template', () => ({
+  useFetchTemplate: jest.fn(),
+  useFetchTemplates: jest.fn(),
+}));
+
+// Project-scoped label gating is not under test here; keep the area off.
+jest.mock('@odh-dashboard/plugin-core/areas', () => ({
+  useIsAreaAvailable: jest.fn(() => ({ status: false })),
+  SupportedArea: { DS_PROJECT_SCOPED: 'ds-project-scoped' },
+}));
+
+// Version labels rendered from the serving runtime itself are noise for these assertions.
+jest.mock('@odh-dashboard/model-serving/shared/components', () => ({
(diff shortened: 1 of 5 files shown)


packages/kserve/src/components/deploymentServingDetails.tsx
```diff
@@ -1,13 +1,121 @@
...
+        <Stack>
+          <StackItem>{getDisplayNameFromServingRuntimeTemplate(servingRuntime)}</StackItem>
+          <StackItem>
+            <LabelGroup numLabels={5}>
+              {renderDeploymentResourceVersionLabels(servingRuntime, { isCompact: true })}
+              {versionStatus && (
+                <ServingRuntimeVersionStatus
+                  isOutdated={versionStatus === ServingRuntimeVersionStatusLabel.OUTDATED}
+                  version={getServingRuntimeVersion(servingRuntime) || ''}
+                  templateVersion={getServingRuntimeVersion(template) || ''}
```
mturley: These `|| ''` fallbacks are unreachable. This block only renders when `versionStatus` is truthy, and `getServingRuntimeVersionStatus` returns a value only when both versions are already defined non-empty strings, so neither `getServingRuntimeVersion` call can be empty here. Drop the fallbacks, or tighten the prop typing if the empty string was covering a type mismatch.

packages/kserve/src/components/__tests__/deploymentServingDetails.spec.tsx
```diff
@@ -0,0 +1,127 @@
...
+    const server = mockServingRuntimeK8sResource({
+      templateName: 'ovms',
+      version: '2.0.0',
+      scope: SERVING_RUNTIME_SCOPE.Project,
+    });
+    const projectTemplate = mockServingRuntimeTemplateK8sResource({
+      name: 'ovms',
+      version: '2.0.0',
+    });
+    mockUseFetchTemplate.mockReturnValue(
```
mturley: This test mocks `useFetchTemplate` entirely, so it verifies the return value is handled but not that the component calls the hook with the right arguments. The fix depends on fetching from the serving runtime's own namespace, so add a `toHaveBeenCalledWith('ovms', <server namespace>, true)` assertion to guard against a regression that passes the wrong namespace or shouldFetch.

packages/kserve/src/components/deploymentServingDetails.tsx
```diff
@@ -1,13 +1,121 @@
...
+  const error = shouldCheckProject ? projectTemplateError : globalTemplatesError;
+
+  const versionStatus = React.useMemo(() => {
+    return getServingRuntimeVersionStatus(
+      getServingRuntimeVersion(servingRuntime),
+      getServingRuntimeVersion(template),
+    );
+  }, [template, servingRuntime]);
+
+  const isTemplateRemoved = !!templateName && (allLoaded || error) && !template;
```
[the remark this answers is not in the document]
emilys314: fixed

mturley (approved):
Verified in emily2 with the nimWizard flag on:
Image

Also verified in my own project with non-NIM deployments without the flag on:
Image

/lgtm
/approve

PR #9361, opened 2026-08-18
feat: migrate monorepo foundation from npm to pnpm

dpanshug:
https://issues.redhat.com/browse/RHOAIENG-83227

## Description

Migrates the ODH Dashboard first-party monorepo from **npm workspaces** to **pnpm 11.22.0** for local development, CI, security automation, and supported container builds. The migration preserves existing application behavior and keeps independently maintained upstream frontends on their native npm workflows.

### Workspace and dependency management

- Add `pnpm-workspace.yaml` as the workspace source of truth and pin `pnpm@11.22.0` in the root `package.json`.
- Replace first-party `package-lock.json` files with one root `pnpm-lock.yaml` and convert internal `@odh-dashboard/*` dependencies to `workspace:*` where appropriate.
- Keep `packages/model-registry/upstream/frontend` and `packages/notebooks/upstream/workspaces/frontend` as explicit npm islands with their upstream lockfiles.
- Retain `shamefully-hoist=true` temporarily for npm-era compatibility. [RHOAIENG-83228](https://issues.redhat.com/browse/RHOAIENG-83228) tracks removing it after undeclared dependency edges are corrected.

### Developer workflow and tooling

- Migrate root/package scripts, Husky, Makefile wrappers, Cypress commands, contract tests, subtree sync, workspace discovery, security audit tooling, and repository guidance to pnpm.
- Replace `npm query .workspace` with `scripts/query-workspace-packages.js`, which delegates workspace expansion to pnpm and provides normalized metadata to Module Federation, Cypress discovery, coverage, and port validation consumers.
- Add pnpm-aware Jest and Webpack/Rspack helpers for the virtual-store path layout.
- Add [ADR 0004](docs/adr/0004-use-pnpm-workspaces.md), including upstream exceptions, temporary hoisting, developer rollout, and the rollback procedure tracked by RHOAIENG-83229.

### CI and containers

- Add the shared `.github/actions/pnpm-setup` action for pinned pnpm setup, store caching, and frozen installation.
- Make the existing module-cache action pnpm-compatible and retain it only where caching separately installed upstream npm-island dependencies is useful.
- Migrate first-party GitHub workflows, dependency validation, Dependabot/audit handling, Cypress artifact preparation, and build validation to pnpm.
- Build supported module/distribution workspace images from repository-root context with `pnpm install --frozen-lockfile`.
- Preserve hybrid root-pnpm/upstream-npm builds for Model Registry and Notebooks.

### Compatibility fixes retained after review

- Upgrade direct `@rspack/core` declarations to 2.2.4 to fix an upstream CopyRspackPlugin symlink-walk regression exposed by pnpm. Controlled measurements restored production build, dev startup, HMR, and incremental type-check latency to approximate npm/main parity.
- Use pnpm-aware Jest transform/module mapping and contract-test package-root schema resolution.
- Use `scripts/cypress/mock-federation-servers.js` to own static Module Federation test servers and cleanup directly, avoiding hangs from nested long-lived Turbo server tasks while preserving the existing static HTTP behavior.
- Preserve the supported Cypress coverage task and filter only virtual Module Federation data URLs from merged coverage.

No product runtime or UI behavior change is intended. Unrelated Gen AI namespace behavior, Kubernetes/model-serving behavior, mock changes, container runtime changes, and other application fixes found during review were removed from this PR.

### Local setup after merge

```bash
npm install --global pnpm@11.22.0  # one-time bootstrap of the repository-pinned pnpm version
pnpm install
```

Do not create package-local lockfiles for first-party workspaces. Commands run inside the Model Registry and Notebooks upstream frontend directories continue to follow their upstream npm instructions.

### Not in this PR

- Removing temporary shameful hoisting: [RHOAIENG-83228](https://issues.redhat.com/browse/RHOAIENG-83228).
- Migrating the independently maintained Model Registry or Notebooks upstream frontends to pnpm.
- Fixing the repository's broad Markdown ESLint processor/project configuration; migration documentation was validated independently where that existing configuration cannot parse nested Markdown.

## How Has This Been Tested?

Local and review validation completed across the migration includes:

- `pnpm install --frozen-lockfile` and a frozen offline installation.
- Full workspace type-check: 33/33 tasks.
- Full workspace unit suite: 22/22 projects.
- Dashboard Jest: 373 suites and 4,172 tests in the final application-scope validation.
- Workspace discovery, App Config, Jest helper, contract runner, security-audit, and Cypress federation-server unit tests.
- Representative dashboard, Gen AI, Module Federation, Cypress coverage, targeted-spec, and restored-artifact test paths.
- All workspace production build tasks after the Rspack compatibility correction.
- Real Data Connect Hub and Gen AI container builds, including runtime user/command and expected artifact inspection.
- Dockerfile and Makefile dry runs, actionlint, shell syntax, package lint, documentation command/link validation, and diff checks.
- Controlled same-machine npm/main versus pnpm performance measurements for production builds, dev startup, HMR, and incremental type checking.

Current-head CI status at the latest review:

- 189 checks pass, including all 15 Prow contexts and the current Konflux component builds.
- Type-check, unit, contract, BFF, frontend, Cypress setup/shards, build simulation, audit, and image-mirror checks pass.
- `Quality-Gate (lint)` was canceled by a hosted-runner shutdown after 36/49 Turbo tasks; logs showed warnings but no lint errors before cancellation. The aggregate `Tests` check failed because of that canceled dependency and needs a clean rerun.
- Disconnected Readiness reports four false positives from pnpm package-extension keys plus one tagged image in a vendored Notebooks schema. Approved scorer PR [opendatahub-io/disconnected-readiness-scorer#112](https://github.com/opendatahub-io/disconnected-readiness-scorer/pull/112) addresses the four pnpm metadata findings; the remaining upstream finding needs a central exception or approved upstream/product resolution.

## Test Impact

No new product behavior was introduced, so no new product-flow test is required. Existing tests were updated or added where migration tooling behavior changed, including:

- workspace discovery and Module Federation package metadata;
- pnpm Jest transform and module mapping;
- Cypress federation server discovery, artifact validation, argument forwarding, and process cleanup;
- contract-test schema path handling;
- security-audit classification and pnpm lockfile behavior.

Existing unit, contract, Cypress, build, and container suites provide regression coverage for the migrated workflows. The temporary `shamefully-hoist=true` compatibility setting is explicitly tracked for removal in [RHOAIENG-83228](https://redhat.atlassian.net/browse/RHOAIENG-83228) rather than being treated as the final dependency-isolation state.

## Request review criteria:

[RHOAIENG-83228]: https://redhat.atlassian.net/browse/RHOAIENG-83228?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[RHOAIENG-83228]: https://redhat.atlassian.net/browse/RHOAIENG-83228?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dpanshug:
Removing everyone from PR reviewer for now, so you won't receive CI failure emails

dpanshug:
Added `ok-to-skip-audit` for the dependency-validation audit failure.

**Why:** This PR introduces `pnpm-lock.yaml` for the first time. The audit workflow compares `pnpm audit` on the PR head against `main`, but `main` has no `pnpm-lock.yaml`, so the base is treated as zero advisories. The 17 reported GHSA IDs are existing transitive findings (tar, undici, fastify, immutable, etc.) — not new vulnerabilities introduced by the pnpm migration.

After merge, `main` will have a pnpm lockfile baseline and future PRs will only fail on genuinely new advisories.

dpanshug:

dpanshug:

.github/workflows/claude-preflight.yml
```diff
@@ -206,8 +206,7 @@ jobs:
             --permission-mode acceptEdits
             --allowedTools "Bash(gh *)"
             --allowedTools "Bash(git *)"
-            --allowedTools "Bash(npm *)"
-            --allowedTools "Bash(npx *)"
+            --allowedTools "Bash(pnpm *)"
```
christianvogt: add `pnpm` but don't remove `npm`

.agentic-ci/config.yml
```diff
@@ -1,5 +1,5 @@
 setup:
-  - name: Install Node.js
-    run: dnf module enable nodejs:22 -y && dnf install -y --nodocs nodejs npm && npm install -g npm@11.8.0
+  - name: Install Node.js and pnpm
+    run: dnf module enable nodejs:22 -y && dnf install -y --nodocs nodejs npm && npm install -g pnpm@11.22.0
   - name: Install dependencies
-    run: npm ci
+    run: pnpm install --frozen-lockfile
```
christianvogt, line 5: I think we still need npm here.

cc @Gkrumbach07

docs/pnpm-migration-spike.md
dpanshug, line 1: we can remove this spike doc

.github/workflows/test.yml
```diff
@@ -10,85 +10,82 @@ concurrency:
...
+      - name: Setup pnpm and install dependencies
+        uses: ./.github/actions/pnpm-setup
         with:
           node-version: ${{ env.NODE_VERSION }}
       - name: Cache turbo
         uses: actions/cache@0057852bfaa89a56745cba8c7296529d2fc39830 # v4
         with:
-          path: |
-            **/.turbo
+          path: ${{ env.TURBO_CACHE_PATHS }}
```
christianvogt: I believe we can actually just use `path: .turbo` according to https://turborepo.dev/docs/guides/ci-vendors/github-actions#remote-caching-with-github-actionscache

Lucifergene (commented):
GenAI Package Review: Found no blocking issues in the changed GenAI code.

Validation completed:
- pnpm install --frozen-lockfile
- GenAI frontend lint and type-check
- Rspack production build
- Jest: 117 suites / 1,833 tests passed
- GenAI contract tests: 18 passed
- Targeted Cypress specs passed
- Local smoke test with pnpm run dev and make dev-start
- Verified dashboard → GenAI federation, BFF, LlamaStack port-forward, MLflow, namespace discovery, LSD status/models, and AAA models

scripts/cypress/mock-federation-servers.js
```diff
@@ -0,0 +1,228 @@
+/**
+ * Start or wait for Cypress mock module-federation static servers.
+ *
+ * Usage (from frontend/):
+ *   node ../scripts/cypress/mock-federation-servers.js start
+ *   node ../scripts/cypress/mock-federation-servers.js wait
+ *   node ../scripts/cypress/mock-federation-servers.js urls
+ */
```
christianvogt, line 17: These mock-federation-server scripts seem to be temporary to work around how we use turborepo to run scripts in the repo on all workspace packages. This PR brings in additional modules that weren't previously in the workspace and therefore it was running extra scripts.

Ideally we address the script names so that it's clear which scripts we need to run in which packages for CI.

Until then this can remain as a workaround but the comment at the top should describe it being temporary and why it exists.

packages/agent-ops/frontend/tsconfig.json
```diff
@@ -1,5 +1,6 @@
 {
   "compilerOptions": {
+    "rootDir": "../../..",
```
christianvogt: Why are the tsconfig changes required. Validate if they still are.

scripts/validate-module-ports.js
```diff
@@ -14,6 +14,8 @@ const { execSync } = require('child_process');
 const fs = require('fs');
 const path = require('path');
 
+const WORKSPACE_QUERY_SCRIPT = path.join(__dirname, 'query-workspace-packages.js');
```
christianvogt, line 17: Suggestion — how we invoke query-workspace-packages

The script already exports listWorkspacePackagesFromManifest / findRepoRoot, but most callers still spawn node with different relative paths (REPO_ROOT/scripts/…, ../../../scripts/…, $MONOREPO_ROOT/scripts/…).

Consider:

Root package.json bin (same pattern as unit-test-coverage-aggregator) so shell/CI/Makefiles use pnpm exec query-workspace-packages or $MONOREPO_ROOT/node_modules/.bin/query-workspace-packages instead of hard-coded scripts/ paths.

In-repo JS/TS — require the module (or an @odh-dashboard/app-config export) in-process instead of execSync('node "…"'), so rspack/discovery code doesn’t depend on __dirname depth.

packages/mlflow/frontend/config/rspack.common.js
```diff
@@ -53,28 +51,9 @@ module.exports = (env) => ({
...
-          ),
-          path.resolve(RELATIVE_DIRNAME, 'node_modules/@patternfly/patternfly/assets/fonts'),
-          path.resolve(RELATIVE_DIRNAME, 'node_modules/@patternfly/patternfly/assets/pficon'),
-          // Root node_modules for @odh-dashboard/internal imports
-          path.resolve(ROOT_NODE_MODULES, 'patternfly/dist/fonts'),
-          path.resolve(ROOT_NODE_MODULES, '@patternfly/react-core/dist/styles/assets/fonts'),
-          path.resolve(ROOT_NODE_MODULES, '@patternfly/react-core/dist/styles/assets/pficon'),
-          path.resolve(ROOT_NODE_MODULES, '@patternfly/patternfly/assets/fonts'),
-          path.resolve(ROOT_NODE_MODULES, '@patternfly/patternfly/assets/pficon'),
-        ],
```
christianvogt, line 77: I'd really like to avoid having to make the rspack changes because it just expands the risk of this PR.

Is it necessary still?

.agentic-ci/config.yml
dpanshug: can you explain more on this, why we still need npm here?

scripts/validate-module-ports.js
dpanshug: I verified this in a clean worktree with no `node_modules`: the current `pnpm run validate:ports` path succeeds, so we need to preserve pre-install workspace querying. I also tested the `root-bin` approach. The existing root unit-test-coverage-aggregator bin is not available through pnpm exec, and adding a root `query-workspace-packages` bin was likewise still reported as “command not found,” including after pnpm installed dependencies. 

We could directly import the current module from node callers, but providing a stable pnpm exec CLI would need a linked workspace tooling package or a different command contract such as a root package script. Since the current anchored paths work, suggest handling the complete API/CLI cleanup separately rather than partially converting callers in this migration PR.

packages/mlflow/frontend/config/rspack.common.js
dpanshug: Restoring MLflow’s old rspack configuration under the pnpm install causes both production and development builds to fail with roughly 270 PatternFly CSS parse errors

Because assets imported through nested workspace dependencies fall outside the old hard coded include paths, these Rspack changes are still required for the pnpm layout

scripts/validate-module-ports.js
christianvogt: @dpanshug I'm fine with looking for a post PR solution.

Griffin-Sullivan (commented):
MaaS changes look fine to me and it's running good after installing

.agentic-ci/config.yml
Gkrumbach07: this is so that the agent in auto fix doesnt spend a bunch of time installing deps. you need to check that the open shell image they use supports pnpm

christianvogt (approved):

manaswinidas (approved):
/approve

PR #2290, opened 2023-12-11
Edge pipeline topology

dpanshug:
Closes: [RHOAIENG-238](https://issues.redhat.com/browse/RHOAIENG-238)  #2250

## Description
Added Edge pipeline topology views

Pipeline page

image

Pipeline page YAML

image

Pipeline run page

image

@vconzola 

## How Has This Been Tested?
Create pipeline and pipelineRun from [here](https://github.com/opendatahub-io/ai-edge/blob/9a3c0e48fa3c7395be3ccdfc89575c0088b9841a/pipelines/tekton/aiedge-e2e/aiedge-e2e.pipeline.yaml)

Redirect to `http://localhost:4010/edgePipelines/pipeline/view/aiedge-e2e` for pipeline topology page
and to `http://localhost:4010/edgePipelines/pipelineRun/view/<pipelineRunName>` for pipeline run topology page

## Request review criteria:

Self checklist (all need to be checked):
- [x] The developer has manually tested the changes and verified that the changes work
- [x] Commits have been squashed into descriptive, self-contained units of work (e.g. 'WIP' and 'Implements feedback' style messages have been removed)
- [x] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [ ] The developer has added tests or explained why testing cannot be added (unit tests & storybook for related changes)

If you have UI changes: 

- [x] Included any necessary screenshots or gifs if it was a UI change.
- [x] Included tags to the UX team if it was a UI/UX change (find relevant UX in the [SMEs](https://github.com/opendatahub-io/odh-dashboard/tree/main/docs/smes.md) section).

After the PR is posted & before it merges:
- [ ] The developer has tested their solution on a cluster by using the image produced by the PR to `main`

PR #9362, opened 2026-08-18
fix: MLflow e2e test retry failure due to stale experiment names

crackcodecamp:
## Summary

- Fix `testMlflowExperiments` e2e test consistently failing on retry across builds #2272, #2290, #2299
- Root cause: MLflow's delete API only soft-deletes experiments, retaining the UNIQUE name constraint. On retry, creating an experiment with the same name returns `RESOURCE_ALREADY_EXISTS`, causing the embedded MLflow UI to show an error modal instead of navigating to the detail page
- Add `freeExperimentName()` utility (restore → rename → delete) to properly free experiment names between retries
- Clear browser state between retries to prevent stale module federation cache

## Changes

1. **`packages/cypress/cypress/utils/oc_commands/mlflow.ts`**:
   - `renameMlflowExperimentViaAPI()` — rename via `/experiments/update`
   - `restoreMlflowExperimentViaAPI()` — restore soft-deleted via `/experiments/restore`
   - `freeExperimentName()` — composite: lookup → restore → rename with timestamp suffix → delete

2. **`packages/cypress/cypress/tests/e2e/mlflowExperiments/testMlflowExperiments.cy.ts`**:
   - Add cleanup step in `retryableBefore` to free both experiment and renamed experiment names
   - Add `beforeEach` to clear cookies/localStorage/sessionStorage for clean module federation state

## Test plan

- [x] Verified on live cluster (`ag-gcp-34-pool-k6spq`): restore → rename → delete successfully frees experiment name, allowing create with same name to succeed
- [x] Confirmed MLflow soft-delete retains UNIQUE constraint — `RESOURCE_ALREADY_EXISTS` error reproduced and fixed
- [ ] Run full e2e suite on Jenkins to verify the fix resolves the retry failures

### DIFF
--- a/packages/cypress/cypress/tests/e2e/mlflowExperiments/testMlflowExperiments.cy.ts
+++ b/packages/cypress/cypress/tests/e2e/mlflowExperiments/testMlflowExperiments.cy.ts
@@ -7,6 +7,7 @@ import {
   deleteMlflowExperimentViaAPI,
   getMlflowExperimentIdByName,
   logMlflowRunsViaAPI,
+  freeExperimentName,
 } from '../../../utils/oc_commands/mlflow';
 import { deleteOpenShiftProject, createOpenShiftProject } from '../../../utils/oc_commands/project';
 import { retryableBefore } from '../../../utils/retryableHooks';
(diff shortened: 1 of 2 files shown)


crackcodecamp:
Closing in favor of cherry-picking #8870 which already addresses this same issue on main/3.5/3.6 with a better approach (suffix incrementing instead of restore/rename/delete, plus prompt management coverage). Will cherry-pick #8870 to v3.4.0-fixes instead to avoid divergence.

PR #9373, opened 2026-08-18
fix(autorag): reconfiguring run does not retain selected models

daniduong:
https://redhat.atlassian.net/browse/RHOAIENG-85392

## Description

Reconfiguring an AutoRAG run did not retain the foundation (generation) and embedding model selections from the source run — the model selection was always reset to "all models" instead.

**Root cause:** `AutoragConfigure` uses two effects to manage model selection:
1. An effect that clears `generation_models` / `embedding_models` and flags them for re-initialization whenever the Open GenAI Stack secret (`ogx_secret_name`) changes.
2. A second effect that re-populates those fields once the models query resolves, preserving the current selection if it is non-empty, otherwise defaulting to all available models.

During reconfigure, the form is pre-populated on mount with the previous run's `ogx_secret_name` and model selections. Effect 1 was a plain `useEffect`, so it fired on the initial mount too (since its dependency already had a value), wiping the pre-populated selections to `[]` before effect 2 ran. Effect 2 then saw empty arrays and fell back to selecting all models — reproducing the reported bug.

**Fix:** Switch effect 1 to `useReconfigureSafeEffect`, the hook already used elsewhere in this component specifically to skip firing on mount so pre-populated reconfigure values aren't wiped before the user sees them. Also removed a couple of now-unnecessary `?? []` fallbacks flagged by `@typescript-eslint/no-unnecessary-condition` in the adjacent code.

## How Has This Been Tested?

- Added a Jest regression test that renders `AutoragConfigure` in the reconfigure flow with a pre-set `ogx_secret_name` and a subset of previously selected foundation/embedding models, while the mocked models query returns a larger set of available models. Verified the previously selected subset is retained instead of being reset to all models.
- Ran the full `AutoragConfigure.spec.tsx` suite (51/51 passing).
- Ran `eslint` and `tsc --noEmit` on the changed file — both clean.

Before | After
--- | ---
<video src="https://github.com/user-attachments/assets/9b92fbd1-8232-42ef-a4f9-b81894582240" /> | <video src="https://github.com/user-attachments/assets/281c0b58-bb83-4565-ad7e-b0da5cfa22ed" />

## Test Impact

- Added `should retain the previously selected foundation/embedding models instead of resetting to all models` in `packages/autorag/frontend/src/app/components/configure/__tests__/AutoragConfigure.spec.tsx`, which fails against the previous implementation and passes with this fix.

## Request review criteria:

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved selected models when reconfiguring, provided they remain available.
  * Removed unavailable selections and displayed a warning.
  * Automatically restored all available models when selections are empty or no longer valid.
  * Retained valid models when only some previously selected models are unavailable.

### DIFF
--- a/packages/autorag/frontend/src/app/components/configure/AutoragConfigure.tsx
+++ b/packages/autorag/frontend/src/app/components/configure/AutoragConfigure.tsx
@@ -232,28 +232,42 @@ function AutoragConfigure({
 
   // When the secret changes, mark models as needing re-initialization and
   // immediately clear stale selections so the UI reflects the transition.
-  useEffect(() => {
+  // Uses useReconfigureSafeEffect (skips on mount) because ogxSecretName is
+  // already populated on mount during reconfigure; a plain useEffect would
+  // wipe the pre-populated model selections before they could be restored.
+  useReconfigureSafeEffect(() => {
     modelsInitialized.current = false;
     setValue('generation_models', []);
     setValue('embedding_models', []);
   }, [ogxSecretName, setValue]);
 
   useEffect(() => {
-    // Initialize available generation and embedding models into the form data
+    // Initialize available generation and embedding models into the form data.
+    // Preserve existing selections (reconfigure flow) when they are already
+    // populated; only default to all models on a fresh create.
     if (allModelsData?.models && !modelsInitialized.current && !isModelsError) {
       modelsInitialized.current = true;
+
+      const currentValues = getValues();
+      const currentGenModels = currentValues.generation_models;
+      const currentEmbModels = currentValues.embedding_models;
+
+      const allLlmModels = allModelsData.models
+        .filter((model) => model.type === 'llm')
+        .map((model) => model.id)
+        .toSorted((a, b) => a.localeCompare(b));
+
+      const allEmbeddingModels = allModelsData.models
+        .filter((model) => model.type === 'embedding')
+        .map((model) => model.id)
+        .toSorted((a, b) => a.localeCompare(b));
+
       reset({
-        ...getValues(),
+        ...currentValues,
         // eslint-disable-next-line camelcase
-        generation_models: allModelsData.models
-          .filter((model) => model.type === 'llm')
-          .map((model) => model.id)
-          .toSorted((a, b) => a.localeCompare(b)),
+        generation_models: currentGenModels.length > 0 ? currentGenModels : allLlmModels,
         // eslint-disable-next-line camelcase
-        embedding_models: allModelsData.models
-          .filter((model) => model.type === 'embedding')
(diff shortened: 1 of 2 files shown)


daniduong:
@coderabbitai review

nickmazzi:
Thanks @daniduong!
None of the AI comments seems mandatory. Approving.
CI Failure is unrelated to autorag or your changes.

/lgtm
/approve

# AI Assisted review
## Overall Assessment

**Approve with minor suggestions.** This is a well-written, well-tested bug fix. The root cause analysis is correct, the fix is targeted and appropriate, and the test coverage is thorough with three distinct scenarios (valid retention, fully stale, partially stale). The CodeRabbit suggestions for filtering stale models and asserting exact model IDs have both been addressed.


## PR Comment Review

| # | Reviewer | Comment Summary | File | Status | Notes |
|---|----------|----------------|------|--------|-------|
| 1 | @coderabbitai[bot] | Assert preserved model IDs, not just counts | `AutoragConfigure.spec.tsx:1242` | Resolved | Added `getLatestFormValues()` assertions for exact model IDs (lines 1238–1239 in new code) |
| 2 | @coderabbitai[bot] | Filter restored model IDs against currently available models to prevent stale selections | `AutoragConfigure.tsx:265` | Resolved | Implemented `retainedGenerationModels` / `retainedEmbeddingModels` filtering with fallback-to-all and warning notification (lines 280–295) |

**Comment stats**: 2 resolved, 0 not resolved, 0 intentionally skipped out of 2 total action items.

## Suggestions

1. **Assert warning notification**: Add assertions in the "drop restored" and "partial stale" tests that `notification.warning` was called with `'Some previously selected models are unavailable'`, and in the "should retain" test that it was *not* called. This captures a named mock (`jest.fn()` at line 110) for the `warning` key but it's never asserted on.

2. **Reset `latestForm` between tests**: Add `latestForm = undefined;` to a `beforeEach` or `afterEach` block for the reconfigure describe block to ensure strict test isolation.

3. **Consider `Set` for lookup**: The `retainedGenerationModels.filter(id => allLlmModels.includes(id))` is O(n*m). For the expected small number of models this is fine, but if model counts grow, using a `Set` for `allLlmModels` would be more efficient. Not blocking — just a note for awareness.

# Test Validation

## autorag

### Overall Summary

| Suite      | Result |
|------------|--------|
| Frontend   | ✅     |
| BFF        | ✅     |
| Contract   | ✅     |
| BFF Lint   | ✅     |
| Lint       | ✅     |
| Type Check | ✅     |

# Manual validation
<table><tr><Td>Before</td><td>After</td></tr>
<tr><td>
image
</td><td>
image
</td></tr></table>

PR #9380, opened 2026-08-18
feat(RHOAIENG-80010): add bidirectional filtering between API groups and resources

srtanish1992:
https://issues.redhat.com/browse/RHOAIENG-80010

## Description

Implements cross-field coordination in the Add Rule modal (Custom Role creation) so that API Groups and Resource Types selections are contextually linked:

- **API Group → Resources filtering**: Selecting an API group narrows the Resources dropdown to only show resources belonging to that group
- **Resource → API Group auto-population**: Selecting a resource automatically adds its corresponding API group
- **Cascading deselection**: Removing an API group removes resources that only belonged to that group
- **Wildcard handling**: `*` (all) in either field shows all options in the other
- **Custom entries**: User-typed values that don't map to known resources/groups are preserved without side effects

https://github.com/user-attachments/assets/10aaa984-b671-4214-9324-96e012e0d2f1

### Behavior Summary

| Trigger | Effect |
| --- | --- |
| Select API group | Resources dropdown filters to show only resources in selected groups |
| Select resource (no matching API group selected) | Auto-add the resource's API group |
| Remove API group | Auto-remove resources whose API group is no longer selected |
| Remove resource | Keep the API group (no cascade up) |
| No API groups selected | Show all resources (unfiltered) |
| Wildcard API group (`*`) | Show all resources (unfiltered) |
| Wildcard resources (`*`) | No auto-populate of API groups |
| Custom resource (unknown group) | No auto-populate |

## How Has This Been Tested?

- Unit tests for `AddRuleModal` orchestration logic (14 test cases covering all scenarios)
- Unit tests for `ResourcesTreeSelect` `filterByApiGroups` prop (7 test cases)
- Manual testing of the full Add Rule modal flow

## Test Impact

- Added `AddRuleModal.spec.tsx` with 14 tests covering auto-population, cascading deselection, wildcards, custom entries, and edit mode
- Added 7 tests to `ResourcesTreeSelect.spec.tsx` for the `filterByApiGroups` prop

## Request review criteria:

If you have UI changes:

## Summary by CodeRabbit

* **Improvements**
  * Resource and API group selections now stay synchronized when creating or editing project role rules.
  * Resource lists can be filtered by selected API groups while retaining wildcard and “All resources” options.
  * Expanded Core, Storage, and RBAC categories to include additional Kubernetes resources and API groups.
  * Duplicate or unnamed API groups are excluded from selection lists.
  * API group wildcard selections are normalized when rules are saved.
  * Only one selection menu remains open at a time for a cleaner experience.

### DIFF
--- a/frontend/src/pages/projects/projectRoles/AddRuleModal.tsx
+++ b/frontend/src/pages/projects/projectRoles/AddRuleModal.tsx
@@ -8,7 +8,8 @@ import ApiGroupsTreeSelect from './ApiGroupsTreeSelect';
 import useApiResources from './useApiResources';
 import type { RuleEntry } from './types';
 import { normalizeVerbs } from './ruleModalUtils';
-import { ALL_RESOURCES_WILDCARD } from './resourceCategories';
+import { ALL_RESOURCES_WILDCARD, RESOURCE_CATEGORIES } from './resourceCategories';
+import { ALL_API_GROUPS_WILDCARD } from './apiGroupCategories';
 
 type AddRuleModalProps = {
   existingRule?: RuleEntry;
@@ -33,6 +34,81 @@ const AddRuleModal: React.FC<AddRuleModalProps> = ({ existingRule, onSave, onClo
 
   const [selectedVerbs, setSelectedVerbs] = React.useState<string[]>(
     () => existingRule?.verbs ?? [],
+  );
+
+  const resolvedApiResourcesData = apiResourcesLoaded
+    ? apiResourcesData
+    : { apiGroups: [], resources: [] };
+
+  const resourceToApiGroupMap = React.useMemo(() => {
+    const map = new Map<string, string>();
+    for (const category of RESOURCE_CATEGORIES) {
+      for (const r of category.resources) {
+        map.set(r.name, r.apiGroup);
+      }
+    }
+    for (const r of resolvedApiResourcesData.resources) {
+      if (!map.has(r.name)) {
+        map.set(r.name, r.apiGroup);
+      }
+    }
+    return map;
+  }, [resolvedApiResourcesData.resources]);
+
+  const handleResourcesChange = React.useCallback(
+    (newResources: string[]) => {
+      setSelectedResources(newResources);
+
+      if (newResources.includes(ALL_RESOURCES_WILDCARD)) {
+        return;
+      }
+
+      const addedResources = newResources.filter((r) => !selectedResources.includes(r));
+      const groupsToAdd = new Set<string>();
+
+      for (const resource of addedResources) {
+        const apiGroup = resourceToApiGroupMap.get(resource);
+        if (apiGroup !== undefined && !selectedApiGroups.includes(apiGroup)) {
+          groupsToAdd.add(apiGroup);
+        }
+      }
+
+      if (groupsToAdd.size > 0) {
+        setSelectedApiGroups((prev) => [...prev, ...groupsToAdd]);
+      }
+    },
+    [selectedResources, selectedApiGroups, resourceToApiGroupMap],
+  );
+
+  const handleApiGroupsChange = React.useCallback(
+    (newApiGroups: string[]) => {
+      setSelectedApiGroups(newApiGroups);
+
+      if (newApiGroups.length === 0 || newApiGroups.includes(ALL_API_GROUPS_WILDCARD)) {
+        return;
+      }
+
+      if (selectedResources.includes(ALL_RESOURCES_WILDCARD)) {
+        return;
+      }
+
+      const removedGroups = selectedApiGroups.filter((g) => !newApiGroups.includes(g));
+      if (removedGroups.length === 0) {
+        return;
+      }
+
+      const allowedGroups = new Set(newApiGroups);
+      const orphanedResources = selectedResources.filter((r) => {
+        const group = resourceToApiGroupMap.get(r);
+        return group !== undefined && !allowedGroups.has(group);
+      });
+
+      if (orphanedResources.length > 0) {
+        const orphanSet = new Set(orphanedResources);
+        setSelectedResources((prev) => prev.filter((r) => !orphanSet.has(r)));
+      }
+    },
+    [selectedApiGroups, selectedResources, resourceToApiGroupMap],
   );
 
   const canSave =
@@ -93,10 +169,8 @@ const AddRuleModal: React.FC<AddRuleModalProps> = ({ existingRule, onSave, onClo
           >
             <ApiGroupsTreeSelect
               selectedApiGroups={selectedApiGroups}
-              onSelectedApiGroupsChange={setSelectedApiGroups}
-              apiResourcesData={
-                apiResourcesLoaded ? apiResourcesData : { apiGroups: [], resources: [] }
-              }
+              onSelectedApiGroupsChange={handleApiGroupsChange}
+              apiResourcesData={resolvedApiResourcesData}
             />
           </FormGroup>
           <FormGroup
@@ -109,10 +183,9 @@ const AddRuleModal: React.FC<AddRuleModalProps> = ({ existingRule, onSave, onClo
           >
             <ResourcesTreeSelect
               selectedResources={selectedResources}
-              onSelectedResourcesChange={setSelectedResources}
-              apiResourcesData={
-                apiResourcesLoaded ? apiResourcesData : { apiGroups: [], resources: [] }
-              }
+              onSelectedResourcesChange={handleResourcesChange}
+              filterByApiGroups={selectedApiGroups}
+              apiResourcesData={resolvedApiResourcesData}
             />
           </FormGroup>
           <FormGroup label="Permitted operations" fieldId="rule-verbs" isRequired>
--- a/frontend/src/pages/projects/projectRoles/AddRuleModal.tsx
+++ b/frontend/src/pages/projects/projectRoles/AddRuleModal.tsx
@@ -10,6 +10,9 @@ import type { RuleEntry } from './types';
 import { normalizeVerbs } from './ruleModalUtils';
 import { ALL_RESOURCES_WILDCARD, RESOURCE_CATEGORIES } from './resourceCategories';
 import { ALL_API_GROUPS_WILDCARD } from './apiGroupCategories';
+import type { ApiResourcesData } from './useApiResources';
+
+const EMPTY_API_RESOURCES_DATA: ApiResourcesData = { apiGroups: [], resources: [] };
 
 type AddRuleModalProps = {
   existingRule?: RuleEntry;
@@ -36,9 +39,7 @@ const AddRuleModal: React.FC<AddRuleModalProps> = ({ existingRule, onSave, onClo
     () => existingRule?.verbs ?? [],
   );
 
-  const resolvedApiResourcesData = apiResourcesLoaded
-    ? apiResourcesData
-    : { apiGroups: [], resources: [] };
+  const resolvedApiResourcesData = apiResourcesLoaded ? apiResourcesData : EMPTY_API_RESOURCES_DATA;
 
   const resourceToApiGroupMap = React.useMemo(() => {
     const map = new Map<string, string>();
@@ -97,6 +98,7 @@ const AddRuleModal: React.FC<AddRuleModalProps> = ({ existingRule, onSave, onClo
         return;
       }
 
+      const allGroupsWereSelected = selectedApiGroups.includes(ALL_API_GROUPS_WILDCARD);
       const allowedGroups = new Set(newApiGroups);
       const orphanedResources = selectedResources.filter((r) => {
         const group = resourceToApiGroupMap.get(r);
@@ -104,7 +106,7 @@ const AddRuleModal: React.FC<AddRuleModalProps> = ({ existingRule, onSave, onClo
           return false;
         }
         if (newApiGroups.length === 0) {
-          return removedGroups.includes(group);
+          return allGroupsWereSelected || removedGroups.includes(group);
         }
         return !allowedGroups.has(group);
       });
--- a/frontend/src/pages/projects/projectRoles/AddRuleModal.tsx
+++ b/frontend/src/pages/projects/projectRoles/AddRuleModal.tsx
@@ -5,12 +5,11 @@ import FieldGroupHelpLabelIcon from '@odh-dashboard/ui-core/components/FieldGrou
 import VerbsTreeSelect from './VerbsTreeSelect';
 import ResourcesTreeSelect from './ResourcesTreeSelect';
 import ApiGroupsTreeSelect from './ApiGroupsTreeSelect';
-import useApiResources from './useApiResources';
+import useApiResources, { type ApiResourcesData } from './useApiResources';
 import type { RuleEntry } from './types';
 import { normalizeVerbs } from './ruleModalUtils';
 import { ALL_RESOURCES_WILDCARD, RESOURCE_CATEGORIES } from './resourceCategories';
 import { ALL_API_GROUPS_WILDCARD } from './apiGroupCategories';
-import type { ApiResourcesData } from './useApiResources';
 
 const EMPTY_API_RESOURCES_DATA: ApiResourcesData = { apiGroups: [], resources: [] };
 
--- a/frontend/src/pages/projects/projectRoles/AddRuleModal.tsx
+++ b/frontend/src/pages/projects/projectRoles/AddRuleModal.tsx
@@ -38,6 +38,12 @@ const AddRuleModal: React.FC<AddRuleModalProps> = ({ existingRule, onSave, onClo
     () => existingRule?.verbs ?? [],
   );
 
+  // selectedApiGroups: chips + YAML (user-selected and auto-added from resources).
+  // explicitApiGroups: groups chosen in the API groups field this session. Drives
+  // Resource types filtering. Starts empty in add and edit so loaded groups do not
+  // lock the list; auto-adding (notebooks → kubeflow.org) also must not lock it.
+  const [explicitApiGroups, setExplicitApiGroups] = React.useState<string[]>([]);
+
   const resolvedApiResourcesData = apiResourcesLoaded ? apiResourcesData : EMPTY_API_RESOURCES_DATA;
 
   const resourceToApiGroupMap = React.useMemo(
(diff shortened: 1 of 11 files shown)


frontend/src/pages/projects/projectRoles/__tests__/AddRuleModal.spec.tsx
```diff
@@ -0,0 +1,274 @@
...
+        resources: ['deployments', 'mycustomresource'],
+        verbs: ['get'],
+      });
+
+      act(() => {
+        capturedApiGroupsProps.onSelectedApiGroupsChange([]);
+      });
+
+      expect(capturedResourcesProps.selectedResources).toContain('mycustomresource');
+    });
```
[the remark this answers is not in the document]
srtanish1992: Fixed in bc49d53. The test now starts with two API groups (`apps` + core), removes `apps` while retaining core, and asserts:
- `deployments` (belonging to removed `apps`) is cascaded out
- `pods` (belonging to retained core group) remains
- `mycustomresource` (unknown group) is preserved

This properly exercises the orphan-resource cascade path instead of hitting the early return.

frontend/src/pages/projects/projectRoles/AddRuleModal.tsx
```diff
@@ -35,6 +36,81 @@ const AddRuleModal: React.FC<AddRuleModalProps> = ({ existingRule, onSave, onClo
...
+        map.set(r.name, r.apiGroup);
+      }
+    }
+    for (const r of resolvedApiResourcesData.resources) {
+      if (!map.has(r.name)) {
+        map.set(r.name, r.apiGroup);
+      }
+    }
+    return map;
+  }, [resolvedApiResourcesData.resources]);
```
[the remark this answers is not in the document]
srtanish1992: Not applicable — in Kubernetes, each resource belongs to exactly **one** API group. This is a fundamental constraint of the K8s API model. There is no scenario where the same resource name exists in multiple API groups simultaneously. The `Map<string, string>` correctly models this 1:1 relationship. Converting to `Map<string, Set<string>>` would add unnecessary complexity for a scenario that doesn't occur in practice.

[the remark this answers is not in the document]
srtanish1992: Fair point on the theoretical model — `GroupResource` does use both fields for identity, and CRDs could technically introduce duplicate resource names across groups.

However, in our practical context:

1. **RHOAI/ODH clusters don't have duplicate resource names across groups.** All standard K8s resources (`pods`, `deployments`, `services`, etc.) and all RHOAI CRDs (`notebooks`, `inferenceservices`, `servingruntime`, etc.) have unique names.

2. **Our `ResourcesTreeSelect` filtering already uses `r.apiGroup` directly** — each resource in the category data carries its own `apiGroup` field, so filtering by API groups works correctly regardless of the map.

3. **The map is only used for two convenience UX behaviors** (auto-populate and cascade), not for RBAC correctness. The actual Role YAML is assembled from what the user explicitly selects.

4. **The fix adds meaningful complexity** (`Map<string, Set<string>>`, intersection checks in both directions, ambiguous auto-populate behavior when a resource exists in multiple groups) for an edge case that doesn't exist on our target platform.

I'll add a code comment documenting this assumption. If duplicate resource names across groups are reported on a real cluster, we can upgrade the map in a follow-up.

DaoDaoNoCode:

d0w (commented):
While I'm not 100% sure on what apigroups should match to what resource types, it looks like filling out API groups first and the subsequent resource types filtering works.

there were a couple issues i found in the reverse direction:
- selecting resource type then deleting the associated api group, the resource types are still selected (unsure if this is expected behavior). Then when you select another resource type belonging to a different api group, the resource type is still selected but missing from the tree.
- Selecting some items such as any item in rbac or jobs under applications will populate the rest of the resource group items incorrectly (e.g. Applications will get authorization.k8s.io related items). These items will not be checkable but will still add the api group.
- Some items such as `nodes` feel like they should be under `Core` since they both have the core API group but I may be wrong to assume that here
- Many items under `Other` are unselectable. Selecting these items will clear all previously selected resource types
- Many items such as `node` are duplicated multiple times (which may be why they are unselectable)

Image

Image

frontend/src/pages/projects/projectRoles/AddRuleModal.tsx
```diff
@@ -35,6 +36,81 @@ const AddRuleModal: React.FC<AddRuleModalProps> = ({ existingRule, onSave, onClo
...
+      for (const r of category.resources) {
+        map.set(r.name, r.apiGroup);
+      }
+    }
+    for (const r of resolvedApiResourcesData.resources) {
+      if (!map.has(r.name)) {
+        map.set(r.name, r.apiGroup);
+      }
+    }
+    return map;
```
d0w: might suggest moving some of the logic to `ruleModalUtils.ts` since the file exists already (though only for 1 function). This isn't really reused anywhere else so it's not necessary but something to think about imo.

frontend/src/pages/projects/projectRoles/AddRuleModal.tsx
```diff
@@ -35,6 +36,81 @@ const AddRuleModal: React.FC<AddRuleModalProps> = ({ existingRule, onSave, onClo
...
+
+      const addedResources = newResources.filter((r) => !selectedResources.includes(r));
+      const groupsToAdd = new Set<string>();
+
+      for (const resource of addedResources) {
+        const apiGroup = resourceToApiGroupMap.get(resource);
+        if (apiGroup !== undefined && !selectedApiGroups.includes(apiGroup)) {
+          groupsToAdd.add(apiGroup);
+        }
+      }
```
d0w, line 76: same here. could make this a `getGroupsToAdd` util

frontend/src/pages/projects/projectRoles/AddRuleModal.tsx
d0w: this memoization also runs twice on modal load. Probably because resources are continuously loading in. I've logged two memoizations on load on my side but it could be more/less. I would add a check to see if `apiResourcesLoaded` is true before running the full logic.

Example:
```
 if (!apiResourcesLoaded) {
      return EMPTY_API_RESOURCES_DATA;
    }
```

frontend/src/pages/projects/projectRoles/AddRuleModal.tsx
```diff
@@ -35,6 +36,81 @@ const AddRuleModal: React.FC<AddRuleModalProps> = ({ existingRule, onSave, onClo
     () => existingRule?.verbs ?? [],
   );
 
+  const resolvedApiResourcesData = apiResourcesLoaded
+    ? apiResourcesData
+    : { apiGroups: [], resources: [] };
```
d0w: might help to turn this into a constant so other components can use this too such as in all the tests.

frontend/src/pages/projects/projectRoles/__tests__/AddRuleModal.spec.tsx
d0w, line 1: i would probably add this to the cypress tests instead since it involves a lot of user interaction

frontend/src/pages/projects/projectRoles/__tests__/ResourcesTreeSelect.spec.tsx
d0w, line 1: should add a `filterByResources` test group since the changes involve bidirectional filtering.

d0w (commented):
didn't look too much at the filtering code for this one since it'll probably change due to the bugs I noted in my previous comment.

frontend/src/pages/projects/projectRoles/AddRuleModal.tsx
srtanish1992: Good catch. Fixed in 6550d1e — extracted a module-level `EMPTY_API_RESOURCES_DATA` constant so the ternary returns a stable reference when `apiResourcesLoaded` is false. This prevents the downstream `resourceToApiGroupMap` memo from re-running on each render cycle during the loading phase.

frontend/src/pages/projects/projectRoles/AddRuleModal.tsx
srtanish1992: Done in 6550d1e — extracted as `EMPTY_API_RESOURCES_DATA` constant with the `ApiResourcesData` type. This also serves as the stable fallback reference to prevent memo re-runs during loading.

frontend/src/pages/projects/projectRoles/AddRuleModal.tsx
srtanish1992: Agreed it could be cleaner. For now keeping it inline since it's only used here and the logic is tightly coupled to the component's state. Will revisit if we consolidate the tree-select logic into a shared hook (planned follow-up from PR #9256).

frontend/src/pages/projects/projectRoles/__tests__/AddRuleModal.spec.tsx
srtanish1992: Fair point — the bidirectional interaction involves a lot of UI state transitions that are better validated end-to-end. The Jest tests here cover the orchestration logic (auto-populate, cascade, wildcards) at the unit level, but we'll add Cypress mock tests for the full user interaction flow as part of the broader custom roles E2E test coverage.

frontend/src/pages/projects/projectRoles/__tests__/ResourcesTreeSelect.spec.tsx
srtanish1992: Already added — there's a `filterByApiGroups` describe block at line 558 of `ResourcesTreeSelect.spec.tsx` with 7 tests covering: undefined/empty/wildcard filter, single group filter, multiple groups, core resources, and the "Other" category behavior under filtering.

frontend/src/pages/projects/projectRoles/AddRuleModal.tsx
srtanish1992: Same reasoning as above — keeping inline for now since it's a single-use callback tightly coupled to component state. Will extract during the shared hook consolidation follow-up.

frontend/src/pages/projects/projectRoles/AddRuleModal.tsx
```diff
@@ -35,6 +36,87 @@ const AddRuleModal: React.FC<AddRuleModalProps> = ({ existingRule, onSave, onClo
...
+      const allowedGroups = new Set(newApiGroups);
+      const orphanedResources = selectedResources.filter((r) => {
+        const group = resourceToApiGroupMap.get(r);
+        if (group === undefined) {
+          return false;
+        }
+        if (newApiGroups.length === 0) {
+          return removedGroups.includes(group);
+        }
+        return !allowedGroups.has(group);
```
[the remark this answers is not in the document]
srtanish1992: Good catch — fixed in 6550d1e. Added `allGroupsWereSelected` flag that detects when the previous selection included `'*'`, and treats it as removal of all concrete groups. Added a regression test for `['*'] → []` followed by verifying all mapped resources are deselected while custom resources survive.

frontend/src/pages/projects/projectRoles/resourceCategories.ts
```diff
@@ -73,6 +76,28 @@ export const RESOURCE_CATEGORIES: ResourceCategory[] = [
...
+        label: 'Self subject access reviews',
+        apiGroup: 'authorization.k8s.io',
+      },
+      {
+        name: 'selfsubjectrulesreviews',
+        label: 'Self subject rules reviews',
+        apiGroup: 'authorization.k8s.io',
+      },
+      { name: 'tokenreviews', label: 'Token reviews', apiGroup: 'authentication.k8s.io' },
+      { name: 'tokenrequests', label: 'Token requests', apiGroup: 'authentication.k8s.io' },
```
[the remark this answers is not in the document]
srtanish1992: Valid — `tokenrequests` is indeed not a top-level K8s resource. The TokenRequest API is a subresource of ServiceAccount (accessed via `serviceaccounts/token`). Removed the entry in 6550d1e.

Note: We intentionally do not add `serviceaccounts/token` as a static category entry because our `parseDiscoveryItems` filters out slash-containing subresources by design — the UI focuses on top-level resources and the user can always type custom subresource paths manually.

srtanish1992:
Thanks for the thorough manual testing @d0w! All issues have been addressed across commits 9dd78c15 and 6550d1e:

| Issue | Fix |
|-------|-----|
| Resource stays selected after removing API group | Fixed cascade to trigger even when all groups are removed. Also fixed wildcard `['*']→[]` case. |
| RBAC/auth items appearing incorrectly, uncheckable | Root cause: duplicate resource names across API groups (e.g., `nodes` in core + `metrics.k8s.io`) creating duplicate option IDs. Fixed by deduplicating `otherResources` by name. |
| `nodes` should be under Core | Added `nodes`, `secrets`, `serviceaccounts` to Core category. |
| Many Other items unselectable / clears all | Same dedup fix — duplicate IDs caused PF Select to misbehave. |
| Items duplicated multiple times | Same dedup fix. |

Additionally:
- Added `authorization.k8s.io` and `authentication.k8s.io` to the RBAC API group category with their resources (`subjectaccessreviews`, `tokenreviews`, etc.)
- Removed phantom `tokenrequests` entry (not a real K8s top-level resource)
- Extracted stable `EMPTY_API_RESOURCES_DATA` constant to prevent memo re-runs during loading

frontend/src/pages/projects/projectRoles/__tests__/ResourcesTreeSelect.spec.tsx
d0w: Not sure if I'm understanding the tests correctly. Though it seems it's only testing what resources get filtered when API groups are selected. But I don't see the reverse: selecting resources should auto select the appropriate API group.

frontend/src/pages/projects/projectRoles/AddRuleModal.tsx
d0w: this still runs twice. While resolvedApiResourceData is still not resolved, the useMemo's initial loop can still run since it relies on a constant causing it to memoize another time once the rest of the api resources are loaded. I would suggest checking to see if apiResourceData is finished loading within the memoization.

frontend/src/pages/projects/projectRoles/AddRuleModal.tsx
```diff
@@ -35,6 +38,86 @@ const AddRuleModal: React.FC<AddRuleModalProps> = ({ existingRule, onSave, onClo
     () => existingRule?.verbs ?? [],
   );
 
+  const resolvedApiResourcesData = apiResourcesLoaded ? apiResourcesData : EMPTY_API_RESOURCES_DATA;
+
+  const resourceToApiGroupMap = React.useMemo(() => {
+    const map = new Map<string, string>();
```
d0w: Also just realized maps can only have unique keys so this will ignore any duplicate resource names. Are we okay with that?

d0w (commented):
at a first glance, the changes work on my end. just a few comments

frontend/src/pages/projects/projectRoles/__tests__/ResourcesTreeSelect.spec.tsx
srtanish1992: The reverse direction (resource selection → auto-populate API group) is tested in `AddRuleModal.spec.tsx` line 94 — there's a full `'auto-populate API groups when resource is selected'` describe block with tests for: adding a resource auto-adds its group, handling wildcard resources, and custom resources with unknown groups. It lives in the AddRuleModal tests because that's where `handleResourcesChange` (the orchestration callback) is defined.

frontend/src/pages/projects/projectRoles/AddRuleModal.tsx
srtanish1992: The double-run is intentional — the first run (with `EMPTY_API_RESOURCES_DATA.resources = []`) still iterates static `RESOURCE_CATEGORIES` to build the map from known mappings. This ensures auto-populate and cascade work for statically-mapped resources even before the API responds. If we skipped it entirely, a user who interacts quickly (e.g., selects 'pods' before discovery completes) would get no auto-populate for the core group. The discovered-data loop (`for (const r of [])`) is a no-op during loading, so the cost is just the static category iteration (~30 entries).

frontend/src/pages/projects/projectRoles/AddRuleModal.tsx
srtanish1992: Yes, this is a known and accepted limitation. In RHOAI/ODH clusters, resource names don't collide across API groups in practice. Static categories take priority (added first), then discovered data fills gaps. We discussed this with CodeRabbit in an earlier review — implementing `Map<string, Set<string>>` was considered but introduces a security concern: auto-populating *all* matching API groups when a user selects a resource could inadvertently grant broader permissions than intended. The single-group mapping is the safer default for RBAC.

d0w:

DaoDaoNoCode:
Screenshot 2026-08-21 at 10 02 48 AM

The 2 dropdown menus could be opened at the same time, which is not ideal here. For the dropdown menu, whenever you click outside, it should trigger a close on the menu. Ideally, the 2 dropdowns could never open at the same time.

srtanish1992:
> Screenshot 2026-08-21 at 10 02 48 AM
> The 2 dropdown menus could be opened at the same time, which is not ideal here. For the dropdown menu, whenever you click outside, it should trigger a close on the menu. Ideally, the 2 dropdowns could never open at the same time.

@DaoDaoNoCode I have fixed this issue.

frontend/src/pages/projects/projectRoles/resourceCategories.ts
```diff
@@ -73,8 +82,57 @@ export const RESOURCE_CATEGORIES: ResourceCategory[] = [
...
+  }
+  for (const category of RESOURCE_CATEGORIES) {
+    for (const r of category.resources) {
+      if (!map.has(r.name)) {
+        map.set(r.name, r.apiGroup);
+      }
+    }
+  }
+  return map;
+};
```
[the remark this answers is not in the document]
srtanish1992: Skipping this. The name-only map is only used for auto-add / cascade in the form (first discovered group wins, core from `/api` first). Save still emits a normal Kubernetes `PolicyRule` with independent `apiGroups` and `resources` lists — that is what the API server enforces, and it is what YAML preview shows.

Group-qualified option IDs in `resources[]` would not be valid Kubernetes resource names. Splitting one Add Rule into paired `{apiGroup, name}` PolicyRules would change the one-rule-per-modal model without fixing an incorrect grant relative to that YAML. Duplicate names (`events`, `pods`) are a known K8s overlap; we are not going to over-complicate the selectors for that.

DaoDaoNoCode:

frontend/src/pages/projects/projectRoles/AddRuleModal.tsx
```diff
@@ -35,6 +38,76 @@ const AddRuleModal: React.FC<AddRuleModalProps> = ({ existingRule, onSave, onClo
...
+
+      for (const resource of addedResources) {
+        const apiGroup = resourceToApiGroupMap.get(resource);
+        if (apiGroup !== undefined && !selectedApiGroups.includes(apiGroup)) {
+          groupsToAdd.add(apiGroup);
+        }
+      }
+
+      if (groupsToAdd.size > 0) {
+        setSelectedApiGroups((prev) => [...prev, ...groupsToAdd]);
```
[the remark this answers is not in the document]
srtanish1992: Valid on the previous updater (closure guard + `[...prev, ...groupsToAdd]`), but not a realistic path here: the tree select fires one `onSelectedResourcesChange` per action, and `groupsToAdd` is a `Set`.

Already addressed in 5fee03d — the updater now skips groups already in `prev`/`next` before pushing. Resolving.

DaoDaoNoCode (approved):
/lgtm
e2e failures are not related

PR #9378, opened 2026-08-18
test(e2e): Move model-serving validation to Model deployment settings spec

mturley:
https://issues.redhat.com/browse/RHOAIENG-85391

## Description

`testAdminClusterSettings.cy.ts` (`Admin should access Cluster Settings and see UI fields matching OpenShift configurations`) failed on every `@ci-dashboard-regression-tags` e2e run. The single-model-serving-platform switch was removed from the **Cluster Settings** page (`/settings/cluster/general`) and now lives only on the **Model deployment settings → General settings** tab (`/settings/model-resources-operations/model-deployment-settings/general-settings`), but the spec still visited Cluster Settings before asserting the switch — so it looked for a control that no longer exists there.

Rather than fix it in place, this moves the model-serving validation to its own feature-area spec so the assertion lives with the page it targets:

- **New** `packages/cypress/cypress/tests/e2e/settings/modelDeploymentSettings/testDeploymentGeneralSettings.cy.ts`:
  - Admin test validating the single-platform switch on the General settings tab against the `OdhDashboardConfig` (the moved assertion).
  - New contributor no-access test: `Test User - should not have access rights to view the Model Deployment Settings tab`.
  - Both `it` blocks keep `@ci-dashboard-regression-tags` / `@SettingsCI`, so the regression suite that surfaced this still covers the assertion — now against the correct page.
- **New** `packages/cypress/cypress/utils/modelDeploymentSettingsUtils.ts`: `validateModelServingPlatforms` moved here from `clusterSettingsUtils.ts`, now navigating to the General settings tab before asserting.
- **Updated** `testAdminClusterSettings.cy.ts`: dropped the stale model-serving validation (still validates PVC size and idle-notebook settings on Cluster Settings).
- **Updated** `clusterSettingsUtils.ts`: removed `validateModelServingPlatforms` and the stale `TODO`.

The new `modelDeploymentSettings/` e2e folder and `modelDeploymentSettingsUtils.ts` also lay the groundwork for moving/creating more Model deployment settings e2e tests as part of [RHOAIENG-69029](https://issues.redhat.com/browse/RHOAIENG-69029) (Admin Setup - e2e tests).

## How Has This Been Tested?

- `npm run type-check` from the repo root — passes (30/30 tasks, including `@odh-dashboard/cypress`).
- `npm run lint` in `packages/cypress` — 0 errors (2 warnings are pre-existing in unrelated files).
[...]

PR #8778, opened 2026-07-23
test: Add E2E test for llm-d routing configurations admin page

YuliaKrimerman:
## Description

Adds  E2E test for the admin routing configurations page (`/settings/model-resources-operations/llmd-routing-configurations`),

Tags: `@Smoke`, `@Dashboard`, `@NonConcurrent`, `@LLMDServingCI`

**Note:** This test requires a cluster with the `LLMInferenceServiceConfig` CRD installed (v1alpha2). It cannot run on clusters without llm-d support (e.g. the rhoai e2e cluster which lacks the CRD). Tested on the zaffre ROSA cluster.

## How Has This Been Tested?

Screenshot 2026-07-23 at 11 13 19 AM

## Test Impact
- No existing tests modified

## Request review criteria:

Self checklist (all need to be checked):
- [x] The developer has manually tested the changes and verified that the changes work
- [x] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [x] The developer has added tests or explained why testing cannot be added (unit or cypress tests for related changes)
- [x] The code follows our [Best Practices](/docs/best-practices.md) (React coding standards, PatternFly usage, performance considerations)

After the PR is posted & before it merges:
- [ ] The developer has tested their solution on a cluster by using the image produced by the PR to `main`

## Summary by CodeRabbit

* **Tests**
  * Added an end-to-end Admin Settings suite for LLMD routing configurations, verifying the configurations table (including the enabled switch) and Edit/Duplicate/Delete flows, and validating cleanup after the run.
  * Extended coverage through the model-serving deploy wizard: sets model location URI, selects generative model type, chooses the LLM-inference-service-llmd deployment method, and confirms the routing dropdown lists the duplicated routing option.
* **Chores**
  * Added new Cypress fixtures for routing configuration and model-serving routing config.
  * Updated Cypress page helpers (routing settings delete confirmation, create-form dropdown interactions, and routing option lookup) and introduced a typed routing test-data shape.

PR #9393, opened 2026-08-19
test(e2e): Consolidate model deployment settings page objects, fix stale tests, add CRUD e2e test for topology

mturley:
https://issues.redhat.com/browse/RHOAIENG-69029

## Description

The Model deployment settings epic ([RHOAIENG-68972](https://issues.redhat.com/browse/RHOAIENG-68972)) consolidated four standalone admin pages into tabs on `/settings/model-resources-operations/model-deployment-settings/` and removed the standalone pages ([RHOAIENG-80077](https://redhat.atlassian.net/browse/RHOAIENG-80077)). This left several E2E tests and their page objects pointing at the old locations. This PR does the E2E-side cleanup, fixes the llm-d admin specs so they pass end-to-end against the migrated tabs, makes the admin specs concurrency-safe, and adds missing live-cluster coverage.

All touched specs were **verified passing end-to-end on a live cluster**, including the deploy-wizard steps.

**1. Consolidate the tab page objects.** Moved the four Cypress page objects for the Model deployment settings tabs into `packages/cypress/cypress/pages/modelDeploymentSettings/` (alongside the already-moved `generalSettings.ts` from https://github.com/opendatahub-io/odh-dashboard/pull/9378) and renamed them to match the tab identifiers:

| From | To | Export |
| --- | --- | --- |
| `pages/servingRuntimes.ts` | `pages/modelDeploymentSettings/servingRuntimeTemplates.ts` | `servingRuntimes` → `servingRuntimeTemplates` |
| `pages/llmAcceleratorConfigs.ts` | `pages/modelDeploymentSettings/llmAcceleratorConfigurations.ts` | `llmAcceleratorConfigs` → `llmAcceleratorConfigurations` |
| `pages/llmdTopologySettings.ts` | `pages/modelDeploymentSettings/topologyConfigurations.ts` | `llmdTopologySettingsPage` → `topologyConfigurations` |
| `pages/llmdRoutingSettings.ts` | `pages/modelDeploymentSettings/routingConfigurations.ts` | `llmdRoutingSettingsPage` → `routingConfigurations` |

Secondary exports (`unsupportedStatusAcceptanceModal`, `llmdRoutingCreatePage`, `deleteRouteModal`) keep their names. All importers were updated, including the scoped `@odh-dashboard/cypress` imports in the `model-serving` mock tests. Moves are tracked as git renames (history preserved).

**2. Fix existing tests broken by the standalone-page removal.**
- `testSingleServingRuntimeCreation.cy.ts`: post-save URL assertion `/serving-runtimes$/` → `/serving-runtime-templates$/` (the form now lands on the serving-runtime-templates tab).
- `servingRuntimeTemplates.ts` + `testAdminClusterSettings.cy.ts`: the serving-runtimes list is now a tab-route page whose title uses `app-tab-page-title`. Added `findTabPageTitle()` for the tab and used it, leaving `findAppTitle()` (`app-page-title`) for the add/edit/duplicate form pages.

**3. Stale-path cleanups.**
- `testGenAiModelDeployment.yaml`: `servingRuntimesPath` → the serving-runtime-templates tab path (the standalone page now only works via a compatibility redirect).
- `testLlmAcceleratorConfigsUnsupported.cy.ts`: dropped the no-op `devFeatureFlags=modelDeploymentSettings=true` (the flag is fully removed), keeping `vLLMDeploymentOnMaaS=true`.

**4. Fix the llm-d routing spec against the migrated tab UI.** `testLlmdRoutingConfigurations.cy.ts` was `@NonConcurrent` since it was added ([#8778](https://github.com/opendatahub-io/odh-dashboard/pull/8778)), so GitHub E2E CI never ran it — and its on-disk selectors had drifted from the actual form. Each of the following was confirmed necessary by tracing the failure to the current component source:
- **Topology-type option testid** `topology-type-workload-single-node` → `workload-single-node`. The `SimpleSelect` renders each option's `data-testid` as its `key`, which is the `TopologyType` enum value (`workload-single-node`) — there was never a `topology-type-` prefix.
- **YAML editor interaction** `.find('textarea')` → a `getYamlEditor()` helper backed by `DashboardCodeEditor.setValue()`. The config field is a PatternFly Monaco `CodeEditor` (`@patternfly/react-code-editor`) with no `<textarea>`.
- **Duplicate name** `${name}-copy` → `copy-of-${name}`. Duplicating produces display name `Copy of <name>`, whose k8s `metadata.name` (and row/option testids) is `copy-of-<name>`. The `after()` cleanup targeted the wrong name too, so the duplicated config was never cleaned up — also fixed.
- **Delete confirmation**: the delete modal gates its danger button on typing the config name, so the spec now types the name before submitting.

**5. Grant the login user project access for the deploy-wizard steps.** The routing and topology admin specs create their test project via `oc` (as a cluster admin) and then drive the deploy wizard logged into the dashboard as `LDAP_ADMIN_USER` — a dashboard admin, not a cluster admin. That user was never a member of the oc-created project, so it never appeared in their Projects list and the "open deploy wizard" step failed with "No results found". Both specs now grant the login user admin access to the project after creation and wait for it to propagate (`addUserToProject` + `waitForUserProjectAccess`), matching the pattern already used by the passing serving-runtime and accelerator-config admin specs. The topology spec is switched to `LDAP_ADMIN_USER` so both llm-d admin specs use the same login user.

**6. Make the admin specs concurrency-safe (drop `@NonConcurrent`).** Three specs were tagged `@NonConcurrent` because they created shared, admin-scoped resources (serving-runtime `Template`s, `LLMInferenceServiceConfig`s) under fixed fixture-derived names, so parallel runs on the same cluster would collide — which meant CI's E2E run (it skips `@NonConcurrent`) never exercised them. Their resource names are now UUID-suffixed so they can run concurrently, and the tag is dropped:
- `testLlmdRoutingConfigurations.cy.ts` — already UUID-scoped every resource; just dropped the tag.
- `testLlmdTopologyConfigurations.cy.ts` — templated the seeded config's `metadata.name` and render+apply it with a UUID-suffixed name.
- `testSingleServingRuntimeCreation.cy.ts` — templated the runtime YAML's name/display-name, rendered with UUID values, and upload it via the real file-picker path (`selectFile` with a file path) so disk-upload coverage is preserved.

**7. Add missing live-cluster coverage: topology config UI-driven CRUD.** New spec `testLlmdTopologyConfigurationsCrud.cy.ts` covers a gap no existing test hit — an admin creating an llm-d topology config **through the UI form** and it persisting as a real `LLMInferenceServiceConfig` CR. (The existing topology e2e only seeds via `oc` and reads back; the mock tests cover form/table/navigation UI but never submit the create form against the API server.) The spec creates a single-node config via the UI (name, resource name, YAML editor), verifies it appears in the table AND persisted on the cluster, then deletes it via the UI and verifies removal. Resource names are UUID-suffixed (concurrency-safe).

**8. Enter the config tabs via the Add button, not the table.** On a clean cluster the routing and topology tabs render an empty state (no table) until the first config exists, so waiting for `*-configurations-table` on arrival timed out in CI (which starts from an empty cluster) even though it passed locally against a populated one. The page objects now wait for the Add button — which the app renders in both the empty state and the populated table header (same `data-testid`) — and the topology CRUD spec drops its up-front table assertion. Post-create table assertions are unchanged, since a config is guaranteed to exist by then.

**9. Tag the llm-d config specs consistently as `@Featureflagged`.** All four llm-d config admin tabs are gated behind the `llmdTemplates` Tech Preview feature flag, so the topology specs now carry `@Featureflagged` like the routing and accelerator specs, and the topology spec's stray `@Smoke` tag (which had no matching `@SmokeSet*` shard) is removed. This follows the established convention that feature-flagged Tech-Preview specs are categorized by the flag rather than slotted into the `@Smoke`/`@Sanity` priority suites — matching the MCP, eval-hub, autorag, and most automl feature-flagged e2e specs.

## How Has This Been Tested?

- **Live cluster (E2E):** all touched specs were run headless against a live OpenShift cluster and pass end-to-end, including the deploy-wizard steps in the routing and topology specs and the new topology CRUD create/persist/delete flow.
- `npm run type-check` from the repo root — passes (30/30 tasks, including `@odh-dashboard/cypress` and `model-serving-cypress`).
- `npm run lint` in `packages/cypress` — 0 errors.
- **E2E CI:** the touched specs carry `@ModelServing` + `@ModelServingCI` (plus their owning-package tags: `@KServeCI` / `@LLMDServingCI`). GitHub E2E auto-detection keys off *changed packages'* `e2eCiTags`, and this PR only changes the Cypress workspaces (`model-serving-cypress`, `@odh-dashboard/cypress`), so the `test:ModelServingCI` label is applied to run the model-serving specs in the GitHub E2E matrix (`testAdminClusterSettings` also runs via the always-on `@ci-dashboard-regression-tags`).

## Test Impact

Test-only change. No product code is modified. Stale E2E assertions that targeted removed standalone pages are repointed to the consolidated tabs; the E2E page objects for those tabs are relocated and renamed for consistency with the tab structure. The llm-d routing spec is fixed to match the migrated tab UI, and both llm-d admin specs now grant their login user project access so the deploy-wizard steps work. Three admin specs are made concurrency-safe (UUID-suffixed resource names) so they run in GitHub E2E CI instead of being skipped as `@NonConcurrent`. One new e2e test adds live-cluster CRUD coverage for the topology configurations tab. No existing test coverage is removed.

## Request review criteria:

[RHOAIENG-69029]: https://redhat.atlassian.net/browse/RHOAIENG-69029?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[RHOAIENG-68972]: https://redhat.atlassian.net/browse/RHOAIENG-68972?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[RHOAIENG-80077]: https://redhat.atlassian.net/browse/RHOAIENG-80077?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

- **Refactor**
  - Updated model deployment settings terminology and navigation for Serving Runtime Templates, Routing Configurations, Topology Configurations, and LLM Accelerator Configurations.
  - Improved configuration workflows with YAML editor access and clearer tab-specific page identification.

- **Tests**
  - Expanded coverage for configuration creation, editing, duplication, deletion, enabling, and unsupported states.
  - Added end-to-end coverage for creating and deleting topology configurations.
  - Improved test isolation with unique configuration and runtime names.
  - Updated route, tab, and model-serving test classifications.

### DIFF
--- a/packages/cypress/cypress/fixtures/resources/modelServing/llmd-topology-config-editor-body.yaml
+++ b/packages/cypress/cypress/fixtures/resources/modelServing/llmd-topology-config-editor-body.yaml
@@ -1,0 +1,23 @@
+# Minimal valid llm-d topology config spec pasted into the create form's YAML
+# editor by testLlmdTopologyConfigurationsCrud.cy.ts. The config's metadata.name
+# comes from the form's resource-name field, so this body only needs a valid spec.
+apiVersion: serving.kserve.io/v1alpha2
+kind: LLMInferenceServiceConfig
+metadata:
+  name: placeholder-overridden-by-form
+  labels:
+    opendatahub.io/config-type: workload-single-node
+    opendatahub.io/dashboard: "true"
+spec:
+  replicas: 1
+  template:
+    containers:
+      - name: main
+        image: quay.io/pierdipi/vllm-cpu:latest
+        resources:
+          limits:
+            cpu: "2"
+            memory: 8Gi
+          requests:
+            cpu: "1"
+            memory: 4Gi
--- a/packages/cypress/cypress/tests/e2e/settings/clusterSettings/testAdminClusterSettings.cy.ts
+++ b/packages/cypress/cypress/tests/e2e/settings/clusterSettings/testAdminClusterSettings.cy.ts
@@ -9,7 +9,7 @@ import { storageClassesPage } from '../../../../pages/storageClasses';
 import { notebookImageSettings } from '../../../../pages/notebookImageSettings';
 import { hardwareProfile } from '../../../../pages/hardwareProfile';
 import { connectionTypesPage } from '../../../../pages/connectionTypes';
-import { servingRuntimes } from '../../../../pages/servingRuntimes';
+import { servingRuntimeTemplates } from '../../../../pages/modelDeploymentSettings/servingRuntimeTemplates';
 import { modelRegistrySettings } from '../../../../pages/modelRegistrySettings';
 import type {
   CommandLineResult,
@@ -222,7 +222,18 @@ describe('Verify that only the Cluster Admin can access Cluster Settings', () =>
 
   it(
     'Cluster Admin user can access all Settings pages',
-    { tags: ['@Smoke', '@SmokeSet2', '@ODS-1688', '@Dashboard', '@SettingsCI'] },
+    {
+      tags: [
+        '@Smoke',
+        '@SmokeSet2',
+        '@ODS-1688',
+        '@Dashboard',
+        '@ModelServing',
+        '@SettingsCI',
+        '@KServeCI',
+        '@ModelServingCI',
+      ],
+    },
     function testClusterAdminSettingsAccess() {
       // Skip on BYOIDC clusters - User Management page requires groups API which returns 404 on BYOIDC
       skipIfBYOIDC(
@@ -257,8 +268,8 @@ describe('Verify that only the Cluster Admin can access Cluster Settings', () =>
       connectionTypesPage.findPageTitle().should('exist');
 
       cy.step('Access Settings -> Model resources and operations -> Serving runtimes');
-      servingRuntimes.visit();
-      servingRuntimes.findAppTitle().should('exist');
+      servingRuntimeTemplates.visit();
+      servingRuntimeTemplates.findTabPageTitle().should('exist');
 
       cy.step('Access Settings -> Model resources and operations -> Model registry settings');
       modelRegistrySettings.visit();
--- a/packages/cypress/cypress/tests/e2e/settings/clusterSettings/testAdminClusterSettings.cy.ts
+++ b/packages/cypress/cypress/tests/e2e/settings/clusterSettings/testAdminClusterSettings.cy.ts
@@ -222,18 +222,7 @@ describe('Verify that only the Cluster Admin can access Cluster Settings', () =>
 
   it(
     'Cluster Admin user can access all Settings pages',
-    {
-      tags: [
-        '@Smoke',
-        '@SmokeSet2',
-        '@ODS-1688',
-        '@Dashboard',
-        '@ModelServing',
-        '@SettingsCI',
-        '@KServeCI',
-        '@ModelServingCI',
-      ],
-    },
+    { tags: ['@Smoke', '@SmokeSet2', '@ODS-1688', '@Dashboard', '@SettingsCI'] },
     function testClusterAdminSettingsAccess() {
       // Skip on BYOIDC clusters - User Management page requires groups API which returns 404 on BYOIDC
       skipIfBYOIDC(
--- a/packages/cypress/cypress/tests/e2e/settings/llmdTopologyConfigurations/testLlmdTopologyConfigurations.cy.ts
+++ b/packages/cypress/cypress/tests/e2e/settings/llmdTopologyConfigurations/testLlmdTopologyConfigurations.cy.ts
@@ -63,7 +63,7 @@ describe('LLMD Topology Configurations - Admin Settings', () => {
   it(
     'Admin can manage topology configurations and verify wizard visibility',
     {
-      tags: ['@Smoke', '@Dashboard', '@ModelServing', '@LLMDServingCI', '@ModelServingCI'],
+      tags: ['@Featureflagged', '@Dashboard', '@ModelServing', '@LLMDServingCI', '@ModelServingCI'],
     },
     () => {
       cy.step('Log in with topology configs feature flag');
--- a/packages/cypress/cypress/tests/e2e/settings/llmdTopologyConfigurations/testLlmdTopologyConfigurations.cy.ts
+++ b/packages/cypress/cypress/tests/e2e/settings/llmdTopologyConfigurations/testLlmdTopologyConfigurations.cy.ts
@@ -1,15 +1,15 @@
 import { HTPASSWD_CLUSTER_ADMIN_USER } from '../../../../utils/e2eUsers';
 import { retryableBefore } from '../../../../utils/retryableHooks';
 import { topologyConfigurations } from '../../../../pages/modelDeploymentSettings/topologyConfigurations';
-import {
-  createCleanLLMInferenceServiceConfig,
-  cleanupLLMInferenceServiceConfig,
-} from '../../../../utils/oc_commands/llmInferenceServiceConfig';
+import { cleanupLLMInferenceServiceConfig } from '../../../../utils/oc_commands/llmInferenceServiceConfig';
 import { projectDetails, projectListPage } from '../../../../pages/projects';
 import { modelServingGlobal, modelServingWizard } from '../../../../pages/modelServing';
 import { ModelLocationSelectOption, ModelTypeLabel } from '../../../../utils/modelServingConstants';
 import { createCleanProject } from '../../../../utils/projectChecker';
 import { deleteOpenShiftProject } from '../../../../utils/oc_commands/project';
+import { applyOpenShiftYaml } from '../../../../utils/oc_commands/baseCommands';
+import { renderYamlFileWithReplacements } from '../../../../utils/oc_commands/templates';
+import { getFixturePath } from '../../../../utils/fileImportUtils';
 import { generateTestUUID } from '../../../../utils/uuidGenerator';
 import { loadDSPFixture } from '../../../../utils/dataLoader';
 import type { DataScienceProjectData } from '../../../../types';
@@ -22,6 +22,9 @@ type TopologyTestData = DataScienceProjectData & {
 let testData: TopologyTestData;
 const uuid = generateTestUUID();
 let projectName: string;
+// Unique per-run config name so the test is safe to run concurrently.
+let topologyConfigName: string;
+const applicationNamespace = Cypress.env('APPLICATIONS_NAMESPACE');
 
 describe('LLMD Topology Configurations - Admin Settings', () => {
   retryableBefore(() => {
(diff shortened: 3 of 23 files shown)


packages/cypress/cypress/tests/e2e/settings/clusterSettings/testAdminClusterSettings.cy.ts
```diff
@@ -222,7 +222,18 @@ describe('Verify that only the Cluster Admin can access Cluster Settings', () =>
...
   it(
     'Cluster Admin user can access all Settings pages',
-    { tags: ['@Smoke', '@SmokeSet2', '@ODS-1688', '@Dashboard', '@SettingsCI'] },
+    {
+      tags: [
+        '@Smoke',
+        '@SmokeSet2',
+        '@ODS-1688',
+        '@Dashboard',
+        '@ModelServing',
```
sridarna: Not sure, Does the Settings Page come under ModelServing Component. We are also not calling Deployment page/model Deployment wizard in this E2E. So I think those tags are not needed.

mturley: Ah, that was a stray add when I was adding it to the other tests. Will remove.

mturley: Pushed

packages/cypress/cypress/tests/e2e/settings/llmdTopologyConfigurations/testLlmdTopologyConfigurations.cy.ts
```diff
@@ -22,40 +26,55 @@ type TopologyTestData = DataScienceProjectData & {
...
+        // login user admin access so it appears in their Projects list.
+        addUserToProject(projectName, LDAP_ADMIN_USER.USERNAME, 'admin');
+        return waitForUserProjectAccess(projectName, LDAP_ADMIN_USER.USERNAME);
+      });
   });
 
   after(() => {
-    cleanupLLMInferenceServiceConfig(testData.topologyConfigName);
+    cleanupLLMInferenceServiceConfig(topologyConfigName);
     deleteOpenShiftProject(projectName, { wait: true, ignoreNotFound: true, timeout: 300000 });
```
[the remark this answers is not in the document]
mturley: Leaving this as-is. Both cleanup helpers delete with `--ignore-not-found`, so if the names are `undefined` the commands are no-ops, since nothing is named `undefined`, and nothing destructive happens. It also matches the rest of the admin-settings suite, where the serving-runtime, accelerator, and routing specs all use the same unguarded `after()`.

packages/cypress/cypress/tests/e2e/settings/llmdTopologyConfigurations/testLlmdTopologyConfigurations.cy.ts
```diff
@@ -22,40 +26,55 @@ type TopologyTestData = DataScienceProjectData & {
...
   });
 
   it(
     'Admin can manage topology configurations and verify wizard visibility',
-    { tags: ['@Smoke', '@Dashboard', '@NonConcurrent', '@LLMDServingCI'] },
+    {
+      tags: ['@Smoke', '@Dashboard', '@ModelServing', '@LLMDServingCI', '@ModelServingCI'],
```
sridarna: @mturley This should have tag '@SmokeSet3' added or remove the smoke tag and add tag Featureflagged to be able to run in our e2e test analysis.

packages/cypress/cypress/tests/e2e/settings/llmdTopologyConfigurations/testLlmdTopologyConfigurationsCrud.cy.ts
```diff
@@ -0,0 +1,94 @@
...
+    });
+  });
+
+  after(() => {
+    cleanupLLMInferenceServiceConfig(resourceName);
+  });
+
+  it(
+    'Admin creates a topology config via the UI, it persists on the cluster, and delete removes it',
+    { tags: ['@Dashboard', '@ModelServing', '@LLMDServingCI', '@ModelServingCI'] },
```
sridarna: Add a Smoke,SmokeSet3 or Featureflagged Tag to be visible in E2E analysis test report.

packages/cypress/cypress/tests/e2e/settings/llmdTopologyConfigurations/testLlmdTopologyConfigurations.cy.ts
sridarna: SmokeSet3 is suggested here because usually model serving tests exists there.
But it can be anywhere. We can also consider Sanity tag if we feel these features are medium priority.
I think we can consider Sanity for the tests.

packages/cypress/cypress/fixtures/resources/modelServing/llmd-topology-config-editor-body.yaml
```diff
@@ -0,0 +1,23 @@
...
+  name: placeholder-overridden-by-form
+  labels:
+    opendatahub.io/config-type: workload-single-node
+    opendatahub.io/dashboard: "true"
+spec:
+  replicas: 1
+  template:
+    containers:
+      - name: main
+        image: quay.io/pierdipi/vllm-cpu:latest
```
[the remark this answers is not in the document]
mturley: Leaving this as `:latest`. This image reference is inert test data — this spec creates, reads, and deletes an `LLMInferenceServiceConfig` (a config CR) and never deploys a model, so no controller ever pulls the image. The same `quay.io/pierdipi/vllm-cpu:latest` tag is already used across the suite (the sibling `llmd-topology-config.yaml`, `llmd-inference-service-config.yaml`, `testVLLMOnMaaS`, the MaaS fixtures, and `mockLLMInferenceServiceK8sResource`), so pinning only this copy would diverge from the established convention for this image. The CWE-829 concern doesn't apply to a fixture that's never fetched.

packages/cypress/cypress/tests/e2e/settings/llmdTopologyConfigurations/testLlmdTopologyConfigurationsCrud.cy.ts
```diff
@@ -0,0 +1,91 @@
...
+      );
+      checkLLMInferenceServiceConfigState(resourceName, namespace);
+
+      cy.step('Delete the config from the UI');
+      topologyConfigurations.getRow(resourceName).findKebabAction('Delete').click();
+      deleteModal.findInput().clear().type(displayName);
+      deleteModal.findSubmitButton().should('be.enabled').click();
+
+      cy.step('Verify the config is removed from the table');
+      topologyConfigurations.getRow(resourceName).find().should('not.exist');
```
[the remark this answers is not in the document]
mturley: Leaving as-is. The create path is already validated against the cluster (not just the UI): `checkLLMInferenceServiceConfigState(resourceName, namespace)` confirms the config persisted as a real CR. For deletion, `getRow(resourceName).should('not.exist')` is the supported absence assertion and matches how the sibling routing and topology specs assert removal, so this follows the suite convention. `after()`'s `cleanupLLMInferenceServiceConfig` is idempotent best-effort teardown, not a mechanism that would mask a failed UI deletion.

sridarna:
Tested in Jenkins too.
job/dashboard/job/dashboard-e2e-tests/2375/Test_20Reports/
The failures are unrelated to PR.

/lgtm
/approve
Screenshot 2026-08-20 at 1 00 38 PM
Screenshot 2026-08-20 at 5 14 51 PM

mturley:
/approve

Looks like Sri’s didn’t go through

PR #9403, opened 2026-08-19
fix(gen-ai): Remove port restrictions from cross-namespace egress to support MCP servers

NickGagan:
https://redhat.atlassian.net/browse/RHOAIENG-85502

## Description

MCP servers can run on any port in any cluster namespace. The previous `gen-ai-allow-ports` NetworkPolicy restricted cross-namespace egress to only ports 8321 and 8443, which blocked connectivity to MCP servers on other ports.

This removes the port restriction from the `namespaceSelector: {}` egress rule, restoring the original open-namespace behaviour that existed before the OGX/NemoGuardrails port additions narrowed it.

## How Has This Been Tested?

Verified locally on a cluster — gen-ai-ui pod can now reach MCP servers in arbitrary namespaces on arbitrary ports.

## Test Impact

No test changes required — this is a manifest-only change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated network access rules to allow outbound traffic on all ports across namespaces.
  * Preserved existing DNS, application-specific, and external HTTPS/HTTP access rules.

### DIFF
--- a/manifests/modules/gen-ai/networkpolicy.yaml
+++ b/manifests/modules/gen-ai/networkpolicy.yaml
@@ -43,11 +43,6 @@ spec:
           protocol: TCP
     - to:
         - namespaceSelector: {}
-      ports:
-        - port: 8321
-          protocol: TCP
-        - port: 8443
-          protocol: TCP
     - to:
         - ipBlock:
             cidr: 0.0.0.0/0


lucferbux (approved):
/lgtm
/approve

@NickGagan we need to create a jira follow up to cover the proper implementation of the feature

andyatmiami (approved):
/lgtm
/approve

---

Discussions in a [very long slack thread ](https://redhat-internal.slack.com/archives/C05SMJ09DD2/p1784725360585439)(that probably isn't accessible to non-Red Hatters) has collectively led us to the conclusion this is the best change we can deliver to unblock 3.5.

External endpoints, etc - more "dynamic resources" - makes this challenging to statically define ports a priori.. so for now we'll relax this restriction and moving forward look how to lock this back down.

PR #9439, opened 2026-08-20
fix(RHOAIENG-87447): Support Markdown rendering and improve HTML visualization in artifact drawer

DaoDaoNoCode:
https://redhat.atlassian.net/browse/RHOAIENG-87447
https://redhat.atlassian.net/browse/RHOAIENG-56884

## Description

https://github.com/user-attachments/assets/c37e2324-d62e-4db4-9963-44c2b1ad9aeb

Fixes two related issues with the artifact Visualization tab in pipeline run details:

1. **Markdown rendered as plain text** ([RHOAIENG-87447](https://redhat.atlassian.net/browse/RHOAIENG-87447)): `Output[Markdown]` artifacts were loaded in an iframe that displayed raw markdown text without formatting. Now, markdown artifacts are fetched as raw text and rendered natively using the project's `MarkdownComponent` (react-markdown + remark-gfm + rehype-sanitize). Falls back to iframe if the fetch fails.

2. **HTML visualization unreadable** ([RHOAIENG-56884](https://redhat.atlassian.net/browse/RHOAIENG-56884)): HTML artifacts (e.g. AutoML leaderboard tables) were rendered in an iframe with no width/height styling, making them nearly invisible. The iframe now has proper sizing (`width: 100%` via PF utility class, `min-height: 300px`).

3. **Expand to fullscreen**: Both HTML and Markdown artifacts now have an expand button that opens a Modal at `variant="large"`, allowing users to view the full content outside the narrow drawer panel.

All changes are also applied to the compare-runs markdown view (`MarkdownCompare` + `useFetchMarkdownMaps`).

### Files changed

| File | Change |
|------|--------|
| `ArtifactVisualization.tsx` | Split MARKDOWN vs HTML rendering paths; markdown uses `MarkdownComponent`, HTML keeps iframe; both get expand button + fullscreen modal |
| `ArtifactVisualization.scss` | New — iframe sizing and markdown scroll containers (only styles PF can't handle) |
| `MarkdownCompare.tsx` | Extended `MarkdownAndTitle` type with `markdownContent?`; renders `MarkdownComponent` when available; fullscreen modal |
| `useFetchMarkdownMaps.ts` | For MARKDOWN artifacts, fetches raw text via download URL; passes `markdownContent` through to config |

## How Has This Been Tested?

- Manually tested with a KFP pipeline that produces both `system.Markdown` and `system.HTML` artifacts
- Verified markdown renders with proper formatting (headings, tables, code blocks, lists)
- Verified HTML iframe renders with correct sizing
- Verified expand button opens fullscreen modal for both types
- Verified fallback to iframe when markdown content fetch fails
- Verified existing Cypress tests are unaffected (fallback chain preserves iframe behavior)

## Test Impact

Existing Cypress tests (`artifacts.cy.ts`, `compareRuns.cy.ts`) continue to pass — the download URL fallback ensures the iframe path is still exercised when the download URL is not mocked.

## Request review criteria:

If you have UI changes:

[RHOAIENG-87447]: https://redhat.atlassian.net/browse/RHOAIENG-87447?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[RHOAIENG-56884]: https://redhat.atlassian.net/browse/RHOAIENG-56884?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Markdown artifacts and comparisons now render directly in the interface when available.
  * Added expand controls and fullscreen modal views for artifact visualizations.
  * Large or unavailable Markdown content automatically falls back to the existing viewer.

* **Style**
  * Improved sizing, scrolling, borders, and fullscreen layouts for clearer visualization.

* **Tests**
  * Added coverage for inline Markdown rendering and fallback behavior.

### DIFF
--- a/frontend/src/concepts/pipelines/content/compareRuns/metricsSection/markdown/useFetchMarkdownMaps.ts
+++ b/frontend/src/concepts/pipelines/content/compareRuns/metricsSection/markdown/useFetchMarkdownMaps.ts
@@ -7,6 +7,7 @@ import {
   getFullArtifactPaths,
 } from '#~/concepts/pipelines/content/compareRuns/metricsSection/utils';
 import { ArtifactType, PipelineRunKF } from '#~/concepts/pipelines/kfTypes';
+import { readBoundedText } from '#~/concepts/pipelines/content/pipelinesDetails/pipelineRun/artifacts/utils';
 import { allSettledPromises } from '#~/utilities/allSettledPromises';
 
 const useFetchMarkdownMaps = (
@@ -17,7 +18,7 @@ const useFetchMarkdownMaps = (
   configsLoaded: boolean;
 } => {
   const [configsLoaded, setConfigsLoaded] = React.useState(false);
-  const { getStorageObjectRenderUrl } = useArtifactStorage();
+  const { getStorageObjectRenderUrl, getStorageObjectDownloadUrl } = useArtifactStorage();
 
   const [configMapBuilder, setConfigMapBuilder] = React.useState<
     Record<string, MarkdownAndTitle[]>
@@ -40,22 +41,39 @@ const useFetchMarkdownMaps = (
           const { run } = path;
           let sizeBytes: number | undefined;
           let url: string | undefined;
-          if (
-            path.linkedArtifact.artifact.getType() === ArtifactType.MARKDOWN ||
-            path.linkedArtifact.artifact.getType() === ArtifactType.HTML
-          ) {
+          let markdownContent: string | undefined;
+          const artifactType = path.linkedArtifact.artifact.getType();
+
+          if (artifactType === ArtifactType.MARKDOWN) {
+            try {
+              const downloadUrl = await getStorageObjectDownloadUrl(path.linkedArtifact.artifact);
+              if (downloadUrl) {
+                const response = await fetch(downloadUrl);
+                if (response.ok) {
+                  markdownContent = await readBoundedText(response);
+                }
+              }
+            } catch {
+              // Fall back to render URL for iframe display
+            }
+            if (!markdownContent) {
+              url = await getStorageObjectRenderUrl(path.linkedArtifact.artifact).catch(
+                () => undefined,
+              );
+            }
+          } else if (artifactType === ArtifactType.HTML) {
             url = await getStorageObjectRenderUrl(path.linkedArtifact.artifact).catch(
               () => undefined,
             );
           }
 
-          if (url === undefined) {
+          if (url === undefined && markdownContent === undefined) {
             return null;
           }
-          return { run, sizeBytes, url, path };
+          return { run, sizeBytes, url, markdownContent, path };
         }),
 
-    [fullArtifactPaths, getStorageObjectRenderUrl],
+    [fullArtifactPaths, getStorageObjectRenderUrl, getStorageObjectDownloadUrl],
   );
 
   React.useEffect(() => {
@@ -66,13 +84,14 @@ const useFetchMarkdownMaps = (
     allSettledPromises(fetchStorageObjectPromises).then(([successes]) => {
       successes.forEach((result) => {
         if (result.value) {
-          const { url, sizeBytes, run, path } = result.value;
+          const { url, markdownContent: mdContent, sizeBytes, run, path } = result.value;
           setRunMapBuilder((runMap) => ({ ...runMap, [run.run_id]: run }));
 
-          const config = {
+          const config: MarkdownAndTitle = {
             title: getFullArtifactPathLabel(path),
-            config: url,
+            config: url ?? '',
             fileSize: sizeBytes,
+            markdownContent: mdContent,
           };
 
           setConfigMapBuilder((configMap) => ({
--- a/frontend/src/concepts/pipelines/content/pipelinesDetails/pipelineRun/artifacts/ArtifactVisualization.scss
+++ b/frontend/src/concepts/pipelines/content/pipelinesDetails/pipelineRun/artifacts/ArtifactVisualization.scss
@@ -1,0 +1,16 @@
+.odh-artifact-visualization {
+  &__iframe {
+    min-height: 300px;
+    border: none;
+  }
+
+  &__iframe--fullscreen {
+    height: 75vh;
+    border: none;
+  }
+
+  &__markdown--fullscreen {
+    max-height: 75vh;
+    overflow: auto;
+  }
+}
--- a/frontend/src/concepts/pipelines/content/pipelinesDetails/pipelineRun/artifacts/ArtifactVisualization.tsx
+++ b/frontend/src/concepts/pipelines/content/pipelinesDetails/pipelineRun/artifacts/ArtifactVisualization.tsx
@@ -1,16 +1,26 @@
 import React from 'react';
 
 import {
+  Button,
   Bullseye,
   EmptyState,
   EmptyStateBody,
   EmptyStateVariant,
   Flex,
+  FlexItem,
+  // eslint-disable-next-line @odh-dashboard/no-restricted-imports -- ContentModal adds a constrained body height that causes double scrollbars with the iframe/markdown scroll containers
+  Modal,
+  // eslint-disable-next-line @odh-dashboard/no-restricted-imports
+  ModalBody,
+  // eslint-disable-next-line @odh-dashboard/no-restricted-imports
+  ModalHeader,
   Spinner,
   Stack,
   StackItem,
   Title,
+  Tooltip,
 } from '@patternfly/react-core';
+import { ExpandIcon } from '@patternfly/react-icons';
 import { TableVariant, Td, Tr } from '@patternfly/react-table';
 
 import { Table } from '@odh-dashboard/ui-core';
@@ -27,7 +37,9 @@ import { buildConfusionMatrixConfig } from '#~/concepts/pipelines/content/artifa
 import { isConfusionMatrix } from '#~/concepts/pipelines/content/compareRuns/metricsSection/confusionMatrix/utils';
 import { usePipelinesAPI } from '#~/concepts/pipelines/context';
 import { useArtifactStorage } from '#~/concepts/pipelines/apiHooks/useArtifactStorage';
-import { getArtifactProperties } from './utils';
+import MarkdownComponent from '#~/components/markdown/MarkdownComponent';
+import { getArtifactProperties, readBoundedText } from './utils';
+import './ArtifactVisualization.scss';
 
 interface ArtifactVisualizationProps {
   artifact: Artifact;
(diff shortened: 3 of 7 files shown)


srtanish1992:
@DaoDaoNoCode I did some manual testing, the bugs are fixed but below are few issues I have found.

1. When comparing runs, there is now dual expand action buttons/icons. Maybe conditionally hide when on compare runs page?

Screenshot 2026-08-20 at 6 11 17 PM

2. If the pipeline server storage has strict CORS and doesn't allow dashboard origin then the `fetch` will fail and the Markdown content will be displayed as a raw content. Now I am not sure about the customer data and how they create the storage OR is it okay to put this in documentation and ask them to fix CORS policy. The right fix would be in DSPA to provide an endpoint to stream bytes so that frontend can just consume it OR a BFF for pipelines UI.

Screenshot 2026-08-20 at 7 12 53 PM

DaoDaoNoCode:
Hey @srtanish1992 Thanks for the review!
For 1, good catch and I submitted the fix.
For 2, you're right that strict CORS on the object store would cause the `fetch()` to fail. In that case our fallback kicks in and renders via iframe (same behavior as before this PR — no regression). The ideal fix is a DSPA content-streaming endpoint or a dashboard BFF proxy so the frontend doesn't need direct object store access. That's a separate effort, I think we can have a pipeline BFF for this once we move the pipelines into a package. For this PR, the fallback chain ensures no worse behavior than the status quo.

frontend/src/concepts/pipelines/content/pipelinesDetails/pipelineRun/artifacts/ArtifactVisualization.tsx
```diff
@@ -47,17 +61,45 @@ export const ArtifactVisualization: React.FC<ArtifactVisualizationProps> = ({ ar
...
         const renderArtifact = async () => {
-          await getStorageObjectRenderUrl(memoizedArtifact)
-            .then((url) => setRenderUrl(url))
-            .catch(() => null);
+          if (artifactType === ArtifactType.MARKDOWN) {
+            let content: string | undefined;
+            try {
+              const url = await getStorageObjectDownloadUrl(memoizedArtifact);
+              if (url) {
+                const response = await fetch(url);
```
srtanish1992: `fetch()` of the presigned `download_url` is cross-origin (S3/MinIO). Iframe `src` is not.

If the bucket has no CORS (default **AWS S3 / IBM COS**) or a strict allow-list that omits the dashboard origin, this fails and the iframe fallback is **RHOAIENG-87447** again (plaintext markdown). Default in-cluster MinIO (`*`) is why this is easy to miss locally. Customers often have stricter CORS; we should not require them to open the bucket.

Please load the bytes same-origin (DSPA streams the object, or dashboard backend GETs the signed URL **by artifact id**). Do not treat iframe fallback as success for markdown. Same pattern in `useFetchMarkdownMaps.ts`.

frontend/src/concepts/pipelines/content/compareRuns/metricsSection/markdown/useFetchMarkdownMaps.ts
```diff
@@ -40,22 +40,39 @@ const useFetchMarkdownMaps = (
...
-            path.linkedArtifact.artifact.getType() === ArtifactType.HTML
-          ) {
+          let markdownContent: string | undefined;
+          const artifactType = path.linkedArtifact.artifact.getType();
+
+          if (artifactType === ArtifactType.MARKDOWN) {
+            try {
+              const downloadUrl = await getStorageObjectDownloadUrl(path.linkedArtifact.artifact);
+              if (downloadUrl) {
+                const response = await fetch(downloadUrl);
```
srtanish1992, line 51: Same CORS-sensitive `fetch(downloadUrl)` as `ArtifactVisualization`. Compare-runs markdown will also silently fall back to the plaintext iframe when object storage blocks the dashboard origin.

Load content same-origin (DSPA or dashboard backend keyed by artifact id), not via browser `fetch` of the signed URL.

frontend/src/concepts/pipelines/content/pipelinesDetails/pipelineRun/artifacts/ArtifactVisualization.tsx
```diff
@@ -47,17 +61,45 @@ export const ArtifactVisualization: React.FC<ArtifactVisualizationProps> = ({ ar
...
+            await getStorageObjectRenderUrl(memoizedArtifact)
+              .then((url) => setRenderUrl(url))
+              .catch(() => null);
+          }
           setLoading(false);
         };
         setLoading(true);
         setRenderUrl(undefined);
+        setMarkdownContent(undefined);
         renderArtifact();
```
srtanish1992: This effect has no abort / `cancelled` flag. In-flight `getArtifact` + `fetch` can still call `setMarkdownContent` after unmount or after the artifact changes.

In `start:dev:ext` (`React.StrictMode`) this also shows up as duplicate `view=DOWNLOAD` + S3 calls (same payload/timing). Production will not Strict-Mode-double, but switching nodes has the same missing cleanup.

Please abort on effect cleanup. Same for `useFetchMarkdownMaps`.

frontend/src/concepts/pipelines/content/pipelinesDetails/pipelineRun/artifacts/ArtifactVisualization.scss
```diff
@@ -0,0 +1,21 @@
...
+    border: none;
+  }
+
+  &__iframe--fullscreen {
+    height: 75vh;
+    border: none;
+  }
+
+  &__markdown {
+    max-height: 400px;
```
srtanish1992: RHOAIENG-87447 asks for markdown to be **fully visible**. Native `MarkdownComponent` is the right renderer, but `max-height: 400px` still clips long artifacts behind a scrollbar.

Can we drop this cap (or only constrain height via the drawer layout) so long markdown is actually fully visible?

frontend/src/concepts/pipelines/content/pipelinesDetails/pipelineRun/artifacts/ArtifactVisualization.tsx
```diff
@@ -154,20 +196,75 @@ export const ArtifactVisualization: React.FC<ArtifactVisualizationProps> = ({ ar
...
+            <Flex
+              justifyContent={{ default: 'justifyContentSpaceBetween' }}
+              alignItems={{ default: 'alignItemsCenter' }}
+            >
+              <FlexItem>
+                <Title headingLevel="h3">Artifact details</Title>
+              </FlexItem>
+              <FlexItem>
+                <Tooltip content="Expand to fullscreen">
+                  <Button
```
srtanish1992, line 263: Nit: tooltip says “Expand to fullscreen” but the modal is `variant="large"` (not fullscreen). Please align the copy with the actual modal, or use a fullscreen variant if that was the intent.

frontend/src/concepts/pipelines/content/pipelinesDetails/pipelineRun/artifacts/__tests__/ArtifactNodeDrawerContent.spec.tsx
```diff
@@ -10,6 +10,11 @@ import { PipelineTask } from '#~/concepts/pipelines/topology';
 import { Artifact, Value } from '#~/third_party/mlmd';
 import { ArtifactNodeDrawerContent } from '#~/concepts/pipelines/content/pipelinesDetails/pipelineRun/artifacts/ArtifactNodeDrawerContent';
 
+jest.mock('#~/components/markdown/MarkdownComponent', () => ({
```
srtanish1992, line 13: This only mocks `MarkdownComponent` so the drawer still mounts. It does not cover native markdown vs iframe fallback, expand/modal, or a failed `fetch`.

The new Visualization paths need unit tests (same gap on `PipelineRunDrawerRightContent.spec.tsx`). Codecov is already flagging this patch.

srtanish1992 (changes requested):
Thanks for dropping the second Expand on Compare — that looks fixed.

Remaining issues are inline. The important one is still `fetch(presigned download_url)`: iframe fallback on CORS failure is the original markdown plaintext bug, not an acceptable success path for RHOAIENG-87447. Please load artifact bytes same-origin (DSPA stream or dashboard backend keyed by artifact id). Also missing tests for the native markdown / fallback / expand paths.

srtanish1992:
> Hey @srtanish1992 Thanks for the review! For 1, good catch and I submitted the fix. For 2, you're right that strict CORS on the object store would cause the `fetch()` to fail. In that case our fallback kicks in and renders via iframe (same behavior as before this PR — no regression). The ideal fix is a DSPA content-streaming endpoint or a dashboard BFF proxy so the frontend doesn't need direct object store access. That's a separate effort, I think we can have a pipeline BFF for this once we move the pipelines into a package. For this PR, the fallback chain ensures no worse behavior than the status quo.

For the 2nd one, I agree that it is out of scope of this PR but I would inform the person who reported this issue and PM that in the scenario where there are stricter policies, the "bug" will be back because if that "customer" has CORS policy that doesn't allow dashboard origin than they know what to do and have the right expectation about this fix.

I just finished code review. Feel free to ignore the feedback related to CORS.

frontend/src/concepts/pipelines/content/pipelinesDetails/pipelineRun/artifacts/ArtifactVisualization.tsx
```diff
@@ -27,37 +37,89 @@ import { buildConfusionMatrixConfig } from '#~/concepts/pipelines/content/artifa
...
+
+    const renderArtifact = async () => {
+      if (artifactType === ArtifactType.MARKDOWN) {
+        let content: string | undefined;
+        try {
+          const url = await getStorageObjectDownloadUrl(memoizedArtifact);
+          if (url && !signal.aborted) {
+            const response = await fetch(url, { signal });
+            if (response.ok) {
+              content = await response.text();
```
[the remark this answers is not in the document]
DaoDaoNoCode: Added a Content-Length guard (5MB cap) before calling response.text(). If the header declares a size over MAX_MARKDOWN_BYTES, we skip decoding and fall back to the render URL iframe. This handles the common case where S3 pre-signed URLs include Content-Length in the response.

I opted for the simpler Content-Length check over the streaming reader approach — the S3/MinIO responses from pipeline artifacts reliably include this header, and a full ReadableStream chunked reader adds complexity for a scenario (missing Content-Length on a multi-MB markdown artifact) that's unlikely in practice.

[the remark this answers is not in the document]
DaoDaoNoCode: Addressed — replaced the Content-Length-only check with a shared readBoundedText utility that enforces the 5MB cap via streaming ReadableStream reader when Content-Length is absent. When the header is present and within limits, it uses response.text() directly. The utility is shared between both call sites in utils.ts.

srtanish1992:
/lgtm

@DaoDaoNoCode do you wan't to create ticket for fixing the CORS issue for documentation purpose?

DaoDaoNoCode:
@srtanish1992 Hi, I'd prefer to do that when we refactoring the pipelines package. We will remember that haha... 😛

PR #9447, opened 2026-08-21
fix(RHOAIENG-85544): Add changes requested by UX sign-off

Philip-Carneiro:
https://redhat.atlassian.net/browse/RHOAIENG-85544

## Description

Add changes requested by UX
Before:
before-2
before-3

After:
after-2
image

## How Has This Been Tested?

It was manually tested, no test added since it's just cosmetics.

## Request review criteria:

If you have UI changes: 

## Summary by CodeRabbit

* **New Features**
  * Runtime argument fields are now always visible in the deployment wizard.
  * Runtime argument settings are automatically inferred from entered content, removing the separate enablement checkbox.
  * Runtime arguments can be edited when creation is allowed and are read-only otherwise.
  * Improved spacing and readability in validated configuration details.

* **Tests**
  * Updated deployment workflows and validation checks to reflect the always-visible runtime argument field.

### DIFF
--- a/packages/model-serving/cypress/tests/mocked/modelServing/modelServingNim.cy.ts
+++ b/packages/model-serving/cypress/tests/mocked/modelServing/modelServingNim.cy.ts
@@ -462,7 +462,6 @@ describe('NIM Models Deployments', () => {
     modelServingWizardEdit.findExternalRouteCheckbox().should('be.checked');
     modelServingWizardEdit.findTokenAuthenticationCheckbox().should('be.checked');
     modelServingWizardEdit.findServiceAccountNameInput().should('have.value', 'my-existing-sa');
-    modelServingWizardEdit.findRuntimeArgsCheckbox().should('be.checked');
     modelServingWizardEdit
       .findRuntimeArgsTextBox()
       .should('have.value', '--verbose\n--log-level=debug');
--- a/packages/model-serving/cypress/tests/mocked/modelServing/modelServingNim.cy.ts
+++ b/packages/model-serving/cypress/tests/mocked/modelServing/modelServingNim.cy.ts
@@ -464,7 +464,7 @@ describe('NIM Models Deployments', () => {
     modelServingWizardEdit.findServiceAccountNameInput().should('have.value', 'my-existing-sa');
     modelServingWizardEdit
       .findRuntimeArgsTextBox()
-      .should('have.value', '--verbose\n--log-level=debug');
+      .should('contain.value', '--verbose\n--log-level=debug');
     modelServingWizardEdit.findEnvVariablesCheckbox().should('be.checked');
     modelServingWizardEdit.findEnvVariableName('0').should('have.value', 'CUSTOM_VAR');
     modelServingWizardEdit.findEnvVariableValue('0').should('have.value', 'custom-value');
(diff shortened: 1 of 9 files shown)


packages/model-serving/cypress/tests/mocked/modelServing/modelServingNim.cy.ts
```diff
@@ -462,10 +462,9 @@ describe('NIM Models Deployments', () => {
     modelServingWizardEdit.findExternalRouteCheckbox().should('be.checked');
     modelServingWizardEdit.findTokenAuthenticationCheckbox().should('be.checked');
     modelServingWizardEdit.findServiceAccountNameInput().should('have.value', 'my-existing-sa');
-    modelServingWizardEdit.findRuntimeArgsCheckbox().should('be.checked');
     modelServingWizardEdit
       .findRuntimeArgsTextBox()
-      .should('have.value', '--verbose\n--log-level=debug');
+      .should('contain.value', '--verbose\n--log-level=debug');
```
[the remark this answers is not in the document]
Philip-Carneiro: The contain.value assertions are intentional here. The RuntimeArgsField component now adds a trailing newline on first render (before user interaction) as a UX convenience — this is controlled by a useRef flag that flips on the first keystroke. Using exact have.value would couple the tests to this display-only detail:

- The exact textarea value depends on whether the component just mounted (trailing \n) or the user has already interacted (no trailing \n). Tests shouldn't need to track that internal state.
- If the trailing newline behavior is adjusted later, every assertion would break without any actual bug being introduced.
- contain.value still catches any real regression (wrong args, missing args, wrong order). The only gap vs have.value is extra content beyond the \n, which can't happen given the component's controlled value logic.

In short: the assertions validate the semantic content (correct args are displayed) without coupling to a cosmetic implementation detail.

packages/model-serving/src/components/deploymentWizard/fields/RuntimeArgsField.tsx
```diff
@@ -134,26 +132,27 @@ export const RuntimeArgsField: React.FC<RuntimeArgsFieldProps> = ({
...
+      <Stack>
+        <TextArea
+          id="runtime-args-textarea"
+          data-testid="runtime-args-textarea"
+          aria-labelledby="runtime-args-label"
+          placeholder={`--arg\n--arg2=value2\n--arg3 value3`}
+          value={displayValue}
+          onChange={handleTextAreaChange}
+          isDisabled={!allowCreate}
+          rows={4}
```
emilys314, line 144: We don't need this. Also when it's empty, the field is larger than it needs to be by default. It would only help if you have stuff, but the new line will be show with the `autoResize`

```suggestion
```

packages/model-serving/src/components/deploymentWizard/fields/RuntimeArgsField.tsx
```diff
@@ -68,28 +67,27 @@ export const RuntimeArgsField: React.FC<RuntimeArgsFieldProps> = ({
   allowCreate = true,
   predefinedArgs,
 }) => {
-  const handleCheckboxChange = (event: React.FormEvent<HTMLInputElement>, checked: boolean) => {
-    const newData = { ...data, enabled: checked };
-    onChange?.(newData);
-  };
+  const hasTyped = React.useRef(false);
+  const displayValue = hasTyped.current ? data.args.join('\n') : `${data.args.join('\n')}\n`;
```
emilys314, line 71: When the field is empty, you'll have to new lines in the text input

Image

This would probably fix it?
```suggestion
  const displayValue = hasTyped.current && data.args.length > 0 ? data.args.join('\n') : `${data.args.join('\n')}\n`;
```

emilys314 (commented):
Looks good, I tested created a new model, adding args, removing args,

emilys314 (approved):
Making a followup ticket to fix the minor double new line issue

PR #9449, opened 2026-08-21
feat(RHOAIENG-79808): lock NIM model location and image when editing deployments

YuliaKrimerman:
https://issues.redhat.com/browse/RHOAIENG-79808

## Description

Prevents risky cross-platform edits in the deployment wizard by locking NIM-specific fields when editing existing deployments.

**Non-NIM deployments (KServe InferenceService / NIMService with S3, URI, OCI, PVC, or existing connection):**
- Hide **NVIDIA NIM** from the model location dropdown on edit
- Other model location options remain available (users can still switch between non-NIM locations)

**NIM deployments (legacy NIM-on-KServe and operator NIMService):**
- Disable the entire **Model location** dropdown when location is NVIDIA NIM, with a hover tooltip explaining why
- Lock **NIM image** selection on edit when a valid catalog image is present (matches legacy `ManageNIMServingModal` behavior)
- Allow re-selecting a NIM image when repository/tag is missing or the saved image is not found in the catalog (avoids invalid form state)

**Other changes:**
- Pass `isEditing` into `GenericFieldRenderer` on the model source step so `NIMImageField` receives edit mode (previously only model type did)
- Extract pure helpers: `isEditingNimModelLocation`, `shouldHideNimModelLocationOption`, `isNIMImageSelectionLocked`

**Files touched:**
- `packages/model-serving/src/components/deploymentWizard/fields/ModelLocationSelectField.tsx`
- `packages/model-serving/src/components/deploymentWizard/steps/ModelSourceStep.tsx`
- `packages/model-serving/src/components/deploymentWizard/utils.ts`
- `packages/nim-serving/src/pages/deploymentWizard/fields/NIMImageField.tsx`
- Unit tests + Cypress mock test updates

## How Has This Been Tested?

**Automated:**
- `npm run test-unit` in `packages/model-serving` — `ModelLocationSelectField.test.tsx`, `modelLocationEditUtils.spec.ts` (37 tests)
- `npm run test-unit` in `packages/nim-serving` — `NIMImageField.spec.tsx`, `isNIMImageSelectionLocked.spec.ts` (16 tests)
- Lint and type-check on changed files

**Manual (planned / in progress on `emily2` cluster):**

https://github.com/user-attachments/assets/e6918365-6fee-4a45-b20b-8ff6640c586c
Screenshot 2026-08-21 at 2 34 15 PM

1. Edit **fraud zip** (non-NIM, URI) → confirm NVIDIA NIM is hidden from model location; other locations still switchable
2. Edit **nim wizard** (NIM-on-KServe) → confirm model location dropdown is disabled with tooltip; NIM image is locked when catalog image is found
3. Edit a NIM deployment with missing/not-found image → confirm image selector stays enabled
4. **Deploy model** (create flow) → confirm NVIDIA NIM still appears and works on new deployments

## Test Impact

**Added:**
- `packages/model-serving/src/components/deploymentWizard/__tests__/modelLocationEditUtils.spec.ts`
- `packages/nim-serving/src/pages/deploymentWizard/fields/__tests__/isNIMImageSelectionLocked.spec.ts`

**Updated:**
- `packages/model-serving/src/components/deploymentWizard/fields/__tests__/ModelLocationSelectField.test.tsx` — edit-mode location restrictions
- `packages/nim-serving/src/pages/deploymentWizard/fields/__tests__/NIMImageField.spec.tsx` — locked vs unlocked image selection on edit
- `packages/model-serving/cypress/tests/mocked/modelServing/modelServingNim.cy.ts` — asserts disabled model location and no clear button on locked NIM image

## Request review criteria:

If you have UI changes:

## Summary by CodeRabbit

* **New Features**
  * Improved deployment editing controls for model locations.
  * Prevented changing or clearing the selected NIM image when it is unavailable or locked.
  * Added explanatory tooltips for disabled model-location controls.
  * Hid incompatible model-location options while editing deployments.

* **Bug Fixes**
  * Improved disabled-state behavior for selection fields, including hiding unavailable clear actions.
  * Preserved valid image selections when catalog data is incomplete or still loading.


mturley (changes requested):
Found a bug: In the "image not found" warning case, the field is not disabled (as intended):

Image

However, in that case (image not found, field enabled) if I choose a different image, the field reverts to disabled:

Image

I would expect that when we're in the "image doesn't exist and I'm letting you choose a new one" state, the field remains enabled until you submit the wizard.

Also, not a blocker, but one thing I notice is that the disabled NIM image dropdown has its text in the regular black color and not the faded disabled style color (you can see that above). Reviewing the code I don't see why that would happen, we could ignore for now.

packages/model-serving/src/components/deploymentWizard/fields/ModelLocationSelectField.tsx
```diff
@@ -28,13 +29,17 @@ import { hasOnlyExtensionFields, ModelLocationInputFields } from './ModelLocatio
...
+import { isEditingNimModelLocation, shouldHideNimModelLocationOption } from '../utils';
 import {
   isModelLocationType,
   ModelLocationData,
   ModelLocationType,
 } from '../../../shared/types/form-data';
 import { UseModelDeploymentWizardState } from '../useDeploymentWizard';
 
+const NIM_EDIT_LOCATION_DISABLED_TOOLTIP =
+  'Model location cannot be changed when editing an NVIDIA NIM deployment.';
```
emilys314: We shouldn't have NIM specific code in /model-serving

packages/model-serving/src/components/deploymentWizard/fields/ModelLocationSelectField.tsx
```diff
@@ -157,7 +191,7 @@ export const ModelLocationSelectField: React.FC<ModelLocationSelectFieldProps> =
     if (uriConnectionTypes.length > 0 || hasURISelected) {
       options.push({ key: uriOption.key, label: uriOption.label });
     }
-    if (isNimWizardEnabled) {
+    if (isNimWizardEnabled && !hiddenLocationKeys.has(NIMModelLocationOption.key)) {
```
mturley, line 194: Interesting: pre-existing, so out of scope for this PR, but I wonder if the `isNimWizardEnabled` code should also not exist here, which would require us to add some kind of extension that can conditionally add location type options. The hiddenLocationKeys solution via modelLocationOverrides seems like it could maybe be unnecessary if the extension that provides the option can have internal logic for hiding it, although I’m not certain how clean that can be. If we extract that to a tech debt ticket we should have that ticket’s scope include reevaluating whether the new ModelLocationFieldOverride is necessary after that refactor.

I’m curious about @emilys314’s thoughts on this one.

packages/model-serving/src/components/deploymentWizard/fields/__tests__/ModelLocationSelectField.test.tsx
```diff
@@ -12,10 +12,26 @@ import type {
...
+
+const nimModelLocationOverride: ModelLocationFieldOverride = {
+  id: 'modelLocation',
+  type: 'modifier',
+  isActive: () => true,
+  locationKey: NIMModelLocationOption.key,
+  disableWhenEditing: true,
+  disabledTooltip: 'Model location cannot be changed when editing an NVIDIA NIM deployment.',
+  hideOptionWhenEditingOtherLocation: true,
+};
```
emilys314: Duplicating and recreating the extension object inside the test means your not testing the actual one. If the real one changes, this will be out of sync

packages/model-serving/src/components/deploymentWizard/fields/ModelLocationSelectField.tsx
emilys314: +1 I'm not a fan of the Override, those are more of a last resort and just force an extension for a specific use case. Although it may be a necessary evil here? (I haven't dug too deep though)

There are some tech debt tickets to look into refactoring the model location as whole. to enable stuff like this. This field desperately needs updating

packages/nim-serving/src/pages/deploymentWizard/fields/NIMImageField.tsx
```diff
@@ -168,19 +181,34 @@ const NIMImageFieldComponent: React.FC<NIMImageFieldComponentProps> = ({
     [images, value],
   );
 
+  const [isReselectionUnlocked, setIsReselectionUnlocked] = React.useState(false);
+  React.useEffect(() => {
+    if (isEditing && existingOptionNotFound && externalData?.loaded && images.length > 0) {
+      setIsReselectionUnlocked(true);
+    }
+  }, [isEditing, existingOptionNotFound, externalData?.loaded, images.length]);
```
emilys314: Is there a way to not have a useEffect?

emilys314 (commented):
I haven't tested it yet

packages/model-serving/src/components/deploymentWizard/fields/ModelLocationSelectField.tsx
YuliaKrimerman: Agree this is interim. ModelLocationFieldOverride mirrors NIMModelTypeOverride until the broader model-location refactor (separate tech debt).

packages/model-serving/cypress/tests/mocked/modelServing/modelServingNim.cy.ts
```diff
@@ -428,11 +428,14 @@ describe('NIM Models Deployments', () => {
...
       .findModelLocationSelect()
-      .should('contain.text', ModelLocationSelectOption.NIM);
+      .should('contain.text', ModelLocationSelectOption.NIM)
+      .find('button')
+      .should('be.disabled');
     modelServingWizardEdit.nim
       .findImageSelect()
       .find('input')
       .should('have.value', 'Snowflake Arctic Embed Large Embedding - 1.0.1');
+    modelServingWizardEdit.nim.findImageSelect().find('[aria-label="Clear input value"]').should('not.exist');
```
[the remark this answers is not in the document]
mturley: This seems acceptable. PatternFly controls the disabled state and only shows the clear control when the field is enabled.

packages/model-serving/src/components/deploymentWizard/fields/__tests__/ModelLocationSelectField.test.tsx
```diff
@@ -256,6 +264,15 @@ const mockAreaStatus = (status: boolean): IsAreaAvailableStatus => ({
...
 const mockPvcs: PersistentVolumeClaimKind[] = [];
+const NIMModelLocationOverride: ModelLocationFieldOverride = {
+  id: 'modelLocation',
+  type: 'modifier' as const,
+  isActive: () => true,
+  locationKey: NIMModelLocationKey,
+  disableWhenEditing: true,
+  disabledTooltip: 'Model location cannot be changed when editing an NVIDIA NIM deployment.',
+  hideOptionWhenEditingOtherLocation: true,
+};
```
mturley, line 275: @YuliaKrimerman this test still recreates the extension object instead of using the real one, I think we could just not mock `useWizardFieldOverrides`. If we have to mock that we could import the real extension object to use here as the mock return value maybe?

mturley (changes requested):
Looks like a coderabbit comment about keyboard accessibility might be accurate, and a +1 to one of Emily’s earlier comments here. Testing it

packages/model-serving/src/components/deploymentWizard/fields/__tests__/ModelLocationSelectField.test.tsx
mturley: As discussed, this is challenging from a circular-dependency perspective and the test contract is really the relevant properties of the extension. We could follow up to move the test to the nim-serving package or add a cypress test if we need to make sure we’re testing the real extension.

mturley (approved):
Tested again, working well, changes look good

/approve
/lgtm

@YuliaKrimerman to open a small followup issue about extension test coverage

packages/model-serving/src/components/deploymentWizard/fields/__tests__/ModelLocationSelectField.test.tsx
YuliaKrimerman: Opened a small follow up ticket for that https://redhat.atlassian.net/browse/RHOAIENG-89397

PR #9477, opened 2026-08-25
Move deployment extension to model-serving

ppadti:
Closes: https://redhat.atlassian.net/browse/RHOAIENG-83823

## Description
This PR aims to move deployment extension used in model catalog is moving to model-serving.

## How Has This Been Tested?
Run the application and try to deploy a model from model catalog and make sure it works the way it is.

## Test Impact
NA 

## Request review criteria:

If you have UI changes: 

## Summary by CodeRabbit

* **New Features**
  * Added a model catalog deployment action that integrates with the deployment wizard.
  * Deployment controls now reflect available serving platforms and KServe requirements.
  * Added informative tooltips when deployment is unavailable, loading, or requires administrator configuration.

* **Bug Fixes**
  * Prevented deployment actions from being activated when required serving capabilities are unavailable.

* **Tests**
  * Added coverage for enabled and disabled deployment states across supported platform configurations.


ppadti:
Need to test after resolving the merge conflict, both of the green cluster are not working. Once I have the cluster will test and update this branch

packages/model-registry/upstream/frontend/src/odh/hooks/useCatalogDeployPrefillData.ts
```diff
@@ -0,0 +1,55 @@
...
+    return {
+      modelName: model.name,
+      modelUri: uri,
+      returnRouteValue: '/ai-hub/models/deployments/',
+      cancelReturnRouteValue: cancelReturnRoute,
+      wizardStartIndex: 1,
+      prefillAlertText: `The ${model.name} model details have been imported from the model catalog.`,
+      ...getValidatedConfigurationsForModel(model, isToolCallingEnabled),
+    };
+  }, [model, uri, cancelReturnRoute, isToolCallingEnabled]);
```
Philip-Carneiro: The `deployPrefill` object does not include the `catalogModelId` field. The old deleted component (`CatalogDeployAction.tsx`) computed:
```typescript
catalogModelId: [decodedParams.sourceId || model.source_id, model.name]
  .filter(Boolean)
  .join('/'),
```
**Impact**:
1. **Validated configurations section breaks** — `deployWizardTracking.ts:110` requires `Boolean(navState.catalogModelId)` to show the validated arguments section. Without it, users deploying from catalog lose access to validated configurations.
2. **Analytics tracking corrupted** — `deployWizardTracking.ts:91` only returns entry point `'model_details'` when `navState.fromCatalog && navState.catalogModelId`. Without it, falls back to wrong entry point in analytics.

The hook **has access** to both `sourceId` (param line 19) and `model.source_id` (param line 15) but doesn't use them.

The type `DeployPrefillData.catalogModelId` is optional, so TypeScript doesn't catch this regression. The downstream consumers (`useNavigateToDeploymentWizardWithData.ts:86`, `deployWizardTracking.ts:91,110`) read this field and fail silently when it's undefined.

packages/model-serving/modelRegistry/getDeployButtonState.ts
```diff
@@ -0,0 +1,29 @@
+const DEPLOY_BUTTON_TOOLTIP = {
+  ENABLE_SINGLE_MODEL_SERVING:
+    'To deploy this model, an administrator must first enable single-model serving in the cluster settings.',
+  ENABLE_MODEL_SERVING_PLATFORM:
+    'To enable model serving, an administrator must first select a model serving platform in the cluster settings.',
+};
+
```
Philip-Carneiro, line 2: I see duplication of this in 3 different places.
`frontend/src/pages/modelServing/screens/const.ts`
`packages/model-registry/upstream/frontend/src/odh/const.ts`

Should be a good idea to extract it form all these places and re-use?

packages/model-serving/modelRegistry/getDeployButtonState.ts
```diff
@@ -0,0 +1,29 @@
...
+    return {
+      enabled: false,
+      tooltip: DEPLOY_BUTTON_TOOLTIP.ENABLE_SINGLE_MODEL_SERVING,
+    };
+  }
+
+  return {
+    enabled: true,
+  };
+};
```
Philip-Carneiro, line 24: This entire block is duplicated as well 
`packages/model-registry/upstream/frontend/src/odh/utils.ts`

Philip-Carneiro (changes requested):
Have some points I added here, but should we add unit test to cover these 3 new files? Because atm there is no testing to it.

**Files**:
- `packages/model-serving/modelRegistry/getDeployButtonState.ts`
- `packages/model-registry/upstream/frontend/src/odh/hooks/useCatalogDeployPrefillData.ts`
- `packages/model-serving/modelCatalog/CatalogDeployAction.tsx`

ppadti:
> Have some points I added here, but should we add unit test to cover these 3 new files? Because atm there is no testing to it.
> 
> **Files**:
> 
> * `packages/model-serving/modelRegistry/getDeployButtonState.ts`
> * `packages/model-registry/upstream/frontend/src/odh/hooks/useCatalogDeployPrefillData.ts`
> * `packages/model-serving/modelCatalog/CatalogDeployAction.tsx`

will add tests for `packages/model-serving/modelRegistry/getDeployButtonState.ts` and  `packages/model-registry/upstream/frontend/src/odh/hooks/useCatalogDeployPrefillData.ts` - we should not add unit test for the component file- those will be covered in cypress

packages/model-serving/modelRegistry/getDeployButtonState.ts
ppadti: yeah, that is expected. The getDeployButtonState in utils folder will be moved/reused in [another issue](https://redhat.atlassian.net/browse/RHOAIENG-84796).

packages/model-serving/modelRegistry/getDeployButtonState.ts
ppadti: updated model-serving packages, for the model-registry package we have anther issue which would also move this to model-serving package

ppadti:
I have to test this PR on jenkins - so do not merge this until then. I will run a test as soon as I get lgtm

Philip-Carneiro:
/lgtm
/approve

The process of deployment of the model is ok.

ppadti:
Tested this on jenkins, id: 2611 - the one failing test is unrelated, caused due conflict in sync - created [a task](https://redhat.atlassian.net/browse/RHOAIENG-90203) to resolve that.

packages/model-serving/modelCatalog/CatalogDeployAction.tsx
```diff
@@ -0,0 +1,60 @@
+import React from 'react';
+import { Button, ButtonVariant, Tooltip } from '@patternfly/react-core';
+import type { CatalogDeployActionComponentProps } from '@odh-dashboard/model-registry/shared';
```
mturley: This component now imports `@odh-dashboard/model-registry/shared`, but `@odh-dashboard/model-serving/package.json` does not declare `@odh-dashboard/model-registry`. Since model-registry already depends on model-serving, this creates a circular package dependency (even though the import is type-only). Could `CatalogDeployActionComponentProps` also be moved to the model-serving package? Its name implies that it belongs with the component itself. It could also have a more generic name though.

mturley (commented):
One small concern. Not a blocker because it doesn’t seem to break the build.

mturley (approved):
Tested against psi-16. Looks good. Failing e2e test seems unrelated.

/approve
/lgtm

PR #9479, opened 2026-08-25
fix(notebooks): enable TLS for standalone notebooks-ui module federation

harshad16:
Add OpenShift serving-cert wiring in standalone manifests and set module-federation tls: true. BFF TLS support lives in opendatahub-io/workbenches (fix/backend-tls-federation) and will be synced via the notebooks subtree.

## Description

JIRA: https://redhat.atlassian.net/browse/RHOAIENG-88028

On fresh RHOAI installs, the notebooks plugin fails to load via module federation because the operator sets tls: true for the notebooks entry in federation-config, but notebooks-ui was deployed to serve HTTP only. rhods-dashboard logs show:

ERR_SSL_PACKET_LENGTH_TOO_LONG
Module federation service 'notebooks' is unavailable
This PR adds the dashboard-side TLS wiring for the standalone notebooks module, aligned with gen-ai/maas/mlflow manifests. No changes under packages/notebooks/upstream/ — BFF TLS is implemented in [opendatahub-io/workbenches](https://github.com/opendatahub-io/workbenches) and will arrive via subtree sync.

Depends-on: https://github.com/opendatahub-io/workbenches/pull/25

## How Has This Been Tested?

Prerequisites: workbenches TLS PR merged + notebooks-ui image built with TLS BFF support.

- Deploy/reconcile dashboard with this PR’s manifests.
- Confirm OpenShift creates notebooks-proxy-tls secret for odh-dashboard-notebooks-ui Service.
- Confirm notebooks-ui pod is Ready (HTTPS probes on /api/v1/healthcheck).
- Check rhods-dashboard logs — no ERR_SSL_PACKET_LENGTH_TOO_LONG for /_mf/notebooks/remoteEntry.js.
- Hard-refresh browser — Workspaces / Workspace Kinds visible under AI Hub when workbenchesV2: true.
- Confirm no manual federation-config patch is required (notebooks.tls stays true).

## Test Impact

Manifest-only change plus package.json federation metadata. No new unit tests. Validation is cluster/integration testing with the updated notebooks-ui image.

## Request review criteria:

If you have UI changes: 

## Summary by CodeRabbit

* **New Features**
  * Enabled HTTPS for the notebooks interface.
  * Added automatic TLS certificate provisioning and secure service configuration.
  * Updated health checks to verify HTTPS availability.
  * Enabled TLS in the notebooks module-federation setup.

### DIFF
--- a/manifests/modules/notebooks/deployment.yaml
+++ b/manifests/modules/notebooks/deployment.yaml
@@ -59,6 +59,23 @@ spec:
                   items:
                     - key: service-ca.crt
                       path: service-ca.crt
+        - name: proxy-tls
+          secret:
+            secretName: notebooks-proxy-tls
+        - name: odh-trusted-ca-cert
+          configMap:
+            name: odh-trusted-ca-bundle
+            optional: true
+            items:
+              - key: ca-bundle.crt
+                path: odh-trusted-ca-bundle.crt
+        - name: odh-ca-cert
+          configMap:
+            name: odh-trusted-ca-bundle
+            optional: true
+            items:
+              - key: odh-ca-bundle.crt
+                path: odh-ca-bundle.crt
       containers:
         - name: notebooks-ui
           image: notebooks-ui-image
@@ -67,7 +84,7 @@ spec:
             httpGet:
               path: /api/v1/healthcheck
               port: 9043
-              scheme: HTTP
+              scheme: HTTPS
             initialDelaySeconds: 30
             timeoutSeconds: 15
             periodSeconds: 30
@@ -77,7 +94,7 @@ spec:
             httpGet:
               path: /api/v1/healthcheck
               port: 9043
-              scheme: HTTP
+              scheme: HTTPS
             initialDelaySeconds: 15
             timeoutSeconds: 15
             periodSeconds: 30
@@ -98,12 +115,33 @@ spec:
           volumeMounts:
             - name: tmp
               mountPath: /tmp
+            - name: proxy-tls
+              mountPath: /etc/tls/private
+              readOnly: true
+            - name: odh-trusted-ca-cert
+              mountPath: /etc/pki/tls/certs/odh-trusted-ca-bundle.crt
+              subPath: odh-trusted-ca-bundle.crt
+              readOnly: true
+            - name: odh-trusted-ca-cert
+              mountPath: /etc/ssl/certs/odh-trusted-ca-bundle.crt
+              subPath: odh-trusted-ca-bundle.crt
+              readOnly: true
+            - name: odh-ca-cert
+              mountPath: /etc/pki/tls/certs/odh-ca-bundle.crt
+              subPath: odh-ca-bundle.crt
+              readOnly: true
+            - name: odh-ca-cert
+              mountPath: /etc/ssl/certs/odh-ca-bundle.crt
+              subPath: odh-ca-bundle.crt
+              readOnly: true
             - name: modules-sa-token
               mountPath: /var/run/secrets/kubernetes.io/serviceaccount
               readOnly: true
-          env:
-            - name: PORT
-              value: "9043"
+          args:
+            - '--port=9043'
+            - '--cert-file=/etc/tls/private/tls.crt'
+            - '--key-file=/etc/tls/private/tls.key'
+            - '--static-assets-dir=/static'
           securityContext:
             allowPrivilegeEscalation: false
             runAsNonRoot: true
--- a/manifests/modules/notebooks/service.yaml
+++ b/manifests/modules/notebooks/service.yaml
@@ -2,6 +2,9 @@ apiVersion: v1
 kind: Service
 metadata:
   name: odh-dashboard-notebooks-ui
+  annotations:
+    service.beta.openshift.io/serving-cert-secret-name: notebooks-proxy-tls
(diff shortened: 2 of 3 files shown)


manifests/modules/notebooks/deployment.yaml
```diff
@@ -98,12 +115,33 @@ spec:
...
+              subPath: odh-trusted-ca-bundle.crt
+              readOnly: true
+            - name: odh-ca-cert
+              mountPath: /etc/pki/tls/certs/odh-ca-bundle.crt
+              subPath: odh-ca-bundle.crt
+              readOnly: true
+            - name: odh-ca-cert
+              mountPath: /etc/ssl/certs/odh-ca-bundle.crt
+              subPath: odh-ca-bundle.crt
+              readOnly: true
```
andyatmiami, line 136: Why do we mount the same thing in 2 different places?

manifests/modules/notebooks/deployment.yaml
```diff
@@ -98,12 +115,33 @@ spec:
...
             - name: modules-sa-token
               mountPath: /var/run/secrets/kubernetes.io/serviceaccount
               readOnly: true
-          env:
-            - name: PORT
-              value: "9043"
+          args:
+            - '--port=9043'
+            - '--cert-file=/etc/tls/private/tls.crt'
+            - '--key-file=/etc/tls/private/tls.key'
```
andyatmiami, line 143: Given we mark the configMap-backed volumes as `optional` ... that means we don't get application startup on their presence, right?

Which means the golang server could potentially start **BEFORE** these files exist? Meaning we'd likely need to hav some type of wait loop before configuring the server?  otherwise `optional: true` doesn't really buy us much other than a different race condition during application server startup?

manifests/modules/notebooks/service.yaml
```diff
@@ -2,6 +2,9 @@ apiVersion: v1
 kind: Service
 metadata:
   name: odh-dashboard-notebooks-ui
+  annotations:
+    service.beta.openshift.io/serving-cert-secret-name: notebooks-proxy-tls
+    service.beta.kubernetes.io/backend-protocol: HTTPS
```
andyatmiami, line 7: Just out of curiosity (i have no idea!) - do we really need to set `backend-protocol` if we are defining `serving-cert-secret-name` ? seems kinda redundant 🤔

manifests/modules/notebooks/deployment.yaml
```diff
@@ -59,6 +59,23 @@ spec:
                   items:
                     - key: service-ca.crt
                       path: service-ca.crt
+        - name: proxy-tls
+          secret:
+            secretName: notebooks-proxy-tls
```
andyatmiami, line 64: where does this come from? did it already exist? just confused (personally) since it seemed like the original state of code before this PR was "no TLS" - so not sure how/why we had a tls-secret beforehand then...

andyatmiami (commented):
Nothing concerning that could/would prevent moving forward with this PR - but I have some questions as I just don't understand the motivation behind some of the changes (even though I don't really have an issue with the changes themselves).

manifests/modules/notebooks/deployment.yaml
harshad16: Great question, 
i believe this is becauses different package find crt on different path.
Took inspiration from: https://github.com/opendatahub-io/odh-dashboard/blob/main/manifests/modules/maas/deployment.yaml#L118-L135

manifests/modules/notebooks/deployment.yaml
harshad16: `Optional ` marker seems to be standard practice in modules 
we would like the server not to break on TLS, and atleast start serving with error, instead of stopping it
took inspiration from here: https://github.com/opendatahub-io/odh-dashboard/blob/main/manifests/modules/maas/deployment.yaml#L65-L78

manifests/modules/notebooks/deployment.yaml
harshad16: The annotation `service.beta.openshift.io/serving-cert-secret-name: notebooks-proxy-tls`, based on that built-in openshift Service CA operator get the secret created with tls certs.

manifests/modules/notebooks/service.yaml
harshad16: Took inspiration from here: https://github.com/opendatahub-io/odh-dashboard/blob/main/manifests/modules/maas/service.yaml#L6-L7

seems like both functions differently 
1. for creating cert secret
2. sharing with load-balancer the mode to function in.

manifests/modules/notebooks/deployment.yaml
andyatmiami: whoops, i'm sorry.. i got my volume mounts confused...

i was thinking 

```
--cert-file=/etc/tls/private/tls.crt
--key-file=/etc/tls/private/tls.key
```

came from one of the `optional`  voiumes - but that is **NOT THE CASE**

so, yeah, we might get handshake errors if CA stores not set up (yet)... but that seems a reasonable "eventual consistency" pattern..

the files we pass in **as config** to the server **will exist**

andyatmiami (approved):
/lgtm
/approve

---

changes seem rationale/justified/correct... and follow patterns from components stable/mature in production

very happy with responses provided by @harshad16 to my original areas of uncertainty.

PR #9480, opened 2026-08-25
fix: Remove inaccessible elasticstack.slack.com URL

crackcodecamp:
Removes the private Slack workspace link that was causing 403 Forbidden errors in test validation. Users can still access community support via the public discuss.elastic.co forum.

### DIFF
--- a/manifests/rhoai/shared/apps/elastic/elastic-app.yaml
+++ b/manifests/rhoai/shared/apps/elastic/elastic-app.yaml
@@ -57,7 +57,7 @@ spec:
 
     - Elasticsearch can be deployed where your apps are: on-prem, your cloud provider of choice, or air-gapped environments.
 
-    - Enjoy being part of a worldwide community of developers where inspiration and support are never far away. Find the Elastic community on [Slack](https://elasticstack.slack.com), our [discuss](https://discuss.elastic.co/) forums, or social media.
+    - Enjoy being part of a worldwide community of developers where inspiration and support are never far away. Find the Elastic community on our [discuss](https://discuss.elastic.co/) forums or social media.
 
     ## Quick links
 


manifests/rhoai/shared/apps/elastic/elastic-app.yaml
```diff
@@ -57,7 +57,7 @@ spec:
 
     - Elasticsearch can be deployed where your apps are: on-prem, your cloud provider of choice, or air-gapped environments.
 
-    - Enjoy being part of a worldwide community of developers where inspiration and support are never far away. Find the Elastic community on [Slack](https://elasticstack.slack.com), our [discuss](https://discuss.elastic.co/) forums, or social media.
+    - Enjoy being part of a worldwide community of developers where inspiration and support are never far away. Find the Elastic community on our [discuss](https://discuss.elastic.co/) forums or social media.
```
andrewballantyne, line 60: Interesting -- not sure I understand this change and how it's not a problem for us in 3.6 https://github.com/opendatahub-io/odh-dashboard/blob/main/manifests/rhoai/apps/elastic/elastic-app.yaml#L60

@antowaddle can you help out here?

antowaddle: The Original PR has all the context https://github.com/opendatahub-io/odh-dashboard/pull/9355

`Investigation showed:

The link works fine in a real browser.
Repeated curl requests to the same URL (varying headers, User-Agent, etc.) show a probabilistic ~403 rate, consistent with Slack's edge/WAF bot-detection flagging automated (non-browser) traffic rather than an actual broken/removed link.
All other URLs in the manifest set (~82 links) validate consistently across repeated runs; only this one is flaky.`

antowaddle (changes requested):
Let's instead backport this fix https://github.com/opendatahub-io/odh-dashboard/pull/9355

lucferbux:
I was about to mention the same @antowaddle so @crackcodecamp if we can cherry-pick the fix on main would be amazing, thanks a lot!

PR #9484, opened 2026-08-25
Allow deployments to stop with a deleted hardware profile, and inject updated values when restarts

nananosirova:
https://redhat.atlassian.net/browse/RHOAIENG-78897
https://redhat.atlassian.net/browse/RHOAIENG-78895

## Description

When a hardware profile bound to a deployment is deleted, the `InferenceService` keeps dangling `opendatahub.io/hardware-profile-name` and `-namespace` annotations. The hardware profile webhook rejects any patch that leaves those annotations in place, including the unrelated stop/start patch, so affected deployments could not be stopped or restarted.

Fixed by patching the hardware profile annotation removal with `patchInferenceServiceStoppedStatus` and `patchDeploymentStoppedStatus`, mirroring the existing workbench fix.

Separately, resources, tolerations and nodeSelector are written into the inference CR when a deployment is created, and the webhook only fills in paths that are absent, it never overwrites values already present. A deployment therefore kept the settings captured at deploy time no matter how many times it was restarted, even though the hardware profile Updated state tells the user that restarting will apply the profile's new settings.

Fixed by stripping those paths on start so the webhook repopulates them from the profile's current values, and bringing the stored resource version up to date so the deployment stops reporting the Updated state. 

## How Has This Been Tested?

This should be tested in 3.4.4.
**Deleted hardware profile**
I deployed a model from model catalog, selecting a hardware profile. While it's running, I delete the hardware profile, and then stop the deployment. It should stop successfully. 
**Updated hardware profile**
Deploy a model with a hardware profile whose default and max counts differ. Stop the deployment, edit the profile's CPU/memory, then start it again. It should come back with the profile's new values rather than the ones captured at deploy time.

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/concepts/hardwareProfiles/__tests__/utils.spec.ts
+++ b/frontend/src/concepts/hardwareProfiles/__tests__/utils.spec.ts
@@ -7,6 +7,7 @@ import {
   getExistingHardwareProfileData,
   assemblePodSpecOptions,
   applyHardwareProfileConfig,
+  getUpdatedHardwareProfilePatches,
 } from '#~/concepts/hardwareProfiles/utils';
 import { HardwareProfileKind } from '#~/k8sTypes';
 import {
@@ -18,6 +19,7 @@ import {
 } from '#~/types';
 import { UseHardwareProfileConfigResult } from '#~/concepts/hardwareProfiles/useHardwareProfileConfig';
 import { NOTEBOOK_HARDWARE_PROFILE_PATHS } from '#~/concepts/notebooks/const.ts';
+import { HardwareProfileBindingState } from '#~/concepts/hardwareProfiles/const';
 
 global.structuredClone = (val: unknown) => JSON.parse(JSON.stringify(val));
 
@@ -676,5 +678,28 @@ describe('applyHardwareProfileConfig', () => {
     expect((result as any).metadata?.annotations?.['opendatahub.io/hardware-profile-name']).toBe(
       'custom-path-test',
     );
+  });
+});
+
+describe('getUpdatedHardwareProfilePatches', () => {
+  it('should only remove paths present on the resource', () => {
+    const patches = getUpdatedHardwareProfilePatches(
+      {
+        state: HardwareProfileBindingState.UPDATED,
+        profile: mockHardwareProfile({ resourceVersion: '999' }),
+      },
+      mockNotebookK8sResource({}),
+      NOTEBOOK_HARDWARE_PROFILE_PATHS,
+    );
+
+    expect(patches).toStrictEqual([
+      { op: 'remove', path: '/spec/template/spec/containers/0/resources' },
+      { op: 'remove', path: '/spec/template/spec/tolerations' },
+      {
+        op: 'replace',
+        path: '/metadata/annotations/opendatahub.io~1hardware-profile-resource-version',
+        value: '999',
+      },
+    ]);
   });
 });
--- a/frontend/src/concepts/hardwareProfiles/__tests__/utils.spec.ts
+++ b/frontend/src/concepts/hardwareProfiles/__tests__/utils.spec.ts
@@ -1,5 +1,6 @@
 import { mockHardwareProfile } from '#~/__mocks__/mockHardwareProfile';
 import { mockNotebookK8sResource } from '#~/__mocks__';
+import { mockInferenceServiceK8sResource } from '#~/__mocks__/mockInferenceServiceK8sResource';
 import {
   getProfileScore,
   sortIdentifiers,
@@ -7,6 +8,7 @@ import {
   getExistingHardwareProfileData,
   assemblePodSpecOptions,
   applyHardwareProfileConfig,
+  getDeletedHardwareProfilePatches,
   getUpdatedHardwareProfilePatches,
 } from '#~/concepts/hardwareProfiles/utils';
 import { HardwareProfileKind } from '#~/k8sTypes';
@@ -19,7 +21,11 @@ import {
 } from '#~/types';
 import { UseHardwareProfileConfigResult } from '#~/concepts/hardwareProfiles/useHardwareProfileConfig';
 import { NOTEBOOK_HARDWARE_PROFILE_PATHS } from '#~/concepts/notebooks/const.ts';
-import { HardwareProfileBindingState } from '#~/concepts/hardwareProfiles/const';
+import {
+  HardwareProfileBindingState,
+  INFERENCE_SERVICE_HARDWARE_PROFILE_PATHS,
+  REMOVE_HARDWARE_PROFILE_ANNOTATIONS_PATCH,
+} from '#~/concepts/hardwareProfiles/const';
 
 global.structuredClone = (val: unknown) => JSON.parse(JSON.stringify(val));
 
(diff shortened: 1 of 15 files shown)


frontend/src/pages/modelServing/screens/global/InferenceServiceTableRow.tsx
```diff
@@ -67,25 +76,65 @@ const InferenceServiceTableRow: React.FC<InferenceServiceTableRowProps> = ({
...
   const onStop = React.useCallback(() => {
     if (dontShowModalValue) {
       setIsStarting(false);
       setIsStopping(true);
-      patchInferenceServiceStoppedStatus(inferenceService, 'true').then(refresh);
+      patchInferenceServiceStoppedStatus(
+        inferenceService,
+        'true',
+        deletedHardwareProfilePatches,
+      ).then(refresh);
```
DaoDaoNoCode: **Suggestion (pre-existing, not introduced by this PR)**

Neither stop path has a `.catch()` handler. If the patch fails (network error, webhook rejection, conflict), `isStopping` stays `true` with no recovery for the user.

Compare with `onStart` which resets state on failure:
```ts
.catch(() => setIsStarting(false));
```

The same gap exists at line 264 (confirm-modal path) and in `DeploymentsTableRow.tsx` (lines 117-119 and 248-249).

Since this PR specifically fixes the webhook-rejection scenario, a `.catch(() => setIsStopping(false))` would guard against remaining edge cases where `deletedHardwareProfilePatches` might still be insufficient (e.g., binding state not yet loaded when user clicks stop).

frontend/src/concepts/hardwareProfiles/__tests__/utils.spec.ts
```diff
@@ -678,3 +680,26 @@ describe('applyHardwareProfileConfig', () => {
...
+      { op: 'remove', path: '/spec/template/spec/containers/0/resources' },
+      { op: 'remove', path: '/spec/template/spec/tolerations' },
+      {
+        op: 'replace',
+        path: '/metadata/annotations/opendatahub.io~1hardware-profile-resource-version',
+        value: '999',
+      },
+    ]);
+  });
+});
```
DaoDaoNoCode, line 787: **Suggestion — missing edge-case and companion-function coverage**

This test covers `getUpdatedHardwareProfilePatches` with the happy path (`UPDATED` state, notebook resource). Consider adding:

1. **`getDeletedHardwareProfilePatches`** — no unit test exists anywhere. This function is a new production dependency in both `InferenceServiceTableRow` and `DeploymentsTableRow`. At minimum: returns `REMOVE_HARDWARE_PROFILE_ANNOTATIONS_PATCH` when `state === DELETED` and annotations exist; returns `[]` when state is not `DELETED` or annotations are absent.

2. **Null/undefined guards** for `getUpdatedHardwareProfilePatches`:
   - `bindingState` is `null` → should return `[]`
   - `paths` is `undefined` → should return `[]` (this happens in `DeploymentsTableRow` before extensions resolve)
   - `resourceVersion` is `undefined` → should return `[]`

3. **InferenceService paths** — production code uses `INFERENCE_SERVICE_HARDWARE_PROFILE_PATHS` (e.g., `spec.predictor.model.resources`) but the test only exercises `NOTEBOOK_HARDWARE_PROFILE_PATHS`. Testing with an InferenceService mock and its paths would verify the path-to-JSON-Pointer conversion works for that shape too.

packages/model-serving/src/components/deployments/__tests__/DeploymentsTableRow.spec.tsx
```diff
@@ -61,6 +61,13 @@ jest.mock('../row/DeploymentHardwareProfileCell', () => ({
   DeploymentHardwareProfileCell: () => <td>Hardware Profile</td>,
 }));
 
+jest.mock(
+  '@odh-dashboard/internal/concepts/hardwareProfiles/useHardwareProfileBindingState',
+  () => ({
+    useHardwareProfileBindingState: () => [null, true, undefined],
+  }),
```
DaoDaoNoCode, line 83: **Suggestion — mock prevents crashes but does not verify behavior**

This returns `[null, true, undefined]`, so both `getDeletedHardwareProfilePatches` and `getUpdatedHardwareProfilePatches` return `[]`. The test confirms the component renders without errors but never exercises the scenario where `bindingStateInfo.state === DELETED` and patches flow through to `patchDeploymentStoppedStatus`.

The Cypress test in `modelServingGlobal.cy.ts` covers the legacy `InferenceServiceTableRow` path, but no test (unit or Cypress) verifies the extension-point `DeploymentsTableRow` path with a deleted or updated hardware profile.

packages/cypress/cypress/tests/mocked/modelServing/modelServingGlobal.cy.ts
```diff
@@ -823,4 +825,41 @@ describe('Model Serving Global', () => {
...
+            value: 'true',
+          },
+        ]);
+        expect(interception.request.body).to.containSubset(
+          REMOVE_HARDWARE_PROFILE_ANNOTATIONS_PATCH,
+        );
+      });
+    });
+  });
 });
```
DaoDaoNoCode, line 865: **Suggestion — consider adding a START-with-updated-profile test**

This test validates the stop-with-deleted-profile flow well. The PR also adds `updatedHardwareProfilePatches` to the start flow (stripping stale `resources`/`tolerations`/`nodeSelector` and bumping the resource-version annotation when the profile has been updated since last deploy). There is no integration test for this path.

A complementary test that sets up a stopped InferenceService whose hardware profile has a different `resourceVersion`, clicks Start, and asserts the PATCH body includes `remove` ops for the resource paths and a `replace` for the resource-version annotation would cover the other half of this fix.

DaoDaoNoCode (commented):
Review from 5-agent parallel analysis (Core Patch Logic, Legacy TableRow, Extension-Point TableRow, Package-Level Status, Test Coverage). Each inline comment is a validated finding with classification. Overall: no blockers found — the core logic is correct, type changes are backward-compatible, and guards prevent unsafe JSON Patch operations.

frontend/src/pages/modelServing/screens/global/InferenceServiceTableRow.tsx
```diff
@@ -67,25 +76,65 @@ const InferenceServiceTableRow: React.FC<InferenceServiceTableRowProps> = ({
...
     } else {
```
DaoDaoNoCode, line 125: **Suggestion (pre-existing, not introduced by this PR)**

Neither stop path has a `.catch()` handler. If the patch fails (network error, webhook rejection, conflict), `isStopping` stays `true` indefinitely with no recovery for the user.

Compare with `onStart` at line 107 which resets state on failure:
```ts
.catch(() => setIsStarting(false));
```

The same gap exists at line 264 (confirm-modal path) and in `DeploymentsTableRow.tsx` (lines 117-119 and 248-249).

Since this PR specifically fixes the webhook-rejection scenario, a `.catch(() => setIsStopping(false))` would guard against remaining edge cases where `deletedHardwareProfilePatches` might still be insufficient (e.g., binding state not yet loaded when user clicks stop).

frontend/src/concepts/hardwareProfiles/__tests__/utils.spec.ts
```diff
@@ -678,3 +680,26 @@ describe('applyHardwareProfileConfig', () => {
     );
   });
 });
+
+describe('getUpdatedHardwareProfilePatches', () => {
+  it('should only remove paths present on the resource', () => {
```
DaoDaoNoCode: **Suggestion — test coverage gaps**

This test only covers the happy path (`UPDATED` state with a notebook). Consider adding:

1. **Guard-clause tests for `getUpdatedHardwareProfilePatches`**:
   - `bindingState` is `null` → returns `[]`
   - `bindingState.state` is not `UPDATED` → returns `[]`
   - `paths` is `undefined` → returns `[]`
   - `resourceVersion` is `undefined` → returns `[]`
   - Fields at dot-notation paths don't exist on the CR → those paths are skipped

2. **Companion test for `getDeletedHardwareProfilePatches`** — also new in this PR, currently has zero test coverage. It's a simple delegation to `REMOVE_HARDWARE_PROFILE_ANNOTATIONS_PATCH`, but a test would lock down that behavior.

3. **InferenceService-path test** — the PR uses this with `INFERENCE_SERVICE_HARDWARE_PROFILE_PATHS`, which have different dot-notation paths than notebook paths. A test with those paths would validate the actual production usage.

packages/model-serving/src/components/deployments/__tests__/DeploymentsTableRow.spec.tsx
DaoDaoNoCode, line 83: **Suggestion — mock prevents crash but doesn't exercise the new behavior**

Returning `[null, true, undefined]` means the test never enters the DELETED or UPDATED code paths. Consider adding a test case that mocks `useHardwareProfileBindingState` to return `HardwareProfileBindingState.DELETED` (or `UPDATED`) and verifies the correct patches are passed to `patchDeploymentStoppedStatus`. Without that, there's no spec-level coverage proving the new feature works end-to-end in this component.

packages/cypress/cypress/tests/mocked/modelServing/modelServingGlobal.cy.ts
```diff
@@ -823,4 +825,41 @@ describe('Model Serving Global', () => {
       errorPopoverTitle.should('be.visible');
     });
   });
+  describe('Model Serving Deleted Hardware Profile', () => {
+    it('should strip the deleted hardware profile annotations when stopping a deployment', () => {
```
DaoDaoNoCode, line 829: **Suggestion — missing START-with-updated-profile test**

This new describe block tests STOP with a deleted hardware profile — great. Consider adding a companion test for **START with an `UPDATED` hardware profile**, which exercises `getUpdatedHardwareProfilePatches` and the resource/toleration/nodeSelector removal + resource-version annotation update. That's the other major code path introduced by this PR and is currently untested at the Cypress level.

frontend/src/pages/modelServing/screens/global/InferenceServiceTableRow.tsx
nananosirova: Fixed

frontend/src/concepts/hardwareProfiles/__tests__/utils.spec.ts
nananosirova: Tests added, but `getDeletedHardwareProfilePatches` is not new to this PR.

packages/model-serving/src/components/deployments/__tests__/DeploymentsTableRow.spec.tsx
nananosirova: Tests added

packages/cypress/cypress/tests/mocked/modelServing/modelServingGlobal.cy.ts
nananosirova: Covered

DaoDaoNoCode:

DaoDaoNoCode:
For the mocked test failure, can you cherry pick the commit in https://github.com/opendatahub-io/odh-dashboard/pull/9466? that should fix the CI failures on this branch. I did the same thing for 3.3 branch and 2.25 branch

frontend/src/concepts/hardwareProfiles/const.ts
```diff
@@ -85,3 +85,6 @@ export const INFERENCE_SERVICE_HARDWARE_PROFILE_PATHS: CrPathConfig = {
   tolerationsPath: 'spec.predictor.tolerations',
   nodeSelectorPath: 'spec.predictor.nodeSelector',
 };
+
+export const HARDWARE_PROFILE_RESOURCE_VERSION_PATH =
+  '/metadata/annotations/opendatahub.io~1hardware-profile-resource-version';
```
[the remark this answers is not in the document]
nananosirova: That annotation is left behind on purpose, it's the breadcrumb the hook uses at `useHardwareProfileBindingState` to know a profile was once bound after the name is gone.

packages/model-serving/src/components/deployments/row/DeploymentsTableRow.tsx
```diff
@@ -71,26 +86,43 @@ export const DeploymentRow: React.FC<{
...
+  ]);
 
   const onStop = React.useCallback(() => {
     if (dontShowModalValue) {
       startStopActionExtension?.properties
         .patchDeploymentStoppedStatus()
-        .then((resolvedFunction) => resolvedFunction(deployment, true));
+        .then((resolvedFunction) =>
+          resolvedFunction(deployment, true, deletedHardwareProfilePatches),
+        );
```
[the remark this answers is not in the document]
nananosirova: This row has no local stopping state, the toggle reads `deployment.status.stoppedStates` derived from the CR, there's nothing for a catch to reset and a failed patch just leaves it showing Running.

nananosirova:

DaoDaoNoCode (approved):
/lgtm

PR #9491, opened 2026-08-26
RHOAIENG-88561: Register data-registry module in dashboard-operator

Fiona-Waters:
## Description

Registers the data-registry module in dashboard-operator to enable standalone deployment and orchestration by the operator.

This PR is part of the Konflux onboarding for the data-registry component and addresses Phase 5a (operator registration).

**Note:** This PR was rebased onto midstream/main (was originally based on upstream/main).

### Operator Module Registration

**dashboard-operator/internal/controller/modules.go:**
- Added dataRegistry module entry with:
  - Port: 9143 (resolved port conflict with notebooks)
  - DSC component gate: feastoperator
  - TLS: enabled
  - Container name: data-registry-ui
  - Image env var: RELATED_IMAGE_ODH_MOD_ARCH_DATA_REGISTRY_IMAGE
  - Manifest slug: data-registry

**dashboard-operator/internal/controller/module_deploy.go:**
- Added proxy path entry (default: /data-registry/api → /api)

**dashboard-operator/internal/controller/support.go:**
- Added image mapping (data-registry-ui-image → RELATED_IMAGE_ODH_MOD_ARCH_DATA_REGISTRY_IMAGE)

**dashboard-operator/charts/dashboard/values.yaml:**
- Added RELATED_IMAGE_ODH_MOD_ARCH_DATA_REGISTRY_IMAGE to relatedImages

### Port Conflict Resolution

Updated data-registry from port 9043 to 9143 (notebooks already uses 9043):

**packages/data-registry/package.json:**
- Updated service.port to 9143

**manifests/modules/data-registry/deployment.yaml:**
- Updated containerPort to 9143
- Updated --port flag to 9143
- Updated probe ports to 9143

**manifests/modules/data-registry/service.yaml:**
- Updated port and targetPort to 9143

### Test Updates

**dashboard-operator/internal/controller/modules_test.go:**
- Updated module count from 9 to 10 across all test cases
- Added dataRegistry to expected module lists
- Updated unknown module test to expect 11 items (10 modules + 1 unknown)

## Related PRs

- Upstream: opendatahub-io/odh-dashboard#9491 (this PR)
- Downstream: red-hat-data-services/odh-dashboard#XXXX (Dockerfile.konflux.data-registry)
- OpenShift CI: Fiona-Waters/release#XXXX (ci-operator config)

## Testing

All dashboard-operator tests passing:
```bash
cd dashboard-operator && go test ./internal/controller/... -count=1
ok  	github.com/opendatahub-io/odh-dashboard/dashboard-operator/internal/controller	1.172s
```

## Notes

- Port 9143 was chosen to avoid conflict with notebooks (9043)
- The dataRegistry module depends on the feastoperator DSC component being available

## Summary by CodeRabbit

- **New Features**
  - Added Data Registry UI support to the dashboard, including routing through the Data Registry API.
  - Data Registry is enabled when the required Feast Operator component is available and disabled when it is removed.

- **Configuration**
  - Added Data Registry image configuration to the dashboard chart.
  - Updated the Data Registry UI port to 9143 across deployment, service, health checks, networking, and plugin configuration.

- **Tests**
  - Expanded module registry, status, routing, and integration coverage to include Data Registry.


Fiona-Waters:
Closing in favor of #9492 which is correctly based on midstream/main instead of upstream/main.

manifests/modules/data-registry/service.yaml
```diff
@@ -11,5 +11,5 @@ spec:
   ports:
     - name: data-registry-ui
       protocol: TCP
-      port: 9043
-      targetPort: 9043
+      port: 9143
```
liday-rh, line 14: The port migration also needs to update `manifests/modules/data-registry/networkpolicy.yaml`, which still permits ingress only on `9043`. Since that policy selects the data-registry pod, traffic forwarded by this Service to `9143` will be denied. Please change the NetworkPolicy ingress port to `9143` as part of this migration.

liday-rh (changes requested):
The operator registration and port updates are generally consistent, but the port migration leaves the data-registry NetworkPolicy on the old port. This will block ingress to the newly configured `9143` listener.

manifests/modules/data-registry/service.yaml
Fiona-Waters: Updated network policy to 9143. Thank you

dashboard-operator/internal/controller/modules.go
```diff
@@ -119,6 +119,16 @@ var moduleRegistry = map[string]ModuleDefinition{
 		ManifestSlug:  "notebooks",
 		TLS:           true,
 	},
+	"dataRegistry": {
+		Name:                  "dataRegistry",
+		ContainerName:         "data-registry-ui",
+		Port:                  9143,
+		ImageEnvVar:           "RELATED_IMAGE_ODH_MOD_ARCH_DATA_REGISTRY_IMAGE",
```
liday-rh, line 131: [P1] Register this image in the platform operator manifest before enabling the module. The disconnected-readiness check reports that RELATED_IMAGE_ODH_MOD_ARCH_DATA_REGISTRY_IMAGE is absent from the operator manifest, so disconnected installations will not inject or mirror this image and the standalone module cannot start.

dashboard-operator/internal/controller/modules.go
```diff
@@ -119,6 +119,16 @@ var moduleRegistry = map[string]ModuleDefinition{
 		ManifestSlug:  "notebooks",
 		TLS:           true,
 	},
+	"dataRegistry": {
```
liday-rh, line 127: [P2] Extend the all-modules federation integration fixture for this registry entry. allModuleSlugs does not include data-registry, and the test Component map does not mark feastoperator Managed. As a result, TestIntegration_FederationConfigMap_AllModulesEnabled currently fails because dataRegistry is missing from the generated federation config.

liday-rh (changes requested):
Two blocking issues remain: the new image is not available through the disconnected operator manifest, and the all-modules federation integration fixture was not extended for dataRegistry.

dashboard-operator/internal/controller/modules.go
Fiona-Waters: Updated in this PR.

dashboard-operator/internal/controller/modules.go
Fiona-Waters: If I understand this correctly, we can't do this until we have the image url which we won't have until after konflux onboarding and build?

liday-rh: You're right. My bad!

dashboard-operator/internal/controller/modules_test.go
```diff
@@ -137,9 +140,36 @@ func TestResolveModuleStatuses(t *testing.T) {
...
+				"dataRegistry": v1alpha1.ModulePhaseDisabled,
+			},
+			wantReason: map[string]string{
+				"dataRegistry": "ComponentNotAvailable",
+			},
+		},
 		{
 			name:    "DSC component absent from non-nil map disables module",
-			wantLen: 9,
+			wantLen: 10,
```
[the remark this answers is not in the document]
liday-rh: @Fiona-Waters, resolve this and then we are good to go!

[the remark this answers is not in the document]
Fiona-Waters: Done, and squashed commits. Thank you

liday-rh (approved):
/lgtm

lucferbux:
/lgtm
/approve

Reviewed the operator onboarding changes for the `data-registry` module.

**Port conflict resolution — verified correct:**
- `dataRegistry` is assigned port `9143`, which is unique across the entire operator module registry (8043, 8143, 8243, 8343, 8443/perses, 8543, 8643, 8743, 8843, 8943/core-bff, 9043/notebooks, 9143). No collision.
- Port `9143` is consistent across all touched surfaces: `modules.go` registry, `deployment.yaml` (containerPort, `--port`, liveness/readiness probes), `service.yaml` (port/targetPort), `networkpolicy.yaml`, and `package.json` service port. The old `9043` (shared with notebooks) is fully retired for this module.

**Operator changes — verified correct:**
- Registry entry (container name, image env var, manifest slug, TLS, `feastoperator` DSC gate) is well-formed; `feastoperator` matches the DSC component naming convention (cf. `mlflowoperator`).
- The image param (`data-registry-ui-image` → `RELATED_IMAGE_ODH_MOD_ARCH_DATA_REGISTRY_IMAGE`) is derived automatically by the `init()` loop over `moduleRegistry` in `support.go`, so no manual map edit is needed (PR description is slightly stale on this, but the code is right).
- Federation entry correctly consumes `mod.TLS`/`mod.Port`; `values.yaml` relatedImages and `manifests/modules/kustomization.yaml` are updated.
- Tests: module counts bumped 9→10 (and 10→11 for the unknown-module case); added feastoperator Managed/Removed cases and the all-modules federation case. Ran `go build`, `go vet -tags integration`, and `go test ./internal/controller/...` locally — all pass. Dynamic count assertions elsewhere use `len(ModuleNames())` and auto-adapt.

**Non-blocking nit:** the explicit `ProxyPaths: [{/data-registry/api → /api}]` is byte-for-byte identical to what `proxyPathsFor` already produces by default for slug `data-registry`, so it could be left `nil` (as notebooks does). Harmless — not worth blocking on.

PR #9497, opened 2026-08-26
feat(nim): prioritize NVIDIA hardware profiles in NIM wizard

YuliaKrimerman:
https://issues.redhat.com/browse/RHOAIENG-82818

## Description

NIM deployments only work with NVIDIA GPUs, but the hardware profile dropdown in the deployment wizard listed profiles in default alphabetical order — making NVIDIA profiles harder to find.

This change prioritizes NVIDIA hardware profiles at the top of the hardware profile selector when the NIM wizard is active (`ModelLocationType.NIM`):

- Adds `isNvidiaHardwareProfile` to detect profiles with an `nvidia.com/*` accelerator identifier (covers `nvidia.com/gpu` and MIG variants).
- Adds `prioritizeHardwareProfiles` to reorder profiles while preserving existing alpha / dashboard `hardwareProfileOrder` sorting within each group.
- Extends `HardwareProfileSelect` and `HardwareProfileFormSection` with an optional `isHardwareProfilePreferred` callback.
- Wires NIM-specific preference in `ModelDeploymentStep` via `ModelServingHardwareProfileSection`.

KServe and other wizard flows are unchanged.

## How Has This Been Tested?
Screenshot 2026-08-26 at 12 23 29 PM

- Ran `npx turbo run lint --filter=@odh-dashboard/hardware-profiles --filter=@odh-dashboard/model-serving` — passed (pre-existing warnings only).
- Added unit tests for `isNvidiaHardwareProfile` and `prioritizeHardwareProfiles` in `packages/hardware-profiles/src/pages/__tests__/utils.spec.ts`.
- Manual testing in the NIM deployment wizard is recommended: open the hardware profile dropdown and confirm NVIDIA GPU profiles appear before CPU-only / non-NVIDIA profiles.

## Test Impact

- **Unit tests added:** `isNvidiaHardwareProfile` and `prioritizeHardwareProfiles` coverage in `packages/hardware-profiles/src/pages/__tests__/utils.spec.ts`.
- **Cypress:** No new Cypress tests. NIM wizard profile ordering could be covered in `packages/model-serving/cypress/tests/mocked/modelServing/modelServingNim.cy.ts` as a follow-up.
- **Note:** `hardware-profiles` does not currently have its own `test-unit` script; confirm the new tests are picked up by CI or wire a runner for that package.

## Request review criteria:

If you have UI changes:

## Summary by CodeRabbit

* **New Features**
  * Hardware profile selections can now prioritize profiles matching the deployment’s preferred accelerator.
  * NVIDIA-compatible hardware profiles are automatically prioritized for NIM deployments.
  * Deployment wizard state now exposes computed field overrides.

* **Tests**
  * Added coverage for accelerator detection, profile prioritization, and NIM hardware profile preferences.

### DIFF
--- a/packages/model-serving/src/components/deploymentWizard/fields/ModelServingHardwareProfileSection.tsx
+++ b/packages/model-serving/src/components/deploymentWizard/fields/ModelServingHardwareProfileSection.tsx
@@ -8,16 +8,18 @@ import type {
   HardwarePodSpecOptionsState,
   PodSpecOptions,
 } from '@odh-dashboard/hardware-profiles/shared';
+import type { HardwareProfileKind } from '@odh-dashboard/k8s-core';
 
 type ModelServingHardwareProfileSectionComponentProps = {
   hardwareProfileConfig: UseHardwareProfileConfigResult;
   project?: string;
   isEditing?: boolean;
+  isHardwareProfilePreferred?: (profile: HardwareProfileKind) => boolean;
 };
 
 export const ModelServingHardwareProfileSection: React.FC<
   ModelServingHardwareProfileSectionComponentProps
-> = ({ hardwareProfileConfig, project, isEditing = false }) => {
+> = ({ hardwareProfileConfig, project, isEditing = false, isHardwareProfilePreferred }) => {
   const podSpecOptionsState: HardwarePodSpecOptionsState<PodSpecOptions> = React.useMemo(
     () => ({
       hardwareProfile: hardwareProfileConfig,
@@ -36,6 +38,7 @@ export const ModelServingHardwareProfileSection: React.FC<
       podSpecOptionsState={podSpecOptionsState}
       isEditing={isEditing}
       isHardwareProfileSupported={() => true}
+      isHardwareProfilePreferred={isHardwareProfilePreferred}
       visibleIn={MODEL_SERVING_VISIBILITY}
     />
   );
--- a/packages/model-serving/src/components/deploymentWizard/steps/ModelDeploymentStep.tsx
+++ b/packages/model-serving/src/components/deploymentWizard/steps/ModelDeploymentStep.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import { Form, FormSection, Spinner } from '@patternfly/react-core';
-import { isNvidiaHardwareProfile } from '@odh-dashboard/hardware-profiles/shared';
+import { isHardwareProfileWithAcceleratorPrefix } from '@odh-dashboard/hardware-profiles/shared';
+import type { HardwareProfileKind } from '@odh-dashboard/k8s-core';
 import K8sNameDescriptionField from '@odh-dashboard/ui-core/components/K8sNameDescriptionField';
 import { UseModelDeploymentWizardState } from '../useDeploymentWizard';
 import ProjectSection from '../fields/ProjectSection';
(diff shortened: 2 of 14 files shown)


packages/model-serving/src/components/deploymentWizard/steps/ModelDeploymentStep.tsx
```diff
@@ -30,6 +32,7 @@ export const ModelDeploymentStepContent: React.FC<ModelDeploymentStepProps> = ({
   hideProjectSection,
 }) => {
   const hideHwp = isNonSingleNodeTopologyActive(wizardState.state);
+  const isNimWizard = wizardState.state.modelLocationData.data?.type === ModelLocationType.NIM;
```
emilys314: This is NIM platform leaking into the generic model-serving package.

Two ways to do this with extensions:
A. You can a hardware profile "preferredAccelerator" to `WizardStateOverrides` in `packages/model-serving/src/shared/types/form-data.ts` and then make the `NIMImageField` have a `getFieldOverrides` 
B. Create a `DeploymentWizardFieldOverride` for the hardware profile

packages/model-serving/src/components/deploymentWizard/steps/ModelDeploymentStep.tsx
```diff
@@ -30,6 +32,15 @@ export const ModelDeploymentStepContent: React.FC<ModelDeploymentStepProps> = ({
...
   const hideHwp = isNonSingleNodeTopologyActive(wizardState.state);
+  const preferredAccelerator = wizardState.computedOverrides.hardwareProfile?.preferredAccelerator;
+
+  const isHardwareProfilePreferred = React.useMemo(
+    (): ((profile: HardwareProfileKind) => boolean) | undefined =>
+      preferredAccelerator
+        ? (profile) => isHardwareProfileWithAcceleratorPrefix(profile, preferredAccelerator)
+        : undefined,
+    [preferredAccelerator],
+  );
```
emilys314, line 43: We shouldn't have field logic outside of the field component or the field hook. The Step component should only be responsible for rendering the fields it containers in the correct order. 

You should pass in `wizardState.computedOverrides.hardwareProfile?.preferredAccelerator` directly into the `ModelServingHardwareProfileSection` field

mturley (commented):
Still reviewing, but is it a problem that we have non-Nvidia hardware profiles marked “compatible”? I know that logic is pre-existing and it’s probably out of scope of this PR but I wonder if we should follow up on that. cc @emilys314

packages/model-serving/src/components/deploymentWizard/fields/ModelServingHardwareProfileSection.tsx
```diff
@@ -36,6 +38,7 @@ export const ModelServingHardwareProfileSection: React.FC<
       podSpecOptionsState={podSpecOptionsState}
       isEditing={isEditing}
       isHardwareProfileSupported={() => true}
+      isHardwareProfilePreferred={isHardwareProfilePreferred}
```
emilys314, line 41: So the UI looks a bit confusing as is here, the other profiles show as "Compatible" even though we know they won't work.

Image

One suggestion is to pass the same fn to `isHardwareProfileSupported` then it would correctly show "Compatible"
```suggestion
isHardwareProfileSupported={isHardwareProfilePreferred}
      isHardwareProfilePreferred={isHardwareProfilePreferred}
```

Image

mturley (approved):
/approve
/lgtm

Tested and working as expected

image

I do think we should consider following up with another PR or maybe a separate ticket for re-evaluating the “compatible” label in this context.

PR #9509, opened 2026-08-27
fix: remove TrainJob node scaling 

PR3MM:
https://issues.redhat.com/browse/RHOAIENG-88673

## Description

Kubeflow Trainer 2.2 made spec.trainer immutable (kubeflow/trainer#3157), so the dashboard's scale flow — which PATCHes spec.trainer.numNodes after create — is now rejected by the TrainJob validating webhook with spec.trainer: Forbidden: field is immutable. This is what's been failing testPauseAndScaleTrainjob.cy.ts on 3.6 EA1.

## What changed:
- Removed the scale modal, its API client, and all three entry points (table row pencil, table row kebab, drawer kebab + Resources tab)
- Node count is still shown, read-only, in both the table and the drawer
- Pause/resume is untouched (it patches spec.suspend, unaffected)
- RayJob scaling is untouched — different CRD, still works

Only affects the 3.6 stream; 3.5 is on Trainer 2.1 and doesn't have this problem. The hook keeps its original name and carries a comment pointing at this ticket, so reverting this commit restores the feature when upstream supports it again.

## How Has This Been Tested?

- packages/model-training Jest suite: 438 passing
- Full model-training mock Cypress suite: 142/142 across 7 specs, including all 41 Ray tests (the scale page objects are shared with Ray, so this was the main regression risk)
- testPauseAndScaleTrainjob.cy.ts run against a RHOAI 3.6 EA1 cluster — pause, drawer, and the read-only node count all verified
- ESLint and tsc --noEmit clean on both changed packages

## BEFORE 

https://github.com/user-attachments/assets/3c33c855-8533-49fe-a3e2-4e1489c72a4f

## AFTER
No edit option for Nodes 
SCR-20260831-llma

All tests passing expect for the `modelTrainingNodeScaling.cy.ts` since it is skipped intentionally
SCR-20260831-lhit

## Test Impact

- Deleted modelTrainingNodeScaling.cy.ts and api/__tests__/scaling.spec.ts — both tested only the removed flow
- Added two mock tests asserting the action is gone: one for the drawer kebab (modelTrainingPauseResume.cy.ts), one for the table row kebab and pencil (modelTrainingTable.cy.ts)
- Updated modelTrainingDetailsResources.cy.ts — the node count was a disabled button, now plain text
- Trimmed the hook spec to the node-count cases. Worth flagging: the old ClusterTrainingRuntime fallback test was a false positive (themock's default numNodes matched the runtime's, so the fallback branch never ran) — the replacement builds a job with no trainer spec and asserts a distinct value
- testPauseAndScaleTrainjob.cy.ts drops the scale steps; pause/resume/complete/delete coverage is unchanged

## Self checklist (all need to be checked):

## Summary by CodeRabbit

* **Changes**
  * Training job node counts are now read-only after job creation.
  * Removed node-count editing actions and scaling dialogs from training job lists and details.
  * Pause, resume, completion, and initial node-count displays remain available.
  * Node-scaling test coverage is temporarily disabled while this capability is unavailable.


claudialphonse78 (commented):
@PR3MM why are you removing the code for the scaling component altogether you just have to remove the place where its getting called in the code the react component and the E2E test case the code part where that is happening that would be removed, we dont have to remove all the scaling code

pnaik1 (approved):

manaswinidas (approved):
/approve

PR #9521, opened 2026-08-30
fix(RHOAIENG-81919): watch platform config ConfigMap for version changes

lucferbux:
Fixes [RHOAIENG-81919](https://issues.redhat.com/browse/RHOAIENG-81919)

## Description

The `dashboard-operator` writes `status.releases[platform].version` from the `odh-dashboard-config` ConfigMap on every reconcile, but the controller did not **watch** that ConfigMap. When the platform operator updates `platformVersion` (or any other platform-managed key/annotation) in `odh-dashboard-config`, no reconcile was triggered, so `status.releases[platform].version` stayed stale until an unrelated event (e.g. touching the CR, an owned-resource change, or the periodic requeue) happened to fire.

This change wires the config ConfigMaps that feed reconcile inputs into the controller's watch set:

- Adds a `Watches(&corev1.ConfigMap{}, ...)` to `SetupWithManager` that enqueues the singleton `default-dashboard` Dashboard when a watched config ConfigMap changes.
- A map function (`mapConfigMapToDashboard`) scopes the enqueue to `odh-dashboard-config` (`platformVersion` + distribution identity) and `dashboard-operator-config` (`reconcileInterval`) **in the operator namespace** only.
- A predicate (`configMapPredicate`) filters events to those ConfigMaps and, on updates, fires only when `Data` or `Annotations` actually change — avoiding no-op reconciles from metadata churn (e.g. `resourceVersion` bumps or managed-field rewrites). Annotations are compared because the distribution identity can arrive as orchestrator annotations, not just data keys (see `readDistributionConfig`).

Notes:
- No RBAC change is required — the operator already has `get;list;watch;...` on `configmaps` (it `Owns(&corev1.ConfigMap{})`).
- The existing `Owns(&corev1.ConfigMap{})` and the new `Watches(&corev1.ConfigMap{}, ...)` coexist: owned ConfigMaps (e.g. `federation-config`) are filtered out of the new watch by the name/namespace predicate, so they are not double-enqueued.

## How Has This Been Tested?

Automated checks run locally from `dashboard-operator/`:

- `go build ./...` — passes
- `go vet ./...` — passes
- `make lint` (golangci-lint v2, standard preset) — `0 issues`
- `go test -race -count=1 ./internal/controller/...` — passes (full package, including the new tests)

Manual reproduction from the ticket (ConfigMap patch → reconcile → status refresh) is best exercised in a live cluster / DAG E2E; that end-to-end coverage is tracked separately in the linked RHOAIENG-82790.

## Test Impact

Added `dashboard-operator/internal/controller/config_watch_test.go` with table-driven unit tests (internal funcs exposed via `export_test.go`):

- `mapConfigMapToDashboard` — enqueues the singleton for each watched ConfigMap in the operator namespace; ignores watched names in other namespaces and unrelated ConfigMap names.
- `configMapPredicate` — Create/Delete/Generic fire only for watched ConfigMaps in the operator namespace; Update fires on `Data` change and on `Annotations` change, but not on metadata-only churn, and never for unrelated names or other namespaces.

## Request review criteria:

[RHOAIENG-81919]: https://redhat.atlassian.net/browse/RHOAIENG-81919?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * The dashboard now automatically reconciles when relevant platform or operator configuration changes.

* **Bug Fixes**
  * Configuration event handling now ignores unrelated resources, namespaces, and metadata-only updates, reducing unnecessary reconciliations.
  * Changes to configuration data and supported platform metadata are detected reliably.

* **Tests**
  * Added coverage for configuration change detection, event filtering, and dashboard reconciliation triggers.

### DIFF
--- a/dashboard-operator/internal/controller/config_watch_test.go
+++ b/dashboard-operator/internal/controller/config_watch_test.go
@@ -187,3 +187,44 @@ func TestConfigMapPredicate_Update(t *testing.T) {
 		})
 	}
 }
+
+// TestConfigMapWatch_HonorsCustomDistributionConfigMapName verifies the watch
+// tracks the resolved distribution config name rather than the hardcoded default.
+// The distribution config ConfigMap name is user-settable (chart value
+// config.name, plumbed in via OPERATOR_CONFIGMAP_NAME); readDistributionConfig /
+// readPlatformVersion resolve it through resolveDistributionConfigMapName. If the
+// watch kept the literal default, a non-default install would miss updates to its
+// real ConfigMap and status.releases[platform].version would go stale — the exact
+// failure this PR fixes (RHOAIENG-81919).
+func TestConfigMapWatch_HonorsCustomDistributionConfigMapName(t *testing.T) {
+	const customName = "custom-distribution-config"
+	t.Setenv("OPERATOR_CONFIGMAP_NAME", customName)
+
+	r := &ctrlpkg.DashboardReconciler{Namespace: testNamespace}
+	p := r.ConfigMapPredicate()
+
+	custom := newConfigMap(customName, testNamespace, map[string]string{"platformVersion": "2.20.0"}, nil)
+
+	t.Run("map func enqueues for the custom-named ConfigMap", func(t *testing.T) {
+		reqs := r.MapConfigMapToDashboard(context.Background(), custom)
+		require.Len(t, reqs, 1)
+		assert.Equal(t, v1alpha1.DashboardInstanceName, reqs[0].Name)
+	})
+
+	t.Run("the hardcoded default name is not watched once a custom name is set", func(t *testing.T) {
+		def := newConfigMap(distributionConfigMapName, testNamespace, map[string]string{"platformVersion": "2.20.0"}, nil)
+		assert.Empty(t, r.MapConfigMapToDashboard(context.Background(), def))
+		assert.False(t, p.Create(event.CreateEvent{Object: def}))
+	})
+
+	t.Run("predicate fires on data change for the custom ConfigMap", func(t *testing.T) {
+		assert.True(t, p.Create(event.CreateEvent{Object: custom}))
+		updated := newConfigMap(customName, testNamespace, map[string]string{"platformVersion": "2.21.0"}, nil)
+		assert.True(t, p.Update(event.UpdateEvent{ObjectOld: custom, ObjectNew: updated}))
+	})
+
+	t.Run("predicate fires on consumed annotation change for the custom ConfigMap", func(t *testing.T) {
+		annoUpdated := newConfigMap(customName, testNamespace, map[string]string{"platformVersion": "2.20.0"}, map[string]string{annotations.PlatformVersion: "2.21.0"})
+		assert.True(t, p.Update(event.UpdateEvent{ObjectOld: custom, ObjectNew: annoUpdated}))
+	})
+}
--- a/dashboard-operator/internal/controller/config_watch_test.go
+++ b/dashboard-operator/internal/controller/config_watch_test.go
@@ -10,6 +10,8 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/event"
 
+	"github.com/opendatahub-io/odh-platform-utilities/pkg/metadata/annotations"
+
 	v1alpha1 "github.com/opendatahub-io/odh-dashboard/dashboard-operator/api/v1alpha1"
 	ctrlpkg "github.com/opendatahub-io/odh-dashboard/dashboard-operator/internal/controller"
 )
@@ -109,6 +111,16 @@ func TestConfigMapPredicate_Update(t *testing.T) {
 	r := &ctrlpkg.DashboardReconciler{Namespace: testNamespace}
 	p := r.ConfigMapPredicate()
 
+	// metadata churn: identical Data and consumed annotations, but the object
+	// metadata differs (resourceVersion bump + a rewritten label). The predicate
+	// must ignore this — it is the exact regression the predicate guards against.
+	churnOld := newConfigMap(distributionConfigMapName, testNamespace, map[string]string{"platformVersion": "2.20.0"}, nil)
+	churnOld.ResourceVersion = "100"
+	churnOld.Labels = map[string]string{"example.com/synced-at": "old"}
+	churnNew := newConfigMap(distributionConfigMapName, testNamespace, map[string]string{"platformVersion": "2.20.0"}, nil)
+	churnNew.ResourceVersion = "101"
+	churnNew.Labels = map[string]string{"example.com/synced-at": "new"}
+
 	tests := []struct {
 		name    string
 		old     *corev1.ConfigMap
@@ -123,15 +135,21 @@ func TestConfigMapPredicate_Update(t *testing.T) {
 		},
 		{
 			name:    "identical data with only metadata churn does not fire",
-			old:     newConfigMap(distributionConfigMapName, testNamespace, map[string]string{"platformVersion": "2.20.0"}, nil),
-			updated: newConfigMap(distributionConfigMapName, testNamespace, map[string]string{"platformVersion": "2.20.0"}, nil),
+			old:     churnOld,
+			updated: churnNew,
 			want:    false,
 		},
 		{
-			name:    "annotation change fires a reconcile",
+			name:    "consumed annotation change fires a reconcile",
 			old:     newConfigMap(distributionConfigMapName, testNamespace, map[string]string{"platformVersion": "2.20.0"}, nil),
-			updated: newConfigMap(distributionConfigMapName, testNamespace, map[string]string{"platformVersion": "2.20.0"}, map[string]string{"platform.opendatahub.io/version": "2.21.0"}),
+			updated: newConfigMap(distributionConfigMapName, testNamespace, map[string]string{"platformVersion": "2.20.0"}, map[string]string{annotations.PlatformVersion: "2.21.0"}),
 			want:    true,
+		},
+		{
+			name:    "unrelated annotation churn does not fire",
+			old:     newConfigMap(distributionConfigMapName, testNamespace, map[string]string{"platformVersion": "2.20.0"}, map[string]string{"kubectl.kubernetes.io/last-applied-configuration": "{}"}),
+			updated: newConfigMap(distributionConfigMapName, testNamespace, map[string]string{"platformVersion": "2.20.0"}, map[string]string{"kubectl.kubernetes.io/last-applied-configuration": "{\"data\":{}}"}),
+			want:    false,
 		},
 		{
 			name:    "unrelated ConfigMap does not fire even on data change",
--- a/dashboard-operator/internal/controller/config_watch_test.go
+++ b/dashboard-operator/internal/controller/config_watch_test.go
@@ -146,6 +146,21 @@ func TestConfigMapPredicate_Update(t *testing.T) {
 			want:    true,
 		},
 		{
+			// The PlatformType/PlatformVersion annotations are consumed only from
+			// odh-dashboard-config; dashboard-operator-config contributes reconcile
+			// inputs via Data alone, so an annotation change on it is a no-op.
+			name:    "consumed annotation change on operator config does not fire",
+			old:     newConfigMap(operatorConfigMapName, testNamespace, map[string]string{"reconcileInterval": "30s"}, nil),
+			updated: newConfigMap(operatorConfigMapName, testNamespace, map[string]string{"reconcileInterval": "30s"}, map[string]string{annotations.PlatformVersion: "2.21.0"}),
+			want:    false,
+		},
+		{
+			name:    "operator config data change fires a reconcile",
+			old:     newConfigMap(operatorConfigMapName, testNamespace, map[string]string{"reconcileInterval": "30s"}, nil),
+			updated: newConfigMap(operatorConfigMapName, testNamespace, map[string]string{"reconcileInterval": "60s"}, nil),
+			want:    true,
+		},
+		{
 			name:    "unrelated annotation churn does not fire",
 			old:     newConfigMap(distributionConfigMapName, testNamespace, map[string]string{"platformVersion": "2.20.0"}, map[string]string{"kubectl.kubernetes.io/last-applied-configuration": "{}"}),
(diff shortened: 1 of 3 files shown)


dashboard-operator/internal/controller/dashboard_reconciler.go
```diff
@@ -875,11 +882,80 @@ func SetupWithManager(mgr ctrl.Manager, opts Options) error {
...
 		Complete(r)
 }
+
+// watchedConfigMaps returns the set of ConfigMap names, in the operator
+// namespace, whose contents feed reconcile inputs: platform version and
+// distribution identity (odh-dashboard-config), and the reconcile interval
+// (dashboard-operator-config). Changes to these must trigger a reconcile so the
+// Dashboard status stays fresh even when nothing else touches the CR
+// (RHOAIENG-81919).
+func watchedConfigMaps() map[string]bool {
```
liday-rh: **Low: per-event map allocation for a constant value.**

`watchedConfigMaps()` builds a fresh two-entry map on every call, and it is called from both `isWatched` (line 939) and `mapConfigMapToDashboard` (line 922). Because the ConfigMap informer is cluster-wide (established by the pre-existing `Owns(&corev1.ConfigMap{})`), these run on ConfigMap events beyond just this operator's own resources.

Scenario: short-circuit evaluation keeps the blast radius to ConfigMap events in the operator namespace, so this is not a hot path in most clusters — but it is still a heap allocation per event for a value that never changes, and on an update to a watched ConfigMap it allocates twice (predicate, then map func).

Suggested change: make it a package-level `var watchedConfigMaps = map[string]bool{...}` and drop the function call, or use a `sets.Set[string]`.

dashboard-operator/internal/controller/dashboard_reconciler.go
```diff
@@ -875,11 +882,80 @@ func SetupWithManager(mgr ctrl.Manager, opts Options) error {
...
+	return map[string]bool{
+		distributionConfigMapName: true,
+		operatorConfigMapName:     true,
+	}
+}
+
+// mapConfigMapToDashboard enqueues a reconcile for the singleton Dashboard when
+// one of the watched config ConfigMaps in the operator namespace changes.
+func (r *DashboardReconciler) mapConfigMapToDashboard(_ context.Context, obj client.Object) []reconcile.Request {
+	if obj.GetNamespace() != r.Namespace || !watchedConfigMaps()[obj.GetName()] {
```
liday-rh: **Low: unguarded `obj` dereference, inconsistent with the nil-checked sibling.**

This line calls `obj.GetNamespace()` with no nil check, while `isWatched` at line 939 explicitly guards with `obj != nil && ...`. Two helpers over the same input disagree about whether nil is possible.

Scenario: today the predicate always runs before the handler, so nil never reaches this function and there is no live bug. But the safety is entirely incidental. If the `builder.WithPredicates(r.configMapPredicate())` option is ever dropped or reordered out of the `Watches(...)` call, or if `mapConfigMapToDashboard` is reused for a second watch without a predicate, this panics in the controller's event path instead of returning nil — and it is exported for tests via `export_test.go`, so a test can reach it directly too.

Suggested change: add the same `obj == nil` guard here, or better, have both paths share one `isWatchedConfigMap(obj)` helper so the nil handling cannot drift.

dashboard-operator/internal/controller/dashboard_reconciler.go
```diff
@@ -875,11 +882,80 @@ func SetupWithManager(mgr ctrl.Manager, opts Options) error {
...
+			}
+
+			oldCM, oldOK := e.ObjectOld.(*corev1.ConfigMap)
+			newCM, newOK := e.ObjectNew.(*corev1.ConfigMap)
+			if !oldOK || !newOK {
+				return true
+			}
+
+			return !maps.Equal(oldCM.Data, newCM.Data) ||
+				!maps.Equal(oldCM.Annotations, newCM.Annotations)
```
liday-rh: **Low, but the one worth fixing: the predicate compares the entire annotation map, not the two keys it actually cares about.**

The doc comment above says the point of this predicate is "avoiding no-op reconciles from metadata churn." Comparing all of `Annotations` partially defeats that: only `annotations.PlatformType` and `annotations.PlatformVersion` are consumed by `readDistributionConfig`, but any annotation delta at all returns `true` here.

Scenario: a `kubectl apply` against `odh-dashboard-config` rewrites `kubectl.kubernetes.io/last-applied-configuration`; a `helm upgrade` rewrites `meta.helm.sh/release-name` / `meta.helm.sh/release-namespace`; an Argo CD sync stamps `argocd.argoproj.io/tracking-id`. None of these change any value the controller reads, yet each one enqueues a full reconcile: kustomize render of every manifest, `writeParamsEnv` disk writes under `ManifestsBasePath`, and a server-side apply of every rendered resource. In a GitOps-managed cluster that re-syncs on a timer, this is recurring churn that the predicate was specifically added to prevent.

Suggested change: compare only the keys that feed reconcile inputs, e.g.

```go
annotationChanged := oldCM.Annotations[annotations.PlatformType] != newCM.Annotations[annotations.PlatformType] ||
    oldCM.Annotations[annotations.PlatformVersion] != newCM.Annotations[annotations.PlatformVersion]

return !maps.Equal(oldCM.Data, newCM.Data) || annotationChanged
```

That keeps the behaviour the comment describes (distribution identity arriving as orchestrator annotations) while dropping the false positives.

dashboard-operator/internal/controller/config_watch_test.go
```diff
@@ -0,0 +1,156 @@
...
+		want    bool
+	}{
+		{
+			name:    "data change fires a reconcile",
+			old:     newConfigMap(distributionConfigMapName, testNamespace, map[string]string{"platformVersion": "2.20.0"}, nil),
+			updated: newConfigMap(distributionConfigMapName, testNamespace, map[string]string{"platformVersion": "2.21.0"}, nil),
+			want:    true,
+		},
+		{
+			name:    "identical data with only metadata churn does not fire",
```
liday-rh, line 137: **Low: this case does not exercise the metadata churn it is named for.**

The `old` and `updated` ConfigMaps on the next two lines are byte-identical: same name, same namespace, same `Data`, nil `Annotations`, and no `ResourceVersion`, `Generation`, or `ManagedFields` set on either. There is no metadata difference for the predicate to ignore.

Scenario: the comment on `configMapPredicate` calls out `resourceVersion` bumps and managed-field rewrites as the exact regression to guard against. As written, this test would still pass if someone changed the implementation to compare `ResourceVersion` or to use `reflect.DeepEqual` on the whole object — i.e. it does not actually protect against the regression it exists for.

Suggested change: give the two objects differing metadata that the predicate must ignore, e.g. set `ResourceVersion: "100"` on `old` and `"101"` on `updated` (and optionally differing `ManagedFields` or a differing label), keeping `Data` and `Annotations` equal.

liday-rh (commented):
Review of the ConfigMap watch. Verification I actually ran, all against this diff applied to a scratch copy of `dashboard-operator/`:

- `go build ./...`, `go vet ./internal/controller/`, and `go test -count=1 ./internal/controller/` all pass.
- **Namespace alignment**: `readOperatorConfig`, `readDistributionConfig`, and `readPlatformVersion` in `internal/controller/config.go` all read from `r.Namespace`, which is the same namespace the new predicate and map func filter on. `opts.Namespace` comes from `--namespace`/`OPERATOR_NAMESPACE` and is enforced non-empty in `cmd/manager/main.go`, so the filter can never degenerate to an empty-string match.
- **RBAC**: confirmed sufficient. The ClusterRole in `charts/dashboard/templates/rbac.yaml` already grants cluster-wide `get;list;watch` on `configmaps`, and the pre-existing `Owns(&corev1.ConfigMap{})` already established the cluster-wide ConfigMap informer, so the new `Watches` adds no informer and no cache cost.
- **No self-triggered reconcile loop**: neither `odh-dashboard-config` nor `dashboard-operator-config` is ever written by the operator (only `cli.Get`). The chart-created `odh-dashboard-config` carries no owner ref to the Dashboard CR, no kustomize `configMapGenerator` under `manifests/` emits either name, and `teardownManagedResources` filters on `platform.opendatahub.io/part-of`, which the chart labels do not set. So the operator cannot trigger itself through this watch.
- **`Owns` + `Watches` on the same GVK**: legal in controller-runtime v0.24.1. `builder.doWatch` registers them as independent sources over the shared informer, and `builder.WithPredicates` is scoped per-watch (it does not leak into `For`/`Owns`).
- **No stale-read race**: the reconciler reads through the cached client fed by the same informer that raised the event, so the read is at least as fresh as the trigger.

**I found no correctness bugs.** The fix does what the ticket describes. Four low-severity items follow inline; item 3 (whole-annotation-map comparison) is the only one I would actually push to fix, since it partially defeats the stated purpose of the predicate.

dashboard-operator/internal/controller/dashboard_reconciler.go
lucferbux: Good catch — done in dd87445. `watchedConfigMaps` is now a package-level `var map[string]bool` built once, no longer rebuilt per event.

dashboard-operator/internal/controller/dashboard_reconciler.go
lucferbux: Done in dd87445. Both the predicate and the map func now route through a single `isWatchedConfigMap(obj)` method that includes the `obj != nil` guard, so the nil handling and name/namespace scoping can no longer drift between the two paths.

dashboard-operator/internal/controller/dashboard_reconciler.go
lucferbux: Agreed, fixed in dd87445. The update predicate now compares only `annotations.PlatformType` and `annotations.PlatformVersion` (the keys `readDistributionConfig` actually consumes) alongside `Data`, instead of the whole annotation map. GitOps/Helm bookkeeping annotation churn (last-applied-configuration, meta.helm.sh/*, argocd tracking-id) no longer enqueues a no-op reconcile. Added a dedicated `unrelated annotation churn does not fire` test case covering this.

dashboard-operator/internal/controller/config_watch_test.go
lucferbux: Done in dd87445. The `identical data with only metadata churn does not fire` case now uses two objects with differing `ResourceVersion` (100 vs 101) and a differing label, while keeping `Data` and the consumed annotations equal — so it would fail if the implementation regressed to comparing `ResourceVersion` or `reflect.DeepEqual` on the whole object.

dashboard-operator/internal/controller/dashboard_reconciler.go
```diff
@@ -875,11 +883,87 @@ func SetupWithManager(mgr ctrl.Manager, opts Options) error {
...
+			}
+
+			oldCM, oldOK := e.ObjectOld.(*corev1.ConfigMap)
+			newCM, newOK := e.ObjectNew.(*corev1.ConfigMap)
+			if !oldOK || !newOK {
+				return true
+			}
+
+			annotationChanged := oldCM.Annotations[annotations.PlatformType] != newCM.Annotations[annotations.PlatformType] ||
+				oldCM.Annotations[annotations.PlatformVersion] != newCM.Annotations[annotations.PlatformVersion]
```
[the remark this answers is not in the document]
lucferbux: Fixed in 8f2ada231e. The `annotationChanged` check is now scoped to `newCM.Name == distributionConfigMapName`, so PlatformType/PlatformVersion annotation churn on `dashboard-operator-config` (which contributes reconcile inputs via `Data` only) no longer enqueues a no-op reconcile. Added two test cases: `consumed annotation change on operator config does not fire` and `operator config data change fires a reconcile`.

liday-rh (commented):
One new cross-PR integration finding after re-reviewing the current head. I verified the existing review comments are addressed and `make test` / `git diff --check` pass. The currently failing required CI result comes from an unrelated envtest asset download failure.

dashboard-operator/internal/controller/dashboard_reconciler.go
```diff
@@ -970,11 +978,95 @@ func SetupWithManager(mgr ctrl.Manager, opts Options) error {
...
+
+// watchedConfigMaps is the set of ConfigMap names, in the operator namespace,
+// whose contents feed reconcile inputs: platform version and distribution
+// identity (odh-dashboard-config), and the reconcile interval
+// (dashboard-operator-config). Changes to these must trigger a reconcile so the
+// Dashboard status stays fresh even when nothing else touches the CR
+// (RHOAIENG-81919). It never changes at runtime, so it is a package-level value
+// rather than rebuilt per event.
+var watchedConfigMaps = map[string]bool{
+	distributionConfigMapName: true,
```
liday-rh: **MEDIUM — this watch will miss the configurable distribution ConfigMap introduced by PR #9522.**

PR #9522 now injects the chart-rendered `config.name` as `OPERATOR_CONFIGMAP_NAME`, and `readDistributionConfig` / `readPlatformVersion` resolve that value. This set still contains only the literal default `odh-dashboard-config`, and the annotation check below also compares against that literal.

After both PRs merge, set `config.name` to a non-default value, then update `platformVersion` in that ConfigMap: reconcile reads the custom ConfigMap when it happens for another reason, but this watch rejects the update, so `status.releases[platform].version` remains stale—the exact failure this PR fixes.

Please derive the watched distribution name from the same resolver used by the readers (and use it for the annotation guard), rather than keeping an independent hardcoded name.

lucferbux: Good catch — fixed in 6daed1117f, after rebasing onto main (which now includes #9522). The watch no longer keys off the hardcoded default: `isWatchedConfigMap` and the update predicate's annotation guard both resolve the distribution name through the same `resolveDistributionConfigMapName()` helper the readers use, so a non-default `OPERATOR_CONFIGMAP_NAME` install watches its real ConfigMap. I dropped the static `watchedConfigMaps` map in favor of a direct two-name check (distribution config resolved via the helper + the hardcoded `dashboard-operator-config`, whose name #9522 left non-configurable). Added `TestConfigMapWatch_HonorsCustomDistributionConfigMapName`, which sets a custom `OPERATOR_CONFIGMAP_NAME` and asserts the custom ConfigMap is watched (map func + create + data/annotation update) while the default literal is not. `make test`, `make test-integration` (all 15 pass), and `make lint` are green.

PR #9392, opened 2026-08-19
RHOAIENG-61022: remove deprecated sidecar deployment mode

lucferbux:
https://issues.redhat.com/browse/RHOAIENG-61022

## Description

The dashboard completed its modular architecture transition to standalone deployment (each BFF module as its own Deployment). The sidecar mode (all BFF modules as containers in one pod) is fully deprecated and no longer used.

This PR removes all sidecar deployment artifacts:

- **Operator CRD**: Removed `DeploymentMode` type, `Sidecar`/`Standalone` constants, and the `deploymentMode` field from `DashboardSpec`
- **Reconciler**: Removed `reconcileSidecar()` and mode-branching logic; renamed `reconcileStandalone()` → `reconcileDeployment()` as the single code path; renamed `deleteSidecarResources()` → `cleanupLegacySidecarResources()` (kept for upgrade safety on clusters transitioning from sidecar)
- **Support/Actions**: Renamed overlay path maps (`standaloneOverlaysSourcePaths` → `overlaysSourcePaths`, old `overlaysSourcePaths` → `platformPaths` for observability); removed sidecar `manifestSets()` and sidecar params.env writing block
- **Manifests**: Deleted `manifests/sidecar/` (12 files), `manifests/odh/kustomization.yaml`, `manifests/odh/params.env`, `manifests/rhoai/base/` (15 files); cleaned sidecar references from standalone overlays
- **CI workflows**: Renamed `detect-sidecar-changes`/`sidecar-build-validation` → `detect-bff-changes`/`bff-build-validation`; updated release-auto-merge, validate-kustomize, and validate-manifests.sh
- **Documentation**: Updated 14+ docs, agent rules, and skills to remove sidecar mode descriptions
- **Helm chart**: Regenerated CRD YAML (no more `deploymentMode` enum), updated Chart.yaml keywords

72 files changed, +200 / -2,640 lines.

## How Has This Been Tested?

- `make generate && make manifests` — CRD regenerated without `deploymentMode`
- `make test` — all operator unit tests pass (controller, webhook, api, charts)
- `make lint` — 0 issues
- `kustomize build manifests/odh/standalone` and `kustomize build manifests/rhoai/standalone` — both succeed
- Verified no stray sidecar references remain (except intentional `cleanupLegacySidecarResources` for upgrade path)

## Test Impact

[...]

PR #9522, opened 2026-08-30
test(RHOAIENG-83640): expand envtest integration tests for dashboard operator

lucferbux:
Related issues:
- https://issues.redhat.com/browse/RHOAIENG-83645
- https://issues.redhat.com/browse/RHOAIENG-83646
- https://issues.redhat.com/browse/RHOAIENG-83647
- https://issues.redhat.com/browse/RHOAIENG-83648
- https://issues.redhat.com/browse/RHOAIENG-83649
- Epic (context): https://issues.redhat.com/browse/RHOAIENG-83640

## Description

Expands the envtest integration-test coverage for the **Dashboard Module Controller** (`dashboard-operator/`), implementing the remaining child tasks of the *Envtest Integration Test Foundation* epic ([RHOAIENG-83640](https://redhat.atlassian.net/browse/RHOAIENG-83640)) in a single PR. The foundation itself (`TestMain`, the helper harness, and the first four `TestIntegration_*` tests) landed earlier in #8742; this PR builds on that harness.

envtest runs the real reconcile loop against a real kube-apiserver (no kubelet/pod scheduling), so these tests catch CRD/CEL validation, status-subresource behaviour, SSA ownership, teardown/label semantics, and REST-mapper edge cases that the fake-client unit tests cannot.

What's added (by ticket):

- **[RHOAIENG-83645](https://redhat.atlassian.net/browse/RHOAIENG-83645) — standalone deployment reconciliation** (`deployment_integration_test.go`): asserts the core ConfigMap and per-module Deployments/Services are applied with the `part-of=dashboard` label, that the federation ConfigMap lists the enabled modules plus `coreBff`, and that every registered module gets a status entry (enabled → deployed/degraded, explicitly-disabled → `Disabled`/`ExplicitOverride`).
- **[RHOAIENG-83646](https://redhat.atlassian.net/browse/RHOAIENG-83646) — `managementState: Removed` teardown** (`managementstate_integration_test.go`): flipping a deployed Dashboard to `Removed` tears down module operands while preserving the CR and the operator's own resources (skipped by name), and reports `Phase=NotReady`, empty URL, nil module statuses, and `ProvisioningSucceeded=False (Removed)`. A second test asserts repeated reconciles are idempotent. Writing this test surfaced a teardown bug (see *Production fixes* below), which this PR also fixes.
- **[RHOAIENG-83647](https://redhat.atlassian.net/browse/RHOAIENG-83647) — observability reconciliation** (`observability_integration_test.go` + one unit test): covers `ObservabilityAvailable` = `Disabled`, `PersesCRDNotFound`, and `Deployed`. The CEL-unadmittable `InvalidConfig` case (enabled + no `persesService`) is unreachable from envtest, so it is covered by a unit test in `actions_test.go` that calls `deployObservabilityManifests` directly.
- **[RHOAIENG-83648](https://redhat.atlassian.net/browse/RHOAIENG-83648) — legacy sidecar cleanup** (`legacy_cleanup_integration_test.go`): a normal reconcile removes the resources left behind by the now-removed sidecar deployment mode (upgrade safety) while leaving unrelated resources untouched.
- **[RHOAIENG-83649](https://redhat.atlassian.net/browse/RHOAIENG-83649) — inter-BFF params + federation ConfigMap** (`federation_integration_test.go`): asserts gen-ai's `params.env` gains `BFF_MAAS_SERVICE_NAME`/`PORT` from its inter-BFF dependency (and maas gets none), that an all-modules-enabled config yields a federation entry per module plus `coreBff`/`mlflowEmbedded` (and no `perses` when observability is off), and that the main-dashboard pod-template federation-config hash changes when the enabled-module set changes.

### Reinterpretation note

Two tickets were written before the sidecar deployment mode was removed (#9392). The controller now has a single (standalone) deployment mode; there is no `DeploymentMode` field and no sidecar/standalone switching. The affected tickets were reinterpreted to match the current codebase and their Jira descriptions updated accordingly:
- [RHOAIENG-83645](https://redhat.atlassian.net/browse/RHOAIENG-83645) ("sidecar mode reconciliation") → standalone deployment reconciliation.
- [RHOAIENG-83648](https://redhat.atlassian.net/browse/RHOAIENG-83648) ("mode switching sidecar↔standalone") → legacy sidecar cleanup (upgrade safety).

### Implementation note (REST-mapper isolation)

The observability `_Deployed` test installs the `PersesDashboard` CRD and drives the reconcile with an **isolated** client whose RESTMapper is warmed with the perses mapping. This keeps the shared test client's mapper free of a cached perses mapping, which the `_PersesCRDNotFound` case depends on staying absent. `_Deployed` is also ordered last so its CRD create/delete never bleeds into the other cases.

## How Has This Been Tested?

All checks were run locally from `dashboard-operator/` (hermetic envtest — no live cluster):

- `make test-integration` — all 15 `TestIntegration_*` tests pass (11 new + 4 pre-existing). The observability group was run repeatedly (`-run TestIntegration_Observability`, several iterations) to confirm the RESTMapper isolation is not flaky.
- `make test` — unit suite green, including the new `TestDeployObservabilityManifests_PersesServiceRequired`; confirms the `//go:build integration` tag keeps the two suites isolated.
- `make lint` — 0 issues.
- `make build` — builds clean.
- `git status --porcelain` — no generated-file drift (no changes under `api/` or `config/`).

## Production fixes

Although the goal of this PR is test coverage, writing the tests surfaced two real defects in the operator, which are fixed here (both in `internal/controller/`):

- **Teardown deleted operator-owned resources** (`dashboard_reconciler.go`). `teardownManagedResources` deletes everything labeled `platform.opendatahub.io/part-of=dashboard` in the applications namespace, but the operator's own Helm chart stamps that same `commonLabel` on the resources it renders — including the webhook `Service` and the operator config `ConfigMap`, both of which live in that namespace. On `managementState: Removed` the teardown would delete the webhook Service, breaking the `failurePolicy: Fail` `ValidatingWebhookConfiguration`. Fix: expand `operatorOwnedResourceNames()` (webhook Service, its serving-cert Secret, config ConfigMap) and apply the by-name skip inside `deleteTyped` so Service/ConfigMap/Secret (and NetworkPolicy/Role/RoleBinding) all preserve operator-owned resources.
- **Stale inter-BFF params were never cleared** (`module_deploy.go`). `params.env` persists across reconciles, so once a dependency was disabled its `BFF_*_SERVICE_NAME`/`PORT` keys lingered in the module's rendered `<slug>-params` ConfigMap. `addInterBFFParams` now clears each dependency's keys before (re)writing them. Note this corrects the rendered ConfigMap only — gen-ai, the sole current consumer, hardcodes the maas coordinates in its Deployment, so the keys are inert for its running pods today; the fix keeps the ConfigMap correct for whenever a module starts consuming it.

## Test Impact

Primarily a test PR — it adds integration and unit tests for the dashboard-operator, plus the two small production fixes described above that the new tests surfaced. No manual/cluster testing was performed; verification was via the local `make` targets above. CI picks the new tagged tests up automatically through the existing "Integration tests (envtest)" step in `.github/workflows/dashboard-operator-tests.yml` — no workflow changes required.

## Request review criteria:

- [ ] The developer has manually tested the changes and verified that the changes work — *not applicable: hermetic envtest run locally, no cluster testing; see How Has This Been Tested.*
- [ ] The code follows our [Best Practices](/docs/best-practices.md) — *N/A: Go operator code; follows the `dashboard-operator` conventions in `.claude/rules/envtest-integration-tests.md` and `operator-controller.md`; `make lint` is clean.*

[RHOAIENG-83640]: https://redhat.atlassian.net/browse/RHOAIENG-83640?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[RHOAIENG-83645]: https://redhat.atlassian.net/browse/RHOAIENG-83645?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

- **Bug Fixes**
  - Removed outdated inter-module service parameters when dependent modules are disabled, preventing stale configuration after updates.

- **Tests**
  - Added integration coverage for module deployments, services, federation configuration, status reporting, and configuration-hash updates.
  - Verified observability behavior when disabled, unavailable, or successfully deployed.
  - Added checks for management-state removal, idempotent teardown, and cleanup of legacy resources.
  - Added validation for required observability configuration and inter-module service parameters.

### DIFF
--- a/dashboard-operator/internal/controller/dashboard_reconciler.go
+++ b/dashboard-operator/internal/controller/dashboard_reconciler.go
@@ -40,22 +40,61 @@ const conditionMaasConsumerPortalAvailable = "MaasConsumerPortalAvailable"
 var operatorDeploymentName = getOperatorDeploymentName()
 
 func getOperatorDeploymentName() string {
-	if name := os.Getenv("OPERATOR_DEPLOYMENT_NAME"); name != "" {
-		return name
-	}
-	return "dashboard-operator"
+	return envOrDefault("OPERATOR_DEPLOYMENT_NAME", "dashboard-operator")
 }
 
-func operatorOwnedResourceNames() map[string]bool {
+// operatorServiceAccountName and resolveDistributionConfigMapName resolve the names the operator's
+// Helm chart rendered for its own ServiceAccount and config ConfigMap. Both derive from user-settable
+// chart values (serviceAccount.name, config.name), so the chart plumbs the rendered names in via
+// env — the Go side must not re-derive them by convention, or a non-default install would delete
+// the operator's own resources during managementState: Removed teardown. Defaults match the chart.
+func operatorServiceAccountName() string {
+	// Defaults to the deployment name: the chart's default serviceAccount.name renders to exactly
+	// that, and unit tests that override the deployment name expect the SA to track it.
+	return envOrDefault("OPERATOR_SERVICE_ACCOUNT_NAME", operatorDeploymentName)
+}
+
+// resolveDistributionConfigMapName returns the name of the chart-rendered config ConfigMap
+// (dashboard.configMapName / .Values.config.name, default odh-dashboard-config). This is the
+// ConfigMap the operator reads for distribution identity and which teardown must preserve. Its
+// name is user-settable, so the chart plumbs it in via OPERATOR_CONFIGMAP_NAME.
+func resolveDistributionConfigMapName() string {
+	return envOrDefault("OPERATOR_CONFIGMAP_NAME", distributionConfigMapName)
+}
+
+func envOrDefault(key, fallback string) string {
+	if v := os.Getenv(key); v != "" {
+		return v
+	}
+	return fallback
+}
+
+// operatorOwnedResources returns the set of the operator's own resources that teardown must never
+// delete, keyed by "<Kind>/<name>". Teardown selects resources by the part-of=dashboard label the
+// chart's commonLabels stamp on everything it renders — including operator-owned resources — so
+// these must be skipped by name. Keying on kind+name (not name alone) keeps a genuinely
+// module-owned resource that happens to share a name (e.g. a module ConfigMap named
+// odh-dashboard-config) from silently surviving teardown.
+//
+// The webhook serving-cert Secret is intentionally absent: cert-manager issues it from the chart's
+// Certificate (which has no secretTemplate), so it never carries the part-of label and is never
+// selected by teardown's label query — no skip entry is needed.
+func operatorOwnedResources() map[string]bool {
 	base := operatorDeploymentName
 	return map[string]bool{
-		base:                      true,
-		base + "-role":            true,
-		base + "-rolebinding":     true,
-		base + "-webhook":         true, // webhook Service (dashboard.webhookServiceName)
-		base + "-webhook-tls":     true, // webhook serving-cert Secret (dashboard.webhookCertSecretName)
-		distributionConfigMapName: true, // operator config ConfigMap (odh-dashboard-config)
+		"Deployment/" + base:                              true,
+		"ServiceAccount/" + operatorServiceAccountName():  true,
+		"Service/" + base + "-webhook":                    true, // dashboard.webhookServiceName
+		"ConfigMap/" + resolveDistributionConfigMapName(): true, // dashboard.configMapName
+		"ClusterRole/" + base + "-role":                   true,
+		"ClusterRoleBinding/" + base + "-rolebinding":     true,
 	}
+}
+
+// isOperatorOwned reports whether a resource of the given kind and name is one of the operator's
+// own resources that teardown must preserve.
+func isOperatorOwned(resources map[string]bool, kind, name string) bool {
+	return resources[kind+"/"+name]
 }
 
 var persesdashboardGVK = schema.GroupVersionKind{
@@ -754,9 +793,9 @@ func (r *DashboardReconciler) teardownManagedResources(ctx context.Context, dash
 	// The operator's own Helm chart stamps platform.opendatahub.io/part-of=dashboard
 	// (commonLabels) onto every resource it renders — including the webhook Service and
 	// the operator ConfigMap, both of which land in the applications namespace. Teardown
-	// must skip these by name so it never deletes operator-owned resources (e.g. deleting
+	// must skip these by kind+name so it never deletes operator-owned resources (e.g. deleting
 	// the webhook Service breaks the failurePolicy: Fail ValidatingWebhookConfiguration).
-	operatorResources := operatorOwnedResourceNames()
+	operatorResources := operatorOwnedResources()
 
 	deleteTyped := func(list client.ObjectList, kind string, opts ...client.ListOption) error {
 		if err := r.List(ctx, list, opts...); err != nil {
@@ -765,7 +804,7 @@ func (r *DashboardReconciler) teardownManagedResources(ctx context.Context, dash
 
 		items := extractItems(list)
 		for i := range items {
-			if operatorResources[items[i].GetName()] {
+			if isOperatorOwned(operatorResources, kind, items[i].GetName()) {
 				continue
 			}
 			logger.Info("Deleting managed resource", "kind", kind, "name", items[i].GetName())
@@ -783,7 +822,7 @@ func (r *DashboardReconciler) teardownManagedResources(ctx context.Context, dash
 	}
 	for i := range deployments.Items {
 		dep := &deployments.Items[i]
-		if dep.Name == operatorDeploymentName {
+		if isOperatorOwned(operatorResources, "Deployment", dep.Name) {
 			continue
 		}
 		logger.Info("Deleting managed resource", "kind", "Deployment", "name", dep.Name)
@@ -808,7 +847,7 @@ func (r *DashboardReconciler) teardownManagedResources(ctx context.Context, dash
 	}
 	for i := range serviceAccounts.Items {
 		sa := &serviceAccounts.Items[i]
-		if operatorResources[sa.Name] {
+		if isOperatorOwned(operatorResources, "ServiceAccount", sa.Name) {
 			continue
 		}
 		logger.Info("Deleting managed resource", "kind", "ServiceAccount", "name", sa.Name)
@@ -843,7 +882,7 @@ func (r *DashboardReconciler) teardownManagedResources(ctx context.Context, dash
 	}
 	for i := range clusterRoles.Items {
 		cr := &clusterRoles.Items[i]
-		if operatorResources[cr.Name] {
+		if isOperatorOwned(operatorResources, "ClusterRole", cr.Name) {
 			continue
 		}
 		logger.Info("Deleting managed resource", "kind", "ClusterRole", "name", cr.Name)
@@ -858,7 +897,7 @@ func (r *DashboardReconciler) teardownManagedResources(ctx context.Context, dash
 	}
 	for i := range clusterRoleBindings.Items {
 		crb := &clusterRoleBindings.Items[i]
-		if operatorResources[crb.Name] {
+		if isOperatorOwned(operatorResources, "ClusterRoleBinding", crb.Name) {
 			continue
 		}
 		logger.Info("Deleting managed resource", "kind", "ClusterRoleBinding", "name", crb.Name)
--- a/dashboard-operator/internal/controller/dashboard_reconciler.go
+++ b/dashboard-operator/internal/controller/dashboard_reconciler.go
@@ -49,9 +49,12 @@ func getOperatorDeploymentName() string {
 func operatorOwnedResourceNames() map[string]bool {
 	base := operatorDeploymentName
 	return map[string]bool{
-		base:                  true,
-		base + "-role":        true,
-		base + "-rolebinding": true,
+		base:                      true,
+		base + "-role":            true,
+		base + "-rolebinding":     true,
+		base + "-webhook":         true, // webhook Service (dashboard.webhookServiceName)
+		base + "-webhook-tls":     true, // webhook serving-cert Secret (dashboard.webhookCertSecretName)
+		distributionConfigMapName: true, // operator config ConfigMap (odh-dashboard-config)
 	}
 }
 
@@ -748,6 +751,13 @@ func (r *DashboardReconciler) teardownManagedResources(ctx context.Context, dash
 	}
 	inNamespace := client.InNamespace(r.ApplicationsNamespace)
 
+	// The operator's own Helm chart stamps platform.opendatahub.io/part-of=dashboard
+	// (commonLabels) onto every resource it renders — including the webhook Service and
+	// the operator ConfigMap, both of which land in the applications namespace. Teardown
+	// must skip these by name so it never deletes operator-owned resources (e.g. deleting
+	// the webhook Service breaks the failurePolicy: Fail ValidatingWebhookConfiguration).
+	operatorResources := operatorOwnedResourceNames()
+
 	deleteTyped := func(list client.ObjectList, kind string, opts ...client.ListOption) error {
 		if err := r.List(ctx, list, opts...); err != nil {
 			return fmt.Errorf("listing %s: %w", kind, err)
@@ -755,6 +765,9 @@ func (r *DashboardReconciler) teardownManagedResources(ctx context.Context, dash
 
 		items := extractItems(list)
 		for i := range items {
+			if operatorResources[items[i].GetName()] {
+				continue
+			}
 			logger.Info("Deleting managed resource", "kind", kind, "name", items[i].GetName())
 			if err := r.Delete(ctx, items[i]); client.IgnoreNotFound(err) != nil {
 				return fmt.Errorf("deleting %s %s: %w", kind, items[i].GetName(), err)
@@ -788,8 +801,6 @@ func (r *DashboardReconciler) teardownManagedResources(ctx context.Context, dash
 	if err := deleteTyped(&configmaps, "ConfigMap", matchLabels, inNamespace); err != nil {
 		return err
 	}
-
-	operatorResources := operatorOwnedResourceNames()
 
 	var serviceAccounts corev1.ServiceAccountList
 	if err := r.List(ctx, &serviceAccounts, matchLabels, inNamespace); err != nil {
--- a/dashboard-operator/internal/controller/federation_integration_test.go
+++ b/dashboard-operator/internal/controller/federation_integration_test.go
@@ -1,0 +1,195 @@
+//go:build integration
+
+package controller_test
+
+import (
+	"context"
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+
+	"github.com/opendatahub-io/odh-platform-utilities/pkg/cluster"
+	"github.com/opendatahub-io/odh-platform-utilities/pkg/metadata/labels"
+
+	v1alpha1 "github.com/opendatahub-io/odh-dashboard/dashboard-operator/api/v1alpha1"
+	ctrlpkg "github.com/opendatahub-io/odh-dashboard/dashboard-operator/internal/controller"
+)
+
+// allModuleSlugs lists every module's ManifestSlug — used to build a full manifest
+// tree for the all-modules-enabled federation test.
+var allModuleSlugs = []string{
+	"model-registry", "gen-ai", "mlflow", "maas", "eval-hub",
+	"automl", "autorag", "agent-ops", "notebooks",
+}
+
+// TestIntegration_AddInterBFFParams verifies the reconciler injects service-discovery
+// env vars into a module's params.env for its inter-BFF dependencies: gen-ai depends
+// on maas, so gen-ai's params.env must gain BFF_MAAS_SERVICE_NAME/PORT while maas —
+// which has no such dependency — must not. (RHOAIENG-83649)
+func TestIntegration_AddInterBFFParams(t *testing.T) {
+	manifests := createIntegrationManifests(t, []string{"gen-ai", "maas"})
+	r := newManifestReconciler(manifests)
+
+	dashboard := newDashboard(v1alpha1.DashboardSpec{
+		Gateway: &v1alpha1.GatewaySpec{Domain: "test.example.com"},
+		Modules: disableAllModulesExcept("genAi", "maas"),
+	})
+
+	ctx := context.Background()
+	require.NoError(t, k8sClient.Create(ctx, dashboard))
+
+	t.Cleanup(func() {
+		deleteDashboard(t)
+		cleanupModuleResources(t)
+	})
+
+	reconcile(t, r)
+	reconcile(t, r)
+
+	// gen-ai depends on maas — its params.env gets the maas service coordinates.
+	genAiParams := readParamsEnv(t, filepath.Join(manifests, "modules", "gen-ai", "params.env"))
+	assert.Equal(t, "odh-dashboard-maas-ui", genAiParams["BFF_MAAS_SERVICE_NAME"])
+	assert.Equal(t, "8243", genAiParams["BFF_MAAS_SERVICE_PORT"])
+
+	// maas has no inter-BFF dependency — no BFF_MAAS_* keys should be written.
+	maasParams := readParamsEnv(t, filepath.Join(manifests, "modules", "maas", "params.env"))
+	assert.NotContains(t, maasParams, "BFF_MAAS_SERVICE_NAME")
+	assert.NotContains(t, maasParams, "BFF_MAAS_SERVICE_PORT")
+}
+
+// TestIntegration_FederationConfigMap_AllModulesEnabled verifies that with all
+// modules enabled (DSC gates satisfied), the federation ConfigMap lists every
+// module plus the synthetic coreBff and mlflowEmbedded entries, and omits perses
+// when observability is off. (RHOAIENG-83649)
+func TestIntegration_FederationConfigMap_AllModulesEnabled(t *testing.T) {
+	manifests := createIntegrationManifests(t, allModuleSlugs)
+	r := newManifestReconciler(manifests)
+
+	// Satisfy every module's RequiredDSCComponents so none are gated off.
+	dashboard := newDashboard(v1alpha1.DashboardSpec{
+		Gateway: &v1alpha1.GatewaySpec{Domain: "test.example.com"},
+		Components: map[string]v1alpha1.ComponentAvailability{
+			"modelregistry":  {ManagementState: "Managed"},
+			"mlflowoperator": {ManagementState: "Managed"},
+			"trustyai":       {ManagementState: "Managed"},
+			"aipipelines":    {ManagementState: "Managed"},
+		},
+	})
+
+	ctx := context.Background()
+	require.NoError(t, k8sClient.Create(ctx, dashboard))
+
+	t.Cleanup(func() {
+		deleteDashboard(t)
+		cleanupModuleResources(t)
+	})
+
+	reconcile(t, r)
+	reconcile(t, r)
+
+	entries := parseFederationEntries(t, getFederationConfigMap(t))
+
+	// Every registered module has a federation entry.
+	for _, name := range ctrlpkg.ModuleNames() {
+		assert.NotNil(t, findFederationEntry(entries, name), "module %s should be in federation config", name)
+	}
+
+	// Synthetic entries that always/conditionally accompany the modules.
+	assert.NotNil(t, findFederationEntry(entries, "coreBff"), "coreBff entry should always be present")
+	assert.NotNil(t, findFederationEntry(entries, "mlflowEmbedded"), "mlflowEmbedded should be present when mlflow is deployed")
+
+	// Observability is off, so there must be no perses entry.
+	assert.Nil(t, findFederationEntry(entries, "perses"), "perses should be absent when observability is disabled")
+}
+
+// TestIntegration_FederationConfigMap_HashAnnotation verifies the reconciler stamps
+// the main dashboard Deployment's pod template with a federation-config hash and that
+// the hash changes when the set of enabled modules changes — the mechanism that
+// triggers a rolling restart when federation config drifts. (RHOAIENG-83649)
+func TestIntegration_FederationConfigMap_HashAnnotation(t *testing.T) {
+	manifests := createIntegrationManifests(t, []string{"model-registry"})
+	r := newManifestReconciler(manifests)
+
+	ctx := context.Background()
+
+	// The main dashboard Deployment is not part of the minimal core manifests, so
+	// pre-create it (labeled part-of=dashboard so cleanupModuleResources removes it).
+	deployName := ctrlpkg.MainDashboardDeploymentName(cluster.OpenDataHub)
+	mainDeploy := newMainDashboardDeployment(deployName)
+	require.NoError(t, k8sClient.Create(ctx, mainDeploy))
+
+	dashboard := newDashboard(v1alpha1.DashboardSpec{
+		Gateway: &v1alpha1.GatewaySpec{Domain: "test.example.com"},
+		Modules: disableAllModulesExcept("modelRegistry"),
+	})
+	require.NoError(t, k8sClient.Create(ctx, dashboard))
+
+	t.Cleanup(func() {
+		deleteDashboard(t)
+		cleanupModuleResources(t)
+	})
+
+	reconcile(t, r)
+	reconcile(t, r)
+
+	hash1 := federationHashOf(t, deployName)
+	require.NotEmpty(t, hash1, "federation hash annotation should be set after reconcile")
+
+	// Disable all modules — the federation config (and thus its hash) must change.
+	dashboard = getDashboard(t)
+	dashboard.Spec.Modules = disableAllModulesExcept()
+	require.NoError(t, k8sClient.Update(ctx, dashboard))
+
+	reconcile(t, r)
+
+	hash2 := federationHashOf(t, deployName)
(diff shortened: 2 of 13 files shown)


dashboard-operator/internal/controller/federation_integration_test.go
```diff
@@ -0,0 +1,195 @@
...
+	// gen-ai depends on maas — its params.env gets the maas service coordinates.
+	genAiParams := readParamsEnv(t, filepath.Join(manifests, "modules", "gen-ai", "params.env"))
+	assert.Equal(t, "odh-dashboard-maas-ui", genAiParams["BFF_MAAS_SERVICE_NAME"])
+	assert.Equal(t, "8243", genAiParams["BFF_MAAS_SERVICE_PORT"])
+
+	// maas has no inter-BFF dependency — no BFF_MAAS_* keys should be written.
+	maasParams := readParamsEnv(t, filepath.Join(manifests, "modules", "maas", "params.env"))
+	assert.NotContains(t, maasParams, "BFF_MAAS_SERVICE_NAME")
+	assert.NotContains(t, maasParams, "BFF_MAAS_SERVICE_PORT")
+}
```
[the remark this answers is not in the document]
lucferbux: Good catch — addressed in 3af8d49026.

While adding the removal case I found this exposed a latent bug, not just a test gap: `addInterBFFParams` never deleted previously-written keys, and because gen-ai has only a *soft* inter-BFF dependency on maas (no hard `InterModuleDependencies`), disabling maas leaves gen-ai deployed and its `params.env` is rewritten each reconcile — so the stale `BFF_MAAS_SERVICE_NAME`/`BFF_MAAS_SERVICE_PORT` coordinates would persist indefinitely.

Fix has two parts:
- **Production** (`module_deploy.go`): `addInterBFFParams` now `delete`s each dependency's env keys before conditionally re-adding them, so a dependency that is no longer Deployed/Degraded leaves no dangling coordinates.
- **Test** (`federation_integration_test.go`): `TestIntegration_AddInterBFFParams` now covers AC1's removal case — disable maas, reconcile, assert both `BFF_MAAS_*` keys are gone from gen-ai's `params.env`.

Verified via `make test-integration` (all green).

dashboard-operator/internal/controller/managementstate_integration_test.go
```diff
@@ -0,0 +1,181 @@
...
+	require.NoError(t, k8sClient.Update(ctx, dashboard))
+
+	reconcile(t, r)
+
+	// Module operands are gone.
+	assert.Empty(t, listDeployments(t, "model-registry"), "module Deployment should be removed")
+	assert.Empty(t, listServices(t, "model-registry"), "module Service should be removed")
+
+	// Operator-owned resources are preserved (skipped by name during teardown).
+	assertExists(t, &appsv1.Deployment{}, types.NamespacedName{Name: "dashboard-operator", Namespace: integrationNamespace})
```
liday-rh, line 76: **HIGH — this assertion block understates what teardown actually deletes, and there is a real gap behind it.**

These four `assertExists` calls cover exactly the four kinds `teardownManagedResources` skips by name: Deployment, ServiceAccount, ClusterRole, ClusterRoleBinding. But `Service`, `ConfigMap` and `Secret` go through `deleteTyped(...)` with **no name skip at all** — every object labeled `platform.opendatahub.io/part-of=dashboard` in the applications namespace is deleted unconditionally.

That matters because the operator's own Helm chart stamps `commonLabels` — which include `platform.opendatahub.io/part-of: dashboard` (`charts/dashboard/values.yaml:17`) — onto *every* resource it renders, including the webhook `Service` (`odh-dashboard-operator-webhook`) and the operator `ConfigMap` (`odh-dashboard-config`). Both land in `metadata.namespace`, and `APPLICATIONS_NAMESPACE` defaults to that same namespace (`cmd/manager/main.go:73-75`, `config/manager/manager.yaml:34`), so they are squarely in teardown's blast radius.

**Reproduced in envtest.** I added a throwaway test alongside this one that seeds a `Service` named `odh-dashboard-operator-webhook` and a `ConfigMap` named `odh-dashboard-config` (both labeled `part-of=dashboard`, both in `integrationNamespace`), reconciles twice, then flips `managementState: Removed` and reconciles once more. Result:

```
webhook Service gone=true ; operator ConfigMap gone=true
```

**Concrete failure scenario.** The `ValidatingWebhookConfiguration` is *not* deleted — teardown does not handle that kind — and it has `failurePolicy: Fail` on `CREATE` of `dashboards` (`charts/dashboard/templates/webhook.yaml:43`). So after a single `managementState: Removed`, the webhook points at a Service that no longer exists and every attempt to create a Dashboard CR is rejected by admission until the chart is re-applied. Neither object is recreated by this operator (they are chart-owned), so recovery needs the ODH operator to re-render the chart — which in turn means the two operators fight, with this one deleting the Service on every Removed reconcile and the other re-applying it. The operator `ConfigMap` disappearing likewise silently reverts log level / distribution / reconcile-interval to defaults.

**Suggested fix**, either:
1. Extend the name-skip to Services/ConfigMaps/Secrets — i.e. run them through the same `operatorResources[...]` check already used for ServiceAccounts and ClusterRoles rather than bare `deleteTyped`; or
2. Scope teardown to operands only, e.g. require `app.kubernetes.io/component` in addition to `part-of=dashboard`, which is already how `deleteModuleResources` distinguishes module operands from everything else.

Either way, `seedOperatorOwnedResources` should grow a Service and a ConfigMap so this test actually pins the guarantee its comment claims ("Operator-owned resources are preserved"). As written the test passes while the gap is wide open, which is the part worth fixing before this merges — a green test here is going to be read as "teardown is safe".

dashboard-operator/internal/controller/module_deploy.go
```diff
@@ -345,6 +345,12 @@ func addInterBFFParams(params map[string]string, moduleName string, statuses map
 		return
 	}
 	for _, dep := range mod.InterBFFDeps {
+		// Clear any previously written coordinates first. params.env persists across
```
liday-rh, line 363: **MEDIUM — this fix is inert in production; the test that guards it asserts on the wrong artifact.**

The stale-coordinate removal is correct in isolation (I mutation-tested it: dropping these two `delete` calls fails `TestIntegration_AddInterBFFParams`). The problem is that nothing downstream consumes the `params.env` keys being written and deleted, so the user-visible bug described in the PR body is not actually fixed.

**Evidence — rendered the real module kustomization.** I appended a sentinel to `manifests/modules/gen-ai/params.env`:

```
BFF_MAAS_SERVICE_NAME=ZZZ-PROOF
```

and ran `kustomize build manifests/modules/gen-ai`. The sentinel shows up in exactly one place — the generated ConfigMap:

```yaml
data:
  BFF_MAAS_SERVICE_NAME: ZZZ-PROOF   # gen-ai-params ConfigMap
```

while the container env is untouched:

```yaml
- name: BFF_MAAS_SERVICE_NAME
  value: odh-dashboard-maas-ui       # hardcoded literal
```

That value comes from the hardcoded literals at `manifests/modules/gen-ai/deployment.yaml:157-160`. The kustomization has no `replacements` entry for these keys (its three `replacements` cover only `gen-ai-ui-image`, `pgvector-image`, `gateway-domain`) and the Deployment has no `envFrom`/`configMapRef` pulling in `gen-ai-params`. A repo-wide grep for `BFF_MAAS_SERVICE_NAME` confirms nothing else reads the params key.

**So the concrete scenario the PR claims to fix still happens:** enable gen-ai + maas, then disable maas. `params.env` is correctly cleaned up by this change, but the running gen-ai pod still has `BFF_MAAS_SERVICE_NAME=odh-dashboard-maas-ui` in its env, pointing at a Service that no longer exists. `TestIntegration_AddInterBFFParams` asserts against the operator's scratch `params.env` file rather than the applied Deployment, so it goes green while the defect persists.

Worth noting the ConfigMap write is a no-op operationally too: `generatorOptions.disableNameSuffixHash: true` means changing `gen-ai-params` content produces no name change and therefore no pod rollout, so even if something *did* consume it via `envFrom`, the value would not reach a running pod without a separate restart trigger.

**Suggested fix**, either:
1. Wire the params through for real — add kustomize `replacements` from `gen-ai-params` into the container env (matching how `gateway-domain` is already handled), or switch the Deployment to `envFrom` the generated ConfigMap and re-enable the name-suffix hash so content changes roll pods — and then retarget the test to assert on the rendered/applied container env rather than `params.env`; or
2. If inter-BFF service discovery is genuinely meant to be static, drop `addInterBFFParams` and `InterBFFDeps` as dead code rather than growing test coverage around a mechanism that has no effect.

Option 1 is the one that matches the PR's stated intent.

liday-rh (commented):
## Review: verified by running the suite

I checked this out and actually ran it rather than reading only:

- `make test-integration` (envtest 1.31, `-race`, `-count=1`): green — all 244 assertions, all 16 `TestIntegration_*` (11 new + 4 pre-existing + the new unit test) pass.
- Mutation-tested the production change: deleting the two `delete(params, ...)` lines in `addInterBFFParams` makes `TestIntegration_AddInterBFFParams` fail at `federation_integration_test.go:75`, so the new test genuinely covers the fix.
- Investigated the RESTMapper-ordering concern flagged in the PR description (the `_Deployed` / `_PersesCRDNotFound` isolation). Ran the full suite 4x with `-shuffle=on` and the observability group 6x, including runs where `_Deployed` executed *before* `_PersesCRDNotFound`. It did not reproduce — the isolated-client approach holds. Dropping that as a concern; the ordering comment in the source is accurate but the scheme is more robust than the note implies.

Two findings follow as inline comments — one HIGH on the teardown coverage gap (reproduced in envtest), one MEDIUM on the inter-BFF params fix being inert in production. Neither blocks; filing as a comment.

### Correction to the PR description

The PR body states "This PR is test-only ... changes no production code", and the self-checklist marks manual/cluster testing as *not applicable* on that basis. That is not accurate: `dashboard-operator/internal/controller/module_deploy.go` changes behaviour on the reconcile path (`addInterBFFParams` now mutates `params.env` content that gets rendered into a deployed ConfigMap). It is a small change and low-risk, but the "test-only" framing is what justified skipping cluster verification, so it is worth restating.

### Checked and cleared

- `overlayStandaloneReadiness` runs *after* `deployModuleManifests`, so the spec-derived statuses handed to `addInterBFFParams` are the right ones — the single trailing reconcile in `TestIntegration_AddInterBFFParams` is sufficient, not a race.
- `TestDeployObservabilityManifests_PersesServiceRequired` compiles against the imports already present in `actions_test.go` and hits the `PersesService == nil` guard before touching the fake client, so the empty scheme is fine.
- Teardown name-skips (`operatorOwnedResourceNames()`) match the seeded `dashboard-operator` / `-role` / `-rolebinding` names.
- `cleanupCrossNamespaceResources` early-returns because `obsNS == r.ApplicationsNamespace` in the test setup, so it never lists PersesDashboards on the shared client — the shared RESTMapper stays clean as intended.
- `allModuleSlugs` matches the 9-entry `moduleRegistry`; `t.Cleanup` LIFO ordering is correct in all the new tests; every cluster-scoped seed is cleaned up.

dashboard-operator/internal/controller/managementstate_integration_test.go
lucferbux: Confirmed — this is a real HIGH teardown bug, thanks for the detailed catch. `deleteTyped` deleted `Service`/`ConfigMap`/`Secret` unconditionally, so on `managementState: Removed` the teardown would delete the webhook `Service` (`odh-dashboard-operator-webhook`) — breaking the `failurePolicy: Fail` `ValidatingWebhookConfiguration` — and the `odh-dashboard-config` ConfigMap the operator reads.

Fixed in `5f072e4d14`:
- Expanded `operatorOwnedResourceNames()` with the webhook `Service` (`<base>-webhook`), its serving-cert `Secret` (`<base>-webhook-tls`), and the config `ConfigMap` (`odh-dashboard-config`).
- Hoisted `operatorResources := operatorOwnedResourceNames()` above the `deleteTyped` closure and added the by-name skip inside it, so `Service`/`ConfigMap`/`Secret` (and `NetworkPolicy`/`Role`/`RoleBinding`, which also route through `deleteTyped`) now all preserve operator-owned resources.
- Strengthened `seedOperatorOwnedResources` + both teardown tests to seed and assert survival of the webhook `Service`, cert `Secret`, and config `ConfigMap`.

dashboard-operator/internal/controller/module_deploy.go
lucferbux: Agreed on both points — thanks for mutation-testing it and rendering the real kustomization.

You're right that the `delete` calls maintain the rendered `<slug>-params` ConfigMap only, and that gen-ai (the sole current consumer) hardcodes the maas coordinates in its Deployment env, so the keys are inert for its running pods today. The stale-clear behavior is still the correct thing for the operator to do to that ConfigMap, and it's in scope for RHOAIENG-83649, so I've kept the fix + test but corrected the over-claim:

- **PR body:** replaced the "user-visible bug" framing with a *Production fixes* section that scopes this to the rendered ConfigMap and explicitly notes it is inert for gen-ai's pods today.
- **Test doc comment** (`TestIntegration_AddInterBFFParams`): now states it asserts on the operator's rendered `params.env` (the source the ConfigMap is generated from), not a running pod's env, and that whether a module consumes the keys is up to its kustomization.
- **Code doc comment** on `addInterBFFParams`: documents that it only maintains the `<slug>-params` keys, that a module must wire them via `envFrom`/replacement, and that they're inert for gen-ai until it stops hardcoding them.

All in `5f072e4d14`.

dashboard-operator/internal/controller/dashboard_reconciler.go
```diff
@@ -48,9 +48,12 @@ func getOperatorDeploymentName() string {
...
 	return map[string]bool{
-		base:                  true,
-		base + "-role":        true,
-		base + "-rolebinding": true,
+		base:                      true,
+		base + "-role":            true,
+		base + "-rolebinding":     true,
+		base + "-webhook":         true, // webhook Service (dashboard.webhookServiceName)
+		base + "-webhook-tls":     true, // webhook serving-cert Secret (dashboard.webhookCertSecretName)
+		distributionConfigMapName: true, // operator config ConfigMap (odh-dashboard-config)
```
liday-rh: **MEDIUM — two of these names are re-derived by convention rather than injected, so a non-default install still hits the bug this PR fixes.**

`distributionConfigMapName` is hardcoded and the ServiceAccount is absent from the map entirely. But the chart derives both from user-settable values:

- `dashboard.configMapName` = `.Values.config.name`
- `dashboard.serviceAccountName` = `namePrefix + (.Values.serviceAccount.name | default "dashboard-operator")`

The ServiceAccount survives today only by coincidence — with the default `serviceAccount.name: dashboard-operator` it happens to equal `base`, so the bare `base: true` entry covers it. Override `serviceAccount.name` or `config.name` and `managementState: Removed` deletes the operator's own ServiceAccount / ConfigMap: exactly the failure mode this PR is closing.

Suggest plumbing both in via env the way `OPERATOR_DEPLOYMENT_NAME` already is, so the Go side never has to guess what Helm rendered.

_Secondary:_ the map is keyed on name only and is consulted for **every** kind in `deleteTyped`, so a genuinely module-owned Service/Secret/Role that happened to be named `odh-dashboard-config` would also silently survive teardown. Keying on kind+name would be tighter.

dashboard-operator/internal/controller/dashboard_reconciler.go
```diff
@@ -48,9 +48,12 @@ func getOperatorDeploymentName() string {
...
 	base := operatorDeploymentName
 	return map[string]bool{
-		base:                  true,
-		base + "-role":        true,
-		base + "-rolebinding": true,
+		base:                      true,
+		base + "-role":            true,
+		base + "-rolebinding":     true,
+		base + "-webhook":         true, // webhook Service (dashboard.webhookServiceName)
+		base + "-webhook-tls":     true, // webhook serving-cert Secret (dashboard.webhookCertSecretName)
```
liday-rh: **LOW — this entry is dead code.**

The webhook serving-cert Secret is created by cert-manager from the chart's `Certificate` (`charts/dashboard/templates/webhook.yaml:68-88`), which has no `secretTemplate`. cert-manager does not propagate the Certificate's labels to the Secret it issues, so this Secret never carries `platform.opendatahub.io/part-of: dashboard` and is therefore never returned by teardown's label selector. And with `webhook.certManager.enabled=false` the Secret isn't created at all.

Harmless as a belt-and-braces entry, but worth a comment saying so — otherwise it reads as load-bearing protection that isn't.

dashboard-operator/internal/controller/dashboard_reconciler.go
```diff
@@ -677,13 +680,23 @@ func (r *DashboardReconciler) teardownManagedResources(ctx context.Context, dash
...
+	operatorResources := operatorOwnedResourceNames()
+
 	deleteTyped := func(list client.ObjectList, kind string, opts ...client.ListOption) error {
 		if err := r.List(ctx, list, opts...); err != nil {
 			return fmt.Errorf("listing %s: %w", kind, err)
 		}
 
 		items := extractItems(list)
 		for i := range items {
+			if operatorResources[items[i].GetName()] {
```
liday-rh: **LOW — the Deployment loop below (~L715) doesn't go through this map.**

Every kind routed through `deleteTyped` now honours `operatorResources`, but the Deployment loop still guards with its own `dep.Name == operatorDeploymentName`. Functionally equivalent today, since `base` is in the map — but it's a trap: the next name added to `operatorOwnedResourceNames()` will be honoured for Services/ConfigMaps/Secrets/RBAC and silently ignored for Deployments.

Same applies to the hand-rolled ServiceAccount / ClusterRole / ClusterRoleBinding loops, though those at least do consult `operatorResources`.

dashboard-operator/internal/controller/module_deploy.go
```diff
@@ -339,12 +339,25 @@ func (r *DashboardReconciler) deleteModuleResources(
...
 func addInterBFFParams(params map[string]string, moduleName string, statuses map[string]v1alpha1.ModuleStatus, platform cluster.Platform) {
 	mod := moduleRegistry[moduleName]
 	if mod.InterBFFDeps == nil {
 		return
 	}
 	for _, dep := range mod.InterBFFDeps {
+		// Clear any previously written coordinates first. params.env persists across
+		// reconciles, so a dependency that is no longer deployed must not leave stale
+		// service coordinates behind.
+		delete(params, dep.EnvServiceName)
```
liday-rh, line 366: **LOW/MEDIUM — the clear runs unconditionally, before the registry lookup and the phase check.**

That means these keys are stripped from `params.env` in *all* not-deployed cases, including when the module's own shipped `params.env` supplies a static default for them.

The pattern that breaks is the one your new doc comment points at: a module that wires these in via a kustomize `replacement` sourced from `data.BFF_MAAS_SERVICE_NAME` — exactly how gen-ai sources `data.gen-ai-ui-image` in `manifests/modules/gen-ai/kustomization.yaml` — would fail to render with a "field not found" error whenever maas is disabled. That converts a stale-value bug into a hard reconcile failure.

Safe today only because gen-ai ships no such default or replacement. Either delete only when the dep is genuinely being cleared, or document that these keys are operator-owned and must not be shipped in a module's `params.env`.

dashboard-operator/internal/controller/managementstate_integration_test.go
```diff
@@ -0,0 +1,198 @@
...
+	// Flip to Removed and reconcile once to run the teardown path.
+	dashboard = getDashboard(t)
+	dashboard.Spec.ManagementState = "Removed"
+	require.NoError(t, k8sClient.Update(ctx, dashboard))
+
+	reconcile(t, r)
+
+	// Module operands are gone.
+	assert.Empty(t, listDeployments(t, "model-registry"), "module Deployment should be removed")
+	assert.Empty(t, listServices(t, "model-registry"), "module Service should be removed")
```
liday-rh, line 68: **MEDIUM — the teardown test has no negative guard for the kinds the new skip touches.**

Deletion is asserted only for Deployments and Services (`model-registry`). Lines 65-71 then assert the operator-owned resources survive. Nothing pins that a *managed* ConfigMap or Secret is **still deleted** — which is precisely the direction the new name-based skip can regress.

As written, broadening the skip to a whole kind (or an over-eager name match) would pass CI green. `createIntegrationManifests` already renders a `dashboard-core-config` ConfigMap that could anchor the assertion; there's no `listConfigMaps` / `listSecrets` / `assertGone` helper yet, so this needs a small addition alongside the existing `listDeployments` / `listServices`.

dashboard-operator/internal/controller/managementstate_integration_test.go
```diff
@@ -0,0 +1,198 @@
...
+		&corev1.ServiceAccount{ObjectMeta: metav1.ObjectMeta{Name: "dashboard-operator", Namespace: integrationNamespace, Labels: partOf}},
+		// Webhook Service: deleting it would break the failurePolicy: Fail ValidatingWebhookConfiguration.
+		&corev1.Service{
+			ObjectMeta: metav1.ObjectMeta{Name: "dashboard-operator-webhook", Namespace: integrationNamespace, Labels: partOf},
+			Spec: corev1.ServiceSpec{
+				Ports: []corev1.ServicePort{{Port: 443}},
+			},
+		},
+		// Webhook serving-cert Secret.
+		&corev1.Secret{ObjectMeta: metav1.ObjectMeta{Name: "dashboard-operator-webhook-tls", Namespace: integrationNamespace, Labels: partOf}},
```
liday-rh: Related to my note on `dashboard_reconciler.go:55` — seeding this Secret with `partOf` by hand means the survival assertion at L68 can't detect that the `base + "-webhook-tls"` skip entry protects nothing in production (cert-manager never stamps that label on the issued Secret). The fixture manufactures the precondition the real cluster doesn't have, so this reads as coverage but is closer to false confidence.

dashboard-operator/internal/controller/observability_integration_test.go
```diff
@@ -0,0 +1,247 @@
...
+	reconcile(t, r)
+
+	cond := conditions.FindStatusCondition(getDashboard(t), conditionObservabilityAvailable)
+	require.NotNil(t, cond, "ObservabilityAvailable condition should be present")
+	assert.Equal(t, metav1.ConditionTrue, cond.Status)
+	assert.Equal(t, "Deployed", cond.Reason)
+
+	// The federation config should carry a perses proxy entry.
+	entries := parseFederationEntries(t, getFederationConfigMap(t))
+	assert.NotNil(t, findFederationEntry(entries, "perses"),
```
liday-rh: **LOW — this assertion can't fail for the reason its message states.**

`buildFederationConfigMap` (`module_deploy.go:425`) adds the perses entry purely from `spec.Observability.Enabled && PersesService != nil`, with no dependency on the deploy succeeding. The identical assertion would pass in the `_PersesCRDNotFound` case, so it adds no signal about "when observability is deployed".

The `perses-dashboard-config` check just below at L169 is the one that actually distinguishes the deployed path — this one could either be dropped or tightened to assert something the failed path wouldn't produce.

dashboard-operator/internal/controller/integration_test.go
```diff
@@ -351,6 +360,69 @@ func disableAllModulesExcept(enabled ...string) map[string]v1alpha1.ModuleOverri
 	return modules
 }
 
+// newReconciler builds a DashboardReconciler wired to the shared k8sClient.
```
liday-rh: Nit: comment says `newReconciler`, function is `newManifestReconciler`.

Worth flagging *why* CI didn't catch it: `make lint` is a bare `golangci-lint run` with no `--build-tags=integration`, so **none** of the new `*_integration_test.go` files in this PR are linted at all. The "make lint — 0 issues" line in the PR body doesn't cover the new test code. Running it tagged surfaces only 3 pre-existing `errcheck` hits on `testEnv.Stop()`, so adding the tag to the lint target looks cheap.

liday-rh (commented):
Re-reviewed at `5f072e4d1`. All three findings from my previous pass are addressed — thanks.

I traced the teardown fix end-to-end and it **holds in a real Helm deploy**: `deployment.yaml:88-89` injects `OPERATOR_DEPLOYMENT_NAME: {{ include "dashboard.deploymentName" . }}`, so `operatorDeploymentName` resolves to `odh-dashboard-operator` at runtime and every concatenated skip name matches its `_helpers.tpl` definition exactly (`-role` → `clusterRoleName`, `-rolebinding` → `clusterRoleBindingName`, `-webhook` → `webhookServiceName`, `-webhook-tls` → `webhookCertSecretName`). The skip is correctly placed before `Delete`, and hoisting `operatorResources` above the `deleteTyped` closure is sound.

Verification run: `go build ./...`, `go vet -tags integration`, `make lint`, and `make test-integration` all pass (15/15 `TestIntegration_*`). I also ran the suite under `-shuffle` with 7 seeds — including orders placing `_Deployed` before `_PersesCRDNotFound` — to probe the RESTMapper-isolation concern. The isolation holds; the tests are order-independent.

Nothing in the diff is a correctness regression. Inline findings below — **the two I'd want addressed before merge are the re-derived skip names (`dashboard_reconciler.go:56`) and the missing negative guard in the teardown test (`managementstate_integration_test.go:62`)**; the rest are follow-up material.

dashboard-operator/internal/controller/dashboard_reconciler.go
lucferbux: Fixed in c37e868b37. The two user-settable names are now injected rather than re-derived by convention: the operator Deployment passes `OPERATOR_SERVICE_ACCOUNT_NAME` (= `dashboard.serviceAccountName`) and `OPERATOR_CONFIGMAP_NAME` (= `dashboard.configMapName`) via `charts/dashboard/templates/deployment.yaml`, and the Go side reads them through `operatorServiceAccountName()` / `resolveDistributionConfigMapName()`. `operatorOwnedResources()` now includes both the ServiceAccount and the ConfigMap keyed by their resolved names, so a non-default `namePrefix` / `config.name` / `serviceAccount.name` install is protected too.

dashboard-operator/internal/controller/dashboard_reconciler.go
lucferbux: Fixed in c37e868b37 — dropped the `-webhook-tls` Secret entry from the owned set. You're right that cert-manager issues that Secret from the chart's `Certificate` (no `secretTemplate`), so it never carries `platform.opendatahub.io/part-of: dashboard` and is never selected by teardown; the entry was protecting nothing.

dashboard-operator/internal/controller/dashboard_reconciler.go
lucferbux: Fixed in c37e868b37 — the Deployment loop now goes through the shared map via `isOperatorOwned(operatorResources, "Deployment", dep.Name)`, same as the `deleteTyped` paths. The trap is gone: adding a name to the owned set now protects it consistently across every kind.

dashboard-operator/internal/controller/module_deploy.go
lucferbux: The unconditional clear is intentional and required for correctness, so I documented the contract (c37e868b37) rather than making it conditional. Making the delete conditional on the target being deployed would reintroduce exactly the round-1 bug: gen-ai has only a *soft* dependency on maas, so disabling maas leaves gen-ai deployed and its `params.env` rewritten each reconcile — a conditional delete would leave the stale `BFF_MAAS_*` keys behind.

So these keys must be operator-owned: cleared unconditionally, then re-set only when the target is deployed. The new doc comment makes that contract explicit — a consuming module MUST NOT ship static defaults for these keys and MUST consume them via `envFrom` on the generated `<slug>-params` ConfigMap, never via a build-time kustomize `replacement` (which fails rendering when the key is cleared). gen-ai, the only current consumer, hardcodes the maas coordinates in its Deployment env, so the keys are inert for it today; keeping the write/clear logic correct means the ConfigMap is right whenever a module starts consuming it via `envFrom`.

dashboard-operator/internal/controller/managementstate_integration_test.go
lucferbux: Fixed in c37e868b37 — added a negative-deletion guard. The test now creates the managed `dashboard-core-config` ConfigMap (which carries `part-of=dashboard`, like the operator-owned resources), asserts it exists before teardown, and asserts it is *gone* after (`assertGone`), while `odh-dashboard-config` survives. Broadening the name-based skip to swallow a managed ConfigMap would now fail the test.

dashboard-operator/internal/controller/managementstate_integration_test.go
lucferbux: Fixed in c37e868b37 — removed the hand-seeded `-webhook-tls` Secret from `seedOperatorOwnedResources` and dropped its survival assertion, along with the dead owned-set entry (your L55 note). The fixture no longer manufactures a precondition the real cluster doesn't have; a doc comment now records that cert-manager issues the Secret without the `part-of` label, so teardown never selects it and a survival check would only give false confidence.

dashboard-operator/internal/controller/observability_integration_test.go
lucferbux: Fixed in c37e868b37 — removed the perses federation-entry assertion from `_Deployed`. You're right it's driven purely by `spec.Observability.Enabled && PersesService != nil` with no dependency on the deploy succeeding, so it would pass identically in `_PersesCRDNotFound` and added no signal. The test now asserts on the `perses-dashboard-config` ConfigMap that the deploy path actually renders (with a comment explaining why that's the discriminating signal for this case).

dashboard-operator/internal/controller/integration_test.go
lucferbux: Both fixed in c37e868b37. The comment now reads `newManifestReconciler`. More importantly, the `lint` / `lint-fix` Makefile targets now run `golangci-lint run --build-tags=integration`, so the `*_integration_test.go` files are actually linted — that surfaced the 3 pre-existing issues you called out, which are fixed in the same commit. `make lint` now reports 0 issues *with* the integration tag, and the PR body has been corrected accordingly.

PR #9546, opened 2026-08-31
fix(cypress): use curl for AutoX pipeline readiness

jefho-rh:
## Description

The AutoX Cypress readiness helper used wget inside the ds-pipeline-api-server container to discover managed pipelines. On the current EA2 CI cluster, wget exits with a missing libpsl.so.5 dependency even though the KFP API and managed pipelines are healthy.

This changes the probe to use curl with insecure TLS, silent output, show errors, and fail-on-HTTP-error behavior. The current container image already provides curl.

## How Has This Been Tested?

- The equivalent curl probe was run inside the live EA2 ds-pipeline-api-server container and returned all four managed pipelines.
- The repository commit hook ran ESLint successfully.
- Full Cypress execution was not run locally; the next validation is the PR AutoML and AutoRAG CI suites on the EA2 cluster.

## Test Impact

No new test was added. This is a one-line command substitution in the shared waitForManagedPipelines helper used by the AutoML and AutoRAG E2E flows. The PR CI suites provide the relevant end-to-end coverage.

## Request review criteria:

- [ ] The developer has tested their solution on a cluster by using the image produced by the PR to main

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when checking managed pipeline availability during automated workflows.
  * Updated the internal network request command while preserving secure certificate handling and quiet output.

### DIFF
--- a/packages/cypress/cypress/utils/autoXPipelines.ts
+++ b/packages/cypress/cypress/utils/autoXPipelines.ts
@@ -95,7 +95,7 @@ export const provisionProjectForAutoX = (
  * (up to 5 min) to confirm pipelines exist before any BFF request is made.
  */
 export const waitForManagedPipelines = (projectName: string): void => {
-  const pipelineListCmd = `oc exec deploy/ds-pipeline-dspa -n ${projectName} -c ds-pipeline-api-server -- wget --no-check-certificate -qO- https://localhost:8888/apis/v2beta1/pipelines 2>/dev/null`;
+  const pipelineListCmd = `oc exec deploy/ds-pipeline-dspa -n ${projectName} -c ds-pipeline-api-server -- curl -ksSf https://localhost:8888/apis/v2beta1/pipelines 2>/dev/null`;
   const maxAttempts = 20;
   const intervalMs = 15000;
 


nickmazzi:
/lgtm
/approve

## AI assisted review
This is a very small, focused change. Here's my assessment:

**The change:** Replaces `wget --no-check-certificate -qO-` with `curl -ksSf` inside the `ds-pipeline-api-server` container for the AutoX pipeline readiness probe.

**No concerns.** The curl flags are correct and equivalent:

| wget flag | curl equivalent | meaning |
|---|---|---|
| `--no-check-certificate` | `-k` | skip TLS verification |
| `-q` | `-sS` | silent + show errors |
| `-O-` | *(default)* | output to stdout (curl does this by default) |
| *(none)* | `-f` | fail on HTTP errors (actually an improvement) |

The `-f` flag is a nice addition — `wget` would already exit non-zero on HTTP errors, but `curl` by default exits 0 even on 4xx/5xx. Adding `-f` preserves the retry-until-success behavior the `waitForManagedPipelines` loop depends on.

The `2>/dev/null` redirect stays, which suppresses stderr from curl's progress/error output — consistent with the silent intent.

Clean, well-scoped fix. Ship it.

PR #9467, opened 2026-08-25
test(e2e): add Kueue model deployment lifecycle Cypress test

kanishka-commits:
https://issues.redhat.com/browse/RHOAIENG-85482

## Description

Adds Cypress E2E for Kueue **model deployment status** ([RHOAIENG-85482](https://issues.redhat.com/browse/RHOAIENG-85482)) — **Inadmissible** and **Queued** only.

- **Inadmissible:** zero-quota ClusterQueue → deploy → assert status, queue position, status modal Resources tab
- **Queued:** 3 CPU / 9Gi quota → first deploy admits → second deploy → assert Queued + modal

Setup: Kueue project + model-serving HP (`scheduling.type: Queue`), wizard deploy (`hf://facebook/opt-125m`, vLLM CPU). Skipped when `IS_NON_ADMIN_RUN` is set.

## How Has This Been Tested?

- `eslint` on changed Cypress files
- Live run on **dash-e2e-rhoai** as `ldap-admin1` — both tests passed

```bash
export CY_TEST_CONFIG=/path/to/test-variables.yml
oc login -u ldap-admin1 --server=<OCP_API_URL>
cd frontend
npm run cypress:run -- --spec ../packages/cypress/cypress/tests/e2e/dataScienceProjects/models/testModelDeploymentKueueLifecycle.cy.ts
```

## Test Impact
Screenshot 2026-08-27 at 3 08 23 PM

New spec: `testModelDeploymentKueueLifecycle.cy.ts`  
Tags: `@Kueue`, `@Dashboard`, `@ModelServing`, `@Featureflagged`, `@NonConcurrent`

## Request review criteria

- [x] Manually tested on cluster
- [x] Testing instructions added
- [x] Tests added
- [x] Follows [Best Practices](/docs/best-practices.md)
- [ ] Tested with PR image before merge
CJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring hardware profiles for Kueue-managed model serving.
  * Added end-to-end coverage for model deployments that are blocked by unavailable quota or queued until resources become available.
  * Deployment status checks now include status subtitles and support selecting enabled or disabled hardware profiles.

* **Tests**
  * Added automated validation for workload admission, queue status, resource quotas, and deployment lifecycle behavior.

[...]

PR #9554, opened 2026-09-01
fix(cypress): remove profileName fallback in model serving wizard

kanishka-commits:
https://issues.redhat.com/browse/RHOAIENG-90165

## Description
Follow-up to #9467. 

**Why this approach?**

The optional `profileName` parameter was introduced only for the Kueue test in PR #9467. All other callers already use the helper with a single argument, which represents the UI-visible profile name.

Kueue is different because its Kubernetes resource name and display name differ. Since the helper interacts with the **UI dropdown**, it should use the display name rather than the Kubernetes resource name. The `profileName || profileDisplayName` fallback was only added to handle this difference and caused the disabled Kueue case to fail.

We therefore:

* Keep the existing one-argument API.
* Use `findByRole` with the UI-visible display name.
* Use `includes()` for options with additional text.
* Remove the `profileName || profileDisplayName` fallback.
* Apply the same behavior across all helper copies.

Kubernetes resource names can still be used for Kubernetes operations; the page-object helper only needs the name rendered by the UI. This also follows the existing `findByRole` pattern used elsewhere for hardware profile selection.

## Test plan
- [ ] Jenkins E2E: `testModelDeploymentKueueLifecycle.cy.ts`,`testModifyAppliedHardwareProfile.cy.ts`, `testModelServingTolerations.cy.ts`, `testNotebookTolerations.cy.ts`, `testWorkbenchTolerations.cy.ts`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved profile selection in the model serving workflow by consistently using the appropriate profile resource identifier.
  * Preserved validation and logging behavior for disabled profiles.
  * This improves reliability when selecting profiles that may be enabled or disabled during the workflow.


packages/cypress/cypress/pages/modelServing.ts
```diff
@@ -1258,7 +1258,7 @@ class ModelServingWizard extends Wizard {
     cy.findByRole('option', { name }).click();
   }
 
-  selectPotentiallyDisabledProfile(profileDisplayName: string, profileName?: string): void {
+  selectPotentiallyDisabledProfile(profileDisplayName: string): void {
```
claudialphonse78: @kanishka-commits why have you removed this parameter ? this function `selectPotentiallyDisabledProfile` is being used in multiple E2E tests across testing it only for our case (kueue )does not satisfy the removal

kanishka-commits:
### Jenkins build /2975 - Passing

Screenshot 2026-09-16 at 3 59 46 PM
Screenshot 2026-09-16 at 3 59 53 PM

claudialphonse78 (approved):

claudialphonse78 (approved):

PR #9577, opened 2026-09-02
fix(e2e): wait for dashboard label and Active before project list lookup

kanishka-commits:
https://redhat.atlassian.net/browse/RHOAIENG-88567

## Description

Fixes flaky Cypress E2E failures after `oc new-project` when navigating the Projects list.

- Wait for namespace readiness before UI lookup.
- Improve Projects table polling and loading handling.
- Fix name-filter input clearing.
- Return chainables from `createCleanProject`.

## How Has This Been Tested?

- **Cluster:** `ods-qe-psi-03` — RHOAI 3.6.0-ea.1, OCP 4.20.15
- **Specs:** `testWorkbenchStatus.cy.ts`, `testProjectEditing.cy.ts` — passing.
- **Lint:** Changed Cypress files — clean.

## Test Impact

Cypress E2E page-object/helper changes only. Existing specs using `createCleanProject` → `filterProjectByName` benefit from the fix.

## Request Review Criteria

- [x] Manually tested and verified.
- [x] Testing instructions added.
- [x] Tests added or testing rationale provided.
- [x] Code follows project best practices.

## Before Merge

- [ ] Test the solution on a cluster using the PR image built for `main`.

## Summary by CodeRabbit

* **Tests**
  * Improved project-list test reliability by waiting for loading and empty states, refining project-name filtering, and adding retry handling.
  * Enhanced project creation checks to confirm namespaces are ready, labels are applied successfully, and projects become active.
  * Updated project setup and cleanup checks to wait for operations to complete before continuing.

PR #9579, opened 2026-09-02
fix(e2e): stabilize serving runtime row and project list lookups

antowaddle:
https://redhat.atlassian.net/browse/RHOAIENG-88567

## Description

Two Cypress E2E flakes that show up on Jenkins (full OOTB runtime lists / `oc new-project` racing the Projects table), not in a typical local cluster.

**Serving runtime templates:** After creating a custom runtime, the test asserted `cy.contains(displayName).should('be.visible')`. Jenkins installs enough out-of-the-box runtimes that the new row is in the DOM but clipped by page overflow, so Cypress treats it as hidden. Assert the row exists via `data-testid` and open delete from the row kebab (same pattern as `testServingRuntimesUnsupported.cy.ts`).

**Projects list (port of https://github.com/opendatahub-io/odh-dashboard/pull/9577):** `oc new-project` can return before the namespace is labeled and `Active`, so the A.I. projects filter misses the row. Wait for the namespace, require `opendatahub.io/dashboard=true` (fail if labeling fails), wait until labeled + Active, return chainables from `createCleanProject` / `createAndVerifyProject`, and stabilize `filterProjectByName` (wait for table vs loading title, type into the real filter input, poll by href, reload/retry).

`should('be.visible', { timeout: 60000 })` does not apply that timeout in Cypress; the 60s wait is on `findProjectsTable(60000)` instead.

## How Has This Been Tested?

- Lint-staged (eslint) passed on the changed Cypress files.
- Serving-runtime visibility failure was observed on Jenkins (`dash-e2e-odh`); locally the shorter runtime list does not reproduce it.
- Project-list helper behavior matches #9577 (`testWorkbenchStatus.cy.ts` / `testProjectEditing.cy.ts` passed on `ods-qe-psi-03` in that PR). Combined branch still needs a Jenkins rerun of the affected E2E specs.

## Test Impact

Cypress E2E only. No new specs.

- `testSingleServingRuntimeCreation.cy.ts` — assert row existence instead of viewport visibility.
- `createOpenShiftProject` / `createCleanProject` / `filterProjectByName` — used by project-list specs (`testWorkbenchStatus`, `testFeatureStoreWorkbenchConnect`, `testProjectEditing`, and others that filter by name after `oc new-project`).

## Request review criteria:

## Summary by CodeRabbit

* **Tests**
  * Improved project list test reliability by waiting for table or empty-state content and allowing additional time for projects to appear.
  * Updated project filtering and deletion checks to reflect current interface behavior.
  * Strengthened serving runtime creation and cleanup verification.
  * Improved project setup checks by confirming labels and active status before tests continue.

### DIFF
--- a/packages/cypress/cypress/pages/projects.ts
+++ b/packages/cypress/cypress/pages/projects.ts
@@ -19,7 +19,7 @@ class ProjectListToolbar extends Contextual<HTMLElement> {
     return this.find().findByLabelText(`Filter by ${name}`);
   }
 
-  findNameFilter(): Cypress.Chainable<JQuery<HTMLElement>> {
+  findNameFilter(): Cypress.Chainable<JQuery<HTMLInputElement>> {
     // data-testid is on the PF TextInputGroup wrapper; type/clear need the input.
     return this.find().findByTestId('project-list-name-filter').find('input');
   }
--- a/packages/cypress/cypress/pages/projects.ts
+++ b/packages/cypress/cypress/pages/projects.ts
@@ -20,7 +20,8 @@ class ProjectListToolbar extends Contextual<HTMLElement> {
   }
 
   findNameFilter(): Cypress.Chainable<JQuery<HTMLElement>> {
-    return this.find().findByTestId('project-list-name-filter');
+    // data-testid is on the PF TextInputGroup wrapper; type/clear need the input.
+    return this.find().findByTestId('project-list-name-filter').find('input');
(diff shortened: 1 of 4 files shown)


sridarna:
The e2e tests are passing with the change.
/job/components/job/dashboard/job/dashboard-e2e-tests/2633/Test_20Reports/
/lgtm
/approve

PR #9562, opened 2026-09-01
feat: serve the MaaS Consumer Portal via Core BFF

caponetto:
https://redhat.atlassian.net/browse/RHOAIENG-90190

## Description

Build and package the MaaS Consumer Portal with the existing `odh-core-bff` image.

- Build the portal production bundle alongside the existing Core-BFF frontend and copy it to `/static/maas-consumer-portal`.
- Install the MaaS and GenAI frontend build inputs required by the statically composed portal.
- Keep static composition safe by resolving the eager federation singleton runtime packages from one copy, while retaining package-local dependencies such as GenAI's Markdown and Monaco stack.
- Fail the container build if the portal `index.html` was not packaged.
- Change portal logout to use the gateway-supported `/oauth2/sign_out` endpoint.
- Make the temporary MaaS Consumer Portal ConsoleLink `params.env` location writable for an arbitrary OpenShift runtime UID.
- Ignore nested `node_modules` directories from Docker build context.
- Retry MaaS Consumer Portal ConsoleLink reconciliation and requeue when the ConsoleLink drifts.
- Validate Core-BFF module proxy service port ranges and proxy paths before registering routes.
- Avoid false module-proxy path collision failures for valid sibling routes.

The portal uses the same Core-BFF image as the dashboard; no portal-specific related image, repository, or tag is introduced.

## How Has This Been Tested?

Automated checks:

- `npm run test-unit --workspace @odh-dashboard/maas-consumer-portal-distribution -- --runInBand`
- `npm run type-check --workspace @odh-dashboard/maas-consumer-portal-distribution`
- `npm run lint --workspace @odh-dashboard/maas-consumer-portal-distribution`
- `npx rspack --config distributions/maas-consumer-portal/config/rspack.prod.js --output-path /private/tmp/portal rspack-runtime-check`
- Core-BFF tests: `make test` from `distributions/core-bff/bff` (provisions envtest assets, then runs all Go packages)
- Dashboard operator unit tests: `make test` from `dashboard-operator` (runs all non-integration Go packages with race detection and coverage)
- Dashboard operator envtest integration tests: `make test-integration` from `dashboard-operator`

Local Core-BFF and mocked BFF validation was performed with the following 3-terminal setup. Run the commands from the `opendatahub-io-odh-dashboard` repository root.

[...]

PR #9581, opened 2026-09-02
feat: add MaaS Consumer Portal operand manifests

caponetto:
[https://redhat.atlassian.net/browse/RHOAIENG-90189](https://redhat.atlassian.net/browse/RHOAIENG-90189)

## Description

Adds the MaaS Consumer Portal operand manifest bundle under `manifests/distributions/maas-consumer-portal`.

The bundle defines the one-replica standalone Core-BFF portal Deployment, Service, ServiceAccount, least-privilege RBAC, NetworkPolicy, HTTPRoute, and ConsoleLink. It uses the shared `odh-core-bff` image, TLS serving certificates, projected service-account credentials, trusted CA bundles, and the portal federation ConfigMap interface.

The ConsoleLink manifest is migrated into the portal bundle while its current independent reconciliation remains ConsoleLink-only. Full portal lifecycle, the portal federation ConfigMap, and injection of the platform-provisioned nested portal hostname are follow-up work.

Also updates the dashboard-operator image permission for the migrated `params.env`, adds the bundle to Kustomize CI validation, and adds a rendering contract test for the real bundle.

Depends on [#9562](https://github.com/opendatahub-io/odh-dashboard/pull/9562) merging first, which packages the MaaS Consumer Portal bundle in the shared `odh-core-bff` image and makes the pre-migration `params.env` writable.

## How Has This Been Tested?

- `make lint` from `dashboard-operator/`
- `make test` from `dashboard-operator/`
- `make test-integration` from `dashboard-operator/`
- Rendered the checked-in portal bundle in an isolated temporary directory with all operator-provided values injected; the test verifies rendered resources, labels, security settings, replacements, service TLS settings, HTTPRoute, ConsoleLink, RBAC, and NetworkPolicy egress.

`npm run type-check` and `npm run lint` from the repository root remain blocked by pre-existing unrelated issues in `packages/eval-hub` and Core-BFF frontend tooling, respectively.

End-to-end cluster validation will be possible after the platform provisions the nested portal hostname and the shared BFF and portal lifecycle reconciliation work is available.

## Test Impact

Adds `TestRenderMaasConsumerPortalManifestBundle`, a real-manifest rendering test. It protects the Kustomize replacements and the portal operand contract without mutating the checked-in `params.env` file.

## Request review criteria:

## Summary by CodeRabbit

* **New Features**
  * Added a deployable MaaS Consumer Portal with HTTPS access, routing, service discovery, authentication, and secure runtime configuration.
  * Added required permissions, networking rules, service account, and TLS integration.
  * Added configurable portal URL, hostname, gateway, image, and federation settings.

* **Bug Fixes**
  * Updated deployment validation to include the MaaS Consumer Portal distribution.
  * ConsoleLink reconciliation now uses the portal distribution while continuing to expose the ConsoleLink resource.


manifests/distributions/maas-consumer-portal/cluster-role.yaml
```diff
@@ -0,0 +1,24 @@
...
+      - subjectaccessreviews
+    verbs:
+      - create
+  - apiGroups:
+      - datasciencecluster.opendatahub.io
+    resources:
+      - datascienceclusters
+    verbs:
+      - get
+      - list
```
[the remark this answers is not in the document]
caponetto: Not applying: these permissions match the established MaaS/GenAI RBAC baseline. list on datascienceclusters supports MaaS discovery, and get on ingresses is already used by the related module roles.

manifests/distributions/maas-consumer-portal/deployment.yaml
```diff
@@ -0,0 +1,153 @@
...
+            periodSeconds: 30
+            successThreshold: 1
+            failureThreshold: 3
+          resources:
+            requests:
+              cpu: 50m
+              memory: 128Mi
+              ephemeral-storage: 10Mi
+            limits:
+              memory: 256Mi
```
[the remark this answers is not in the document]
caponetto: Not applying: the portal resources intentionally match the existing Core-BFF pattern—CPU request with memory-only limit. No workload measurements or established CPU-limit value support adding one.

manifests/distributions/maas-consumer-portal/httproute.yaml
```diff
@@ -0,0 +1,19 @@
...
+      namespace: openshift-ingress
+  hostnames:
+    - portal-hostname
+  rules:
+    - matches:
+        - path:
+            type: PathPrefix
+            value: /
+      backendRefs:
+        - name: maas-consumer-portal
```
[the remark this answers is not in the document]
caponetto: Not applying: this repository’s established Gateway pattern uses the service-serving certificate annotation plus service.beta.kubernetes.io/backend-protocol: HTTPS, without BackendTLSPolicy or a CA ConfigMap. The portal follows that same main-dashboard and module pattern; adding new Gateway API resources would be unsupported scope expansion.

manifests/distributions/maas-consumer-portal/networkpolicy.yaml
```diff
@@ -0,0 +1,49 @@
...
+        - protocol: UDP
+          port: 5353
+        - protocol: TCP
+          port: 5353
+    # Core-BFF uses the projected user token for Kubernetes API calls. The API
+    # endpoint address is platform-managed and has no stable label selector;
+    # this is restricted to its HTTPS control-plane port.
+    - to:
+        - ipBlock:
+            cidr: 0.0.0.0/0
```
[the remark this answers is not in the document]
caponetto: Not applying: the control-plane CIDR and a platform-native API-server selector are not stable contracts available to these manifests. The task explicitly permits 0.0.0.0/0:6443 when justified, and the manifest documents why that constrained exception is required.

manifests/distributions/maas-consumer-portal/deployment.yaml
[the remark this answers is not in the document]
caponetto: The task requirement is concrete requests/limits consistent with the Core-BFF container; it does not require a CPU limit. The portal already has CPU, memory, and ephemeral-storage requests plus a memory limit matching Core-BFF. Adding an unmeasured CPU limit would diverge from the established workload policy and risks CPU throttling. We should revisit only if performance measurements establish an appropriate value.

manifests/distributions/maas-consumer-portal/networkpolicy.yaml
```diff
@@ -0,0 +1,49 @@
...
+      ports:
+        - protocol: TCP
+          port: 6443
```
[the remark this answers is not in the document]
caponetto: Not applying: the portal has no required external HTTPS dependency. Its Kubernetes API access follows the existing module policy contract on control-plane port 6443; adding 0.0.0.0/0:443 would broaden egress to arbitrary HTTPS endpoints without a portal use case.

manifests/distributions/maas-consumer-portal/params.env
```diff
@@ -0,0 +1,9 @@
+# Injected by the Dashboard Module Controller. portal-hostname and
+# maas-consumer-portal-url are independently resolved by the selected URL model.
+core-bff-image=quay.io/opendatahub/odh-core-bff:main
```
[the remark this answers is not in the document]
caponetto: Not applying: params.env defaults follow the existing dashboard manifest convention and are only fallback values for standalone rendering. During reconciliation, RELATED_IMAGE_ODH_CORE_BFF_IMAGE overrides core-bff-image with the release-provided immutable image reference. Hardcoding a digest here would bypass that promotion path and require manual updates.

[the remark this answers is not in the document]
caponetto: Thanks—correct that the current ConsoleLink-only reconciler does not inject the image into this bundle. However, this task does not apply the Deployment, and direct standalone application of this operator-owned bundle is not a supported install path. The mutable fallback also matches existing dashboard manifest conventions.

When the follow-up task deploys the full bundle, it must inject RELATED_IMAGE_ODH_CORE_BFF_IMAGE; that is an explicit acceptance criterion.

caponetto:
@lucferbux this is ready for review, though it must wait for https://github.com/opendatahub-io/odh-dashboard/pull/9562 to land first

/hold until https://github.com/opendatahub-io/odh-dashboard/pull/9562 lands first

dashboard-operator/Dockerfile
```diff
@@ -25,7 +25,8 @@ COPY --from=builder /tmp/manager .
 COPY --chown=65532:0 manifests/ /opt/manifests/dashboard/
 RUN chmod -R g+w /opt/manifests/dashboard/modules/ \
     && chmod g+w /opt/manifests/dashboard/odh/params.env \
-    && chmod g+w /opt/manifests/dashboard/rhoai/params.env
+    && chmod g+w /opt/manifests/dashboard/rhoai/params.env \
+    && chmod g+w /opt/manifests/dashboard/distributions/maas-consumer-portal/params.env
```
lucferbux, line 29: Heads-up for the rebase: this PR removes `manifests/maas-consumer-portal-consolelink/` entirely, but #9562 (which lands first) adds `chmod g+w /opt/manifests/dashboard/maas-consumer-portal-consolelink/rhoai/params.env` to this same `RUN` block. Once you rebase on top of #9562 you'll hit a conflict here — make sure the resolution drops that consolelink chmod line, otherwise the image build fails on a `chmod` against a path that no longer exists.

manifests/distributions/maas-consumer-portal/params.env
```diff
@@ -0,0 +1,9 @@
...
+dashboard-namespace=
+gateway-name=
+maas-consumer-portal-url=
+portal-federation-config=
```
lucferbux, line 7: This placeholder is the consumer side of the federation seam with #9625: the Deployment mounts a ConfigMap whose name is replaced from `data.portal-federation-config`, but it's empty here. #9625 produces that ConfigMap as `maas-consumer-portal-federation-config` — whoever wires the portal lifecycle has to inject that exact name into this param, or Core-BFF starts with no remotes config and the federation mount points at nothing. Worth a comment noting the expected value so the two halves don't drift.

manifests/distributions/maas-consumer-portal/kustomization.yaml
```diff
@@ -0,0 +1,115 @@
...
+        fieldPaths:
+          - subjects.0.namespace
+      - select:
+          kind: Deployment
+          name: maas-consumer-portal
+        # Kustomize cannot select a scalar argument by flag name. Keep
+        # --namespace=dashboard-namespace at args.5; the manifest rendering
+        # test verifies the resulting argument.
+        fieldPaths:
+          - spec.template.spec.containers.[name=maas-consumer-portal].args.5
```
lucferbux, line 102: nit: pinning the namespace replacement to `args.5` by position is fragile — reorder the container args in deployment.yaml and this silently rewrites the wrong flag. The render test guards it (it'd fail since `--namespace=` wouldn't get the injected value), so not blocking; just flagging the positional coupling in case the args list grows.

manifests/distributions/maas-consumer-portal/cluster-role.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: maas-consumer-portal
+rules:
```
lucferbux, line 5: This is exactly the least-privilege scope I'd want here — SAR create, DSC get/list, ingress get and nothing else — and the render test pins it to these three rules so it can't quietly grow later. Nice.

manifests/distributions/maas-consumer-portal/deployment.yaml
```diff
@@ -0,0 +1,153 @@
...
+              mountPath: /var/run/secrets/kubernetes.io/serviceaccount
+              readOnly: true
+          args:
+            - --deployment-mode=standalone
+            - --auth-method=user_token
+            - --auth-token-header=x-forwarded-access-token
+            - --auth-token-prefix=
+            - --platform-type=OpenShift
+            - --namespace=dashboard-namespace
+            - --static-assets-dir=/static/maas-consumer-portal
```
lucferbux, line 141: Good to see the other half of #9562 wired through here — the shared core-bff image serving the portal bundle from `/static/maas-consumer-portal` via this flag. The two PRs line up cleanly.

dashboard-operator/internal/controller/maas_consumer_portal_manifest_test.go
```diff
@@ -0,0 +1,183 @@
...
+	// runtime, so rendering the source directory directly would mutate the worktree.
+	source := filepath.Join("..", "..", "..", "manifests", "distributions", maasConsumerPortalName)
+	dir := filepath.Join(t.TempDir(), maasConsumerPortalName)
+	require.NoError(t, os.CopyFS(dir, os.DirFS(source)))
+
+	params := readExistingParams(filepath.Join(dir, "params.env"))
+	params["core-bff-image"] = maasConsumerPortalCoreBFFImage
+	params["dashboard-namespace"] = "portal-test"
+	params["gateway-name"] = "portal-gateway"
+	params["maas-consumer-portal-url"] = "https://portal.apps.example.com/"
```
lucferbux, line 31: Minor: the test sets all seven params, but the runtime ConsoleLink-only path only injects `maas-consumer-portal-url` and `section-title` and renders the full bundle with the rest empty. That empty-param render isn't covered — might be worth a case that renders with just those two set, to lock in that the ConsoleLink still comes out clean when gateway-name/hostname/federation-config are blank.

lucferbux (commented):
Went through this alongside #9562 and the portal bundle scope. The ConsoleLink migration into `manifests/distributions/maas-consumer-portal` reads clean — the deploy action still filters the render down to just the ConsoleLink, so we're not shipping the Deployment/HTTPRoute ahead of the lifecycle work in #9625, and deploy/delete stay symmetric. RBAC is genuinely least-privilege, the security context is fully hardened, and the render contract test is thorough and renders a temp copy so it doesn't mutate the checked-in params.env. I also confirmed the ConsoleLink `MarkTrue` isn't clobbered by the module-demand condition (reconcile ordering is fine).

Left inline notes: one real heads-up on the Dockerfile re: merge order with #9562, the federation-config seam with #9625, and a couple of nits. Nothing blocking here — leaving these as comments.

manifests/distributions/maas-consumer-portal/params.env
lucferbux: This is the consumer side of the federation seam with #9625: that PR produces the ConfigMap `maas-consumer-portal-federation-config`, but this placeholder is empty and the Deployment mounts it through the `portal-federation-config` replacement (kustomization.yaml:115 → the volume's `configMap.name`). Whatever task actually deploys the portal bundle has to set this to `portal-federation-config=maas-consumer-portal-federation-config`, or the volume points at a ConfigMap that doesn't exist and core-bff comes up with no remotes config. I know the PR body scopes it as a follow-up — just flagging it on the line so it doesn't get lost between the stacked PRs.

lucferbux (commented):
Note: this entry is a leftover from a concurrent review run on my end — my full review of this PR is the companion COMMENT above (the Dockerfile rebase hazard vs #9562, the empty `portal-federation-config` placeholder that #9625 has to fill, the `args.5` positional coupling, the least-privilege RBAC affirmation, and the empty-param render test gap). I trimmed the duplicated inline notes here and left just one extra nit inline, on ephemeral-storage sizing.

manifests/distributions/maas-consumer-portal/deployment.yaml
```diff
@@ -0,0 +1,153 @@
...
+            initialDelaySeconds: 15
+            timeoutSeconds: 15
+            periodSeconds: 30
+            successThreshold: 1
+            failureThreshold: 3
+          resources:
+            requests:
+              cpu: 50m
+              memory: 128Mi
+              ephemeral-storage: 10Mi
```
lucferbux, line 103: nit: with `readOnlyRootFilesystem`, the `tmp` emptyDir (sizeLimit 100Mi) is the only writable path, but the ephemeral-storage request here is 10Mi with no limit. If the portal ever writes past 10Mi to /tmp it could be evicted under node pressure before it hits the 100Mi cap. Might be worth bumping the request or adding a matching limit so scheduling reflects the real ceiling. Non-blocking.

dashboard-operator/internal/controller/actions.go
```diff
@@ -275,7 +275,7 @@ func deployMaasConsumerPortalConsoleLink(
 		logger.Info("MaaS Consumer Portal ConsoleLink is an RHOAI-only feature; deploying it on Open Data Hub will use RHOAI branding")
 	}
 
-	m := maasConsumerPortalConsoleLinkManifestInfo(basePath)
+	m := maasConsumerPortalManifestInfo(basePath)
```
lucferbux, line 278: Cross-PR heads-up with #9625 (also open against `main`, not stacked): this PR renames `maasConsumerPortalConsoleLinkManifestInfo` -> `maasConsumerPortalManifestInfo`, moves the bundle to `distributions/maas-consumer-portal/`, and deletes `manifests/maas-consumer-portal-consolelink/`. #9625's `maas_consumer_portal_consolelink.go` re-adds the old function still pointing at `maas-consumer-portal-consolelink/rhoai` — the exact path this PR removes. This is separate from the federation-config seam and the #9562 Dockerfile note: whoever merges second needs a semantic rebase (fn rename + path move), not just a textual conflict resolution, or the render blows up on a missing dir. Worth agreeing the merge order explicitly.

lucferbux (commented):
Follow-up after re-reading #9581 and #9625 side by side — one thing my first pass didn't flag that only shows up when you look at both PRs together. Not blocking, just needs merge coordination. Inline.

caponetto:
Superseded by https://github.com/opendatahub-io/odh-dashboard/pull/9637

PR #9588, opened 2026-09-02
fix: fire Model Updated instead of Model Deployed in edit flow

YuliaKrimerman:
https://issues.redhat.com/browse/RHOAIENG-82390

## Description

Follow-up to the model deployment tracking PR. The `fireModelDeployed` helper already accepted
an `isEdit` flag to emit `"Model Updated"` vs `"Model Deployed"`, but the flag was never
threaded through from the wizard to either tracking call site. As a result, every event fired
as `"Model Deployed"` — even when editing an existing deployment — breaking the
_"Model Updated"_ Amplitude reports and diverging from the legacy `ManageKServeModal` behaviour.

**Root cause:** Two call sites both ignored the available `isEdit` signal:

1. **Submit path** — `useModelDeployedTracking` did not accept or forward `isEdit` to
   `fireModelDeployed`.
2. **Cancel path** — `useExitDeploymentWizard` did not accept or forward `isEdit` to
   `fireModelDeployed`.

**Fix:** Thread `isEdit` (`!!existingDeployment`) from the wizard down to both call sites:

## How Has This Been Tested?

- Existing unit tests for `fireModelDeployed` already cover the `isEdit: true` → `"Model Updated"` and `isEdit: false` → `"Model Deployed"` branches (see `deploymentTracking.spec.ts`).
- Ran the full `shared/tracking` unit test suite locally — 24 tests pass.
- Manual verification: lint (`LINT OK`) and type-check (`tsc --noEmit` with no errors).

## Test Impact

No new tests required — the fix is purely plumbing the existing `isEdit` parameter through.
The correctness of `fireModelDeployed(isEdit=true)` → `"Model Updated"` is already covered
by `packages/model-serving/src/shared/tracking/__tests__/deploymentTracking.spec.ts`
(`"should fire Model Updated event when isEdit is true"`).

## Request review criteria:

### DIFF
--- a/packages/model-serving/src/components/deploymentWizard/ModelDeploymentWizard.tsx
+++ b/packages/model-serving/src/components/deploymentWizard/ModelDeploymentWizard.tsx
@@ -56,7 +56,7 @@ const ModelDeploymentWizard: React.FC<ModelDeploymentWizardProps> = ({
 }) => {
   const onRefresh = useRefreshWizardPage(existingDeployment);
   const { isExitModalOpen, openExitModal, closeExitModal, handleExitConfirm, exitWizardOnSubmit } =
-    useExitDeploymentWizard({ returnRoute, cancelReturnRoute });
+    useExitDeploymentWizard({ returnRoute, cancelReturnRoute, isEdit: !!existingDeployment });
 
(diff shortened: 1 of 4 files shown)


emilys314 (approved):
Looks good, i see the "Model Updated" in the logs

Image

PR #9608, opened 2026-09-03
Show connected nim pvcs

emilys314:
Closes https://redhat.atlassian.net/browse/RHOAIENG-88034

## Description

Adds an extension for the cluster storage tab to fetch resources and to check if a pvc shows up in that resource(s). This is added to kserve generically because kserve knows how to fetch and interpret InferenceService + ServingRuntimes best, and the logic technically isn't really specific to NIM. 

The extension has a hook which is ran at the table level. It returns a Generic result based on the extension. The extension then has a parser function which links each PVC in the row with an array of names it's connected to. 

In `useClusterStorageConnectedResources`, I use both `useExtensions` and `useResolvedExtensions`. because useExtensions knows the list of extensions at compile time while the resolved variant only tells you the number after it is finished resolved it. This optimizes the loader and the visibility of the column a little bit

Also in the kserve extension hook, it will return `loaded: true` early if there are 0 inferenceServices.

image

## How Has This Been Tested?

Baseline behavior
1. Create a new project
2. Create a new cluster storage, it should have no connected resources

New NIM
1. Enable NIM, deploy a model, and select creating a new PVC
2. Go to the cluster storage tab, it should show the PVC with the name of the NIM model you created

New NIM with existing PVC
1. Deploy a new NIM model, select an existing PVC,  the one from the other NIM model
2. It should show both models as connected resources

image

## Test Impact

kserve connected resource logic has jest tests
the nim cluster storage cypress mock test was updated to check one as a connected resource

## Request review criteria:

If you have UI changes: 

## Summary by CodeRabbit

* **New Features**
  * Cluster storage now displays resources connected to each PVC, including deployed model names.
  * Connected-resource information can be provided by supported extensions and appears with resource-specific labels.
  * Added loading indicators and notifications while connected-resource details are retrieved.
* **Bug Fixes**
  * Improved handling of empty, loading, and unmatched connected-resource states.
  * Added support for displaying connected resources when extensions are available, even if workbenches are disabled.

### DIFF
--- a/frontend/src/pages/projects/screens/detail/connections/ConnectedResources.tsx
+++ b/frontend/src/pages/projects/screens/detail/connections/ConnectedResources.tsx
@@ -18,7 +18,7 @@ const CONNECTED_RESOURCE_LABEL_STYLES: Record<
   ConnectedResourceLabel['kind'],
   Pick<React.ComponentProps<typeof ResourceLabel>, 'resourceType' | 'outlineColor'>
 > = {
-  'deployed-model': { resourceType: ProjectObjectType.deployedModels, outlineColor: 'purple' },
+  'connected-models': { resourceType: ProjectObjectType.connectedModels, outlineColor: 'purple' },
 };
 
 export type ConnectedResourcesProps = EitherNotBoth<
--- a/frontend/src/pages/projects/screens/detail/connections/ConnectedResources.tsx
+++ b/frontend/src/pages/projects/screens/detail/connections/ConnectedResources.tsx
@@ -2,6 +2,7 @@ import * as React from 'react';
 import { LabelGroup, Spinner } from '@patternfly/react-core';
 import type { PersistentVolumeClaimKind } from '@odh-dashboard/k8s-core';
 import type { EitherNotBoth } from '@odh-dashboard/foundation';
+import type { ConnectedResourceLabel } from '@odh-dashboard/plugin-core/extension-points';
 import { getDisplayNameFromK8sResource } from '@odh-dashboard/k8s-core';
 import {
   useRelatedNotebooks,
@@ -12,12 +13,28 @@ import { ProjectObjectType } from '#~/concepts/design/utils';
 import ResourceLabel from '#~/pages/projects/screens/detail/connections/ResourceLabel';
 import { useInferenceServicesForConnection } from '#~/pages/projects/useInferenceServicesForConnection';
 
+// Host owns how each contributed resource kind renders (icon + color); extensions only pick a kind.
+const CONNECTED_RESOURCE_LABEL_STYLES: Record<
+  ConnectedResourceLabel['kind'],
+  Pick<React.ComponentProps<typeof ResourceLabel>, 'resourceType' | 'outlineColor'>
+> = {
+  'deployed-model': { resourceType: ProjectObjectType.deployedModels, outlineColor: 'purple' },
+};
+
 export type ConnectedResourcesProps = EitherNotBoth<
   { connection: Connection },
   { pvc: PersistentVolumeClaimKind }
->;
+> & {
+  additionalResources?: ConnectedResourceLabel[];
+  additionalResourcesLoaded?: boolean;
+};
 
-const ConnectedResources: React.FC<ConnectedResourcesProps> = ({ connection, pvc }) => {
+const ConnectedResources: React.FC<ConnectedResourcesProps> = ({
+  connection,
+  pvc,
+  additionalResources = [],
+  additionalResourcesLoaded = true,
+}) => {
   const { notebooks: connectedNotebooks, loaded: notebooksLoaded } = useRelatedNotebooks(
     connection
       ? ConnectedNotebookContext.EXISTING_DATA_CONNECTION
@@ -26,11 +43,11 @@ const ConnectedResources: React.FC<ConnectedResourcesProps> = ({ connection, pvc
   );
   const connectedModels = useInferenceServicesForConnection(connection ?? pvc);
 
-  if (!notebooksLoaded) {
+  if (!notebooksLoaded || !additionalResourcesLoaded) {
     return <Spinner size="sm" />;
   }
 
-  if (!connectedNotebooks.length && !connectedModels.length) {
+  if (!connectedNotebooks.length && !connectedModels.length && !additionalResources.length) {
     return '--';
   }
 
@@ -54,10 +71,16 @@ const ConnectedResources: React.FC<ConnectedResourcesProps> = ({ connection, pvc
       />
     ));
 
+  const renderAdditionalResourceLabels = () =>
+    additionalResources.map(({ key, title, kind }) => (
+      <ResourceLabel key={key} title={title} {...CONNECTED_RESOURCE_LABEL_STYLES[kind]} />
+    ));
+
   return (
     <LabelGroup>
       {renderNotebookLabels()}
       {renderModelLabels()}
+      {renderAdditionalResourceLabels()}
     </LabelGroup>
   );
 };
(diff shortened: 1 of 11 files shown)


frontend/src/pages/projects/screens/detail/storage/useClusterStorageConnectedResources.tsx
```diff
@@ -0,0 +1,69 @@
...
+      resolvedExtensions.flatMap((extension) => {
+        const entry = connectedResourceData[extension.uid];
+        return entry ? extension.properties.getConnectedResources(pvc, entry.data) : [];
+      }),
+    [resolvedExtensions, connectedResourceData],
+  );
+
+  const hookNotifications = resolvedExtensions.map((extension) => (
+    <HookNotify
+      key={extension.uid}
```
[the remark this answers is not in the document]
emilys314: Hm I'm not able to reproduce this. I feel like the dom goes through too many changes to be able to persist the same hooks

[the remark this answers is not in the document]
emilys314: The two real ways to switch project:
Edit the URL → full document reload → everything mounts fresh → useState reset.
Via the all-projects page → you route to /projects (renders <ProjectView/>, a different route element → provider unmounts), then mount fresh for project B.

The page can't change the project in-place I'm pretty sure

packages/plugin-core/src/extension-points/cluster-storage.ts
```diff
@@ -24,3 +24,48 @@ export type ClusterStorageContextExtension = Extension<
 
 export const isClusterStorageContextExtension =
   createExtensionGuard<ClusterStorageContextExtension>('app.cluster-storage/storage-context');
+
+export type ClusterStorageConnectedResources<TData = unknown> = {
+  loaded: boolean;
+  data: TData;
+};
+
+export type ConnectedResourceKind = 'deployed-model';
```
YuliaKrimerman: better go for enum instead of string-literal union here

`export enum ConnectedResourceKind {
  DeployedModel = 'deployed-model',
}
`

packages/kserve/src/clusterStorage/connectedResources.tsx
```diff
@@ -0,0 +1,60 @@
...
+export const getConnectedKServeResourceLabels = (
+  pvc: PersistentVolumeClaimKind,
+  { inferenceServices, servingRuntimes }: KServeConnectedResourcesData,
+): ConnectedResourceLabel[] =>
+  inferenceServices
+    .filter((inferenceService) => {
+      const servingRuntime = servingRuntimes.find(
+        (runtime) => runtime.metadata.name === inferenceService.spec.predictor.model?.runtime,
+      );
+      return servingRuntime ? servingRuntimeUsesPVC(servingRuntime, pvc.metadata.name) : false;
```
YuliaKrimerman, line 60: will a a generic KServe deployment mounting the same PVC also appear?

frontend/src/pages/projects/screens/detail/connections/ConnectedResources.tsx
```diff
@@ -12,12 +13,28 @@ import { ProjectObjectType } from '#~/concepts/design/utils';
 import ResourceLabel from '#~/pages/projects/screens/detail/connections/ResourceLabel';
 import { useInferenceServicesForConnection } from '#~/pages/projects/useInferenceServicesForConnection';
 
+// Host owns how each contributed resource kind renders (icon + color); extensions only pick a kind.
+const CONNECTED_RESOURCE_LABEL_STYLES: Record<
+  ConnectedResourceLabel['kind'],
+  Pick<React.ComponentProps<typeof ResourceLabel>, 'resourceType' | 'outlineColor'>
+> = {
+  'deployed-model': { resourceType: ProjectObjectType.deployedModels, outlineColor: 'purple' },
```
YuliaKrimerman: and also on lines, 47-54, For PVC-backed model storage where both paths match the same inference service, the user could see duplicate labels with different icons. Consider deduplicating by inference service name in ConnectedResources, or aligning on one ProjectObjectType.

YuliaKrimerman (commented):
Couple of small things, haven't tested yet.

packages/plugin-core/src/extension-points/cluster-storage.ts
emilys314: yeah enum vs string-literal unions are an interesting topic of debate, I don't think one is better than the other in a vacuum

I went with a type here because the extension-points is a mostly types only file. Enums are objects (usually) once they get compiled into javascript, while string-literals have 0 runtime weight after typescript compiles away into javascript

frontend/src/pages/projects/screens/detail/connections/ConnectedResources.tsx
emilys314: Oh so they are the same rocket icon, not sure why they are different. It makes sense though to use the same one, i'll change it to `connectedModels`

packages/kserve/src/clusterStorage/connectedResources.tsx
emilys314: Yes but generic KServe doesn't really use this type of PVC usage.

Kserve models typically put data on the InferenceService
legacy NIM models typically put data on the ServingRuntime

We should move the renderModelLabels into this function too at some point, right now it's tech debt that we have model serving code in the frontend there

packages/kserve/src/clusterStorage/__tests__/connectedResources.spec.tsx
```diff
@@ -0,0 +1,121 @@
...
+        name: 'is-other',
+        displayName: 'Other Model',
+        runtimeName: 'sr-other',
+      }),
+    ];
+
+    const labels = getConnectedKServeResourceLabels(pvc, { inferenceServices, servingRuntimes });
+    // Only the inference service whose runtime mounts the PVC, labeled with the inference service
+    // display name (not the serving runtime name).
+    expect(labels).toEqual([{ key: 'is-match', title: 'My NIM Model', kind: 'deployed-model' }]);
```
mturley: Other code now uses `kind: 'connected-models’` but this assertion still expects `'deployed-model’`, so this test is failing

mturley (changes requested):

packages/kserve/src/clusterStorage/__tests__/connectedResources.spec.tsx
emilys314: Sorry I missed this one, fixed and rebased, all are passing now
image

packages/kserve/src/clusterStorage/connectedResources.tsx
```diff
@@ -0,0 +1,60 @@
...
+  servingRuntimes: ServingRuntimeKind[];
+};
+
+/**
+ * Lists the project's inference services and their serving runtimes.
+ */
+export const useConnectedKServeResources = (
+  project: ProjectKind,
+): ClusterStorageConnectedResources<KServeConnectedResourcesData> => {
+  const [inferenceServices, inferenceServicesLoaded] = useWatchInferenceServices(project);
```
mturley: The watch errors are dropped here. `useWatchInferenceServices` and `useWatchServingRuntimes` return errors as their third tuple element, but this ignores those values and gates `loaded` only on the boolean flags. If either has an error, `additionalResourcesLoaded` stays false and every PVC row in the table stays on the spinner indefinitely with no error shown. Reproduced by blocking the requests in browser dev tools.

mturley (changes requested):
@emilys314 found a problem I didn’t notice before, sorry

packages/kserve/src/clusterStorage/connectedResources.tsx
emilys314: Ah that's a good catch, classic

I made it 'ignore' the error since there is no existing way to show network errors for the connected resources

image

I added jest tests to track this too

mturley (approved):
/approve
/lgtm

PR #9618, opened 2026-09-03
RHOAIENG-83655: add status condition fidelity tests

liday-rh:
https://redhat.atlassian.net/browse/RHOAIENG-83655

## Description

Adds envtest coverage for the Dashboard operator status trust contract:

- verifies `Ready` tracks provisioning success and recovers after a render failure
- verifies module Deployment health drives `Degraded`, `Ready`, and `Phase`
- verifies `status.observedGeneration` follows spec generation changes
- verifies `Phase` transitions from Ready to NotReady and back to Ready

The new degradation coverage exposed that `Degraded=True` was informational and therefore left `Ready=True` and `Phase=Ready`. This change treats module degradation as an error condition and explicitly marks the negative-polarity `Ready` condition false, preserving the `ModulesDegraded` reason and degraded-module count.

## How Has This Been Tested?

From `dashboard-operator`:

```text
make test-integration
make test
make lint
git diff --check
```

All commands pass. `make lint` reports 0 issues.

## Test Impact

Adds the following envtest integration tests:

- `TestIntegration_ConditionReady_TracksSubConditions`
- `TestIntegration_ConditionDegraded_ReflectsModuleHealth`
- `TestIntegration_ObservedGeneration_UpdatedPerReconcile`
- `TestIntegration_PhaseTransitions_ReadyNotReady`

Also adds focused unit coverage for degraded-condition aggregation, reason/message fidelity, severity, and Ready roll-up.

## Request review criteria:

If you have UI changes:
- [ ] Included any necessary screenshots or gifs if it was a UI change. (Not applicable: no UI changes.)
- [ ] Included tags to the UX team if it was a UI/UX change. (Not applicable: no UI/UX changes.)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved dashboard status reporting when modules are degraded.
  - The **Degraded** condition now provides a clear message identifying the number of affected modules.
  - The **Ready** condition is explicitly marked false while modules remain degraded.
  - Dashboard readiness, degraded status, observed generation, and phase transitions now update consistently during failed and recovered reconciliations.

### DIFF
--- a/dashboard-operator/internal/controller/dashboard_reconciler.go
+++ b/dashboard-operator/internal/controller/dashboard_reconciler.go
@@ -661,8 +661,14 @@ func (r *DashboardReconciler) reconcileDegradedCondition(
 	if degradedModules > 0 {
 		cm.MarkTrue(string(common.ConditionTypeDegraded),
 			conditions.WithReason("ModulesDegraded"),
-			conditions.WithError(fmt.Errorf("%d module(s) degraded", degradedModules)),
-			conditions.WithSeverity(common.ConditionSeverityInfo))
+			conditions.WithMessage("%d module(s) degraded", degradedModules),
+			conditions.WithSeverity(common.ConditionSeverityError))
+		// Degraded is a negative-polarity condition. The condition manager only
+		// rolls up false/unknown dependents, so explicitly make Ready unhappy when
+		// Degraded=True.
+		cm.MarkFalse(string(common.ConditionTypeReady),
+			conditions.WithReason("ModulesDegraded"),
+			conditions.WithMessage("%d module(s) degraded", degradedModules))
 	}
 }
 
--- a/dashboard-operator/internal/controller/dashboard_reconciler_test.go
+++ b/dashboard-operator/internal/controller/dashboard_reconciler_test.go
@@ -766,6 +766,35 @@ func TestReconcile_StatusContract(t *testing.T) {
 	}
 }
 
+func TestReconcileDegradedCondition(t *testing.T) {
+	dashboard := &v1alpha1.Dashboard{}
+	cm := conditions.NewManager(
+		dashboard,
+		string(common.ConditionTypeReady),
+		string(common.ConditionTypeDegraded),
+	)
+	r := &ctrlpkg.DashboardReconciler{}
+
+	r.ReconcileDegradedCondition(cm, map[string]v1alpha1.ModuleStatus{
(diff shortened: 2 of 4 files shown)


lucferbux:
Reviewed against [RHOAIENG-83655](https://redhat.atlassian.net/browse/RHOAIENG-83655). The four envtest cases map cleanly to the acceptance criteria (Ready sub-condition tracking, Degraded module-health reflection with count, observedGeneration-per-reconcile, and Phase transitions), plus focused unit coverage for the degraded roll-up. The production fix correctly addresses a real gap: `Degraded` is a negative-polarity condition, and the condition manager's `RecomputeHappiness`/`findUnhappyDependent` only treats `False`/`Unknown` dependents as unhappy, so `Degraded=True` alone never rolled Ready down — the explicit `MarkFalse(Ready)` fixes that, and the tests confirm both the failure and recovery paths. Switching `WithError` -> `WithReason("ModulesDegraded")` + `WithMessage(...)` is also an improvement since `WithError` was previously clobbering the reason with `ConditionReasonError`.

One non-blocking edge case worth a follow-up (not a regression — behavior here is strictly better than before): because `SetCondition` re-runs `RecomputeHappiness` on every mark, any dependent re-marked *after* `reconcileDegradedCondition` can flip Ready back to `True`. Concretely, the RBAC restore at `dashboard_reconciler.go:505-508` does `MarkTrue(Degraded, "NamespacedRBACFailed")` after the explicit `MarkFalse(Ready)`; since `Degraded=True` isn't counted as an unhappy dependent, that recompute resets `Ready=AllDependentsHealthy=True`. So a Dashboard that is both module-degraded and RBAC-failed ends up `Ready=True`. The same underlying limitation means RBAC/Federation degradation paths (lines 471, 482) also leave Ready happy. A more robust long-term fix would live in the roll-up (teach it negative-polarity conditions, or compute Ready once at the end of the pipeline) rather than per-call `MarkFalse(Ready)`. Fine to handle separately.

Note: `ci/prow/images` is red, but that's the monorepo-wide image build unrelated to these Go-only operator changes; the operator PR image-mirror jobs and CodeRabbit are green.

/lgtm
/approve


[RHOAIENG-83655]: https://redhat.atlassian.net/browse/RHOAIENG-83655?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #9625, opened 2026-09-03
feat(operator): orchestrate shared MaaS Consumer Portal module demand

caponetto:
[https://redhat.atlassian.net/browse/RHOAIENG-90166](https://redhat.atlassian.net/browse/RHOAIENG-90166)

Depends on:

- [https://github.com/opendatahub-io/odh-dashboard/pull/9562](https://github.com/opendatahub-io/odh-dashboard/pull/9562)
- [https://github.com/opendatahub-io/odh-dashboard/pull/9581](https://github.com/opendatahub-io/odh-dashboard/pull/9581)

## Description

Implements operand-aware, aggregate-demand orchestration for the shared MaaS and GenAI BFFs used by the core dashboard and MaaS Consumer Portal.

- Resolve module demand from both the core dashboard and MaaS Consumer Portal management states. MaaS and GenAI run when either operand requires them; no other module is enabled by the portal alone.
- Preserve explicit `spec.modules.<module>.state=Disabled` overrides and report `MaasConsumerPortalAvailable=False` with an actionable dependency reason when a portal-required module is unavailable.
- Keep shared MaaS and GenAI Deployments and Services during core-dashboard-only, portal-only, and dual-managed operation; remove them only when neither operand needs them.
- Generate and reconcile `maas-consumer-portal-federation-config`, containing deterministic MaaS and GenAI federation entries derived from the module registry. The ConfigMap uses MaaS Consumer Portal ownership labels and changes trigger a portal Deployment rollout through a configuration-hash annotation.
- Allow ingress from Pods labeled `deployment: maas-consumer-portal` to the MaaS and GenAI NetworkPolicies on ports 8243 and 8143.
- Preserve a primary MaaS Consumer Portal availability failure reason rather than allowing later dependency or ConfigMap checks to mask it.

### File Restructuring

MaaS Consumer Portal-specific controller code and tests were extracted from the general controller files into focused files:

- `maas_consumer_portal.go` for portal availability conditions and lifecycle state.
- `maas_consumer_portal_consolelink.go` for ConsoleLink reconciliation.
- `maas_consumer_portal_federation.go` for portal federation ConfigMap generation, reconciliation, and rollout hashing.
- Dedicated MaaS Consumer Portal unit, module-deploy, export, and envtest integration test files.

The shared module-demand and federation-entry logic remains in the general module controller code because it is intentionally consumed by both operands.

> **Note**: This PR produces the portal federation ConfigMap. The MaaS Consumer Portal Deployment manifest must mount `maas-consumer-portal-federation-config` so Core-BFF can consume its JSON.

## How Has This Been Tested?

  - make -C dashboard-operator fmt
  - make -C dashboard-operator vet
  - make -C dashboard-operator lint
  - make -C dashboard-operator test
  - make -C dashboard-operator test-integration
  - make -C dashboard-operator build
  - make -C dashboard-operator chart-lint

## Test Impact

- Added unit coverage for portal module availability, explicit overrides, ConfigMap contents, hash rollout behavior, condition precedence, and shared module-demand reconciliation.
- Added envtest coverage for all four core-dashboard/MaaS Consumer Portal Managed/Removed state combinations, including validation that both operands share a single MaaS and GenAI BFF deployment/service pair.

## Request Review Criteria

- [x] Testing instructions have been added in the PR body
- [x] The developer has added tests for the related changes
- [x] The code follows repository best practices
- [ ] The developer has tested the PR image on a cluster before merge

## Summary by CodeRabbit

* **New Features**
  * Added MaaS Consumer Portal integration, including ConsoleLink provisioning and federation configuration.
  * Added portal availability reporting when required modules or federation configuration are unavailable.
  * Enabled MaaS Consumer Portal traffic to required GenAI and MaaS services.

* **Bug Fixes**
  * Preserved module status details and transition times during resource teardown.
  * Improved handling of shared modules required by multiple dashboard components.

### DIFF
--- a/dashboard-operator/internal/controller/dashboard_reconciler.go
+++ b/dashboard-operator/internal/controller/dashboard_reconciler.go
@@ -35,7 +35,6 @@ import (
 
 const dashboardFinalizer = "components.platform.opendatahub.io/cleanup"
 const conditionObservabilityAvailable = "ObservabilityAvailable"
-const conditionMaasConsumerPortalAvailable = "MaasConsumerPortalAvailable"
 
 var operatorDeploymentName = getOperatorDeploymentName()
 
@@ -208,14 +207,27 @@ func (r *DashboardReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	if dashboard.Spec.ManagementState == "Removed" {
 		logger.Info("ManagementState is Removed, tearing down resources")
 
-		if err := r.teardownManagedResources(ctx, dashboard); err != nil {
+		// MaaS and GenAI are shared dependencies. Reconcile their aggregate
+		// demand before the core teardown so MaaS Consumer Portal-only operation retains them.
+		nextStatuses, err := r.reconcileModuleDemand(ctx, dashboard)
+		if err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to reconcile MaaS Consumer Portal-required modules: %w", err)
+		}
+		preserveModuleStatusTransitionTimes(dashboard.Status.ModuleStatuses, nextStatuses)
+		dashboard.Status.ModuleStatuses = nextStatuses
+		r.setMaasConsumerPortalModuleCondition(cm, dashboard, nextStatuses)
+		if err := r.deployMaasConsumerPortalFederationConfigMap(ctx, dashboard, nextStatuses); err != nil {
+			r.markMaasConsumerPortalFederationConfigMapFailed(cm, err)
+			logger.Error(err, "Failed to deploy MaaS Consumer Portal federation ConfigMap")
+		}
+
+		if err := r.teardownManagedResources(ctx, dashboard, nextStatuses); err != nil {
 			return ctrl.Result{}, fmt.Errorf("failed to tear down resources: %w", err)
 		}
 
 		dashboard.Status.ObservedGeneration = dashboard.Generation
 		dashboard.Status.Phase = common.PhaseNotReady
 		dashboard.Status.URL = ""
-		dashboard.Status.ModuleStatuses = nil
 		dashboard.Status.Distribution = nil
 
 		cm.MarkFalse(string(common.ConditionTypeProvisioningSucceeded),
@@ -429,40 +441,23 @@ func (r *DashboardReconciler) reconcileDeployment(
 		return ctrl.Result{}, fmt.Errorf("failed to deploy resources: %w", err)
 	}
 
-	nextStatuses := resolveModuleStatuses(&dashboard.Spec)
-
-	if err := r.deployModuleManifests(ctx, dashboard, nextStatuses); err != nil {
+	nextStatuses, err := r.reconcileModuleDemand(ctx, dashboard)
+	if err != nil {
 		cm.MarkFalse(string(common.ConditionTypeProvisioningSucceeded),
 			conditions.WithReason("ModuleDeployFailed"),
 			conditions.WithError(err))
 		return ctrl.Result{}, fmt.Errorf("failed to deploy module manifests: %w", err)
 	}
 
-	// GC disabled modules
-	if err := r.deleteModuleResources(ctx, nextStatuses); err != nil {
-		logger.Error(err, "Failed to clean up disabled module resources")
-	}
-
 	cm.MarkTrue(string(common.ConditionTypeProvisioningSucceeded),
 		conditions.WithReason("ResourcesApplied"),
 		conditions.WithMessage("Dashboard and module manifests applied successfully"))
 
-	// Overlay readiness from module deployments (before federation ConfigMap
-	// so the ConfigMap reflects actual deployment health, e.g. Degraded modules)
-	r.overlayStandaloneReadiness(ctx, nextStatuses)
-
 	// Persist module statuses now so early returns from steps 6-9 don't leave
 	// stale status on the CR (the outer Reconcile always calls Status().Update).
-	for name, next := range nextStatuses {
-		if prev, ok := dashboard.Status.ModuleStatuses[name]; ok &&
-			prev.Phase == next.Phase &&
-			prev.Reason == next.Reason &&
-			prev.Message == next.Message {
-			next.LastTransitionTime = prev.LastTransitionTime
-			nextStatuses[name] = next
-		}
-	}
+	preserveModuleStatusTransitionTimes(dashboard.Status.ModuleStatuses, nextStatuses)
 	dashboard.Status.ModuleStatuses = nextStatuses
+	r.setMaasConsumerPortalModuleCondition(cm, dashboard, nextStatuses)
 
 	// Reconcile cross-namespace RBAC (notebooks, model-registry)
 	rbacErr := r.reconcileNamespacedRBAC(ctx, dashboard)
@@ -484,6 +479,10 @@ func (r *DashboardReconciler) reconcileDeployment(
 			conditions.WithError(err))
 		logger.Error(err, "Failed to deploy federation ConfigMap")
 		return ctrl.Result{}, fmt.Errorf("federation ConfigMap: %w", err)
+	}
+	if err := r.deployMaasConsumerPortalFederationConfigMap(ctx, dashboard, nextStatuses); err != nil {
+		r.markMaasConsumerPortalFederationConfigMapFailed(cm, err)
+		logger.Error(err, "Failed to deploy MaaS Consumer Portal federation ConfigMap")
 	}
 
 	if err := r.patchDeploymentFederationHash(ctx, fedData); err != nil {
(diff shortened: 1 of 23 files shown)


dashboard-operator/internal/controller/dashboard_reconciler.go
```diff
@@ -208,14 +207,27 @@ func (r *DashboardReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+		if err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to reconcile MaaS Consumer Portal-required modules: %w", err)
+		}
+		preserveModuleStatusTransitionTimes(dashboard.Status.ModuleStatuses, nextStatuses)
+		dashboard.Status.ModuleStatuses = nextStatuses
+		r.setMaasConsumerPortalModuleCondition(cm, dashboard, nextStatuses)
+		if err := r.deployMaasConsumerPortalFederationConfigMap(ctx, dashboard, nextStatuses); err != nil {
+			r.markMaasConsumerPortalFederationConfigMapFailed(cm, err)
+			logger.Error(err, "Failed to deploy MaaS Consumer Portal federation ConfigMap")
+		}
```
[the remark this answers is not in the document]
caponetto: Skip: portal federation ConfigMap failures are intentionally non-blocking. The failure is recorded as MaasConsumerPortalAvailable=False and persisted through the normal status update in both branches, while core deployment/teardown continues. Returning the error would make a portal-only failure block core reconciliation, conflicting with the portal’s independent-operand design.

dashboard-operator/internal/controller/maas_consumer_portal_module_deploy_test.go
```diff
@@ -0,0 +1,100 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+
+	"github.com/opendatahub-io/odh-platform-utilities/pkg/cluster"
+
+	v1alpha1 "github.com/opendatahub-io/odh-dashboard/dashboard-operator/api/v1alpha1"
+	ctrlpkg "github.com/opendatahub-io/odh-dashboard/dashboard-operator/internal/controller"
+)
+
+func TestBuildMaasConsumerPortalFederationConfigMap(t *testing.T) {
```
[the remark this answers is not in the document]
caponetto: Skip: these tests cover different units and behaviors—the ConfigMap builder versus reconciliation-time deletion. Combining them into one table-driven test would reduce clarity without improving coverage.

dashboard-operator/internal/controller/maas_consumer_portal_federation.go
```diff
@@ -0,0 +1,138 @@
...
+	entries := make([]federationEntry, 0, 2)
+	for _, name := range maasConsumerPortalRequiredModuleNames() {
+		mod := moduleRegistry[name]
+		status := statuses[name]
+		if status.Phase != v1alpha1.ModulePhaseDeployed && status.Phase != v1alpha1.ModulePhaseDegraded {
+			continue
+		}
+		entries = append(entries, r.moduleFederationEntry(name, mod))
+	}
+	data, err := json.MarshalIndent(entries, "    ", "  ")
```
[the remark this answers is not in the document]
caponetto: Skip: the portal and core federation ConfigMaps currently use the same json.MarshalIndent(entries, "    ", "  ") format. There is no formatting divergence, and consumers parse the JSON rather than rely on its whitespace.

caponetto:
@lucferbux this is ready for review, though it must wait for https://github.com/opendatahub-io/odh-dashboard/pull/9562 and https://github.com/opendatahub-io/odh-dashboard/pull/9581 to land first

/hold until https://github.com/opendatahub-io/odh-dashboard/pull/9562 and https://github.com/opendatahub-io/odh-dashboard/pull/9581 land first

dashboard-operator/internal/controller/maas_consumer_portal_federation.go
```diff
@@ -0,0 +1,138 @@
...
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/opendatahub-io/odh-platform-utilities/pkg/deploy"
+	"github.com/opendatahub-io/odh-platform-utilities/pkg/metadata/labels"
+
+	v1alpha1 "github.com/opendatahub-io/odh-dashboard/dashboard-operator/api/v1alpha1"
+)
+
+const maasConsumerPortalFederationConfigMapName = "maas-consumer-portal-federation-config"
```
lucferbux, line 22: The deterministic build around this (sorted module names, stable sha256, hash annotation on the pod template) is exactly right — no map-ordering churn on the rollout. One cross-PR seam worth calling out: this ConfigMap name (`maas-consumer-portal-federation-config`) is the producer side of #9581's consumer, where the portal Deployment mounts a ConfigMap whose name comes from an empty `portal-federation-config` param. They have to resolve to the same name or Core-BFF gets no remotes config. Since these are landing as stacked PRs, worth a note on both sides so the value doesn't drift.

dashboard-operator/internal/controller/maas_consumer_portal.go
```diff
@@ -0,0 +1,55 @@
...
+) {
+	if dashboard.Spec.MaasConsumerPortal == nil || dashboard.Spec.MaasConsumerPortal.ManagementState != "Managed" {
+		return
+	}
+	if maasConsumerPortalUnavailable(cm) {
+		return
+	}
+	for _, name := range maasConsumerPortalRequiredModuleNames() {
+		status := statuses[name]
+		if status.Phase == v1alpha1.ModulePhaseDeployed {
```
lucferbux, line 36: Minor maintainability point: this availability check treats only `Deployed` as healthy, but the preserve set (`maas_consumer_portal_federation.go:46`) and the federation entries (`:92`) both treat `Deployed || Degraded` as in-play. So a `Degraded` required module gets kept during teardown and wired into the federation config, yet reported here as `RequiredModuleUnavailable`. It's internally consistent and defensible, but the intent is split across three sites — a future edit to one predicate won't obviously flag the other two. Might be worth extracting `modulePresent()` (Deployed||Degraded) vs `moduleHealthy()` (Deployed) helpers, or at least a comment noting the deliberate asymmetry. Not blocking.

lucferbux (commented):
Traced the orchestration across all four core/portal Managed|Removed combinations and the hard parts hold up: union/OR module demand, explicit-`Disabled` override precedence, first-failure-wins on the availability condition (with recovery), the gen-ai slug-vs-name preserve matching, and the federation ConfigMap surviving core teardown via its own `part-of` label. Federation output is deterministic (sorted names, stable sha256) with the rollout hash landing on the pod template, the ConfigMap carries an owner ref for GC, and the NetworkPolicy ports aren't swapped (gen-ai 8143 / maas 8243, portal podSelector appended without dropping the existing sources). Solid work.

Two inline notes — the federation-config name seam with #9581, and a maintainability point on the module-phase predicates. One very-low-confidence aside: the Pass-1 reorder in `resolveModuleStatuses` flips the reported `Reason` for a module that's both explicitly Disabled and missing its DSC component (`ComponentNotAvailable` → `ExplicitOverride`); it ends disabled either way and `ExplicitOverride` is arguably more accurate, so likely intentional — just noting it. Leaving as comments.

lucferbux (commented):
Follow-up after re-reading #9625 and #9581 together — two things my first pass didn't cover: the manifestInfo/path divergence with #9581 (distinct from the federation-config name seam), and a module-GC behavior change. Both inline, neither blocking.

dashboard-operator/internal/controller/maas_consumer_portal_consolelink.go
```diff
@@ -0,0 +1,225 @@
...
+
+// maasConsumerPortalHostPrefix is prepended to Gateway.Domain to derive the
+// MaaS Consumer Portal host (e.g. maas-consumer-portal.<domain>).
+const maasConsumerPortalHostPrefix = "maas-consumer-portal"
+
+// maasConsumerPortalConsoleLinkManifestInfo points at the portal ConsoleLink
+// bundle. It always uses the /rhoai source: the portal is an RHOAI feature
+// and deploys on both self-managed and managed RHOAI when enabled, so it is
+// intentionally not gated by the platformPaths /not-supported overlay.
+func maasConsumerPortalConsoleLinkManifestInfo(basePath string) render.ManifestInfo {
```
lucferbux, line 51: Cross-PR divergence with #9581 (also open against `main`, not stacked): this re-adds `maasConsumerPortalConsoleLinkManifestInfo` pointing at `maas-consumer-portal-consolelink/rhoai`, but #9581 renames it to `maasConsumerPortalManifestInfo`, moves the bundle to `distributions/maas-consumer-portal/`, and deletes this directory outright. This is a different issue from the federation-config name seam — it's the render source path itself. Whoever merges second needs a semantic rebase (reconcile the fn name + render path), and the integration test's `writeMaasConsumerPortalManifest` helper writes to the same old path, so it needs the same fix. Worth coordinating merge order explicitly.

dashboard-operator/internal/controller/module_deploy.go
```diff
@@ -596,6 +600,22 @@ func (r *DashboardReconciler) deployFederationConfigMap(
...
+// lifecycles. Shared modules retain the dashboard ownership label because they
+// are common dependencies rather than resources owned by a single operand.
+func (r *DashboardReconciler) reconcileModuleDemand(ctx context.Context, dashboard *v1alpha1.Dashboard) (map[string]v1alpha1.ModuleStatus, error) {
+	statuses := resolveModuleStatuses(&dashboard.Spec)
+	if err := r.deployModuleManifests(ctx, dashboard, statuses); err != nil {
+		return nil, err
+	}
+	if err := r.deleteModuleResources(ctx, statuses); err != nil {
+		log.FromContext(ctx).Error(err, "Failed to clean up disabled module resources")
+		return nil, err
```
lucferbux, line 613: Behavior change worth confirming is intentional: `deleteModuleResources` used to be best-effort in the old `reconcileDeployment` (error logged and swallowed), but `reconcileModuleDemand` now returns it, so a module-GC failure fails the whole reconcile (`ModuleDeployFailed`) — including the Removed path, where a stuck delete now blocks teardown instead of letting the rest proceed. Probably the right call since it surfaces stuck GC rather than hiding it, just flagging the semantics change. Minor: it's also double-logged (here and again at the caller).

caponetto:
Superseded by https://github.com/opendatahub-io/odh-dashboard/pull/9637

PR #9637, opened 2026-09-04
feat: deploy MaaS Consumer Portal through Dashboard Module Controller

caponetto:
[https://redhat.atlassian.net/browse/RHOAIENG-90191](https://redhat.atlassian.net/browse/RHOAIENG-90191)
[https://redhat.atlassian.net/browse/RHOAIENG-90190](https://redhat.atlassian.net/browse/RHOAIENG-90190)
[https://redhat.atlassian.net/browse/RHOAIENG-90189](https://redhat.atlassian.net/browse/RHOAIENG-90189)
[https://redhat.atlassian.net/browse/RHOAIENG-90166](https://redhat.atlassian.net/browse/RHOAIENG-90166)

## Description

Implements MaaS Consumer Portal deployment through the Dashboard Module Controller with a platform-managed nested portal hostname. The controller can now deploy, observe, update, and remove the MaaS Consumer Portal independently of the main dashboard while MaaS and GenAI BFFs are reconciled from aggregate demand.

This PR supersedes the following open, stacked PRs, whose changes are included in this branch:

- #9562 — packages the portal bundle in the shared `odh-core-bff` image and hardens Core-BFF proxy validation.
- #9581 — adds the portal operand manifest bundle.
- #9625 — adds aggregate MaaS/GenAI module demand and portal federation configuration.

### Included work

- Builds the portal production bundle into `odh-core-bff` at `/static/maas-consumer-portal`; no portal-specific image, related-image variable, or promotion path is introduced.
- Uses the gateway-supported `/oauth2/sign_out` endpoint and validates Core-BFF module-proxy service ports and paths.
- Adds the complete portal distribution: one-replica standalone Core-BFF Deployment, Service, ServiceAccount, TLS serving certificate, least-privilege RBAC, NetworkPolicy, HTTPRoute, ConsoleLink, parameters, and Kustomization.
- Runs the portal as non-root and read-only, with `/tmp` storage reserved at its `100Mi` ceiling, projected service-account credentials, trusted CA mounts, HTTPS probes, and explicit OpenShift/namespace/static-assets/federation-file arguments.
- Reconciles MaaS and GenAI BFFs when either the main dashboard or portal requires them; preserves explicit module disables, shares one BFF deployment/service pair, and removes shared workloads only when neither operand needs them.
- Generates deterministic portal-only federation JSON from the module registry, labels it as portal-owned, and rolls the portal Deployment when it changes.
- Reconciles the full portal lifecycle independently of the core dashboard state, reports portal URL and availability, preserves the last-known-good URL during transient failures, and removes portal-owned resources—including the service-CA Secret—on portal removal or Dashboard deletion.
- Adds ownership watches for portal resources and Gateway API registration.

The controller derives the nested portal hostname from the gateway domain. The platform operator must provision its DNS, certificate, and gateway listener/Route support, and validate the shared gateway authentication flow for that host.

## How Has This Been Tested?

Automated checks run on this rebased branch:

- `make test` from `dashboard-operator/` — passed (format, vet, race-enabled Go tests, and coverage).
- `make lint` from `dashboard-operator/` — passed with 0 issues.
- `npm run lint` from `distributions/core-bff/` — passed after synchronizing the nested frontend dependencies.
- Focused portal manifest rendering and controller tests — passed during implementation.

Repository-wide checks were also run. They remain blocked by unrelated Eval Hub failures:

- `npm run type-check` fails in `packages/eval-hub/frontend/src/app/hooks/useEvaluationJobDetailPolling.ts`.
- `npm run lint` fails on existing Eval Hub Prettier errors.
- `npm run test-unit` fails in existing Eval Hub polling and Evaluations Page tests.

Manual tests

## Test Impact

Adds and updates dashboard-operator unit and envtest coverage for:

- portal manifest rendering and parameter substitutions;
- portal deployment, route, health, URL, status, update, and removal behavior;
- portal-only, core-only, dual-managed, and both-removed transitions;
- aggregate MaaS/GenAI module demand, explicit-disable handling, deterministic federation JSON, and rollout hashing;
- portal resource cleanup and ownership watches.

Adds Core-BFF module-proxy validation tests and portal logout coverage.

## Request review criteria:

## Summary by CodeRabbit

* **New Features**
  * Added independent MaaS Consumer Portal deployment with HTTPS routing, health checks, secure runtime settings, and access controls.
  * Added federation with MaaS and GenAI modules.
  * Dashboard status now reports the last known reachable portal URL.
  * Portal sign-out redirects through the OAuth sign-out endpoint.
* **Bug Fixes**
  * Improved proxy validation for invalid ports, paths, and route collisions.
  * Preserved required module status during dashboard removal and portal teardown.
* **Documentation**
  * Updated Dashboard Operator documentation for the independent portal operand and status fields.


dashboard-operator/internal/controller/maas_consumer_portal.go
```diff
@@ -0,0 +1,355 @@
...
+	for i := range rendered {
+		if rendered[i].GetKind() != "ConfigMap" || rendered[i].GetName() != "maas-consumer-portal-params" {
+			resources = append(resources, rendered[i])
+		}
+	}
+	if err := deploy.NewDeployer(deploy.WithFieldOwner("dashboard-operator"), deploy.WithLabel(labels.PlatformPartOf, maasConsumerPortalPartOf), deploy.WithApplyOrder()).Deploy(ctx, deploy.DeployInput{Client: r.Client, Owner: dashboard, Release: deploy.ReleaseInfo{Type: string(r.Platform)}, Resources: resources}); err != nil {
+		return fmt.Errorf("deploying MaaS Consumer Portal bundle: %w", err)
+	}
+	cm := &corev1.ConfigMap{}
+	if err := r.Get(ctx, client.ObjectKey{Name: maasConsumerPortalFederationConfigMapName, Namespace: r.ApplicationsNamespace}, cm); err != nil {
```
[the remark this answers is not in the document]
caponetto: Good catch. I added the IsNotFound guard so a temporarily unavailable federation ConfigMap defers only the rollout-hash patch, which is applied on the next reconcile. The missing-ConfigMap path is now covered by a unit test.

dashboard-operator/internal/controller/maas_consumer_portal.go
```diff
@@ -0,0 +1,355 @@
...
+	if !ok {
+		cm.MarkFalse(conditionMaaSConsumerPortalAvailable, conditions.WithReason("MaaSConsumerPortalDomainRequired"), conditions.WithMessage("MaaS Consumer Portal is enabled but gateway domain is not set"))
+		return 0
+	}
+	if err := r.deployMaaSConsumerPortalBundle(ctx, dashboard, url); err != nil {
+		// The module and federation steps run before the bundle. Preserve their
+		// specific failure conditions instead of replacing them with a generic
+		// bundle-apply failure, while still applying the portal's desired bundle.
+		if maasConsumerPortalUnavailable(cm) {
+			return maasConsumerPortalRetryInterval
```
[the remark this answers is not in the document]
caponetto: Good catch. I added an error log when bundle deployment fails while a prior portal condition takes precedence, so retries remain diagnosable without replacing the more specific condition.

dashboard-operator/internal/controller/maas_consumer_portal.go
```diff
@@ -0,0 +1,355 @@
...
+	}
+	return false
+}
+
+func (r *DashboardReconciler) deployMaaSConsumerPortalBundle(ctx context.Context, dashboard *v1alpha1.Dashboard, url string) error {
+	m := maasConsumerPortalManifestInfo(r.ManifestsBasePath)
+	params := readExistingParams(filepath.Join(m.String(), "params.env"))
+	maps.Copy(params, resolveImageParams())
+	params["dashboard-namespace"] = r.ApplicationsNamespace
+	params["gateway-name"] = "data-science-gateway"
```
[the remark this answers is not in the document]
caponetto: Dashboard.spec.gateway currently configures only the domain; the Gateway resource remains the platform-standard data-science-gateway. I made that dependency explicit in the route-unavailable condition so a misconfigured or unavailable Gateway is diagnosable. Making its name configurable would require extending the Dashboard API and platform contract.

manifests/distributions/maas-consumer-portal/networkpolicy.yaml
```diff
@@ -0,0 +1,49 @@
...
+    # endpoint address is platform-managed and has no stable label selector;
+    # this is restricted to its HTTPS control-plane port.
+    - to:
+        - ipBlock:
+            cidr: 0.0.0.0/0
+      ports:
+        - protocol: TCP
+          port: 6443
+    - to:
+        - podSelector:
```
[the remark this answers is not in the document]
caponetto: Implemented. MaaS and GenAI egress peers now pair their pod selectors with a kubernetes.io/metadata.name namespace selector, populated from the injected dashboard namespace.

distributions/maas-consumer-portal/src/components/__tests__/UserDropdown.spec.tsx
```diff
@@ -0,0 +1,15 @@
+jest.mock('mod-arch-core', () => ({
```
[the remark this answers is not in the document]
caponetto: Done.

dashboard-operator/internal/controller/maas_consumer_portal_integration_test.go
```diff
@@ -0,0 +1,436 @@
...
+	for i := range dashboard.Status.Conditions {
+		if dashboard.Status.Conditions[i].Type == conditionType {
+			return dashboard.Status.Conditions[i].Reason
+		}
+	}
+
+	return ""
+}
+
+func TestIntegration_MaaSConsumerPortalConsoleLink(t *testing.T) {
```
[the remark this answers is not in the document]
caponetto: Skipped. The finding is a non-functional refactor, and the tests do not share a single lifecycle shape: the main test verifies create → ready → update → apply failure → removal, while the two smaller tests verify distinct teardown combinations. Combining them into a table would add indirection and obscure those flows without reducing meaningful duplication.

lucferbux (commented):
Nice — this pulls #9562 / #9581 / #9625 together cleanly, and the three cross-PR seams I was worried about are all reconciled: the old `maas-consumer-portal-consolelink` manifest dir is gone (consolelink folded into the distributions bundle), `maasConsumerPortalManifestInfo` now points at `distributions/maas-consumer-portal`, the federation-config name is consistent producer → params.env → volume, and the Dockerfile chmod moved to the new params.env path. No dangling refs to the deleted paths.

The big calls from the design doc landed the way we agreed:
- single core-bff image with both FEs baked in, selected via `--static-assets-dir` (the `test -s .../index.html` build guard is a nice touch)
- RHOAI-only gate, ODH reports UnsupportedPlatform + cleans up
- portal-only readiness when core is Removed (shared BFFs preserved via component label), portal-owned federation config + serving-cert surviving core teardown
- least-priv RBAC (SAR create / DSC get,list / ingresses get)

One thing I want on record: we shipped **sidecar-free** — core-bff in `user_token` mode reading `x-forwarded-access-token` off the gateway, no per-pod kube-rbac-proxy. That reverses my earlier note in the doc where I said kube-rbac-proxy was the only way to translate the session token to a k8s user; the core-bff SAR path does the same thing, so I'm fine with it — I just want it explicit that this is the deliberate landing, and that data-science-gateway is guaranteed to inject (and sanitize) that header. Left an inline on it.

Rest is minor — a couple of leftover multi-task comments to scrub, the positional kustomize arg, and confirming the federation mount vs the doc. Not blocking. Holding my approval until the auth-model confirmation, but this is in good shape.

dashboard-operator/internal/controller/maas_consumer_portal_federation.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+	return requiredModules
+}
+
+// patchMaaSConsumerPortalDeploymentFederationHash triggers a rollout only when
+// the MaaS Consumer Portal remote configuration changes. Task 2/4 own creation
+// of this Deployment, so an absent MaaS Consumer Portal workload is an expected no-op.
+func (r *DashboardReconciler) patchMaaSConsumerPortalDeploymentFederationHash(ctx context.Context, configData string) error {
+	var deployment appsv1.Deployment
+	key := client.ObjectKey{Name: "maas-consumer-portal", Namespace: r.ApplicationsNamespace}
```
lucferbux: Two small things here:
- `Task 2/4 own creation of this Deployment` is leftover framing from when this was split across PRs — now that it's all one PR, worth rewording to just explain the no-op guard.
- `client.ObjectKey{Name: "maas-consumer-portal"}` hardcodes the deployment name where the rest of the bundle repeats that string too — worth a shared const so it can't drift from the manifests.

manifests/distributions/maas-consumer-portal/kustomization.yaml
```diff
@@ -0,0 +1,115 @@
...
+        fieldPaths:
+          - subjects.0.namespace
+      - select:
+          kind: Deployment
+          name: maas-consumer-portal
+        # Kustomize cannot select a scalar argument by flag name. Keep
+        # --namespace=dashboard-namespace at args.5; the manifest rendering
+        # test verifies the resulting argument.
+        fieldPaths:
+          - spec.template.spec.containers.[name=maas-consumer-portal].args.5
```
lucferbux, line 102: Positional `args.5` injection is a sharp edge — reorder the container args and this silently rewrites the wrong flag. I see the comment + the manifest rendering test guard it, so it's covered; just calling it out. (Kustomize genuinely can't select by flag name, so not much choice here.)

manifests/distributions/maas-consumer-portal/deployment.yaml
```diff
@@ -0,0 +1,153 @@
...
+              subPath: odh-ca-bundle.crt
+              readOnly: true
+            - name: portal-sa-token
+              mountPath: /var/run/secrets/kubernetes.io/serviceaccount
+              readOnly: true
+          args:
+            - --deployment-mode=standalone
+            - --auth-method=user_token
+            - --auth-token-header=x-forwarded-access-token
+            - --auth-token-prefix=
```
lucferbux, line 146: This is the sidecar-free landing — core-bff reading `x-forwarded-access-token` directly instead of a per-pod kube-rbac-proxy. It settles the doc back-and-forth in favor of the core-bff `user_token` + SAR path, which I'm on board with. One thing to confirm: data-science-gateway always injects `x-forwarded-access-token` on this route and strips any client-supplied one, so we're not trusting a spoofable header. If that's guaranteed at the gateway, good to go.

manifests/distributions/maas-consumer-portal/deployment.yaml
```diff
@@ -0,0 +1,153 @@
...
+                      path: service-ca.crt
+        - name: portal-tls
+          secret:
+            secretName: maas-consumer-portal-tls
+        - name: maas-consumer-portal-federation-config
+          configMap:
+            name: maas-consumer-portal-federation-config
+            items:
+              - key: module-federation-config.json
+                path: maas-consumer-portal-federation-config.json
```
lucferbux, line 57: This mount isn't `optional`, while the adjacent CA mounts are. The design note leaned optional, but required actually seems more correct — the BFF can't come up without the remotes file — so this just relies on the operator creating the CM before the Deployment (which it does). Fine to keep required, flagging the delta from the doc.

manifests/distributions/maas-consumer-portal/params.env
```diff
@@ -0,0 +1,12 @@
+# Injected by the Dashboard Module Controller. maas-consumer-portal-hostname and
+# maas-consumer-portal-url are independently resolved by the selected URL model.
```
lucferbux: nit: "resolved by the selected URL model" reads like leftover design-doc language (same with `URL-model-specific` in the federation builder comment). Now that it's just subdomain-based, worth making these concrete or dropping them.

dashboard-operator/internal/controller/module_deploy_test.go
```diff
@@ -66,6 +71,77 @@ func allDeployedStatuses() map[string]v1alpha1.ModuleStatus {
...
+		MaaSConsumerPortal: &v1alpha1.MaaSConsumerPortalSpec{ManagementState: "Removed"},
+	}}
+
+	statuses, err := reconciler.ReconcileModuleDemand(context.Background(), dashboard)
+	require.NoError(t, err)
+	for module, status := range statuses {
+		assert.Equalf(t, v1alpha1.ModulePhaseNotDeployed, status.Phase, "%s should not be deployed", module)
+		assert.Equalf(t, "NotRequired", status.Reason, "%s should be marked not required", module)
+	}
+}
```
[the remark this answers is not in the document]
caponetto: Skipped. This is a style-only refactor: the cited tests exercise distinct setup, error injection, and lifecycle assertions. Combining them into broad tables would increase callback-based indirection without improving coverage or behavior validation.

dashboard-operator/internal/controller/support_test.go
```diff
@@ -84,16 +84,14 @@ func TestComputeKustomizeVariables(t *testing.T) {
...
-	// regardless of platform (it deploys on both self-managed and managed).
-	info := maasConsumerPortalConsoleLinkManifestInfo("/base")
+func TestMaaSConsumerPortalManifestInfo(t *testing.T) {
+	info := maasConsumerPortalManifestInfo("/base")
 	assert.Equal(t, "/base", info.Path)
-	assert.Equal(t, "maas-consumer-portal-consolelink", info.ContextDir)
-	assert.Equal(t, "/rhoai", info.SourcePath)
+	assert.Equal(t, "distributions", info.ContextDir)
+	assert.Equal(t, "maas-consumer-portal", info.SourcePath)
 }
```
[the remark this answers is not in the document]
caponetto: Skipped. The test has one deterministic input and one invariant result; adding a one-row table would not add coverage or make future distribution cases easier to express.

manifests/distributions/maas-consumer-portal/httproute.yaml
```diff
@@ -0,0 +1,19 @@
...
+  hostnames:
+    - maas-consumer-portal-hostname
+  rules:
+    - matches:
+        - path:
+            type: PathPrefix
+            value: /
+      backendRefs:
+        - name: maas-consumer-portal
+          port: 8443
```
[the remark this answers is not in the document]
caponetto: Skipped. This is not a safe standalone manifest change: a BackendTLSPolicy requires a same-namespace CA ConfigMap plus a hostname matching the service certificate, and it also requires the cluster’s Gateway implementation/version to support that policy API. Neither trust source nor compatibility contract is defined in this bundle today. Gateway API TLS requirements (https://gateway-api.sigs.k8s.io/guides/user-guides/tls/), OpenShift re-encrypt requirements (https://docs.redhat.com/en/documentation/openshift_container_platform/4.22/html/ingress_and_load_balancing/configuring-gateway-api)

Adding it correctly would need a platform-level design covering CA injection, API/RBAC availability, lifecycle cleanup, and supported OCP versions.

manifests/distributions/maas-consumer-portal/kustomization.yaml
caponetto: Agreed. The positional replacement is unavoidable with Kustomize; the inline comment and render test guard the expected --namespace= result, so no change needed.

manifests/distributions/maas-consumer-portal/params.env
caponetto: Updated both comments to use concrete portal terminology.

dashboard-operator/internal/controller/maas_consumer_portal_federation.go
caponetto: Addressed: removed the stale split-work wording and centralized the portal Deployment name in a shared constant to prevent controller-side drift.

manifests/distributions/maas-consumer-portal/deployment.yaml
caponetto: Thanks for pointing this out. I'll double-check it!

manifests/distributions/maas-consumer-portal/cluster-role.yaml
```diff
@@ -0,0 +1,24 @@
...
+kind: ClusterRole
+metadata:
+  name: maas-consumer-portal
+rules:
+  - apiGroups:
+      - authorization.k8s.io
+    resources:
+      - subjectaccessreviews
+    verbs:
+      - create
```
lucferbux: Tied to the auth-model discussion: this grants privileged SARs (checking access *as the SA*), but the portal never does that — every authz check is a `SelfSubjectAccessReview` run with the user's token (`token_k8s_client.go`), which doesn't consume the SA's RBAC at all. I grepped the whole core-bff binary: only `SelfSubjectAccessReviews().Create`, never `SubjectAccessReviews().Create`. So this rule looks like dead weight — drop it for least-privilege unless there's a planned SA-level check I'm missing. (The `datascienceclusters` / `ingresses` reads below are the legit SA-scoped ones.)

lucferbux (commented):
Went back through the auth path after the kube-rbac-proxy question — and first I want to correct my last review, where I called this the "core-bff `user_token` + SAR path." That was sloppy wording. What the portal actually does in `user_token` mode is the fastify-parity passthrough we want, not SA + SSAR:

- `ExtractRequestIdentity` reads `x-forwarded-access-token` (`factory.go`), and `NewTokenKubernetesClient` builds the k8s client with `cfg.BearerToken = <user token>` and clears the SA token file (`token_k8s_client.go:90-95`). Every k8s call runs **as the user**.
- The `/api/k8s/` proxy sets `Authorization: Bearer <user token>` outbound and strips the inbound `x-forwarded-access-token` / cookies / `x-forwarded-*` (`k8s_proxy.go` + `SensitiveIngressHeaders`).
- Every authz check is a `SelfSubjectAccessReview` evaluated **as the user** — there's no privileged `SubjectAccessReviews().Create` in the whole binary. The code even documents "Parity: mirrors Fastify's adminUtils.ts isUserAdmin()".

That's the same thing `backend/` does — `userUtils.ts` pulls `x-forwarded-access-token` and calls k8s with `Bearer <accessToken>`. So we're mimicking the fastify backend correctly here. The only small divergence: the backend prefers the kube-rbac-proxy `x-auth-request-user` header for the username (4.19+) and falls back to `SelfSubjectReview`, whereas the portal always resolves via `SelfSubjectReview` — functionally equivalent, no concern.

On the kube-rbac-proxy itself: the portal is sidecar-free, so it isn't running its own kube-rbac-proxy — it consumes the `x-forwarded-access-token` the data-science-gateway injects, the same header the backend consumes. That's fine as long as the gateway is the one injecting it and stripping any client-supplied copy (the inbound trust boundary just moves from a per-pod sidecar to the shared gateway). Still the one thing worth confirming, but the app-side handling is right.

One concrete follow-on: since we're passthrough and never SA + SAR, the SA doesn't need `subjectaccessreviews: create` — flagged inline.

manifests/distributions/maas-consumer-portal/cluster-role.yaml
caponetto: Removed the unused subjectaccessreviews.create permission. The portal service account now retains only the DSC and ingress reads it uses.

dashboard-operator/internal/controller/module_deploy_test.go
```diff
@@ -68,6 +72,77 @@ func allDeployedStatuses() map[string]v1alpha1.ModuleStatus {
 	return statuses
 }
 
+func TestReconcileModuleDemand_WhenNeitherOperandRequiresModules(t *testing.T) {
```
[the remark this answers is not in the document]
caponetto: Skipped. The cases have materially different setup and assertions—normal no-demand resolution, resource cleanup, injected deletion failure, and readiness overlay. Combining them would add indirection without improving coverage or behavior.

caponetto:
@lucferbux I have manually tested this code, and overall it looks good to go in. Here's a status:

### Currently Working

- The MaaS Consumer Portal is enabled and reports `MaaSConsumerPortalAvailable=True`.
- The portal Deployment is ready and uses the intended Core-BFF image.
- The portal Service, serving TLS Secret, ServiceAccount and RBAC, NetworkPolicy, federation ConfigMap, HTTPRoute, and ConsoleLink are created.
- The HTTPRoute is accepted and its backend reference resolves.
- The portal URL is derived from the configured gateway domain: `https://maas-consumer-portal.<gateway-domain>/`
- The federation ConfigMap contains the MaaS and GenAI remote definitions, and both target Services have live endpoints.
- The running Core-BFF pod contains the portal static assets and mounted federation configuration. It is configured for `user_token` authentication using `x-forwarded-access-token`.
- A temporary direct OpenShift Route can serve the portal UI for visual inspection.
- Portal teardown is verified: Removed deletes portal resources and reports Disabled; the protected ServiceAccount remains.


### Remaining Platform Work

Platform must make the derived portal hostname a supported, authenticated gateway host:

- Provision DNS for `maas-consumer-portal.<gateway-domain>`.
- Provide TLS certificate coverage for that nested hostname.
- Configure the gateway listener or Route to forward that hostname to the portal HTTPRoute.
- Configure OAuth to support login, callback, and session handling on the portal
hostname.
- Inject a trusted `x-forwarded-access-token` for portal requests and remove any client-provided copy before forwarding.



### End-to-End Limitation

The portal's MaaS and GenAI proxy targets are configured and reachable in-cluster, but authenticated proxy calls cannot yet be verified end to end. The temporary direct Route bypasses gateway authentication, so Core-BFF correctly rejects its API requests when `x-forwarded-access-token` is absent. The supported nested hostname currently lacks the required Platform DNS, TLS, and gateway OAuth configuration.

lucferbux:
I tested this end-to-end on a live ODH cluster (ROSA) — built the operator plus a core-bff image carrying the portal bundle from this branch, deployed with the new CRD, and enabled `spec.maasConsumerPortal.managementState: Managed`.

**The operator side works well.** The portal Deployment goes Available, the HTTPRoute is `Accepted`/`ResolvedRefs`, the portal-specific federation ConfigMap correctly carries only `maas` + `genAi`, `status.maasConsumerPortalUrl` and the `MaaSConsumerPortalAvailable` condition populate, and the ConsoleLink is created. The bundle serves correctly (verified via port-forward: `/` → 200, `/app.js` → 200, `/healthcheck` OK, `/api/v1/config` → 401 without a token). Nice work — federation is already handled here, no extra operator section needed.

**One important integration finding:** clicking the ConsoleLink lands the user on the **main dashboard**, not the portal. Routing is actually correct — the gateway has a dedicated Envoy virtual host for the portal FQDN pointing at the portal service. The blocker is **auth**: the shared `kube-auth-proxy` (oauth2-proxy) fronting `data-science-gateway` is anchored to the base host (`--redirect-url` / `--cookie-domain` = base host, no `--whitelist-domain`), and the original subdomain is reduced to `rd=/` **before** the OAuth round-trip — so after login the callback runs on the base host and redirects to `/` there = the dashboard. This is a platform/gateway concern (opendatahub-operator `GatewayConfig`), **not** the operator code in this PR.

It also surfaces a **design question**: the portal is the only gateway service on a separate subdomain — dashboard, mlflow, and model-catalog are all paths on the single base host. That one choice drives three gaps (DNS, TLS cert, and this auth round-trip).

Full write-up (verification details, root-cause diagnosis, and a subdomain-vs-path options analysis): https://docs.google.com/document/d/1NWEMSRfPcpa3LpDNUW0ZpZNBfw0N5RFcXQL1dFdh7wg/edit?usp=sharing

I've left my `/lgtm` + `/approve` in place since the PR is correct for its scope, but we should sync on the gateway/auth integration (and the subdomain-vs-path decision) before it's usable end-to-end. cc @caponetto — let's talk.

PR #9627, opened 2026-09-03
Add NIM properties to model deployment tracking events

mturley:
https://redhat.atlassian.net/browse/RHOAIENG-88015

## Description

Adds NIM-specific properties to the model deployment analytics events and fixes catalog image-name lookup.

The `Model Deployed` and `Model Updated` events now include the NIM image repository/tag, catalog image display name when available, PVC storage mode, and new-storage size/class properties. User-entered deployment names, PVC names, subpaths, and credentials are not included.

`externalData` was added to the tracking-property extension contract and passed through the model-serving tracking collection path because the catalog display name is available only in the NIM wizard field's external data. The wizard state contains the selected repository and tag, but not the catalog metadata needed to derive the safe display name. Passing the existing external-data map keeps the model-serving hub independent of the NIM package while allowing the NIM extension to resolve the catalog entry. The implementation validates the nested `nimImages.images` shape and omits `nimImageName` when the catalog entry is unavailable.

The edit flow emits `Model Updated` rather than `Model Deployed`, including for existing-storage NIM deployments.

## How Has This Been Tested?

- Ran the targeted NIM tracking unit tests:
  - `npm run test-unit --workspace=@odh-dashboard/nim-serving -- --runInBand src/tracking/__tests__/nimWizardTracking.spec.ts`
- Ran ESLint on the changed implementation and test files.
- Ran the NIM package type check:
  - `npm run type-check --workspace=@odh-dashboard/nim-serving`
- Manually validated in the local dashboard with `nimWizard=true` using Playwright:
  - Created a NIM deployment with new storage (`75 GiB`, `gp3-csi`).
  - Edited a deployment using a catalog image that exists.
  - Edited `nim-bad-image` using an image absent from the catalog.
  - Captured all three console telemetry payloads.
  - Confirmed catalog-backed events include `nimImageName: "Active Speaker Detection"`.
  - Confirmed the unavailable-image event omits `nimImageName` while retaining the safe raw image value.
  - Confirmed user-entered deployment/PVC names and subpaths were absent.

Captured browser console output:

```text
[...]

PR #9644, opened 2026-09-04
feat(nim-serving): add NIM segment tracking for deploy and project account flows

YuliaKrimerman:
https://issues.redhat.com/browse/RHOAIENG-88014

## Description

Adds Segment analytics for NIM project settings account enable/replace and remove flows in `packages/nim-serving/`. This PR is scoped to project settings only — it does not include deployment wizard tracking (that work is handled separately in #9627).

**Events fired:**
- `Model Serving NIM Account Enabled` — when a user adds or replaces a personal API key via `NIMApiKeyModal`
- `Model Serving NIM Account Removed` — when a user removes the NIM account via `NIMSettingsCard`

## How Has This Been Tested?

- `npm run lint` — pass (`packages/nim-serving`)
- `npm run type-check` — pass (`packages/nim-serving`)
- `npm run test-unit` — pass, 263 tests (`packages/nim-serving`)

**Manual testing (recommended before merge):**
1. Open a project's NIM settings card
2. Add a personal API key → confirm validation success path completes normally
3. Replace an existing key → confirm revalidation flow completes normally
4. Remove the NIM account → confirm delete polling completes or times out as expected
5. Verify Segment events in browser devtools / analytics debugger (no API key values in payloads)

## Test Impact

**New unit tests:**
- `nimTrackingConstants.spec.ts` — event firing and payload shape
- `useNimAccountEnabledTracking.spec.ts` — enable/replace success, replace pending guard, ref reset on resubmit, API failure
- `NIMApiKeyModal.spec.tsx` — enable success, API failure, replace validation failure, analytics payload safety (no API key leakage)
- `NIMSettingsCard.spec.tsx` — delete polling timeout after exactly 10 refresh attempts

**Updated unit tests:**
- `NIMSettingsCard.spec.tsx` — extended DeleteModal mock to exercise remove confirm flow

No Cypress changes — analytics wiring only; existing project settings UI behavior is unchanged.

## Request review criteria:

## Summary by CodeRabbit

* **New Features**
  * Added analytics tracking for NIM account enablement, replacement, and removal outcomes.
  * Tracks success, validation failures, API failures, deletion failures, and deletion timeouts using non-sensitive categories.
  * Prevents duplicate removal events and excludes API keys and underlying error messages from tracking.

* **Tests**
  * Added coverage for enablement, replacement, deletion, timeout, retry, and duplicate-event scenarios.

### DIFF
--- a/packages/nim-serving/src/pages/projectSettings/NIMApiKeyModal.tsx
+++ b/packages/nim-serving/src/pages/projectSettings/NIMApiKeyModal.tsx
@@ -45,7 +45,7 @@ const NIMApiKeyModal: React.FC<NIMApiKeyModalProps> = ({
   const [submitted, setSubmitted] = React.useState(false);
   const [createError, setCreateError] = React.useState<string>();
   const [submitMode, setSubmitMode] = React.useState(NimAccountEnabledMode.ENABLE);
-  const { resetTrackingState, trackSubmitApiFailure } = useNimAccountEnabledTracking(
+  const { trackSubmitApiFailure } = useNimAccountEnabledTracking(
     submitted,
     accountStatus,
     submitMode,
@@ -55,7 +55,6 @@ const NIMApiKeyModal: React.FC<NIMApiKeyModalProps> = ({
     const trimmedKey = apiKey.trim();
     setIsCreating(true);
     setCreateError(undefined);
-    resetTrackingState();
 
     const accountExists =
       accountStatus !== NIMAccountStatus.NOT_FOUND && accountStatus !== NIMAccountStatus.LOADING;
@@ -78,15 +77,7 @@ const NIMApiKeyModal: React.FC<NIMApiKeyModalProps> = ({
     } finally {
       setIsCreating(false);
     }
-  }, [
-    apiKey,
-    accountStatus,
-    namespace,
-    startRevalidation,
-    refresh,
-    resetTrackingState,
-    trackSubmitApiFailure,
-  ]);
+  }, [apiKey, accountStatus, namespace, startRevalidation, refresh, trackSubmitApiFailure]);
 
   const handleClose = React.useCallback(() => {
     onClose();
--- a/packages/nim-serving/src/pages/projectSettings/NIMApiKeyModal.tsx
+++ b/packages/nim-serving/src/pages/projectSettings/NIMApiKeyModal.tsx
@@ -21,6 +21,8 @@ import {
 import ContentModal from '@odh-dashboard/ui-core/components/ContentModal';
 import { NIMAccountStatus } from '../../api/accounts/hooks';
 import { createNIMResources, createOrReplaceSecret } from '../../api/accounts/api';
+import { NimAccountEnabledMode } from '../../tracking/nimTrackingConstants';
+import { useNimAccountEnabledTracking } from '../../tracking/useNimAccountEnabledTracking';
 
 type NIMApiKeyModalProps = {
   onClose: () => void;
@@ -42,14 +44,23 @@ const NIMApiKeyModal: React.FC<NIMApiKeyModalProps> = ({
   const [isCreating, setIsCreating] = React.useState(false);
   const [submitted, setSubmitted] = React.useState(false);
   const [createError, setCreateError] = React.useState<string>();
+  const [submitMode, setSubmitMode] = React.useState(NimAccountEnabledMode.ENABLE);
+  const { resetTrackingState, trackSubmitApiFailure } = useNimAccountEnabledTracking(
+    submitted,
+    accountStatus,
+    submitMode,
+  );
 
   const handleSubmit = React.useCallback(async () => {
     const trimmedKey = apiKey.trim();
     setIsCreating(true);
     setCreateError(undefined);
+    resetTrackingState();
 
     const accountExists =
       accountStatus !== NIMAccountStatus.NOT_FOUND && accountStatus !== NIMAccountStatus.LOADING;
+    const mode = accountExists ? NimAccountEnabledMode.REPLACE : NimAccountEnabledMode.ENABLE;
+    setSubmitMode(mode);
 
     try {
       if (accountExists) {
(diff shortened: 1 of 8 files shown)


packages/nim-serving/src/pages/projectSettings/NIMApiKeyModal.tsx
```diff
@@ -61,11 +71,13 @@ const NIMApiKeyModal: React.FC<NIMApiKeyModalProps> = ({
       setSubmitted(true);
```
mturley: A retry after a validation failure does not get a fresh tracking state. After the first attempt reaches `ERROR`, `useNimAccountEnabledTracking` sets `hasTrackedSubmitRef.current = true`. The modal keeps `submitted` as `true` when the user submits again, so the hook’s reset effect never runs. As a result, a subsequent retry that transitions through `PENDING` and then `READY` is ignored, and no successful `Model Serving NIM Account Enabled` event is emitted for that attempt.

We should reset the tracking refs at the start of every new submission and possibly add coverage for retrying after a validation failure/API failure and then succeeding.

packages/nim-serving/src/pages/projectSettings/NIMApiKeyModal.tsx
```diff
@@ -61,11 +71,13 @@ const NIMApiKeyModal: React.FC<NIMApiKeyModalProps> = ({
...
       await refresh();
     } catch (e) {
-      setCreateError(e instanceof Error ? e.message : 'Failed to create NIM resources.');
+      const errorMessage = e instanceof Error ? e.message : 'Failed to create NIM resources.';
```
mturley: **Severity: P1**

`refresh()` is awaited in the same `try` block as the resource creation calls. If resource creation succeeds but the subsequent refresh rejects, this catch reports `API_FAILED` and displays a creation error even though the account operation already succeeded. Please separate the resource operation from the refresh and only call `trackSubmitApiFailure` when `createNIMResources` or `createOrReplaceSecret` rejects.

packages/nim-serving/src/pages/projectSettings/NIMApiKeyModal.tsx
```diff
@@ -61,11 +71,13 @@ const NIMApiKeyModal: React.FC<NIMApiKeyModalProps> = ({
...
+      setCreateError(errorMessage);
```
mturley: **Severity: P2**

The raw `Error.message` is rendered through `createError`. API/server error messages can expose internal implementation details or echoed sensitive values. Please use a fixed user-facing message here and keep only the allowlisted failure category in the analytics payload.

packages/nim-serving/src/pages/projectSettings/NIMSettingsCard.tsx
```diff
@@ -76,22 +78,44 @@ const NIMSettingsCard: React.FC<NIMSettingsCardProps> = ({ namespace }) => {
...
     deleteStatusIntervalRef.current = setInterval(async () => {
+      retries -= 1;
       try {
         const result = await refresh();
         if (!result) {
           stopPollingDeleteStatus();
           setIsDeleteModalOpen(false);
-        } else if (retries === 0) {
-          stopPollingDeleteStatus(new Error('NIM resources were not deleted in time.'));
+          fireNimAccountRemoved({
```
mturley: **Severity: P1**

The polling callback is asynchronous, so `clearInterval` cannot cancel a callback that is already awaiting `refresh()`. If another callback is already in flight when this success path stops the interval, that callback can later enter its timeout or error path and emit a second `fireNimAccountRemoved` event for the same deletion. Please add a per-deletion terminal-outcome guard before emitting any removal event.

packages/nim-serving/src/tracking/useNimAccountEnabledTracking.ts
```diff
@@ -0,0 +1,73 @@
...
+} => {
+  const hasTrackedSubmitRef = React.useRef(false);
+  const hasSeenValidationPendingRef = React.useRef(false);
+
+  React.useEffect(() => {
+    if (!submitted) {
+      hasTrackedSubmitRef.current = false;
+      hasSeenValidationPendingRef.current = false;
+    }
+  }, [submitted]);
```
mturley: `submitted` never gets set back to `false`, so when a validation fails and the user retries this doesn't trigger. We need submitted to stay true so errors are shown, so I think instead of resetting the refs here we should pass back some kind of resetTrackingRefs function and have the component call it when the user submits again. Otherwise a failure sets `hasTrackedSubmitRef.current = true;`, a retry doesn't reset it, and the successful retry doesn't fire a tracking event.

packages/nim-serving/src/pages/projectSettings/NIMApiKeyModal.tsx
```diff
@@ -61,11 +72,21 @@ const NIMApiKeyModal: React.FC<NIMApiKeyModalProps> = ({
...
+      trackSubmitApiFailure(mode);
```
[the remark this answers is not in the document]
mturley: @YuliaKrimerman this coderabbit comment is still valid. We need separate try/catch blocks for the submission and the refresh so that a failed refresh after a successful submit doesn’t fire a submit failure event.

packages/nim-serving/src/pages/projectSettings/NIMApiKeyModal.tsx
```diff
@@ -61,11 +72,21 @@ const NIMApiKeyModal: React.FC<NIMApiKeyModalProps> = ({
       setSubmitted(true);
       await refresh();
     } catch (e) {
-      setCreateError(e instanceof Error ? e.message : 'Failed to create NIM resources.');
+      const errorMessage = e instanceof Error ? e.message : 'Failed to create NIM resources.';
+      setCreateError(errorMessage);
```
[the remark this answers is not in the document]
mturley: This coderabbit comment is pre-existing and (I think) invalid. We show raw server error messages to the user all the time. As long as they aren’t being submitted in the tracking event it’s ok to show them in the UI.

packages/nim-serving/src/pages/projectSettings/NIMSettingsCard.tsx
```diff
@@ -86,12 +94,28 @@ const NIMSettingsCard: React.FC<NIMSettingsCardProps> = ({ namespace }) => {
         if (!result) {
           stopPollingDeleteStatus();
           setIsDeleteModalOpen(false);
+          fireNimAccountRemoved({
```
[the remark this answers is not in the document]
mturley: @YuliaKrimerman I think this bot comment is still valid, although IMO not major. We could add a `hasTrackedDeleteRef` as the inline suggestion shows here so that we make sure not to fire duplicate delete events. I don't think it's blocking, though.

mturley (changes requested):
One new thing (`submitted` reset effect issue in `useNimAccountEnabledTracking`) and a couple +1s to bot replies. I don't think the bot ones necessarily need to block the PR though.

packages/nim-serving/src/tracking/useNimAccountEnabledTracking.ts
emilys314, line 1: The 2 useEffects are a little sus so I'd recommend 0 useEffects if possible, but if they work that's fine. It's just it can be harder to predict their behavior and when they will trigger and usually lead to buggier code.  I haven't tested it yet

mturley (approved):
Tested and events are emitted properly.

/approve
/lgtm

@emilys314 at first glance I don’t see a good way around the useEffects since we’re reacting to externally changing state we have to poll for, but do you think we should open a task to take another look at factoring them away?

PR #9658, opened 2026-09-07
ci: extend Konflux build simulator with operator build, CRD/CR reconciliation, and manifest validation

lucferbux:
Jira: [RHOAIENG-87691](https://issues.redhat.com/browse/RHOAIENG-87691)

## Description

Extends the Konflux build simulator (`.github/workflows/pr-build-validation.yml`) so it exercises the `dashboard-operator` and the manifest overlays in addition to the frontend/backend image. This completes the 5-phase testing pyramid the simulator was designed around:

- **Phase 1 — operator image build (`operator-build` job):** Builds `dashboard-operator/Dockerfile` from the repo root (`--build-arg OPERATOR_VERSION=ci-test`) and publishes the image as the `dashboard-operator-image` artifact so Phase 4 can reuse it without rebuilding. Catches Go compilation, missing COPY deps, and FIPS build failures.
- **Phase 4 — operator CRD + CR reconciliation (extends `operator-integration`):** Loads the operator image into the Kind cluster, applies the `Dashboard` CRD (`components.platform.opendatahub.io`) and waits for `Established`, applies the operator ClusterRole (`config/rbac/role.yaml`) plus a ServiceAccount/ClusterRoleBinding, deploys a Kind-friendly operator Deployment (no cert-manager webhook/metrics TLS), creates a minimal `Dashboard` CR, and asserts the controller starts reconciling cleanly — the finalizer (`components.platform.opendatahub.io/cleanup`) and `status.observedGeneration` are set with **0 operator restarts**. This catches controller startup panics, scheme/registration errors, and reconcile crashes that unit + envtest tests can miss on a real API server.
- **Phase 5 — manifest validation (`manifest-validation` job):** Runs `kustomize build` over `manifests/base`, `manifests/odh`, and `manifests/rhoai`, piping each through `kubeconform -strict -ignore-missing-schemas` (K8s 1.31.0). Independent of the Docker build, so manifest regressions fail fast.

Also adds `dashboard-operator/**` to the workflow trigger paths and documents the additions in `docs/KONFLUX_BUILD_SIMULATION.md`.

The chunk-size regression detection item from the ticket was intentionally deferred out of scope for this PR (core + testable scope).

## How Has This Been Tested?

- Validated the workflow YAML parses and all 12 jobs are present (including the new `operator-build` and `manifest-validation` jobs).
- Ran `bash -n` on all extracted `run:` shell blocks — all pass.
- Ran the operator reconciliation flow end-to-end locally in a fresh Kind cluster using the exact scripts from the workflow: the operator started, the `Dashboard` CR reconciled (finalizer + `observedGeneration=1`, 0 restarts) → `✅ PASS`.
- Verified `kustomize build` + `kubeconform` succeeds for `manifests/base`, `manifests/odh`, and `manifests/rhoai` locally.

### CI run on this PR

The simulator ran on this PR and all 12 jobs passed, including the three new additions — run [`34096375157`](https://github.com/opendatahub-io/odh-dashboard/actions/runs/34096375157):

| Job | Result |
|---|---|
| Phase 0: Hermetic Build Preflight | ✅ success |
| Phase 1: Docker Build (ODH mode) | ✅ success |
| Phase 1: Docker Build (RHOAI mode) | ✅ success |
| **Phase 1: Operator Image Build** | ✅ success |
| Phase 2-3: Runtime & Module Federation (ODH) | ✅ success |
| Phase 2-3: Runtime & Module Federation (RHOAI) | ✅ success |
| **Phase 4: Operator Integration** | ✅ success |
| **Phase 5: Manifest Validation** | ✅ success |
| Build Validation Summary | ✅ success |

Phase 4 reconciliation smoke test evidence from the runner:
- Dashboard CRD applied → `condition met` (Established)
- Dashboard CR created; controller added the finalizer and set `observedGeneration: 1`
- `✅ PASS: operator reconciled the Dashboard CR (finalizer + observedGeneration set, 0 restarts)`

As designed, an expected non-fatal error appears in the operator logs (`namespaces "openshift-ingress" not found`) — the operand can't fully provision on Kind without OpenShift, so the test asserts the controller *starts reconciling cleanly* rather than requiring `Ready=True`. The 0-restart check distinguishes "can't provision on Kind" (expected) from "controller is crashing" (a real failure).

## Test Impact

This PR is CI infrastructure — it adds automated validation (operator build, Kind CRD/CR reconciliation smoke test, manifest schema validation) rather than product code. No unit/Cypress tests apply; the new coverage is the workflow jobs themselves, verified locally as described above.

## Request review criteria:

[RHOAIENG-87691]: https://redhat.atlassian.net/browse/RHOAIENG-87691?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

- **Chores**
  - Expanded pull request validation to cover dashboard operator changes.
  - Added automated operator image builds and integration checks for deployment, reconciliation, and restart health.
  - Added Kubernetes manifest rendering and schema validation for supported overlays.
  - Improved validation reporting, security, and job dependencies for clearer build results.
  - Added simulation-only deployment and access configuration for operator smoke tests.

- **Documentation**
  - Updated build simulation guidance with operator validation steps, independent manifest checks, revised timing estimates, and current workflow triggers.


.github/workflows/pr-build-validation.yml
```diff
@@ -322,6 +323,144 @@ jobs:
...
+        run: |
+          set -euo pipefail
+          docker build \
+            --build-arg OPERATOR_VERSION=ci-test \
+            --tag dashboard-operator:sim-test \
+            --file dashboard-operator/Dockerfile \
+            .
+
+      - name: Save operator image
+        run: docker save dashboard-operator:sim-test | gzip > /tmp/dashboard-operator.tar.gz
```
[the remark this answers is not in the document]
lucferbux: Fixed in 72775df177 — added `set -euo pipefail` to the Save operator image step so a `docker save` failure surfaces here instead of uploading a corrupt artifact.

.github/workflows/pr-build-validation.yml
```diff
@@ -322,6 +323,144 @@ jobs:
...
+      K8S_SCHEMA_VERSION: '1.31.0'
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.2.2
+
+      - name: Install kustomize and kubeconform
+        run: |
+          set -euo pipefail
+          echo "::group::Installing kustomize ${KUSTOMIZE_VERSION}"
+          curl -sSL "https://github.com/kubernetes-sigs/kustomize/releases/download/kustomize%2F${KUSTOMIZE_VERSION}/kustomize_${KUSTOMIZE_VERSION}_linux_amd64.tar.gz" \
```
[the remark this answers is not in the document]
lucferbux: Fixed in 72775df177 — kustomize (v5.4.3) and kubeconform (v0.6.7) are now downloaded to disk and verified with `sha256sum --check` against pinned hashes before extraction, matching the Install Kind pattern. Hashes were cross-checked against the projects' published checksum files.

.github/workflows/pr-build-validation.yml
```diff
@@ -869,15 +1020,176 @@ jobs:
...
+                      initialDelaySeconds: 5
+                      periodSeconds: 10
+                    resources:
+                      requests:
+                        cpu: 10m
+                        memory: 128Mi
+          EOF
+
+          echo "Waiting for operator rollout..."
+          kubectl rollout status deployment/dashboard-operator -n opendatahub --timeout=120s
```
antowaddle: kubectl rollout status ... --timeout=120s runs under set -e at the end of the Deploy operator step, but the ::group::Operator logs dump lives in the next step (line 1129), which never runs on failure. The exact failure this phase exists to catch (controller panic, scheme/registration error) surfaces as a bare timed out waiting for the condition — no logs, no describe, no events. Dump inline on failure, or split out an if: failure() diagnostics step.

.github/workflows/pr-build-validation.yml
```diff
@@ -869,15 +1020,176 @@ jobs:
...
+
+      - name: Deploy operator
+        run: |
+          set -euo pipefail
+          # Minimal deployment: the real manager.yaml mounts webhook/metrics cert secrets
+          # (cert-manager) that don't exist in Kind. We run the manager with the webhook
+          # disabled (default) and metrics served over HTTP so no certs are required —
+          # enough to prove the controller starts and reconciles.
+          kubectl apply -f - <<'EOF'
+          apiVersion: apps/v1
```
antowaddle: We should probably create a file with this spec and reference it here.

.github/workflows/pr-build-validation.yml
```diff
@@ -869,15 +1020,176 @@ jobs:
...
+          echo "::group::Applying Dashboard CRD"
+          kubectl apply -f dashboard-operator/config/crd/bases/components.platform.opendatahub.io_dashboards.yaml
+          kubectl wait --for=condition=Established --timeout=60s \
+            crd/dashboards.components.platform.opendatahub.io
+          echo "::endgroup::"
+
+          echo "::group::Applying operator RBAC (ClusterRole + SA + binding in opendatahub)"
+          kubectl apply -f dashboard-operator/config/rbac/role.yaml
+          kubectl apply -f - <<'EOF'
+          apiVersion: v1
```
antowaddle: We should create a file here and reference it in this file.

.github/workflows/pr-build-validation.yml
```diff
@@ -869,15 +1020,176 @@ jobs:
...
+                    resources:
+                      requests:
+                        cpu: 10m
+                        memory: 128Mi
+          EOF
+
+          echo "Waiting for operator rollout..."
+          kubectl rollout status deployment/dashboard-operator -n opendatahub --timeout=120s
+
+      - name: Create Dashboard CR and verify reconciliation starts
```
antowaddle, line 1129: The Dashboard CR reconcile server-side-applies manifests/base+manifests/odh into opendatahub as field owner dashboard-operator. manifests/base/deployment.yaml:30 is image: odh-dashboard-image / imagePullPolicy: Always, resolving to quay.io/opendatahub/odh-dashboard:main. deploy.MergeDeployments preserves replicas and resources but not the image, so the deliberate patch to odh-dashboard:odh-test / Never is reverted and the node starts pulling upstream :main during the 120s poll. This job has no Free disk space step (unlike docker-build-* and the new operator-build) and already holds two loaded images. It also silently destroys the "we're testing the PR image" invariant. Consider a separate job/cluster, managementState: Removed, or a throwaway APPLICATIONS_NAMESPACE.

.github/workflows/pr-build-validation.yml
```diff
@@ -322,6 +323,150 @@ jobs:
...
+            if [ "${DOC_COUNT:-0}" -eq 0 ]; then
+              echo "❌ FAIL: $overlay rendered no resources"
+              FAILURES=$((FAILURES + 1))
+              echo "::endgroup::"
+              continue
+            fi
+
+            # Schema-validate. -ignore-missing-schemas skips CRDs with no public schema
+            # (Dashboard, OdhDashboardConfig, ...) but still fails on malformed core resources.
+            if ! kubeconform \
```
antowaddle: -ignore-missing-schemas turns a misspelled kind into a silent pass — :446
Verified locally: kind: Deployment → Deploymnet in manifests/base yields Valid: 39, Invalid: 0, Errors: 0, Skipped: 10, exit 0 — the typo is just counted as "no schema found." manifests/rhoai skips 54 of 95 resources and is never kubectl apply-ed anywhere, so a typo'd kind there ships undetected. Assert the skipped count against a baseline, or feed the repo's own CRDs via -schema-location.

.github/workflows/pr-build-validation.yml
```diff
@@ -322,6 +323,150 @@ jobs:
...
+          tar -xz -C /tmp -f /tmp/kubeconform.tar.gz kubeconform
+          sudo mv /tmp/kubeconform /usr/local/bin/kubeconform
+          kubeconform -v
+          echo "::endgroup::"
+
+      - name: Validate kustomize overlays
+        run: |
+          set -euo pipefail
+          # Deployable platform overlays that must render and schema-validate cleanly.
+          OVERLAYS=(
```
antowaddle, line 416: Phase 5 omits every overlay the operator actually renders at runtime — :411
OVERLAYS covers only base, odh, rhoai. The operator also kustomize-renders manifests/modules/<slug> (module_deploy.go:205), manifests/observability/{odh,rhoai} (actions.go:205), and manifests/distributions/maas-consumer-portal (maas_consumer_portal.go:201). A broken resources: path in manifests/modules/gen-ai/kustomization.yaml passes the job named "Manifest Validation" and fails only on a live cluster. All of these render cleanly today, so adding them costs nothing.

antowaddle (changes requested):
Thanks @lucferbux for this, exactly what's needed to continue our shift-left effort for build/image testing. I've added a few comments can you review them please?

Additionally, I think it would be useful to update https://github.com/antowaddle/Red-Hat-Quality-Tiger-Team/tree/main/.claude/skills/konflux-build-simulator with the same type of general logic. wdyt?

.github/workflows/pr-build-validation.yml
lucferbux: Fixed. The rollout status check now runs inside an `if ! kubectl rollout status ...; then` block in the same "Deploy operator" step that, on failure, dumps `kubectl describe deployment/dashboard-operator`, `kubectl get pods`, current + `--previous` logs, and recent events before `exit 1` — so a controller panic / scheme error is diagnosable inline instead of surfacing as a bare "timed out".

.github/workflows/pr-build-validation.yml
lucferbux: Done. Extracted into a checked-in file and referenced via `kubectl apply -f`. Both sim-only specs now live under `.github/pr-build-simulation/` (`operator-deployment.yaml` and `operator-rbac.yaml`) — a dedicated dir, not `manifests/`, so kustomize/overlay validation never picks them up.

.github/workflows/pr-build-validation.yml
lucferbux: Done. Extracted into `.github/pr-build-simulation/operator-rbac.yaml` (ServiceAccount + ClusterRoleBinding) and applied via `kubectl apply -f`. See the companion `operator-deployment.yaml` in the same dir for the operator spec.

.github/workflows/pr-build-validation.yml
lucferbux: Good catch on the invariant. Rather than patch `imagePullPolicy` (futile — `deploy.MergeDeployments` re-owns the container spec via SSA each reconcile, so the operator would clobber `image`/`imagePullPolicy` right back), I point the operator at the PR image: `operator-deployment.yaml` sets `RELATED_IMAGE_ODH_DASHBOARD_IMAGE=odh-dashboard:odh-test`, so the operator SSA-applies the already-loaded PR tag instead of resolving to `quay.io/opendatahub/odh-dashboard:main`. That preserves the "we test the PR image" invariant and avoids the `:main` network pull. I also added the `free-disk-space` step to the `operator-integration` job. The dashboard pod may show a benign `ErrImagePull` on the local-only `odh-test` tag (no registry copy), which is fine — Phase 4 asserts only on operator reconcile evidence (finalizer + `observedGeneration`), and a comment in the CR step documents this.

.github/workflows/pr-build-validation.yml
lucferbux: Fixed. Phase 5 no longer relies on silent skipping: kubeconform now runs with `-output json`, and a jq filter fails the job if any skipped resource is a Kubernetes built-in — group `""` (core), `*.k8s.io`, or apps/batch/policy/autoscaling/extensions — which is exactly what a misspelled kind like `Deploymnet` produces. CRD-backed kinds (Dashboard, ConsoleLink, …) are still allowed to skip. Skipped kinds are printed for visibility.

.github/workflows/pr-build-validation.yml
lucferbux: Fixed. OVERLAYS now mirrors what the operator actually renders: `manifests/base`, `manifests/odh`, `manifests/rhoai`, `manifests/observability/{odh,rhoai}`, `manifests/maas-consumer-portal-consolelink/rhoai`, plus dynamic discovery of every `manifests/modules/<slug>/` containing a `kustomization.yaml` (future-proof for new modules). Note: the `manifests/distributions/maas-consumer-portal` path is stale — the operator renders `manifests/maas-consumer-portal-consolelink/rhoai` (per `maas_consumer_portal.go`), which is what I validate.

lucferbux:
@antowaddle you were right that the fixes belong in the generator too — I've folded these review lessons back into the `konflux-build-simulator` skill so regenerated workflows stop reproducing them: antowaddle/Red-Hat-Quality-Tiger-Team#37 (operator reconcile smoke test hardening in Step 2.3b, operator-derived overlay coverage + kubeconform typo guard in Step 2.4b, plus LEARNINGS.md / SKILL.md updates).

.github/workflows/pr-build-validation.yml
```diff
@@ -322,6 +323,196 @@ jobs:
...
+            # emit JSON and fail if any skipped resource belongs to a core Kubernetes group
+            # (core "" or *.k8s.io or apps/batch/policy/autoscaling/extensions) — those
+            # always have a schema, so a skip there means the kind/apiVersion is wrong.
+            KC_JSON=$(mktemp)
+            set +e
+            kubeconform \
+                -strict \
+                -ignore-missing-schemas \
+                -kubernetes-version "$K8S_SCHEMA_VERSION" \
+                -output json \
```
antowaddle, line 473: P1 — Skip detection is currently inert — .github/workflows/pr-build-validation.yml:471
With kubeconform v0.6.7, -output json without -verbose produces {"resources":[]} for valid/skipped resources. Thus BAD_SKIPS never sees anything, and a typo such as kind: Deploymnet still passes. Add -verbose and test the guard with an intentionally invalid kind.

.github/workflows/pr-build-validation.yml
```diff
@@ -322,6 +323,196 @@ jobs:
...
+            SKIPPED_KINDS=$(jq -r '[.resources[] | select(.status=="statusSkipped") | .kind] | unique | join(", ")' "$KC_JSON" 2>/dev/null || echo "")
+            [ -n "$SKIPPED_KINDS" ] && echo "  ℹ️  skipped (no schema): $SKIPPED_KINDS"
+
+            # Fail if a *built-in* kind was skipped — that indicates a typo, not a CRD.
+            BAD_SKIPS=$(jq -r '
+              .resources[]
+              | select(.status=="statusSkipped")
+              | . as $r
+              | (if ($r.version | contains("/")) then ($r.version | split("/")[0]) else "" end) as $g
+              | select($g == "" or ($g | endswith(".k8s.io")) or ($g | IN("apps","batch","policy","autoscaling","extensions")))
```
antowaddle: P1 — The skip classifier will fail valid current manifests — .github/workflows/pr-build-validation.yml:493
Once -verbose is added, the blanket *.k8s.io match flags intentional skips such as CustomResourceDefinition (apiextensions.k8s.io/v1) and HTTPRoute (gateway.networking.k8s.io/v1). Use an explicit built-in allowlist or exclude known CRD groups/kinds.

antowaddle (changes requested):
@lucferbux 2 additional changed. 

Can you also check why new simulator manifests do not trigger this workflow — .github/workflows/pr-build-validation.yml:16 Maybe
Add .github/pr-build-simulation/** to on.pull_request.paths; otherwise changes solely to operator-deployment.yaml or operator-rbac.yaml will not rerun validation.?

.github/workflows/pr-build-validation.yml
lucferbux: Good catch — confirmed locally with kubeconform v0.6.7: \`-output json\` without \`-verbose\` returns \`{"resources":[]}\` for valid/skipped resources, so the guard never saw a skip. Added \`-verbose\`; a \`kind: Deploymnet\` typo is now reported as a skipped built-in and fails the job (verified locally). Fixed in 32f4871.

.github/workflows/pr-build-validation.yml
lucferbux: Right again — with \`-verbose\` the blanket \`*.k8s.io\` match false-fails on \`apiextensions.k8s.io/CustomResourceDefinition\` and \`gateway.networking.k8s.io/HTTPRoute\` (both skipped, no schema). Replaced the suffix heuristic with explicit built-in **group** + **kind** allowlists: a skip is a typo if its group is a core built-in group (catches a misspelled kind) OR its kind is a well-known built-in kind (catches a misspelled group). Verified 0 false positives across all 17 rendered overlays and that both \`Deploymnet\` and \`aps/v1\` typos are caught. Fixed in 32f4871.

lucferbux:
@antowaddle addressed the two P1 kubeconform findings (both threads resolved) in 32f4871, plus your third point from the review summary:

> Can you also check why new simulator manifests do not trigger this workflow — Add `.github/pr-build-simulation/**` to `on.pull_request.paths`?

Done — added `.github/pr-build-simulation/**` to `on.pull_request.paths`, so a change touching only `operator-deployment.yaml` / `operator-rbac.yaml` now re-runs the validation.

Summary of this push:
- `-verbose` added to kubeconform (without it the JSON `resources` array is empty for valid/skipped resources, so the built-in-skip typo guard was inert).
- Replaced the `.k8s.io` suffix heuristic with explicit built-in **group + kind** allowlists (the suffix match false-failed on `apiextensions.k8s.io/CustomResourceDefinition` and `gateway.networking.k8s.io/HTTPRoute`). Verified 0 false positives across all 17 rendered overlays and that `Deploymnet` and `aps/v1` typos are both caught.
- Trigger paths now include `.github/pr-build-simulation/**`.

.github/workflows/pr-build-validation.yml
```diff
@@ -322,6 +324,209 @@ jobs:
...
+  operator-build:
+    name: "Phase 1: Operator Image Build"
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    needs: check-skip
+    if: needs.check-skip.outputs.should-skip != 'true'
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.2.2
```
antowaddle, line 346: One last tiny nit and this is good

Can we add this please? This ensures we don't expose the GH token through the dockerfile. 
```
with:
  persist-credentials: false
```

lucferbux: Done in 170cfca — added `persist-credentials: false`. Applied it to every checkout step in the workflow (not just this one), since several jobs `docker build` from the repo root and none of them push, so the persisted token is never needed. `fetch-depth: 0` is preserved where it was set.

antowaddle (approved):
/lgtm

lucferbux:
@antowaddle can you add the lgtm again please?

PR #9675, opened 2026-09-08
code modified for check inference service

diwakarsharma2:
## Description

Modified code to add isvc state check in between retries so that if there is any failure, code will quit in between rather than waiting for complete retries. https://redhat.atlassian.net/browse/RHOAIENG-91826

## Request review criteria:

If you have UI changes: 

## Summary by CodeRabbit

- **Bug Fixes**
  - Deployment checks now identify non-recoverable inference service failures immediately instead of continuing to poll.
  - Transient runtime issues continue polling until recovery or timeout.
  - Deployment status validation occurs before waiting between checks.
  - Retry intervals are reduced from 50 seconds to 30 seconds, with a longer maximum wait window.
  - Error messages now provide consistent condition details for easier troubleshooting.
  - Model-serving workflows now support standard and raw deployment configurations.
  - Runtime-argument controls are detected more reliably.


diwakarsharma2:
Jenkins run result: 2795

packages/cypress/cypress/utils/oc_commands/modelServing.ts
```diff
@@ -284,15 +327,12 @@ export const checkInferenceServiceState = (
...
+          ${buildConditionDetails()}`;
 
         cy.log(errorMessage);
         throw new Error(errorMessage);
       } else {
-        return cy.wait(50000).then(() => {
-          failOnDeploymentStatus(serviceName);
-          return checkState();
-        });
+        return cy.wait(30000).then(() => checkState());
```
ConorOM1, line 335: this is the core issue, this needs to match the documented timeout of 8 minutes cc: @sridarna

diwakarsharma2: @ConorOM1 for s390x, isvc takes time to come up. 39th attempt with 50 secs of wait in each attempt

Screenshot 2026-09-09 at 12 29 27 PM

packages/cypress/cypress/utils/oc_commands/modelServing.ts
```diff
@@ -284,15 +327,12 @@ export const checkInferenceServiceState = (
...
       }
```
sridarna, line 336: the checkInferenceServiceState function makes maxAttempts=96.
So cy.wait(30000)  The total time executed if the model is not ready would be  
48 minutes.
Which is long wait time and if the model is not ready, waiting for 48 minutes would not generate the Jenkins test report as it is set to timeout to 30minutes.

As previous observation, the old wait time 5000ms for each attempt has worked for both predictive and generative model If the model is healthy.
If it is taking more than 8 minutes. The E2E should fail since with some issue.

sridarna: Please update the cy.wait to 5000.

packages/cypress/cypress/utils/oc_commands/modelServing.ts
sridarna: +1 to what added by @ConorOM1 
If it is taking time more than 8 minutes, the Jenkins test report would not be generated. 
Also, As per cypress-e2e rules, this is a long wait time.
@diwakarsharma2 Please test it with documented timeout of 8 minutes.
which would update cy.wait(5000)

diwakarsharma2: @sridarna can we do onething here, for x86 lets do the wait time of 5 secs and for s390x wait time of 50 secs?

sridarna: @antowaddle @FedeAlonso What are your thoughts on this?

PR #9678, opened 2026-09-08
feat: serve MaaS Consumer Portal under gateway path

caponetto:
https://redhat.atlassian.net/browse/RHOAIENG-93250

## Description

Serve the MaaS Consumer Portal at `https://<spec.gateway.domain>/maas-consumer-portal/` instead of a derived hostname. The portal now shares the Gateway hostname and authentication session, while retaining its independently managed operand lifecycle.

- Replace the portal HTTPRoute hostname contract with the Gateway domain, a 302 trailing-slash redirect, and a prefix rewrite to the existing Core-BFF routes.
- Publish the path URL in `status.maasConsumerPortalUrl` and the ConsoleLink only after the Deployment and HTTPRoute are ready; retain the last-known-good URL during failed rollouts.
- Configure portal-specific asset public paths, router basename, browser API prefixes, and local dev proxy rewrites.
- Remove obsolete hostname parameters/replacements and document the Gateway filter prerequisite, migration behavior, and portal-only operation.

No new MaaS or GenAI BFF Deployments are introduced. Gateway-owned OAuth callback and sign-out paths remain unchanged.

## How Has This Been Tested?

Run the following from the repository root:

```bash
npm run test-unit --workspace=@odh-dashboard/maas-consumer-portal-distribution
npm run type-check --workspace=@odh-dashboard/maas-consumer-portal-distribution
npm run lint --workspace=@odh-dashboard/maas-consumer-portal-distribution
npm run build --workspace=@odh-dashboard/maas-consumer-portal-distribution
make -C dashboard-operator test
make -C dashboard-operator test-integration
```

## Test Impact

- Added/updated coverage for portal URL derivation, manifest rendering, ConsoleLink URL, HTTPRoute redirect/rewrite configuration, status retention/removal, portal-only lifecycle, and in-place HTTPRoute/ConsoleLink updates.
- Added portal router-basename/root-redirect coverage.
- The production Gateway data plane and integrated OAuth/BYOIDC flow require real-cluster validation.

## Request review criteria:

## Summary by CodeRabbit

- **New Features**
  - MaaS Consumer Portal is now available at `https://<gateway-domain>/maas-consumer-portal/`.
  - Added path-based routing for portal pages, assets, redirects, and API requests.
  - Portal links and dashboard URLs now use the shared gateway domain.

- **Bug Fixes**
  - Improved route and URL updates when the gateway domain changes.
  - Preserved the last verified dashboard URL while status updates are pending or deployment fails.

- **Documentation**
  - Updated deployment, routing, and local development guidance for the gateway-based portal URL.


manifests/distributions/maas-consumer-portal/httproute.yaml
```diff
@@ -8,12 +8,29 @@ spec:
...
+          requestRedirect:
+            path:
+              type: ReplaceFullPath
+              replaceFullPath: /maas-consumer-portal/
+            statusCode: 308
     - matches:
         - path:
             type: PathPrefix
-            value: /
+            value: /maas-consumer-portal
```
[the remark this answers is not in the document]
caponetto: No change needed. The finding is not valid for Gateway API PathPrefix: matching is performed path-element by path-element, so /maas-consumer-portal matches that path and its descendants, but not /maas-consumer-portal-other/.... The current manifest also matches the requested contract exactly. Gateway API reference (https://gateway-api.sigs.k8s.io/reference/api-spec/main/spec/)

manifests/distributions/maas-consumer-portal/httproute.yaml
```diff
@@ -8,12 +8,29 @@ spec:
...
+      filters:
+        - type: URLRewrite
+          urlRewrite:
+            path:
+              type: ReplacePrefixMatch
+              replacePrefixMatch: /
       backendRefs:
```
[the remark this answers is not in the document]
caponetto: No code change needed. The route intentionally relies on the existing gateway-owned authentication session; adding a route filter would violate the shared-hostname design.

The requested OAuth/BYOIDC deep-link verification is real-cluster validation, which you explicitly excluded from this work. It remains documented as a deployment validation prerequisite.

caponetto:

caponetto:
@lucferbux everything is working as expected. I'm just gonna wait for https://github.com/opendatahub-io/odh-dashboard/pull/9667 to be merged to do a final test.

distributions/maas-consumer-portal/config/rspack.dev.js
```diff
@@ -127,9 +135,11 @@ const buildProxyConfig = () => {
       console.info('Dashboard host:', dashboardHost);
       return [
         {
-          context: ['/maas/api', '/gen-ai/api'],
+          context: [`${BASE_PATH}/maas/api`, `${BASE_PATH}/gen-ai/api`],
           target: `https://${dashboardHost}`,
-          secure: false,
+          pathRewrite: { [`^${BASE_PATH}`]: '' },
+          secure: true,
```
lucferbux, line 146: This flips cluster-proxy TLS verification on by default (`secure: false` → `true`). Right hardening and it's documented, but it's a silent behavior change for existing Mode B setups: anyone proxying to a dashboard with an internally-signed cert now fails the handshake unless they set `ODH_DASHBOARD_CA_FILE`. Consider logging a hint when verification is on but no CA file is set.

lucferbux (commented):
Nice follow-up to #9637 — this is exactly the path-based approach from the strategy doc (§3b), and it pairs with the token-forwarding fix in #9667. The gateway ReplacePrefixMatch makes the portal Service see the same paths it did under the old subdomain, so Core-BFF SPA fallback and the maas/gen-ai BFF contracts keep working unchanged. Verified end-to-end: router basename handles browser URLs, API_URL_PREFIX/URL_PREFIX handle raw fetches, and mod-arch-core only uses URL_PREFIX for API paths (no double-prefix). Operator + manifest test coverage is solid, and the missing-domain retry fix is a good catch.

A few non-blocking things below, mostly consistency/maintainability. Two that aren't inline:
- `rspack.dev.js` and `rspack.common.js` each redeclare `const BASE_PATH = '/maas-consumer-portal'`; worth sourcing both (and `portalPaths.ts`) from one place so a rename can't drift.
- The portal HTTPRoute now shares the gateway hostname with the Dashboard `/` catch-all. Correctness relies on the Gateway impl merging same-hostname routes by path precedence (portal prefix winning over `/`). The docs call out the RequestRedirect/URLRewrite prerequisite but not this same-hostname merge — worth adding, since a non-conforming impl would let portal requests fall through to the Dashboard route.

packages/gen-ai/frontend/src/app/utilities/const.ts
```diff
@@ -1,6 +1,6 @@
 import { asEnumMember, DeploymentMode } from 'mod-arch-core';
 
-const URL_PREFIX = process.env.URL_PREFIX || '/gen-ai';
+const URL_PREFIX = process.env.URL_PREFIX || `${process.env.BASE_PATH || ''}/gen-ai`;
```
lucferbux: Asymmetric with maas here: maas keeps `URL_PREFIX` = `/maas` for routing and adds a separate `API_URL_PREFIX` for the base path, but gen-ai folds `BASE_PATH` straight into `URL_PREFIX`. It's fine today only because gen-ai uses `URL_PREFIX` for API paths only (its routes are relative) — if someone later builds a route off it, it double-prefixes under the portal basename. Worth matching maas's split, or a comment noting `URL_PREFIX` here is API-only.

Also the `process.env.URL_PREFIX ||` short-circuit sits in front of `BASE_PATH`: if `URL_PREFIX` ever gets defined in a portal build, the base path is silently dropped and gen-ai API calls 404 through the gateway.

distributions/maas-consumer-portal/src/PortalContextProvider.tsx
```diff
@@ -4,7 +4,7 @@ import PortalAreaContextProvider from './PortalAreaContextProvider';
 
 const modularArchConfig: ModularArchConfig = {
   deploymentMode: DeploymentMode.Standalone,
-  URL_PREFIX: '/maas',
+  URL_PREFIX: `${process.env.BASE_PATH || ''}/maas`,
```
lucferbux, line 7: This hand-rebuilds maas's `API_URL_PREFIX`. Since the PR now exports `API_URL_PREFIX` from the maas package for exactly this purpose, importing it avoids the two definitions drifting if maas's prefix ever changes.

distributions/maas-consumer-portal/src/portalPaths.ts
```diff
@@ -0,0 +1,5 @@
+/** Browser-visible mount point for the MaaS Consumer Portal. */
+export const PORTAL_BASE_PATH = '/maas-consumer-portal';
```
lucferbux, line 5: `/maas-consumer-portal` is now hardcoded here, in the Go operator (`maasConsumerPortalBasePath`, with a trailing slash), in `httproute.yaml` (Exact + ReplaceFullPath + PathPrefix), and in both rspack configs. No single source of truth and the slash conventions differ across them — a rename would need all in sync. Not blocking, but at least a cross-referencing comment would help.

distributions/maas-consumer-portal/config/rspack.dev.js
caponetto: Valid. Added a cluster-proxy startup hint when TLS verification is enabled without ODH_DASHBOARD_CA_FILE.

packages/gen-ai/frontend/src/app/utilities/const.ts
caponetto: Valid; fixed.

  - URL_PREFIX is now basename-relative (/gen-ai or its override).
  - New API_URL_PREFIX prepends BASE_PATH and is used by all GenAI browser API calls and modular-architecture configuration.
  - This prevents a portal build from dropping its external prefix when URL_PREFIX is set.

distributions/maas-consumer-portal/src/PortalContextProvider.tsx
caponetto: Skipped:  API_URL_PREFIX is not exported by @odh-dashboard/maas; it remains an internal MaaS frontend detail. Exporting it solely for this host configuration would introduce an unnecessary public cross-package   contract. The portal’s modular-architecture configuration intentionally owns its BASE_PATH + /maas value.

distributions/maas-consumer-portal/src/portalPaths.ts
caponetto: Valid and addressed. Added a synchronization comment to PORTAL_BASE_PATH identifying the portal HTTPRoute and rspack BASE_PATH constants as the related definitions.

caponetto:
## Review summary

  - Added a cluster-proxy startup hint when TLS verification is enabled without `ODH_DASHBOARD_CA_FILE`.
  - Split GenAI routing and API prefixes to match MaaS: `URL_PREFIX` is basename-relative; `API_URL_PREFIX` applies `BASE_PATH` for all browser/BFF requests.
  - Skipped exporting MaaS `API_URL_PREFIX`: it is an internal frontend detail, and exporting it only for the portal host config would add an unnecessary public cross-package contract.
  - Documented that the portal base path must stay synchronized with the HTTPRoute and Rspack configuration.
  - Documented the Gateway prerequisite to merge same-hostname routes with path precedence, ensuring the portal route wins over the dashboard catch-all.

I'm still testing this out. I will `/ unhold` this PR as soon as I'm done.

caponetto:
/unhold

@lucferbux this is ready!

PR #9696, opened 2026-09-09
feat: remove the MaaS Consumer Portal OpenShift ConsoleLink

caponetto:
[https://redhat.atlassian.net/browse/RHOAIENG-93529](https://redhat.atlassian.net/browse/RHOAIENG-93529)

## Description

- Remove the MaaS Consumer Portal ConsoleLink so the portal no longer appears in the OpenShift application menu.
- Preserve the standalone portal at `https://<gateway-domain>/maas-consumer-portal/`, including its HTTPRoute, Deployment, service, federation configuration, readiness checks, and status URL handling.
- Leave the core Dashboard ConsoleLink and its controller/RBAC support unchanged.
- **Bonus**: Set `Cache-Control: no-cache` on the Core BFF SPA entry document and client-side-route fallback so browsers revalidate after gateway logout. Static assets remain cacheable.

## How Has This Been Tested?

```bash
make -C dashboard-operator test
make -C dashboard-operator test-integration
make -C distributions/core-bff/bff test
```

Manual tests on a real cluster.

## Test Impact

- Updated MaaS Consumer Portal manifest, lifecycle, and envtest coverage to remove ConsoleLink creation and cleanup assertions while retaining Deployment and HTTPRoute lifecycle coverage.
- Extended SPA fallback coverage to verify the cache-control header for `/` and a client-side route.

## Request review criteria:

## Summary by CodeRabbit

* **Changes**
  * Removed the MaaS Consumer Portal ConsoleLink resource and related configuration.
  * Portal Deployments and HTTPRoutes continue to be deployed, managed, and preserved during dependency issues.
  * Updated portal cleanup behavior to focus on managed HTTPRoutes.

* **Bug Fixes**
  * SPA entry-point responses now include `Cache-Control: no-cache`, including client-side routes and fallback responses.
  * Nested `index.html` files remain cacheable.

* **Documentation**
  * Updated dashboard operator documentation to reflect the portal resource changes.


caponetto:
/unhold

@lucferbux This is ready for review

dashboard-operator/internal/controller/maas_consumer_portal.go
```diff
@@ -290,15 +287,6 @@ func (r *DashboardReconciler) deleteLabeledMaaSConsumerPortalCustomResources(ctx
...
-	consoleLinks := &unstructured.UnstructuredList{}
-	consoleLinks.SetGroupVersionKind(consoleLinkListGVK)
-	if err := r.List(ctx, consoleLinks, client.MatchingLabels{labels.PlatformPartOf: maasConsumerPortalPartOf}); err != nil {
-		if !meta.IsNoMatchError(err) {
-			errs = append(errs, err)
-		}
-	} else {
-		errs = append(errs, r.deleteMaaSConsumerPortalUnstructuredItems(ctx, consoleLinks.Items))
-	}
 	return errors.Join(errs...)
```
lucferbux, line 290: Removing the ConsoleLink from the bundle **and** deleting the ConsoleLink cleanup branch here means any `maas-consumer-portal-link` ConsoleLink created by a *prior* operator build stays orphaned after upgrade: the SSA deployer doesn't prune resources absent from the desired set, and owner-ref GC only fires on Dashboard CR deletion. On such clusters the portal link keeps showing in the OpenShift app menu — the exact thing RHOAIENG-88481 wants gone.

If this operand has only ever run in dev/test (never shipped to a real cluster), this is moot — feel free to resolve. Otherwise a one-time delete-by-label (`part-of=maas-consumer-portal`, GVK ConsoleLink) on reconcile would close the upgrade gap. Non-blocking — your call on whether it's needed now or an accepted follow-up.

distributions/core-bff/bff/internal/api/routes.go
```diff
@@ -71,14 +72,20 @@ func (app *App) newStaticHandler() http.Handler {
 	fileServer := http.FileServer(staticDir)
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		ctxLogger := helpers.GetContextLoggerFromReq(r)
+		if r.URL.Path == "/" || r.URL.Path == "/"+constants.IndexHTMLFileName {
+			// The SPA entry point must revalidate after gateway logout so a cached
+			// document cannot render with an expired authentication session.
+			w.Header().Set(constants.HeaderCacheControl, constants.CacheControlNo)
```
lucferbux, line 78: Minor: `Cache-Control: no-cache` is now set in two places in this handler (here for the entry doc served via `fileServer`, and again on the SPA fallback below at the `ServeFile` path). Both are correct and cover disjoint runtime paths, but a tiny helper or a shared comment tying both to the "entry-document must revalidate" contract would read a bit cleaner. Non-blocking.

lucferbux (commented):
Reviewed at effort **high**. The change is tightly scoped and reads well: the portal ConsoleLink is cleanly removed across Go, manifest, kustomization params, and docs while the standalone portal (Deployment/HTTPRoute) and the core-dashboard ConsoleLink are preserved, and the `no-cache` SPA-entry logic in core-bff is correct and well covered by the new `app_test.go` cases. A couple of non-blocking notes inline.

One test-coverage note that didn't fit a line: `TestIntegration_MaaSConsumerPortalConsoleLinkRemovedWhenDisabled` was deleted with no replacement. That test covered the *combined* teardown path (core `managementState: Removed` **and** portal `Removed`) asserting full portal cleanup. With it gone, a regression in `deleteMaaSConsumerPortalResources` under that specific combined-removal path would no longer be caught. If the combined path is still meaningful, consider keeping a slimmed-down version (asserting Deployment/HTTPRoute teardown, minus the ConsoleLink assertions).

lucferbux (changes requested):
Requesting changes on the one substantive item (see inline comment on `maas_consumer_portal.go`): removing the ConsoleLink from the bundle **and** deleting the ConsoleLink cleanup branch leaves any `maas-consumer-portal-link` ConsoleLink created by a prior operator build orphaned after upgrade — the SSA deployer doesn't prune resources absent from the desired set, and owner-ref GC only fires on Dashboard CR deletion. On already-provisioned clusters the portal link would keep showing in the OpenShift app menu, which is the exact behavior RHOAIENG-88481 aims to remove.

Suggested fix: a one-time delete-by-label (`part-of=maas-consumer-portal`, ConsoleLink GVK) on reconcile to close the upgrade gap.

If this operand has only ever run in dev/test and never shipped to a real cluster, this is moot — just confirm that and I'm happy to approve as-is. The other two notes (duplicate `Cache-Control` set, deleted combined-removal test) are non-blocking.

dashboard-operator/internal/controller/maas_consumer_portal.go
caponetto: This is intentionally out of scope. The MaaS Consumer Portal has not been released, so the operator does not need upgrade-time cleanup for previously deployed portal ConsoleLinks. The revised lifecycle intentionally stops managing ConsoleLinks altogether. Any links created in development or test clusters can be removed manually if needed.

distributions/core-bff/bff/internal/api/routes.go
caponetto: Both writes are intentional and cover separate paths: direct entry-document requests (/ and /index.html) go through http.FileServer, while SPA deep links use the explicit index.html fallback. The existing comment captures the revalidation rationale, so I’d keep this small duplication rather than add a helper for a single header assignment.

caponetto:
The ConsoleLink cleanup concern is intentionally out of scope: the MaaS Consumer Portal has not been released, so there is no supported upgrade population requiring migration or operator-managed cleanup of previously created portal ConsoleLinks. Any development or test-cluster links can be removed manually.

The combined-removal coverage remains in TestIntegration_MaaSConsumerPortalLifecycle. It starts with the core dashboard managementState: Removed and the portal Managed, then sets the portal to Removed and verifies cleanup of the Deployment, HTTPRoute, Service, NetworkPolicy, federation ConfigMap, RBAC, and serving-certificate Secret. The only removed assertion is for the removed ConsoleLink. The focused lifecycle envtest passes.

Other answers are inline. No code changes needed.

cc @lucferbux

lucferbux (approved):
/approve
/lgtm

PR #9702, opened 2026-09-09
Edit nim pvc subpath

emilys314:
Closes https://redhat.atlassian.net/browse/RHOAIENG-88011

## Description

This is a PR to support managing existing NIM PVCs without too much scope creep. It adds the "NIM Storage" radio option when editing a NIM PVC in the 'cluster storage' tab. 
- For simplicity, you can't change the context type
- You also can't create a new NIM storage, they only show up from being created in the wizard. 
- The extension is limited to only update annotations, which is fine since all context types just add different annotations

Note: There are 2 states. First is the PVC data object, 2nd is the the useState in the field. On initial render, technically the PVC data object doesn't have any value for the contextTypeAnnotations, instead the field state inits it by reading the existingPvc object directly. On save, if you didn't change anything, the value still remains even though the save function using the PVC data object, this is because it using _.merge with the old object, so the existing pvc annotations remains

Screenshot 2026-09-09 at 2 02 51 PM

## How Has This Been Tested?

**Regression checks**
1. Create a new PVC, the NIM option shouldn't show
2. Edit a General purpose PVC, the NIM option shouldn't show
3. Edit a model storage PVC, it too shouldn't show

**NIM flow**
1. Deploy a new NIM model, create a new storage
2. Go to the "Cluster storage" tab and edit it
3. You should be able to see the NIM radio option selected, and show the subpath of the PVC
4. Change the subpath, hit save. Verify the change persisted and the annotation is updated
5. If you create a new NIM model, and use that PVC, it should auto-populate the subpath with the new value.

## Test Impact

The cypress manage nim pvcs test has been expanded to assert the changing of the subpath and that it persists on save
I also added some simple jest unit tests for the field that is rendered and the isNIMPVC helper function.

## Request review criteria:

If you have UI changes: 

cc @justinxhale

## Summary by CodeRabbit

- **New Features**
  - Added support for storage context types when creating and editing persistent storage.
  - Extension-based storage settings now display read-only selections with extension-specific configuration fields.
  - Added loading placeholders while storage context options are retrieved.
  - NIM storage now supports editing its model subpath.

- **Bug Fixes**
  - Storage context annotations are preserved and updated when storage configurations change.
  - PVC update detection now recognizes added, removed, or changed context annotations.

### DIFF
--- a/frontend/src/pages/projects/screens/detail/storage/data.ts
+++ b/frontend/src/pages/projects/screens/detail/storage/data.ts
@@ -6,7 +6,7 @@ import { StorageTableData } from './types';
 import { getContextStorageTypeExplanation, StorageContextType } from './useStorageContextType';
 
 export const columns = (data: {
-  storageContextTypes: StorageContextType[];
+  storageContextTypes?: StorageContextType[];
 }): SortableData<StorageTableData>[] => [
   {
     field: 'name',
@@ -48,7 +48,7 @@ export const columns = (data: {
     width: 15,
     sortable: false,
     info: {
-      popover: getContextStorageTypeExplanation(data.storageContextTypes),
+      popover: getContextStorageTypeExplanation(data.storageContextTypes ?? []),
       popoverProps: {
         showClose: true,
       },
--- a/frontend/src/pages/projects/screens/spawner/storage/PVCContextField.tsx
+++ b/frontend/src/pages/projects/screens/spawner/storage/PVCContextField.tsx
@@ -6,13 +6,23 @@ import {
   HelperText,
   HelperTextItem,
   Radio,
+  Skeleton,
   TextInput,
 } from '@patternfly/react-core';
+import type { PersistentVolumeClaimKind } from '@odh-dashboard/k8s-core';
 import FieldGroupHelpLabelIcon from '@odh-dashboard/ui-core/components/FieldGroupHelpLabelIcon';
 import {
   GENERAL_PURPOSE_PVC_CONTEXT_TYPE,
+  getPVCContextStorageType,
   MODEL_STORAGE_PVC_CONTEXT_TYPE,
+  StorageContextType,
 } from '#~/pages/projects/screens/detail/storage/useStorageContextType';
+
+export const PVCContextFieldSkeleton: React.FC = () => (
+  <FormGroup label="Storage context" fieldId="storage-context">
+    <Skeleton height="80px" screenreaderText="Loading storage context options" />
+  </FormGroup>
+);
 
 type PVCContextFieldProps = {
   setModelName: (name: string) => void;
@@ -21,6 +31,9 @@ type PVCContextFieldProps = {
   modelPath: string;
   setValid: (isValid: boolean) => void;
   removeModelAnnotations: () => void;
+  existingPvc?: PersistentVolumeClaimKind;
+  storageContextTypes?: StorageContextType[];
+  setContextTypeAnnotations: (annotations: Record<string, string>) => void;
 };
 
 const PVCContextField: React.FC<PVCContextFieldProps> = ({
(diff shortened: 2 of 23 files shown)


packages/plugin-core/src/extension-points/cluster-storage.ts
```diff
@@ -10,6 +21,8 @@ export type ClusterStorageContextProperties = {
   description?: string;
   /** Returns `true` when the given PVC belongs to this storage context. */
   isPVCUsingStorageContextType: CodeRef<(pvc: PersistentVolumeClaimKind) => boolean>;
+  /** Settings fields shown when editing an existing PVC of this context type. */
+  PVCStorageContextSettingsFields: ComponentCodeRef<PVCStorageContextSettingsFieldsProps>;
```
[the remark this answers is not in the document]
emilys314: Outdated extensions are not a concern in our codebase and also the extension isn't released yet

frontend/src/pages/projects/screens/detail/storage/data.ts
```diff
@@ -48,7 +48,7 @@ export const columns = (data: {
     width: 15,
     sortable: false,
     info: {
-      popover: getContextStorageTypeExplanation(data.storageContextTypes),
+      popover: getContextStorageTypeExplanation(data.storageContextTypes ?? []),
```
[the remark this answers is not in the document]
emilys314: This shouldn't happen, but i updated the 0 length edge case

packages/nim-serving/src/pages/clusterStorage/clusterStorage.tsx
```diff
@@ -0,0 +1,47 @@
...
+      <FormGroup label="Subpath" fieldId="nim-subpath">
+        <TextInput
+          id="nim-subpath"
+          data-testid="nim-subpath-input"
+          value={subPath}
+          onChange={(_event, value) => {
+            setSubPath(value);
+            onChange({ [NIM_PVC_SUBPATH_ANNOTATION]: value });
+          }}
+          placeholder="/"
```
YuliaKrimerman: should we probably omit the annotation in this case and not set it to ""?

packages/nim-serving/src/pages/clusterStorage/clusterStorage.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+export const isNIMPVC = (pvc: PersistentVolumeClaimKind): boolean =>
+  !!pvc.metadata.annotations && Object.hasOwn(pvc.metadata.annotations, NIM_PVC_ANNOTATION);
+
+const Fields: React.FC<PVCStorageContextSettingsFieldsProps> = ({ existingPvc, onChange }) => {
+  const [subPath, setSubPath] = React.useState(
+    existingPvc.metadata.annotations?.[NIM_PVC_SUBPATH_ANNOTATION] ?? '',
+  );
+
+  return (
```
YuliaKrimerman, line 18: not a must but SubPathField in NIMPVCField.tsx is nearly identical (label, data-testid, helper text, placeholder). we can extract to shared component for reusability

frontend/src/pages/projects/screens/spawner/storage/PVCContextField.tsx
```diff
@@ -30,121 +43,163 @@ const PVCContextField: React.FC<PVCContextFieldProps> = ({
...
+    <FormGroup
+      label="Storage context"
+      fieldId="storage-context"
+      labelHelp={
+        <FieldGroupHelpLabelIcon
+          content={
+            <p>
+              {isLocked
+                ? 'The context indicates the purpose of the storage. It cannot be changed.'
+                : 'The context indicates the purpose of the storage: general purpose, or model storage.'}
```
YuliaKrimerman: Consider reusing getContextStorageTypeExplanation(storageContextTypes ?? []) for the unlocked case, same as the table column popover.

frontend/src/pages/projects/screens/spawner/storage/PVCContextField.tsx
```diff
@@ -30,121 +43,163 @@ const PVCContextField: React.FC<PVCContextFieldProps> = ({
...
+        description={GENERAL_PURPOSE_PVC_CONTEXT_TYPE.description}
+        isChecked={contextType.title === GENERAL_PURPOSE_PVC_CONTEXT_TYPE.title}
+        isDisabled={isLocked}
+        onChange={() => {
+          setContextType(GENERAL_PURPOSE_PVC_CONTEXT_TYPE);
+          setValid(true);
+          removeModelAnnotations();
+        }}
+      />
+      <br />
```
YuliaKrimerman, line 116: I know this is pre-existing in couple of places but if you're up to it PF Stack / Flex instead of <br />. but this is a nit

YuliaKrimerman (commented):
couple of small things

packages/nim-serving/src/pages/clusterStorage/clusterStorage.tsx
emilys314: That's a fair point, the wizard doesn't create the subpath if it's an empty string either. Updated

packages/nim-serving/src/pages/clusterStorage/clusterStorage.tsx
emilys314: Good point, updated

frontend/src/pages/projects/screens/spawner/storage/PVCContextField.tsx
emilys314: Good callout. Updated, although the workbenches spawner page also uses this modal, and it doesn't provide any context storage types as a prop, so in that scenario I manually sent in the 2 context types that are hardcoded and always present

frontend/src/pages/projects/screens/spawner/storage/PVCContextField.tsx
emilys314: I don't think I feel up for it in this pr 🙁 😂

mturley:
@emilys314 from the expanded content of the preflight bot - seems out of scope, but I wonder if it’s legitimate enough for a followup issue

> 🧹 Nitpick comments (1)
> frontend/src/pages/projects/screens/spawner/storage/PVCContextField.tsx (1)
> `64–75`: _🧹 Nit_ · _Claude review_
> 
> **`validatePath` closes over stale `contextType` state.**
> 
> When the user switches to the Model Storage radio and a non-empty path already exists, `validatePath(modelPath)` is called before `setContextType` has re-rendered the component. At that moment `contextType` is still the old value, so the `contextType.title === GENERAL_PURPOSE_PVC_CONTEXT_TYPE.title` branch can fire incorrectly and call `setValid(true)` even for an invalid path. This is a pre-existing pattern (the old `isGeneralPurpose` boolean had the same stale-closure issue) and only affects an edge case, but consider passing the target type explicitly:
> 
> ```diff
> -  const validatePath = (path: string) => {
> +  const validatePath = (path: string, type: StorageContextType = contextType) => {
>      const trimmedPath = path.trim();
>      const pathRegex = /^[A-Za-z0-9._-]+(?:\/[A-Za-z0-9._-]+)*\/?$/;
>      if (
> -      contextType.title === GENERAL_PURPOSE_PVC_CONTEXT_TYPE.title ||
> +      type.title === GENERAL_PURPOSE_PVC_CONTEXT_TYPE.title ||
> ```
> 
> And pass the new type at call sites:
> 
> ```diff
> -  validatePath(modelPath);
> +  validatePath(modelPath, MODEL_STORAGE_PVC_CONTEXT_TYPE);
> ```

mturley (approved):
/approve
/lgtm

Code LGTM, existing comments look to be addressed. Tested: displayed an existing NIM storage's context, edited its subpath, cleared it to use the PVC root, confirmed edits saved correctly and persisted on reopening the form, confirmed that reusing the PVC in the wizard prepopulates the updated subpath. Also tested regression paths by creating and editing both general purpose and model storage PVCs: NIM storage was correctly unavailable for new PVCs and absent from the edit options for those existing PVCs and the model path remained intact.

PR #9709, opened 2026-09-09
Prompt for deleting PVC when deleting NIM deployment

mturley:
https://redhat.atlassian.net/browse/RHOAIENG-88022

## Description

Adds optional PVC cleanup to legacy NVIDIA NIM model deletion. The delete confirmation now lets users select the associated cache PVC, checks whether other model deployments use it, and warns before deleting a shared cache.

PVC cleanup is protected by a deletion preflight: selected resources are dry-run first, and real deletion requests are sent only after all selected dry runs succeed. If preflight validation fails, the modal reports which resource failed and makes clear that nothing was deleted.

### Screenshots

Delete modal with new checkbox:

Screenshot 2026-09-10 at 1 43 24 PM

Loading state after checking checkbox:

Screenshot 2026-09-10 at 1 27 14 PM

Result when nothing shares the PVC:

image

Result when other deployments share the PVC:

image

Dry-run deletion validation failure; nothing was deleted:

Screenshot 2026-09-11 at 1 57 52 PM

Real deletion failure (if that somehow happens after the dry run - screenshot was taken before dry run was implemented):

Screenshot 2026-09-10 at 1 44 49 PM

On successful deletion, the modal closes.

## Scope and deferred work

This PR is limited to `nimWizard` KServe-backed NIM deployments represented by an `InferenceService` and `ServingRuntime`. NIMService-operator deployments are intentionally excluded from this PR and covered by the follow-up Story [RHOAIENG-94199](https://redhat.atlassian.net/browse/RHOAIENG-94199).

The implementation:
- Derives the NIM cache PVC by following the `kserve-container` mount at `/mnt/models/cache` to its corresponding PVC-backed ServingRuntime volume.
- Keeps PVC deletion opt-in so deleting a deployment does not remove shared or reusable model cache by default.
- Queries other KServe deployments using the same PVC, excluding the deployment currently being deleted.
- Shows an accessible loading state while PVC dependents are checked.
- Prevents submission while PVC dependency checking is pending.
- Blocks selected PVC deletion when dependency lookup fails; users can uncheck PVC deletion to proceed with primary deployment deletion only.
- Shows the affected deployment names when the PVC is shared, and a confirmation message when it is not.
- Dry-runs the selected InferenceService, ServingRuntime, and PVC deletions before issuing any real deletion requests.
- Reports the specific resource when dry-run deletion validation fails and states that nothing was deleted.
- Deletes the primary InferenceService and ServingRuntime first, then deletes the PVC only when explicitly selected and after preflight succeeds.
- Preserves the existing error handling if an actual primary deployment deletion fails or PVC deletion returns an explicit unsuccessful Kubernetes status.
- Tolerates Kubernetes PVC delete responses that return the deleted PVC resource rather than a `K8sStatus`, as well as empty or undefined responses.

## How Has This Been Tested?

- `npm run lint --workspace=model-serving-cypress`
- `npm run type-check --workspace=model-serving-cypress`
- `cd packages/cypress && npm run type-check`
- `cd packages/cypress && npm run lint -- --quiet`
- `npm run test-unit --workspace=@odh-dashboard/nim-serving -- src/nimKServe/__tests__/delete.spec.ts --runInBand`
- `npm run type-check --workspace=@odh-dashboard/nim-serving`
- `npm run test-unit --workspace=@odh-dashboard/model-serving -- src/concepts/__tests__/extensionUtils.spec.ts --runInBand`
- `npm run type-check --workspace=@odh-dashboard/model-serving`
- `npm run type-check --workspace=@odh-dashboard/kserve`
- `git diff --check`
- Manual browser testing with `nimWizard=true` covered the private-PVC flow, shared-PVC dependent warning, loading behavior, unchecked deletion, checked deletion, no-cache deployment deletion, and PVC response handling while running as a non-cluster-admin.
- Manual browser testing also verified the dry-run failure state: the failed resource is identified and the message states that nothing was deleted.

## Test Impact

- Added unit coverage for PVC deletion responses without a string status result, including `undefined`, an empty response, and a PVC-resource response containing only a pending phase.
- Added unit coverage proving that deployment delete extensions evaluate `isActive(deployment)`, exclude inactive extensions, and select the highest-priority active extension.
- Added unit coverage for dry-run-before-delete ordering and resource-specific PVC dry-run failure handling.
- Added mocked Cypress coverage for the shared-cache flow, including the dependent deployment warning, dependent deployment list, and all three DELETE requests.
- Added mocked Cypress coverage for the non-shared-cache flow and its informational confirmation message.
- Added mocked Cypress coverage for the loading-state submission guard and failed dependency lookup behavior.
- Added Cypress page-object locators for the PVC deletion controls, loading state, dependent-deployment alert, and dependent deployment list.

## Request review criteria:

[RHOAIENG-94199]: https://redhat.atlassian.net/browse/RHOAIENG-94199

## Summary by CodeRabbit

- **New Features**
  - Added a specialized deletion dialog for NIM KServe model deployments.
  - Users can choose whether to delete the associated cache PVC.
  - The dialog identifies and lists other deployments using a shared PVC.
  - Added loading and warning messages while PVC usage is checked.
  - Added support for platform-specific deployment deletion dialogs.

- **Bug Fixes**
  - Prevented deletion of shared cache PVCs without clear confirmation.
  - Improved handling of PVC deletion failures and incomplete responses.
  - Disabled deletion when PVC dependency checks cannot be completed.


packages/kserve/src/clusterStorage/connectedResources.tsx
```diff
@@ -39,10 +45,92 @@ export const useConnectedKServeResources = (
...
+  return inferenceServices.flatMap((inferenceService) => {
+    const { name } = inferenceService.metadata;
+    const runtimeName = inferenceService.spec.predictor.model?.runtime;
+    if (
+      typeof name !== 'string' ||
+      !name ||
+      name === excludeInferenceServiceName ||
+      typeof runtimeName !== 'string' ||
+      !runtimeNamesUsingPVC.has(runtimeName)
+    ) {
```
emilys314: nitpick but typescript knows it's a string, so the `typeof` check isn't needed


```suggestion
      !name ||
      name === excludeInferenceServiceName ||
      !runtimeName ||
      !runtimeNamesUsingPVC.has(runtimeName)
    ) {
```

packages/nim-serving/src/nimKServe/deleteUtils.ts
```diff
@@ -0,0 +1,41 @@
+import type { KServeDeployment } from '@odh-dashboard/kserve/types';
+import { KSERVE_CONTAINER_NAME, NIM_CACHE_MOUNT_PATH } from '../constants';
+
+export type NIMKServePVCReference = {
+  name: string;
+  namespace: string;
+};
+
+/** Returns the legacy NIM cache PVC mounted by the selected ServingRuntime. */
+export const getNIMCachePVCReference = (
```
emilys314: This function is fairly similar to `extractNIMPVCFieldData` in https://github.com/opendatahub-io/odh-dashboard/blob/main/packages/nim-serving/src/nimKServe/fields/nimPVCApplyExtract.ts

The other function gives some other fluff and no namespace, but we always know the namespace is the same as the deployment

packages/model-serving/extension-points/index.ts
```diff
@@ -215,13 +215,28 @@ export const isModelServingDeploymentsTableExtension = <D extends Deployment = D
...
+
 export type ModelServingDeleteModal<D extends Deployment = Deployment> = Extension<
   'model-serving.platform/delete-deployment',
   {
     platform: D['modelServingPlatformId'];
     onDelete: CodeRef<(deployment: D) => Promise<void>>;
     title: string;
     submitButtonLabel: string;
+    /** Whether this extension applies to the deployment. Active extensions with the highest priority win. */
+    isActive?: CodeRef<(deployment: D) => boolean> | true;
```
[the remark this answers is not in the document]
mturley: Addressed

packages/model-serving/extension-points/index.ts
```diff
@@ -215,13 +215,28 @@ export const isModelServingDeploymentsTableExtension = <D extends Deployment = D
...
+    /** Priority among active delete extensions for the same platform. */
+    priority?: number | 0;
```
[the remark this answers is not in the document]
mturley: Addressed

packages/model-serving/src/components/deleteModal/DeleteModelServingModal.tsx
```diff
@@ -43,28 +63,38 @@ const DeleteModelServingModal: React.FC<DeleteModelServingModalProps> = ({
...
+
+  return ResolvedDeleteModalComponent ? (
+    <ResolvedDeleteModalComponent
+      key={`${deployment.model.metadata.namespace}/${deployment.model.metadata.name}`}
+      deployment={deployment}
+      onClose={onBeforeClose}
+      onDelete={deletePrimaryResource}
+      title={deleteModal.properties.title}
+      submitButtonLabel={deleteModal.properties.submitButtonLabel}
+    />
```
emilys314, line 84: A full modal replacement isn't ideal I think. Other platforms could want a similar feature to give control over for deleting related deployment resources. Also in theory, you could mess up the existing patterns by doing a full replacement, since the contract is very open ended.

This is fine for now though

packages/model-serving/extension-points/index.ts
```diff
@@ -215,13 +215,28 @@ export const isModelServingDeploymentsTableExtension = <D extends Deployment = D
...
     platform: D['modelServingPlatformId'];
     onDelete: CodeRef<(deployment: D) => Promise<void>>;
     title: string;
     submitButtonLabel: string;
+    /** Whether this extension applies to the deployment. Active extensions with the highest priority win. */
+    isActive?: CodeRef<(deployment: D) => boolean> | true;
+    /** Priority among active delete extensions for the same platform. */
+    priority?: number;
+    /** Optional platform-specific modal. It must call onClose(true) only after deletion succeeds. */
+    DeleteModalComponent?: CodeRef<React.ComponentType<ModelServingDeleteModalComponentProps<D>>>;
```
emilys314: nit pick but I think a `ComponentCodeRef` is a more precise type here

emilys314 (commented):
Works good though 

Image

packages/model-serving/src/components/deleteModal/DeleteModelServingModal.tsx
mturley: That’s a good point, I think it requires some care to find the right abstraction here though. I briefly poked around with it and it got messy quickly. Opened a followup: https://redhat.atlassian.net/browse/RHOAIENG-94265

packages/model-serving/extension-points/index.ts
mturley: Ah yep

packages/nim-serving/src/nimKServe/deleteUtils.ts
mturley: Fair, will reuse that

packages/kserve/src/clusterStorage/connectedResources.tsx
mturley: Done

packages/nim-serving/src/nimKServe/deleteUtils.ts
mturley: Done

packages/model-serving/extension-points/index.ts
mturley: Done

packages/model-serving/src/components/deleteModal/DeleteModelServingModal.tsx
emilys314: Sounds good! That makes sense

packages/kserve/src/clusterStorage/connectedResources.tsx
```diff
@@ -39,10 +45,91 @@ export const useConnectedKServeResources = (
...
+  const [inferenceServices, servingRuntimes] = await Promise.all([
+    k8sListResourceItems<InferenceServiceKind>({
+      model: InferenceServiceModel,
+      queryOptions: { ns: namespace },
+    }),
+    k8sListResourceItems<ServingRuntimeKind>({
+      model: ServingRuntimeModel,
+      queryOptions: { ns: namespace },
+    }),
+  ]);
```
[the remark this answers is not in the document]
mturley: I don’t think this is valid. These resources will not be malformed, they are server-validated and the callbacks are statically typed with `InferenceServiceKind` and `ServingRuntimeKind`. If we assume they could always be malformed we would have bigger problems across the entire codebase. Safe to trust TypeScript here.

packages/nim-serving/src/nimKServe/delete.ts
```diff
@@ -0,0 +1,32 @@
...
+
+/** Deletes a legacy NIM deployment, optionally deleting its cache PVC afterward. */
+export const deleteLegacyNIMDeployment = async ({
+  deletePrimaryDeployment,
+  namespace,
+  pvcName,
+  deletePVC,
+  deletePVCResource = deletePvc,
+}: DeleteLegacyNIMDeploymentArgs): Promise<void> => {
+  await deletePrimaryDeployment();
```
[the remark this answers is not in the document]
mturley: Instead of the complexity of a retry UX, I added a dry-run pass. If any of the 3 dry run deletes fail (IS, SR or PVC) nothing will be deleted and this error will show.

Screenshot 2026-09-11 at 1 57 52 PM

packages/nim-serving/src/nimKServe/NIMKServeDeleteModal.tsx
```diff
@@ -0,0 +1,167 @@
...
+      ) : null}
+    </Stack>
+  ) : null;
+
+  return (
+    <DeleteModal
+      title={title}
+      onClose={handleClose}
+      submitButtonLabel={submitButtonLabel}
+      onDelete={handleDelete}
```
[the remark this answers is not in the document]
mturley: Done.

emilys314 (approved):
Looks good, still works good

packages/model-serving/cypress/tests/mocked/modelServing/modelServingNim.cy.ts
```diff
@@ -76,6 +79,146 @@ describe('NIM Models Deployments', () => {
     modelServingGlobal.getModelRow('Test Name').findKebabAction('Delete').should('exist');
   });
 
+  it('should warn before deleting a shared NIM cache PVC', () => {
+    const selectedDeployment = mockNimInferenceService();
+    const sharedDeployment = mockInferenceServiceK8sResource({
+      name: 'shared-model',
+      displayName: 'Shared model',
+      runtimeName: 'test-name',
```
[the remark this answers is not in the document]
mturley: ServingRuntimes are not shared between InferenceServices in this model. This is invalid.

PR #9713, opened 2026-09-10
E2E: Add Workbench Kueue Queued scenario

kanishka-commits:
https://redhat.atlassian.net/browse/RHOAIENG-93223

## Description

Adds a Cypress E2E test for the **Queued workbench Kueue lifecycle**. When the ClusterQueue quota is fully consumed by one workbench, a second workbench shows **Queued** status, and the resources modal displays the relevant queue and quota context.

Extends the shared Kueue workbench lifecycle fixture and types to support the Queued scenario. The Queued test runs in a separate spec to avoid resource-name collisions with the existing Inadmissible test.

## How Has This Been Tested?

- Ran `testWorkbenchKueueQueuedLifecycle.cy.ts` on a Kueue-enabled cluster with `devFeatureFlags=true`.
- Verified the second workbench enters the **Queued** state when the available quota is fully consumed.
- Verified the resources modal displays the expected queue/quota information.

## Test Impact

- New E2E test: `testWorkbenchKueueQueuedLifecycle.cy.ts`
- Updated Kueue workbench lifecycle fixture/types to support queued quota values:
  - `queuedCpuQuota`
  - `queuedMemoryQuota`
  - etc.

### Self Checklist

- [x] The developer has manually tested the changes and verified that they work.
- [x] Testing instructions have been added in the PR body for changes that are not immediately obvious.
- [x] The developer has added tests or explained why testing cannot be added.
- [x] The code follows our Best Practices.

## Summary by CodeRabbit

- **Tests**
  - Expanded end-to-end coverage for workbench lifecycle behavior under different quota conditions.
  - Added validation for workbenches becoming queued when quota is partially available.
  - Improved checks for quota-related messaging, queue position, resource details, and readiness states.
  - Added isolated scenarios and cleanup to make lifecycle validation more reliable.

### DIFF
--- a/packages/cypress/cypress/tests/e2e/dataScienceProjects/workbenches/testWorkbenchKueueLifecycle.cy.ts
+++ b/packages/cypress/cypress/tests/e2e/dataScienceProjects/workbenches/testWorkbenchKueueLifecycle.cy.ts
@@ -208,10 +208,7 @@ describe('Workbench Kueue Lifecycle Tests', () => {
 
         openWorkbenchesTab(projectCtx);
         createWorkbench(projectCtx, firstWorkbenchName);
-        pollUntilWorkloadAdmitted(projectCtx.projectName, {
-          maxAttempts: 120,
-          pollIntervalMs: 5000,
-        });
+        pollUntilWorkloadAdmitted(projectCtx.projectName);
 
         createWorkbench(projectCtx, secondWorkbenchName);
         pollUntilAnyWorkloadMessageMatches(projectCtx.projectName, QUEUED_MESSAGE);
--- a/packages/cypress/cypress/tests/e2e/dataScienceProjects/workbenches/testWorkbenchKueueLifecycle.cy.ts
+++ b/packages/cypress/cypress/tests/e2e/dataScienceProjects/workbenches/testWorkbenchKueueLifecycle.cy.ts
@@ -3,7 +3,7 @@ import {
   deleteOpenShiftProject,
   createOpenShiftProject,
 } from '../../../../utils/oc_commands/project';
-import { retryableBefore } from '../../../../utils/retryableHooks';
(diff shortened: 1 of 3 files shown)


kanishka-commits:
## Jenkins build 2878/ - Passing

On: tangerine-odh
OCP: 4.21.27
Operator: RHODS
Image: quay.io/rhoai/rhoai-fbc-fragment:rhoai-3.6-ea.1

Screenshot 2026-09-11 at 2 20 12 PM

packages/cypress/cypress/tests/e2e/dataScienceProjects/workbenches/testWorkbenchKueueLifecycle.cy.ts
```diff
@@ -21,117 +25,210 @@ import {
...
+          throw new Error('Test setup did not complete');
+        }
+        const projectCtx = ctx;
+
+        const firstWorkbenchName = `kueue-wb-q1-${projectCtx.uuid}`;
+        const secondWorkbenchName = `kueue-wb-q2-${projectCtx.uuid}`;
+
+        openWorkbenchesTab(projectCtx);
+        createWorkbench(projectCtx, firstWorkbenchName);
+        pollUntilWorkloadAdmitted(projectCtx.projectName, {
```
claudialphonse78: umm 120 ? isnt this a lot of attempts

kanishka-commits: Good catch, 120 attempts (10 min) was copied from the model deployment lifecycle test. Removed the override to use the default 60 attempts (5 min), which should be sufficient for Kueue admission.

claudialphonse78 (commented):
/lgtm

claudialphonse78 (approved):

PR #9716, opened 2026-09-10
RHOAIENG-83656: Add real-cluster E2E framework

liday-rh:
https://redhat.atlassian.net/browse/RHOAIENG-83656

## Description

Add the initial real-cluster E2E framework for dashboard-operator:

- add an `e2e`-tagged `TestMain` that requires `KUBECONFIG` and `TEST_NAMESPACE`, builds a controller-runtime client, and verifies cluster connectivity, the test namespace, and the served Dashboard CRD version;
- add shared polling, atomic create, UID-guarded cleanup, JQ matcher, and platform contract helpers for follow-up E2E scenarios;
- add `make test-e2e` and `make build-e2e` targets, including selective `-run` and standalone binary support;
- add the odh-platform-utilities framework dependency; and
- document local, containerized, RBAC, isolation, and cleanup requirements.

The framework complements the existing blocking envtest suite. It does not install CRDs or run against shared Cypress clusters.

## How Has This Been Tested?

- `cd dashboard-operator && make lint`
- `cd dashboard-operator && make test`
- `cd dashboard-operator && make test-integration`
- `cd dashboard-operator && make build`
- `cd dashboard-operator && make build-e2e`
- `cd dashboard-operator && make check-chart-crds`
- `cd dashboard-operator && go mod tidy -diff`
- verified `make test-e2e` fails early with a clear error when the required environment is missing

### Manual cluster testing

- Built commit `8b946353488399f980f03e44991888967e3f7b63` and deployed `quay.io/liday/dashboard-operator:pr-9716-8b9463534` (`sha256:39d64f4d95edc74e8706f49ec17078b7eabbc45ea04c87ec4a593ff16ce9c191`) to an OpenShift 4.21 / RHOAI 3.5 cluster.
- Upgraded the Dashboard CRD and installed the PR dashboard-operator in `redhat-ods-applications`.
- Ran the complete real-cluster E2E target:

  ```bash
  KUBECONFIG=/Users/liday/.kube/config \
    TEST_NAMESPACE=redhat-ods-applications \
    make test-e2e
  ```

  `TestE2EFrameworkPreflight` passed.
- Verified the operator was 1/1 available, both dashboard replicas were available, all ten standalone module deployments were available, and `Dashboard/default-dashboard` reported `Ready=True` and `Degraded=False` after the run.
- Applied an additive service-account-token mount to the live `core-bff` container to work around an existing omission in the upstream `main` dashboard manifest; this omission is not introduced by this PR.

## Test Impact

This adds the E2E test bootstrap and shared helpers required by follow-up operand, lifecycle, and contract scenarios. It also keeps E2E-tagged files in the normal lint gate and preserves the existing unit and envtest suites.

## Request review criteria:

If you have UI changes:

## Summary by CodeRabbit

- **New Features**
  - Added a real-cluster end-to-end testing framework for dashboard-operator.
  - Added commands to build and run standalone E2E test binaries.
  - Added validation for kubeconfig, test namespace, cluster connectivity, and Dashboard CRD availability.
  - Added reusable helpers for applying, monitoring, validating, and safely cleaning up Dashboard resources.
  - Added coverage for Dashboard lifecycle behavior and expression-based result validation.

- **Documentation**
  - Documented E2E prerequisites, environment variables, execution options, and test authoring guidance.
  - Updated integration testing documentation to reference the E2E framework.

### DIFF
--- a/dashboard-operator/test/e2e/e2e_test.go
+++ b/dashboard-operator/test/e2e/e2e_test.go
@@ -79,7 +79,6 @@ func initializeE2E() error {
 	if err != nil {
 		return fmt.Errorf("build REST config from KUBECONFIG: %w", err)
 	}
-	config.Timeout = preflightTimeout
 
 	c, err := client.New(config, client.Options{Scheme: scheme})
 	if err != nil {
@@ -121,13 +120,17 @@ func initializeE2E() error {
 
 func newE2EScheme() (*runtime.Scheme, error) {
 	scheme := runtime.NewScheme()
-	for name, addToScheme := range map[string]func(*runtime.Scheme) error{
-		"Kubernetes":    clientgoscheme.AddToScheme,
-		"apiextensions": apiextensionsv1.AddToScheme,
-		"Dashboard":     dashboardv1alpha1.AddToScheme,
-	} {
-		if err := addToScheme(scheme); err != nil {
-			return nil, fmt.Errorf("register %s APIs with E2E scheme: %w", name, err)
+	registrations := []struct {
+		name        string
+		addToScheme func(*runtime.Scheme) error
+	}{
+		{name: "Kubernetes", addToScheme: clientgoscheme.AddToScheme},
+		{name: "apiextensions", addToScheme: apiextensionsv1.AddToScheme},
+		{name: "Dashboard", addToScheme: dashboardv1alpha1.AddToScheme},
+	}
+	for _, registration := range registrations {
+		if err := registration.addToScheme(scheme); err != nil {
+			return nil, fmt.Errorf("register %s APIs with E2E scheme: %w", registration.name, err)
 		}
 	}
 
--- a/dashboard-operator/test/e2e/e2e_test.go
+++ b/dashboard-operator/test/e2e/e2e_test.go
@@ -18,6 +18,7 @@ import (
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/tools/clientcmd"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	gatewayv1 "sigs.k8s.io/gateway-api/apis/v1"
 )
 
 const (
@@ -127,6 +128,7 @@ func newE2EScheme() (*runtime.Scheme, error) {
 		{name: "Kubernetes", addToScheme: clientgoscheme.AddToScheme},
 		{name: "apiextensions", addToScheme: apiextensionsv1.AddToScheme},
 		{name: "Dashboard", addToScheme: dashboardv1alpha1.AddToScheme},
+		{name: "Gateway API", addToScheme: gatewayv1.Install},
 	}
 	for _, registration := range registrations {
 		if err := registration.addToScheme(scheme); err != nil {
--- a/dashboard-operator/test/e2e/e2e_test.go
+++ b/dashboard-operator/test/e2e/e2e_test.go
@@ -1,0 +1,135 @@
+//go:build e2e
+
+package e2e
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"os"
+	"testing"
+	"time"
+
+	dashboardv1alpha1 "github.com/opendatahub-io/odh-dashboard/dashboard-operator/api/v1alpha1"
+	corev1 "k8s.io/api/core/v1"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	k8svalidation "k8s.io/apimachinery/pkg/util/validation"
+	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/tools/clientcmd"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	dashboardCRDName = "dashboards.components.platform.opendatahub.io"
+	preflightTimeout = 30 * time.Second
+)
+
+var (
+	k8sClient     client.Client
+	testNamespace string
+)
+
+func TestMain(m *testing.M) {
+	if err := initializeE2E(); err != nil {
+		_, _ = fmt.Fprintf(os.Stderr, "E2E preflight failed: %v\n", err)
+		os.Exit(1)
+	}
+
+	os.Exit(m.Run())
+}
+
+func TestE2EFrameworkPreflight(t *testing.T) {
+	if k8sClient == nil {
+		t.Fatal("controller-runtime client was not initialized")
+	}
+	if testNamespace == "" {
+		t.Fatal("test namespace was not initialized")
+	}
+}
+
+func initializeE2E() error {
+	kubeconfig := os.Getenv("KUBECONFIG")
+	if kubeconfig == "" {
+		return errors.New("KUBECONFIG must point to a kubeconfig file")
+	}
+
+	info, err := os.Stat(kubeconfig)
+	if err != nil {
+		return fmt.Errorf("cannot access KUBECONFIG: %w", err)
+	}
+	if !info.Mode().IsRegular() {
+		return errors.New("KUBECONFIG must point to a regular file")
+	}
+
+	namespace := os.Getenv("TEST_NAMESPACE")
+	if namespace == "" {
+		return errors.New("TEST_NAMESPACE must name an existing namespace")
+	}
+	if problems := k8svalidation.IsDNS1123Label(namespace); len(problems) > 0 {
+		return fmt.Errorf("TEST_NAMESPACE %q is invalid: %v", namespace, problems)
+	}
+
+	scheme, err := newE2EScheme()
+	if err != nil {
+		return err
+	}
+
+	config, err := clientcmd.BuildConfigFromFlags("", kubeconfig)
+	if err != nil {
+		return fmt.Errorf("build REST config from KUBECONFIG: %w", err)
+	}
+	config.Timeout = preflightTimeout
+
+	c, err := client.New(config, client.Options{Scheme: scheme})
+	if err != nil {
+		return fmt.Errorf("create controller-runtime client: %w", err)
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), preflightTimeout)
+	defer cancel()
+
+	if err := c.Get(ctx, client.ObjectKey{Name: namespace}, &corev1.Namespace{}); err != nil {
+		return fmt.Errorf("verify cluster connectivity and TEST_NAMESPACE %q: %w", namespace, err)
+	}
+
+	crd := &apiextensionsv1.CustomResourceDefinition{}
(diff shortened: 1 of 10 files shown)


dashboard-operator/test/e2e/helpers_test.go
```diff
@@ -0,0 +1,296 @@
...
+			"refuse to delete Dashboard %q without %s=%s ownership label",
+			key.Name,
+			e2eManagedByKey,
+			e2eFieldOwner,
+		)
+	}
+
+	uid := dashboard.UID
+	resourceVersion := dashboard.ResourceVersion
+	preconditions := client.Preconditions{UID: &uid, ResourceVersion: &resourceVersion}
```
lucferbux: The `ResourceVersion` precondition makes cleanup racy: if the operator writes status between the `Get` above and this `Delete`, the RV moves and we get a 409, so cleanup fails intermittently. The `UID` precondition alone already gives us the "don't delete a recreated object" guarantee without the flake — I'd drop `ResourceVersion` here.

dashboard-operator/test/e2e/helpers_test.go
```diff
@@ -0,0 +1,296 @@
...
+	key := client.ObjectKey{Name: dashboardv1alpha1.DashboardInstanceName}
+
+	if err := c.Get(ctx, key, dashboard); err != nil {
+		if apierrors.IsNotFound(err) {
+			return nil
+		}
+		return fmt.Errorf("get Dashboard %q before cleanup: %w", key.Name, err)
+	}
+
+	if dashboard.Labels[e2eManagedByKey] != e2eFieldOwner {
```
lucferbux, line 161: This guard is weaker than it reads. `applyDashboardCR` SSA-stamps the `managed-by` label onto whatever `default-dashboard` already exists, and the documented prerequisite is that the platform controller is scaled down, not that its singleton is gone. So if a platform-created Dashboard is still there, we stamp it on apply and then happily delete it here — i.e. the label check protects nothing in the realistic case. Suggest capturing the UID at apply time and only cleaning up the object we actually created.

dashboard-operator/test/e2e/helpers_test.go
```diff
@@ -0,0 +1,296 @@
...
+			deployment := &appsv1.Deployment{}
+			if err := c.Get(ctx, client.ObjectKey{Namespace: namespace, Name: name}, deployment); err != nil {
+				if apierrors.IsNotFound(err) {
+					return false, nil
+				}
+				return false, err
+			}
+
+			for _, condition := range deployment.Status.Conditions {
+				if condition.Type == appsv1.DeploymentAvailable && condition.Status == corev1.ConditionTrue {
```
lucferbux, line 101: `DeploymentAvailable=True` can still be reported by the old ReplicaSet mid-rollout, so this can return ready against stale pods — which is exactly what'll bite the upgrade / spec-change scenarios this helper is meant to support. Worth also gating on `Status.ObservedGeneration >= Generation` and `UpdatedReplicas == desired`.

dashboard-operator/test/e2e/helpers_test.go
```diff
@@ -0,0 +1,296 @@
...
+	t.Helper()
+
+	matched, failureMessage := evaluateJQMatch(actual, expression, args...)
+	if !matched {
+		t.Fatal(failureMessage)
+	}
+}
+
+func evaluateJQMatch(actual any, expression string, args ...any) (bool, string) {
+	matcher := jq.Match(expression, args...)
```
lucferbux: Two issues here:

1. `jq.Match` runs `fmt.Sprintf(format, args...)` on the expression, so `assertJQMatch` passing the raw expression as the format with no args means any developer query containing a literal `%` (e.g. `test("a%b")`) gets printf-mangled before it runs, and we silently evaluate a broken query. The redaction test masks this because it always passes the expression as a `%s` arg. Suggest calling `jq.Match("%s", expression)` in the helper path.
2. The failure messages collapse everything to `"failed to evaluate JQ assertion"` / `"JQ assertion did not match"`, dropping the expression and the gojq compile error. Those describe the query, not cluster data, so they're safe to keep and we'll want them to debug real cluster failures — redact only the actual value, not the expression.

dashboard-operator/test/e2e/e2e_test.go
```diff
@@ -0,0 +1,138 @@
...
+
+func newE2EScheme() (*runtime.Scheme, error) {
+	scheme := runtime.NewScheme()
+	registrations := []struct {
+		name        string
+		addToScheme func(*runtime.Scheme) error
+	}{
+		{name: "Kubernetes", addToScheme: clientgoscheme.AddToScheme},
+		{name: "apiextensions", addToScheme: apiextensionsv1.AddToScheme},
+		{name: "Dashboard", addToScheme: dashboardv1alpha1.AddToScheme},
```
lucferbux, line 130: nit: the operator deploys Gateway/HTTPRoute operands, so I'd register `sigs.k8s.io/gateway-api` in the scheme now — otherwise the first operand-health scenario that reads a Gateway typed will hit "no kind is registered" and have to re-add the scheme instead of reusing the framework's.

lucferbux (changes requested):
Really nice first step — the scope is exactly right for a module operator (we own the `Dashboard` CR, so operand health / CR reconciliation / platform-contract / cleanup are ours to test), and I like that it consumes `validation.ValidatePlatformContract` from platform-utilities instead of reinventing conformance. Condition-based waits, non-forced SSA, the ownership-label cleanup guard, and the fail-fast preflight are all the right instincts.

A few things I'd like to sort before we merge, mostly because the follow-up scenario stories will inherit these helpers as-is:

- The helpers ship unexercised (all `//nolint:unused`) and nothing runs `make test-e2e` in CI, so any bug here is invisible until someone builds on it. Not asking for full scenarios in this PR, but could we land at least one thin apply -> wait -> cleanup scenario (or a manual/nightly CI job) so the framework is actually driven once?
- A couple of real robustness bugs in the wait/cleanup helpers (inline).
- The one test that does run today is gated behind the cluster preflight, so the redaction guarantee never executes in normal PR CI (inline).

Details in the inline comments.

dashboard-operator/test/e2e/helpers_test.go
liday-rh: Fixed in cf9d19a3f. Cleanup now uses only the captured UID as the delete precondition, so operator status updates cannot cause a resourceVersion conflict while protection against deleting a recreated singleton remains.

dashboard-operator/test/e2e/helpers_test.go
liday-rh: Fixed in cf9d19a3f. Deployment readiness now requires observedGeneration >= generation and both updatedReplicas and readyReplicas to equal the desired replica count before accepting DeploymentAvailable=True.

dashboard-operator/test/e2e/e2e_test.go
liday-rh: Fixed in cf9d19a3f. The E2E scheme now registers Gateway API v1 with gatewayv1.Install (the supported v1.4 registration API).

dashboard-operator/test/e2e/helpers_test.go
liday-rh: Fixed in cf9d19a3f. JQ evaluation now calls jq.Match("%s", expression), preserves the expression and gojq error in diagnostics, and never formats the actual cluster value. The regression test covers literal percent signs, mismatches, evaluation errors, and redaction, and it now runs in normal make test CI rather than behind the e2e preflight.

dashboard-operator/test/e2e/helpers_test.go
liday-rh: Fixed in cf9d19a3f. The apply helper now refuses to modify a pre-existing default-dashboard, returns the UID of the object it created, and cleanup requires both that UID and the E2E ownership label. The lifecycle scenario also skips before mutation when the singleton already exists.

dashboard-operator/test/e2e/helpers_test.go
```diff
@@ -0,0 +1,287 @@
...
+}
+
+func applyDashboardCR(c client.Client, spec dashboardv1alpha1.DashboardSpec) (types.UID, error) {
+	ctx := context.Background()
+	key := client.ObjectKey{Name: dashboardv1alpha1.DashboardInstanceName}
+	existing := &dashboardv1alpha1.Dashboard{}
+	if err := c.Get(ctx, key, existing); err == nil {
+		return "", fmt.Errorf("refuse to apply Dashboard %q because it already exists", key.Name)
+	} else if !apierrors.IsNotFound(err) {
+		return "", fmt.Errorf("check for existing Dashboard %q: %w", key.Name, err)
```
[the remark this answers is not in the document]
liday-rh: Fixed in f8de975cf. The helper now performs one atomic client.Create call, returns a contextual error on AlreadyExists or other create failures, and captures the API-assigned UID directly from the successfully created Dashboard. It no longer has a check-then-apply window and cannot adopt an existing singleton.

dashboard-operator/test/e2e/helpers_test.go
```diff
@@ -0,0 +1,287 @@
...
+		applyConfiguration,
+		client.FieldOwner(e2eFieldOwner),
+	); err != nil {
+		return "", fmt.Errorf("server-side apply Dashboard %q: %w", dashboard.GetName(), err)
+	}
+
+	created := &dashboardv1alpha1.Dashboard{}
+	if err := c.Get(ctx, key, created); err != nil {
+		return "", fmt.Errorf("get applied Dashboard %q: %w", key.Name, err)
+	}
```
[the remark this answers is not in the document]
liday-rh: Fixed structurally in f8de975cf. Replacing SSA plus a follow-up Get with client.Create removes the post-apply Get failure path entirely: the successful create response populates the Dashboard UID directly, so the caller can register UID-guarded cleanup immediately without an intermediate API request that could fail.

lucferbux:
/lgtm
/approve

Re-reviewed after the previous round — all six inline points (UID-only delete precondition, ownership-label/adoption guard via atomic \`client.Create\`, deployment-readiness generation/replica gating, \`jq.Match("%s", expression)\` with preserved-and-redacted diagnostics, Gateway API scheme registration) and the CodeRabbit check-then-act race are resolved in the current code, and a thin apply→wait→cleanup lifecycle scenario plus the JQ redaction test (running in normal CI) now exercise the framework. Verified the reconciler marks `ProvisioningSucceeded=False (reason=Removed)`, matching the lifecycle test's expectation. Only minor nits remain (see below), none blocking.

Non-blocking nits for a follow-up if you like:
- `make test-e2e` guard `test -f "$KUBECONFIG"` and the in-code single-file assumption break if `KUBECONFIG` is a colon-separated path list; itnull is documented as one file, so just a sharp-edge note.
- `jq.go` is untagged (so the redaction test can run in normal CI), which pulls the framework matcher deps (gojq/gomega) into the non-e2e build graph — acceptable trade-off, worth a comment.
- Manual cluster run in the PR body was on the initial commit `8b946353`, before the atomic-create/hardening refactors; the changed create/cleanup path was not re-validated on a real cluster.

PR #9726, opened 2026-09-10
feat(model-serving): support secretKeyRef env vars in deployment wizard

YuliaKrimerman:
https://issues.redhat.com/browse/RHOAIENG-93848

## Description

Adds **Value / Secret** support for custom runtime environment variables in the deployment wizard. Users can reference an existing Kubernetes Secret via `valueFrom.secretKeyRef` instead of entering a plain-text value.

Part of epic [RHOAIENG-90745](https://issues.redhat.com/browse/RHOAIENG-90745) (private HF catalog secret deploy). Enables Story 2 to wire HF token input to Secret-mode env vars.

## How Has This Been Tested?

**Manual (cluster)**
1. Create a Secret in the target namespace (e.g. `hf-token-test` with key `HF_TOKEN`)
2. Open deployment wizard → enable **Add custom runtime environment variables**
3. Add a row, set type to **Secret**, enter env name (`HF_TOKEN`), secret name, and key
4. On Review, confirm display: `HF_TOKEN, secret/hf-token-test:HF_TOKEN`
5. Deploy and verify ISVC/NIMService spec has `valueFrom.secretKeyRef`
6. Edit deployment and confirm Secret fields round-trip correctly
Screenshot 2026-09-10 at 3 02 33 PM
Screenshot 2026-09-10 at 3 02 45 PM
Screenshot 2026-09-10 at 3 02 51 PM

## Test Impact

**Added**
- `packages/model-serving/src/shared/__tests__/environmentVariablesUtils.spec.ts` — mapping, normalization, review formatting
- `EnvironmentVariablesField.spec.tsx` — migrated from `.spec.ts`; secret schema validation, UI interaction tests

**Updated**
- `packages/kserve/src/__tests__/deployUtils.spec.ts` — `secretKeyRef` apply/extract
- `packages/nim-serving/src/api/nimservices/__tests__/k8s.spec.ts` — NIM env assembly with secrets
- `packages/nim-serving/src/pages/deploymentWizard/__tests__/extractNIMFormData.spec.ts` — extract round-trip

Cypress coverage for the Secret dropdown is out of scope here (tracked separately in [RHOAIENG-93850](https://redhat.atlassian.net/browse/RHOAIENG-93850)).

## Request review criteria:

If you have UI changes:

[RHOAIENG-90745]: https://redhat.atlassian.net/browse/RHOAIENG-90745?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[RHOAIENG-93850]: https://redhat.atlassian.net/browse/RHOAIENG-93850?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

- **New Features**
  - Deployment wizards support value-based and secret-sourced environment variables, including secret name and key fields.
  - Secret references can be marked optional where supported.
  - Environment variable values and secret references receive type-specific validation.
  - Environment variable settings are preserved consistently across model-serving deployment workflows.
  - Review screens format value-based and secret-based variables appropriately.
  - Selection controls can optionally hide the selected-item indicator.

- **Tests**
  - Added coverage for variable types, validation, optional references, and deployment conversion.


packages/model-serving/src/shared/environmentVariablesUtils.ts
```diff
@@ -0,0 +1,141 @@
...
+export const mapEnvironmentVariableToK8sEnv = (
+  envVar: EnvironmentVariable,
+): K8sEnvironmentVariable => {
+  if (envVar.type === EnvironmentVariableType.Secret) {
+    return {
+      name: envVar.name,
+      valueFrom: {
+        secretKeyRef: {
+          name: envVar.secretName,
+          key: envVar.secretKey,
```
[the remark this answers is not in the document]
YuliaKrimerman: Thanks for flagging this. Referencing Secrets via valueFrom.secretKeyRef is standard Kubernetes workload behavior — any user who can create InferenceServices/Pods in a namespace can already reference Secrets in that namespace via YAML or kubectl, regardless of this UI change. The notebook spawner uses the same pattern.

This PR doesn't introduce new access paths; it exposes the existing (and preferred) secretKeyRef mechanism instead of storing values in plain text. If we want admission-time enforcement of Secret access, that would be a separate platform/security initiative (admission webhook / OPA), not something the dashboard form can enforce without listing all Secrets (which itself requires get permission).

Tracking as out of scope for RHOAIENG-93848.

packages/model-serving/src/shared/environmentVariablesUtils.ts
```diff
@@ -0,0 +1,171 @@
...
+export type EnvironmentVariable = ValueEnvironmentVariable | SecretEnvironmentVariable;
+
+const SECRET_NAME_MAX_LENGTH = 253;
+const SECRET_DATA_KEY_MAX_LENGTH = 253;
+
+/** Secret metadata.name — DNS subdomain (RFC 1123), not a DNS label. */
+const SECRET_NAME_REGEX = /^[a-z0-9]([-a-z0-9.]*[a-z0-9])?$/;
+
+/** Secret data key — valid path segment name. */
+const SECRET_DATA_KEY_REGEX = /^[-]?([_a-zA-Z0-9]([-._a-zA-Z0-9]*[_a-zA-Z0-9])?)$/;
```
mturley: `SECRET_DATA_KEY_REGEX` rejects valid Kubernetes Secret data keys that begin with `.` or `_`. For example, `.dockerconfigjson` is the standard key in a `kubernetes.io/dockerconfigjson` Secret, but it fails the current regex even though it is valid for `secretKeyRef.key`.

Please align this with Kubernetes’ data-key validation (for example, `^[-._a-zA-Z0-9]+$` with the existing 253-character length guard) and add coverage for `.dockerconfigjson` and a key beginning with `_`.

mturley, line 32: This regex rejects keys that begin with `.`, like `.dockerconfigjson` which is standard for OCI registry secrets. We should align `isValidSecretDataKey` with k8s’ internal `IsConfigMapKey` ([source here within k8s itself](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/util/validation/validation.go#L405), thanks AI) which is also already present in notebooks [here](https://github.com/opendatahub-io/odh-dashboard/blob/main/packages/notebooks/upstream/workspaces/frontend/src/app/pages/Workspaces/Form/properties/secrets/SecretsCreateModal.tsx#L42-L43).

```ts
/^[-._a-zA-Z0-9]+$/
```

We could factor it out somewhere shared, but that would be tricky because this is upstream kubeflow notebooks code, so it’s probably fine to duplicate.

The real k8s validator also rejects whole values that could be path strings, like `.` and `..`, but the notebooks code doesn’t do that. IMO it’s fine for us to match notebooks and leave that extra validation server-side.

mturley (changes requested):

packages/model-serving/src/components/deploymentWizard/fields/EnvironmentVariablesField.tsx
```diff
@@ -87,6 +147,15 @@ export const useEnvironmentVariablesField = (
...
 };
 
+type EnvironmentVariableUpdates = {
+  type?: EnvironmentVariableType;
+  name?: string;
+  value?: string;
+  secretName?: string;
+  secretKey?: string;
+  optional?: boolean;
+};
```
emilys314: I wonder if there is a way to reuse the existing `EnvironmentVariable` type union instead? This would help reduce duplication of code

It's better type safety to not have all the properties of exclusive types available all at once

packages/model-serving/src/components/deploymentWizard/fields/EnvironmentVariablesField.tsx
```diff
@@ -109,33 +178,80 @@ export const EnvironmentVariablesField: React.FC<EnvironmentVariablesFieldProps>
...
+    if (!data.enabled) {
+      return;
+    }
+
+    const currentVar = normalizeEnvironmentVariable(data.variables[index]);
+    const nextType = updates.type ?? currentVar.type;
+
+    const updatedVar: EnvironmentVariable =
+      nextType === EnvironmentVariableType.Secret
+        ? {
```
emilys314: Related to https://github.com/opendatahub-io/odh-dashboard/pull/9726/changes#r3991689784

If we had a `isSecretEnvVar(nextType)`, it would make the subsequent code less wordy

packages/model-serving/src/components/deploymentWizard/fields/EnvironmentVariablesField.tsx
```diff
@@ -198,36 +314,113 @@ export const EnvironmentVariablesField: React.FC<EnvironmentVariablesFieldProps>
...
+                    : SECRET_NAME_VALIDATION_ERROR
+                  : '';
+              const secretKeyError =
+                normalizedEnvVar.type === EnvironmentVariableType.Secret
+                  ? normalizedEnvVar.secretKey.trim() === ''
+                    ? 'Secret key is required'
+                    : isValidSecretDataKey(normalizedEnvVar.secretKey)
+                    ? ''
+                    : SECRET_DATA_KEY_VALIDATION_ERROR
+                  : '';
```
emilys314: This would be less wordy if we could reuse the existing zod validation schemas we have for this type. Right now we have 2 branches of validation, 1 is what's used for enabling / disabling the next button (zod), the 2nd is this handmade error checking.

packages/model-serving/src/shared/environmentVariablesUtils.ts
```diff
@@ -0,0 +1,171 @@
...
+  valueFrom: Record<string, unknown>,
+): { name: string; key: string; optional?: boolean } | undefined => {
+  const { secretKeyRef } = valueFrom;
+  if (
+    secretKeyRef &&
+    typeof secretKeyRef === 'object' &&
+    'name' in secretKeyRef &&
+    'key' in secretKeyRef &&
+    typeof secretKeyRef.name === 'string' &&
+    typeof secretKeyRef.key === 'string'
```
emilys314: I always want to challenge having a bunch of type coercion.

Can't we change `unknown` into `K8sEnvironmentVariable`?

packages/model-serving/src/shared/environmentVariablesUtils.ts
```diff
@@ -0,0 +1,171 @@
...
+
+export const createDefaultEnvironmentVariable = (): ValueEnvironmentVariable => ({
+  type: EnvironmentVariableType.Value,
+  name: '',
+  value: '',
+});
+
+export const normalizeEnvironmentVariable = (
+  envVar: Partial<EnvironmentVariable> & { name: string },
+): EnvironmentVariable => {
```
emilys314, line 116: This looks really nice to have a consistent way of treating env vars 😀

packages/model-serving/src/shared/environmentVariablesUtils.ts
```diff
@@ -0,0 +1,171 @@
...
+      ...(envVar.optional ? { optional: true } : {}),
+    };
+  }
+
+  return {
+    type: EnvironmentVariableType.Value,
+    name: envVar.name,
+    value: 'value' in envVar ? envVar.value ?? '' : '',
+  };
+};
```
emilys314, line 132: If we use a typescript type guard we can cut down on the wordiness 

```suggestion
export const normalizeEnvironmentVariable = (envVar: EnvironmentVariable): EnvironmentVariable => {
  if (isSecretEnvVar(envVar)) {
    return {
      type: EnvironmentVariableType.Secret,
      name: envVar.name,
      secretName: envVar.secretName,
      secretKey: envVar.secretKey,
      ...(envVar.optional ? { optional: true } : {}),
    };
  }

  return {
    type: EnvironmentVariableType.Value,
    name: envVar.name,
    value: envVar.value,
  };
};
```

emilys314 (commented):
Haven' tested it yet

packages/model-serving/src/shared/environmentVariablesUtils.ts
```diff
@@ -0,0 +1,234 @@
...
+  | {
+      name: string;
+      valueFrom: K8sEnvironmentVariableValueFrom;
+    };
+
+/** Env var as read from CRs (looser than write shape). */
+export type K8sEnvironmentVariableInput = {
+  name: string;
+  value?: string | number;
+  valueFrom?: K8sEnvironmentVariableValueFrom | Record<string, unknown>;
```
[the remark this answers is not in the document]
YuliaKrimerman: Good catch on the round-trip risk for unsupported valueFrom shapes. This behavior predates the secretKeyRef change — the wizard has always only supported plain value env vars (and now secretKeyRef). Unsupported refs were already mapped to { value: '' } on extract and overwritten on apply.

For RHOAIENG-93848 scope (Value + Secret in the deployment wizard), we're not adding passthrough support for configMapKeyRef / fieldRef / etc. If we want to harden edit flows for deployments with operator-managed env vars, I'd track that as a follow-up (preserve raw entries + merge on apply + read-only UI).

Happy to open a follow-up ticket if the team wants it.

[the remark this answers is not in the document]
mturley: > Good catch on the round-trip risk for unsupported valueFrom shapes. This behavior predates the secretKeyRef change — the wizard has always only supported plain value env vars (and now secretKeyRef). Unsupported refs were already mapped to { value: '' } on extract and overwritten on apply.
> 
> For RHOAIENG-93848 scope (Value + Secret in the deployment wizard), we're not adding passthrough support for configMapKeyRef / fieldRef / etc. If we want to harden edit flows for deployments with operator-managed env vars, I'd track that as a follow-up (preserve raw entries + merge on apply + read-only UI).
> 
> Happy to open a follow-up ticket if the team wants it.

@YuliaKrimerman it seems like a followup ticket is a good idea here, at least just to merge on apply so we're never clobbering things.

mturley: The `| Record<string, unknown>` here that propagates elsewhere feels a little loose. Can we have the `K8sEnvironmentVariableValueFrom` type match the actual k8s type (include `configMapKeyRef`, `fieldRef`, and `resourceFieldRef`) even though we won’t support all those, just to match the real data structure? We can include a comment explaining those aren’t supported. Then you could remove `| Record<string, unknown>` everywhere you’re passing it, and instead narrow things based on whether `valueFrom.secretKeyRef` is defined.

packages/model-serving/src/shared/environmentVariablesUtils.ts
```diff
@@ -0,0 +1,234 @@
...
+  const { secretKeyRef } = valueFrom;
+  return (
+    secretKeyRef !== null &&
+    typeof secretKeyRef === 'object' &&
+    'name' in secretKeyRef &&
+    'key' in secretKeyRef &&
+    typeof secretKeyRef.name === 'string' &&
+    typeof secretKeyRef.key === 'string'
+  );
+};
```
mturley: If you drop `Record<string, unknown>` you don’t need this guard. Instead, you can just have code do `if (!valueFrom.secretKeyRef)`. If that property is defined we can trust that its contents match the TS type, otherwise they’d have been rejected by k8s.

packages/model-serving/src/shared/environmentVariablesUtils.ts
```diff
@@ -0,0 +1,234 @@
...
+
+/** Env var as written to serving CRs (value or secretKeyRef, never both). */
+export type K8sEnvironmentVariable =
+  | { name: string; value: string }
+  | {
+      name: string;
+      valueFrom: K8sEnvironmentVariableValueFrom;
+    };
+
+/** Env var as read from CRs (looser than write shape). */
```
mturley: Do we need it to be looser than the write shape? The InferenceService, LLMInferenceService and NIMService CRDs specify `valueFrom` strictly, so we can trust that if `valueFrom.secretKeyRef` is defined, it will have the required `key` and `name`. Let’s have our TypeScript types reflect the required fields, and then we can trust that the operator has validated them server-side and avoid having so much type coercion.

- https://github.com/kserve/kserve/blob/master/config/crd/full/serving.kserve.io_inferenceservices.yaml#L822
- https://github.com/kserve/kserve/blob/master/config/crd/full/llmisvc/serving.kserve.io_llminferenceservices.yaml
- https://github.com/NVIDIA/k8s-nim-operator/blob/main/config/crd/bases/apps.nvidia.com_nimservices.yaml#L1330

mturley (changes requested):
A few comments inline. Also, preflight agent comments (collapsed in its main review body) still not addressed:

> **`packages/llmd-serving/src/deployments/model.ts:159`** · _Jira Eval review_
> 
> The KServe and NIM apply/extract paths both have new unit tests for the secretKeyRef round-trip. The LLM-d equivalent paths (`applyModelEnvVarsAndArgs` in `model.ts` and `extractEnvironmentVariables` in `hardware.ts`) were updated to use `mapEnvironmentVariablesToK8sEnv` / `mapK8sEnvToEnvironmentVariable`, but no tests verify the secretKeyRef behavior for this package. Consider adding a test case to match the coverage in kserve and nim-serving.
> 🧹 Nitpick comments (1)
> packages/model-serving/src/shared/environmentVariablesUtils.ts
> 
> Multiple files use `ReturnType<typeof mapK8sEnvToEnvironmentVariable>[]` in return-type annotations. Since `mapK8sEnvToEnvironmentVariable` already returns `EnvironmentVariable`, these can be simplified to `EnvironmentVariable[]` — making the return types self-documenting without chasing the utility signature.
> 
> Affected: `packages/kserve/src/hardware.ts:52`, `packages/llmd-serving/src/deployments/hardware.ts:52`, `packages/llmd-serving/src/deployments/model.ts:159`, `packages/nim-serving/src/pages/deploymentWizard/extractNIMFormData.ts:47`.

At least the test coverage is probably worth doing here and the other would be good for readability.

YuliaKrimerman:
@mturley Addressed the comments and opened the ticket https://redhat.atlassian.net/browse/RHOAIENG-94808 from the thread above

mturley:
@YuliaKrimerman there’s a quay outage, we’ll need that to be resolved before we retest here https://status.redhat.com/incidents/0g8rsg3tbjpz

packages/model-serving/src/shared/environmentVariablesUtils.ts
emilys314: This can still be updated

packages/model-serving/src/components/deploymentWizard/fields/EnvironmentVariablesField.tsx
```diff
@@ -198,36 +234,98 @@ export const EnvironmentVariablesField: React.FC<EnvironmentVariablesFieldProps>
...
+                          <FormHelperText>
+                            <HelperText>
+                              <HelperTextItem variant="error" icon={<ExclamationCircleIcon />}>
+                                {secretNameError}
+                              </HelperTextItem>
+                            </HelperText>
+                          </FormHelperText>
+                        )}
+                      </SplitItem>
+                      <SplitItem isFilled>
```
emilys314, line 307: Is it possible to reduce the amount of layout-shifting? It looks a little unpolished with the inconsistency 

https://github.com/user-attachments/assets/16c6fed5-c918-4120-8de7-060a55e940af

emilys314 (commented):
I tested the Pr with both an InferenceService and an LLMInferneceService and the env propogates into the pod!!

Image

Image

packages/model-serving/src/components/deploymentWizard/fields/EnvironmentVariablesField.tsx
YuliaKrimerman: Resolved @emilys314 

https://github.com/user-attachments/assets/7da1373e-23f4-4589-9f60-2a55b1d2b936

packages/model-serving/src/components/deploymentWizard/fields/EnvironmentVariablesField.tsx
```diff
@@ -21,34 +21,174 @@ import {
...
+
+type EnvVarTextInputProps = {
+  'data-testid': string;
+  'aria-label': string;
+  value: string;
+  hasError: boolean;
+  isRequired?: boolean;
+  onChange: (value: string) => void;
+  inputRef?: React.Ref<HTMLInputElement>;
+};
```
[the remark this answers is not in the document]
YuliaKrimerman: fixed in the commit below

packages/kserve/src/hardware.ts
```diff
@@ -51,17 +55,14 @@ export const extractRuntimeArgs = (
 
 export const extractEnvironmentVariables = (
   kserveDeployment: KServeDeployment,
-): { enabled: boolean; variables: { name: string; value: string }[] } => {
+): { enabled: boolean; variables: EnvironmentVariable[] } => {
   const envVars =
     kserveDeployment.model.spec.predictor.model?.env?.filter(
       (envVar) => !isDashboardManagedHfTokenEnvVar(envVar),
```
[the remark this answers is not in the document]
mturley: Feels like this is fine, do you agree @YuliaKrimerman ?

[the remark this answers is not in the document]
YuliaKrimerman: Agree — leave as-is. We can’t see the Secret’s dashboard label from the env var alone, and any HF_TOKEN secretKeyRef shouldn’t appear in Additional env vars (it’d show as an empty value and be unsafe to round-trip). The merged HF already owns auth via the dedicated API key field + SA path.

emilys314 (approved):
Looks good, i tested both LLMInferenceService and an InferenceService and it updated as expected

PR #9740, opened 2026-09-11
fix(RHOAIENG-87447): support Markdown rendering and improve HTML visu…

nananosirova:
…alization in artifact drawer (#9439)

Markdown artifacts were rendered as plain text in an unsized iframe. HTML artifacts (e.g. AutoML leaderboards) were unreadable due to missing iframe dimensions. Both now have an expand button that opens a fullscreen modal.

- Render system.Markdown artifacts natively via MarkdownComponent (react-markdown + remark-gfm + rehype-sanitize) instead of iframe
- Add proper iframe sizing for system.HTML artifacts
- Add expand-to-fullscreen modal for both artifact types
- Apply the same fixes to the compare-runs markdown view

(cherry picked from commit 568e592a0ee63f996b03573654004972d22858ec)

Ticket: https://github.com/opendatahub-io/odh-dashboard/pull/9740

## Request review criteria:

If you have UI changes: 

### DIFF
--- a/frontend/src/concepts/pipelines/content/compareRuns/metricsSection/markdown/MarkdownCompare.tsx
+++ b/frontend/src/concepts/pipelines/content/compareRuns/metricsSection/markdown/MarkdownCompare.tsx
@@ -12,6 +12,8 @@ import { CompareRunsEmptyState } from '#~/concepts/pipelines/content/compareRuns
 import { PipelineRunArtifactSelect } from '#~/concepts/pipelines/content/compareRuns/metricsSection/PipelineRunArtifactSelect';
 import { PipelineRunKF } from '#~/concepts/pipelines/kfTypes';
 import { CompareRunsNoMetrics } from '#~/concepts/pipelines/content/compareRuns/CompareRunsNoMetrics';
+import MarkdownComponent from '#~/components/markdown/MarkdownComponent';
+import '#~/concepts/pipelines/content/pipelinesDetails/pipelineRun/artifacts/ArtifactVisualization.scss';
 
 type MarkdownCompareProps = {
   configMap: Record<string, MarkdownAndTitle[]>;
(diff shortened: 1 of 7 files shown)


frontend/src/concepts/pipelines/content/compareRuns/metricsSection/markdown/MarkdownCompare.tsx
```diff
@@ -52,7 +56,17 @@ const MarkdownCompare: React.FC<MarkdownCompareProps> = ({
   const renderMarkdownWithSize = (config: MarkdownAndTitle) => (
     <Stack hasGutter>
       <StackItem>
-        <iframe data-testid="markdown-compare" src={config.config} title="markdown view" />
+        {config.markdownContent ? (
+          <MarkdownComponent data={config.markdownContent} dataTestId="markdown-compare" />
+        ) : (
+          <iframe
+            className="odh-artifact-visualization__iframe pf-v6-u-w-100"
+            sandbox="allow-scripts"
```
danreed-rh, line 64: I see that we are adding allow-scripts into the iframe where we didn't have it before, why are we adding it here now? Is resizing or fullscreen being triggered from within the iframe?

danreed-rh (changes requested):
## Requested changes

1. **Hardware-profile deletion:** Remove `opendatahub.io/hardware-profile-resource-version` along with the name and namespace annotations. Leaving it makes the binding hook classify the resource as permanently `DELETED`. Add a regression test.

2. **Compare-runs race:** In-flight artifact requests can update the cleared maps after the selected artifacts change. Add cancellation or a generation guard, and abort download fetches where possible.

3. **Artifact drawer race:** Recheck `signal.aborted` after awaiting the Markdown fallback render URL and before setting `renderUrl`, so an old request cannot overwrite a newer artifact.

4. **HTML sandbox compatibility:** `sandbox="allow-scripts"` blocks popups, downloads, and same-origin access. Confirm the intended HTML-report contract, allow only necessary capabilities, and test a representative HTML artifact.

5. **Markdown link semantics:** `MarkdownComponent` renders Markdown links as buttons opening a new window rather than anchors. Preserve link semantics and normal browser behavior for artifact Markdown.

PR #9745, opened 2026-09-11
adjust testCreateConnectionTypes.cy.ts name flaky test

Taj010:
For https://redhat.atlassian.net/browse/RHOAIENG-93964 

## Description

Fixes flaky failures in `testCreateConnectionTypes.cy.ts` when the duplicate/edit/delete test runs alongside other S3 connection types or after a failed local run.

The spec was asserting that `custom-type-select` disappeared after delete, but that dropdown only means “multiple S3-compatible types exist,” not “this copy is gone.” It could also duplicate the wrong row when a leftover `Copy of S3…` name substring-matched the OOTB row.

Changes: 
- Use a unique duplicate name per run (`Copy of … - {uuid}`) and clean up stale configmaps in `before`/`after`
- Match connection type table rows by exact display name (not substring)
- After delete, assert the duplicate option is absent when the S3 dropdown exists; pass if the dropdown is gone (only OOTB S3 left)
- Default `deleteConnectionTypeByName` namespace to `redhat-ods-applications` when `APPLICATIONS_NAMESPACE` is unset

## How Has This Been Tested?

- validated test using the  `npm run run:e2e -- --spec cypress/tests/e2e/settings/connectionTypes/testCreateConnectionTypes.cy.ts`
- Also ran Jenkins test -
image

## Test Impact

Adjusted `testCreateConnectionTypes.cy.ts` file
## Request review criteria:

If you have UI changes: 

## Tests

- Improved connection type test coverage for names containing special characters and exact name matching.
- Added validation that connection type creation and duplication complete without footer errors.
- Strengthened setup and cleanup handling for connection type resources.
- Updated deletion checks to account for cases where the custom-type dropdown is not displayed.

### DIFF
--- a/packages/cypress/cypress/tests/e2e/settings/connectionTypes/testCreateConnectionTypes.cy.ts
+++ b/packages/cypress/cypress/tests/e2e/settings/connectionTypes/testCreateConnectionTypes.cy.ts
@@ -17,6 +17,9 @@ import { deleteOpenShiftProject } from '../../../../utils/oc_commands/project';
 import { deleteConnectionTypeByName } from '../../../../utils/oc_commands/connectionTypes';
 import { modelServingWizard, modelServingGlobal } from '../../../../pages/modelServing';
 
+const toConnectionTypeConfigMapName = (displayName: string): string =>
+  `ct-${displayName.toLowerCase().replace(/[\s-]+/g, '-')}`;
+
 describe('Verify Connection Type Creation', () => {
   let testData: OOTBConnectionTypesData;
   let connectionTypeName: string;
@@ -43,7 +46,7 @@ describe('Verify Connection Type Creation', () => {
       projectName = `${testData.projectResourceName}-${uuid}`;
       connectionTypeName = `${testData.connectionTypeName}-${uuid}`;
       existingConnectionTypeName = `${testData.s3}`;
-      duplicateConnectionTypeName = `Copy of ${existingConnectionTypeName}`;
+      duplicateConnectionTypeName = `Copy of ${existingConnectionTypeName} - ${uuid}`;
       connectionTypeDescription = `${testData.connectionTypeDescription}`;
       connectionTypeCategory = testData.connectionTypeCategory;
       connectionTypeModelServingCompatibleType = testData.connectionTypeModelServingCompatibleType;
@@ -58,16 +61,19 @@ describe('Verify Connection Type Creation', () => {
         throw new Error('Project name is undefined or empty in the loaded fixture');
       }
       cy.log(`Loaded project name: ${projectName}`);
+      deleteConnectionTypeByName(toConnectionTypeConfigMapName(connectionTypeName));
+      deleteConnectionTypeByName(toConnectionTypeConfigMapName(duplicateConnectionTypeName));
+      deleteConnectionTypeByName(
+        toConnectionTypeConfigMapName(`Copy of ${existingConnectionTypeName}`),
+      );
       createCleanProject(projectName);
     }),
   );
 
   after(() => {
     // Delete createdconfigmaps in this e2e test
-    deleteConnectionTypeByName(`ct-${connectionTypeName.toLowerCase().replace(/[\s-]+/g, '-')}`);
-    deleteConnectionTypeByName(
-      `ct-${duplicateConnectionTypeName.toLowerCase().replace(/[\s-]+/g, '-')}`,
-    );
+    deleteConnectionTypeByName(toConnectionTypeConfigMapName(connectionTypeName));
+    deleteConnectionTypeByName(toConnectionTypeConfigMapName(duplicateConnectionTypeName));
     // Delete provisioned Project - wait for completion due to RHOAIENG-19969 to support test retries, 5 minute timeout
     // TODO: Review this timeout once RHOAIENG-19969 is resolved
     deleteOpenShiftProject(projectName, { wait: true, ignoreNotFound: true, timeout: 300000 });
@@ -136,6 +142,7 @@ describe('Verify Connection Type Creation', () => {
 
       cy.step('Submit the form to create the connection type');
       createConnectionTypePage.findSubmitButton().click();
+      cy.findByTestId('connection-type-footer-error').should('not.exist');
 
       cy.step('Verify we are redirected to Connection Types list page');
       connectionTypesPage.shouldHaveConnectionTypes();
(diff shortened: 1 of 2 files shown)


packages/cypress/cypress/tests/e2e/settings/connectionTypes/testCreateConnectionTypes.cy.ts
```diff
@@ -136,6 +142,7 @@ describe('Verify Connection Type Creation', () => {
 
       cy.step('Submit the form to create the connection type');
       createConnectionTypePage.findSubmitButton().click();
+      cy.findByTestId('connection-type-footer-error').should('not.exist');
```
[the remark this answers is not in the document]
Taj010: If create or duplicate actually fails, the test would still fail on the next steps - we check the connection type shows up in the list and we find the duplicate row and edit it. The should('not.exist') on the footer error is just an extra check, not the main assertion. Adding intercept + wait feels like a separate improvement and is out of scope for this flake fix

packages/cypress/cypress/tests/e2e/settings/connectionTypes/testCreateConnectionTypes.cy.ts
```diff
@@ -265,7 +271,14 @@ describe('Verify Connection Type Creation', () => {
       modelServingGlobal.selectSingleServingModelButtonIfExists();
       modelServingGlobal.findDeployModelButton().click();
       modelServingWizard.findModelLocationSelectOption(modelLocation).click();
-      modelServingWizard.findCustomModelLocationSelect().should('not.exist');
+      // Dropdown only renders when multiple S3-compatible types exist; after delete it may be gone.
+      cy.get('body').then(($body) => {
+        if ($body.find('[data-testid="custom-type-select"]').length > 0) {
```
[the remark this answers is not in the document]
Taj010: This condition is intentional to fix the flake. The old test always found CustomModelLocationSelect ().should('not.exist'), but that dropdown only appears when there are multiple S3-compatible connection types. After deletion, if only the OOTB S3 type remains, the dropdown won't render at all — and that's expected, not a failure.

packages/cypress/cypress/tests/e2e/settings/connectionTypes/testCreateConnectionTypes.cy.ts
d0w, line 145: there should be a utility function `findFooterError` we can use here

packages/cypress/cypress/tests/e2e/settings/connectionTypes/testCreateConnectionTypes.cy.ts
```diff
@@ -216,10 +223,9 @@ describe('Verify Connection Type Creation', () => {
       const exisitingRow = connectionTypesPage.getConnectionTypeRow(existingConnectionTypeName);
       exisitingRow.findKebab().click();
       connectionTypesPage.findDuplicateAction().click();
-      createConnectionTypePage
-        .findConnectionTypeName()
-        .should('have.value', duplicateConnectionTypeName);
+      createConnectionTypePage.findConnectionTypeName().clear().type(duplicateConnectionTypeName);
       createConnectionTypePage.findSubmitButton().should('be.enabled').click();
+      cy.findByTestId('connection-type-footer-error').should('not.exist');
```
d0w, line 228: same here

d0w (commented):
tested on my side and it works well. just some nitpicks

d0w:

packages/cypress/cypress/tests/e2e/settings/connectionTypes/testCreateConnectionTypes.cy.ts
Taj010: Done!

d0w:

Taj010:

packages/cypress/cypress/tests/e2e/settings/connectionTypes/testCreateConnectionTypes.cy.ts
```diff
@@ -58,16 +61,19 @@ describe('Verify Connection Type Creation', () => {
         throw new Error('Project name is undefined or empty in the loaded fixture');
       }
       cy.log(`Loaded project name: ${projectName}`);
+      deleteConnectionTypeByName(toConnectionTypeConfigMapName(connectionTypeName));
+      deleteConnectionTypeByName(toConnectionTypeConfigMapName(duplicateConnectionTypeName));
+      deleteConnectionTypeByName(
```
DaoDaoNoCode: **Blocking:** This deletes the unsuffixed `ct-copy-of-s3-compatible-object-storage-v1` ConfigMap even though the updated test only creates the UUID-suffixed duplicate. On a shared E2E cluster that name is not owned by this run and could belong to another user or test. Please remove this cleanup, or verify ownership before deleting it; cleanup of the two UUID-owned names is sufficient for retry safety.

packages/cypress/cypress/tests/e2e/settings/connectionTypes/testCreateConnectionTypes.cy.ts
```diff
@@ -216,10 +223,9 @@ describe('Verify Connection Type Creation', () => {
       const exisitingRow = connectionTypesPage.getConnectionTypeRow(existingConnectionTypeName);
       exisitingRow.findKebab().click();
       connectionTypesPage.findDuplicateAction().click();
-      createConnectionTypePage
-        .findConnectionTypeName()
-        .should('have.value', duplicateConnectionTypeName);
+      createConnectionTypePage.findConnectionTypeName().clear().type(duplicateConnectionTypeName);
```
DaoDaoNoCode: Could we preserve the original duplicate-name behavior check before overriding the value? For example, first assert `Copy of ${existingConnectionTypeName}`, then clear and enter the UUID-suffixed name. As written, the test would no longer catch a regression where the Duplicate action stops prefilling the expected name.

packages/cypress/cypress/tests/e2e/settings/connectionTypes/testCreateConnectionTypes.cy.ts
```diff
@@ -265,7 +271,14 @@ describe('Verify Connection Type Creation', () => {
       modelServingGlobal.selectSingleServingModelButtonIfExists();
       modelServingGlobal.findDeployModelButton().click();
       modelServingWizard.findModelLocationSelectOption(modelLocation).click();
-      modelServingWizard.findCustomModelLocationSelect().should('not.exist');
+      // Dropdown only renders when multiple S3-compatible types exist; after delete it may be gone.
+      cy.get('body').then(($body) => {
```
DaoDaoNoCode: **Non-blocking:** The absent/present branching is consistent with the UI contract: the selector is omitted when only one S3 type remains. However, this raw `cy.get("body")` and `data-testid` lookup should be encapsulated in the model-serving page object per the E2E page-object convention, so selector knowledge stays out of the spec.

PR #9764, opened 2026-09-15
fix(federation): force federated DEPLOYMENT_MODE for workspace module images

lucferbux:
Fixes https://issues.redhat.com/browse/RHOAIENG-83821

## Description

Federated module images built with `Dockerfile.workspace` are expected to compile the frontend with `DEPLOYMENT_MODE=federated` so Module Federation shares the host's `@odh-dashboard/*` and framework packages (`import: false`) instead of bundling duplicate copies of React / PatternFly / ODH SDK.

Today every `Dockerfile.workspace` declares only `ARG DEPLOYMENT_MODE` and never promotes it to `ENV`, and CI/Konflux does not pass `--build-arg`. So `npm run build:prod` runs with `process.env.DEPLOYMENT_MODE` unset, the parent `.env` files are not in the Docker context, and the build silently falls back to each package's `dotenv.js` default (often `standalone` or `kubeflow`). Remotes then load duplicate React/ODH packages and break host/remote singleton sharing with no obvious error.

Separately, the `data-registry` (and the newer `data-connect-hub`) BFF module Deployments omit `--deployment-mode=federated` even though their Go binaries define the flag.

### Changes

**Frontend — force federated compile in every `Dockerfile.workspace`:**
- Promote the (previously dead) build arg to a defaulted arg + env var before the build, so images compile federated by default while remaining overridable via `--build-arg`:
  ```dockerfile
  ARG DEPLOYMENT_MODE=federated
  ENV DEPLOYMENT_MODE=${DEPLOYMENT_MODE}
  ```
- Applied to: `core-bff`, `agent-ops`, `automl`, `autorag`, `data-connect-hub`, `data-registry`, `eval-hub`, `gen-ai`, `maas`, `mlflow`, `model-registry`, `notebooks`.
- Excluded `plugin-template` (sample only) per the ticket.

**BFF — add the deployment-mode flag to the two manifests that were missing it:**
- `manifests/modules/data-registry/deployment.yaml` → `--deployment-mode=federated`
- `manifests/modules/data-connect-hub/deployment.yaml` → `--deployment-mode=federated` (module onboarded after the ticket; same defect)
- Gen AI keeps omitting the flag (its BFF binary does not define it and would crash if passed). Notebooks has no Go BFF, so no flag is needed.

**Tests — lock in the federated Module Federation shared-config contract:**
- Added a `federated remote must-share contract` suite to `packages/app-config/src/webpack/__tests__/webpack-plugins.spec.ts` asserting that, on a remote build, every must-come-from-host module (React, routers, `@patternfly/react-core`/`react-styles`, dynamic-plugin-sdk) gets `import: false` and is not eager; on a host build they are eager and never `import: false`; and only host-provided ODH packages get `import: false` while federated-only packages keep their fallback.

> Note: `data-connect-hub` is a module onboarded after the ticket was filed. It exhibited the same frontend and BFF gaps, so it is included here to keep all federated modules consistent.

## How Has This Been Tested?

- `npx jest src/webpack/__tests__/webpack-plugins.spec.ts` in `packages/app-config` — 17/17 pass (3 new).
- ESLint and `tsc --noEmit` clean on the changed test file / `app-config`.
- Both edited `deployment.yaml` manifests validated as parseable YAML.
- Verified via grep that every module Deployment except `gen-ai` and `notebooks` now passes `--deployment-mode=federated`, and that all 12 target `Dockerfile.workspace` files (not `plugin-template`) now set `ENV DEPLOYMENT_MODE`.

## Test Impact

- Added unit tests for the webpack Module Federation shared-module policy (host vs. remote `import: false` / `eager` behavior) covering the full must-share module set.
- The Dockerfile and manifest changes are build/deploy configuration and are not exercised by unit tests; they were verified by static inspection and a documented `docker build` spot-check is recommended before merge.

## Request review criteria:


christianvogt (approved):

christianvogt:
@lucferbux what's the disconnected readiness check error ?

lucferbux:
@christianvogt No idea at all, but I discovered I was adding this to core-bff, so I reverted that and rebased and now everything seems fine. I think is good to go.

PR #9773, opened 2026-09-15
Rhoaieng 88477/edit external models

sridarna:
closes https://redhat.atlassian.net/browse/RHOAIENG-88477
## Description

**Adds the Edit external model flow for MaaS.**

- External models table Edit kebab action navigates to the edit page.
- Form is prefilled with display name, description, project, and provider references; Save/Cancel return to the external models list.

**Provider Ref Validations:**

- Provider ref modal Add/Save(Edit) button  is disable when target model ID& path is empty, invalid path format Ex: path not starting with '/', unresolved placeholders for path field.

**External Model List Updates:**

- Expanded provider details row shows weight and traffic percentage (e.g. 100 (100%), 60 (60%)). reusing the provider ref table row.

**other:**
Update ModelName to populate metadata.name.
ProviderSource  constants (existing, create-new) extracted in const.tsx.

## How Has This Been Tested?

Testing Steps:
Open AI Hub → Models → Deployments → External models.
Kebab → Edit on a model; confirm form is prefilled.
Edit a provider ref (modal), add another ref (wizard), change weights, Save — confirm redirect and updated data.
Expand a model row — confirm weight shows with percentage, e.g. 6 (60%).

## Test Impact

Add Mock tests for edit flow & weights column in list external models 
## Request review criteria:

If you have UI changes: 

## New Features
- Added editing for external models, including model details and provider references.
- Added Edit actions and navigation from external-model listings.
- Added provider weight and percentage displays.

## Bug Fixes
- Improved validation for incomplete paths, placeholders, and provider configuration.
- Corrected provider weight normalization and handling.

## Tests
- Expanded coverage for editing, validation, provider references, saving, and navigation.


packages/maas/frontend/src/app/pages/external-models/CreateExternalModel/CreateExternalModelForm.tsx
```diff
@@ -45,20 +50,34 @@ import ProviderReferencesTable from './ProviderReferencesTable';
...
+  );
 
-  const { data: nameDescData, onDataChange: onNameDescChange } = useK8sNameDescriptionFieldData({
-    namespace,
-  });
-
-  const [providerRefs, setProviderRefs] = React.useState<ProviderRef[]>([]);
+  const [providerRefs, setProviderRefs] = React.useState<ProviderRef[]>(
+    () => externalModel?.providerRefs ?? [],
+  );
```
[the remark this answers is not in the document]
sridarna: Addressed

packages/maas/frontend/src/app/pages/external-models/CreateExternalModel/CreateExternalModelForm.tsx
```diff
@@ -106,22 +125,37 @@ const CreateExternalModelForm: React.FC<CreateExternalModelFormProps> = ({
...
-    };
-
     try {
       const apiOpts: APIOptions = {};
-      await createExternalModel()(apiOpts, request);
+
+      if (externalModel) {
+        const request: UpdateExternalModelRequest = {
+          displayName: trimmedName,
+          modelName: trimmedName,
```
Griffin-Sullivan: shouldn't these be different? same below

packages/maas/frontend/src/odh/odhExtensions/odhExtensions.ts
```diff
@@ -130,6 +130,16 @@ const ODH_EXTENSIONS: ODHExtensions[] = [
       component: () => import('./ExternalModelsCreateWrapper'),
     },
   },
+  {
+    type: 'app.route',
+    flags: {
+      required: [MODEL_AS_SERVICE_ID, EXTERNAL_MODELS_ID],
```
Griffin-Sullivan: You can remove the requirement for EXTERNAL_MODELS_ID here

Griffin-Sullivan (commented):
Overall changes look good to me and it's working just a couple comments

packages/maas/frontend/src/app/pages/external-models/CreateExternalModel/CreateExternalModelForm.tsx
sridarna: As per this https://github.com/opendatahub-io/ai-gateway-payload-processing/blob/main/config/crd/bases/inference.opendatahub.io_externalmodels.yaml
 `ModelName is the client-facing model name used in inference request bodies. Clients send this value in the "model" field of chat completion requests.Defaults to metadata.name if not set. Use this field when the desired model name contains characters not allowed in Kubernetes resource names(dots, colons, slashes, uppercase).`
So, the display name field  in create external model modal  is used to set ModelName.
We are populating displayName the same value  since displayName is used in list external models for UI.

packages/maas/frontend/src/odh/odhExtensions/odhExtensions.ts
sridarna: The same goes for ExternalModelsCreateWrapper, ExternalProvidersWrapper too. Right?

Griffin-Sullivan: I think @katieperry4 's PR will handle those?

packages/maas/frontend/src/app/pages/external-models/CreateExternalModel/CreateExternalModelForm.tsx
Griffin-Sullivan: so should the model name be the same as the k8s name and the display name is just the display name on the ExternalModel

sridarna: No, the modelName should be the display name. As it mentioned it is client-facing model name.
But we do have displayName field in BFF for create, we are populating it too with the same value as modelName.
Without displayName being populated at create, the UI displays k8s name as external model name in list view.

packages/maas/frontend/src/odh/odhExtensions/odhExtensions.ts
sridarna: Removed only for EditWrapper

packages/maas/frontend/src/app/pages/external-models/CreateExternalModel/CreateExternalModelForm.tsx
Griffin-Sullivan: i'm confused because if your display name has spaces and special chars like `Griffin's model` then that's what people will have to put in the curl request when they are using the model. I think that's not desired here and it should be like a k8s name where it's only letters, dashes, and numbers

Griffin-Sullivan: I think that's why the default behavior is if we don't set the model name it will just use whatever the `metadata.name` or k8sname is

katieperry4 (commented):
Read through and did some testing:
- edited name, desc, providers, made new providers from edit, removed them, edited the providers references themselves and all the fields from there
everything worked without issue!

/lgtm
once the name/modelName discussion gets resolved, let me know if I need to add the lgtm again

packages/maas/frontend/src/app/pages/external-models/CreateExternalModel/CreateExternalModelForm.tsx
sridarna: Yeah. modelName purpose looks like used for curl request. I will update the modelName to set the k8sname.

packages/maas/frontend/src/app/pages/external-models/CreateExternalModel/CreateExternalModelForm.tsx
```diff
@@ -141,7 +141,7 @@ const CreateExternalModelForm: React.FC<CreateExternalModelFormProps> = ({
           name: nameDescData.k8sName.value,
           namespace,
           displayName: trimmedName,
-          modelName: trimmedName,
+          modelName: nameDescData.k8sName.value,
```
Griffin-Sullivan: on update you do `externalModel.name` and on create you do `nameDescData.k8sName.value`. I know on edit we can't update the k8s name is that why you have to use externalModel.name? Just want to make sure we have the same value regardless. Technically on update you won't be able to change the model name since you can't change the k8s name

sridarna: make sense. looking into.

sridarna: removed modelName from update.
So, if a user updates modelName in  external model yaml, during edit from UI we are not updating modelName. looks like it make sense.
I think the bff still have modelName in updateRequest. currently leaving it if it's okay

Griffin-Sullivan (approved):

katieperry4 (commented):
just tested the description removal!
All looks good!
/lgtm

sridarna:
FYI: cypress e2e tests for maas are passing in CI https://github.com/opendatahub-io/odh-dashboard/actions/runs/35139874754

christianvogt:
@sridarna can you validate that  Cypress E2E Tests: `Verify User can deploy a model by selecting a MaaS subscription and API key` which is failing is not a result of this PR.

sridarna:
@christianvogt The e2e is failing with insufficient resource. This pr changes doesn't affect the E2E.

PR #9795, opened 2026-09-16
test(RHOAIENG-83659): add platform contract E2E coverage

liday-rh:
https://issues.redhat.com/browse/RHOAIENG-83659

## Description

Adds real-cluster Dashboard platform-contract conformance coverage using the shared E2E framework. The suite verifies the utility contract, CEL and validating-webhook singleton enforcement independently, webhook readiness, and semantic dashboard/platform release metadata.

## How Has This Been Tested?

- `make test`
- `make lint`
- `make test-integration`
- `make build-e2e`
- `make test-e2e E2E_TEST_ARGS="-run ^TestE2E_PlatformContractConformance$"`

Manual cluster validation used OpenShift 4.21.28 with RHOAI 3.5.0-ea.2. The dashboard-operator was built from this branch with version `0.0.0-rhoaieng.83659`, deployed with webhook TLS/CA injection enabled, and the selective E2E test passed in 9.3 seconds. The Dashboard was restored to `Ready=True` after the test.

## Test Impact

Adds pure unit coverage for webhook/release assertions, strengthens controller and webhook unit tests, and adds a selective real-cluster conformance scenario.

## Request review criteria:

If you have UI changes:

## Summary by CodeRabbit

* **Tests**
  * Added end-to-end coverage for platform contract conformance, including validating webhook configuration, CEL validation rules, admission responses, and required release metadata.
  * Expanded reconciliation and singleton webhook tests to verify release identity, response codes, and denial messages.
  * Added validation coverage for webhook selection, CEL rejection, duplicate instances, and release metadata.

* **Documentation**
  * Added guidance for required webhook, TLS, permissions, and release configuration.
  * Documented how to run the platform contract conformance scenario independently.


dashboard-operator/test/e2e/contract_test.go
```diff
@@ -0,0 +1,225 @@
...
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const contractTimeout = 10 * time.Minute
+
+var contractModuleNames = []string{
```
lucferbux: nit: this hardcodes the full module set, which duplicates the canonical `moduleRegistry` in `internal/controller/modules.go`. We already export `ctrlpkg.ModuleNames()` (it's what `disableAllModulesExcept` uses in the controller integration tests) — if we derive the list from that, a newly-registered module gets disabled here automatically instead of silently slipping through and getting provisioned during the contract run. Right now the two lists happen to match, but they'll drift the moment someone adds a module.

dashboard-operator/test/e2e/contract_assertions.go
```diff
@@ -0,0 +1,212 @@
...
+			return true
+		}
+	}
+
+	return false
+}
+
+func matchesRuleValue(values []string, expected string) bool {
+	for _, value := range values {
+		if value == expected || value == "*" {
```
lucferbux: nit: accepting `*` here means a cluster-wide policy webhook (Gatekeeper/Kyverno with `apiGroups: ["*"]`/`resources: ["*"]`) also matches `webhookHandlesDashboardCreates`, so `selectDashboardWebhook` can bail with `expected exactly one` or `failurePolicy Fail` for reasons that have nothing to do with our webhook. Might be worth matching the dashboard webhook explicitly (exact group/resource, or by name) so the contract test doesn't depend on whatever admission controllers happen to be installed on the cluster.

dashboard-operator/test/e2e/contract_test.go
```diff
@@ -0,0 +1,225 @@
...
+	})
+	if err != nil {
+		if lastValidationErr != nil {
+			return fmt.Errorf("wait for required Dashboard releases: %w: %w", err, lastValidationErr)
+		}
+
+		return fmt.Errorf("wait for required Dashboard releases: %w", err)
+	}
+	if lastValidationErr != nil {
+		return errors.New("required Dashboard releases remained invalid after polling")
```
lucferbux: nit: this branch is unreachable — `PollUntilContextTimeout` only returns a nil error when the condition returned `true`, i.e. `lastValidationErr == nil`, so we can never land here with a non-nil `lastValidationErr`. Can drop it.

dashboard-operator/test/e2e/contract_test.go
```diff
@@ -0,0 +1,225 @@
...
+	crd := &apiextensionsv1.CustomResourceDefinition{}
+	require.NoError(t, k8sClient.Get(t.Context(), client.ObjectKey{Name: dashboardCRDName}, crd))
+
+	for _, servedVersion := range crd.Spec.Versions {
+		if servedVersion.Name != dashboardv1alpha1.GroupVersion.Version || servedVersion.Schema == nil ||
+			servedVersion.Schema.OpenAPIV3Schema == nil {
+			continue
+		}
+		for _, rule := range servedVersion.Schema.OpenAPIV3Schema.XValidations {
+			if rule.Rule == dashboardCELRule && rule.Message == dashboardCELValidationMessage {
```
lucferbux: nit: exact-string match on the CEL expression is a bit brittle — a whitespace/formatting change to the rule in the CRD would fail this even though it's semantically identical. Also we're iterating all versions but not checking `servedVersion.Served`. Minor, only bites if the rule text ever gets reformatted.

lucferbux (commented):
Looks solid overall, nice conformance coverage and I like that the assertion helpers are split out untagged so they get real unit tests. Left a few nits inline, nothing blocking. Also +1 to CodeRabbit's note that `validateRequiredReleases` only enforces single-entry for `dashboard`/`platform` and would let a duplicate `future-component` through, agree it's minor but easy to tighten while we're here.

lucferbux:
Reviewed for correctness at high effort. Verified locally (worktree of the PR head): builds in both non-e2e and `-tags e2e` modes, `go vet -tags e2e ./test/e2e/...` clean, and the new/changed unit tests pass (`TestSelectDashboardWebhook`, `TestValidateRequiredReleases`, `TestValidateWebhookRejection`, `TestReconcile_PlatformVersionHandshake`, `TestSingletonHandler_WithExistingInstance`).

Cross-checked the hardcoded contract constants against the sources: group `components.platform.opendatahub.io`, `common.ReleasePlatform == "platform"`, repo URL and `Version` default in the reconciler, and the singleton denial message/403 in `odh-platform-utilities/pkg/webhook`. The split of pure assertion logic (untagged `contract_assertions.go` + unit test) from the cluster-only `//go:build e2e` scenario is a nice touch and gives real unit coverage without a cluster.

One minor, non-blocking nit: `waitForDashboardWebhook` treats a hard misconfiguration (wrong `failurePolicy`, missing CA bundle) the same as "not ready yet" and polls for the full `contractTimeout` before surfacing `lastErr`. Fine to keep as-is given webhook config can appear asynchronously after operator start.

Looks good.

/lgtm
/approve

christianvogt (approved):

liday-rh:
@mergifyio queue

christianvogt:
@liday-rh please rebase

